PR #9010, opened 2019-07-30
introduce --remote_default_exec_properties

buchgr:
`--remote_default_exec_properties` serves the same purpose as
`--remote_default_platform properties` but is easier to use in that it
doesn't require the user to specify a text protobuf on the command line.

Use it like:
```
bazel build :foo --remote_cache=host:port \
  --remote_default_exec_properties=OSFamily=linux \
  --remote_default_exec_properties=ISA=x86-64 \
  --remote_default_exec_properties=ContainerImage=docker://...
```
The naming is also inline with the newly introduced `exec_properties`
attribute.

RELNOTES: --remote_default_platform_properties has been deprecated in favor
of --remote_default_exec_properties.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -328,7 +328,7 @@ public final class RemoteOptions extends OptionsBase {
 
   @Option(
       name = "remote_default_exec_properties",
-      defaultValue = "null",
+      defaultValue = "[]",
       documentationCategory = OptionDocumentationCategory.REMOTE,
       effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
       converter = AssignmentConverter.class,
@@ -358,7 +358,7 @@ public final class RemoteOptions extends OptionsBase {
   }
 
   public SortedMap<String, String> getRemoteDefaultExecProperties() throws UserExecException {
-    boolean hasExecProperties = remoteDefaultExecProperties != null;
+    boolean hasExecProperties = !remoteDefaultExecProperties.isEmpty();
     boolean hasPlatformProperties = !remoteDefaultPlatformProperties.isEmpty();
 
     if (hasExecProperties && hasPlatformProperties) {
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -14,18 +14,27 @@
 
 package com.google.devtools.build.lib.remote.options;
 
+import build.bazel.remote.execution.v2.Platform;
+import build.bazel.remote.execution.v2.Platform.Property;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSortedMap;
+import com.google.devtools.build.lib.actions.UserExecException;
 import com.google.devtools.build.lib.util.OptionsUtils;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.common.options.Converters;
+import com.google.devtools.common.options.Converters.AssignmentConverter;
 import com.google.devtools.common.options.EnumConverter;
 import com.google.devtools.common.options.Option;
 import com.google.devtools.common.options.OptionDocumentationCategory;
 import com.google.devtools.common.options.OptionEffectTag;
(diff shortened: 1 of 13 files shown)


buchgr:
I decided to not make it an incompatible change for now and just have these two flags that do the same thing live side by side for a while.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -309,13 +318,25 @@ public RemoteOutputsStrategyConverter() {
...
+      deprecationWarning = "--remote_default_platform_properties has been deprecated in favor of --remote_default_exec_properties.",
       help =
           "Set the default platform properties to be set for the remote execution API, "
               + "if the execution platform does not already set remote_execution_properties. "
               + "This value will also be used if the host platform is selected as the execution "
               + "platform for remote execution.")
   public String remoteDefaultPlatformProperties;
 
+  @Option(
+      name = "remote_default_exec_properties",
```
ishikhman: Do we really need "remote_default" prefix for this flag? wouldn't it be simpler just `exec_properties`?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -309,13 +318,25 @@ public RemoteOutputsStrategyConverter() {
...
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
+      converter = AssignmentConverter.class,
+      allowMultiple = true,
+      help = "Set the default exec properties to be used as the remote execution platform "
+              + "if an execution platform does not already set exec_properties.")
+  public List<Map.Entry<String, String>> remoteDefaultExecProperties;
```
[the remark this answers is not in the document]
buchgr: That's how the options parser works. If you set allowMultiple=true then you need to make it a list. I suppose it makes sense as when having a list the options parses doesn't need to do any merging / overrides.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -309,13 +318,25 @@ public RemoteOutputsStrategyConverter() {
...
       help =
           "Set the default platform properties to be set for the remote execution API, "
               + "if the execution platform does not already set remote_execution_properties. "
               + "This value will also be used if the host platform is selected as the execution "
               + "platform for remote execution.")
   public String remoteDefaultPlatformProperties;
 
+  @Option(
+      name = "remote_default_exec_properties",
+      defaultValue = "null",
```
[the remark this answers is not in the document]
buchgr: good idea. will do.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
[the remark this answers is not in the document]
ishikhman: gotcha, thanks for explaining!

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
buchgr: I think just exec_properties is misleading. It's important to state that this flag is a default value that's overriden by platforms. Theoretically we could drop the "remote_" prefix, but then again we have it for all the other remote options as well and in case exec_properties would be introduced for a different execution strategy some time in the future it would for sure require a different default_exec_properties flag given that the values are really specific to the execution environment.

So I think we should keep the naming, well aware that the flag name is a bit long.

ishikhman: Alright, fair enough.
The only thing - I don't think it's clear from the current description that 
> this flag is a default value that's overriden by platforms

Could you then just adjust the description?

buchgr: > Set the default exec properties to be used as the remote execution platform if an execution platform does not already set exec_properties.

How would you like me to rephrase it?

ishikhman: > Set the default exec properties to be used as the remote execution platform if an execution platform does not already set exec_properties.

Ah, sorry, good enough, missed this :confused:

PR #9013, opened 2019-07-30
Don't crash if the server returns a long response on uploading.

benjaminp:
HttpObjectAggregator raises runtime error TooLongFrameException if the response body exceeds the aggregation limit. We need to turn this exception into an IOException, so it doesn't crash Bazel.


artem-zinnatullin:
Nice catch 👍

buchgr:
nice catch indeed

Issue #7503, opened 2019-02-22
Windows: Bash-less alternative for genrule


laszlocsomor:
### Description of the problem / feature request:

Provide a genrule implementation (or a similar rule) that does not use Bash, but uses Cmd or PowerShell instead.

### Feature requests: what underlying problem are you trying to solve with this feature?

Goal: a genrule-like rule that supports simple shell actions (e.g. copy files, run programs on inputs to produce outputs, etc.) without having to install any extra shell (i.e. should work on a clean Win10 installation).

This is part of the effort to cut Bazel's Bash dependency (#4319).

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

0.22.0

###  Have you found anything relevant by searching the web?

maprule() contains most of the necessary logic for this: https://github.com/bazelbuild/bazel-skylib/blob/master/rules/maprule.bzl

laszlocsomor:
To avoid repeating the same mistake we did with Bash about assuming that certain tools are always installed with Bash (e.g. zip, grep, tar, perl), (https://github.com/bazelbuild/bazel/issues/5265), the rule should:
- only support cmd.exe's built-in commands (at least with the cmd_genrule variant)
- run with an empty PATH (this also fosters reproducibility and cache-ability of artifacts)

I don't know PowerShell enough to draw are constraints about what the rule could assume is available and what isn't. Maybe someone with more expertise could chime in on this.

laszlocsomor:
Progress!

- `native_binary` ([source](https://github.com/bazelbuild/bazel-skylib/blob/master/rules/native_binary.bzl), [documentation](https://github.com/bazelbuild/bazel-skylib/blob/master/docs/native_binary_doc.md#native_binary)): wraps a native binary in a `*_binary` rule, which you can `bazel run` or use in `genrule.tools`
- `native_test` ([source](https://github.com/bazelbuild/bazel-skylib/blob/master/rules/native_binary.bzl), [documentation](https://github.com/bazelbuild/bazel-skylib/blob/master/docs/native_binary_doc.md#native_test)): wraps a native binary in a `*_test` rule, which you can `bazel test`
[...]

PR #9022, opened 2019-07-31
Refactor CommandHelper.java

meteorcloudy:
Extract language specific functions to an interface.

Working towards: https://github.com/bazelbuild/bazel/issues/7503

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
@@ -1,0 +1,52 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.analysis;
+
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.analysis.actions.FileWriteAction;
+import com.google.devtools.build.lib.vfs.PathFragment;
+
+/**
+ * The class for constructing command line for Bash.
+ */
+public class BashCommandConstructor implements CommandConstructor {
+
+  private final PathFragment shellPath;
+  private final String scriptPostFix;
+
+  BashCommandConstructor(PathFragment shellPath, String scriptPostFix) {
+    this.shellPath = shellPath;
+    this.scriptPostFix = scriptPostFix;
+  }
+
+  @Override
+  public ImmutableList<String> buildCommandLineArgvWithArtifact(Artifact scriptFileArtifact) {
+    return ImmutableList.of(shellPath.getPathString(), scriptFileArtifact.getExecPathString());
+  }
+
+  @Override
+  public Artifact buildCommandLineArtifact(RuleContext ruleContext, String command) {
+    String scriptFileName = ruleContext.getTarget().getName() + scriptPostFix;
+    String scriptFileContents = "#!/bin/bash\n" + command;
+    return FileWriteAction.createFile(
+        ruleContext, scriptFileName, scriptFileContents, /*executable=*/true);
+  }
+
+  @Override
+  public ImmutableList<String> buildCommandLineSimpleArgv(String command) {
+    return ImmutableList.of(shellPath.getPathString(), "-c", command);
+  }
+}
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.analysis.actions.FileWriteAction;
+import com.google.devtools.build.lib.vfs.PathFragment;
+
+/**
+ * The class for constructing command line for Bash.
+ */
+public class BashCommandConstructor implements CommandConstructor {
```
laszlocsomor: final

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.actions.Artifact;
+
+/**
+ * The interface to construct command line for different shells (Bash, Batch, Powershell).
+ * Used in {@link com.google.devtools.build.lib.analysis.CommandHelper}
+ */
+public interface CommandConstructor {
+
+  // Given a string of command, return the arguments to run the command.
```
laszlocsomor: Comment as JavaDoc:
```
/** Given a string of command, return the arguments to run the command. */
```

Also below.

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+  ImmutableList<String> buildCommandLineSimpleArgv(String command);
+
+  // Write the command to a script and return the artifact of the script.
+  Artifact buildCommandLineArtifact(RuleContext ruleContext, String command);
```
laszlocsomor: How about: `commandAsScript`?

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.devtools.build.lib.actions.Artifact;
+
+/**
+ * The interface to construct command line for different shells (Bash, Batch, Powershell).
+ * Used in {@link com.google.devtools.build.lib.analysis.CommandHelper}
+ */
+public interface CommandConstructor {
+
+  // Given a string of command, return the arguments to run the command.
+  ImmutableList<String> buildCommandLineSimpleArgv(String command);
```
laszlocsomor: How about: `asExecArgv`?

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
laszlocsomor: Also please add an example, showing how the function transforms the input. E.g. `buildCommandLineSimpleArgv("foo bar") -> ["/bin/bash", "-c", "foo bar"]`

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
```diff
@@ -0,0 +1,34 @@
...
+
+  // Write the command to a script and return the artifact of the script.
+  Artifact buildCommandLineArtifact(RuleContext ruleContext, String command);
+
+  // Given an artifact of a script, return the arguments to run this command.
+  ImmutableList<String> buildCommandLineArgvWithArtifact(Artifact scriptFileArtifact);
```
laszlocsomor: Since this function does the same as `buildCommandLineSimpleArgv` but with a script instead of a command, can we name them the same?

src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private final PathFragment shellPath;
+  private final String scriptPostFix;
```
laszlocsomor: How about `scriptNameTail` or `scriptNameSuffix`?
First I thought this stored script code (to be appended to the command script) instead of the end of the file's name.

src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/analysis/BashCommandConstructor.java
meteorcloudy: Oh, I was using the original name, but `scriptNameSuffix` does sound better.

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
meteorcloudy: Done

meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/analysis/CommandConstructor.java
meteorcloudy: Good idea!

Issue #6531, opened 2018-10-26
Expansion of $(location) on path with spaces double-quotes argument.


ob:
Tested with Bazel 0.18.0 and master as of d4e3ad8951. 

Small test to reproduce at: https://github.com/ob/bazel-tests/tree/master/spaces

If I run:
```
$ bazel build --spawn_strategy=standalone -s --sandbox_debug //spaces:no_spaces
```
I get this output (paths removed and output elided for clarity):

```
DEBUG: arule.bzl:3:5: Expanding Location Placeholders...
DEBUG: arule.bzl:4:5: Values: ["-c", "$(location adir/foo.c)"]
DEBUG: arule.bzl:5:5: Expansion: ["-c", "spaces/adir/foo.c"]

SUBCOMMAND: # //spaces:no_spaces [action 'SkylarkAction spaces/adir/foo.o']
(cd /private/var/tmp/_bazel_obonilla/124df79d9592378b17b94075fb66c278/execroot/__main__ && \
  exec env - \
  cc -c spaces/adir/foo.c -o bazel-out/darwin-fastbuild/bin/spaces/adir/foo.o)
```

And the build succeeds. However, if I run:

```
$ bazel build --spawn_strategy=standalone -s --sandbox_debug //spaces:spaces
```
DEBUG: arule.bzl:3:5: Expanding Location Placeholders...
DEBUG: arule.bzl:4:5: Values: ["-c", "$(location a dir/foo.c)"]
DEBUG: arule.bzl:5:5: Expansion: ["-c", "'spaces/a dir/foo.c'"]

SUBCOMMAND: # //spaces:spaces [action 'SkylarkAction spaces/a dir/foo.o']
(cd /private/var/tmp/_bazel_obonilla/124df79d9592378b17b94075fb66c278/execroot/__main__ && \
  exec env - \
  cc -c ''\''spaces/a dir/foo.c'\''' -o 'bazel-out/darwin-fastbuild/bin/spaces/a dir/foo.o')
ERROR: BUILD:13:1: SkylarkAction spaces/a dir/foo.o failed (Exit 1): cc failed: error executing command 
  (cd /private/var/tmp/_bazel_obonilla/124df79d9592378b17b94075fb66c278/execroot/__main__ && \
  exec env - \
  cc -c ''\''spaces/a dir/foo.c'\''' -o 'bazel-out/darwin-fastbuild/bin/spaces/a dir/foo.o')
[...]

PR #9024, opened 2019-07-31
Add cmd_bash, cmd_ps, cmd_bat attributes to genrule

### LINKED ISSUE #7503: Windows: Bash-less alternative for genrule

meteorcloudy:
Based on https://github.com/bazelbuild/bazel/pull/9022

Users now can use Batch or Powershell command in genrule on Windows. 
The attributes have the following priority:
`cmd_ps` > `cmd_bat` > `cmd_bash` > `cmd`, but `cmd_ps` and `cmd_bat` will only be taken into consideration on Windows.

Fixes #7503


src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
```diff
@@ -0,0 +1,51 @@
...
+package com.google.devtools.build.lib.analysis;
+
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.analysis.actions.FileWriteAction;
+
+/**
+ * The class for constructing command line for Batch.
+ */
+public class BatchCommandConstructor implements CommandConstructor {
```
laszlocsomor: final

laszlocsomor: "Batch" is an overloaded term.
Could we call it something else? How about `CmdExeCommandConstructor`?

Also in other places where it's now "batch".

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
```diff
@@ -0,0 +1,51 @@
...
+
+  // `cmd.exe` exists at C:\Windows\System32, which is in the default PATH on Windows.
+  private static final String BATCH_BIN = "cmd.exe";
```
laszlocsomor: How about `CMD_EXE`?

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
```diff
@@ -0,0 +1,51 @@
...
+  private String scriptPostFix;
+
+  BatchCommandConstructor(String scriptPostFix) {
+    this.scriptPostFix = scriptPostFix;
+  }
+
+  @Override
+  public ImmutableList<String> buildCommandLineSimpleArgv(String command) {
+    return ImmutableList.of(BATCH_BIN, "/c", command);
```
laszlocsomor: You probably want these too:
https://github.com/bazelbuild/bazel/blob/5a8e9c520b8ecc4b0b36c4885de38cf030f4b629/src/main/java/com/google/devtools/build/lib/util/CommandBuilder.java#L158-L164

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class PowershellCommandConstructor implements CommandConstructor {
+
+  // `powershell.exe` exists at C:\Windows\System32\WindowsPowerShell\v1.0, which is in the default
+  // PATH on Windows.
+  private static final String POWERSHELL_BIN = "powershell.exe";
+  private static final String[] POWERSHELL_SETUP_COMMANDS = {
+      // 1. Use Set-ExecutionPolicy to allow users to run scripts unsigned.
+      "Set-ExecutionPolicy -Scope CurrentUser RemoteSigned",
+      // 2. Set $errorActionPreference to Stop so that we exit immediately if a command fails.
```
laszlocsomor: Will only stop if a CmdLet fails, not when an external command fails. I think we should clarify that.

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The class for constructing command line for Powershell.
+ */
+public class PowershellCommandConstructor implements CommandConstructor {
+
+  // `powershell.exe` exists at C:\Windows\System32\WindowsPowerShell\v1.0, which is in the default
+  // PATH on Windows.
+  private static final String POWERSHELL_BIN = "powershell.exe";
+  private static final String[] POWERSHELL_SETUP_COMMANDS = {
```
laszlocsomor: Could you store the joined string instead of the parts to join?
Also I believe we prefer `com.google.common.base.Joiner.on("; ").join(...)` to `String.join("; ", ...)`.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
```diff
@@ -115,28 +149,35 @@ public ConfiguredTarget create(RuleContext ruleContext)
       return null;
     }
 
-    String baseCommand = ruleContext.attributes().get("cmd", Type.STRING);
+    CommandType cmdType = cmdTypeAndAttr.first;
+    String cmdAttr = cmdTypeAndAttr.second;
+    boolean expandToWindowsPath = cmdType != CommandType.BASH;
```
laszlocsomor: Hm, doesn't PowerShell also understand Unix-style paths?
(And PowerShell runs on Linux too!)

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.devtools.build.lib.analysis.actions.FileWriteAction;
+
+/**
+ * The class for constructing command line for Powershell.
+ */
+public class PowershellCommandConstructor implements CommandConstructor {
+
+  // `powershell.exe` exists at C:\Windows\System32\WindowsPowerShell\v1.0, which is in the default
+  // PATH on Windows.
+  private static final String POWERSHELL_BIN = "powershell.exe";
```
laszlocsomor: I believe PowerShell runs on Linux too. I understand the motivation here is to support Windows PS, but we might want to support Linux PS later too. Maybe explain this in a comment, or rename the class to `WindowsPowerShellCommandConstructor`?

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
```diff
@@ -77,6 +80,35 @@ protected RuleConfiguredTargetBuilder updateBuilder(
     return builder;
   }
 
+  enum CommandType {
+    BASH,
+    BATCH,
```
laszlocsomor: How about `WINDOWS_CMD_EXE`?

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
```diff
@@ -77,6 +80,35 @@ protected RuleConfiguredTargetBuilder updateBuilder(
...
+    POWERSHELL,
```
laszlocsomor: How about `WINDOWS_POWERSHELL`?

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
```diff
@@ -77,6 +80,35 @@ protected RuleConfiguredTargetBuilder updateBuilder(
...
+  }
+
+  private static Pair<CommandType, String> determineCommandTypeAndAttribute(RuleContext ruleContext) {
+    AttributeMap attributeMap = ruleContext.attributes();
+    // TODO(pcloudy): This should match the execution platform instead of using OS.getCurrent()
+    if (OS.getCurrent() == OS.WINDOWS) {
+      if (attributeMap.isAttributeValueExplicitlySpecified("cmd_ps")) {
```
laszlocsomor: No need to check if it's explicitly defined. It's enough to check if it's empty or not. An empty command cannot possibly make sense (it cannot generate any outputs, so the action would fail anyway), so we can ignore `cmd_ps` if the user deliberately sets it to `""`. 

Same for other attributes.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
```diff
@@ -115,28 +149,35 @@ public ConfiguredTarget create(RuleContext ruleContext)
...
     }
 
-    // Add the genrule environment setup script before the actual shell command.
-    command = String.format("source %s; %s",
-        ruleContext.getPrerequisiteArtifact("$genrule_setup", Mode.HOST).getExecPath(),
-        command);
+    if (cmdType == CommandType.BASH) {
+      // Add the genrule environment setup script before the actual shell command.
+      command = String.format("source %s; %s",
+          ruleContext.getPrerequisiteArtifact("$genrule_setup", Mode.HOST).getExecPath(),
```
laszlocsomor: Do you plan to allow the same for cmd_exe / ps commands?

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
```diff
@@ -166,8 +166,75 @@ are always generated into a predictable location (available via <code>$(@D)</cod
...
+        The Batch command to run on Windows.
+        <p> This attribute has higher priority than `cmd` and `cmd_bash`. The command runs in
+            the similar way as the `cmd` attribute, with the following differences:
+        </p>
+        <ul>
+          <li>
+            This command only applies on Windows.
+          </li>
+          <li>
+            The command runs with `cmd.exe /c`.
```
laszlocsomor: If you use more flags (as I suggested), remember to update here.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
```diff
@@ -166,8 +166,75 @@ are always generated into a predictable location (available via <code>$(@D)</cod
...
+          </li>
+          <li>
+            After <a href="${link make-variables#location}">$(location)</a> and
+            <a href="${link make-variables}">"Make" variable</a> substitution, the paths will be
+            expanded to Windows style paths (with forward slash).
```
laszlocsomor: did you mean backslash? (`\`)

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+/** Tests of {@link BazelGenRule} on Windows.*/
+@RunWith(JUnit4.class)
+public class GenRuleWindowsTest extends BuildViewTestCase {
+
+  private static final Pattern BATCH_COMMAND_PATTERN =
+      Pattern.compile("(?<command>.*)");
+
+  private static final Pattern POWERSHELL_COMMAND_PATTERN =
+      Pattern.compile(".*'utf8';\\s+(?<command>.*)");
```
laszlocsomor: I think we should assert that the script is written as utf-8. That's what we really want. Whether we tell PS this as a command line flag is irrelevant.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests of {@link BazelGenRule} on Windows.*/
+@RunWith(JUnit4.class)
+public class GenRuleWindowsTest extends BuildViewTestCase {
+
+  private static final Pattern BATCH_COMMAND_PATTERN =
+      Pattern.compile("(?<command>.*)");
```
laszlocsomor: Hm, isn't this regex equivalent to `.*`, i.e. matching any string?

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
```diff
@@ -0,0 +1,142 @@
...
+    Artifact messageArtifact = getFileConfiguredTarget("//genrule1:message.txt").getArtifact();
+    SpawnAction shellAction = (SpawnAction) getGeneratingAction(messageArtifact);
+
+    assertThat(shellAction).isNotNull();
+    assertThat(shellAction.getOutputs()).containsExactly(messageArtifact);
+
+    String expected = "echo \"Hello, Batch cmd.\" >" + getWindowsPath(messageArtifact);
+    assertThat(shellAction.getArguments().get(0)).isEqualTo("cmd.exe");
+    assertThat(shellAction.getArguments().get(1)).isEqualTo("/c");
+    assertBatchCommandEquals(expected, shellAction.getArguments().get(2));
```
laszlocsomor: I think you can replace `assertBatchCommandEquals` with `assertThat(...).isEqualTo(...)`

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    Artifact messageArtifact = getFileConfiguredTarget("//genrule1:message.txt").getArtifact();
+    SpawnAction shellAction = (SpawnAction) getGeneratingAction(messageArtifact);
+
+    assertThat(shellAction).isNotNull();
+    assertThat(shellAction.getOutputs()).containsExactly(messageArtifact);
+
+    String expected = "echo \"Hello, Powershell cmd.\" >" + getWindowsPath(messageArtifact);
+    assertThat(shellAction.getArguments().get(0)).isEqualTo("powershell.exe");
+    assertThat(shellAction.getArguments().get(1)).isEqualTo("/c");
```
laszlocsomor: Could you verify that the generated file is indeed utf-8?

src/test/py/bazel/genrule_test.py
```diff
@@ -0,0 +1,148 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import unittest
+from src.test.py.bazel import test_base
+
+
+class GenRuleTest(test_base.TestBase):
```
laszlocsomor: You should also assert quoting -- what if an input or output file (`$@`) has a space? (`$(location)` also supports spaces.)

See also https://github.com/bazelbuild/bazel/issues/6531

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
meteorcloudy: That was exactly what I did. But I broke some tests because `cmd` is no longer a mandatory attribute. So I also want to ensure at least one cmd attribute got specified. If I don't use `isAttributeValueExplicitlySpecified`, I cannot tell the value is empty is because the attribute is not set or it's deliberately set to "". And as you said, setting an cmd attribute to "" doesn't make sense, I prefer to not silently ignore it.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
meteorcloudy: Nope, we can setup the environment without a setup script in those two.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
meteorcloudy: Yes! Thanks!

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
meteorcloudy: Yeah!

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: No, if we don't set this and you do an `echo hello > foo` in powershell, `foo` will be encoded in UTF-16, that's probably not users want.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: Ah, that's true. Let me remove the redundancy.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: Thanks! Done.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: I don't think the action is executed here. So I cannot check. But in the python test, if I don't set `$PSDefaultParameterValues['*:Encoding'] = 'utf8'`, the generated file will be in utf-16, which caused the test to fail.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
meteorcloudy: That's a good point! If Powershell supports Unix-style path well enough, I think we should just use path with forward slash. I'll change this and see if tests still pass.

src/test/py/bazel/genrule_test.py
meteorcloudy: I checked `$@` won't add any quotes if the path contains spaces, but `$(location)` does add single quote in that case. Single quote will work in powershell but not in cmd.exe. However, I don't think that's a behavior related to my change and we already have tests for that in other places.  See https://github.com/bazelbuild/bazel/commit/aacc305ba5a2035155cdc9afc9a300c739282829

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
laszlocsomor: I thought of it as a batch-style CommandConstructor, in the sense of `bazel --batch`.

I actually don't know whether "Batch" is the name of the language. I don't know if this language even has a name... I know `.bat` files are called batch files, I guess because they batch commands together.

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
laszlocsomor: I don't know.

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
laszlocsomor: SGTM!

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
laszlocsomor: Thanks!

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
laszlocsomor: 1. If `cmd` is no longer a mandatory attr, isn't that an incompatible change? Is it mandatory to define at least one of cmd, cmd_bash, cmd_bat, and cmd_ps?

2. Setting the attribute to "" (or None) is customary inside macros:

    ```
    def my_genrule(name, cmd, cmd_bat=None, cmd_ps=None, **kwargs):
      native.genrule(name = name, cmd = cmd, cmd_bat=cmd_bat, cmd_ps=cmd_ps, **kwargs)
    ```

    These attributes are explicitly defined and may be legitimately empty.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
laszlocsomor: Agreed! I think the test has the good intention but does the wrong assertion.

I meant to suggest you verify that the output file indeed is utf8-encoded. We know the way to do that is passing some flag to PS, but that's not the end result we want to test for.

To illustrate what I mean, if you wanted to integration-test the `zip` tool, you'd want to assert the contents of the produced `zip` file, not necessarily the command arguments that lead to generating that zip file. Does that make sense?

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
meteorcloudy: Ah, I see. Indeed, this could be confusing. But it is commonly called "Windows Batch Script", see this link: https://en.wikibooks.org/wiki/Windows_Batch_Scripting

So to make it more clear, I think we can Rename it to "WindowsBatchCommandConstructor" and also for powershell "WindowsPowershellCommandConstructor"

src/main/java/com/google/devtools/build/lib/analysis/PowershellCommandConstructor.java
meteorcloudy: No problem, as long as we make it clear in documentation.

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
laszlocsomor: SGTM!

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
laszlocsomor: Can we test in Python that the file is indeed utf-8 and not utf-16?

I mean, if we only test that we pass the right flag to PS it's only as good as testing that you sent a message to someone -- it doesn't verify the recipient actually got your message.

laszlocsomor: I think @ulfjack found such a bug in Bazel a few weeks ago : the integration tests only asserted that Bazel executed the right command, they didn't assert the output of that command.

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: Ah, I see. But this is not an integration test here. This is actually only a configured target test. We want to test the correct command is generated for the genrule, but the test won't actually execute the command. We have the integration test in `genrule_test.py`, in that test, we read the generated file with utf-8 encoding and compared the content is expected. Is that enough?

src/test/py/bazel/genrule_test.py
laszlocsomor: Thanks

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: And I'll rename the test to `GenruleWindowsConfiguredTargetTest`

src/test/py/bazel/genrule_test.py
```diff
@@ -0,0 +1,148 @@
...
+    exit_code, _, stderr = self.RunBazel(['build', '//foo:x'])
+    self.AssertExitCode(exit_code, 0, stderr)
+
+    script = os.path.join(bazel_bin, 'foo', 'x.genrule_script.ps1')
+    hello = os.path.join(bazel_bin, 'foo', 'hello_world')
+    self.assertTrue(os.path.exists(script))
+    self.assertTrue(os.path.exists(hello))
+
+    expected_content = '\n'.join(['Hello world' for _ in range(0, 1000)])
+    self.AssertFileContentContains(hello, expected_content)
```
meteorcloudy: @laszlocsomor This will fail without `$PSDefaultParameterValues['*:Encoding'] = 'utf8'`, because we read the file content with the default encoding (utf-8). This is how I discovered Powershell is using utf-16 as default encoding. I think it does exact what you suggested in other comments?

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
laszlocsomor: SGTM

src/test/java/com/google/devtools/build/lib/bazel/rules/genrule/GenRuleWindowsTest.java
meteorcloudy: I agreed, I think we already have such a test in genrule_test.py, that is how I discovered we need to set `$PSDefaultParameterValues['*:Encoding'] = 'utf8'`

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
meteorcloudy: 1. It's mandatory to define at least one of cmd, cmd_bash on Linux and at least one of cmd, cmd_bash, cmd_bat, and cmd_ps on Windows.
For current users that only use `cmd` in genrule, I don't think it is an incompatible change because it will still fail when `cmd` is not set explicitly.

2. I tested, if `cmd_ps = None`, `isAttributeValueExplicitlySpecified("cmd_ps")` is `False`, so it will be ignored. That means if users want to override the value in the macro, they should set the default value to `None` instead of `""`.

src/main/java/com/google/devtools/build/lib/analysis/BatchCommandConstructor.java
meteorcloudy: DOne

src/test/py/bazel/genrule_test.py
laszlocsomor: Yes. Thanks!

src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBase.java
laszlocsomor: Thanks!

meteorcloudy:
Thanks! \o/

PR #9028, opened 2019-07-31
Refactor SimpleBlobStore#put

buchgr:
Split it into `SimpleBlobStore#uploadFile` and `SimpleBlobStore#uploadBlob`.
The return type changes to `ListenableFuture<Void>` but all
implementations are still blocking for now.

This is a refactoring towards the larger goal of removing `SimpleBlobStoreActionCache`
and making GrpcRemoteCache a SimpleBlobStore implementation.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
@@ -92,7 +92,7 @@ public final class SimpleBlobStoreActionCache extends AbstractRemoteActionCache 
     upload(result, actionKey, action, command, execRoot, files, /* uploadAction= */ true);
     if (outErr.getErrorPath().exists()) {
       Digest stdErrDigest = digestUtil.compute(outErr.getErrorPath());
-      uploadFile(stdErrDigest, outErr.getOutputPath());
+      uploadFile(stdErrDigest, outErr.getErrorPath());
       result.setStderrDigest(stdErrDigest);
     }
     if (outErr.getOutputPath().exists()) {
--- a/src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
@@ -21,6 +21,7 @@ import build.bazel.remote.execution.v2.Digest;
 import build.bazel.remote.execution.v2.Directory;
 import build.bazel.remote.execution.v2.DirectoryNode;
 import build.bazel.remote.execution.v2.FileNode;
+import com.google.common.base.Throwables;
 import com.google.common.hash.HashingOutputStream;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
@@ -38,14 +39,13 @@ import com.google.devtools.build.lib.util.io.FileOutErr;
 import com.google.devtools.build.lib.vfs.Path;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
-import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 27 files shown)


buchgr:
@borkaehw

borkaehw:
Do we have an ETA for this PR?

SrodriguezO:
`SimpleBlobStoreActionCache` extends `AbstractRemoteActionCache`, which is what the cache is abstracted away to in `RemoteModule.java`. 

If we're removing `SimpleBlobStoreActionCache`, is the plan to have `SimpleBlobStore` extend `AbstractRemoteActionCache` directly?

buchgr:
@SrodriguezO the plan is for the `AbstractRemoteActionCache` to no longer be abstract but instead for it to has-a `SimpleBlobStore`. Grpc, Http and Disk should all be of type `SimpleBlobStore`. 

The do the following renames:

`AbstractRemoteActionCache -> RemoteCache`
`SimpleBlobStore -> RemoteCacheProtocol`

SrodriguezO:
Ah that makes sense. Sounds great! Thanks for clarifying

src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
```diff
@@ -135,54 +125,30 @@ public void upload(
...
+      Throwables.propagateIfPossible(e.getCause(), IOException.class, InterruptedException.class);
+      throw new IOException(String.format("Uploading of file '%s' failed: %s", file.getPathString(), e.getCause()));
     }
   }
 
-  public Digest uploadStream(Digest digest, InputStream in)
-      throws IOException, InterruptedException {
-    final String hash = digest.getHash();
-
-    if (storedBlobs.putIfAbsent(hash, true) == null) {
```
ishikhman: It seems that with this logic we tried to avoid uploading to BlobStore duplicates. Is it safe to just remove it now?

src/main/java/com/google/devtools/build/lib/remote/http/HttpBlobStore.java
```diff
@@ -552,36 +553,64 @@ public void close() {
...
-        HttpResponse response = ((HttpException) e).response();
-        if (authTokenExpired(response)) {
-          refreshCredentials();
-          // The error is due to an auth token having expired. Let's try again.
-          if (!reset(in)) {
-            // The InputStream can't be reset and thus we can't retry as most likely
-            // bytes have already been read from the InputStream.
-            throw e;
+    boolean success = false;
+    if (storedBlobs.putIfAbsent((casUpload ? "cas/" : "ac/") + key, true) == null) {
```
ishikhman: Move "cas/" and "ac/" to constants?

src/test/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCacheTest.java
```diff
@@ -283,39 +283,6 @@ public void testDownloadDirectoriesWithSameHash() throws Exception {
     assertThat(DIGEST_UTIL.compute(execRoot.getRelative("a/foo/file"))).isEqualTo(fileDigest);
   }
 
-  @Test
```
ishikhman: These tests got removed together with the logic from `SimpleBlobStoreActionCache`, but the logic is reused now in `HttpBlobStore`, maybe add these tests for new logic in `HttpBlobStore`?

src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
buchgr: I just moved the logic to `HttpBlobStore`. I believe that's where it belongs. For the `GrpcRemoteCache` the `ByteStreamUploader` also has its duplication logic. I can see good arguments for either having it in the AbstractRemoteCache or the protocol implementation, but I'd prefer to go with doing the de-duplication in the protocol implementation for simplicity.

We can re-evaluate once SimpleblobstoreActionCache is no more?

src/main/java/com/google/devtools/build/lib/remote/http/HttpBlobStore.java
buchgr: done

src/main/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCache.java
ishikhman: ok, sounds good! Let's revisit it later :)

src/test/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCacheTest.java
buchgr: good point. done!

Issue #7630, opened 2019-03-05
"bes upload" uploads unnecessary files


EricBurnett:
### Description of the problem / feature request:

Bazel appears to attempt to upload every file referenced by BES to the CAS, even ones it just downloaded *from* the CAS. This is not harmful, but somewhat wasteful - one extra RPC per file, and at least one round-trip worth of file contents (more if the CAS implementation doesn't short-circuit `OK` blobs it already has without waiting).

![image](https://user-images.githubusercontent.com/201307/53811762-10867600-3f28-11e9-892c-d281a2c05c90.png)

Consider this an optimization request, but not super high priority.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run a remote build with BES enabled, and look at the RPCs issued.

### What operating system are you running Bazel on?

Linux

### What's the output of `bazel info release`?

0.23.0

buchgr:
Thanks for the report. I have been wanting to fix this for a while, but it's never been a high enough priority. I ll probably get to this by the end of this week, because it's a strict requirement for "Remote Builds without the Bytes".

coeuvre:
is this still an issue?

EricBurnett:
I have no idea. Let's call it obsolete - if it shows up again as material we can do something about it, but if either it's no longer happening, or it is but doesn't seem to be important for performance in practice, it's probably not worth spending time on.

Issue #8673, opened 2019-06-18
RE combined with BES is slow with large output artifacts


cphang99:
### Description of the problem / feature request:

When BES is used in conjunction with remote-execution, output artifacts tagged `no-remote` and `no-cache` are uploaded to the CAS. When these outputs are large in size, this can result in significant performance penalties.

More concretely, we have seen significant performance penalities when BES is used in conjunction with RE, but *not* when each was used in isolation.

It would be beneficial to introduce a bazel command line flag which would disable BES artifact upload for output artifacts corresponding to targets that are marked `no-remote` and `no-cache`. The paths for output files corresponding to these targets should be local (or none at all) and not bytestream:/// pathnames.

### What operating system are you running Bazel on?

macOS 10.14.x

### What's the output of `bazel info release`?

release 0.25.3

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/issues/8249 refers to how file URIs are automatically translated to bytestream:/// paths in remote execution/remote cache contexts

buchgr:
Interesting. Are you suggesting that we should not reference these files at all then in the BEP?

ishikhman:
Hm, can it be because those tags are not properly propagated? 
What rules were you using for those tagged targets?

cphang99:
> Interesting. Are you suggesting that we should not reference these files at all then in the BEP?

I think so. We should not reference them with bytesream paths if that results them being uploaded to the CAS.

cphang99:
> Hm, can it be because those tags are not properly propagated?
> What rules were you using for those tagged targets?

We’re using a custom rule, but we can verify that the `no-remote` and `no-cache` tags are being correctly propagated.

For a particular target that produces 2GB artefacts, with `no-remote` and `no-cache` tags set, a bazel build of that target is 10x slower, but *only* when BES + remote execution is enabled. Interestingly, this is not shown in the elapsed time, but only on using a `time` command.
```
INFO: Elapsed time: 4.706s, Critical Path: 0.63s
INFO: 0 processes.

real    0m40.333s
user    0m0.040s
sys     0m0.028s
```

cphang99:
One suggestion:

We have observed that enabling `—experimental_remote_download_outputs` i.e. builds without the bytes makes the issue go away in 0.25.3. We have tracked this down to here 

[...]

Issue #8573, opened 2019-06-06
Release 1.0 - September 2019


dslomov:
Target RC date - September 3rd, 2019.
See the  [blog post](https://blog.bazel.build/2019/06/06/Bazel-Semantic-Versioning.html)  for some details

kkiningh:
There's quite a few issues with the [1.0 tag](https://github.com/bazelbuild/bazel/labels/bazel%201.0) left, including some that seem fairly significant. Which ones does the Bazel team still expect to address before the release?

dslomov:
> There's quite a few issues with the [1.0 tag](https://github.com/bazelbuild/bazel/labels/bazel%201.0) left, including some that seem fairly significant. Which ones does the Bazel team still expect to address before the release?

We will do a review of those. If you think a specific issue is particularly important to address for 1.0, please comment on it.

katre:
Note that aaff22cadcb59f5497b9a84879ca2ca2a3b1323b and 7ed66f7a31bf5c174f05c295c07fe90cc2aaf2b9 are rollbacks of commits that are present in 0.29.0. They should be rolled forward again before 1.0 is cut, but we need to make sure of that when choosing the baseline.

(They were rolled back due to an internal issue that does not affect Bazel. See http://b/138789815 for the internal discussion.)

cgruber:
In particular, both early phases of the rules_kotlin and rules_android roadmaps seem not to have any indications of their progress, and I have had several assurances that their "mid-2019" goals would roll out as a part of the Bazel 1.0 process.  Can someone please confirm this, or otherwise inform folks who are waiting on those?

dslomov:
Current list of blockers is https://github.com/bazelbuild/bazel/labels/bazel%201.0

dslomov:
97a8264 is a 1.0 baseline I am going to try. Failing downstream projects:
* TensorFlow: https://github.com/bazelbuild/bazel/issues/7362#issuecomment-527115823              
   *  Need fixes from TensorFlow
* rules_foreign_cc: https://github.com/bazelbuild/rules_foreign_cc/pull/309
  *  Needs fix from rules_foreign_cc
* rules_typescript: fix in flight
* rules_nodejs: rules_nodejs/.bazelci/presubmit.yml was changed, a wrong target was added but the following fixes didn’t get in because the pipeline was red for some other reason. Filed https://github.com/bazelbuild/rules_nodejs/issues/1096

dslomov:
[...]

PR #9031, opened 2019-07-31
Introduce flag to conditionally disable BES artifact upload

arlyon:
Issue #8673 describes a pain point with the BES when dealing with large output artifacts from RE. This patch introduces a flag to the bazel CLI that conditionally disables BES upload for targets marked with `no-remote` or `no-cache` (even without builds without the bytes).

When the flag `--bep_disable_artifact_upload` is provided, the BES artifact uploader will respect the tags when determining which outputs should be sent to the BES, and ignore files referenced by any events if the target that triggered that event is tagged with either of the aforementioned tags.

This significantly reduces build times in some cases, but is contingent on being able to mark all the offending targets as no-cache or no-remote. I understand that this may not be ideal for all use cases. Some potential solutions:

1) make the flag enable or disable the functionality for _all_ targets, not just those with specific tags (this is the functionality present in the first commit in this patchset)
2) make the tags that enable this functionality configurable
3) introduce a new tag that controls this functionality

Criticisms or opinions welcome.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventstream/BuildEvent.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventstream/BuildEvent.java
@@ -53,10 +53,18 @@ public interface BuildEvent extends ChainableEvent, ExtendedEventHandler.Postabl
 
     public final Path path;
     public final LocalFileType type;
+    public final boolean shouldUploadArtifacts; /** whether the local file should be uploaded to the BEP */
 
     public LocalFile(Path path, LocalFileType type) {
       this.path = Preconditions.checkNotNull(path);
       this.type = Preconditions.checkNotNull(type);
+      this.shouldUploadArtifacts = true;
+    }
+
+    public LocalFile(Path path, LocalFileType type, boolean shouldUploadArtifacts) {
+      this.path = Preconditions.checkNotNull(path);
+      this.type = Preconditions.checkNotNull(type);
+      this.shouldUploadArtifacts = shouldUploadArtifacts;
     }
 
     @Override
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
@@ -91,7 +91,7 @@ class ByteStreamBuildEventArtifactUploader implements BuildEventArtifactUploader
                 }
                 DigestUtil digestUtil = new DigestUtil(file.getFileSystem().getDigestFunction());
                 Digest digest = digestUtil.compute(file);
-                if (isRemoteFile(file)) {
+                if (isRemoteFile(file) || !files.get(file).shouldUploadArtifacts) {
                   return Futures.immediateFuture(new PathDigestPair(file, digest));
                 }
                 Chunker chunker = Chunker.builder().setInput(digest.getSizeBytes(), file).build();
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
@@ -26,15 +26,21 @@ import com.google.devtools.build.lib.buildeventstream.BuildEvent.LocalFile;
 import com.google.devtools.build.lib.buildeventstream.BuildEventArtifactUploader;
 import com.google.devtools.build.lib.buildeventstream.PathConverter;
 import com.google.devtools.build.lib.remote.util.DigestUtil;
+import com.google.devtools.build.lib.util.Pair;
 import com.google.devtools.build.lib.vfs.Path;
 import io.grpc.Context;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Collection;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
@@ -48,20 +54,25 @@ class ByteStreamBuildEventArtifactUploader implements BuildEventArtifactUploader
   private final String remoteServerInstanceName;
 
   private final AtomicBoolean shutdown = new AtomicBoolean();
+  private final String remoteInstanceName;
+  private final CasDigestLookup digestLookup;
 
   ByteStreamBuildEventArtifactUploader(
       ByteStreamUploader uploader,
       String remoteServerName,
       Context ctx,
       @Nullable String remoteInstanceName,
-      int maxUploadThreads) {
+      int maxUploadThreads,
+      CasDigestLookup digestLookup) {
     this.uploader = Preconditions.checkNotNull(uploader);
     String remoteServerInstanceName = Preconditions.checkNotNull(remoteServerName);
-    if (!Strings.isNullOrEmpty(remoteInstanceName)) {
-      remoteServerInstanceName += "/" + remoteInstanceName;
+    this.remoteInstanceName = remoteInstanceName;
+    if (!Strings.isNullOrEmpty(this.remoteInstanceName)) {
+      remoteServerInstanceName += "/" + this.remoteInstanceName;
     }
     this.ctx = ctx;
     this.remoteServerInstanceName = remoteServerInstanceName;
+    this.digestLookup = digestLookup;
     // Limit the maximum threads number to 1000 (chosen arbitrarily)
     this.uploadExecutor =
         MoreExecutors.listeningDecorator(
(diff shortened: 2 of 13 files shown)


buchgr:
@arlyon thanks for the patch!

I'd like to understand the issue that you are trying to solve better:

- Are you using remote execution or remote caching?
- If remote execution, why are you tagging targets with no-remote instead of using builds without the bytes?
- Bazel currently has a bug where it always tries to upload files to the remote cache even if the cache already contains them. If this bug was fixed would you still need the flag you are proposing?

Thanks!

arlyon:
Hey @buchgr 

1) Both, with the BES enabled, but with build without the bytes disabled.
2) I am not able to elaborate.
3) Without remote execution (but with remote caching) the problem does not occur, likely due to the fact that BES and the Remote Cache share the same `ByteStreamUploader`, which is able to prevent duplicate file hashes from being submitted twice. When remote execution is introduced, this hash set is not consistent between the local bazel client and the remote execution server. I assume that that is the bug you are referring to?

After some further discussion today, we discovered that this inconsistency is essentially what causes the symptoms that this PR fixes. A side effect of that being "bursty" event upload, and excessive CAS usage. Obviously fixing the root cause is better than this patch which merely addresses the symptoms. Is there an open issue for this that I can look at? I would be keen to fix it.

arlyon:
I think it may be better to push only the first commit (which globally disables artifact upload), which will resolve the duplicate upload problem until a proper fix is implemented.

buchgr:
@arlyon https://github.com/bazelbuild/bazel/issues/7630

buchgr:
@arlyon there's also `--bes_upload_mode=nowait_for_upload_complete` that might be of interest to you.

jjardon:
Hi @buchgr , I've just wanted to confirm you are ok with going with the first commit of this PR (which globally disables artifact upload and will solve the initial issue), until a more elaborated fix is implemented

buchgr:
I don't think we can do that unfortunately as this will break BES for a lot of people. I think #7630 isn't that hard to fix but I don't have time for it this week :/.

- Add an abstraction that allows sharing a set of hashes between `AbstractRemoteActionCache` and `BuildEventArtifactUploader`.
- For every action output that `AbstractRemoteActionCache` downloads make an entry of its content hash in the set. In BuildEventArtifactUploader check if the hash of the file is in the set before doing the upload. If it's in the set you know the remote side has the data already.

arlyon:
If we were to get that patch up this week, would it be possible to hit the 1.0 merge window? I am starting work on #7630 now, and aim to be done this week. In the case that it isn't, is it possible to get the first commit merged for that deadline with the guarantee that the full fix is in progress?

The first commit is transparent for anyone who doesn't enable the flag and will not break BES. Those who do enable the flag disable artifact upload (keeping event upload in tact) and will only "break" the BES when they are using remote execution without caching at which point the artifacts are never uploaded to the CAS, and the BES references files that do not exist. It is assumed that those who are using the flag are aware of the consequences.

arlyon:
ByteStreamUploader also has some functionality to detect duplicate uploads; is it maybe an idea to have that also consult the same hash set?

buchgr:
1.0 will be cut on September 2nd. I believe that this should be enough time.

arlyon:
Hey @buchgr after further investigation, while your proposal to fix #7630 does partially solve the problem (by removing duplicate uploads), there is an edge case; since the cached blob set is ephemeral, any files that are pulled from the local cache bypass it because they are never downloaded. So, in situations where all the needed files already exist in the local cache, the BES will upload all the artifacts.

A solution to this is to, on start, hash the files in the local cache and to do a "get missing blobs" to determine which ones the CAS has and pre-populate our hashset from that. Does this seem reasonable?

Other than that, the fix is working (and tested).

buchgr:
> A solution to this is to, on start, do a "get missing blobs" on the files in the local cache to determine which ones the CAS has and pre-populate our hashset from that. Does this seem reasonable?

The "get missing blobs" call is a great idea. I believe this might be a good enough general fix and save us from maintaining the hashset. what do you think?

arlyon:
> The "get missing blobs" call is a great idea. I believe this might be a good enough general fix and save us from maintaining the hashset. what do you think?

The implementation I settled on is a hybrid. It maintains a hashset locally of hashes known to be cached, and for the digests it doesn't have an answer for it consults the CAS.

buchgr:
I'd prefer to only do a simple FindMissingBlobs() call in the BytestreamBuildEventArtifactUploader and leave other parts of the code untouched. My reasoning is that it's 1) simple 2) a strict improvement and good enough as far as we know.

arlyon:
Hey @buchgr thanks for the feedback, I've simplified the patch to do missing blob requests. There's a bit of messing around with futures because of a synchronization point due to the requirement that all the submitted are checked in a single request but it is fairly easy to understand if you are familiar with Futures.

Keen to hear your feedback and get this merged soon.

buchgr:
Thanks. Instead of duplicating the code for `getMissingDigests` could you reuse the one in GrpcRemoteCache?

arlyon:
Hey @buchgr I've incorporated your suggestions and fixed the failing CI test case. Let me know what you think.

Alex

src/main/java/com/google/devtools/build/lib/remote/CasDigestLookup.java
```diff
@@ -0,0 +1,114 @@
...
+import io.grpc.Context;
+import io.grpc.StatusRuntimeException;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class CasDigestLookup {
```
buchgr: Why have this class? Why no reuse getMissingDigests from GrpcRemoteCache?

src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
```diff
@@ -48,74 +54,159 @@
...
-                return Futures.transform(
-                    upload, unused -> new PathDigestPair(file, digest), uploadExecutor);
+
+                return Futures.allAsList(cachedPairFutures);
               },
-              MoreExecutors.directExecutor());
-      uploads.add(digestFuture);
-      }
+              MoreExecutors.directExecutor()
+          );
```
buchgr: I think the code would be easier to read if the lambdas passed to `Futures.transformAsync` would be methods.

src/main/java/com/google/devtools/build/lib/remote/CasDigestLookup.java
arlyon: From my interpretation there is a clear boundary between the GrpcRemoteCache and the ByteStreamBuildEventArtifactUploader and I didn't want to have the latter be aware of the former. Perhaps having the GrpcRemoteCache implement a CasDigestLookup interface would be a better approach, I'll do that.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
arlyon: Will do.

cphang99:
Hi @buchgr thanks very much for the reviews. Assuming you're happy with this set of reviews, can we get this in before the 1.0 cut (#8573) on Monday?

src/main/java/com/google/devtools/build/lib/buildeventstream/BuildEvent.java
```diff
@@ -53,10 +53,18 @@
 
     public final Path path;
     public final LocalFileType type;
+    public final boolean shouldUpload; /** whether the local file should be uploaded to the BEP */
```
buchgr: afaik this is unused?

arlyon: It is used for the bep_disable_all_artifact_upload flag here: https://github.com/arlyon/bazel/blob/eb94fe6caddfe89bcc396be556160f30a1bb618f/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java#L103

arlyon: And here is where the flag is detected: https://github.com/arlyon/bazel/blob/0a5961bd0e6594800ad5cd9a26a641472d6466cd/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceUploader.java#L178

If the flag is provided, any artifacts referenced by build events are not uploaded. Perhaps a better name would be `shouldUploadArtifacts`

dslomov:
Is this really blocking for 1.0?

buchgr:
Yes it's a big performance issue. Will try to get it merged today.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
```diff
@@ -78,44 +89,111 @@ private static boolean isRemoteFile(Path file) {
...
+          toUpload.put(digest, new LinkedList<>());
+        }
+        toUpload.get(digest).add(path);
+      } else {
+        pathDigestFutures.add(Futures.immediateFuture(pathPair.second));
+      }
+    }
+
+    Context prevContext = ctx.attach();
+    ImmutableSet<Digest> missingDigests = digestLookup.getMissingDigests(toUpload.keySet());
```
buchgr: That's a blocking call. This won't fly :\.

buchgr:
@arlyon given that this needs to go in to Bazel *today* I have taken the liberty to create a PR based on your change https://github.com/bazelbuild/bazel/pull/9320 that addresses a few issues:

* The getMissingDigests call must not be blocking
* Moves the digestUtil.compute(...) to the thread pool
* Does not introduce a flag to disable BES upload. I am not a fan of the idea as it complicates things quite a bit and I would like to see how far calling getMissingDigests gets us. We may revisit at a later point.

arlyon:
> @arlyon given that this needs to go in to Bazel _today_ I have taken the liberty to create a PR based on your change #9320 that addresses a few issues:
> 
>     * The getMissingDigests call must not be blocking
> 
>     * Moves the digestUtil.compute(...) to the thread pool
> 
>     * Does not introduce a flag to disable BES upload. I am not a fan of the idea as it complicates things quite a bit and I would like to see how far calling getMissingDigests gets us. We may revisit at a later point.

I had the call to `getMissingDigests` return a Future previously (and can revert it to do that) but changed it to just return the raw set because that is what `GrpcRemoteCache` does. Shame about the flag, but understandable. Glad to see this part get merged nonetheless. Feel free to close as you see fit.

buchgr:
This has been merged as https://github.com/bazelbuild/bazel/commit/120cd54c3cdb6e90470ffd3ff275531f776f54a8

PR #9034, opened 2019-08-01
Remove dummy toolchains and add the correct toolchain types.

iirina:


aiuto:
@iirina Is this still valid or should it be closed?

PR #9037, opened 2019-08-01
Hard links for files in build-runfiles to allow non-elevated runs.

excitoon:
Hi.

What do you think about using hard links for files on Windows instead of soft links? I suppose, it will simplify things for users, because hard links do not require administrator rights.

@meteorcloudy


meteorcloudy:
@excitoon Sorry, I don't think we should use hard link for runfiles. We've considered this, but a hard link refers to the same file. If the link or the original file is in opened by another process, you cannot delete it or even modify it. This could cause some problems. Also, creating hard link is quite slow and you cannot create more than 1023 hard links to the same file.

excitoon:
Okay!

PR #9039, opened 2019-08-01
Refactor GrpcRemoteCache to use uploadFile and uploadBlob.

buchgr:
The methods have the same signature as the ones in SimpleBlobStore. This gets us one step closer to merging AbstractRemoteActionCache and SimpleBlobStoreActionCache.


buchgr:
@borkaehw

borkaehw:
I like the plan we discuss here https://github.com/bazelbuild/bazel/pull/9028#issuecomment-517194971, just wondering are all the works going to happen in this single PR?

PR #9044, opened 2019-08-01
Add stub @rules_python to third_party/

brandjon:
This stub contains only one relevant file, @rules_python//python:defs.bzl,
which mimics the file at the same path in bazelbuild/rules_python. Having this
repo gives us a way to inject this defs.bzl file into our protobuf dependency,
which is loaded as a separate (local) external repo and therefore cannot access
the existing //tools/python:private/defs.bzl in Bazel's own workspace. It also
means we'll be compatible with the future upstream migration to fix protobuf
for #9006.

A separate PR will add this to the Bazel root WORKSPACE file (since
third_party/ must be updated in a separate commit).

Break-out of #9019. Work toward #9006.

RELNOTES: None

### DIFF
--- a/third_party/rules_python/BUILD
+++ b/third_party/rules_python/BUILD
@@ -1,5 +1,5 @@
 filegroup(
     name = "srcs",
     srcs = glob(["**"]) + ["//third_party/rules_python/python:srcs"],
-    visibility = ["//visibility:public"],
+    visibility = ["//third_party:__pkg__"],
 )
--- a/third_party/rules_python/BUILD
+++ b/third_party/rules_python/BUILD
@@ -1,0 +1,5 @@
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]) + ["//third_party/rules_python/python:srcs"],
+    visibility = ["//visibility:public"],
+)
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -35,6 +35,7 @@ filegroup(
         "//third_party/py/gflags:srcs",
         "//third_party/py/mock:srcs",
         "//third_party/py/six:srcs",
+        "//third_party/rules_python:srcs",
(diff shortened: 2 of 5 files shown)


third_party/rules_python/BUILD
```diff
@@ -0,0 +1,5 @@
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]) + ["//third_party/rules_python/python:srcs"],
+    visibility = ["//visibility:public"],
```
katre: Limit the visibility to just //third_party:srcs. Same for python:srcs

brandjon: Done, let's see how CI likes it

third_party/rules_python/BUILD
```diff
@@ -0,0 +1,5 @@
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]) + ["//third_party/rules_python/python:srcs"],
+    visibility = ["//third_party:srcs"],
```
katre: I meant:
visibility = ["//third_party:__pkg__"],
or similar.

brandjon:
Update from offline discussions: Visibility is fixed in most recent commit but we're confirming CI passes. (Turns out my workstation failed bootstrapping tests because of low disk space.)

[Buildkite](https://buildkite.com/bazel/bazel-bazel/builds/9240) testing this commit without visibility fix (in addition to the one auto-triggered by this PR)

[Buildkite](https://buildkite.com/bazel/bazel-bazel/builds/9241) testing this commit with visibility fix

brandjon:
[Passed](https://buildkite.com/bazel/bazel-bazel/builds/9241#7b9a5c0e-b3ff-4144-8042-00d2f67c2684) in manual CI run of branch containing this PR's content.

Issue #8986, opened 2019-07-25
Bazel integration tests get default repos from the actual WORKSPACE file.


aiuto:
We reference some repositories in WORKSPACE then we replace the reference in test harnesses. This is brittle and can lead to tests which do not match the repo version we actually use. Below is the case for rules_cc and shell test. The problem exists for Java and Python tests too. This will get worse as we move more code out of native and //tools and into external repos.

In WORKSPACE:
```
http_archive(
    name = "rules_cc",
    sha256 = "36fa66d4d49debd71d05fba55c1353b522e8caef4a20f8080a3d17cdda001d89",
    strip_prefix = "rules_cc-0d5f3f2768c6ca2faca0079a997a97ce22997a0c",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_cc/archive/0d5f3f2768c6ca2faca0079a997a97ce22997a0c.zip",
        "https://github.com/bazelbuild/rules_cc/archive/0d5f3f2768c6ca2faca0079a997a97ce22997a0c.zip",
    ],
)
```

https://github.com/bazelbuild/bazel/blob/64239e0af91514bb89f4da07794149c58813ff74/src/test/shell/testenv.sh#L452
```
function add_rules_cc_to_workspace() {
  cat >> "$1"<<EOF
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "rules_cc",
    sha256 = "36fa66d4d49debd71d05fba55c1353b522e8caef4a20f8080a3d17cdda001d89",
    strip_prefix = "rules_cc-0d5f3f2768c6ca2faca0079a997a97ce22997a0c",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_cc/archive/0d5f3f2768c6ca2faca0079a997a97ce22997a0c.zip",
        "https://github.com/bazelbuild/rules_cc/archive/0d5f3f2768c6ca2faca0079a997a97ce22997a0c.zip",
    ],
)
EOF
}
```

It is more than just one file. 
```
grep -r 36fa66d4d49debd71d05fba55c1353b522e8caef4a20f8080a3d17cdda001d89 .
./src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE:    sha256 = "36fa66d4d49debd71d05fba55c1353b522e8caef4a20f8080a3d17cdda001d89",
[...]

Issue #9029, opened 2019-07-31
Make Bazel itself depend on `@rules_python`


brandjon:
As part of the work for #9006, we're migrating Bazel's own use of native Python rules to load these rules from a bzl file. Ideally, the load statements should reference the macros in `@rules_python//python:defs.bzl` just like user code should. However, this isn't possible for the `@bazel_tools` repository, which is not allowed to have any external dependencies. Therefore, [we fake it](https://github.com/bazelbuild/bazel/commit/db063a85a7196747cacf6bed44944021108a7ff8) by creating `//tools/python:private/defs.bzl`, which implements the same magic `@rules_python` does, for use by Bazel only.

Of course, there are other uses of Python rules in Bazel's source tree besides those that appear in `@bazel_tools`. Currently some of these also reference the same `//tools/python:private/defs.bzl` file, while others reference a stub `@rules_python` repository (defined in Bazel's own WORKSPACE) that only contains a stub version of `@rules_python//python:defs.bzl`; see #9019.

This issue tracks potentially replacing the stub `@rules_python` with the real thing, and avoiding non-`@bazel_tools` uses of the `//tools/python:private/defs.bzl` path.

Counterarguments for doing the former: The stub shouldn't be hard to keep in sync with the real rules_python, and using a stub avoids extra build overhead (and finagling with the WORKSPACE). Counterarguments for doing the latter: It's not particularly harmful to refer directly to the workspace-relative path, and it allows us to easily have a path that works both in `@io_bazel` and `@bazel_tools` (in the case of the `six` dependency).

aiuto:
Bumping this up.  This is interfering with tests that depend on rules that, in turn, depend on rules_python behind a maybe().  I'm seeing errors like:
```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/3202/execroot/io_bazel/_tmp/69f12340b17380b035a0a8ebbbfbe1ac/root9496731726608493967/c258b9cf248a239110f64db75ed82839/external/rules_pkg/BUILD:15:11: in py_library rule @rules_pkg//:archive: Direct access to the native Python rules is deprecated. Please load py_library from the rules_python repository. See http://github.com/bazelbuild/rules_python and https://github.com/bazelbuild/bazel/issues/9006. You can temporarily bypass this error by setting --incompatible_load_python_rules_from_bzl=false.
[...]

PR #9046, opened 2019-08-01
Update blackbox tests to use rules_pkg instead of built-ins for pkg_tar

aiuto:
We add rules_pkg to the generated WORKSPACE for integration tests.

A better solution may be to figure out why we need tar for the repo tests. If we can eliminate that, a lot of problems go away., but that job belongs to the external deps team.

WIP: Alas, things fail because starlark is getting rid of the outputs attribute, and the test runs with the incompatible changes on. We have to fix that in rules_pkg and use version 0.2.1 of that

WIP: This solution is super brittle because we have to repeat repo urls and sha256  identically in many places. #8986.  Since I have a few weeks before I have to do this, perhaps we can come up with a better design for #8986 now and work towards that.  @philwo: do you want me to propose one or do you want to take a stab at it.


philwo:
Sorry, I won't have time for this, happy to review a working PR, but can't do anything else :/

philwo:
I saw this error in your tests:

```
ERROR: An error occurred during the fetch of repository 'rules_pkg':
   java.io.IOException: Error downloading [https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.2.5/rules_pkg-0.2.5.tar.gz, https://github.com/bazelbuild/rules_pkg/releases/download/0.2.5/rules_pkg-0.2.5.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/3184/execroot/io_bazel/_tmp/69f12340b17380b035a0a8ebbbfbe1ac/root6921373952495163105/a98fd7f2a261cbcdaf147d0d0ed54db5/external/rules_pkg/rules_pkg-0.2.5.tar.gz: Unknown host: github.com
ERROR: no such package '@rules_pkg//': java.io.IOException: Error downloading [https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.2.5/rules_pkg-0.2.5.tar.gz, https://github.com/bazelbuild/rules_pkg/releases/download/0.2.5/rules_pkg-0.2.5.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/3184/execroot/io_bazel/_tmp/69f12340b17380b035a0a8ebbbfbe1ac/root6921373952495163105/a98fd7f2a261cbcdaf147d0d0ed54db5/external/rules_pkg/rules_pkg-0.2.5.tar.gz: Unknown 
```

We run tests without network access by default now and override repositories used by integration tests to be the same as the repos from the "outer" Bazel. Relevant changes that explain what's going on here:

- https://github.com/bazelbuild/bazel/commit/99c0e6dda968c861f5c0c2cbd8c772f50e58fdbc
- https://github.com/bazelbuild/bazel/commit/17506af3667b16850fb8feca82306edded8b1733
- https://github.com/bazelbuild/bazel/commit/9ae6368647f62654e578b620730574127407453d

You probably just have to add `rules_pkg` to this list to fix the test failures: https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/blackbox/bazel/DefaultToolsSetup.java#L30

Also we have to make sure that the version of rules_pkg in Bazel's WORKSPACE file and the one specified in the test's WORKSPACE file is the same. So we should bump rules_pkg in Bazel's WORKSPACE also to 0.2.5 and add a comment where to keep the two in sync.

aiuto:
> https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/blackbox/bazel/DefaultToolsSetup.java#L30

Thanks. I missed that one.

> Also we have to make sure that the version of rules_pkg in Bazel's WORKSPACE file and the one specified in the test's WORKSPACE file is the same. So we should bump rules_pkg in Bazel's WORKSPACE also to 0.2.5 and add a comment where to keep the two in sync.

That is a nice to have, but not actually needed.  Since we only use pkg_tar to package bazel, and for anything else, this integration test could use any version of it.  That points to some of the strangeness here. It makes a tarball, but then never uses it. I have mail out to the original author to confirm my suspicion that pkg_tar was just a convenient way to have a rule that writes content based on another rule and have it done at workspace time.

aiuto:
Damm. Broken possibly because of #9029. Need to come back to this next month.

aiuto:
> > https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/blackbox/bazel/DefaultToolsSetup.java#L30
> 
> Thanks. I missed that one.
> 
> > Also we have to make sure that the version of rules_pkg in Bazel's WORKSPACE file and the one specified in the test's WORKSPACE file is the same. So we should bump rules_pkg in Bazel's WORKSPACE also to 0.2.5 and add a comment where to keep the two in sync.
> 
> That is a nice to have, but not actually needed. Since we only use pkg_tar to package bazel, and for anything else, this integration test could use any version of it. That points to some of the strangeness here. It makes a tarball, but then never uses it. I have mail out to the original author to confirm my suspicion that pkg_tar was just a convenient way to have a rule that writes content based on another rule and have it done at workspace time.

Oh. My bad. I didn't read your comment about forcing repositories to be the outer ones. So yes. I do have to update all of Bazel. PR #11191 will do that first.

aiuto:
I'm rethinking this.  These tests should not have to depend on @rules_pkg.  They should be rewritten to be self contained.

aiuto:
This PR won't work any more.

PR #9048, opened 2019-08-01
Rp outside tp

brandjon:
Next breakout of #9019. Will rebase after #9044 is merged. Manually triggered CI run [here](https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/3930). (Hope I'm not slowing down our Buildkite by doing these manual runs, but I want monotonic CI progress, and updating these PRs kills our CI runs.)


brandjon:
The fix for the test has a bug, will handle it on the internal import side of things.

PR #9052, opened 2019-08-01
Remove worker log by default

keith:
This hash change log doesn't seem actionable by the user, so we might as
well only enable it for the verbose mode.


keith:
This test failure doesn't look related, looks like the bazel download failed

keith:
I saw this for the first time today, with another build it went away so I assume it was an edge case around me switching bazel versions or doing a clean or something

keith:
@philwo is this good to merge?

jin:
Importing for review to philwo@.

Issue #6937, opened 2018-12-15
Versioned Docs: the navigation sidebar should also be versioned


jin:
Currently, the doc contents are versioned. However, the navigation sidebar isn't, and will link to 404s when the requested page does not exist for a particular release, or that the links simply don't exist in the sidebar if it's been removed from git-master.

Neither of these are regressions from the previously git-master-only docs, but it would be nice to have the sidebar versioned as well.

spomorski:
Me and axgillies@ recently discovered this problem recently and it's important that we fix this ASAP.

Not having a matching nav tree per versioned docset makes the versioned docsets feature a lot less useful since users will look for docs that have been moved, renamed, or deleted, not find them in the nav tree when viewing an older docset, and assume the docs are gone.

Can we give this a high priority? I think it's a serious usability problem that negatively impacts the Bazel user experience.

jin:
@fweikert I discovered this and assigned it an initial P2. Could you please take a look and re-triage its priority? Thanks!

fweikert:
@jin - who should own this issue? You're the expert for versioned documentation :)

aiuto:
If we can do any quick fix, that would be fine for now.
For Federation releases, I want to do versioned doc sets that include rules too, with links to bazel at the appropriate federation version. We will have to redo the docs pipeline at that time. 
So then the question is if this EngProd, Product, or DevExp at that time.

jin:
bumping this up to p1. I'll take a look at it sometime next week.

jin:
Moving to team-XProduct as part of the recent team merge.

jin:
Closing this: we can now apply version conditionals in the navigation bar.

jin:
Reopening this issue to actually version the sidebar without conditionals

PR #9057, opened 2019-08-02
Only add "Updating Bazel" link on master or > 0.28 docs.

### LINKED ISSUE #9056: "Updating Bazel" link on https://docs.bazel.build/versions/0.28.0/bazel-overview.html 404s
The leftnav bar is not versioned with the rest of the documentation, and if the link target in the navbar does not exist in a particular version,  the link 404s.

It has happened  with recently added https://docs.bazel.build/versions/master/updating-bazel.html (added in https://github.com/bazelbuild/bazel/commit/4285b4bf14eecbd0f5b7a0b09cad617109c6abab)

dslomov:
Fixes #9056.

### DIFF
--- a/site/_layouts/documentation.html
+++ b/site/_layouts/documentation.html
@@ -200,7 +200,6 @@ nav: docs
               </li>
 
               <li>
-
                 <a class="sidebar-nav-heading" data-toggle="collapse"
                     href="#cpp-menu" aria-expanded="false"
                     aria-controls="cpp-menu">
--- a/site/_layouts/documentation.html
+++ b/site/_layouts/documentation.html
@@ -199,7 +199,7 @@ nav: docs
                 </ul>
               </li>
 
-              <li>{%   assign minor_version = version_components[0] %}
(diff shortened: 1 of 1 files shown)


dslomov:
Principled fix would be versioning of _layouts/documentation.html, but it is more work and I have no quick idea how to start with that.

site/_layouts/documentation.html
```diff
@@ -189,7 +199,8 @@ <h3>Rules</h3>
                 </ul>
               </li>
 
-              <li>
+              <li>{%   assign minor_version = version_components[0] %}
```
laszlocsomor: could you explain why this is needed?

dslomov: accident, removed

jin:
The versioning problem of _layouts.documentation.html is tracked in https://github.com/bazelbuild/bazel/issues/6937. This seems like a good workaround for versioning the sidebar until we find a better solution.

PR #9062, opened 2019-08-02
flipped a flag 'incompatible_tls_enabled_removed'

### LINKED ISSUE #8061: incompatible_tls_enabled_removed: flip it and enable tls by default while deprecating tls_enabled flag
`--tls_enabled` flag is deprecated and should be removed in one of the next releases.
If `--incompatible_tls_enabled_removed` flag set to true, an error is rised in case tls_enabled is still in use.

**Migration**
* gRPC + enabled TLS:
Replace `--tls_enabled=true` and `--remote_cache=localhost:1234` with : `--remote_cache=grpcs://localhost:1234` or `--remote_cache=localhost:1234`
* gRPC + disabled TLS:
Replace `--tls_enabled=false` and `--remote_cache=localhost:1234` with : `--remote_cache=grpc://localhost:1234`

**Note**: the same migration plan applies to all of the following flags: `-remote_cache`, `--remote_executor `or `--bes_backend`. Use `--incompatible_tls_enabled_removed=true` to test.

**Note2**: http cache/executor is not affected

**Expected timeline**
Released with Bazel 0.26 (May 2019), off by default
Enabled by default with Bazel 0.30 (September 2019)

ishikhman:
Set `--incompatible_tls_enabled_removed` to true by default.

RELNOTES: by default all remote connections considered to be via `gRPC` with TLS enabled, unless other specified. To disable TLS use `grpc://` prefix for you endpoints. All remote connections via `gRPC` affected - `--remote_cache`, `--remote_executor` or `--bes_backend`. http cache/executor is not affected. See #8061 for details.

Fixes #8061.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/AuthAndTLSOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/AuthAndTLSOptions.java
@@ -61,19 +61,6 @@ public class AuthAndTLSOptions extends OptionsBase {
   )
   public String googleCredentials;
 
-  @Deprecated
-  @Option(
-      name = "tls_enabled",
-      defaultValue = "false",
-      documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
-      deprecationWarning =
-          "Deprecated. Please specify a valid protocol in the URL (https or grpcs).",
-      effectTags = {OptionEffectTag.UNKNOWN},
-      help =
-          "DEPRECATED. Specifies whether to use TLS for remote execution/caching and "
-              + "the build event service (BES). See #8061 for details.")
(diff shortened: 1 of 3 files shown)


buchgr:
It's too late to flip this flag since 0.29 has already been cut. Next flag flip window will be in March 2020 :'(

ishikhman:
I've decided to keep both `--incompatible_tls_enabled_removed` and `--tls_enabled` as `deprecated` and `no-op` for some time to keep migration smoother and avoid breakage of the builds when one or both flags are still present.

ishikhman:
As of https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1131 no breakage for downstream projects deteceted after merge.

Issue #8830, opened 2019-07-09
incompatible_allow_tags_propagation


ishikhman:
Tags are not currently being propagated from a bazel's target to the actions' execution requirements. which would be very useful in order to be able to mar certain targets as 'no-remote' or 'no-cache'. At the moment, this is possible only on action/rule level or via adjustments to the rules. 
Detailed description of the issue is in #7766.
Design is described in the [proposal](https://docs.google.com/document/d/1X2GtuuNT6UqYYOK5lJWQEdPjAgsbdB3nFjjmjso-XHo/edit?ts=5ca70ed0#heading=h.5mcn15i0e1ch).

In order to make the transition smoother, a flag `--incompatible_allow_tags_propagation `  has been introduced. 

## What does it change?
* If the flag  `--incompatible_allow_tags_propagation ` set to `true`, bazel will take tags specified on target level, filter them based on **prefix** and propagate to the actions, that are created for the target.
**Propagated prefixes**: "block-", "requires-", "no-", "supports-", "disable-", "local", "cpu:"

* If set to `false` tags are not propagated.

## Migration notes
Check the tags with the prefixes: "block-", "requires-", "no-", "supports-", "disable-", "local", "cpu:". Most likely they were not propagated and with the flag `--incompatible_allow_tags_propagation` they will be, therefore the build behavior might change. 

## Plan 
**By Bazel 1.0:** Make functionality usable
- [x] Starlark rules are able to propagate tags under `--incompatible_allow_tags_propagation ` flag
- [x] Native rules (C++, for example) are able to propagate tags under `--incompatible_allow_tags_propagation ` flag

**After Bazel 1.0:**: Stabilize
- [ ] fix everything that is failing/unstable incrementally
- [ ] move 'tags propagation' from _experimental_ to _supported_ feature

laurentlb:
What's the status of this flag?
If it's ready, add the label `migration-ready` (or `migration-0.x`) and `breaking-change-1.0`.

dslomov:
What is the extent of the breakage? How hard it is to migrate? Do we have tooling available?

ishikhman:
**Before the change**: tags were ignored for starlark rules
**After the change**: tags are propagated for starlark rules
Migration would be needed only for the cases where tags were specified for targets where they are not actually required.

What do you mean by 'tooling'?

dslomov:
> What do you mean by 'tooling'?
> 

[...]

PR #9076, opened 2019-08-05
made --incompatible_tags_propagation to be true by default (only star…

ishikhman:
`--incompatible_allow_tags_propagation` flag protects the tags propagation from targets to actions.

RELNOTES: bazel will take tags specified on target level, filter them based on prefix and propagate to the actions, that are created for the target. Propagated prefixes: "block-", "requires-", "no-", "supports-", "disable-", "local", "cpu:". Only Starlark rules affected. See #8830 for details.


ishikhman:
Closing, because flag flip is not stable.

Issue #7974, opened 2019-04-08
incompatible_windows_escape_python_args: on Windows, enables correct escaping of py_binary.args and py_test.args


laszlocsomor:
## Description

Add --incompatible_windows_escape_python_args flag (default: false).

**This is a no-op on Linux/macOS.**

On Linux/macOS/non-Windows: the flag is a no-op.

On Windows: This flag affects how py_binary and py_test targets are built, how their launcher escapes command line flags. When this flag is true, the launcher escapes command line flags using Windows-style escaping (correct behavior). When the flag is false, the launcher uses Bash-style escaping (buggy behavior).

## Background

The output of a py_binary or py_test is an `.exe` file of two parts: a stub (launcher) and a payload (zip file). The payload is the actual Python program with all its runfiles..

To build py_binary/py_test rules, Bazel appends the zip file and some metadata to the launcher. The launcher reads the data from its tail, then sets up the environment and runs Python, passing the command line arguments along.

However the launcher escapes arguments badly. This is the same bug for Python as https://github.com/bazelbuild/bazel/issues/7486 was for Java. The solution is similar: use `WindowsEscapeArg2` in the launcher:
https://github.com/bazelbuild/bazel/blob/20a78a82b69124cf972612cd9c7e6e303c5d570b/src/tools/launcher/python_launcher.cc#L52
https://github.com/bazelbuild/bazel/blob/20a78a82b69124cf972612cd9c7e6e303c5d570b/src/tools/launcher/util/launcher_util.cc#L173-L182

Related bug: https://github.com/bazelbuild/bazel/issues/7958

## The new flag

The `--incompatible_windows_escape_python_args` flag affects the metadata Bazel embeds in the launcher, which governs whether the launcher uses bad or good escaping.

## Migration recipe

None, as of 2019-04-08.

We don't expect any breakages when this flag is enabled. However if it breaks your build or if it breaks your py_binaries or py_tests, please let us know so we can help fixing it and provide a migration recipe.

## Rollout plan

- Bazel 0.25.0 will not support this flag.
- Bazel 0.26.0 is expected to support this flag, with default value being false.
- Bazel 0.27.0 is expected to flip this flag to true, but keep the flag.

PR #9088, opened 2019-08-06
Remove --[no]incompatible_windows_escape_python_args

laszlocsomor:
The flag was flipped in Bazel 0.27.0.

Removing the flag, the code path for "false"
value, and tests.

Fixes https://github.com/bazelbuild/bazel/issues/7974

RELNOTES[INC]: Python, Windows: the --[no]incompatible_windows_escape_python_args is no longer supported. (It was flipped to true in Bazel 0.27.0)


laszlocsomor:
FYI @dslomov

PR #9093, opened 2019-08-06
Cycle remote execute/wait on unresponsive end

### LINKED ISSUE #8605: remote: retry Execute() with status OK but empty ExecuteResponse
If a server closes an Execute()/WaitExecute() call with status OK and an ExecuteResponse.done=false then Bazel should retry the call. This was discovered / requested by @werkt.

werkt:
If an execute/waitExecution stream ends, it can be the result of a
server termination. Ignore this state and rewait for the execution
assuming we are configured to do so.

Fixes #8605

Change-Id: Ib68063795e0482b9a67c2565263cb58098f4c8b6

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -129,49 +129,53 @@ class GrpcRemoteExecutor {
     try {
       return retrier.execute(
           () -> {
-            final Iterator<Operation> replies;
-            if (waitExecution.get()) {
-              WaitExecutionRequest wr =
-                  WaitExecutionRequest.newBuilder().setName(operation.get().getName()).build();
-              replies = execBlockingStub().waitExecution(wr);
-            } else {
-              replies = execBlockingStub().execute(request);
-            }
-            try {
-              while (replies.hasNext()) {
-                Operation o = replies.next();
-                operation.set(o);
-                waitExecution.set(!operation.get().getDone());
-                ExecuteResponse r = getOperationResponse(o);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -129,49 +129,53 @@ public ExecuteResponse executeRemotely(ExecuteRequest request)
...
-                ExecuteResponse r = getOperationResponse(o);
-                if (r != null) {
-                  return r;
-                }
-              }
-            } catch (StatusRuntimeException e) {
-              if (e.getStatus().getCode() == Code.NOT_FOUND) {
-                // Operation was lost on the server. Retry Execute.
-                waitExecution.set(false);
+            for (;;) {
```
buchgr: Why do you feel you need to use an infinite loop as opposed to handling this in the retrier?

werkt: Because the completion is not an innate exception, and shouldn't count against the retry limit. The existing request behavior - waiting if necessary without a deadline, forever - indicates that a completion doesn't represent an error condition, and that only request exceptions can represent transiency worth retrying and expending retry attempts.

buchgr: Apologies for my comment I was thinking of a different problem that you are trying to fix. Makes total sense. May I ask you to add a test so that we don't regress on it?

werkt: Test added, fails without change.

PR #9109, opened 2019-08-07
Enable  incompatible_auto_configure_host_platform

katre:


src/test/java/com/google/devtools/build/lib/rules/platform/PlatformTest.java
```diff
@@ -35,8 +35,11 @@
   @Rule public ExpectedException expectedException = ExpectedException.none();
 
   @Test
+  // TODO(https://github.com/bazelbuild/bazel/issues/6849): Remove this test when the functionality
```
aragos: Is this line still within the char limit? Unsure whether this can be visualized in github review...

src/test/java/com/google/devtools/build/lib/rules/platform/PlatformTest.java
```diff
@@ -35,8 +35,11 @@
...
+  // is removed, but untill then it still needs to be verified.
```
aragos: One l too many in untill.

src/test/java/com/google/devtools/build/lib/rules/platform/PlatformTest.java
katre: Yes, 99 characters long according to vim.

src/test/java/com/google/devtools/build/lib/rules/platform/PlatformTest.java
katre: Fixed.

Issue #7081, opened 2019-01-10
incompatible_auto_configure_host_platform: Switch to new auto-configured host platforms


katre:
Flag: --incompatible_auto_configure_host_platform
Available since: 0.23
Will be flipped in: Unknown

See https://github.com/bazelbuild/proposals/blob/master/designs/2018-10-22-autoconfigured-host-platform.md for more context.
The default host and target platforms are defined in `@bazel_tools//platforms`, and there is special-purpose code in the Platform rule implementation to support this. This complicates the platform rules implementation and makes them dependent on the configuration.

Once this flag is enabled, Bazel will instead default to using a host platform generated by a repository rule, located at `@local_config_platform//:host`, and which will be created only once for a given build. The target platform will also default to this. Once migration is complete, the legacy targets and code can be removed.

Migration notes:
This only affects builds where the `--host_platform` and `--platforms` flags are _not_ used. Any builds currently setting those flags will not be affected.

Any users who rely on `--host_platform` and `--platforms` being affected by the `--host_cpu` and `--cpu` flags will need to begin directly setting `--host_platform` and `--platforms`.

katre:
Changed the targetted flag-flip release, since this will require extensive testing and may need to wait for the projected migration plan for the cpu flag.

hlopko:
Hey @katre, I just re-discovered this flag, what's needed to flip this? Or at least to get to a place where we can ask users to try and to re-enable it on the CI to see what's broken on downstream?

hlopko:
Related: bazelbuild/rules_go#2089, bazelbuild/rules_cc#24, #8763, https://github.com/bazelbuild/bazel/issues/8766.

katre:
Major work left:
1. Set the right tags so that we get migration readiness reports from CI, which I apparently never get right.
2. Write directions so that users who currently use cross-compilation with `--cpu` and `--host_cpu` and platform-related rules can add a platform mapping and continue work. I suspect this is a small set of users but I can't prove that.

hlopko:
Looks like only Bazel is to be migrated: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/175

katre:
Thanks for the pointer!

katre:
Logs: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/191#annotation---incompatible_auto_configure_host_platform

meteorcloudy:
[...]

Issue #6849, opened 2018-12-05
Improve host platform autodetection


katre:
See the proposal at https://github.com/bazelbuild/proposals/blob/master/designs/2018-10-22-autoconfigured-host-platform.md

This issue is to track progress on the work.

katre:
This feature is now enabled. Leaving this open to track the cleanup and removal of old code (after the 1.0 release).

katre:
This is now implemented and should have been closed.

PR #9110, opened 2019-08-07
Fix local_config_platform to also be handled during `bazel sync`.

katre:
Part of #7081 and #6849.


katre:
Sorry for the extra review, I didn't realize this was the cause of the other failures I was seeing.

PR #9114, opened 2019-08-07
Improve performance of string.replace

Quarz0:
This PR improves the performance, in terms of time, of `replace` method by replacing regex by linear scanning which is much faster.
Using the benchmark [here](https://github.com/Quarz0/starlark/blob/benchmark_runner/benchmark_suite/data/string_replace.star), the results obtained on my local machine were as follows:
```
=== data/string_replace.star ===
                   Time (MS)  Memory (KB)
java               10777.098    89340.000    # Before
java                1728.092   184904.000    # After
go                    10.259     4496.000
rust                6615.044     4500.000
python3               50.873     5640.000
```

~~This also fixes the case where `maxsplits` is a negative value. According to the [spec](https://github.com/bazelbuild/starlark/blob/master/spec.md#stringreplace), a negative value for the number of splits should be ignored and this is also what python and the Go implementation do.~~
```python
# Before
"abc".replace("a", "z", -1)    # "abc"
# After
"abc".replace("a", "z", -1)    # "zbc"
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -286,9 +286,23 @@ public final class StringModule {
   public String replace(
       String self, String oldString, String newString, Object maxSplitO)
       throws EvalException {
-    int maxSplit = -1;
-    if (maxSplitO != Runtime.NONE && (Integer) maxSplitO >= 0) {
-      maxSplit = (Integer) maxSplitO;
+    int maxSplit = Integer.MAX_VALUE;
+    if (maxSplitO != Runtime.NONE) {
+      maxSplit = Math.max(0, (Integer) maxSplitO);
+    }
+    if (oldString.equals("")) {
+      StringBuilder sb = new StringBuilder();
+      for (int i = 0; i < self.length(); i++) {
+        if (maxSplit > 0) {
+          sb.append(newString);
+          maxSplit--;
+        }
+        sb.append(self.charAt(i));
+      }
+      if (maxSplit > 0) {
+        sb.append(newString);
+      }
+      return sb.toString();
     }
     return StringUtils.replace(self, oldString, newString, maxSplit);
   }
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -32,7 +32,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import org.apache.commons.lang.StringUtils;
 
 /**
  * Skylark String module.
@@ -290,21 +289,25 @@ public final class StringModule {
     if (maxSplitO != Runtime.NONE) {
       maxSplit = Math.max(0, (Integer) maxSplitO);
     }
-    if (oldString.equals("")) {
-      StringBuilder sb = new StringBuilder();
-      for (int i = 0; i < self.length(); i++) {
-        if (maxSplit > 0) {
-          sb.append(newString);
-          maxSplit--;
-        }
-        sb.append(self.charAt(i));
-      }
-      if (maxSplit > 0) {
+    StringBuilder sb = new StringBuilder();
+    int start = 0;
+    for (int i = 0; i < maxSplit; i++) {
+      if (oldString.isEmpty()) {
         sb.append(newString);
+        if (start < self.length()) {
+          sb.append(self.charAt(start++));
+        } else {
+          break;
+        }
+      } else {
+        int end = self.indexOf(oldString, start);
+        if (end < 0) break;
+        sb.append(self, start, end).append(newString);
+        start = end + oldString.length();
       }
-      return sb.toString();
     }
-    return StringUtils.replace(self, oldString, newString, maxSplit);
+    sb.append(self, start, self.length());
+    return sb.toString();
   }
 
   @SkylarkCallable(
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -32,6 +32,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.apache.commons.lang.StringUtils;
 
 /**
  * Skylark String module.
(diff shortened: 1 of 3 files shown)


laurentlb:
cc @alandonovan, @meisterT

laurentlb:
Does your time column include the startup time? If so, it's hard to make language comparisons. You could check the time to run a "Hello World" and subtract that.

meisterT:
> Does your time column include the startup time? If so, it's hard to make language comparisons. You could check the time to run a "Hello World" and subtract that.

Yes, it does include the startup time, see also my comments on https://github.com/bazelbuild/starlark/pull/75

Quarz0:
> Does your time column include the startup time? If so, it's hard to make language comparisons. You could check the time to run a "Hello World" and subtract that.

I guess subtracting the startup time is a good idea, but I thought in this case, just comparing the before and after performance of Java alone would be a good indicator of performance improvement, regardless of comparing Java with other implementations.

laurentlb:
Yes, the first two lines show a very nice performance improvement, thanks!
@meisterT will run additional benchmarks (in our codebase) to measure the impact.

Can you separate the performance improvement from the semantic change (negative number of splits)? We might want an incompatible flag for that, even if a breakage is unlikely.

Quarz0:
> Can you separate the performance improvement from the semantic change (negative number of splits)? We might want an incompatible flag for that, even if a breakage is unlikely.

Done. I also fixed the case when `oldString` is the empty string which was incorrectly handled in the previous commit. There shouldn't be any changes in semantics now.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -281,25 +282,29 @@ public String strip(String self, Object charsOrNone) {
...
+      if (maxSplit > 0) {
+        sb.append(newString);
       }
-      m.appendTail(sb);
-    } catch (IllegalStateException e) {
-      throw new EvalException(loc, e.getMessage() + " in call to replace");
+      return sb.toString();
     }
-    return sb.toString();
+    return StringUtils.replace(self, oldString, newString, maxSplit);
```
alandonovan: The replace function needed here is pretty easy to write; I don't think it warrants another dependency:

```
        int start = 0;
        for (int i = 0; i < maxSplit; i++) {
            int end = self.indexOf(old, start);
            if (end < 0) break;
            sb.append(self, start, end).append(newString);
            start = end + old.length();
        }
        sb.append(self, start, self.length());
```

Quarz0: I was actually planning to use `StringUtils` in other methods too like `count` and `split` to improve their performance as well.

alandonovan: To be clear: the performance improvement in this change comes not from anything clever in the StringUtils package, but from avoiding the regexp package, which has substantial overheads. By all means, avoid the regexp package in count and split too, but I don't think there's any need for an additional dependency to do that. Both functions can be implemented as a simple loop around indexOf. For example:
```
    List<String> split(String str, String sep, int max) {
        List<String> res = new ArrayList<>();
        int n = 0, i = 0;
        for (;;) {
            int j = str.indexOf(sep, i);
            if (j < 0 || ++n == max) {
                res.add(str.substring(i));
                break;
            }
            res.add(str.substring(i, j));
            i = j + sep.length();
        }
        return res;
    }
```

src/test/starlark/testdata/string_misc.sky
```diff
@@ -41,6 +41,18 @@ assert_eq('b$()n$()n$()'.replace('$()', '$($())'), "b$($())n$($())n$($())")
 assert_eq('b\\n\\n\\'.replace('\\', '$()'), "b$()n$()n$()")
 
 assert_eq('banana'.replace('a', 'e', 2), "benena")
+assert_eq('banana'.replace('a', 'e', 0), "banana")
+assert_eq('banana'.replace('a', 'e', -1), "banana")
```
alandonovan: Is this right? I would expect benene.

src/test/starlark/testdata/string_misc.sky
```diff
@@ -41,6 +41,18 @@ assert_eq('b$()n$()n$()'.replace('$()', '$($())'), "b$($())n$($())n$($())")
...
+assert_eq('banana'.replace('a', 'e', -10), "banana")
```
alandonovan: same here

src/test/starlark/testdata/string_misc.sky
```diff
@@ -41,6 +41,18 @@ assert_eq('b$()n$()n$()'.replace('$()', '$($())'), "b$($())n$($())n$($())")
...
+
+assert_eq('banana'.replace('', '-'), "-b-a-n-a-n-a-")
+assert_eq('banana'.replace('', '-', -2), "banana")
```
alandonovan: same here

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -281,24 +281,32 @@ public String strip(String self, Object charsOrNone) {
...
-      for (int i = 0; i < maxSplit && m.find(); i++) {
-        m.appendReplacement(sb, Matcher.quoteReplacement(newString));
+    int maxSplit = Integer.MAX_VALUE;
+    if (maxSplitO != Runtime.NONE) {
+      maxSplit = Math.max(0, (Integer) maxSplitO);
+    }
+    StringBuilder sb = new StringBuilder();
+    int start = 0;
+    for (int i = 0; i < maxSplit; i++) {
+      if (oldString.length() == 0) {
```
alandonovan: oldString.isEmpty()

src/test/starlark/testdata/string_misc.sky
Quarz0: Yes, that's a bug in the current implementation, but as Laurent mentioned above that we'd better not include semantic changes in this PR.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
Quarz0: Done

src/test/starlark/testdata/string_misc.sky
alandonovan: Ah, so he did.

In general if you identify a bug but don't intend to fix it in the same PR, it's good practice to at least identify the problem with comments (both at the logic and at the faulty assertions). But let's not hold up this PR any further. LGTM

PR #9132, opened 2019-08-09
Add @rules_java to BUILD.tools files.

iirina:


jin:
Does this mean that all Android builds with Bazel 1.0 will require @rules_java? Or will @rules_java be provided with default WORKSPACE suffix?

aiuto:
Ping.  I'm giving this a week before I close it unless I hear activity.

PR #9136, opened 2019-08-09
Allow patch arguments to be specified as function parameters

asuffield:
There are many scenarios where the patch function is useful, other
than when directly applying a single set of patches from the invoking
rule. This allows the caller to specify the patch parameters directly
from starlark.

### DIFF
--- a/tools/build_defs/repo/utils.bzl
+++ b/tools/build_defs/repo/utils.bzl
@@ -66,41 +66,62 @@ def workspace_and_buildfile(ctx):
 def _is_windows(ctx):
     return ctx.os.name.lower().find("windows") != -1
 
-def _use_native_patch(ctx):
+def _use_native_patch(patch_tool, patch_args):
     """If patch_tool is empty and patch_args only contains -p<NUM> options, we use the native patch implementation."""
-    if ctx.attr.patch_tool:
+    if patch_tool:
         return False
-    for arg in ctx.attr.patch_args:
+    for arg in patch_args:
         if not arg.startswith("-p"):
             return False
     return True
 
-def patch(ctx):
+def patch(ctx, patches=None, patch_cmds=None, patch_cmds_win=None, patch_tool=None, patch_args=None):
     """Implementation of patching an already extracted repository.
 
-    This rule is inteded to be used in the implementation function of a
-    repository rule. It assumes that the parameters `patches`, `patch_tool`,
-    `patch_args`, `patch_cmds` and `patch_cmds_win` are present in `ctx.attr`.
+    This rule is inteded to be used in the implementation function of
+    a repository rule. Ifthe parameters `patches`, `patch_tool`,
+    `patch_args`, `patch_cmds` and `patch_cmds_win` are not specified
+    then they are taken from `ctx.attr`.
 
     Args:
       ctx: The repository context of the repository rule calling this utility
         function.
+      patches: The patch files to apply. List of strings, Labels, or paths.
+      patch_cmds: Bash commands to run for patching, passed one at a
+        time to bash -c. List of strings
+      patch_cmds_win: Powershell commands to run for patching, passed
+        one at a time to powershell /c. List of strings.
+      patch_tool: Path of the patch tool to execute for applying
+        patches. String.
+      patch_args: Arguments to pass to the patch tool. List of strings.
+
     """
     bash_exe = ctx.os.environ["BAZEL_SH"] if "BAZEL_SH" in ctx.os.environ else "bash"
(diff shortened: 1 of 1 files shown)


asuffield:
ping

jin:
What's the use case here? Are you writing repository rules that use this function but don't have the attrs declared?

asuffield:
Yes, two that have come up so far:

- writing repository rules that download and patch things, using a patch owned by the rule, so the caller doesn't know or care that there's patching involved
- writing rules which have more than one patching step; consider something like go_repository, which would benefit from being able to patch before and after running gazelle, but then needs to manage the attributes itself

asuffield:
Ping, would be nice to get this in for 1.0

jin:
@aehlig, could you ptal at this?

tools/build_defs/repo/utils.bzl
```diff
@@ -66,64 +66,85 @@ def workspace_and_buildfile(ctx):
...
+        patches = ctx.attr.patches
+    if patch_cmds == None and hasattr(ctx.attr, "patch_cmds"):
+        patch_cmds = ctx.attr.patch_cmds
+    if patch_cmds_win == None and hasattr(ctx.attr, "patch_cmds_win"):
+        patch_cmds_win = ctx.attr.patch_cmds_win
+    if patch_tool == None and hasattr(ctx.attr, "patch_tool"):
+        patch_tool = ctx.attr.patch_tool
+    if patch_args == None and hasattr(ctx.attr, "patch_args"):
+        patch_args = ctx.attr.patch_args
+
```
aehlig: As now the scope of the `patch` function changes&mdash;from being used only for rules we know have all the `ctx.attr` set to a general-purpose patching utility&mdash;should we then start providing useful defaults here, e.g., by adding

```
if patches == None:
  patches = []
if patch_cmds == None:
  patch_cmds = []
...
```
here?

aehlig: > @aehlig, could you ptal at this?

The change is backwards-compatible, so things are fine from a technical point of view. And, personally, I don't mind having some useful utility functions for writing repository rules bundled with bazel (even though, to be general-purpose useful for all contexts, the key-word arguments should all be optional, even if the context does not provide them; see my comment in the code).

However, I know that at least @laurentlb prefers utility functions to be collected in skylib; the reason `patch` is in bazel is a bootstrap issue: `patch` is needed for `http_archive` which is the standard way of pulling in skylib. I'm willing to accept the argument that other repository rules have the same bootstrap issue, so the tools needed for writing rules fetching Starlark libraries should all be bundled.

aehlig:
> @aehlig, could you ptal at this?

The hange is backwards-compatible, so things are fine from a technical point of view. And, personally, I don't mind having some useful utility functions for writing repository rules bundled with bazel (even though, to be general-purpose useful for all contexts, the key-word arguments should all be optional, even if the context does not provide them; see my comment in the code).

However, I know that at least @laurentlb prefers utility functions to be collected in skylib; the reason patch is in bazel is a bootstrap issue: patch is needed for `http_archive` which is the standard way of pulling in skylib. I'm willing to accept the argument that other repository rules have the same bootstrap issue, so the tools needed for writing rules fetching Starlark libraries should all be bundled.

tools/build_defs/repo/utils.bzl
asuffield: Makes sense. I reworked the default values a bit, which involved shuffling some of the other code around, but I think it all does the same thing now. PTAL.

asuffield:
Yes, that matches my thinking about putting it here - I have a custom variation of http_archive that I use to fetch skylib (dealing with the tedious problem of operating without direct internet access). It's not the one which motivated me to construct this change, but I wouldn't ever want this code to exist twice, so this is the only place it can reasonably live.

tools/build_defs/repo/utils.bzl
aehlig: Thanks for fixing.

aehlig:
> LGTM. Thanks. Will import.

Actually, I must have missed something, as is witnessed by a [failing test](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/a94b983a-db57-4a6c-b9d2-9e6ee6a3e757/src/test/shell/bazel/external_patching_test/attempt_1.log).

asuffield:
I bet it's patch_tool == "". Pushed a fix, let's see if it passes.

aehlig:
> I bet it's patch_tool == "". Pushed a fix, let's see if it passes.

Yes, that's it. Thanks.

PR #6810, opened 2018-11-30
remote: add aws auth for remote caching

buchgr:

PR #4889, opened 2018-03-20
Initial attempt at direct s3 remote cache

GregBowyer:
This is an initial attempt to support #4844 

It directly supports S3 as a simple remote cache. After spending some time looking at how AWS credentials are done it looks like its easier to use the SDK, instead of hand rolling the auth directly.

Since the SDK provides the backbone for auth support, it makes sense to include the S3 parts of the sdk as well.

In this PR we use the S3 SDK directly to provide S3 remote cache support. This support should be on par with the google datastore support, and has similar options in its usage.

Thoughts?

# TODO
- [x] Remove requirement on S3 SDK
- [x] Remove external jar requirements
- [x] Chase down SSL failures (not related to S3 I think) (was #4944)
- [x] Chase down connection failures (not related to S3 I think) (was #4944)
- [x] Make work correctly with S3-Accelerator
- [x] Make Region crash more user friendly 
- [x] Real world test
- [x] Documentation 
- [x] Allow for credential profiles to be configured directly
- [x] I want to take a bit more work at `AwsHttpAuthAdapter` as its constructor is messy for testing
- [x] Unit tests for changes
- [x] EC2 instance credentials test
- [x] Figure out how to support security tokens
- [x] Tidy up command line options (could use some help here as to what the bazel team want the options to look like)
- [x] Replace the ThreadLocal date formatters with `javax.time` ones
- [x] Stare at, and maybe profile the Aws auth code for inefficiencies
- [x] Make sure the remote documentation makes sense for the option
- [x] Solve the terrible, slow, horrible and unstable downloads ~~~this patch~~~ (was #4944) causes

PR #9137, opened 2019-08-09
Initial attempt at direct s3 remote cache (with AuthHeadersProvider)

borkaehw:
This is the split-up of #4889 with AuthHeadersProvider. `AuthHeaderRequest.java` is required in this PR but the new file is created in https://github.com/bazelbuild/bazel/pull/9138. It needs https://github.com/bazelbuild/bazel/pull/9138 to get merged first or we could move `AuthHeaderRequest.java` to here.
___
Removes the directy dependency of the remote and buildeventservice
packages on the authentication libraries by using the new AuthHeadersProvider.

The authandtls package is split into the "authencation/google" and
"grpc" package.

This is a step towards merging #6810.

This change is a no-op.


borkaehw:
Close this based on discussion in https://github.com/bazelbuild/bazel/pull/4889

Issue #7847, opened 2019-03-26
"Server terminated abruptly" due to java.lang.IllegalStateException on invalid flag value


jmillikin-stripe:
### Description of the problem / feature request:

`bazel build --experimental_generate_json_trace_profile=$PWD/trace.json` causes the Bazel server to crash. This incorrect flag value should be reported to the user as such, instead of crashing the server.

Stack trace:

```
java.lang.IllegalStateException: Unable to set option '--experimental_generate_json_trace_profile' to value 'null'.
        at com.google.devtools.common.options.OptionsParserImpl.getParsedOptions(OptionsParserImpl.java:597)
        at com.google.devtools.common.options.OptionsParser.getOptions(OptionsParser.java:736)
        at com.google.devtools.build.lib.runtime.CommandEnvironment.<init>(CommandEnvironment.java:176)
        at com.google.devtools.build.lib.runtime.BlazeWorkspace.initCommand(BlazeWorkspace.java:204)
        at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.execExclusively(BlazeCommandDispatcher.java:284)
        at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.exec(BlazeCommandDispatcher.java:206)
        at com.google.devtools.build.lib.server.GrpcServerImpl.executeCommand(GrpcServerImpl.java:749)
        at com.google.devtools.build.lib.server.GrpcServerImpl.access$1600(GrpcServerImpl.java:103)
        at com.google.devtools.build.lib.server.GrpcServerImpl$2.lambda$run$0(GrpcServerImpl.java:818)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.IllegalArgumentException: Can not set boolean field com.google.devtools.build.lib.runtime.CommonCommandOptions.enableTracer to null value
        at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(Unknown Source)
        at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(Unknown Source)
        at java.base/jdk.internal.reflect.UnsafeBooleanFieldAccessorImpl.set(Unknown Source)
[...]

PR #9145, opened 2019-08-12
Bazel server won't crash for bad flag value

laszlocsomor:
Two of the first things the Bazel server does in
BlazeCommandDispatcher.execExclusively is parsing
command line options and initializing modules.
(Modules are Bazel-internal components.)

Parsing the flags consists of two phases:

1. Tokenizing the flags and storing the raw values
   as strings. This step is done eagerly.

2. Parsing values out of the strings and assigning
   them to typed class members. This step is done
   lazily when something asks the OptionsParser to
   return a specific Options instance.

Bazel catches bad flag values in step 1 already
and stores an error code, but it carries on with
initializing all modules. Bazel tries to
initialize all modules, even if some of them fail
to do so, and only then will Bazel observe the
error code.

This is the problem though: if a module that's
being initialized tries to read an option that got
an invalid value (like SandboxModule trying to
read SandboxOptions.sandboxDebug, but
"--sandbox_debug" got "/foo" as a value), then the
OptionsParser will fail to parse the right value
and fail to assign it to the sandboxDebug field,
and Bazel crashes.

This PR makes Bazel silently ignore the failure to
set the value. This is fine: Bazel still reports
the bad flag, but it also leaves the field at its
default value.

Fixes https://github.com/bazelbuild/bazel/issues/7847

Change-Id: I5b20332af57eaad8106018412ab06d4507468736


src/main/java/com/google/devtools/common/options/OptionsParserImpl.java
```diff
@@ -621,8 +621,8 @@ private boolean shouldIgnoreOption(OptionDefinition optionDefinition) {
       try {
         optionDefinition.getField().set(optionsInstance, value);
       } catch (IllegalArgumentException e) {
-        throw new IllegalStateException(
-            String.format("Unable to set %s to value '%s'.", optionDefinition, value), e);
+        // May happen when a boolean option got a string value. Just ignore this error without
```
meteorcloudy: Just curious, how does Bazel report wrong value of the flag?

laszlocsomor: It says:
```
bazel build --verbose_failures=/foo
(...)
ERROR: While parsing option --verbose_failures=/foo: '/foo' is not a boolean
```

meteorcloudy: Perfect, thanks!

PR #9146, opened 2019-08-12
Improve performance of string methods count and split

Quarz0:
The following 2 benchmark results show the performance improvements of this change:

- `string.count` [benchmark](https://github.com/Quarz0/starlark/blob/benchmark_runner/benchmark_suite/data/string_count.star):
```
=== data/string_count.star ===
          Time (MS)  Memory (KB)
java       8802.588   756580.000    # before
java       4831.417    99908.000    # after
go          702.808     8320.000
python3    2298.714     7560.000
```

- `string.split/rsplit` [benchmark](https://github.com/Quarz0/starlark/blob/benchmark_runner/benchmark_suite/data/string_split.star):
```
=== data/string_split.star ===
          Time (MS)  Memory (KB)
java      10451.067   746772.000    # before
java       4317.741   742340.000    # after
go         8855.651    11380.000
python3    1478.717    11852.000
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -27,11 +27,11 @@ import com.google.devtools.build.lib.skylarkinterface.SkylarkModuleCategory;
 import com.google.devtools.build.lib.syntax.SkylarkList.MutableList;
 import com.google.devtools.build.lib.syntax.SkylarkList.Tuple;
 import com.google.devtools.build.lib.syntax.Type.ConversionException;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.Collections;
 
 /**
  * Skylark String module.
@@ -336,7 +336,7 @@ public final class StringModule {
     if (maxSplitO != Runtime.NONE) {
       maxSplit = (Integer) maxSplitO;
     }
-    List<String> res = new ArrayList<>();
+    ArrayList<String> res = new ArrayList<>();
     int start = 0;
     while (true) {
       int end = self.indexOf(sep, start);
@@ -347,7 +347,7 @@ public final class StringModule {
       res.add(self.substring(start, end));
       start = end + sep.length();
     }
-    return MutableList.copyOf(env, res);
+    return MutableList.wrapUnsafe(env, res);
   }
 
   @SkylarkCallable(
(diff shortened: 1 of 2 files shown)


laurentlb:
cc @alandonovan @meisterT

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -332,12 +332,22 @@ public String replace(
...
+    while (true) {
+      int end = self.indexOf(sep, start);
+      if (end < 0 || maxSplit-- == 0) {
+        res.add(self.substring(start));
+        break;
+      }
+      res.add(self.substring(start, end));
+      start = end + sep.length();
+    }
+    return MutableList.copyOf(env, res);
```
laurentlb: We can avoid this copy by declaring res as a MutableList.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -368,58 +378,25 @@ public String replace(
...
+      if (start < 0 || maxSplit-- == 0) {
+        res.addFirst(self.substring(0, Math.max(0, end + 1)));
+        break;
+      }
+      res.addFirst(self.substring(start + sep.length(), end + 1));
+      end = start - 1;
     }
-
-    return MutableList.copyOf(env, result);
+    return MutableList.copyOf(env, res);
```
laurentlb: same here: use a MutableList, and simply reverse it before returning the result.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -368,58 +378,25 @@ public String replace(
...
-      result.addFirst(current);
-
-      ++splitsSoFar;
-      remainingLength -= sepLen + current.length();
-    }
-
-    if (splitsSoFar == maxSplits && remainingLength >= 0)   {
-      result.addFirst(input.substring(0, remainingLength));
+    ArrayDeque<String> res = new ArrayDeque<>();
+    int end = self.length() - 1;
```
laurentlb: Comment from Alan:
> The arithmetic is easier if you use half-open intervals. Use end = self.length() here, pass end-1 to lastIndexOf, and set end=start end the end of the loop body. Then the max expression is no longer needed because end can never be negative.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
Quarz0: But MutableList does not support `set`, so reversing it will still require a copy.
MutableList also does not support `add`, it supports `append` instead which needs `loc` and `env` arguments.

alandonovan: Quite; I forgot that MutableList doesn't support any of the standard update operations and defines its own variants. wrapUnsafe looks like a good compromise.

PR #9154, opened 2019-08-13
Replace wrapped_ar on macOS with libtool

keith:
Previously there were some inconsistencies with which tool was used to
link in different scenarios. If you had the Xcode command line tools,
but not Xcode, `libtool` would be used via `unix_cc_configure.bzl`. Also
if you were linking for Objective-C, not C++, libtool would be used.
This C++ linking case was the only place this was used, and it can
easily be replaced by the libtool wrapper so all linking on macOS uses
libtool.


sergiocampama:
@trybka can you take a look at this one?

jin:
Importing.

Issue #9155, opened 2019-08-13
Building bazel Docs missing install info


antofthy:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

> Looking at the page...
>    https://docs.bazel.build/versions/master/install-compile-source.html
> 
> It seems relatively straight forward to build a version of bazel (for the current machine) using the downloaded bazel binary  to build itself.
> 
> So I did as the instructions said..
> ```
>    bazel build //src:bazel
> ```
> All is well.. except   **Where is the built binary?** and **How do I Install it?**
> 
> There is just no information on this, making the whole exercise futile!
> 

### Feature requests: what underlying problem are you trying to solve with this feature?

> Original problem.. building "tenserflow" producing "`this rule is missing dependency declarations`" errors, that nothing seems to resolve!

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

> Replace this line with your answer.

### What operating system are you running Bazel on?

> Red Hat Enterprise Linux Server release 7.6 (Maipo)

### What's the output of `bazel info release`?

> release 0.26.1

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

> Replace this line with your answer.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

> No GIT!
> Installed binary   bazel-0.26.1-installer-linux-x86_64.sh
> Compiling from source tar file from same version.

###  Have you found anything relevant by searching the web?

> Replace these lines with your answer.
>
> Places to look:
> - StackOverflow: http://stackoverflow.com/questions/tagged/bazel
> - GitHub issues: https://github.com/bazelbuild/bazel/issues
> - email threads on https://groups.google.com/forum/#!forum/bazel-discuss

### Any other information, logs, or outputs that you want to share?

> Replace these lines with your answer.
>
> If the files are large, upload as attachment or provide link.

PR #9169, opened 2019-08-14
Docs: explain building bazel with bazel

laszlocsomor:
Fixes https://github.com/bazelbuild/bazel/issues/9155

Change-Id: I909c5ca359ad89243667ca3f1dc09c9913ff8ce3

### DIFF
--- a/site/docs/install-compile-source.md
+++ b/site/docs/install-compile-source.md
@@ -14,22 +14,138 @@ To build Bazel from source, you can do one of the following:
 
 <h2 id="build-bazel-using-bazel">Build Bazel using Bazel</h2>
 
-If you already have a Bazel binary, you can build Bazel from a GitHub checkout.
+TL;DR:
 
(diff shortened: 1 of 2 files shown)


site/docs/install-compile-source.md
```diff
@@ -14,22 +14,138 @@ To build Bazel from source, you can do one of the following:
...
+
+2.  `cd` into the directory where you extracted (or cloned) Bazel's sources.
+
+    For example if you extracted the sources under your home directory, run:
+
+        cd %USERPROFILE%\bazel-src
+
+3.  Build Bazel from source:
+
+        bazel build //src:bazel.exe
```
jin: Do we want to recommend //src:bazel-dev.exe instead?

laszlocsomor: Good idea! Yes.
Will do in internal review.

PR #9173, opened 2019-08-14
Remove contains() and containsActionResult from SimpleBlobStore

buchgr:
contains() and containsActionResult() only make sense for the OnDiskBlobStore
because they can be implemented with a (quickly returning) stat() call. They are
not generally useful for remote caching/execution. Thus this change
moves them down to the OnDiskBlobStore.

In the same breath I removed any non-test uses of the InMemoryBlobStore
and moved the class to the test package. The InMemoryBlobStore has never
been used by Bazel itself but only by its tests and the LRE. The LRE
usage has never really made sense either and so I removed it from there
as well.


buchgr:
FYI @borkaehw

PR #9175, opened 2019-08-14
Add sandbox_add_mount_pair support to docker sandbox

mukerjee:
The flag --sandbox_add_mount_pair allows multiple 'source:target' pairs to be
added to the Linux sandbox. The docker sandbox should support this as well.


mukerjee:
@googlebot I signed it!

Issue #9084, opened 2019-08-06
enumerate, join, extend functions not accepting dicts


Quarz0:
### Description of the problem:

When trying to pass a `dict` object to any of the Starlark functions `enumerate`, `string.join` or `list.extend`, the error `expected value of type 'sequence' for parameter ...` is thrown.

According to the spec:
- A `dict` is both an `Iterable` and a `Sequence`. (see [sequence types](https://github.com/bazelbuild/starlark/blob/master/spec.md#sequence-types))
- The arguments of [`enumerate`](https://github.com/bazelbuild/starlark/blob/master/spec.md#enumerate), [`extend`](https://github.com/bazelbuild/starlark/blob/master/spec.md#listextend) and [`join`](https://github.com/bazelbuild/starlark/blob/master/spec.md#stringjoin) are iterable sequences.

Also, both Python and the Go implementation accept dictionaries being passed to the above functions.

PR #9179, opened 2019-08-15
Allow enumerate, join, extend functions to accept dicts

Quarz0:
Fixes: #9084

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/SkylarkList.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/SkylarkList.java
@@ -28,6 +28,7 @@ import com.google.devtools.build.lib.skylarkinterface.StarlarkContext;
 import com.google.devtools.build.lib.syntax.StarlarkMutable.BaseMutableList;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Collection;
 import java.util.List;
 import java.util.RandomAccess;
 import javax.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/syntax/SkylarkList.java
```diff
@@ -521,15 +522,15 @@ public void set(int index, E value, Location loc, Mutability mutability) throws
...
       },
       useLocation = true,
       useEnvironment = true
     )
     public Runtime.NoneType extend(
-        SkylarkList<E> items, Location loc, Environment env)
+        Object items, Location loc, Environment env)
         throws EvalException {
-      addAll(items, loc, env.mutability());
+      addAll((Collection<? extends E>) EvalUtils.toCollection(items, loc, env), loc, env.mutability());
```
alandonovan: That cast isn't sound, and it should result in an "unchecked" warning. Do you know how (and why) these are suppressed in the Bazel codebase? They should really be treated as errors.

Quarz0: I am not sure actually why they are suppressed. I removed the cast anyway.

src/main/java/com/google/devtools/build/lib/syntax/SkylarkList.java
```diff
@@ -521,15 +522,15 @@ public void set(int index, E value, Location loc, Mutability mutability) throws
...
       },
       useLocation = true,
       useEnvironment = true
     )
     public Runtime.NoneType extend(
-        SkylarkList<E> items, Location loc, Environment env)
+        Object items, Location loc, Environment env)
         throws EvalException {
-      addAll(items, loc, env.mutability());
+      addAll(new ArrayList(EvalUtils.toCollection(items, loc, env)), loc, env.mutability());
```
alandonovan: This incurs an unnecessary copy that was not present in the old code. Essentially the problem has been pushed to 'new ArrayList', which has a raw type, which ought to be a compile error if our build was correctly configured. Let's revert to the previous version.

(I think the right solution is to abolish the type parameter in StarlarkList, so that it extends `Collection<Object>`. If you pass a `StarlarkList<X>` to the interpreter, you have no guarantee that it doesn't insert objects of class Y into the list. The only reason this doesn't happen today is that Blaze never passes mutable lists (other than of `<Object>)` to Starlark functions, but an application that did would start to see very confusing crashes. 
But that's a much bigger change.)

laurentlb: fyi, I've merged this PR using Friday's state (without the unnecessary copy, but with the cast).

Issue #9172, opened 2019-08-14
Windows: Bazel cannot share cpp cache between two projects with different workspace name


meteorcloudy:
If you try to share cache between two projects with different workspace name, when the cache is hit for a C++ compilation, the header check will fail because the header paths printed by MSVC's `/showInclude` option are absolute paths, which contains the workspace name.

You can reproduce with:
https://github.com/meteorcloudy/my_tests/tree/master/cpp_cache_with_different_workspace_name

```
$ cd A
$ bazel build --disk_cache=C:/src/tmp/bazel_disk_cache :hello
$ cd ../B
$ bazel build --disk_cache=C:/src/tmp/bazel_disk_cache :hello
Starting local Bazel server and connecting to it...
INFO: Invocation ID: e98dabd3-b255-4c34-86e5-ab5070b346f3
INFO: Analyzed target //:hello (13 packages loaded, 72 targets configured).
INFO: Found 1 target...
ERROR: C:/tools/msys64/home/pcloudy/workspace/my_tests/cpp_cache_with_different_workspace_name/b/BUILD:1:1: undeclared inclusion(s) in rule '//:hello':
this rule is missing dependency declarations for the following files included by 'hello.cc':
  'C:/users/pcloudy/_bazel_pcloudy/7tppxn5m/execroot/a/hello.h'
Target //:hello failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 10.570s, Critical Path: 0.09s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

Note the error message when we build //:hello in B,
**'C:/users/pcloudy/_bazel_pcloudy/7tppxn5m/execroot/a/hello.h'** is from project A.

This is the root cause of https://buildkite.com/bazel/bazel-federation/builds/19#ee3795da-1088-4c27-b81e-2636e30e3587

FYI @fweikert @laszlocsomor

meteorcloudy:
We had a similar bug before because output base was different. https://github.com/bazelbuild/bazel/issues/6847#issuecomment-446966864
That issue was fixed at https://github.com/bazelbuild/bazel/commit/a1aa5c14782b38d8edf86953230f76ba2662378d, but it doesn't take different workspace names into consideration.

laszlocsomor:
Thanks for the report!

> So the question is, should we allow sharing cache between two projects with different workspace name?

Why wouldn't we?

meteorcloudy:
[...]

PR #9188, opened 2019-08-16
Windows: Ignore workspace name differences while doing header checking

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel/issues/9172

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/rules/cpp/ShowIncludesFilterTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/cpp/ShowIncludesFilterTest.java
@@ -38,7 +38,7 @@ public class ShowIncludesFilterTest {
 
   @Before
   public void setUpOutputStreams() throws IOException {
-    showIncludesFilter = new ShowIncludesFilter("foo.cpp", "__main__");
+    showIncludesFilter = new ShowIncludesFilter("foo.cpp");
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/devtools/build/lib/rules/cpp/ShowIncludesFilterTest.java
```diff
@@ -93,17 +93,31 @@ public void testMatchAllOfNotePrefix() throws IOException {
...
     assertThat(output.toString()).isEmpty();
     filterOutputStream.write(getBytes("\n"));
     // It's a match, output should be filtered, dependency on bar.h should be found.
     assertThat(output.toString()).isEmpty();
-    assertThat(showIncludesFilter.getDependencies()).contains("bar.h");
+    assertThat(showIncludesFilter.getDependencies()).contains("foo\\bar\\bar.h");
+  }
+
+  @Test
+  public void testFindHeaderFromAbsolutePathOutsideExecroot() throws IOException {
```
laszlocsomor: In a comment, please reference the bug for which this is a regression test.

meteorcloudy: Thanks, done!

meteorcloudy:
Integration test added, PTAL~

PR #9191, opened 2019-08-16
Remove defaults from Apple device types and versions

keith:
These defaults are very prone to getting out of date, which makes these
fields detrimental to rule implementations. In the case of rules_apple
we'd like to make the tests default to these versions, if a version
isn't passed as part of the test target, unfortunately we cannot
differentiate between the user didn't pass a value, and these old
versions. Overall these defaults aren't very helpful because with each
Xcode update old versions no longer run on developer machines unless
they install the old simulator versions manually.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcCommandLineOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcCommandLineOptions.java
@@ -30,7 +30,7 @@ import java.util.List;
 public class ObjcCommandLineOptions extends FragmentOptions {
   @Option(
     name = "ios_simulator_version",
-    defaultValue = "9.3",
+    defaultValue = "null",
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
Sorry for being pedantic, what's the behavior with these `null` defaults for existing a new users?

keith:
If people are passing this flag now, there is no change, if they aren't passing this flag and running a `bazel test Something` then it would attempt to run on iOS 9.3 simulators, which isn't support by the newest versions of Xcode so it's unlikely anyone is doing that.

keith:
If people are passing this flag now, there is no change, if they aren't passing this flag and running a `bazel test Something` then it would attempt to run on iOS 9.3 simulators, which isn't support by the newest versions of Xcode so it's unlikely anyone is doing that.

jin:
Importing.

dstiers-1109:
Thanks

PR #9195, opened 2019-08-18
Add JMH benchmarking tool for Starlark

Quarz0:
Related: https://github.com/bazelbuild/starlark/pull/75

This is a benchmark tool for Starlark that uses OpenJDK JMH (Java Microbenchmark Harness).
The tool loads and executes Starlark files then looks up every method that starts with `bench_` and benchmarks each using JMH.

I used the files from the repo [jmh_rules](https://github.com/buchgr/rules_jmh) to get JMH to work. I only made a small change to [add `main_class`](https://github.com/Quarz0/bazel/blob/87f5fee7d3602b9050ba7aa78535fbde218585cd/third_party/rules_jmh/defs.bzl#L39) to `jmh_java_benchmarks`, to be able to overwrite its default value.

The `Bench.java` class might still need some refactoring or changing JMH configurations, but I thought an early feedback might be helpful.

Sample output:
```
$ bazel run //src/main/java/com/google/devtools/starlark:Bench_benchmarks/search_star
...
# JMH version: 1.21
# VM version: JDK 1.8.0_121, Java HotSpot(TM) 64-Bit Server VM, 25.121-b13
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/jre/bin/java
# VM options: <none>
# Warmup: 5 iterations, 5 s each
# Measurement: 5 iterations, 5 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: com.google.devtools.starlark.Bench.testBenchmark
# Parameters: (benchName = bench_binary_search, path = src/main/java/com/google/devtools/starlark/benchmarks/search.star)

# Run progress: 0.00% complete, ETA 00:03:20
# Fork: 1 of 2
# Warmup Iteration   1: 19.621 us/op
# Warmup Iteration   2: 16.848 us/op
# Warmup Iteration   3: 16.601 us/op
# Warmup Iteration   4: 16.760 us/op
# Warmup Iteration   5: 16.561 us/op
Iteration   1: 16.656 us/op
Iteration   2: 16.603 us/op
Iteration   3: 16.683 us/op
Iteration   4: 16.623 us/op
Iteration   5: 17.103 us/op

...

Result "com.google.devtools.starlark.Bench.testBenchmark":
  16.319 ±(99.9%) 0.693 us/op [Average]
  (min, avg, max) = (15.876, 16.319, 17.103), stdev = 0.459
  CI (99.9%): [15.626, 17.013] (assumes normal distribution)

...

# Run complete. Total time: 00:03:23

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

Benchmark                    (benchName)                                                             (path)  Mode  Cnt     Score    Error  Units
Bench.testBenchmark  bench_binary_search  src/main/java/com/google/devtools/starlark/benchmarks/search.star  avgt   10    16.319 ±  0.693  us/op
Bench.testBenchmark  bench_linear_search  src/main/java/com/google/devtools/starlark/benchmarks/search.star  avgt   10  5094.581 ± 56.421  us/op
```


buchgr:
You might want to mention that this code was copied from https://github.com/buchgr/rules_jmh.

Quarz0:
> You might want to mention that this code was copied from https://github.com/buchgr/rules_jmh.

Yes, sorry about that. I actually mentioned that at the beginning (see edit history), but then removed it when I tried to include it directly via a git_repository rule (while trying to fix the failing tests).

Quarz0:
cc @laurentlb @alandonovan

WORKSPACE
```diff
@@ -712,3 +712,16 @@ load("@local_config_winsdk//:toolchains.bzl", "register_local_rc_exe_toolchains"
 register_local_rc_exe_toolchains()
 
 register_toolchains("//src/main/res:empty_rc_toolchain")
+
+
+new_local_repository(
+    name = "rules_jmh",
+    path = "./third_party/rules_jmh",
+    build_file = "//third_party/rules_jmh:BUILD",
+    workspace_file = "//third_party/rules_jmh:rules_jmh.WORKSPACE",
```
jin: Consider adding some information about the snapshot of this vendored directory.

laurentlb:
@alandonovan, any opinion on this approach?

laurentlb:
Tests are failing

Quarz0:
> Tests are failing

Yes, the `bazel_bootstrap_distfile_test` is failing which I think has something to do with compiling the JMH library I included, but I couldn't figure out how to make this work or why is it failing.

laurentlb:
@aehlig, can you please help?

jin:
Looks like this PR is pulling the JMH jars using rules_jvm_external. That doesn't work with our compile script, which uses checked-in jars: https://github.com/bazelbuild/bazel/blob/80c8052fa443e08a05a64072659b483410f2f072/scripts/bootstrap/compile.sh#L25

You may have to vendor the JMH jars in a separate PR, java_import them and modify the rules to use the java_import targets.

Quarz0:
> You may have to vendor the JMH jars in a separate PR, java_import them and modify the rules to use the java_import targets. And, if they are a compile dependency for Bazel itself, you may need to modify compile.sh as well.

I tried adding the JMH jars and its dependencies to third_party, java_import them and getting rid of `rules_jvm_external`. The benchmark binary is still working but now the `bazel_bootstrap_distfile_test` keeps running forever and doesn't terminate.

jin:
@Quarz0 can you please update the diff where you switched rules_jvm_external to java_import and let the tests run?

Quarz0:
The test passed, thanks. I initially thought there might be something wrong because the test took too long when I added the jars compared to before, sorry for that.

jin:
Looks like this PR touches third_party and non-third_party code. Could you please split them up? We're not able to import hybrid third_party changes.

ulfjack:
We can split the PR on import into third_party and non-third_party changes. However, I have another comment: if you move the Bench.java out of the src/main tree, then you wouldn't need to vendor the jar files, because the bootstrap script wouldn't pick up the Bench.java (which does not need to be compiled into Bazel). I think.

aiuto:
cc @alandonovan

src/main/java/com/google/devtools/build/lib/syntax/FuncallExpression.java
```diff
@@ -236,7 +236,7 @@ public FuncallException(String msg) {
 
   private final int numPositionalArgs;
 
-  FuncallExpression(Expression function, ImmutableList<Argument.Passed> arguments) {
+  public FuncallExpression(Expression function, ImmutableList<Argument.Passed> arguments) {
```
alandonovan: The lib.syntax package has changed a lot in the last few months, so you'll need to sync up.

src/main/java/com/google/devtools/starlark/BUILD
```diff
@@ -25,6 +26,22 @@ java_binary(
     ],
 )
 
+[
+    jmh_java_benchmarks(
```
alandonovan: There's no need to create one build target per file in benchmarks/*. The Java code is 100% identical for all of them. Instead just pass the list of filenames to a single Java program and do a loop in Java.

src/main/java/com/google/devtools/starlark/BUILD
```diff
@@ -25,6 +26,22 @@ java_binary(
...
+        name = "Bench_" + file.replace('.', '_'),
+        srcs = ["Bench.java"],
+        main_class = "com.google.devtools.starlark.Bench",
+        deps = [
+            "//src/main/java/com/google/devtools/build/lib:events",
+            "//src/main/java/com/google/devtools/build/lib:packages-internal",
```
alandonovan: Not sure why this dependency is needed. Perhaps it no longer is?

src/main/java/com/google/devtools/starlark/benchmarks/search.star
```diff
@@ -0,0 +1,22 @@
+ar = [i for i in range(100000)]
```
alandonovan: array = list(range(10000)) or just range(10000)

src/main/java/com/google/devtools/starlark/benchmarks/search.star
```diff
@@ -0,0 +1,22 @@
...
+target = 33333
+
+def bench_binary_search():
+    st, en = 0, len(ar) - 1
+    for i in range(1000000000):
+        if st > en:
+            break
+        mid = (st + en) // 2
+        if ar[mid] == target:
```
alandonovan: compute ar[mid] once

src/main/java/com/google/devtools/starlark/benchmarks/search.star
```diff
@@ -0,0 +1,22 @@
+ar = [i for i in range(100000)]
+target = 33333
+
+def bench_binary_search():
+    st, en = 0, len(ar) - 1
+    for i in range(1000000000):
```
alandonovan: Is this a poor man's infinite loop?

third_party/rules_jmh/defs.bzl
```diff
@@ -0,0 +1,23 @@
+def jmh_java_benchmarks(name, srcs, deps=[], tags=[], plugins=[], main_class="org.openjdk.jmh.Main", **kwargs):
```
alandonovan: Within Google's code base, we have a rule called java_jmh_benchmarks which supports name, srcs, deps, and tags (though apparently not plugins yet). Could you use that name, to make it easier for us to converge the two sets of rules on a single open-source one?

src/main/java/com/google/devtools/starlark/benchmarks/search.star
```diff
@@ -0,0 +1,22 @@
+ar = [i for i in range(100000)]
+target = 33333
+
+def bench_binary_search():
+    st, en = 0, len(ar) - 1
```
alandonovan: start, end

Better: lo, hi

aiuto:
ping

laurentlb:
Closing for inactivity

PR #9200, opened 2019-08-19
Refactor SimpleBlobStore#get(ActionResult)(...)

buchgr:
The change update methods signatures to be compatible with those in
AbstractRemoteActionCache:
  1) Future<Boolean> get(String, OutputStream) -> Future<Void> downloadBlob(Digest, OutputStream)
  2) Future<Boolean> getActionResult(String, OutputStream) -> Future<ActionResult> downloadActionResult(ActionKey)

The refactoring uncovered a bug in CombinedDiskHttpBlobStore. We did not
close the output stream *before* moving a file to the disk cache, thus
potentially moving a corrupted file to the disk cache as any errors
would only be reported afterwards.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteCache.java
@@ -49,6 +49,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.google.devtools.build.lib.actions.ActionInput;
 import com.google.devtools.build.lib.concurrent.ThreadSafety.ThreadSafe;
 import com.google.devtools.build.lib.remote.RemoteRetrier.ProgressiveBackoff;
+import com.google.devtools.build.lib.remote.common.CacheNotFoundException;
 import com.google.devtools.build.lib.remote.common.SimpleBlobStore.ActionKey;
 import com.google.devtools.build.lib.remote.merkletree.MerkleTree;
 import com.google.devtools.build.lib.remote.options.RemoteOptions;
@@ -254,6 +255,10 @@ public class GrpcRemoteCache extends AbstractRemoteActionCache {
     uploader.uploadBlobs(inputsToUpload, /* forceUpload= */ true);
   }
 
+  private String digestToString(Digest digest) {
+    return digest.getHash() + "/" + digest.getSizeBytes();
+  }
+
   @Override
   protected ListenableFuture<Void> downloadBlob(Digest digest, OutputStream out) {
     if (digest.getSizeBytes() == 0) {
@@ -263,7 +268,7 @@ public class GrpcRemoteCache extends AbstractRemoteActionCache {
     if (!options.remoteInstanceName.isEmpty()) {
       resourceName += options.remoteInstanceName + "/";
     }
-    resourceName += "blobs/" + digestUtil.toString(digest);
+    resourceName += "blobs/" + digestToString(digest);
 
     @Nullable Supplier<HashCode> hashSupplier = null;
     if (options.remoteVerifyDownloads) {
@@ -348,7 +353,7 @@ public class GrpcRemoteCache extends AbstractRemoteActionCache {
               public void onError(Throwable t) {
                 Status status = Status.fromThrowable(t);
                 if (status.getCode() == Status.Code.NOT_FOUND) {
-                  future.setException(new CacheNotFoundException(digest, digestUtil));
+                  future.setException(new CacheNotFoundException(digest));
                 } else {
                   future.setException(t);
                 }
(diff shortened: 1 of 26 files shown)


src/main/java/com/google/devtools/build/lib/remote/common/CacheNotFoundException.java
```diff
@@ -25,8 +24,8 @@
 public final class CacheNotFoundException extends IOException {
   private final Digest missingDigest;
 
-  CacheNotFoundException(Digest missingDigest, DigestUtil digestUtil) {
-    super("Missing digest: " + digestUtil.toString(missingDigest));
+  public CacheNotFoundException(Digest missingDigest) {
+    super("Missing digest: " + missingDigest.getHash() + "/" + missingDigest.getSizeBytes());
```
artem-zinnatullin:
```suggestion
    super("Missing digest: " + DigestUtil.toString(missingDigest));
```

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteCache.java
```diff
@@ -268,7 +269,7 @@ public void ensureInputsPresent(
     if (!options.remoteInstanceName.isEmpty()) {
       resourceName += options.remoteInstanceName + "/";
     }
-    resourceName += "blobs/" + digestUtil.toString(digest);
+    resourceName += "blobs/" + digest.getHash() + "/" + digest.getSizeBytes();
```
artem-zinnatullin:
```suggestion
    resourceName += "blobs/" + DigestUtil.toString(digest);
```

src/test/java/com/google/devtools/build/lib/remote/SimpleBlobStoreActionCacheTest.java
```diff
@@ -430,4 +440,71 @@ public void testDownloadFailsOnDigestMismatch() {
             () -> getFromFuture(client.downloadFile(fs.getPath("/exec/root/foo"), digest)));
     assertThat(e).hasMessageThat().contains(digest.getHash());
   }
+
+  private class ConcurrentMapBlobStore implements SimpleBlobStore {
```
artem-zinnatullin: supernit: Maybe extract it into a separate file in test sources? Just to keep this file smaller :)

Could be package private if reusability is a concern

src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/OnDiskBlobStoreActionCache.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public OnDiskBlobStoreActionCache(
+      RemoteOptions options,
+      Path cacheDir,
+      DigestUtil digestUtil) {
+    super(options, new OnDiskBlobStore(cacheDir), digestUtil);
+  }
+
+  public boolean containsKey(Digest digest) {
+    return ((OnDiskBlobStore) blobStore).contains(digest);
```
artem-zinnatullin: Maybe add a `private final` field that's already casted instead of casting every time? 

I'm afraid casting every time will have runtime overhead before (if) JIT figures out that cast always succeeds, given how often `containsKey` will be called in large builds

Docs: [`checkcast`](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html)

borkaehw:
@buchgr Is it also related to making GrpcRemoteCache a SimpleBlobStore implementation?

buchgr:
@borkaehw yes

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteCache.java
buchgr: This doesn't work due to breakages due to dependency issues. Seems like a small price to pay for now.

src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/OnDiskBlobStoreActionCache.java
buchgr: Good point in general but this code is not performance critical as it's only used in tests.

src/main/java/com/google/devtools/build/lib/remote/common/CacheNotFoundException.java
buchgr: Same reason as above :\.

src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/OnDiskBlobStoreActionCache.java
ishikhman: Still, it could be a good idea to 'not' make tests run longer ;)

buchgr:
> Still, it could be a good idea to 'not' make tests run longer ;)

They won't. This is a micro optimization that _may_ shave off a few micro seconds.

SrodriguezO:
@buchgr what's the status on this PR?

ishikhman:
@SrodriguezO it's on a way to be merged, but @buchgr is away for a few weeks, so it will be finalised in a week or so

SrodriguezO:
@buchgr friendly ping : )

PR #9202, opened 2019-08-19
Add load statement for @rules_proto to //tools/jdk:BUILD.java_tools

Yannic:


Yannic:
I think the tests fail because we're a bit inconsistent with the version of rules_proto :(
I'll try to update them everywhere after https://github.com/bazelbuild/rules_proto/pull/17 lands.

Yannic:
Didn't intent to close this

iirina:
Ugh I don't like adding more things to the implicit suffix, but I don't see another way :(

src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE
```diff
@@ -262,4 +262,16 @@ maybe(
     strip_prefix = "rules_cc-0d5f3f2768c6ca2faca0079a997a97ce22997a0c",
 )
 
+# Needed only because of java_tools.
+maybe(
+    http_archive,
+    "rules_proto",
+    sha256 = "7d05492099a4359a6006d1b89284d34b76390c3b67d08e30840299b045838e2d",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/9cd4f8f1ede19d81c6d48910429fe96776e567b1.zip",
```
aiuto: Why is this a different revision from this; https://github.com/bazelbuild/bazel/blob/209175ff8ffeb05628ed8a187dd414a3d2935c55/WORKSPACE#L229

PR #9210, opened 2019-08-20
Stop lexer from scanning characters in ints if first token in not a zero

Quarz0:
If the first token of a number is not '0', then the lexer should not scan any characters because the number should be a decimal.
This solves some cases where omitting whitespace between tokens produced errors. For example
```
# Before
6in() # False
6or() # invalid base-10 integer constant: 6o

# After
6in() # False
6or() # 6
```
The first example works correctly in both cases since the token 'i' never advances the buffer, while 'o' can appear in integers.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/Lexer.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/Lexer.java
@@ -673,6 +673,8 @@ final class Lexer {
         case 'd': case 'D':
         case 'e': case 'E':
         case 'f': case 'F':
+          if (buffer[oldPos] != '0')
+            return bufferSlice(oldPos, pos);
         case '0': case '1':
         case '2': case '3':
         case '4': case '5':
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/syntax/Lexer.java
```diff
@@ -673,6 +673,8 @@ private String scanInteger() {
         case 'd': case 'D':
         case 'e': case 'E':
         case 'f': case 'F':
+          if (buffer[oldPos] != '0')
+            return bufferSlice(oldPos, pos);
```
laurentlb: Add comment to explain this.

nit, always use curly braces after `if`.

aiuto:
Can any of you who commented please add yourself as owner to shepherd in the change.

laurentlb:
This should be allowed: `0o1or()`

Quarz0:
> This should be allowed: `0o1or()`

Thanks, Fixed.

PR #9211, opened 2019-08-20
Improve error message about missing artifact with owner

irengrig:


src/main/java/com/google/devtools/build/lib/skyframe/ArtifactFunction.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014 The Bazel Authors. All rights reserved.
+// Copyright 2019 The Bazel Authors. All rights reserved.
```
aiuto: We should not be updating copyright notices. The date of first release remains corect.

jin:
@irengrig let's import this?

PR #9212, opened 2019-08-20
Enable unpacking of dicts into positional args

Quarz0:
```python
def varargs(*args): return args
x15 = {"one": 1}

# Before
varargs(*x15) # argument after * must be an iterable, not dict
# After
varargs(*x15) # ("one",)
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/FuncallExpression.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/FuncallExpression.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -841,12 +842,15 @@ public final class FuncallExpression extends Expression {
       if (arg.isPositional()) {
         posargs.add(value);
       } else if (arg.isStar()) { // expand the starArg
-        if (!(value instanceof Iterable)) {
+        Collection<?> items;
+        try {
+          items = EvalUtils.toCollection(value, getLocation(), env);
+        } catch (EvalException e) {
           throw new EvalException(
               getLocation(),
               "argument after * must be an iterable, not " + EvalUtils.getDataTypeName(value));
         }
-        for (Object starArgUnit : (Iterable<Object>) value) {
+        for (Object starArgUnit : items) {
           posargs.add(starArgUnit);
         }
       } else if (arg.isStarStar()) { // expand the kwargs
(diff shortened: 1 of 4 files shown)


laurentlb:
(cc @alandonovan)

src/main/java/com/google/devtools/build/lib/syntax/FuncallExpression.java
```diff
@@ -841,12 +842,15 @@ private void evalArguments(
...
+        Collection<?> items;
+        try {
+          items = EvalUtils.toCollection(value, getLocation(), env);
+        } catch (EvalException e) {
           throw new EvalException(
               getLocation(),
               "argument after * must be an iterable, not " + EvalUtils.getDataTypeName(value));
         }
-        for (Object starArgUnit : (Iterable<Object>) value) {
+        for (Object starArgUnit : items) {
```
alandonovan: posargs.addAll(items)

src/test/java/com/google/devtools/build/lib/syntax/SkylarkEvaluationTest.java
```diff
@@ -1251,6 +1251,19 @@ public void testJavaCallWithPositionalAndKwargs() throws Exception {
         .testLookup("b", "with_params(1, true, false, true, false, a, [\"a\", \"b\"])");
   }
 
+  @Test
+  public void testPositionalArgTypes() throws Exception {
```
alandonovan: Can we use Starlark for all new tests of the evaluator? The syntactic cost of Java quotation is far larger than the expression of interest, and every change to the test case forces a recompile.

src/main/java/com/google/devtools/build/lib/syntax/FuncallExpression.java
```diff
@@ -841,12 +842,15 @@ private void evalArguments(
       if (arg.isPositional()) {
         posargs.add(value);
       } else if (arg.isStar()) { // expand the starArg
-        if (!(value instanceof Iterable)) {
+        Collection<?> items;
+        try {
+          items = EvalUtils.toCollection(value, getLocation(), env);
```
alandonovan: toCollection allocates a copy when all we need is to iterate over the existing value without retaining it. Is there a variant of toCollection that returns an iterable? If not, it may be time to add one.

ulfjack:
This change does not apply cleanly anymore. Can you update?

meisterT:
What's the status of this PR?

alandonovan:
I think we can revert this. The bug appears to have been fixed by CL 282682290 "bazel syntax: eliminate unnecessary array copying" (how does one find the corresponding git commit?) but we could still use a test.

brandjon:
> how does one find the corresponding git commit?

Search for the CL number in the search box at the top left. It'll find the git commit with the CL number in its description (`PiperOrigin-RevId: NNNNNNNNN`). Make sure to click the Commits item on the left.

alandonovan:
Thanks!

PR #9217, opened 2019-08-21
Fix proguard target for default conditions.

### LINKED ISSUE #9116: Incorrect reference to `@remote_java_tools_linux//tools/jdk` in `@bazel_tools//tools/jdk:proguard`

When executing: `bazel query 'deps(//...)'`.

The following error occurs, both in 0.28 and 0.27 and 0.29 rcs. (it's fine in 0.26)

```
ERROR: /private/var/tmp/_bazel_gruber/3f8511af2a1bd2e817b9401d1ab7b770/external/bazel_tools/tools/jdk/BUILD:252:1: no such package '@remote_java_tools_linux//tools/jdk': BUILD file not found in directory 'tools/jdk' of external repository @remote_java_tools_linux and referenced by '@bazel_tools//tools/jdk:proguard'
```

digging in, the remote_java_tools_linux workspace is created, and includes lots of tools like singlejar, ijar, jacoco, and proguard, but there is no tools/jdk/BUILD in that workspace. There ***is*** a build file in the root, so `@remote_java_tools_linux//:proguard` would have worked, but the `@bazel_tools//tools/jdk:proguard` target is specified as:

```
alias(
    name = "proguard",
    actual = select({
        "//src/conditions:darwin": "@remote_java_tools_darwin//:proguard",
        "//src/conditions:darwin_x86_64": "@remote_java_tools_darwin//:proguard",
        "//src/conditions:windows": "@remote_java_tools_windows//:proguard",
        "//src/conditions:linux_x86_64": "@remote_java_tools_linux//:proguard",
        # On different platforms the linux repository can be used because the
        # proguard deploy jar is platform-agnostic.
        "//conditions:default": "@remote_java_tools_linux//tools/jdk:proguard",
    }),
)
```

This happens on both our CI machines (linux) and on my mac (darwin) which suggests that something about the platform information is off, causing it to drop to the conditional default - which is mis-specified.

Using the bazel `build`, `test` and `run` doesn't invoke this behavior, leading me to believe that `bazel query` is failing to properly set the platform information, causing the fall-through behavior, instead of hitting the proper platform conditional.  So there are two issues to fix here.  (a) the conditional default target needs to be corrected, as it refers to a non-existant packaget and (b) the query behavior needs to set the platform properly. 

1. Create a workspace
2. In 0.27.x and 0.28.x run `bazel query 'deps(@bazel_tools//tools/jdk:proguard)'`

0.29.0rc4
0.28.0
0.27.1

No.

No

iirina:
Fixes #9116.

### DIFF
--- a/tools/jdk/BUILD
+++ b/tools/jdk/BUILD
@@ -261,7 +261,7 @@ alias(
         "//src/conditions:linux_x86_64": "@remote_java_tools_linux//:proguard",
         # On different platforms the linux repository can be used because the
         # proguard deploy jar is platform-agnostic.
-        "//conditions:default": "@remote_java_tools_linux//tools/jdk:proguard",
+        "//conditions:default": "@remote_java_tools_linux//:proguard",
     }),
 )
 


iirina:
@oquenchil can you take a look?

PR #9220, opened 2019-08-22
Use rules_android in Android tutorial

artem-zinnatullin:
This change updates [Bazel Android Tutorial](https://docs.bazel.build/versions/master/tutorial/android-app.html) to use [rules_android](https://github.com/bazelbuild/rules_android) so that users aren't exposed to the deprecated native rules and are better prepared for the future use of [rules_android](https://github.com/bazelbuild/rules_android).

Caveat: I had to use commit instead of a tag for [rules_android](https://github.com/bazelbuild/rules_android) since it hasn't been released in a while even though it gets changed.


jin:
Thanks! Merging.

jin:
Closing this for now; we'll revisit this when the Starlark rules_android are ready to be recommended on the site.

Issue #8147, opened 2019-04-25
incompatible_restrict_named_params: make a number of params positional-only


c-parsons:
This flag makes a number of parameters of Starlark builtin functions positional-only.

**Flag:** --incompatible_restrict_named_params

**Motivation**
The affected Starlark parameters were originally intended to be positional-only, but, due to an infrastructure bug, they were able to be specified by keyword for several releases.

An example case is that, today, one can specify:

```
len(x = [])
```

But `len`'s one parameter should be positional-only.

See #5010 for original bug.

**Migration**
We will incorporate automatic migration using buildifier for affected parameters. Migration can also be achieved manually by specifying affected parameters positionally (look to the Bazel Starlark builtins docs for positional ordering, when in doubt!)

thomasvl:
Shouldn't the docs get updated to not use a name and have their description cover the name when describing behaviors?  i.e. - https://docs.bazel.build/versions/master/skylark/lib/globals.html#getattr, `default` appears to now only be positional, but given how it is documented, it really seems to call it out as something that is named instead.

c-parsons:
Yes, you are correct. Lets track this as part of this issue. (I will own that change)

dslomov:
Is migration tooling available?

c-parsons:
@vladmos IIRC we discussed writing buildifier warnings for this, but I'm difficulty finding these conversations (and therefore I can't recall if we did any work towards this), do you recall?

vladmos:
@c-parsons the actual pull request is https://github.com/bazelbuild/buildtools/pull/692 . There was a prototype in a different PR earlier that was closed by Github because I accidentally deleted the branch.

laurentlb:
@vladmos Can you provide instructions for migration?

vladmos:
The command is `buildifier --lint=fix --warnings=keyword-positional-params`, you'll need to either build Buildifier manually from head or wait for 0.29 which should be released this week.

Note that it will only fix (and warn about) the following top-level functions: https://github.com/vladmos/buildtools/blob/fe14be0e63262168393387d0f7c87250c44be5d2/warn/warn_bazel_api.go#L842-L856
Various methods such as `string.split` are not (yet) supported, and even if they were, buildifier can't guarantee to find all of their usages due to lack of static typing.

meteorcloudy:
rules_rust is failing in downstream due to the flip of this flag:
[...]

PR #9228, opened 2019-08-22
Handle negative count arg in string.replace

### LINKED ISSUE #9181: string.replace incorrectly handles negative value for count

According to the [spec](https://github.com/bazelbuild/starlark/blob/master/spec.md#stringreplace), a negative value for `count` in `string.replace` should be ignored and this is also what python and the Go implementation do.
```
# Current implementation
"abc".replace("a", "z", -1)    # "abc"

# Python/Go
"abc".replace("a", "z", -1)    # "zbc"
```

**Side Note:** The parameter `count` of `string.replace` is incorrectly named `maxsplit` in the definition of `replace` [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java#L277).

zegl:
Also update the name of the argument to be "count" instead of "maxsplit",
to match the Starlark spec.

This fixes #9181

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -279,14 +279,24 @@ final class StringModule implements StarlarkValue {
             type = Integer.class,
             noneable = true,
             defaultValue = "None",
-            doc = "The maximum number of replacements. A negative value is ignored.")
-      })
-  public String replace(String self, String oldString, String newString, Object count)
+            doc = "The maximum number of replacements. A negative value is ignored (if --incompatible_string_replace_count is true).")
+      },
+      useStarlarkThread = true)
+  public String replace(String self, String oldString, String newString, Object count, StarlarkThread thread)
       throws EvalException {
     int maxReplaces = Integer.MAX_VALUE;
-    if (count != Starlark.NONE && (Integer) count >= 0) {
-      maxReplaces = (Integer) count;
+
+    StarlarkSemantics semantics = thread.getSemantics();
+    if (semantics.incompatibleStringReplaceCount()) {
+      if (count != Starlark.NONE && (Integer) count >= 0) {
+        maxReplaces = (Integer) count;
+      }
+    } else {
+      if (count != Starlark.NONE) {
+        maxReplaces = Math.max(0, (Integer) count);
+      }
     }
+
     StringBuilder sb = new StringBuilder();
     int start = 0;
     for (int i = 0; i < maxReplaces; i++) {
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -274,16 +274,17 @@ final class StringModule implements StarlarkValue {
             type = String.class,
             doc = "The string to replace with."),
         @Param(
-            name = "count",
+            // TODO(#8147): rename param to "count" once it's positional-only.
+            name = "maxsplit",
             type = Integer.class,
             noneable = true,
             defaultValue = "None",
-            doc = "The maximum number of replacements.")
+            doc = "The maximum number of replacements. A negative value is ignored.")
       })
   public String replace(String self, String oldString, String newString, Object count)
       throws EvalException {
     int maxReplaces = Integer.MAX_VALUE;
-    if (count != Starlark.NONE && (Integer) count > -1) {
+    if (count != Starlark.NONE && (Integer) count >= 0) {
       maxReplaces = (Integer) count;
     }
     StringBuilder sb = new StringBuilder();
--- a/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
+++ b/src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
@@ -274,21 +274,21 @@ final class StringModule implements StarlarkValue {
             type = String.class,
             doc = "The string to replace with."),
         @Param(
-            name = "maxsplit",
+            name = "count",
             type = Integer.class,
             noneable = true,
             defaultValue = "None",
             doc = "The maximum number of replacements.")
       })
-  public String replace(String self, String oldString, String newString, Object maxSplitO)
+  public String replace(String self, String oldString, String newString, Object count)
       throws EvalException {
-    int maxSplit = Integer.MAX_VALUE;
-    if (maxSplitO != Starlark.NONE) {
-      maxSplit = Math.max(0, (Integer) maxSplitO);
+    int maxReplaces = Integer.MAX_VALUE;
+    if (count != Starlark.NONE && (Integer) count > -1) {
+      maxReplaces = (Integer) count;
     }
     StringBuilder sb = new StringBuilder();
     int start = 0;
-    for (int i = 0; i < maxSplit; i++) {
+    for (int i = 0; i < maxReplaces; i++) {
       if (oldString.isEmpty()) {
         sb.append(newString);
         if (start < self.length()) {
--- a/src/main/java/com/google/devtools/build/lib/packages/StarlarkSemanticsOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/StarlarkSemanticsOptions.java
@@ -508,6 +508,18 @@ public class StarlarkSemanticsOptions extends OptionsBase implements Serializabl
       help = "If set to true, the command parameter of actions.run_shell will only accept string")
   public boolean incompatibleRunShellCommandString;
 
+  @Option(
+      name = "incompatible_string_replace_count",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.STARLARK_SEMANTICS,
+      effectTags = {OptionEffectTag.BUILD_FILE_SEMANTICS},
+      metadataTags = {
+        OptionMetadataTag.INCOMPATIBLE_CHANGE,
+        OptionMetadataTag.TRIGGERED_BY_ALL_INCOMPATIBLE_CHANGES
+      },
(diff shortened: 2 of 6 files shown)


iirina:
@laurentlb @c-parsons can you take a look? Thanks!

zegl:
I've rebased and force-pushed to resolve conflicts in `EvaluationTest.java`.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -278,23 +278,23 @@ public String strip(String self, Object charsOrNone) {
             type = String.class,
             doc = "The string to replace with."),
         @Param(
-            name = "maxsplit",
+            name = "count",
```
ulfjack: I think renaming this would be a breaking change. May need to keep the old name for now.

zegl: Yes, it's clearly breaking! The question is if it's a change that's allowed to be made according to the compatability policy, the [Backwards Compatability document](https://docs.bazel.build/versions/master/backward-compatibility.html) does not mention what's allowed and not in these edge cases.

To draw some inspiration from other projects: the [Go1 compatability document](https://golang.org/doc/go1compat) states:

>  If a compiler or library has a bug that violates the specification, a program that depends on the buggy behavior may break if the bug is fixed. We reserve the right to fix such bugs.

I think that this is an example of that, as the Starlark implementation in Bazel does not satisfy the [spec](https://github.com/bazelbuild/starlark/blob/master/spec.md#stringreplace).

I'll leave it up to you and the team to figure this out, but I do think that it would be nice to allow to make some breaking changes without increasing the major-version number in the future.

c-parsons: This is definitely a breaking change which is not allowed by our policy -- apologies this is not clear by the policy doc, but this is definitely going to break some users and cause some pain if we change it.

We should avoid renaming this for now.
The parameter is tagged with `legacyNamed = true`, so, subject to the rollout of https://github.com/bazelbuild/bazel/issues/8147, this parameter will be positional-only in the future. We can then rename the parameter safely for documentation purposes.

src/test/java/com/google/devtools/build/lib/syntax/SkylarkStringModuleTest.java
```diff
@@ -0,0 +1,41 @@
+// Copyright 2014 The Bazel Authors. All rights reserved.
```
c-parsons: 2019*

src/test/java/com/google/devtools/build/lib/syntax/SkylarkStringModuleTest.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.devtools.build.lib.syntax.util.EvaluationTestCase;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for SkylarkStringModule.
+ */
+@RunWith(JUnit4.class)
+public class SkylarkStringModuleTest extends EvaluationTestCase {
```
c-parsons: StringModuleTest ?

src/test/java/com/google/devtools/build/lib/syntax/SkylarkStringModuleTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.devtools.build.lib.testutil.MoreAsserts.assertThrows;
+
+import com.google.devtools.build.lib.syntax.util.EvaluationTestCase;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for SkylarkStringModule.
```
c-parsons: StringModule ?

aiuto:
Ping

aiuto:
Another ping

zegl:
@aiuto Are you pinging me?

What would be the process of getting this change in now? I'm guessing that it's still not allowed breakage, or is it?

zegl:
I resolved the conflicts (Runtime.NONE has been renamed to Starlark.NONE), and fixed the issues found by c-parsons.

laurentlb:
Can you take a look at https://www.bazel.build/breaking-changes-guide.html?
If you put the change behind a flag, we'll be able to import it and ensure a smooth rollout.

Thanks!

aiuto:
I was pinging all the stale PRs so we can finish them or close them.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -274,21 +274,21 @@ public String strip(String self, Object charsOrNone) {
...
             type = Integer.class,
             noneable = true,
             defaultValue = "None",
             doc = "The maximum number of replacements.")
```
brandjon, line 281: Please update:

> "The maximum number of replacements. A negative value is ignored."

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -274,21 +274,21 @@ public String strip(String self, Object charsOrNone) {
...
       })
-  public String replace(String self, String oldString, String newString, Object maxSplitO)
+  public String replace(String self, String oldString, String newString, Object count)
       throws EvalException {
-    int maxSplit = Integer.MAX_VALUE;
-    if (maxSplitO != Starlark.NONE) {
-      maxSplit = Math.max(0, (Integer) maxSplitO);
+    int maxReplaces = Integer.MAX_VALUE;
+    if (count != Starlark.NONE && (Integer) count > -1) {
```
brandjon, line 286: Super nit: I feel like `count >= 0` is more readable than `count > -1`.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
brandjon: To clarify, there's two breaking changes: the rename and the behavior for negative values.

The rename is subsumed by #8147, so we should definitely not do any extra work here. You can add a comment reminding us to get back to it:

> // TODO(#8147): rename param to "count" once it's positional-only.

The behavioral change for negatives is also breaking. I feel like it's not very likely people will be broken by it, but it seems like the kind of thing we should guard with a flag. Look for other examples of using incompatible flags in Starlark builtins -- for instance, `incompatibleDisableDepsetItems` in the depset constructor later in this file, and its corresponding definitions in `StarlarkSemantics` and `StarlarkSemanticsOptions`.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
aiuto: That's not a super nit. >=0 simply *is* more readable than > -1.

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
zegl: Ok, let's do this the proper way, by marking the negative (or zero/default) value behaviour change as breaking.

Other than adding the flag and the handling of it, is there anything else that needs to be done? What's the process for creating issues with the `incompatible-change` and `breaking-change-*` labels?

laurentlb: When you file the GitHub issue, you can cc me so that I add the labels.
If your change is ready this week, we can include it in Bazel 3.2.

Thanks!

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
```diff
@@ -274,21 +274,32 @@ public String strip(String self, Object charsOrNone) {
...
+  public String replace(String self, String oldString, String newString, Object count, StarlarkThread thread)
       throws EvalException {
-    int maxSplit = Integer.MAX_VALUE;
-    if (maxSplitO != Starlark.NONE) {
-      maxSplit = Math.max(0, (Integer) maxSplitO);
+    int maxReplaces = Integer.MAX_VALUE;
+
+    StarlarkSemantics semantics = thread.getSemantics();
+    if (semantics.incompatibleStringReplaceCount()) {
+      if (count != Starlark.NONE && (Integer) count >= 0) {
```
alandonovan, line 291: Also: 'None' is not a legal argument for count, according to the Starlark spec (which matches Pythons 2 and 3 and go.starlark.net implementation).

brandjon: Hm, I'll write a quick followup to change the default value to the internal UNBOUND sentinel and have the incompatible flag also reject the value None.

brandjon:
FYI in the internal merge we retained some of the test cases in `string_misc.sky` and added a TODO to port `StringModuleTest` back over to those once the incompatible flag is no longer a blocker. We do prefer that tests of pure Starlark features be written in Starlark where possible (we've got enough verbose Java tests :) ).

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
laurentlb: Thanks. I suggest to:

- Change the default value to unbound
- Raise an error if the argument is None and the incompatible flag is enabled
- After the flag flip, remove noneable=True

src/main/java/com/google/devtools/build/lib/syntax/StringModule.java
brandjon: @alandonovan points out to me that we actually already flipped *and deleted* `--incompatible_restrict_named_params`, so I'll do the rename as part of the follow-up.

PR #9232, opened 2019-08-23
Client: introduce Path class, convert server_dir

laszlocsomor:
Benefits of a Path abstraction instead of raw
strings:

- type safety
- no need to convert paths on Windows all the time
- no need to check if paths are absolute


src/test/cpp/util/path_posix_test.cc
```diff
@@ -184,4 +184,52 @@ TEST(PathPosixTest, MakeAbsoluteAndResolveEnvvars) {
             JoinPath(GetCwd(), "%PATH%"));
 }
 
+TEST(PathPosixTest, NormalizeAbsPath) {
+  EXPECT_EQ(TestOnly_NormalizeAbsPath(""), "");
```
meteorcloudy: Can you also a test for path starting with `../`, eg. `../x/y`

laszlocsomor: Done.

laszlocsomor:
Thanks for the review! Done.

Issue #6799, opened 2018-11-29
incompatible_remove_native_maven_jar


dkelmer:
The native maven_jar rule will no longer be available with this flag set to true. 

---

**(Update: managed by @jin as of July 12)**

A migration tool will be provided to automatically convert `maven_jar` targets to `rules_jvm_external`'s `maven_install` rule, and modify BUILD files to use the new labels.

See https://docs.google.com/document/d/1CqxPv-TwvFWBHgtg7_QhVfck4vH4rnmCU_IuMdC7O2M/edit# for a design document of this migration tool.

---

Proposal by @dkelmer as of Nov 28, which is another valid migration solution that is done manually:

Use 
```python
load("@bazel_tools//tools/build_defs/repo:java.bzl", "java_import_external")
```
or the convenience wrapper

```python
load("@bazel_tools//tools/build_defs/repo:jvm.bzl", "jvm_maven_import_external")
```
instead. 

Given a `WORKSPACE` file that looks like the following:

```python
maven_jar(
    name = "truth",
    artifact = "com.google.truth:truth:0.30",
    sha1 = "9d591b5a66eda81f0b88cf1c748ab8853d99b18b",
)
```

It will need to look like this after updating:
```python
load("@bazel_tools//tools/build_defs/repo:jvm.bzl", "jvm_maven_import_external")
jvm_maven_import_external(
    name = "truth",
    artifact = "com.google.truth:truth:0.30",
    artifact_sha256 = "59721f0805e223d84b90677887d9ff567dc534d7c502ca903c0c2b17f05c116a",
    server_urls = ["http://central.maven.org/maven2"],
    licenses = ["notice"],  # Apache 2.0
)
```

Notably
*   the `licenses` attribute is mandatory
*   sha1 is no longer supported, only sha256 is
*   the `server_urls` attribute is mandatory. If your `maven_jar` rule did not specify a url
    then you should use the default server ("http://central.maven.org/maven2"). If your rule
    did specify a url then keep using that one.

dkelmer:
Broken projects: 
- [ ] android testing: https://github.com/googlesamples/android-testing 
- [ ] BUILD_file_generator: https://github.com/bazelbuild/BUILD_file_generator
- [ ] intellij/CLion plugin: https://github.com/bazelbuild/intellij 
- [ ] migration tooling: https://github.com/bazelbuild/migration-tooling
- [ ] rules_appengine: https://github.com/bazelbuild/rules_appengine
[...]

PR #9235, opened 2019-08-23
Wire up --incompatible_disallow_unverified_http_downloads for maven_server

jin:
Force usage of either HTTPS or HTTP w/ SHA-1. Note that SHA-1 is still susceptible to collision attacks, but this should reduce the exploitable surface of the current implementation that allows plain HTTP without checksums.

Also see https://github.com/bazelbuild/bazel/issues/6799#issuecomment-524382068

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/MavenJarFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/MavenJarFunction.java
@@ -14,6 +14,7 @@
 
 package com.google.devtools.build.lib.bazel.repository;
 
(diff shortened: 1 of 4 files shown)


JLLeitschuh:
Will there be another warning just simply acknowledging that sha1 is considered cryptographically broken?

jin:
@JLLeitschuh The issue with that now is warning spam (especially for projects with many maven_jar rules) with no simple and immediate actionable, like switching to SHA-256 immediately because that attribute does not exist yet. I can look into adding a SHA-256 attribute to maven_jar.

dstiers-1109:
Thank u

Issue #8880, opened 2019-07-12
Bazel allows users to use sha1 Hash for Jar Verification


JLLeitschuh:
This is a public disclosure of a 0-day vulnerability in Bazel that has not been fixed within the 90-day vulnerability disclosure window given to the Google security team.

The original report to the Google Security Team can be found here:
https://issuetracker.google.com/issues/130163464

# Summary: CWE-327: Use of a Broken or Risky Cryptographic Algorithm for Security Verification of Build Dependencies

Bazel's build uses SHA-1 hashes to verify the integrity of build artifacts resolved from artifact servers to ensure that they haven't been maliciously compromised.

https://docs.bazel.build/versions/master/be/workspace.html#maven_jar_args

> sha1 String; optional
>
> A SHA-1 hash of the desired jar.
> If the downloaded jar does not match this hash, Bazel will error out. It is a security risk to omit the SHA-1 as remote files can change. At best omitting this field will make your build non-hermetic. It is optional to make development easier but should be set before shipping.

The specific quote that's important here is "It is a security risk to omit the SHA-1 as remote files can change."

According to Wikipedia: https://en.wikipedia.org/wiki/SHA-1
"Since 2005 SHA-1 has not been considered secure against well-funded opponents"

A practical SHA-1 attack has been demonstrated by Google:
https://shattered.io/

I want to note that some uses of Bazel I've found in Non-Google public repositories are using HTTP to resolve dependencies from artifact server, thus opening them to MITM attacks where an SHA-1 collision artifact is served. See #8607

### Attack scenario:

Large, well-funded organizations with the capabilities to invest lots of CPU time into this attack.

> Who is capable of mounting this attack?
> This attack required over 9,223,372,036,854,775,808 SHA1 computations. This took the equivalent processing power as 6,500 years of single-CPU computations and 110 years of single-GPU computations.

https://shattered.io/

By creating an SHA-1 collision of some JAR file, hacking a popular artifact server (like JFrog or Maven Central) and replacing those artifacts, Bazel users may be using maliciously compromised artifacts without knowing that's what they are doing.

----

The second preimage attack now only costs ~$100,000 according to new research published May 13th, 2019.

> Our initial estimations were $1 million to compute the chosen-prefix collision, which is an amount of money we simply don't have. Thanks to our latest improvements, the cost went down below $100,000 and we are currently working on computing the first chosen-prefix collision for SHA-1.
[...]

PR #9237, opened 2019-08-23
Add sha256 and sha256_src attributes to maven_jar

jin:
..and print warnings if sha256 or sha256_src aren't used, like this:

> `WARNING: /usr/local/google/home/jingwen/code/copybara/WORKSPACE:192:1: maven_jar rule @junit//jar: Not using a checksum to verify the integrity of the artifact or the usage of SHA-1 is not secure (see https://shattered.io) and can result in an non-reproducible build. Please specify the SHA-256 checksum with: sha256 = "90a8e1603eeca48e7e879f3afbc9560715322985f39a274f6f6070b43f9d06fe",`

> `WARNING: /usr/local/google/home/jingwen/code/copybara/WORKSPACE:192:1: maven_jar rule @junit//jar: Not using a checksum to verify the integrity of the artifact or the usage of SHA-1 is not secure (see https://shattered.io) and can result in an non-reproducible build. Please specify the SHA-256 checksum with: sha256_src = "694f4694a51f67dadea4d2045742d38fb4efb92d82d42744b15e26ce653bcd3e",`

The warning message is designed to be copy paste-able directly into the WORKSPACE file. 

https://github.com/bazelbuild/bazel/issues/6799
https://github.com/bazelbuild/bazel/issues/8880

RELNOTES: Added `sha256` and `sha256_src` attributes to `maven_jar`. Please consider migrating to SHA-256 as SHA-1 has been deemed cryptographically insecure ([https://shattered.io]()). Or, use [`rules_jvm_external`](https://github.com/bazelbuild/rules_jvm_external) to manage your transitive Maven dependencies with artifact pinning and SHA-256 verification support.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/workspace/MavenJarRule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/workspace/MavenJarRule.java
@@ -55,6 +55,20 @@ public class MavenJarRule implements RuleDefinition {
         <p>Either this or <code>repository</code> can be specified.</p>
         <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
         .add(attr("server", Type.STRING))
+        /* <!-- #BLAZE_RULE(maven_jar).ATTRIBUTE(sha256) -->
+        A SHA-256 hash of the desired jar.
+
+        <p>If the downloaded jar does not match this hash, Bazel will error out. It is a security
+        risk to download a file without verifying cryptographic secure hash.</p>
+        <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
+        .add(attr("sha256", Type.STRING))
+        /* <!-- #BLAZE_RULE(maven_jar).ATTRIBUTE(sha256_src) -->
+        A SHA-256 hash of the desired jar source file.
+
+        <p>If the downloaded source jar does not match this hash, Bazel will error out. It is a
+        security risk to download a file without verifying cryptographic secure hash.</p>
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/bazel/rules/workspace/MavenJarRule.java
```diff
@@ -55,21 +55,37 @@ public RuleClass build(RuleClass.Builder builder, RuleDefinitionEnvironment envi
...
-        <code>rules_jvm_external</code></a> and pin your Maven artifacts with their SHA-256
-        checksums.</p>
+        purposes. Please either use the 'sha256' attribute or migrate to
+        <a href="https://github.com/bazelbuild/rules_jvm_external/"><code>rules_jvm_external</code>
+        </a> and pin your Maven artifacts with their SHA-256 checksums.</p>
         <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
         .add(attr("sha1", Type.STRING))
         /* <!-- #BLAZE_RULE(maven_jar).ATTRIBUTE(sha1_src) -->
-        A SHA-1 hash of the desired jar source file.
+        A SHA-1 hash of the desired jar source file. Please consider using 'sha256_src' instead.
```
aiuto: I don''t know if we need the "please" in all of these. This is a strong recommendation and my hunch is we will fully remove sha1 in 2.0. I think you are fine with
"A SHA-1 hash of the desired jar source file. (Deprecated: use sha256)"

PR #9244, opened 2019-08-26
Remove on-demand runfile tree creation from test strategy

buchgr:
Remove on-demand runfile tree creation from test strategy
  
Bazel builds runfiles trees when building an executable target. This
allows a binary to be run directly (without bazel test/run). The flag
--nobuild_runfile_trees disables this behavior. In order to still be
able to run tests that require a runfiles tree the
Local/StandaloneTestStrategy has code to create / update it before
executing the test.

However, most execution strategies in Bazel don't use the runfiles tree:
  * Actions being executed using sandboxing don't have access to the
    runfiles tree as their working directory is an ephemeral directory
    created by the sandbox. Instead, the sandbox treats runfiles as normal
    inputs and creates a new tree for every action.
  * Actions being executed with the remote strategy don't have access to
    the runfiles tree either as they run on a different machine. Just like
    the sandbox the remote strategy treats runfiles as normal inputs and
    uploads them to the remote worker.

The runfiles tree is only used by actions executed with the standalone/local
or worker (with --worker_sandboxing=false) strategy as they use the exec
root as their working directory.

This change makes it so that if --nobuild_runfile_links is specified a
runfiles tree is only created if the execution strategy actually uses
it.

We found that this can lead to significant performance improvements in
particular for very large runfile trees. Also, before this change in
remote execution we would create a runfiles tree even for remote cache
hits.

As a side effect this change also removes an imparity between binaries
and tests, in that binaries had no such logic to on-demand create a
runfiles tree if --nobuild_runfile_links is set.

I will send out a follow up change with a similar optimization the
input / output manifest.

Note to Reviewers:
 * RunfilesTreeUpdater replaces TestStrategy#getLocalRunfilesDirectory.
 * The locking logic in RunfilesTreeUpdater implements one monitor per
   runfiles directory. The monitor ensures that a runfiles directory
   can't be modified concurrently. It replaces synchronized(inputManifest)
   from getLocalRunfilesDirectory. We can't use synchronized(inputManifest)
   for two reasons:
     1) The execution strategies don't have access to it.
     2) As noted above, in a follow up change I want to enable this logic
     to also work when --nobuild_runfile_manifests is set in which case no
     inputManifest object to use as monitor will exist.


buchgr:
@werkt I believe you will like this.

buchgr:
I would typically be able to wait but this week is special because next Monday Bazel 1.0 will be cut :). I'd appreciate reviews before that. I do realize that many folks knowledge able in this area are currently on vacation :\.

buchgr:
@haxorz I wasn't trying to pick on you to do the review. I requested reviews from folks that I was thought might be knowledgeable in this area.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -444,4 +446,14 @@ private static CommandLine computeArgs(
         ruleContext.getExpander().withDataLocations().tokenized("args"),
         additionalArgs);
   }
+
+  /** Provided the path of the runfiles directory returns the path of the input manifest. */
```
laszlocsomor: The comment just spells out the function signature. Can you make it more useful? Explaining the input manifest's use could be helpful. How about something along these lines:

> Returns the input manifest for runfiles creation.
> The input manifest lists pairs of paths: execroot-relative paths of the runfiles with the runfiles-root-relative path they should be mapped to.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -444,4 +446,14 @@ private static CommandLine computeArgs(
...
+  public static Path inputManifestPath(Path runfilesDir) {
+    return FileSystemUtils.replaceExtension(runfilesDir, INPUT_MANIFEST_EXT);
+  }
+
+  /** Provided the path of the runfiles directory returns the path of the output manifest. */
```
laszlocsomor: Same here as line 450.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -95,7 +97,7 @@ private static RunfilesSupport create(
       RuleContext ruleContext, Artifact executable, Runfiles runfiles, CommandLine args) {
     Artifact owningExecutable = Preconditions.checkNotNull(executable);
     boolean createManifest = ruleContext.getConfiguration().buildRunfilesManifests();
-    boolean createSymlinks = createManifest && ruleContext.getConfiguration().buildRunfiles();
+    boolean createSymlinks = ruleContext.getConfiguration().shouldCreateRunfilesTree();
```
laszlocsomor: With the exception of this line, the rest of the changes in this file are refactorings that are independent of the logic change.
Could you create a separate PR just for refactorings? That'd make the logic-changing parts easier to review.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -73,6 +73,8 @@
 @AutoCodec
```
laszlocsomor: All changes in this file are refactorings that are independent of the business logic change.

Could you create a separate PR just for refactorings? That'd make the logic-changing parts easier to review.

src/main/java/com/google/devtools/build/lib/analysis/config/BuildConfiguration.java
```diff
@@ -864,6 +864,11 @@ public String getHostCpu() {
     return options.hostCpu;
   }
 
+  public static boolean shouldCreateRunfilesTree(CoreOptions options) {
+    return options.buildRunfilesManifests && options.buildRunfiles;
+  }
+
+  // TODO(buchgr): https://github.com/bazelbuild/bazel/issues/9248
```
laszlocsomor: What's the TODO about? Please note the desired action and its condition.

src/main/java/com/google/devtools/build/lib/standalone/StandaloneModule.java
```diff
@@ -51,13 +54,20 @@ public void executorInit(CommandEnvironment env, BuildRequest request, ExecutorB
             env.getBlazeWorkspace().getBinTools(),
             testTmpRoot);
 
+    CoreOptions coreOptions = env.getOptions().getOptions(CoreOptions.class);
+    boolean runfilesEnabled = BuildConfiguration.runfilesEnabled(coreOptions);
+    // True if all runfile trees will be created when building binaries/test. If so, we won't need
+    // the RunfilesTreeUpdater.
+    boolean createRunfilesAtBuild = BuildConfiguration.shouldCreateRunfilesTree(coreOptions);
```
lberki: This will not work if someone changes the value of this flag in a configuration transition. Sure, it would be kind of weird to do so, but one can do that from Starlark (same in `WorkerModule`). You'll need to find a way to convey that bit from the configured target itself.

src/main/java/com/google/devtools/build/lib/exec/RunfilesTreeUpdater.java
```diff
@@ -0,0 +1,159 @@
...
+      // TODO(buchgr): Add logic to create a runfiles tree if --nobuild_runfile_manifests is set
+      throw new UserExecException("Can't build local runfiles tree because "
+          + "--nobuild_runfile_manifests is set.");
+    }
+
+    Path outputManifest = RunfilesSupport.outputManifestPath(runfilesDirPath);
+    try {
+      // Avoid rebuilding the runfiles directory if the manifest in it matches the input manifest,
+      // implying the symlinks exist and are already up to date. If the output manifest is a
+      // symbolic link, it is likely a symbolic link to the input manifest, so we cannot trust it as
```
lberki: Is it? I thought that we create it anew so as to avoid this exact issue, don't we?

src/main/java/com/google/devtools/build/lib/exec/RunfilesTreeUpdater.java
```diff
@@ -0,0 +1,159 @@
...
+        // monitorsWithRefcnt map.
+        return null;
+      }
+      if (monitorWithRefcnt.compareAndSet(count, count + 1)) {
+        return monitorWithRefcnt;
+      }
+    }
+  }
+
+  private AtomicInteger getMonitor(PathFragment runfilesDirectory) {
```
lberki: I think this works, but I'm kinda afraid of this kind of multi-threaded code. Do you have any other alternatives that do not require this? For example (off the top of my head, I haven't given this too much thought), you could synchronize on the `Path` object referring to one of the manifests (not an `Artifact`, though, that's not guaranteed to be unique)

Or you could just add the lock to the `TestRunnerAction`. That has the advantage of not abusing the monitor of a `Path`. Of course, this only works if this is the only case when a runfiles tree is reused (I don't know what happens when the same binary and its runfiles tree is used as a tool in two local genrules executing in parallel)

buchgr:
> As an alternative, /cc @aragos is considering making it so that strategy selection is decided in the analysis phase. If that comes to pass, you can just make this decision in the analysis phase.and not worry about RunfilesTreeUpdater.

I am aware of this and I agree. However, there will be lots of water flowing down the Isar until then :).

lberki:
> I am aware of this and I agree. However, there will be lots of water flowing down the Isar
Wat, that saying works in any language other than Hungarian, too? (we say Danube, but same difference :) )

src/main/java/com/google/devtools/build/lib/standalone/StandaloneModule.java
buchgr: Interesting point. Would you consider this a blocker for this change or can this be done as an immediate follow up?

src/main/java/com/google/devtools/build/lib/exec/RunfilesTreeUpdater.java
buchgr: I copied this comment from the old code. I also could not find a place where the output manifest is a symlink to the input manifest. It might be time to retire the comment and remove the check?

src/main/java/com/google/devtools/build/lib/standalone/StandaloneModule.java
lberki: Blocker. Let's not submit known-incorrect code.

src/main/java/com/google/devtools/build/lib/exec/RunfilesTreeUpdater.java
buchgr: > For example (off the top of my head, I haven't given this too much thought), you could synchronize on the Path object referring to one of the manifests (not an Artifact, though, that's not guaranteed to be unique)

Exactly. What this code does is map equal path objects to a unique object. I believe it's the most principled way of solving this problem, in particular because it pushes the complexity down the stack where it belongs.

> Or you could just add the lock to the TestRunnerAction. 

I think this would serialize execution of shards?

> I don't know what happens when the same binary and its runfiles tree is used as a tool in two local genrules executing in parallel

What would happen is exactly what this code is trying to prevent :).

lberki: Let me word this more carefully: you are probably thinking `public synchronized runTest() { }`, I'm thinking

```
Object runfilesLock;

public void runTest() {
  if (runfilesTreeNeeded()) {
    synchronized (runfilesLock) {
      ...
    }
  }
}
```

...and what would that be in case of the two-genrules-using-the-same-tool situation?

buchgr: @lberki 

This would serialize runfiles tree creation for all tests where runfilesTreeNeeded() returns false. This is a regression.

>...and what would that be in case of the two-genrules-using-the-same-tool situation?

As my change description states runfiles tree creation when --nobuild_runfile_links is true does not work for genrules. It only works for tests (because there is custom code). A side effect of this change is that it will make it work for genrules as well.

buchgr:
> Wat, that saying works in any language other than Hungarian, too? (we say Danube, but same difference :) )

I also say Danube (makes sense right?) but adjusted it to Isar given our location. I am not aware of anyone else saying it that way :).

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
buchgr: done (inside google)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
buchgr: done (inside google)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
buchgr: done (inside google)

src/main/java/com/google/devtools/build/lib/analysis/config/BuildConfiguration.java
buchgr: done (inside google)

buchgr:
@lberki @laszlocsomor 

I have imported the CL, addressed all comments except for one and split it up into five smaller changes. Let's continue the review internally.

The one comment I did not fully address is @lberki's comments regarding getting rid of `RunfilesTreeUpdater`. I found that even the solution we discussed offline with `RunfilesSupport.lock` to not work satisfactory. Instead, I simplified `RunfilesTreeUpdater`quite a bit by replacing all the lock-free code with a synchronized block which should be good enough.

PR #9245, opened 2019-08-26
Client: convert output_base to Path

laszlocsomor:
Benefits of a Path abstraction instead of raw
strings:
- type safety
- no need to convert paths on Windows all the time
- no need to check if paths are absolute


laszlocsomor:
ping @meteorcloudy

Issue #5055, opened 2018-04-19
bazelrc: multiplatform support


laszlocsomor:
### Description of the problem / feature request:

I'd like to have some mechanism to use platform-specific flags in bazelrc files.

### Feature requests: what underlying problem are you trying to solve with this feature?

Motivation is to let checked-in, project-specific bazelrc files work on all platforms. Flags dealing with paths (e.g. `--output_user_root`), envvars (e.g. `--test_env`), and external programs or shells (e.g. `--workspace_status_command`) are all platform-specific.

cvcal:
We've talked about different solutions to this problem in the past, but the easiest solution is to accept multiple rc files (#4502), which will let users keep platform specific flags in the global rc while keeping project specific flags in a workspace rc file. The workspace one can be shared between different platforms. 
Other solutions we've considered include checking for a "<platform>.bazelrc" filename, or making the rc file syntax more complicated so it can include some flags on some platforms and not others. Neither of these options seem to solve a problem that #4502 doesn't already address, though, and 4502 has other reasons for needing to happen. 
Thoughts?

cvcal:
The general platform issue for bazel flags is very much not solved, the solution I mentioned leaves us in basically the following state:
- host platform: goes in global bazelrc
- execution platform: if host, in global bazelrc, if remote, likely in the workspace file. 
- target platform: the workspace rc file can contain --config definitions that correspond to the different configurations needed.

I realize this is not ideal, but the "right" solution is not clear to me. The fixes in the other issue will only get us out of the unusable into the "works, I guess" territory.

ittaiz:
+1 to Chloe’s description

laszlocsomor:
> Thoughts?

I don't really have any, you're the domain expert. :)

> The fixes in the other issue will only get us out of the unusable into the "works, I guess" territory.

SGTM.

davido:
> ... and external programs or shells (e.g. --workspace_status_command) are all platform-specific

This is indeed a serious problem. Gerrit Code Review has this content of `tools/bazel.rc` (committed in the repository):

```
build --workspace_status_command=./tools/workspace-status.sh
```

Stamping machinery shells out to cmd.exe, so that we have to provide a different script: `hazel build --workspace_status_command=./tools/workspace-status.cmd ...`. There is currently no way to specify different option, except to overwrite it on bazel invocation.

[...]

PR #9246, opened 2019-08-26
bazelrc: support --enable_platform_specific_config

meteorcloudy:
When the value of this flag is true, a host platform specific config
section will be enabled if it exists. Bazel recognizes your host
platform as linux, osx, windows or freebsd. It's equivalent to add
--config=linux on Linux platforms and --config=windows on Windows, etc.

Fixes https://github.com/bazelbuild/bazel/issues/5055

RELNOTES[NEW]: Platform-specific bazelrc: with --enable_platform_specific_config you can 
enable flags in bazelrc according to your host platform.
GOOGLE:

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
@@ -20,7 +20,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
-import com.google.common.collect.Multiset;
 import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.EventHandler;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
@@ -330,10 +329,9 @@ public final class BlazeOptionHandler {
       return false;
     }
 
-    Multiset<String> keys = commandToRcArgs.keys();
     for (String commandName : getCommandNamesToParse(commandAnnotation)) {
       String defaultConfigDef = commandName + ":" + OS.getCurrent().getCanonicalName();
-      if (keys.contains(defaultConfigDef)) {
+      if (commandToRcArgs.containsKey(defaultConfigDef)) {
         return true;
       }
     }
--- a/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
@@ -18,13 +18,16 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Multiset;
 import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.EventHandler;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
 import com.google.devtools.build.lib.runtime.commands.ProjectFileSupport;
 import com.google.devtools.build.lib.runtime.proto.InvocationPolicyOuterClass.InvocationPolicy;
 import com.google.devtools.build.lib.util.ExitCode;
+import com.google.devtools.build.lib.util.OS;
 import com.google.devtools.build.lib.vfs.FileSystemUtils;
 import com.google.devtools.build.lib.vfs.Path;
 import com.google.devtools.common.options.InvocationPolicyEnforcer;
@@ -316,6 +319,28 @@ public final class BlazeOptionHandler {
   }
 
   /**
+   * If --enable_platform_specific_config is true and the corresponding config definition exists,
+   * we should enable the platform specific config.
+   */
+  private boolean shouldEnablePlatformSpecificConfig(
+      OptionValueDescription enablePlatformSpecificConfigDescription,
+      ListMultimap<String, RcChunkOfArgs> commandToRcArgs) {
+    if (enablePlatformSpecificConfigDescription == null ||
+        !(boolean) enablePlatformSpecificConfigDescription.getValue()) {
+      return false;
+    }
+
+    Multiset<String> keys = commandToRcArgs.keys();
+    for (String commandName : getCommandNamesToParse(commandAnnotation)) {
+      String defaultConfigDef = commandName + ":" + OS.getCurrent().getCanonicalName();
+      if (keys.contains(defaultConfigDef)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
    * Expand the values of --config according to the definitions provided in the rc files and the
    * applicable command.
    */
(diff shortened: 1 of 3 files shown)


laszlocsomor:
> RELNOTES[NEW]: Support --enable_platform_specific_config, you can use
> this flag to enable flags in bazelrc according to your host platform.
> GOOGLE:

How about this instead:

> RELNOTES[NEW]: Platform-specific bazelrc: with --enable_platform_specific_config you can use
> enable flags in bazelrc according to your host platform.

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
```diff
@@ -315,6 +318,28 @@ ExitCode parseOptions(List<String> args, ExtendedEventHandler eventHandler) {
...
 
+  /**
+   * If --enable_platform_specific_config is true and the corresponding config definition exists,
+   * we should enable the platform specific config.
+   */
+  private boolean shouldEnablePlatformSpecificConfig(
+      OptionValueDescription enablePlatformSpecificConfigDescription,
+      ListMultimap<String, RcChunkOfArgs> commandToRcArgs) {
+    if (enablePlatformSpecificConfigDescription == null ||
+        !(boolean) enablePlatformSpecificConfigDescription.getValue()) {
```
laszlocsomor: Is this cast safe? Shouldn't it be `(Boolean)` and check for null?

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
```diff
@@ -315,6 +318,28 @@ ExitCode parseOptions(List<String> args, ExtendedEventHandler eventHandler) {
...
+      return false;
+    }
+
+    Multiset<String> keys = commandToRcArgs.keys();
```
laszlocsomor: If you inlined this, you wouldn't need to import Multiset.
Also, call `containsKey` instead of `keys().contains` -- it's possible that the latter constructs a set and the former doesn't.

meteorcloudy:
Great, thanks!

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
```diff
@@ -325,23 +350,30 @@ void expandConfigOptions(
...
     OptionValueDescription configValueDescription =
         optionsParser.getOptionValueDescription("config");
-    if (configValueDescription == null || configValueDescription.getCanonicalInstances() == null) {
-      // No --config values were set, we can avoid this whole thing.
-      return;
+    if (configValueDescription != null && configValueDescription.getCanonicalInstances() != null) {
+      // Find the base set of configs. This does not include the config options that might be
+      // recursively included.
+      ImmutableList<ParsedOptionDescription> configInstances =
+          ImmutableList.copyOf(configValueDescription.getCanonicalInstances());
```
laszlocsomor: Why make a copy?

(I realize this is old code, just asking.)

src/test/java/com/google/devtools/build/lib/runtime/BlazeOptionHandlerTest.java
```diff
@@ -297,6 +308,59 @@ public void testExpandConfigOptions_withConfig() throws Exception {
         .containsExactly("Found applicable config definition c0:config in file rc1: b");
   }
 
+  @Test
+  public void testExpandConfigOptions_withPlatformSpecificConfigEnabled() throws Exception {
```
laszlocsomor: The new test methods are only subtly different, I find it difficult to understand what each of them tests.
Could you add short descriptions to explain what a test method does and how it's different from the related test methods? (E.g. one uses `--enable_platform_specific_config` therefore X, the other one uses `--config=platform_config` therefore Y.)

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
```diff
@@ -56,6 +56,19 @@
               + "your build may break in the future due to deprecations or other changes.")
   public Void allIncompatibleChanges;
 
+  @Option(
+      name = "enable_platform_specific_config",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      help =
+          "When the value of this flag is true, a host platform specific config section will be "
```
laszlocsomor: > When the value of this flag is true

Could be rewritten as "If true".

> ... a host platform specific config section will be enabled if it exists

I find that confusing. Using active voice and straight word order feels clearer to me. How about: "... Bazel picks up host-OS-specific config lines from the bazelrc files."

I feel like an example would greatly help too: "For example, if the host OS is Linux and you run bazel build, Bazel would pick up lines starting with build:linux"

> linux, osx, windows

macOS is the official name so let's use "macos" instead of "osx".

> It's equivalent to add --config=linux on Linux platforms and --config=windows on Windows, etc.

We could be a little clearer here: "Enabling this flag is equivalent to using --config=linux on Linux, --config=windows on Windows, etc."

Putting it together:

> If true,  Bazel picks up host-OS-specific config lines from bazelrc files. For example, if the host OS is Linux and you run bazel build, Bazel picks up lines starting with build:linux. Supported OS identifiers are linux, macos, windows, and freebsd. Enabling this flag is equivalent to using --config=linux on Linux, --config=windows on Windows, etc.

meteorcloudy: Wow, thanks for helping with the documentation. This is great improvement!

As for the name for mac, it seems we still use `osx` in a lot of places in our code base. For example, https://github.com/bazelbuild/bazel/blob/71a213c42a4572f64b323b3db731e1334b6a0c38/tools/platforms/BUILD#L63

It prefer to keep the consistence util we decided to migrate everything to `macos`. WDYT?

src/test/java/com/google/devtools/build/lib/runtime/BlazeOptionHandlerTest.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
meteorcloudy: Sorry, I actually don't know the reason..

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
meteorcloudy: Nice, thanks for the suggestion!

src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java
meteorcloudy: Yes, as long as there is a default value, it won't be null and will be converted to the correct type. 
See https://github.com/bazelbuild/bazel/blob/1f684e1b87cd8881a0a4b33e86ba66743e32d674/src/main/java/com/google/devtools/common/options/OptionDefinition.java#L255

laszlocsomor: Thanks!

src/test/java/com/google/devtools/build/lib/runtime/BlazeOptionHandlerTest.java
laszlocsomor: Thanks!

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
laszlocsomor: Could you please ask the Apple team what's their preference?

philwo:
This was merged via https://github.com/bazelbuild/bazel/commit/59755455034a998cdedfb7b086aea3ad78419381.

gunan:
This looks quite useful to me.
Is there documentation on how to use this?

meteorcloudy:
@gunan The documentation is here: https://docs.bazel.build/versions/master/command-line-reference.html#flag--enable_platform_specific_config

Here is an example in `.bazelrc` file:
```
--enable_platform_specific_config
build:windows --workspace_status_command=command.bat
build:linux --workspace_status_command=command.sh
build:macos --workspace_status_command=command.sh
```
Then the corresponding option will be enabled for each platform. Be aware, Bazel will only enable those options based on your host platform, instead of execution platform or target platform.

gunan:
Is there a way to use this to change definition of certain build configs?

such as:
I want --config=avx to be "--copt=-mavx" in linux, but "--copt=/arch=AVX" on windows.

meteorcloudy:
No, there is no way to do that currently. But it sounds like a very useful feature.

PR #9259, opened 2019-08-27
WindowsFileSystem.readSymbolicLink: Map IOException to NotASymlinkException

tomaszstrejczek:
Fixes symbolic link outputs on Windows.

FileSystem.resolveOneLink expects that readSymbolicLink throws NotASymlinkException when the provided path is valid but is not a symbol link. UnixFileSystem.readSymbolicLink throws the exception when necessary, but WindowsFileSystem.readSymbolicLink - doesn't (throws IOException instead).

The PR fixes WindowsFileSystem.readSymbolicLink. It's implemented in the same way as Unix implementation.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java
+++ b/src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java
@@ -36,9 +36,10 @@ import java.nio.file.attribute.DosFileAttributes;
 public class WindowsFileSystem extends JavaIoFileSystem {
 
   public static final LinkOption[] NO_OPTIONS = new LinkOption[0];
-  public static final LinkOption[] NO_FOLLOW = new LinkOption[] {LinkOption.NOFOLLOW_LINKS};
+  public static final LinkOption[] NO_FOLLOW = new LinkOption[] { LinkOption.NOFOLLOW_LINKS };
 
-  public WindowsFileSystem() throws DefaultHashFunctionNotSetException {}
+  public WindowsFileSystem() throws DefaultHashFunctionNotSetException {
+  }
 
   public WindowsFileSystem(DigestHashFunction hashFunction) {
     super(hashFunction);
@@ -46,8 +47,10 @@ public class WindowsFileSystem extends JavaIoFileSystem {
 
   @Override
   public String getFileSystemType(Path path) {
-    // TODO(laszlocsomor): implement this properly, i.e. actually query this information from
-    // somewhere (java.nio.Filesystem? System.getProperty? implement JNI method and use WinAPI?).
+    // TODO(laszlocsomor): implement this properly, i.e. actually query this
+    // information from
+    // somewhere (java.nio.Filesystem? System.getProperty? implement JNI method and
+    // use WinAPI?).
     return "ntfs";
   }
 
@@ -67,10 +70,8 @@ public class WindowsFileSystem extends JavaIoFileSystem {
 
   @Override
   protected void createSymbolicLink(Path linkPath, PathFragment targetFragment) throws IOException {
-    Path targetPath =
-        targetFragment.isAbsolute()
-            ? getPath(targetFragment)
-            : linkPath.getParentDirectory().getRelative(targetFragment);
+    Path targetPath = targetFragment.isAbsolute() ? getPath(targetFragment)
+        : linkPath.getParentDirectory().getRelative(targetFragment);
     try {
       java.nio.file.Path link = getIoFile(linkPath).toPath();
       java.nio.file.Path target = getIoFile(targetPath).toPath();
@@ -92,7 +93,11 @@ public class WindowsFileSystem extends JavaIoFileSystem {
   @Override
   protected PathFragment readSymbolicLink(Path path) throws IOException {
     java.nio.file.Path nioPath = getNioPath(path);
-    return PathFragment.create(WindowsFileOperations.readSymlinkOrJunction(nioPath.toString()));
+    try {
+      return PathFragment.create(WindowsFileOperations.readSymlinkOrJunction(nioPath.toString()));
+    } catch (IOException e) {
+      throw e.getMessage().endsWith("path is not a link") ? new NotASymlinkException(path) : e;
+    }
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


laszlocsomor:
Thanks for fixing this!

Can you explain what bug it caused, or how you found it?

Also, can you write a test that demonstrates what this PR really fixes, i.e. shows the implication of throwing the right exception type?

tomaszstrejczek:
The bug was discovered when creating on Windows an output which is a symbolic link. 
I've provided a test (//src/test/py/bazel:bazel_windows_symlinks_test) which fails on Windows without the fix.
After the fix the test works.

src/test/py/bazel/bazel_windows_symlinks_test.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
laszlocsomor: nit: 2019

src/test/py/bazel/bazel_windows_symlinks_test.py
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import unittest
+from src.test.py.bazel import test_base
+
+
+class BazelWindowsTest(test_base.TestBase):
```
laszlocsomor: nit: BazelWindowsSymlinksTest

src/test/py/bazel/bazel_windows_symlinks_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def createProjectFiles(self):
+        self.CreateWorkspaceWithDefaultRepos('WORKSPACE')
+        self.ScratchFile(
+            'foo/BUILD', ['genrule(name="x", srcs=[":sample"], outs=["link"], cmd="python -c \\"import os; os.symlink(os.path.join(os.getcwd(), \'$<\'),\'$@\')\\"",)'])
```
laszlocsomor: Don't assume `python` is on the PATH, it isn't always.
Either create a py_binary with this script and declare it in genrule.tools and run the py_binary from the genrule.cmd, or run `in=$<; out=$@; cmd.exe /C mklink ${in//\//\\} ${out//\//\\}`.

src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java
```diff
@@ -92,7 +92,11 @@ protected void createSymbolicLink(Path linkPath, PathFragment targetFragment) th
   @Override
   protected PathFragment readSymbolicLink(Path path) throws IOException {
     java.nio.file.Path nioPath = getNioPath(path);
-    return PathFragment.create(WindowsFileOperations.readSymlinkOrJunction(nioPath.toString()));
+    try {
+      return PathFragment.create(WindowsFileOperations.readSymlinkOrJunction(nioPath.toString()));
+    } catch (IOException e) {
+      throw e.getMessage().endsWith("path is not a link") ? new NotASymlinkException(path) : e;
```
laszlocsomor: Don't branch on the exception message, let `WindowsFileOperations.readSymlinkOrJunction` throw the right exception instead.

tomaszstrejczek: WindowsFileOperations.readSymlinkOrJunction doesn't have access to NotASymlinkException (it's low-level jni wrapper). 
Similarly (I suspect that for the same reason) UnixFileSystem.readSymbolicLink branches on the exception message.
I would leave it the way it is.

laszlocsomor: Please change `WindowsFileOperations.readSymlinkOrJunction` to return an error enum then, instead of throwing `IOException`. `WindowsFileOperations` is Bazel's own internal class so we can change it freely.

`UnixFileSystem.readSymbolicLink` sets a bad example: text output meant for humans should not be an API contract because it's too easy to break without noticing. (I know there's code in Bazel that violates this because of [Hyrum's law](https://www.hyrumslaw.com), but that's bad and we shouldn't introduce more if possible.)

src/test/py/bazel/bazel_windows_symlinks_test.py
```diff
@@ -0,0 +1,46 @@
...
+class BazelWindowsSymlinksTest(test_base.TestBase):
+
+    def createProjectFiles(self):
+        self.CreateWorkspaceWithDefaultRepos('WORKSPACE')
+        self.ScratchFile('foo/BUILD', [
+               'py_binary(name="symlinker", srcs=[":symlinker.py"])',
+               'genrule(name="x",srcs=[":sample"],outs=["link"],cmd="$(location symlinker) $< $@",tools=[":symlinker"])'
+        ])
+        self.ScratchFile('foo/symlinker.py', [
+            'import os; import sys; os.symlink(os.path.join(os.getcwd(), sys.argv[1]), sys.argv[2])',
```
laszlocsomor: `os.symlink` does not exist under Python 2.

Again, I suggest the `cmd.exe` approach.

tomaszstrejczek: Both things fixed as requested.

src/test/py/bazel/bazel_windows_symlinks_test.py
```diff
@@ -0,0 +1,42 @@
...
+import unittest
+from src.test.py.bazel import test_base
+
+
+class BazelWindowsSymlinksTest(test_base.TestBase):
+
+    def createProjectFiles(self):
+        self.CreateWorkspaceWithDefaultRepos('WORKSPACE')
+        self.ScratchFile('foo/BUILD', [
+               'genrule(name="x",srcs=[":sample"],outs=["link"],cmd="IN=$< OUT=$@ OUT2=$${OUT//\\//\\\\\\\\\\\\} IN2=$${IN//\\//\\\\\\\\\\\\} cmd.exe /C \\"mklink %OUT2% %cd%\\\\%IN2%\\"")'
```
laszlocsomor: This is incorrect, the backslashes don't work out. I tried fixing it but realized it's easier to use a `.bat` file.

Here's the diff to do that (and also to add a negative test verifying that a dangling link is reported):

```diff
$ git d ; echo ---
diff --git a/src/test/py/bazel/bazel_windows_symlinks_test.py b/src/test/py/bazel/bazel_windows_symlinks_test.py
index 2ec741a91e..e376cb00ab 100644
--- a/src/test/py/bazel/bazel_windows_symlinks_test.py
+++ b/src/test/py/bazel/bazel_windows_symlinks_test.py
@@ -22,8 +22,25 @@ class BazelWindowsSymlinksTest(test_base.TestBase):
     def createProjectFiles(self):
         self.CreateWorkspaceWithDefaultRepos('WORKSPACE')
         self.ScratchFile('foo/BUILD', [
-               'genrule(name="x",srcs=[":sample"],outs=["link"],cmd="IN=$< OUT=$@ OUT2=$${OUT//\\//\\\\\\\\\\\\} IN2=$${IN//\\//\\\\\\\\\\\\} cmd.exe /C \\"mklink %OUT2% %cd%\\\\%IN2%\\"")'
+            'genrule(',
+            '    name = "x",',
+            '    srcs = ["sample"],',
+            '    outs = ["link"],',
+            '    exec_tools = ["sym.bat"],',
+            '    cmd = "$(location sym.bat) $< $@",',
+            ')',
+            'genrule(',
+            '    name = "y",',
+            '    outs = ["dangling-link"],',
+            '    exec_tools = ["sym.bat"],',
+            '    cmd = "$(location sym.bat) does-not-exist $@",',
+            ')',
         ])
+        self.ScratchFile('foo/sym.bat', [
+            '@set IN=%1',
+            '@set OUT=%2',
+            r'@mklink %OUT:/=\% %cd%\%IN:/=\%',
+        ], executable = True)
         self.ScratchFile('foo/sample', [
             'sample',
         ])
@@ -31,12 +48,14 @@ class BazelWindowsSymlinksTest(test_base.TestBase):
     def testWindowsSymlinkedOutput(self):
         self.createProjectFiles()

-        exit_code, _, stderr = self.RunBazel([
-            '--batch', 'build',
-            '//foo:x',
-        ])
+        exit_code, _, stderr = self.RunBazel(['build', '//foo:x'])
         self.AssertExitCode(exit_code, 0, stderr)

+        exit_code, _, stderr = self.RunBazel(['build', '//foo:y'])
+        self.AssertNotExitCode(exit_code, 0, stderr)
+
+        if not any(['is a dangling symbolic link' in l for l in stderr]):
+            self.fail('FAIL:\n | stderr:\n | %s' % '\n | '.join(stderr))

 if __name__ == '__main__':
     unittest.main()

```

tomaszstrejczek: Done

laszlocsomor:
Thank you, LGTM.

PR #9267, opened 2019-08-28
propagated for cc_library and cc_binary

ishikhman:
Part of #8830 

RELNOTES[NEW]: tags: use `--experimental_allow_tags_propagation` flag to propagate tags to the action's execution requirements from cc_library or cc_binary targets. Such tags should start with: `no-`, `requires-`, `supports-`, `block-`, `disable-`, `cpu:`. See #8830 for details.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -102,8 +102,6 @@ import com.google.devtools.build.lib.util.OrderedSetMultimap;
 import com.google.devtools.build.lib.util.StringUtil;
 import com.google.devtools.build.lib.vfs.FileSystemUtils;
 import com.google.devtools.build.lib.vfs.PathFragment;
-
-import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -112,6 +110,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /**
  * The totality of data available during the analysis of a rule.
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -125,6 +125,15 @@ import javax.annotation.Nullable;
 public final class RuleContext extends TargetContext
     implements ActionConstructionContext, ActionRegistry, RuleErrorConsumer {
 
+  public boolean isAllowTagsPropagation() {
+    try {
+      return this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation();
+    } catch (InterruptedException e) {
+      // fallback to false if value of experimentalAllowTagsPropagation cannot be fetched
+      return false;
+    }
+  }
+
   /**
    * The configured version of FilesetEntry.
    */
@@ -550,6 +559,40 @@ public final class RuleContext extends TargetContext
 
   @Override
   public void registerAction(ActionAnalysisMetadata... action) {
+//List<ActionAnalysisMetadata> newList = new ArrayList<>();
+//    try {
+//        if (this.getAnalysisEnvironment() == null
+//                || this.getAnalysisEnvironment().getSkylarkSemantics() == null
+//                || !this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation()) {
+//            getAnalysisEnvironment().registerAction(action);
+//            return;
+//      }
+//      for (ActionAnalysisMetadata metadata : action) {
+//        if (metadata instanceof ExecutionInfoSpecifier && !(metadata instanceof StarlarkAction)) {
+//          ExecutionInfoSpecifier executionInfoSpecifier = (ExecutionInfoSpecifier) metadata;
+//          ImmutableMap<String, String> filteredExecutionInfo = TargetUtils.getFilteredExecutionInfo(
+//                  executionInfoSpecifier.getExecutionInfo(),
+//                  rule,
+//                  this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation());
+//
+//          Map<String, String> executionInfoBuilder = new HashMap<>();
+//          // in case something was already in action.executionInfo - we copy it to the final Map
+//          executionInfoBuilder.putAll(executionInfoSpecifier.getExecutionInfo());
+//          // adding filtered execution requirements to the execution info map
+//          executionInfoBuilder.putAll(filteredExecutionInfo);
+//
+//          ActionAnalysisMetadata updatedAction = executionInfoSpecifier.addExecutionInfo(ImmutableMap.copyOf(executionInfoBuilder));
+//          newList.add(updatedAction);
+//        } else {
+//          newList.add(metadata);
+//        }
+//      }
+//    } catch (InterruptedException e) {
+//      // todo: what?
+//      e.printStackTrace();
+//      throw new RuntimeException("Something horrible happened!");
+//    }
+//    getAnalysisEnvironment().registerAction(newList.toArray(new ActionAnalysisMetadata[0]));
     getAnalysisEnvironment().registerAction(action);
   }
 
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -126,9 +126,10 @@ import java.util.stream.Collectors;
 public final class RuleContext extends TargetContext
     implements ActionConstructionContext, ActionRegistry, RuleErrorConsumer {
 
-    public boolean isAllowTagsPropagation() throws InterruptedException {
-        return this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation();
-    }
+  @Override
+  public boolean isAllowTagsPropagation() throws InterruptedException {
+    return this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation();
+  }
 
   /**
    * The configured version of FilesetEntry.
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -559,40 +559,6 @@ public final class RuleContext extends TargetContext
 
   @Override
   public void registerAction(ActionAnalysisMetadata... action) {
-//List<ActionAnalysisMetadata> newList = new ArrayList<>();
-//    try {
-//        if (this.getAnalysisEnvironment() == null
-//                || this.getAnalysisEnvironment().getSkylarkSemantics() == null
-//                || !this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation()) {
-//            getAnalysisEnvironment().registerAction(action);
-//            return;
-//      }
-//      for (ActionAnalysisMetadata metadata : action) {
-//        if (metadata instanceof ExecutionInfoSpecifier && !(metadata instanceof StarlarkAction)) {
-//          ExecutionInfoSpecifier executionInfoSpecifier = (ExecutionInfoSpecifier) metadata;
-//          ImmutableMap<String, String> filteredExecutionInfo = TargetUtils.getFilteredExecutionInfo(
-//                  executionInfoSpecifier.getExecutionInfo(),
-//                  rule,
-//                  this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation());
-//
-//          Map<String, String> executionInfoBuilder = new HashMap<>();
-//          // in case something was already in action.executionInfo - we copy it to the final Map
-//          executionInfoBuilder.putAll(executionInfoSpecifier.getExecutionInfo());
-//          // adding filtered execution requirements to the execution info map
-//          executionInfoBuilder.putAll(filteredExecutionInfo);
-//
-//          ActionAnalysisMetadata updatedAction = executionInfoSpecifier.addExecutionInfo(ImmutableMap.copyOf(executionInfoBuilder));
-//          newList.add(updatedAction);
-//        } else {
-//          newList.add(metadata);
-//        }
-//      }
-//    } catch (InterruptedException e) {
-//      // todo: what?
-//      e.printStackTrace();
-//      throw new RuntimeException("Something horrible happened!");
-//    }
-//    getAnalysisEnvironment().registerAction(newList.toArray(new ActionAnalysisMetadata[0]));
     getAnalysisEnvironment().registerAction(action);
   }
 
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -102,6 +102,8 @@ import com.google.devtools.build.lib.util.OrderedSetMultimap;
 import com.google.devtools.build.lib.util.StringUtil;
 import com.google.devtools.build.lib.vfs.FileSystemUtils;
 import com.google.devtools.build.lib.vfs.PathFragment;
+
+import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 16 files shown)


src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
```diff
@@ -125,6 +125,15 @@
 public final class RuleContext extends TargetContext
     implements ActionConstructionContext, ActionRegistry, RuleErrorConsumer {
 
+  public boolean isAllowTagsPropagation() {
+    try {
+      return this.getAnalysisEnvironment().getSkylarkSemantics().experimentalAllowTagsPropagation();
+    } catch (InterruptedException e) {
+      // fallback to false if value of experimentalAllowTagsPropagation cannot be fetched
+      return false;
```
buchgr: InterruptedException does not mean that the value can't be fetched but that something (i.e. ctrl+c) told it to stop fetching. I think the proper thing to do here is to add a `throws InterruptedException`.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -28,12 +25,12 @@
...
 import com.google.devtools.build.lib.syntax.Type;
 import com.google.devtools.build.lib.util.Pair;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+
 import javax.annotation.Nullable;
+import java.util.*;
```
buchgr: for future: add the google-java-style in your IDE.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -234,6 +231,23 @@ private static boolean hasConstraint(Rule rule, String keyword) {
...
+  /**
+   * Returns the execution info from the tags declared on the target. These include only some tags
+   * {@link #legalExecInfoKeys} as keys with empty values.
+   *
+   * @param rule a rule instance to get tags from
+   * @param allowTagsPropagation if set to true, tags will be propagated from a target to the
+   *        actions' execution requirements, for more details {@see
+   *        SkylarkSematicOptions#experimentalAllowTagsPropagation}
+   */
+  public static Map<String, String> getExecutionInfo(Rule rule, boolean allowTagsPropagation) {
```
buchgr: this method is only called in three different sides. Remove it and instead use ternary operator at the call side?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
```diff
@@ -1548,6 +1554,9 @@ private CppCompileActionBuilder initializeCompileAction(Artifact sourceArtifact)
     builder.setCcCompilationContext(ccCompilationContext);
     builder.setCoptsFilter(coptsFilter);
     builder.setFeatureConfiguration(featureConfiguration);
+    if (ruleContext != null) {
```
buchgr: if (ruleContext != null && ruleContext.isAllowTagsPropagation()) {
  builder.addExecutionInfo(TargetUtils.getExecutionInfo(ruleContext.getRule()))
}

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -141,19 +143,21 @@ public CcLinkingOutputs getCcLinkingOutputs() {
    * @param ccToolchain the C++ toolchain provider for the build
    * @param fdoContext the C++ FDO optimization support provider for the build
    * @param configuration the configuration that gives the directory of output artifacts
+   * @param ruleContext
```
buchgr: add a comment here explaining ruleContext

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -124,15 +126,15 @@ public CcLinkingOutputs getCcLinkingOutputs() {
   private final ActionRegistry actionRegistry;
   private final RuleErrorConsumer ruleErrorConsumer;
   private final SymbolGenerator<?> symbolGenerator;
+  @Nullable private final RuleContext ruleContext;
```
buchgr: why nullable?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -827,17 +832,21 @@ private CppLinkActionBuilder newLinkActionBuilder(
...
+            .setTestOrTestOnlyTarget(isTestOrTestOnlyTarget)
+            .setLinkType(linkType)
+            .setLinkerFiles(
+                    (cppConfiguration.useSpecificToolFiles()
+                            && linkType.linkerOrArchiver() == LinkerOrArchiver.ARCHIVER)
+                            ? ccToolchain.getArFiles()
+                            : ccToolchain.getLinkerFiles())
+            .setLinkArtifactFactory(linkArtifactFactory)
+            .setUseTestOnlyFlags(useTestOnlyFlags);
+    if (ruleContext != null) {
```
buchgr: if (ruleContext != null && ruleContext.isAllowTagsPropagation()) {
  ...
}

src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
```diff
@@ -1432,7 +1432,7 @@ public boolean isCcToolchainResolutionEnabled(SkylarkRuleContext skylarkRuleCont
                     .getActionConstructionContext()
                     .getConfiguration()
                     .getFragment(CppConfiguration.class),
-                ((BazelStarlarkContext) starlarkContext).getSymbolGenerator())
+                ((BazelStarlarkContext) starlarkContext).getSymbolGenerator(), null)
```
buchgr: /* ruleContext= */ null. Also in other places.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
```diff
@@ -148,6 +148,7 @@ public Artifact create(
   //  of them.
   private boolean isTestOrTestOnlyTarget;
   private boolean isStampingEnabled;
+  private Map<String, String> executionInfo = new LinkedHashMap<>();
```
buchgr: can this be final?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
```diff
@@ -1022,15 +1023,15 @@ public CppLinkAction build() throws InterruptedException, RuleErrorException {
     // If the crosstool uses action_configs to configure cc compilation, collect execution info
     // from there, otherwise, use no execution info.
     // TODO(b/27903698): Assert that the crosstool has an action_config for this action.
-    Map<String, String> executionRequirements = new LinkedHashMap<>();
+//    Map<String, String> executionRequirements = new LinkedHashMap<>();
```
buchgr: remove?

src/test/shell/bazel/BUILD
```diff
@@ -767,6 +767,14 @@ sh_test(
     tags = ["no_windows"],
 )
 
+sh_test(
+    name = "tags_propagation_native_test",
```
buchgr: @hlopko should this test be part of the bazel_rules_cc_test.sh?

src/test/shell/bazel/tags_propagation_native_test.sh
```diff
@@ -0,0 +1,177 @@
...
+  cat > test/test.cc <<EOF
+#include <iostream>
+int main() { std::cout << "Hello test!" << std::endl; return 0; }
+EOF
+
+  bazel aquery --experimental_allow_tags_propagation '//test:test' > output1 2> $TEST_log \
+      || fail "should have generated output successfully"
+
+  if [[  "${PLATFORM}" = "darwin"  ]]; then
+    assert_contains "ExecutionInfo: {local: '', no-cache: '', no-remote: '', requires-darwin: ''}" output1
```
buchgr: Why is 'requires-darwin' relevant to this change?

src/test/shell/bazel/tags_propagation_native_test.sh
```diff
@@ -0,0 +1,177 @@
...
+  else
+    assert_contains "ExecutionInfo: {local: '', no-cache: '', no-remote: ''}" output1
```
buchgr: I think it would be nicer if the test didn't depend on the order. Consider

assert_contains "local: ''"
assert_contains "no-cache: ''"
assert_contains "no-remote: ''"

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
ishikhman: makes sense. Fixed.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
ishikhman: fixed :)

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
ishikhman: done

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -234,6 +231,23 @@ private static boolean hasConstraint(Rule rule, String keyword) {
...
+    if (allowTagsPropagation){
+      return getExecutionInfo(rule);
+    } else {
+      return Collections.emptyMap();
```
hlopko: suuuper nit: ImmutableMap.of() and a change in signature to return ImmutableMap. Or inline as Jakob suggested.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -352,7 +352,7 @@ public static ConfiguredTarget init(CppSemantics semantics, RuleContext ruleCont
                 semantics,
                 featureConfiguration,
                 ccToolchain,
-                fdoContext)
+                fdoContext, ruleContext)
```
hlopko: We try hard not to pass RuleContext into CcCompilationHelper and CcLinkingHelper because of its huge powers (e.g. it allows you to access rule attributes which is something we don't want to allow in our helpers to keep our code layering clean). Can we instead use ActionConstructionContext that is already passed to CcLinkingInfo? If not, let's pass executionInfo, not rule context, into the constructor of the Helper?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
```diff
@@ -276,16 +278,17 @@ public CcCompilationContext getCcCompilationContext() {
...
+          ActionConstructionContext actionConstructionContext,
+          Label label,
+          @Nullable Artifact grepIncludes,
+          CppSemantics semantics,
+          FeatureConfiguration featureConfiguration,
+          SourceCategory sourceCategory,
+          CcToolchainProvider ccToolchain,
+          FdoContext fdoContext,
+          BuildConfiguration buildConfiguration,
+          @Nullable RuleContext ruleContext) {
```
hlopko: Why nullable?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
ishikhman: done

src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
hlopko: Why is null here ok? You can get to action construction context using actions.getActionConstructionContext() there.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
hlopko: Nit: Use ImmutableMap.builder()

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
ishikhman: done

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
ishikhman: done

src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
ishikhman: I'll check that!

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
ishikhman: Because it can be null in case of sandwich, that we discussed offline yesterday. See this line: https://github.com/bazelbuild/bazel/pull/9267/files/49ca1c4f3364a34a398a03b39b29a83552ce1e69#r318975035

src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
ishikhman: yes, done

src/test/shell/bazel/BUILD
hlopko: I actually think we don't need an integration test for this, wdyt about adding an unit test to CcLibraryConfiguredTargetTest? We're testing analysis-only behavior anyway. You'll assert that the action in constructed as expected.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
ishikhman: done

src/test/shell/bazel/BUILD
hlopko: (and also a test to CcBinaryConfiguredTargetTest, GenruleConfiguredTargetTest...)

src/test/shell/bazel/tags_propagation_native_test.sh
ishikhman: because if you run a test on iOS platform, then testRunner (or smth like that) will add "requires-darwin" to the actions' execution requirements.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
ishikhman: removed the method :)

src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
ishikhman: Because it can be null in case of sandwich, like we understood it with Jakob yesterday. See: https://github.com/bazelbuild/bazel/pull/9267/files/49ca1c4f3364a34a398a03b39b29a83552ce1e69#r318975035

I'll also have a look at ActionConstructionContext, maybe we could use that => this @Nullable will not be needed at all, as well as RuleContext

src/test/shell/bazel/BUILD
ishikhman: hm, we could do that. I just prefer integration tests in this case - to make sure that in 'real' env it actually works. I also find them more readable here :)

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
ishikhman: hm, it might work! 
I've replaced RuleContext directly with a Rule and pushed isAllowTagsPropagation up to the ActionConstructionContext, @hlopko ptal?

src/test/shell/bazel/BUILD
hlopko: Well, you find them more readable because you didn't spend much time in the area :) We actually like our BuildViewTestCases more becasue they allow us to inspect the state of the world more easily. Instead of comparing handwritten line of text that aquery outputs you can have assertThat(executionInfo).containsExactly(...).

Plus the integration test will have exactly same coverage as the BuildViewTestCase.

ishikhman:
@hlopko ok, I'll have a look at the tests and BuildViewTestCases :) 

What do you think about changes from RuleContext -> Rule and ActionCreationContext?

Issue #9176, opened 2019-08-14
Windows: glob(["**"]) in external root is non-recursive


moroten:
### Description of the problem

`glob(["**"])` inside a `new_local_repository()` does not search subdirectories on Windows, it only reports files in the root directory. The external repository seems to be created with all root files copied in, including the `BUILD.bazel` file, and all subdirectories linked using junctions (`dir | select FullName,LinkType,Target` in Powershell tells me it's a junction).

I would expect `glob([**])` to search subdirectories on both Windows and Linux.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
mkdir -p my-external/subdir/inner
touch my-external/root.txt
touch my-external/subdir/outer.txt
touch my-external/subdir/inner/inner.txt
echo 'new_local_repository(name = "myext", path = "my-external", build_file = "BUILD.myext")' > WORKSPACE
echo 'filegroup(name = "all_files", srcs = glob(["**"]))
filegroup(name = "subdir_files", srcs = glob(["subdir/**"]))' > BUILD.myext
```

It fails in the root on Windows:
```
$ bazel query 'deps(@myext//:all_files)'
@myext//:all_files
@myext//:root.txt
@myext//:WORKSPACE
@myext//:BUILD.bazel
```
It works as expected in the root on Linux:
```
$ bazel query 'deps(@myext//:all_files)'
@myext//:all_files
@myext//:subdir/outer.txt
@myext//:subdir/inner/inner.txt
@myext//:root.txt
@myext//:WORKSPACE
@myext//:BUILD.bazel
```
The subdirectory works on both Windows and Linux:
```
$ bazel query 'deps(@myext//:subdir_files)'
@myext//:subdir_files
@myext//:subdir/outer.txt
@myext//:subdir/inner/inner.txt
```

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

release 0.28.1

###  Have you found anything relevant by searching the web?

Search GitHub for `glob+windows`, nothing relevant found.

laszlocsomor:
This is glob() bug with junctions on Windows, not an external repository bug.

Here's a simpler repro in the main workspace:
```
C:\>md foo\gh9176

C:\>cd foo\gh9176

C:\foo\gh9176>type nul > WORKSPACE

C:\foo\gh9176>echo filegroup(name='x',srcs=glob(['foo/**'])) > BUILD

[...]

PR #9269, opened 2019-08-28
Windows, VFS: stat treats junctions as symlinks

laszlocsomor:
FileStatus.isSpecialFile() no longer returns true
for junctions.

This allows FileSystem.direntFromStat(FileStatus)
to return SYMLINK instead of UNKNOWN for
junctions, allowing recursive glob() to traverse
junctions.

Fixes https://github.com/bazelbuild/bazel/issues/9176


laszlocsomor:
I finally found out why the test is failing: it runs "bazel build //..." twice, and the second build picks up the generated "bazel-<directory_name_of_workspace>" junction (now treated as a symlink, not as a special file anymore) and treats it as a package.

Here's a simple repro:
```
echo workspace(name="foo_ws") > WORKSPACE

echo genrule(name="x",outs=["x.txt"],cmd="echo $$PWD $@ > $@") > BUILD

bazel-dev.exe build //...
(...)

bazel-dev.exe build //...
FAILED: Build did NOT complete successfully (7 packages loaded)
    currently loading: bazel-gh4327-test/external/bazel_tools/third_party/def_parser
```

Somehow the same thing works on Linux, except when I rename the workspace directory after the first build, which then breaks the second build on Linux too. (Still trying to understand why.)

Issue #7386, opened 2019-02-08
Deprecate automatic creation of empty `__init__.py` files


limdor:
### Description of the problem / feature request:

Bazel is creating __init__.py files automatically. This is an issue in several situations (see https://github.com/bazelbuild/rules_python/issues/55).

This behavior can be suppressed using:
https://docs.bazel.build/versions/master/be/python.html#py_test.legacy_create_init

But in case that you have a large code base with a lot of python rules this is not a good solution. It would mean that this flag needs to be set in every single rule in case that you have a folder in the root of the workspace with the same name as a python package.
When this happens, it is not so easy to figure out what is going on. For modern Python developers __init__.py files are like source, these are not generated files. Thus, is not intuitive that the error could be because some init files are automatically generated.

### Feature requests: what underlying problem are you trying to solve with this feature?

In a first step: There should be an option to decide the behavior of legacy_create_init globally.
In a second step: The default option should be not to generate __init__.py files by default.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Same like https://github.com/bazelbuild/rules_python/issues/55 or any setup that at the root of the workspace you have a folder called like a builtin python module (time, logging, platform, queue, etc.)

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

release 0.22.0

brandjon:
The end state shouldn't be a global option but rather removing the legacy create init behavior altogether. This might look like the following.

1. Go through an incompatible change migration to set the default value of `legacy_create_init` to false.

2. Then go through an incompatible change migration to force the value to false for everyone, effectively making the attribute a no-op.

3. Then go through an incompatible change migration to delete the redundant attribute.

brandjon:
The first step of the above is implemented in @Faqa's PR #9271, which is in the process of being merged. It introduces an incompatible change flag tracked by #10076.

Steps 2 and 3 above can come after that. They can actually be merged into a single step. But that'll come down the line, since I don't anticipate flipping the first flag for a while. (We need to see how difficult the migration is, and if it requires new tooling, that'll be hard to prioritize.)

We can use this bug to track the overall deprecation effort.

brandjon:
A concern came up in internal review of #9271: How will a user know that a failure at execution time (from missing `__init__.py`) is due to flipping this flag?

[...]

PR #9271, opened 2019-08-28
Add 'remove_legacy_create_init_py' flag

Faqa:
Does the first step detailed by @brandjon  in #7386, adding a migration flag to set the value of `legacy_create_init` to False.


hlopko:
I'm sure Jon will take good care of this PR :)

brandjon:
Thank you for this PR! The implementation looks good but I'd like to request a change to how this behavior is rolled out.

There are three levels of behavioral change to consider:

1. Flipping the default value of `legacy_create_init` from true to false.
2. Making `legacy_create_init` a no-op (always false regardless of what the target says).
3. Deleting `legacy_create_init` (error if target tries to specify it).

This PR skips (1) and goes directly to implementing (2) but not (3). Skipping (1) means that users can't incrementally migrate their builds -- they can't temporarily opt out of the change for one or two stubborn targets, even if everything else in their build works. Skipping (3) means a separate incompatible change will still be needed just to cleanup the no-op attribute.

I suggest that instead we structure it so that this PR implements (1). Then a later change (to be flipped after Bazel 2.0!) could skip (2) and go directly to (3).

The biggest barrier I see is that migration for many builds may be cumbersome, since you have to add a `__init__.py` file to each Python package and include it in srcs. But in any case, we can implement this now and worry about whether migration is feasible yet later.

Faqa:
@brandjon  - I've re-implemented the change in line with your comment. I did have to change the type of "legacy_create_init" from Boolean to TriState, but the tests verify that anyone who has set the attribute using existing boolean values will still have it work. 

I tried to utilize the codebase's existing "default value" mechanism instead, but the amount of private classes and methods there made me think this was not an encouraged method.

Faqa:
@brandjon - ping.

brandjon:
Thanks for the changes. I'm merging this internally with a few tweaks:

- Factored PyExecutable logic into a static helper method
- Renamed the flag (and corresponding fields) to `--incompatible_default_to_explicit_init_py`
- Moved tests to BazelPyBinaryConfiguredTargetTest.java

Awaiting internal code review, expect the push in a day or two.

Issue #9284, opened 2019-08-29
Bazel 0.29 contains a regression with Remote Build Execution backend


v-mr:
### Description of the problem / feature request:

Bazel 0.29.0 (https://github.com/bazelbuild/bazel/issues/8572) has regression with Remote Build Execution backend.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

ERROR: <redacted>/test/BUILD:205:5:  failed (Exit 34). Note: Remote connection/protocol failed with: execution failed FAILED_PRECONDITION: Precondition check failed.
 
### What operating system are you running Bazel on?

Linux 4.19.37-5+deb10u1rodete1-amd64 #1 SMP Debian 4.19.37-5+deb10u1rodete1 (2019-07-22 > 2018) x86_64 GNU/Linux

### What's the output of `bazel info release`?

release 0.29.0

katre:
I am investigating with @v-mr

lizan:
We're seeing same issue in envoyproxy/envoy#8074, seems this only happens when `--nocache_test_results` is passed to bazel. I can reproduce it with `--nocache_test_results` but not without it.

katre:
Thank you for the data point, @lizan.

@v-mr has given me a repoduction case and I am bisecting now.

katre:
@lizan It's possible that the difference with `--nocache_test_results` is that when it is not set the tests aren't executed, which seems to be where the failure is.

lizan:
> It's possible that the difference with --nocache_test_results is that when it is not set the tests aren't executed, which seems to be where the failure is.

@katre I doubt so because I changed the our RBE image SHA at same time, which should revoke all cache in the first time when I ran without `--nocache_test_results`.

katre:
Bisect has eventually told me that the culprit is e4ccba45376f74d24ce3506fb4057e1268f01a59. I have no idea why that is, but reverting it from master causes my reproduction to be fixed.

I'm going to debug this further to try and determine _why_ this change causes this failure.

katre:
That commit is the one that caused the action key to change, which is apparently what causes the bug to be visible. It's not the commit that actually causes the error.

buchgr:
Trying to repro. PRECONDITION_FAILED hints at inputs not being uploaded.

buchgr:
[...]

PR #9287, opened 2019-08-30
Fix a serious regression in remote execution

buchgr:
Action inputs would not be uploaded to the CAS for targets tagged
with 'no-remote-cache' or 'no-cache'.

The regression was introduced by 8860c3ee71e6ee6aeb216720276e0dcaeb6deca2


katre:
Please make sure to note that this is for #9284 in the description.

PR #9295, opened 2019-08-30
Include cc configure headers in the cache key

hlopko:
This PR adds a file containing builtin include directories paths as an input to every C++ compilation action. The result is that whenever these headers change (as a result of reconfiguring `cc_configure` repository rule), we include this file in the action cache key.

Fixes https://github.com/bazelbuild/bazel/issues/9296.


meteorcloudy:
Wait.. it's failing on Windows. But only because of some syntax error, please fix.

hlopko:
Oki, importing :P

PR #9298, opened 2019-08-30
WIP: Have remote_outputs=toplevel download runfiles. Fixes #8899

### LINKED ISSUE #8899: support run command for experimental_remote_download_outputs

experimental_remote_download_outputs does not properly download all runtime dependencies.

Attempting to use the new remote builds without the bytes (#6862)

https://github.com/phb/bazel_minimal_missingrunfiles
and specifically see 
https://github.com/phb/bazel_minimal_missingrunfiles/blob/master/EXAMPLE_OUTPUT

macosx.

Tried various versions including latest HEAD as of today.

This was initially reported in #6862 and split into it's own ticket. CC @buchgr

buchgr:
This change makes it so that when building a binary remotely and
having --experimental_remote_download_toplevel set the runfiles of the
binary will also be downloaded. This applies also to building a test
target. However, when executing a test remotely with bazel test neither
the test binary nor its runfiles are downloaded. Instead Bazel fetches
the test result: test.xml and test.log

This change also enables bazel run to be used with
--experimental_remote_download_toplevel.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
@@ -49,7 +49,7 @@ final class RemoteActionContextProvider extends ActionContextProvider {
   private final DigestUtil digestUtil;
   @Nullable private final Path logDir;
   private final AtomicReference<SpawnRunner> fallbackRunner = new AtomicReference<>();
-  private ImmutableSet<ActionInput> topLevelOutputs = ImmutableSet.of();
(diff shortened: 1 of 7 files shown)


buchgr:
@phb

phb:
Tested this and it works great for my usecase. thank you!

buchgr:
@SrodriguezO I believe you also need `bazel run`. Would be great if you could test it too!

SrodriguezO:
@buchgr that's awesome! Haven't had a chance to try it out yet, but if worked for others I imagine it's good. Thanks for doing this!

PR #9301, opened 2019-08-30
Small documentation fixes for bazel emscripten tutorial.

bzztbomb:
- Fix error message in step 4, it differs from what is display when following the tutorial.
- Remove cpu from cc_toolchain, it's not needed in newer versions of Bazel.


dstiers-1109:
thinks

dstiers-1109:
Thanks

PR #9302, opened 2019-09-01
Update ios-app.md to load git_repository

kateinoigakukun:


dstiers-1109:
lets do this

dstiers-1109:
lets do this

PR #9310, opened 2019-09-03
fewer words, easier reading

nicoaragon:
Proposing grammar fixes.

### DIFF
--- a/site/docs/skylark/language.md
+++ b/site/docs/skylark/language.md
@@ -7,21 +7,13 @@ title: Starlark Language
 
 <!-- [TOC] -->
 
-The page is an overview of [Starlark](https://github.com/bazelbuild/starlark)
-(formerly known as Skylark), the language used in Bazel. This should be
-enough to get you started. Please refer to the
-[Starlark repository on GitHub](https://github.com/bazelbuild/starlark/)
-for more information about the language.
-For a complete list of functions and types, please check the
-[API reference](lib/skylark-overview.html).
+The page is an overview of [Starlark](https://github.com/bazelbuild/starlark), formerly known as Skylark, the language used in Bazel. For a complete list of functions and types, check [Starlark's API reference](lib/skylark-overview.html).
+
+For more information about the language, see [Starlark's GitHub repo](https://github.com/bazelbuild/starlark/).
 
 ## Syntax
 
-Starlark is designed to be small, simple, and thread-safe. Although it is
-inspired from Python, it is not a general-purpose language and most Python
-features are not included.
-
-Here is an example to show the syntax:
+Starlark's syntax is inspired by Python3. This is valid syntax in Starlark:
 
 ```python
 def fizz_buzz(n):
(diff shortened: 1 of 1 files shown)


site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
-[list](lib/list.html), [string](lib/string.html). On top of that, two new
-types are specific to Bazel: [depset](lib/depset.html) and
-[struct](lib/struct.html).
-
-Starlark is syntactically a subset of Python 3, and will remain so through at
-least the 1.x release lifecycle. This ensures that Python-based tooling can at
-least parse Starlark code. Although Starlark is not *semantically* a subset of
-Python, behavioral differences are rare (excluding cases where Starlark raises
-an error).
+Starlark's semmantics can differ from Python, and behavioral differences are rare, except for cases where Starlark raises an error. The following Python types are supported:
```
laurentlb: semantics

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
-In a build, there are many "evaluation contexts": each `.bzl` file and each
-`BUILD` file is loaded in a different context. Each rule is also analyzed in a
-separate context. We allow side-effects (e.g. appending a value to a list or
-deleting an entry in a dictionary) only on objects created during the current
-evaluation context. Once the code in that context is done executing, all of its
-values are frozen.
+This is because Bazel builds use parallel execution. During a build, each `.bzl` file and each `BUILD` file get their own execution context. Each rule is also analyzed in its own context.
 
-For example, here is the content of the file `foo.bzl`:
+Let's go through an example. Let the below be the content of the file `foo.bzl`:
```
laurentlb: This sounds weird to me: "Let the below be the content" (but I'm not native speaker)

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
+Bazel creates `var` when `foo.bzl` loads. `var` is thus part of `foo.bzl`'s context. When `fct()` runs, it does so within the context of `foo.bzl`. After evaluation for `foo.bzl` completes, the environment contains an immutable entry, `var`, with the value `[5]`.
 
-It is possible for multiple other files to load symbols from `foo.bzl` at the
-same time. For this reason, the following code is not legal:
+When another, `bar.bzl` loads symbols from `foo.bzl`, loaded values remain immutable. For this reason, the following code in `bar.bzl` is illegal:
 
 ```python
-load(":foo.bzl", "var", "fct")
+# `bar.bzl`
+load(":foo.bzl", "var", "fct") #loads `var`, and `fct` from `./foo.bzl`
```
laurentlb: space after `#`

site/docs/skylark/language.md
nicoaragon: fixed. Thank you!

site/docs/skylark/language.md
nicoaragon: fixed. Thank you!

nicoaragon:
Thank you, @laurentlb !

What are next steps?

laurentlb:
Thanks for the changes, and for the PR. I like the simpler phrasing.

I am going import the change. If another reviewer has comments, they'll add them here; otherwise your PR will be merged and closed.

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
-[list](lib/list.html), [string](lib/string.html). On top of that, two new
-types are specific to Bazel: [depset](lib/depset.html) and
-[struct](lib/struct.html).
-
-Starlark is syntactically a subset of Python 3, and will remain so through at
-least the 1.x release lifecycle. This ensures that Python-based tooling can at
-least parse Starlark code. Although Starlark is not *semantically* a subset of
-Python, behavioral differences are rare (excluding cases where Starlark raises
-an error).
+Starlark's semantics can differ from Python, and behavioral differences are rare, except for cases where Starlark raises an error. The following Python types are supported:
```
brandjon: Nit: "and behavioral differences are rare" -> "but behavioral difference are rare"

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
 
-The variable `var` is created when `foo.bzl` is loaded. `fct()` is called during
-the same context, so it is safe. At the end of the evaluation, the environment
-contains an entry mapping the identifier `var` to a list `[5]`; this list is
-then frozen.
+Bazel creates `var` when `foo.bzl` loads. `var` is thus part of `foo.bzl`'s context. When `fct()` runs, it does so within the context of `foo.bzl`. After evaluation for `foo.bzl` completes, the environment contains an immutable entry, `var`, with the value `[5]`.
 
-It is possible for multiple other files to load symbols from `foo.bzl` at the
-same time. For this reason, the following code is not legal:
+When another, `bar.bzl` loads symbols from `foo.bzl`, loaded values remain immutable. For this reason, the following code in `bar.bzl` is illegal:
```
brandjon: Nit: "When another," -> "When another" (delete comma)

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
 fct()          # runtime error, fct() attempts to modify a frozen list
 ```
 
-Evaluation contexts are also created for the analysis of each custom rule. This
-means that any values that are returned from the rule's analysis are frozen.
-Note that by the time a custom rule's analysis begins, the .bzl file in which
-it is defined has already been loaded, and so the global variables are already
-frozen.
-
+Just like the above example using `bzl` files, values returned by rules are immutable too.
```
brandjon: I'd preserve the caveat that rules can't mutate globals in the same file. Maybe add the sentence:

"Since a rule function runs after the .bzl file it's in has already been loaded, rules can't update global variables."

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
-
+There are two syntactic restrictions in `BUILD` files: 1) declaring functions is illegal, and 2) `*args` and `**kwargs` arguments are not allowed.
 
 ## Differences with Python
 
-In addition to the mutability restrictions, there are also differences with
-Python:
-
-* Global variables cannot be reassigned.
+* Global variables are immutable.
```
brandjon: This can be misread as saying that you can't do

    a = []
    a.append(1)

at the global level, which is actually allowed. Maybe keep the current wording for this point.

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
 
-* `for` statements are not allowed at the top-level; factor them into functions
-  instead. In BUILD files, you may use list comprehensions.
+* `for` statements are not allowed at the top-level. Use them within functions instead. In BUILD files, you may use list comprehensions.
```
brandjon: "you may use list comprehensions, but consider using macros instead for readability."

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
+* In tuples, a trailing comma is valid only when the tuple is between parentheses, e.g. write `(1,)` instead of `1,`.
 
-* Dictionary literals cannot have duplicated keys. For example, this is an
-  error: `{"a": 4, "b": 7, "a": 1}`.
+* Dictionary literals cannot have duplicated keys. For example, this is an error: `{"a": 4, "b": 7, "a": 1}`.
 
-* Variable of a comprehension may not be used after the comprehension. This is
-  stricter than Python 2 and Python 3, which have different behavior (shadowing
-  vs reassignment).
+* The variable used in a comprehension may not be used after the comprehension. This is stricter than Python 2 and Python 3, which have different behavior (shadowing vs reassignment).
```
brandjon: The Starlark semantics here are pretty messy, so we need to be careful if we want to be technically accurate.

First, let's distinguish between a variable and an identifier. Python 3 says you can't access the comprehension's variable after the comprehension, but you can still access a different variable that happens to share the same identifier. Starlark is stricter by saying you can't even do that.

Second, this restriction only applies to local function variables. You can still access a global that occurs after the comprehension.

So a possible rewording might be:

"If a function contains both a local variable `x`, and a comprehension that has a local variable `x` (which shadows the outer one), then the name `x` cannot be referenced after the comprehension. This is stricter than both Python 2 and Python 3."

Whether it's worth saying all that here, I'm not sure.

site/docs/skylark/language.md
```diff
@@ -37,142 +29,97 @@ def fizz_buzz(n):
...
 
-* Strings are represented with double-quotes (e.g. when you
-  call [repr](lib/globals.html#repr)).
+* Strings are represented with double-quotes (e.g. when you call [repr](lib/globals.html#repr)).
```
brandjon: "Strings are conventionally represented with"

This is to help avoid implying that the literal syntax `s = 'abc'` is invalid.

site/docs/skylark/language.md
nicoaragon: thank you! changed to "but".

site/docs/skylark/language.md
nicoaragon: Thank you! deleted it!

site/docs/skylark/language.md
nicoaragon: Thank you! I added the sentence.

site/docs/skylark/language.md
nicoaragon: Great callout! I updated the statement to make it accurate.

site/docs/skylark/language.md
nicoaragon: I agree with the clarification. I updated the statement accordingly.

site/docs/skylark/language.md
nicoaragon: thank you! I updated it.

nicoaragon:
@brandjon can you re-open my pull request? I tried but I do not have permissions.

laurentlb:
Sorry - that's a race condition. I noticed Jon's comments too late.
The change has been merged. Can you create a new PR?

nicoaragon:
No problem! I just did: see [PR 9326](https://github.com/bazelbuild/bazel/pull/9326). Thank you!

PR #9311, opened 2019-09-03
Fix Starlark name typo in guide

mantasak:


mantasak:
@googlebot I signed it!

PR #9314, opened 2019-09-03
Java stub perf improvements

anchlovi:
When the `classpath` is too long the stub writes it to a manifest file

At Wix it is not unusual to find tests targets with a very long `classpath`. For a test target with 1.1MB long `classpath` and with over 2600 entries it takes over 55 seconds to construct and write the manifest file with the current stub.

With the suggested fix it takes ~20 millis to construct manifest file for the same target


ittaiz:
Thanks @anchlovi! @iirina can you help with this?

Issue #8581, opened 2019-06-07
--experimental_remote_download_outputs breaks symlink outputs


philwo:
### Description of the problem / feature request:

When using `--experimental_remote_download_outputs` no matter whether `=minimal` or `=toplevel`, Bazel fails to build rules_go with these errors:

```
(13:01:33) ERROR: /workdir/tests/core/go_path/BUILD.bazel:5:2: output 'tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/cmd/bin.a' is a dangling symbolic link
(13:01:33) ERROR: /workdir/tests/core/go_path/BUILD.bazel:5:2: output 'tests/core/go_path/link_path/pkg/plan9_arm/example.com/repo/cmd/bin.a' is a dangling symbolic link
(13:01:33) ERROR: /workdir/tests/core/go_path/BUILD.bazel:5:2: output 'tests/core/go_path/link_path/pkg/linux_amd64/example.com/repo/pkg/lib.a' is a dangling symbolic link
(13:01:33) ERROR: /workdir/tests/core/go_path/BUILD.bazel:5:2: Couldn't build file tests/core/go_path/link_path/src/example.com/repo/cmd/bin/bin.go: not all outputs were created or valid
```

Failing build with the flag: https://buildkite.com/bazel-testing/rules-go-golang/builds/181
Working build without the flag: https://buildkite.com/bazel-testing/rules-go-golang/builds/182

### What operating system are you running Bazel on?

Fails on Linux and macOS.

### What's the output of `bazel info release`?

Bazel 0.26.1

philwo:
@buchgr Friendly ping. This is blocking the „remote builds without bytes“ rollout on CI.

coeuvre:
Is this still an issue today?

PR #9315, opened 2019-09-03
Disable runfile tree building when --remote_download_minimal

buchgr:
Since we don't download any outputs there's no point in building the runfile tree. After this change there should no longer be dangling symlinks in the output base. It can also improve performance if symlink tree creation is slow.


ishikhman:
Related to #8581

PR #9320, opened 2019-09-03
Call FindMissingBlobs before BES artifact upload. Fixes #8673, #7630

### LINKED ISSUE #8673: RE combined with BES is slow with large output artifacts

buchgr:
Bazel unconditionally uploads files referenced by BEP events to the remote cache. This change makes it so that Bazel calls FindMissingBlobs() before uploading a file.

This should drastically improve BES performance for when files already exist remotely (i.e. when using remote execution).

A side effect of this change is that it also moves calls to potentially blocking Path#getDigest() on the thread pool.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/AbstractRemoteActionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/AbstractRemoteActionCache.java
@@ -85,18 +85,13 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
-/**
- * A cache for storing artifacts (input and output) as well as the output of running an action.
- */
+/** A cache for storing artifacts (input and output) as well as the output of running an action. */
 @ThreadSafety.ThreadSafe
-public abstract class AbstractRemoteActionCache implements AutoCloseable {
+public abstract class AbstractRemoteActionCache implements MissingDigestsFinder, AutoCloseable {
 
-  /**
-   * See {@link SpawnExecutionContext#lockOutputFiles()}.
-   */
+  /** See {@link SpawnExecutionContext#lockOutputFiles()}. */
   @FunctionalInterface
   interface OutputFilesLocker {
-
     void lock() throws InterruptedException;
   }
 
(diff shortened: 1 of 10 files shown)


buchgr:
@arlyon can you review please? Ideally, test it our for your use case?

arlyon:
> @arlyon can you review please? Ideally, test it our for your use case?

I'll run some checks on it and review. Aiming for 4pm UTC.

arlyon:
Code looks good, local testing with smaller builds show no immediate issues. Currently running a larger build that should take around 30-60 mins so I'll update as that runs.

buchgr:
@arlyon can you please reply with `@googlebot I consent.` as the @googlebot god wants from us?

arlyon:
Hey @buchgr, having some problems (unrelated to the patch) in our test env. What is the latest you need this done by?

buchgr:
@arlyon I believe we'll cut tomorrow end-of-day. the earlier the better :).

buchgr:
@arlyon can you please reply with the googlebot command anyway? Else I won't be able to credit you with this change.

arlyon:
@googlebot I consent.

arlyon:
Hey, we're happy on our end. Thanks for helping us get this through.

buchgr:
Thanks @arlyon!

buchgr:
@arlyon has signed the cla according to https://github.com/bazelbuild/bazel/pull/9031. I also manually looked up the author's e-mail in the CLA list. He also responded with `@googlebot I consent` but the googlebot did not recognize it.

PR #9330, opened 2019-09-05
Fix destructive mv of bazel internal files

AustinSchuh:
The genrule was moving a $SRC.  In some sandboxes, ths is a permission
problem.  On Debian Stretch, this results in the following error:

ERROR: /home/austin/local/bazel2/scripts/packages/BUILD:57:2: Executing
genrule //scripts/packages:zip-bazel-exe_nojdk failed (Exit 1) bash
failed: error executing command /bin/bash -c ... (remaining 1
argument(s) skipped)

touch: cannot touch '/tmp/tmp.EhwgH9R4/bazel.exe': Permission denied

So let's copy instead.

### DIFF
--- a/scripts/packages/BUILD
+++ b/scripts/packages/BUILD
@@ -61,7 +61,7 @@ genrule(
     cmd = "\n".join([
         "tmpdir=$$(mktemp -d $${TMPDIR:-/tmp}/tmp.XXXXXXXX)",
         "trap \"rm -fr $${tmpdir}\" EXIT",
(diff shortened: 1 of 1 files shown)


philwo:
I'm curious - isn't this target only built on Windows (which doesn't have sandboxing yet)?

AustinSchuh:
I'm pretty certain it is.  Either that, or I still don't know what the list of targets is.  I pulled the list of targets out of the bazelci yaml, and confirmed it by looking at the command lines executed as part of a CI build.

PR #9331, opened 2019-09-05
updated documentation to clarify TLS support 

ishikhman:
updated documentation to clarify TLS support for --remote_cache, --remote_executor and --bes_backend

Follow-up of #8061

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
@@ -34,6 +34,8 @@ public class BuildEventServiceOptions extends OptionsBase {
     help =
         "Specifies the build event service (BES) backend endpoint as HOST or HOST:PORT. "
             + "Disabled by default."
+            + "The supported schemas are grpc and grpcs (grpc with TLS enabled). "
+            + "If no schema is provided bazel'll default to grpcs. Specify grpc:// schema to disable TLS."
   )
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
```diff
@@ -34,6 +34,8 @@
     help =
         "Specifies the build event service (BES) backend endpoint as HOST or HOST:PORT. "
             + "Disabled by default."
+            + "The supported schemas are grpc and grpcs (grpc with TLS enabled). "
+            + "If no schema is provided bazel'll default to grpcs. Specify grpc:// schema to disable TLS."
```
agoulti: This seems to imply that bes_backend should default to grpcs, but from what I've seen it doesn't.
Where's the disconnect?

ishikhman: It is, starting from version 1.0 (it is not a part of 0.29, I can see that there was a confusion around this, sorry for that). Before 1.0, in order to get this behaviour one have to specify `--incompatible_tls_enabled_removed` flag.
Starting 1.0 this won't be necessary.
see #8061 for details.

agoulti: Ooh, I missed that! Yes, it works with Bazel 1.0 release candidate!

Issue #8216, opened 2019-05-01
--disk_cache overrides --noremote_upload_local_results


JaredNeil:
### Description of the feature request:

I would like to be able to use the `--disk_cache` as a read/write cache, but the remote cache as a read-only cache. I thought `--noremote_upload_local_results` would do this, but it seems that flag is just overridden by the `--disk_cache` option. It seems this is the expected behavior, since in #7512, I see it says:
>On put, Bazel will store items in both the disk and the HTTP cache.

Could this be changed to support this mixed use case?

### What underlying problem are you trying to solve with this feature?

We are still in the process of moving our build to bazel, and not all builds on dev machines are safe to be pushed to the shared remote cache, so we only populate the remote cache from CI builds. Being able to have a local cache is very helpful when switching between dev branches for super fast builds. Our current workaround is to run a remote cache on every machine that acts as a proxy to the real shared cache, but we would like to move away from all of that overhead.

buchgr:
I agree. disk cache should always be read/write and the `remote_upload_local_results` flag should only apply to remote caching.

SrodriguezO:
Hey @ishikhman! Have you had a chance to look any further into this?

buchgr:
@SrodriguezO we love contributions :)

SrodriguezO:
We're happy to help! Just wanted to avoid duplicate efforts, since the issue's assigned. We can start looking into it next week.

SrodriguezO:
We'll still hold off to hear back from @ishikhman though (to avoid duplicate efforts)

ishikhman:
> We'll still hold off to hear back from @ishikhman though (to avoid duplicate efforts)

unfortunately, not yet. That would be really great if you could look into it! Removed myself from the assignment ;)

SrodriguezO:
Silly me.. I just realized this won't actually solve our problem :( 

We're exploring a move to RBE and are hoping to utilize its remote cache, but it isn't compatible with local disk caching in the first place :/

buchgr:
@SrodriguezO it's just not been implemented yet. It should be a straightforward PR. Would also love your contribution :)

SrodriguezO:
@buchgr I dug around a bit and it seems a bit involved. I opened a draft PR to test things out as well as a feature request to formalize the process and to get some input. I'd love your feedback. https://github.com/bazelbuild/bazel/issues/8690

keith:
@buchgr any chance we could mark this as a blocker for 1.0?

[...]

PR #9338, opened 2019-09-05
Fix noremote_upload_local_results behavior in combined blobstore

borkaehw:
Disable the upload action to remote cache when `--noremote_upload_local_results` is set.

This problem would only happen in `CombinedDiskHttpBlobStore` at this moment when users want read/write permission to disk cache but only read permission to remote cache.

Once the GrpcCache refactoring (making GrpcCache a SimpleBlobStore implementation) is done, `CombinedDiskHttpBlobStore` should be renamed to `CombinedDiskRemoteBlobStore`. Then this fix will apply to both GrpcCache and HttpCache.

Fix https://github.com/bazelbuild/bazel/issues/8216

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
@@ -114,15 +114,10 @@ final class RemoteSpawnCache implements SpawnCache {
   public CacheHandle lookup(Spawn spawn, SpawnExecutionContext context)
       throws InterruptedException, IOException, ExecException {
     if (!Spawns.mayBeCached(spawn)
-        || (!Spawns.mayBeCachedRemotely(spawn) && isRemoteCache(options))) {
+        || (!Spawns.mayBeCachedRemotely(spawn) && useRemoteCache(options))) {
       // returning SpawnCache.NO_RESULT_NO_STORE in case the caching is disabled or in case
       // the remote caching is disabled and the only configured cache is remote.
       return SpawnCache.NO_RESULT_NO_STORE;
-    }
-    boolean checkCache = options.remoteAcceptCached;
-
-    if (checkCache) {
-      context.report(ProgressStatus.CHECKING_CACHE, "remote-cache");
     }
 
     SortedMap<PathFragment, ActionInput> inputMap = context.getInputMapping(true);
@@ -150,7 +145,9 @@ final class RemoteSpawnCache implements SpawnCache {
         TracingMetadataUtils.contextWithMetadata(buildRequestId, commandId, actionKey);
 
     Profiler prof = Profiler.instance();
-    if (checkCache) {
+    if (options.remoteAcceptCached ||
+        (options.incompatibleRemoteResultsIgnoreDisk && useDiskCache(options))) {
+      context.report(ProgressStatus.CHECKING_CACHE, "remote-cache");
       // Metadata will be available in context.current() until we detach.
       // This is done via a thread-local variable.
       Context previous = withMetadata.attach();
@@ -211,7 +208,8 @@ final class RemoteSpawnCache implements SpawnCache {
 
     context.prefetchInputs();
 
-    if (options.remoteUploadLocalResults) {
+    if (options.remoteUploadLocalResults ||
+        (options.incompatibleRemoteResultsIgnoreDisk && useDiskCache(options))) {
       return new CacheHandle() {
         @Override
         public boolean hasResult() {
@@ -297,7 +295,11 @@ final class RemoteSpawnCache implements SpawnCache {
     }
   }
 
-  private static boolean isRemoteCache(RemoteOptions options) {
+  private static boolean useRemoteCache(RemoteOptions options) {
     return !isNullOrEmpty(options.remoteCache);
+  }
+
+  private static boolean useDiskCache(RemoteOptions options) {
+    return options.diskCache != null && !options.diskCache.isEmpty();
   }
 }
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -144,6 +144,25 @@ public final class RemoteOptions extends OptionsBase {
   public boolean remoteUploadLocalResults;
 
   @Option(
+      name = "incompatible_remote_results_ignore_disk",
+      defaultValue = "false",
+      category = "remote",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      metadataTags = {
+        OptionMetadataTag.INCOMPATIBLE_CHANGE,
+        OptionMetadataTag.TRIGGERED_BY_ALL_INCOMPATIBLE_CHANGES
+      },
+      help =
+          "If set to true, --noremote_upload_local_results and --noremote_accept_cached "
+              + "will not apply to the disk cache. "
+              + "If a combined cache is used:"
+              + "\n\t--noremote_upload_local_results will cause results to be written to the disk cache, but not uploaded to the remote cache."
+              + "\n\t--noremote_accept_cached will result in Bazel checking for results in the disk cache, but not in the remote cache."
+              + "\nSee #8216 for details.")
+  public boolean incompatibleRemoteResultsIgnoreDisk;
+
+  @Option(
       name = "remote_instance_name",
       defaultValue = "",
       documentationCategory = OptionDocumentationCategory.REMOTE,
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java
@@ -47,11 +47,12 @@ public final class RemoteCacheClientFactory {
       PathFragment diskCachePath,
       boolean remoteVerifyDownloads,
       DigestUtil digestUtil,
-      RemoteCacheClient remoteCacheClient)
+      RemoteCacheClient remoteCacheClient,
+      RemoteOptions options)
       throws IOException {
     DiskCacheClient diskCacheClient =
         createDiskCache(workingDirectory, diskCachePath, remoteVerifyDownloads, digestUtil);
-    return new DiskAndRemoteCacheClient(diskCacheClient, remoteCacheClient);
+    return new DiskAndRemoteCacheClient(diskCacheClient, remoteCacheClient, options);
   }
 
   public static ReferenceCountedChannel createGrpcChannel(
@@ -159,7 +160,7 @@ public final class RemoteCacheClientFactory {
 
     RemoteCacheClient httpCache = createHttp(options, cred, digestUtil);
     return createDiskAndRemoteClient(
-        workingDirectory, diskCachePath, options.remoteVerifyDownloads, digestUtil, httpCache);
+        workingDirectory, diskCachePath, options.remoteVerifyDownloads, digestUtil, httpCache, options);
   }
 
   public static boolean isDiskCache(RemoteOptions options) {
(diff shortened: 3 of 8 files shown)


buchgr:
@borkaehw Thanks for the patch! I agree that we should have this but it's incompatible behavior. You need to make it so that this new behavior is guarded behind an incompatible flag that is disabled by default.

ishikhman:
@borkaehw here is a link for the policy with some instructions: https://bazel.build/breaking-changes-guide.html

buchgr:
Hmm. I am wondering if it would make more sense if we gave the disk cache it's own knobs and introduced `--disk_write_local_results` and `--disk_accept_cached` instead?

ishikhman:
> Hmm. I am wondering if it would make more sense if we gave the disk cache it's own knobs and introduced `--disk_write_local_results` and `--disk_accept_cached` instead?

Hm, do you know a use case for that?

SrodriguezO:
I can't think of a strong use case for using the disk cache as read-only. To disable it altogether, you might as well not specify a `--disk_cache` in the config. 

In any case, I think this would fall outside the scope of this PR, which is to resolve #8216 (and the changes to `remote_upload_local_results`'s behavior would have to be done anyway).

borkaehw:
I have made the incompatible flag and written a better commit message. I suppose GitHub issue should be created after we merge. Let me know if I did anything wrong. Thanks.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -139,6 +139,23 @@
...
+      defaultValue = "false",
+      category = "remote",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      metadataTags = {
+        OptionMetadataTag.INCOMPATIBLE_CHANGE,
+        OptionMetadataTag.TRIGGERED_BY_ALL_INCOMPATIBLE_CHANGES
+      },
+      help =
+          "If set to true, when --noremote_upload_local_results and "
```
SrodriguezO: Suggested rephrasing:

"If set to true, --noremote_upload_local_results will not apply to the disk cache. If a combined cache is used, --noremote_upload_local_results will cause results to be written to the disk cache, but not uploaded to the remote cache. See #8216 for details."

borkaehw: Thanks.

src/main/java/com/google/devtools/build/lib/remote/disk/CombinedDiskHttpBlobStore.java
```diff
@@ -39,17 +40,21 @@
...
     this.remoteCache = Preconditions.checkNotNull(remoteCache);
+    this.options = options;
   }
 
   @Override
   public void putActionResult(ActionKey actionKey, ActionResult actionResult)
       throws IOException, InterruptedException {
     diskCache.putActionResult(actionKey, actionResult);
-    remoteCache.putActionResult(actionKey, actionResult);
+    if (!options.incompatibleUploadLocalResultsDiskOnly || options.remoteUploadLocalResults) {
```
ishikhman: I find the following version a bit more readable, but it's up to you: to change it or not :)
 `if (options.incompatibleUploadLocalResultsDiskOnly && !options.remoteUploadLocalResults)`

buchgr:
> In any case, I think this would fall outside the scope of this PR, which is to resolve #8216 (and the changes to remote_upload_local_results's behavior would have to be done anyway).

I think it's important to be keep the behaviour consistent. This change makes it so that --remote_upload_local_results not longer applies to --disk_cache whether a --remote_cache is specified or not. So I think to stay sane it we should introduce the same behaviour for --remote_accept_cached?

SrodriguezO:
>I think it's important to be keep the behaviour consistent

That's a fair point. 

Personally, I can't think of a use case for this, but I don't have a strong opinion either way. @borkaehw wdyt?

src/main/java/com/google/devtools/build/lib/remote/disk/CombinedDiskHttpBlobStore.java
SrodriguezO: I agree with @ishikhman here; not a strong preference, but it seems a bit cleaner

SrodriguezO: You'd need to negate the entire clause for those to be equivalent: 
`!(options.incompatibleUploadLocalResultsDiskOnly && !options.remoteUploadLocalResults)`

I think either way works, but I prefer the original : )

borkaehw: I agree with @SrodriguezO, also prefer the original. More suggestions are welcome.

buchgr:
> Personally, I can't think of a use case for this, but I don't have a strong opinion either way. 

I don't think we necessarily need the flags that I proposed but I think that we'll need to also disable `--remote_accept_cached` for the `--disk_cache` along with `--remote_upload_local_results`.

Does that sound reasonable?

SrodriguezO:
>but I think that we'll need to also disable --remote_accept_cached for the --disk_cache along with --remote_upload_local_results

Oh I see. Yeah we should definitely do that.

borkaehw:
@buchgr thanks for suggestions, it sounds reasonable to me. I will take a closer look.

borkaehw:
Just pushed changes to address comments, please take a look. Thanks.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
```diff
@@ -118,11 +118,6 @@ public CacheHandle lookup(Spawn spawn, SpawnExecutionContext context)
       // the remote caching is disabled and the only configured cache is remote.
       return SpawnCache.NO_RESULT_NO_STORE;
     }
-    boolean checkCache = options.remoteAcceptCached;
-
-    if (checkCache) {
```
SrodriguezO: How come you removed this?

SrodriguezO: hmm I see that you moved the progress status down to the other block, but will the logic between here and there cause a delay?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
```diff
@@ -206,7 +203,8 @@ public CacheHandle lookup(Spawn spawn, SpawnExecutionContext context)
 
     context.prefetchInputs();
 
-    if (options.remoteUploadLocalResults) {
+    if (options.remoteUploadLocalResults ||
+        (options.incompatibleRemoteResultsDiskOnly && options.diskCache != null && !options.diskCache.isEmpty())) {
```
SrodriguezO: Can we move the `options.diskCache != null && !options.diskCache.isEmpty())` part to a helper `usesDiskCache` method?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -139,6 +139,26 @@
       help = "Whether to upload locally executed action results to the remote cache.")
   public boolean remoteUploadLocalResults;
 
+  @Option(
+      name = "incompatible_remote_results_disk_only",
```
SrodriguezO: Can we rename to `incompatible_remote_results_flags_ignore_disk`? The current name makes me think this flag makes remote results flags apply to the disk cache only (which is the opposite of what they do)

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
borkaehw: It passes the tests. I don't see a strong reason to keep it separate. I might be wrong. @ishikhman do you have more context on it?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -139,6 +139,26 @@
...
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      metadataTags = {
+        OptionMetadataTag.INCOMPATIBLE_CHANGE,
+        OptionMetadataTag.TRIGGERED_BY_ALL_INCOMPATIBLE_CHANGES
+      },
+      help =
+          "If set to true, --noremote_upload_local_results and --noremote_accept_cached "
+              + "will not apply to the disk cache. "
+              + "If a combined cache is used, --noremote_upload_local_results will cause "
```
SrodriguezO:
```
+ "If a combined cache is used: "
+ "\n\t--noremote_upload_local_results will cause results to be written to the disk cache, but not uploaded to the remote cache"
+ "\n\t--noremote_accept_cached will result in Bazel checking for results in the disk cache, but not the remote cache."
+ "See #8216 for details."
```

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
borkaehw: Good idea.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
borkaehw: Sounds good. I prefer just `incompatible_remote_results_ignore_disk` to make it a bit shorter.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
borkaehw: Done.

ulfjack:
Ping @buchgr 

I think there might be people who use a read-only disk cache on NFS, which is written by CI, and read by everyone else.

borkaehw:
@ishikhman friendly ping here.

I think this is valuable for people. I am happy to address feedback and pushing it toward merged.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
ishikhman: I don't see a bit difference to be honest. Should be fine both ways.

ishikhman:
@borkaehw still LGTM :) 
Pinging @buchgr for merging part

SrodriguezO:
@buchgr friendly ping : )

SrodriguezO:
@buchgr can you please merge this when you have a minute? It was approved nearly a month ago and is just pending merge. Or maybe @ishikhman can merge it? We've been using this on our fork without issue, and we'd hate to see it rot.

borkaehw:
Since https://github.com/bazelbuild/bazel/pull/9200 was merged. Force pushed to address conflicts.

borkaehw:
Since #10200 was merged. Force pushed to address conflicts.

kastiglione:
Just wanted to chime in and say we've been waiting for many releases (since 0.17 or earlier), hoping for this fix soon. Another friendly ping @buchgr :)

borkaehw:
Since #10233 was merged. Force pushed to address conflicts.

Also added tests for combined disk and grpc cache.

borkaehw:
Not sure why `//src/test/shell/bazel:bazel_proto_library_test` is failing. It doesn't seem to be related to my change.

borkaehw:
Force push to fix `//src/test/shell/bazel:bazel_proto_library_test`.

borkaehw:
@buchgr friendly ping. Thanks for making grpc-disk combined cache #10233, it improves our developer workflow dramatically. This fix is the final piece blocking us from using official release of Bazel (we have carried this commit on custom Bazel for a long time). This simple PR will bring a huge value to the community. Would you be able to take a quick look?

borkaehw:
This PR breaks the CI on `master`, but not the CI on PRs. Here is the fix. https://github.com/bazelbuild/bazel/pull/10462

Issue #9327, opened 2019-09-04
[Security] netrc support passes original credentials on HTTP redirects


keith:
### Description of the problem / feature request:

When you have a `~/.netrc` file with these contents:

```
machine github.com
  login something
  password something
```

Downloads of GitHub tarballs fail with this error:

```
skylib.0.8.0.tar.gz: GET returned 400 Bad Request
ERROR: no such package '@bazel_skylib//': java.io.IOException: Error downloading [https://github.com/bazelbuild/bazel-skylib/releases/download/0.8.0/bazel-skylib.0.8.0.tar.gz] to 
```

If you copy the curl request of what bazel is doing, you can see why:

```
% curl -H 'User-Agent: Bazel/release 0.29.0' -H 'Authorization: Basic somebase64' -H 'Host: github.com' -H 'Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' 'http://github.com/bazelbuild/bazel-skylib/releases/download/0.8.0/bazel-skylib.0.8.0.tar.gz' -L
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidArgument</Code><Message>Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified</Message><ArgumentName>Authorization</ArgumentName><ArgumentValue>token anything</ArgumentValue><RequestId>8A1908B7B4F61B0B</RequestId><HostId>G3CxaBga4CoVWwqf1fu4WMrr2LjLJ28s2H8+Kn90PFFFygw8KKQ6+JINoeUTipnpbtbcxmKhr9A=</HostId></Error>
```

Removing the Authorization header from the request fixes this issue. Switching the `Basic` auth to `bearer` or `token` and providing that directly doesn't solve this issue.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

`bazel build //... --repository_cache=`

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 0.29.0

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/pull/7978

keith:
cc @genrym

jin:
cc @aehlig

dslomov:
cc @laurentlb this looks like a release blocker for 0.29 /cc #9293

katre:
[...]

Issue #9293, opened 2019-08-30
Release 0.29.1 - August 2019 (stable)


katre:
A patch release is needed after 0.29.0 (#8572) to address the following issues:

- #9284 - Failures in remote execution when tests are not cached
- *TO BE FILED* - Regression in git_repository.

hlopko:
Can we consider a fix for https://github.com/bazelbuild/bazel/issues/9296 to be cherrypicked into 0.29.1?

katre:
Yes, I think #9296 is valid.

jmillikin-stripe:
Filed https://github.com/bazelbuild/bazel/issues/9299 for the git_repository regression from v0.28

katre:
@jmillikin-stripe Thanks!

katre:
Since the fix for #9296 isn't ready, and at this point in the day, before a US holiday weekend, I am going to wait until Tuesday (Sept 3) to cut the first RC for 0.29.1.

katre:
Planning to cut RC1 today. Assuming everything works, we can release on Thursday.

katre:
Release command:
```
$ ./scripts/release/release.sh create --force_rc=1 0.29.1 0.29.0 2c046486f5febb9e475a5589b59abae77f14245e 9e1d65a1fcebff7169a63acad95c7b44302b89c4 8b0bfaf6716327ca6b60692704e3d57766f7ab1f
```

katre:
Release announcement sent: https://releases.bazel.build/0.29.1/rc1/index.html

Waiting for tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1179

katre:
Downstream tests all pass (after several retries).

keith:
FYI it seems like a new feature in 0.29.0 broke github archive downloads in some cases https://github.com/bazelbuild/bazel/issues/9327 I'm not sure if this warrants a revert or not

buchgr:
@katre 

An important internal project found a serious bug in remote execution where output files would sometimes not get deleted between builds. In this particular instance we found this to lead to zombie test results. That is test cases showing up as failed that no longer exist in the codebase. I have fixed this problem in master for different reasons not understanding the impact of this bug: https://github.com/bazelbuild/bazel/commit/5c02b92c45b5422971b3164685d9edb771367a1b

katre:
@buchgr Please file a github issue so I can track the reason for the cherrypick.

[...]

PR #9341, opened 2019-09-05
Revert "Make http_archive honor $HOME/.netrc"

katre:
This reverts commit c26e3396cb341e173b4be5c780cafac3ec386df5.

Work on #9327.

### DIFF
--- a/src/test/shell/bazel/skylark_repository_test.sh
+++ b/src/test/shell/bazel/skylark_repository_test.sh
@@ -1852,41 +1852,4 @@ EOF
       || fail "Expected success despite needing a file behind basic auth"
 }
 
-function test_implicit_netrc() {
-  mkdir x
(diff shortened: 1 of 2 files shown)


katre:
This attempts to mitigate #9327 so that 0.29.1 (#9293) can be released. Once that is done the code can be restored and fixed.

katre:
I cannot reproduce the actual issue, can @keith or @davido please verify this fixes the problem?

keith:
I can confirm this fixes the issue for me

davido:
I can confirm, that after applying this revert, the line in ~/.netrc:

```
machine github.com login foo password bar
```

is ignored, and `bazel-dev build headless` in gerrit just works.

katre:
Dropping this PR since the issue was fixed separately.

PR #5525, opened 2018-07-06
Add aarch64 embedded JDK config, MacOS mock toolchain settings

ArielleA:
Hi, 

This PR is for additional work to get the tests running correctly on Aarch64. One key issue was that the default embedded jvm used by default was X86_64 and does not work on other architectures.
To support this I have added a new constraint, however I will refine this in case a version of Darwin becomes available on aarch64.

I'll do the work to find a good arm32 toolchain in a future CL. I do not have access for ppc or z390 to repair those architectures.

Either way, hope this is good for you.

PR #9346, opened 2019-09-06
fix build for Power

clnperez:
The zip -d flag will error if it doesn't find that extension.
This is a fix for Power, but doesn't fix the default case.

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

### DIFF
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -11,6 +11,12 @@ filegroup(
 )
 
 config_setting(
+    name = "linux_ppc",
+    values = {"cpu": "ppc"},
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
     name = "linux_x86_64",
     values = {"cpu": "k8"},
     visibility = ["//visibility:public"],
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -528,6 +528,7 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     # The .so file is an x86 one, so we can just remove it if the CPU is not x86
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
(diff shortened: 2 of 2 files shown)


gregestren:
Has the default case never worked, then?

clnperez:
>Has the default case never worked, then?

@gregestren probably not. that code was introduced 4 months ago. since it's new-ish maybe hasn't been hit until now.

src/conditions/BUILD
```diff
@@ -10,6 +10,12 @@ filegroup(
     visibility = ["//src:__pkg__"],
 )
 
+config_setting(
+    name = "linux_ppc",
```
gregestren: Should`ppc` should imply `Linux`? I see there's already [precedent](https://github.com/bazelbuild/bazel/pull/5525/files#diff-5aeb8330f796a1fbec8885e37f3ca5b9R56) for `aarch64` so I guess it's not a huge deal.

clnperez:
did this get merged? i'm not sure why it's closed

src/conditions/BUILD
clnperez: yah that's just a convention i've seen and stuck with. but you're right, for now bazel doesn't run on aix or i :)

jin:
yes: https://github.com/bazelbuild/bazel/commit/5cff4f1edf8b95bf0612791632255852332f72b5

We don't merge pull requests directly in this repo. They get imported to the internal review system and exported out as commits.

clnperez:
It looks like only one of those files got merged @jin 

https://github.com/bazelbuild/bazel/blob/master/third_party/BUILD#L531
doesn't have the power condition from https://github.com/clnperez/bazel/blob/99775df8356ab6c7dd77493b35e59de5f6a773f8/third_party/BUILD#L531

jin:
Ah, I see it.

third_party changes need to be broken out from PRs into a separate PR. Please open a new PR containing the change for third_party and I (or @gregestren) can import it manually./

PR #9352, opened 2019-09-09
Add documentation for creating "custom verbs" with Starlark macros

clintharrison:
This was originally submitted as https://github.com/bazelbuild/bazel-blog/pull/189/, this is merely moving it to the normal docs site rather than the blog.

cc @laurentlb @jin who commented on the original PR

### DIFF
--- a/site/docs/skylark/tutorial-custom-verbs.md
+++ b/site/docs/skylark/tutorial-custom-verbs.md
@@ -1,0 +1,156 @@
+---
+layout: documentation
+title: Using macros to create "custom verbs"
+---
+
+# Using macros to create "custom verbs"
+
+Day-to-day interaction with Bazel happens primarily through a few commands:
+`build`, `test`, and `run`. At times, though, these can feel limited: you may want to
+push packages to a repository, publish documentation for end-users, or deploy
+an application with Kubernetes. But Bazel doesn't have a `publish` or
+`deploy` command – where do these actions fit in?
+
+## The versatility of `bazel run`
+
+Bazel's focus on hermeticity, reproducibility, and incrementality means the
+`build` and `test` commands aren't helpful the above tasks. These actions may
+run in a sandbox, with limited network access, and aren't guaranteed to be
+re-run with every `bazel build`.
+
+Instead, we rely on `bazel run`: the workhorse for tasks we *want* to have side
+effects. Bazel users are accustomed to rules that create executables, and rule
+authors can follow a common set of patterns to extend this to "custom verbs".
+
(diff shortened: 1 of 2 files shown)


kastiglione:
Thanks for documenting this.

site/docs/skylark/tutorial-custom-verbs.md
```diff
@@ -0,0 +1,157 @@
...
+an additional "docs.publish" target is also created, which builds a script for
+publishing the site. Once we check the output of the primary target, we can use
+`bazel run :docs.publish` to publish it for public consumption, just like an
+imaginary `bazel publish` command.
+
+
+It's not immediately obvious what the implementation of the `_sphinx_publisher`
+rule might look like. Often, actions like this write a _launcher_ shell script.
+This method typically involves using
+[`ctx.actions.expand_template`](https://docs.bazel.build/versions/0.28.0/skylark/lib/actions.html#expand_template)
```
laurentlb: Use a relative link: `lib/actions.html#expand_template`

laurentlb: It looks like having an existing rule for this would be useful. I've filed https://github.com/bazelbuild/bazel-skylib/issues/191 for this.

site/docs/skylark/tutorial-custom-verbs.md
```diff
@@ -0,0 +1,157 @@
...
+together.
+
+
+In `rules_k8s`, this is indeed what `.apply` does:
+[`expand_template`](https://github.com/bazelbuild/rules_k8s/blob/f10e7025df7651f47a76abf1db5ade1ffeb0c6ac/k8s/object.bzl#L213-L241)
+writes a very simple Bash script, based on
+[`apply.sh.tpl`](https://github.com/bazelbuild/rules_k8s/blob/f10e7025df7651f47a76abf1db5ade1ffeb0c6ac/k8s/apply.sh.tpl),
+which runs `kubectl` with the output of the primary target. This script can
+then be build and run with `bazel run :staging.apply`, effectively providing a
+`k8s-apply` command for `k8s_object` targets.
```
laurentlb: Ideally, we would have an example on https://github.com/bazelbuild/examples/ and link it from here.

clintharrison: I agree that would be nice to have! Do you consider that a blocker for merging this documentation change, since it'll have to be two PRs anyway?

laurentlb: Doesn't have to be a blocker. We can always update documentation later.

PR #9371, opened 2019-09-11
Add explicit --sdk argument to xcrun calls

keith:
Xcode 11 (and possibly earlier versions) has a bug where, when you don't
pass `-sdk macosx`, it can either pick the global
`/Library/Developer/CommandLineTools` SDK path, or the SDK path bundled
within Xcode itself. This seems to depend on if you have other versions
of Xcode installed, and which versions of the command line tools you
have installed. Passing `-sdk macosx` seems to always force it to pick
the Xcode bundled version, which is what we prefer. This fixes issues
with remote cache misses caused by these tools not being the same
because of this difference. At least 2 apple bugs have been filed
about this FB7147521 FB7253366

You can see which path it's picking by running:

```
env -i xcrun -l -n clang
```


kastiglione:
Since this affects cache, can it be picked into the next release (1.0)?

Issue #9364, opened 2019-09-10
Error messaging when BUILD file not present at top level directory is unclear


nlopezgi:
### Description of the problem / feature request:

`go_repository` now reads a configuration file, which, by default is WORKSPACE in the main workspace. 
The implementation of this feature is such that the error messaging when the file is not present is as follows: 
```
ERROR: Analysis of target '//src/hack/project:image' failed; build aborted: no such package '@com_github_google_go_containerregistry//pkg/v1': Unable to load package for //:WORKSPACE: BUILD file not found in any of the following directories.
 - /local/path/to/my/git/repository
```
This error message is already causing issues for rules_docker customers, as now this BUILD file is needed in all projects that depend transitively on gazelle. 

Please see https://github.com/bazelbuild/bazel-gazelle/issues/609#issuecomment-530017922

Also related: 
bazelbuild/rules_docker#1139
bazelbuild/rules_docker#1126

### Feature requests: what underlying problem are you trying to solve with this feature?

I would like to have better messaging for this type of errors.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Depend on gazelle (latest version) from a project that does not have a root BUILD file

fyi @laurentlb

aehlig:
> `BUILD file not found at the top level directory`
> would be a good start

Such a message without context would give the wrong impression that there were a requirement to have a `BUILD` file next next to the `WORKSPACE` file, which, however, is not a requirement. It is perfectly fine to not have a package for the empty string, as long as no one references the empty string as a package name. In that respect, "top level" is no different than any other package name: only those packages referenced somewhere have to exist.

Therefore, I also wouldn't call ...

> So user has to work out that the two first references above are misleading,

... the first two references misleading; instead they give the needed context on where the offending reference was found (keep in mind that `bazel` has no means of knowing whether your source layout does or does not intend to have a package with name `''`). It simply follows the standard pattern of all `bazel` error messages that a lot of users find useful.
- First it states which of targets `bazel` was asked to build failed; this information is not redundant, as the command line accepts target patters that can, in general, expand to several targets.
- Then it mentions the root cause, in this case a package depended upon that does not exist.
- Finally, it gives information on the root cause. In this case, why that package does not exist: it has a definition, but that definition depends on an invalid label `//:WORKSPACE`.
- It also explains why the label `//:WORKSPACE` is invalid: that label requires the package `''` to exist, but it found no `BUILD` file for that package.
[...]

PR #9385, opened 2019-09-13
Improve package loading error message on directories without BUILD files

jin:
Let's inform the user that BUILD files turn directories into packages.

Fixes https://github.com/bazelbuild/bazel/pull/9364


indera-shsp:
Does this mean a `BUILD.bazel` or a `BUILD` file?

PR #9388, opened 2019-09-16
Readd support for authentication and .netrc

aehlig:
Now that the root cause of #9327 is fixed in e38d83871b39ed5f69cad903fa97e90b47a59875.

This reverts commit a0e3bb207fe2044120a2555a37162ee1f2b17500.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
@@ -80,6 +80,7 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


aehlig:
cc @keith 

Potentially also relevant Bazel 1.0 (#8573)

dslomov:
cc @artem-zinnatullin

keith:
Works for me!

artem-zinnatullin:
Hm, I think there are issues with new implementation (I might be wrong tho), left few comments on the commit https://github.com/bazelbuild/bazel/commit/e38d83871b39ed5f69cad903fa97e90b47a59875

We better not rush it as it can either break functionality or create another CVE :)

aehlig:
> Hm, I think there are issues with new implementation[...]

As [mentioned on the commit](https://github.com/bazelbuild/bazel/commit/e38d83871b39ed5f69cad903fa97e90b47a59875#r35114101), I believe the mentioned issue can only lead to too few authentication headers being sent, which seems a safe change, and moreover, a same-host policy would interact strangely with the current `ctx.download` interface of expecting a map from URL to authentication information.

aehlig:
> [...] moreover, a same-host policy would interact strangely with the current `ctx.download` interface of expecting a map from URL to authentication information.

Just to clarify the tragedy here: the problem is the slightly(!) generalized interface. Simply passing a `.netrc` file, or the fully generalized interface of passing a function (that could, e.g., be generated from a `.netrc`) would both  allow (or in the first case even automatically force) a same-host policy. But a map from URLs to authentication credentials is on the one hand flexible enough to describe policies that use different credentials on the same host (which might even make sense in some cases), but on the other hand is not powerful enough to describe the full host-to-credential mapping that a `.netrc` file describes.

cc @dslomov

aehlig:
>[...] I believe the mentioned issue can only lead to too few authentication headers being sent, which seems a safe change [...]

I wonder, if the amount of authentication sent is enough for the original use case.

cc @ittaiz @genrym

PR #9395, opened 2019-09-16
Fix framework_search_paths var name

keith:
In the documentation this is showing up incorrectly because of the
variable name.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -21,7 +21,7 @@ import static com.google.devtools.build.lib.rules.objc.ObjcProvider.DEFINE;
 import static com.google.devtools.build.lib.rules.objc.ObjcProvider.DYNAMIC_FRAMEWORK_DIR;
 import static com.google.devtools.build.lib.rules.objc.ObjcProvider.DYNAMIC_FRAMEWORK_FILE;
 import static com.google.devtools.build.lib.rules.objc.ObjcProvider.FORCE_LOAD_LIBRARY;
-import static com.google.devtools.build.lib.rules.objc.ObjcProvider.FRAMEWORK_SEARCH_PATH_ONLY;
+import static com.google.devtools.build.lib.rules.objc.ObjcProvider.FRAMEWORK_SEARCH_PATHS;
 import static com.google.devtools.build.lib.rules.objc.ObjcProvider.FRAMEWORK_SUFFIX;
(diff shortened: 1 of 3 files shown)


keith:
😬 actually this value might be read now? so the write value is one thing from starlark but the read is another?

ob:
This started as a question of how to propagate `-F`'s. I looked at `rules_apple` and they use `framework_search_paths` [here](https://github.com/bazelbuild/rules_apple/blob/821ca56c920f88679c90a807e6d7bd071950d7f8/apple/internal/apple_framework_import.bzl#L154), but that attribute is not [in the documentation](https://docs.bazel.build/versions/0.29.1/skylark/lib/ObjcProvider.html) for ObjcProvider, however, `framework_search_path_only` [is there](https://docs.bazel.build/versions/0.29.1/skylark/lib/ObjcProvider.html#framework_search_path_only).

I discovered that you can't _read_ from `framework_search_paths`, but you can write to it, and you can't write to `framework_search_path_only` but you can _read_ from it.

So now my rule reads from one and writes to the other, which seems *wrong*. 

What is the intent with these two?

sergiocampama:
@googlewalt fyi

hlopko:
Can we have a test for this PR?

ulfjack:
The Starlark API is defined in ObjcProviderApi.java, not here, so this does not change the API. (It may be that this change then doesn't have the expected impact, but this rename seems like an improvement to me.)

keith:
Ok yea, this is just a variable naming improvement then. I think the general issue that @ob mentions is still the case, but I guess this is safe to merge

hlopko:
Ok, approving the import.

ob:
Since this just cleans up the variable name and doesn't address the original question, should I open a separate issue with the original question? 

https://github.com/bazelbuild/bazel/pull/9395#issuecomment-531986434

hlopko:
Only if you want to have it fixed, or to be super honest, to fix it yourself. In the long term, objc library will merge with cc library and the ObjcProvider api will dissapear. But we're making a very slow progress there.

googlewalt:
FWIW, here are some details on the current implementation:

The ObjcProvider Starlark API (ObjcProviderApi.java) is read-only.  There is no way to write to an existing ObjcProvider.  However, there is an API for creating a new one.  That interface is defined in AppleCommonApi.java and implemented in AppleSkylarkCommon.java.  The latter uses the names of the items listed in KEYS_FOR_SKYLARK in ObjcProvider.

It is definitely confusing that the read and write APIs don't use consistent names.  Probably the simplest fix is to add "framework_search_paths" to ObjcProviderApi.java, without removing "framework_search_paths_only," to avoid incompatibility change.

Issue #9390, opened 2019-09-16
sh_binary wrapper fails on upper case letters in file extension on Windows


aherrmann:
### Description of the problem:

Windows filenames are case-insensitive, and also can be shortened to 8.3 format, which is upper-case by default. The `sh_binary` wrapper fails with the following error when calling the binary with an upper-case file extension.
```
LAUNCHER ERROR: Cannot open the binary to read launch data
LAUNCHER ERROR: Failed to parse launch info.
```

This is an issue if the `sh_binary` is meant to be called by another program which uses upper-case filenames or 8.3 format. E.g. autotools `./configure` scripts.

The `sh_binary` wrapper should accept all combinations of upper and lower case in the file extension.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create a binary target as follows:
```
cat >sh-binary.sh <<EOF
echo "Hello world"
EOF
cat >BUILD <<EOF
sh_binary(
    name = "sh-binary",
    srcs = ["sh-binary.sh"],
)
EOF
```

Build and execute
```
> bazel build //:sh-binary

> bazel-bin\sh-binary.exe
Hello world

> BAZEL-~1\SH-BIN~1.exe
Hello world

> bazel-bin\sh-binary.EXE
LAUNCHER ERROR: Cannot open the binary to read launch data
LAUNCHER ERROR: Failed to parse launch info.

> BAZEL-~1\SH-BIN~1.EXE
LAUNCHER ERROR: Cannot open the binary to read launch data
LAUNCHER ERROR: Failed to parse launch info.

```

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

release 0.28.1

PR #9400, opened 2019-09-17
Windows, *_binary launcher: bazel-bin/x.EXE works

laszlocsomor:
Fixes https://github.com/bazelbuild/bazel/issues/9390


aherrmann:
@dslomov would it be possible to cherry-pick this fix into 1.0? We would like to depend on this fix in rules_haskell.

laszlocsomor:
@aherrmann : Sorry, 1.0 was released without this fix, but 1.1 will have it.

Issue #8536, opened 2019-05-31
--incompatible_strict_action_env breaks autodetecting Python toolchain on Mac


brandjon:
Enabling `--incompatible_strict_action_env` without providing an explicit `--action_env` causes `PATH` to default to `/bin:/usr/bin`.

https://github.com/bazelbuild/bazel/blob/e495e6b7afff5930e85a149f7809720b72199005/src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java#L169-L171

https://github.com/bazelbuild/bazel/blob/e495e6b7afff5930e85a149f7809720b72199005/src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java#L449-L454

On our Mac CI workers, `python3` is [installed via Homebrew](https://github.com/bazelbuild/continuous-integration/issues/578#issuecomment-490642532) in `/usr/local/bin`. I expect other Mac users would have similar configurations.

This causes downstream failures in my presubmit to flip `--incompatible_use_python_toolchains` for Bazel 0.27. The flag is being flipped anyway. I expect those projects will have to either modify their `--action_env` or somehow modify their buildkite CI to add a custom Python toolchain. 

+cc @philwo for CI, @buchgr for strict action env.

brandjon:
Gerrit breakage tracked [here](https://bugs.chromium.org/p/gerrit/issues/detail?id=10953). Note that their bazelrc contains

```
build --experimental_strict_action_env
build --action_env=PATH
```

I'm not sure why `/usr/local/bin` wouldn't be propagated through `PATH`. It should be in the default environment of the CI workers.

brandjon:
rules_nodejs breakage tracked in bazelbuild/rules_nodejs#809.

brandjon:
Gerrit [sets](https://gerrit.googlesource.com/gerrit/+/refs/heads/master/.bazelrc) `--incompatible_strict_action_env` but also passes in `--action_env=PATH`, so you'd think they'd just get the client's PATH. Yet the error message indicates that the action is running with the default strict action env PATH of `/bin:/usr/bin`.

@buchgr, any idea why that would be?

davido:
IIRC Gerrit sets `--incompatible_strict_action_env` to make dick cache reliably work across different checkout directories (~/projects/gerrit, ~/projects/gerrit2, ...).

buchgr:
[...]

PR #9402, opened 2019-09-18
Specify 2.7 instead of 2 for python version

keith:
This changes the python setup to prefer an executable named `python2.7`
over `python2`. This fixes an issue on macOS where it doesn't ship with
a `python2` executable, but you can install one from homebrew, but
really you'd prefer to be using the system `python2.7` binary. I assume
that this isn't a huge issue because bazel likely doesn't support
versions of python before 2.7.

Follow up to https://github.com/bazelbuild/bazel/commit/3a4be3c93813987a27a97dade3f9ebbc5770e349

### DIFF
--- a/tools/python/utils.bzl
+++ b/tools/python/utils.bzl
@@ -21,7 +21,7 @@ file is less likely to cause bootstrapping issues.
 
 def _expand_pyversion_template_impl(ctx):
     for output, version, strict in [
-        (ctx.outputs.out2, "2", "1"),
+        (ctx.outputs.out2, "2.7", "1"),
         (ctx.outputs.out3, "3", "1"),
         (ctx.outputs.out2_nonstrict, "2", "0"),
         (ctx.outputs.out3_nonstrict, "3", "0"),


keith:
cc @brandjon (we also discussed this in slack), this is meant to fix this issue https://github.com/bazelbuild/bazel/issues/8536#issuecomment-532472396

keith:
Another option in our case is to stop passing `--experimental_strict_action_env` since we set `PATH` explicitly with `--action_env=PATH` and `env -i PATH=... bazel`

keith:
Any thoughts on something like this @brandjon  ?

brandjon:
Hi Keith,

This change makes it so the autodetecting toolchain's PY2 runtime would 1) look for the command `python2.7` (instead of `python2`) on `PATH`, and 2) fallback on `python` if that can't be found (same as before). This feels a bit ad hoc. Currently we don't have a notion of distinguishing between Python minor versions except in the toolchain; why should the default toolchain single out 2.7?

To get precise control over what interpreter is picked, I'd use an explicit Python toolchain instead of relying on the default autodetecting one.

keith:
The core issue I'm trying to fix here is that `python2` isn't provided by macOS, so we have to use `python2.7` specifically. I figured singling out 2.7 was pretty safe since I assume support for <python 2.7 isn't really something we're worried about since folks likely aren't on older versions. We could totally switch to our own toolchain to work around this but this is a subtle bit of non-hermaticity described here https://github.com/bazelbuild/bazel/issues/8536#issuecomment-532472396 that I think it would be nice to fix for everyone

lzell:
Is there any harm in looking for the OS X system-installed `python2.7` binary first, then falling back to `python2` if not found?

ulfjack:
Ping @brandjon

aiuto:
ping @brandjon

keith:
I assume at this point this won't matter for much longer because python 2 support will be removed?

brandjon:
We're not removing support for building and running Python 2 code using Bazel. We've only removed the expectation that Bazel itself (and some language-specific tooling) will work with Python 2.

The issue with this PR is that it is a backwards-incompatible change that solves the problem in an ad hoc way, by changing the default for everyone, to accommodate a situation that is particular to a given environment and IMO better addressed by choosing an explicit toolchain.

(That said, there is some precedent for tweaking the Python interpreter location based on platform in #10432, although that case was for the stub script shebang, and would be eliminated by a future improvement to allow toolchains to specify the shebang string.)

aiuto:
Given the last comments, I'm assuming this PR is dead on arrival.
I'll close in a week if I don't hear from anyone.

keith:
FWIW this is still an issue. I haven't heard an alternative path to fixing this but we might want to. I assume with the current state of python2 at this point folks are feeling super motivated to support this though.

PR #9403, opened 2019-09-18
Add explicit version minimums for macOS tools

keith:
This is a follow up to https://github.com/bazelbuild/bazel/pull/9371
which helps maintain cache hits across different macOS versions. By
default when you compile something on macOS with clang the minimum OS
version is set to the current OS version. This means if you have
developers on multiple OS versions they may not get cache hits.

10.9 was chosen to match rules_apple https://github.com/bazelbuild/rules_apple/commit/86ce42562e784cc5cbd17aeb5f5a5334c856ba9b


keith:
cc @DavidGoldman

keith:
Thanks!

PR #9404, opened 2019-09-18
Update deprecated flag value

thii:
According to the User Manual, `standalone` is a deprecated value for the `--spawn_strategy` flag. This update the Remote Caching page to use the new value `local` instead.


aiuto:
Why are you suggesting this?
What is the context?
Please add that to the PR description.

thii:
Updated.

PR #9409, opened 2019-09-20
Bazel client: faster startup flag diff'ing

laszlocsomor:
Speed up computing the set of changed startup
flags.

Replace linear search (quadratic complexity) with
using an std::unordered_set (linear complexity).

Change-Id: Ic8caf00f2898a5e5d329cbdf79cb5507589254c8

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -1048,19 +1048,30 @@ static DurationMillis ExtractData(const string &self_path,
   }
 }
 
-// Returns true if the server needs to be restarted to accommodate changes
-// between the two argument lists.
-static bool AreStartupOptionsDifferent(
-    const vector<string> &running_server_args,
-    const vector<string> &requested_args) {
+static bool IsVolatileArg(const string& arg) {
   // TODO(ccalvarin) when --batch is gone and the startup_options field in the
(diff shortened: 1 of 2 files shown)


src/main/cpp/blaze.cc
```diff
@@ -1073,58 +1084,37 @@ static bool AreStartupOptionsDifferent(
     options_different = true;
   }
 
-  // Args in running_server_args that are not in requested_args.
-  bool found_missing_args = false;
+  std::unordered_set<string> missing_from_new;
+  vector<string> missing_from_old;
+
+  // Add all of 'running_server_args' to 'missing_from_new', then remove all of
```
philwo: I have to say, I don't really get this reasoning, but it looks very elegant, is less code than before and I'm also tired, so maybe this is fine. 😅

laszlocsomor: It's a roundabout way to compute the difference of two sets A and B. (Except the sets are given as vectors, although all elements should be unique, because they are the canonical set of startup flags which should not have duplicates.)

The logic adds all elements from `A` to a set, then removes all elements of `B` from the set. What remains in the set is `A \ B`. If an element could not be removed from the set, then it was not there, so I add those to a vector, which yields `B \ A`.

I'm adding a test to verify this.

Issue #8799, opened 2019-07-05
Windows, macOS: package names and input-file-target names are case-ignoring


laszlocsomor:
### Description of the problem / feature request:

Bazel does not enforce correct casing in package names, nor in target names (if they refer to input files).

Bazel only enforces correct casing in target names that are rules or output files. (Because these are defined as strings in the BUILD file.)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Repro on Windows -- in cmd.exe:
```
md c:\src\tmp\case\Foo\Bar

cd c:\src\tmp\case

type nul > WORKSPACE

echo Servus > Foo\Bar\Hello

echo genrule(name="x",srcs=["foo/bar/hello"],outs=["x.out"],cmd="(echo $< ; cat $< ;) > $@") > BUILD

bazel build //:x & type bazel-bin\x.out
(...)
INFO: Build completed successfully, 2 total actions
foo/bar/hello
Servus
```

Repro on macOS -- in Bash:
```
mkdir -p /tmp/case/Foo/Bar

cd /tmp/case

touch WORKSPACE

echo Servus > Foo/Bar/Hello

echo 'genrule(name="x",srcs=["foo/bar/hello"],outs=["x.out"],cmd="(echo $< ; cat $< ;) > $@")' > BUILD

bazel build //:x && cat bazel-bin/x.out
(...)
INFO: Build completed successfully, 2 total actions
foo/bar/hello
Servus
```

### What operating system are you running Bazel on?

Windows 10
macOS 10.15.1

### What's the output of `bazel info release`?

release 1.2.1

###  Have you found anything relevant by searching the web?

Same culprit as https://github.com/bazelbuild/bazel/issues/7627.

PR #9435, opened 2019-09-25
Windows: implement getCorrectCasing

laszlocsomor:
Implement getCorrectCasing: a function that
normalizes a path and returns the case-correct
version of it.

See https://github.com/bazelbuild/bazel/issues/8799

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/windows/jni/WindowsFileOperations.java
+++ b/src/main/java/com/google/devtools/build/lib/windows/jni/WindowsFileOperations.java
@@ -231,12 +231,31 @@ public class WindowsFileOperations {
     }
   }
 
-  public static String getCorrectCasing(String path) throws IOException {
+  /**
+   * Returns the case-corrected copy of <code>absPath</code>
+   *
+   * <p>Windows paths are case-insensitive, but the filesystem preserves the casing. For example you
+   * can create <code>C:\Foo\Bar</code> and access it as <code>c:\FOO\bar</code>, but the correct
+   * casing would be <code>C:\Foo\Bar</code>. Sometimes Bazel needs the correct casing (see
+   * https://github.com/bazelbuild/bazel/issues/8799 for example).
+   *
+   * <p>This method fixes paths to have the correct casing, up to the last existing segment of the
+   * path.
+   *
+   * @param absPath an absolute Windows path. May not be normalized (i.e. have <code>./</code> and
+   *                <code>../</code> segments) and may have <code>/</code> separators instead of
+   *                <code>\\</code>.
+   * @return an absolute, normalized Windows path. The path is case-corrected up to the last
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/devtools/build/lib/windows/jni/WindowsFileOperations.java
```diff
@@ -228,4 +230,14 @@ public static boolean deletePath(String path) throws IOException {
         throw new IOException(String.format("Cannot delete path '%s': %s", path, error[0]));
     }
   }
+
```
meisterT: Please a comment here that explains what "correct" refers to.

src/main/native/windows/file.cc
```diff
@@ -781,5 +781,46 @@ bool GetCwd(std::wstring* result, DWORD* err_code) {
   }
 }
 
+std::wstring GetCorrectCasing(const std::wstring& abs_path, bool with_unc) {
```
meisterT: this looks like an absolute path is required - I don't see this requirement on the Java side. Do I miss something?

src/main/native/windows/file.cc
```diff
@@ -781,5 +781,46 @@ bool GetCwd(std::wstring* result, DWORD* err_code) {
...
+  if (!HasDriveSpecifierPrefix(abs_path.c_str())) {
+    return L"";
+  }
+  std::wstring path = Normalize(abs_path);
+  std::unique_ptr<wchar_t[]> result(new wchar_t[4 + path.size() + 1]);
+  wcscpy(result.get(), L"\\\\?\\");
+  wcscpy(result.get() + 4, path.c_str());
+  result[4] = towupper(result[4]);
+  wchar_t* start = result.get() + 7;
```
meisterT: There is a lot of implicit knowledge in this code which makes it hard to read, e.g. where does the 7 come from?

meisterT: is this from UNC (if so make it conditional)?

src/main/native/windows/file.cc
meisterT: currently the only caller passes with_unc=false, right? Do you have plans to change that?

src/main/java/com/google/devtools/build/lib/windows/jni/WindowsFileOperations.java
laszlocsomor: Done.

src/main/native/windows/file.cc
laszlocsomor: Removed.

src/main/native/windows/file.cc
laszlocsomor: Added comment.

laszlocsomor:
Thanks for the quick and thorough review!

PR #9450, opened 2019-09-29
Java_toolchain: Remove Java 7 compatibility mode

### LINKED ISSUE #8772: `java_proto_library` results in spurious warning message
When compiling Java protos, the following error message is emitted:

```
warning: -parameters is not supported for target value 1.7. Use 1.8 or later.
```

This is because we force protos to be compiled with `-source 7 -target 7` for some reason:

https://github.com/bazelbuild/bazel/blob/master/tools/jdk/BUILD.java_tools#L14

then we separately add `-parameters`, which is not compatible with target level 7:

https://github.com/bazelbuild/bazel/blob/master/tools/jdk/BUILD.java_tools#L53

Would be nice if we didn't spam the console. I don't see a very simple way of not putting `-parameters` on the protoc Java compile command lines, though: we don't have a catchall branch in `compatible_javacopts` and we can't disable a flag there, either.

davido:
Closes #8772.

Change-Id: Ia3467b2a33ee4bfbd10cfe3f1027e292d3a967b7

### DIFF
--- a/tools/jdk/BUILD.java_tools
+++ b/tools/jdk/BUILD.java_tools
@@ -17,15 +17,6 @@ SUPRESSED_WARNINGS = select({
 java_toolchain(
     name = "toolchain",
     bootclasspath = ["@bazel_tools//tools/jdk:platformclasspath"],
-    compatible_javacopts = {
-        # Restrict protos to Java 7 so that they are compatible with Android.
-        "proto": [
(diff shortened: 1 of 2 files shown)


davido:
Superseded with: https://github.com/bazelbuild/bazel/pull/9494

Issue #9321, opened 2019-09-03
Tree artifacts don't seem to work on Windows


ale64bit:
### Description of the problem / feature request:

Tree artifacts don't seem to work on Windows.

### Feature requests: what underlying problem are you trying to solve with this feature?

Generate a `cc_library` from an unknown number of source files, e.g. because they are produced by a tool.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Same as in the accepted answer here: https://stackoverflow.com/questions/48417712/how-to-build-static-library-from-the-generated-source-files-using-bazel-build/48524539#48524539

But with a small modification to make it compatible with Windows:

genccs.bzl: verbatim.

BUILD: verbatim.

genccs.cpp:
```cpp
#include <fstream>
#include <Windows.h>// BEFORE: <sys/stat.h>
using namespace std;

int main (int argc, char *argv[]) {
  CreateDirectory(argv[1], NULL); // BEFORE: mkdir(argv[1], S_IRWXU);
  ofstream myfile;
  myfile.open(string(argv[1]) + string("/foo.cpp"));
  myfile << "int main() { return 42; }";
  return 0;
}
```

WORKSPACE: empty file.

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

release 0.29.0

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

Relevant question and answer: https://stackoverflow.com/questions/48417712/how-to-build-static-library-from-the-generated-source-files-using-bazel-build/48524539#48524539

Other relevant links:
https://github.com/bazelbuild/bazel/issues/5093

### Any other information, logs, or outputs that you want to share?

When running `bazel build main` with the above example, I'm getting:
```
INFO: Analyzed target //:main (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: C:/src/treeartifact/BUILD:7:1: output '_dotd/main/gen_tree/foo.obj.d' was not created
ERROR: C:/src/treeartifact/BUILD:7:1: not all outputs were created or valid
Target //:main failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 3.087s, Critical Path: 2.23s
INFO: 4 processes: 4 local.
[...]

PR #9454, opened 2019-10-01
Windows: Fix compiling cpp with tree artifact as source

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel/issues/9321

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
@@ -1398,12 +1398,16 @@ public final class CcCompilationHelper {
             /* additionalBuildVariables= */ ImmutableMap.of()));
     semantics.finalizeCompileActionBuilder(configuration, featureConfiguration, builder);
     // Make sure this builder doesn't reference ruleContext outside of analysis phase.
+    SpecialArtifact dotdTreeArtifact = null;
+    if (!featureConfiguration.isEnabled(CppRuleClasses.PARSE_SHOWINCLUDES)) {
+      dotdTreeArtifact = CppHelper.getDotdOutputTreeArtifact(
+          actionConstructionContext, label, sourceArtifact, outputName, usePic);
+    }
     CppCompileActionTemplate actionTemplate =
         new CppCompileActionTemplate(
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
```diff
@@ -1398,12 +1398,16 @@ private Artifact createCompileActionTemplate(
             /* additionalBuildVariables= */ ImmutableMap.of()));
     semantics.finalizeCompileActionBuilder(configuration, featureConfiguration, builder);
     // Make sure this builder doesn't reference ruleContext outside of analysis phase.
+    SpecialArtifact dotdTreeArtifact = null;
```
ulfjack: Can you add a comment here what this is for?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileActionTemplate.java
```diff
@@ -127,9 +127,12 @@
         TreeFileArtifact outputTreeFileArtifact =
             ActionInputHelper.treeFileArtifactWithNoGeneratingActionSet(
                 outputTreeArtifact, PathFragment.create(outputName), artifactOwner);
-        TreeFileArtifact dotdFileArtifact =
-            ActionInputHelper.treeFileArtifactWithNoGeneratingActionSet(
-                dotdTreeArtifact, PathFragment.create(outputName + ".d"), artifactOwner);
+        TreeFileArtifact dotdFileArtifact = null;
+        if (dotdTreeArtifact != null) {
```
ulfjack: Can you please also mark dotdTreeArtifact as @Nullable?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileActionTemplate.java
meteorcloudy: Thanks, done!

Issue #8009, opened 2019-04-11
Case insensitive labels poisons cache


ozio85:
### Description of the problem / feature request:

When first building with a correct tag on Windows e.g.
bazel build //MyLabel/Test:MyName

Then building with an incorrect tag (but will still work):
bazel build //mylabel/Test:MyName

Will in some cases cause the cache to be poisoned (and will in all cases make c-file includes to generated files to be invalid) 

### Feature requests: what underlying problem are you trying to solve with this feature?

Require all platforms (not only linux) to have exact casing, otherwise throw an error.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

See problem description above

### What operating system are you running Bazel on?

Windows

### What's the output of `bazel info release`?

0.23.2

###  Have you found anything relevant by searching the web?

-GitHub issues:
#6401
#7627

### Any other information, logs, or outputs that you want to share?

Nope :) Great project otherwise! :)

meisterT:
cc @laszlocsomor since this seems to be a Windows issue.

laszlocsomor:
This is not a Windows bug, it just shows up on Windows where the filesystem is case-preserving but case-insensitive.

The cause is complex: Label creation and Package loading use Path/PathFragment objects, I assume they check file existence (e.g. "does foo/BUILD exist? if so, //foo is a package") which ignores casing on Windows.

The naive fix is to make the WindowsFileSystem case-sensitive, but I think that's too aggressive. Bazel should accepts incorrectly-cased paths in some cases (e.g. if they come from a flags such as `--output_user_root`), and reject in others (if a Label is mis-cased).

I think the correct fix is to validate casing just in Label creation and Package lookup. Target names are already case-sensitive because Bazel matches them against strings read from a BUILD file.

laszlocsomor:
More thinking later, I realize fixing this in the package loading logic would be impractical.

Package loading involves package lookup, which involves checking if a BUILD file exists under some path. Skyframe (Bazel's "engine") evaluates the PackageLookupFunction which requests a FileValue for the BUILD file. Special-casing these (but not other) FileValue lookups to be case-sensitive could have consequences I cannot foresee; I'm afraid Skyframe could draw an inconsistent image of the filesystem and crash.

[...]

Issue #6401, opened 2018-10-16
Label case insensitivity causes duplicate outputs 


jerrymarino:
When Bazel builds 2 case insensitive `Label` strings, it will throw various combinations of errors or exceptions and fail.

Additionally, the duplicated outputs for multiple actions won't fail the validation check.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Build bazel targets from the command line using a lowercased version of the label

```
bazel build //A/B:C //a/b:c
```
_tested with native C++ compilation rules_

### What's the output of `bazel info release`?
```
$ tools/bazelwrapper info release
release 0.18.0
```

dslomov:
* Which platform is this on?
* what is `tools/bazelwrapper`?

aiuto:
Passing back to @jerrymarino  for more details. We would love to know what host systems and if the workspace is located on a case insensitive or sensitive partition.

jerrymarino:
Hey @dslomov and @aiuto thanks for following up - I can reproduce this on my current machine:
```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.6
BuildVersion:   17G65
```
And any other mac running Bazel.

`bazelwrapper` is a helper script I'm using to version and automatically download bazels - none the less, I can reproduce the issue by running the bazel binary directly.

gregestren:
Jerry,

Exactly which error do you see?

On my Mac, when I make a simple `cc_binary` in `testapp/BUILD` and just change the target case I get:

```
$ bazel build //testapp:cc //testapp:CC
ERROR: Skipping '//testapp:CC': no such target '//testapp:CC': target 'CC' not declared in package 'testapp' (did you mean 'cc'?) defined by /Users/greg/bazel/testapp/BUILD
WARNING: Target pattern parsing failed.
ERROR: no such target '//testapp:CC': target 'CC' not declared in package 'testapp' (did you mean 'cc'?) defined by /Users/gregce/bazel/bazel/testapp/BUILD
INFO: Elapsed time: 0.050s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

So even if the filesystem is case-insensitive Bazel is still keeping the label case-sensitive.

When I change the package's case though things break down:

```
$ bazel build //testapp:cc //TESTAPP:cc
INFO: Analysed 2 targets (1 packages loaded).
INFO: Found 2 targets...
[...]

PR #9456, opened 2019-10-01
WIP: Support case-sensitive package loading

laszlocsomor:
DO NOT SUBMIT: PackageIdentifier just became case-sensitive; must figure out how to put that behind a flag.

The --incompatible_validate_package_path_casing
flag enables a new validation step in package
loading: when enabled, Bazel requires that package
paths be case-correct.

When the user runs "bazel build //Foo/Bar:Baz",
Bazel looks up the package "Foo/Bar" by checking
"Foo/Bar/BUILD". If the file exists then the
package exists. Every Label that Bazel creates in
this package is derived from the "Foo/Bar" string.

On Linux the filesystem is case-sensitive, so the
path casing must be correct for stat() to succeed.
It's impossible to load the package "//Foo/Bar" as
"//FOO/baR".

On Windows and macOS the filesystem ignores the
difference between lower and upper case letters,
so stat("foo/bAr/BUILD") and stat("FOO/bar/build")
both succeed, and "bazel build //foo/bAr:Baz" and
"bazel build //FOO/bar:Baz" both work. This leads
to bugs such as #6401, #8009, and #8799.

This PR adds a new SkyFunction
(PathCasingLookupFunction) that compares a
RootedPath's casing to the actual casing on disk.
The SkyKey for PathCasingLookupFunction stores not
just the RootedPath reference but also the String
representation, to distinguish differntly cased
but otherwise equal RootedPaths.

The PackageLookupFunction depends on the
PathCasingLookupValue for the path in the
PackageIdentifier, and denies loading the package
if the path was incorrectly cased.

Change-Id: I9d55cd76b2499172cb55252032163d936b83690e


laszlocsomor:
This PR still needs:
- passing the CI tests
- adding a RELNOTES[NEW] tag
- creating a bug for the incompatible flag

laszlocsomor:
@ulfjack: PTAL at the description.

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
```diff
@@ -116,7 +119,9 @@ public static PackageIdentifier discoverFromExecPath(PathFragment execPath, bool
   private PackageIdentifier(RepositoryName repository, PathFragment pkgName) {
     this.repository = Preconditions.checkNotNull(repository);
     this.pkgName = Preconditions.checkNotNull(pkgName);
-    this.hashCode = Objects.hash(repository, pkgName);
+    this.repositoryCasing = repository.getName();
```
ulfjack: I don't see why you need to make a copy here. The RepositoryName and pkgName aren't changing underneath you, are they?

ulfjack:
In what cases can the RootedPath be different from the String? Why not just construct a new RootedPath with the correct casing?

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
laszlocsomor: RepositoryName.equals is case-ignoring. If I only stored the RepositoryName reference, I couldn't tell apart PackageIdentifiers with differently cased RepositoryNames. So I need to also store the String representation.

jin:
Closing this for now, please ping/reopen if you're intending to work on this.

PR #9704, opened 2019-10-04
Fix error in C++ use cases documentation

danielmartin:
In the sample scenario, the legacy library includes `include/some_lib.h`, so `copts` should not add the final `include` directory to the include path.


danielmartin:
@googlebot I signed it!

hlopko:
Oh good point. But since compilers look into the directory of the .cc file by default, copts attribute is not needed in this case. Let's change the text from `"include/some_lib.h"` to `"some_lib.h"`. Now the copts is needed and correct. What do you think?

danielmartin:
@hlopko Looks good. I've updated the patch.

PR #9872, opened 2019-10-07
Support remote execution and BES reporting via unix domain sockets.

jmillikin-stripe:
This extends the existing `--remote_proxy` flag to also apply to gRPC channels created for `--remote_executor` and `--remote_cache`. It also adds a `--bes_proxy` flag, which applies to gRPC channels created by BES.

Logic in `newNettyChannelBuilder()` was derived from the existing unix socket setup in `HttpBlobStore`.


jmillikin-stripe:
cc @adunham-stripe @ianoc-stripe

jmillikin-stripe:
Gentle ping -- I'd like to get this into v1.1 if possible.

PR #9972, opened 2019-10-08
Chocolatey: Remove strict MSYS2 version requirement

philwo:
The version given is very old and I don't see a reason why Bazel should depend on this exact version of MSYS2.

On Bazel's CI we just test against the latest MSYS2 version available from Chocolatey and it works fine, so one could argue that this old version is actually a very untested environment.

Users are also complaining about this (I guess because it prevents their MSYS2 version from being upgraded?): https://chocolatey.org/packages/bazel#comment-4643886227

### DIFF
--- a/scripts/packages/chocolatey/bazel.nuspec.template
+++ b/scripts/packages/chocolatey/bazel.nuspec.template
@@ -76,7 +76,7 @@ Supply like `--params="/option:'value' ..."` ([see docs for --params](https://gi
     </dependencies>-->
     <dependencies>
       <dependency id="chocolatey-core.extension" version="1.0.7"/>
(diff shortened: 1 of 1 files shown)


philwo:
FYI @petemounce - WDYT?

philwo:
@petemounce Thanks!

@meteorcloudy How do you feel about not restricting the supported version for msys2? Personally, I think it's fine - anything else than the latest version isn't tested on our CI anyway, because we always install the latest one from Choco and not the one specified as compatible in our own Choco package definition...

meteorcloudy:
I think the latest version of MSYS should work fine with Bazel. We can use that in our choco definition

Issue #4561, opened 2018-02-01
Building Windows DLL without exports fails


excitoon:
### Description of the problem / feature request:

In order to produce import library for a DLL bazel uses `/IMPLIB` flag of the linker and expect that it will force creating import library in the place specified:
https://msdn.microsoft.com/en-us/library/67wc07b9.aspx

Unfortunately, this flag is not managing the fact of creating or not creating import library but only sets the name for it. If DLL does not export anything, linker do not create it by design. Unluckily, we actively use DLLs without imports for a reason.

The output of bazel is as follows:
```
C:\Users\Chebotarev_V\Documents\bazel-issues\windows-dll-without-exports>bazel build ...
INFO: Analysed target //:hellolib.dll (8 packages loaded).
INFO: Found 1 target...
ERROR: C:/users/chebotarev_v/documents/bazel-issues/windows-dll-without-exports/BUILD:3:1: output 'hellolib.ifso' was not created
ERROR: C:/users/chebotarev_v/documents/bazel-issues/windows-dll-without-exports/BUILD:3:1: not all outputs were created or valid
Target //:hellolib.dll failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 49.652s, Critical Path: 1.14s
FAILED: Build did NOT complete successfully
```

Please find an example in my repository:
https://github.com/excitoon/bazel-issues/tree/master/windows-dll-without-exports

### Feature requests: what underlying problem are you trying to solve with this feature?

We need building DLLs without exports.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

https://github.com/excitoon/bazel-issues/tree/master/windows-dll-without-exports

### What operating system are you running Bazel on?

Windows 10 x64

### What's the output of `bazel info release`?

```
Build label: 0.10.0
Build target: bazel-out/x64_windows-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Nov 9 22:42:19 +50056 (1517474731339)
Build timestamp: 1517474731339
Build timestamp as int: 1517474731339
```

###  Have you found anything relevant by searching the web?

Nothing expect mentioned.

excitoon:
The same error in tensorflow:
```
[...]

PR #9976, opened 2019-10-09
Windows: Allow building DLL without any symbol exported

meteorcloudy:
Previously, when building a DLL without any symbol exported, the action
will fail because the import library won't be generated by the linker.
In this change, we use a trivial DEF file to make sure the linker always
generate the import library when linking a DLL.

Fixes https://github.com/bazelbuild/bazel/issues/4561

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
@@ -459,7 +459,7 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
                   ruleContext, defParser, objectFiles.build(), binary.getFilename());
         }
         customDefFile = common.getWinDefFile();
-        trivialDefFile = CppHelper.createTrivialDefFileAction(ruleContext, binary.getFilename());
+        trivialDefFile = CppHelper.createTrivialDLLDefFileAction(ruleContext, binary.getFilename());
       }
     }
 
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
@@ -430,7 +430,6 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
     CcLinkingContext depsCcLinkingContext = collectCcLinkingContext(ruleContext);
 
     Artifact generatedDefFile = null;
-    Artifact customDefFile;
     Artifact winDefFile = null;
     if (isLinkShared(ruleContext)) {
       if (featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
@@ -458,8 +457,7 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
               CppHelper.createDefFileActions(
                   ruleContext, defParser, objectFiles.build(), binary.getFilename());
         }
-        customDefFile = common.getWinDefFile();
-        winDefFile = CppHelper.getWindowsDefFileForLinking(ruleContext, customDefFile, generatedDefFile, featureConfiguration);
+        winDefFile = CppHelper.getWindowsDefFileForLinking(ruleContext, common.getWinDefFile(), generatedDefFile, featureConfiguration);
       }
     }
 
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
@@ -430,8 +430,8 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
     CcLinkingContext depsCcLinkingContext = collectCcLinkingContext(ruleContext);
 
     Artifact generatedDefFile = null;
-    Artifact customDefFile = null;
-    Artifact trivialDefFile = null;
+    Artifact customDefFile;
+    Artifact winDefFile = null;
     if (isLinkShared(ruleContext)) {
       if (featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
         ImmutableList.Builder<Artifact> objectFiles = ImmutableList.builder();
@@ -459,7 +459,7 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
                   ruleContext, defParser, objectFiles.build(), binary.getFilename());
         }
         customDefFile = common.getWinDefFile();
-        trivialDefFile = CppHelper.createTrivialDefFileAction(ruleContext);
+        winDefFile = CppHelper.getWindowsDefFileForLinking(ruleContext, customDefFile, generatedDefFile, featureConfiguration);
       }
     }
 
@@ -509,9 +509,7 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
             cppConfiguration,
             linkType,
             pdbFile,
-            generatedDefFile,
-            customDefFile,
-            trivialDefFile);
+            winDefFile);
 
     CcLinkingOutputs ccLinkingOutputsBinary = ccLinkingOutputsAndCcLinkingInfo.first;
 
@@ -700,9 +698,7 @@ public abstract class CcBinary implements RuleConfiguredTargetFactory {
       CppConfiguration cppConfiguration,
       LinkTargetType linkType,
       Artifact pdbFile,
-      Artifact generatedDefFile,
-      Artifact customDefFile,
-      Artifact trivialDefFile)
+      Artifact winDefFile)
       throws InterruptedException, RuleErrorException {
     CcCompilationOutputs.Builder ccCompilationOutputsBuilder =
         CcCompilationOutputs.builder()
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -885,6 +886,26 @@ public static Artifact createDefFileActions(
...
 
+  /**
+   * Create action for generating trivial DEF file without any exports, should only be used when
+   * targeting Windows.
+   *
+   * @param dllName The DLL name to be written into the DEF file, it specifies which DLL is required
+   *     at runtime
+   * @return The DEF file artifact.
+   */
+  public static Artifact createTrivialDefFileAction(RuleContext ruleContext, String dllName) {
```
maximyurchuk: Seems name of function is not very good, because def file created by this function is suitable only for `.dll`. Maybe createTrivial**Dll**DefFileAction is better?

meteorcloudy: I think the fact we are creating a DEF file implies it's for a Windows DLL, there's no DEF file for shared library on other platforms, right?

maximyurchuk: I mean that def can be used for linking `.exe`, not only for `.dll`. But def file created by `createTrivialDefFileAction` not suitable for `.exe` creation, because of `LIBRARY ...` string, which doesn't work for `.exe`.  So I propose emphasize that def file suitable for `.dll`, not `.exe`.

meteorcloudy: I see, thanks for the notice. Done.

maximyurchuk:
@meteorcloudy I'm not familiar with this code yet, could you please say will empty def file creation work 

- while using `cc_common.link`
- manual linker invoke using `cc_common.create_link_variables` + `cc_common.get_mem usiory_inefficient_command_line` 

?

meteorcloudy:
@maximyurchuk Oh, I just did an experiment. Actually an empty DEF file will still force the linker to generate a import library. So it does work! I'll use an empty DEF file instead.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
```diff
@@ -312,10 +312,11 @@ public static void init(
         || !ccCompilationOutputs.isEmpty()) {
       if (featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
         // If user specifies a custom DEF file, then we use it.
-        Artifact defFile = common.getWinDefFile();
-
-        Artifact defParser = common.getDefParser();
+        Artifact customDefFile = common.getWinDefFile();
         Artifact generatedDefFile = null;
+        Artifact trivialDefFile = CppHelper.createTrivialDefFileAction(ruleContext);
```
laszlocsomor: Can we spare creating `trivialDefFile` here too, and create it only in line 346?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
```diff
@@ -333,16 +334,16 @@ public static void init(
...
-        // object files to generate DEF file and use it to export symbols - if we have a parser.
-        // Otherwise, use no DEF file.
-        if (defFile == null
-            && CppHelper.shouldUseGeneratedDefFile(ruleContext, featureConfiguration)) {
-          defFile = generatedDefFile;
-        }
-
-        if (defFile != null) {
-          linkingHelper.setDefFile(defFile);
+        // 1. If a custom DEF file is specified in win_def_file attribute, use it.
```
laszlocsomor: This decision logic is the same as in cc_library, can you extract it to a function?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -885,6 +886,23 @@ public static Artifact createDefFileActions(
     return defFile;
   }
 
+  /**
+   * Create action for generating trivial DEF file without any exports, should only be used when
```
laszlocsomor: "Trivial" carries little meaning for me. Can we find a more precise name? What do you think about "no-export" or "fallback"?

And how about calling the function `createDefFileForDllWithoutExports` or something similar?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -458,6 +459,7 @@ public static ConfiguredTarget init(CppSemantics semantics, RuleContext ruleCont
                   ruleContext, defParser, objectFiles.build(), binary.getFilename());
         }
         customDefFile = common.getWinDefFile();
+        trivialDefFile = CppHelper.createTrivialDefFileAction(ruleContext);
```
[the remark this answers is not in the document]
meteorcloudy: Thanks for the advice! I refactored the logic, should be better now.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -823,10 +827,16 @@ public static ConfiguredTarget init(CppSemantics semantics, RuleContext ruleCont
...
+    // 2. If the windows_export_all_symbols feature is enabled, parse object files to
+    //    generate DEF file and use it to export symbols - if we have a parser.
+    // 3. Otherwise, we use a trivial DEF file to ensure the import library will be generated.
     if (customDefFile != null) {
       ccLinkingHelper.setDefFile(customDefFile);
-    } else if (CppHelper.shouldUseGeneratedDefFile(ruleContext, featureConfiguration)) {
+    } else if (generatedDefFile != null && CppHelper.shouldUseGeneratedDefFile(ruleContext, featureConfiguration)) {
       ccLinkingHelper.setDefFile(generatedDefFile);
+    } else {
+      ccLinkingHelper.setDefFile(trivialDefFile);
```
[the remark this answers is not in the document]
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
meteorcloudy: Done,

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
meteorcloudy: I replaced it with "empty" here, because it's essentially an empty file. Does it sound good?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -430,7 +430,8 @@ public static ConfiguredTarget init(CppSemantics semantics, RuleContext ruleCont
     CcLinkingContext depsCcLinkingContext = collectCcLinkingContext(ruleContext);
 
     Artifact generatedDefFile = null;
-    Artifact customDefFile = null;
+    Artifact customDefFile;
```
laszlocsomor: Could you remove `customDefFile`? Seems like we no longer use it. You can inline `common.getWinDefFile()` into `getWindowsDefFileForLinking`.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -430,7 +430,8 @@ public static ConfiguredTarget init(CppSemantics semantics, RuleContext ruleCont
...
+    Artifact winDefFile = null;
```
laszlocsomor: `generatedDefFile` is used again in line 641. I wonder, do you have to use `winDefFile` there instead? If so, please update and remove `generatedDefFile` here and declare it in the `if` block's scope below.

meteorcloudy: No, it should still be geenratedDefFile in line 641. Because that's for users to get the generated DEF file via `def_file` output group, it should work even when `windows_export_all_symbols` feature is not enabled.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -885,6 +887,42 @@ public static Artifact createDefFileActions(
     return defFile;
   }
 
+  /**
+   * Create action for generating an empty DEF file without any exports, should only be used when
+   * targeting Windows.
+   *
+   * @return The artifact of an empty DEF file.
+   */
+  public static Artifact createEmptyDefFileAction(RuleContext ruleContext) {
```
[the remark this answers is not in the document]
meteorcloudy: Thanks, done.

PR #9978, opened 2019-10-09
Hide sidebar item 'Building With Platforms'

lemoncurry:
While researching on Bazel I noticed a broken link in the documentation for item 'Building With Platforms' on the bazel website.
According to PR #9446 this item should not appear before Bazel 1.0.
This is why I added version conditionals so this item will appear only for the intended versions of the documentation.


lemoncurry:
Fixed by PR #9985.

PR #8415, opened 2019-05-21
Switch to command -v over which

keith:
This fixes an issue on macOS where `which` failed to return any results
because the `PATH` wasn't propagated. `command` is more standard for
this use case as well https://github.com/koalaman/shellcheck/wiki/SC2230

Fixes https://github.com/bazelbuild/bazel/issues/8414

PR #9984, opened 2019-10-10
Python: Look for which before attempting to look for python

jwendell:
If the system does not have `which`, looking for python executables
will fail, and the error message will not help, because it was
`which` that was not found, not python.

### DIFF
--- a/tools/python/pywrapper_template.txt
+++ b/tools/python/pywrapper_template.txt
@@ -44,6 +44,11 @@ die() {
 #     https://github.com/bazelbuild/bazel/issues/8414
 #     https://github.com/bazelbuild/bazel/issues/8415
 
+# Check if `which` is present first
+if ! which --help >/dev/null 2>&1; then
+  die "which utility not found."
+fi
+
 # Try the "python%VERSION%" command name first, then fall back on "python".
 PYTHON_BIN="$(PATH="$PATH" which python%VERSION% 2> /dev/null)"
 USED_FALLBACK="0"


brandjon:
Out of curiosity, what environment are you using that lacks `which`?

I'm beginning to rue the day I chose to implement the autodetecting toolchain with a shell wrapper.

jwendell:
Containers. I'm using a Fedora 30 container. By default `which` is not installed on it.

keith:
I think this came up in a previous conversation but `command -v` might work for all platforms (this is what shellcheck recommends too)

keith:
Ah here: https://github.com/bazelbuild/bazel/pull/8415

nlopezgi:
We also saw this error when we didn't have a custom python toolchain in rules_docker for `py_image`. Most of our `xx_image` containers are based off distroless, which does not have `which` installed.

keith:
I think the downside of command -v not checking the executable bit is probably not a big deal and it would instead be worth solving these cases and sacrificing that.

brandjon:
From a quick test locally, it looks like `command -v` chooses an executable file in preference to a non-executable one.

```sh
$ cd ~
# Prefer ~ over /usr/bin
$ export PATH=$PWD:$PATH

$ touch python
$ which python
/usr/bin/python
$ command -v python
/usr/bin/python

$ chmod 755 python
$ which python
/usr/local/google/home/brandjon/python
$ command -v python
/usr/local/google/home/brandjon/python
```

The difference is just that `command -v` also returns a non-executable one if nothing executable's available.

```sh
$ touch blah
$ chmod 644 blah
$ which blah
$ command -v blah
/usr/local/google/home/brandjon/blah
```

However, we have to worry about shell environments that don't provide `command -v`, and which would be broken if we were to substitute it for the existing uses of `which`. Therefore, if we were to use `command -v` I'd want to have it be a fallback for when `which` isn't already available. I'd also want to avoid the same problem of having a cryptic error message when both `which` and `command` are unavailable, or when the file chosen by `command -v` ends up being non-executable.

Given that doing this rigorously would require factoring up the calls to `which` into a helper and adding more test coverage to [pywrapper_test.py](https://github.com/bazelbuild/bazel/blob/master/tools/python/pywrapper_test.py), it sounds much simpler to just stick with `which` and improve the error message when it's unavailable.

keith:
Do you know of any places where command -v isn't available? Since it is POSIX compliant where which is not https://stackoverflow.com/questions/592620/how-to-check-if-a-program-exists-from-a-bash-script/677212#677212

brandjon:
`csh` is one that I just tried that appears to have no `command` command. A lot of other ones seem to have it.

Is there anything in POSIX prohibiting `/bin/sh` from being `csh`?

keith:
I'm not sure. From a quick google it's hard to tell if any popular OS's use csh as a default. I think overall using that would be safer, but in theory you could still use which and fall back to command or vice versa

aiuto:
@jwendell Are you going to address the review comments or drop this?

meisterT:
What's the state of this PR?

meisterT:
Closing for lack of activity. Please reply if you want to pick it up again.

Issue #8245, opened 2019-05-06
Allow passing extra HTTP headers to remote cache


ob:
It's currently not possible to pass extra HTTP headers to the remote cache. I would like to be able to do something like this:

```
bazel build --remote_http_cache=$SOME_URL --remote_http_header="Build-Id: some-id" --remote_http_header="Product-Name: Some-Name"
```

The idea is to allow for custom logic in the remote http cache to keep metrics about cache usage on a per-product or per-build basis.

buchgr:
SGTM. It seems like a good idea to also support this for gRPC as it's HTTP/2 after all.

I think I'd prefer a format like `--remote_header=Header=Value`.

simon0191:
@ob are you planning to add support for GRPC headers?

ob:
I'm not sure how to do that. I'm not familiar with that section of the code. I'll take a look but if you have any pointers on what would need to be done I'd appreciate it.

excitoon:
@ob I would rather add "headers-adder" to the pipeline. That will be pretty clean solution.

ob:
I believe this issue is fixed.

PR #10015, opened 2019-10-14
Pass --remote_header's headers to grpc endpoints

AlessandroPatti:
Extends #8245 to gRPC.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -272,7 +272,8 @@ public final class RemoteModule extends BlazeModule {
             new GrpcRemoteExecutor(
                 execChannel.retain(),
                 GoogleAuthUtils.newCallCredentials(authAndTlsOptions),
-                execRetrier);
+                execRetrier,
+                remoteOptions);
         execChannel.release();
         RemoteExecutionCache remoteCache =
             new RemoteExecutionCache(cacheClient, remoteOptions, digestUtil);
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
@@ -122,6 +122,7 @@ public class GrpcCacheClient implements RemoteCacheClient, MissingDigestsFinder 
   private ContentAddressableStorageFutureStub casFutureStub() {
     return ContentAddressableStorageGrpc.newFutureStub(channel)
         .withInterceptors(TracingMetadataUtils.attachMetadataFromContextInterceptor())
+        .withInterceptors(TracingMetadataUtils.newCacheHeadersInterceptor(options))
         .withCallCredentials(credentials)
         .withDeadlineAfter(options.remoteTimeout, TimeUnit.SECONDS);
   }
@@ -129,6 +130,7 @@ public class GrpcCacheClient implements RemoteCacheClient, MissingDigestsFinder 
   private ByteStreamStub bsAsyncStub() {
     return ByteStreamGrpc.newStub(channel)
         .withInterceptors(TracingMetadataUtils.attachMetadataFromContextInterceptor())
+        .withInterceptors(TracingMetadataUtils.newCacheHeadersInterceptor(options))
         .withCallCredentials(credentials)
         .withDeadlineAfter(options.remoteTimeout, TimeUnit.SECONDS);
   }
@@ -136,6 +138,7 @@ public class GrpcCacheClient implements RemoteCacheClient, MissingDigestsFinder 
   private ActionCacheBlockingStub acBlockingStub() {
     return ActionCacheGrpc.newBlockingStub(channel)
         .withInterceptors(TracingMetadataUtils.attachMetadataFromContextInterceptor())
+        .withInterceptors(TracingMetadataUtils.newCacheHeadersInterceptor(options))
         .withCallCredentials(credentials)
         .withDeadlineAfter(options.remoteTimeout, TimeUnit.SECONDS);
   }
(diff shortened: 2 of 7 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -182,14 +185,37 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+      if (remoteOptions.remoteHeaders.size() > 0){
+        interceptors.add(createExtraHeadersClientInterceptor(remoteOptions.remoteHeaders));
+      }
+      if (remoteOptions.remoteExecHeaders.size() > 0) {
+        execInterceptors = new ArrayList<>(interceptors);
+        execInterceptors.add(createExtraHeadersClientInterceptor(remoteOptions.remoteExecHeaders));
+      }
+      if (remoteOptions.remoteCacheHeaders.size() > 0) {
+        cacheInterceptors = new ArrayList<>(interceptors);
+        cacheInterceptors.add(createExtraHeadersClientInterceptor(remoteOptions.remoteCacheHeaders));
```
buchgr: Instead of adding the headers in the module, consider passing the headers to the RemoteSpawnRunner. That way it's easier to write a unit test.

AlessandroPatti:
> Can you please explain why headers for caching and execution have different flags? I am somewhat surprised given that Bazel doesn't allow you to mix caching and execution in one invocation.

Bazel does allow to have two different endpoints for cache and executor doesn't it? The idea there was that if both `--remote_cache=<host>` and `--remote_executor=<host>`, it is possible to attach different headers to the cache and execution requests.

buchgr:
@AlessandroPatti yes we do but I don't know of anyone setting these two to different values or both at the same time. If you set `--remote_executor` then `--remote_cache` automatically defaults to the value of `--remote_executor`.

So for now I'd prefer to just reuse `--remote_headers` flag for everything until the use case comes up.

simon0191:
> @AlessandroPatti yes we do but I don't know of anyone setting these two to different values or both at the same time. If you set `--remote_executor` then `--remote_cache` automatically defaults to the value of `--remote_executor`.
> 
> So for now I'd prefer to just reuse `--remote_headers` flag for everything until the use case comes up.

We are using separate services for the CAS and the Scheduler, and therefore require different endpoints for both. I though that Bazel now creates 2 different connections.
CC @eliangidoni @AlessandroPatti

AlessandroPatti:
We do have such use case already. We want to add different headers to cache and execution requests. Now that I think about it, it is not really necessary for the two endpoint to be different, different headers might be necessary according to the request type.

buchgr:
Makes sense. Let's figure out how to keep flags sane then :-). I don't want to have 3 different flags.

How about:

* `--remote_header`: if set applies to all requests for both http and grpc. 
* `--remote_exec_header`: if set applies only to Execute() RPCs. Allows to override a header set in --remote_header.

simon0191:
If we want to pass a header for the CAS and a header for execution the request would look like:

`bazel build --remote_header='CAS_SPECIFIC_HEADER=XXX' --remote_exec_header='EXEC_SPECIFIC_HEADER=YYY'`

I think it would be better to have a flag that allow us to specify a cas specific header so that it's more explicit what's happening. 

`bazel build --remote_cas_header='CAS_SPECIFIC_HEADER=XXX' --remote_exec_header='EXEC_SPECIFIC_HEADER=YYY'`

And, alternatively have a `--remote_header=GENERIC_HEADER=ZZZ` which will be passed to both, CAS and Executor, but will be overriden by the specific ones.

AlessandroPatti:
> If we want to pass a header for the CAS and a header for execution the request would look like:
> 
> `bazel build --remote_header='CAS_SPECIFIC_HEADER=XXX' --remote_exec_header='EXEC_SPECIFIC_HEADER=YYY'`
> 
> I think it would be better to have a flag that allow us to specify a cas specific header so that it's more explicit what's happening.
> 
> `bazel build --remote_cas_header='CAS_SPECIFIC_HEADER=XXX' --remote_exec_header='EXEC_SPECIFIC_HEADER=YYY'`
> 
> And, alternatively have a `--remote_header=GENERIC_HEADER=ZZZ` which will be passed to both, CAS and Executor, but will be overriden by the specific ones.

I'd also add that it would not be possible to pass a header to the CAS and not to the executor.

@buchgr

simon0191:
@werkt thoughts on this?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -182,14 +185,37 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+    }
+    if (Strings.isNullOrEmpty(remoteOptions.remoteExecutor) && remoteOptions.remoteCacheHeaders.size() > 0){
+      throw new AbruptExitException(
+          "--remote_exec_header can only be used if --remote_executor is set.",
+          ExitCode.COMMAND_LINE_ERROR);
+    }
     try {
       List<ClientInterceptor> interceptors = new ArrayList<>();
+      List<ClientInterceptor> execInterceptors = interceptors;
+      List<ClientInterceptor> cacheInterceptors = interceptors;
```
werkt: This will use the same referenced list for all of the interceptors, which definitely is not what you want. You would need separate List instances, likely with interceptors populated below copied in, to effect different sets of interceptors.

werkt: I see the reassignment below accomplishes this, but a bit awkwardly. Composition via a list view would be preferred, as jakob mentioned below, as a parameter to RemoteSpawnRunner instantiation, just so you don't have the potential for reference duplication, and don't have to rely on reference equality.

AlessandroPatti: I am not sure what you meant here. How should the headers be passed with the `RemoteSpawnRunner`? Also, that module takes care only of remote execution, doesn't it? In that case, how to pass the headers to the cache too?

AlessandroPatti:
Moved the code outside the `RemoteModule` class and added some tests. @buchgr can you please review the changes?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
werkt: Coordinated offline. New changes look good, ty.

elianuber:
cc @ola-rozenfeld @ishikhman 
Could you take a look at this PR ? It seems that the most important comments were fixed by @AlessandroPatti

rizzatti:
👍

AlessandroPatti:
@buchgr Could you please review these changes? It is the last missing piece for us and we'd love to see it in the next release

PR #10029, opened 2019-10-14
Suppress last-ditch download exceptions w/cleanup

werkt:
Create an encapsulating DownloadException to represent the aggregate
exception set of an ActionResult download. IOExceptions will be retained
through exception suppression, and the outer exception has a property to
indicate if it only represents a sequence of CacheNotFoundExceptions.

InterruptedExceptions interception is improved to cancel pending work
and wrap, through suppression, any DownloadException that also occurred
during the download. InterruptedException being thrown on the download
control thread, it does not require suppression of further interrupts,
and can represent an outer download exception. Thread interrupt status
is suppressed for cancellations, and conveyed on throw.

These exception wrapping efforts allow non-asynchronous frame
representation in stack traces, and much clearer identification of
sources within remote strategy execution which produce failures based on
remote errors.

Any DownloadException in the last-ditch output download under
handleError in RemoteSpawnRunner is added as suppressed to the
initiating exception. Other exceptions (likely local IO) present clear
immediate traces and do not require specialized treatment.


buchgr:
I am not sure how I feel about this change. In part because it seems to be treating a symptom instead of the root cause: If you see a stacktrace from Bazel then I'd consider this a bug. Would you agree?

werkt:
I think that might be an oversimplification of the presentation of a stack trace - fwiw this doesn't change the current behavior of only making these traces visible with `--verbose_failures`, it simply includes logic that ensures that if a trace *is* printed as a result of a download failure with `--verbose_failures` enabled and without `--remote_local_fallback`, it contains lines that refer to the actual bazel code which invoked the requests which threw exceptions, to provide a bare minimum of identification of what it was doing, as opposed to the grpc-only frames present during client calls.

werkt:
@jin ping on triage for this

To any future reviewers, this does not add stack traces where they did not exist before. This only improves the quality of existing RuntimeExceptions from skyframe or exceptions printed with --verbose_failures.

jin:
Sorry, was OOO. This looks like a general developer experience improvement. @philwo wdyt?

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -260,6 +262,36 @@ private static Path toTmpDownloadPath(Path actualPath) {
...
   }
 
+  /**
+   * Exception which represents a collection of IOExceptions for the purpose
+   * of distinguishing remote communication exceptions from those which occur
+   * on filesystems locally. This exception serves as a trace point for the actual
+   * download, so that the intented operation can be observed in a stack, with all
+   * constituent exceptions available for observation.
+   */
+  static class DownloadException extends Exception {
```
ulfjack: It looks like this is essentially a public API. I think it would be better to move it to a top-level .java file.

ulfjack: Why not make this an IOException?

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -260,6 +262,36 @@ private static Path toTmpDownloadPath(Path actualPath) {
...
+    // true since no empty DownloadException is ever thrown
+    boolean allCacheNotFoundException = true;
```
ulfjack: Prefer private

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -260,6 +262,36 @@ private static Path toTmpDownloadPath(Path actualPath) {
...
+
+    DownloadException() {
+    }
+
+    DownloadException(IOException e) {
+      add(e);
+    }
+
+    void add(IOException e) {
```
ulfjack: Why not override addSuppressed?

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -348,12 +388,7 @@ public void download(
           tmpOutErr.clearErr();
         }
       } catch (IOException e) {
-        if (downloadException != null && e != downloadException) {
-          e.addSuppressed(downloadException);
-        }
-        if (interruptedException != null) {
-          e.addSuppressed(interruptedException);
-        }
+        downloadException.add(e);
```
ulfjack: This seems to be missing initialization for downloadException - how do you know it's not null at this point?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
```diff
@@ -169,6 +170,8 @@ public CacheHandle lookup(Spawn spawn, SpawnExecutionContext context)
                 prof.profile(ProfilerTask.REMOTE_DOWNLOAD, "download outputs")) {
               remoteCache.download(
                   result, execRoot, context.getFileOutErr(), context::lockOutputFiles);
+            } catch (DownloadException e) {
+              throw new IOException(e);
```
ulfjack: Adding another level of exception wrapping here doesn't sound so good. I already had problems debugging issues with the capabilities call because it was wrapping three levels deep.

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
werkt: Nominally to expose this checked exception uniquely for handlers - there are too many IOException interceptors already, and I wanted to have a clear shot to the top.

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
werkt: I wanted a filter on this, so that it couldn't contain/suppress anything but IOExceptions. That is probably unnecessary, but I felt it needed scope.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
werkt: No disagreement there, this all just hinges on making DownloadException an IOException, which it seems like it should be.

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
werkt: Truth. Will fix.

werkt: Actually, my treatment of this exception is wrong, I'll just redo it.

werkt: And this was another reason specifically not to have DownloadException be an IOException (despite being badly behaved here) - we cannot possibly be handling the same IOException to be a suppression parent if DownloadException is *not* an IOException (see my previous change to add the `e != downloadException` here).

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
werkt: And as it turns out, addSuppressed is `final`

src/main/java/com/google/devtools/build/lib/remote/DownloadException.java
```diff
@@ -0,0 +1,49 @@
...
+  private boolean allCacheNotFoundException = true;
+
+  DownloadException() {
+  }
+
+  DownloadException(IOException e) {
+    add(e);
+  }
+
+  void add(IOException e) {
```
ulfjack: Can you add documentation that this works like addSuppressed and that we can't override addSuppressed?

werkt:
@ulfjack Added your doc, and sweetened the pot here by fixing #10846 with this by preventing any presumed cancellation of concurrently uploadable blobs, and made it into a very similar capture of a now `BulkTransferException`. I think this fits the mold of the exception type anyway.

dws-uber:
Any update on this?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
```diff
@@ -57,20 +57,40 @@ private void uploadMissing(Map<Digest, Path> files, Map<Digest, ByteString> blob
...
+        if (bulkTransferException != null) {
+          bulkTransferException = new BulkTransferException();
+        }
+        Throwable cause = e.getCause();
+        IOException ioEx;
+        if (cause instanceof IOException) {
+          ioEx = new IOException(cause);
+        } else {
+          ioEx = (IOException) cause;
+        }
```
ulfjack:
```suggestion
        IOException ioEx = cause instanceof IOException ? (IOException) cause : new IOException(cause);
```

src/main/java/com/google/devtools/build/lib/remote/BulkTransferException.java
```diff
@@ -0,0 +1,56 @@
...
+   * Add an IOException to the suppressed list.
+   *
+   * The Java standard addSuppressed is final and this method stands in
+   * its place to selectively filter and record whether all suppressed
+   * exceptions are CacheNotFoundExceptions
+   */
+  void add(IOException e) {
+    if (allCacheNotFoundException) {
+      allCacheNotFoundException = e instanceof CacheNotFoundException;
+    }
```
ulfjack:
```suggestion
    allCacheNotFoundException &= e instanceof CacheNotFoundException;
```

src/main/java/com/google/devtools/build/lib/remote/BulkTransferException.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * Add an IOException to the suppressed list.
+   *
+   * The Java standard addSuppressed is final and this method stands in
+   * its place to selectively filter and record whether all suppressed
+   * exceptions are CacheNotFoundExceptions
+   */
+  void add(IOException e) {
```
ulfjack: I wonder if this can ever be called in a multi-threaded context.

werkt: It cannot in any of the uses I've added here - all the aggregations happen on a single thread.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
werkt: There's a better solution to this that looks more like what's available in RemoteCache.java, i'll refactor that and replace this.

werkt:
Can we see this through? The problem that this corrects in >1.1.0 is keeping us from upgrading to it. @philwo?

jin:
I can bring it in.

dws-uber:
> I can bring it in.

If this could be part of 3.1 that would be excellent.

jin:
> If this could be part of 3.1 that would be excellent.

cc @laurentlb (release manager for 3.1 https://github.com/bazelbuild/bazel/issues/11076)

laurentlb:
Sounds good. Please leave a comment on https://github.com/bazelbuild/bazel/issues/11076 with the commit hash.

werkt:
Any movement here? We need a commit in master before we can get ourselves into the running for 3.1

jin:
Yes. Phillipp is OOO so I'm requesting for another reviewer internally.

laurentlb:
The fix was cherrypicked on Friday.
If you try the release candidate, let me know if there's any other blocker (we might do the release tomorrow, but can delay it if it helps).

werkt:
@laurentlb we've been building since rc3 was cut and haven't found any other blockers, thanks for the cherry pick!

Issue #10003, opened 2019-10-12
Windows installation experience is subpar


mmrath:
### Description of the problem / feature request:

The current windows installation experience is subpar. I was trying to get some internal projects to Bazel to show the wider team. I generally use MacOS but in my company Windows is default. When I tried to install on windows in company environment I was stumbled upon too many issue. Reporting some important once here. 

1. Dependency on MSYS, some of the dev don't even know what it is. And in my company environment you cannot install anything apart from `C:\Installs` and regular `Program Files` directory. Need to go through infra and security for any exceptions.
2.  If some how I install MSYS the next problem is to control Bazel extract directory from where bazel server runs. Any executable will need to be on `C:\Installs` for it to execute, but bazel's work directory is not there
 

### Feature requests: what underlying problem are you trying to solve with this feature?

One click installer which installs everything to get started with Bazel

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Install Bazel on windows as per documentation

### What operating system are you running Bazel on?

Windows

### What's the output of `bazel info release`?

Not able to install bazel

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

NA

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

NA

###  Have you found anything relevant by searching the web?

Yes but solution is not good enough for large scale deployment

### Any other information, logs, or outputs that you want to share?

NA

pauldraper:
FYI circa 0.24.0, Bazel itself no longer requires MSYS.

Many rules however do rely on POSIX utilities.

mmrath:
@pauldraper  but Bazel documentation mentions MSYS2 as prerequisite.

laszlocsomor:
@mmrath : You're right, the docs are stale, let me update them. Bazel 1.0 doesn't need MSYS anymore unless you build or run targets that need Bash (e.g. `genrule` and `sh_test`). The only hard requirement is the Visual C++ Redistributble DLLs.

Re: configuring output directories: I agree a one-click installer would be easier, but in the meantime you can use certain flags to control this:
- [--output_user_root](https://docs.bazel.build/versions/master/command-line-reference.html#flag--output_user_root) flag defines the root directory where Bazel will extract its embedded files and where it will write the output by default.
- [--output_base](https://docs.bazel.build/versions/master/command-line-reference.html#flag--output_base) flag sets the output directory. 

Demo:
```
[...]

PR #10041, opened 2019-10-15
Docs: remove MSYS requirement for Windows

laszlocsomor:
See https://github.com/bazelbuild/bazel/issues/10003

Change-Id: I573000c150bd7a273ec0d2ff3389e774677cb6d5

### DIFF
--- a/site/docs/install-windows.md
+++ b/site/docs/install-windows.md
@@ -17,21 +17,12 @@ Also supported:
 
 *   64 bit Windows Server 2008 R2 or newer
 
+*   Older Windows 10 versions, disabled "Developer Mode" (enabling the mode just lets you use the
+    `--enable_runfiles` Bazel flag)
+
 ## 2. Install the prerequisites
 
 *   [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
-
-*   [MSYS2 x86_64](https://www.msys2.org/)
-
-    You should use the default install path.
-
-*   MSYS2 packages
-
-    Open the MSYS2 terminal, and run this command:
-
-    ```
-    pacman -S zip unzip patch diffutils git
-    ```
 
 ## 3. Download Bazel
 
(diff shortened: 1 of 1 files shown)


site/docs/install-windows.md
```diff
@@ -42,16 +33,67 @@ Recommended: rename this binary to `bazel.exe` and move it to a directory on the
...
+## 4. Optional: configure output directories
+
+**You can skip this step. Bazel can work without configuring the output directories, and will use
+its default values.**
+
+By default, Bazel writes to two directories:
+
+-   The "output user root", configurable with the `--output_user_root` flag.
+
+    This is where Bazel extracts from itself some embedded tools, its own Java runtime, and where it
```
meteorcloudy: Nit: Hmm, do we need to mention the Java runtime? Some people may not like this, although it's not a really issue for them.

site/docs/install-windows.md
```diff
@@ -68,7 +110,7 @@ We recommend installing:
 
     You will need this to build Java code on Windows.
 
-    Also supported: Java 8, 9
+    Also supported: Java 8 and 9
 
 *   [Python 2.7 for Windows x86-64](https://www.python.org/downloads/windows/)
```
meteorcloudy: Can we change the default to Python 3?

laszlocsomor: I'm wary of recommending py3 unless we know it works. Py2 is deprecated but it works.

Which version do we use on CI? If it's py3, that'd give me confidence that we can recommend it.

site/docs/install-windows.md
laszlocsomor: Good idea. I'm not happy about omitting it (why should we hide it?), but I know it has been controversial (for questionable reasons). Removing.

site/docs/install-windows.md
meteorcloudy: We installed both versions, but use python3 as default on Windows machines. I'm also using Python3 for everyday develop, so it should be fine. But we can also do it after the python 3 migration we're going to do.

laszlocsomor: SGTM, let's defer updating this section.

PR #10044, opened 2019-10-16
Add tool for dumping remote execution grpc log

keith:
This is a small script for dumping a human readable version of the log
produced by `--experimental_remote_grpc_log`.

Usage:

```sh
$ bazel run tools/remote_execution_log_dump path/to/log
metadata {
  tool_details {
    tool_name: "bazel"
    tool_version: "1.0.0"
  }
  action_id: "capabilities"
  tool_invocation_id: "..."
  correlated_invocations_id: "..."
}
...
```

### DIFF
--- a/src/main/protobuf/BUILD
+++ b/src/main/protobuf/BUILD
@@ -182,6 +182,11 @@ proto_library(
     ],
 )
 
+cc_proto_library(
+    name = "remote_execution_log_cc_proto",
+    deps = [":remote_execution_log_proto"],
+)
+
 java_proto_library(
     name = "remote_execution_log_java_proto",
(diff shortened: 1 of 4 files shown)


buchgr:
Thanks @keith. I think that works. A better solution might be to instead add a flag that dumps it in a human readable format in the first place?

keith:
That would be fine too yea

buchgr:
@keith I think I'd prefer that over adding a tool?

buchgr:
Closing this in favor of adding a flag to write it in a human readable format.

keith:
Should I file a separate ticket to track that flag?

Issue #10039, opened 2019-10-15
Support custom skylark build flags in external dependencies


moroten:
### Description of the problem / feature request:

The implementation of Skylark support for custom build flags
(#5577) does not work for custom build flags in external repositories.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Given the following setup:
```
mkdir my-external
touch my-external/WORKSPACE nothing.txt myext.txt top.txt

cat >my-external/BUILD.bazel <<EOF
load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
bool_flag(
    name = "flag_in_myext",
    build_setting_default = False, visibility = ["//visibility:public"],
)
EOF

cat >BUILD.bazel <<EOF
load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
bool_flag(
    name = "flag_at_top",
    build_setting_default = False,
)
alias(
    name = "alias_for_flag_in_myext",
    actual = "@myext//:flag_in_myext",
)
config_setting(
    name = "is_flag_at_top_enabled",
    flag_values = {":flag_at_top": "True"},
)
config_setting(
    name = "is_flag_in_myext_enabled",
    flag_values = {":alias_for_flag_in_myext": "True"},
)
filegroup(
    name = "everything",
    srcs = select({
        ":is_flag_at_top_enabled": ["top.txt"],
        ":is_flag_in_myext_enabled": ["myext.txt"],
        "//conditions:default": ["nothing.text"],
    }),
)
EOF

cat >WORKSPACE <<EOF
local_repository(name = "myext", path = "my-external")
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "bazel_skylib",
    url = "https://github.com/bazelbuild/bazel-skylib/archive/b113ed5d05ccddee3093bb157b9b02ab963c1c32.zip",
    sha256 = "cea47b31962206b7ebf2088f749243868d5d9305273205bdd8651567d3e027fc",
    strip_prefix = "bazel-skylib-b113ed5d05ccddee3093bb157b9b02ab963c1c32",
)
EOF
```
the output looks like
```
$ bazel cquery 'kind("source file", deps(:everything))'
//:nothing.text (null)
$ bazel cquery 'kind("source file", deps(:everything))' --//:flag_at_top
//:top.txt (null)
[...]

PR #10052, opened 2019-10-17
Support flag overrides of config settings in external repositories.

jmillikin-stripe:
The user-defined flags documented at https://docs.bazel.build/versions/1.0.0/skylark/config.html can't currently be set across repositories with `--@some_repo//:some_flag`. It looks like this is just a simple oversight in the flag parser, and after changing it to accept `--@` as a Starlark flag prefix the rest of the functionality is working.

Fixes https://github.com/bazelbuild/bazel/issues/10039

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/BlazeCommandDispatcher.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/BlazeCommandDispatcher.java
@@ -653,6 +653,7 @@ public class BlazeCommandDispatcher implements CommandDispatcher {
         OptionsParser.builder()
             .optionsData(optionsData)
             .skippedPrefix("--//")
+            .skippedPrefix("--@")
             .allowResidue(annotation.allowResidue())
             .build();
     return parser;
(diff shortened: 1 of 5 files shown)


gregestren:
That's a surprisingly simple fix. Looks solid to me.

Could you add tests covering the two problem scenarios from https://github.com/bazelbuild/bazel/issues/10039 (`--@myext//:flag_in_myext` and `--//:alias_for_flag_in_myext`)?

Maybe at https://github.com/bazelbuild/bazel/blob/7fa64c493a42e41e10aea48b3fc4e4197d75cd65/src/test/java/com/google/devtools/build/lib/skylark/StarlarkOptionsParsingTest.java for the basic syntax and https://github.com/bazelbuild/bazel/blob/77497817b011f298b7f3a1138b08ba6a962b24b8/src/test/shell/integration/starlark_configurations_test.sh for the alias?

jmillikin-stripe:
Tests added.

Aliases don't work for `config_setting` regardless of which workspace they're defined in, so I've only tested the non-alias case from that issue.

Issue #9165, opened 2019-08-13
Expose pdb output from cc_common.link on Windows


maximyurchuk:
### Description of the problem / feature request:
There is no pdb output on Windows in `cc_common.link`. This information should be added. 

### Feature requests: what underlying problem are you trying to solve with this feature?
I need bazel rule which builds shared libraries. Rule need satisfy following properties:
1) Rule provides import libraries and pdb on windows
2) I can write my `BUILD` files in crossplatform style (no ".dll" in label names)
3) I can use custom names for output files (for example same label has output name foo.dll on windows and libbar.so on linux)
4) It has transitive linkage

Seems using `cc_common.link` in my rule is a good choice, but it does not provide pdb files as a result. 

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
Following code shows that there is no pdb
**my_rule.bzl**
```
load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")

def print_val(prefix, name, type, value):
    # print("{}.{}, type: {}, value: {}".format(prefix, name, type, value))
    print("{}.{} = {}".format(prefix, name, value))
 
def dump(obj, prefix):
    for field in dir(obj):
        attr_value =  getattr(obj, field)
        attr_value_type =  type(getattr(obj, field))
        print_val(prefix, field, attr_value_type, attr_value)
        dump2(attr_value, prefix + "." + field)
    
def dump2(obj, prefix):        
    for field in dir(obj):
        attr_value =  getattr(obj, field)
        attr_value_type =  type(getattr(obj, field))        
        print_val(prefix, field, attr_value_type, attr_value)
    

def _cc_bin_impl(ctx):
    cc_toolchain = find_cpp_toolchain(ctx)
    feature_configuration = cc_common.configure_features(
        ctx = ctx,
        cc_toolchain = cc_toolchain,
        requested_features = ctx.features,
        unsupported_features = ctx.disabled_features,
    )
    
    (_compilation_context, compilation_outputs) = cc_common.compile(
        name = ctx.label.name,
        actions = ctx.actions,
        feature_configuration = feature_configuration,
        cc_toolchain = cc_toolchain,
        srcs = ctx.files.srcs,
        compilation_contexts = [],
    )
    
    linking_outputs = cc_common.link(
        name = ctx.label.name,
        actions = ctx.actions,
        feature_configuration = feature_configuration,
        cc_toolchain = cc_toolchain,
[...]

PR #10065, opened 2019-10-18
expose pdb file from cc_common.link

maximyurchuk:
#9165 expose pdb file from cc_common.link

cc_common.link now has PDB file in linking_outputs.pdb_file


src/main/java/com/google/devtools/build/lib/rules/cpp/Link.java
```diff
@@ -156,6 +156,14 @@ private Link() {} // uninstantiable
         ArtifactCategory.INTERFACE_LIBRARY,
         Executable.NOT_EXECUTABLE),
 
+    /** A pdb file for dynamic library. */
+    PDB_FILE(
+        LinkerOrArchiver.LINKER,
+        CppActionNames.CPP_LINK_DYNAMIC_LIBRARY,
```
maximyurchuk: Don't know how to declare output for both CPP_LINK_EXECUTABLE and CPP_LINK_DYNAMIC_LIBRARY in one enum. But seems it works fine.

Can someone take an advice about right way to fix that?

maximyurchuk: Actully not used now.

maximyurchuk:
@oquenchil reasonable argumentation for me. But how close is this "future"?:) 

Anyway, I have another proposal for pdb generation in cc_common.link:
- I will manually add /PDB:foo.pdb and other flags in `cc_common.link` in `user_link_flags` argument in my code
- In that case there another field need to be added in cc_common.link for additional outputs. Because pdb output need be accessable from starlark. 

Is it good for you? If yes, I will create an issue.

oquenchil:
A good way to solve this is a getter in linking outputs for any file/s related to debugging, whatever that is. If you want to submit a PR where linking outputs has a generic getter for these debugging files (.dwo, .pdb, etc.), with tests for Windows, Linux and Mac, then you are welcome to do it.

It will be a while before we can work on this, not in the near future.

maximyurchuk:
@oquenchil could you explain what do you mean by "getter in linking outputs for any file/s related to debugging, whatever that is."? As I understood you said it about starlark build api, if it is true could you give an example of interface which is acceptable? For example I can change `pdb_file` in api to `symbol_files` (or `debug_files`) which can be different on different platforms (on windows it will be list with one pdb, on gcc it will be dwo file, on macos something else). If this acceptable, imho it is not necessary to implement all of these types in one pull request, because missing files on other platform can be added later.

oquenchil:
Yes, that's exactly what I mean. A more generic name that can be used for other platforms. If we expose this, we'd like it to be exposed for debugging artifacts of every platform for which we have debugging artifacts internally. Otherwise, when Linux users see that in the API they will wonder why they are not getting any debugging files for Linux.

maximyurchuk:
@oquenchil as I can see internally bazel has support (regarding С++) only for 
1) pdb for windows 
2) dwo/dwp for gcc


Also 'dsym' debug symbols exists on macos, but I can't find any mentions about that in rules/cpp, only in objc  rules. 

Is it acceptable to implement new API only for .dwo/.dwp and .pdb file types? (since only these types are internally suported for C++ rules for now). Or I missed some types?

oquenchil:
Nope, that sounds good!

So a field in linker_outputs that exposes .pdb and .dwo/.dwp files. As opposed to passing the file as an input to the API and using that file.

aiuto:
friendly ping to everyone

maximyurchuk:
Sorry, no time for this. I suppose it is better close PR and recreate another if I return to this issue.

PR #10072, opened 2019-10-21
Update "six" library from 1.10.0 to 1.12.0

laszlocsomor:
The newer version has features we need to easily
migrate our .py scripts from PY2 to PY3 syntax.

Change-Id: Ifee50f35ef63796911de1de1459332ae174b77d0


laszlocsomor:
Fixed by https://github.com/bazelbuild/bazel/commit/866133e8f9d53691761e55e32ca9fafbdb34550a

Issue #10027, opened 2019-10-14
A test_suite that references an invalid BUILD file does not fail when tests are run


cheister:
### Description of the problem / feature request:

If a `test_suite` references a target from an invalid BUILD file it will report the invalid BUILD file error but tests will succeed. 

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

The following files reproduce the issue:

`BUILD`
```
test_suite(
    name = "test",
    tests = ["//src/test/java:test"]
)
```

`src/test/java/BUILD`
```
random_undefined()

test_suite(name = "test")
```

```
>  bazel test //:test
INFO: Writing tracer profile to '/private/var/tmp/_bazel_cheister/52b334e042662833bef6eb9e0e6e2c56/command.profile.gz'
ERROR: /private/tmp/bazel-test/src/test/java/BUILD:1:1: name 'random_undefined' is not defined
ERROR: /private/tmp/bazel-test/src/test/java/BUILD:1:1: name 'random_undefined' is not defined
INFO: Analyzed 0 targets (2 packages loaded, 0 targets configured).
INFO: Found 0 test targets...
INFO: Elapsed time: 0.118s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
```

### What operating system are you running Bazel on?

`Darwin Kernel Version 18.7.0`

### What's the output of `bazel info release`?

`release 1.0.0-homebrew`

PR #10080, opened 2019-10-22
Handle packages with errors in TestsInSuiteFunction.

benjaminp:
Clients of PackageFunction have to be careful to handle partially valid package results. Packages in an error state at least need to be flagged, so higher levels don't think the whole build succeeded. If --nokeep_going is in effect, an exception should generally be raised to abort the build.

Fixes https://github.com/bazelbuild/bazel/issues/10027.


haxorz:
i'm pretty busy this week, but i'll try to get to this review by EOD friday.

@ulfjack would also be a good person to do the review.

ulfjack:
Looks good to me. Was 'only' missing a hasError = true there, aborting if --nokeep_going is a nice touch.

haxorz:
thanks so much for looking at this, ulf!

PR #10100, opened 2019-10-25
Remove //third_party/py/abseil

laszlocsomor:
This makes migrating Bazel's Python scripts to PY3
easier, because we don't have to update to a newer
Abseil version.

It also makes Bazel's distribution archive smaller
and speeds up git checkouts (except shallow ones).

No package depends on Abseil since e92b2d997875db8db70e0c4ff37e52a94c0d4a35.

Change-Id: I8e9a57ed4ba96bb2dca714f612cf66ead58303d4


philwo:
This was closed in https://github.com/bazelbuild/bazel/commit/a7bc98f5f5f65a5982c4c72735f6f1a75a5e7259.

Issue #8778, opened 2019-07-02
default_constraint_value should not add a constraint to toolchains


jayconrod:
### Description of the problem / feature request:

This issue came up when analyzing bazelbuild/rules_go#2115. Two toolchains are needed to build a target: a Go toolchain, and a Docker toolchain. The target platform is `@io_bazel_rules_go//go/toolchain:linux_amd64`, which has the following constraints:

* `@bazel_tools//platforms:linux`
* `@bazel_tools//platforms:x86_64`
* `@io_bazel_rules_go//go/toolchain:cgo_off` (the constraint setting is `cgo_constraint` in the same package)
* `@io_bazel_rules_go//go/toolchain:not_darwin` (the constraint setting is `darwin_constraint` in the same package)

`cgo_constraint` and `darwin_constraint` have `default_constraint_value` set to `cgo_on` and `is_darwin`, respectively. This is so that when the default target platform is used, these values can be used in `select` statements. In particular, `cgo_on` is used to add an additional dependency on a target that requests a C/C++ toolchain.

The problem is that Bazel rejects all registered Docker toolchains for the target platform. These toolchains don't specify any constraints related to `cgo_constraint` or `darwin_constraint`, since those settings are irrelevant to Docker. It appears that because of the `default_constraint_value` attribute in the definition of these settings, the Docker toolchains implicitly require `cgo_on` and `is_darwin`. Because of this, Docker toolchains cannot be used:

```
INFO: ToolchainResolution: Looking for toolchain of type @io_bazel_rules_docker//toolchains/docker:toolchain_type...
INFO: ToolchainResolution:   Considering toolchain @docker_config//:toolchain...
INFO: ToolchainResolution:     Toolchain constraint @io_bazel_rules_go//go/toolchain:darwin_constraint has value @io_bazel_rules_go//go/toolchain:is_darwin, which does not match value @io_bazel_rules_go//go/toolchain:not_darwin from the target platform @io_bazel_rules_go//go/toolchain:linux_amd64
INFO: ToolchainResolution:     Toolchain constraint @io_bazel_rules_go//go/toolchain:cgo_constraint has value @io_bazel_rules_go//go/toolchain:cgo_on, which does not match value @io_bazel_rules_go//go/toolchain:cgo_off from the target platform @io_bazel_rules_go//go/toolchain:linux_amd64
INFO: ToolchainResolution:   Rejected toolchain @docker_config//:toolchain, because of target platform mismatch
```

[...]

Issue #8274, opened 2019-05-09
Toolchain exec/target_compatible_with using default_constraint_value


m3rcuriel:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, ask it on StackOverflow instead: https://stackoverflow.com/questions/tagged/bazel
> - if this is a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

I am not sure if this is a bug or a feature, but currently toolchains with exec_compatible_with/target_compatible_with seem to pull in a default constraint value. This means that you can't do something like the example below.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
constraint_setting(
    name = "distro",
    default_constraint_value = ":ubuntu-bionic"
)

constraint_value(
   name = "ubuntu-bionic",
   constraint_setting = ":distro",
) 

constraint_value(
   name = "ubuntu-trusty",
   constraint_setting = ":distro",
) 

constraint_value(
   name = "ubuntu-disco",
   constraint_setting = ":distro",
) 
```

Now the behavior I would expect would be that if I defined a toolchain with
```
exec_compatible_with = [
    "@//bazel_tools/platforms:whatever",
]
```
to work just fine. However, it uses the `default_constraint_value` for the toolchain instead.

This is presumably because
```
216:        toolchainConstraints.diff(platform.constraints());
```
is used to resolve the toolchains, and diff falls through on a default value on _either_ `other.constraints()` or `this.constraints()`, and so the diff finds the default value.

Like I said, I'm not sure if this is a feature or a bug or if my intuition for `constraint_settings` is wrong. The issue from my point of view is that in order for toolchains to resolve correctly they must list out a value for every single constraint setting. This only makes sense if constraint settings are exclusively used to distinguish toolchains, in _all toolchain_types_.

### What operating system are you running Bazel on?

Linux not4u 5.0.8-arch1-1-ARCH #1 SMP PREEMPT Wed Apr 17 14:56:15 UTC 2019 x86_64 GNU/Linux

### What's the output of `bazel info release`?
 AUR:

Build label: 0.25.0- (@non-git)
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
[...]

PR #10102, opened 2019-10-25
Avoid breaking toolchain resolution if platform has defaulted constraint.

jmillikin:
### Summary
See https://github.com/bazelbuild/bazel/issues/8778 and https://github.com/bazelbuild/bazel/issues/8274 for context / motivation. In summary, the presence of a `constraint_setting` with `default_constraint_value` in a Platform's constraints will cause toolchain resolution to break for all toolchains that do not explicitly set that constraint value. This makes `default_constraint_value` dangerous because setting it can break other rulesets composed into the same top-level workspace.

cc @jayconrod @m3rcuriel @katre

### Test Plan
I updated the unit test `SingleToolchainResolutionFunction`, verified it failed, then altered the implementation to pass. I also built `//src:bazel` and ran it with a test workspace to verify the end result worked as expected.

Test workspace:
```
$ cat WORKSPACE
register_toolchains("//:my_toolchain")
$ cat BUILD
load(":rules.bzl", "my_toolchain_info", "my_rule")

my_rule(name = "hello")
toolchain_type(name = "my_toolchain_type")
my_toolchain_info(name = "my_toolchain_info")
toolchain(
    name = "my_toolchain",
    toolchain = "//:my_toolchain_info",
    toolchain_type = "//:my_toolchain_type",
)

constraint_setting(
    name = "setting",
    default_constraint_value = ":value_a",
)
constraint_value(
    name = "value_a",
    constraint_setting = ":setting",
)
constraint_value(
    name = "value_b",
    constraint_setting = ":setting",
)
platform(
    name = "platform_a",
    constraint_values = ["//:value_a"],
)
platform(
    name = "platform_b",
    constraint_values = ["//:value_b"],
)

$ cat rules.bzl
def _my_toolchain_info(ctx):
    return platform_common.ToolchainInfo()

my_toolchain_info = rule(_my_toolchain_info)

def _my_rule(ctx):
    pass

my_rule = rule(_my_rule, toolchains = ["//:my_toolchain_type"])
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
@@ -185,12 +185,26 @@ public abstract class ConstraintCollection
       return true;
     }
 
-    // Then, check the parent, directly to ignore defaults.
+    // Then, check the parent.
     if (parent() != null) {
       return parent().has(constraint);
     }
 
     return constraint.hasDefaultConstraintValue();
+  }
+
+  public boolean hasWithoutDefault(ConstraintSettingInfo constraint) {
+    // First, check locally.
+    if (constraints().containsKey(constraint)) {
+      return true;
+    }
+
+    // Then, check the parent, directly to ignore defaults.
+    if (parent() != null) {
+      return parent().hasWithoutDefault(constraint);
+    }
+
+    return false;
   }
 
   /**
--- a/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
@@ -214,7 +214,21 @@ public class SingleToolchainResolutionFunction implements SkyFunction {
     // Check every constraint_setting in either the toolchain or the platform.
     ImmutableSet<ConstraintSettingInfo> mismatchSettings =
         toolchainConstraints.diff(platform.constraints());
+    boolean matches = true;
     for (ConstraintSettingInfo mismatchSetting : mismatchSettings) {
+      // If a constraint_setting has a default_constraint_value, and the platform
+      // sets a non-default constraint value for the same constraint_setting, then
+      // even toolchains with no reference to that constraint_setting will detect
+      // a mismatch here. This manifests as a toolchain resolution failure (#8778).
+      //
+      // To allow combining rulesets with their own toolchains in a single top-level
+      // workspace, toolchains that do not reference a constraint_setting should not
+      // be forced to match with it.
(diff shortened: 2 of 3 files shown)


jmillikin:
It looks like there's a test asserting the current behavior, added in https://github.com/bazelbuild/bazel/commit/b629238a8af71ae5262e69e423d18cbe8f0687af by @katre.

Is there any way we could get this changed during the v1.0 series, or are we stuck with it until >= v2.0?

m3rcuriel:
I personally am all for this and I like how concise the change is.

gregestren:
Sanity check: does https://docs.bazel.build/versions/1.1.0/be/platform.html#constraint_setting.default_constraint_value imply to you that what you propose is already the intended behavior?

jmillikin:
My read of
> If a constraint setting has a default value, then whenever a platform does not include any constraint value for that setting, it is the same as if the platform had specified the default value. Otherwise, if there is no default value, the constraint setting is considered to be unspecified by that platform. In that case, the platform would not match against any constraint list (such as for a `config_setting`) that requires a particular value for that setting.
is that the behavior of default values is specified with regard to platforms, but unspecified with regard to toolchains. So both "use" and "ignore" are permitted by the docs. The test suite is asserting behavior not covered by the docs.

gregestren:
Thanks.

What you describe makes sense to me. That's certainly an awkward user experience. Two issues give me pause:

1) John's comment at https://github.com/bazelbuild/bazel/issues/8274#issuecomment-491374361
2) This would be a backwards-incompatible change, so I think it'd have to be Bazel 2.0-gated.

Under the circumstances I'd like to get support from @katre before proceeding. He's on vacation but will be back Nov 4. Can we pick up the conversation then?

katre:
This change looks good in general. I think that this is not a change that breaks backwards compatibility (in that the only changes are that builds that previously did not work now will work). However, I'd like to run the change against all our downstream CI projects to verify this.

@jmillikin: Can you update the change to fix the failing test in toolchain_test.sh? Once you have a green CI run there, I will handle scheduling the downstream projects run and we can see what the result looks like and decide whether or not this requires an incompatible flag.

jmillikin:
> Can you update the change to fix the failing test in toolchain_test.sh? Once you have a green CI run there, I will handle scheduling the downstream projects run and we can see what the result looks like and decide whether or not this requires an incompatible flag.

Done.

katre:
Thanks, tests currently running at https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1265

katre:
Re-run of the tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1266
The tensorflow failure seems to have been because the first attempt was 100 commits behind master, and missing a needed bugfix.

katre:
Everything passes, so I think we can submit this as-is. I'll merge it internally.

PR #10103, opened 2019-10-25
Restore //third_party/py/abseil

laszlocsomor:
Apparently python-gflags is deprecated and advises
to use abseil-py.

See https://pypi.org/project/python-gflags

I looked at argparse, it's nice and does all we
need, but only supported above Python 3.2.

Requiring that would be an incompatible change,
while using Abseil allows running tools both with
PY2 and PY3. This is important for running the
Android tools, for example.

This reverts commit a7bc98f5f5f65a5982c4c72735f6f1a75a5e7259.


laszlocsomor:
I tried updating `//third_party/py/gflags` and noticed it was deprecated.
It seems we need abseil-py after all. Hopefully this means I can delete gflags instead.

philwo:
Can we just get rid of gflags and replace it with argparse?

laszlocsomor:
Hm, good idea. Let me see how hard would that be.

laszlocsomor:
I looked at `argparse`, it's nice and does all we need, but only supported above Python 3.2.

Requiring that would be an incompatible change, while using Abseil allows running tools both with PY2 and PY3. This is important for running the Android tools, for example.

philwo:
Thanks for your research!

Two thoughts on this:
- The world is dropping support for Python 2 end of this year (https://pythonclock.org/). Yes, it’d be an incompatible change, but luckily we have a major release lined up in time. We could do it.
- Python 2.7 also supports argparse (https://docs.python.org/2/library/argparse.html) and anything older than 2.7 is so incredibly out of date, it doesn’t matter.

I think migrating to argparse would be fine, WDYT?

laszlocsomor:
> Two thoughts on this:
> 
> * The world is dropping support for Python 2 end of this year (https://pythonclock.org/). Yes, it’d be an incompatible change, but luckily we have a major release lined up in time. We could do it.

Existing PY2 installations will keep working though, just not receive any updates, right? So users who don't update their runtime for whatever (potentially good) reason, would be broken.

To do the incompatible change, we somehow had to put this behind an incompatible flag. Somehow. I don't know how we could. So preserving PY2 compatibility seems like the least risky approach. WDYT?

> * Python 2.7 also supports argparse (https://docs.python.org/2/library/argparse.html) and anything older than 2.7 is so incredibly out of date, it doesn’t matter.

Oh, I completely missed that! Thanks for pointing that out.

> I think migrating to argparse would be fine, WDYT?

We could in the long run, and so reduce our dependencies and shrink the code base. Those would be undeniable benefits. However, I think that'd break Python 3.0 .. 3.2 users.

Abseil is available and migrating from gflags is trivial while migrating to argparse requires some work. So it seems there's no immediate benefit to migrate to argparse.

philwo:
Feel free to go with Abseil for now and we can discuss later when I’m back how to migrate to argparse in a second step :)

laszlocsomor:
SGTM!

laszlocsomor:
Closed by https://github.com/bazelbuild/bazel/commit/5ada1402f7945a7c0336e3ffc8b3e3dda93fc856

benjaminp:
BTW, I would not worry about Python 3.0 or 3.1. 3.1 was last released in 2012.

laszlocsomor:
@benjaminp : Thanks, that's long ago indeed. We can migrate to `argparse` after the PY3 migration's done.

PR #10105, opened 2019-10-25
Add rules_rust as recommended rules

hlopko:


laurentlb:
I discussed with rules_rust owners and decided not to include it here. The issue is that the rules are not actively developed, don't quite provide a first-class experience, and don't have many users.

hlopko:
Ack! Thank you!

Issue #7390, opened 2019-02-09
Host executable with non-empty DefaultInfo files missing from sandbox


jmillikin:
### Description of the problem / feature request:

Rules marked `executable = True` are allowed to return a `DefaultInfo` with non-empty `files`, which means that `bazel build //some:exec` will print a different output path than what gets invoked by `bazel run //some:exec`. This behavior is useful when writing rules for "script" binaries, when the content needed for hermetic execution might not function when copied outside of the Bazel execroot.

However, when such a binary target is used in host mode (i.e. as a dependency of an action), Bazel will construct the sandbox filesystem incorrectly. The path passed to `execvp()` is different than the file present on disk.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```python
# BUILD
load(":rules.bzl", "bin_rule", "out_rule")
bin_rule(name = "hello_bin")
out_rule(name = "hello_out")
```

```python
# rules.bzl
def _bin_rule(ctx):
    out_a = ctx.actions.declare_file(ctx.attr.name + ".a.sh")
    out_b = ctx.actions.declare_file(ctx.attr.name + ".b.sh")
    ctx.actions.write(
        output = out_a,
        content = "#!/bin/sh\necho 'hello a' > $@",
        is_executable = True,
    )
    ctx.actions.write(
        output = out_b,
        content = "#!/bin/sh\necho 'hello b' > $@",
        is_executable = True,
    )
    return DefaultInfo(
        files = depset(direct = [out_a]),
        executable = out_b,
    )

def _out_rule(ctx):
    out = ctx.actions.declare_file(ctx.attr.name + ".txt")
    ctx.actions.run(
        # I expect this to run `hello_bin.b.sh`
        executable = ctx.executable._hello_bin,
        outputs = [out],
        arguments = [out.path],
        mnemonic = "HelloOut",
    )
    return DefaultInfo(
        files = depset(direct = [out]),
    )

bin_rule = rule(_bin_rule, executable = True)
out_rule = rule(_out_rule, attrs = {
    "_hello_bin": attr.label(
        default = "//:hello_bin",
        executable = True,
        cfg = "host",
    ),
})
```

```
$ bazel-0.22 build -s //:hello_out
[...]
SUBCOMMAND: # //:hello_out [action 'HelloOut hello_out.txt']
(cd /private/var/tmp/_bazel_john/591a601d7d8436419d69a3dcb9b24d54/execroot/__main__ && \
[...]

Issue #6607, opened 2018-11-06
omitting the executable from data_runfiles crashes Bazel


benjaminp:
```
$ cat rule.bzl
def _broken_executable(ctx):
    other_file = ctx.actions.declare_file('other')
    ctx.actions.write(content = 'foo', output = other_file)
    ctx.actions.write(content = 'bar', output = ctx.outputs.executable)
    return struct(default_runfiles = ctx.runfiles(files=[other_file]))

broken_executable = rule(
    implementation = _broken_executable,
    executable = True,
)
$ cat BUILD
load(":rule.bzl", "broken_executable")

broken_executable(name = "broken")
$ bazel build //:broken
Internal error thrown during build. Printing stack trace: java.lang.RuntimeException: Unrecoverable error while evaluating node '//:broken BuildConfigurationValue.Key[b4f63d729fb5712a9002341842f4fb65] false' (requested by nodes )
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:499)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:368)
	at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
Caused by: java.lang.IllegalStateException: main program File:[[<execution_root>]bazel-out/k8-fastbuild/bin]broken not included in runfiles
	at com.google.devtools.build.lib.analysis.RunfilesSupport.create(RunfilesSupport.java:118)
	at com.google.devtools.build.lib.analysis.RunfilesSupport.withExecutable(RunfilesSupport.java:417)
[...]

PR #10110, opened 2019-10-26
Add tool executables (from FilesToRunProvider) to action inputs.

jmillikin:
Fixes https://github.com/bazelbuild/bazel/issues/7390

Summary: rules are allowed to set `DefaultInfo(files = depset(), executable = ...)`. If the resulting targets are used as tool inputs to another target, the executable will not be added to the downstream targets' action inputs. This manifests as either an `execvp()` error (from `ctx.actions.run()`) or a Bash error in genrules.

This PR fixes that so the use case of non-default executables works as expected. This is useful for building scripts with complex runfiles, where `bazel run //:something` and `bazel build //:something && bazel-bin/something` cannot be supported by the same output file.

cc @ccate @jmmv

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
@@ -44,9 +44,6 @@ public final class FilesToRunProvider
       NestedSet<Artifact> filesToRun,
       @Nullable RunfilesSupport runfilesSupport,
       @Nullable Artifact executable) {
-    if (executable != null) {
-      filesToRun = NestedSetBuilder.fromNestedSet(filesToRun).add(executable).build();
-    }
     this.filesToRun = filesToRun;
     this.runfilesSupport = runfilesSupport;
     this.executable  = executable;
--- a/src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
@@ -44,6 +44,9 @@ public final class FilesToRunProvider
       NestedSet<Artifact> filesToRun,
       @Nullable RunfilesSupport runfilesSupport,
       @Nullable Artifact executable) {
+    if (executable != null) {
+      filesToRun = NestedSetBuilder.fromNestedSet(filesToRun).add(executable).build();
+    }
     this.filesToRun = filesToRun;
     this.runfilesSupport = runfilesSupport;
     this.executable  = executable;
--- a/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
@@ -197,7 +197,6 @@ public final class CommandHelper {
         Artifact executableArtifact = tool.getExecutable();
         // If the label has an executable artifact add that to the multimaps.
         if (executableArtifact != null) {
-          resolvedToolsBuilder.add(executableArtifact);
           mapGet(tempLabelMap, label).add(executableArtifact);
           // Also send the runfiles when running remotely.
           toolsRunfilesBuilder.add(tool.getRunfilesSupplier());
--- a/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
@@ -197,6 +197,7 @@ public final class CommandHelper {
         Artifact executableArtifact = tool.getExecutable();
         // If the label has an executable artifact add that to the multimaps.
         if (executableArtifact != null) {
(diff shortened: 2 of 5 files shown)


lberki:
Well, that's an interesting one. I didn't expect that we allow the creation of a `FilesToRunProvider` where the executable is not in `getFiles()`. 

Since we use `FilesToRunProvider` in many places other than `CommandHelper`, I think this pull request only fixes a part of the problem. I see two viable solutions:

1. Make it an error to create a `FilesToRunProvider` where the executable is not in `files`. This would be an incompatible change and it appears that there are places where this is used.
2. Just silently add the executable to `FilesToRunProvider#getFiles()` when it's not already there. This would cause a minor amount of RAM wastage.

I think the best approach is to do something halfway between: do (1) for call sites from Java and (2) for call sites from Starlark. 

@laurentlb , WDYT?

benjaminp:
A similar situation happens with the presence of the executable in the runfiles. That case is handled with an implementation similar to what lberki suggests here. Broken native rules receive an `IllegalStateException`. Starlark rules typically implicitly have the executable added to their runfiles or receive a Starlark-level error in some more obscure cases (at least since #6607 fixed the Java-level crash leaking through).

jmillikin:
New commit added to implement "Just silently add the executable to `FilesToRunProvider#getFiles()` when it's not already there." -- there seems to be only one place in the code where `FilesToRunProvider` can get constructed  with that condition, and I couldn't figure out how to tell whether the caller is from Starlark or Java.

src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
```diff
@@ -44,6 +44,9 @@ public FilesToRunProvider(
       NestedSet<Artifact> filesToRun,
       @Nullable RunfilesSupport runfilesSupport,
       @Nullable Artifact executable) {
+    if (executable != null) {
+      filesToRun = NestedSetBuilder.fromNestedSet(filesToRun).add(executable).build();
```
lberki: This will create an extra `NestedSet` instance for every call site, most importantly, in the constructor of `FileConfiguredTarget`, which means an extra object for each input file, which is _probably_ unacceptable overhead.

How about special-casing at least `#fromSingleExecutableArtifact()` so that this overhead is not present?

Given that the only other call site is in `RuleConfiguredTargetBuilder#build()`, I'd prefer putting logic there because, well, that place already has logic. In addition to being a personal preference, you also have more information, notably, `ruleContext.getRule().getRuleClassObject().isSkylark()`, which you can use to determine whether this is a Starlark rule and thus adding the binary to the nested set is necessary.

Even better is changing `SkylarkRuleConfiguredTargetUtil#addSimpleProviders()` -- that place has control over both the `executable and `filesToBuild` fields of `RuleConfiguredTargetBuilder` and is specific to Starlark rules, so you can just add the "executable is forcibly added to `filesToRun`" logic there.

lberki:
Added a comment -- TL;DR: I think it's better to tweak `SkylarkRuleConfiguredTargetUtil#addSimpleProviders()` because that's specific to Starlark.

What I'm slightly unhappy about is that you probably can't just assert whether the executable is in `filesToRun` for Java rules because you'd need to iterate over the contents of the latter nested set just for error checking, which sounds wasteful.

src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
jmillikin: `addSimpleProviders()` seems to be used only for constructing `filesToBuild`, but this change is about adding the executable to `filesToRun`.

I've expanded `RuleConfiguredTargetBuilder::build()` (via `buildFilesToRun()`) to inject the executable. It's applied to all rules, since as you note it's difficult to figure out whether an `Artifact` is present in a `NestedSet` without iterating over it.

jmillikin:
`addSimpleProviders()` seems to be used only for constructing `filesToBuild`, but this change is about adding the executable to `filesToRun`.

I've expanded `RuleConfiguredTargetBuilder::build()` (via `buildFilesToRun()`) to inject the executable. It's applied to all rules, since as you note it's difficult to figure out whether an `Artifact` is present in a `NestedSet` without iterating over it.

src/main/java/com/google/devtools/build/lib/analysis/FilesToRunProvider.java
jmillikin: > Given that the only other call site is in `RuleConfiguredTargetBuilder#build()`, I'd prefer putting logic there because, well, that place already has logic.

Done.

jmillikin:
Gentle ping @iberki

jmillikin:
Gentle ping @lberki

philwo:
@lberki As requested, here's your ping. :) Please let me know if I can import and cherry-pick this into Bazel 1.2.0.

Issue #7255, opened 2019-01-25
bazel doesn't work when $HOME path contains non-ASCII characters


davispuh:
### Description of the problem / feature request:

bazel doesn't work when $HOME path contains non-ASCII characters

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
$ mkdir /tmp/tēst
$ export HOME=/tmp/tēst
$ bazel version
```

### What operating system are you running Bazel on?

ArchLinux

### What's the output of `bazel version`?

```
Build label: 0.21.0- (@non-git)
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 20 19:15:52 2018 (1545333352)
Build timestamp: 1545333352
Build timestamp as int: 1545333352
```

### Any other information, logs, or outputs that you want to share?

Note that `LANG=lv_LV.UTF-8`

When using ArchLinux provided binary
```
[bazel INFO src/main/cpp/option_processor.cc:359] Looking for the following rc files: /etc/bazel.bazelrc,/mnt/Build/bazel-git/src/bazel-git/.bazelrc,/tmp/tēst/.bazelrc
[bazel INFO src/main/cpp/blaze.cc:1552] Debug logging requested, sending all client log statements to stderr
[bazel INFO src/main/cpp/blaze.cc:1572] Acquired the client lock, waited 0 milliseconds
[bazel USER src/main/cpp/blaze.cc:853] Starting local Bazel server and connecting to it...
[bazel USER src/main/cpp/blaze.cc:892] Server crashed during startup. Now printing /tmp/tēst/.cache/bazel/_bazel_Viesis/49d083d1a0368b9a8f85b056b4534a36/server/jvm.out
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: /tmp/t??st/.cache/bazel/_bazel_Viesis/install/90846db665bd444d85cd9a305cffe931/_embedded_binaries/A-server.jar
        at java.base/sun.nio.fs.UnixPath.encode(UnixPath.java:145)
        at java.base/sun.nio.fs.UnixPath.<init>(UnixPath.java:69)
        at java.base/sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:280)
        at java.base/java.io.File.toPath(File.java:2290)
[...]

Issue #4555, opened 2018-02-01
Darwin: different encodings expected, even for intern vfs library


aehlig:
Internally, our libraries, when executed on Darwin, expect two different encodings. Everything that goes through native code, i.e., everything where our `C` code calls `fopen(3)` expects bazel's internal standard of using the unicode carachters 0 to 255 to represent a sequence of octets as a Java String. Everything that goes through a standard java library (that then does the system call) expects a unicode String, that, when encoded UTF-8, gives the desired octets (note that these are less possible file names).

That leads to the need of strange reencodings, as in 7e6837cc1d1aa4259f5c27ba3606b277b5f6c3e9. Also note, that even bazel's internal `vfs` library didn't standardise on a single encoding: `Path#chmod` expects the first, `Path#asByteSink` the second.

aehlig:
Choosing category "misc/misc" as the closest proxy to "internal cleanup". Not tagging it MacOs specific, as most likely a similar issue is present on Windows.

Issue #374, opened 2015-08-13
$ are not allowed in filename of a filegroup


damienmg:
Ultimately any character can be part of a filename. We should probably allow that.

Some mangling to generate the corresponding label should probably be done.

Original report on the mailing-list:
 https://groups.google.com/d/msgid/bazel-discuss/CAN0GiO3__5jXo5rZqroSj0mFxpqCzUZZVkY%3DSNsJK1%2BZ1BdJLg%40mail.gmail.com

abergmeier:
1. So are we talking Unicode?
2. Where does _any character_ stop?
3. how do you treat characters that are not allowed on certain platforms?
4. When using mangling, how do you handle collisions?

kayasoze:
In POSIX, filenames are "bags of bytes"--there is no encoding; however, `NUL` and `/` are not allowed.  Windows has a few more restrictions.  Perhaps the BUILD file should be parsed in the encoding of the system locale, usually `UTF-8`, and filenames run though a `ValidForCurrentPlatform()` function which checks for disallowed characters.  However, opting for strict platform neutrality in this way means that Bazel would have to represent filenames as a bag of bytes and not a Unicode string, as there is no guarantee that the filename will roundtrip through Unicode correctly.  The problem can probably be simplified by restricting filenames to be `UTF-8 or `UTF-16`, which should cover most people's needs even though that's not strict POSIX.

ulfjack:
Well, I think we can probably require valid UTF-8 file names and strongly recommend that people use UTF-8 for their file system. For labels / BUILD files, we probably need an escaping scheme, at least for the control characters. If there's a file that isn't valid UTF-8, we give an error message?

btelles:
Our company codes mainly in C++, but our frontend uses a lot of JS and nodejs modules which have all sorts of characters in the filenames--for example, -, #, @, (, and ).

Right now this is a major blocker for getting all our codebase under one build system since we can't reference files with semi-special characters. I don't think Bazel should decide what characters are acceptable in file names, as that reduces file names to those that fit both (1) supported languages and (2) supported platforms. This seems unnecessarily restrictive, and is becoming a major pain point for us.

ulfjack:
Agreed. Unfortunately, it's a bit tricky to fix, as a lot of code assumes that the mapping from labels to file names (and vice versa) is trivial, and doesn't require escaping. Any suggestions on an escaping scheme?

damienmg:
URL based?

[...]

PR #10111, opened 2019-10-27
Support non-ASCII source artifact paths on UNIX platforms.

jmillikin:
Fixes https://github.com/bazelbuild/bazel/issues/7255, and mitigates some of the most annoying limitations of https://github.com/bazelbuild/bazel/issues/374.

Summary: On UNIX platforms, Bazel uses `readdir()` via `NativePosixFiles` but opens paths with `java.io.File`. These two libraries use different representations of non-ASCII filesystem paths, which prevents Bazel from reading source artifacts.

There is a workaround, because `java.io.File` can also accept the path as a `URI` with percent-encoded octets. Using this mechanism for paths containing characters outside the ASCII range allows Bazel to happily consume source artifacts with Unicode filenames.

cc @davispuh for https://github.com/bazelbuild/bazel/issues/7255
cc @aehlig for https://github.com/bazelbuild/bazel/issues/4555
cc @alandonovan who, per https://github.com/bazelbuild/bazel/issues/374#issuecomment-523120425, was working on a fix but ran into unknown difficulties.


jmillikin:
Also, small note: due to limitations within `java.io` this only supports Unicode paths, not the "arbitrary blob of bytes" semantics found on Linux or BSD. According to https://stackoverflow.com/questions/14171565/java-read-write-unicode-utf-8-filenames-not-contents we'd have to switch Bazel to opening files with `java.nio`, which is a larger change than I wanted to make.

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
```diff
@@ -48,35 +52,65 @@ protected InputStream getInputStream(Path path) throws IOException {
...
+      } catch (AccessDeniedException e) {
+        throw new IOException(e.getMessage() + " (Permission denied)", e);
       }
+      break;
     }
+
+    // On some platforms, whether a path can be read as a file won't be checked until
+    // the first read().
+    //
+    // http://mail.openjdk.java.net/pipermail/nio-dev/2014-December/002877.html
```
lberki: wat

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
```diff
@@ -48,35 +52,65 @@ protected InputStream getInputStream(Path path) throws IOException {
...
+    // The underlying stream might not be seekable, and wrapping in a BufferedInputStream
+    // causes test failures in the integration suite, so create our own little one-byte
+    // buffer here.
+    byte[] buf = new byte[1];
+    int buflen = 0;
+    try {
+      buflen = stream.read(buf, 0, 1);
+    } catch (IOException e) {
+      if (e.getMessage().equals("Is a directory")) {
+        throw new IOException(path.toString() + " (Is a directory)", e);
```
lberki: Why not just `throw e;`? (granted, this one has a slightly nicer stack trace)

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
```diff
@@ -48,35 +52,65 @@ protected InputStream getInputStream(Path path) throws IOException {
...
+    //
+    // http://mail.openjdk.java.net/pipermail/nio-dev/2014-December/002877.html
+    //
+    // The underlying stream might not be seekable, and wrapping in a BufferedInputStream
+    // causes test failures in the integration suite, so create our own little one-byte
+    // buffer here.
+    byte[] buf = new byte[1];
+    int buflen = 0;
+    try {
+      buflen = stream.read(buf, 0, 1);
```
lberki: In what cases does this matter? From reading the above thread, it seems that this doesn't have anything to do with whether the path contains code points >128, does it?

src/main/java/com/google/devtools/build/lib/unix/UnixFileSystem.java
```diff
@@ -432,6 +436,39 @@ public void deleteTreesBelow(Path dir) throws IOException {
...
+  @Override
+  protected java.nio.file.Path createJavaNioPath(Path path) throws FileNotFoundException {
+    final String pathStr = path.getPathString();
+    if (pathStr.chars().allMatch(c -> c < 128)) {
+      return Paths.get(pathStr);
+    }
+
+    // Paths returned from NativePosixFiles are Strings containing raw bytes from the filesystem,
+    // but Java's IO subsystem expects paths to be encoded in the current locale. We can avoid this
+    // assumption by converting the path to a URI, which permits percent-encoding of any octet.
```
lberki: clever :)

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
jmillikin: The error message doesn't have the filename in it, which makes it (1) unhelpful for users and (2) fail a test covering this exact case.

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
jmillikin: Nothing to do with encoding, this particular pile of nonsense is for the "no error until first read" issue from migrating `java.io` -> `java.nio`. There's a test that calling `getInputStream()` on a directory will fail.

This is reasonable! It should fail!

But in `java.nio` it won't fail, it will return a valid `InputStream` that throws an exception when the very first byte is read.

I have no idea whether this code is good -- the last time I wrote Java professionally was during the days of applets -- but it does allow the test to pass.

lberki: Oh, I see. I missed the part that this is only a problem with `java.nio`.

src/test/shell/bazel/unicode_filenames_test.sh
```diff
@@ -0,0 +1,131 @@
...
+ls_srcs(
+  name = "ls_srcs",
+  srcs = glob(["srcs/**/*"]),
+)
+EOF
+
+  cat >pkg/rules.bzl <<'EOF'
+def _ls_srcs(ctx):
+  out = ctx.actions.declare_file(ctx.attr.name)
+  ctx.actions.run(
```
laszlocsomor: Running a Bash command? Call 0800-RUN-SHELL!
See `ctx.actions.run_shell`.

src/test/shell/bazel/unicode_filenames_test.sh
```diff
@@ -0,0 +1,131 @@
...
+  _ls_srcs,
+  attrs = {"srcs": attr.label_list(allow_files = True)},
+)
+EOF
+}
+
+function test_utf8_source_artifact() {
+  unicode_filenames_test_setup
+
+  python -c 'open(b"pkg/srcs/regular file.txt", "wb").close()'
```
laszlocsomor: Python isn't guaranteed to be on the PATH.
Could you either use a Bash command to do this, or move this to a py_binary (or cc_binary) that the test data-depends on and uses `rlocation` to look up and run here?

src/test/shell/bazel/unicode_filenames_test.sh
```diff
@@ -0,0 +1,131 @@
...
+  python -c 'open(b"pkg/srcs/\xc3\xbcn\xc3\xafc\xc3\xb6d\xc3\xab f\xc3\xafl\xc3\xab.txt", "wb").close()'
+
+  bazel build //pkg:ls_srcs >$TEST_log 2>&1 || fail "Should build"
+  assert_contains "pkg/srcs/regular file.txt" bazel-bin/pkg/ls_srcs
+  assert_contains "pkg/srcs/subdir/file.txt" bazel-bin/pkg/ls_srcs
+  assert_contains "pkg/srcs/ünïcödë fïlë.txt" bazel-bin/pkg/ls_srcs
+}
+
+function test_traditional_encoding_source_artifact() {
+  # Windows and macOS both validate filesystem paths. Linux and the traditional BSDs
```
laszlocsomor: Could you explain what you mean by "validate" here?

NTFS encodes as UTF-16, APFS as normalized UTF-8.

src/main/java/com/google/devtools/build/lib/unix/UnixFileSystem.java
```diff
@@ -432,6 +436,39 @@ public void deleteTreesBelow(Path dir) throws IOException {
...
+    final byte[] pathBytes = path.getPathString().getBytes(StandardCharsets.ISO_8859_1);
+    final StringBuilder builder = new StringBuilder(pathBytes.length * 3);
+    builder.append("file://");
+    for (byte b : pathBytes) {
+      // Keep a few known-safe bytes unescaped for debugging.
+      if (b == 0x2F // '/'
```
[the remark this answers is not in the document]
jmillikin: Done.

src/test/shell/bazel/unicode_filenames_test.sh
jmillikin: Done.

src/test/shell/bazel/unicode_filenames_test.sh
jmillikin: Done (with bash + printf).

src/test/shell/bazel/unicode_filenames_test.sh
jmillikin: Reworded a bit, is it better now? I want to communicate that Windows and macOS paths are always valid Unicode, but Linux/FreeBSD/etc are just bags of bytes.

src/main/java/com/google/devtools/build/lib/vfs/AbstractFileSystem.java
```diff
@@ -48,35 +52,65 @@ protected InputStream getInputStream(Path path) throws IOException {
...
     final String name = path.toString();
     if (profiler.isActive()
         && (profiler.isProfiling(ProfilerTask.VFS_READ)
             || profiler.isProfiling(ProfilerTask.VFS_OPEN))) {
       long startTime = Profiler.nanoTimeMaybe();
       try {
-        // Replace default FileInputStream instance with the custom one that does profiling.
-        return new ProfiledFileInputStream(name);
+        // Replace default InputStream instance with the custom one that does profiling.
+        return new ProfiledInputStream(Files.newInputStream(nioPath), name);
```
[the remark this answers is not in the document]
jmillikin: This PR originally used `new FileInputStream`, but the test failed in CI on CentOS. It looks like that issue is Windows-specific, so I've shuffled the code a bit to ensure only UNIX platforms use `Files.newInputStream`.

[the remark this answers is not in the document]
laszlocsomor: Thanks!

src/test/shell/bazel/unicode_filenames_test.sh
laszlocsomor: Yes, better. Thanks!

laszlocsomor:
Update: I tried importing this change but couldn't, because some Google-internal tests failed. Sorry about that, I'll try to look at those as soon as I can, hopefully tomorrow.

src/main/java/com/google/devtools/build/lib/unix/UnixFileSystem.java
```diff
@@ -432,6 +441,76 @@ public void deleteTreesBelow(Path dir) throws IOException {
...
+    }
+  }
+
+  @Override
+  protected InputStream createFileInputStream(Path path)
+      throws FileNotFoundException, IOException {
+    final java.nio.file.Path nioPath = createJavaNioPath(path);
+    InputStream stream = null;
+    try {
+      stream = Files.newInputStream(nioPath, java.nio.file.StandardOpenOption.READ);
```
laszlocsomor: After much debugging, I found two culprits of the breaking internal tests.

One is this line. I don't know why, but returning `new FileInputStream(nioPath.toFile())` works, but `Files.newInputStream` doesn't. Maybe something somewhere checks for `instanceof FileInputStream` -- which of course they shouldn't, but do so anyway -- and while I'd love to fix that, I'm not sure I can. So, would you be open to using FileInputStream here?

The other error is the logic in lines 490..511. Again, frustratingly, I have no idea why it's wrong. But considering that this logic -- no matter how reasonable -- doesn't serve the goal of this PR (i.e. to support non-ASCII characters), nor is it essential to it, would you be OK with removing it?

jmillikin: > So, would you be open to using FileInputStream here?

Well, using `java.nio.Files` is a hard requirement to make Bazel's CI pass.

> The other error is the logic in lines 490..511. Again, frustratingly, I have no idea why it's wrong. But considering that this logic -- no matter how reasonable -- doesn't serve the goal of this PR (i.e. to support non-ASCII characters), nor is it essential to it, would you be OK with removing it?

I didn't add this logic for fun. The CI tests failed unless I added the seek-once (since tests assert trying to read a directory fails), and using the normal buffering stream caused a different CI failure.

There doesn't seem to be an easy way around using NIO. I've been digging into the JDK source, and Bazel's current behavior of forcing `LANG=en_US.ISO-8859-1` has a side effect on Linux of setting `sun.jnu.encoding` to ASCII-only mode, which prevents `java.io.File` from being able to represent bytes > 0x7F at all.

There are ... a lot of things to unpack here, and it's possible that I'll find something that can be made to work. But it would be much easier if the dependencies on `FileInputStream` could be identified and characterized.

jmillikin:
Implementation rebased to `java.io.File`, by changing some of the locale logic in the Bazel launcher to pass through `LANG`, `LC_ALL` in some circumstances. @laszlocsomor PTAL.

I had an idea this morning about how the JVM encoding interacts with Bazel and it seems to be working (though with limitations):
* On machines with a ISO-8859-1 locale, the JVM can accept arbitrary paths.
* On machines _without_ ISO-8859-1, we can still support the most common case (UTF-8) by re-encoding the file paths.
  * The same code path works for Darwin because the JVM hardcodes the file encoding to UTF-8 regardless of `LC_ALL`.

The only case from `java.nio` that can't be supported now is a Linux/BSD machine, without ISO-8859-1 locales, that has non-UTF-8 file paths. I think this case is uncommon enough to not worry about. If there's anyone out there who needs this, then they'll have to either use `java.nio` or go direct to the POSIX native C API.

laszlocsomor:
Internal tests pass. \o/

Thank you for reworking the PR. Let me get it reviewed again.

> The only case from java.nio that can't be supported now is a Linux/BSD machine, without ISO-8859-1 locales, that has non-UTF-8 file paths. I think this case is uncommon enough to not worry about. If there's anyone out there who needs this, then they'll have to either use java.nio or go direct to the POSIX native C API.

But this is not a regression, since such systems simply still won't support non-ascii characters, right?

jmillikin:
> But this is not a regression, since such systems simply still won't support non-ascii characters, right?

Correct, no change in behavior there. Doesn't work before, still won't work.

laszlocsomor:
Thanks for confirming that.

PR #10113, opened 2019-10-27
Allow user-specific `user.bazelrc` per "Best Practices for Bazel"

jmillikin:
https://docs.bazel.build/versions/master/best-practices.html describes allowing a user-specific `user.bazelrc` in a workspace, but Bazel itself doesn't do this. This makes it a bit annoying to set flags every time.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -42,4 +42,5 @@ build:ubuntu1604_java8 --config=remote_shared
 # Alias
 build:remote --config=ubuntu1604_java8
 
-
+# User-specific .bazelrc
+try-import user.bazelrc
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@
 /.sass-cache
 # Bazelisk version file
 .bazelversion
+# User-specific .bazelrc
+user.bazelrc


philwo:
I guess it's fine (I didn't even know about the try-import feature 😅).

laszlocsomor:
@jmillikin : I'm the internal reviewer for this PR. Could you explain more about the use case? What (kind of) flags do you need to set just for Bazel, that fit neither in your [home RC](https://docs.bazel.build/versions/master/guide.html#where-are-the-bazelrc-files) nor the [workspace RC](https://docs.bazel.build/versions/master/guide.html#where-are-the-bazelrc-files)? Asking because I just want to better understand this use case scenario.

jmillikin:
These flags are for the workspace bazelrc, but if I set them in there then I need to stash them before rebasing, remember to exclude them from commits, and sometimes I can't pull master because there's a change to that file. All of the symptoms you'd expect from carrying a long-lived patch in the local checkout.

Example flags I set on my laptop to make `bazel build //src:bazel` (1) work, (2) be fast:
```diff
diff --git a/.bazelrc b/.bazelrc
index 1ad4a27b8f..66eac79fa5 100644
--- a/.bazelrc
+++ b/.bazelrc
@@ -43,3 +43,10 @@ build:ubuntu1604_java8 --config=remote_shared
 build:remote --config=ubuntu1604_java8
 
 
+build --javabase=@bazel_tools//tools/jdk:remote_jdk
+build --override_repository=rules_python=/Users/john/src/third_party/github.com/bazelbuild/rules_python
+build --extra_toolchains=@rules_python//python:autodetecting_toolchain_nonstrict
+build --incompatible_strict_action_env
+build --action_env=PATH
+build --disk_cache=bazel-caches/DISK_CACHE
+build --repository_cache=bazel-caches/REPO_CACHE
```

laszlocsomor:
Thanks, that explains it.

To me the flags don't look specific to Bazel. Any reason you don't put these in your home RC?

laszlocsomor:
Makes sense, thanks!

PR #10114, opened 2019-10-27
Add grpc URL example to the remote caching doc.

mostynb:


site/docs/remote-caching.md
```diff
@@ -246,6 +246,11 @@ Use the following flag to read from and write to the remote cache:
 build --remote_cache=http://replace-with-your.host:port
 ```
 
+If your cache supports the gRPC based protocol, you can instead use:
+```
+build --remote_cache=grpc://replace-with-your.host:port
+```
```
buchgr: I wonder if should instead just list the supported protocols?

```
Use the following flag to read from and write to the remote cache:

build --remote_cache=(http|https|grpc|grpcs)://replace-with-your.host:port
```

mostynb: How about listing them in a sentence afterwards? (Pushed in an update.)

Issue #10011, opened 2019-10-14
remote cache hits but is not used!


emranbm:
### Description of the problem / feature request:

Bazel remote cache hits, but the build runs completely before the cache hit!

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Reproducer docker image:
```dockerfile
FROM ubuntu:16.04

WORKDIR /opt/

COPY bazel-0.29.1-installer-linux-x86_64.sh .
COPY src/ src/
COPY distdir/ /opt/distdir/

RUN apt-get update
RUN apt-get install unzip
RUN apt-get install -y g++
RUN ./bazel-0.29.1-installer-linux-x86_64.sh --user \
&& rm ./bazel-0.29.1-installer-linux-x86_64.sh \
&& ln -s /root/bin/bazel /usr/bin/bazel \
&& bazel version
RUN touch WORKSPACE
```
`src` directory:
```
src/
├── BUILD    >>> defines `main` target
└── main.cc  >>> simple hello world
```
docker-compose.yml:
```yml
version: "3.7"
services:
  cache-test:
    build: .
    command: "bazel build -s src:main --distdir=/opt/distdir/ --remote_cache=http://cacheserver:9090"
```
Now just run `docker-compose up` and the log seems like:
```
Attaching to cache_cache-test_1
cache-test_1  | Starting local Bazel server and connecting to it...
cache-test_1  | INFO: Invocation ID: 4ea41648-9c33-411e-8567-f45417954581
cache-test_1  | Loading: 
cache-test_1  | Loading: 0 packages loaded
cache-test_1  | Analyzing: target //src:main (1 packages loaded, 0 targets configured)
cache-test_1  | INFO: Analyzed target //src:main (13 packages loaded, 68 targets configured).
cache-test_1  | INFO: Found 1 target...
cache-test_1  | [0 / 6] [Prepa] BazelWorkspaceStatusAction stable-status.txt ... (3 actions, 0 running)
cache-test_1  | SUBCOMMAND: # //src:main [action 'Compiling src/main.cc']
cache-test_1  | (cd /root/.cache/bazel/_bazel_root/68ab1268abb7e5f5688810968df97d6b/execroot/__main__ && \
cache-test_1  |   exec env - \
cache-test_1  |     PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
cache-test_1  |     PWD=/proc/self/cwd \
[...]

PR #10123, opened 2019-10-29
Add a note about stdout/stderr using remote cache

emranbm:
See issue #10011 for more details.

### DIFF
--- a/site/docs/remote-caching.md
+++ b/site/docs/remote-caching.md
@@ -52,7 +52,7 @@ The remote cache stores two types of data:
 * The action cache, which is a map of action hashes to action result metadata.
 * A content-addressable store (CAS) of output files.
 
-**Note:** The remote cache also stores the output of actions (i.e. the `stdout`/`stderr`); so when a cache hit occurs during a build, you may see the exact outputs as without having any cache hit. Keep in mind that just seeing expected std outputs doesn't mean it's really running, when using remote cache.
+Note that the remote cache additionally stores the stdout and stderr for every action. Inspecting the stdout/stderr of Bazel thus is not a good signal for [estimating cache hits](https://docs.bazel.build/versions/master/remote-caching-debug.html).
 
 ### How a build uses remote caching
 
--- a/site/docs/remote-caching.md
+++ b/site/docs/remote-caching.md
@@ -52,6 +52,8 @@ The remote cache stores two types of data:
 * The action cache, which is a map of action hashes to action result metadata.
 * A content-addressable store (CAS) of output files.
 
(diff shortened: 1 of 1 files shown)


emranbm:
@googlebot I signed it!

site/docs/remote-caching.md
```diff
@@ -52,6 +52,8 @@ The remote cache stores two types of data:
 * The action cache, which is a map of action hashes to action result metadata.
 * A content-addressable store (CAS) of output files.
 
+**Note:** The remote cache also stores the output of actions (i.e. the `stdout`/`stderr`); so when a cache hit occurs during a build, you may see the exact outputs as without having any cache hit. Keep in mind that just seeing expected std outputs doesn't mean it's really running, when using remote cache.
```
aiuto: I don't think we can universally say something like this here.  The choice of what each cache server emits to output is up to the server implementation.

My preference is to not make this comment, but I'll pass it around to other's who know more about the various remote cache implementations.

aiuto:
@buchgr/@ola-rozenfeld: Can one of you take a look and see if some sort of comment about this makes sense, or if we should just drop this PR?

site/docs/remote-caching.md
buchgr: I agree with @aiuto. I think what might be useful is to clarify what "action result metadata" in the previous sentence means.

`Action result metadata includes the action's stdout, stderr, paths of output files and exit code.`

WDYT @aiuto @emranbm ?

emranbm: > The choice of what each cache server emits to output is up to the server implementation
As we know, the cache servers are just simple HTTP servers required to just implement `PUT` and `GET` requests. So I think the implementation of the cache server doesn't matter, but the files Bazel actually `PUT`s.

ola-rozenfeld: I like @emranbm 's comment more, because these docs are about how Bazel behaves with remote cache, not just remote caching in general. And Bazel does put the stdout/err in the cache, and display it when fetching from the cache, which may be misleading. Also, for servers that support remote execution in addition to remote caching, it's not up to the server implementation whether to put the stdout/err as part of the action results -- they are part of the results, it's not a choice.

emranbm: Thanks, for your reply.
Any decision made?

ola-rozenfeld: Bowing out of decisions on this one, up to @buchgr .

buchgr: What I think could be improved from the original note is to remove the **bold** formatting as well as a link to the documentation on cache hit/miss. How about this statement?

Note that the remote cache additionally stores the stdout and stderr for every action. Inspecting the stdout/stderr of Bazel thus is not a good signal for [estimating cache hits](https://docs.bazel.build/versions/master/remote-caching-debug.html).

emranbm: @buchgr I just updated the PR with your suggestion. Check that, please.

PR #10137, opened 2019-10-31
Attempt to git fetch only specified commit

### LINKED ISSUE #8969: git_repository should --depth=1 fetch of sha1

Since some years back, it is possible to do `git fetch origin --depth=1 <sha1>` if the server has enabled the configuration `uploadpack.allowReachableSHA1InWant`. When specifying a sha1 in the `git_repository` rule, first try a shallow fetch (`--depth=1`, not `--shallow-since`) of the specific sha1 before doing `git fetch --all`.

By fetching just the specific commit that is wanted, nothing more, the network traffic can be reduced. It is also a clean interface in specifying exactly what is wanted.

Implemented this way, the `shallow_since` argument will not be needed. It could be implemented to recommend that either `shallow_since=<timestamp>` or `shallow_sha1=True` is set, depending on what the server is capable of. This would make it possible to write a full `git_repository` call with only the sha1 and without finding out the commit date.

I could not find any GitHub issue relating to this.

HackAttack:
The Git server config setting uploadpack.allowReachableSHA1InWant and friends allow git fetch to request a particular commit, which is more efficient than fetching all refs and then checking out the commit locally. Attempt to do this, and fall back to fetching everything if it fails.

Resolves #8969

### DIFF
--- a/tools/build_defs/repo/git_worker.bzl
+++ b/tools/build_defs/repo/git_worker.bzl
@@ -21,8 +21,8 @@ _GitRepoInfo = provider(
         "shallow": "Defines the depth of a fetch. Either empty, --depth=1, or --shallow-since=<>",
         "reset_ref": """Reference to use for resetting the git repository.
 Either commit hash, tag or branch.""",
-        "fetch_ref": """Reference for fetching. Can be empty (HEAD), tag or branch.
-Can not be a commit hash, since typically it is forbidden by git servers.""",
+        "fetch_ref": """Reference for fetching.
+Either commit hash, tag or branch.""",
         "remote": "URL of the git repository to fetch from.",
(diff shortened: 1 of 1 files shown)


aiuto:
Is there a corresponding issue?

HackAttack:
Also happy to add tests if someone can provide guidance on where is the appropriate place (git_repository_test.sh? skylark_git_repository_test.sh? GitRepositoryBlackBoxTest.java?).

aiuto:
This seems reasonable and non-breaking, so we won't have to guard it with an incompatible flag. I *think* it addresses @aehlig\'s comment about it being opt-in, and no overhead if they server does not support the feature. At least, that was how I read the code.

@HackAttack Do you think you can add a test to any of the places you listed?  Maybe the black box test.

PR #7381, opened 2019-02-07
Make --force_pic apply to host configurations

keith:
Alternatively it might make sense to have a `--host_force_pic`, or just require users to pass `--host_copt=-fPIC`

PR #10143, opened 2019-10-31
Enable PIC for cc regardless of compilation_mode

keith:
If the feature for supports_pic is enabled, this should be enabled for
all compilation modes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@
 # User-specific .bazelrc
 user.bazelrc
 
+/third_party
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
@@ -539,8 +539,7 @@ public class CppHelper {
       CppConfiguration cppConfiguration,
       FeatureConfiguration featureConfiguration) {
     return cppConfiguration.forcePic()
-        || (toolchain.usePicForDynamicLibraries(cppConfiguration, featureConfiguration)
-            && cppConfiguration.getCompilationMode() != CompilationMode.OPT);
+        || toolchain.usePicForDynamicLibraries(cppConfiguration, featureConfiguration);
   }
 
   /**
--- a/src/test/java/com/google/devtools/build/lib/rules/cpp/CcToolchainTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/cpp/CcToolchainTest.java
@@ -178,7 +178,7 @@ public class CcToolchainTest extends BuildViewTestCase {
             mockToolsConfig, CcToolchainConfig.builder().withFeatures(CppRuleClasses.SUPPORTS_PIC));
     invalidatePackages();
     assertThat(usePicForBinariesWithConfiguration("--cpu=k8")).isTrue();
-    assertThat(usePicForBinariesWithConfiguration("--cpu=k8", "-c", "opt")).isFalse();
+    assertThat(usePicForBinariesWithConfiguration("--cpu=k8", "-c", "opt")).isTrue();
   }
 
   private boolean usePicForBinariesWithConfiguration(String... configuration) throws Exception {
--- a/src/test/java/com/google/devtools/build/lib/rules/cpp/SkylarkCcCommonTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/cpp/SkylarkCcCommonTest.java
@@ -5111,7 +5111,7 @@ public class SkylarkCcCommonTest extends BuildViewTestCase {
   }
 
   @Test
-  public void testCreatePicAndNoPic() throws Exception {
+  public void testCreateOnlyPicWithOpt() throws Exception {
     getAnalysisMock()
         .ccSupport()
         .setupCcToolchainConfig(
@@ -5121,7 +5121,7 @@ public class SkylarkCcCommonTest extends BuildViewTestCase {
     assertThat(getConfiguredTarget("//foo:bin")).isNotNull();
     ConfiguredTarget target = getConfiguredTarget("//foo:skylark_lib");
     assertThat(getFilenamesToBuild(target)).contains("skylark_lib.pic.o");
-    assertThat(getFilenamesToBuild(target)).contains("skylark_lib.o");
+    assertThat(getFilenamesToBuild(target)).doesNotContain("skylark_lib.o");
   }
 
   @Test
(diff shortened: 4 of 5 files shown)


keith:
cc @hlopko for context on this one. We couldn't figure out why this ended up here, but this causes issues with Swift on Linux which requires pic. Using `--force_pic` solves it for some builds, but for builds in the `host` configuration this would solve it (alternatively https://github.com/bazelbuild/bazel/pull/7381 would too)

keith:
This became an issue with rules_swift with this change https://github.com/bazelbuild/rules_swift/commit/ec3ce441e4a4ccac4b2b16c3f3d8bd0e85255cfc

allevato:
> This became an issue with rules_swift with this change [bazelbuild/rules_swift@ec3ce44](https://github.com/bazelbuild/rules_swift/commit/ec3ce441e4a4ccac4b2b16c3f3d8bd0e85255cfc)

Specifically, before this change I was manually looping through the dependencies' `LibrariesToLink` and pulling out the `pic_static_library` if it was present, only falling back to the `static_library` if it didn't have the former. Now, `cc_common.link` apparently uses logic that involves the compilation mode to determine whether PIC artifacts should be used, for reasons that aren't entirely clear to me looking at the code history.

Since Swift requires PIC, if there was a corresponding feature I could enable in my feature configuration, I would just do that, but the closest thing that appears to be available is `force_pic_flags` and that only controls the `-pie` flag _conditionally_ if `--force_pic` is set **and** it doesn't control which LibraryToLink archives get passed to the linker.

allevato:
Digging deeper, I stumbled across a [Stack Overflow post](https://stackoverflow.com/questions/18026333/what-does-compiling-with-pic-dwith-pic-with-pic-actually-do) that included this blurb:

> So why would you want to build something without PIC?
>
> Well, first of all it slows down you program by few percents, because at the start of each function an additional code is run to adjust register, and a precious register is not available for the optimizer (x86 only). Often function can't know if it's called from same library or from another, and thus even internal calls suffer from the penalty. So if you want to optimized for speed - try to compile without PIC.

So the `--compilation_mode=opt` check is there to disable PIC in those builds to squeeze out additional runtime performance improvements, perhaps?

In that case, we may need another way to force it in the API (a feature, an argument to `cc_common.link`, or something else) so that individual rules that absolutely _do_ need it, like Swift, can do so.

allevato:
Another idea that comes to mind after discussing with some colleagues, which would require no changes to Bazel: Have `swift_{binary,test}` transition its `deps` on `//command_line_options:features = ["pic"]`. That would force-enable it for any dependencies, regardless of compilation mode.

I'm not sure whether that's better or worse than updating the linking API to let the caller choose. Transitions have the ability to harm remote caching, but by transitioning on a feature instead of a `--copt`, I wonder if that would end up converging to the same cache key since features can be deduplicated in a way that flags can't.

kastiglione:
> Well, first of all it slows down you program by few percents

a few percent? Citation needed

hlopko:
Hey @keith thanks for the PR. As you can imagine, we cannot merge this as is. Bazel follows [backwards compatibility policy](https://docs.bazel.build/versions/master/backward-compatibility.html) and this is a breaking change, and in addition to that we don't really wan't to change the default of using NOPIC with -c opt for executables for performance reasons. 

The only reason we use PIC objects by default in other compilation modes is to speed up the build - we only have to build objects once for both executables and for shared libraries (= tests), and we accept the performance penalty in outputs as a reasonable tradeoff.

I'm thinking why not just make swift binaries transition on `--force-pic`? That won't affect caching as long as the produced actions are the same, which I believe they would be. 

And I'm not sure if enabling "pic" results in what you expect (@allevato did you try if it works?). If I had to guess I'd say it results in adding -fPIC also to nopic compilation actions. So actions that only differend in -fPIC flag will now be identical, but they will still be registered twice and executed twice (with the second one hitting the cache).

I would accept a PR that adds 'force_pic' attribute to the cc_common.link and adds tests that also verify that the right thing happens when a dependency doesn't provide PIC objects/static library.

allevato:
> I'm thinking why not just make swift binaries transition on `--force-pic`? That won't affect caching as long as the produced actions are the same, which I believe they would be.

This sounds like a good way to go in the short-term if it works, especially if the transition doesn't destroy caching.

> I would accept a PR that adds 'force_pic' attribute to the cc_common.link and adds tests that also verify that the right thing happens when a dependency doesn't provide PIC objects/static library.

But this also sounds like a good idea long-term, since having behavior of an API that can only be toggled via command-line flags (or now, through transitions) feels a bit mysterious.

hlopko:
Oh wait it does affect caching - the output artifact will be stored elsewhere, so it's effectivelly a different action :(

ulfjack:
C++ generates both pic and no-pic actions, right? So it should be possible to choose the pic actions at the binary level?

allevato:
> C++ generates both pic and no-pic actions, right? So it should be possible to choose the pic actions at the binary level?

That should be theoretically possible, but it's not doable today because the APIs don't support it. We'd need a parameter like the one @hlopko described above:

> I would accept a PR that adds 'force_pic' attribute to the cc_common.link and adds tests that also verify that the right thing happens when a dependency doesn't provide PIC objects/static library.

hlopko:
In any case, this PR cannot be merged as is. Keith do you want to repurpose it for the 'force_pic' attribute? Or you (or somebody else) will create a fresh PR and this one can be closed?

keith:
I created this issue to track that https://github.com/bazelbuild/bazel/issues/10303 I might be able to do this after the holidays

oquenchil:
I'm a bit late to this thread. I rather not add force_pic to cc_common.link if there is a different way so that we can keep the API as small as possible.

Can you try by iterating over every library_to_link and for each one build a new library_to_link with just the pic_static_library instead of having the static_library in there too. I believe then that the logic defaults to just taking the pic libraries without any complaints.

Please let me know if that works.

allevato:
If I'm understanding correctly, that would require flattening the `depset` of `LibrariesToLink` at analysis time to iterate over them, which would be an unfortunate performance hit for this requirement. At least the `force_pic` flag would allow the choice to be deferred until the crosstool variables are expanded when the action is executed, right?

I also considered whether we could do it with an aspect attached to each `CcInfo`-providing dependency and remove the `static_library` at that stage, then bubble the modified providers up as a `SwiftCcInfo` or something like that... but by the time the aspect gets the `CcInfo` of the target it's applied to, I think its `LibrariesToLink` would have already been merged with its descendants, so that would be an even worse hit.

ulfjack:
(Let me add a side note that C++ linking is currently quadratic and that's
something we may have to fix.)

oquenchil:
> If I'm understanding correctly, that would require flattening the depset of LibrariesToLink at analysis time to iterate over them, which would be an unfortunate performance hit for this requirement. At least the force_pic flag would allow the choice to be deferred until the crosstool variables are expanded when the action is executed, right?

That is correct that you would have to flatten the nested sets during analysis. But that is currently what we do anyway, so I'm not sure it would be a big performance hit. We just pass artifacts to the linking action, not LibraryToLink objects. I think we also flatten in order to properly build the command line.

As far as I know it has always been like this. I don't know if the new command line arguments builder or nested sets are powerful enough now to allow us to refactor all the linking code to avoid this.

PR #10145, opened 2019-10-31
Fix typo in j2objc URL

rmaz:
Fix the mirror download URL for j2objc.


jin:
This has been fixed in https://github.com/bazelbuild/bazel/commit/8d6eb83c179e6f3cad8c2988a013d409131d5efd

Thank you, anyway!

PR #10155, opened 2019-11-01
Correct references to --compilationMode

kastiglione:
Replaces two references to `--compilationMode` with  `--compilation_mode`.


hlopko:
(Will merge on Monday when I'm at work)

hlopko:
And by Monday I meant Friday, but better late than never :)

PR #10157, opened 2019-11-03
Clarify instructions for running inside the container

regisd:
Clarify the instructions for running Bazel in interactive mode in the container. In particular, `git clone` needs to be done within the container, and for that we need to start a shell in interactive mode first.

### DIFF
--- a/site/docs/bazel-container.md
+++ b/site/docs/bazel-container.md
@@ -64,10 +64,17 @@ docker run \
 ## Build Abseil project from inside the container
 
 The instructions in this section allow you to build using the Bazel container
-with the sources checked out inside the container. A container is started up at
-the beginning of your dev workflow, and all changes to sources must be done
-inside the container. Build results are cached only inside the container and
-will be lost once container is stopped.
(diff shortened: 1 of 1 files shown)


site/docs/bazel-container.md
```diff
@@ -64,10 +64,17 @@ docker run \
...
-inside the container. Build results are cached only inside the container and
-will be lost once container is stopped.
+with the sources inside the container. By starting a container at the beginning
+of your developement workflow and doing changes in the worskpace within the
+container, build results will be cached.
+
+Start a shell in the Bazel container:
+
+```bash
+docker run --interactive --entrypoint=/bin/bash l.gcr.io/google/bazel:0.17.1
```
aiuto: Is the 0.17.1 a Bazel version?  If so, can you bump it to 1.1?  
Too often people blindly copy/paste examples, so when they point to obsolete tools it gets in the way of moving the community along.

regisd: Yes. I was going to propose a follow-up change for this, but I can do it now if you prefer.

Issue #3885, opened 2017-10-10
Support WORKSPACE.bazel as alias for WORKSPACE


davidstanke:
We support naming the BUILD file "BUILD.bazel"... for consistency, let's add support for WORKSPACE.bazel, too.

katre:
It'll be a decent amount of work, there are a lot of bare references to "WORKSPACE" that will need to expand to accept WORKSPACE or WORKSPACE.bazel.

Also, any code that generates an internal WORKSPACE file should update to create WORKSPACE.bazel.

damienmg:
I'd rather not as @katre said, it is a fair amount of work. We can have WORKSPACE.bazel be parsed easily but since we use WORKSPACE to get the root of the workspace directory everywhere, I'd rather just not do it before 1.0

cgrushko:
`BUILD.bazel` became supported because of Windows not distinguishing between `BUILD` and `build`, which was often the name of a directory.
`WORKSPACE` doesn't probably have this problem?

damienmg:
That was because of macOS actually but yes we do not have a collision issue
on WORKSPACE.

cgrushko:
FWIW, supporting WORKSPACE.bazel makes it slightly easier to export code to open-source.

jin:
+1, one less Copybara step.

mikest:
+1, Also useful to disambiguate the mime type in code editors for Syntax highlighting.

alexeagle:
Raising to P2 per discussion with @dslomov and @meteorcloudy - this will also make Bazel more palatable in Angular CLI since users won't wonder "What is this strange WORKSPACE file that's ignored by git". And the `.gitignore` can just be `*.bazel`

PR #10175, opened 2019-11-06
Support WORKSPACE.bazel file

meteorcloudy:
Similar to BUILD.bazel file, Bazel now supports WORKSPACE.bazel as the workspace file. If both WORKSPACE and WORKSPACE.bazel exist, the latter one takes priority.

By supporting WORKSPACE.bazel file, users are more conscious about which files are Bazel related files.

Fixes https://github.com/bazelbuild/bazel/issues/3885


laszlocsomor:
Is there a bug? Could you add it to the PR message?

src/main/java/com/google/devtools/build/lib/rules/repository/LocalRepositoryFunction.java
```diff
@@ -122,7 +122,15 @@ protected static FileValue getWorkspaceFile(RootedPath directory, Environment en
             directory.getRoot(),
             directory
                 .getRootRelativePath()
-                .getRelative(BuildFileName.WORKSPACE.getFilenameFragment()));
+                .getRelative(LabelConstants.WORKSPACE_DOT_BAZEL_FILE_NAME));
+
+    if (!workspaceRootedFile.asPath().exists()) {
+      workspaceRootedFile = RootedPath.toRootedPath(
```
laszlocsomor: Logic is not trivial, so please extract to helper function.

meteorcloudy:
Yes! Added into the description. Looking into the test failure now.

src/main/java/com/google/devtools/build/lib/skyframe/LocalRepositoryLookupFunction.java
```diff
@@ -96,7 +96,15 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
                   .getRootRelativePath()
-                  .getRelative(BuildFileName.WORKSPACE.getFilenameFragment()));
+                  .getRelative(LabelConstants.WORKSPACE_DOT_BAZEL_FILE_NAME));
+      if (!workspaceRootedFile.asPath().exists()) {
+        workspaceRootedFile =
+            RootedPath.toRootedPath(
+                directory.getRoot(),
+                directory
+                    .getRootRelativePath()
+                    .getRelative(LabelConstants.WORKSPACE_FILE_NAME));
```
laszlocsomor: Same logic as in LocalRepositoryFunction. Can you extract to a common helper function?

src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
```diff
@@ -295,17 +295,29 @@ private PackageLookupValue getPackageLookupValue(
...
+            env,
+            packagePathEntries,
+            LabelConstants.EXTERNAL_PACKAGE_IDENTIFIER,
+            BuildFileName.WORKSPACE_DOT_BAZEL);
+    if (resultForWorkspaceDotBazel == null) {
+      return null;
+    }
+    if (resultForWorkspaceDotBazel.packageExists()) {
+      return resultForWorkspaceDotBazel;
+    }
```
laszlocsomor: The same logic is repeated again.
Can you extract to a helper function?

src/test/py/bazel/bazel_workspace_test.py
```diff
@@ -0,0 +1,58 @@
...
+  def testWorkspaceDotBazelFileInMainRepo(self):
+    self.ScratchFile("WORKSPACE.bazel")
+    self.ScratchFile("BUILD", [
+      "py_binary(",
+      "  name = 'bin',",
+      "  srcs = ['bin.py'],",
+      ")",
+    ])
+    self.ScratchFile("bin.py")
+    exit_code, _, stderr = self.RunBazel(['build', '//:bin'])
```
laszlocsomor: Please also add:
- a negative test: delete or rename WORKSPACE.bazel and assert the build no longer works.
- a test for renaming to "WORKSPACE"

opt: No need for the py_binary, just an empty BUILD file. If `bazel build //:BUILD` works, then the workspace is valid. Also in tests below.

src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
meteorcloudy: I cannot refactor this code, because the logic is a bit different from others. It involves checking WORKSPACE file under multiple package paths.

src/test/py/bazel/bazel_workspace_test.py
meteorcloudy: Thanks, I added code for testing the build after removing the `WORKSPACE.bazel` file. One without `WORKSPACE`, the other case has `WORKSPACE`.

meteorcloudy: I'll leave the python targets as it is because I want to test the build actually works and works with external dependencies.

src/main/java/com/google/devtools/build/lib/rules/repository/LocalRepositoryFunction.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/skyframe/LocalRepositoryLookupFunction.java
meteorcloudy: Done,

src/main/java/com/google/devtools/build/lib/skyframe/LocalRepositoryLookupFunction.java
```diff
@@ -91,20 +91,7 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
-                  .getRelative(LabelConstants.WORKSPACE_DOT_BAZEL_FILE_NAME));
-      if (!workspaceRootedFile.asPath().exists()) {
-        workspaceRootedFile =
-            RootedPath.toRootedPath(
-                directory.getRoot(),
-                directory
-                    .getRootRelativePath()
-                    .getRelative(LabelConstants.WORKSPACE_FILE_NAME));
-      }
+      RootedPath workspaceRootedFile = WorkspaceFileHelper.getWorkspaceRootedFile(directory);
```
laszlocsomor: And here, and also in other places where we deal with Skyframe.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryLoaderFunction.java
```diff
@@ -47,14 +46,8 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
     }
-
-    RootedPath workspaceFilePath = RootedPath.toRootedPath(
-        Root.fromPath(repository.getPath()), LabelConstants.WORKSPACE_DOT_BAZEL_FILE_NAME);
-    if (!workspaceFilePath.asPath().exists()) {
-      workspaceFilePath = RootedPath.toRootedPath(
-          Root.fromPath(repository.getPath()), LabelConstants.WORKSPACE_FILE_NAME);
-    }
-
+    RootedPath workspaceFilePath =
```
laszlocsomor: Same here

src/main/java/com/google/devtools/build/lib/rules/repository/LocalRepositoryFunction.java
```diff
@@ -117,21 +117,7 @@ public boolean isLocal(Rule rule) {
...
-
-    if (!workspaceRootedFile.asPath().exists()) {
-      workspaceRootedFile = RootedPath.toRootedPath(
-          directory.getRoot(),
-          directory
-              .getRootRelativePath()
-              .getRelative(LabelConstants.WORKSPACE_FILE_NAME));
-    }
-
+    RootedPath workspaceRootedFile = WorkspaceFileHelper.getWorkspaceRootedFile(directory);
```
laszlocsomor: I just realized you must not use Path.exists() in SkyFunctions. You must request the FileValue for "WORKSPACE.bazel", check its type, and only if it's missing or not a regular file, then request FileValue for "WORKSPACE".

This way the SkyFunction will correctly track changes on the filesystem (e.g. if the file "WORKSPACE.bazel" is replaced by a directory of the same name -- actually, could you add a test for that?)

 think this logic is incorrect: it circumvents Skyframe. You have to ask Skyframe about the file existing, you mustn't use Path.exists().

For taht, request the FileValue for "WORKSPACE.bazel", and check is type: if it's not an existing file, then get the FileValue for "WORKSPACE".
This way Skyframe correctly tracks changes to these files and in the directory.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
```diff
@@ -581,8 +582,7 @@ public static void addExternalFilesDependencies(
       if (isDirectory || repositoryPath.segmentCount() > 1) {
         if (!isDirectory
             && rule.getRuleClass().equals(LocalRepositoryRule.NAME)
-            && (repositoryPath.endsWith(BuildFileName.WORKSPACE.getFilenameFragment())
-                || repositoryPath.endsWith(BuildFileName.WORKSPACE_DOT_BAZEL.getFilenameFragment()))) {
+            && WorkspaceFileHelper.endsWithWorkspaceFileName(repositoryPath)) {
```
laszlocsomor: Same here -- we are in a SkyFunction, you must ask Skyframe about file existence, you shouldn't rely on Path.exists()

src/main/java/com/google/devtools/build/lib/skyframe/LocalRepositoryLookupFunction.java
laszlocsomor: (Context: see my longer comment below.)

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryLoaderFunction.java
laszlocsomor: (Context: see my longer comment below.)

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
laszlocsomor: Ah sorry, this is not about existence check, merely checking the file name. This is fine.

src/main/java/com/google/devtools/build/lib/rules/repository/LocalRepositoryFunction.java
meteorcloudy: Got it! Thanks for reminding me of that! This is extremely important!

src/main/java/com/google/devtools/build/lib/skyframe/LocalRepositoryLookupFunction.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryLoaderFunction.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
meteorcloudy: No problem, thanks!

src/main/java/com/google/devtools/build/lib/rules/repository/LocalRepositoryFunction.java
meteorcloudy: I had to move WorkspaceFileHelper.java to another package to avoid dependencies problem.

src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryFunction.java
```diff
@@ -232,7 +232,7 @@ public void setUseNativePatch(boolean useNativePatch) {
           new IOException(rule + " must create a directory"), Transience.TRANSIENT);
     }
 
-    if (!outputDirectory.getRelative(LabelConstants.WORKSPACE_FILE_NAME).exists()) {
+    if (!WorkspaceFileHelper.doesWorkspaceFileExistUnder(outputDirectory)) {
```
laszlocsomor: I'm concerned about the correctness of calling Path.exists() (without requesting the FileValue) and of creating files (making a cached FileValue stale without telling Skyframe about it).
Let me ask around how to do this.

src/main/java/com/google/devtools/build/lib/rules/repository/WorkspaceFileHelper.java
```diff
@@ -0,0 +1,88 @@
...
+
+    SkyKey workspaceFileKey = FileValue.key(workspaceRootedFile);
+    FileValue value;
+    value = (FileValue) env.getValueOrThrow(workspaceFileKey, IOException.class);
+    if (value == null) {
+      return null;
+    }
+
+    if (value.exists() && !value.isDirectory()) {
+      return workspaceRootedFile;
```
laszlocsomor: Returning here means `WORKSPACE.bazel` exists and it's not a directory.

Returning in line 63 means `WORKSPACE.bazel` is missing (or it's a directory), but says nothing about `WORKSPACE` because it's not checked.

So the caller can't assume the returned path exists and must request it again from Skyframe. (Which Skyframe might even report as an error.)

So the caller won't know if the file exists or not.

Maybe request the second FileValue as well, and return the RootedPath + FileValue pair?

src/main/java/com/google/devtools/build/lib/rules/repository/WorkspaceFileHelper.java
```diff
@@ -0,0 +1,88 @@
...
+                .getRelative(LabelConstants.WORKSPACE_DOT_BAZEL_FILE_NAME));
+
+    SkyKey workspaceFileKey = FileValue.key(workspaceRootedFile);
+    FileValue value;
+    value = (FileValue) env.getValueOrThrow(workspaceFileKey, IOException.class);
+    if (value == null) {
+      return null;
+    }
+
+    if (value.exists() && !value.isDirectory()) {
```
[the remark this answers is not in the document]
meteorcloudy: I see, thanks!

src/main/java/com/google/devtools/build/lib/rules/repository/WorkspaceFileHelper.java
meteorcloudy: Indeed, the caller shouldn't assume the returned path exists. This function just returns the path of the workspace file, then the caller should request the value again. In some case, it would be FileValue, but other cases, it would be `PackageValue` or `WorkspaceFileValue`. So requesting the FileValue of `WORKSPACE` is actually unnecessary.

I'll add a comment to make this clear, WDYT?

laszlocsomor: If this function were inlined, then such callers would just request the PackageValue or WorkspaceFileValue for "WORKSPACE.bazel" and if missing then for "WORKSPACE", without ever requesting a FileValue, right? Because presumable PackageValue and WorkspaceFileValue already depend on the FileValue.

So this logic is actually not generalizable, and documenting and understanding the fine details is more effort than duplicating the code and slightly adjusting it as needed at its call sites. So I recommend inlining this method after all.

meteorcloudy: There are actually three steps in the logic at every places this function gets called.

1. Get FileValue of `WORKSPACE.bazel`

2. Check the FileValue to determine we should use `WORKSPACE` or `WORKSPACE.bazel`

3. Get FileValue/PackageValue/WorkspaceFileValue of the given workspace file.

As you can see, the first two steps are the same, so they could be refactored out. Inline this function would cause this logic to be implemented at 4 different places. So I guess we should keep this?

laszlocsomor: Thinking more about this, I have no strong argument or better idea. Let's keep it like this.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryLoaderFunction.java
```diff
@@ -46,11 +49,21 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
+    RootedPath workspaceFilePath;
+    try {
+      workspaceFilePath = WorkspaceFileHelper
+          .getWorkspaceRootedFile(Root.fromPath(repository.getPath()), env);
+      if (workspaceFilePath == null) {
+        return null;
+      }
+    } catch (IOException e) {
+      throw new RepositoryLoaderFunctionException(
+          new IOException("Could not determine workspace file (WORKSPACE.bazel / WORKSPACE): "
```
laszlocsomor: nit: "/" here might look like a directory separator. How about `"WORKSPACE.bazel" or "WORKPACE"` (with these quotes)?

meteorcloudy: Sounds good!

PR #10180, opened 2019-11-06
Add cc_library option to ensure include_prefix behaves the same regardless of library path

elklein:
Currently include_prefix behaves differently (at least for system includes) when include_prefix matches the path of the library within a larger project.

If you have a library located at path foo/bar (with BUILD file at foo/bar/BUILD), and an include file named Log.h, currently if you specify an include_prefix of "foo/bar", then dependent code cannot do the following and have it work:

#include <foo/bar/Log.h>

Log.h in that case will not be found by the dependent code. If you switch the <> to "", it will be found (because Bazel currently sticks in an "-iquote ." argument to the compile command line). Furthermore, if you change the include_prefix to be "foobar", and change the includes to be:

#include <foobar/Log.h>

it will be also be found.

In cases where it's not possible to change the #include in dependent source code, then the only WAR I've found so far is to put "-isystem ." in the copts of all dependent code, but that's not a really great solution.

This code change allows cc_library rules to opt-in to having include_prefix always treated the same, regardless of library path.

It's possible that this feature: https://docs.google.com/document/d/1PKTDhE9rHzd0WQJY2czjg_vPT28fufGcl2Gain6mUaI/edit?pli=1#heading=h.9grrl7f2c55j
will address the issue as well, but it's not clear what the status of that is, and what the timeline for adoption would be.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppRuleClasses.java
@@ -447,6 +447,19 @@ public class BazelCppRuleClasses {
           prefix is added.
           <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
           .add(attr("include_prefix", STRING))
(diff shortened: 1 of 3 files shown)


elklein:
@googlebot I fixed it

Issue #10184, opened 2019-11-06
Remote caching broken on macOS with Xcode


kastiglione:
### Description of the problem  request:

Remote caching with Bazel 1.1 (https://github.com/bazelbuild/bazel/issues/9982) was broken by https://github.com/bazelbuild/bazel/pull/8495. That changes embeds an absolute path in `cc_wrapper.sh`. For example:

```
# Call the C++ compiler
/private/var/tmp/_bazel_iosci/1ec83befa5e62d8b1339d3693738f87a/external/local_config_cc/wrapped_clang "$@"
```

This has both username and workspace path hash. This wrapper is an input to pretty much every compilation command, causing an extremely degraded cache hit rate.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

We saw this with remote builds. An easier repro might be to use local `--disk_cache` between to separate local checkouts of a project.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 1.1.0

meisterT:
cc @allevato

allevato:
That was an unanticipated effect of that change.

So the issue here is that the C++ actions in crosstool invoke `cc_wrapper.sh` instead of `wrapped_clang` (which the Objective-C actions do), and previously `cc_wrapper.sh` was invoking `clang` directly (via the `cc` substitution). That meant that those actions didn't get any of the Bazel placeholder substitutions that `wrapped_clang` does.

`cc_wrapper.sh` **needs** to call `wrapped_clang` in order to have the right behavior, but since `wrapped_clang` is also built from source when the repository is initialized, it ends up in an output directory that has all those non-hermetic path segments in it.

I'm not sure what the best solution here is. @keith, you were experimenting with something that would replace the `cc` substitution in `osx_cc_configure.bzl` with a shell script fragment that computed `cc_wrapper.sh`'s location and executed `wrapped_clang` relative to it. #10187 notwithstanding, did that work?

@hlopko, does that sound like a reasonable approach?

kastiglione:
@allevato: Keith's PR for calling `wrapped_clang` relatively is here:  https://github.com/bazelbuild/bazel/pull/10185

PR #8495, opened 2019-05-29
Use wrapped_clang for C++ link actions on Apple targets.

allevato:
The `cc_wrapper.sh` script doesn't perform substitution of Bazel placeholder strings like `__BAZEL_XCODE_DEVELOPER_DIR__` in command line arguments, so without this, the link command will fail if it contains those.

This is currently blocking https://github.com/bazelbuild/rules_swift/pull/243, which aims to clean up the logic in `swift_binary` by leveraging more of the `cc_common` APIs.

PR #10185, opened 2019-11-07
Replace macOS CC path with relative path

keith:
After this change https://github.com/bazelbuild/bazel/pull/8495 the path
to CC embedded in this file was absolute, which caused cache misses for
all compile actions. This change depends on this file relative to the
cc_wrapper itself instead.

Fixes https://github.com/bazelbuild/bazel/issues/10184

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -83,11 +83,12 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
         # the C++ actions behave consistently.
         cc = repository_ctx.path("wrapped_clang")
 
+        cc_path = '"$(dirname "$0")"/wrapped_clang'
         repository_ctx.template(
             "cc_wrapper.sh",
             paths["@bazel_tools//tools/cpp:osx_cc_wrapper.sh.tpl"],
             {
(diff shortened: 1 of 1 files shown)


hlopko:
@aehlig @c-parsons I'm wondering if repository_ctx.path("some/relative/path") should return absolute path ever? What do you think about just returning repo-relative path in that case?

keith:
In the meantime can we merge something like this to fix this major regression?

keith:
@hlopko ping since I don't want this to get lost 😄

hlopko:
Sg, let's merge this in and figure the proper fix async :)

philwo:
This PR was cherry-picked into 1.2.0rc1, but unfortunately broke a few downstream projects with these errors:

  = note: external/local_config_cc/cc_wrapper.sh: line 56: dirname: command not found
          external/local_config_cc/cc_wrapper.sh: line 56: /wrapped_clang: No such file or directory
 
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1275#b3a42dd5-7043-4f7c-a24d-78fff6759694

Any idea how to go forward with this?

kastiglione:
Any chance those projects can install whichever package contains `dirname`?

PR #10200, opened 2019-11-08
Refactor remote package to only have one remote cache impl.

buchgr:
Historically the remote package shipped two implementations
for remote caching: GrpcRemoteCache and SimpleBlobStoreActionCache.
Today there is no good reason to keep this structure and the
duplication of code and tests that comes with it. This is the
final refactoring of a long series:

 * 5fc91a7
 * 12ebb84
 * 8ac6bdc
 * 9fb83b4
 * 36611c3
 * a6b5b05
 * d40933d
 * 797f292

This change makes it so that there is one RemoteCache that
has a RemoteCacheClient. The RemoteCacheClient has one
implementation per caching protocol:

 * DiskCacheClient (formerly OnDiskBlobStore)
 * GrpcCacheClient (formerly GrpcRemoteCache)
 * HttpCacheClient (formerly HttpBlobStore)
 * CombinedDiskHttpCacheClient (formerly CombinedDiskHttpBlobStore)

A single RemoteCache type for all protocols will allow
composition of caching protocols. In particular, grpc and disk
caching. Additionally, this change will allow us to fix a longstanding
bug where for the HTTP


buchgr:
FYI @sergiocampama @borkaehw

Issue #10201, opened 2019-11-08
Do not validate extensions in cc_common


maximyurchuk:
### Description of the feature request:
Remove validation of extensions in C++ rules. If someone try to use custom header extension following error will be generated 
`'build/bazel/tests/test_header_without_ext/qqq.foo' has wrong extension. The list of possible extensions for 'private_hdrs' are: .h,.hh,.hpp,.ipp,.hxx,.h++,.inc,.inl,.tlh,.tli,.H`

Headers without extension also will lead to an error. 

There are several issues about this problem, for example 
#302
#5182
#6805
#7242
#8111

I suppose bazel can't know about all extensions which people can use, so I propose remove validation of extension. 

### Feature requests: what underlying problem are you trying to solve with this feature?
In our code we want to use own STL implementation and there are no extensions in header files, e.g. `#include <iostream>`. Bazel doesn't support this. 

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
Sample code can be found [here](https://github.com/maximyurchuk/bazel-issues/tree/master/custom_headers_extension)

### What operating system are you running Bazel on?
Windows 10, 1903

### What's the output of `bazel info release`?
`release 1.0.0`

scentini:
As per the reply on the PR, C++ team would prefer to keep validating extensions.

> Hey thank you for the change but we want to continue validating extensions. If you need more flexibility to customize which extensions are allowed through the toolchain, you can send a PR that allows this customization through the toolchain.

https://github.com/bazelbuild/bazel/pull/10202#issuecomment-559461855

PR #10202, opened 2019-11-08
allow empty and custom extensions in cc_common

maximyurchuk:
#10201 
Remove validation of extension of src and hdr attributes in cc_common.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/rules/cpp/SkylarkCcCommonTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/cpp/SkylarkCcCommonTest.java
@@ -67,6 +67,7 @@ import com.google.devtools.build.lib.view.config.crosstool.CrosstoolConfig.MakeV
 import com.google.devtools.build.lib.view.config.crosstool.CrosstoolConfig.ToolPath;
 import com.google.protobuf.TextFormat;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/devtools/build/lib/rules/cpp/SkylarkCcCommonTest.java
```diff
@@ -5734,23 +5703,16 @@ private static void createFiles(
...
-    getConfiguredTarget("//bar:skylark_lib");
-    assertContainsEvent(
-        "has wrong extension. The list of possible extensions for '" + attrName + "'");
-  }
-
-  private void doTestPossibleExtensionsOfSrcsAndHdrs(String attrName, List<String> extensions)
+  private void doTestPossibleExtensionsOfSrcsAndHdrs(String attrName, List<String> knownExtensions)
       throws Exception {
+    List<String> extensions = new ArrayList<String>(knownExtensions);
+    String unknownExtension = ".unknown_extension";
```
maximyurchuk: Test passed, but bazel not working with unknown header extensions with following stack trace: 

```
Caused by: java.lang.IllegalStateException
        at com.google.common.base.Preconditions.checkState(Preconditions.java:491)
        at com.google.devtools.build.lib.rules.cpp.CcCompilationHelper.addPrivateHeader(CcCompilationHelper.java:424)
        at com.google.devtools.build.lib.rules.cpp.CcCompilationHelper.addPrivateHeaders(CcCompilationHelper.java:408)
        at com.google.devtools.build.lib.rules.cpp.CcModule.compile(CcModule.java:1556)
        at com.google.devtools.build.lib.bazel.rules.cpp.BazelCcModule.compile(BazelCcModule.java:88)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
```

It's because failing precondition in `com.google.devtools.build.lib.rules.cpp.CcCompilationHelper.addPrivateHeader`:
```
    Preconditions.checkState(isHeader || isTextualInclude);
```

Looks like this test is not correct and it test only first element in `extensions` list. If I add `invalidatePackages();` in loop below, test will fail (same as bazel binary).

Could someone say is it correct way of fixing the test? Or maybe I didn't understand something.

oquenchil:
Hey thank  you for the change but we want to continue validating extensions. If you need more flexibility to customize which extensions are allowed through the toolchain, you can send a PR that allows this customization through the toolchain.

PR #10216, opened 2019-11-11
Fix some of the bazel Windows tools code to work with GCC.

pmuetschard:
This fixes bazel build failures on Windows when using MinGW GCC instead of MSVC for building C++.


meisterT:
cc @laszlocsomor

PR #10222, opened 2019-11-12
Hardcode path to dirname on macOS

keith:
It seems that some repos have customized PATH so this isn't available.
This should fix that.

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -83,7 +83,7 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
         # the C++ actions behave consistently.
         cc = repository_ctx.path("wrapped_clang")
 
-        cc_path = '"$(dirname "$0")"/wrapped_clang'
+        cc_path = '"$(/usr/bin/dirname "$0")"/wrapped_clang'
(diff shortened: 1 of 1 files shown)


keith:
https://github.com/bazelbuild/bazel/pull/10185#issuecomment-552968409

keith:
cc @philwo can we run that failed test with this change? I don't really see how this is possible still since if users removed `/usr/bin` on macOS it seems like nothing would work.

jmillikin-stripe:
@keith It looks like the failures are specific to rules_rust, which is running commands under `env -`.

keith:
So they run with no `PATH` defined? I'm surprised that really works at all, but I guess this is a safe change then

philwo:
I'm also surprised that this works at all. 🤷‍♂ Let's merge this.

keith:
@philwo should we cherry pick this to fix the downstream tests in the rc?

philwo:
@keith Yes, I will cherry-pick this into 1.2.0rc2.

kastiglione:
Thanks @philwo

thii:
FWIW, `rules_swift` also runs commands under `env -`.

keith:
Let me know if you see issues with that and this change, it should be ok now

PR #10231, opened 2019-11-13
Fix error in bash syntax in bisect script.

katre:
Also clean first in case the error being bisected is a failure in
incrementality.

### DIFF
--- a/scripts/bazel-bisect.sh
+++ b/scripts/bazel-bisect.sh
@@ -79,7 +79,8 @@ cd "$WORKING_DIR"
 # Call the newly built Bazel. If it fails, save the exit code but don't stop
 # the script.
 result=0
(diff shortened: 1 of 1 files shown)


scripts/bazel-bisect.sh
```diff
@@ -79,7 +79,8 @@ cd "$WORKING_DIR"
 # Call the newly built Bazel. If it fails, save the exit code but don't stop
 # the script.
 result=0
-"$TMP_BIN" $BAZEL_ARGUMENTS || result=$\? && true
+"$TMP_BIN" clean || true
```
jin: Let's key this on a BAZEL_BISECT_WITH_CLEAN=1?

katre: Done, except BAZEL_BISECT_WITH_CLEAN=y because integer booleans are wrong. :)

PR #10247, opened 2019-11-15
Add support for selectors dict concatenation

AlessandroPatti:
Currently it is possible to concatenate multiple select like
`[...] + select({"...": [...]}) + select({"...": [...]})`
However it is not possible to perform a concatenation when the attribute type is a dictionary.
This add support for the syntax:
`{...} + select({"...": {"...": "..."}}) + select({"...": {"...": "..."}})`
All the dictionaries are merged in order, so that each dictionary overrides the keys of the predecessors if the key is already present.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/Type.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/Type.java
@@ -515,6 +515,15 @@ public abstract class Type<T> {
     }
 
     @Override
+    public Map<KeyT, ValueT> concat(Iterable<Map<KeyT, ValueT>> iterable) {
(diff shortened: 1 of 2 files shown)


laurentlb:
Since we don't allow dict concatenation, I don't think we want to add this construct.

gregestren:
@laurentlb - is there a link to that policy for clarification?

aiuto:
I am asserting that this is a good enough link: https://docs.bazel.build/versions/master/skylark/lib/dict.html

TL;DR; Starlark follows python for basic types and python does not allow dict + dict.
Now, we should have a Starlark language definition that is more precise, but that is a different issue.

Issue #10250, opened 2019-11-15
Port Bazel to OpenBSD [duplicate of #9397]


aldersondrive:
Currently, building Bazel on OpenBSD fails. I have been working on porting it. The current status is that I can build a Bazel binary that appears to work.

Here's an issue to track related PRs. Feel free to assign it to me if you like.

aldersondrive:
For anyone following along:

* Other relevant commits include https://github.com/bazelbuild/bazel/commit/ba6be8ac7801f4912a4d2441d4896ffa01bfc94a and https://github.com/bazelbuild/bazel/commit/ebf6a46e25d4be3bf0bbd77cb21dbff0276d5d04.

* A bootstrap build of Bazel on OpenBSD now succeeds and yields a `bazel` binary that works. One caveat is that the bootstrap build requires `python` to be on the `PATH`; it can just be a symlink to `/usr/local/bin/python3`.

aldersondrive:
FYI, the build is now [packaged as an OpenBSD port](https://marc.info/?l=openbsd-ports&m=159163098121456&w=2) (although currently the port is not merged into OpenBSD's ports tree).

aldersondrive:
FYI, the port is now updated to [Bazel 6.0.0-pre.20220922.1](https://github.com/aldersondrive/bazel_openbsd_port).

aldersondrive:
The port is now updated to Bazel 6.0.0, and available at https://github.com/aldersondrive/bazel_openbsd_port.

PR #10251, opened 2019-11-15
Update WORKSPACE to use new version of bazelbuild/platforms repository.

aldersondrive:
The motivation for this change is to pick up commit https://github.com/bazelbuild/platforms/commit/46993efdd33b73649796c5fc5c9efb193ae19d51, for the OpenBSD port in https://github.com/bazelbuild/bazel/issues/10250.


ulfjack:
It looks right to me. The file is there and the checksum also matches.

PR #10271, opened 2019-11-19
docs: fix ios-app tutorial

SCENEE:
The patch allows `bazel build //ios-app:ios-app` work well following "Create a BUILD file" instructions in [Introduction to Bazel: Building an iOS App](https://docs.bazel.build/versions/master/tutorial/ios-app.html).

### DIFF
--- a/site/docs/tutorial/ios-app.md
+++ b/site/docs/tutorial/ios-app.md
@@ -104,12 +104,24 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 git_repository(
     name = "build_bazel_rules_apple",
     remote = "https://github.com/bazelbuild/rules_apple.git",
-    tag = "0.4.0",
+    tag = "0.19.0",
 )
+git_repository(
+    name = "build_bazel_rules_swift",
+    remote = "https://github.com/bazelbuild/rules_swift.git",
+	tag = "0.13.0",
+)
+
+git_repository(
+    name = "build_bazel_apple_support",
+    remote = "https://github.com/bazelbuild/apple_support.git",
+	tag = "0.7.2",
+)
+
 git_repository(
     name = "bazel_skylib",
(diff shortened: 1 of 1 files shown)


aiuto:
I don't know enough about iOS to review this.

site/docs/tutorial/ios-app.md
```diff
@@ -174,7 +186,6 @@ objc_library(
          "UrlGet/main.m",
     ],
     hdrs = glob(["UrlGet/*.h"]),
-    xibs = ["UrlGet/UrlGetViewController.xib"],
```
DavidGoldman: I believe this should instead be in the `data` attribute, see [here](https://github.com/bazelbuild/rules_apple/blob/master/doc/resources.md#migration).

site/docs/tutorial/ios-app.md
```diff
@@ -198,6 +209,7 @@ ios_application(
     ],
     minimum_os_version = "9.0",
     infoplists = [":UrlGet/UrlGet-Info.plist"],
+    launch_storyboard = "UrlGet/UrlGetViewController.xib",
```
DavidGoldman: Instead I think you'll want to add a dummy storyboard instead of reusing the view controller's XIB

site/docs/tutorial/ios-app.md
SCENEE: Thank you for your comment. I agree with you! I will update here following your advice.

site/docs/tutorial/ios-app.md
SCENEE: I just referred the [BUILD](https://github.com/bazelbuild/examples/blob/master/tutorial/ios-app/BUILD) file, but I didn't intend to add it so I'm going to remove `launch_storyboard` attribute. This is because the addition needs to update `source-only` branch of [bazelbuild/examples](https://github.com/bazelbuild/examples/). Thank you.

PR #10272, opened 2019-11-19
Support .bazelversion and $USE_BAZEL_VERSION in Bazel's wrapper script.

philwo:
This will allow users to install multiple versions of Bazel and have the
wrapper automatically determine and run the correct version for each
project.

### DIFF
--- a/scripts/packages/bazel.sh
+++ b/scripts/packages/bazel.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2015 The Bazel Authors. All rights reserved.
+# Copyright 2019 The Bazel Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


scripts/packages/bazel.sh
```diff
@@ -61,29 +88,77 @@ function get_realpath() {
...
+  echo ""
+  echo "You can specify which Bazel version to use using these methods:"
+  echo "1. Set \$USE_BAZEL_VERSION to a version number (e.g. export USE_BAZEL_VERSION=0.29.1)."
+  echo "2. Add a .bazelversion file that contains a version number next to your WORKSPACE file."
+  echo "3. Otherwise, the latest Bazel version installed as ${wrapper_dir}/bazel-* will be used."
+  echo ""
+  echo "For case 1 and 2, a binary called \"bazel-\$version\" (e.g. \"bazel-0.29.1\") must be present in ${wrapper_dir}."
+  echo ""
+  echo "You might be able to install specific Bazel versions via apt-get:"
+  echo "  $ apt-get update && apt-get install bazel-0.29.1") >&2
```
meteorcloudy: This should be: `apt-get install bazel=0.29.1` when the multiple version support is available.

philwo: Unfortunately apt-get doesn't allow users to install multiple versions of a package like this (this would instead replace the current Bazel version with 0.29.1 on the system).

I think we have to create separate side-by-side installable Bazel packages for this.

kylecordes:
It could be confusing to have both this mechanism and `bazelisk` meeting the same need.

philwo:
@kylecordes I made this alternative mechanism for people who don't want to or can't use Bazelisk, for example because their company or environment prohibits downloading binaries from the internet. These users usually prefer to install Bazel via `apt-get` and then can use this mechanism.

This wrapper script is only included in Bazel's binary installers and Debian packages, so shouldn't affect any other users.

@meteorcloudy @kylecordes WDYT, I could also add a hint to the error message like this: "Hint: Do you want to get automatic upgrades and downloads of missing versions? Check out Bazelisk: [...]".

scripts/packages/bazel.sh
meteorcloudy: https://github.com/bazelbuild/bazel/pull/10278 will make both possible!

meteorcloudy: What do you think we replace "0.29.1" with the actual bazel version requested `$bazel_version`?

meteorcloudy: Oh, sorry, never mind. I saw more specific instruction for this later.

meteorcloudy:
If I have `.bazelversion` file that contains "1.1.0" and current latest release version is 1.1.0, does this change require me to do both `apt-get install bazel` and `apt-get install bazel-1.1.0`? Even though the current latest version of Bazel is 1.1.0 and already installed by `apt-get install bazel`.

meteorcloudy:
Well, maybe it's not that bad users has to do "apt-get install bazel-<version>", as we already print the command to execution for users, all they have to do is copy & paste & enter. I think this looks good once we have "bazel-<version>" packages available also for internal users.

PR #10277, opened 2019-11-20
Adds explicit support for comments in bazelignore file

TheMarex:
This adds explicit support for comments in the `.bazelignore` file, by removing all lines starting with `#`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/BlacklistedPackagePrefixesFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/BlacklistedPackagePrefixesFunction.java
@@ -127,7 +127,7 @@ public class BlacklistedPackagePrefixesFunction implements SkyFunction {
 
     @Override
     public boolean processLine(String line) throws IOException {
-      if (!line.isEmpty()) {
+      if (!line.isEmpty() && !line.startsWith("#")) {
         fragments.add(PathFragment.create(line));
       }
       return true;
--- a/src/test/shell/bazel/bazelignore_test.sh
+++ b/src/test/shell/bazel/bazelignore_test.sh
@@ -25,6 +25,27 @@ source "${CURRENT_DIR}/../integration_test_setup.sh" \
 
 set -e
 
+test_ignores_comment_lines() {
+    rm -rf work && mkdir work && cd work
+    create_workspace_with_default_repos WORKSPACE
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/BlacklistedPackagePrefixesFunction.java
```diff
@@ -97,7 +97,7 @@ public SkyValue compute(SkyKey key, Environment env)
 
     @Override
     public boolean processLine(String line) throws IOException {
-      if (!line.isEmpty()) {
+      if (!line.isEmpty() && !line.startsWith("#")) {
```
laszlocsomor: Is it possible that a directory name starts with `#`?
`/` is surely disallowed in directory names, and absolute paths are disallowed (I hope) in the `.bazelignore`, so `//` as the comment prefix seems safer than `#`. WDYT?

laszlocsomor: Right, I just noticed you test this in the shell test, but IMO the behavior is incorrect now.
So my question is still open :)

TheMarex: Hm, I think my main motivation for using was to make this work the same way `gitignore` works. `#` seems to be the pseudo-standard for ignore files as far as I can tell. So I think user-expectation of a line `#some/path` is that it will be ignored.

Directory names starting with `#` seems very unlikely to me. But I get your point that this is technically breaking. I bet there are more `bazelignore` files where comments like `#some helpful text` "works" merely based on the fact that there is no directory with that name anyway.

laszlocsomor: That's a good point about convention. You know what, that's actually a pretty compelling argument. Let's stay with `#`.

laszlocsomor:
One more thing, could you please add a release note to the PR message like so:

```
RELNOTES[NEW]: bazelignore files now support line comments, with # as the comment character
```

or something like this.

TheMarex:
@laszlocsomor Thanks! Sorry for dropping the ball here, forgot about this PR. :santa: Added the release notes.

PR #10278, opened 2019-11-20
Update release process of apt repo

meteorcloudy:
With the new release process, Bazel APT repo will now support
installing Bazel at a specific version.
Users can run "apt-get install bazel=1.0.0", can call "bazel" from command line.

We also publish side-by-side packages named "bazel-<version>", which
made it possible to install multiple Bazel versions at the same time.
Users can run "apt-get install bazel-1.0.0", then call "bazel-1.0.0" from command line.
This will support https://github.com/bazelbuild/bazel/pull/10272

### DIFF
--- a/scripts/ci/build.sh
+++ b/scripts/ci/build.sh
@@ -208,6 +208,86 @@ function ensure_gpg_secret_key_imported() {
   fi
 }
 
+# Generate new content of Release file
+function print_new_release_content() {
+  local distribution="$1"
+  # Print the headers of the original Release file
+  head -n 7 dists/${1}/Release 2>>1
+  metadata_files=("jdk1.8/binary-amd64/Packages" "jdk1.8/binary-amd64/Packages.gz" "jdk1.8/binary-amd64/Release" "jdk1.8/source/Sources.gz" "jdk1.8/source/Release")
+  # Re-generate hashes for all metadata fiels
+  echo MD5Sum:
+   for file in ${metadata_files[*]}; do
+    path="dists/${distribution}/$file"
+    echo "" "$(md5sum ${path} | cut -d " " -f1)" "$(ls -l ${path} | cut -d " " -f5)" "$file"
+   done
+  echo SHA1:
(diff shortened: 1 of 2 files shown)


scripts/ci/build.sh
```diff
@@ -208,6 +208,87 @@ function ensure_gpg_secret_key_imported() {
...
+
+  # Get bazel version
+  bazel_version=$(grep "Version:" deb-output/DEBIAN/control | cut -d " " -f2)
+  bazel_version=${bazel_version/\~/}
+
+  # Change package name to bazel-{bazel_version}
+  versioned_deb_pkg_name="bazel-${bazel_version}-versioned-package-amd64.deb"
+  sed -i "s/Package:\ bazel/Package:\ bazel-${bazel_version}/g" "deb-output/DEBIAN/control"
+
+  # Delete conffiles, bash completion files and bash wrapper to avoid conflict when mulitple versions are installed.
```
philwo: We remove so many things from the package, would it be maybe better to just start with an empty directory and copy over the files we need?

I worry that one day we might accidentally ship some other files here that we don't intend to. For example, I recently noticed that bash-completions are no longer supposed to be in /etc but in /usr/share. If someone would fix this but forget to update this script, we would keep the files in the versioned package and might cause a conflict when trying to install it.

scripts/ci/build.sh
```diff
@@ -208,6 +208,87 @@ function ensure_gpg_secret_key_imported() {
   fi
 }
 
+# Generate new content of Release file
+function print_new_release_content() {
+  local distribution="$1"
+  # Print the headers of the original Release file
+  head -n 7 dists/${1}/Release 2>>1
```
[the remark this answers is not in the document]
meteorcloudy: That's way better! Thanks!

scripts/ci/build.sh
meteorcloudy: Makes perfect sense, done!

meteorcloudy:
This is ready for review again ;)

Issue #10187, opened 2019-11-07
Absolute paths in `local_config_cc/builtin_include_directory_paths` break caching


keith:
### Description of the problem / feature request:

On macOS when the `builtin_include_directory_paths` file is generated it includes an absolute path:

```
This file is generated by cc_configure and contains builtin include directories
that /private/var/tmp/_bazel_ksmiley/36ab5c790411e201d438edcb7aed1d38/external/local_config_cc/wrapped_clang reported. This file is a dependency of every compilation action and
changes to it will be reflected in the action cache key. When some of these
paths change, Bazel will make sure to rerun the action, even though none of
declared action inputs or the action commandline changes.

/Applications/Xcode-11.2.0-beta2.app/Contents/Developer
/Applications/Xcode-11.2.1-GM1.app/Contents/Developer
/Applications/Xcode-10.3.0.app/Contents/Developer
/Applications/Xcode-10.2.1.app/Contents/Developer
/Applications/Xcode-11.2.0.app/Contents/Developer
/Applications/Xcode-11.2.0-beta1.app/Contents/Developer
/Applications/Xcode-11.1.0-GM1.app/Contents/Developer
/Applications/Xcode-11.0.0-GM2.app/Contents/Developer
/Applications/
```

This is caused by writing `cc` to the file:

https://github.com/bazelbuild/bazel/blob/bd0a40254697ec7e0cd9a522cf5ba282ea76389c/tools/cpp/lib_cc_configure.bzl#L242

Since macOS CC is defined by this absolute path:

https://github.com/bazelbuild/bazel/blob/bd0a40254697ec7e0cd9a522cf5ba282ea76389c/tools/cpp/osx_cc_configure.bzl#L84

In the meantime developers can disable this behavior entirely by setting `BAZEL_IGNORE_SYSTEM_HEADERS_VERSIONS=1` in their environment.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 1.1.0

###  Have you found anything relevant by searching the web?

This is very related to https://github.com/bazelbuild/bazel/issues/10184 which can hopefully be fixed by https://github.com/bazelbuild/bazel/pull/10185 but I wanted to track it separately because that PR will not fix this issue.

slam:
This bug is as severe as #10184. They both break remote cache on MacOS. Is it possible to also fix this bug in the upcoming 1.2.0 release?

keith:
[...]

PR #10281, opened 2019-11-20
Remove builtin directory file on macOS

keith:
On macOS the paths produced in this file are specific to Xcode paths.
Since Xcode versions are already managed by `local_config_xcode`, having
this duplicates that validation, and also breaks caching.

This fixes https://github.com/bazelbuild/bazel/issues/10187

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -19,7 +19,6 @@ load(
     "@bazel_tools//tools/cpp:lib_cc_configure.bzl",
     "escape_string",
     "resolve_labels",
-    "write_builtin_include_directory_paths",
 )
 load(
     "@bazel_tools//tools/cpp:unix_cc_configure.bzl",
@@ -140,7 +139,6 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
                  error_msg)
 
         escaped_include_paths = _get_escaped_xcode_cxx_inc_directories(repository_ctx, cc, xcode_toolchains)
-        write_builtin_include_directory_paths(repository_ctx, cc, escaped_include_paths)
         escaped_cxx_include_directories = []
         for path in escaped_include_paths:
             escaped_cxx_include_directories.append(("        \"%s\"," % path))
--- a/tools/osx/crosstool/BUILD.tpl
+++ b/tools/osx/crosstool/BUILD.tpl
@@ -47,7 +47,6 @@ cc_toolchain_suite(
     filegroup(
         name = "osx_tools_" + arch,
         srcs = [
-            ":builtin_include_directory_paths",
             ":cc_wrapper",
             ":libtool",
             ":make_hashed_objlist.py",


keith:
cc @hlopko

aragos:
Not sure if this is a safe change as it changes the behavior when `BAZEL_IGNORE_SYSTEM_HEADERS_VERSIONS` is present. Adding @trybka for additional review.

hlopko:
Alternatively, we can not put CC into the system headers file (on all platforms). That would fix all the issues as well, right?

keith:
We could do that, but I thought your original intent was to capture the system header files especially for linux distros? If we didn't do that what would this file contain?

trybka:
I don't think I'm an appropriate reviewer here. The cc_configure stuff is outside of my wheelhouse.

keith:
@hlopko @aragos anything I can do to move this forward?

aiuto:
@jmmv Is this something your team can look into now?

aiuto:
Trying new reviewers

keith:
@jmmv think you could take a look?

tools/cpp/osx_cc_configure.bzl
```diff
@@ -140,7 +139,6 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
                  error_msg)
 
         escaped_include_paths = _get_escaped_xcode_cxx_inc_directories(repository_ctx, cc, xcode_toolchains)
-        write_builtin_include_directory_paths(repository_ctx, cc, escaped_include_paths)
```
tetromino, line 143: I'm not sure if it's safe to remove the include directories altogether (see https://github.com/bazelbuild/bazel/issues/10187#issuecomment-556205226).

Would it be sufficient to just replace the `cc` path here with a fixed string? e.g.

```
write_builtin_include_directory_paths(repository_ctx, "wrapped_clang", escaped_include_paths)
```

keith: What from that comment specifically do you mean makes it unsafe?

I don't think replacing `cc` with `wrapped_clang` would affect anything here since the core piece I want to remove is the explicit path to Xcode.app in the generated file:

```
 /usr/local/include
 /Applications/Xcode-11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1
 /Applications/Xcode-11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/11.0.0/include
 /Applications/Xcode-11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /Applications/Xcode-11.3.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode-11.3.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
```

Here I would expect the output to be the same whether you can `clang` or `wrapped_clang` on macOS

meisterT:
ping @tetromino

keith:
bump

trybka:
Does this have any effect on sandboxing? (This is not my area of expertise, but that's the primary concern I would have).

keith:
It shouldn't, we flip the current bool for this off currently and building with sandboxing works fine.

sventiffe:
@tetromino friendly ping

tetromino (approved):

tetromino:
I apologize for the wait. Verified that this works correctly on systems with multiple Xcode versions installed.

sventiffe:
So are we good to merge? Will you initiate that?

tetromino:
@aiuto was initiating the merge.

PR #10284, opened 2019-11-21
Only request grpc write when not complete

werkt:
If a queryWriteStatus yields a committedSize which leaves no content
remaining to be uploaded, immediately succeed a blob upload. This can
easily occur if a competing blob write completes asynchronously between
abnormal write termination and a query.


buchgr:
@werkt please add a test for this :stuck_out_tongue:

Issue #10294, opened 2019-11-22
HttpDownloader: java.io.IOException: thread interrupted


beasleyr-vmw:
### Description of the problem / feature request:

Our Bazel builds intermittently fail during analysis when transferring Maven artifacts from an internal Artifactory server. The console only displays a vague error like the following:

```
ERROR: An error occurred during the fetch of repository 'org_apache_neethi_neethi_jar_sources_2_0_4':
   java.io.IOException: thread interrupted
ERROR: .../92b8e37af6d4337248d7dd4b9c25861a/external/vlsi_maven_deps/BUILD:224:1: @vlsi_maven_deps//:org_apache_neethi_neethi_jar_sources_2_0_4_extension depends on @org_apache_neethi_neethi_jar_sources_2_0_4//file:file in repository @org_apache_neethi_neethi_jar_sources_2_0_4 which failed to fetch. no such package '@org_apache_neethi_neethi_jar_sources_2_0_4//file': java.io.IOException: thread interrupted
ERROR: Analysis of target '//vim/vlsi/dist:dist' failed; build aborted: Analysis failed
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Difficult to say. Run lots of builds which download small artifacts from a flaky server?

### What operating system are you running Bazel on?

CentOS 7.2 and macOS High Sierra and Mojave.

### What's the output of `bazel info release`?

release 1.1.0+vmware

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

na

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

na

###  Have you found anything relevant by searching the web?

- https://github.com/bazelbuild/bazel/issues/3851
- https://github.com/bazelbuild/bazel/issues/8974#issuecomment-516590432

### Any other information, logs, or outputs that you want to share?

All failures encountered having this signature involved files < 32KB in size, possibly implicating the 32KB precheck logic in HttpStream.java.

----

I applied a small patch to try to get more context, and the next failure produced the following:

```
ERROR: An error occurred during the fetch of repository 'com_google_protobuf_protobuf_java_util_jar_sources_3_4_0':
   java.io.IOException: thread interrupted
Printing stack trace: com.google.devtools.build.lib.syntax.EvalExceptionWithStackTrace: java.io.IOException: thread interrupted
[...]

PR #10295, opened 2019-11-22
http: handle SocketTimeoutException directly

beasleyr-vmw:
SocketTimeoutException derives from InterruptedIOException, but its occurrence
in HttpConnectorMultiplexer (HCM) is truly distinct from a HCM-prompted job
control interruption, and it should be resported as such.  Otherwise users
may only see "java.net.IOException: thread interrupted" when a download fails.

NB: I'm not a Java developer.  (Should `catch InterruptedIOException`
include an `if (Thread.interrupted())` check instead?)

Relates to #10294.


laurentlb:
@aehlig Can you review this PR?

aiuto:
ping

aiuto:
another ping, but with new reviewers

ahumesky:
thanks!

PR #10307, opened 2019-11-26
Add validation that all targets used as platforms actually advertise

### LINKED ISSUE #10282: Cycle error when --platforms is not a platform

If you try to use a number of things as a platform, you get a massive garbage error from the cycle detector.

Examples include:
* `sh_binary`
* `rust_binary`
* `config_setting`

This can be reproduced in the repo at https://github.com/m3rcuriel/bazel/tree/cycle_repro by running `bazel build //... --platforms=//:garbage` or even `bazel build //... --platforms=//:test`.

5.3.7.b-3-hardened #1 SMP PREEMPT @1572019254 x86_64 GNU/Linux

`release 0.29.1- (@non-git)`

This version comes from the AUR. Also fails with bazel HEAD (11/20/19) built with the bazel from AUR.

```
git@github.com:bazelbuild/bazel
2b7e21db100bdc0bd319f37106ed3d9e7c8dbe74
2b7e21db100bdc0bd319f37106ed3d9e7c8dbe74
```

Inspired by https://github.com/bazelbuild/rules_rust/issues/267.

katre:
they provide PlatformInfo.

Fixes #10282.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
@@ -16,8 +16,8 @@ package com.google.devtools.build.lib.skyframe;
 
 import static com.google.common.base.Predicates.equalTo;
 import static com.google.common.base.Predicates.not;
-import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static com.google.devtools.build.lib.skyframe.RegisteredExecutionPlatformsFunction.hasPlatformInfo;
 import static java.util.stream.Collectors.joining;
 
 import com.google.common.collect.ImmutableList;
@@ -42,7 +42,6 @@ import com.google.devtools.build.skyframe.ValueOrException;
 import com.google.devtools.build.skyframe.ValueOrException3;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.function.Function;
 import javax.annotation.Nullable;
 
 /** Helper class that looks up {@link PlatformInfo} data. */
@@ -90,44 +89,44 @@ public class PlatformLookupUtil {
       ImmutableList<ConfiguredTargetKey> platformKeys, Environment env)
       throws InterruptedException, InvalidPlatformException {
     // Load the packages. This should already be in Skyframe and thus not require a restart.
-    Map<ConfiguredTargetKey, PackageIdentifier> targetsToPackageIdentifiers =
-        platformKeys.stream()
-            .distinct()
-            .collect(
-                toImmutableMap(Function.identity(), ctk -> ctk.getLabel().getPackageIdentifier()));
     ImmutableSet<PackageValue.Key> packageKeys =
-        targetsToPackageIdentifiers.values().stream()
+        platformKeys.stream()
+            .map(ConfiguredTargetKey::getLabel)
+            .map(Label::getPackageIdentifier)
             .distinct()
             .map(PackageValue::key)
             .collect(toImmutableSet());
 
-    Map<PackageIdentifier, Package> packages = new HashMap<>();
     Map<SkyKey, ValueOrException<NoSuchPackageException>> values =
         env.getValuesOrThrow(packageKeys, NoSuchPackageException.class);
-    if (env.valuesMissing()) {
-      return;
-    }
+    boolean valuesMissing = env.valuesMissing();
+    Map<PackageIdentifier, Package> packages = valuesMissing ? null : new HashMap<>();
     for (Map.Entry<SkyKey, ValueOrException<NoSuchPackageException>> value : values.entrySet()) {
       try {
         PackageValue packageValue = (PackageValue) value.getValue().get();
-        packages.put(packageValue.getPackage().getPackageIdentifier(), packageValue.getPackage());
+        if (!valuesMissing && packageValue != null) {
+          packages.put(packageValue.getPackage().getPackageIdentifier(), packageValue.getPackage());
+        }
       } catch (NoSuchPackageException e) {
+        throw new InvalidPlatformException(e);
       }
+    }
+    if (valuesMissing) {
+      return;
     }
 
     // Now check each platform.
-    HasPlatformInfo hasPlatformInfo = HasPlatformInfo.create();
     for (ConfiguredTargetKey platformKey : platformKeys) {
       try {
         Label platformLabel = platformKey.getLabel();
         Target target =
             packages.get(platformLabel.getPackageIdentifier()).getTarget(platformLabel.getName());
-        if (!hasPlatformInfo.hasPlatformInfo(target)) {
+        if (!hasPlatformInfo(target)) {
           // validation failure
           throw new InvalidPlatformException(platformLabel);
         }
       } catch (NoSuchTargetException e) {
-
+        throw new InvalidPlatformException(e);
       }
     }
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
```diff
@@ -66,6 +85,53 @@
...
+    Map<SkyKey, ValueOrException<NoSuchPackageException>> values =
+        env.getValuesOrThrow(packageKeys, NoSuchPackageException.class);
+    if (env.valuesMissing()) {
+      return;
+    }
+    for (Map.Entry<SkyKey, ValueOrException<NoSuchPackageException>> value : values.entrySet()) {
+      try {
+        PackageValue packageValue = (PackageValue) value.getValue().get();
+        packages.put(packageValue.getPackage().getPackageIdentifier(), packageValue.getPackage());
+      } catch (NoSuchPackageException e) {
```
aragos: It might be useful to add a comment here why this exception cannot happen/can safely be ignored.

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
```diff
@@ -66,6 +85,53 @@
     return platforms;
   }
 
+  /** Validate that all keys are for actual platform targets. */
+  private static void validatePlatformKeys(
+      ImmutableList<ConfiguredTargetKey> platformKeys, Environment env)
+      throws InterruptedException, InvalidPlatformException {
+    // Load the packages. This should already be in Skyframe and thus not require a restart.
+    Map<ConfiguredTargetKey, PackageIdentifier> targetsToPackageIdentifiers =
```
aragos: This map only seems to be used to create another collection out of its values. Consider merging the two stream operations and/or making this a set of package identifiers.

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
```diff
@@ -66,6 +85,53 @@
...
+    for (ConfiguredTargetKey platformKey : platformKeys) {
+      try {
+        Label platformLabel = platformKey.getLabel();
+        Target target =
+            packages.get(platformLabel.getPackageIdentifier()).getTarget(platformLabel.getName());
+        if (!hasPlatformInfo.hasPlatformInfo(target)) {
+          // validation failure
+          throw new InvalidPlatformException(platformLabel);
+        }
+      } catch (NoSuchTargetException e) {
```
aragos: See comment above.

src/main/java/com/google/devtools/build/lib/skyframe/RegisteredExecutionPlatformsFunction.java
```diff
@@ -234,7 +234,10 @@ public boolean shouldRetain(Target target, boolean explicit) {
       if (explicit) {
         return true;
       }
+      return hasPlatformInfo(target);
+    }
 
+    boolean hasPlatformInfo(Target target) {
```
aragos: The owning class for this method is very strange - it's an AutoValue without fields? (Documentation would help!)

In any case this method should likely be static now and possibly live elsewhere so it can be used from here and its new usage properly.

src/main/java/com/google/devtools/build/lib/skyframe/RegisteredExecutionPlatformsFunction.java
```diff
@@ -234,7 +234,10 @@ public boolean shouldRetain(Target target, boolean explicit) {
...
       // If the rule requires platforms or toolchain resolution, it can't be used as a platform.
```
aragos: This comment appears misplaced here and indicates that the associated logic maybe doesn't belong in this method?

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
katre: Added error handling.

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
katre: Done.

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
katre: Added error handling.

src/main/java/com/google/devtools/build/lib/skyframe/RegisteredExecutionPlatformsFunction.java
katre: Moved and cleaned up the comment.

src/main/java/com/google/devtools/build/lib/skyframe/RegisteredExecutionPlatformsFunction.java
katre: Cleaned up the comment.

src/main/java/com/google/devtools/build/lib/skyframe/RegisteredExecutionPlatformsFunction.java
katre: The AutoValue is present to give hashCode and equals implementations (needed for AutoCodec to work properly), and a toString. I can remove it and add simple versions of those, which is better?

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
```diff
@@ -90,44 +89,44 @@ private static void validatePlatformKeys(
...
             .map(PackageValue::key)
             .collect(toImmutableSet());
 
-    Map<PackageIdentifier, Package> packages = new HashMap<>();
     Map<SkyKey, ValueOrException<NoSuchPackageException>> values =
         env.getValuesOrThrow(packageKeys, NoSuchPackageException.class);
-    if (env.valuesMissing()) {
-      return;
-    }
+    boolean valuesMissing = env.valuesMissing();
```
aragos: I don't quite understand why you don't return here if values are missing.

src/test/shell/bazel/toolchain_test.sh
```diff
@@ -1375,6 +1375,36 @@ EOF
...
+)
+sh_binary(
+  name = "target",
+  srcs = ["target.sh"],
+)
+EOF
+
+  bazel build --platforms=//:hello //:target &> $TEST_log && fail "Build succeeded unexpectedly"
+  expect_log "While resolving toolchains for target //:target: Target //:hello was referenced as a platform, but does not provide PlatformInfo"
+}
```
aragos: Can you also add a test for a missing platform target?

src/main/java/com/google/devtools/build/lib/skyframe/PlatformLookupUtil.java
katre: This is a source of many Skyframe errors: you have to propagate errors *before* you restart due to missing dependencies, or else the errors will be swallowed and either a) not ever be raised, or b) be raised for the parent of the current skyfunction call. Neither of these is good, and so checking for errors first is essential.

In the simple one-key `env.getValueOrThrow` case this happens automatically, because the `getValueOrThrow` method throws the error immediately and your exception handler catches it, but when requesting multiple keys at once via `env.getValuesOrThrow` it is surprisingly difficult to get it right.

This is why all of my skyfunctions have a lot of error-testing code, because I've been bitten by this too many times **and I still get it wrong the first time**.

aragos: Ugh. I'll reserve my rant about this for another time. :)

src/test/shell/bazel/toolchain_test.sh
katre: See test_platforms_options_error_invalid_target: https://source.bazel.build/bazel/+/master:src/test/shell/bazel/toolchain_test.sh;l=865?q=toolchain_test.sh

PR #10316, opened 2019-11-26
Allow building the Bazel C++ code on OpenBSD.

aldersondrive:
This PR is smaller than it appears, because the *_openbsd.cc files are largely copy/pasted from the corresponding *_freebsd.cc files. The GitHub web UI does not show this similarity, AFAIK. You could try:

```
diff src/main/cpp/blaze_util_freebsd.cc src/main/cpp/blaze_util_openbsd.cc
diff src/main/native/unix_jni_freebsd.cc src/main/native/unix_jni_openbsd.cc
```

This change, split out of the larger PR https://github.com/bazelbuild/bazel/pull/10274, is part of the OpenBSD port in https://github.com/bazelbuild/bazel/issues/10250.

### DIFF
--- a/src/main/cpp/blaze_util_openbsd.cc
+++ b/src/main/cpp/blaze_util_openbsd.cc
@@ -1,0 +1,168 @@
+// Copyright 2015 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>  // errno, ENAMETOOLONG
+#include <limits.h>
+#include <pwd.h>
+#include <signal.h>
+#include <spawn.h>
+#include <stdlib.h>
+#include <string.h>  // strerror
+#include <sys/mount.h>
+#include <sys/param.h>
+#include <sys/queue.h>
+#include <sys/socket.h>
+#include <sys/sysctl.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#include "src/main/cpp/blaze_util.h"
+#include "src/main/cpp/blaze_util_platform.h"
+#include "src/main/cpp/util/errors.h"
+#include "src/main/cpp/util/exit_code.h"
+#include "src/main/cpp/util/file.h"
+#include "src/main/cpp/util/logging.h"
+#include "src/main/cpp/util/path.h"
+#include "src/main/cpp/util/port.h"
+#include "src/main/cpp/util/strings.h"
+
+namespace blaze {
+
+using blaze_util::GetLastErrorString;
+using std::string;
+
+string GetOutputRoot() {
+  char buf[2048];
+  struct passwd pwbuf;
+  struct passwd *pw = NULL;
+  int uid = getuid();
+  int r = getpwuid_r(uid, &pwbuf, buf, 2048, &pw);
+  if (r != -1 && pw != NULL) {
+    return blaze_util::JoinPath(pw->pw_dir, ".cache/bazel");
+  } else {
+    return "/tmp";
+  }
+}
+
+void WarnFilesystemType(const blaze_util::Path &output_base) {
+  struct statfs buf = {};
+  if (statfs(output_base.AsNativePath().c_str(), &buf) < 0) {
+    BAZEL_LOG(WARNING) << "couldn't get file system type information for '"
+                       << output_base.AsPrintablePath()
+                       << "': " << strerror(errno);
+    return;
+  }
+
+  if (strcmp(buf.f_fstypename, "nfs") == 0) {
+    BAZEL_LOG(WARNING) << "Output base '" << output_base.AsPrintablePath()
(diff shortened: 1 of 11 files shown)


aldersondrive:
@hlopko FYI

aldersondrive:
Oops, I just realized that the code in `src/main/native` is part of the server, not part of the client. I've renamed the PR accordingly. If you prefer that I move the JNI C++ changes to a separate PR, just let me know.

src/main/cpp/blaze_util_openbsd.cc
```diff
@@ -0,0 +1,168 @@
+// Copyright 2015 The Bazel Authors. All rights reserved.
```
laszlocsomor: The diff seems short indeed. Could you reuse the FreeBSD file without forking it? It's a risk the two files won't stay in sync.

src/main/cpp/util/port.h
```diff
@@ -71,6 +71,21 @@
 #endif
 
 
+// CAN_FIND_OWN_EXECUTABLE_PATH
+//
+// Indicates that a running process can find a path to its own executable.
+#if !defined(__OpenBSD__)
```
laszlocsomor: This logic assumes every other platform can find their own executable. That isn't guaranteed. What I believe you actually want is:
```
#if defined(__OpenBSD__)
#define CANNOT_FIND_OWN_EXECUTABLE_PATH=1
#endif
```

This tells a different story: we know on OpenBSD we can't find our own executable, without assuming anything of other platforms.

src/main/cpp/util/port.h
```diff
@@ -71,6 +71,21 @@
...
+#define CAN_FIND_OWN_EXECUTABLE_PATH
+#endif
+
+// HAVE_EMULTIHOP
+//
+// Indicates that errno.h defines EMULTIHOP.
+#if !defined(__OpenBSD__)
+#define HAVE_EMULTIHOP
```
laszlocsomor: Same here as above.

Defining `HAS_NO_EMULTIPHOP=1` allows achieving the same effect as `HAVE_EMULTIPOP` but without making assumptions of other platforms.

rupertks:
@jmmv FYI

src/main/cpp/blaze_util_openbsd.cc
aldersondrive: The following functions are the same in `blaze_util_freebsd.cc` and `blaze_util_openbsd.cc`:

```
GetOutputRoot
WarnFilesystemType
GetMillisecondsMonotonic
SetScheduling
IsSharedLibrary
ConfigureDaemonProcess
WriteSystemSpecificProcessIdentifier
VerifyServerProcess
ExcludePathFromBackup
GetExplicitSystemLimit
```

The function bodies are 65 lines altogether.

Options include:

1. Do nothing. Keep the `blaze_util_freebsd.cc` and `blaze_util_openbsd.cc` files as they are.

2. Combine them into a single `blaze_util_bsd.cc` file that contains a bunch of `#ifdef`s. This would be messy, hard to read, and based on the flawed premise that all BSD variants will share substantially similar code.

3. Move the common functions into a `blaze_util_bsd_common.inc` file, which both `blaze_util_freebsd.cc` and `blaze_util_openbsd.cc` can `#include`. This would be more complex and indirect, and possibly not worth it for only 65 lines.

4. Something else.

How would you like to proceed? (IMO option 1 might be the least bad option.)

BTW, a few functions are also duplicated among the Linux and Darwin files.

src/main/cpp/util/port.h
aldersondrive: That would also work. Before I make this change, though: Are you sure?

- The concern is that if we use `CAN_FIND_OWN_EXECUTABLE_PATH` and forget to update `port.h` when adding a new platform, we will silently assume that the new platform supports binaries finding their own executable paths. But, the same is equally true of the `CANNOT_FIND_OWN_EXECUTABLE_PATH` alternative: If we forget to update `port.h` when adding a new platform, then `CANNOT_FIND_OWN_EXECUTABLE_PATH` won't be `#define`d for the new platform, and thus the rest of the code will assume that the platform supports binaries finding their own executable paths.

- The negative naming is hard to read. Code is easier to read when symbols say what *is* true, not what *isn't* true. For example, we probably wouldn't want to refactor `HAVE_ATTRIBUTE` as `DONT_HAVE_ATTRIBUTE`.

So, AFAICS the proposed change would not meet the stated goal, and meanwhile would make the code less readable.

A robust (but longer) solution might be something like this:

```
#if defined(__linux) || defined(__APPLE__) || __defined(__FreeBSD__) || defined(_WIN32) || defined(__CYGWIN__)
#define CAN_FIND_OWN_EXECUTABLE_PATH
#elif defined(__OpenBSD__)
// This platform does not support processes finding their own executable paths.
#else
#error Unrecognized platform.
#endif
```

What do you think?

src/main/cpp/blaze_util_openbsd.cc
laszlocsomor: You're right, let's go with (1).

src/main/cpp/util/port.h
laszlocsomor: Thanks for explaining that. You're right, my solution also makes assumptions about other platforms.

The robust solution looks safer than the other two approaches without much more complexity, so yes I think that looks like the best solution.

aldersondrive: Done.

src/main/cpp/util/port.h
aldersondrive: I realized that we could eliminate `HAVE_EMULTIHOP` entirely. It no longer exists.

aldersondrive:
Thank you for the review.

Does anything else need to happen before this PR is merged?

laszlocsomor:
I already imported it; needs to pass internal code review.
Here's the process: https://bazel.build/contributing.html#patch-acceptance-process

PR #10321, opened 2019-11-27
Fix VS15 toolchain configure: hide output from VsDevCmd.bat

nikola-sh:
VsDevCmd.bat outputs header
```
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.3.10
** Copyright (c) 2019 Microsoft Corporation
**********************************************************************
```
which interferes with 
```
echo %VCINSTALLDIR%
```
and doesn't allow to configure VS toolchain.

### DIFF
--- a/tools/cpp/windows_cc_configure.bzl
+++ b/tools/cpp/windows_cc_configure.bzl
@@ -157,7 +157,7 @@ def find_vc_path(repository_ctx):
         repository_ctx.file(
             "get_vc_dir.bat",
             "@echo off\n" +
-            "call \"" + script + "\"\n" +
+            "call \"" + script + "\" > NUL\n" +
             "echo %VCINSTALLDIR%",
             True,
         )


nikola-sh:
> @googlebot I signed it!

tools/cpp/windows_cc_configure.bzl
```diff
@@ -157,7 +157,7 @@ def find_vc_path(repository_ctx):
         repository_ctx.file(
             "get_vc_dir.bat",
             "@echo off\n" +
-            "call \"" + script + "\"\n" +
+            "call \"" + script + "\" > NUL\n" +
```
laszlocsomor: This hides all output, when all we want to hide is the banner.
Looking into vsdevcmd.bat, I see it's enough to set `%VSCMD_ARG_no_logo%` to some non-empty value:

```
"set VSCMD_ARG_no_logo=1" +
"call \"" + script + "\"\n" +
```

drakonka: I just ran into this  yesterday as well and  `>nul` was also my workaround. I am curious as to why we don't want to hide _all_ output from the `VsDevCmd.bat` call? We echo `%VCINSTALLDIR%` right after and rely on that echo output to get the `vc_dir` path from what I can see. So it seems like _any_ output from the `VsDevCmd.bat` (or any output outside of `echo %VCINSTALLDIR%` would still result in a misformatted path.

laszlocsomor: Ah, of course, you're right.

PR #10333, opened 2019-11-28
Fix example

spudtrooper:
Running example as is fails with: type 'depset' is not iterable


spudtrooper:
@googlebot I signed it!

PR #10345, opened 2019-12-02
Support case-sensitive package loading

laszlocsomor:
The new --[no]experimental_check_label_casing flag
enables a new validation step in package loading:
when enabled, Bazel requires that package paths be
case-correct.

When the user runs "bazel build //Foo/Bar:Baz",
Bazel looks up the package "Foo/Bar" by stat'ing
"Foo/Bar/BUILD". If the file exists, then the
package exists. Every Label that Bazel creates in
this package is derived from the "Foo/Bar" string.

On Linux the filesystem is case-sensitive, so the
path casing must be correct for stat() to succeed,
so it's impossible to load the package "//Foo/Bar"
as "//FOO/baR".

On Windows and macOS the filesystem ignores the
difference between lower and upper case letters,
so stat("foo/bAr/BUILD") and stat("FOO/bar/build")
would both suceed, so "bazel build //foo/bAr:Baz"
and "bazel build //FOO/bar:Baz" both work. This
leads to bugs such as #6401, #8009, and #8799.

Change-Id: I5ad86a0ede6f85aa9cac8e8e56072630738db55c


laszlocsomor:
I'll split up this PR to smaller chunks.

PR #10357, opened 2019-12-03
VFS: add RootedPathAndCasing

laszlocsomor:
This class stores a RootedPath and the String
value of its relative part.

This class allows distinguishing RootedPaths on
case-insensitive filesystems, if two RootedPaths
refer to the same file but use different casing.

The motivation is to support validating package
path casing. (See issue #8799.) My plan is to add
a new SkyFunction called PathCasingLookupFunction,
which validates that a certain RootedPath is
correctly cased. The SkyKey can't just take a
RootedPath, because on Windows the path semantics
dictate that RootedPath objects are compared
without regard to casing. Thus we must also store
the particular casing in the SkyKey, and so we
need the RootedPathWithCasing class.

The class only compares the casing of the relative
parts, not of the root parts. Reason is, this
class will be used to validate the casing of
package paths, which are always relative to the
package root whose casing we needn't worry about.

See https://github.com/bazelbuild/bazel/issues/8799

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/vfs/RootedPathAndCasing.java
+++ b/src/main/java/com/google/devtools/build/lib/vfs/RootedPathAndCasing.java
@@ -1,0 +1,78 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


meteorcloudy:
The change itself looks good. But can you explain a bit more how are we going to use this? Replacing RootedPath? And will it cause significant memory regression?

laszlocsomor:
Sure! I explained its purpose in the commit message.

Good question about memory tax. I haven't measured it.
What I do know is I'll put the new path-casing validation logic behind an experimental flag. Unless you enable that flag, no extra SkyFunctions will be computed, hence no memory cost.

This class will not replace RootedPath; it builds on top of it.

PR #10360, opened 2019-12-03
Require non-null descriptions for MiddlemanActions

werkt:
Prevent an uninformative "null for <target>" name from being presented
for MiddlemanAction prettyPrint by mandating a description not be null.
Use a descriptive name for the CcCompilationHelper middleman based on
the objc implementation, populated from CcCommon. Limit the objc
interpretation within the helper to the matching objc purposes.


hlopko:
Thank you @werkt, importing.

PR #10361, opened 2019-12-04
Populate SpawnMetrics and improve CP output

werkt:
Prevent non-remote executions from remote stat output in
CriticalPathComponent.
Populate metrics recorded in remote execution.
Emit SpawnExecutedEvent to supply metrics to CriticalPathComputer
subscriber after 'Actual execution' in AbstractSpawnStrategy.


werkt:
@buchgr ping on this

src/main/java/com/google/devtools/build/lib/runtime/CriticalPathComponent.java
```diff
@@ -185,7 +188,10 @@ public int getId() {
    * the longestPhaseSpawnRunnerName to the longest running spawn runner name across all phases if
    * it exists.
    */
-  void addSpawnResult(SpawnMetrics metrics, @Nullable String runnerName) {
+  void addSpawnResult(SpawnMetrics metrics, @Nullable String runnerName, boolean wasRemote) {
+    if (wasRemote) {
+      this.remote = true;
+    }
```
buchgr: this.remote = wasRemote?

src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java
```diff
@@ -213,6 +213,9 @@ String getDetailMessage(
   /** Returns a file path to the action metadata log. */
   Optional<MetadataLog> getActionMetadataLog();
 
+  /** Whether the spawn result was obtained through remote strategy. */
```
buchgr: s/remote strategy/remote execution/

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -284,11 +300,20 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
...
+                Stopwatch uploadTime = Stopwatch.createStarted();
                 remoteCache.ensureInputsPresent(merkleTree, additionalInputs);
+                spawnMetrics.setUploadTime(
+                    uploadTime.elapsed().minus(
+                        networkTime.getDuration().minus(networkTimeStart)));
               }
               ExecuteResponse reply;
               try (SilentCloseable c = prof.profile(REMOTE_EXECUTION, "execute remotely")) {
+                // remove accounting for execution wait time
+                networkTime.disable();
```
buchgr: It seems cleaner to special case the execute method in the client interceptor.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -323,19 +350,55 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
...
 
+  /** conversion utility for protobuf Timestamp difference to java.time.Duration */
+  private static Duration between(Timestamp from, Timestamp to) {
+    return Duration.ofNanos(Durations.toNanos(Timestamps.between(from, to)));
+  }
+
+  private static void spawnMetricsAccounting(SpawnMetrics.Builder spawnMetrics, ExecutedActionMetadata executionMetadata) {
+    if (!executionMetadata.getWorker().isEmpty()) {
+      // Accumulate queueTime from any previous attempts
+      Duration remoteQueueTime = spawnMetrics.build().remoteQueueTime().plus(
```
buchgr: I have a hard time understanding why you are summing up the queue times. Can you explain this a bit more please?

src/main/java/com/google/devtools/build/lib/runtime/CriticalPathComponent.java
werkt: it would be more accurate to say `this.remote = wasRemote || this.remote;`: I don't want a subsequent local spawn (which seems to be a thing now with discoverInputs) to reset the flag for a component (i.e. action) which had a previous remote spawn result. This is a 'flip high, ignore low' switch.

src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java
werkt: strategy was intentional here - both pure remote cache and remote execution with their independent RemoteModule configuration emit true here - any SpawnResults created under the `build/lib/remote` package will have this set.

src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
```diff
@@ -0,0 +1,123 @@
...
+import io.grpc.ClientInterceptor;
+import io.grpc.Context;
+import io.grpc.ForwardingClientCall;
+import io.grpc.ForwardingClientCallListener;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.time.Duration;
+
+/** Reentrant wall clock stopwatch and grpc interceptor for network waits. */
```
buchgr: Why do you need the reentrant part? I don't follow why this interceptor should be more complex than take time on first call and stop taking time on close?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: I'd be on board with that, let me see how it goes.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -323,19 +350,55 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
     }
   }
 
+  /** conversion utility for protobuf Timestamp difference to java.time.Duration */
+  private static Duration between(Timestamp from, Timestamp to) {
+    return Duration.ofNanos(Durations.toNanos(Timestamps.between(from, to)));
+  }
+
+  private static void spawnMetricsAccounting(SpawnMetrics.Builder spawnMetrics, ExecutedActionMetadata executionMetadata) {
```
buchgr: consider adding a test for this method.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -284,11 +300,20 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
                 Map<Digest, Message> additionalInputs = Maps.newHashMapWithExpectedSize(2);
                 additionalInputs.put(actionKey.getDigest(), action);
                 additionalInputs.put(commandHash, command);
+                Duration networkTimeStart = networkTime.getDuration();
+                Stopwatch uploadTime = Stopwatch.createStarted();
                 remoteCache.ensureInputsPresent(merkleTree, additionalInputs);
+                spawnMetrics.setUploadTime(
+                    uploadTime.elapsed().minus(
+                        networkTime.getDuration().minus(networkTimeStart)));
```
buchgr: Can you please explain why you are subtracting here?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: Yes, that was based on the commentary in the SpawnMetrics class: the remoteQueueTime getter in SpawnMetrics has a comment explaining this [here](https://github.com/bazelbuild/bazel/blob/bca23e8bb9fc220a6c523a59b2aa88979deb7981/src/main/java/com/google/devtools/build/lib/actions/SpawnMetrics.java#L110-L113)

src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
werkt: Because uploads and downloads occur concurrently and asynchronously during ensureInputsProvided and remoteCache.download respectively, and have start and stop times interspersed throughout their blocking periods of operations, with possible times during which no network activity is outstanding

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: Agreed

src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTree.java
```diff
@@ -68,21 +69,37 @@ public ByteString getBytes() {
...
   public Digest getRootDigest() {
     return rootDigest;
   }
 
+  /** Returns the number of files represented by this merkle tree */
+  public long getInputFiles() {
+    return inputFiles;
+  }
+
+  /** Returns the number of bytes represented by this merkle tree, including directories */
```
buchgr: Returns the sum of file sizes plus protobuf sizes used to represent this merkle tree?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: Because the tacit understanding in SpawnMetrics, evident in this [otherTime()](https://github.com/bazelbuild/bazel/blob/c35746d7f3708acb0d39f3082341de0ff09bd95f/src/main/java/com/google/devtools/build/lib/actions/SpawnMetrics.java#L162-L171) calculation, is an expectation of cumulative totalTime representing all component times, which are all expected to match the wall clock time calculations.

Time spent in one metric, i.e. the wall time spent uploading all inputs, must not include the time spent during another metric, in this case the time spent waiting for network responses, so that they can be used to determine a percentage of total wall clock time each metric represents.

src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTree.java
werkt: Yes. I assume that's what you want me to change the comment to be?

src/main/java/com/google/devtools/build/lib/runtime/CriticalPathComponent.java
buchgr: Consider adding a comment that remote about this.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
buchgr: Great. It should be quite straightforward. The gRPC stubs also expose a MethodDescriptor for each RPC.

src/main/java/com/google/devtools/build/lib/runtime/CriticalPathComponent.java
werkt: Done

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: Added a comment detailing the reasoning here.

werkt:
Any update on this?

werkt:
Again, a request for status on this. What is the holdup?

werkt:
@jin you are a reliable source of triage. Everyone on this review has declared their intention to be removed from bazel ownership. Can you please relocate this to someone who can be responsible for moving it forward?

There are obvious changes going on with blaze that are making use of these objects that are not in bazel (and creating merge conflicts). The last piece of information I have is that this change was under an opaque internal review. Can someone please provide an update, reason, approval, or rejection of this?

jin:
I'll escalate this to someone who can take a look.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -183,7 +186,7 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
                 remoteOptions.remoteExecutor,
                 remoteOptions.remoteProxy,
                 authAndTlsOptions,
-                loggingInterceptor);
+                interceptors.toArray(new ClientInterceptor[0]));
```
ulfjack: I'd prefer to pass the list in instead of converting it to an array.

src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTree.java
```diff
@@ -128,31 +145,35 @@ public static MerkleTree build(
...
+            inputBytes.addAndGet(protoDirDigest.getSizeBytes());
           }
           Directory protoDir = b.build();
           Digest protoDirDigest = digestUtil.compute(protoDir);
           digestDirectoryMap.put(protoDirDigest, protoDir);
           m.put(dirname, protoDirDigest);
         });
-    return new MerkleTree(digestDirectoryMap, digestPathMap, m.get(PathFragment.EMPTY_FRAGMENT));
+    Digest rootDigest = m.get(PathFragment.EMPTY_FRAGMENT);
+    return new MerkleTree(digestDirectoryMap, digestPathMap, rootDigest, tree.numFiles(), inputBytes.addAndGet(rootDigest.getSizeBytes()));
```
ulfjack: I would pull out the inputBytes update into a separate line.

src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
ulfjack: I think internally, Google is computing network time as total wall time minus all known sub-intervals. Maybe that's good enough here as well?

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -320,19 +345,56 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
...
 
+  /** conversion utility for protobuf Timestamp difference to java.time.Duration */
+  private static Duration between(Timestamp from, Timestamp to) {
+    return Duration.ofNanos(Durations.toNanos(Timestamps.between(from, to)));
+  }
+
+  @VisibleForTesting
+  static void spawnMetricsAccounting(SpawnMetrics.Builder spawnMetrics, ExecutedActionMetadata executionMetadata) {
+    if (!executionMetadata.getWorker().isEmpty()) {
+      // Accumulate queueTime from any previous attempts
```
ulfjack: Can you add a comment on what happens when some of these aren't set?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
werkt: This was an artifact of using variadic, and was fortunately just conflicted by someone adding another interceptor. Won't appear in the updated patch

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
werkt: Commenting here on network time, UI won't let me address that individually.

I'm not sure why "Good Enough" is good enough here, especially when it comes to blaming the network for things that bazel is doing in between. I see the logic - technically we can't know how much lag occurs throughout the action - but with each concurrent request happening and being metered separately for uploads and downloads' times, this is a solid tool for measuring the explicit wait time for responses after the requests have gone out.

Misleading cost measurement tools cost more than not having them at all, and I want to avoid blaming the network if it's full GC occurring attributable to Other.

ulfjack: You can't control which bucket a full GC gets attributed to. If Bazel starts a full GC after the reply came back, but before you looked at the clock, then it gets officially counted as network time. There's no way for us to distinguish between the cases where the full GC delayed processing (and shouldn't be counted as network time), and where it happened while a request was in-flight, but the GC didn't actually delay processing.

I'm just concerned about the amount of complexity the network timing introduces, and it's not clear to me that it's worth it. At the same time, it's important to get this change in, because we currently have *no*  visibility into the metrics, and wall-time is just wrong.

ulfjack:
I think this can be merged. It's not perfect, but it's also not a public API, and I don't have a conter-proposal (and we really want the spawn metrics to be tracked more accurately).

@philwo

philwo:
Thanks, Ulf! I've imported it and will try to get it submitted.

ulfjack:
Ping?

philwo:
Hi! Sorry for the delay. This was quite tricky to merge, as I had to modify some of our internal code to build against the modified interface, but it's all good now! Submitting :)

PR #10366, opened 2019-12-04
WIP: Add case-checking PackageIdentifier

laszlocsomor:
If the "bazel.check_label_casing" JVM property is
"1", then PackageIdentifier objects will store not
only the RepositoryName and PathFragment, but also
the toString() values of those.

These augmented PackageIdentifiers (actually:
CaseCheckingPackageIdentifiers) allow telling
apart PackageIdentifiers whose RepositoryName and
PathFragment values are equal but are cased
differently. This scenario is common on
case-insensitive filesystems, where the paths
"foo/x" and "FoO/X" are equal but differently
cased.

Motivation is to allow case-checking the package
names in PackageLookupFunction.

The SkyKey of PackageLookupFunction contains a
PackageIdentifier, which holds a reference to the
RepositoryName and PathFragment (package
fragment). On case-insensitive filesystems, both
of these objects check equality without regard to
case, and so do PackageIdentifiers as a
consequence.

But PackageIdentifier objects are interned for
better memory usage, which means that on a
case-insensitive filesystem it's impossible to
create two PackageIdentifiers whose RepositoryName
and PathFragment are equal but differently cased.
And that means it's also impossible to check in
the PackageLookupFunction whether the requested
package path (and thus the PackageIdentifier) has
correct casing.

With this PR, PackageIdentifier.create() will
return CaseCheckingPackageIdentifier when case
checking is enabled, so PackageLookupFunction will
know exactly what casing the user requested, and
can therefore validate that casing and deny
loading the package if the casing is wrong. And
that is what would fix https://github.com/bazelbuild/bazel/issues/8799

Change-Id: I46106f83c0a5b16fa347dd8a07e3c5342248c605

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/cmdline/LabelConstants.java
+++ b/src/main/java/com/google/devtools/build/lib/cmdline/LabelConstants.java
@@ -25,4 +25,18 @@ public class LabelConstants {
   public static final PathFragment WORKSPACE_DOT_BAZEL_FILE_NAME =
       PathFragment.create("WORKSPACE.bazel");
   public static final String DEFAULT_REPOSITORY_DIRECTORY = "__main__";
+
+  /**
+   * Whether Bazel should check (true) or trust (false) the casing of Labels.
+   *
+   * <p>A Package "//foo" exists if its BUILD file "foo/BUILD" exists, i.e. if stat("foo/BUILD")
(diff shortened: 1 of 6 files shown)


jin:
Closing this for now, please ping/reopen if you're intending to work on this.

PR #10380, opened 2019-12-07
Refactor GetProcessCWD to return a unique_ptr.

aldersondrive:
This way we can use `nullptr`, rather than a sentinel value (the empty string), to indicate that the directory is unknown.

This change resolves a TODO. It was originally requested by @jmmv.

`std::make_unique` and `absl::make_unique` seem to be unavailable, so I've made do with `std::unique_ptr<blaze_util::Path>(new blaze_util::Path(...))`.

### DIFF
--- a/src/main/cpp/blaze_util_bsd.cc
+++ b/src/main/cpp/blaze_util_bsd.cc
@@ -154,10 +154,10 @@ void SetScheduling(bool batch_cpu_scheduling, int io_nice_level) {
   // Stubbed out so we can compile.
 }
 
-blaze_util::Path GetProcessCWD(int pid) {
+std::unique_ptr<blaze_util::Path> GetProcessCWD(int pid) {
 #if defined(HAVE_PROCSTAT)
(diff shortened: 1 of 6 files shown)


aldersondrive:
Ping?

src/main/cpp/blaze_util_bsd.cc
```diff
@@ -184,9 +184,9 @@ blaze_util::Path GetProcessCWD(int pid) {
     procstat_freeprocs(procstat, p);
   }
   procstat_close(procstat);
-  return blaze_util::Path(cwd);
+  return std::unique_ptr<blaze_util::Path>(new blaze_util::Path(cwd));
```
jmmv: I think the blessed idiom these days is to use `std::make_unique`. Not sure that'll pass CI though.

aldersondrive: I tried `std::make_unique`. The build failed.

PR #10405, opened 2019-12-12
Pass execution_requirements through sha256 run

werkt:
Make execution_requirements specifiable by callers of sha256 to enable
propagation of common.tags and control resultant action behavior


buchgr:
LGTM

PR #10409, opened 2019-12-12
jvm.bzl: Load Java rules from @rules_java

### LINKED ISSUE #10046: Prepare {java|jvm}_import_external for --incompatible_load_java_rules_from_bzl option flip
See #8746 for details.

beasleyr-vmw:
Resolves #10046.

### DIFF
--- a/tools/build_defs/repo/jvm.bzl
+++ b/tools/build_defs/repo/jvm.bzl
@@ -252,6 +252,10 @@ def jvm_maven_import_external(
     srcjar_urls = kwargs.pop("srcjar_urls", None)
 
     rule_name = kwargs.pop("rule_name", "java_import")
+    rule_load = kwargs.pop(
+        "rule_load",
+        'load("@rules_java//java:defs.bzl", "java_import")',
+    )
 
     if fetch_sources:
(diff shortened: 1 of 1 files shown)


beasleyr-vmw:
cc @jin

jin:
This will incur a breaking change on all downstream users of jvm_import_external, since we have not flipped the incompatible change on `@rules_java`. Users do not need to add `@rules_java` now, so if we merge this, projects will break.

This would be better coordinated with someone on the server rules team (@iirina?).

beasleyr-vmw:
I'm confused. Hasn't Bazel implicitly satisified rules_java since 0.29 via jdk.WORKSPACE? (AFAIK, jdk.WORKSPACE is magically appended to the user's WORKSPACE file, so all users automatically have a rules_java repo.)

https://github.com/bazelbuild/bazel/blob/55c75ed16e8c456d92510df13da4ffe01e0b998d/src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE#L260-L269

jin:
Oh, hum, my bad! LGTM.

Issue #8685, opened 2019-06-20
Allow customizing Python stub script's shebang in the Python toolchain


brandjon:
Currently the stub script requires (via its shebang) that the target system have a `python` command available on `PATH`. If the target is PY3, the user might reasonable expect that it work if `python` is not available but `python3` is. We can modify the shebang in this case to be `#!/usr/bin/env python3`. However doing so might break if the system has a (Python 3) `python` command but not a `python3` command.

See #8665 for a case where the shebang has been a problem in practice.

brandjon:
More generally, we can support customizing the shebang as part of the Python toolchain. I'm not clear on how useful this feature would be in the short to medium term. Long term we'd like to eliminate the Python stub altogether and replace it with a C++ one.

brandjon:
Repurposing this bug to track the FR to customize the shebang string in the toolchain. That supersedes requests to make it work with a particular interpreter version / system environment xyz. See also #11269.

beasleyr-vmw:
@brandjon I'm not sure if the Bazel maintainers prefer to resolve this independent of issues like #8446 or #137, but I'm pursuing this internally and would be happy to generate a PR within the next day or two.

- Any naming suggestions for the new `py_runtime` parameter?
- Are `PyRuntimeInfoTest` test cases sufficient, or do I need to get down and dirty with `python_stub_test.sh`?

chrislovecnm:
@brandjon what is the status on this issue?  I just ran smack into it with trying to do a hermetic python build.  When was this issue introduced?

fahhem:
Potentially fixed by: https://github.com/bazelbuild/bazel/pull/6632

meteorcloudy:
This issue shouldn't be P4 anymore since Python 2 has already been deprecated.

meteorcloudy:
@brandjon Is there any plan to fix this?

codeschlosser:
Seems like the problem holds if `--build_python_zip` flag is used. At least when running `bazel test` (still puts `#!/usr/bin/env python` as shebang for the binary, everything is OK in the zip archive)

meteorcloudy:
@codeschlosser Are you testing this on Windows? 
Probably the problem is in the python launcher for Windows:
https://github.com/bazelbuild/bazel/blob/ba4435cc81baee8621b75337b33986cc4fa13731/src/tools/launcher/python_launcher.cc#L45

codeschlosser:
[...]

PR #10432, opened 2019-12-17
Tweak Python semantics for Python 3 and OpenBSD.

aldersondrive:
As discussed at https://github.com/bazelbuild/bazel/pull/10432#issuecomment-571746342, Bazel is already switching to assume that the Python binary is called `python3`, not `python`.

This change is also useful for the OpenBSD port in https://github.com/bazelbuild/bazel/issues/10250. (On OpenBSD, the python executable is called `python3`, and no `python` binary exists.)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
@@ -235,13 +235,14 @@ public class BazelPythonSemantics implements PythonSemantics {
 
       if (OS.getCurrent() != OS.WINDOWS) {
         PathFragment shExecutable = ShToolchain.getPathOrError(ruleContext);
+        String pythonExecutableName = OS.getCurrent() == OS.OPENBSD ? "python3" : "python";
         ruleContext.registerAction(
             new SpawnAction.Builder()
                 .addInput(zipFile)
                 .addOutput(executable)
                 .setShellCommand(
                     shExecutable,
-                    "echo '#!/usr/bin/env python' | cat - "
+                    "echo '#!/usr/bin/env " + pythonExecutableName + "' | cat - "
                         + zipFile.getExecPathString()
                         + " > "
                         + executable.getExecPathString())
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
@@ -235,6 +235,8 @@ public class BazelPythonSemantics implements PythonSemantics {
 
       if (OS.getCurrent() != OS.WINDOWS) {
         PathFragment shExecutable = ShToolchain.getPathOrError(ruleContext);
+        // TODO(#8685): Remove this special-case handling as part of making the proper shebang a
+        // property of the Python toolchain configuration.
         String pythonExecutableName = OS.getCurrent() == OS.OPENBSD ? "python3" : "python";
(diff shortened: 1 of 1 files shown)


benjaminp:
Maybe you can just unconditionally use `env python3`? It's been announced that Bazel assumes Python 3 now. It would be in the spirit of 29d45006f700186c700bf3fa5348ca906dc19dc9.

aldersondrive:
> Maybe you can just unconditionally use `env python3`? It's been announced that Bazel assumes Python 3 now. It would be in the spirit of [29d4500](https://github.com/bazelbuild/bazel/commit/29d45006f700186c700bf3fa5348ca906dc19dc9).

That would work for OpenBSD, but would also change behavior on all platforms. Notably, FreeBSD has no `python3` binary, so that change would break Bazel on FreeBSD.

On the other hand, if Bazel is indeed going to assume the existence of a `python3` binary (per https://github.com/bazelbuild/bazel/commit/29d45006f700186c700bf3fa5348ca906dc19dc9), then what you describe would be okay and the FreeBSD port needs to change in this respect anyway.

How would you like to proceed? I'm fine with whichever way you folks think best. I just hesitate to push a PR that changes behavior on all platforms and breaks one of them without confirmation from Bazel team that doing so is really the right way to go.

aldersondrive:
> How would you like to proceed?

Update: I checked a clean FreeBSD system, and it *does* have a `python3` in the `PATH`. Maybe there was something wrong with the other FreeBSD system I checked earlier.

In any case, it looks like Bazel already assumes the existence of `python3`, and apparently `python3` does exist on FreeBSD after all, so I've gone ahead and made the change you suggested. PTAL.

brandjon:
The issue with putting python3 in the shebang is that 1) it's an incompatible change that 2) affects *all* `py_binary` targets. In contrast, 29d4500 only affects a test used in the development of Bazel itself. While we require Python 3 for Bazel development and for certain baked-into-bazel tools that users may depend on, we do not require it for arbitrary user `py_binary` targets.

For supporting this use case, I think we'd need an attribute on the Python toolchain to set the shebang to use in the stub script.

See #8685 for additional commentary along these lines. I think we can dedup this against that one.

aldersondrive:
Would you be okay with [the first version of this PR](https://github.com/bazelbuild/bazel/pull/10432/commits/fa58eadc615c9133c962a0425272c33cd6bbf673)?

It's not an incompatible change, because it affects behavior only on OpenBSD, where Bazel support is still under development.

https://github.com/bazelbuild/bazel/issues/8685 sounds like a much larger and more complex change that likely won't be done anytime soon.

brandjon:
I'm hesitant to add additional ad hoc logic to the native rules. That said, given that

1. this only affects a platform that isn't already fully supported,
2. the new support is experimental, and
3. we can migrate to #8685 by adding a toolchain that mimics the proposed `OS.getCurrent()` check,

it may be ok. @katre, do you see any platform-related concerns with a migration path from this native check to a future built-into-bazel platform definition for OpenBSD? (And can you confirm our support status for OpenBSD?)

katre:
I am fine with this change, especially if we are moving towards having the python binary in the toolchain (so we can remove it). I'd like to see a TODO to remove the special casing in the future (possibly referencing #8685).

Yes, we want to support OpenBSD and if there aren't existing constraints for it, I'd support a PR to add them.

brandjon:
Ok, sounds good. @aldersondrive, can you please revert to the first version and add a TODO referencing #8685? (Standard format is `// TODO(#8685): blah blah blah`)

brandjon:
Should be merged tomorrow.

PR #10436, opened 2019-12-18
Add C++ toolchain configuration for OpenBSD.

aldersondrive:
In my limited testing, these changes appear to work. When I built Bazel on OpenBSD incorporating these changes, the resulting `bazel` binary could build C++ libraries and binaries. Likewise on FreeBSD.

This change, split out of the larger PR https://github.com/bazelbuild/bazel/pull/10274, is part of the OpenBSD port in https://github.com/bazelbuild/bazel/issues/10250.


jin:
cc @jmmv, who may know something about BSD here :)

jmmv:
I believe this works based on the description and what I see in the diff.

If the new OpenBSD-specific file was generated with a simple copy+find/replace, and given that both the old and new file are Starlark... could we generalize them instead? Similar to what we did with a single file for all BSDs in the C++ client code.

aldersondrive:
> I believe this works based on the description and what I see in the diff.
> 
> If the new OpenBSD-specific file was generated with a simple copy+find/replace, and given that both the old and new file are Starlark... could we generalize them instead? Similar to what we did with a single file for all BSDs in the C++ client code.

Done. PTAL.

tools/cpp/bsd_cc_toolchain_config.bzl
```diff
@@ -57,20 +57,20 @@ all_link_actions = [
 def _impl(ctx):
     cpu = ctx.attr.cpu
     compiler = "compiler"
-    toolchain_identifier = "local_freebsd" if cpu == "freebsd" else "stub_armeabi-v7a"
-    host_system_name = "local" if cpu == "freebsd" else "armeabi-v7a"
-    target_system_name = "local" if cpu == "freebsd" else "armeabi-v7a"
-    target_libc = "local" if cpu == "freebsd" else "armeabi-v7a"
-    abi_version = "local" if cpu == "freebsd" else "armeabi-v7a"
-    abi_libc_version = "local" if cpu == "freebsd" else "armeabi-v7a"
+    toolchain_identifier = "local_{}".format(cpu) if cpu == "freebsd" or cpu == "openbsd" else "stub_armeabi-v7a"
```
jmmv: I'd do:

`is_bsd = cpu == "freebsd" || cpu == "openbsd"`

and then use `is_bsd` in all lines to avoid duplication and ensure they all remain consistent. Would also help cut this long line.

tools/cpp/cc_toolchain_config.bzl
```diff
@@ -685,7 +698,7 @@ def _impl(ctx):
                 ),
             ],
         )
-    elif (ctx.attr.cpu == "freebsd"):
+    elif (ctx.attr.cpu == "freebsd" or ctx.attr.cpu == "openbsd"):
```
jmmv: nit: Break in two lines for consistency with the rest?

tools/cpp/bsd_cc_toolchain_config.bzl
aldersondrive: Much better. Done.

tools/cpp/cc_toolchain_config.bzl
aldersondrive: Done.

PR #10444, opened 2019-12-19
output_directory CompileBuildVariable

werkt:
Provide a substitution for the output_directory, which is always
provided as the output_file's containing directory. This is useful
for providing a directory to tools which explicitly demand a scratch
location for outputs.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CompileBuildVariables.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CompileBuildVariables.java
@@ -43,6 +43,8 @@ public enum CompileBuildVariables {
   UNFILTERED_COMPILE_FLAGS("unfiltered_compile_flags"),
   /** Variable for the path to the compilation output file. */
   OUTPUT_FILE("output_file"),
+  /** Variable for the path to the compilation output file directory. */
+  OUTPUT_DIRECTORY("output_directory"),
   /** Variable for the dependency file path */
   DEPENDENCY_FILE("dependency_file"),
   /** Variable for the module file name. */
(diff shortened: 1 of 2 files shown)


jin:
@oquenchil Have you had the chance to take a look at this PR?

oquenchil:
Hey George,

please help me understand a bit more about why this is needed and to see if there are no alternatives. I have a bunch of questions:

> This is useful for providing a directory to tools which explicitly demand a scratch location for outputs.

1. Can you give me a concrete example of this? 
2. What is a scratch location? A location where the tool can write to? Why do the object files have to be there?
3. How often does this use case appear? In your project or other people's project. Is it just this single tool that requires this? 
4. Would a symlink work?
5. How would you set this build variable? Would you set it in the crosstool and would apply for every compile action from every rule or would it be set just in a custom Starlark rule?

On a relative separate note, in the case that you have your custom Starlark rule and that's where you would set the build variable. How useful would it be to be able to add new variables straight from Starlark without having to touch Bazel Java code?

oquenchil:
I see, I understand now, thank you very much. 

Would it be possible to take the output_file path append something to it and pass that as the directory in the crosstool?
E.g.:
`flags = ["--keep-dir=%{output_file}_dir"],`

Or would it fail because the directory doesn't exist and it cannot create it? If it can't create it, why not?

werkt:
It does happen that bare nvcc fails because the directory doesn't exist - keep seems to imply to nvcc that the directory management should occur externally. As to why it cannot, aside from the expectations, it is quite simply a closed source tool that I have no (direct) influence over.

In general, this can be solved by script wrappers, and even without an additional argument (just find the -o equivalent and perform the transform on it, creating it if necessary), but having the output_directory be qualified by bazel, and having that be the only location that is written to in addition to having the directory created, seemed the more appropriate solution.

oquenchil:
Yes, I agree that script wrappers are ugly. But I generally prefer not adding anything at all to Java code in Bazel if it can be avoided.

Even though the change is quite small the cost is relatively high. 
- It will use more memory for every compilation action. We have had changes rolled back in the past that added a new build variable just because of the memory usage. 

- It would be one more variable for people to understand 

- We would need a test for this, so even more code

But this is only for one specific tool `nvcc`, so I think the cost is not justified.

oquenchil:
Hey George,

if a wrapper works I will close this PR.

werkt:
That's fine. I will make a wrapper work.

oquenchil:
Alright, thanks!

Issue #10442, opened 2019-12-18
Feature request: for Bazel to integrate gcs:// and s3:// transports


nkoroste:
Related issues:
https://github.com/bazelbuild/rules_jvm_external/issues/256
https://github.com/coursier/coursier/issues/1399

It would be nice to have the ability to use a maven repo from an internal GCS bucket. 

Currently https://github.com/bazelbuild/bazel/blob/d2a5b330fe3fe1a7f913a3ce2cf2969a872f7d75/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java
Doesn't support gcs://

gmendonca:
Maybe we could make this broader and also support `s3://`

Related:
https://github.com/coursier/coursier/issues/1488
https://github.com/bazelbuild/rules_scala/issues/911

jin:
Modified the FR to include `s3://`.

uri-canva:
Would the new remote downloader feature help here in the meantime? For example if urls in `repo_ctx.download` were passed to the remote downloader without being checked for support, it would be possible for the remote downloader to support the transports, even if the downloader built into bazel didn't (it's possible that's already the case, I haven't checked).

PR #10445, opened 2019-12-19
Extend http downloader to add support for non-standard authorization headers

chancila:
In order to leverage downloading from certain cloud storage providers that don't support basic authentication, this PR extends the existing authentication mechanism in bazel to forward an oauth2 token to the request header.

Sample netrc

```
machine storage.googleapis.com
        oauth2-token aaabbbbccccdddd
```

See https://github.com/bazelbuild/bazel/issues/10442

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
@@ -79,10 +79,8 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


chancila:
@googlebot I signed it!

chancila:
I realize that .netrc doesn't really have any mechanism for specifying an oauth token.

Perhaps setting some combination of the existing attributes with a keyword and the oauth2 token is a better approach.

I'm also open to other suggestions.

I verified it locally with a trivial sample but I can write a unit test once the maintainers are okay with some specific design.

Thank you!

nkoroste:
Thanks @chancila I verified that together with this patch and a fork of jvm_rules/coursier to support GCS urls I can successfully download artificats from our internal GCS buckets.

chancila:
Anyone from google got any comments? Thanks!

jin:
Thanks! Please add a test to src/test/shell/bazel/skylark_repository_test.sh.

chancila:
I can tweak it a bit to not break netrc configs...

I also saw that someone requested s3 support in the issue, this wouldn't fix that problem because from what I'm seeing amazon has their own custom http header for API access, basically means adding another one-off in the java code for s3 auth, would allowing a mechanism for forwarding http headers from starlark to the http client in java be better?

jin:
>  would allowing a mechanism for forwarding http headers from starlark to the http client in java be better?

I would prefer this. I'm hesistant on adding a new symbol to the netrc file that's not listed here (https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html).

chancila:
@jin I actually looked through the initial design docs and PRs for the netrc change and noticed the original implementer had some initial designs for this that made sense but was never merged when the feature was added. I basically took inspiration from that and added it here instead of forwarding http headers, which would be a much more intrusive change.

let me know if it makes sense and I can add unit tests to make sure all of this is covered.

chancila:
@jin any updates here? I managed to get this entire thing set up with internal forks of bazel/jvm_external/coursier to get gcs based artifact fetching working. Would be nice to contribute this stuff back so I don't have to maintain a fork here.

laurentlb:
@jin Any update on this?
Or should we look for another reviewer?

jin:
Yep, sorry for the delay, I'll be reviewing this in the coming week.

nkoroste:
@jin now that this is submitted what do you think is the best way to propagate it to jvm_rules/coursier?

PR #10455, opened 2019-12-19
Add OpenSUSE support

Hui-Zhi:
Add OpenSUSE support for installation.

### DIFF
--- a/site/docs/install-suse.md
+++ b/site/docs/install-suse.md
@@ -5,6 +5,9 @@ title: Installing Bazel on openSUSE
 
 # Installing Bazel on openSUSE Tumbleweed & Leap
 
+NOTE: The Bazel team does not maintain openSUSE support. For issues
+using bazel on openSUSE please file ticket at bugzilla.opensuse.org.
+
 The commands below must be run either via `sudo` or while logged in as `root`.
 
 ## Installing Bazel on openSUSE Tumbleweed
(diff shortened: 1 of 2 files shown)


philwo:
https://software.opensuse.org/package/bazel shows that the last version they package is Bazel 0.22.0. I don't think we should add official looking documentation to our own website for these packages unless they are better maintained, or alternatively we should make it very clear that we do not maintain these packages.

Hui-Zhi:
@philwo I will ask SUSE R&D team to update Bazel version.

Hui-Zhi:
@philwo SUSE R&D team has built bazel 2.0 here: https://build.opensuse.org/request/show/761595,
and it will take a few days to have it here: https://software.opensuse.org/package/bazel, also I will refine the steps for instruction later.

Hui-Zhi:
@philwo I have updated the installation guide, and the latest version has been built at https://build.opensuse.org/request/show/761595. It will be available at after an SUSE internal review: https://software.opensuse.org/package/bazel

Please review this.

jin:
Could you please add a note at the top of the instructions that the Bazel team does not maintain openSUSE support, and add a point of contact to the maintainers?

Hui-Zhi:
@jin I have added a note for this, but there's no need to add maintainers, users just need to file a ticket at bugzilla.opensuse.org then SUSE maintainers will take care of it.

philwo:
Thank you! I'll import this change.

PR #10463, opened 2019-12-19
Clarify precise platform/toolchain matching semantics.

gregestren:


gregestren:
I've had to look this up zillions of times to refresh my own memory before deciding to finally document what I keep on re-learning. :)

Issue #10476, opened 2019-12-22
bazel.sh attempts to exec directory {WORKSPACE_DIR}/tools/bazel


aaliddell:
### Description of the problem / feature request:

Using Bazel installed via apt, since 2.0.0 it attempts to exec `{WORKSPACE}/tools/bazel` if it exists, even as directory. This path is commonly used as a directory for storing Bazel related tools and the path is not obviously documented as being special.

The error seen is:

```
/usr/bin/bazel: line 196: <path to workspace snipped>/tools/bazel: Is a directory
/usr/bin/bazel: line 196: exec: <path to workspace snipped>/tools/bazel: cannot execute: Is a directory
```

The commit that introduced this was https://github.com/bazelbuild/bazel/commit/d3f8efca72a0e2013467c033df1bf58dd76e8a10 by @philwo 

In particular, these lines in bazel.sh:

https://github.com/bazelbuild/bazel/blob/d5ae460f1581ddf27514b4be18255481b47b4075/scripts/packages/bazel.sh#L193-L197

Prior to this commit, the wrapper was also checked if it was not a directory:

https://github.com/bazelbuild/bazel/blob/40f464c32f4b14d92689dfc1db4c00ecd5d58645/scripts/packages/bazel.sh#L79-L81

Proposed solution is to re-add the directory exclusion check, which I will do shortly in a PR.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create an standard workspace with a folder at `tools/bazel`. Using Bazel intalled via apt, run any bazel command (just `bazel` will do)

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

release 2.0.0

###  Have you found anything relevant by searching the web?

Previous similar issue: https://github.com/bazelbuild/bazel/issues/3422

philwo:
Sorry for the unintended breakage and thank you for catching and fixing it!

PR #10477, opened 2019-12-22
Check that tools/bazel is not a directory before attempting to exec

aaliddell:
Fixes https://github.com/bazelbuild/bazel/issues/10476

### DIFF
--- a/scripts/packages/bazel.sh
+++ b/scripts/packages/bazel.sh
@@ -191,7 +191,7 @@ if [[ ! -x $BAZEL_REAL ]]; then
 fi
 
 readonly wrapper="${workspace_dir}/tools/bazel"
-if [[ -x "$wrapper" ]]; then
+if [[ -x "$wrapper" && ! -d "$wrapper" ]]; then
   export BAZEL_REAL
   exec -a "$0" "${wrapper}" "$@"
 fi


philwo:
Thank you! I'll merge this.

scripts/packages/bazel.sh
```diff
@@ -191,7 +191,7 @@ if [[ ! -x $BAZEL_REAL ]]; then
 fi
 
 readonly wrapper="${workspace_dir}/tools/bazel"
-if [[ -x "$wrapper" ]]; then
+if [[ -x "$wrapper" && ! -d "$wrapper" ]]; then
```
philwo: Code review brought up the question of why not just use `-f` here instead of `! -d`:

```
if [[ -x "$wrapper" && -f "$wrapper" ]]; then
```

It looks fine to me, because symlinks are handled correctly that way, too ("If file is a symbolic link, test will fully dereference it and then evaluate the expression against the file referenced"), but wanted to check with you.

aaliddell: Honestly I just went with what it was before, so I didn’t really think about alternatives. If there’s a preferred change, I’m fine with that 👍

aaliddell: I’m also not able to make the changes to the pr currently, so feel free to edit it yourself if you want

PR #10482, opened 2019-12-23
DottedVersion.Option implements StarlarkValue

AlessandroPatti:
As of 2.0, it is not possible to transition on these values anymore, since `DottedVersion` does implement StarlarkValue, but `DottedVersion.Option` does not.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
@@ -25,6 +25,7 @@ import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
 import com.google.devtools.build.lib.skylarkbuildapi.apple.DottedVersionApi;
 import com.google.devtools.build.lib.syntax.Printer;
+import com.google.devtools.build.lib.syntax.StarlarkValue;
 import java.util.ArrayList;
 import java.util.Objects;
 import java.util.regex.Matcher;
@@ -89,7 +90,7 @@ public final class DottedVersion implements DottedVersionApi<DottedVersion> {
    * directory names)</p>
    * */
   @Immutable
-  public static final class Option {
+  public static final class Option implements StarlarkValue {
     private final DottedVersion version;
 
     private Option(DottedVersion version) {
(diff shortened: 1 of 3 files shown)


tetromino:
@jmmv - ping for review?

jmmv:
This seems fine, but what does this actually do? Could we have a test that shows what this offers?

AlessandroPatti:
> This seems fine, but what does this actually do? Could we have a test that shows what this offers?

This allows accessing `DottedVersion` and `DottedVersion.Option` in Starlark. Added a small test case.

tetromino:
@jmmv - ping for review

jmmv:
I *think* this is fine, but would appreciate @c-parsons review. I have never dealt with exposing stuff to Starlark so I don't know what the considerations for this are.

jmmv:
I actually just imported this change and will ask for @c-parsons review internally. Will come back if there are concerns.

aiuto:
Ping everyone.  Since this was approved once, can we dust it off to fix the build errors and try again. Or close it.

AlessandroPatti:
> How does one practically create / consume a DottedVersion.Option in Starlark?

The way this was intended to be used was in a transition. This change would allow to check the current version. E.g. `str(settings["//command_line_option:ios_minimum_os"])) == "1.0.0"`, which throws an error at the moment.
> I am wary of adding such a specific type to Starlark. It would make sense to use Strings for this purpose, and then do String->Option conversion natively.

I believe this is already supported. String values as transition output for `"//command_line_option:ios_minimum_os"` was correctly converted to option in my test.

Rebased the branch to test these changes again. However, we are not using this anymore, so feel free to close this if you think it would not bring much value or rather issues.

c-parsons:
* Prior to your change, does `settings["//command_line_option:ios_minimum_os"])` throw an error, or return an opaque value?
* With your change, what does settings["//command_line_option:ios_minimum_os"]) return? A string? A new StarlarkValue with a string representation?  (I would prefer if this simply returned a string as opposed to producing a new Starlark type at this time, especially if the current use case is simply to stringify)
* Can you add a test for accessing this value in a transition? Perhaps in src/test/java/com/google/devtools/build/lib/analysis/StarlarkRuleTransitionProviderTest.java

meisterT:
What's the status of this PR?

susinmotion:
Can we send this home? @jmmv

lberki:
I would come down on the side of not importing this change, since it would widen our public API for an unclear benefit. 

Let me know if you feel strongly otherwise!

PR #10493, opened 2019-12-27
proto_lang_toolchain: Add original sources of ProtoInfo to blacklistedProtos

### LINKED ISSUE #10484: proto_lang_toolchain and blacklisted_protos with ProtoInfo not propagated correctly

With protobuf's `proto_lang_toolchain`, after this change https://github.com/protocolbuffers/protobuf/commit/7b28278c7d4f4175e70aef2f89d304696eb85ae3 it seems the `blacklisted_protos` are no longer working correctly and removing that field has no effect on the build.

See https://github.com/protocolbuffers/protobuf/issues/7046 for a repro case

macOS

release 1.2.0 (I also tried with 2.0.0 with no change)

The protobuf commit linked above referenced this bazel change https://github.com/bazelbuild/bazel/commit/a5ee2c4d979441e663cf53b2ace4daa05e5033d8

I'm also wondering if the issue is with protobuf's use of `strip_import_prefix` on the `proto_library` targets they define: https://github.com/protocolbuffers/protobuf/commit/a03d332aca5d33c5d4b2cd25037c9e37d57eff02

Yannic:
Fixes #10484


rebello95:
Thanks for opening this, @Yannic!

I was able to confirm that cleaning/building Envoy Mobile with this patch applied to the bazel binary resolves the issue we're seeing in https://github.com/lyft/envoy-mobile/issues/617 🎉

keith:
@Yannic any change this could get merged soon so we can make the next cut?

gnossen:
@Yannic Bump. Great change! This PR resolves an issue that's been blocking gRPC from [upgrading its version of protobuf](https://github.com/grpc/grpc/pull/21590).

Yannic:
@cushon landed a similar change (https://github.com/bazelbuild/bazel/commit/01df1e5c3aaf709e9388293514ae978e59ae2cbb), but the issue was the difference between direct sources and original direct sources, so I don‘t think the issue is resolved by that commit.

This issue is blocking users from upgrading past Protobuf 3.9(?), so it would be great if we could get it merged soon.

/cc @laurentlb

lberki:
I was about to lament that `#getOriginalDirectProtoSources()` is arguably a hack and demand that you at least add a test case but then I realized that:

1. You actually have a test case
2. It was *me* who did not add a test case in https://github.com/bazelbuild/bazel/commit/46c95dcbe2889e515a2b44b9dec00d6c13f7e9fa
3 (I don't know what I was thinking)
3. It might actually be better to check for the direct source file since (at least in theory) the same source file might give rise to multiple virtual ones if multiple `proto_library` rules reference it (I haven't thought it through as to whether that's a realistic risk)

Let me try to run this through our internal test battery.

lberki:
...aaand, failures :(

Yannic:
I agree that this is a hack, but `blacklisted_protos = ["a.proto", "b.proto"]` will always reference the original file, so we have to filter by original files for now. We probably can (and should!) get rid of `getOriginalDirectProtoSources()` (& friends) when `blacklisted_protos` requires the `ProtoInfo` provider.

Let me know if there's anything I can do to fix the failing tests.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
```diff
@@ -44,8 +44,12 @@ public ConfiguredTarget create(RuleContext ruleContext)
       ProtoInfo protoInfo = protos.get(ProtoInfo.PROVIDER);
       // TODO(cushon): it would be nice to make this mandatory and stop adding files to build too
       if (protoInfo != null) {
-        blacklistedProtos.addTransitive(protoInfo.getTransitiveProtoSources());
+        // TODO(yannic): Switch to getTransitiveProtoSources() and remove references to
+        // original proto sources when ProtoInfo is mandatory.
+        blacklistedProtos.addAll(protoInfo.getOriginalDirectProtoSources());
```
cushon: Using direct instead of transitive sources un-does a change made in 01df1e5c3aaf709e9388293514ae978e59ae2cbb that I think is going to break some uses of `proto_lang_toolchain.blacklisted_protos`.

lberki: Yep, this is exactly what I found out after debugging the aforementioned test failures. Adding _both_ here isn't very nice so I'll try adding the set of transitive blacklisted sources (I'd totally ask Yannic to do so, but since the tests that break are internal, he'd be flying blind...)

lberki: ...and it blends! I'm sending this out (+my changes) for internal review.

meteorcloudy:
This PR is causing some breakage in the downstream:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1349

Culprit finder result on Envoy:
https://buildkite.com/bazel/culprit-finder/builds/248#ca25feb7-a89b-43a4-ac25-4c8522c0e197

Yannic:
Suspecting that this shouldn't have been changed to transitive original sources: https://github.com/bazelbuild/bazel/commit/4162cc54ba0b128b616c0bd05b65bf9ad5e66f2e#diff-42ff97786b96eb7e34507430633100d5L279

This will fail for all `proto_library` targets that have at least one blacklisted source as a transitive dependency.

meteorcloudy:
@Yannic Can you send a fix for this?

meteorcloudy:
@Yannic Sorry, I'll rollback this change to get our CI green again.

PR #10513, opened 2020-01-02
Update error for running non-executable targets

keith:
For custom rules this must be set and it wasn't previously obvious from
the error that this was the issue.


keith:
Updated the comment. It would be great if those reviews could happen publicly instead!

jin:
@lberki, does this look better?

lberki:
Yep. There is no need to guess whether the rule is a Starlark rule because you have a `Target` available, `Rule` is-a `Target` and you can use `rule.getRuleClassObject().isSkylark()` to learn whether it is.

jin:
@keith could you try using @lberki's `isSkylark()` approach to tailor the error message?

aiuto:
ping

Issue #3481, opened 2017-07-31
some java tests are never run


benjaminp:
Not every java test file is included in a test target, which presumably means such files are never run.

```
$ ~/bin/bazel query -k 'filter("\.*\.java$", deps(filter("srcs", //src/test/java/...)) - deps(tests(//src/test/java/...))'
//src/test/java/com/google/devtools/build/lib:worker/WorkerPoolConfigTest.java
//src/test/java/com/google/devtools/build/lib:worker/ExampleWorkerOptions.java
//src/test/java/com/google/devtools/build/lib:worker/ExampleWorker.java
//src/test/java/com/google/devtools/build/lib:worker/ErrorMessageTest.java
//src/test/java/com/google/devtools/build/lib:util/ResourceFileLoaderTest.java
//src/test/java/com/google/devtools/build/lib:rules/test/TestTargetPropertiesTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ValueTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/TargetDeviceFamilyTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/RuleType.java
//src/test/java/com/google/devtools/build/lib:rules/objc/OnlyNeedsSourcesRuleType.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcSkylarkTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcRuleTestCase.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcProviderTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcProtoProviderTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcProtoLibraryTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcProtoAspectTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcLibraryTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcImportTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcFrameworkTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcCommonTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcBundleTest.java
//src/test/java/com/google/devtools/build/lib:rules/objc/ObjcBundleLibraryTest.java
[...]

PR #10537, opened 2020-01-08
"Stream" action cache digest computations.

benjaminp:
Now that ActionAnalysisMetadata.getInputs() returns a NestedSet, we don't need allocate an intermediate (path -> metadata) map during action digest computation. Encapsulate the logic for this in a OrderIndependentHasher class. This class also caches fingerprints and digest arrays to avoid allocations.

Update the size constant in CompactPersistentActionCacheTest.testToStringIsntToBig because the digests are all 32 byte arrays instead of 1 byte arrays.

Additionally, fix ActionCache.Entry so it's stateless and immutable by making a builder class for it. Delete CompactPersistentActionCacheTest.testEntryToStringIsIdempotent because the property it wants to proven by the type system.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/cache/DigestUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/cache/DigestUtils.java
@@ -19,7 +19,6 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheStats;
 import com.google.common.primitives.Longs;
-import com.google.devtools.build.lib.actions.FileArtifactValue;
 import com.google.devtools.build.lib.clock.BlazeClock;
 import com.google.devtools.build.lib.profiler.Profiler;
 import com.google.devtools.build.lib.profiler.ProfilerTask;
@@ -51,7 +50,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * <p>Note that the (path, metadata) tuples must be unique, otherwise the XOR-based approach will
  * fail.
  */
-public class DigestUtils {
+public final class DigestUtils {
 
   // Object to synchronize on when serializing large file reads.
   private static final Object DIGEST_LOCK = new Object();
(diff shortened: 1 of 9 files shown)


benjaminp:
@ulfjack, as discussed at BazelCon...

ulfjack:
This breaks CompactPersistentActionCacheTest. Can you take a look?

Otherwise looks good.

benjaminp:
> This breaks CompactPersistentActionCacheTest. Can you take a look?

This is now fixed.

It looks like #3481 is still biting.

> 
> Otherwise looks good.

Thanks for the review.

src/main/java/com/google/devtools/build/lib/actions/cache/DigestUtils.java
```diff
@@ -331,4 +302,35 @@ public static void write(byte[] digest, OutputStream sink) throws IOException {
     }
     return xor(rhs, lhs);
   }
+
+  /** Utility class for combining a set of (path, metadata) pairs into a single digest. */
+  public static final class OrderIndependentHasher {
```
ericfelly: why not add this as a top-level class?

benjaminp: I just put it here because it replaces functionality that was in `DigestUtils`. (Also, it shares the private `xor()` method with the outer class, but we could just increase the visibility of that.)

I don't have a strong opinion about its location.

benjaminp: I just promoted it to it's own file.

meisterT:
Benjamin, I am sorry, we had to roll this back today since it broke correctness on incremental builds.

If I understand it correctly, on an incremental build it sometimes happens that there is an incorrect action cache hit.

I'll try to construct a repro that works with Bazel.

benjaminp:
Oops, sorry for the bug!

Shot in the dark: Is it something involving shared actions producing distinct artifacts with the same exec path?

meisterT:
Yes, it seems to be exactly that. I tried to add a CompactHashSet in the hasher but the performance win  is gone :-/

benjaminp:
I suppose we could factor the action owner into the artifact's digest. But I'm not sure there's a middle ground that distinguishes actions with the same execpath but doesn't introduce too much rigidity into the digest. We probably don't want to include the configuration hash in derived artifact digests.

PR #10545, opened 2020-01-09
Add attr for recursively setting up git submodules

keith:
If a repo has multiple levels of submodules this hydrates those.

### DIFF
--- a/src/test/shell/bazel/starlark_git_repository_test.sh
+++ b/src/test/shell/bazel/starlark_git_repository_test.sh
@@ -371,6 +371,64 @@ EOF
   expect_log "Pluto is a planet"
 }
 
+function test_new_git_repository_submodules() {
+  local outer_planets_repo_dir=$TEST_TMPDIR/repos/outer-planets
+
+  # Create a workspace that clones the outer_planets repository.
+  cd $WORKSPACE_DIR
+  cat >> $(create_workspace_with_default_repos WORKSPACE) <<EOF
+load('@bazel_tools//tools/build_defs/repo:git.bzl', 'new_git_repository')
+new_git_repository(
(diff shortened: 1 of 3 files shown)


keith:
If we're not ok with this being a breaking change maybe we can provide another attribute like `recursively_init_submodules` for this?

jin:
> If we're not ok with this being a breaking change maybe we can provide another attribute like recursively_init_submodules for this?

Yes, I'd prefer that this is an attribute on `git_repository`.

keith:
Updated with that approach!

jin:
Could you add a test for this in src/test/shell/bazel/skylark_git_repository_test.sh?

keith:
I added a test that uses the existing fixtures, so it doesn't actually verify the recursion, just that any are setup, I'm happy to add a new fixture for that if I should?

jin:
Sorry for the delay, yes, we should have a fixture that sets up and verifies the `--recursive` flag.

jin:
Please add a test with a recursive fixture, and I'll import this. Thanks!

keith:
I spent a while looking into this and it's pretty unclear to me how these fixtures are supposed to work. I'll try to look into it again at some point.

keith:
The test repo stuff here is a little too hard to follow, I can't figure out how it's supposed to work. Can we merge this knowing it doesn't break any existing behavior?

jin:
Hmm, could we at least test that the attribute triggers the `--recursive` argument to show up in the repository rule executed command?

keith:
Is there a starlark test somewhere that has access to that info? All I see is the shell tests where it doesn't seem like I could access the command

keith:
Thanks!

PR #10547, opened 2020-01-09
Put HTTP portions of external downloads behind an interface.

jmillikin-stripe:
This PR is in preparation for a remote implementation of external downloads. It is intended to have no behavior changes, and to split the current `HttpDownloader` logic into:
* `HttpDownloader`, which does the actual download with all the fancy concurrency and error handling.
* `DownloadManager`, which interacts with `RepositoryCache` and the distdirs. It doesn't know about HTTP any more, instead going through a `Downloader` interface.

A future PR will implement a `Downloader` in terms of the bazelbuild/remote-apis protocol and put it behind an `--experimental_remote_downloader` flag.

R: @buchgr @dslomov 
CC: @sstriker

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Downloader.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Downloader.java
@@ -1,0 +1,49 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


buchgr:
LGTM. Thanks!

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Downloader.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param checksum valid checksum which is checked, or absent to disable
+   * @param output path to the destination file to write
+   * @throws IOException if download was attempted and ended up failing
+   * @throws InterruptedException if this thread is being cast into oblivion
+   */
+  void download(
+      List<URL> urls,
+      Map<URI, Map<String, String>> authHeaders,
+      Optional<Checksum> checksum,
+      String canonicalId,
```
aehlig: If I understand this change correctly, the downloader should manage the actual download but not care about the cache. So why do we keep the `canonicalId` part of the interface? (The only implementation of this interface does not use it.)

buchgr: Good catch. I fixed this in the internal review.

jmillikin-stripe: The gRPC remote downloader client will need to pass this canonical ID through to the backend.

PR #10563, opened 2020-01-10
Add gRPC's cc_grpc_library to //third_party/grpc/bazel

anakanemison:
This adds a simplified version of gRPC's cc_grpc_library to
//third_party/grpc/bazel. I intend to replace the hacky cc_grpc_library
implemented in //tools/build_rules:genproto.bzl and used by
//src/main/protobuf:command_server_cc_proto with this one.

Aside from a general cleanup, this version of the macro supports proto
files/rules with dependencies. Future work with fine-grained failure
reporting will add dependencies to command_server.proto.

The added files, all BUILD/bzl, are taken from gRPC's v1.26.0
(commithash 'de893acb', see https://github.com/grpc/grpc/tree/v1.26.0).
Because these resources have no dependencies on the rest of gRPC, there
are no conflicts with the current imported version of gRPC's C++
sources.

This includes a patch file, bazel.patch, which adapts gRPC's BUILD/bzl
sources for use by Bazel. The cc_grpc_library macro is significantly
simplified, the deprecated well_known_proto-related functionality is
removed, and load statements are relativized. A "srcs" filegroup is
added to the package's BUILD file and included by
//third_party/grpc:srcs.

This also deletes the obsolete grpc.patch file.


benjaminp:
I'm curious why it's desirable to keep backporting patches (e.g., also 302acaa669236356df0af4f4be41bac357569439) to the vendored gRPC copy rather than simply updating it.

anakanemison:
I'd be happy to see Bazel's imported gRPC C++ sources brought up to v1.26.0. I'm not aware of anything blocking that effort besides someone taking a close look at the changes, figuring out whether anything needs adapting, and testing the result.

This PR isn't a backport. It's adding some macros and Starlark rules from gRPC which weren't previously imported in any form.

benjaminp:
> This PR isn't a backport. It's adding some macros and Starlark rules from gRPC which weren't previously imported in any form.

Ah. I see I misparsed the commit message. You've copied in and simplified files from the upstream gRPC releases. Still, it seems this PR amounts to a fork of these files. If gRPC was updated, could vanilla `cc_grpc_library` be used?

anakanemison:
Hey @benjaminp, sorry for the delay in responding--I was out on vacation.

I think there would be a way for gRPC's own `cc_grpc_library` to evolve so that it could be used, without modification, by Bazel and other projects. For example, it would be helpful if the macro's handling of kwargs was like it is in this version because the original only supported attributes common to all rules.

I'd guess that their use of external targets (via `//external` and vanilla external repos) is generally compatible with other projects but I'm not sure. Maybe it would work with Bazel too, though Bazel's `third_party` vendoring strategy and bootstrapping requirement makes me uncertain about it. I'm not an expert in those areas.

jin:
@anakanemison who would the best reviewer for this change? I can help to import this into //third_party once it's approved.

anakanemison:
@jin Based on the history of who has made updates to third_party/grpc, maybe @buchgr or @tetromino? 

A couple of months ago, when I was first considering making this change, I chatted with @tetromino about it. There's a small chance he retains some context about it :)

third_party/grpc/bazel.patch
```diff
@@ -0,0 +1,288 @@
+From 6eb4a396945bbe89831ce1d205106c1348f376e3 Mon Sep 17 00:00:00 2001
+From: Mark Schaller <mschaller@google.com>
+Date: Thu, 9 Jan 2020 19:50:57 -0500
+Subject: [PATCH] Adapt gRPC bazel files for use in Bazel
+
+Simplifies cc_grpc_library. Removes well_known_proto-related
```
tetromino: The only change I would suggest is in the diff's commit message, which imho should be more detailed to make manual rebasing easier for future maintainers.

Something like this:

Simplifies cc_grpc_library:
* removes grpc_only arg (forced to True);
* removes proto_only, well_known_protos (forced to False);
* adds extra_cc_library_kwargs

Removes well_known_proto-related functionality throughout.
Renames generate_cc's _protoc attr to protoc. Relativizes
load statements. Adds a filegroup, "srcs", to the package.

anakanemison: Thanks, good idea! I've changed the 2nd-to-last paragraph of the commit message to this:

```
This includes a patch file, bazel.patch, which adapts gRPC's BUILD/bzl
sources for use by Bazel.

This simplifies the cc_grpc_library macro and its deps:
* removes cc_grpc_library's grpc_only parameter (now always true)
* removes its proto_only, well_known_protos parameters (now always
    false)
* removes its use_external parameter (it was previously unused)
* adds to it an extra_cc_library_kwargs parameter (forwarded to its
    generated cc_library rule)
* removes "well_known_proto" functionality from generate_cc.bzl and
    protobuf.bzl
* changes generate_cc's _protoc attribute to protoc
* sets cc_grpc_library's generate_cc call's plugin and protoc arguments
    to Bazel's targets ("//third_party/grpc:cpp_plugin" and
    "//third_party/protobuf:protoc", respectively)
* relativizes load statements throughout
* adds "srcs" filegroup to the package's BUILD file and includes it in
    //third_party/grpc:srcs

```

tetromino: I'm sorry for the unclear wording. I meant adding more detail to the message *in the new `bazel.patch` file* - because that's what someone will be looking at 1 year from now when they run `patch -p3 < bazel.patch` and get conflicts.

anakanemison: Oh, I see! No prob--done.

tetromino:
Assigning to @laszlocsomor (or next green team sheriff) to push the third_party/ change.

laszlocsomor:
Schedule got messed up; actual sheriff is @c-parsons.

c-parsons:
This was pushed as https://github.com/bazelbuild/bazel/commit/7714c808cdf96de572c2067f062d95e8a68ca67c

Issue #10132, opened 2019-10-30
Release 2.1 - January 2020 (stable)


dslomov:
Cut date: 2020-01-07

c16a:
Hope 934c4fd4a4a96093a2cd0b08cf619bef04d0b582 is included in this change?

laurentlb:
# Status of Bazel 2.1

- Target baseline: 2020-01-08
- Expected release date: January 15th-20th
- [List of release blockers](https://github.com/bazelbuild/bazel/labels/Release%20blocker)

To report a release-blocking bug, please file a bug using the `Release blocker` label, and cc me.

Task list:

- [ ] Update GitHub issues for incompatible changes
- [ ] Pick release baseline:
- [ ] Create release candidate:
- [ ] Check downstream projects:
- [ ] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [ ] Send for review the release announcement PR:
- [ ] Push the release, notify package maintainers:
- [ ] Update the documentation
- [ ] Push the blog post
- [ ] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

laurentlb:
All changes submitted before today should be included.
I'll update the previous status message later today.

Yannic:
Could we cherry-pick https://github.com/bazelbuild/bazel/commit/4162cc54ba0b128b616c0bd05b65bf9ad5e66f2e into Bazel 2.1? (Context: https://github.com/bazelbuild/bazel/issues/10484)

werkt:
Requesting cherry-pick of 9a823d9dab82631ff4600eff38d597807e8d1221 into Bazel 2.1 (Context: #10566), addresses a NPE resulting in build failures.

meteorcloudy:
Please cherry-pick 7e4c9d5 to fix https://github.com/bazelbuild/bazel/issues/10572

Helcaraxan:
For those not following the GitHub issue in question but #10356, which was [re-opened yesterday](https://github.com/bazelbuild/bazel/issues/10356#issuecomment-576386798), might be considered a release-blocker for 2.1.0.

keith:
Any plans for when there will be another RC or the official 2.1.0 release?

laurentlb:
I'll make another RC tomorrow.

laurentlb:
[...]

PR #7846, opened 2019-03-26
Progressively retrying uploads to bytestream

werkt:
Attempt to continue uploads through retriable exceptions, using the
ProgressiveBackoff to reset as progress is made. This adds a write
status query after each retriable exception to determine whether
progress has been made. A service that does not support write
resumption, either with persistent write resets or failure status
responses, the behavior is identical to the non-progressive
implementation.

The ProgressiveBackoff was also added to the read side for executeAsync
invocations to ensure that it was used instead of the retrier supplier,
and the injected backoff executeAsync Retrier method was exposed.

PR #10566, opened 2020-01-12
Prevent NPE on backwards seek on Chunker

werkt:
Ensure that a reset() is not immediately followed by a data member
access with a call to maybeInitialize(). Practically, this can occur if
an offset has progressed locally, but the remote indicates a non-0
committed size, and would always have thrown an NPE on a requisite seek,
proven with the accompanied test case.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
@@ -142,14 +142,11 @@ public final class Chunker {
    * <p>Closes any open resources (file handles, ...).
    */
   public void seek(long toOffset) throws IOException {
-    boolean resetAndSkip = toOffset < offset;
     if (toOffset < offset) {
       reset();
     }
-    if (toOffset != 0 && toOffset != offset) {
-      maybeInitialize();
-      data.skip(toOffset - offset);
-    }
+    maybeInitialize();
+    data.skip(toOffset - offset);
     offset = toOffset;
   }
 
--- a/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
@@ -142,13 +142,12 @@ public final class Chunker {
    * <p>Closes any open resources (file handles, ...).
    */
   public void seek(long toOffset) throws IOException {
-    maybeInitialize();
+    boolean resetAndSkip = toOffset < offset;
     if (toOffset < offset) {
       reset();
-      if (toOffset != 0) {
-        data.skip(toOffset);
-      }
-    } else if (offset != toOffset) {
+    }
+    if (toOffset != 0 && toOffset != offset) {
+      maybeInitialize();
       data.skip(toOffset - offset);
     }
     offset = toOffset;
--- a/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
@@ -139,14 +139,14 @@ public final class Chunker {
   /**
    * Seek to an offset, if necessary resetting or initializing
    *
-   * <p>Closes any open resources (file handles, ...).
+   * <p>May close open resources in order to seek to an earlier offset.
    */
   public void seek(long toOffset) throws IOException {
     if (toOffset < offset) {
       reset();
     }
     maybeInitialize();
-    data.skip(toOffset - offset);
+    ByteStreams.skipFully(data, toOffset - offset);
     offset = toOffset;
   }
 
--- a/src/test/java/com/google/devtools/build/lib/remote/ChunkerTest.java
+++ b/src/test/java/com/google/devtools/build/lib/remote/ChunkerTest.java
@@ -159,6 +159,20 @@ public class ChunkerTest {
     assertThat(next.getData()).hasSize(8);
   }
 
+  @Test
+  public void seekBackwards() throws IOException {
+    byte[] data = new byte[10];
+    Chunker chunker = Chunker.builder().setInput(data).setChunkSize(10).build();
+
+    chunker.seek(4);
+    chunker.seek(2);
+
+    Chunk next = chunker.next();
+    assertThat(next).isNotNull();
+    assertThat(next.getOffset()).isEqualTo(2);
+    assertThat(next.getData()).hasSize(8);
+  }
+
   private void assertNextEquals(Chunker chunker, byte... data) throws IOException {
     assertThat(chunker.hasNext()).isTrue();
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -142,13 +142,12 @@ public void reset() throws IOException {
    * <p>Closes any open resources (file handles, ...).
    */
   public void seek(long toOffset) throws IOException {
-    maybeInitialize();
+    boolean resetAndSkip = toOffset < offset;
```
buchgr: that's unused?

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -142,13 +142,12 @@ public void reset() throws IOException {
...
     if (toOffset < offset) {
       reset();
-      if (toOffset != 0) {
-        data.skip(toOffset);
```
buchgr: if you added a maybeInitialize() above this line above data.skip() wouldn't that be enough?

werkt: Yeah, there's probably too much cleverness going on in this already - I was avoiding the initialize and skip if it was a noop or could be left in the reset condition.

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -142,15 +142,11 @@ public void reset() throws IOException {
...
-      }
-    } else if (offset != toOffset) {
-      data.skip(toOffset - offset);
     }
+    maybeInitialize();
+    data.skip(toOffset - offset);
```
buchgr: While you are at it: I just noticed another bug with this code. We ignore the return value of InputStream#skip. According to its documentation it's free to skip less bytes than specified. Similar to read.

Do you want to fix this here or should we do it in a different change?

werkt: I'll do it here

werkt: There are crazy directions to go in with the response from the skip here: more than expected, 0 with a possible non-eof. I'll try to accommodate them all here, and throw if we get into a bad state. lmk if it's too much

werkt: Nevermind, I'll just switch to using ByteStreams.skipFully, it accomplishes all that I wanted it to.

buchgr: Today I learned. Beautiful. Thanks!

buchgr:
LGTM. I ll import.

laurentlb:
Is it a recent regression? (did this work with a recent version of Bazel)

werkt:
@laurentlb No, this has been present since 0.26.0, introduced with https://github.com/bazelbuild/bazel/pull/7846

laurentlb:
I don't think it qualifies for a cherry-pick, but it will definitely be part of Bazel 2.2 in February.

werkt:
Just an update here, I'm pretty sure this is a regression in the transition to 2.0. It may simply be emergent, that Jacobs refactoring made everything faster and induced it, but we have only started seeing it with clients using 2, and not 1

jlisee:
> I don't think it qualifies for a cherry-pick, but it will definitely be part of Bazel 2.2 in February.

I represent the bazel users at Uber ATG that leverage @werkt Buildfarm project.  We hit the NPEs often enough that we had to delay our roll out of Bazel 2 by what will eventually be 1.5 to 2 months.

This is feels like an ideal candidate for a cherry pick and patch release, because:
 - It delays users upgrading
 - Is a non-breaking change
 - Fixes an underlying stability issues
 - Comes with an accompanying unit test

If you have another patch release of schedule before 2.2 it would be great to see this included in that release.

buchgr:
I can see that the bug is severe enough that a patch release could be warranted. I am surprised there aren't more bug reports about this.

laurentlb:
Thanks for the feedback, I'll take a look at it.

cc @philwo

laurentlb:
@jlisee, @werkt 
See Bazel 2.1.1rc: https://github.com/bazelbuild/bazel/issues/10132#issuecomment-585747185

I'd interested to hear what we can do to help you roll out new versions of Bazel. I think it would also be beneficial for everyone if you had a simple way to try new versions of Bazel (or even build it at head every day) and report issues more quickly.

We can easily mark bugs as "release critical" when they've just been introduced. Older bugs are usually not blocking.

werkt:
tl;dr: @laurentlb we can't keep up with you, yet.

> I'd interested to hear what we can do to help you roll out new versions of Bazel. I think it would also be beneficial for everyone if you had a simple way to try new versions of Bazel (or even build it at head every day) and report issues more quickly.

This is an extremely difficult to address problem for us. The cadence of bazel is intense, and it is not completely a goal for us to stay up to date with the latest bleeding edge commits there, for a variety of reasons. Even for releases, *every single minor version* since 0.4, including the post-1 minor releases, has required changes to our code or build definitions. Upgrades are major events, not the least of which because they are the most likely places for all remote action keys to change their composition, resulting in novel cache keying.

We use `bazilisk --migrate` internally, but the inconsistency of this with regard to the duration that incompatible flags remain in bazel to the release cadence means that we see red herring incompatibles (that we can only verify are red herrings through release candidate evaluation or commit diving) that don't prevent us from moving to the next version of bazel, but complicate our evaluation of the results.

And even with feature deprecation-based migration, many releases simply come with a newly broken (hyrum-esque) feature, that we find out when we can finally evaluate the release. Non-json profiling has been broken for some time. This bug, caused still by an unknown coincidence of behaviors in the client, broke for our particular (standard-compliant) remote implementation. Performance degradation in a language-specific command generation section of the composition will have caused us to reevaluate overall build timeouts.

We are trying to keep up and attempting to put in place ways to get the latest bazel tested, but sometimes reporting new bugs in a timely fashion loses out over other priorities, and a release slips us for a couple of months. And then we're a major release behind.

This is blatantly selfish, or at least shortsighted, but the best way for us to roll out new versions of bazel is to have a lower frequency of new versions of bazel. Better still, fewer versions with any *known* incompatibilities.
> 
> We can easily mark bugs as "release critical" when they've just been introduced. Older bugs are usually not blocking.

Similar to the above, I don't think we're watching every bug that comes in, mostly due to the onerous filtering that requires beyond just being 'release critical' - we build on a specific OS, with a specific workflow, with a specific feature set, with specific strategies, and with very specific languages and tooling. The firehose is just too much.

PR #10577, opened 2020-01-14
Fixes #10176 Updating struct.to_json to escape the key string.

### LINKED ISSUE #10176: struct.to_json doesn't escape keys

struct.to_json doesn't escape keys.

```python
print(struct(**{ "\"": 1 }).to_json()) # outputs {""":1}
```

macOS

release 0.27.0

N/A

https://github.com/bazelbuild/bazel/blob/a3421e204f22e478382bb8487b96df9dcd8e463a/src/main/java/com/google/devtools/build/lib/packages/StructImpl.java#L300

Keys are not escaped.

dwhu:
Please let me know where to place an appopriate test. No existing test
class appeared appropriate as far as I could tell.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/BUILD
@@ -466,6 +466,7 @@ java_library(
         "//third_party:guava",
         "//third_party:jsr305",
         "//third_party/protobuf:protobuf_java",
+        "//third_party:apache_commons_lang"
     ],
 )
 
--- a/src/main/java/com/google/devtools/build/lib/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/BUILD
@@ -466,7 +466,6 @@ java_library(
         "//third_party:guava",
         "//third_party:jsr305",
         "//third_party/protobuf:protobuf_java",
-        "//third_party:apache_commons_lang"
     ],
 )
 
--- a/src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
@@ -32,7 +32,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
-import org.apache.commons.lang.StringEscapeUtils;
 
 /**
  * An abstract base class for Starlark values that have fields, have to_json and to_proto methods,
@@ -271,9 +270,7 @@ public abstract class StructImpl implements Info, ClassObject, StructApi {
       for (String field : ((ClassObject) value).getFieldNames()) {
         sb.append(join);
         join = ",";
-        sb.append("\"");
-        sb.append(field);
-        sb.append("\":");
+        jsonEscapeString(sb, field);
         printJson(((ClassObject) value).getValue(field), sb, "struct field", field);
       }
       sb.append("}");
@@ -290,9 +287,7 @@ public abstract class StructImpl implements Info, ClassObject, StructApi {
               container,
               key != null ? " '" + key + "'" : "");
         }
-        sb.append("\"");
-        sb.append(StringEscapeUtils.escapeJava(String.valueOf(entry.getKey())));
-        sb.append("\":");
+        jsonEscapeString(sb, (String) entry.getKey());
         printJson(entry.getValue(), sb, "dict value", String.valueOf(entry.getKey()));
       }
       sb.append("}");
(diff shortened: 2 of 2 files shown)


aiuto:
@alandonovan  You've touched this code most in the last year. Can you take a look.

src/main/java/com/google/devtools/build/lib/BUILD
```diff
@@ -513,6 +513,7 @@ java_library(
         "//third_party:guava",
         "//third_party:jsr305",
         "//third_party/protobuf:protobuf_java",
+        "//third_party:apache_commons_lang"
```
alandonovan: This is a new dependency, and it's totally unnecessary. There is already code in StructImpl.java to escape Java strings into JSON notation (see L319); please re-use it.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
```diff
@@ -299,7 +300,7 @@ private void printJson(Object value, StringBuilder sb, Location loc, String cont
           throw new EvalException(loc, errorMessage);
         }
         sb.append("\"");
-        sb.append(entry.getKey());
+        sb.append(StringEscapeUtils.escapeJava(String.valueOf(entry.getKey())));
```
alandonovan: Don't use String.valueOf(x) here; it's equivalent to x.toString() except when x is null, which can't happen because this statement is dominated by the check on L291 (and even if it could, I would prefer to see a crash in that case than to silently emit incorrect JSON).

Use: (String) entry.getKey()

alandonovan: Could you also apply escaping to the field names on L280? Field names should be valid identifiers, but this is not guaranteed, and we don't want to emit corrupt JSON in that case. 

At that point, there are three places that emit quote escape(s) quote; could you refactor jsonEscapeString to accept a StringBuilder and emit the quotes too? Thanks.

src/main/java/com/google/devtools/build/lib/BUILD
dwhu: Sounds good.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
dwhu: Done.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
```diff
@@ -333,10 +327,16 @@ private void printJson(Object value, StringBuilder sb, Location loc, String cont
...
 
-  private String jsonEscapeString(String string) {
-    return escapeDoubleQuotesAndBackslashesAndNewlines(string)
+  private void jsonEscapeString(StringBuilder stringBuilder, String string, boolean appendColon) {
+    stringBuilder.append("\"");
+    stringBuilder.append(escapeDoubleQuotesAndBackslashesAndNewlines(string)
         .replace("\r", "\\r")
-        .replace("\t", "\\t");
+        .replace("\t", "\\t"));
+    if (appendColon) {
```
alandonovan: The appendColon logic has no business being in here, as it can be composed separately an no extra cost, resulting in a simpler function.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
```diff
@@ -333,10 +327,16 @@ private void printJson(Object value, StringBuilder sb, Location loc, String cont
     }
   }
 
-  private String jsonEscapeString(String string) {
-    return escapeDoubleQuotesAndBackslashesAndNewlines(string)
+  private void jsonEscapeString(StringBuilder stringBuilder, String string, boolean appendColon) {
+    stringBuilder.append("\"");
```
alandonovan: If you like, use append('"') to avoid the internal loop.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
```diff
@@ -333,10 +327,16 @@ private void printJson(Object value, StringBuilder sb, Location loc, String cont
     }
   }
 
-  private String jsonEscapeString(String string) {
-    return escapeDoubleQuotesAndBackslashesAndNewlines(string)
+  private void jsonEscapeString(StringBuilder stringBuilder, String string, boolean appendColon) {
```
alandonovan: Let's call this function appendJSONStringLiteral, the StringBuilder 'out', and the String s.

dwhu: Sounds good.

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
dwhu: Not sure I completely understand your suggestion. Was my change what you had in mind?

src/main/java/com/google/devtools/build/lib/packages/StructImpl.java
dwhu: Done.

aiuto:
Can someone please resolve the merge conflicts so we can merge this.
@alandonovan @dwhu

PR #10579, opened 2020-01-14
Note default cache directory root on macOS

gibfahn:
On Darwin systems it looks like the cache is written to this location.

### DIFF
--- a/site/docs/output_directories.md
+++ b/site/docs/output_directories.md
@@ -26,9 +26,12 @@ The solution that's currently implemented:
 
 * Bazel must be invoked from a directory containing a WORKSPACE file. It reports
   an error if it is not. We call this the _workspace directory_.
-* The _outputRoot_ directory is ~/.cache/bazel. (Unless `$TEST_TMPDIR` is
-  set, as in a test of bazel itself, in which case this directory is used
-  instead.)
+* The _outputRoot_ directory defaults to `~/.cache/bazel` on Linux,
+  `/private/var/tmp` on macOS, and on Windows it defaults to `%HOME%` if set,
+  else `%USERPROFILE%` if set, else the result of calling
+  `SHGetKnownFolderPath()` with the `FOLDERID_Profile` flag set. If the
+  environment variable `$TEST_TMPDIR` is set, as in a test of bazel itself,
+  then that value overrides the default.
 * We stick the Bazel user's build state beneath `outputRoot/_bazel_$USER`. This
   is called the _outputUserRoot_ directory.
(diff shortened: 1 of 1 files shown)


site/docs/output_directories.md
```diff
@@ -26,9 +26,9 @@ The solution that's currently implemented:
 
 * Bazel must be invoked from a directory containing a WORKSPACE file. It reports
   an error if it is not. We call this the _workspace directory_.
-* The _outputRoot_ directory is ~/.cache/bazel. (Unless `$TEST_TMPDIR` is
-  set, as in a test of bazel itself, in which case this directory is used
-  instead.)
+* The _outputRoot_ directory is ~/.cache/bazel on Linux, and /private/var/tmp
```
aiuto: Why don't we take the opportunity to improve the readability.  How about

The _outputRoot_ directory defaults to ~/.cache/bazel on Linux, and /private/var/tmp on macOS. If the environment variable `$TEST_TMPDIR` is set, as in a test of bazel itself, then that value overrides the default.

@laszlocsomor Do you know, offhand what we have for Windows. We might as well add that here too, while fixing it.

laszlocsomor: On Windows, we use `%HOME%` if set, otherwise `%USERPROFILE%` if set, otherwise call `SHGetKnownFolderPath` with `FOLDERID_Profile`.

https://github.com/bazelbuild/bazel/blob/22ca5ee2d2805300393f71ae95bb1e5a9f5a168b/src/main/cpp/workspace_layout.cc#L33-L35

https://github.com/bazelbuild/bazel/blob/22ca5ee2d2805300393f71ae95bb1e5a9f5a168b/src/main/cpp/blaze_util_windows.cc#L395-L407

https://github.com/bazelbuild/bazel/blob/22ca5ee2d2805300393f71ae95bb1e5a9f5a168b/src/main/cpp/blaze_util_windows.cc#L409-L440

gibfahn: >Why don't we take the opportunity to improve the readability. How about

Sure, why not, took your suggestion and added in the Windows info.

gibfahn: >Why don't we take the opportunity to improve the readability. How about

Sure, took your suggestion and added in the Windows info (which I wanted to add but couldn't find, thanks for the info @laszlocsomor ).

laszlocsomor:
@aiuto : could you please be the main reviewer and import this PR once it looks good?

gibfahn:
Gentle nudge @aiuto , I believe the existing feedback has been addressed.

Issue #10572, opened 2020-01-13
Bazel does not report details of the failed commands on windows.


Artem-B:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:
Bazel does not report details of the failed commands on windows.. When compilation fails, bazel does not print any info about the failed command. --verbose_failures does not help.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
BUILD:
cc_library(
  name = "lib",
  srcs = ["lib.cc"],
)

lib.cc:
#error This file fails to compile

```

Output on Windows (no info about the failed command and its options, with or without --verbose_faqilures):
```
$ bazel200.exe build lib --verbose_failures
INFO: Invocation ID: 13dab351-628a-4c68-9628-e912a25f15ec
Loading: 
Loading: 0 packages loaded
Analyzing: target //:lib (0 packages loaded, 0 targets configured)
INFO: Analyzed target //:lib (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
[0 / 3] [Scann] Compiling lib.cc
ERROR: C:/ktmpfs/repro/BUILD:1:1: C++ compilation of rule '//:lib' failed (Exit 2)
lib.cc(1): fatal error C1189: #error:  this file does not compile
Target //:lib failed to build
INFO: Elapsed time: 0.211s, Critical Path: 0.08s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
FAILED: Build did NOT complete successfully
```

Compare with the output on Linux, which does print the failed command:
```
Starting local Bazel server and connecting to it...
INFO: Analyzed target //:lib (13 packages loaded, 46 targets configured).
INFO: Found 1 target...
ERROR: /mnt/sda_crypt/tra/tf-win-break/BUILD:1:1: C++ compilation of rule '//:lib' failed (Exit 1) gcc failed: error executing command 
  (cd /usr/local/google/home/tra/work/.cache/bazel/_bazel_tra/5e7d1dac0bc55dad9def8e6cf6dbaef0/sandbox/linux-sandbox/1/execroot/__main__ && \
  exec env - \
    PATH=[....] \
    PWD=/proc/self/cwd \
[...]

PR #10581, opened 2020-01-14
Make --verbose_failures work again

meteorcloudy:
--verbose_failures is broken since Bazel 0.28.0, Bazel won't print the
failing command for local execution even --verbose_failures is enabled.

This change fixes the problem and adds a test for it.

Fixes https://github.com/bazelbuild/bazel/issues/10572

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/SpawnExecException.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/SpawnExecException.java
@@ -65,16 +65,9 @@ public class SpawnExecException extends ExecException {
     if (messagePrefix == null) {
       messagePrefix = action.describe();
     }
-    // Note: we intentionally do not include the ExecException here, unless verboseFailures is true,
-    // because it creates unwieldy and useless messages. If users need more info, they can run with
-    // --verbose_failures.
     String message =
-        result.getDetailMessage(messagePrefix, getMessage(), isCatastrophic(), forciblyRunRemotely);
-    if (verboseFailures) {
-      return new ActionExecutionException(message + ": " + getMessage(), this, action, isCatastrophic(), getExitCode());
-    } else {
-      return new ActionExecutionException(message, action, isCatastrophic(), getExitCode());
-    }
+        result.getDetailMessage(messagePrefix, getMessage(), verboseFailures, isCatastrophic(), forciblyRunRemotely);
+    return new ActionExecutionException(message + ": " + getMessage(), this, action, isCatastrophic(), getExitCode());
   }
 
   /** Return exit code depending on the spawn result. */
--- a/src/main/java/com/google/devtools/build/lib/exec/SpawnExecException.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/SpawnExecException.java
@@ -37,7 +37,7 @@ public class SpawnExecException extends ExecException {
     super(message, result.isCatastrophe());
     checkArgument(
         !Status.SUCCESS.equals(result.status()),
-        "Can't create exception with successful" + " spawn result.");
+        "Can't create exception with successful spawn result.");
     this.result = Preconditions.checkNotNull(result);
     this.forciblyRunRemotely = forciblyRunRemotely;
   }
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/devtools/build/lib/standalone/StandaloneSpawnStrategyTest.java
```diff
@@ -205,34 +209,56 @@ private ActionExecutionContext createContext() {
...
+      // the caveat is that the command will have some output to stdout.
+      // Default exit code of help is 1
+      return "C:/windows/system32/help.exe";
+    }
     return OS.getCurrent() == OS.DARWIN ? "/usr/bin/false" : "/bin/false";
   }
 
   private static String getTrueCommand() {
+    if (OS.getCurrent() == OS.WINDOWS) {
+      // No false command on Windows, with use whoami.exe as an alternative,
```
Artem-B: No **true** command on windows...

src/test/java/com/google/devtools/build/lib/standalone/StandaloneSpawnStrategyTest.java
```diff
@@ -205,34 +209,56 @@ private ActionExecutionContext createContext() {
...
+    }
     return OS.getCurrent() == OS.DARWIN ? "/usr/bin/true" : "/bin/true";
   }
 
   @Test
   public void testBinEchoPrintsArguments() throws Exception {
-    Spawn spawn = createSpawn("/bin/echo", "Hello,", "world.");
+    Spawn spawn;
+    if (OS.getCurrent() == OS.WINDOWS) {
+      spawn = createSpawn("C:/windows/system32/cmd.exe", "/c", "echo", "Hello,", "world.");
```
Artem-B: Are the tests assumed to be running on an arbitrary windows machine or do they operate in some sort of controlled environment (I.e. no need for the real windows installation in c:\windows).
If the tests may be executed on a real machine, I think it's a bit too optimistic to assume that windows is always installed in C:\WINDOWS. You should probably be using `%WinDir%` and `%SystemDrive%` environment variables in order to figure out the actual location.

src/test/java/com/google/devtools/build/lib/standalone/StandaloneSpawnStrategyTest.java
meteorcloudy: Thanks..

src/test/java/com/google/devtools/build/lib/standalone/StandaloneSpawnStrategyTest.java
meteorcloudy: The test mostly runs on Bazel CI, where we can safely assume Windows is installed in C:\Windows. Using environment variable doesn't work when launching a binary through spawn action, the underlying CreateProcess system call won't translate it. So I'll keep C:\Windows, but I did some refactor to make it easier to change the Windows system dir in case someone wants to run this test on a machine with a different Windows system dir.

src/main/java/com/google/devtools/build/lib/exec/SpawnExecException.java
```diff
@@ -71,7 +71,7 @@ public ActionExecutionException toActionExecutionException(
     String message =
         result.getDetailMessage(messagePrefix, getMessage(), isCatastrophic(), forciblyRunRemotely);
     if (verboseFailures) {
-      return new ActionExecutionException(message, this, action, isCatastrophic(), getExitCode());
+      return new ActionExecutionException(message + ": " + getMessage(), this, action, isCatastrophic(), getExitCode());
```
buchgr: From looking at the code I don't quite follow how this fixes it. The value of getMessage() is already appended to message by the implementation of SpawnResult#getDetailMessage. Why does appending it twice help?

meteorcloudy: No, I think in most case, getDetailMessage doesn't actually append the value of getMessage().
See 
https://github.com/bazelbuild/bazel/blob/2d0046a39e383613dac12b2dfe9845dd22df100f/src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java#L343

buchgr: Ah looked too quick :). Did you consider fixing this in getDetailMessage instead of here? I think this would allow us to avoid printing the same error message twice?

meteorcloudy: Make sense!

meteorcloudy: Please take a look again. ;)

buchgr:
LGTM

PR #10585, opened 2020-01-15
A few FilesystemValueChecker improvements.

benjaminp:
- Pass possibly-modified keys around as a Collection rather than an Iterable.
- Remove an atomic that counted the possibly-modified keys.
- Use a lambda in getDiff.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
@@ -113,7 +113,7 @@ public class FilesystemValueChecker {
    */
   public ImmutableBatchDirtyResult getNewAndOldValues(
       Map<SkyKey, SkyValue> valuesMap,
-      Iterable<SkyKey> keys,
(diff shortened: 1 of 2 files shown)


laurentlb:
What's the status of this? Should it be merged?

jin:
For primary reviewers, please read go/import-bazel-pr (internal link) for import instructions.

ulfjack:
Looks reasonable to me. @benjaminp, did you see any improvements with this change?

benjaminp:
> Looks reasonable to me. @benjaminp, did you see any improvements with this change?

The context of this change was that I was trying to make a more complex modification to `FileSystemValueChecker`. Part of the way through that I noticed the change would be very large, so I split this out. I don't think Bazel users will find this change wildly beneficial, but at least the world has one less thread-local atomic variable.

ulfjack:
IMO, this can be merged. @philwo, maybe you could import it and send it to Eric or Nathan as an internal review? It might make sense to mention that this is a preparatory change - I'm definitely interested in what @benjaminp is cooking up.

Issue #7152, opened 2019-01-17
incompatible_disable_legacy_proto_provider: Remove legacy `dep.proto.` access to the proto provider


lberki:
We are migrating off of legacy providers that are addressed by name to providers addressed: https://bazel.build/designs/skylark/declared-providers.html

This bug is for migrating the provider containing information about `.proto` files. The following changes are required in Starlark sources:

* Change the way the provider is accessed from `ctx.attr.dep.proto.` to `ctx.attr.dep[ProtoInfo]`
* Change the way the presence of provider is checked from `hasattr(dep, "proto")` to `ProtoInfo in dep`
* Change the way a rule requires a provider from `providers = [["​proto"]]` to `providers = [[ProtoInfo]]`

philwo:
This flag was not flipped in time for the Bazel 0.23.0 release and will thus be postponed to Bazel 0.24.0.

drigz:
`ProtoInfo` was introduced in 0.22.0. Is there a way for code to be compatible with both 0.21.0 and 0.24.0 Bazel versions?

laurentlb:
As it would break many projects on the CI and we plan to cut Bazel 0.24 tomorrow, the flag flip will be postponed.

katre:
This issue was tagged as "breaking-change-0.24" but does not appear ready to be flipped in the [0.24.0 release](https://github.com/bazelbuild/bazel/issues/6968). If this is incorrect please comment on that issue and discuss with me.

dkelmer:
Seems like the flag wasn't flipped in time for the 0.25 release. Changing window.

hlopko:
And it also wasn't flipped for 0.26 or 0.27, so 1.0 it is :)

lberki:
Fix for rules_rust is at https://github.com/bazelbuild/rules_rust/pull/242

lberki:
For fix Bazel is at https://github.com/bazelbuild/bazel/pull/8938

lberki:
Fix for `ts_proto_library` (and thus Bazel Codelabs) is at https://github.com/bazelbuild/rules_nodejs/pull/925 .

lberki:
Pull request to fix Envoy: https://github.com/envoyproxy/envoy/pull/7850

lberki:
Bug for rules_nodejs: https://github.com/bazelbuild/rules_nodejs/issues/979

lberki:
[...]

PR #10591, opened 2020-01-15
Remove unused LEGACY_SKYLARK_NAME from ProtoInfo

Yannic:
The legacy `.proto` provider has been removed in https://github.com/bazelbuild/bazel/issues/7152.


lberki:
Merci beaucoup! Let me import this.

PR #10596, opened 2020-01-15
Use minutes instead of seconds

JayBazuzi:
to be more human-friendly


aiuto:
But the timeout values are specified in seconds. If we give an example which converts to minutes, that would be confusing. It would lead some people down the path to thing that they can specify timeout values in minutes.

PR #10599, opened 2020-01-15
Fix a memory leak in linux sandbox

guibou:
`getcwd` result is allocated using `malloc` and should be deallocated with
`free`.

However `opt.working_dir` won't own it because that's an `std::string` and
the current `operator=` call will allocate a new buffer and copy the
content of `getcwd` result.

Hence, in the previous solution, the result of `getcwd` was just leaked at
the end of the function.

This commit adds the necessary `free`.

*edit*: I found that when running a C++ ASAN sanitizer during my build. I have a rule which setups the environment with `LD_PRELOAD and apparently bazel is calling the sandbox inside the build environment, hence it inherit the `LD_PRELOAD` setting.

### DIFF
--- a/src/main/tools/linux-sandbox-options.cc
+++ b/src/main/tools/linux-sandbox-options.cc
@@ -263,6 +263,9 @@ void ParseOptions(int argc, char *argv[]) {
 
   if (opt.working_dir.empty()) {
     char *working_dir = getcwd(nullptr, 0);
+    if(!working_dir) {
+      DIE("getcwd");
+    }
     opt.working_dir = working_dir;
     free(working_dir);
   }
--- a/src/main/tools/linux-sandbox-options.cc
+++ b/src/main/tools/linux-sandbox-options.cc
@@ -262,6 +262,8 @@ void ParseOptions(int argc, char *argv[]) {
   }
 
   if (opt.working_dir.empty()) {
-    opt.working_dir = getcwd(nullptr, 0);
+    char *working_dir = getcwd(nullptr, 0);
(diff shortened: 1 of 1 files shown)


jmmv:
While we are at this... `getcwd` can return `NULL` and that's not being gracefully handled. Could you fix that as well?

guibou:
@jmmv which semantic would you like to see? Default value of a fatal error?

jmmv:
I think a fatal error. `linux-sandbox` is a short-lived process, and if something goes wrong in it, we should just abort the action. (See all other `DIE` calls within the file.)

guibou:
@jmmv done. With a simple error message as it looks like that's the way to do it when I'm reading similar files.

src/main/tools/linux-sandbox-options.cc
```diff
@@ -263,6 +263,9 @@ void ParseOptions(int argc, char *argv[]) {
 
   if (opt.working_dir.empty()) {
     char *working_dir = getcwd(nullptr, 0);
+    if(!working_dir) {
```
susinmotion: Could you check against nullptr here?

guibou: You mean, `working_dir != nullptr`? May I ask why?

susinmotion: Since it's not a boolean -- passing along the (nit) comment from internal review. I changed it there, though, so no action needed.

guibou:
Thank you!

Issue #10606, opened 2020-01-17
glob() is non-deterministic if a source file path starts with `@`


jmillikin-stripe:
### Description of the problem / feature request:

When globbing a directory containing source files or directories starting with `@`, the output of `glob()` is not deterministic.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
$ cat BUILD
genrule(
    name = "test",
    srcs = ["//subdir:srcs"],
    outs = ["test_output.txt"],
    cmd = "find . > $@",
)

$ cat subdir/BUILD
srcs = glob(["**/*"])
print("srcs: ", srcs)
filegroup(
    name = "srcs",
    srcs = srcs,
    visibility = ["//visibility:public"],
)

$ cat subdir/@ohno.txt
$
```

```
$ bazel build //:test
DEBUG: /Users/jmillikin/src/bazeltest-globcolon/subdir/BUILD:2:1: srcs:  [":@ohno.txt", "BUILD"]
[...]
INFO: Build completed successfully, 2 total actions

$ touch subdir/BUILD
$ bazel build //:test
DEBUG: /Users/jmillikin/src/bazeltest-globcolon/subdir/BUILD:2:1: srcs:  ["@ohno.txt", "BUILD"]
ERROR: /Users/jmillikin/src/bazeltest-globcolon/subdir/BUILD:3:1: no such package '@ohno.txt//': The repository '@ohno.txt' could not be resolved and referenced by '//subdir:srcs'
ERROR: Analysis of target '//:test' failed; build aborted: no such package '@ohno.txt//': The repository '@ohno.txt' could not be resolved
```

### What operating system are you running Bazel on?

macOS and Linux

### What's the output of `bazel info release`?

`release 2.0.0`

PR #10610, opened 2020-01-17
Do glob result sorting and @-escaping at the last moment.

benjaminp:
c4f2d80270f1ce947fcf7fb0a4e5f0afb3a7062d changed legacy globbing (without a test) to prepend ':' to any targets starting with '@'. It did not change the skyframe globbing logic, which means incrementality bugs. Fix this issue by doing the escaping just before glob() returns its result, which is a common path for both glob implementations.

Fixes https://github.com/bazelbuild/bazel/issues/10606.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/Globber.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/Globber.java
@@ -39,11 +39,11 @@ public interface Globber {
       throws BadGlobException, InterruptedException;
 
   /**
-   * Fetches the result of a previously started glob computation. The returned list must be ordered
-   * deterministically. For more obvious correctness, implementations should generally sort the list
-   * they return.
+   * Fetches the result of a previously started glob computation. The returned list has an arbitrary
+   * order.
    */
-  List<String> fetch(Token token) throws BadGlobException, IOException, InterruptedException;
+  List<String> fetchUnsorted(Token token)
+      throws BadGlobException, IOException, InterruptedException;
 
   /** Should be called when the globber is about to be discarded due to an interrupt. */
   void onInterrupt();
--- a/src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
@@ -46,6 +46,7 @@ import com.google.devtools.build.lib.syntax.Tuple;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
@@ -100,7 +101,7 @@ public class SkylarkNativeModule implements SkylarkNativeModuleApi {
     try {
       Globber.Token globToken =
           context.globber.runAsync(includes, excludes, excludeDirs != 0, allowEmpty);
-      matches = context.globber.fetch(globToken);
+      matches = context.globber.fetchUnsorted(globToken);
     } catch (IOException e) {
       String errorMessage =
           String.format(
@@ -118,7 +119,17 @@ public class SkylarkNativeModule implements SkylarkNativeModuleApi {
       throw new EvalException(null, "illegal argument in call to glob", e);
     }
 
-    return StarlarkList.copyOf(thread.mutability(), matches);
+    ArrayList result = new ArrayList<>(matches.size());
+    for (String match : matches) {
+      if (match.charAt(0) == '@') {
+        // Add explicit colon to disambiguate from external repository.
+        match = ":" + match;
+      }
+      result.add(match);
+    }
+    result.sort(Comparator.naturalOrder());
+
+    return StarlarkList.copyOf(thread.mutability(), result);
   }
 
   @Override
--- a/src/main/java/com/google/devtools/build/lib/packages/GlobCache.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/GlobCache.java
@@ -180,10 +180,6 @@ public class GlobCache {
       // invalid as a label, plus users should say explicitly if they
       // really want to name the package directory.
       if (!relative.isEmpty()) {
-        if (relative.charAt(0) == '@') {
-          // Add explicit colon to disambiguate from external repository.
-          relative = ":" + relative;
-        }
         result.add(relative);
       }
     }
--- a/src/main/java/com/google/devtools/build/lib/packages/PackageFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/PackageFactory.java
@@ -73,7 +73,6 @@ import com.google.devtools.build.lib.vfs.RootedPath;
 import com.google.devtools.build.lib.vfs.UnixGlob;
 import java.io.IOException;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -265,11 +264,9 @@ public final class PackageFactory {
   /** {@link Globber} that uses the legacy GlobCache. */
   public static class LegacyGlobber implements Globber {
     private final GlobCache globCache;
-    private final boolean sort;
 
-    private LegacyGlobber(GlobCache globCache, boolean sort) {
+    private LegacyGlobber(GlobCache globCache) {
       this.globCache = globCache;
-      this.sort = sort;
     }
 
     private static class Token extends Globber.Token {
@@ -299,7 +296,7 @@ public final class PackageFactory {
     }
 
     @Override
-    public List<String> fetch(Globber.Token token)
+    public List<String> fetchUnsorted(Globber.Token token)
         throws BadGlobException, IOException, InterruptedException {
       List<String> result;
       Token legacyToken = (Token) token;
@@ -309,9 +306,6 @@ public final class PackageFactory {
               legacyToken.excludes,
               legacyToken.excludeDirs,
               legacyToken.allowEmpty);
-      if (sort) {
-        Collections.sort(result);
-      }
       return result;
     }
 
@@ -792,7 +786,7 @@ public final class PackageFactory {
 
   /** Returns a new {@link LegacyGlobber}. */
   public static LegacyGlobber createLegacyGlobber(GlobCache globCache) {
-    return new LegacyGlobber(globCache, /*sort=*/ true);
+    return new LegacyGlobber(globCache);
   }
 
   @Nullable
--- a/src/main/java/com/google/devtools/build/lib/skyframe/PackageFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/PackageFunction.java
@@ -76,7 +76,6 @@ import com.google.devtools.build.skyframe.ValueOrException2;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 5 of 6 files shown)


benjaminp:
@haxorz, you may find this interesting.

haxorz:
Yes, tyvm for the Github mention. Also tyvm for the good commit message.

This bug & commit interests me for many reasons, one of which is that, bleh, there are been many glob incrementality bugs like this in the past (c9fbe95 and bb5d5ef come to mind).

benjaminp:
You must have a strong reason for keeping the "legacy" globbing logic around.

haxorz:
Yes. The reasons are purely technical in nature and have absolutely nothing to do with Google-Blaze stuff. I don't have a Github links handy (sorry!), so I will try to briefly summarize the issues here:
- SkyFunction restart stuff (SkyFunction#compute is supposed to return null when their deps aren't ready), coupled with the fact that state from the previous #compute call isn't retained or reused, means that if a SkyFunction does a lot of work and then gets restarted, it has to do the work again.
- Starlark execution is logically single threaded and uses a blocking execution model.
- Starlark execution for the code in a BUILD file (e.g. in the "loading phase") happens in a single Skyframe-land thread (specifically, doing PackageFunction-land work)
- The Starlark `glob` function takes an arbitrary string list for the `include` argument
- globbing work is inherently embarrassingly parallel. But the way to use Skyframe-land parallelism is to declare deps, but declaring GlobValue deps means you might get restarted, which means you might need to redo work.
- Now consider adversarial .bzl code like

```
def a(x):
  for y in glob(b(x)):
    c(y)
```
- LegacyGlobber lets Bazel do glob evaluation intra-Skyframe-thread, by making use of a threadpool external to Skyframe.
- My commit message in 3a95f35 (Feb 2016) summarizes the above
- There have been some casual attempts to address in the past and they've all failed. One example is 833fd12fd86ddc32e89b1b0cff57b248a58dc8ab => acb02f52533e70a807d528aceb232b774ec0ee56. I can go into more details if you want. A concerted effort to address some of the above issues could probably succeed.
- Fwiw, as embarrassing as the existing business logic in Bazel is, and as bad as it is to have LegacyGlobber-induced bugs, I don't think the actual *wall time cost* from this LegacyGlobber junk is a big deal in practice. That's why the above bullet point hasn't been prioritized.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
```diff
@@ -118,7 +119,18 @@
       throw new EvalException(loc, "illegal argument in call to glob", e);
     }
 
-    return StarlarkList.copyOf(thread.mutability(), matches);
+    String[] result = new String[matches.size()];
```
haxorz: I'm a bit afraid of moving the sorting out of PackageFunction and into here, since it invites a future bug where `glob` evaluation goes through something other than SkylarkNativeModule. @c-parsons thoughts?

@meisterT are you sure c4f2d80270f1ce947fcf7fb0a4e5f0afb3a7062d needed to add support for the '@' character? In general, I wish Bazel supported less stuff...

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
```diff
@@ -118,7 +119,18 @@
...
+    for (int i = 0; i < result.length; i++) {
+      String match = matches.get(i);
+      if (match.charAt(0) == '@') {
+        // Add explicit colon to disambiguate from external repository.
+        match = ":" + match;
+      }
+      result[i] = match;
+    }
+    Arrays.sort(result);
```
haxorz: Minor performance concern: If the glob is a Skyframe-cache-hit, then matches will be already sorted. I know that TimSort is good on already sorted arrays, but still... 

Sorry for lack of concrete suggestion; I'm pretty busy today. I'll try to take another look next week (will involve reading over all the code).

benjaminp: `PackageFunction.HybridToken.resolveAsync`, which isn't cached by Skyframe, does the sorting of Skyframe results. So, unless I'm misunderstanding, this doesn't actually change the number of sorts of cached globs.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
benjaminp: I have no personal opinion about whether `@` should be supported or not. (One might note that #374 has more thumbs up and advocacy that the average Bazel issue.) I expect at this point, we couldn't delete this behavior with a deprecation period.

I agree removing sorting from `PackageFunction` isn't unproblematic. But as far as fixing the linked issue is concerned, to not sort here, we'd have to think of some other place shared by legacy and skyframe globbing to put the '@'-prefixing logic. I also don't conceptually think this prefixing behavior belongs to globbing. Globbing should return filesystem paths not targets.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
haxorz: > [me in previous github comment] If the glob is a Skyframe-cache-hit, then matches will be already sorted

Er, this may wrong, sorry. PackageFunction.HybridToken.resolve had (past-tense b/c you're deleting it in this commit) a comment "Skyframe glob results are unsorted". I wrote this years ago and may not be true now (or ever). Someone should think about this (I can next week if I have time).

> [benjaminp] So, unless I'm misunderstanding, this doesn't actually change the number of sorts of cached globs.

Yes you're correct. Thanks for the correction.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
haxorz: > I expect at this point, we couldn't delete this behavior with a deprecation period.

Yes, ofc.

> I also don't conceptually think this prefixing behavior belongs to globbing. Globbing should return filesystem paths not targets.

Completely agree.

I was curious about the '@' support b/c if it hypothetically didn't exist then we wouldn't be talking about how the inappropriate location of the prefixing code.

clint-stripe: `@` is quite common in npm package names, e.g. [`@bazel/typescript`](https://www.npmjs.com/package/@bazel/typescript), and consequently shows up in filenames (particularly as a directory name...)

meisterT: Yes, npm was the driving factor behind my change.

haxorz:
I'm planning to review this by EOD.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
haxorz: Okay. How about the following:
- Rename Globber#fetch to Globber#fetchUnsorted (and update javadoc accordingly). Your changes in this CL are making the javadoc for Globber#fetch not match either implementation, and I also don't want future code readers to assume the returned List<String> has any meaningful order.
- Keep your colon-prefixing and sorting here in SkylarkNativeModule.java. I really like your reasoning that colon-prefixing doesn't belong in the core globbing code.

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
```diff
@@ -118,7 +119,18 @@
...
+
+    return StarlarkList.wrap(thread.mutability(), result);
```
haxorz: Nit: How about instead doing

  `StarlarkList.copyOf(thread.mutability(), Arrays.asList(matches))`

This way you don't need to make #wrap public. I don't think the Starlark folks want that one to be public. In particular, @alandonovan has been trying to clean up java-land visibility and API issues in the Starlark implementation.

benjaminp: This is all because I wanted to avoid a unnecessary copy. This is a situation where it's definitely okay to "donate" the array to `StarlarkList`.

A possible alternative to this would be having public `StarkList` APIs to preallocate a list with the correct size and provide some way to sort it.

Possibly also I just shouldn't worry about it.

haxorz: > This is all because I wanted to avoid a unnecessary copy. 

I initially thought the same and then was going to suggest implement Iterable wrapper around an Array, and then I remembered Arrays.asList: https://docs.oracle.com/javase/9/docs/api/java/util/Arrays.html#asList-T...-

src/main/java/com/google/devtools/build/lib/packages/Globber.java
```diff
@@ -39,11 +39,12 @@ Token runAsync(
       throws BadGlobException, InterruptedException;
 
   /**
-   * Fetches the result of a previously started glob computation. The returned list must be ordered
-   * deterministically. For more obvious correctness, implementations should generally sort the list
-   * they return.
+   * Fetches the result of a previously started glob computation. The returned list has an arbitrary
+   * order. For more obvious correctness, implementations should generally sort the list they
```
haxorz: Let's just delete the second sentence ("For more obvious correctness..."). Or if you want you can rephrase to something like "Consumers probably want to sort this list".

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
benjaminp: But `matches` is a `List` not an array.

haxorz: Gosh I'm sorry -- bad copy/paste on my part.

Should be `StarlarkList.copyOf(thread.mutability(), Arrays.asList(result))`; that'd work, right?

src/main/java/com/google/devtools/build/lib/packages/Globber.java
benjaminp: deleted

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
benjaminp: It adds a copy because `StarlarkList.copyOf` will call `Iterables.toArray` on it, which will make a copy.

haxorz: Okay, now I'm embarrassed :/

Let's just keep what you have right now (i.e. public #wrap) and someone on the Starlark team can take a look some other time.

alandonovan: Please do not make wrap public. The arrays referenced by Tuple and StarlarkList must be completely encapsulated, like ImmutableList. You must incur one copy; copyOf(Arrays.asList(x)) is optimal.

haxorz:
Hi @benjaminp,

I'd like to publicly apologize for the additional delay.

For the past 8 days, I intentionally did nothing because I was under the outdated and incorrect assumption that someone else on the Bazel team would merge this PR. I was curious why there was no movement here, so I asked around last night and learned the terrible truth that my inaction was at fault. I'm so sorry for my ignorance! It's been a while since I've reviewed a PR.

I [and the rest of the team, trust me, ] truly appreciate your dedication to making Bazel awesome.

Have a great weekend,

-Nathan

src/main/java/com/google/devtools/build/lib/packages/SkylarkNativeModule.java
```diff
@@ -118,7 +119,17 @@
       throw new EvalException(null, "illegal argument in call to glob", e);
     }
 
-    return StarlarkList.copyOf(thread.mutability(), matches);
+    ArrayList result = new ArrayList<>(matches.size());
```
haxorz: Fyi, in 3b39e1e I changed this variable declaration to "ArrayList<String>"

benjaminp:
No problem. Thanks for the review and the extensive comment about legacy versus skyframe globbing.

PR #10611, opened 2020-01-17
Send non-quiet repository rule output to stderr

### LINKED ISSUE #8077: repository_rule with quiet=false breaks output of bazel query
Users expect that bazel query output can be piped to other commands, e.g. for Angular releases we do something like

`bazel query --output=label 'kind(".*_package", //packages/...)' | xargs bazel build`

However if this runs a repository rule that has some logging output (doesn't use quiet=True to hide the stdout from actions it runs) then this output appears in the bazel query output in a way that can't be separated from the query result.

Originally filed at https://github.com/bazelbuild/rules_nodejs/issues/583 where you can see an end-user affected by it.

DavidANeil:
Resolves #8077


aiuto:
Ping reviewers

laurentlb:
Thanks a lot!

@DavidANeil I've missed the notification, and I'm very sorry for the delay. I'm importing this change right now.

PR #10614, opened 2020-01-19
Always add '-fmodules' flag when objc_library's enable_modules is set to True

thii:
I wasn't able to find any documentation on why modules was not enabled even though users explicitly specify it.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -1229,7 +1229,7 @@ public class CompilationSupport {
       }
     }
 
-    if (attributes.enableModules() && !getCustomModuleMap(ruleContext).isPresent()) {
+    if (attributes.enableModules()) {
       copts.add("-fmodules");
     }
     if (copts.contains("-fmodules")) {


thii:
Requested a review from @laszlocsomor because you were the author of the `objc_library`’s custom module map feature.

laszlocsomor:
I don't recall that. Which commit was it?

Passing review to @c-parsons .

thii:
This one: https://github.com/bazelbuild/bazel/commit/fba07bb72570245d26bd8795709c5a004fc9026a.

laszlocsomor:
Ah, thanks. No, I didn't author that, just committed it. /cc @allevato

allevato:
Unfortunately I don't have the historical context for why that check is in there either. There was a corresponding check in LegacyCompilatonSupport.java that adds `-fmodule-maps` instead of `-fmodules`, which means "use module maps to verify integrity of headers, but don't enable modules", and in that context checking for the custom module map makes sense—there's no point in checking the integrity of the headers if you're generating the module map from the headers in the build target. The original author could have just added the condition in both places 🤷‍♂ 

At the same time, I'm not sure whether it's worth keeping `enable_modules` on life support. It's not the way modules should be implemented, and given that there's a migration currently in the works to unify the Objective-C rules and the C++ rules, focusing additional work on features of the Obj-C rules that won't exist in the future seems like the wrong approach.

You can get the same effect of `enable_modules = True` by just passing `-fmodules` in your `copts`. That's pretty much all that this attribute does anyway from what I can tell; the only other thing it does is pass `-fmodule-maps` when it's *not* enabled, but `-fmodules` implies and overrides `-fmodule-maps` anyway, so passing `-fmodules` as a copt should work in all cases. We should probably just remove `enable_modules` at this point, unless there's something about its behavior that I'm missing.

laszlocsomor:
Thanks for the detailed answer!

tetromino:
Closing the PR given the reviewers' comments.

Issue #10133, opened 2019-10-30
Release 2.2 (stable) - February 2020


dslomov:
# Status of Bazel 2.2.0

- Target baseline: 78055efad0917b848078bf8d97b3adfddf91128d
- Expected release date: 2020-02-20
- [List of release blockers](https://github.com/bazelbuild/bazel/labels/Release%20blocker)

To report a release-blocking bug, please file a bug using the `Release blocker` label, and cc me.

Task list:

- [x] Update GitHub issues for incompatible changes
- [x] Pick release baseline: 78055efad0917b848078bf8d97b3adfddf91128d
- [x] Create release candidate: https://releases.bazel.build/2.2.0/rc1/index.html
- [x] Check downstream projects: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1389
- [x] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [ ] Send for review the release announcement PR:
- [X] Push the release, notify package maintainers: @vbatts @petemounce @excitoon
- [ ] Update the documentation
- [ ] Push the blog post
- [X] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

jmillikin-stripe:
Scheduling note: if possible, I'd love to get https://github.com/bazelbuild/bazel/pull/10622 into this release -- it's been delayed a fairly long time on design reviews, and our internal builds are suffering for it.

philwo:
@meteorcloudy Can we make Bazel 2.2 our first release built on Kokoro? :)

meteorcloudy:
Yes!

philwo:
@jmillikin-stripe I'm sorry, the PR you wanted to have in Bazel 2.2 still hasn't been merged yet, but I'm already super late with cutting the baseline for Bazel 2.2. The next Bazel release will already be cut on March 1st. 

 pinged Jakob, but he said he can't get it done right now, so I have to go ahead and prepare Bazel 2.2 now.

philwo:
I'm preparing Bazel 2.2.0 now:

```
RELEASE_NUMBER=2.2.0
BASELINE_COMMIT=78055efad0917b848078bf8d97b3adfddf91128d
git clone https://github.com/bazelbuild/bazel.git ~/bazel-release-$RELEASE_NUMBER
cd ~/bazel-release-$RELEASE_NUMBER
[...]

PR #10622, opened 2020-01-21
Implement RemoteDownloader w/ `--experimental_remote_downloader`

jmillikin-stripe:
This is the Bazel client implementation of https://github.com/bazelbuild/proposals/pull/160. It allows downloading of external dependencies to be delegated to a remote service.

TODOs:
- [x] Once https://github.com/bazelbuild/remote-apis/pull/112 is merged, the vendored copy of `bazelbuild/remote-apis` should be updated. I've used a [WIP] placeholder for now.
- [x] If the general approach looks reasonable then I'll add tests. Currently I've been testing with an in-house implementation of the downloader server.

R: @buchgr @dslomov
CC: @EricBurnett @sstriker @ulfjack


buchgr:
Thanks for the PR. I think the code should rather live in the downloader package next to the http implementation. I don't really see a good reason for it being in the remote package. WDYT?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
```diff
@@ -0,0 +1,88 @@
...
+ * <p>The actual checksum is computed gradually as the output is written. If it
+ * doesn't match, then an {@link IOException} will be thrown when {@link #close()}
+ * is called. This error will be thrown multiple times if these methods are called
+ * again for some reason.
+ *
+ * <p>This class is not thread safe, but it is safe to message pass this object
+ * between threads.
+ */
+@ThreadCompatible
+public final class HashOutputStream extends OutputStream {
```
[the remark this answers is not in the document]
jmillikin-stripe: I started with a wrapper around `HashingOutputStream` but the code was about the same size as copying `HashInputStream` from this directory. The important behavior (throwing an exception on mismatch from an expected `HashCode`) isn't in Guava.

jmillikin-stripe:
> Thanks for the PR. I think the code should rather live in the downloader package next to the http implementation. I don't really see a good reason for it being in the remote package. WDYT?

When I tried to do this I hit circular dependency problems -- quite a lot of dependency problems, actually, because of how large the `java_library` targets are here.

I could try moving it back but it might require creating shim interfaces for most of the remote execution utilities (e.g. the tracing metadata injector).

I could also try creating a `RemoteDownloader` in `downloader/`, and then a separate `GrpcRemoteDownloader extends RemoteDownloader` that implements some part of the interface. This seemed unnecessary given that there's no plan for an `HttpRemoteDownloader`.

buchgr:
It's hard for me to say how bad the circular dependency problem is. I'd much prefer if this functionality was implemented outside of the remote module. I think we should treat the Fetcher service and remote execution/caching as separate. Both make sense to be used without the other.

For the flag naming how about `--external_dep_proxy`? Just an idea.

jmillikin-stripe:
> I think we should treat the Fetcher service and remote execution/caching as separate.

The remote downloader has a hard dependency on the remote CAS. It's not possible to use `--remote_downloader` without `--remote_cache` (and there's a check in `RemoteOptions` to enforce this).

jmillikin-stripe:
Done -- Rebased to master, un-WIP'd the remote-apis vendor commit, and relocated the `GrpcRemoteDownloader` to the `downloader/` directory.

Does this look better? If so, I'll start on the tests.

jmillikin-stripe:
Timed out waiting for response. I've added a unit test `GrpcRemoteDownloaderTest` to `DownloaderTestSuite`, cribbing heavily from `HttpDownloaderTest` and `GrpcCacheClientTest`.

Also removing [WIP] since the TODOs are resolved.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+      Path destination,
+      com.google.common.base.Optional<Checksum> checksum) throws IOException {
+    OutputStream out = destination.getOutputStream();
+    if (checksum.isPresent()) {
+      out = new HashOutputStream(out, checksum.get());
+    }
+    return out;
+  }
+
+  private static String authHeadersJson(Map<URI, Map<String, String>> authHeaders) {
```
buchgr: How does this interact with the authentication methods from the remote module? Why do we want this?

src/main/java/com/google/devtools/build/lib/remote/BUILD
```diff
@@ -18,13 +18,22 @@ filegroup(
 
 java_library(
     name = "remote",
-    srcs = glob(["*.java"]),
+    srcs = glob(["*.java"], exclude = [
+        "ExecutionStatusException.java",
+        "ReferenceCountedChannel.java",
+        "RemoteRetrier.java",
+        "RemoteRetrierUtils.java",
+        "Retrier.java",
```
buchgr: That's surprising. Why do you need this?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
jmillikin-stripe: `ctx.download()` allows arbitrary HTTP headers to be passed to the HTTP library for authenticating with the remote endpoint. This is part of the `netrc` support added in ... v2.0? v2.1? Fairly recently.

This code implements that by passing it along to the remote download server, which is free to use (or ignore) that data. The download server would use it when the place being downloaded from (e.g. example.com) needs special headers such as HTTP Basic Auth.

src/main/java/com/google/devtools/build/lib/remote/BUILD
jmillikin-stripe: The `RemoteRetrier` class implements the logic of `--remote_retries`, which causes Bazel to retry failed gRPC calls sent to the remote executor/storage/downloader.

jmillikin-stripe:
(rebased for https://github.com/bazelbuild/bazel/issues/10679)

src/main/java/com/google/devtools/build/lib/remote/BUILD
buchgr: Let me be more clear: This target looks odd. Why do you need to exclude the sources? Is there a way to have more meaningful targets rather than one large one?

jmillikin-stripe: I would personally prefer smaller targets, and I also think it looks weird, but this is all pre-existing weirdness in Bazel.
* `//src/main/java/com/google/devtools/build/lib/remote` is a glob of every Java file in the directory.
* Since there is now a dependency of `RemoteModule` -> `downloader` -> `RemoteRetrier`, we need to have at least some of the Java files in `remote/` be a separate target.
* Other targets handle this case with globs that exclude some targets.

If you feel like adjusting the `remote/` directory to have finer-grained targets that sounds like a good idea to me. I would like to prefer doing any major surgery to the Bazel build graph as part of this PR.

laurentlb:
@buchgr, do you think this PR can be merged in time for Bazel 2.2?

(@jmillikin-stripe asked for it in https://github.com/bazelbuild/bazel/issues/10133#issuecomment-581330035)

jmillikin-stripe:
gentle ping @buchgr 

Please let me know if this PR is too large to easily review, I can split some parts (like the `remote-apis` vendor update) into separate PRs if that would be better.

buchgr:
Sorry I haven't forgotten about it. I am currently overloaded with an internal project. We'll ship it in 2.2.

jmillikin-stripe:
I'm going to try splitting this up so it's easier to land incrementally, with the hopes of getting it into v2.3.

First subset is vendoring the updated remote-apis protos: https://github.com/bazelbuild/bazel/pull/10818

laurentlb:
@buchgr Any update on this?
Should someone else review the PR?

jmillikin-stripe:
Ping: March is approaching.

jin:
@buchgr ping, could you please bring this in? Is this in a mergeable state?

jmillikin-stripe:
Rebased for the update of vendored remote-apis protos.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.devtools.build.lib.concurrent.ThreadSafety.ThreadCompatible;
+import java.io.IOException;
+import java.io.OutputStream;
+import javax.annotation.Nullable;
+import javax.annotation.WillCloseWhenClosed;
+
+/**
+ * Output stream that guarantees its contents matches a hash code.
+ *
+ * <p>The actual checksum is computed gradually as the output is written. If it
```
ulfjack: Please add that the data is still written through to the underlying stream, even if it doesn't match.

src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+package com.google.devtools.build.lib.remote;
+
+import com.google.common.base.Preconditions;
+import com.google.devtools.build.lib.bazel.repository.downloader.Downloader;
+import com.google.devtools.build.lib.bazel.repository.downloader.DownloaderFactory;
+import java.util.Optional;
+
+public final class RemoteDownloaderFactory implements DownloaderFactory {
+
+  private volatile Downloader remoteDownloader;
```
ulfjack: Prefer synchronized over volatile. You almost never need it from a performance point of view, and it's much harder to reason about.

src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    Preconditions.checkState(this.remoteDownloader == null);
+    this.remoteDownloader = remoteDownloader;
+  }
+
+  public void reset() {
+    this.remoteDownloader = null;
+  }
+
+  @Override
+  public Optional<Downloader> create() {
```
ulfjack: This implementation technically violates the specification on the interface, which says:
Returns a ***new*** Downloader, or empty.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -95,18 +97,25 @@
   private final RepositoryRemoteExecutorFactoryDelegate repositoryRemoteExecutorFactoryDelegate =
       new RepositoryRemoteExecutorFactoryDelegate();
 
+  private final RemoteDownloaderFactory remoteDownloaderFactory = new RemoteDownloaderFactory();
```
ulfjack: Not asking you to make any changes right now, but why is this part of the RemoteModule? Would it make sense to use the remote downloader without remote execution?

src/main/java/com/google/devtools/build/lib/remote/RemoteRetrierUtils.java
```diff
@@ -14,6 +14,7 @@
 
 package com.google.devtools.build.lib.remote;
 
+import com.google.devtools.build.lib.remote.ExecutionStatusException;
```
ulfjack: Revert this line?

src/main/java/com/google/devtools/build/lib/runtime/BlazeRuntime.java
```diff
@@ -1589,7 +1597,8 @@ public BlazeRuntime build() throws AbruptExitException {
           productName,
           serverBuilder.getBuildEventArtifactUploaderMap(),
           serverBuilder.getAuthHeadersProvidersMap(),
-          serverBuilder.getRepositoryRemoteExecutorFactory());
+          serverBuilder.getRepositoryRemoteExecutorFactory(),
+          serverBuilder.getDownloaderFactory());
```
ulfjack: This looks wrong. This means it's not possible to have different Downloader factories for different builds in the same client. Shouldn't this be part of the CommandEnvironment instead?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/BUILD
```diff
@@ -10,7 +10,9 @@ filegroup(
 
 java_library(
     name = "downloader",
-    srcs = glob(["*.java"]),
+    srcs = glob(["*.java"], exclude = [
```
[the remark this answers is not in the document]
jmillikin-stripe: Done.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+import java.net.URI;
+import java.net.URL;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.TreeMap;
+
+/** A Downloader implementation that uses gRPC calls to a remote server. */
```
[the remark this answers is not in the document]
jmillikin-stripe: Done. I linked to the upstream repo for the Remote Execution APIs, of which Fetch is a part.

Note that I copied this comment from `GrpcRemoteExecutor` -- if more commentary is expected for these classes, it might also be good to add some there.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+      final Digest blobDigest = response.getBlobDigest();
+
+      retrier.execute(() -> requestCtx.call(() -> {
+        try (OutputStream out = newOutputStream(destination, checksum)) {
+          Utils.getFromFuture(cacheClient.downloadBlob(blobDigest, out));
+        }
+        return null;
+      }));
+    } catch (StatusRuntimeException e) {
+      throw new IOException(e);
```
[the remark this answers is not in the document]
jmillikin-stripe: I copied this exception handling from `GrpcCacheClient`, and I don't know enough about grpc-java to understand the exact issue you describe. Would it be possible for a Bazel dev to make the existing code act as you describe? Then I could copy it, and/or use a shared helper that does the right thing w.r.t. error handling.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  static FetchBlobRequest newFetchBlobRequest(
+      String instanceName,
+      List<URL> urls,
+      Map<URI, Map<String, String>> authHeaders,
+      com.google.common.base.Optional<Checksum> checksum,
```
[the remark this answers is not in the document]
jmillikin-stripe: This is part of the existing interface. I don't really want to make any extensive refactorings to the Bazel download<->repository_ctx path as part of this PR.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+    for (URL url : urls) {
+      requestBuilder.addUris(url.toString());
+    }
+    if (checksum.isPresent()) {
+      requestBuilder.addQualifiers(
+        Qualifier.newBuilder()
+            .setName("checksum.sri")
+            .setValue(checksum.get().toSubresourceIntegrity()).build());
+    }
+    if (!Strings.isNullOrEmpty(canonicalId)) {
```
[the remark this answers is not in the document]
jmillikin-stripe: The purpose and semantics of `canonical_id` is documented at https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,183 @@
...
+      requestBuilder.addQualifiers(
+        Qualifier.newBuilder()
+            .setName("bazel.canonical_id")
+            .setValue(canonicalId).build());
+    }
+    if (!authHeaders.isEmpty()) {
+      requestBuilder.addQualifiers(
+        Qualifier.newBuilder()
+            .setName("bazel.auth_headers")
```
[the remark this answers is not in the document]
jmillikin-stripe: There's no other way to get authentication data passed through to a remote downloader, and one of the core use cases for this code is to allow fully isolating Bazel behind a validating proxy. We can't do that if some downloads still need to go direct for header reasons.

Given that the API side of this ended up so generic (an unstructured `map<string, string>`) there doesn't seem to be much Bazel can do about potential security issues if a download server is incorrectly sharing content between different trust domains.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
jmillikin-stripe: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
jmillikin-stripe: `RemoteModule` handles all three remote capabilities: execution, caching, and downloading.

Remote downloading has a hard dependency on the remote cache, because that's where it puts downloaded content for consumption by Bazel.

src/main/java/com/google/devtools/build/lib/remote/RemoteRetrierUtils.java
jmillikin-stripe: Done.

src/main/java/com/google/devtools/build/lib/runtime/BlazeRuntime.java
jmillikin-stripe: I don't understand the question, but handling of the remote downloader should match that of the other remote capabilities (caching and execution). The calls to `getRepositoryRemoteExecutorFactory()` and `getDownloaderFactory()` belong together.

If they belong together _somewhere else_, could we do that separately? I'm very out of my depth when it comes to this level of the Bazel client architecture.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/GrpcRemoteDownloader.java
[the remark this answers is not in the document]
jmillikin-stripe: Also, my understanding is that `Optional` is preferred over nullable values in Bazel's codebase.

src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java
jmillikin-stripe: This logic is copied from the existing delegate/factory stacks, such as `RepositoryRemoteExecutorFactoryDelegate`. To be honest I have no idea why this level of indirection is necessary, and would be happy to have some guidance (or, like, direct commits to this PR) from someone who writes Java regularly.

I tried using `synchronized` instead, but the build fails:
```
src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java:24: error: modifier synchronized not allowed here
  private synchronized Downloader remoteDownloader;
```

src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java
jmillikin-stripe: Changed `DownloaderFactory::create()` to not say "new".

jmillikin-stripe:
@ulfjack Thank you so much for the review! I've addressed most comments, but there are a few places where I either didn't understand the request (I don't write much Java), or where the issue was not introduced by my change.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloaderDelegate.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.devtools.build.lib.events.ExtendedEventHandler;
+import com.google.devtools.build.lib.vfs.Path;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class DownloaderDelegate implements Downloader {
```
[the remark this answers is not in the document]
ulfjack: This is still open...

src/main/java/com/google/devtools/build/lib/remote/RemoteDownloaderFactory.java
ulfjack: Oh, sorry. The synchronized goes on all methods that access the field, like so:

private Downloader remoteDownloader;

public synchronized void init(Downloader remoteDownloader) {
}

src/main/java/com/google/devtools/build/lib/runtime/BlazeRuntime.java
michajlo: I think what ulf means is, assuming you're familiar with the client-server setup, the DownloaderFactory instance (and thus implementation) is tied to the lifetime of the server. It cannot be changed across different commands to the same server. I think this is fine since it's no worse than the current setup for other remote stuff, and there's only one impl.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
```diff
@@ -0,0 +1,92 @@
...
+    delegate.close();
+    check();
+  }
+
+  private void check() throws IOException {
+    if (actual == null) {
+      actual = hasher.hash();
+    }
+    if (!code.equals(actual)) {
+      throw new UnrecoverableHttpException(
```
michajlo: This side-effect is unfortunately hidden, but I guess there's prior art for this.

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
```diff
@@ -0,0 +1,314 @@
...
+  private final MutableHandlerRegistry serviceRegistry = new MutableHandlerRegistry();
+  private final String fakeServerName = "fake server for " + getClass();
+  private Server fakeServer;
+  private Context withEmptyMetadata;
+  private Context prevContext;
+  private static ListeningScheduledExecutorService retryService;
+
+  @BeforeClass
+  public static void beforeEverything() {
+    retryService = MoreExecutors.listeningDecorator(Executors.newScheduledThreadPool(1));
```
michajlo: why not put this in the before and after? Otherwise we can leak threads across tests within this test class

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
```diff
@@ -0,0 +1,314 @@
...
+    return new GrpcRemoteDownloader(
+        channel.retain(),
+        Optional.<CallCredentials>empty(),
+        retrier, withEmptyMetadata, cacheClient, remoteOptions);
+  }
+
+  private static byte[] downloadBlob(
+      GrpcRemoteDownloader downloader, URL url, Optional<Checksum> checksum)
+      throws IOException, InterruptedException {
+    final List<URL> urls = ImmutableList.<URL>of(url);
```
michajlo: Modern java should be smart enough that you can leave the "<URL>" out of "ImmutableList.<URL>of(...)", so rm it. Same for other occurrences where the compiler is fine with it.

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
```diff
@@ -0,0 +1,314 @@
...
+    fakeServer.awaitTermination();
+  }
+
+  @AfterClass
+  public static void afterEverything() {
+    retryService.shutdownNow();
+  }
+
+  private GrpcRemoteDownloader newDownloader(RemoteCacheClient cacheClient) throws IOException {
+    final RemoteOptions remoteOptions = Options.getDefaults(RemoteOptions.class);
```
michajlo: btw, final shouldn't be necessary on most local variables, but it's not necessary to remove

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
```diff
@@ -0,0 +1,314 @@
...
+            responseObserver.onCompleted();
+          }
+        });
+
+    final RemoteCacheClient cacheClient = new InMemoryCacheClient();
+    final GrpcRemoteDownloader downloader = newDownloader(cacheClient);
+
+    getFromFuture(cacheClient.uploadBlob(contentDigest, ByteString.copyFromUtf8("wrong content")));
+
+    IOException e = assertThrows(UnrecoverableHttpException.class, () -> downloadBlob(
```
michajlo: Fix formatting to improve readability:

    IOException e = assertThrows(
        UnrecoverableHttpException.class,
        () -> downloadBlob(
            downloader,
             ...,
             ...));

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Downloader.java
```diff
@@ -43,6 +43,7 @@ void download(
       List<URL> urls,
       Map<URI, Map<String, String>> authHeaders,
       Optional<Checksum> checksum,
+      String canonicalId,
```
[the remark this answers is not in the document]
jmillikin-stripe: The purpose and semantics of `canonical_id` is documented at https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download

Note that this field was originally present in the PR that added the `Downloader` interface. It got _removed_ as part of the import into Google, and I just need back, because we have propagate this data to the backend somehow.

jmillikin-stripe:
Per Dmitry's latest comment on https://github.com/bazelbuild/bazel/issues/10742, it sounds like this has missed the merge window for v3.0. I am going to take the parts that the reviewer(s) seem happy with and them into a separate PR (https://github.com/bazelbuild/bazel/pull/10914), in the hopes of trimming this one down further.

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,192 @@
...
+public class GrpcRemoteDownloader implements AutoCloseable, Downloader {
+
+  private final ReferenceCountedChannel channel;
+  private final Optional<CallCredentials> credentials;
+  private final RemoteRetrier retrier;
+  private final Context requestCtx;
+  private final RemoteCacheClient cacheClient;
+  private final RemoteOptions options;
+
+  private AtomicBoolean closed = new AtomicBoolean();
```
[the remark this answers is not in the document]
jmillikin-stripe: Done in forked PR https://github.com/bazelbuild/bazel/pull/10914

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,192 @@
...
+          Utils.getFromFuture(cacheClient.downloadBlob(blobDigest, out));
+        }
+        return null;
+      }));
+    } catch (StatusRuntimeException e) {
+      throw new IOException(e);
+    }
+  }
+
+  static FetchBlobRequest newFetchBlobRequest(
```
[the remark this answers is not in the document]
jmillikin-stripe: Done in forked PR https://github.com/bazelbuild/bazel/pull/10914

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
jmillikin-stripe: I copy-pasted this from the existing downloader test case, and don't have enough context in Bazel to know whether there's a pre-existing bug here.

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
jmillikin-stripe: Done in forked PR https://github.com/bazelbuild/bazel/pull/10914

src/test/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java
jmillikin-stripe: Done in forked PR https://github.com/bazelbuild/bazel/pull/10914

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,192 @@
...
+      final Digest blobDigest = response.getBlobDigest();
+
+      retrier.execute(() -> requestCtx.call(() -> {
+        try (OutputStream out = newOutputStream(destination, checksum)) {
+          Utils.getFromFuture(cacheClient.downloadBlob(blobDigest, out));
+        }
+        return null;
+      }));
+    } catch (StatusRuntimeException e) {
+      throw new IOException(e);
```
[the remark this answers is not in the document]
jmillikin-stripe: I haven't found a use for extra text in this particular exception site. The error message (from gRPC) is already pretty complete.

Note that I copied this from the remote cache code, so if we decide that more info is needed here, it should probably be added to `src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java` also.

katre:
Left comments on #10914. What's remaining in this PR after that? Can you give a high-level description?

jmillikin-stripe:
After that PR is merged, this one would be just the plumbing of the new downloader into `RemoteModule`, `BazelRuntime`, and so on. I'm hopeful that since I'm just copying what Bazel is already doing, I can link each new code block to the existing code and thus simplify the reviewer's task.

katre:
#10914 is now merged, do you want to rebase this and I will review the remainder?

jmillikin-stripe:
Rebased to master.

Per earlier comments, I've rewritten this final commit. Instead of using the existing design of `RemoteRepositoryRemoteExecutorFactory` as a reference, it passes around a `Supplier<Downloader>`.

I don't know if the new version is as idiomatic for Java, but I find it easier to follow. Please let me know if you'd prefer the previous way.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -336,6 +376,17 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+
+      if (enableRemoteDownloader) {
+        remoteDownloaderSupplier.set(new GrpcRemoteDownloader(
+            downloaderChannel.retain(),
+            Optional.ofNullable(credentials),
+            retrier,
+            repoContext,
+            cacheClient,
+            remoteOptions));
+        downloaderChannel.release();
```
katre: Why the immediate release after the retain?

jmillikin-stripe: Trying to match the existing code. Ctrl+F `new ByteStreamUploader(` or `new GrpcRemoteExecutor(` for where I'm looking as examples.

katre: Thanks for the explanation. I'll import this today and try to cut a new RC of 3.0.0.

jmillikin-stripe:
@katre Thank you so much for your help! Both for the review, and for handling the cherry-pick.

I no longer have Peer Bonus access, but if you're ever in Tokyo I'll buy you lunch or something :P

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -160,6 +172,13 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
     boolean enableHttpCache = RemoteCacheClientFactory.isHttpCache(remoteOptions);
     boolean enableGrpcCache = GrpcCacheClient.isRemoteCacheOptions(remoteOptions);
     boolean enableRemoteExecution = shouldEnableRemoteExecution(remoteOptions);
+    boolean enableRemoteDownloader = shouldEnableRemoteDownloader(remoteOptions);
+
+    if (enableRemoteDownloader && !enableGrpcCache) {
```
Qinusty, line 177: This seems to require `--remote_cache` be set alongside `--remote_executor` are there any side effects which could arise from changing this to the below snippet?

```java
if (enableRemoteDownloader && (!enableGrpcCache || !enableRemoteExecution)) {
```

jmillikin: Remote execution is not required to use remote downloads. Remote caching (specifically gRPC remote caching) _is_ required, because that's where the downloads get placed.

Qinusty: I've opened https://github.com/bazelbuild/bazel/issues/11913 to track this issue.

purkhusid:
@jmillikin Is there any way to invalidate the Remote Downloader cache? I've tried using `--remote_instance_name` but it does not seem to affect the Remote Downloader cache.

jmillikin:
I think that would be a question for the author of your remote downloader implementation. The Bazel client only uses one RPC, which is approximately "download file and return the CAS checksum + size". Other RPCs for cache management and invalidation are implementation-specific.

purkhusid:
@jmillikin Ok, but I wonder if it might be worth a feature request to make the Remote Assets API sensitive to the `--remote_instance_name` flag? That could be implemented as a client side feature, right?

Issue #10574, opened 2020-01-13
add `label_kind` to `--output` for cquery


ahumesky:
`bazel cquery` with `--output label_kind` gives

    ERROR: Invalid output format 'label_kind'. Valid values are: label, transitions, proto, textproto, build

gregestren:
Contributions welcome. I expect this'd be easy by simply adding a new entry here:

https://github.com/bazelbuild/bazel/blob/605eaccabe2fed0a1d188d77baa2206f6346becd/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java#L207-L209

i.e. clone and modify this:

https://github.com/bazelbuild/bazel/blob/966dc2328a8565103e514f898673fd8eb037b9c0/src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java

amanbora:
Hi, I would like to work on this issue, since I'm new to bazel i might need some guidance in the future.

gregestren:
Hi Aman,

Happy to hear!. Feel free to reach out to me at any point and I'll offer help as I can (here or gregce@bazel.build).

amanbora:
I have updated 'label' to 'label_kind' in [https://github.com/bazelbuild/bazel/blob/966dc2328a8565103e514f898673fd8eb037b9c0/src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java](url),

and added a new entry in [https://github.com/bazelbuild/bazel/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java](url)

gregestren:
Great! I'll respond more on https://github.com/bazelbuild/bazel/pull/10632.

PR #10632, opened 2020-01-22
added 'label_kind' to '--output' for  cquery 

amanbora:
https://github.com/bazelbuild/bazel/issues/10574

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
@@ -220,8 +220,6 @@ public class ConfiguredTargetQueryEnvironment
     return ImmutableList.of(
         new LabelAndConfigurationOutputFormatterCallback(
             eventHandler, cqueryOptions, out, skyframeExecutor, accessor),
-        new LabelKindAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, getAccessor()),
         new TransitionsOutputFormatterCallback(
             eventHandler,
             cqueryOptions,
@@ -488,4 +486,3 @@ public class ConfiguredTargetQueryEnvironment
         SkyQueryEnvironment.DEFAULT_THREAD_COUNT);
   }
 }
-
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
@@ -209,7 +209,6 @@ public class ConfiguredTargetQueryEnvironment
   public ImmutableList<NamedThreadSafeOutputFormatterCallback<ConfiguredTarget>>
       getDefaultOutputFormatters(
           TargetAccessor<ConfiguredTarget> accessor,
-          ConfiguredTargetAccessor confiredTargetAccessor,
           ExtendedEventHandler eventHandler,
           OutputStream out,
           SkyframeExecutor skyframeExecutor,
@@ -220,9 +219,9 @@ public class ConfiguredTargetQueryEnvironment
         cqueryOptions.aspectDeps.createResolver(packageManager, eventHandler);
     return ImmutableList.of(
         new LabelAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, true),
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, true),
         new LabelAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, false),
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, false),
         new TransitionsOutputFormatterCallback(
             eventHandler,
             cqueryOptions,
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
@@ -209,6 +209,7 @@ public class ConfiguredTargetQueryEnvironment
   public ImmutableList<NamedThreadSafeOutputFormatterCallback<ConfiguredTarget>>
       getDefaultOutputFormatters(
           TargetAccessor<ConfiguredTarget> accessor,
+          ConfiguredTargetAccessor confiredTargetAccessor,
           ExtendedEventHandler eventHandler,
           OutputStream out,
           SkyframeExecutor skyframeExecutor,
@@ -219,9 +220,9 @@ public class ConfiguredTargetQueryEnvironment
         cqueryOptions.aspectDeps.createResolver(packageManager, eventHandler);
     return ImmutableList.of(
         new LabelAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, true),
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, true),
         new LabelAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, false),
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, false),
         new TransitionsOutputFormatterCallback(
             eventHandler,
             cqueryOptions,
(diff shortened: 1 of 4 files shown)


amanbora:
@googlebot I signed it!

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
```diff
@@ -220,6 +220,8 @@ private static BuildConfiguration mergeEqualBuildConfiguration(
     return ImmutableList.of(
         new LabelAndConfigurationOutputFormatterCallback(
             eventHandler, cqueryOptions, out, skyframeExecutor, accessor),
+        new LabelKindAndConfigurationOutputFormatterCallback(
```
gregestren: Does your PR included a file that defines this class?

amanbora:
I added the declaration just like the one before it.

gregestren:
Have you tried building bazel and trying `bazel cquery` with this change?

amanbora:
No, I'll build it then and report

amanbora:
I have tried to build adding some more commits to it, having reference from query.
Screenshot 2020-01-23 at 12 29 02 AM

This is the error I'm getting. `getTargetKind` needs to be defined and another boolean field needs to be added to `CqueryThreadsafeCallback`

gregestren:
The problem is the `cquery` version is a [`ConfiguredTarget`](https://github.com/bazelbuild/bazel/blob/11e5fab47acdc1ce5c6940a141335fc30e31d2e0/src/main/java/com/google/devtools/build/lib/analysis/ConfiguredTarget.java) while the `query` version is a [`Target`](https://github.com/bazelbuild/bazel/blob/1f684e1b87cd8881a0a4b33e86ba66743e32d674/src/main/java/com/google/devtools/build/lib/packages/Target.java). Only the latter has `getTargetKind()`.

To get the target from the configured target, you need to also pass a `ConfiguredTargetAccessor` to the constructor. That has [`getTargetFromConfiguredTarget`](https://github.com/bazelbuild/bazel/blob/f2cb3c69413c74569f19232f89a5f0e86d4f15ac/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java#L162), which can get you back to the corresponding `Target`.

gregestren:
Aman, I noticed you pushed some more changes but the build isn't yet compiling (see PR checks). I'm not you're still working on it, so just let me know when you're ready for the next review round or if you need more advice.

amanbora:
@gregestren I'm working on it, I could definitely use some advice
Screenshot 2020-01-23 at 11 26 14 PM
I'm currently getting this error

src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java
```diff
@@ -24,33 +24,46 @@
...
       ExtendedEventHandler eventHandler,
       CqueryOptions options,
       OutputStream out,
       SkyframeExecutor skyframeExecutor,
-      TargetAccessor<ConfiguredTarget> accessor) {
-    super(eventHandler, options, out, skyframeExecutor, accessor);
+      TargetAccessor<ConfiguredTarget> accessor,
+      ConfiguredTargetAccessor configuredTargetAccessor,
+      boolean showKind) {
+    super(eventHandler, options, out, skyframeExecutor, accessor, configuredTargetAccessor, showKind);
```
gregestren: Ok, so what I'm reading from your last error report is compilation fails because the parameters you're passing to CqueryThreadsafeCallback's constructor are different than what the constructor expects:  https://github.com/bazelbuild/bazel/blob/ebca24c525ec93f34bb65053370376082765d800/src/main/java/com/google/devtools/build/lib/query2/cquery/CqueryThreadsafeCallback.java#L48-L53

The link above doesn't know what to do with an extra `ConfiguredTargetAccessor` argument.

But I just realized something that can make this all simpler! You're already accepting a `TargetAccessor<ConfiguredTarget> accessor` above, and `ConfiguredTargetAccessor` is already an instance of `TargetAccessor`!

https://github.com/bazelbuild/bazel/blob/f2cb3c69413c74569f19232f89a5f0e86d4f15ac/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java#L56

So you don't have to pass the new `ConfiguredTargetAccessor configuredTargetAccessor` parameter at all (my bad). Instead, simply call `accessor.getTargetKind()` in `processOutput` below. And there's no need to change anything  here in the constructor.

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
```diff
@@ -219,7 +220,9 @@ private static BuildConfiguration mergeEqualBuildConfiguration(
         cqueryOptions.aspectDeps.createResolver(packageManager, eventHandler);
     return ImmutableList.of(
         new LabelAndConfigurationOutputFormatterCallback(
-            eventHandler, cqueryOptions, out, skyframeExecutor, accessor),
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, true),
+        new LabelAndConfigurationOutputFormatterCallback(
+            eventHandler, cqueryOptions, out, skyframeExecutor, accessor, confiredTargetAccessor, false),
```
gregestren: Following my longer comment: you don't need `configuredTargetAccessor` here or in the above line.

src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java
```diff
@@ -24,33 +24,46 @@
...
-          new StringBuilder()
-              .append(configuredTarget.getLabel())
-              .append(" (")
-              .append(config != null && config.isHostConfiguration() ? "HOST" : config)
-              .append(")");
+      StringBuilder output;
+      if(showKind){
+            output =
+            new StringBuilder()
+                .append(configuredTargetAccessor.getTargetFromConfiguredTarget(configuredTarget))
```
gregestren: Following my longer comment: just change this to `accessor.getTargetKind(configuredTarget)` and I think it'll just work.

amanbora:
i was just doing what you suggested right now, deja vu!
The only error I'm getting is that I can't pass showKind (boolean) as it violates the `CqueryThreadsafeCallback` definition

src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java
```diff
@@ -18,39 +18,53 @@
...
   public LabelAndConfigurationOutputFormatterCallback(
       ExtendedEventHandler eventHandler,
       CqueryOptions options,
       OutputStream out,
       SkyframeExecutor skyframeExecutor,
-      TargetAccessor<ConfiguredTarget> accessor) {
-    super(eventHandler, options, out, skyframeExecutor, accessor);
+      TargetAccessor<ConfiguredTarget> accessor,
+      boolean showKind) {
+    super(eventHandler, options, out, skyframeExecutor, accessor, showKind);
```
gregestren: I don't think you need `showKind` on this line: since it's private to  `LabelAndConfigurationOutputFormatterCallback`, `CqueryThreadsafeCallback` doesn't know (or need to know) it exists.

But you do want to set the private final field again, i.e. `this.showKind = showKind`.

amanbora:
I have committed all the necessary changes, build is successful, you can suggest any other necessary changes.

src/main/java/com/google/devtools/build/lib/query2/cquery/LabelAndConfigurationOutputFormatterCallback.java
```diff
@@ -18,39 +18,55 @@
...
           skyframeExecutor.getConfiguration(eventHandler, configuredTarget.getConfigurationKey());
-
-      StringBuilder output =
-          new StringBuilder()
-              .append(configuredTarget.getLabel())
-              .append(" (")
-              .append(config != null && config.isHostConfiguration() ? "HOST" : config)
-              .append(")");
+      StringBuilder output = new StringBuilder();
+      if(showKind){
```
gregestren: Since most of the lines here and in the other branch are repeated, what if `if (showKind)` just sets some string `result` (or some similar name)? Then you only have to write:

    ...
    append(result)

and all those other lines once.

src/test/shell/integration/configured_query_test.sh
```diff
@@ -90,6 +90,20 @@ EOF
...
+  local -r pkg=$FUNCNAME
+  mkdir -p $pkg
+  cat > $pkg/BUILD <<EOF
+sh_library(name='maple', deps=[':japanese'])
+sh_library(name='japanese')
+EOF
+
+ bazel cquery --output=label_kind "deps(//$pkg:maple)" > output 2>"$TEST_log" || fail "Expected success"
+
+ assert_contains "name: \"//$pkg:maple\"" output
```
gregestren: 99% of the way there. But this and the line below are going to fail because it's not actually outputting `//$pkg:maple` now, which is what the `label_kind` option does, right?

From the red test results attached to this PR:

https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/270d91f7-5738-4269-a656-c3c98f7c3550/src/test/shell/integration/configured_query_test/attempt_1.log

You can see the actual output looks like:

    sh_library rule 
    platform rule
    constraint_value rule
    ...

amanbora: Yea the task what we wanted is being achieved before only. We dont require these two lines, right?

PR #10634, opened 2020-01-22
Intercept capabilities and uploader requests and add custom grpc headers

AlessandroPatti:
Following #10015. Some requests do not use the custom headers.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
@@ -38,6 +38,8 @@ import io.grpc.CallCredentials;
 import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ClientInterceptors;
 import io.grpc.Context;
 import io.grpc.Metadata;
 import io.grpc.Status;
@@ -87,6 +89,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
 
   @GuardedBy("lock")
   private final Map<HashCode, ListenableFuture<Void>> uploadsInProgress = new HashMap<>();
+  private final ClientInterceptor[] interceptors;
 
   @GuardedBy("lock")
   private boolean isShutdown;
@@ -108,7 +111,8 @@ class ByteStreamUploader extends AbstractReferenceCounted {
       ReferenceCountedChannel channel,
       @Nullable CallCredentials callCredentials,
       long callTimeoutSecs,
-      RemoteRetrier retrier) {
+      RemoteRetrier retrier,
+      ClientInterceptor... interceptors) {
     checkArgument(callTimeoutSecs > 0, "callTimeoutSecs must be gt 0.");
 
     this.instanceName = instanceName;
@@ -116,6 +120,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
     this.callCredentials = callCredentials;
     this.callTimeoutSecs = callTimeoutSecs;
     this.retrier = retrier;
+    this.interceptors = interceptors;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


elianuber:
Thanks Ale for patching this bug !
Maybe @buchgr can you give a quick review ?

EdSchouten:
Friendly ping!

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -87,6 +89,7 @@
 
   @GuardedBy("lock")
   private final Map<HashCode, ListenableFuture<Void>> uploadsInProgress = new HashMap<>();
+  private final ClientInterceptor[] interceptors;
```
ulfjack: I'd prefer to use ImmutableList here instead of array. One problem with using arrays is that anyone with a reference to the array can modify the contents asynchronously. Technically, ensuring the safety of this code requires reviewing all the callers to check if they're doing anything sinister. Even if it's unlikely that anyone would do so intentionally, it wouldn't be obvious if anyone is doing it accidentally.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -218,7 +218,10 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
               (execChannel != null ? execChannel : cacheChannel),
               credentials,
               remoteOptions.remoteTimeout,
-              retrier);
+              retrier,
+              (execChannel != null ?
+                  TracingMetadataUtils.newExecHeadersInterceptor(remoteOptions) :
```
ulfjack: I find it slightly odd that this has different options for caching and execution. If someone is asking Bazel to do one or the other, they can also provide different options, no?

(I realize that's a pre-existing condition, so feel free to ignore this comment.)

src/test/java/com/google/devtools/build/lib/remote/RemoteServerCapabilitiesTest.java
```diff
@@ -103,6 +105,10 @@ public static void afterEverything() {
       assertThat(meta.getToolDetails().getToolName()).isEqualTo("bazel");
       assertThat(meta.getToolDetails().getToolVersion())
           .isEqualTo(BlazeVersionInfo.instance().getVersion());
+      assertThat(headers.get(Metadata.Key.of("Key1", Metadata.ASCII_STRING_MARSHALLER)))
```
ulfjack: Having checks in the common case means that any future test case is forced to also test this by setting up custom remoteOptions. That seems like a bad direction, generally speaking.

AlessandroPatti: Right, I'll create a new test case for it.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
AlessandroPatti: Changed this one to an `ImmutableList` as suggested.
There was also a similar bit in `RemoteServerCapability`, for which I opted for making a copy of the array instead, as it makes it easier to pass it directly to the `withInterceptors` method, that only accepts an array.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
AlessandroPatti: Updated to run the checks twice if the endpoints differ.

AlessandroPatti:
> Do you know why it wasn't done this way when the extra header options were added to Bazel? Just an oversight?

Yes, just an oversight.

ulfjack:
Looks good to me.

@buchgr @meisterT

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -108,14 +113,16 @@ public ByteStreamUploader(
       ReferenceCountedChannel channel,
       @Nullable CallCredentials callCredentials,
       long callTimeoutSecs,
-      RemoteRetrier retrier) {
+      RemoteRetrier retrier,
+      ClientInterceptor... interceptors) {
```
buchgr: Did you consider passing in a channel that already has the interceptors attached instead? It seems preferable to me to not add knowledge about interceptors to the ByteStreamUploader if not necessary.

AlessandroPatti: Updated

PR #10638, opened 2020-01-23
Improve mechanism to distribute precompiled jars (e.g. for bootstrapping)

Yannic:
This change adds a new directory `derived/jars` to `bazel-distfile.{tar,zip}` that contains jars that we can't (or don't want to) build during bootstrapping (e.g. the Protobuf runtime).
Delaying the compilation of these jars until we create the distfile archive removes the need to check-in these precompiled jars.

This should allow us to replace some of the checked-in third-party repositories with `http_archive`.


Yannic:
@philwo I think this could be split into a third-party and a non-third-party commit without breaking CI.

@lberki fyi^

philwo:
@Yannic Thanks! If you can easily split this up into two separate PRs, I'll make sure to get them in asap. :)

Yannic:
@philwo Thank you! Third_party parts are in https://github.com/bazelbuild/bazel/pull/10703 now.

philwo:
Merging this now.

PR #10567, opened 2020-01-13
Add remaining changes for bootstrap build on OpenBSD.

aldersondrive:
I have tested these changes via a bootstrap build on OpenBSD 6.6-current (amd64 architecture). The resulting `bazel` binary appeared to work in some simple testing involving `cc_library`, `cc_binary`, `java_library`, and `java_binary`.

Known issues/limitations:

- Building a `java_binary`'s deploy JAR fails because the `singlejar` tool fails to build. I intend to investigate soon.

- Running Bazel requires a `--host_javabase=@local_jdk//:jdk` flag.

- Sandboxing is unsupported.

This change, split out of the larger PR https://github.com/bazelbuild/bazel/pull/10274, is part of the OpenBSD port in https://github.com/bazelbuild/bazel/issues/10250. Most of the changes in that larger PR have already been committed via several smaller PRs (see the PRs linked from https://github.com/bazelbuild/bazel/issues/10250). This PR collects the remaining changes.

PR #10639, opened 2020-01-23
Tweak third_party/BUILD for OpenBSD.

aldersondrive:
This change was part of https://github.com/bazelbuild/bazel/pull/10567, which was merged as commit https://github.com/bazelbuild/bazel/commit/d6145a0836e5856e5b6183bd484c4a6a3f6927a0, but somehow that commit contains all of the changes in the PR *except* the changes to `third_party/BUILD`. Apparently the changes to that file were lost when the PR was imported into Google's internal code repository. Here is another PR for those changes that disappeared.

@aehlig FYI.

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -469,6 +469,7 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
+    "//src/conditions:openbsd": "*.so *.jnilib *.dll",
     # Play it safe -- better have a big binary than a slow binary
     # zip -d does require an argument. Supply something bogus.
     "//conditions:default": "*.bogusextension",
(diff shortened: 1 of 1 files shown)


third_party/BUILD
```diff
@@ -469,6 +469,7 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
+    "//src/conditions:openbsd": "*.so *.jnilib *.dll",
```
jmmv: Any idea why FreeBSD doesn't show up here? It's surprising.

aldersondrive: FreeBSD's Bazel port does [patches `third_party/BUILD`](https://github.com/freebsd/freebsd-ports/blob/f5c3ba467a9cc5e834c157b0364d327425981c24/devel/bazel/files/patch-third_party__BUILD) instead.

I considered adding the following line also:
```
    "//src/conditions:freebsd": "*.so *.jnilib *.dll",
```
.. but haven't so far, to avoid mixing up unrelated changes in a single commit and because @aehlig said nothing about this issue earlier. I can certainly add this tweak if you folks like, though.

aldersondrive: In the absence of anyone expressing an opinion about this, I'll go ahead and close this thread. Feel free to reopen it if you'd like this to change in this PR.

philwo:
Merged in e274da343cecc2507688552d5fc13170906e89ce.

Issue #10457, opened 2019-12-19
C++ Coverage on MacOS is completely broken


ulfjack:
Coverage with gcov:

1. GCOV_PREFIX_STRIP is set incorrectly
With a value of 3, the paths end up like this:
bazel-testlogs/cpp/foo_test/_coverage/_bazel_ulfjack/312a82d9f5d7857ee9e5303755329142/sandbox/darwin-sandbox/4/execroot/__main__/bazel-out/darwin-fastbuild/bin/cpp/_objs/foo_test/main.gcda

The collect-coverage.sh script doesn't allow overriding GCOV_PREFIX_STRIP. I have a patch to not set it if it's already set in the test env. On this machine, I have to set it to 10.

2. The latest gcov does not support the -i flag
The -i flag is hard-coded in collect_cc_coverage.sh with no override possible.

3. The gcov log uses lower-case "creating"
The collect_cc_coverage.sh only checks for upper-case "Creating".

4. The generated output is not in intermediate gcov format as expected by lcov_merger.

Coverage with llvm-profdata (--features=llvm_coverage_map_format):

1. COVERAGE_GCOV_PATH is set to gcov
This is obviously wrong. It needs to be set to llvm-profdata. However, llvm-profdata is not on the path, and there does not seem to be a way to override the location of gcov. I tried setting GCOV=/path/to/llvm-profdata before calling Bazel, but that had no effect.

2. The lcov_merger simply copies the .profdata file to coverage.dat
At least this allows users to post-process manually, but it's still unfortunate.

jeffwheeler:
For anyone else that tries to run coverage on Mac, try putting this in your `~/.bazelrc` file:
```
coverage --action_env=GCOV_PREFIX_STRIP=10
```

Thanks to @ulfjack, it is possible to set the correct prefix size through this argument. It'd be nice to fix this bug in Bazel.

ulfjack:
It bugs me that you have to guess the correct value for `GCOV_PREFIX_STRIP` on MacOS in order to use it. I looked into it, but so far haven't been able to come up with an automated method.

ulfjack:
IIUC, the gcno and gcda files have to be next to each other, which is why Bazel copies them into the same tree; however, matching up the names in shell is not straightforward to do in O(n) time.

josh-swerdlow:
Is there any update on this? I'm still hoping I can get coverage reports on my mac using bazel instead of switching to some other compiler mechanism (as done in the aforementioned link).

[...]

PR #10640, opened 2020-01-23
Partially fix MacOS llvm coverage collection

ulfjack:
This allows setting GCOV on the command line to overwrite the default setting, which points to /usr/bin/gcov. In order to use this, you also need to disable the gcov coverage feature and enable the llvm coverage feature instead. The full command-line looks like this:

GCOV=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/llvm-profdata bazel coverage --features=llvm_coverage_map_format --features=-gcc_coverage_map_format //cpp:test

Progress on #10457.

Change-Id: Iae80e3a5830711b6402a3e015bf9ae12b536f429

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -139,6 +139,13 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
                  "https://github.com/bazelbuild/bazel/issues with the following:\n" +
                  error_msg)
 
+        tool_paths = {}
+        gcov_path = repository_ctx.os.environ.get("GCOV")
+        if gcov_path != None:
+            if not gcov_path.startswith("/"):
+                gcov_path = repository_ctx.which(gcov_path)
+            tool_paths["gcov"] = gcov_path
+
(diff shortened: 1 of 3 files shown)


keith:
IMO ideally this should be discovered based on the currently passed --xcode_version. I think this change would mean it would be too early for that. Would that be possible some other way?

ulfjack:
As far as I know, xcode supports both gcov and llvm-cov modes, so the user has to decide which one to pick. You're right that the toolchain should automatically detect the right location of the binaries, and I think that this change will help with that. I'm not sure yet how the user can decide which mode to use, though.

keith:
I more mean that right now your example has GCOV set to an absolute path, where really it should probably be set to `xcrun llvm-profdata` and then that would be picked up based on `DEVELOPER_DIR` when it was actually run, but right now it could use the `llvm-profdata` executable from a different version of Xcode than the one specified with `--xcode_version`

ulfjack:
Ah, yes. I think my patch won't make it harder to do that in the future, and it adds a little bit of required infrastructure. Unfortunately, I'm not going to be able to spend more time on this.

ulfjack:
And it got rolled back.

PR #9945, opened 2019-10-07
third_party condition for ppc64le dyn lib

clnperez:
Pick up the `third_party` portion of PR #9346

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

Issue #9263, opened 2019-08-27
Failure builing on ppc64le: Executing genrule filter_netty_dynamic_libs


clnperez:
### Description of the problem / feature request:

I can no longer build bazel on Power. I'm trying to use the dist bundle and building version 0.28.1. I'm not sure what this zip error is all about: 
`zip error: Nothing to do! (bazel-out/ppc-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar) `

```
1,242 / 1,750][Prepa] ...//:google_bytestream_bytestream_java_grpc_srcs 
SUBCOMMAND:  //third_party:filter_netty_dynamic_libs [action 'Executing genrule //third_party:filter_netty_dynamic_libs'] 
 (cd /tmp/bazel_xA2x7yX3/out/execroot/io_bazel && \ 
    exec env - \                                                                                                                                                                                    
    PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \                                                                                                                           
  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cp third_party/netty_tcnative/netty-tcnative-boringssl-static-2.0.24.Final.jar bazel-out/ppc-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar && zip -qd bazel-out/ppc-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar *.bogusextension')                                                  
[1,243 / 1,750] [Prepa] //third_party:filter_netty_dynamic_libs                                                                                                                        
ERROR:/bazel/third_party/BUILD:535:1: Executing genrule //third_party:filter_netty_dynamic_libs failed (Exit 12)
[1,244 / 1,750] checking cached actions
[1,244 / 1,750]  checking cached actions
zip error: Nothing to do! (bazel-out/ppc-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar)    Target //src:bazel_nojdk failed to build                                                                                                                                          
                                                                                                                                  
ERROR: Could not build Bazel
[...]

PR #10643, opened 2020-01-24
Correct failure of netty build component on s390x (due to zip -qd needing something to delete)

jonpspri:
Build on s390x fails because the process to remove unnecessary libraries from netty was not configured for x86.  As a result, zip -d failed because it had nothing to remove.  This PR does two things:

1.  Adds conditions for s390x to the src and third_party BUILD files, and
2.  Adds a condition to the genrule so that 'zip -d' does not execute if there's nothing to do.

See also Issue #9263; Pull Requests #9346 and #9945

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -469,9 +469,9 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
+    "//src/conditions:linux_s390x": "*.so *.jnilib *.dll",
     # Play it safe -- better have a big binary than a slow binary
-    # zip -d does require an argument. Supply something bogus.
-    "//conditions:default": "*.bogusextension",
+    "//conditions:default": "",
 })
 
 # Remove native libraries that are for a platform different from the one we are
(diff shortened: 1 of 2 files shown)


third_party/BUILD
```diff
@@ -469,9 +469,9 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
+    "//src/conditions:linux_s390x": "*.so *.jnilib *.dll",
     # Play it safe -- better have a big binary than a slow binary
-    # zip -d does require an argument. Supply something bogus.
-    "//conditions:default": "*.bogusextension",
+    "//conditions:default": "",
```
gregestren: This fulfills point #2, right?

jonpspri: Correct.  The empty string means nothing is removed.  The test for the empty string in the shell execution keeps unzip from being run as a no-op (causing a failure).

gregestren: Thanks for clarifying!

gregestren:
Jonathan,

It looks like we have to follow a different merge process for patches that include `third_party`. I can merge `src/conditions/BUILD` in this PR, then we'll need a separate PR just for `third_party/BUILD`.

Does that ordering sequence work and do you mind setting up the PR structure that way?

jonpspri:
@gregestren Yes, that works.  Go ahead and merge `src/conditions/BUILD` from this PR.  Once it's in place and percolated through, I'll open a new PR for `third_party/BUILD`.  It looks like there's a conflict there to resolve anyway.

Do I need to undo the third_party/BUILD changes in this branch for that to work?

gregestren:
I believe it's okay as-is. I'll try to merge - I'll add a comment here if anything doesn't work.

gregestren:
This patch is in - ready for third_party/BUILD when you are.

jonpspri:
@gregestren Perfect.  I'll rebase and submit a new pull request later today.  Kinda busy doing my "real job" this morning.

Issue #10621, opened 2020-01-21
`bazel run` before `bazel test` fails with `Could not chdir` on Windows


gregmagolan:
### Description of the problem / feature request:

On Windows, if you run a test target with `bazel run` before you run that same test with `bazel test` (after a clean) then the test will fail with `Could not chdir` on Windows.

This was originally reported in #7387 and marked as resolved so it may have regressed as I'm seeing this in Bazel 2.0.0.

The culprit is that under Windows with runfiles not enabled by default, `bazel run` does not create the root workspace directory in the runfiles directory (`/path/to/test.runfiles/my_workspace_name`) while `bazel test` does. If you run `bazel test` first then `bazel run` will succeed for that test from then on as that directory is present under the test runfiles.

The error happens with both the default `--incompatible_windows_native_test_wrapper` and `--noincompatible_windows_native_test_wrapper` (although the error output is slightly different):

```
-----------------------------------------------------------------------------
ERROR(tools/test/windows/tw.cc:462) value: 2 (0x00000002), arg: c:\users\gmago\_bazel~1\wpufrwit\execroot\build_~1\bazel-out\x64_windows-fastbuild\bin\internal\linker\test\unit_tests.bat.runfiles\build_bazel_rules_nodejs: Could not chdir
```
in the first case, and

```
external/bazel_tools/tools/test/test-setup.sh: line 147: cd: /c/users/gmago/_bazel~1/wpufrwit/execroot/build_~1/bazel-out/x64_windows-fastbuild/bin/internal/linker/test/unit_tests.bat.runfiles/build_bazel_rules_nodejs: No such file or directory
Could not chdir /c/users/gmago/_bazel~1/wpufrwit/execroot/build_~1/bazel-out/x64_windows-fastbuild/bin/internal/linker/test/unit_tests.bat.runfiles/build_bazel_rules_nodejs
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

On Windows with Bazel 2.0.0, run any test target with `bazel run` on a clean clone or after a bazel clean before running that test target with `bazel test`.

### What operating system are you running Bazel on?

Windows 10 Pro

### What's the output of `bazel info release`?

```
C:\Users\gmago\google\rules_nodejs>bazel info release
INFO: Invocation ID: 1ecaee15-b295-461c-9656-0c5fe8418002
release 2.0.0
```

### Any other information, logs, or outputs that you want to share?

[...]

PR #10646, opened 2020-01-24
Fix bazel run //foo:test on Windows

laszlocsomor:
"bazel run" will now create a <target>.runfiles
directory even, even if --enable_runfiles=no
(which is the default on Windows but not
elsewhere).

Reason is that "bazel run //foo:test" now works,
because the test expects it can chdir into the
runfiles directory. Previously "bazel run" would
not create this directory (when
--enable_runfiles=no) and the chdir call failed.

Fixes https://github.com/bazelbuild/bazel/issues/10621


laszlocsomor:
Thank you!

PR #10648, opened 2020-01-24
Stop using //third_party/protobuf/3.6.1 in //src

Yannic:


philwo:
If this entire PR passes CI (which seems fine), I can perform a so called manual atomic  override merge. It’s not without risks... but nobody said the job would be easy, right.

@Yannic It’s still marked as draft though - is this ready to be merged?

Yannic:
@philwo It's marked as draft because it contains commits from #10638 (which adds the logic to include pre-built jars in the bootstrap archive).

Maybe let's start by merging #10638 and then I'll rebase this one so you don't have to do a manual atomic override merge.

Yannic:
@philwo Thanks for merging the dependencies of this yesterday. I've also split this into a `//src` change and a `//third_party` change. ptal.

PR #10649, opened 2020-01-24
Fix races in CanAccessDirectory

nikola-sh:
If several bazel processes are starting concurrently some of them can fail to start with error
```Output base directory ... must be readable and writable.```

It happens because CanAccessDirectory doesn't handle `NTSTATUS` `STATUS_DELETE_PENDING` for file.
`STATUS_DELETE_PENDING` occured if `DeleteFile` finished successfully but there are open handles for the file.
`STATUS_DELETE_PENDING` is converted to `ERROR_ACCESS_DENIED` on Win32 level so client code can't distinguish pending deletion from real denied access.

To workaround this problem we can create check-file with unique name, for example with current thread id.

P.S.
This problem is reproducible on Windows versions prior to Windows 10 19H1.

### DIFF
--- a/src/main/cpp/util/file_windows.cc
+++ b/src/main/cpp/util/file_windows.cc
@@ -22,6 +22,7 @@
 #include <memory>  // unique_ptr
 #include <sstream>
 #include <vector>
+#include <string>
 
 #include "src/main/cpp/util/errors.h"
 #include "src/main/cpp/util/exit_code.h"
@@ -534,7 +535,10 @@ bool CanAccessDirectory(const Path& path) {
 
   // The only easy way to know if a directory is writable is by attempting to
   // open a file for writing in it.
-  Path dummy_path = path.GetRelative("bazel_directory_access_test");
+  std::string dummy_name = "bazel_directory_access_test_";
+  dummy_name += std::to_string(::GetCurrentThreadId());
+
+  Path dummy_path = path.GetRelative(dummy_name);
 
   // Attempt to open the dummy file for read/write access.
   // If the file happens to exist, no big deal, we won't overwrite it thanks to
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/main/cpp/util/file_windows.cc [hunk 1]
```diff
@@ -548,9 +552,9 @@ bool CanAccessDirectory(const Path& path) {
       /* dwFlagsAndAttributes */ FILE_ATTRIBUTE_NORMAL,
       /* hTemplateFile */ NULL);
   DWORD err = GetLastError();
-  if (handle == INVALID_HANDLE_VALUE && err != ERROR_ALREADY_EXISTS) {
```
nikola-sh: We should respect any error here, `handle == INVALID_HANDLE_VALUE && err==ERROR_ALREADY_EXISTS` is incorrect invariant for `OPEN_ALWAYS`.

laszlocsomor: Why incorrect? CreateFileW's docs says: "If the specified file exists, the function succeeds and the last-error code is set to ERROR_ALREADY_EXISTS (183)."

nikola-sh: Yes, but if function succeeds handle != INVALID_HANDLE_VALUE

laszlocsomor: Ah, of course.

src/main/cpp/util/file_windows.cc
```diff
@@ -534,7 +535,10 @@ bool CanAccessDirectory(const Path& path) {
 
   // The only easy way to know if a directory is writable is by attempting to
   // open a file for writing in it.
-  Path dummy_path = path.GetRelative("bazel_directory_access_test");
+  std::string dummy_name = "bazel_directory_access_test_";
+  dummy_name += std::to_string(::GetCurrentThreadId());
```
laszlocsomor: Please add a brief comment here saying that the Thread ID avoids races among concurrent Bazel processes.

nikola-sh: @laszlocsomor can you restart checks? They are failed on platforms that I didn't change.

laszlocsomor:
Don't worry, I know the tests aren't your PR's fault.
I already imported the PR, just waiting for internal review.

PR #10651, opened 2020-01-24
Update Protobuf to 3.11.3

Yannic:


lberki:
(I'll hold off reviewing this until https://github.com/bazelbuild/bazel/pull/10648 lands)

philwo:
I think all dependencies of this are merged now!

Yannic:
We're mostly there! #10723 adds a patch for Protobuf to use the existing references for zlib, six, guava, ... When this landed, I'll rebase this PR (already verified locally that it works), and as a last change we get to delete the vendored version of Protobuf 😃.

philwo:
Hi @Yannic,

just saw the failed tests here with this message:

```
ERROR: Analysis of target '//a:c' failed; build aborted: no such package '@bazel_skylib//lib': java.io.IOException: Error downloading [https://github.com/bazelbuild/bazel-skylib/archive/0.9.0.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/2733/execroot/io_bazel/_tmp/53cf69ace19a0796fc4537030d753e89/root/917143f2cfdfb15433399b1574e0bac0/external/bazel_skylib/0.9.0.tar.gz: Unknown host: github.com
```

You can fix this with a little tweak:
- Add `@bazel_skylib//:WORKSPACE` to the `test_repos` filegroup in [src/BUILD](https://source.bazel.build/bazel/+/9ec71817f9b7750300c012f9a9dfa1ce7409ddcb:src/BUILD;l=725).
- Add `bazel_skylib` to the `repos` array in [testenv.sh](https://source.bazel.build/bazel/+/9ec71817f9b7750300c012f9a9dfa1ce7409ddcb:src/test/shell/testenv.sh;l=288).

Then it should work!

Yannic:
@philwo Thanks! That fixed the skylib error. Unfortunately, there is a new error because of zlib being referenced as `@io_bazel//third_party/zlib` and I can't seem to figure out how to fix that.

philwo:
Mhm... yeah, I'm not sure how to solve that. What's the reason that we're patching the references to @zlib and @six in protobuf's BUILD file via https://github.com/bazelbuild/bazel/blob/master/third_party/protobuf/3.11.3.patch?

If we change this: https://github.com/bazelbuild/bazel/blob/master/WORKSPACE#L59-L62 to:
```starlark
local_repository(
    name = "six",
    path = "./third_party/py/six/",
)
```

and add one for zlib:
```starlark
local_repository(
    name = "zlib",
    path = "./third_party/zlib/",
)
```

Maybe it works?

Yannic:
Finally figured it out. I think the problem was actually that bash didn't like the use of \` in a comment.

Also, I'm not sure why, but adding `local_repository(name = "zlib, path = "...")` prohibits from building `//third_party/zlib` (did you mean `@zlib//`), while `new_local_repository(name = "zlib, path = "...", build_file = ".../BUILD")` doesn't. I find that very confusing.

philwo:
@Yannic We're done!! 🎉 

Thank you so much for this contribution and your hard work through the little issues and blockers :)

PR #7860, opened 2019-03-27
Wrap StatusRuntimeExceptions from GrpcRemoteCache

werkt:
Exceptions that occur during remote interactions are expected to be
wrapped in IOException for observation by the RemoteSpawn{Runner,Cache}
layers.

Fixes #7856

PR #10663, opened 2020-01-27
Wrap FindMissingBlobs exceptions in IOException

werkt:
Consistent with #7860, present an IOException with context to ensure
fallback if configured, rather than an unhandled RuntimeException and
bazel crash, when issuing FindMissingBlobs requests.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
@@ -203,7 +203,7 @@ public class GrpcCacheClient implements RemoteCacheClient, MissingDigestsFinder 
             MoreExecutors.directExecutor());
     return Futures.catchingAsync(
         success,
-        Exception.class,
+        RuntimeException.class,
         (e) -> Futures.immediateFailedFuture(
             new IOException(
                 String.format(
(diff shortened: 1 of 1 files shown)


werkt:
Can someone review this, or provide some feedback about these status tags, specifically WIP?

src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
```diff
@@ -201,6 +201,17 @@ public static boolean isRemoteCacheOptions(RemoteOptions options) {
               return result.build();
             },
             MoreExecutors.directExecutor());
+    return Futures.catchingAsync(
+        success,
+        Exception.class,
```
ulfjack: I'd make this more narrow, e.g., RuntimeException (other checked exceptions should be fine)

ulfjack:
Apart from the minor nit in the review, I think this looks good.

werkt:
> Does Bazel retry this call after the change?

Bazel already retries this call due to the executeAsync at https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java#L209. The effect of this change is to catch this exception in https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java#L317 and trigger a fallback to local or informative remote exception to match the behavior of the other StatusRuntimeException -> IOException wrappers.

This is a vestige of the old behavior of Retrier, where it would throw a cause with retries exceeded wrapped in its own RetryException, descended from IOException, rather than the bare cause. All of the other remote methods during the execution loop have been crafted already to wrap their fallback-inducing retry-violations in IOExceptions for this purpose.

src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
werkt: No problem. Done.

ulfjack:
LGTM

@philwo can you merge this or find someone to do so?

PR #10667, opened 2020-01-28
Protype for PackageValidator validator work

kkress:
Proposal for a PackageValidator API in Bazel to support applying cross rule instance validation checks to BUILD files.

Proposal:
https://docs.google.com/document/d/1FfRXzW5RAB7tixbd1Lac51GNqNGQN932ZuTRdg8oUWU/edit#

Not intended for merge, only for preview of changes


meisterT:
Can this be closed?

kkress:
I think so, I'm picking this work up again and things have diverged significantly from the original and the prototype isn't the direction we want to go.

Issue #10541, opened 2020-01-08
Documentation: create a Bazel glossary


jin:
Feature request: create a glossary for abstractions used in the Bazel build system and ecosystem.

Work in progress: https://github.com/jin/bazel-glossary

gregestren:
Super-useful! 

Happy to help with (at least) configuration-related terms.

PR #10677, opened 2020-01-29
Bazel Glossary

jin:
First draft of the Bazel Glossary.

There are a few terms left undefined, which I've marked as comments and TODOs for now.

Another follow up is to add out-links to the full documentation page for particular terms, like "Aspect".

Thanks to all contributors to this glossary: 

* @gregestren 
* @ahumesky 
* @laurentlb 
* @tetromino 
* @jmmv 
* @artem-zinnatullin 

Fixes https://github.com/bazelbuild/bazel/issues/10541


site/docs/glossary.md
```diff
@@ -0,0 +1,545 @@
+---
+layout: documentation
+title: Bazel Glossary
+---
+
+# Bazel Glossary
```
gregestren: Some of these have extensive "dive in deeper" docs elsewhere (e.g. cquery or platforms). What should be the process for that?

* Have a standardized **See also:** section at the bottom of each entry?
* Arbitrary inline text / links?
* Don't link out at all?

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+
+#### Toolchain
+
+A set of tools to build outputs for a language. Typically, a toolchain includes
+compilers, linkers, interpreters or/and linters. A toolchain can also vary by
+platform, that is, a Unix compiler toolchain's components may differ for the
+Windows variant, even though the toolchain is for the same language. Selecting
+the right toolchain for the platform is known as *toolchain resolution*.
+
+#### Top-level
```
laurentlb: `Top-level` alone is ambiguous.
Update to `Top-level target`.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+
+A file used or derived by Bazel. Can also be a directory of files, known as
+"tree artifacts". Tree artifacts are black boxes to Bazel: Bazel does not treat
+the files in tree artifacts as individual artifacts and thus cannot reference
+them directly as action inputs / outputs. An artifact can be an input to
+multiple actions, but must only be generated by at most one action.
+
+#### Aspect
+
+A mechanism for rules to create additional [*actions*](#action) in their
```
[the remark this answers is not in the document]
laurentlb: I think we should have way more links to the actual documentation (I think around 50% of the entries have a corresponding section in the documentation, with much more information than this glossary).

This page already represents a lot of work, so I'm happy if we do that in a followup PR.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+
+The second phase of a build. Processes the target dependency graph specified in
+[*BUILD files*](#build-file) to produce an in-memory [*action
+graph*](#action-graph) that determines the order of actions to run during the
+[*execution phase*](#execution-phase). This is the phase in which rule
+implementations are evaluated.
+
+#### Artifact
+
+A file used or derived by Bazel. Can also be a directory of files, known as
```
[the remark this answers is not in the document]
laurentlb: Do you mean "a source file or a generated file"?

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+#### Action key
+
+The cache key of an [*action*](#action). Computed based on action metadata, such
+as the command line that will be executed, compiler flags, library locations and
+system headers. Enables Bazel to cache or invalidate individual actions
+deterministically.
+
+#### Action cache
+
+The cache of [*action*](#action) metadata to action execution result. Lives on
```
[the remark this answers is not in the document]
jin: Took another stab at this. PTAL.

site/docs/glossary.md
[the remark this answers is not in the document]
jin: Going with "a source file or a generated file".

site/docs/glossary.md
jin: We should link out, and I like the *See also:* approach. I'll add link outs in a follow up PR to keep this iteration scoped down.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
+---
+layout: documentation
+title: Using the Android Native Development Kit with Bazel
+---
+
+# Bazel Glossary
```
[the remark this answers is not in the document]
jin: Hmm, would a table-of-contents be better here?

site/docs/glossary.md
```diff
@@ -0,0 +1,545 @@
+---
+layout: documentation
+title: Bazel Glossary
+---
+
+# Bazel Glossary
+
+#### .bazelrc
+
+Bazel’s configuration file used to change the default values for *startup
```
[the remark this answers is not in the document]
jin: Agree, and done. It looks much better without italics everywhere.

site/docs/glossary.md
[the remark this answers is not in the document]
jin: Yep, I'm happy to do a follow up to add link-outs to full documentation pages.

site/docs/glossary.md
gregestren: SGTM. Happy to help with the content of them if you need.

site/docs/glossary.md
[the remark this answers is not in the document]
gregestren: Soft pitch on my other work: this would look good with those right-side context menus vs. a top ToC that takes up a bunch of vertical real estate.

site/docs/glossary.md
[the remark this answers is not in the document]
artem-zinnatullin: I think external links are great to have but it's probably better to have them listed as a separate optional block under each term

One of the goals of having such a glossary is that it is for the most part self-contained like an Oxford English Dictionary and such :)

artem-zinnatullin:
Super small point to think about: during contributions to the draft of this glossary I've found it pretty challenging to maintain the column width limit (I was using IntelliJ Idea to edit with hard-wrapping on 80 symbols turned on):

- Markdown links break if `[` `]` or `[]` and `()` are on different lines and such
- Unnecessary git diffs appear when you're editing existing term definition, ie you changed line 2 and now all the following lines in that term have to be changed to fit 80 symbols
- Leftover space symbols at the end of lines after editing

I understand that certain editors are better for this, but if we're going to maintain such a restriction it would be great to enforce formatting or ease it somehow (maybe suggest an editor to use or have an [editorconfig](https://editorconfig.org/) in place)

site/docs/glossary.md
```diff
@@ -0,0 +1,563 @@
...
+
+We welcome contributions to this glossary.
+
+The definitions should strive to be crisp, concise, and easy to understand.
+
+Principles:
+
+* Use the simplest words possible. Try to avoid technical jargon.
+* Use the simplest grammar possible. Try not to string too many concepts into a
+* sentence.
```
devcurmudgeon: this word should be part of the previous bullet point

site/docs/glossary.md
```diff
@@ -0,0 +1,563 @@
...
+
+A mechanism for rules to create additional [actions](#action) in their
+dependencies. For example, if target A depends on B, one can apply an aspect on
+A that traverses ]up] a dependency edge to B, and runs additional actions in B
+to generate and collect additional output files. These additional actions are
+cached and reused between targets requiring the same aspect. Created with the
+`aspect()` Starlark Build API function. Can be used, for example, to generate
+metadata for IDEs, and create actions for linting.
+
+]See also:] [Aspects documentation](skylark/aspects.html)
```
devcurmudgeon: Strange ] around the See also: text

jin:
@artem-zinnatullin I hear you. I personally use vim/emacs' `gf` to hard wrap the text. We'll eventually run into internal linters that enforces column widths, so that's something that we can and should set up as an additional buildkite pipeline on GitHub.

site/docs/glossary.md
jin: thanks.

site/docs/glossary.md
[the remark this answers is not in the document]
axgillies: Looks a lot better, thanks!

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+Information outside of rule definitions that impacts how rules generate actions.
+Every build has at least one configuration specifying the target platform,
+action environment variables, and command-line build flags. *Transitions* may
+create additional configurations, e.g. for host tools or cross-compilation.
+
+#### Configuration trimming
+
+The process of only including the pieces of configuration a target actually
+needs. For example, if you build Java binary `//:j` with C++ dependency `//:c`,
+it's wasteful to include `--javacopt` in `//:c`,'s configuration because
```
axgillies: Looking at this line ...

"it's wasteful to include `--javacopt` in `//:c`,'s configuration because ..."

As a general practice, it's a good idea to take a look at places where you are using the possessive case with apostrophe-s ('s). For example, here, the following is a lot clearer:

"... it's wasteful to include `--javacopt` in the configuration of `//:c` because ...:

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+changing `--javacopt` unnecessarily breaks C++ build cacheability.
+
+#### Configured query (cquery)
+
+A query tool that queries over *configured targets* (after the analysis phase
+completes). This means select() and build flags (e.g. `--platforms`) are
```
axgillies: Consider setting up "select()" with backticks here.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+accurately reflected in the results.
+
+#### Configured target
+
+The result of evaluating a *target* with a *configuration*.. The analysis phase
```
axgillies: Remove extra period character after first line.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+produces this by combining the build's options with the targets that need to be
+built. For example, if `//:foo` builds for two different architectures in the
+same build, it has two configured targets: `<//:foo, x86>` and `<//:foo, arm>`.
+
+#### Correctness
+
+The property that a build's output faithfully reflects the state of its
```
axgillies: I'd say something like this to make it a bit more clear:

"A build is correct when its output faithfully reflects the state of its transitive inputs."

... and then replace the second sentence with something meaningful about hermeticity and non-determinism?

site/docs/glossary.md
```diff
@@ -0,0 +1,565 @@
...
+
+#### Depset
+
+A data structure for collecting data on transitive dependencies. Optimized so
+that merging depsets is time and space efficient, because it’s common to have
+very large depsets (e.g. hundreds of thousands of files). Implemented to
+recursively refer to other depsets for space efficiency reasons. [Rule](#rule)
+implementations should not "flatten" depsets by converting them to lists unless
+the rule is at the top level of the build graph. Flattening large depsets incurs
+huge memory consumption. Also known as Nested Sets in Bazel's internal
```
axgillies: Maybe this case is a good one for italics:

"Also known as _nested sets_ in Bazel's internal implementation."

site/docs/glossary.md
```diff
@@ -0,0 +1,565 @@
...
+conjunction with an actual remote blob store.
+
+#### Distdir
+
+A read-only directory containing files that Bazel would otherwise fetch from the
+internet using repository rules. Enables builds to run fully offline.
+
+#### Dynamic execution
+
+An execution strategy that combines both the local and remote execution
```
axgillies: I propose here:

"An execution strategy that selects between local and remote execution based on various heuristics, and uses the execution results of the faster successful method. Certain actions are executed faster locally (for example, linking) and others are faster remotely (for example, highly parallelizable compilation). A dynamic execution strategy can provide the best possible incremental and clean build times."

site/docs/glossary.md
```diff
@@ -0,0 +1,565 @@
...
+strategy, and through various heuristics, uses the execution result of the
+faster successful strategy. Certain [actions](#action) are better executed
+locally (e.g. linking), and others remotely (highly parallelizable compilation),
+so the dynamic execution strategy provides the best possible incremental and
+full build times.
+
+#### Execution phase
+
+The third phase of a build. Executes the [actions](#action) from the [action
```
axgillies: I propose ...

"Executes in the actions in the action graph ..."

and ...

"These actions invoke executable scripts and programs that read and write artifacts."

site/docs/glossary.md
```diff
@@ -0,0 +1,565 @@
...
+executed: locally, remotely, dynamically, sandboxed, docker, and so on.
+
+#### Execution root
+
+A directory in the [workspace](#workspace)’s [output base](#output-base)
+directory where local [actions](#action) are executed in
+non-[sandboxed](#sandboxing) builds. The directory contents are mostly symlinks
+of input [artifacts](#artifact) from the workspace. The execution root also
+contains symlinks to external repositories as other inputs and the `bazel-out`
+directory to store outputs. Prepared during the [loading phase](#loading-phase)
```
axgillies: Maybe:

"Prepared during the loading phase by creating a _symlink forest_ of directories that represent the transitive closure of packages on which a build depends."

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+cached and reused between targets requiring the same aspect. Created with the
+aspect() Starlark Build API function. Can be used, for example, to generate
+metadata for IDEs, and create actions for linting.
+
+#### Aspect-on-aspect
+
+An aspect composition mechanism, where aspects can be applied on other aspects.
+For an aspect A to inspect aspect B, aspect A must declare the *providers* it
+needs from aspect B (with `required_aspect_providers` attribute), and aspect B
+must declare the providers it returns (with provides attribute). For example,
```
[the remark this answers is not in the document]
jin: This is the `provides` attribute in the aspects declaration. Clarified with formatting.

site/docs/glossary.md
jin: Thanks for the advice! Updated -- sounds much clearer.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+depended on by the target(s) to build. Accessible with `bazel info
+execution_root` on the command line.
+
+#### File
+
+See [*Artifact*](#artifact).
+
+#### Hermeticity
+
+The property of a build or test where there are no external influences on the
```
axgillies: I think this is easier to explain with a bit of a different sentence structure:

"A build is hermetic if there are no external influences on its build and test operations, which helps to make sure that results are deterministic and correct. For example, hermetic builds typically disallow network access to actions, restrict access to declared inputs, use fixed timestamps and timezones, restrict access to environment variables, and use fixed seeds for random number generators."

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+usually means disallowing network access to actions, only allowing access to
+declared input [*artifacts*](#artifact), making timestamps and timezones fixed,
+restricting access/resetting environment variables, and using fixed seeds for
+random number generators.
+
+#### Incremental build
+
+A build that reuses the results of earlier builds to reduce build time and
+resource needs. Bazel aims to guarantee correct results for incremental builds.
+This is the opposite of a clean build.
```
axgillies: Maybe a few adjustments along the lines of:

"An incremental build reuses the results of earlier builds to reduce build time and resource usage. Dependency checking and caching aim to produce correct results for this type of build. An incremental build is the opposite of a clean build."

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+
+<!-- TODO: #### Install base -->
+
+#### Label
+
+An identifier for a target. A fully qualified label such as
```
axgillies: Replace "fully qualified" with "fully-qualified."

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+`//path/to/package:target` consists of `//` to mark the workspace root
+directory, path/to/package as the directory that contains the [*BUILD
```
axgillies: Set off `path/to/package` and `:target` with backticks.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+file*](#build-file) declaring the target, and :target as the name of the target
+declared in the aforementioned BUILD file. May also be prefixed with
+`@my_repository//<..>` to indicate that the target is declared in an *external
+repository* named `my_repository`.
+
+#### Loading phase
+
+The first phase of a build. WORKSPACE, BUILD, and .bzl files are parsed to
```
axgillies: Again, take a look at usage of backticks to specify monospaced fonts throughout the document.

Generally, we want to signal typographically whenever a special term is used to indicate a filename, file type, or snippet of command or code that a user might have to enter. So, here, preferentially, we would use `WOKRSPACE`, `BUILD` and `.bzl` to set off the terms, though I imagine this is very inconsistent throughout the doc set.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+create *packages*. Macros and certain functions like glob() are evaluated in
```
axgillies: As above, function name like `glob()` should be set off with backticks.

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+#### Macro
+
+A mechanism to compose multiple rule target declarations together under a single
+Starlark function. Enables reusing common rule declaration patterns across BUILD
+files. Expanded to the underlying rule target declarations during the *loading
+phase*.
+
+#### Mnemonic
+
+A short, human-readable string to quickly understand what an *action* is doing.
```
axgillies: A couple of notes here:

The context of the term is not clear -- where is the mnemonic declared? In BUILD files, in .bzl files? Second, it's not clear who is supposed to be understanding when worded like this.

For example, this might better be something like:

"A *mnemonic* is a short, human-readable string in a `.bzl` file that describes what an action is doing. Mnemonics can be used as identifiers ..."

site/docs/glossary.md
```diff
@@ -0,0 +1,537 @@
...
+
+#### Output base
+
+A unique directory specific to a workspace for containing Bazel output files to
+separate outputs from the source tree. Located in the *output user root*.
+
+#### Output groups
+
+A group of files that is expected to be built when Bazel finishes building a
+target. Rules put their usual outputs in the "default output group" (e.g the
```
axgillies: As above, it helps users out to set apart file types like `.jar`, `.a` and `.so` with backticks as you have here with `cc_library`.

site/docs/glossary.md
jin: Thanks! I think you may be looking at an older revision of the file in this PR - I've done a pass for the backtick change throughout the document.

site/docs/glossary.md
jin: Thanks! I've reworded it to make it clearer. PTAL.

PR #10691, opened 2020-02-01
On OpenBSD, search the PATH to find the Bazel executable.

aldersondrive:
Searching the `PATH` is the only feasible way to find the executable on OpenBSD when `argv[0]` is not an absolute path or a relative path. This change resolves a TODO.

This change moves a preexisting `Which` function for searching the `PATH` out of `blaze_util_linux.cc` and into `blaze_util_posix.cc`, so that the Linux code and the BSD code can share this function.

In my testing on OpenBSD 6.6-current, a bootstrap build of Bazel succeeds and the resulting `bazel` binary can find itself on the `PATH`. (One caveat: For the bootstrap build to succeed, I had to manually apply the unrelated change in https://github.com/bazelbuild/bazel/pull/10639, since it's not merged yet.)

@jmmv FYI.

### DIFF
--- a/src/main/cpp/blaze_util_bsd.cc
+++ b/src/main/cpp/blaze_util_bsd.cc
@@ -131,13 +131,15 @@ string GetSelfPath(const char* argv0) {
     return blaze_util::GetCwd() + "/" + argv0str;
   }
 
-  // TODO(aldersondrive): Try to find the executable by inspecting the PATH.
+  // Otherwise, try to find the executable by searching the PATH.
+  std::string from_search_path = Which(argv0);
+  if (!from_search_path.empty()) {
+    return from_search_path;
+  }
 
   // None of the above worked. Give up.
   BAZEL_DIE(blaze_exit_code::BAD_ARGV)
-      << "Unable to determine the location of this Bazel executable. "
-         "Currently, argv[0] must be an absolute or relative path to the "
-         "executable.";
+      << "Unable to determine the location of this Bazel executable.";
   return "";  // Never executed. Needed so compiler does not complain.
 #else
 # error This BSD is not supported
--- a/src/main/cpp/blaze_util_bsd.cc
+++ b/src/main/cpp/blaze_util_bsd.cc
@@ -132,7 +132,7 @@ string GetSelfPath(const char* argv0) {
   }
 
   // Otherwise, try to find the executable by searching the PATH.
-  std::string from_search_path = Which(argv0);
+  const std::string from_search_path = Which(argv0);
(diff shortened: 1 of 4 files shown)


src/main/cpp/blaze_util_posix.cc
```diff
@@ -217,6 +218,29 @@ string GetHomeDir() { return GetPathEnv("HOME"); }
 
 string GetJavaBinaryUnderJavabase() { return "bin/java"; }
 
+string Which(const string& executable) {
+  string path = GetPathEnv("PATH");
```
jmmv: Take the chance to make all these local variables `const`.

src/main/cpp/blaze_util_posix.cc
```diff
@@ -217,6 +218,29 @@ string GetHomeDir() { return GetPathEnv("HOME"); }
...
+  for (string piece : pieces) {
+    if (piece.empty()) {
+      piece = ".";
+    }
+
+    struct stat file_stat;
+    string candidate = blaze_util::JoinPath(piece, executable);
+    if (access(candidate.c_str(), X_OK) == 0 &&
+        stat(candidate.c_str(), &file_stat) == 0 &&
+        (S_ISREG(file_stat.st_mode) || S_ISLNK(file_stat.st_mode))) {
```
jmmv: Is the `S_ISLNK` check necessary? `stat` follows symlinks, so I would imagine it only returns this type on error -- in which case the candidate isn't valid.

src/main/cpp/blaze_util_bsd.cc
```diff
@@ -131,13 +131,15 @@ string GetSelfPath(const char* argv0) {
     return blaze_util::GetCwd() + "/" + argv0str;
   }
 
-  // TODO(aldersondrive): Try to find the executable by inspecting the PATH.
+  // Otherwise, try to find the executable by searching the PATH.
+  std::string from_search_path = Which(argv0);
```
jmmv: nit: Can be `const`.

src/main/cpp/blaze_util_posix.cc
aldersondrive: Done.

src/main/cpp/blaze_util_bsd.cc
aldersondrive: Done.

src/main/cpp/blaze_util_posix.cc
aldersondrive: Whoops, good catch! Removed.

Issue #10523, opened 2020-01-06
Implement the toolchain transition


katre:
Tracking issue for the implementation of [toolchain transitions](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md).

eagleonhill:
What's the target release for toolchain transition and `exec` transition?

While I agree it's better to use `exec` transition than `host`, the fact that currently entire toolchain has applied `host` transition makes it impossible to include runtime libraries today. If `exec` transition would take more effort, would it be possible to decouple them, so there's some proper way to link runtime?

Just a little more background: https://docs.bazel.build/versions/master/skylark/lib/CcToolchainInfo.html#dynamic_runtime_lib is currently broken because the symlink is located at `bazel-out/host`.

katre:
Are you using the `--incompatible_enable_cc_toolchain_resolution` flag? We haven't flipped that behavior and this is part of the reason it's not the default yet. I'd recommend flipping that back and continuing to use the legacy `cc_toolchain`suite` approach if this is causing you problems.

I am working on a migration plan for the toolchain transition and hope to send it around shortly. Once that's been agreed on I'll begin implementation and start working with rule owners to migrate, since doing this as a single world-breaking migration isn't feasible. Because CC rules don't use toolchain resolution by default, getting that working may be far down the list.

timblakely:
Just ran into what I thought was #6889. Glad to see I wasn't going crazy when querying `ctx.target_platform_has_constraint` and not seeing the right constraints.

My use case is that I need to configure `feature`s and `flag_sets` that are passed to the `cc_common.create_cc_toolchain_config_info` based on the target cpu. My workaround is to have a separate cc_toolchain for each target, but that's not scalable in the long run.

Once toolchain transitions are in, would this use case be supported?

katre:
That is internal to the `cc_toolchain` machinery, and so it's part of toolchain resolution (toolchain resolution will select a single `cc_toolchain` instance for a target, and then that toolchain's dependencies are up to the rule implementation).

You'll need to follow up with the CC rule maintainers to find out if that's a use case they want to support. if so, I can work with them on ways to implement it in their toolchain implementation.

hlopko:
Wohooo \o/ Congrats! :)

katre:
Re-opening, we also need to support aspects that use toolchains.

katre:
Now I can close again.

Issue #6889, opened 2018-12-11
Toolchain dependencies are all in the HOST configuration


damienmg:
__EDIT__: Sorry for the first post, misposted.
 
See https://github.com/bazelbuild/bazel/pull/5650 for more context. I am opening a bug since there is no solution yet and I need to refer to it.

### Description of the problem / feature request:

If you add a target dependencies to a toolchain, the dependency ends-up being a host dependencies in the final binary. https://github.com/bazelbuild/bazel/pull/5650 contains way more information and a fix for it.

katre:
This is a topic I plan to look into soon, sorry that the backlog has been high.

Several other rule authors have also mentioned this issue, thanks for filing it.

katre:
See https://groups.google.com/forum/#!topic/bazel-dev/5osWxhoF0Fk for the design to address this.

katre:
Tracking issue for implementing toolchain transitions (finally): #10523.

katre:
This will be fixed by #7935 and #10523, closing in favor of those issues.

Issue #4067, opened 2017-11-10
Toolchains with deps in target mode


ianthehat:
At the moment all the deps of a toolchain are evaluated in host mode.
It would be nice if toolchains could "inject" deps on target (for instance, the library that goes along with the generated source file produced by the toolchain) but there is no way to achieve this at the moment that works for cross compilation.

damienmg:
@katre: I don't really understand, can you tal and triage?

katre:
@ianthehat and I discussed this issue, I know what is required.

katre:
This will be implemented by #11128.

PR #10694, opened 2020-02-03
Use target config for toolchain

eagleonhill:
Toolchain typically contains two parts: Compiler and runtime.
Runtime should be using target config.
For compilers that needs host/exec config, the toolchain should do own
transition.

Currently the toolchain is using host config if enabled new toolchain
resolution, which caused the runtime not able to be linked as expected.


katre:
This is a known issue that has been reported several times (see #4067 and #6889). Unfortunately, simply using the target configuration is also not always correct: consider the case of a toolchain that wants to build the compiler itself (which would need to use the exec configuration).

I have proposed a solution [called Toolchain Transitions](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md) which allows for both cases, and I have opened an issue tracking that work (#10523).

I am closing this PR but please read the proposal and leave any comments on #10523 if there is a case that isn't covered there.

Issue #7514, opened 2019-02-22
Add Starlark support for Symlink actions.


sergiocampama:
### Description of the problem / feature request:

There are multiple benefits for using symlinks in certain actions, especially when creating directory structures or renaming artifacts' basenames. For example, rules_apple requires heavy usage of [symlinks](https://github.com/bazelbuild/rules_apple/blob/0a9df881d57288f54a1c571fc822f9b275e574ba/apple/bundling/file_actions.bzl).

Given that symlinks are a core foundation of how Bazel works, Starlark should provide an API to register symlink actions, that work both on macOS, Linux and Windows. In native code this already exists, through the [`SymlinkAction`](https://github.com/bazelbuild/bazel/blob/da74aa058422c3b9e32e4b86b17c3a264521a926/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkAction.java) class.  

This could be added to the `actions` interface, something like `ctx.actions.symlink(src, dest)` where both `src` and `dest` are `File` instances created through `ctx.actions.declare_file`, for example. Notice that `File`instances can also represent tree artifacts.

laurentlb:
@Yannic, thanks for the pull request!

Can you briefly clarify your plans? Do you plan to do followup changes after #10695?

Yannic:
@laurentlb I'm planning to investigate allowing symlinks to executables (it's unclear how dangling symlinks interact with them and I don't want to commit to an API yet), and to allow specifying a progress message for these actions (actually, it's not large change so I guess I'll just do it as part of #10695). That said, #10695 can be used by rule authors to create symlinks without any follow-ups.

Yannic:
I think this is implemented and can be closed?

pauldraper:
Yep.

https://docs.bazel.build/versions/master/skylark/lib/actions.html#declare_symlink

https://docs.bazel.build/versions/master/skylark/lib/actions.html#symlink

thiagohmcruz:
Got here while reading some things about `ctx.actions.symlink`, jump bump-ing that apparently this can be closed 👍

brandjon:
Thanks, closing.

Issue #10298, opened 2019-11-24
Stabilize unresolved symbolic links (--experimental_allow_unresolved_symlinks)


jmillikin:
### Description of the problem / feature request:

https://github.com/bazelbuild/bazel/issues/9075 added support for unresolved symlinks, via the `actions.declare_symlink()` and `actions.symlink()` functions. That functionality is currently gated by `--experimental_allow_unresolved_symlinks`.

I've discovered that unresolved symlinks are critical for supporting hermetic Node.JS builds. Specifically, Node's use of symlinks for module resolution semantics mean that we need to be able to create symlink artifacts that reference dependencies and can be used as runfiles, but exist outside the runfiles root.

What are the current blockers to stabilizing unresolved symlinks? Is there a list of known remaining work to get them enabled by default?

cc @lberki @laszlocsomor

lberki:
@laurentlb , @laszlocsomor , what do you think?

The main blockers I see is that RBE doesn't support this yet (the protocol itself does, but we haven't wired it up yet) and that it probably fails on Windows in all sorts of interesting ways. 

Neither of these are insurmountable problems and if I squinted the right way, I could even argue that they are not necessary for declaring the functionality stable, but I just can't sign up for doing them at the moment.

laszlocsomor:
@lberki : Why do you suppose it fails on Windows?
Is there a test case I could try on Windows?

lberki:
Try `//third_party/bazel/src/test/shell/bazel:bazel_unresolved_symlink_test`.

I admit I haven't tried it on Windows. There is a chance that it works since all that's needed is `ActionMetadataHandler#fileArtifactValueFromArtifact()` to wrok which in turn only needs `Path#statIfFound(Symlinks.NOFOLLOW)` and `Path#readSymbolicLink()` .

lberki:
...oh, and `Path#createSymbolicLink()` for `SymlinkAction#execute()`.

laszlocsomor:
Thanks! The test fails here: https://github.com/bazelbuild/bazel/blob/b1e232d6f1d82494639660e8a2772ac7da7c5da8/src/test/shell/bazel/bazel_unresolved_symlink_test.sh#L243

...because the build succeeds though it shouldn't. (Note: I edited the test to use `c:/badsymlink` instead of `/badsymlink`.)

laszlocsomor:
The culprit is that WindowsFileSystem does not create file symlinks:
[...]

Issue #9075, opened 2019-08-05
Add support for unresolved symbolic links


lberki:
We have had a few requests to support unresolved symbolic links, but so far, we have managed to live without.

The use case is creating packages and file system images where one has a symlink (for example`/bin/bash`) and the desired result is not to package up the file the symlink points to (in this case, the shell of the system the action runs on), but to put the actual symlink `/bin/bash` in the package.

RBE wisely decided to implement this ahead of time, which is good. 

I expect there to be some trouble with Windows, but other than that, the change should be relatively straightforward.

We might run in to trouble if we want to support unresolved symlinks within tree artifacts: whether the symlink needs to be resolved may need to be decided on a per-file basis and it's not trivial to convey that information to Bazel. However, we'll cross that bridge when we get to it.

laurentlb:
fyi @laszlocsomor

laszlocsomor:
Thanks @laurentlb .

https://github.com/bazelbuild/bazel/blob/299655ee2600874667fa9e21e6d1e598bb011b56/src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java#L109 is implemented properly, so it should work as long as you go through the VFS (and not Java IO).

lberki:
This is now fixed with the above commit. If there are any bugs, let's tackle separately.

sergiocampama:
is #7514 resolved by this issue?

lberki:
No, i specifically did not want to resolve it... that issue is about "traditional", i.e. resolved symlinks.

I did create a Starlark API for it (`symlink` when called on an artifact produced by `declare_file` does exactly that), but I put it behind an experimental flag and I'm not planning to enable it by default.

PR #10695, opened 2020-02-03
Allow artifact -> artifact symlink actions from Starlark

Yannic:
Updates #7514


src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -221,28 +221,58 @@ public void doNothing(String mnemonic, Object inputs) throws EvalException {
...
+    } else {
+      // Should have been verified by Starlark before this function is called
+      throw new IllegalStateException();
+    }
+  }
+
+  private void registerSymlinkAction(
+      Artifact output, Artifact input, String progressMessage) throws EvalException {
+    if (output.isSymlink()) {
+      throw Starlark.errorf("output of symlink action must be created by declare_file()");
```
lberki: nit: "symlinks pointing to artifacts must be created by declare_file()?"

If you want a dangling symlink, it has to created by `declare_symlink()`, doesn't it?

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -221,28 +221,58 @@ public void doNothing(String mnemonic, Object inputs) throws EvalException {
...
   @Override
-  public void symlink(FileApi output, String path) throws EvalException {
+  public void symlink(FileApi output, Object target) throws EvalException {
     context.checkMutable("actions.symlink");
 
+    Artifact outputArtifact = (Artifact)output;
+    // TODO(yannic): This should be configurable.
+    String progressMessage = "Creating symlink " + outputArtifact.getRootRelativePathString();
+
+    if (target instanceof String) {
```
lberki: @laurentlb , is this kind of behavior change based on the runtime type of the argument okay?

It seems to be reasonable on the surface, but there doesn't seem to be that many instances of this in our code and maybe explicitly specifying which one we want like `actions.symlink(output, file=<file>)` / `actions.symlink(output, symlink=<string>)` would be better?

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,29 @@
   @SkylarkCallable(
       name = "symlink",
       doc =
-          "<p><b>Experimental</b>. This parameter is experimental and may change at any "
```
laurentlb: Why did you remove the "Experimental" annotation?

Yannic: I intend to only mark the "dangling symlink" behavior as experimental, the artifact -> artifact symlink introduced in this PR has been used by native rules for a long time and (IMO) shouldn't need to be behind a flag.

I agree that it would be nice to have an "Experimental" warning for the dangling symlink part somewhere, I just wasn't sure where to put that warning. OTOH, the only way to get a dangling symlink is `ctx.actions.declare_symlink()`, which already is marked as experimental, so 🤷‍♀️.

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
Yannic: TBH, I actually like having `actions.symlink(output, file=<file>)` / `actions.symlink(output, symlink=<string>)` more than overloading `target` based on the runtime type.

lberki: Same here, but I want this to be idiomatic, that's why I asked for @laurentlb 's opinion!

brandjon:
Some thoughts:

1. My understanding is that all symlink functionality in Starlark is currently experimental. Let's not remove any experimental designations in this PR; that requires a separate discussion.

2. It seems the API we're exposing to Starlark is that a symlink is a special kind of `File`, created by `declare_symlink()` instead of `declare_file()` or other means, and that `symlink()` requires a symlink `File`. This should be clearly documented. (I also think the wording around `output` could be made clearer, but we can iterate on documentation separately. Maybe the param should be renamed to `link` for symmetry with `man 1 ln`?)

3. We seem to be in agreement that direct overloading on path string vs `File` is confusing, and that a nice alternative to fully splitting `symlink()` into two methods would be to pass in mutually exclusive keyword args. I like the names `target_path=...` and `target_file=...` for these params.

laurentlb:
> overloading on path string vs File is confusing

I agree. Maybe we need to resolve the situation about the experimental dangling symlink feature. Do we want to support it? Is it going away?

lberki:
We definitely want to support it in the long term. For now, however, the fully supported one is the artifact -> artifact symlink.

Yannic:
> Some thoughts:
> 
> 1. My understanding is that all symlink functionality in Starlark is currently experimental. Let's not remove any experimental designations in this PR; that requires a separate discussion.

"Normal" symlink functionality (i.e. artifact -> artifact) is currently non-existent in Starlark, only unresolved symlinks are available behind `--experimental_allow_unresolved_symlinks`. I don't think guarding artifact -> artifact symlinks with a flag is necessary (and definitely not with `--experimental_allow_unresolved_symlinks`, that would be really confusing to users).

> 2. It seems the API we're exposing to Starlark is that a symlink is a special kind of `File`, created by `declare_symlink()` instead of `declare_file()` or other means, and that `symlink()` requires a symlink `File`. This should be clearly documented. (I also think the wording around `output` could be made clearer, but we can iterate on documentation separately. Maybe the param should be renamed to `link` for symmetry with `man 1 ln`?)

See 1. for `declare_file()` vs `declare_symlink()`. I agree that documentation should be improved as part of this PR :).

> 3. We seem to be in agreement that direct overloading on path string vs `File` is confusing, and that a nice alternative to fully splitting `symlink()` into two methods would be to pass in mutually exclusive keyword args. I like the names `target_path=...` and `target_file=...` for these params.

Seems like we all agree that overloading is not the way to go.

lberki:
Let's go with `target_path=` and `target_file=` then. `ctx.actions.symlink()` is nice enough that I don't want to uglify it and rename it to something like `symlink_file()`.

I just realized that this is inconsistent with the `symlink` function in repository contexts which have `from=` and `to=` attributes, but given that those are very confusing (in fact, I was sure that `from` denotes the symlink and `to` denotes the target until I read the documentation), so I'd much rather keep the `output`/`target` names, unless @laurentlb or @brandjon strongly disagree.

brandjon:
> "Normal" symlink functionality (i.e. artifact -> artifact) is currently non-existent in Starlark, only unresolved symlinks are available behind `--experimental_allow_unresolved_symlinks`. I don't think guarding artifact -> artifact symlinks with a flag is necessary

If it's experimental it needs to be behind a flag.

> I'd much rather keep the `output/target` names

Any thoughts on `link` instead of `output`? I don't have a strong opinion, and `output` might be better after all if we're playing up the symmetry between normal file artifacts and symlink artifacts in Starlark API.

src/test/shell/bazel/bazel_unresolved_symlink_test.sh
```diff
@@ -308,4 +415,122 @@ EOF
...
+EOF
+
+  cat > a/BUILD <<'EOF'
+load(":a.bzl", "a")
+
+a(name="a")
+EOF
+
+  # bazel build //a:a && fail "build succeeded"
+  # [[ "$?" == 1 ]] || fail "unexpected exit code"
```
Yannic: I can't quite figure out why this isn't failing.
Looking at [this](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkAction.java;l=62;drc=79b4a07d35179f9ae29c4c0103bc2ecfcbe5c96c) comment, it seems like the action should fail.
Am I missing something here, or is that comment out-of date (or invalid for Bazel)?

Yannic:
This is ready for review now.

@brandjon I decided to stick with `output` to be consistent with `ctx.actions.write`, `ctx.actions.run` and `ctx.actions.run_shell`. I've also added an "experimental" warning to `target_path` (which is the only experimental part of this API).

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -220,29 +221,89 @@ public void doNothing(String mnemonic, Object inputs) throws EvalException {
...
 
-    PathFragment targetPath = PathFragment.create(path);
-    Artifact outputArtifact = (Artifact) output;
     if (!outputArtifact.isSymlink()) {
       throw Starlark.errorf("output of symlink action must be created by declare_symlink()");
     }
 
-    Action action =
+    if (isExecutable) {
+      // TODO(yannic): Do we allow dangling symlinks to be executable?
```
lberki: Let's be strict first.

src/test/shell/bazel/bazel_unresolved_symlink_test.sh
lberki: Hehe. That's because Bazel makes every output file executable (which is not how I would like it to be, but it would be quite difficult to change now :( ). Probably this is where:

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java#L737

If you want to test this functionality, symlink to a source file instead of one that was built by Bazel itself.

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -220,29 +221,100 @@ public void doNothing(String mnemonic, Object inputs) throws EvalException {
...
+    if (progressMessage != Starlark.NONE) {
+      // Should have been verified by Starlark before this function is called.
+      return (String)progressMessage;
+    }
+
+    return "Creating symlink " + output.getRootRelativePathString();
+  }
+
+  /** Return whether exactly one of the passed object is not {@link Starlark.NONE}. */
+  private boolean exactlyOneNotNone(Object first, Object... others) {
```
lberki, line 235: This is a bit of an overgeneralization, but ok. I would have just used a simple if statement, i.e:

```
if ((targetFile == Starlark.NONE) == (targetPath == Starlark.NONE)) {
}
```

lberki:
Update: internal change is being reviewed by @brandjon

Yannic:
Any updates on this?

brandjon:
Hi Yannic, I had a few comments from the internal review, but first I want to step back and make sure I understand the big picture. I'm particularly cautious because we seem to have skipped a Starlark API review for this change and for lberki@'s earlier dangling symlink feature (which up until now has been marked experimental).

Questions for @lberki:

1. There's an old doc, [Symlinks in Bazel Remote Caching and Execution](https://docs.google.com/document/d/1zqXCrQk1gF6kRvHXxpMiEfZGNBJ7rlVN8PeBl195_Zc/edit), that says dangling symlinks are explicitly not supported in remote execution. #9075 and 2fc2e2b84aa378de68a8f3f87a2f22eb583a0df3 say different things about RBE compatibility but I don't know the nuances. What's the current status of this feature and remote execution, and does it block releasing any symlink functionality as non-experimental?

2. To confirm your earlier [post](https://github.com/bazelbuild/bazel/pull/10695#issuecomment-583421163), you agree that the new functionality added in this PR is ok to be marked non-experimental, right?

For @yannic and @lberki: I want to make sure we're on the same page. Does this describe the API?

    * `declare_file()` creates a regular artifact (`isSymlink() == false`). But the corresponding physical file may or may not be a symlink, depending on whether it was generated by `symlink()` or another action. So users would use `declare_file()` when they want something guaranteed to exist.

    * `declare_symlink()` creates a symlink artifact (`isSymlink() == true`). The corresponding physical file must be a symlink. It may or may not resolve to a path that exists, and that path may or may not be in the output tree. This functionality remains experimental for now.

I'll reply with my comments from the internal review shortly.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,70 @@
...
-              + "time. Please do not depend on it. It may be enabled on an experimental basis by "
-              + "setting <code>--experimental_allow_unresolved_symlinks</code></p><p>"
-              + "Creates a symlink in the file system. If the output file is a regular file, the "
-              + "symlink must point to a file. If the output is an unresolved symlink, a dangling "
-              + "symlink is allowed.",
+          "Creates a symlink in the file system."
```
brandjon, line 162: To match the docs for other actions, reword as:

> Creates an action that writes a symlink in the file system.

This emphasizes that nothing is being written at analysis time.

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -220,29 +221,100 @@ public void doNothing(String mnemonic, Object inputs) throws EvalException {
...
+                target,
+                outputArtifact,
+                progressMessageForSymlink(progressMessage, outputArtifact));
+        registerAction(action);
+        return;
+      }
+      SymlinkAction action =
+          SymlinkAction.toArtifact(
+              ruleContext.getActionOwner(),
+              target,
```
brandjon, line 287: Question: If `symlink()` makes an `Artifact` with `isSymlink() == false` point to another `Artifact` with `isSymlink() == false`, doesn't that mean you can have two symlinks point to each other? Or is that prevented by having the target `Artifact` considered an input to the symlink action?

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,70 @@
...
+              + "<p>If the output file is a regular file (i.e. created by "
+              + "<a href=\"#declare_file\"><code>declare_file()</code></a>), the symlink must "
+              + "point to a file. If the output is an unresolved symlink (i.e. created by "
+              + "<a href=\"#declare_symlink\"><code>declare_symlink()</code></a>), a dangling "
+              + "symlink is allowed.</p>",
```
brandjon, line 167: Suggested rewording:

> This function must be called with exactly one of `target_file` and `target_path` specified.
>
> If `target_file` is used, then `output` must be declared as a regular file (such as by using `declare_file()`). It will actually be created as a symlink that points to the path of `target_file`.
>
> If `target_path` is used instead, then `output` must be declared as a symlink (such as by using `declare_symlink()`). In this case, the symlink will point to whatever the content of `target_path` is. This can be used to create a dangling symlink.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,70 @@
...
+            name = "target_file",
+            type = FileApi.class,
+            named = true,
+            positional = false,
+            noneable = true,
+            defaultValue = "None",
+            doc =
+                "The target of the symlink."
+                    + "<p>If this parameter is set, <code>target_path</code> "
+                    + "must be <code>None</code>.</p>"),
```
brandjon, line 185: With the above wording in place, this can be simplified to:

> The `File` that the output symlink will point to.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,70 @@
...
+            name = "target_path",
+            type = String.class,
+            named = true,
+            positional = false,
+            noneable = true,
+            defaultValue = "None",
+            doc =
+                "The target path of the symlink."
+                    + "<p>If this parameter is set, <code>target_file</code> "
+                    + "must be <code>None</code>.</p>"
```
brandjon, line 196: > The exact path that the output symlink will point to. No normalization or other processing is applied.

(@lberki can you confirm that's correct?)

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,17 +159,70 @@
...
+        @Param(
+            name = "is_executable",
+            type = Boolean.class,
+            named = true,
+            positional = false,
+            defaultValue = "False",
+            doc =
+                "Whether the output file should be executable. "
+                    + "<p>If this is <code>True<code>, <code>target_path</code> must also be "
+                    + "executable.</p>"),
```
brandjon, line 209: `target_path` should be `target_file` here. I'd rewrite as:

> May only be used with `target_file`, not `target_path`. If true, when the action is executed, the `target_file`'s path is checked to confirm that it is executable, and an error is reported if it is not. Setting `is_executable` to `False` does not mean the target is necessarily not executable, just that no verification is done. This feature does not make sense for `target_path` because dangling symlinks might not exist at build time.

@lberki please verify what I wrote makes sense.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -70,9 +70,25 @@ function set_up() {
   mkdir -p symlink
   touch symlink/BUILD
   cat > symlink/symlink.bzl <<EOF
```
brandjon, line 72: For whatever reason github won't let me add a comment near the top of the file -- on line 17 delete the word "unresolved" in the comment. Same for the `run_suite` call at the bottom of the file.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -70,9 +70,25 @@ function set_up() {
...
+    ctx.actions.symlink(
+        output = symlink,
+        target_path = ctx.attr.link_target,
+    )
+    return DefaultInfo(files = depset([symlink]))
+
+dangling_symlink = rule(
+    implementation = _dangling_symlink_impl,
+    attrs = {"link_target": attr.string()},
+)
```
brandjon, line 84: I think this whole definition is identical to the `symlink` rule below. Delete and just test the dangling behavior with the `symlink` rule?

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -88,21 +104,31 @@ EOF
 }
 
 function test_smoke() {
+  if "$is_windows"; then
+    warn "Skipping test on Windows: Unresolved symlinks are not supported yet"
```
brandjon, line 108: Is there an existing github issue for this, that we can reference in a TODO (here and at other occurrences)?

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -250,7 +314,42 @@ EOF
   [[ "$?" == 1 ]] || fail "unexpected exit code"
```
brandjon, line 314: I'm not a fan of these tests that just check that the build failed without seeing why (by writing to $TEST_log and using `expect_log` / `expect_not_log` assertions). These tests could be failing due to a syntax error in the test or other unintended failure. That said, this is a pre-existing issue, so you can leave it alone so long as you've confirmed that this test case produces the expected log output showing that the bazel command fails for the right reasons.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -250,7 +314,42 @@ EOF
...
+EOF
+
+  cat > a/BUILD <<'EOF'
+load(":a.bzl", "bad_symlink")
+
+bad_symlink(name="bs")
+EOF
+
+  bazel build //a:bs && fail "build succeeded"
+  [[ "$?" == 1 ]] || fail "unexpected exit code"
```
brandjon, line 344: See above -- instead of cargo-culting the exit code check, use the `$TEST_log` (same for all other tests added by this PR). See other test suites in this directory for examples of how to do it.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -308,4 +415,135 @@ EOF
...
+EOF
+
+  cat > a/BUILD <<'EOF'
+load(":a.bzl", "a")
+
+a(name="a")
+EOF
+
+  bazel build //a:a || fail "build failed"
+  assert_contains "Hello, World!" bazel-bin/a/a.link
```
brandjon, line 445: Add an assertion that the link file is in fact a symlink (as opposed to a copy).

src/test/shell/bazel/bazel_symlink_test.sh
brandjon: Sure, let's replace the warning with a

    // TODO(#10298): Support unresolved symlinks on Windows.

comment. Warnings are logspam anyway -- the fact that we're not running on windows is relevant for maintainers of the test suite, but not people viewing the test logs.

src/test/shell/bazel/bazel_symlink_test.sh
brandjon: Thanks for fixing the pre-existing occurrences too.

src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java
```diff
@@ -252,69 +242,69 @@ public void symlink(
...
       Object /* String or None */ targetPath,
       Boolean isExecutable,
-      Object /* String or None */ progressMessage)
+      Object /* String or None */ progressMessageUnchecked)
       throws EvalException {
     context.checkMutable("actions.symlink");
 
-    if (!exactlyOneNotNone(targetFile, targetPath)) {
-      throw Starlark.errorf("\"target_file\" and \"target_path\" cannot be set at the same time");
+    if ((targetFile == Starlark.NONE) == (targetPath == Starlark.NONE)) {
```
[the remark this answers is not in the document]
brandjon: This is one of those instances where I should really double check the seemingly baffling simple things before I hit send. Nevermind.

brandjon:
@lberki, when you get a chance please weigh in on these three concerns:

* [normalization of symlink path](https://github.com/bazelbuild/bazel/pull/10695#discussion_r409257298)

* [`is_executable`](https://github.com/bazelbuild/bazel/pull/10695#discussion_r409259264)

* [safe to mark non-experimental](https://github.com/bazelbuild/bazel/pull/10695#issuecomment-614310187)

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -159,12 +159,18 @@
...
-              + "<a href=\"#declare_symlink\"><code>declare_symlink()</code></a>), a dangling "
-              + "symlink is allowed.</p>",
+          "Creates an action that writes a symlink in the file system."
+              + "<p>This function must be called with exactly one of <code>target_file</code> and "
+              + "<code>target_path</code> specified.</p>"
+              + "<p>If <code>target_file</code> is used, then <code>output</code> must be declared "
+              + "as a regular file (such as by using "
+              + "<a href=\"#declare_file\"><code>declare_file()</code></a>). It will actually be "
+              + "created as a symlink that points to the path of <code>target_file</code>.</p>"
+              + "<p>If target_path is used instead, then <code>output</code> must be declared as a "
```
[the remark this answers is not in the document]
Yannic: done

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
```diff
@@ -191,22 +194,22 @@
...
             named = true,
             positional = false,
             defaultValue = "False",
             doc =
-                "Whether the output file should be executable. "
-                    + "<p>If this is <code>True<code>, <code>target_path</code> must also be "
-                    + "executable.</p>"),
+                "May only be used with <code>target_file</code>, not <code>target_path</code>. "
+                    + "If true, when the action is executed, the <code>target_file</code's path is "
+                    + "checked to confirm that it is executable, and an error is reported if it is "
```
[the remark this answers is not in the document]
Yannic: done

src/test/shell/bazel/bazel_symlink_test.sh
Yannic: Ok, done

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -211,18 +265,31 @@ EOF
...
+    warn "Skipping test on Windows: Unresolved symlinks are not supported yet"
+    return 0
+  fi
+
   mkdir -p a
   cat > a/a.bzl <<'EOF'
 def _bad_symlink_impl(ctx):
     symlink = ctx.actions.declare_symlink(ctx.label.name)
-    ctx.actions.write(symlink, ctx.attr.link_target)  # Oops, should be "symlink"
+    # Oops, should be "dangling_symlink"
```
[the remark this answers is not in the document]
Yannic: Probably it is. Done.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
lberki: @lberki here. Yes, this is correct. This is because unresolved symlinks are essentially funky text files that the operating system handles differently.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/SkylarkActionFactoryApi.java
lberki: Yep, exactly. Do I understand correctly that the only concern was that one instance of `target_file` was accidentally `target_path`?

brandjon: The concern was the typo and providing enough detail. I just wanted to be sure the additional detail I added wasn't blatantly wrong. :)

lberki:
re: https://github.com/bazelbuild/bazel/pull/10695#issuecomment-614310187, we are on the same page.

1. The reason why unresolved symlinks are marked as experimental is exactly the interaction with remote execution (maybe it works, but I certainly haven't tried). In addition, it hasn't gone through a proper design review. `ctx.actions.symlink()` itself for non-unresolved (double negation! say, regualr symlinks) is a pretty safe change, it only exposes old internal functionality that hasn't changed since forever.

2. Yep. You may want to ask @laurentlb what he thinks about it, but I don't think he'd object.

brandjon:
Thanks, that should be everything then!

brandjon:
(In internal merge, I added a documentation note to `target_path` that it's experimental and requires setting the flag.)

PR #10705, opened 2020-02-04
Consolidate creation of the ConfiguredAttributeMap and clean up casts in

katre:
DependencyResolver.

Work towards toolchain transition, #10523.
GOOGLE:
RELNOTES: None.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
@@ -262,7 +262,10 @@ public abstract class DependencyResolver {
     BuildConfiguration config = node.getConfiguration();
     OrderedSetMultimap<DependencyKind, Label> outgoingLabels = OrderedSetMultimap.create();
 
-    // TODO(bazel-team): Figure out a way to implement the below using LabelVisitationUtils.
+    // TODO(bazel-team): Figure out a way to implement the below (and partiallyResolveDependencies) using
+    // LabelVisitationUtils.
+    Rule fromRule = null;
+    ConfiguredAttributeMapper attributeMap = null;
     if (target instanceof OutputFile) {
       Preconditions.checkNotNull(config);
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
```diff
@@ -262,7 +262,10 @@ static PartiallyResolvedDependency of(
     BuildConfiguration config = node.getConfiguration();
     OrderedSetMultimap<DependencyKind, Label> outgoingLabels = OrderedSetMultimap.create();
 
-    // TODO(bazel-team): Figure out a way to implement the below using LabelVisitationUtils.
+    // TODO(bazel-team): Figure out a way to implement the below (and partiallyResolveDependencies) using
```
gregestren: Is there even a need to pass configConditions as map in the method signature now? It looks like it's never used anywhere in DependencyResolver except to create the attribute mapper.

katre: The configConditions map is used on line 280 to create the ConfiguredAttributeMapper, and then isn't used again. Are you suggesting hoisting the ConfiguredAttributeMapper even higher, to the callers of dependentNodeMap? I'm not sure that makes sense since the mapper is only needed if the target is a rule, and that's checked here. Ideally we'd replace this with a visitor and remove all of the type checking entirely, but for now this seems the cleanest option.

gregestren: Yes, that's what I meant. But I forgot that `ConfiguredAttributeMapper` itself uses the whole Map.. When I wrote the comment I thought the `Label` part of the `Map<Label, ConfigMatchingProvider>` was unread code. But that's not actually true.

katre:
Committed in b0f47b2128856cd185ef33d16a626d63df313231.

PR #10708, opened 2020-02-05
Make Java code coverage output compatible with genhtml

### LINKED ISSUE #10676: Missing Java branch coverage
Java test coverage does not seem to include branch data by default (using `genhtml`) and I can't find any information on how to enable it.

Is this a missing feature? Or a bug? Or is there some other way to generate a html report that includes branch data?

These are the commands I use:

```bash
bazel coverage --jobs=${COVERAGE_CPUS:-1} --instrumentation_filter //src/main/... --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main -- ...
genhtml -p $(pwd) -title rules_antlr -o $destdir --branch-coverage --legend bazel-out/_coverage/_coverage_report.dat
``` 
`genhtml` always shows "branches...: no data found" even if "lcov_branch_coverage = 1" is set in the lcov configuration file.

I'm using bazel 2.0.0 on Debian.

marcohu:
- Switch output from BA to BRDA that is understood by genhtml
- Ditch the custom branch detail gathering and use the Jacoco data which includes filtering
- Differentiate between branch not executed and not taken
- Display all branches per line, not just a single one

Fixes #10676

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
@@ -123,16 +123,15 @@ public class JacocoCoverageRunner {
     // report. The lcov formatter doesn't seem to care, and we're only using one bundle anyway.
     final IBundleCoverage bundleCoverage = analyzeStructure();
 
-    final Map<String, BranchCoverageDetail> branchDetails = analyzeBranch();
-    createReport(bundleCoverage, branchDetails);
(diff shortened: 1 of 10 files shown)


marcohu:
`src/test/shell/bazel/bazel_coverage_java_test.sh` does not use the modified output generator, but I don't know how to alter the bazel invocations to achieve that.

The expected coverage results have to be adjusted as well, but won't yield the desired results without using the local output generator.

PR #10712, opened 2020-02-05
Add cc_library option to ensure include_prefix behaves the same regar…

elklein:
Currently include_prefix behaves differently (at least for system includes) when include_prefix matches the path of the library within a larger project.

If you have a library located at path foo/bar (with BUILD file at foo/bar/BUILD), and an include file named Log.h, currently if you specify an include_prefix of "foo/bar", then dependent code cannot do the following and have it work:

#include <foo/bar/Log.h>

Log.h in that case will not be found by the dependent code. If you switch the <> to "", it will be found (because Bazel currently sticks in an "-iquote ." argument to the compile command line). Furthermore, if you change the include_prefix to be "foobar", and change the includes to be:

#include <foobar/Log.h>

it will be also be found.

In cases where it's not possible to change the #include in dependent source code, then the only WAR I've found so far is to put "-isystem ." in the copts of all dependent code, but that's not a really great solution.

This code change allows cc_library rules to opt-in to having include_prefix always treated the same, regardless of library path.

It's possible that this feature: https://docs.google.com/document/d/1PKTDhE9rHzd0WQJY2czjg_vPT28fufGcl2Gain6mUaI/edit?pli=1#heading=h.9grrl7f2c55j
will address the issue as well, but it's not clear what the status of that is, and what the timeline for adoption would be.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppRuleClasses.java
@@ -449,6 +449,19 @@ public class BazelCppRuleClasses {
           prefix is added.
           <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
           .add(attr("include_prefix", STRING))
+          /* <!-- #BLAZE_RULE($cc_library).ATTRIBUTE(always_create_virtual_includes) -->
+          Set to ensure that include_prefix behaves the same when it mirrors the
+          workspace directory tree as when it does not.
+
+          <p>Without this, if include_prefix is the relative path from the workspace
+          root to the library directory, it is treated differently. Specifically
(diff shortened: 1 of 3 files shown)


elklein:
This is actually an old pull request, but the old one got accidentally closed due to some branch rearrangement on my end. I had moved the code in question to a side-branch off of my forked master (which I should have done to start with), and pushed my master back to github which accidentally closed the old pull request.

The old request was: https://github.com/bazelbuild/bazel/pull/10181

It's probably just as well because I needed to bring it back up again and see if I could get some eyes on it anyway. Please let me know if you think this is an OK approach to the issue we're facing.

elklein:
@scele: Just a note that this has been re-filed.

elklein:
Closing this as it is no longer relevant. We have rearchitected a bunch of things and have abandoned this.

PR #10717, opened 2020-02-05
Add support for bash completion of --config= names

elklein:
This change supports the following command line arguments:
--bazelrc=<file>
--nosystem_rc
--noworkspace_rc
--nohome_rc

And supports finding bazelrc files in all the documented locations
as-per the docs at:
https://docs.bazel.build/versions/2.0.0/guide.html#bazelrc

It can handle command inheritance. So if you specify a "test" command
it knows to look at test, build, and common configs.

It also can deal with import and try-import commands, and can chase
those down to any nesting level.

### DIFF
--- a/scripts/bazel-complete-template.bash
+++ b/scripts/bazel-complete-template.bash
@@ -378,6 +378,198 @@ _bazel__expand_options() {
   fi
 }
 
+# Usage: _bazel__abspath <file>
+#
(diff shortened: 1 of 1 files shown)


philwo:
@elklein Thank you for the contribution! This is a very nice feature to have. 😃

Issue #10535, opened 2020-01-07
Show debug information about multiplex worker vs regular persistent worker


fenghaolw:
### Description of the problem / feature request:

Currently when we replace our persistent worker with [multiplex worker](https://docs.bazel.build/versions/master/multiplex-worker.html), it is unclear which one is being used. `--worker_verbose` shows "Create new ... worker (id #)", but does not display the debug information for multiplex worker.

### Feature requests: what underlying problem are you trying to solve with this feature?

Add additional debug information for better developer experience.

### Any other information, logs, or outputs that you want to share?

It sounds to me we could add additional information in [WorkerKey](https://github.com/bazelbuild/bazel/blob/8c96c0614684243b0a51d0ad8ddc27b0258b4af1/src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java) and then print this debug information in [WorkerFactory](https://github.com/bazelbuild/bazel/blob/8c96c0614684243b0a51d0ad8ddc27b0258b4af1/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java)?

susinmotion:
@borkaehw @SrodriguezO

borkaehw:
Sounds like a good suggestion. I will check if I have bandwidth in the following few weeks. Anyone is welcome to implement this as well. Thanks.

PR #10732, opened 2020-02-07
Show users that multiplex-worker is in use

borkaehw:
Fix: #10535

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
@@ -55,7 +55,7 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
   public Worker create(WorkerKey key) throws Exception {
     int workerId = pidCounter.getAndIncrement();
     Path logFile =
-        workerBaseDir.getRelative("worker-" + workerId + "-" + key.getMnemonic() + ".log");
+        workerBaseDir.getRelative(key.getWorkerTypeName() + "-" + workerId + "-" + key.getMnemonic() + ".log");
 
     Worker worker;
     boolean sandboxed = workerOptions.workerSandboxing || key.mustBeSandboxed();
@@ -77,9 +77,10 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
       reporter.handle(
           Event.info(
               String.format(
-                  "Created new %s %s worker (id %d), logging to %s",
+                  "Created new %s %s %s (id %d), logging to %s",
                   sandboxed ? "sandboxed" : "non-sandboxed",
                   key.getMnemonic(),
+                  key.getWorkerTypeName(),
                   workerId,
                   logFile)));
     }
@@ -89,7 +90,7 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
   Path getSandboxedWorkerPath(WorkerKey key, int workerId) {
     String workspaceName = key.getExecRoot().getBaseName();
     return workerBaseDir
-        .getRelative("worker-" + workerId + "-" + key.getMnemonic())
+        .getRelative(key.getWorkerTypeName() + "-" + workerId + "-" + key.getMnemonic())
         .getRelative(workspaceName);
   }
 
@@ -110,7 +111,7 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
       reporter.handle(
           Event.info(
               String.format(
-                  "Destroying %s worker (id %d)", key.getMnemonic(), p.getObject().getWorkerId())));
+                  "Destroying %s %s (id %d)", key.getMnemonic(), key.getWorkerTypeName(), p.getObject().getWorkerId())));
     }
     p.getObject().destroy();
   }
@@ -126,8 +127,8 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
       StringBuilder msg = new StringBuilder();
       msg.append(
           String.format(
-              "%s worker (id %d) can no longer be used, because its files have changed on disk:",
-              key.getMnemonic(), worker.getWorkerId()));
+              "%s %s (id %d) can no longer be used, because its files have changed on disk:",
+              key.getMnemonic(), key.getWorkerTypeName(), worker.getWorkerId()));
       TreeSet<PathFragment> files = new TreeSet<>();
       files.addAll(key.getWorkerFilesWithHashes().keySet());
       files.addAll(worker.getWorkerFilesWithHashes().keySet());
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
@@ -54,8 +54,9 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
   @Override
   public Worker create(WorkerKey key) throws Exception {
     int workerId = pidCounter.getAndIncrement();
+    String workTypeName = WorkerKey.makeWorkerTypeName(key.getProxied());
     Path logFile =
-        workerBaseDir.getRelative(WorkerKey.makeWorkerTypeName(key.getProxied()) + "-" + workerId + "-" + key.getMnemonic() + ".log");
+        workerBaseDir.getRelative(workTypeName + "-" + workerId + "-" + key.getMnemonic() + ".log");
 
     Worker worker;
     boolean sandboxed = workerOptions.workerSandboxing || key.mustBeSandboxed();
@@ -80,7 +81,7 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
                   "Created new %s %s %s (id %d), logging to %s",
                   sandboxed ? "sandboxed" : "non-sandboxed",
                   key.getMnemonic(),
-                  WorkerKey.makeWorkerTypeName(key.getProxied()),
+                  workTypeName,
                   workerId,
                   logFile)));
     }
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
@@ -55,7 +55,7 @@ final class WorkerFactory extends BaseKeyedPooledObjectFactory<WorkerKey, Worker
   public Worker create(WorkerKey key) throws Exception {
     int workerId = pidCounter.getAndIncrement();
     Path logFile =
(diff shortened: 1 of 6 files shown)


borkaehw:
cc @fenghaolw this is the PR to address the issue you opened. Sorry for getting back to this late.

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
```diff
@@ -109,12 +110,17 @@ public WorkerSpawnRunner(
...
+      return "multiplex-worker";
+    } else {
+      return "worker";
+    }
   }
 
   @Override
   public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
       throws ExecException, IOException, InterruptedException {
+    this.supportsMultiplexWorkers = Spawns.supportsMultiplexWorkers(spawn);
```
borkaehw: It looks like a bad idea. Can someone help?

philwo: Yes, I don't think this will work (reliably), because the WorkerSpawnRunner instance is shared for all worker strategy executions (multiplex and non-multiplex). I think this might cause the name of all running worker actions to change to "multiplex-worker" or "worker", depending on whether the most recently executed action was a multiplexed one or not.

I can't immediately think of a good solution for this, except maybe instantiating two WorkerSpawnRunners (one for multiplexed workers, one for non-multiplexed workers). But it might make the code much more complex.

Maybe @susinmotion or @jmmv have a good idea.

susinmotion: What about appending 'multiplex' to the existing name of the worker if it's multiplex?

borkaehw: I think the challenge is that when `WorkerSpawnRunner` is constructed, we don't have `Spawn` object to know if it's multiplex or not. I don't think we should set it in `exec()` since `this.supportsMultiplexWorkers` could be undefined at a certain period of time.

@susinmotion do you mean we could undo the change in `getName()`? It's nice to see information when build is in progress though, like
```
<Mnemonic> <Target> 10s multiplex-worker
```

susinmotion: Yes, I did mean undoing the change in getName(), but I also see why that'd be nice info to have.
Let me take a closer look!

jmmv: What we care about is the name that is used for logging, right?

If so, why not just change the `context.report()` calls in this file to specify the exec-specific name instead of calling `getName()`? They all happen within `exec()`, so at that point we have the right context information.

borkaehw: Sounds like a great idea. Just force-pushed, please take another look.

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
```diff
@@ -110,6 +110,15 @@ public boolean getProxied() {
     return proxied;
   }
 
+  /** Getter function for worker type name. */
```
jmmv: nit: Should be something like `Returns a user-friendly name for this worker type.`

https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
```diff
@@ -112,6 +112,14 @@ public String getName() {
     return "worker";
   }
 
+  private String getMultiplexPrefixName(boolean proxied) {
+    if (proxied) {
+      return "multiplex-" + getName();
+    } else {
+      return getName();
+    }
+  }
```
jmmv: This logic is duplicated in `WorkerKey`. Maybe it can be reused?

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
```diff
@@ -112,6 +112,14 @@ public String getName() {
     return "worker";
   }
 
+  private String getMultiplexPrefixName(boolean proxied) {
```
jmmv: nit: This name is a bit confusing given that the return value may not be related to multiplexing at all. Maybe `getNameForReporting` or something? And add a docstring to indicate what this does and why it's separate from `getName`.

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
borkaehw: Done.

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
borkaehw: Good idea, done.

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
borkaehw: Sounds reasonable. I think we have no choice but to keep the logic in `WorkerSpawnRunner` since there is no `WorkerKey` yet when we call the first `context.report()`. But in order to get rid of the duplicate logic, we will have to use a new member variable.

Some explanation of design choices:
1. We could replace the second `context.report()` by
```
context.report(ProgressStatus.EXECUTING, key.getWorkerTypeName());
```
But I like it to be consistent calling the same function `getNameForReporting()` in same file. What do you think?
2. It looks like `Spawns.supportsMultiplexWorkers(spawn)` has been called multiple times in `WorkerSpawnRunner`. We could store it in a variable in `exec()` and pass it all the way down to `actuallyExec()` and `execInWorker()`. It may look cleaner. What do you think?

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
```diff
@@ -68,6 +71,8 @@
     this.mustBeSandboxed = mustBeSandboxed;
     /** Set it to true if this job should be run with WorkerProxy. */
     this.proxied = proxied;
+    /** If multiplex worker is used, it should be "multiplex-worker", otherwise return "worker" */
```
jmmv: This comment doesn't make a lot of sense here. Where do these magic strings come from? If they are important, they should be part of the docstring. And if they must be these specific values, there should be some validation. I would remove this.

src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java
jmmv: Have a static function `WorkerKey#makeWorkerTypeName(String name)` that builds the name, and then call it from wherever you need. No need for member fields.

borkaehw: Thanks, I think it will work. Just to verify, would the function be `WorkerKey#makeWorkerTypeName(boolean proxied)` instead? Passing `String name` to the function doesn't make sense to me, would you mind to elaborate more?

borkaehw: I assume `boolean` makes more sense, just updated. Let me know if you have different thought.

borkaehw:
@jmmv I have addressed all the comments. Please take another look when you get a chance. Thanks.

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
```diff
@@ -110,6 +110,15 @@ public boolean getProxied() {
     return proxied;
   }
 
+  /** Returns a user-friendly name for this worker type. */
+  public static String makeWorkerTypeName(boolean proxied) {
```
jmmv: I think it'd be better if this took a key as input instead of a boolean, given that the input seems to always come from a key.

But humm... hold on. Given that `WorkerKey` knows whether it's proxied or not already, shouldn't this be a `getWorkerTypeName` member function without arguments?

borkaehw: > given that the input seems to always come from a key

Not always, the input on this line comes from `Spawn`.
https://github.com/bazelbuild/bazel/pull/10732/files#diff-4ebed75948b31f1a40fc7f782b6d31ceR127

I think since this is a static function, we don't always have the information about proxy.

Like on the same line https://github.com/bazelbuild/bazel/pull/10732/files#diff-4ebed75948b31f1a40fc7f782b6d31ceR127, we can only tell by passing argument to the static function.

If we prefer to take a key as argument, I would like to know what would the key look like. Previously you suggested `String name`, what is a `name`?

src/main/java/com/google/devtools/build/lib/worker/WorkerFactory.java
```diff
@@ -77,9 +77,10 @@ public Worker create(WorkerKey key) throws Exception {
       reporter.handle(
           Event.info(
               String.format(
-                  "Created new %s %s worker (id %d), logging to %s",
+                  "Created new %s %s %s (id %d), logging to %s",
                   sandboxed ? "sandboxed" : "non-sandboxed",
                   key.getMnemonic(),
+                  WorkerKey.makeWorkerTypeName(key.getProxied()),
```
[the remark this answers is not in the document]
borkaehw: I assume you meant the entire `WorkerKey.makeWorkerTypeName(key.getProxied())` statement, not just the `key.getProxied()`, since `key.getProxied()` three times in this functoin.

Let me know if you have more concerns. Thanks.

borkaehw:
@jmmv Friendly ping here to move this forward.

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
jmmv: All usages of `makeWorkerTypeName` are of the form: `WorkerKey.makeWorkerTypeName(key.getProxied())`, which is a very strange way of saying `key.getProxiedWorkerTypeName()`.

There is only once exception, which is `WorkerKey.makeWorkerTypeName(Spawns.supportsMultiplexWorkers(spawn))`. Maybe that one just needs special-casing: make the two strings constants and duplicate the conditional at that call site? Or add a `WorkerKey.getProxiedWorkerTypeName` member function that calls the static `makeWorkerTypeName`?

I guess it doesn't matter much and any of these "solutions" are actually worse. So I think this is good enough; can't see of an easy way to handle this single corner case.

jmmv:
Alright, now I gotta deal with the import, which I haven't done many times. Hang on. Don't hesitate to keep poking me if not done by Monday...

src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java
borkaehw: Sounds good, thanks! Let's keep it in mind, I am happy to improve this when we have better solution.

borkaehw:
Thanks for merging.

PR #10734, opened 2020-02-08
Tweak third_party/BUILD for FreeBSD.

aldersondrive:
This way there's no need for FreeBSD's `bazel` port to [patch `third_party/BUILD`](https://github.com/freebsd/freebsd-ports/blob/f5c3ba467a9cc5e834c157b0364d327425981c24/devel/bazel/files/patch-third_party__BUILD) anymore. Also, this change moves one step closer to a bootstrap build on FreeBSD succeeding using unmodified sources.

https://github.com/bazelbuild/bazel/pull/10639 made the equivalent change for OpenBSD. AFAICS the same kind of change would be helpful for FreeBSD also.

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -469,6 +469,7 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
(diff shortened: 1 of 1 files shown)


aldersondrive:
I was going to send this to @aehlig, but he has apparently vanished from GitHub. @philwo and @jmmv, maybe one of you could review this change, since it's a follow-up to https://github.com/bazelbuild/bazel/pull/10639?

philwo:
Thanks! Merged in 554e0f7359c7754195125c254954a33bb99a31c9.

PR #10741, opened 2020-02-10
Remove git notes usage from release process

meteorcloudy:
Previously we store releae name, candidate number, cherry-picks and
release notes info in git repo using `git notes`.  To remove this
complexity, we remove all git notes usage in this change.

Instead of using git notes, we parse release name and candidate number
from branch name or tag name.  For cherry-picks, we rely on the git
patch-id to find the original commit hash. And for release notes, there
is no need to store it in git notes, because we always manually modify
the generated release note.

### DIFF
--- a/scripts/release/common.sh
+++ b/scripts/release/common.sh
@@ -41,7 +41,7 @@ function __git_commit_subject() {
 
 # Returns the branch name of the current git repository
 function git_get_branch() {
-  git symbolic-ref --short HEAD 2>/dev/null
+  git symbolic-ref --short HEAD
 }
 
 # Returns the tag name of the current git repository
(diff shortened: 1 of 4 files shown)


scripts/release/release.sh
```diff
@@ -161,43 +124,27 @@ function __apply_cherry_picks() {
...
   shift 2
-  local branch_name="release-${release_name}"
+  local branch_name="release-${release_name}rc${force_rc}"
 
   # Fetch everything from remote repositories to avoid conflicts
   git fetch -f "${RELEASE_REPOSITORY}"
-  git fetch -f "${RELEASE_REPOSITORY}" 'refs/notes/*:refs/notes/*'
 
-  local last_release="$(git rev-parse --verify "${branch_name}" 2>/dev/null || true)"
+
```
philwo: nit: remove duplicate empty line?

meteorcloudy: Done.

scripts/release/common.sh
```diff
@@ -41,22 +41,30 @@ function __git_commit_subject() {
 
 # Returns the branch name of the current git repository
 function git_get_branch() {
-  git symbolic-ref --short HEAD
+  git symbolic-ref --short HEAD 2>/dev/null
```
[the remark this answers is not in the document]
meteorcloudy: Just want to make the output of generate_email more clean. But, indeed it's better to do it in `get_release_candidate` and `get_release_name`, where we don't care when it fails.

Issue #10746, opened 2020-02-10
building bazel libcpu_profiler fails to find jni_md.h (ppc64le)


clnperez:
### Description of the problem / feature request:
compiling from master on ppc64le fails:

> bazel-out/ppc-fastbuild/bin/external/bazel_tools/tools/jdk/include/jni.h:45:10: fatal error: jni_md.h: No such file or directory

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

> bazel build --sandbox_debug -s --verbose_failures --host_javabase=@local_jdk//:jdk //src:bazel

### What operating system are you running Bazel on?
> Ubuntu 18.04.3

### What's the output of `bazel info release`?
building with:
> release 1.1.0- (@non-git)

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

> https://github.com/Unicamp-OpenPower/bazel-build/blob/master/.travis.yml

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

root@ede9b339f10f:/bazel# git remote get-url origin ; git rev-parse master ; git rev-parse HEAD
https://github.com/bazelbuild/bazel.git
1ae1c16133699bcd6ae90f00cec6140ad4146cd3
314cd3353b018420598188cf54c3f97358baa2e3

###  Have you found anything relevant by searching the web?

> no

### Any other information, logs, or outputs that you want to share?

This commit seems to have broken us: https://github.com/bazelbuild/bazel/commit/314cd3353b018420598188cf54c3f97358baa2e3

More or the error here: https://gist.github.com/clnperez/781149e8c6882ea668d06483325e1151

clnperez:
@alandonovan -- i've tried to figure out how to get this to work and can't. i thought maybe it was because we build using the local jdk, but i think that's the default now. i also tried that on x86 and this works fine. so now i'm guessing there's some place i can't find with an `linux_x86_64"  option and a blank `default` for this header file. Any suggestions?

alandonovan:
Take a look at the BUILD.oss file in the linked commit: 
https://github.com/bazelbuild/bazel/commit/314cd3353b018420598188cf54c3f97358baa2e3#diff-96bd71d4e041b52d389d631025d2c09fR167
[...]

PR #10748, opened 2020-02-11
Add aarch64 configuration condition for jni gcc includes

garyschulte:
Very light PR which adds arm64 platform rule for building bazel.

When building bazel from sources on linux_aarch64, the build would fail during c++ compilation, complaining about a missing jni platform include:

```
INFO: Analyzed target //src:bazel (1 packages loaded, 194 targets configured).
INFO: Found 1 target...
ERROR: /home/ubuntu/dev/bazel/src/main/java/com/google/devtools/build/lib/syntax/BUILD:150:1: C++ compilation of rule '//src/main/java/com/google/devtools/build/lib/syntax:libcpu_profiler.so' failed (Exit 1) gcc failed: error executing command /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 24 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from src/main/java/com/google/devtools/build/lib/syntax/cpu_profiler_unimpl.cc:17:0:
bazel-out/aarch64-fastbuild/bin/external/bazel_tools/tools/jdk/include/jni.h:45:10: fatal error: jni_md.h: No such file or directory
 #include "jni_md.h"
          ^~~~~~~~~~
compilation terminated.
Target //src:bazel failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 2.737s, Critical Path: 0.73s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

After adding aarch64 select, the build is successful.

PR #10759, opened 2020-02-11
fix cpu profiler compile on ppc64le

clnperez:
Add conditions for linux/ppc instead of just linux/x86_64.

Other headers might be in the same place (for s390x, arm*, etc.), but
would need to be tested first. This fix is only tested on ppc64le.

Fixes https://github.com/bazelbuild/bazel/issues/10746

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/syntax/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/syntax/BUILD
@@ -165,6 +165,7 @@ cc_library(
     hdrs = ["@bazel_tools//tools/jdk:jni_header"] + select({
         "//src/conditions:linux_x86_64": ["@bazel_tools//tools/jdk:jni_md_header-linux"],
         "//src/conditions:linux_aarch64": ["@bazel_tools//tools/jdk:jni_md_header-linux"],
+        "//src/conditions:linux_ppc": ["@bazel_tools//tools/jdk:jni_md_header-linux"],
         "//src/conditions:darwin": ["@bazel_tools//tools/jdk:jni_md_header-darwin"],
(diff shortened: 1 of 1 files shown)


clnperez:
Looks like the same as https://github.com/bazelbuild/bazel/pull/10748. Maybe we could join forces and just do a bare linux case. :)

alandonovan:
If you can find a condition target for bare Linux (sans architecture), feel free to use it, but I didn't see one. (In any case, the jni rule logically belongs in the Bazel Java rules, not here.)

clnperez:
Can this be merged? I'm fine with separate arch conditions and am not looking into a bare linux one (bandwidth).

jr-santos98:
@alandonovan @clnperez I had the same problem, and I came to the same conclusion. It's build now for ppc64le.

clnperez:
looks like the arm pr got in. i rebased this one and re-pushed.

clnperez:
@alandonovan what else does this need to get merged?

aiuto:
Whoops. Alan is out on vacation this week, so he probably missed on that. 
@iirina You are on triage this week, can you import/merge this in the morning? 
If I don't see it merged I can try to kick it off Friday, but I won't be in the office until  the afternoon, US Eastern time.

clnperez:
Ah. Ok thanks @aiuto

aiuto:
My bad. I got distracted on Friday and let this drop. Sending it through the CI pipeline now.

james-stephenson:
Unfortunately this is still broken on armhf / armv7l (e.g. Raspberry Pi and similar), as well as any other architectures.

It seems like a default "linux" case would solve all of these and not make the `select` a mess.

PR #10760, opened 2020-02-11
Disable dd stdout.

benjaminp:
This fixes noise from the install base key action like the following:
```
1+0 records in
1+0 records out
32 bytes copied, 0.113519 s, 0.3 kB/s
```

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -6,9 +6,7 @@ load(":rule_size_test.bzl", "rule_size_test")
 
 exports_files(["jdeps_modules.golden"])
 
-# We avoid using the `head` tool's `-c` option, since it does not exist on
-# OpenBSD; here we use `dd` instead.
-md5_cmd = "set -e -o pipefail && echo $(SRCS) | sort | xargs %s | %s | dd bs=32 count=1 > $@"
+md5_cmd = "set -e -o pipefail && %s $(SRCS) | %s | %s > $@"
 
 [genrule(
     name = "install_base_key-file" + suffix,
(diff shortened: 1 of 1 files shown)


aiuto:
@jmmv :  You're an alternate OS person. This looks good to me. If you can take a look and LGTM,  I'll shepherd it in since I'm triage this week.

jmmv:
From a portability perspective, this seems fine. Just have a question.

aiuto:
And that question is.... ?

jmmv:
It's above, inline with the diff.

src/BUILD
```diff
@@ -6,9 +6,7 @@ load(":rule_size_test.bzl", "rule_size_test")
 
 exports_files(["jdeps_modules.golden"])
 
-# We avoid using the `head` tool's `-c` option, since it does not exist on
-# OpenBSD; here we use `dd` instead.
-md5_cmd = "set -e -o pipefail && echo $(SRCS) | sort | xargs %s | %s | dd bs=32 count=1 > $@"
+md5_cmd = "set -e -o pipefail && echo $(SRCS) | sort | xargs %s | %s | %s > $@"
```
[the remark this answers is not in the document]
benjaminp: Looks like it's been there for a while (c4e74d1e531b06d2c8be2fd7c265a14abbd2fc4f).

I agree, though, that it does nothing. It also shouldn't really be required as making `$(SRCS)` deterministic is Bazel's problem.

src/BUILD
```diff
@@ -6,9 +6,7 @@ load(":rule_size_test.bzl", "rule_size_test")
 
 exports_files(["jdeps_modules.golden"])
 
-# We avoid using the `head` tool's `-c` option, since it does not exist on
-# OpenBSD; here we use `dd` instead.
-md5_cmd = "set -e -o pipefail && echo $(SRCS) | sort | xargs %s | %s | dd bs=32 count=1 > $@"
+md5_cmd = "set -e -o pipefail && echo $(SRCS) | xargs %s | %s | %s > $@"
```
jmmv: So now, can we do `%s $(SRCS)` and avoid the `echo` and `xargs` calls, or are these command lines too long for this to work?

benjaminp: tests say we can dispense with `xargs`

src/BUILD
```diff
@@ -6,9 +6,7 @@ load(":rule_size_test.bzl", "rule_size_test")
 
 exports_files(["jdeps_modules.golden"])
 
-# We avoid using the `head` tool's `-c` option, since it does not exist on
-# OpenBSD; here we use `dd` instead.
-md5_cmd = "set -e -o pipefail && echo $(SRCS) | sort | xargs %s | %s | dd bs=32 count=1 > $@"
+md5_cmd = "set -e -o pipefail && %s $(SRCS) | %s | %s > $@"
```
aiuto: If this were "set -e -o pipefail && %s $(SRCS) | **sort |** %s | %s > $@"
then that would be immune to Bazel reordering $SRCS. (if we still want to preserve that intent)

benjaminp: The current implementation should be deterministic per (Bazel version, repo version), which should be good enough?

aiuto: Sure. It was more a suggestion. I'm fine with what you have.
I can merge when I am back at my desk.

Issue #7135, opened 2019-01-16
Bazel cannot build itself in arm64 docker container


seank-com:
### Description of the problem / feature request:

Bazel builds successfully but the ```compile.sh``` proceeds to build Bazel with Bazel and that fails. This is likely a bug in qemu, but I don't know where to get enough detail out of Bazel to investigate it any further.

### Feature requests: what underlying problem are you trying to solve with this feature?

```bash
ERROR: /root/bazel-0.21.0/src/BUILD:109:1: PythonZipper src/create_embedded_tools.zip failed (Exit 255): zipper failed: error executing command
  (cd /tmp/bazel_AM1PL7Q9/out/execroot/io_bazel && \
  exec env - \
    PATH=/bin:/usr/bin \
  bazel-out/host/bin/external/bazel_tools/third_party/ijar/zipper cC bazel-out/host/bin/src/create_embedded_tools.zip @bazel-out/host/bin/src/create_embedded_tools.zip-0.params)
Execution platform: @bazel_tools//platforms:host_platform
Target //src:bazel_nojdk failed to build
INFO: Elapsed time: 6149.498s, Critical Path: 1611.76s
INFO: 1641 processes: 1404 local, 237 worker.
FAILED: Build did NOT complete successfully

ERROR: Could not build Bazel
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Dockerfile contents

```dockerfile
FROM multiarch/ubuntu-core:arm64-bionic as base

RUN apt-get update -y
RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
  python3 python3-dev python3-pip 

RUN pip3 install --upgrade pip

FROM base as bazel

RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
  build-essential openjdk-8-jdk zip unzip wget python

WORKDIR /root

RUN wget https://github.com/bazelbuild/bazel/releases/download/0.21.0/bazel-0.21.0-dist.zip
RUN mkdir bazel-0.21.0
RUN unzip bazel-0.21.0-dist.zip -d bazel-0.21.0

WORKDIR /root/bazel-0.21.0
```

From a terminal running in the directory where Dockerfile is

```bash
$ docker run --rm --privileged multiarch/qemu-user-static:register --reset
$ docker build -t multiarch-bld .
$ docker run -it multiarch-bld
# ./compile.sh
```

[...]

PR #10761, opened 2020-02-11
Add READ access of mmap() in docker env

yhwang:
In docker multi-architecture environment, calling the
mmap() with only PROT_WRITE would fail the following
write(). Adding PROT_READ access would solve the issue.
The logic there only needs WRITE access, however, it
should be fine to assign READ access too. Actually,
PROT_WRITE is offen implementated as PROT_READ|PROT_WRITE.

fix: #7135.

Signed-off-by: Yihong Wang <yh.wang@ibm.com>

### DIFF
--- a/third_party/ijar/mapped_file_unix.cc
+++ b/third_party/ijar/mapped_file_unix.cc
@@ -113,7 +113,8 @@ MappedOutputFile::MappedOutputFile(const char* name, size_t estimated_size)
   size_t mmap_length =
       std::min(static_cast<size_t>(estimated_size + sysconf(_SC_PAGESIZE)),
                std::numeric_limits<size_t>::max());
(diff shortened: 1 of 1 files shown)


yhwang:
I verified this fix on:
```
Linux pi64 4.19.76-linuxkit #1 SMP Thu Oct 17 19:31:58 UTC 2019 aarch64 aarch64 aarch64 GNU/Linux
```
and
```
Linux pi4 4.19.76-linuxkit #1 SMP Thu Oct 17 19:31:58 UTC 2019 armv7l armv7l armv7l GNU/Linux
```

and used the `Docker version 19.03.5, build 633a0ea` on MacOS

aiuto:
@philwo  I'm not really sure if this is your area right now. It seems to be correct to me, but I'm not sure who really owns ijar.

PR #10762, opened 2020-02-12
docs.bazel.build design refresh

jin:
Staging site: https://bazel-docs-staging.netlify.com

Discussion: https://groups.google.com/d/msg/bazel-dev/hjRP7JTtZ_U/9SZYZUSBDAAJ

Fixes https://github.com/bazelbuild/bazel/issues/10791

### DIFF
--- a/scripts/docs/Gemfile
+++ b/scripts/docs/Gemfile
@@ -1,20 +1,5 @@
-gem 'jekyll', '3.1.6'
-
-# Jekyll 3.1.6 automatically depends on these Gems:
-# (see https://github.com/jekyll/jekyll/blob/v3.1.6/jekyll.gemspec)
-#
-# 'liquid',       '~> 3.0'
-# 'kramdown',     '~> 1.3'
-# 'mercenary',    '~> 0.3.3'
-# 'safe_yaml',    '~> 1.0'
-# 'colorator',    '~> 0.1'
-# 'rouge',        '~> 1.7'
-# 'jekyll-sass-converter', '~> 1.0'
-# 'jekyll-watch', '~> 1.1'
-
-# These are additional Gems whose versions I fixed to the ones mentioned in Jekyll 3.1.6's Gemfile.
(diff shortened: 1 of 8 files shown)


gregestren:
Overall this looks amazing! I'm seriously impressed at how successfully you've re-stylized this.

Walkthrough comments:

**Linux desktop:**
 - The top links look squeezed together. Can we unsqueeze them some?
navbar
 - I'm torn on the phrasing "Create issue for this page". It's clear. But it's long, so breaks onto two lines easily, which is a bit jarring. Is "Create issue" or "Create doc issue" clear enough or is that too compromising?
* Is there any way to stop scrolling the right navbar when we get to the bottom of the page? 
* On the left navbar, what would a slightly thicker green bar look like?

  leftbar

  I'm curious because of https://reactjs.org/docs/release-channels.html. Does that make the focus any clearer?

* Higher min width for the left navbar? As-is we get overflow on narrow windows:
version

&nbsp;
**13" Macbook**: Looks great!

**iPhone 6:**: Looks great! Maybe consider making the right navbar inline (as a block element at the top of the page, like what we have now in production), since mobile viewers lose the horizontal space. But I wouldn't block this PR on that. And we might want to do some analytics review to verify how much time that's worth.

jin:
> On the left navbar, what would a slightly thicker green bar look like?


Screen Shot 2020-02-19 at 2 56 33 PM

gregestren:
> > On the left navbar, what would a slightly thicker green bar look like?
> 
> Screen Shot 2020-02-19 at 2 56 33 PM

Thanks for trying this.

I'm really not sure which looks better. I trust your judgment whichever choice you prefer - this is a minor nit in the grand scheme of things and I don't want to further block progress on it.

jin:
I actually prefer the thicker one :) 

Pushed a new update with your comments, PTAL. https://bazel-docs-staging.netlify.com/versions/master/bazel-overview.html

I'll hold off the mobile optimizations as a follow up.

PR #10764, opened 2020-02-12
Add targets for creating Bazel dmg file for macOS

meteorcloudy:
1. We add Bazel icon for the Bazel binary.
2. We use a preconfigured DS_Store file and a background image in the dmg file.


scripts/packages/dmg/create_bazel_dmg.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+# Copyright 2015 The Bazel Authors. All rights reserved.
```
philwo: 2020? ^^

PR #10780, opened 2020-02-13
Fix tyop

JayBazuzi:


aiuto:
I'll just do this fix myself since I am on triage this week. Thanks for pointing it out.

PR #10784, opened 2020-02-14
Delete Protobuf 3.6.1 from tree

Yannic:
These files are no longer needed since Protobuf was upgrade
to 3.11.3 in https://github.com/bazelbuild/bazel/pull/10651.


Yannic:
@philwo Thanks a ton for helping with merging all dependencies of this. It's time to profit from all that work :)).

I hereby declare this to be my favorite patch to Bazel so far.

philwo:
Merged in https://github.com/bazelbuild/bazel/commit/bb1258238f6559bafebc767c94a68348a82a9e81 and all looks good, no breakages observed since then - thank you so much for this work :D

Issue #10777, opened 2020-02-13
Release builds should be notarized for macOS Catalina


shs96c:
### Description of the problem / feature request:

In macOS Catalina, Apple introduced a requirement that applications that are downloaded and run on the machine are notarized. Without this, applications refuse to run unless a user follows a non-obvious sequence of steps involving the "Security" preference pane.

Bazel releases should be notarized in order to ensure that it works as expected on macOS. 

### What operating system are you running Bazel on?

macOS Catalina and above

###  Have you found anything relevant by searching the web?

Mozilla also ran into (and resolved!) this problem with both Firefox and `geckodriver`:

* https://firefox-source-docs.mozilla.org/testing/geckodriver/Notarization.html
* https://bugzilla.mozilla.org/show_bug.cgi?id=1588081

Steps for notarizing applications on the command line can be found here: https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow

There may already be machinery within Google that can be harnessed for notarizing releases, which may simplify this issue considerably.

davido:
Duplicate: #9304.

aiuto:
The title here says it better than #9304 
I gave this to both the Apple and EngProd teams since it may require collaboration to get the nuances.

tetromino:
Yep, this affects the latest MacOS Bazel installer if it's downloaded using the browser (as opposed to curl or wget). We should fix this, and until we fix this, we should document a workaround.

tetromino:
Correction: this affects only the Bazel _executable_ (e.g. `bazel-2.1.0-darwin-x86_64`).

The _installer script_ (which is what we suggest users to download in https://docs.bazel.build/versions/master/install-os-x.html) works fine when downloaded from the browser.

Downgrading priority to P2.

meteorcloudy:
I'm moving our release pipeline to Kokoro so that we can do the signing and notarization for macOS.

fweikert:
@meteorcloudy what's the current state of the Kokoro builds?

meteorcloudy:
They are set up, but we haven't switched to them because we still haven't figured out the notarization process..

jin:
Is this still a P1 bug?

meteorcloudy:
It doesn't look like so. Deprioritizing this to P2.

Issue #9304, opened 2019-09-01
Bazel on mac 10.15 fails with "“bazel-real” can’t be opened because its integrity cannot be verified."


paladox:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

When i upgraded to Mac 10.15, bazel stopped working. bazel build release fails with (“bazel-real” can’t be opened because its integrity cannot be verified.)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

* bazel build release (in the gerrit project)

### What operating system are you running Bazel on?

Mac OSX 10.15

### What's the output of `bazel info release`?

Does not work due to "“bazel-real” can’t be opened because its integrity cannot be verified."

(I'm using the packages from https://github.com/bazelbuild/bazel/releases)

I tried upgrading to bazel 0.29, but the installer fails with the same error.

paladox:
Work around is to do:

sudo spctl --master-disable

Which brings backs the option of "Anywhere" to Security & Privacy.

philwo:
@paladox I can reproduce this on a test machine with macOS 10.15.

It is because of the new notarization requirement. We will have to find a way how to correctly notarize our release binaries for macOS.

As a workaround, you can download the Bazel installer or binary not using a web-browser, but using "wget" or "curl", which doesn't mark the file as "having been downloaded from the internet", so macOS will not complain:

```
curl -LO https://github.com/bazelbuild/bazel/releases/download/0.29.0/bazel-0.29.0-installer-darwin-x86_64.sh && bash bazel-0.29.0-installer-darwin-x86_64.sh
```

You could also install Bazel or Bazelisk using Homebrew, which apparently also works fine.

paladox:
Ah thanks! I just went for the "sudo spctl --master-disable" option :)

philwo:
I think Apple relaxed the requirements here, because I can no longer reproduce this in the final version of macOS 10.15.

Reducing priority due to this, however, we'll of course still work on notarizing our official release binaries. :)

cmooreml:
Good article on how to enable/disable: https://www.imore.com/how-open-apps-unidentified-developers-mac

Splaktar:
[...]

PR #10787, opened 2020-02-14
Update install-os-x.md with Catalina download workaround instructions.

tetromino:
This is a workaround for the app notarization requirements in Catalina until
we have a new release workflow in place. Note that at least on MacOS 10.15.2,
the notarization warning is triggered while running the installer script if
the script was downloaded using the browser. See  #9304 and #10777.

RELNOTES: None.


aiuto:
Thanks!  That should give us time until we can do proper notarization.

Issue #8609, opened 2019-06-11
Add support for reading a list of target patterns for from a file


robbertvanginkel:
### Description of the problem / feature request:

Add support for reading a list of target patterns for from a file for `bazel {build,test}` (in addition to supporting target patterns in argv)

### Feature requests: what underlying problem are you trying to solve with this feature?

We have a command based on query that selects a set of targets (based on this [bazel-discuss](https://groups.google.com/d/msg/bazel-discuss/I9udqWIcEdI/iczVgWLOBQAJ) and https://github.com/bazelbuild/bazel/issues/7962) and passes these to bazel build. Roughly something like:
```
bazel query ... | grep | sed | awk > foo
bazel test $(<foo)
```

However, this becomes problematic if the output of query + modifications is long, specifically longer than the  `ARG_MAX` the system supports:
```
# write non-null arg file that is longer than getconf ARG_MAX
$ dd if=/dev/zero bs=1 count=$((1+$(getconf ARG_MAX))) | tr "\0" "\7" > foo
$ bazel build $(<foo)
-bash: /usr/local/bin/bazel: Argument list too long
```

I currently work around this by using xargs, as it splits the incoming stream into multiple bazel invocations if it is too big.
```
bazel query ... | grep | sed | awk | xargs bazel test --
```
I don't however want multiple bazel invocations as this means I need to support reading/merging multiple logs, build event files etc etc.

Ideally bazel would support reading a list of arguments from a file, maybe even generically like:
```
$ cat argfile.txt
--bazelrc=/dev/null
build
--
//some/...
...
//targets:here
$ bazel --argfile=argfile.txt
```
But at least I'd like to find a way to have bazel test and build read the list of target patterns from a file (or any other method of passing it a list of arbitrary length).

### What operating system are you running Bazel on?

MacOS / Linux

### What's the output of `bazel info release`?

release 0.26.1

###  Have you found anything relevant by searching the web?

The [bazel query documentation](https://docs.bazel.build/versions/master/query.html#set
) suggests using shell command substitution like:
```
  bazel query deps(//my:target) --output=label | grep ... | sed ... | awk ... > foo
  bazel query "kind(cc_binary, set($(<foo)))"
```

However, this fails once the query output is longer than `ARG_MAX` as described above.

https://github.com/bazelbuild/bazel/blob/aca672f2dae79451325fbff5e60d79bfcde0669f/site/docs/query.html#L490-L493

[...]

PR #10796, opened 2020-02-15
Add support for invoking bazel with args from file

robbertvanginkel:
Fixes https://github.com/bazelbuild/bazel/issues/8609. Also see https://groups.google.com/d/msg/bazel-discuss/895-OjU2Z2A/mYlw05rpDgAJ. Allows Bazel to be invoked with all arguments coming from a file rather than using commandline arguments. For example:

```
$ cat args.txt
build
//...
$ bazel-bin/src/bazel @args.txt
INFO: Invocation ID: 5a582c6a-4a3c-4c16-bff1-619aba804783
...
```

With the current implementation: only one argfile is allowed, it should be prefixed with an `@` and the first argument (argv[1]). Argfiles should be newline separated (`\n`), one argument per line.

TODO:
- [ ] Decide param file location (tempfile? somewhere in output_base?).
- [ ] Decide `command_server.proto` changes/semantics when using argfile.
- [ ] Investigate special quoting needs of param file for JVM on Windows.
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -261,7 +261,7 @@ class BlazeServer final {
   // is in connected state.
   unsigned int Communicate(
       const std::string &command, const std::vector<std::string> &command_args,
-      const std::string &invocation_policy,
+      bool uses_param_file, const std::string &invocation_policy,
       const std::vector<RcStartupFlag> &original_startup_options,
       const LoggingInfo &logging_info,
       const DurationMillis client_startup_duration,
@@ -1306,7 +1306,7 @@ static ATTRIBUTE_NORETURN void RunClientServerMode(
       CancelServer);
   SignalHandler::Get().PropagateSignalOrExit(server->Communicate(
       option_processor.GetCommand(), option_processor.GetCommandArguments(),
-      startup_options.invocation_policy,
+      option_processor.UsingParamFile(), startup_options.invocation_policy,
       startup_options.original_startup_options_, *logging_info,
       client_startup_duration, extract_data_duration,
       command_wait_duration_ms));
@@ -1964,7 +1964,7 @@ void BlazeServer::KillRunningServer() {
 }
 
 unsigned int BlazeServer::Communicate(
-    const string &command, const vector<string> &command_args,
+    const string &command, const vector<string> &command_args, bool uses_param_file,
     const string &invocation_policy,
     const vector<RcStartupFlag> &original_startup_options,
     const LoggingInfo &logging_info,
@@ -1987,8 +1987,16 @@ unsigned int BlazeServer::Communicate(
   request.set_cookie(request_cookie_);
   request.set_block_for_lock(block_for_lock_);
   request.set_client_description("pid=" + blaze::GetProcessIdAsString());
-  for (const string &arg : arg_vector) {
-    request.add_arg(arg);
+  if (uses_param_file) {
+    blaze_util::Path param_file = output_base_.GetRelative("params");
+    string content;
+    blaze_util::JoinStrings(arg_vector, '\0', &content);
+    blaze_util::WriteFile(content, param_file);
+    request.set_parameter_file(param_file.AsCommandLineArgument());
+  } else {
+    for (const string &arg : arg_vector) {
+      request.add_arg(arg);
+    }
   }
   if (!invocation_policy.empty()) {
     request.set_invocation_policy(invocation_policy);
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -715,6 +715,20 @@ static void RunBatchMode(
 
   jvm_args_vector.insert(jvm_args_vector.end(), command_arguments.begin(),
                          command_arguments.end());
+  if (option_processor.UsingParamFile()) {
+    blaze_util::Path param_file =
+      blaze_util::Path(startup_options.output_base).GetRelative("params");
+    string jvm_execve_exe = jvm_args_vector[0];
+
+    string content;
+    for(auto i = jvm_args_vector.begin()+1; i != jvm_args_vector.end(); ++i) {
+      content.length() == 0 ?
+        content = blaze_util::Quoted(*i) :
+        content += '\n' + blaze_util::Quoted(*i);
+    }
+    blaze_util::WriteFile(content, param_file);
+    jvm_args_vector = {jvm_execve_exe, "@" + param_file.AsCommandLineArgument()};
+  }
 
   GoToWorkspace(workspace_layout, workspace);
 
--- a/src/main/cpp/util/strings.cc
+++ b/src/main/cpp/util/strings.cc
@@ -30,6 +30,8 @@
 
 #include <cassert>
 #include <memory>  // unique_ptr
+#include <sstream>
+#include <iomanip>
 
 #include "src/main/cpp/util/exit_code.h"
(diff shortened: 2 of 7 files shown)


robbertvanginkel:
👋My first time contributing to the bazel client, C++ isn't my area of expertise. Although I realise the PR isn't in a completely mergeable state (missing tests/documentation), I wanted to open a PR earlier rather  than later to ask for feedback (so I don't have to rewrite tests/code etc for  code that might need a bit of an overhaul).

I'll leave some comments on the files I have open questions/doubts about, please give it a look!

src/main/cpp/blaze.cc
```diff
@@ -261,7 +261,7 @@ class BlazeServer final {
   // is in connected state.
   unsigned int Communicate(
       const std::string &command, const std::vector<std::string> &command_args,
-      const std::string &invocation_policy,
+      bool uses_param_file, const std::string &invocation_policy,
```
robbertvanginkel: Maybe refactor `Communicate` to take a ` const OptionProcessor *option_processor`, as the first three arguments are currently: 
```

server->Communicate(
  option_processor.GetCommand(), 
  option_processor.GetCommandArguments(), 
  option_processor.UsingParamFile()
```

src/main/cpp/blaze.cc
```diff
@@ -715,6 +715,20 @@ static void RunBatchMode(
 
   jvm_args_vector.insert(jvm_args_vector.end(), command_arguments.begin(),
                          command_arguments.end());
+  if (option_processor.UsingParamFile()) {
+    blaze_util::Path param_file =
+      blaze_util::Path(startup_options.output_base).GetRelative("params");
```
robbertvanginkel: I'm not sure if these param files are safe to be written to output_base, or if they need unique locations to support multiple clients using the same output_base.

src/main/cpp/blaze.cc
```diff
@@ -715,6 +715,20 @@ static void RunBatchMode(
...
+    string jvm_execve_exe = jvm_args_vector[0];
+
+    string content;
+    for(auto i = jvm_args_vector.begin()+1; i != jvm_args_vector.end(); ++i) {
+      content.length() == 0 ?
+        content = blaze_util::Quoted(*i) :
+        content += '\n' + blaze_util::Quoted(*i);
```
robbertvanginkel: The java argument file format takes all whitespace as separate arguments, so we need to appropriately quote each.
https://docs.oracle.com/javase/9/tools/java.htm#JSWOR-GUID-4856361B-8BFD-4964-AE84-121F5F6CF111

src/main/cpp/util/strings.cc
```diff
@@ -271,6 +273,13 @@ void StringPrintf(string *str, const char *format, ...) {
   delete[] buf;
 }
 
+std::string Quoted(const std::string s)
+{
+  std::ostringstream ss;
+  ss << std::quoted( s.c_str() );
+  return ss.str();
+}
+
```
robbertvanginkel: I'm not sure if there's a better way to achieve quoting that java expects the argfile to have.

src/main/protobuf/command_server.proto
```diff
@@ -58,6 +58,8 @@ message RunRequest {
   // came from. These options have already been parsed and already have had
   // their effect. This information should only be used for logging.
   repeated StartupOption startup_options = 6;
+
+  string parameter_file = 7;
```
robbertvanginkel: I think this should be a `oneof` with `repeated bytes arg = 2;`, that way it is clear for the java code which field to parse. Right now the java code attempts to parse field 2, if its empty it attempts to parse this.

I'm not sure where this proto file is used, and if this would be backward compatible change. Any tips on how to do this best welcome.

philwo:
Hi @robbertvanginkel,

thanks for your contribution! This seems like a pretty cool feature to have. I'll try to find you a reviewer for this!

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
```diff
@@ -595,6 +597,20 @@ private void executeCommand(RunRequest request, BlockingStreamObserver<RunRespon
...
+          try {
+            Scanner sc = new Scanner(file);
+            sc.useDelimiter("\u0000");
+            while (sc.hasNext()) {
+              builder.add(sc.next());
+            }
+            args = builder.build();
+          } catch (IOException e) {
+          }
+        }
```
laurentlb: Do you really need to modify the Java code?
I think it would be simpler to open the file and do the expansion in `option_processor.cc`.
(also skip the change to the proto file)

robbertvanginkel: Yes, after doing the expansion in the cpp client we still need to send it to the java server. As the server/client communicate over gRPC this still has a message size limit, see https://github.com/bazelbuild/bazel/issues/8609#issuecomment-571830880 for some more context.

The only solution I found for that is making sure the client writes the arguments to a file and instructs the server to read that file, for which we'd need a proto change (or implicitly encode this somehow in the existing field for args).

laurentlb: I'm told that the gRPC limit is entirely artificial and we can just increase it.

Call .maxInboundMessageSize(1<<30) on NettyServerBuilder here: 
https://github.com/bazelbuild/bazel/blob/636430a30f3eef86062640d66e049df37ad16711/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java#L439

pcj:
I don't think we should assume that the bazel client and bazel server are always running on the same machine (and sharing the same filesystem), so changing the `command_server.proto` is probably not a good idea.

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
michajlo: Let's keep a sensible limit on the max inbound message size.

robbertvanginkel: @michajlo what would you consider a sensible limit? 10/100s of megabytes? 

What I like about the file approach is that it doesn't impose another seemingly arbitrarily chosen limit. What I'd like to avoid is ending up with a choice that needs to be increased 6months or a year down the line. Maybe the limit should be configurable?

robbertvanginkel:
Thanks all for the comments! There's definitely some extra investigation to do here, I agree with @michajlo that it is probably a good idea to have a larger design discussion around it.

However, I think that might take some more time and I would like to have a reasonable resolution for #8609 soon. The alternative suggested (a `--build_file` option similar to the existing `--query_file`) seems like a pretty straightforward implementation (https://github.com/bazelbuild/bazel/pull/10856), although a bit less generic. Please give that idea a look, I hope that can be merged a bit sooner and be  a stop gap till I can spend the time working out the details for a generic bazel arguments file discussion.

jlisee:
With #10856 landed should this be closed?

robbertvanginkel:
Yes #10856 fixes my immediate issue for https://github.com/bazelbuild/bazel/issues/8609. I don't have the time right now to dive into general argfile support and writeup a proposal.

Closing out this PR to reflect I'm not actively working on it. If anyone ends up looking at argfile support in the future, I hope this points them in the right direction :).

PR #10797, opened 2020-02-15
Hashing cleanups: add SHA-384 and SHA-512 support. Stop using MessageDigest.

EdSchouten:
I am currently experimenting with adding custom hashing algorithms to Bazel and REv2 and assessing their performance. I will likely share the results of that work at some point in the future. Leading up to this, I made some cleanups to the existing code for handling/computing checksums. These cleanups made it a lot easier for me to add custom hashing algorithms. In short, this PR does the following:

- It updates the REv2 definitions used by Bazel to the latest version. This brings in support for SHA-384 and SHA-512.
- It adds SHA-384 and SHA-512 support to Bazel. I personally don't care about these hashing algorithms, but it's low-hanging fruit.
- It then removes all uses of plain Java MessageDigest from the codebase. Everything will now use Guava's HashFunction.

This allows me to add custom hashing algorithms by implementing my own HashFunction, as opposed to needing to implement both a HashFunction and MessageDigest.


src/main/java/com/google/devtools/build/lib/util/Fingerprint.java
```diff
@@ -14,43 +14,36 @@
...
   /** Creates and initializes a new instance. */
   public Fingerprint(DigestHashFunction digestFunction) {
-    messageDigest = digestFunction.cloneOrCreateMessageDigest();
-    // This is a lot of indirection, but CodedOutputStream does a reasonable job of converting
-    // strings to bytes without creating a whole bunch of garbage, which pays off.
-    codedOut =
-        CodedOutputStream.newInstance(
-            new DigestOutputStream(ByteStreams.nullOutputStream(), messageDigest),
-            /*bufferSize=*/ 1024);
+    hashFunction = digestFunction.getHashFunction();
```
ulfjack: The reason why this was reusing the MessageDigest is performance. Creating a new instance is rather expensive. If the guava functions use MessageDigest under the covers, then this could be a net loss.

Another observation: technically, Bazel doesn't *have* to use the same digest for its own action cache as for remote execution (I'm not 100% sure if Fingerprint is used anywhere in the remote execution client).

src/main/java/com/google/devtools/build/lib/vfs/DigestHashFunction.java
```diff
@@ -24,18 +24,12 @@
...
-import java.security.NoSuchAlgorithmException;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map.Entry;
 
 /**
  * Type of hash function to use for digesting files.
- *
- * <p>This tracks parallel {@link java.security.MessageDigest} and {@link HashFunction} interfaces
- * for each provided hash, as Bazel uses both - MessageDigest where performance is critical and
```
ulfjack: see here

third_party/remoteapis/build/bazel/remote/execution/v2/remote_execution.proto
```diff
@@ -323,6 +323,8 @@ service ContentAddressableStorage {
   // If part of the tree is missing from the CAS, the server will return the
   // portion present and omit the rest.
   //
+  // Errors:
```
ulfjack: Updating the re proto must go into a separate commit. Due to how Bazel source is imported / exported from Google, third_party is special. Sorry!

aiuto:
Is this going to go any where or should we drop it?

EdSchouten:
I will likely revisit this at some point in the future, but let's drop it for the time being.

To clarify: at some point I would like to explore the possibility of switching to BLAKE3, so that we can decompose large CAS objects.

steeve:
I'd be very curious to look at the performance improvements by switching to Blake3.

There is a JNI implementation at https://github.com/sken77/BLAKE3jni which might need testing.

PR #10798, opened 2020-02-15
Add s390x and make zip -d step more resilient

jonpspri:
Continuation of PR #10643 for application to third_party directory

Build on s390x fails because the process to remove unnecessary libraries from netty was not configured for x86. As a result, zip -d failed because it had nothing to remove. This PR does two things:

    Adds conditions for s390x to the src and third_party BUILD files, and
    Adds a condition to the genrule so that 'zip -d' does not execute if there's nothing to do.

See also Issue #9263; Pull Requests #9346 and #9945

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -469,11 +469,12 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
-    "//src/conditions:freebsd": "*.so *.jnilib *.dll",
-    "//src/conditions:openbsd": "*.so *.jnilib *.dll",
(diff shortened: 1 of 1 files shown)


third_party/BUILD
```diff
@@ -469,11 +469,12 @@ UNNECESSARY_DYNAMIC_LIBRARIES = select({
     "//src/conditions:arm": "*.so *.jnilib *.dll",
     "//src/conditions:linux_aarch64": "*.so *.jnilib *.dll",
     "//src/conditions:linux_ppc": "*.so *.jnilib *.dll",
-    "//src/conditions:freebsd": "*.so *.jnilib *.dll",
-    "//src/conditions:openbsd": "*.so *.jnilib *.dll",
```
philwo: Why are you removing these?

jonpspri: Um... because despite all my efforts not to I managed to overlook something when I ran the push?  I’ll un-remove them now...

philwo: No worries, thanks!

jonpspri: @philwo Patched up.

PR #10807, opened 2020-02-17
Windows: Lookup environment variables in exact casing

gjasny:
The Bazel respository_ctx provides all environment variables
in a case-sensitive dictionary.

On a Windows 10 system looking up "PROGRAMFILES" will not be
successful whereas "ProgramFiles" will be.

### DIFF
--- a/tools/cpp/windows_cc_configure.bzl
+++ b/tools/cpp/windows_cc_configure.bzl
@@ -25,12 +25,32 @@ load(
     "write_builtin_include_directory_paths",
 )
 
+def _lookup_env_var(env, name, default = None):
+    """Lookup environment variable case-insensitve.
+
+    If a matching (case-insesitive) entry is found in the env dict both
+    the key and the value are returned. The returned key might differ from
+    name in casing.
+
+    If a matching key was found its value is returned otherwise
+    the default is returned.
+
+    Return a (key, value) tuple"""
+    for key, value in env.items():
+        if name.lower() == key.lower():
+            return (key, value)
+    return (name, default)
+
+def _get_env_var(repository_ctx, name, default = None):
+    """Returns a value from an environment variable."""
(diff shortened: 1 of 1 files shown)


gjasny:
I'm not sure this is the proper fix but it seems to be a step into the right direction.

gjasny:
Hmm Buildkite reports environment variables all uppercase:
```
  | PROGRAMFILES=(C:\Program Files)
  | PROGRAMFILES(X86)=(C:\Program Files (x86))
```

Would change `_get_path_env_var` in `windows_cc_configure.bzl` to lookup the environment variable in a case-insensitive manner help?

meteorcloudy:
> Would change _get_path_env_var in windows_cc_configure.bzl to lookup the environment variable in a case-insensitive manner help?

@gjasny Sorry for the late reply. I think we should do this, to ignore the case while checking for environment variable on Windows.

laszlocsomor:
Windows envvar names are case-insensitive, so `GetEnvironmentVariableW` returns the same for `FOO` as for `Foo` I think _get_path_env_var should also be case-insensitive.

tools/cpp/windows_cc_configure.bzl
```diff
@@ -426,12 +426,12 @@ def find_llvm_path(repository_ctx):
 
     # 3. Check default directories for LLVM installation
     auto_configure_warning_maybe(repository_ctx, "Looking for default LLVM installation directory")
-    program_files_dir = _get_path_env_var(repository_ctx, "PROGRAMFILES")
+    program_files_dir = _get_path_env_var(repository_ctx, "ProgramFiles")
```
meteorcloudy: @gjasny Can you instead make this function(`_get_path_env_var`) case-insensitive?

meteorcloudy:
@gjasny I've merged the other PR, can you resolve the conflict and address the reviewer comments?

gjasny:
Working on it ...

PR #10814, opened 2020-02-18
Create package .../devtools/build/lib/{unix,unix/jni}

Yannic:


aiuto:
Is there an issue to motivate this? If not, could you at least add the rationale in the PR description.

philwo:
Splitting packages up like this is a welcome refactoring :) I'll try to merge it!

philwo:
Unfortunately this is a bit more involved than just importing the PR, due to the way we transform our internal BUILD files to Bazel's BUILD files. I have to do some more urgent work right now, but will get back to this in the next days.

ulfjack:
+10 from me, very nice :-D

Yannic:
Motivation for this is mostly to reduce the size of `lib/BUILD` to make it easier to navigate. I can't really see a reason why, e.g., `.../lib:events` isn't moved to `.../lib/events`.

`UnixJniLoader` was moved to `....lib.unix.jni` to match the layout of `...lib.windows`.

kkress:
I'm working on another large refactor of BUILD files in lib.  I'll take ownership of this internally and get it merged.

src/main/java/com/google/devtools/build/lib/unix/jni/BUILD
```diff
@@ -0,0 +1,23 @@
...
+    name = "srcs",
+    srcs = glob(["**"]),
+    visibility = ["//src/main/java/com/google/devtools/build/lib/unix:__pkg__"],
+)
+
+# Java interfaces to various native UNIX system services.
+#
+# All programs that depend on libraries in this package will have
+# JNI (native code) dependencies.
+java_library(
```
alandonovan: Nothing is gained by splitting lib.unix.jni from lib.unix. Neither is useful without the other, and both are primarily concerned with "native functions for POSIX platforms", so they are not in any useful sense separate packages.

But in the interests of workflow efficiency, let's commit this change, which is mostly progress, and deal with the rest in a follow-up.

Ditto lib.windows.

kkress: I've got a stack of changes pending for cleanups, I'll add this to the list.

Issue #10794, opened 2020-02-14
Bazel docs - do lightweight content formatting fixes


gregestren:
For example, make https://docs.bazel.build/versions/2.1.0/bazel-vision.html less wall-o'-texty.

The broader goal is to make docs inviting, accessible, not intimidating. The less encyclopedic (i.e. dense) each page is the better.

https://github.com/bazelbuild/bazel/issues/10791 should also help by using bigger fonts and enforcing better maximum column widths.

PR #10817, opened 2020-02-19
Standardize install pages

gregestren:
- Standardized steps to "Part X: " form
- Shortened some redundant titles
- Made header sizes consistent

Tested with styling changes from https://github.com/bazelbuild/bazel/pull/10762. In particular:
formatted changes for the new right-side page navbar.

Serves https://github.com/bazelbuild/bazel/issues/10794 (but
since these are versioned these will release on a different
timeline than styling changes)

### DIFF
--- a/site/docs/install-compile-source.md
+++ b/site/docs/install-compile-source.md
@@ -38,7 +38,7 @@ TL;DR:
 
 Detailed instructions follow below.
 
-<h3 id="build-bazel-install-bazel">1. Get the latest Bazel release</h3>
+<h3 id="build-bazel-install-bazel">Step 1: Get the latest Bazel release</h3>
 
 **Goal**: Install or download a release version of Bazel. Make sure you can run
 it by typing `bazel` in a terminal.
@@ -67,7 +67,7 @@ scratch (bootstrap)](#bootstrap-bazel).)
 
         You must make the binary executable by running `chmod +x /path/to/bazel`.
 
-<h3 id="build-bazel-git">2. Download Bazel's sources from GitHub</h3>
+<h3 id="build-bazel-git">Step 2: Download Bazel's sources from GitHub</h3>
 
 If you are familiar with Git, then just git clone https://github.com/bazelbuild/bazel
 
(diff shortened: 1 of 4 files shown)


site/docs/install-compile-source.md
```diff
@@ -154,7 +155,7 @@ your PATH so that you can run "bazel" everywhere.)
 
 You can also build Bazel from scratch, without using an existing Bazel binary.
```
aiuto: Bootstrap should move to a separate page. 1 out of 1000 bazel users should ever to that.
Even building from source should be a separate page, since that is a non-preferred path and most people won't do that.

site/docs/install-compile-source.md
```diff
@@ -140,7 +141,7 @@ Unix-like systems.)
 4.  The output will be at `bazel-bin\src\bazel-dev.exe` (or
     `bazel-bin\src\bazel.exe`).
 
-<h3 id="build-bazel-install">5. Install the built binary</h3>
+<h3 id="build-bazel-install">Step 5: Install the built binary</h3>
 
 Actually, there's nothing to install.
```
aiuto: I disagree with this.  You can't keep using the bazel in bazel-bin. You need to 

  You should 
- copy the built binary to some place in your path, making sure it will be selected instead of the previously downloaded bazel
- to bazel test  ???? on the bazel source tree. Make sure all tests pass
Continue to use that new bazel.

site/docs/install-windows.md
```diff
@@ -35,11 +37,11 @@ Alternatively you can:
...
 *   [Download Bazelisk](https://github.com/bazelbuild/bazelisk) instead of Bazel. Bazelisk is a
     Bazel launcher that ensures you always use the latest Bazel release.
-*   [Install Bazel from Chocolatey](#install-using-chocolatey) (see below)
-*   [Install Bazel from Scoop](#install-using-scoop) (see below)
+*   [Install Bazel from Chocolatey](#using-chocolatey)
+*   [Install Bazel from Scoop](#using-scoop)
 *   [Build Bazel from source](install-compile-source.html)
 
-## 4. Optional: configure output directories
+### Step 4: Optionally configure output directories
```
aiuto: I think optional: install   or   (optional) install    is  better than optionally install.
You are designating an optional step, not saying that you should install it in an optional fashon.

gregestren: Done.

site/docs/install-compile-source.md
gregestren: Points taken. I'd like to restrict this PR to formatting, and save deeper content refactoring / splitting up pages for the next phase's effort.

PR #10834, opened 2020-02-20
Fix spelling of 'led' in error message

j-towns:
The past tense of the verb 'lead' is spelled 'led' (see e.g. https://www.grammarly.com/blog/led-lead/).


aiuto:
I'm not sure led is any better.   The whole error message is a bit wonky.
It is a cycle, but then why is it a "missing" definition. 
Lead or led is not an optimal term either. Maybe just
"The chain below shows the cycle"

j-towns:
>I'm not sure led is any better

It's at least spelled correctly. I don't think I'm qualified to improve the semantics of the error message, just wanted to fix an obvious typo.

aiuto:
Sorry if I wasn't clear. I meant I can't even tell what tense was intended. So I'm sure if we want 
"This trace led to the problem" or "This trace will lead to the problem". It should all go, so it's not worth changing now to something that doesn't help.

PR #10856, opened 2020-02-25
Implement --target_pattern_file flag for reading patterns from file

robbertvanginkel:
An alternative to https://github.com/bazelbuild/bazel/pull/10796 to fix https://github.com/bazelbuild/bazel/issues/8609.

As https://github.com/bazelbuild/bazel/issues/8609#issuecomment-590522448 points out there's currently a `--query_file` to read a query pattern from file, but there's no analogous feature available for build/test. Generic parameter file support attempted in https://github.com/bazelbuild/bazel/pull/10796 turns out to be a little more complex than expected due to bazel's client/server architecture and likely requires some design document. 

This PR would fix help https://github.com/bazelbuild/bazel/issues/8609 without rushing into any designs, so we can spend time properly designing generic parameter files while also alleviating the immediate problem with a  known/simpler pattern.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildtool/BuildRequestOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/buildtool/BuildRequestOptions.java
@@ -455,6 +455,17 @@ public class BuildRequestOptions extends OptionsBase {
       help = "If this flag is set, replay action out/err on incremental builds.")
   public boolean replayActionOutErr;
 
+  @Option(
+      name = "target_pattern_file",
+      defaultValue = "",
+      documentationCategory = OptionDocumentationCategory.GENERIC_INPUTS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS},
+      help =
+          "If set, build will read patterns from the file named here, rather than on the command "
+              + "line. It is an error to specify a file here as well as command-line patterns."
+  )
+  public String targetPatternFile;
+
   /**
    * Converter for jobs: Takes keyword ({@value #FLAG_SYNTAX}). Values must be between 1 and
    * MAX_JOBS.
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
@@ -67,8 +67,8 @@ public final class BuildCommand implements BlazeCommand {
     BlazeRuntime runtime = env.getRuntime();
     List<String> targets;
     try {
-      targets = TargetPatternFileSupport.handleTargetPatternFile(env, options);
-    } catch (TargetPatternFileSupport.TargetPatternFileSupportException e) {
+      targets = TargetPatternsHelper.readFrom(env, options);
+    } catch (TargetPatternsHelper.TargetPatternsHelperException e) {
       env.getReporter()
           .handle(Event.error(e.getMessage()));
       return BlazeCommandResult.exitCode(ExitCode.COMMAND_LINE_ERROR);
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
@@ -33,6 +33,7 @@ import com.google.devtools.build.lib.runtime.CommandEnvironment;
 import com.google.devtools.build.lib.runtime.KeepGoingOption;
 import com.google.devtools.build.lib.runtime.LoadingPhaseThreadsOption;
 import com.google.devtools.build.lib.util.DetailedExitCode;
+import com.google.devtools.build.lib.util.ExitCode;
 import com.google.devtools.common.options.OptionsParsingResult;
 import java.util.List;
 
@@ -65,9 +66,12 @@ public final class BuildCommand implements BlazeCommand {
   public BlazeCommandResult exec(CommandEnvironment env, OptionsParsingResult options) {
     BlazeRuntime runtime = env.getRuntime();
     List<String> targets;
-    try (SilentCloseable closeable = Profiler.instance().profile("ProjectFileSupport.getTargets")) {
-      // only takes {@code options} to get options.getResidue()
-      targets = ProjectFileSupport.getTargets(runtime.getProjectFileProvider(), options);
+    try {
+      targets = TargetPatternFileSupport.handleTargetPatternFile(env, options);
+    } catch (TargetPatternFileSupport.TargetPatternFileSupportException e) {
+      env.getReporter()
+          .handle(Event.error(e.getMessage()));
+      return BlazeCommandResult.exitCode(ExitCode.COMMAND_LINE_ERROR);
     }
     if (targets.isEmpty()) {
       env.getReporter()
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
@@ -1,0 +1,63 @@
+package com.google.devtools.build.lib.runtime.commands;
+
+import com.google.common.collect.Lists;
+import com.google.devtools.build.lib.buildtool.BuildRequestOptions;
+import com.google.devtools.build.lib.profiler.Profiler;
+import com.google.devtools.build.lib.profiler.SilentCloseable;
+import com.google.devtools.build.lib.runtime.BlazeCommand;
+import com.google.devtools.build.lib.runtime.CommandEnvironment;
+import com.google.devtools.build.lib.vfs.FileSystemUtils;
+import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.common.options.OptionsParsingResult;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+/**
+ * Provides support for implementations for {@link BlazeCommand} to read target patterns from file.
+ */
+final class TargetPatternFileSupport {
+
+  private TargetPatternFileSupport() {}
+
+  /**
+   * Reads a list of target patterns, either from the command-line residue or by reading newline
+   * delimited target patterns from the --target_pattern_file flag. If --target_pattern_file is
+   * specified and options contain a residue, or file cannot be read it throws an exception instead.
+   */
+  public static List<String> handleTargetPatternFile(
+      CommandEnvironment env, OptionsParsingResult options)
+      throws TargetPatternFileSupportException {
+    List<String> targets = options.getResidue();
+    BuildRequestOptions buildRequestOptions = options.getOptions(BuildRequestOptions.class);
+    if (!targets.isEmpty() && !buildRequestOptions.targetPatternFile.isEmpty()) {
+      throw new TargetPatternFileSupportException(
+          "Command-line target pattern and --target_pattern_file cannot both be specified");
+    } else if (!buildRequestOptions.targetPatternFile.isEmpty()) {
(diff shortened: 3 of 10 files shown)


robbertvanginkel:
@michajlo thanks for your comments in https://github.com/bazelbuild/bazel/pull/10796#pullrequestreview-363636974 and your suggestion in https://github.com/bazelbuild/bazel/issues/8609#issuecomment-590522448. While would like generic parameter file support in Bazel, I would also appreciate a more immediate solution to #8609. Suggesting we go with this much simpler approach for now till I can spend the time on a generic solution to the open questions in https://github.com/bazelbuild/bazel/pull/10796.

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
```diff
@@ -94,4 +100,31 @@ public BlazeCommandResult exec(CommandEnvironment env, OptionsParsingResult opti
     ExitCode exitCode = new BuildTool(env).processRequest(request, null).getExitCondition();
     return BlazeCommandResult.exitCode(exitCode);
   }
+
+  public static Pair<ExitCode, List<String>> getTargets(CommandEnvironment env, OptionsParsingResult options) {
```
robbertvanginkel: Pairs with `null` are probably not the most elegant way of reusing this code between `BuildCommand`/`TestCommand`, suggestions to improve welcome :)

src/main/java/com/google/devtools/build/lib/buildtool/BuildRequestOptions.java
```diff
@@ -455,6 +455,17 @@ public boolean useTopLevelTargetsForSymlinks() {
       help = "If this flag is set, replay action out/err on incremental builds.")
   public boolean replayActionOutErr;
 
+  @Option(
+      name = "build_file",
```
michajlo: let's call this different - "build file" is already claimed... maybe "target_file" or "targets_file" or "target_pattern_file"?

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
```diff
@@ -64,12 +69,29 @@
...
+      }
+      try (SilentCloseable closeable = Profiler.instance().profile("ProjectFileSupport.getTargets")) {
+        // only takes {@code options} to get options.getResidue()
+        targets = ProjectFileSupport.getTargets(runtime.getProjectFileProvider(), options);
+      }
+    } else if (!buildRequestOptions.buildFile.isEmpty()) {
+      // Works for absolute or relative query file.
+      Path residuePath = env.getWorkingDirectory().getRelative(buildRequestOptions.buildFile);
+      try {
+        targets = Lists.newArrayList(FileSystemUtils.readLines(residuePath, StandardCharsets.UTF_8));
```
michajlo: We should probably check if not targets were specified in the files. If we're not doing this in --query_file it was probably an oversight...

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
```diff
@@ -69,29 +70,12 @@
...
-        targets = Lists.newArrayList(FileSystemUtils.readLines(residuePath, StandardCharsets.UTF_8));
-      } catch (IOException e) {
-        env.getReporter()
-            .handle(Event.error("I/O error reading from " + residuePath.getPathString()));
-        return BlazeCommandResult.exitCode(ExitCode.COMMAND_LINE_ERROR);
-      }
-    } else {
+    Pair<ExitCode, List<String>> maybeTargets = getTargets(env, options);
+    if (maybeTargets.getFirst() != null) {
+      BlazeCommandResult.exitCode(maybeTargets.getFirst());
```
michajlo: did you mean to return this?

src/test/shell/integration/build_file_test.sh
```diff
@@ -0,0 +1,102 @@
...
+function test_build_file_test() {
+  setup
+  echo //:y > test.params
+  bazel test --build_file=test.params >& $TEST_log || fail "Expected success"
+  expect_log "1 test passes"
+}
+
+function test_build_file_no_targetpattern() {
+  setup
+  bazel build --build_file=build.params -- //:x >& $TEST_log && fail "expected failure" || true
```
michajlo: ... && fail "Expected failure"

We want to make sure this command didn't succeed.

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
michajlo: Might be cleaner to move to a new class and use a custom exception to signal that whatever was invalid?

src/main/java/com/google/devtools/build/lib/buildtool/BuildRequestOptions.java
robbertvanginkel: Went with `target_pattern_file`.

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
robbertvanginkel: The build command explicitly allows calling build without a resisdue/target pattern.

michajlo: I think this was mostly resolved by later commits - IIRC my concern was the target file could be empty and we'd be skip the "Nothing will be built" warning.

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.devtools.build.lib.profiler.SilentCloseable;
+import com.google.devtools.build.lib.runtime.CommandEnvironment;
+import com.google.devtools.build.lib.vfs.FileSystemUtils;
+import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.common.options.OptionsParsingResult;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+public final class TargetPatternFileSupport {
```
michajlo: make package private (remove "public")? Also add a quick javadoc to this and the visible methods and exception.

jin:
Reviewers: once approved, please read the internal go/bazel-import-pr link for PR import instructions.

robbertvanginkel:
Processed the comments but did not get to the unit test yet. I looked at making one, but standing up a `CommandEnvironment` to pass in is a little tricky (its constructor is not public and in a different package), and when I attempted to make a mockito mock I got an error like 
```
Cannot mock/spy class com.google.devtools.build.lib.runtime.CommandEnvironment
Mockito cannot mock/spy because :
 - final class
```
I'm not sure how best to approach that, if you have any recommendation please let me know and I'll try to finish the unit test.

michajlo:
CommandEnvironment should be mockable as of an hour or so ago.

michajlo:
Let me know when that last test is ready, otherwise everything else lgtm.

robbertvanginkel:
@michajlo the unit test should be there: https://github.com/bazelbuild/bazel/blob/20964788553921e0335d5c34cd4bf70d18c4b7dd/src/test/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupportTest.java

robbertvanginkel:
@michajlo I squashed the changes and rebased them on master to resolve a merge conflict github was showing. It seems like there was a few changes to use `DetailedExitCode` instead of `ExitCode`, I attempted to update the references.

I think all the changes, including unit and integration tests should be there. Please let me know if theres anything else!

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return targets;
+  }
+
+  /**
+   * TargetPatternFileSupportException gets thrown when TargetPatternFileSupport cannot return a
+   * list of targets based on the supplied command line options.
+   */
+  public static class TargetPatternFileSupportException extends Exception {
+    public TargetPatternFileSupportException(String message) { super(message); }
```
michajlo: style thing, mv the super(...) to a new line:

    ... {
      super(message);
    }

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
```diff
@@ -0,0 +1,63 @@
...
+final class TargetPatternFileSupport {
+
+  private TargetPatternFileSupport() {}
+
+  /**
+   * Reads a list of target patterns, either from the command-line residue or by reading newline
+   * delimited target patterns from the --target_pattern_file flag. If --target_pattern_file is
+   * specified and options contain a residue, or file cannot be read it throws an exception instead.
+   */
+  public static List<String> handleTargetPatternFile(
```
michajlo: nit: this isn't always getting the target patterns from a file, in fact it's almost always going to come from the commandline. Can we rename this to better reflect that? I think renaming the class/method to something like "TargetPatternsHelper#readFrom" avoids committing to too much.

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
```diff
@@ -0,0 +1,63 @@
...
+      try (SilentCloseable closeable =
+          Profiler.instance().profile("ProjectFileSupport.getTargets")) {
+        targets = ProjectFileSupport.getTargets(env.getRuntime().getProjectFileProvider(), options);
+      }
+    }
+    return targets;
+  }
+
+  /**
+   * TargetPatternFileSupportException gets thrown when TargetPatternFileSupport cannot return a
```
michajlo: I think you can just simplify this to "Thrown when target patterns were incorrectly specified" since it can happen independent of any file being read.

src/test/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupportTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Test
+  public void testSpecifyNonExistingFileThrows() throws OptionsParsingException {
+    options.parse("--target_pattern_file=patterns.txt");
+
+    assertThrows(
+        TargetPatternFileSupport.TargetPatternFileSupportException.class,
+        () -> TargetPatternFileSupport.handleTargetPatternFile(env, options));
+  }
+}
```
michajlo: add a test that specifying neither is ok?

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
robbertvanginkel: Fixed. Does Bazel have something like `google-java-format` setup? I couldn't find it myself in the contribution guidelines.

robbertvanginkel:
All comments should be taken care of.

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternFileSupport.java
```diff
@@ -0,0 +1,63 @@
+package com.google.devtools.build.lib.runtime.commands;
```
michajlo: this file needs a copyright header

robbertvanginkel:
Thanks @michajlo!

isaprykin:
--query_file accepts space separated list of targets.
It needs to be clarified that this only accepts newline separated list of targets.
I costed me a few hours.

PR #10858, opened 2020-02-25
Update docs for --combined_report flag

keith:
This doesn't accept HTML right now and AFAICT the codepaths aren't setup
for this.


ulfjack:
LGTM

ulfjack:
@meisterT can you import?

PR #10866, opened 2020-02-26
test_exec_properties to control test action

### LINKED ISSUE #10799: test_exec_properties differentiation

> Provide a test_exec_properties option for *_test rules.

> The test execution platform properties should be differentiated from exec_properties. Currently setting an exec_properties of a *_test will apply to the actions used to build the target, as well as the test itself. In order to apply specific runtime configurations to the platform of the test, but not, say, the compiles of a cc_test, a different rule attribute is required.

werkt:
If specified for a test action, these properties will be added to the
effective exec_properties for a target, overriding any keys which match
by name, and only being effective for the test's execution, not any
other actions from the rule's context.

Closes #10799

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
@@ -230,6 +230,8 @@ public class BaseRuleClasses {
                           env.getToolsLabel(DEFAULT_COVERAGE_REPORT_GENERATOR_VALUE))))
           // The target itself and run_under both run on the same machine.
           .add(attr(":run_under", LABEL).value(RUN_UNDER).skipPrereqValidatorCheck())
+          // Exec properties used only for the test execution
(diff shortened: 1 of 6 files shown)


sdtwigg:
I am unsure about this change.

Regardless, I think you also need to add the attribute to SkylarkRuleClassFunctions.java in order for this to also work in Starlark test rules.

werkt:
The associated issue should provide some additional explanation for this change. If there's already a way to target *only* test actions with per target property manipulations, I'm fine with that, but I don't believe there is.

Updated to support starlark test rules.

gregestren:
Hey @werkt. In the interest of moving this PR forward, I think the right next step is to get an update from @juliexxia, who's implemented a major chunk of the work discussed in https://github.com/bazelbuild/bazel/issues/10799. 

Let's aim to get her perspective next week, then decide the right way to close this from there.

juliexxia:
It's being replaced by an exec group to create test actions. Waiting on internal PR reviews.

philwo:
@juliexxia I understand that we can thus close this PR. Please comment and reopen if this is wrong.

PR #10868, opened 2020-02-26
Allow "pool" as an expected NinjaToken after rule name in a declaration

jin:
Repro:

```ninja
rule some_name
    pool = some_pool
    command = echo hello
```

```starlark
ninja_graph(
   name = "simple",
   main = "simple.ninja",
   ninja_srcs = [
       "simple.ninja",
   ],
   output_root = "out",
   output_root_inputs = [
   ],
)
```

```
ERROR: in ninja_graph rule //:simple: Expected [IDENTIFIER], but got pool in fragment:
rule some_name
    pool = some_pool
    command = echo hello

ERROR: Analysis of target '//:simple' failed; build aborted: Analysis of target '//:simple' failed; build aborted
FAILED: Build did NOT complete successfully (6 packages loaded, 11 targets configured)
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/parser/NinjaParserStep.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/parser/NinjaParserStep.java
@@ -180,6 +180,7 @@ public class NinjaParserStep {
     variablesBuilder.put(NinjaRuleVariable.NAME, NinjaVariableValue.createPlainText(name));
 
     parseExpected(NinjaToken.NEWLINE);
+    lexer.interpretPoolAsVariable();
     while (parseIndentOrFinishDeclaration()) {
       String variableName = asString(parseExpected(NinjaToken.IDENTIFIER));
       parseExpected(NinjaToken.EQUALS);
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/parser/NinjaParserStep.java
```diff
@@ -181,7 +182,7 @@ public NinjaRule parseNinjaRule() throws GenericParsingException {
 
     parseExpected(NinjaToken.NEWLINE);
     while (parseIndentOrFinishDeclaration()) {
-      String variableName = asString(parseExpected(NinjaToken.IDENTIFIER));
+      String variableName = asString(parseExpected(ImmutableSet.of(NinjaToken.IDENTIFIER, NinjaToken.POOL)));
```
c-parsons: Why is "pool" treated differently from an identifier?

When going through the existing code to better understand this lexer, I happened upon NinjaLexer.interpretPoolAsVariable()

So it looks like there's already some toggleable logic for treating "pool" as an identifier, no?

src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/parser/NinjaParserStep.java
```diff
@@ -446,26 +457,26 @@ private void throwNotExpectedTokenError(NinjaToken expectedToken, NinjaToken tok
     throw new GenericParsingException(
         String.format(
             "Expected %s, but got %s in fragment:\n%s\n",
-            asString(expectedToken.getBytes()),
+            expectedTokens,
```
c-parsons: It looks like you need to update tests :)

src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/parser/NinjaParserStep.java
jin: NinjaToken.POOL was created together with NinjaToken.IDENTIFIER. Maybe @irengrig had plans for specializing logic with the `pool` token? As far as I can tell, `NinjaToken.POOL` is not used specially anywhere else other than NinjaParserStep.

irengrig:
There is a keyword "pool", which is recognized by the lexer, when the *pool declaration* is parsed.
There is a flag managing where pool is a keyword or is it a variable name: https://github.com/bazelbuild/bazel/blob/4b351c5d8176455d6ada9b5852ac872e202d8de4/src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/lexer/NinjaLexer.java#L49

why is this became important now? what problem is solved here?

irengrig:
(found out why you are fixing that)
please use the flag in NinjaLexer I pointed?

jin:
Hang on, so should `pool` in `pool = some_pool` be parsed as `NinjaToken.POOL` or `IDENTIFIER`? The problem here is that the `parseExpected` call expects to see an `IDENTIFIER`, but the lexer (correctly?) categorizes it as `POOL`, and throws.

In my debugging, `interpretPoolAsVariable` is false in the NinjaParserStep when it encounters `pool`.

irengrig:
I probably only was thinking about the case when
build xxx: yyy
  pool = some-pool-defined

but not
rule xxx: command $in $out
  pool = some-pool-defined

- in both such cases pool is an identifier

here it is a pool lexem:

pool some-pool-defined
   depth = 5

documentation: https://ninja-build.org/manual.html#ref_pool

PR #10877, opened 2020-02-29
Adding links to eleven new languages supported by Bazel

kigster:
Perhaps one of the biggest selling points of Bazel to any org — is the long list of languages it supports, ensuring that even those using more exotic languages can benefit from Bazel. 

The language list, however, keeps on growing, and yet this page shows a rather outdated list. This is why I thought that it is time for the official documentation to mention the rules that are hosted on https://github.com/bazelbuild organization or are at the very least in the late stages of development. This is how I arrived at the current list of an additional eleven languages.

This overhaul would not be complete if it didn't mention the packaging rules and the docker rules, which are most often used in the deployment. 

## TL;DR 

By looking through this list developers unfamiliar with Bazel should start to feel confident that they should be able to migrate their entire process (build, test, deploy) to Bazel, and that is why (IMHO) it is important to emphasize the breadth of Bazel Rules on this page.

### DIFF
--- a/site/docs/rules.md
+++ b/site/docs/rules.md
@@ -35,9 +35,33 @@ Here is a selection of recommended rules:
 * [Shell](be/shell.html)
 * [Webtesting](https://github.com/bazelbuild/rules_webtesting) (Webdriver)
 
+The following rules have been recently accepted into the official [BazelBuild](https://github.com/bazelbuild) Github Org:
+
+* [Closure](https://github.com/bazelbuild/rules_closure)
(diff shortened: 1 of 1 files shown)


laurentlb:
Please see the process for recommended rules: https://www.bazel.build/recommended-rules.html

Whether a rule is in bazelbuild organization or not doesn't make a big difference. We can recommend a ruleset only when we believe it's well maintained and we can offer some guarantees to the users. I've checked the rules in your list a few months ago, and decided (after chatting with their maintainers) that they were not ready.

For a more exhaustive list of rules, see https://awesomebazel.com/

PR #10894, opened 2020-03-04
Pass default flags before custom flags to allow overriding

AlessandroPatti:
Target-specific copts should override "default" ones coming from the objc fragment.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -305,11 +305,11 @@ public class CompilationSupport {
             .addCcCompilationContexts(objcCompilationContext.getDepCcCompilationContexts())
             .setCopts(
                 ImmutableList.<String>builder()
-                    .addAll(getCompileRuleCopts())
                     .addAll(
                         ruleContext
                             .getFragment(ObjcConfiguration.class)
                             .getCoptsForCompilationMode())
+                    .addAll(getCompileRuleCopts())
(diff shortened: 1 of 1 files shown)


oquenchil:
I'm not sure what the extent of breakages would be, but isn't this an incompatible change?

AlessandroPatti:
> I'm not sure what the extent of breakages would be, but isn't this an incompatible change?

I don't think an incompatible change flag would be necessary, the set of flags will remain the same.
Having the flags in the `copts` attributes always overridden looks to me more of an unexpected behaviour than a thing that users would exploit.

oquenchil:
The new policy in Bazel since 1.0 is that if it's an incompatible change, an incompatible change flag is necessary. It could even be that this doesn't break anyone at all but the fact is that we don't know, so unfortunately the policy says that a flag is necessary.

This overriding of flags can definitely break an existing project that was relying the old order.

I'm adding Dmitry in case there have been any changes in the policy that I'm not aware of.

philwo:
@oquenchil If you classify this as a bug, i.e. it was always intended that the override works, but it just didn't, then we don't need a flag. It would still be good to call out in the relnotes then.

oquenchil:
Aha that's a nice loophole. Good to know. Yes, this is definitely a bug.

dslomov:
If there is a widespread breakage caused by this change, I think we might still need a flag. Can we asses that with [downstream testing](https://github.com/bazelbuild/continuous-integration/blob/master/docs/downstream-testing.md)?

oquenchil:
This broke internal tests. Can you please put it behind a flag and we can merge again?

Thanks.

PR #10906, opened 2020-03-05
remote: fail if repository_ctx.execute() has a non-zero exit status

buchgr:
Remote execution systems generally don't cache actions with a non-zero
exit status. A uncachable command significantly increases the runtime
of a repository_rule with remote execution enabled. This change makes
it an error if repository_ctx.execute() returns a non-zero exit status.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
@@ -33,6 +33,7 @@ import com.google.devtools.build.lib.bazel.repository.downloader.Checksum;
 import com.google.devtools.build.lib.bazel.repository.downloader.DownloadManager;
 import com.google.devtools.build.lib.bazel.repository.downloader.HttpUtils;
 import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.ExtendedEventHandler.FetchProgress;
 import com.google.devtools.build.lib.packages.Attribute;
 import com.google.devtools.build.lib.packages.Rule;
@@ -444,6 +445,14 @@ public class SkylarkRepositoryContext
         OutErr outErr = OutErr.SYSTEM_OUT_ERR;
         outErr.printOut(stdout);
         outErr.printErr(stderr);
+      }
+
+      if (result.exitCode() != 0) {
+        String warningMsg = String.format("@%s: '%s' exited with status %d. A remote execution"
+                + " system only caches commands with a zero exit status."
+            + " A non-zero exit status increases the runtime of the repository rule.",
+            rule.getName(), String.join(" ", arguments), result.exitCode());
+        env.getListener().handle(Event.warn(warningMsg));
       }
 
       return new SkylarkExecutionResult(result.exitCode(), stdout, stderr);
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
@@ -448,11 +448,10 @@ public class SkylarkRepositoryContext
       }
 
       if (result.exitCode() != 0) {
-        String warningMsg = String.format("@%s: '%s' exited with status %d. A remote execution"
-                + " system only caches commands with a zero exit status."
-            + " A non-zero exit status increases the runtime of the repository rule.",
-            rule.getName(), String.join(" ", arguments), result.exitCode());
-        env.getListener().handle(Event.warn(warningMsg));
+        String errorTemplate = "Command '%s' exited with status %d. Commands with non-zero exit status are not"
+                + " support with remote execution as they cannot be cached. Update the command to always exit"
+                + " with status 0.";
+        throw Starlark.errorf(errorTemplate, String.join(" ", arguments), result.exitCode());
       }
 
       return new SkylarkExecutionResult(result.exitCode(), stdout, stderr);
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -1781,6 +1781,39 @@ EOF
   expect_log "exceeded deadline"
 }
 
+function test_repo_remote_exec_non_zero_exit_status() {
+    # Test that Bazel fails with an error if repository_ctx.execute returns a non-zero exit code.
+
+  touch BUILD
+
+  cat > test.bzl <<'EOF'
+def _impl(ctx):
+  ctx.execute(["/bin/bash", "-c", "exit 1"])
+  ctx.file("BUILD")
+
+foo_configure = repository_rule(
+  implementation = _impl,
+  remotable = True,
+)
+EOF
+
+  cat > WORKSPACE <<'EOF'
+load("//:test.bzl", "foo_configure")
+
+foo_configure(
+  name = "default_foo",
+)
+EOF
+
+  bazel fetch \
+    --remote_executor=grpc://localhost:${worker_port} \
+    --experimental_repo_remote_exec \
+    @default_foo//:all >& $TEST_log && fail "Should fail" || true
+
+  cat $TEST_log
+}
+
+
 # TODO(alpha): Add a test that fails remote execution when remote worker
 # supports sandbox.
 


buchgr:
+@brandjon since you mentioned that printing warnings in workspace rules can confuse users.

brandjon:
@alandonovan 

I have issues with adding *any* warning to console output -- the vast majority of people who see it will not be able to act on it or care. If it's important enough to warn, it's important enough to fail with an error. Conversely, if you want a way to alert maintainers of repo rules (or performance-conscious users), we should provide a non-default flag to opt into these warnings.

Alan also points out that it's not clear what action we want the repo rule maintainer to take. Wrap their tool invocation in a shell script that returns 0? Hassle the remote execution devs to start caching non-zero exits? Sit in the corner and feel ashamed?

buchgr:
That's a good point. I will change this to be an error and also update the message to ask the user to wrap the command to return exit code 0.

alandonovan:
That sounds reasonable if it's feasible, though I assume it's a breaking change. Does it mean that the (misnamed) `exec_result.return_code` becomes a constant zero?

buchgr:
@alandonovan 

The remote execution functionality is guarded by `--experimental_repo_remote_exec`. I'd like to get more experience with the API before proposing to make it stable.

Correct, `exec_result.return_code` would be a constant zero if remote execution is enabled. That is
```
repository_rule(
  ...
  remotable = True,
)
```

P.S.: [1] is the original change that added remote execution to repository_ctx.execute. We also discussed executing the implementation function on a remote machine. We decided against it due to the simplicity of [1]. I'd be interested if you have any thoughts on the matter. 


[1] https://github.com/bazelbuild/bazel/commit/06a26c37126e3eefabd28c24a4e35059b5b136bc

alandonovan:
Ah, so the restriction would apply only to remote execution (and thus the field value is not truly constant).

Contributors who don't use remote execution would be liable to introduce execution steps that cause the remote build to fail. Is that likely to be a problem in practice?

Why do users need the ability to run commands that fail? (The only example of this in the build proper is the old C++ non-compilation tests, which were a total pain.)

buchgr:
> Contributors who don't use remote execution would be liable to introduce execution steps that cause the remote build to fail. Is that likely to be a problem in practice?

I'll update the change to fail if `remotable=True and exec_result.return_code != 0`. So it fail for rules marked compatible with remote execution even if remote execution is not used. That should make testing and development easier. Sounds good?

> Why do users need the ability to run commands that fail? (The only example of this in the build proper is the old C++ non-compilation tests, which were a total pain.)

Excellent question. I was surprised when I first encountered this in the wild. One example in Tensorflow is getting include paths for the nvcc compiler. Apparently, the best way to get include paths from nvcc is to invoke it as `nvcc -v /dev/null -o /dev/null`. It will fail with a non-zero exit code and print the include paths to stderr. The repository rule then parses stderr.

alandonovan:
Can the (presumably) small handful of cases that expect a negative result be rephrased to: `sh -c '! command'` or `sh -c 'command || true`, then we can make execute unconditionally require success?

brandjon:
I think that users shouldn't have to think about whether the repo rule is marked as remotely executable, just to understand whether or not it's valid to have a non-zero return code.

Sure, if they're hardcoding paths to system tools that don't exist on the remote machine, they're gonna have a bad time. But it seems a little unfair that they could shell out to a perfectly reasonable script, but then switching to remote execution requires restructuring that script instead of just cleaning up their path assumptions.

Do we permit rule actions (e.g. `ctx.actions.run()`) to return non-zero exit codes?

An alternative might be to add an argument to `repository_ctx.execute()`, maybe `cacheable` (default `True`), that must be set to False for non-zero exit codes to be permitted. The error message would have to tell the user what to do, e.g. "repository_ctx.execute() returned non-zero status (set cacheable=False if this is intentional and not an error)".

alandonovan:
> Do we permit rule actions (e.g. ctx.actions.run()) to return non-zero exit codes?

No, non-zero exit is a build failure.

> An alternative might be to add an argument to repository_ctx.execute(), maybe cacheable (default True), that must be set to False for non-zero exit codes to be permitted. 

It's trivial for users to negate or ignore the success code in the rare cases it is necessary, by interposing a shell expression as I showed above. Why should we make the build tool more complex to handle this case?

We shouldn't be asking users to think about the concept of "Cacheable".  We should ask them to use deterministic commands. Caching is an implementation detail of optimizing deterministic builds.

buchgr:
> Can the (presumably) small handful of cases that expect a negative result be rephrased to: sh -c '! command' or sh -c 'command || true', then we can make execute unconditionally require success?

I think so. This is how I fixed Tensorflow's repo rules. In Tensorflow two callsides needed fixing: the nvcc compiler as mentioned above and grep which returns 1 if the pattern can't be found.

I like the idea of expecting all commands to exit with 0. It's simple and the few odd commands are easy enough to fix if necessary.

jsharpe:
I'm currently making use of repository_ctx.execute to detect the type of a particular tool and then create controlflow execution in starlark dependent upon the type of the tool I've detected; I then go on to setup other variables based upon the flavour of the tool I've detected. If I can no longer check return_code in this case then it would mean that I'd end up removing the flow logic from starlark and moving it into a bash or python script.

I guess the question here is whether 'autodetection' of tools for creating toolchain configurations should be written in starlark or in an external script. Which one is easier to make hermetic? I don't think I'm qualified enough to answer that question but my instinct would say that its better to have this logic in starlark rather than an arbitrary choice of scripting language?

buchgr:
Thanks @jsharpe. I'd be great if you could be more specific as to what detecting the type of a particular tool entails.

jin:
Quick stats: there [more than 2000 usages](https://github.com/search?q=%22%5C.return_code%22+filename%3Abzl&type=Code) of `return_code` on GitHub.

jsharpe:
@buchgr - specifically its mpicc. I read the complier wrapper from the environment variable and then attempt to workout which flavour of mpicc I have - I then invoke it with specific arguments to interrogate  and extract the relevant compiler flags. Its basically a port of the FindMPI script from cmake: https://github.com/Kitware/CMake/blob/master/Modules/FindMPI.cmake.

alandonovan:
@jsharpe It is easy to prove that making Bazel assert that return_code==0 does not reduce its expressiveness; users can always adapt to the restriction by interposing a shell expression that writes the command exit status to stdout and then exits zero.

Obviously this could be inconvenient if there are many places where this needs to happen, but from a brief search, nearly all places that compare return_code with 0 immediately return an error in one of the two cases.

jsharpe:
> @jsharpe It is easy to prove that making Bazel assert that return_code==0 does not reduce its expressiveness; users can always adapt to the restriction by interposing a shell expression that writes the command exit status to stdout and then exits zero.
> 
Indeed, I'm not doubting this fact but it does increase the barrier to entry to the new user by not following the python subprocess idioms for a language that to a new user looks a lot like python.

Also creating a shell expression that works on windows as well as linux is probably not that straightforward or does sh just work on windows too? If it doesn't the canonical way to do this would need to be documented well.

> Obviously this could be inconvenient if there are many places where this needs to happen, but from a brief search, nearly all places that compare return_code with 0 immediately return an error in one of the two cases.

Actually looking at my code it would also be broken just by the remote execution as it doesn't look like repository_ctx.which executes in the remote environment as well? I have a list of possible names for the mpi compiler wrapper which I test in turn for existance using repository_ctx.which() and then I try the various different arguments that the mpi compiler wrappers implement to work out which flavour I am. @buchgr did you consider that the command passed to execute may have come from the output of which?

My starlark code ends up looking like:

```
cmd = [mpicc, "--showme:compile"]
res = respository_ctx.execute(cmd)
if res.return_code == 0:
     <>
     return

cmd = [mpicc, "-compile-info"]
res = respository_ctx.execute(cmd)
if res.return_code == 0:
     <>
     return

... <repeated for 4 or 5 cases (with some with different environments set)>

```
I feel this ends up being sufficiently complicated that if I was having to wrap the detection logic in scripts that I might as well just end up shelling out to cmake to determine the flags rather than do it natively.

buchgr:
> Also creating a shell expression that works on windows as well as linux is probably not that straightforward or does sh just work on windows too?

On Windows one would need to use cmd.exe or powershell.exe. They support similar expressions (`cmd.exe /c "..." /b 0`).

> Actually looking at my code it would also be broken just by the remote execution as it doesn't look like repository_ctx.which executes in the remote environment as well?

Correct. It takes some work to make the rule compatible with remote execution. I think longer term we should deprecate `repository_ctx.which` as well and instead provide equivalent functionality in a Starlark library. Here is Tensorflow's implementation of `which` that works with remote execution https://github.com/tensorflow/tensorflow/blob/c3dbc73edcc9966825ba5fc80748268843eb0e3a/third_party/remote_config/common.bzl#L14

> @buchgr did you consider that the command passed to execute may have come from the output of which?

If remote execution is enabled then the expression `repository_ctx.execute(repository_ctx.which("mpicc"), "--showme:compile")` will fail as execute will reject `path` objects (as returned by which).

> I feel this ends up being sufficiently complicated that if I was having to wrap the detection logic in scripts that I might as well just end up shelling out to cmake to determine the flags rather than do it natively.

I think Bazel should provide an `execute` via a Starlark library that wraps a command in the shell logic, has an implementation of `which` and so on. Implementing this in Starlark is easy enough after all. Would that be ok with you?

meisterT:
What's the status of this PR?

src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
```diff
@@ -446,6 +447,13 @@ private SkylarkExecutionResult executeRemote(
         outErr.printErr(stderr);
       }
 
+      if (result.exitCode() != 0) {
+        String errorTemplate = "Command '%s' exited with status %d. Commands with non-zero exit status are not"
+                + " support with remote execution as they cannot be cached. Update the command to always exit"
```
jsharpe, line 452:
```suggestion
                + " supported with remote execution as they cannot be cached. Update the command to always exit"
```

jsharpe:
@buchgr realised I never responded to your question. Yes I'm fine with the idea of shipping a starlark function. It might even make more sense in skylib rather than bazel core?

v2046:
image

buchgr:
Let's close it. I don't have the cycles for this right now.

Issue #10742, opened 2020-02-10
Release 3.0 - March 2020


dslomov:
Cut date 2020-03-02

dslomov:
We continue our policy of major release every 3 month. We expect a very limited set of breaking changes in release 3.0

robbertvanginkel:
I hope the Bazel team can aim to review @jmillikin-stripe's work on the remote repository cache (https://github.com/bazelbuild/bazel/pull/10818, https://github.com/bazelbuild/bazel/pull/10622) on time for it to be included in 3.0.

Just yesterday we ran into https://github.com/GoogleContainerTools/distroless/issues/474, where accidentally an external dependency was removed from a gcp bucket leading to a few build failures for us. Although its not impossible to guard against that by using a resolved workspace file or http proxy, today it is quite hard to setup Bazel to use internal mirrors for all downloadable dependencies (especially the ones transitively pulled in are tricky to change). @jmillikin-stripe's work should make this significantly easier and has been in progress for a while, just want to throw another voice out there that it would solve a pain point and I would love to start using it.

katre:
I will attempt to cut a release on 2020/03/04, in the morning EST.

philwo:
@robbertvanginkel Looks like #10818 is merged, but #10622 is still in the middle of the review. Unfortunately the team members who were most familiar with our remote execution client have changed teams and are no longer available for Bazel code reviews. I hope that we can pick this up in the next months, but in the meantime, merging PRs in that area will take more time than usual, especially of that size. 😔

@ulfjack Would you be interested and available to "own" this part of the code? I'd be happy to merge PRs that you approve!

jmillikin-stripe:
I am extremely motivated to get that change in. A delay of "months" would be overwhelmingly bad. Can I do a pair-programming review session or something over Hangouts to go over it?

adunham-stripe:
Hi folks,

My name is Andrew, and I'm the PTL for Security here at Stripe. This feature is very important for us; we're counting on it to be able to ensure the integrity and provenance of dependencies in our builds.  The fact that it has taken almost 45 days to get a review on an accepted proposal, which was itself submitted nearly a year ago, and especially with someone that has been so willing to take feedback and iterate quickly, is starting to erode my confidence in Bazel's ability to accept contributions from outside Google. Can you please prioritize reviewing and merging @jmillikin-stripe's work?

Best,
--Andrew D

katre:
[...]

PR #10914, opened 2020-03-06
Implementation (but not plumbing) of the gRPC remote downloader

jmillikin-stripe:
Extracted from https://github.com/bazelbuild/bazel/pull/10622

Per discussion on that PR, there's still some unanswered questions about how exactly we plumb the new `Downloader` type into `RemoteModule`. And per https://github.com/bazelbuild/bazel/issues/10742#issuecomment-595633454, it is unlikely that even heroic effort from me will get the full end-to-end functionality into v3.0.

Given this, to simplify the review, I'm taking some of the bits the reviewer is happy with and moving them to a separate PR. After merger, `GrpcRemoteDownloader` and its tests will exist in the source tree, but will not yet be available as CLI options.

R: @michajlo
CC: @adunham-stripe @dslomov @EricBurnett  @philwo @sstriker

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
@@ -1,0 +1,92 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.bazel.repository.downloader;
+
+import com.google.common.hash.HashCode;
+import com.google.common.hash.Hasher;
+import com.google.devtools.build.lib.concurrent.ThreadSafety.ThreadCompatible;
+import java.io.IOException;
+import java.io.OutputStream;
+import javax.annotation.Nullable;
+import javax.annotation.WillCloseWhenClosed;
+
+/**
+ * Output stream that guarantees its contents matches a hash code.
+ *
+ * <p>The actual checksum is computed gradually as the output is written. If it
+ * doesn't match, then an {@link IOException} will be thrown when {@link #close()}
+ * is called. This error will be thrown multiple times if these methods are called
+ * again for some reason.
+ *
+ * <p>Note that as the checksum can only be computed once the stream is closed,
+ * data will be written to the underlying stream regardless of whether it matches
+ * the expected checksum.
+ *
+ * <p>This class is not thread safe, but it is safe to message pass this object
+ * between threads.
+ */
+@ThreadCompatible
+public final class HashOutputStream extends OutputStream {
+
+  private final OutputStream delegate;
+  private final Hasher hasher;
+  private final HashCode code;
+  @Nullable private volatile HashCode actual;
+
+  public HashOutputStream(@WillCloseWhenClosed OutputStream delegate, Checksum checksum) {
+    this.delegate = delegate;
+    this.hasher = checksum.getKeyType().newHasher();
+    this.code = checksum.getHashCode();
+  }
+
+  @Override
+  public void write(int buffer) throws IOException {
+    hasher.putByte((byte) buffer);
+    delegate.write(buffer);
+  }
+
+  @Override
+  public void write(byte[] buffer) throws IOException {
+    hasher.putBytes(buffer);
+    delegate.write(buffer);
+  }
+
+  @Override
+  public void write(byte[] buffer, int offset, int length) throws IOException {
+    hasher.putBytes(buffer, offset, length);
+    delegate.write(buffer, offset, length);
+  }
+
+  @Override
+  public void flush() throws IOException {
+    delegate.flush();
+  }
+
+  @Override
+  public void close() throws IOException {
+    delegate.close();
+    check();
+  }
+
+  private void check() throws IOException {
+    if (actual == null) {
+      actual = hasher.hash();
+    }
+    if (!code.equals(actual)) {
+      throw new UnrecoverableHttpException(
+          String.format("Checksum was %s but wanted %s", actual, code));
+    }
+  }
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>The actual checksum is computed gradually as the output is written. If it
+ * doesn't match, then an {@link IOException} will be thrown when {@link #close()}
+ * is called. This error will be thrown multiple times if these methods are called
+ * again for some reason.
+ *
+ * <p>Note that as the checksum can only be computed once the stream is closed,
+ * data will be written to the underlying stream regardless of whether it matches
+ * the expected checksum.
+ *
+ * <p>This class is not thread safe, but it is safe to message pass this object
```
katre: Is there a meaning to "message pass", or is this just mis-pasted?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
```diff
@@ -0,0 +1,92 @@
...
+ * between threads.
+ */
+@ThreadCompatible
+public final class HashOutputStream extends OutputStream {
+
+  private final OutputStream delegate;
+  private final Hasher hasher;
+  private final HashCode code;
```
katre: It looks like "code" is the passed-in code this checks against, and "actual" is the computed checksum to compare with? Add some comments to clarify this (and to explain why this needs to be "volatile", because I'm not sure why).

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,194 @@
...
+    FetchBlobRequest.Builder requestBuilder =
+        FetchBlobRequest.newBuilder()
+            .setInstanceName(instanceName);
+    for (URL url : urls) {
+      requestBuilder.addUris(url.toString());
+    }
+    if (checksum.isPresent()) {
+      requestBuilder.addQualifiers(
+        Qualifier.newBuilder()
+            .setName("checksum.sri")
```
katre: Should "checksum.sri" be a constant somewhere? Do we have remote-exec specific constants anywhere?

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -0,0 +1,194 @@
...
+    this.options = options;
+  }
+
+  @Override
+  public void close() {
+    if (closed.getAndSet(true)) {
+      return;
+    }
+    cacheClient.close();
+    channel.release();
```
katre: This is handling the release, but where is the corresponding retain call?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
jmillikin-stripe: It's copied from [`HashInputStream.java`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashInputStream.java). I don't know what "message pass" means in the context of Java.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
jmillikin-stripe: These are also copied from `HashInputStream.java`. I have no context on why these names were chosen, or what the purpose of `volatile` is for the `Hash*Stream` classes.

To clarify: I'm happy to _change_ these to be something else, but I don't have the ability to explain why existing Bazel code is the way it is.

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
jmillikin-stripe: Changed to a constant. I can't find any common shared constants file, so I just put it at the top of this one.

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
jmillikin-stripe: It gets retained by the caller of `new GrpcRemoteDownloader`. See the implementation and lifecycle of [`GrpcCacheClient.java`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java) for the code I copied from.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
katre: Ah, I didn't realize the source. Thanks for clarifying.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HashOutputStream.java
katre: Fair point. Let's leave as is, then.

katre:
Turns out I also need to sync open source https://source.bazel.build/bazel/+/master:third_party/remoteapis/ with the internal version. This may slow things down, my apologies.

tjgq:
@jmillikin-stripe With apologies for the thread necromancy: I am trying to understand the rationale for including the authorization headers in the qualifiers for the `FetchBlobRequest` in this PR [1].

My understanding of the spec [2][3] is that qualifiers serve to disambiguate resources residing at the same URI. However, it seems to me that authorization should only control access to a resource and would (should?) not affect its contents.

Do you recall why we decided to do this? Would you object to an incompatible Bazel change to drop authorization headers from the qualifier?

/cc @coeuvre @Yannic

[1] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java;l=199;drc=1ccc0a378a65ad6d7a9b6f1117871fdeda5c26e8
[2] https://cs.opensource.google/bazel/bazel/+/master:third_party/remoteapis/build/bazel/remote/asset/v1/remote_asset.proto;l=282;drc=29ac010f3754c308de2ff13d3480b870dc7cb7f6
[3] https://cs.opensource.google/bazel/bazel/+/master:third_party/remoteapis/build/bazel/remote/asset/v1/remote_asset.proto;l=67;drc=29ac010f3754c308de2ff13d3480b870dc7cb7f6

Issue #10867, opened 2020-02-26
Make %USERPROFILE%\_netrc the default location for the netrc file on Windows


plule-ansys:
### Description of the problem / feature request:

Bazel has a default "netrc" file location under Linux and Mac OS, but not under Windows. This is identified as "TODO" in [http.bzl](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L55)

### Feature requests: what underlying problem are you trying to solve with this feature?

Not having a default location for the netrc file on Windows makes it hard to use, especially as http_archive requires an absolute path ( #10689 ). The only supported way is to use a hard-coded path such as "C:/netrc".

### Proposed Solution

Use "%USERPROFILE%/_netrc" as a default location. While GNU does not define any location for Windows, this location is de facto the one being used. This location is used by:
 - [curl/libcurl](https://ec.haxx.se/usingcurl/usingcurl-netrc)
 - By extension any software using libcurl, including [git credential manager](https://confluence.atlassian.com/bitbucketserver/permanently-authenticating-with-git-repositories-776639846.html)
 - [Go langage](https://github.com/search?q=org%3Agolang+%22_netrc%22&type=Code)
 - [Heroku Ruby Library](https://github.com/heroku/netrc/blob/master/lib/netrc.rb#L31)

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

release 2.1.0

PR #10915, opened 2020-03-06
Set default location for .netrc file for Windows

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel/issues/10867
Also enabled skylark_repository_test on Windows


meteorcloudy:
Tests failed with `ed: command not found`, looks like we should install `ed` in MSYS2 on our Windows machiens.

PR #10921, opened 2020-03-07
Create package .../devtools/build/lib/bugreport

Yannic:
Updates #10930


lberki:
Why is this useful? Did I miss the relevant conversation?

Our internal source tree doesn't have these packages, either so it would be kinda painful to import this. Not that we shouldn't do it if it's an improvement, but I'd like to know what I'm signing up for.

philwo:
@lberki I think this is useful, because it is cleaning up our BUILD file structure. I find it laudable that people are tackling our technical debt. 😊

@Yannic There's just one thing - we neglected doing this for a very long time, but we recently have decided to finally clean this up in a way that synchronizes our internal copy of the BUILD files and Bazel's BUILD files and let's us generate one from the other automatically using [Copybara](https://github.com/google/copybara). This will allow us to use Copybara to do PR imports, too, which will make importing PRs that touch BUILD files much easier and quicker in the future. @kkress has been doing a lot of work here and he has many more changes in flight. Would you be OK with pausing this BUILD file refactoring work for a while, until Kevin is done with his changes? Otherwise you two are duplicating work, which would be a shame.

If you want to help in this area, one thing that would be greatly beneficial even now is if you could try to untangle circular dependencies between packages on the Java side in the ["build/lib" subfolders](https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/devtools/build/lib).

kkress:
> If you want to help in this area, one thing that would be greatly beneficial even now is if you could try to untangle circular dependencies between packages on the Java side in the ["build/lib" subfolders](https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/devtools/build/lib).

 Specifically there are a bunch a cycles between packages in buid/lib:build-base and I thin a few in :runtime too

laurentlb:
@kkress Maybe create a GitHub issue where other people can follow your progress and discuss the next steps?

ulfjack:
Another benefit may be performance. I've been using Bazel to test our
remote execution system and it looks like the build-base library is a
bottleneck. It takes very long to build and there are quite a few things
depending on it.

brandjon:
@alandonovan has expressed interest in making our lib/ subpackage structure more rationale. Anyone aiming to break directory-level circular dependencies should check with him to avoid working toward opposite directions.

jin:
+1 to @laurentlb's idea, please start a Github issue on Bazel package reorganization to avoid conflicting work in the same area.

alandonovan:
Thanks, Yannick, this change is a step in the right direction. Kevin is in the middle of related simplifications to BUILD files, so he should decide when to merge this change.

kkress:
I created [10930](https://github.com/bazelbuild/bazel/issues/10930) to track the work.

I have a pending CL internally that refactors events, bugfiling and util*, but its gonna take me a few days to clean it up and get it out.  It should end up refactoring things identically to this PR

kkress:
Just wanted to drop a comment here indicating I haven't forgotten this.  There have been a steady stream of BUILD file changes coming through.  I'm currently cleaning up src/test/java/... when that is done I'll be pushing through a change very close to this one giving events, bugreport and util their own packages

Yannic:
@kkress @katre I see the `events` part of this change has landed. Should I rebase for the `bugreport` part?

katre:
Yes, please do. Sorry for stepping on your toes like that. I'll be happy to help review the change once you've rebased.

src/main/java/com/google/devtools/build/lib/bugreport/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
+load("@rules_java//java:defs.bzl", "java_library")
```
katre, line 1: Can you rename this file to just BUILD to match the rest of the project?

Yannic: Of course!

Yannic:
Thank you!

katre:
Thanks. This will probably require non-trivial internal merges, but I'll start work today.

katre:
Turned out to be less of a problem than expected. Thanks for the help with this!

PR #10922, opened 2020-03-08
Add toolchain support for x86_windows

gjasny:
This patch adds toolchain support to compile targets for x86 32bit Windows using Visual Studio.
I split the change into two commits for easier reviewability. The first commit copies the x64_windows toolchain build template and the second commit then adjusts it for x86_windows.

The change has been tested with our internal mono repo using Visual Studio 2017 and 2019 as well as with abseil-cpp.

### DIFF
--- a/tools/cpp/BUILD.windows.tpl
+++ b/tools/cpp/BUILD.windows.tpl
@@ -216,6 +216,71 @@ toolchain(
 )
 
 cc_toolchain(
+    name = "cc-compiler-x64_windows",
+    toolchain_identifier = "msvc_x64",
+    toolchain_config = ":msvc_x64",
+    all_files = ":empty",
+    ar_files = ":empty",
+    as_files = ":msvc_compiler_files",
+    compiler_files = ":msvc_compiler_files",
+    dwp_files = ":empty",
+    linker_files = ":empty",
+    objcopy_files = ":empty",
+    strip_files = ":empty",
+    supports_param_files = 1,
+)
+
+cc_toolchain_config(
+    name = "msvc_x64",
+    cpu = "x64_windows",
+    compiler = "msvc-cl",
+    host_system_name = "local",
+    target_system_name = "local",
+    target_libc = "msvcrt",
+    abi_version = "local",
+    abi_libc_version = "local",
+    toolchain_identifier = "msvc_x64",
+    msvc_env_tmp = "%{msvc_env_tmp}",
+    msvc_env_path = "%{msvc_env_path}",
+    msvc_env_include = "%{msvc_env_include}",
+    msvc_env_lib = "%{msvc_env_lib}",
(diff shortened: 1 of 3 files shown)


gjasny:
@meteorcloudy Is Windows 32bit support something that you'd like to support (or tolerate) in Bazel?

meteorcloudy:
@gjasny The change itself looks good, thanks! But my concern is that we don't have any tests for x86 windows and it increases the maintenance burden for Bazel team because the auto detected cc toolchains are already very complicated in Bazel. How do you think we create a third party repository for this? You can use helper functions in `@bazel_tools/tools/cpp:windows_cc_configure.bzl`, see this [commit message](https://github.com/bazelbuild/bazel/commit/a3dd7775d2dbf7e4368712626e11c15be7a23efd). If you can also write some tests that would be even better. I can add this repo to Bazel CI and have tests against both Bazel@HEAD and Bazel@Release.

tools/cpp/windows_cc_configure.bzl
```diff
@@ -551,40 +570,62 @@ def _get_msvc_vars(repository_ctx, paths):
         lib_path = find_llvm_tool(repository_ctx, llvm_path, "llvm-lib.exe")
         if not lib_path:
             lib_path = find_msvc_tool(repository_ctx, vc_path, "lib.exe")
```
aiuto: Doesn't this need an arch arg?

tools/cpp/BUILD.windows.tpl
```diff
@@ -215,6 +216,71 @@ toolchain(
...
+        "cpp": "%{msvc_x86_cl_path}",
+        "gcc": "%{msvc_x86_cl_path}",
+        "gcov": "wrapper/bin/msvc_nop.bat",
+        "ld": "%{msvc_x86_link_path}",
+        "nm": "wrapper/bin/msvc_nop.bat",
+        "objcopy": "wrapper/bin/msvc_nop.bat",
+        "objdump": "wrapper/bin/msvc_nop.bat",
+        "strip": "wrapper/bin/msvc_nop.bat",
+    },
+    default_link_flags = ["/MACHINE:X86"],
```
aiuto: I don't see this getting set in the 64 bit chain.

aiuto:
To echo meteorcloudy's concern, this should really be in a different repository.  We are never going to build bazel for 32bit, so it will be undertested in this repo. rules_cc might be a better choice.

I am also concerned with the style that adds a second version of every executable. That gets fragile quickly.  It would be more maintainable if msvc_vars was a dict indexed by arch, returning another dict of cl_path, ml_path, link_path.

aiuto:
cc: @katre

gjasny:
We stopped the Bazel migration. I might re-visit the topic once I've got time for it, again. Maybe 32bit Windows builds will be gone by then.

Issue #9337, opened 2019-09-05
ProtoInfo created descriptor sets should include the proto file comments


thomasvl:
`ProtoInfo.*_descriptor_sets` appear to be built without the source comments from the proto file. Well done proto generators also relay the comments into the generated source so IDEs that do source indexing and pick them up to provide contextual tooltips/documentation/etc.

While there are way to a developer to invoke bazel with extra args to work around this (`--protocopt=--include_source_info`) we don't really want to encourage developers to have to use lots of custom flags for their builds, and with the use of remote action caches the use/lack of this flag means cache impacts.

So since the descriptor sets are in the provider for generation rules, it seems like they should provide the sets with comments.  This also means the source generated during a build would match what a developer would get if they just invoked protoc by hand with the .proto files available since generated from parsed proto files always includes comments.

Yannic:
Adding `--include_source_info` to all protoc invocations seems to not break anyone that doesn't test that the generated sources match checked-in sources (e.g. [here](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1231#e8b68262-878b-42c4-9c88-13b26f621196)).

I think it's save to add this behind an `--incompatible_` flag and flip for Bazel 2.0.
@lberki WDYT?

Bazel+downstream: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1231

lberki:
AFAIU it's a backward-compatible change since `--include_source_info` adds information to `FileDescriptorProto` so it could be done without an incompatible change.

@hlopko , WDYT? I'd ordinarly just say go ahead, but I burnt myself once, so I'd like to have another pair of eyes.

hlopko:
This also sounds like backwards compatible change to me. Let's see what @haberman thinks, but for me I'd say go ahead.

Josh, should we worry about enabling this by default for all users? Are there performance/binary size/build speed implications that we should be mindful about?

haberman:
I don't know if I have enough info to make that call on the fly. We should gather a bit of data about what the size/speed implications of this are. Perhaps we can get this data for a sizable google3 build?

hlopko:
I'll try to do that. If you don't want to be blocked by me, I'd be fine with guarding this behavior with a build_setting, default it to off, and let interested users to enable it by hand?

Yannic:
There already is an easy way to enable this (passing `--protocopt=--include_source_info`), so I don't think adding a build_setting for it provides many benefits.

[...]

PR #10925, opened 2020-03-08
Create experimental flag to include source info in proto descriptors

Yannic:
`protoc` has an option to include source info (e.g. comments) in the
generated descriptor sets. Well done proto generators also relay the
comments into the generated source so IDEs that do source indexing
and pick them up to provide contextual tooltips/documentation/etc.

This change adds an experimental flag to include this info in the
descriptors generated by `proto_library` so we can collect some
data whether including source info by default is feasible.

See #9337 for more context.

//cc @thomasvl

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
@@ -14,12 +14,12 @@
 
 package com.google.devtools.build.lib.rules.proto;
 
-import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.Iterables.isEmpty;
 import static com.google.devtools.build.lib.collect.nestedset.Order.STABLE_ORDER;
 import static com.google.devtools.build.lib.rules.proto.ProtoCommon.areDepsStrict;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


lberki:
After some more discussion with @oquenchil and @meisterT , we are kind of afraid that we'd never be able to flip this flag within Google.

There is a way around that, which is to make descriptor set generation use a regular `proto_lang_toolchain` through a dependency instead of the ad hoc one created here:

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java#L359

and then we can use a different one within Google and with Bazel. 

However, that's a lot of ceremony. So how about this course of action:

1. We get this change submitted
2. We run benchmarks within Google to figure out if flipping the flag is feasible or not
3. If yes, we flip the flag internally, then in Bazel
4. If not, we got with the above plan with `proto_lang_toolchain` rules

WDYT, @Yannic , @oquenchil , @meisterT ?

meisterT:
Sounds like a simple plan, so I like it :-)

Yannic:
The purpose of the flag is to see if including source info by default is feasible or not, so your plan SGTM.

alexeagle:
@meisterT FYI I'm getting a request to add support for this experimental flag (https://github.com/aspect-build/rules_lint/pull/314) which makes me wish for a tracking issue for removing the `experimental_` prefix, following @lberki proposed plan above.

PR #10929, opened 2020-03-09
Add missing NS_BLOCK_ASSERTIONS for more architectures

keith:
This flag was passed for some architectures in the conditional above
this one, but not for all

RELNOTES: NS_BLOCK_ASSERTIONS is now passed for all Apple architectures

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5324,6 +5324,7 @@ def _impl(ctx):
                                 "-O2",
                                 "-D_FORTIFY_SOURCE=1",
                                 "-DNDEBUG",
+                                "-DNS_BLOCK_ASSERTIONS=1",
                             ],
                         ),
                     ],


jmmv:
From the internal review, I got this comment:

> Not sure how this file is released, but this may cause some code to break. So you may want to figure out how to communicate / announce that?

Any thoughts?

keith:
I would consider this a bug in the first case tho which is how it was discovered since it broke the other way where only in certain configurations it wouldn't do what people expected.

I think given that 3.1.0 is shipping with this commit https://github.com/bazelbuild/bazel/commit/0bbe38176e15d891a4e4cab2d8425e292de6cc5f and it was just a brief mention in the release notes, I'd say this one is safe to go out with the same treatment.

tetromino:
What would be a good blurb to put in RELNOTES?

keith:
I think

> Bazel now passes `-DNS_BLOCK_ASSERTIONS=1` for all architectures for Apple platforms

Would be appropriate

susinmotion:
Any updates on review?

keith:
I've updated the description to include relnotes. Is this good to merge now?

meisterT:
ping @jmmv

PR #10932, opened 2020-03-09
Update testing.md

Vertexwahn:
Target must be named "myrules_test" or command `bazel test //mypkg:myrules_test` needs to be changed to `bazel test //mypkg:myrules_tests`


Vertexwahn:
@googlebot I signed it!

Issue #4867, opened 2018-03-16
add quiet mode to bazel run


ixdy:
### Description of the problem / feature request:
`bazel run` always prints out a few INFO log messages before running the command, and there seems to be no way of suppressing that output.

For example, I want to suppress all of these `INFO:` lines when running the following in kuberentes/kubernetes:
```console
$ bazel run  --show_result=0 --noshow_progress --noshow_loading_progress --logging=0 cmd/kubectl -- version --client
INFO: Analysed target //cmd/kubectl:kubectl (589 packages loaded).
INFO: Found 1 target...
INFO: Elapsed time: 79.858s, Critical Path: 39.71s

INFO: Running command line: bazel-bin/cmd/kubectl/linux_amd64_pure_stripped/kubectl version
Client Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0-alpha.0.1096+05ae431ea0978e", GitCommit:"05ae431ea0978ee2fc0be91e5b9bc82feb97b3a2", GitTreeState:"clean", BuildDate:"2018-03-16T22:46:18Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

### Feature requests: what underlying problem are you trying to solve with this feature?
I'd like to be able to build and run binaries through bazel without any additional information being written to stderr/stdout, since the tool being run also prints to stderr/stdout.

### What operating system are you running Bazel on?
Linux

### What's the output of `bazel info release`?
`release 0.11.1`

###  Have you found anything relevant by searching the web?
Found a related discussion in https://groups.google.com/forum/#!topic/bazel-discuss/JVCotYRXeyk.
I couldn't find any feature requests filed after @ulfjack's last reply there.

philwo:
Agree that this would be nice to have.

Realistically, I won't have time to work on this in the next months, even though this should not be a huge change.

@ixdy Is the [--script_path flag](https://github.com/bazelbuild/bazel/blob/3c5a1098af0c5ae80d4e3b1fc52dd1fef6027d43/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java#L125) a possible workaround for your usecase?

jlisee:
[...]

PR #10936, opened 2020-03-10
Add option to filter events from UI

AlessandroPatti:
At the moment there is not an option to easily control the verbosity of Bazel UI. The `ui_event_filters` option allows to arbitrarily filter events from the UI. The syntax is the same as `build_tag_filters` or `output_group`, so it is possible to add or remove events, using leading `+/-`, or override the default set completely with direct assignment.

Related: #4867

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
@@ -238,10 +238,13 @@ public class UiOptions extends OptionsBase {
       name = "ui_event_filters",
       converter = EventFiltersConverter.class,
       defaultValue = "null",
-      metadataTags = {OptionMetadataTag.HIDDEN},
-      documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
-      effectTags = {OptionEffectTag.UNKNOWN},
-      help = "Specifies which events to show in the UI.",
+      documentationCategory = OptionDocumentationCategory.LOGGING,
+      effectTags = {OptionEffectTag.TERMINAL_OUTPUT},
+      help =
+          "Specifies which events to show in the UI. It is possible to add or remove events "
+              + "to the default ones using leading +/-, or override the default "
+              + "set completely with direct assignment. The set of supported event kinds "
+              + "include INFO, DEBUG, ERROR and more.",
       allowMultiple = true)
   public List<EventKind> eventFilters;
 
--- a/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
@@ -13,13 +13,20 @@
 // limitations under the License.
 package com.google.devtools.build.lib.runtime;
 
+import com.google.devtools.build.lib.events.EventKind;
 import com.google.devtools.build.lib.runtime.UiStateTracker.ProgressMode;
+import com.google.devtools.common.options.Converter;
+import com.google.devtools.common.options.Converters.CommaSeparatedOptionListConverter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -178,6 +234,17 @@ public ProgressModeConverter() {
       help = "Report all events known to the Bazel UI.")
   public boolean experimentalUiDebugAllEvents;
 
+  @Option(
+      name = "ui_event_filters",
+      converter = EventFiltersConverter.class,
+      defaultValue = "null",
+      metadataTags = {OptionMetadataTag.HIDDEN},
+      documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
+      effectTags = {OptionEffectTag.UNKNOWN},
```
philwo: These should be `OptionDocumentationCategory.LOGGING` and `OptionEffectTag.TERMINAL_OUTPUT`.

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -178,6 +234,17 @@ public ProgressModeConverter() {
       help = "Report all events known to the Bazel UI.")
   public boolean experimentalUiDebugAllEvents;
 
+  @Option(
+      name = "ui_event_filters",
+      converter = EventFiltersConverter.class,
+      defaultValue = "null",
+      metadataTags = {OptionMetadataTag.HIDDEN},
```
philwo: Let's not make this flag hidden (this would only be appropriate for experimental flags or flags that are not meant to be specified by a human user).

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -178,6 +234,17 @@ public ProgressModeConverter() {
...
+      documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      help = "Specifies which events to show in the UI.",
```
philwo: Can we expand the help text a bit here and mention the + /  - syntax and some examples of what kind of events can be specified here? (I had to look up the EventKind class to figure out what the supported events are.)

philwo:
Looks like the new test just requires a slight tweak to also catch the message on Windows, otherwise looks good to me. :)

AlessandroPatti:
> Looks like the new test just requires a slight tweak to also catch the message on Windows, otherwise looks good to me. :)

Replaced full path with `.*` regex. Not sure if it is the way to go or there is a cross-platform equivalent of `pwd`.

philwo:
I'm running into an issue with lots of tests failing now, because https://github.com/bazelbuild/bazel/commit/3d2d8b627500307f7317ad9c7b0eecdf20b81674 was submitted in the meantime, which removed the ability to distinguish "the option was never specified" (= `eventFilters` is null) and "the user explicitly specified an empty list" (= `eventFilters` is an empty list).

Any idea how to fix this in a clean way?

AlessandroPatti:
I am storing the filtered events rather than the allowed once. Makes it easier to handle the new default.

AlessandroPatti:
Any update on this?

philwo:
Sorry for the delay! Looks good to me, importing and merging it now.

philwo:
FYI, I had to modify one string in the `expect_log` part in the new test case, because for whatever reason, Bazel wouldn't print the "0 processes" on our internal presubmit tests. I replaced it with "Elapsed time", which appears to be reliably printed.

Issue #10005, opened 2019-10-13
Remove native Protobuf rules


Yannic:
This is the ☔️-bug for removing the native versions of everything related to Protocol Buffers from Bazel.

Blocked on:
  - [x] bazelbuild/rules_proto#6
  - [x] bazelbuild/bazel#10006
  - [x] bazelbuild/bazel#10484
  - [x] bazelbuild/bazel#7514
  - [x] bazelbuild/rules_proto#58
  - [x] bazelbuild/bazel#12037 
  - [x] bazelbuild/bazel#12431
  - [ ] bazelbuild/bazel#12484 
  - [x] bazelbuild/bazel#12486

jin:
cc @lberki

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

PR #10937, opened 2020-03-10
Create proto_toolchain rule

Yannic:
This change adds a new `proto_toolchain` rule which we will use later to resolve
tools and options required to compile Protocol Buffers, thus eventually replacing
some of the command-line options from `ProtoConfiguration`.

To preclude inconsistencies in the migration period, we will start with a
`proto_toolchain` rule that doesn't have any arguments and populate
its provider from the existing command-line options. After all consumers
have been migrated to resolve the tools from the toolchain instead of
reading the command-line, we will replace all proto-specific flags with
attributes on `proto_toolchain` (or similar, language-specific toolchains
for flags that only affect a specific language).

This change also adds a dependency on the new toolchain type to
`proto_library`, `cc_proto_library`, `java_proto_library`, and `java_lite_proto_library`.
Unfortunately, there seems to be no existing way to add a dependency on a
toolchain based on the value of a flag. To avoid breaking everyone, we add
a (default) `proto_toolchain` target to `@bazel_tools//tools/proto:toolchain` that
is registered in a workspace suffix and add a migration flag to disable them (thus
requiring users to upgrade to a compatible `@rules_proto` version. At this point,
it is unclear whether we want to do this migration now (i.e. for Bazel 4.0) or
wait for a later release.

See https://github.com/bazelbuild/rules_proto/pull/54

Updates #10005


Yannic:
This is ready for review now.

meisterT:
What's the state of this PR?

philwo:
We're replacing the "WIP" label with the native GitHub "draft PRs". I thus converted this PR to a draft.

Please mark this as ready for review when it's ready, or close it if you no longer intend to work on it. Thanks! 👍

Yannic:
Things have changed since opening this, so I'll close this and submit a new (updated) one.

Issue #9215, opened 2019-08-21
incompatible_generated_protos_in_virtual_imports: put generated `.proto` files into a virtual proto source root directory


lberki:
If a `proto_library` rule has both generated and non-generated sources, it would give rise to multiple source roots (one in the source tree and one in the output tree). This is unintuitive and incompatible with the `ProtoInfo` provider that has place for only one (`[ProtoInfo].proto_source_root`).

To resolve this, we create a virtual source root under `bazel-bin/<package>/_virtual_imports/<rule>`. 

No user-visible changes are expected except in cases where Starlark code relied on the exact layout of the source tree.

thomasvl:
Can you maybe give an example of what does/doesn't change because of this so one can better assess the impact on LANG_proto_library rules?

lberki:
Consider the following code say, in `x/BUILD`:

```
genrule(name="gen", srcs=[], outs=["a.proto"])
proto_library(name="p", srcs=["a.proto", "b.proto"])
```

At HEAD, this results in the proto_library having a `proto_source_root` of `x`. However, `a.proto` and `b.proto` are not actually siblings (one is under `x`, the other, under `bazel-bin/x`, thus, logic is required to figure out what command line to pass to `protoc`.

When this flag is flipped, the two `.proto` files are both symlinked into a directory `bazel-bin/x/virtual_imports/p`) and that's what `proto_source_root` will be. Simpler and removes guesswork for LANG_proto_library rules and makes it so that each `proto_library` only has one source root.

lberki:
Sorry, I meant `/_virtual_imports/` (including the leading underscore)

I wanted to provide enough information in providers so that you can construct the command line without assuming things about paths, or, even better, provide a Starlark method to do so, but I couldn't make it in time for Bazel 1.0. So I decided that I'll be satisfied with that being possible without backwards-incompatible changes.

Hard-coding `_virtual_imports` is fine -- that won't change after Bazel 1.0.

In the above case, it depends on whether `foo.proto` is generated or not. If it is not, the proto path of the proto compile action of `:bar` will be `.` and `bazel-bin/x/_virtual_imports/p`. If it is, in addition to that, `bazel-bin/m/n/o/_virtual_imports/foo` will also be there. IOW, each `proto_library` that has `import_prefix`, `strip_import_prefix` or generated sources creates its own `_virtual_imports` directory which contains the `.proto` files in its `srcs`.

[...]

PR #10939, opened 2020-03-10
Remove import-path guessing from ProtoCompileActionBuilder

Yannic:
Instead of trying to guess the correct source root and import-path of a proto file,
we save the real source root together with the .proto source when creating the
ProtoInfo provider and re-use that information for codegen actions.


lberki:
I seem to remember that my concern with this exact change was RAM use (that's why we use the otherwise-not-very-reasonable "null as import path" pattern). Let me run the numbers.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -589,168 +563,42 @@ static CustomCommandLine createCommandLineFromToolchains(
...
-     */
-    @Override
-    public void expandToCommandLine(Artifact proto, Consumer<String> args) {
-      for (String directProtoSourceRoot : directProtoSourceRoots.toList()) {
-        PathFragment sourceRootPath = PathFragment.create(directProtoSourceRoot);
-        String arg =
-            guessProtoPathUnderRoot(
-                outputDirectory, sourceRootPath, proto, siblingRepositoryLayout);
-        if (arg != null) {
-          args.accept("-I" + arg + "=" + proto.getExecPathString());
```
lberki: Why was this removed? This is important because this is how `protoc` learns the import map -> file path map.

Yannic: There are two ways to use `protoc`:
  - The "explicit import map"-approach we're using today, and
  - the "C++ #include like"-approach we'll use when this patch lands (i.e. `--proto_path=root1 --proto_path=root2`).

Both can't really deal with ambiguous import paths. `protoc` will happily accept `-Ifoo.proto=/path/to/a.proto -Ifoo.proto=/path/to/b.proto`, but which `foo.proto` it resolves to is, AFAICT, undefined. With `--proto_path`, `protoc` just uses the first matching file it can find.

If we continue with the "import map"-approach, we'd have to provide access to the transitive closure of import path -> file, which sounds like it requires a lot of memory (or still guess them, which would make this whole patch obsolete).

Given the above, and there's no real benefit in using the "explicit import map"-approach vs. "C++ #include like"-approach (`protoc` is negligibly slower with the new approach, but computing the command-line is slightly faster), I decided to drop the import map and let `protoc` handle it.

Yannic:
Let me actually change this patch to always use null if `source_root + import_path == artifact.exec_path` and see whether there are cases left where that doesn't hold. If not, we can get rid of the map altogether.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
lberki: Problem is, you can't really use the `--proto_path` approach because that would result in ambiguous imports (at least in the naive case), 

E.g. if you have both `//a:b.proto' and '@repo//c:d.proto` in your transitive closure, you would need to have both `.` and `external/repo` on your proto path, therefore, `import external/repo/c/d.proto` would work.

The workaround for that would be to create a little import directory for every `proto_library`, but that causes O(n^2) import lookup behavior and is thus unacceptable (I know, I tried because that's the cleanest approach)

The import map does not necessarily require a lot of RAM because the command line is constructed lazily (that's what all those mapping functions are for) and all you need is a description of how the command line is built, which is O(1) number of extra objects.

My game plan was to explicitly store the import path in `ProtoInfo` (i.e. no more `null` values in that `Pair`) -- that's also more RAM, but it's at least only O(n) in the number of .proto files. A bigger hammer would be to store a `NestedSet` of `(import path, [proto file])`, but that's not possible to do without changing the Starlark API.

lberki:
How would using null if `source_root + import_path == exec_path` help? Then you only have the exec path when you generate the command line, and you can't tell where the import path starts. You _might_ be able to get away with some smarter guessing logic -- the important use case at Google is when the import root is the root of the main repository, so if you signal that with a `null`, then RAM use becomes much less.

Yannic:
Let's go a step back and see what we want to achieve here.

The goal of this is to compute the import paths of all protos without guessing it. I see two possible options here:
  - Compute them at `proto_library` level and save them in `ProtoInfo`, or
  - Make it possible to derive them from the artifact and its proto_source_root.

The first option would be the cleanest, but it seems like that's unfeasible because of memory consumption.
For the second option, we need to ensure that `proto_source_root + import_path == exec_path` and have a map proto_source_root -> files (`Map<PathFragment, NestedSet<Artifact>>` [1]) in `ProtoInfo`. That would give us access to everything needed to derive the import path and build the command-line.

[1] possibly implemented as `NestedSet<Pair<PathFragment, NestedSet<Artifact>>>` and only flattened at execution time to keep construction O(direct deps) instead of O(transitive deps). Actually, we may not even need to flatten it into a map at all.

lberki:
Yeah, that `NestedSet<Pair<PathFragment, NestedSet<Artifact>>>` seems like a reasonable way to go. Then you'd still store duplicate information in `ProtoInfo` because e.g. `transitiveProtoSources` and `transitiveProtoSourceRoots` are derivable from it, but I think that's an acceptable amount of collateral damage if the memory use is not too large (the asymptotic memory use is certainly good, but the constant factor may still kill this)

Yannic:
Modified the change to use `NestedSet<Pair<PathFragment, ImmutableList<Artifact>>>` instead of guessing the import path. I was also able to remove some fields from `ProtoInfo` that are no longer needed (`strictImportableProtoSources*` also becomes obsolete, but these are a bit trickier to remove so I'll leave that for a follow-up). ptal.

lberki:
Good job!

I ran our internal benchmark and this results in net _improvement_ of memory use: 0.1%. This significantly exceeds my expectations of any such change having an unacceptable amount of memory overhead. Let me review the pull request in more detail.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -682,72 +617,52 @@ private static String guessProtoPathUnderRoot(
...
     }
   }
 
   @AutoCodec
   @AutoCodec.VisibleForSerialization
-  static final class ExpandToPathFnWithImports implements CapturingMapFn<Pair<Artifact, String>> {
-    private final String outputDirectory;
-    private final NestedSet<String> directProtoSourceRoots;
-    private final boolean siblingRepositoryLayout;
+  static final class ExpandProtosUnderSingleSourceRootToImportPathsArgsFn
```
lberki: Is this unused now (or I just can't find the references to this?)

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -272,12 +271,9 @@ private static String getOutputDirectory(RuleContext ruleContext) {
 
     // Add include maps
     addIncludeMapArguments(
-        getOutputDirectory(ruleContext),
         result,
-        areDepsStrict ? protoInfo.getStrictImportableProtoSourcesImportPaths() : null,
-        protoInfo.getStrictImportableProtoSourceRoots(),
-        protoInfo.getTransitiveProtoSources(),
-        siblingRepositoryLayout);
+        areDepsStrict ? protoInfo.getImportableProtos() : null,
```
lberki: Mind removing the `siblingRepositoryLayout` variable from line 266? (it's unused now as far as I can tell)

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoInfo.java
```diff
@@ -55,11 +55,11 @@ public Provider() {
   private final NestedSet<Pair<Artifact, String>> strictImportableProtoSourcesImportPaths;
   private final NestedSet<Pair<Artifact, String>>
       strictImportableProtoSourcesImportPathsForDependents;
-  private final NestedSet<String> strictImportableProtoSourceRoots;
-  private final NestedSet<Pair<Artifact, String>> exportedProtoSourcesImportPaths;
-  private final NestedSet<String> exportedProtoSourceRoots;
   private final Artifact directDescriptorSet;
   private final NestedSet<Artifact> transitiveDescriptorSets;
+  private final NestedSet<Pair<PathFragment, ImmutableList<Artifact>>> importableProtos;
```
lberki: nit: add back the `strict` prefix, i.e. call this `strictImportableProtos`

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -682,72 +617,52 @@ private static String guessProtoPathUnderRoot(
         }
       };
 
+  private static String computeImportPath(PathFragment protoSourceRoot, Artifact proto) {
+    PathFragment importPath = proto.getExecPath().relativeTo(protoSourceRoot);
```
lberki: nit: why not inline the `importPath` variable?

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -682,72 +617,52 @@ private static String guessProtoPathUnderRoot(
...
-          }
-        }
+    public void expandToCommandLine(Artifact proto, Consumer<String> args) {
+      args.accept(computeImportPath(protoSourceRoot, proto));
+    }
+  }
+
+  @AutoCodec
+  @AutoCodec.VisibleForSerialization
+  static final class ExpandToImportPathsArgsFn
```
lberki: opt: you can make these lambdas instead of inner classes it you want

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java
```diff
@@ -404,44 +399,57 @@ private static Library createLibraryWithVirtualSourceRootMaybe(
...
    */
-  private static NestedSet<String> computeExportedProtoSourceRoots(
-      RuleContext ruleContext, String currentProtoSourceRoot) {
-    return getProtoSourceRootsOfAttribute(ruleContext, currentProtoSourceRoot, "exports");
+  private static NestedSet<Pair<PathFragment, ImmutableList<Artifact>>> computeTransitiveProtos(
+      RuleContext ruleContext,
+      NestedSet<Pair<PathFragment, ImmutableList<Artifact>>> exportedProtos) {
+    NestedSetBuilder<Pair<PathFragment, ImmutableList<Artifact>>> protos =
+        NestedSetBuilder.stableOrder();
+    protos.addTransitive(exportedProtos);
```
lberki: This is inconsistent with how `transitiveProtoSources` (the `NestedSet<Artifact>`) is computed: that is only local sources + deps and this one is local sources + deps + exports. Make this consistent (or else explain in a comment why the discrepancy exists)

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java
```diff
@@ -179,21 +180,18 @@ private ProtoCommon() {
...
         String sourceRoot,
         ImmutableList<Pair<Artifact, String>> importPathSourcePair) {
-      this.sources = sources;
       this.sourceRoot = sourceRoot;
       this.importPathSourcePair = importPathSourcePair;
     }
 
     public ImmutableList<Artifact> getSources() {
-      return sources;
+      return ImmutableList.copyOf(Iterables.transform(importPathSourcePair, p -> p.first));
```
lberki: nit, opt: you can replace `p -> p.first` with `Pair::getFirst` (I'm not sure if it's an improvement)

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java
```diff
@@ -404,44 +399,57 @@ private static Library createLibraryWithVirtualSourceRootMaybe(
...
-    return getProtoSourceRootsOfAttribute(ruleContext, currentProtoSourceRoot, "deps");
+  private static NestedSet<Pair<PathFragment, ImmutableList<Artifact>>> computeExportedProtos(
+      RuleContext ruleContext, Library library) {
+    NestedSetBuilder<Pair<PathFragment, ImmutableList<Artifact>>> protos =
+        NestedSetBuilder.stableOrder();
+    protos.add(new Pair<>(PathFragment.create(library.getSourceRoot()), library.getSources()));
+    for (ProtoInfo info : ruleContext.getPrerequisites("exports", TARGET, ProtoInfo.PROVIDER)) {
+      protos.addTransitive(info.getExportedProtos());
+    }
+    if (library.getSources().isEmpty()) {
```
lberki: It appears that `computeExportedProtos` at HEAD does not special-case rules without sources and this one does. Why is that?

lberki:
I see a bunch of failures in our internal test battery (unfortunately, I am too pressed for time to investigate myself). There are two kinds:

1. When a command line is expected to contain `--allowed_public_imports=` but instead contains `--allowed_public_imports <list of .proto files>`
2. One where the command line is expected to contain `-Ifoo/bar.proto=blaze-out/k8-fastbuild/bin/foo/bar.proto`, but instead it contains `-Iblaze-out/k8-fastbuild/bin/foo/bar.proto=blaze-out/k8-fastbuild/bin/foo/bar.proto`

My guess is that the latter is related to buggy import part computation (I can fix it when I import it if you have no idea) and the former is related to the export set computation you changed (I'll take a look, but for now, I'd like to know why you changed it in the first place)

I was somewhat surprised, but then I realized that we don't have external tests for exports, so no surprise you broke it with this change :( Once I know why you changed the logic, I'll take a look myself.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java
Yannic: Actually it does, but it's not so obvious:
https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java;l=67?q=ProtoCommon
(Tested here: https://source.bazel.build/bazel/+/master:src/test/java/com/google/devtools/build/lib/rules/proto/BazelProtoLibraryTest.java;l=53?q=BazelProtoLibraryTest)

Yannic:
Argh, turns out I misunderstood `exports` to behave like `java_library`'s `exports`, but it's only used to indicate that direct sources may `public import` it (iff it's also in `deps`). I can fix that.

For the second kind of failure you're seeing, I see two possible explanations:
  - The tests don't construct `ProtoInfo` correctly and blindly set `proto_source_root` to `""` and only add the correct source root to `transitive_proto_path`. I had to fix that in `ProtoCompileActionBuilderTest`, so hopefully, that's what you're seeing as well.
  - There are cases inside Google where `ProtoInfo.proto_source_root` is not a prefix of (all) direct sources (e.g. because the files are generated). This is a precondition for this change and not fixable from my side.

lberki:
I have a test case for one of the failures (add this to `BazelProtoLibraryTest.java`):

```
 @Test
  public void testSourceAndGeneratedProtoFiles() throws Exception {
    scratch.file("a/BUILD",
        "genrule(name='g', srcs=[], outs=['g.proto'], cmd = '')",
        "proto_library(name='p', srcs=['s.proto', 'g.proto'])");

    Iterable<String> commandLine = paramFileArgsForAction(getDescriptorWriteAction("//a:p"));
    String genfiles = getTargetConfiguration().getGenfilesFragment().toString();
    assertThat(commandLine).containsAtLeast(
        "-Ia/s.proto=a/s.proto",
        "-Ia/g.proto=" + genfiles + "/a/g.proto");
  }
```

The trick is to fix this in such a way that one doesn't need to create virtual proto source roots for every `proto_library` rule with generated sources (so that RAM use is under control, although if you prefer, I can run the benchmarks that way, too, and see what happens)

I *think* what would work is to have multiple `ProtoCommon.Library` instances for `proto_library` rules that have both source and generated `.proto` files.

lberki:
Re, the first, it appears to be the case. The simplest test case appears to be to test whether `ProtoInfo.getExportedProtos()` returns the `.proto` files in the current rule or not.

Although you should probably not change any behavior in this change, I couldn't resist and tried to grok how the `exports` logic works.

It appears that the proto compile action has the strict importable protos of the rules in its `exports` in the `--allowed_public_imports` flag. And the strict importable protos in turn are computed as:

* If the rule has sources, the sources of the rule and the strict importable protos for dependents of `deps` 
* If the rule has no sources, the strict importable protos of `deps`

And the strict importable protos for dependents are:

* The sources if the rule has them
* The strict importable protos for dependents of the `deps` if the rule has no sources

So it appears that sources of `deps` of `exports` are allowed public imports if we assume that every `proto_library` rule has sources? That doesn't make much sense... but that's what the methods `computeExportedProtos`, `computeStrictImportableProtos` and `computeStrictImportableProtosImportPathsForDependents' methods seem to do?

Yannic:
> I have a test case for the second failure (add this to `BazelProtoLibraryTest.java`):
> 
> ```
>   @Test
>   public void testSourceAndGeneratedProtoFiles() throws Exception {
>     scratch.file("a/BUILD",
>         "genrule(name='g', srcs=[], outs=['g.proto'], cmd = '')",
>         "proto_library(name='p', srcs=['s.proto', 'g.proto'])");
> 
>     Iterable<String> commandLine = paramFileArgsForAction(getDescriptorWriteAction("//a:p"));
>     String genfiles = getTargetConfiguration().getGenfilesFragment().toString();
>     assertThat(commandLine).containsAtLeast(
>         "-Ia/s.proto=a/s.proto",
>         "-Ia/g.proto=" + genfiles + "/a/g.proto");
>   }
> ```
> 
> This is because if you have both source and generated `.proto` files, you need more than one source root for a single `proto_library`.
> 
> The trick is to fix this in such a way that one doesn't need to create virtual proto source roots for every `proto_library` rule with generated sources (so that RAM use is under control, although if you prefer, I can run the benchmarks that way, too, and see what happens)
> 
> I _think_ what would work is to have multiple `ProtoCommon.Library` instances for `proto_library` rules that have both source and generated `.proto` files. This would make `.direct_source_root` a lie, but not more of a lie than it currently is.

Is `.direct_source_root` really a lie, though. At least Bazel seems to already put all `proto_library` targets with generated sources into a virtual proto source root directory [1,2].
I think having multiple (direct) source roots for a single `proto_library` target will make things way more complicated, so yes, please run the benchmark.

[1] https://github.com/bazelbuild/bazel/issues/9215
[2] https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java;l=276?q=ProtoCommon

Yannic:
> Re, the first, it appears to be the case. The simplest test case appears to be to test whether `ProtoInfo.getExportedProtos()` returns the `.proto` files in the current rule or not.
> 
> Although you should probably not change any behavior in this change, I couldn't resist and tried to grok how the `exports` logic works.
> 
> It appears that the proto compile action has the strict importable protos of the rules in its `exports` in the `--allowed_public_imports` flag. And the strict importable protos in turn are computed as:
> 
> * If the rule has sources, the sources of the rule and the strict importable protos for dependents of `deps`
> * If the rule has no sources, the strict importable protos of `deps`
> 
> And the strict importable protos for dependents are:
> 
> * The sources if the rule has them
> * The strict importable protos for dependents of the `deps` if the rule has no sources
> 
> So it appears that sources of `deps` of `exports` are allowed public imports if we assume that every `proto_library` rule has sources? That doesn't make much sense... but that's what the methods `computeExportedProtos`, `computeStrictImportableProtos` and `computeStrictImportableProtosImportPathsForDependents' methods seem to do?

Yes, that seems to be what those methods do.
I don't think my previous understanding of `--direct_dependencies=srcs + exportedSources(deps) and --allowed_public_imports=srcs + exportedSources(exports)` is necessarily a bad idea. It would make `proto_library` more consistent with other `<lang>_library` rules that allow `exports`. I agree that that change shouldn't be part of this PR, though.

I'll refrain from updating the PR for this until we've figured out the other issue.

lberki:
Yep, I agree on both counts :)

What's the "other issue"? The `-I<something under bazel-out>` I mentioned above?

I'll run the benchmark as soon as you update this change. What's the plan? As long as you don't create virtual proto source roots for every `proto_library` with generated sources, I'm fairly certain that RAM usage will be acceptable.

lberki:
/cc @bazelbuild/rules_closure_collaborators 

This will probably conflict with https://github.com/bazelbuild/bazel/pull/10966 (fortunately, it's only a textual conflict, no complex refactoring will be needed)

lberki:
/cc @cheister
This will probably conflict with #10966 (fortunately, it's only a textual conflict, no complex refactoring will be needed)

Yannic:
Yes, the `-I<something under bazel-out>` is the "other issue".

Can you check what's the status-quo on creating virtual source roots for generated protos? Unless I'm missing something really obvious here, Bazel already creates virtual source roots for all targets that have at least one generated source. We're still putting the original source in the `Pair<Artifact, ImportPath>`, but that doesn't make a difference because the virtual file is still there. What am I missing?

lberki:
What you are missing is the very sad existence of the `generatedProtosInVirtualImports` argument of `ProtoCommon.createProtoInfo()`. There are two reasons for that. One is memory use of the extra symlink actions, which I hope is okay. It probably is.

The second, more hairy one is that of implicit outputs: if you write `proto_library(srcs=["foo.proto"])` in e.g. `a/BUILD`, Blaze will automatically have a target `//a:foo.pb.cc` and the root-relative path of that file must be `a/foo.pb.cc` and thus the `.proto` file cannot be in a virtual imports directory.

@c-mita is working on removing this (in fact, I was hoping he'd be done with it last week), and then we can hopefully make Blaze and Bazel consistent.

lberki:
...and as to the fix, you can either wait until @c-mita is done, although that will take at least a week until the submits the change and it percolates through our release process, but more like two, or use the "two `Library` instances" approach.

WDYT?

Yannic:
I'll wait until @c-mita is finished and we can try to put all generated `.proto` files into a virtual source root.

In the meantime, could we move forward with the "artifact -> artifact symlink PR" (https://github.com/bazelbuild/bazel/pull/10695)?

lberki:
@c-mita is done with the removal of implicit outputs! (well, he submitted the change a week ago, but we had to wait for our internal release to make sure that it doesn't break anything). Your path is free!

lberki:
(And I'm running our internal test suite on #10695 as we speak)

Yannic:
Modified the PR to force putting generated files under a virtual source root. Let's run the benchmark to get some numbers.

(No other changes for this benchmark run to remove the possible noise, and maybe also because there were complicated merge conflics with #10966 and I haven't found the time to fix the exports thingy :) ).

lberki:
Sigh. Nothing as easy as it seems :( It appears that your change broke your internal version in some way. I'd say I'll debug it but today is hectic so I'll probably won't have time to do so.

lberki:
Nothing is as easy as it seems :(

I tried this in our internal code base and it appears that in addition to there being multiple places in the code of Bazel that had to be fixed, our code base seems to rely on not creating virtual import directories for generated `.proto` files. It's a pity because it does look like the cleanup is of a manageable size, but it's not something we can tackle anytime soon.

So I propose going with the "two `Library` instances" approach. WDYT? That definitely works.

Yannic:
Found another neat solution: Instead of using `Pair<Artifact, nullable_import_path>`, we can save `Pair<Artifact, real_source_root>` (`ProtoSource` for readability) and always derive the import path. Memory usage should be the same. `ProtoInfo.proto_source_root` is still the same lie as before, but `for all ProtoSource: ProtoSource.source.exec_path == ProtoSource.source_root + ProtoSource.import_path`.

Yannic:
@lberki ptal

Yannic:
pinging this again

philwo:
@lberki Can we merge this?

Yannic:
(rebased to fix conflicts)

lberki:
Thanks for rebasing! I wanted to look at it today, but it's not a trivial change so I'll need to find a chunk of time in which to do it :( I do want to merge this, though, if at all possible.

lberki:
Update: I started reviewing this change, had to resolve conflicts when importing it, then found a bug in javac (!) while fixing an internal test case. Let's see if I will have enough time and brain to continue the review today.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -288,21 +280,16 @@ private static String getOutputDirectory(RuleContext ruleContext) {
       result.add("--disallow_services");
     }
     if (checkStrictImportPublic) {
-      NestedSet<Pair<Artifact, String>> protosInExports =
-          protoInfo.getExportedProtoSourcesImportPaths();
-      if (protosInExports.isEmpty()) {
+      NestedSet<ProtoSource> publicImportSources = protoInfo.getPublicImportSources();
+      if (publicImportSources.isEmpty()) {
         // This line is necessary to trigger the check.
         result.add("--allowed_public_imports=");
```
lberki, line 286: A test case in our internal test battery fails. I can't export it at the moment, but the BUILD file it uses is as follows:

```
proto_library(
  name = 'myProto',
  srcs = ['myProto.proto'],
)
```

then verifies that its proto compile action has `--allowed_public_imports=`. This fails, and instead, it has `--allowed_public_imports test/myProto.proto`. I think I know why (see the other comment)

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCommon.java
```diff
@@ -568,37 +525,27 @@ public static ProtoInfo createProtoInfo(
...
    */
-  private static NestedSet<Pair<Artifact, String>> computeExportedProtos(
-      ImmutableList<ProtoInfo> exports) {
-    NestedSetBuilder<Pair<Artifact, String>> result = NestedSetBuilder.stableOrder();
-    for (ProtoInfo provider : exports) {
-      result.addTransitive(provider.getStrictImportableProtoSourcesImportPaths());
+  private static NestedSet<ProtoSource> computePublicImportProtos(
+      ImmutableList<ProtoSource> directSources, ImmutableList<ProtoInfo> exports) {
+    NestedSetBuilder<ProtoSource> result = NestedSetBuilder.stableOrder();
+    result.addAll(directSources);
```
lberki, line 546: This line was not present in `computeExportedProtos()` and is probably the cause of the aforementioned test breakage

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoCompileActionBuilder.java
```diff
@@ -288,21 +280,16 @@ private static String getOutputDirectory(RuleContext ruleContext) {
...
             "--allowed_public_imports",
             VectorArg.join(":")
-                .each(protosInExports)
-                .mapped(
-                    new ExpandToPathFnWithImports(
-                        getOutputDirectory(ruleContext),
-                        protoInfo.getTransitiveProtoSourceRoots(),
-                        siblingRepositoryLayout)));
+                .each(publicImportSources)
+                .mapped((s, args) -> s.getImportPath().getPathString()));
```
lberki, line 292: Mind extracting these to a separate class just like the other argument map functions? (one of our internal tests requires that, long story)

lberki (commented):
I also ran some memory benchmarks and there is even a slight improvement!

I'll need to take a closer look, but if you don't mind, I'll wait until you shuffe ProtoInfo back so that it's easier to see what's going on.

Yannic (commented):
Sorry for the delay, took me some time to get this stuff back into memory. Tried my best to shuffle everything back. PTAL

> I also ran some memory benchmarks and there is even a slight improvement!

I'm curious how much of an improvement this is. Any chance you can share the numbers?

lberki:
It was a ~0.1% improvement, if my memory serves well, so ntohing very significant, but it was still a pleasant surprise.

I tried to import this, then `SharedProtoLibraryTest` failed because of some completely unnecessary complexity in our test setup.So I recommend moving those tests back to `BazelProtoLibraryTest` for the time being so that sorting out our test framework is not a dependency of submitting this change. 

WDYT?

Yannic:
Sure, merged them into `BazelProtoLibraryTest` and added a TODO for you ;).

lberki:
Wow, eight months already! I fixed some linter errors and I could nitpick a bit more, but instead let's submit this and see if it sticks. It should, but it's an intricate enough problem that I'm not sure.

lberki (approved):

lberki:
Submitted. Let's see whether this stands the test of our internal source tree! (we'll know tomorrow evening with reasonable certainty in Europe if all goes well, if not, a day or two after that)

PR #10947, opened 2020-03-11
Don't consider a workspace directory a marker of being in a workspace

segiddins:
Jenkins workers tend to place all of their Jenkins workspaces in a
directory named `workspace`. On a case-insensitive filesystem,
Bazel would consider anything run by Jenkins as being in a Bazel
workspace, because of that `workspace` directory being in the
filesystem hierarchy of the cwd.

Avoid this by only looking for _files_ named `workspace` when
determining if a directory is potentially a Bazel workspace.


philwo:
Thanks @segiddins! That's a nice fix.

philwo:
@segiddins No, all good - I'll import it now. :)

PR #10958, opened 2020-03-12
Fix call to assert_().fail(String).

katre:
This was removed between Truth 0.45 and Truth 1.0.1

RELNOTES: None.


katre:
This has to go in so that https://github.com/bazelbuild/bazel/pull/10956 will pass tests.

Issue #9412, opened 2019-09-20
target with 'no-remote-exec' is not cached remotely


ishikhman:
The following test is failing at the last check:

```
function test_tag_no_remote_exec() {
  mkdir -p a
  cat > a/BUILD <<'EOF'
genrule(
  name = "foo",
  srcs = [],
  outs = ["foo.txt"],
  cmd = "echo \"foo\" > \"$@\"",
  tags = ["no-remote-exec"]
)
EOF

  bazel build \
    --spawn_strategy=remote,local \
    --remote_executor=grpc://localhost:${worker_port} \
    //a:foo >& $TEST_log || "Failed to build //a:foo"

  expect_log "1 local"
  expect_not_log "1 remote"

  bazel clean

  bazel build \
    --spawn_strategy=remote,local \
    --remote_executor=grpc://localhost:${worker_port} \
    //a:foo >& $TEST_log || "Failed to build //a:foo"

  expect_not_log "1 remote"
  expect_log "remote cache hit"
}
```

**When this is the case?**
When from the very beginning bazel selects non-local strategy for a target:
- if `tags = ["no-remote-exec"]`
- if  `tags = ["no-remote"]`

Qinusty:
Which version of bazel does this affect and is it fixed in the 3.0 release candidates?

Qinusty:
After looking into this for a little bit, this seems to be caused by the combination of the effects of the `--remote_executor=` flag and the tag `no-remote-exec`.

From what I can see, the `--remote-executor=` flag forces the `BlazeExecutor` class to be instantiated with it's spawn cache configured as `NoSpawnCache`. 

Additionally, the use of `no-remote-exec` as a tag forces the `ProxySpawnActionContext.resolve()` function to filter out `RemoteSpawnStrategy` as a possible spawn execution strategy for the action. 

So this leaves us with a Standalone/Sandboxed runner etc, making use of the `AbstractSpawnStrategy` logic for caching (The only alternative to this is the `RemoteSpawnRunner` class which handles remote caching in relation to remote exection).

The `AbstractSpawnStrategy` caching logic depends on the `SpawnCache` for the executor being configured as part of it's ContextMap. All in all, this leaves `no-remote-exec` actions falling into a code path without any remote caching configuration.

I'll continue looking into this, just thought I'd document my findings so far.

**NOTE**: All of this analysis was done on bazel version 1.2.1 but I have confirmed this issue exists in version 2.0 (Unsure about 3.0).

[...]

PR #10964, opened 2020-03-14
Fix no-remote-exec caching functionality

Qinusty:
This commit addresses issue #9412 and provides all
instances of the RemoteSpawnStrategy with a RemoteSpawnCache
to ensure that the AbstractSpawnStrategy caching logic behaves
as expected for no-remote-exec actions.

This patch was written for 1.2.1 and adapted to latest master.

EDIT: Force pushed different changes 19/03/2020, see below for further discussion
EDIT: Rebased as of 28/04/20, unsure if 08efd761721 should be needed

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
@@ -114,6 +114,10 @@ public abstract class AbstractSpawnStrategy implements SandboxedSpawnStrategy {
     if (cache == null) {
       cache = SpawnCache.NO_CACHE;
     }
+    // If we are using the RemoteSpawnRunner, it handles caching internally so avoid caching logic here.
+    if (spawnRunner.getName().equals("remote")) {
+      cache = SpawnCache.NO_CACHE;
+    }
     SpawnResult spawnResult;
     ExecException ex = null;
     try (CacheHandle cacheHandle = cache.lookup(spawn, context)) {
--- a/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
@@ -107,9 +107,7 @@ public abstract class AbstractSpawnStrategy implements SandboxedSpawnStrategy {
     final Duration timeout = Spawns.getTimeout(spawn);
     SpawnExecutionContext context =
         new SpawnExecutionContextImpl(spawn, actionExecutionContext, stopConcurrentSpawns, timeout);
-    // TODO(ulfjack): Provide a way to disable the cache. We don't want the RemoteSpawnStrategy to
-    // check the cache twice. Right now that can't happen because this is hidden behind an
-    // experimental flag.
+
     SpawnCache cache = actionExecutionContext.getContext(SpawnCache.class);
     // In production, the getContext method guarantees that we never get null back. However, our
     // integration tests don't set it up correctly, so cache may be null in testing.
--- a/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
@@ -114,8 +114,8 @@ public abstract class AbstractSpawnStrategy implements SandboxedSpawnStrategy {
     if (cache == null) {
       cache = SpawnCache.NO_CACHE;
     }
-    // If we are using the RemoteSpawnRunner, it handles caching internally so avoid caching logic here.
-    if (spawnRunner.getName().equals("remote")) {
+    // Avoid caching for runners which handle caching internally e.g. RemoteSpawnRunner
+    if (spawnRunner.handlesCaching()) {
       cache = SpawnCache.NO_CACHE;
     }
     SpawnResult spawnResult;
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
@@ -112,16 +112,11 @@ final class RemoteActionContextProvider implements ExecutorLifecycleListener {
   }
 
   /**
-   * Registers a spawn cache action context if this instance was created without an executor,
-   * otherwise does nothing.
+   * Registers a spawn cache action context
    *
    * @param registryBuilder builder with which to register the cache
    */
-  public void registerSpawnCacheIfApplicable(ModuleActionContextRegistry.Builder registryBuilder) {
-    if (executor != null) {
-      return; // No need to register cache if we're using a remote executor.
-    }
-
+  public void registerSpawnCache(ModuleActionContextRegistry.Builder registryBuilder) {
     RemoteSpawnCache spawnCache =
         new RemoteSpawnCache(
             env.getExecRoot(),
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
@@ -314,7 +314,7 @@ final class RemoteSpawnCache implements SpawnCache {
   }
 
   private static boolean useRemoteCache(RemoteOptions options) {
-    return !isNullOrEmpty(options.remoteCache);
+    return !isNullOrEmpty(options.remoteCache) || !isNullOrEmpty(options.remoteExecutor);
   }
 
   private static boolean useDiskCache(RemoteOptions options) {
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
@@ -472,6 +472,11 @@ public class RemoteSpawnRunner implements SpawnRunner {
     return Spawns.mayBeExecutedRemotely(spawn);
   }
 
+  @Override
+  public boolean handlesCaching() {
+    return true;
+  }
+
   private void maybeWriteParamFilesLocally(Spawn spawn) throws IOException {
     if (!executionOptions.shouldMaterializeParamFiles()) {
       return;
--- a/src/main/java/com/google/devtools/build/lib/exec/SpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/SpawnRunner.java
@@ -258,6 +258,9 @@ public interface SpawnRunner {
   /** Returns whether this SpawnRunner supports executing the given Spawn. */
   boolean canExec(Spawn spawn);
 
+  /** Returns whether this SpawnRunner handles caching of actions internally. */
+  boolean handlesCaching();
+
   /** Returns the name of the SpawnRunner. */
   String getName();
 
--- a/src/main/java/com/google/devtools/build/lib/exec/local/LocalSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/local/LocalSpawnRunner.java
@@ -181,6 +181,11 @@ public class LocalSpawnRunner implements SpawnRunner {
     return true;
   }
 
+  @Override
+  public boolean handlesCaching() {
+    return false;
+  }
+
   protected Path createActionTemp(Path execRoot) throws IOException {
     return execRoot.getRelative(
         java.nio.file.Files.createTempDirectory(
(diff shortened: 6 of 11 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
```diff
@@ -113,6 +113,11 @@ public CacheHandle lookup(Spawn spawn, SpawnExecutionContext context)
         || (!Spawns.mayBeCachedRemotely(spawn) && useRemoteCache(options))) {
       // returning SpawnCache.NO_RESULT_NO_STORE in case the caching is disabled or in case
       // the remote caching is disabled and the only configured cache is remote.
+      System.out.println("RemoteSpawnCache has decided not to cache this action...");
```
ulfjack: This can't be checked in.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -228,39 +227,12 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
...
-            spawn, context, inputMap, actionKey, action, command, uploadLocalResults, e);
-      }
-
       ExecuteRequest.Builder requestBuilder =
           ExecuteRequest.newBuilder()
               .setInstanceName(remoteOptions.remoteInstanceName)
               .setActionDigest(actionKey.getDigest())
-              .setSkipCacheLookup(!acceptCachedResult);
+              // Always skip cache lookup as this is done in AbstractSpawnStrategy
+              .setSkipCacheLookup(true);
```
ulfjack: I wouldn't make this change. Note that the old code also made two lookups, so skipping here is a significant departure from before.

However, I also think that Bazel shouldn't call the remote cache at all when remote execution is enabled and instead directly try to execute the action. The cache lookup is an unnecessary additional round-trip.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
```diff
@@ -88,19 +88,20 @@ public void registerActionContexts(ExecutorBuilder executorBuilder) {
...
-              remoteOptions,
-              cache,
-              buildRequestId,
-              commandId,
-              env.getReporter(),
-              digestUtil,
-              filesToDownload);
+    // Always provide a spawnCache as this allows no-remote-exec actions
+    // to remotely cache actions.
+    RemoteSpawnCache spawnCache =
```
ulfjack: It would be even better to avoid using the cache w/ remote execution, and only apply it when the action ends up not being executed remotely.

src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
```diff
@@ -105,15 +105,17 @@ public boolean canExec(Spawn spawn, ActionContext.ActionContextRegistry actionCo
...
-    // experimental flag.
+
     SpawnCache cache = actionExecutionContext.getContext(SpawnCache.class);
     // In production, the getContext method guarantees that we never get null back. However, our
     // integration tests don't set it up correctly, so cache may be null in testing.
     if (cache == null) {
       cache = SpawnCache.NO_CACHE;
     }
+    // If we are using the RemoteSpawnRunner, it handles caching internally so avoid caching logic here.
+    if (spawnRunner.getName().equals("remote")) {
```
Qinusty: Open to suggestions about this, feels a little hacky.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
Qinusty: I think the `spawnCache` needs to be provided here as this registration function is only called once when the executor is initialized. I've introduced a check inside of `AbstractSpawnStrategy` to avoid using this cache if we're using a `RemoteSpawnRunner`

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
```diff
@@ -292,7 +292,7 @@ private void report(Event evt) {
   }
 
   private static boolean useRemoteCache(RemoteOptions options) {
-    return !isNullOrEmpty(options.remoteCache);
+    return !isNullOrEmpty(options.remoteCache) || !isNullOrEmpty(options.remoteExecutor);
```
Qinusty: This also feels slightly wrong, perhaps the function needs to be a class method to allow the `RemoteExecutionCache` to override it.

alanfalloon:
Any news on this PR? We are eagerly awaiting a fix for #9412.

davidcummingsaapl:
What can we do to expedite getting this change reviewed and submitted? It is significantly impacting the performance of our builds.

src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
ulfjack: Maybe add another method to the SpawnRunner interface?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
ulfjack: Oh, I see. The point is that you *want* to use the remote cache for *local actions* even when the remote executor is enabled.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
ulfjack: The class structure here is a bit more complex than I'd like, but this seems fine for now.

ulfjack: Oh, I have another concern - the remote executor may not allow writes to the action cache. In that case, we can't use remote caching even if it's enabled here.

Qinusty: In that case, would it not be a requirement for the bazel client to be ran with [--noremote_upload_local_results](https://docs.bazel.build/versions/master/command-line-reference.html#flag--remote_upload_local_results)?

In our case, we do not currently place this restriction on our remote executor frontend as we're currently transitioning to remote execution and still have some rules which don't work remotely.

I just tested this use case, if bazel attempts to push to the remote executor which rejects action cache writes, the output is as seen below (with the second line being a response from the remote executor.

```
WARNING: Writing to Remote Cache:
UNIMPLEMENTED: This service can only be used to get action results for instance "remote-execution"
```

src/main/java/com/google/devtools/build/lib/remote/RemoteActionContextProvider.java
Qinusty: Yes, in the case of certain rules being incompatible with remote execution, this allows a gradual move to remote execution by tagging targets which fail remotely as `no-remote-exec`. Currently this is quite a pain because of the relating issue causing no caching for these targets.

philwo:
@ulfjack Any concerns about this one? Otherwise I'd be happy to import it.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
ulfjack: IIUC, this change implicitly enables remote caching if a remote executor is enabled, so people who are relying on the current behavior will start to see those warnings, and they'll see one warning for every such action.

I suppose asking them to explicitly also set ```--noremote_upload_local_results``` is a reasonable workaround, as long as Bazel doesn't crash (as you said, and which I also double-checked in the source code).

ulfjack:
LGTM

@philwo, looks good to merge. Thanks!

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java
Qinusty: The change will also effect those who are relying on the current behaviour of not remotely caching `no-remote-exec` actions (To which I'd suggest using `no-remote` as the docs already detail).

philwo:
Just to make sure: What you're all writing here sounds like this is a inc^H^H^H semantic change, but apparently for people relying on certain behaviors at the moment, there's an easy workaround via a flag, is this correct? Could you propose a short text snippet that we'll incorporate into the release notes so that everyone is aware of what changes and what they have to look out for?

Qinusty:
> Just to make sure: What you're all writing here sounds like this is a inc^H^H^H semantic change, but apparently for people relying on certain behaviors at the moment, there's an easy workaround via a flag, is this correct? Could you propose a short text snippet that we'll incorporate into the release notes so that everyone is aware of what changes and what they have to look out for?

Actions tagged `no-remote-exec` and therefore executed locally are now remotely cacheable, this may cause warnings when used with a remote cache which does not allow the uploading of local action results. In this case, targets can be tagged `no-remote` or use [--noremote_upload_local_results](https://docs.bazel.build/versions/master/command-line-reference.html#flag--remote_upload_local_results) to prevent the uploading of all locally executed action results.

ifoox:
Hi there, wondering if there's anything that still needs to be done for this PR or if it's ready to go? (I happed to stumble across this issue as well).

Thanks!

philwo:
@ifoox I think it's fine. I'll merge this when I'm back from my leave (next Monday). If anyone else wants to do so before that, that's also great!

ifoox:
Thanks @philwo, that would be greatly appreciated! No rush, next week would be perfectly fine.

FYI I also tested this patch and it worked well for my instance of this problem. Thanks for sending it @Qinusty!

Qinusty:
I've rebased against latest master and resolved some conflicts around recent changes by @katre (Review appreciated, this PR always initializes the spawnCache as this is needed for no-remote-exec actions).

I've also had to add a commit to fix a build failure I've been getting since rebasing, error seen below. 
```
ERROR: /home/qinusty/Documents/projects/bazel/src/main/java/com/google/devtools/build/lib/syntax/BUILD:147:1: no such target '@bazel_tools//tools/jdk:jni': target 'jni' not declared in package 'tools/jdk' defined by /home/qinusty/.cache/bazel/_bazel_qinusty/58d05914cd15a0c432ca36aac2056b19/external/bazel_tools/tools/jdk/BUILD and referenced by '//src/main/java/com/google/devtools/build/lib/syntax:libcpu_profiler.so'
INFO: Call stack for the definition of repository 'com_google_protobuf' which is a http_archive (rule definition at /home/qinusty/Documents/projects/bazel/tools/build_defs/repo/http.bzl:336:16):
 - /home/qinusty/Documents/projects/bazel/WORKSPACE:96:1
ERROR: Analysis of target '//src/test/shell/bazel/remote:remote_execution_test' failed; build aborted: Analysis failed
```

philwo:
I'm trying to import / merge this now. Sorry for the delay!

philwo:
Import successful, tests / review pending 🎉

src/main/java/com/google/devtools/build/lib/syntax/BUILD
```diff
@@ -157,5 +157,5 @@ cc_binary(
         "//conditions:default": ["cpu_profiler_unimpl.cc"],
     }),
     linkshared = 1,
-    deps = ["@bazel_tools//tools/jdk:jni"],
```
philwo, line 160: What's the explanation for this change? 🤔 It looks fine to me, just wondering how it's related to this change.

Qinusty: I mentioned above, when I rebased I wasn't actually able to build Bazel master! Very strange but this fixed the issue... I'd be interested to hear from someone if this was a correct solution to the problem as this target DOES exist within Bazel.

philwo: @Qinusty Sorry, I didn't see that! The `@bazel_tools` reference is a reference to a repository that's built into each Bazel binary. So, the contents of that repository change depending on which Bazel version you're using. I think what happened was that when you tried to build Bazel master and it failed, you were using a Bazel version that didn't have that target (`@bazel_tools//tools/jdk:jni`) yet.

Does it work with the latest Bazel release now? If yes... maybe that was just a temporary hiccup (we only test that you can build Bazel master branch with the latest stable release that was available at the time of the commit). If it still doesn't work without that change, it's pretty weird.

Qinusty: Ah that is likely the case, I'll drop the commit rebuild and push if successful!

Qinusty: Success!

philwo: Great, thanks for checking! I'll revert that part then. :)

philwo:
@Qinusty Things are looking pretty good, I think this can be submitted tomorrow morning. :)

Issue #8438, opened 2019-05-22
toolchain tool_paths should accept labels


guibou:
### Description of the problem / feature request:

toolchain confiuration `cc_common.create_cc_toolchain_config_info` attributes `tool_paths`/`path` should accept bazel labels.

### Feature requests: what underlying problem are you trying to solve with this feature?

The current status (in bazel 0.25) is that toolchain configuration info attributes for tool paths only accept local or absolute paths, but not bazel label.

That's not convenient if your tools comes from another (possibly repository) rule because you need to generate the path by yourself (I don't even know how bazel correctly knows it had to build the dependent rule before).

Example of what I'd like to do:

```
cc_common.create_cc_toolchain_config_info(
     ...
        tool_paths = [
            tool_path(
                name = "gcc",
                path = "@gcc//:bin/gcc",
...
```

### What operating system are you running Bazel on?

Linux

### What's the output of `bazel info release`?

`release bazel-0.25-2`

katre:
This is not a configuration issue, it is part of cc toolchain definition.

guibou:
Note: It should also provide a way to reference directories (for include paths for example).

hlopko:
CC @scentini

guibou:
May I add to this issue that a lot of rule have `string` argument in place which may be `label`. For example, I'm currently limited by `http_archive` `patch_tool` attribute which is a `string` and not a `label`, meaning that I cannot build my local `patch` tool and depend on it.

katre:
You should file that as a separate issue, so that it can be separately triaged and tracked.

guibou:
@katre done. Considering that I've opened this issue 6 months and that nobody seems interested to at least discuss it, I'm wondering if I may not be doing something totally wrong.

If we stay on the topic of setting `create_cc_toolchain_config_info` with path for the compiler, how do people write bazel toolchain using compilers which are them self provided by bazel?

Right now I'm cheating this way:

- My "external repository rules" for compilers are actually generating output in a directory that I manage, outside of bazel. They are also generating `BUILD` files with variables containing theses paths.
- I can then `load` theses `BUILD` files and use the variables as argument for my toolchain `path`.

This is painful because I need to ensure that the first step is reproducible by myself, a job I'd like bazel to do for me.

katre:
[...]

Issue #7746, opened 2019-03-18
Allow cc_toolchain to be defined across multiple repositories


nathaniel-brough:
### Description of the problem / feature request:

Currently the workflow to define a cc_toolchain is a good fit for a monorepo. However it is difficult to have a toolchain configuration span across multiple repos.

**Why would I like to be able to have a toolchain span multiple repos?**
I would like to share a configurable toolchain as an open source project. Therefore I would like to be able to download and use a "CcToolchainConfigInfo" provider rule from an external repository (e.g. using http_archive).

### Feature requests: what underlying problem are you trying to solve with this feature?

To my knowledge this is not currently possible to span a toolchain config across multiple repositories. This is due to the fact that `tool_path` from `"@bazel_tools//tools/cpp:cc_toolchain_config_lib.bzl"` is relative to the package directory. The problem with this is that when you load a toolchain config rule from an external repository the relative paths are relative to where the rule is instantiated rather than where the rule is defined.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
Follow the tutorial on configuring c++ toolchains. Then attempt to load and use cc_toolchain_config from a different workspace. Doing so will result in an error where the given toolpath cannot be found.

### What operating system are you running Bazel on?

Manjaro Linux Rolling Release

### What's the output of `bazel info release`?

release 0.23.0

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

I have found that repository rules are a potential workaround. This works by passing through the attrs from the repository_rule through to a toolchain config rule. This bypasses the problem of relative paths by instantiating the config rule in the given repository. For our particular use case this is a fairly heavy weight solution as it involves creating a new repository for each of our targets. We currently have about 10 different configurations for different cc_toolchain rules, as we target a number of different embedded devices.

### Any other information, logs, or outputs that you want to share?

I think that a potential solution would be to include a new rule that can reference a target rather than a relative path. e.g.

@toolchain_workspace//:BUILD
```python
sh_binary(
    name = "gcc",
    srcs = ["gcc_wrapper.sh"],
)
```

@toolchain_workspace//:cc_toolchain_config.bzl
```python
def _impl(ctx):
    tool_paths = [
        # Create a new toolchain target rule 
        tool_target(
[...]

PR #10967, opened 2020-03-15
Add optional tool_path_origin to Tools in C++ crosstool

Yannic:
This change adds an optional field to the C++ crosstool proto that
allows configuring the origin tool_path is relative to.

For now, the origin can be one of the following:
  - CROSSTOOL_PACKAGE: If tool_path is a relative path, it's assumed to
    be relative to the package of the crosstool top.
  - FILESYSTEM_ROOT: tool_path is an absolute path. This is checked by
    Bazel.
  - WORKSPACE_ROOT: tool_path must be a relative path and is assumed to
    be relative to the exec-root of the workspace (similar to other
    executables).

Updates bazelbuild/bazel#8438


lberki:
@oquenchil , can you take a look?

oquenchil:
Hi Yannic, this https://github.com/bazelbuild/bazel/issues/7746#issuecomment-517326088 from scentini and a few comments above from hlopko is the last context that I can find about how we should go about this problem and how to allow tools from other repositories.

I don't exactly see how your change is going in that direction but perhaps it's because this is all kind of new to me and I might be missing something.

Could you please clarify?

Riatre:
Not original author but this bite me hard so I'll try to explain:

Currently, `tool_path` must either be an absolute path, or a path relative to the package defining cc_toolchain_config, and it must be canonical (cannot contains `../`), thus it is impossible to reference files from other repositories.

For example, say, you have a cc_toolchain_config rule `//tools/toolchain:config`, you want to use a gcc binary from `@gcc_archive//:bin/gcc`, currently it is not possible because `@gcc_archive//:bin/gcc` would be located at `<execroot>/external/gcc_archive/bin/gcc`, and there is no canonical relative path from `<execroot>/tools/toolchain` to `<execroot>/external/gcc_archive_bin/gcc`.

People currently solve this with one-liner wrapper scripts located under `//tools/toolchain`, it exploits the fact that when we execute tools, $PWD is set to exec-root, so we can `exec external/gcc_archive/bin/gcc "$@"` at that time. Even [official tutorial](https://github.com/bazelbuild/bazel/blob/master/site/docs/tutorial/cc-toolchain-config.md) leverages this pattern, though it has other legitimate reasons to wrap the compiler. But this is less than ideal for at least two reasons:

1. User must construct the exec-path manually.
2. It is not obvious how I can refer to an artifact instead of raw file.
3. As per https://github.com/bazelbuild/bazel/issues/8438#issuecomment-594443436, this hack is not compatible with rules_foreign_cc. (I haven't verified this though)

Yannic:
The use-case I had in mind is primarily to remove the need for the checked-in wrapper-scripts @Riatre mentioned. It's also a step in the right direction for tools from external repositories (although, AFAICT, a bit more work is needed there).

The way CC toolchains/crosstools currently work is that, for relative paths, `tool.path` is expected to be relative to the `cc_toolchain` (`cc_toolchain_suite`?) target, which is what's saved in the proto crosstool. This toolchain-relative path is later converted to an exec-path [1] so Bazel can actually run the compile/link/... action. This change allows doing this exec-path convertion earlier (i.e. when generating the crosstool proto), which removes the need for all of these toolchain-relative wrapper scripts.

It's still neccessary to provide all tool artifacts a second time to `cc_toolchain.*_files` (although that is probably not to hard to fix in a follow-up).

Usage example:
```skylark
load("@rules_cc//cc:cc_toolchain_config_lib.bzl", "tool")

def _my_toolchain_impl(ctx):
    # ...

    # Use system compiler
    compiler = tool(path = "/bin/gcc")

    # Use custom linker (new)
    linker = tool(tool = ctx.executable.ld)

    # ...
```

[1] https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CcToolchainFeatures.java;l=950;drc=f5262c54f0609a7c40a57d7dc70e298068b06b0e

Issue #10913, opened 2020-03-06
Native patch doesn't preserve executable bit


thii:
### Description of the problem / feature request:

Native patch doesn't preserve the executable bit of the patched file. If you're trying to patch an executable file in a repository, the file will become a non-executable file.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
.
├── BUILD
├── WORKSPACE
└── rules_apple.patch
```

BUILD is an empty file.

WORKSPACE:

```
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "build_bazel_rules_apple",
    commit = "3bbee3a593d1f6ddb3391ba3a99ffe3f09a4821b",
    patch_args = ["-p1"],
    patches = ["//:rules_apple.patch"],
    remote = "https://github.com/bazelbuild/rules_apple.git",
    shallow_since = "1582926244 -0800",
)
load(
    "@build_bazel_rules_apple//apple:repositories.bzl",
    "apple_rules_dependencies",
)
apple_rules_dependencies()
```

rules_apple.patch (https://github.com/bazelbuild/rules_apple/pull/724.diff):

```diff
diff --git a/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh b/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
index c271bd80..357edb87 100755
--- a/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
+++ b/tools/dynamic_framework_slicer/dynamic_framework_slicer.sh
@@ -79,9 +79,14 @@ else
         fi
     done
 
+    # Create a thin binary if there's only one needed slice, otherwise create a universal binary
     declare -a lipo_args
-    for slice in "${slices_needed[@]}"; do
-        lipo_args+=(-extract $slice)
-    done
+    if [[ "${#slices_needed[@]}" -eq 1 ]]; then
+        lipo_args+=(-thin ${slices_needed[0]})
+    else
+        for slice in "${slices_needed[@]}"; do
+            lipo_args+=(-extract $slice)
+        done
+    fi
     xcrun lipo "$IN" "${lipo_args[@]}" -output "$OUT"
 fi
```

Run:

```
bazel build @build_bazel_rules_apple//tools/dynamic_framework_slicer
```

Observed error:

```
[...]

PR #10970, opened 2020-03-16
Native patch: handling file permission properly

meteorcloudy:
1. If no file permission is specified in the patch file, preserve the
original file permission.

2. If file permission is specified, then set it as it is.

Fixes https://github.com/bazelbuild/bazel/issues/10913

RELNOTES: Native patch can handle file permission properly

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
@@ -239,17 +239,17 @@ public class PatchUtil {
 
   private static boolean getReadPermission(int permission) {
     // Parse read permission from posix file permission notation
-    return permission / 4 % 2 == 1;
+    return (permission & 4) == 4;
   }
 
   private static boolean getWritePermission(int permission) {
     // Parse write permission from posix file permission notation
-    return permission / 2 % 2 == 1;
+    return (permission & 2) == 2;
   }
 
   private static boolean getExecutablePermission(int permission) {
     // Parse executable permission from posix file permission notation
-    return permission % 2 == 1;
+    return (permission & 1) == 1;
   }
 
   private static int getFilePermissionValue(Path file) throws IOException {
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
@@ -156,7 +156,9 @@ public class PatchUtil {
     GIT_HEADER,
     RENAME_FROM,
     RENAME_TO,
-    GIT_LINE,
+    NEW_MODE,
+    NEW_FILE_MODE,
+    OTHER_GIT_LINE,
     UNKNOWN
   }
 
@@ -203,9 +205,15 @@ public class PatchUtil {
         if (line.startsWith("rename to ")) {
           return LineType.RENAME_TO;
         }
+        if (line.startsWith("new mode ")) {
+          return LineType.NEW_MODE;
+        }
+        if (line.startsWith("new file mode ")) {
+          return LineType.NEW_FILE_MODE;
+        }
         for (String prefix : GIT_LINE_PREFIXES) {
           if (line.startsWith(prefix)) {
-            return LineType.GIT_LINE;
+            return LineType.OTHER_GIT_LINE;
           }
         }
       }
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
```diff
@@ -229,8 +237,35 @@ private static void writeFile(Path file, List<String> content) throws IOExceptio
     FileSystemUtils.writeLinesAs(file, StandardCharsets.UTF_8, content);
   }
 
+  private static boolean getReadPermission(int permission) {
+    // Parse read permission from posix file permission notation
+    return permission / 4 % 2 == 1;
```
philwo: We can use bit-operators here to test for the bits in the permission bitmask :)

Executable is `0b001 = 1`
Writable is `0b010 = 2`
Readable is `0b100 = 4`

So we can replace this with:

```java
return permission & 4;
```

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
```diff
@@ -229,8 +237,35 @@ private static void writeFile(Path file, List<String> content) throws IOExceptio
...
+  }
+
+  private static boolean getWritePermission(int permission) {
+    // Parse write permission from posix file permission notation
+    return permission / 2 % 2 == 1;
```
philwo: Similarly:
```java
return permission & 2;
```

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
```diff
@@ -229,8 +237,35 @@ private static void writeFile(Path file, List<String> content) throws IOExceptio
...
+  }
+
+  private static boolean getExecutablePermission(int permission) {
+    // Parse executable permission from posix file permission notation
+    return permission % 2 == 1;
```
philwo: Similarly:
```java
return permission & 1;
```

src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java
```diff
@@ -378,7 +408,7 @@ public void testMissingBothOldAndNewFile() throws IOException {
     Path patchFile =
         scratch.file(
             "/root/patchfile",
-            "diff --git a/foo.cc b/foo.cc",
+            "diff --git a/ b/",
```
philwo: How is this change related?

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
```diff
@@ -245,6 +280,10 @@ private static void applyPatchToFile(
       oldContent = new ArrayList<>();
     } else {
       oldContent = readFile(inputFile);
+      // Preserve old file permission if no explicit permission is set.
+      if (filePermission == -1) {
```
philwo: If you want to use some more exotic Java data types, you could use `Optional<BitSet>` for the filePermission type (and `BitSet` in general, rather than an `int`, to store the permissions). Then you could use an `enum` as the keys for the `BitSet` (e.g. `EXECUTABLE = 0, WRITABLE = 1, READABLE = 2`).

But using `-1` as the indicator for "no explicit permissions" is also fine, if you prefer it. ;)

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
meteorcloudy: Thanks, I'll use 
```java 
return (permission & 4) == 4
```

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
meteorcloudy: Done.

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
meteorcloudy: Done

src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java
meteorcloudy: So this test case is for testing an error message is thrown when no input and output file name is specified. Before, we only parse file names from `---` and `+++` lines. But now we also parses file names from `diff --git` line, therefore we should also remove them here to keep the test passing.

src/main/java/com/google/devtools/build/lib/bazel/repository/PatchUtil.java
meteorcloudy: I tried to use BitSet, but found converting a character (eg. `7`) to a BitSet also requires some steps. The code overall didn't look much cleaner, so I'll keep use `-1` and bit operations ;)

meteorcloudy:
Thanks for the review!!

PR #10976, opened 2020-03-17
Do not use DOS epoch for timestamps of zip entries in Android rules

benjaminRomano:
**Background**
I noticed cache output inconsistencies between my local and CI builds. I traced it down to the zip library in Java includes the timezone info within the extra field section of each zip entry. This appears to be a problem starting with Java 8 (See comment from Gradle's ZipCopyAction source code for more info: https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java#L42-L56)

**Changes**
* Switch to using `February 1, 1980` instead to avoid this edge case

**Test Plan**
Ensure unit tests pass

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
+++ b/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
@@ -59,8 +59,8 @@ public class AndroidResourceOutputs {
     private static final long MINIMUM_TIMESTAMP_INCREMENT = 2000L;
 
     // The earliest date representable in a zip file, 1-1-1980 (the DOS epoch).
-    private static final long ZIP_EPOCH =
-        new GregorianCalendar(1980, Calendar.JANUARY, 01, 0, 0).getTimeInMillis();
+    private static final long DEFAULT_TIMESTAMP =
+        new GregorianCalendar(1980, Calendar.FEBRUARY, 01, 0, 0).getTimeInMillis();
 
     private final ZipOutputStream zip;
 
@@ -84,9 +84,9 @@ public class AndroidResourceOutputs {
      */
     protected long normalizeTime(String filename) {
       if (filename.endsWith(".class")) {
-        return ZIP_EPOCH + MINIMUM_TIMESTAMP_INCREMENT;
+        return DEFAULT_TIMESTAMP + MINIMUM_TIMESTAMP_INCREMENT;
       } else {
-        return ZIP_EPOCH;
+        return DEFAULT_TIMESTAMP;
       }
     }
 
--- a/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
+++ b/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
@@ -58,7 +58,11 @@ public class AndroidResourceOutputs {
     // see http://www.info-zip.org/FAQ.html#limits
     private static final long MINIMUM_TIMESTAMP_INCREMENT = 2000L;
 
-    // The earliest date representable in a zip file, 1-1-1980 (the DOS epoch).
+    /**
+     * Normalized timestamp for zip entries
+     * We use the system's default timezone and locale and additionally avoid using the unix epoch
+     * to ensure Java's zip implementation does not add the System's timezone into the extra field of the zip entry
+     */
     private static final long DEFAULT_TIMESTAMP =
         new GregorianCalendar(1980, Calendar.FEBRUARY, 01, 0, 0).getTimeInMillis();
 
--- a/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
+++ b/src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
@@ -60,7 +60,7 @@ public class AndroidResourceOutputs {
 
     /**
      * Normalized timestamp for zip entries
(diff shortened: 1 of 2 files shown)


aiuto:
I support the concept of the change, but can not comment on the impact in Java rules. We ended up doing something similar for pkg_zip. It also helps with windows portability to not use the unix epoch.

src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
```diff
@@ -59,8 +59,8 @@
     private static final long MINIMUM_TIMESTAMP_INCREMENT = 2000L;
 
     // The earliest date representable in a zip file, 1-1-1980 (the DOS epoch).
```
ahumesky: let's update this comment with the new date, and why it was chosen

ahumesky:
Thanks for taking a look at this! I was wondering why we weren't seeing this internally... turns out we patched ZipEntry internally to go back to the old behavior...

src/tools/android/java/com/google/devtools/build/android/AndroidResourceOutputs.java
jin: (ping @benjaminRomano)

benjaminRomano: Sorry for delay, I updated the comment with some more context.

benjaminRomano:
@jin PTAL

jin:
@ahumesky Is this good for import?

aiuto:
@ahumesky ping

nkoroste:
Fyi, we're using this internally and it significantly increased our cache hit rate

ulfjack:
@jin, just import and send to @ahumesky for review. ;-)

donaldchai:
The idea behind this change sounds good, but I think some more clarification is needed for others who read this code later.

The code linked in ZipCopyAction mentions "January 1st 1980 then it is treated as a special flag" without any concrete pointers.  AFAICT this is just a bug in the JDK; java.util.zip.ZipEntry.DOSTIME_BEFORE_1980 happens to be exactly 1980-01-01.

donaldchai:
Submitted a report to Oracle as 9065240 (not sure when I'll get a URL to link to).  The underlying bug is apparent with this test code.

```
import java.io.File;
import java.io.FileOutputStream;
import java.util.Calendar;
import java.util.GregorianCalendar;
import java.util.TimeZone;
import java.util.zip.*;

public class MakeZips {
  public static void main(String[] args) throws Exception {
    TimeZone.setDefault(TimeZone.getTimeZone("America/Los_Angeles"));

    // java.util.zip.ZipEntry.DOSTIME_BEFORE_1980 is actually 1980-01-01 00:00:00,
    // meaning that the beginning of the DOS epoch causes extra metadata to be written
    makeZip(
        new File("out1a.zip"),
        new GregorianCalendar(1980, Calendar.JANUARY, 01, 0, 0, 0).getTimeInMillis());

    // ZIP files use 2-second precision, so the next possible timestamp is 1980-01-01 00:00:02;
    // this doesn't have extra metadata.
    makeZip(
        new File("out2a.zip"),
        new GregorianCalendar(1980, Calendar.JANUARY, 01, 0, 0, 2).getTimeInMillis());

    TimeZone.setDefault(TimeZone.getTimeZone("GMT"));

    // java.util.zip.ZipEntry.DOSTIME_BEFORE_1980 is actually 1980-01-01 00:00:00,
    // meaning that the beginning of the DOS epoch causes extra metadata to be written
    // --- this file will differ from out1a.zip
    makeZip(
        new File("out1b.zip"),
        new GregorianCalendar(1980, Calendar.JANUARY, 01, 0, 0, 0).getTimeInMillis());

    // ZIP files use 2-second precision, so the next possible timestamp is 1980-01-01 00:00:02;
    // this doesn't have extra metadata.
    // --- this file will be identical to out1a.zip
    makeZip(
        new File("out2b.zip"),
        new GregorianCalendar(1980, Calendar.JANUARY, 01, 0, 0, 2).getTimeInMillis());
  }

  private static void makeZip(File f, long time) throws Exception {
    try (ZipOutputStream out = new ZipOutputStream(new FileOutputStream(f))) {
      ZipEntry e = new ZipEntry("entry.txt");
      e.setTime(time);
      out.putNextEntry(e);
      out.write(new byte[] {0, 1, 2, 3});
      out.closeEntry();
    }
  }
}
```

donaldchai:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8246129

nkoroste:
Is there any harm in merging this into Bazel as it causes major issues with cache hit rate? It's a pretty trivial change so it shouldn't cause any side effect. Plus we've been using it locally since before this PR was created and haven't seen any issues. Also, this same change exists in Gradle.

donaldchai:
There is no harm.  I'm just looking for a comment linking to an official JDK bug, because a healthy software ecosystem requires fixing bugs at the root cause rather than workarounds (with light documentation).  I can add the comment if needed.

FWIW, the bug I filed (https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8246129) is now fixed in the newest JDK going forward.  I wish someone had filed that 5 years ago, since it took <= 2 weeks from bug filing to the fix.

PR #10978, opened 2020-03-17
Use a set to reduce python_path_entries length

### LINKED ISSUE #10977: Excessively long python paths generated

These two functions return some duplicate packages and don't try to simplify the path.

```
python_path_entries = CreatePythonPathEntries(python_imports, module_space)
python_path_entries += GetRepositoriesImports(module_space, %import_all%)
```

in

`$bazel\src\main\java\com\google\devtools\build\lib\bazel\rules\python\python_stub_template.txt`

This long path causes issues when setting up our ROS nodes.

Convert to a set to reduce the path length

```
  # Remove any duplicates to reduce path length
  python_path_entries = set(python_path_entries)
```

Ubuntu 18.04 LTS

```
(23:15:42) WARNING: Option 'remote_default_platform_properties' is deprecated: --remote_default_platform_properties has been deprecated in favor of --remote_default_exec_properties.
release 1.2.0
```

 No

mfourieshieldai:
Resolves #10977

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -275,6 +275,8 @@ def Main():
   python_path_entries = CreatePythonPathEntries(python_imports, module_space)
   python_path_entries += GetRepositoriesImports(module_space, %import_all%)
 
+  # Remove any duplicates to reduce path length
+  python_path_entries = set(python_path_entries)
   python_path_entries = [GetWindowsPathWithUNCPrefix(d) for d in python_path_entries]
 
   old_python_path = os.environ.get('PYTHONPATH')


src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -275,6 +275,8 @@ def Main():
   python_path_entries = CreatePythonPathEntries(python_imports, module_space)
   python_path_entries += GetRepositoriesImports(module_space, %import_all%)
 
+  # Remove any duplicates to reduce path length
```
brandjon: Nit: It's not the individual paths that are being shortened, but the overall `PYTHONPATH` length.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -275,6 +275,8 @@ def Main():
...
+  python_path_entries = set(python_path_entries)
```
brandjon: A set will scramble the order of these entries, possibly non-deterministically. To preserve existing behavior (in most cases) let's keep the existing order and filter out all but the first occurrence of each entry.

PR #10986, opened 2020-03-19
Fix bug in example code in depset docs

bibachhetri:
You need `order = order` in the create function to ensure that the order we pass in is applied. Otherwise, we return a depset using the default order


bibachhetri:
@googlebot I signed it!

PR #10994, opened 2020-03-20
Normalize java_lang_extras.jar

benjaminRomano:
**Background**
Normalize is not passed into singlejar for `java_lang_extras.jar` target leading to non-reproducible output.

**Change**
* Pass in normalize flag

**Test Plan**
* Verify tests pass

### DIFF
--- a/tools/android/BUILD.tools
+++ b/tools/android/BUILD.tools
@@ -151,6 +151,7 @@ genrule(
         --sources  $(location @bazel_tools//tools/jdk:platformclasspath) \
         --include_prefixes "java/lang/invoke/" \
         --include_prefixes "java/lang/annotation/" \
+        -- normalize \
         --output $@""",
     tools = ["@bazel_tools//tools/jdk:singlejar"],
     visibility = ["//visibility:private"],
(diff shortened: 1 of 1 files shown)


tools/android/BUILD.tools
```diff
@@ -151,6 +151,7 @@ genrule(
         --sources  $(location @bazel_tools//tools/jdk:platformclasspath) \
         --include_prefixes "java/lang/invoke/" \
         --include_prefixes "java/lang/annotation/" \
+        -- normalize \
```
jin: space?

benjaminRomano: Fixed sorry about that

jin:
For completeness, here's what `--normalize` does:

> // Scan all entries, verify that *.class entries have timestamp
>  // 01/01/2010 00:00:02 and the rest have the timestamp of 01/01/2010 00:00:00.

https://github.com/bazelbuild/bazel/blob/85292b1a096927de9c5d54b222670e18a3929441/src/tools/singlejar/output_jar_simple_test.cc#L588-L617

PR #11003, opened 2020-03-23
Reduce exception wrapping for cleaner stack traces

ulfjack:
The StatusRuntimeException often already wraps an IOException, and provides no
value on top of that (because it's created internally in gRPC and has no
relationship to Bazel code). If so, just throw the wrapped IOException directly
to reduce the output spam (if the exception is printed).

Change-Id: Ib2bbad230cf632c491aea0fb24a7721e96178499


buchgr:
LGTM

Issue #8857, opened 2019-07-10
move rules_pkg out of Bazel


aiuto:
//tools/build_defs/pkg:{pkg_deb, pkg_rpm, pkg_tar}  are being pulled out of Bazel core and moved to https://github.com/bazelbuild/rules_pkg/tree/master/pkg. Users should migrate to the new versions.

Migration:

1. Add WORKSPACE stanza as specified in a rule_pkg release 
 https://github.com/bazelbuild/rules_pkg/releases
1. Replace instances of
   ```
   load("//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")
   load("//tools/build_defs/pkg:rpm.bzl", "pkg_rpm")
   ```
   with
   ```
   load("@rules_pkg//:pkg.bzl", "pkg_tar", "pkg_deb")
   load("@rules_pkg//:rpm.bzl", "pkg_rpm")
   ```

    One line solution (for some users)
    ```
   find . -name 'BUILD' -o -name '*.bzl' | xargs sed -i -e '/^load/s#//tools/build_defs/pkg:#@rules_pkg//:#'
    ```

Expected timeline:
- The bazelbuild/rules_pkg repository is available for use now. Issues against packaging rules are being transferred to that repository and will be addressed there.
- Flag to be released in Bazel 0.29.
- Enabled by default in next Bazel incompatible flag flip release after 0.29
- //tools/build_defs/pkg/... removed from Bazel by EOY 2019

dslomov:
Will the migration tooling be available to do this change?

aiuto:
> Will the migration tooling be available to do this change?

I could do special purpose tooling with a small python script. OR.. I could wait for a migration tool that can update WORKSPACE IFF any of the BUILD or bzl files were transformed.

The other thing we can do is not bother flipping now and get people to migrate by only adding new features to the new repo.

hlopko:
Why did you decide to implement the  incompatible flag as a build_setting? Our whole infrastructure relies on incompatible flags being native options. The most notable problems are that our buildkite pipeline showing the progress of the migration will not include this flag, and bazelisk --strict will not flip this flag so users relying on bazelisk won't notice this incompatible change at all.

I mean it's fine to improve the infrastructure to support build_settings in the future, and start using build_settings after that, but not the other way around. I'm afraid you'll have to either manually collect the migration progress data or replace the build_setting with native option.

aiuto:
[...]

PR #11025, opened 2020-03-26
make_deb: fix length computation for extrafiles

matzf:
When adding the `extrafiles` (e.g. `preinst`, `postinst` etc.) with non-ascii characters in the content to the tar, a potentially faulty length was used. This would result in truncated files.
This seems to only occur when using files that are the result of build rules, i.e. the problem does not seem to occur when using a plain file as input for e.g. `postinst`.

This seems to have been be an oversight in 70e3c339bea02f6e9273b13de11d078ca6c2752f, where the same problem was fixed for the main `control` file.
This change makes the handling of the length of the `extrafiles` identical to the main `control` file.

### DIFF
--- a/tools/build_defs/pkg/make_deb.py
+++ b/tools/build_defs/pkg/make_deb.py
@@ -179,9 +179,10 @@ def CreateDebControl(extrafiles=None, **kwargs):
       if extrafiles:
         for name, (data, mode) in extrafiles.items():
           tarinfo = tarfile.TarInfo(name)
-          tarinfo.size = len(data)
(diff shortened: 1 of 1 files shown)


matzf:
@googlebot I signed it!

aiuto:
We are not maintaining make_dep at this location any more. The code here is scheduled to be removed (#8857)

Bazel now uses this for our builds: https://github.com/bazelbuild/rules_pkg/blob/master/pkg/make_deb.py
I can look at the PR there. Please also include a test case with it.

matzf:
See https://github.com/bazelbuild/rules_pkg/pull/144

PR #11029, opened 2020-03-26
make --experimental_repo_remote_exec compatible with rbe_autconfig

buchgr:
rbe_autoconfig has an attribute 'exec_properties' of type string_dict [1].
As do all repository rules if --experimental_repo_remote_exec is set.
Bazel reports an error if both are used together. This change fixes this
so if a repository_rule declares an attribute 'exec_properties' of type
string_dict Bazel no longer tries to add its own.

[1] https://github.com/bazelbuild/bazel-toolchains/blob/b5928f980c79cbdbfd0b5c5a924aeb2cd61eb4f9/rules/rbe_repo.bzl#L758

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
@@ -37,6 +37,7 @@ import com.google.devtools.build.lib.analysis.constraints.EnvironmentRule;
 import com.google.devtools.build.lib.analysis.test.TestConfiguration;
 import com.google.devtools.build.lib.cmdline.Label;
 import com.google.devtools.build.lib.packages.Attribute;
+import com.google.devtools.build.lib.packages.Attribute.Builder;
 import com.google.devtools.build.lib.packages.Attribute.LabelLateBoundDefault;
 import com.google.devtools.build.lib.packages.Attribute.LabelListLateBoundDefault;
 import com.google.devtools.build.lib.packages.Attribute.LateBoundDefault.Resolver;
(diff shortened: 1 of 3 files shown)


buchgr:
FYI @GMNGeoffrey

GMNGeoffrey:
Yay, thank you!

src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
```diff
@@ -345,10 +347,19 @@ public Metadata getMetadata() {
...
+   * Adds an {@code exec_properties} attribute of type {@code STRING_DICT} if the rule does not
+   * already have one.
+   */
+  public static void addOrOverrideExecPropertiesAttribute(RuleClass.Builder builder)
       throws ConversionException {
-    return builder.add(
-        attr(RuleClass.EXEC_PROPERTIES, STRING_DICT).defaultValue(ImmutableMap.of()));
+    Builder<Map<String, String>> attr =
+        attr(RuleClass.EXEC_PROPERTIES, STRING_DICT).defaultValue(ImmutableMap.of());
+    if (builder.contains(RuleClass.EXEC_PROPERTIES)) {
```
meteorcloudy, line 358: What if the existing EXEC_PROPERTIES doesn't have the type string_dict? Should we throw an error here? Otherwise, it could be confusing to user.

GMNGeoffrey:
Any progress on getting this merged? We're still using a hacky local patch and I would love to fix that and close that out.

GMNGeoffrey:
Ping on this? I'm in the process of updating our bazel version and am going to have to update our hacky patch as well

meteorcloudy:
@buchgr Can you resolve the comment and conflict of this PR?

buchgr:
Forgot about it. I ll look into this today. Thanks for the ping.

agluszak:
Fixed by https://github.com/bazelbuild/bazel-toolchains/pull/912

PR #11030, opened 2020-03-26
Implement mutual TLS authentication

### LINKED ISSUE #10735: FR: mutual tls auth for remote cache/executor

I was under impression that [--tls_certificate](https://docs.bazel.build/versions/2.0.0/command-line-reference.html#flag--tls_certificate) can be used for mutual tls, but in code I see that it only configures trusted CA certs: [GoogleAuthUtils.java#L117](https://github.com/bazelbuild/bazel/blob/2.1.0/src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java#L117)

In code it only uses `trustManager`, while mutual tls requires `keyManager` to set up client cert and private key. Example https://github.com/grpc/grpc-java/issues/4004#issuecomment-360576997

I need to use remote cache grpcs with mutual tls auth

Debian 10

release 2.1.0

ulfjack:
Add a pair of flags tls_client_certificate / tls_client_key to specify a
certificate and corresponding key, which allow Bazel to authenticate itself
over TLS to a remote cache or remote executor.

Before this change, Bazel only supports Google Cloud authentication, which
requires an open network connection to Google Cloud on the client as well as
on the server.

I have heard from one user that they are tunneling their traffic over a VPN
and then perform no client authentication in their remote caching system. I
heard from another user that they have locally patched Bazel to enable mTLS
(but have not upstreamed the patch). Finally, there is also a pending
feature request.

Compared to other authentication mechanisms, mTLS is already supported by
gRPC out of the box.

I added test coverage by also adding a --tls_ca_certificate to the local
remote worker, and updating the existing TLS test to also run with an mTLS
configuration.

I had to generate a new ca cert/key pair in order to sign a new client
certificate (and also re-generate the server cert/key pair); these are
checked in as testdata. Note that the generator script given there already
supports generating both server and client cert/key pairs, so no change to
the documentation was necessary.

Fixes #10735.

Change-Id: I8c9fdab11d172a4cc8a2b80de43faa48086cc893


buchgr:
looks good!

ulfjack:
@philwo @meisterT I'd appreciate if someone could merge this change. :-)

meisterT:
Jakob already started the import but it hangs in some weird copybara error.

buchgr:
it complains about checked in private keys. Can you disguise them somehow?

aiuto:
cc @mikelalcon
Can you publish the error message, that may help determine the best way to disguise them. I did a cursory look at the copybara code and do not see a scanner looking for keys directly. If it is only checking by file name, it would be easy to fix by renaming. If not, then uglier tricks are needed.

meisterT:
The error message doesn't come from copybara but from git.

It's this one here:
https://cloud.google.com/source-repositories/docs/detecting-security-keys#how_the_security-key_checking_feature_works

ulfjack:
Thanks!

PR #11034, opened 2020-03-27
Fix horizontal scrollbar on desktop docs

jin:
docs.bazel.build on desktop now shows a horizontal scrollbar. Let's not change the bootstrap default css for `.container`, but modify our child div's margin to override the parent's instead.

### DIFF
--- a/site/_sass/sidebar.scss
+++ b/site/_sass/sidebar.scss
@@ -8,6 +8,9 @@ $sidebar-hover-border-color: #66bb6a;
   padding-top: 40px;
   padding-left: 1.5rem;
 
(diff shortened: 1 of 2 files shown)


jin:
Sorry for not having enough details. 

> FWIW, in the current live page, I only se a scroll bar when I do not need it. That is, when my window is wide enough to show left nav, content, AND right nav. Then chrome is adding a horizontal scroll, even though it is not visually needed. If I narrow the window, the left and right panes disappear and the scrollbar does too. Instead, the text flows to my window width (yeah!).

This PR fixes exactly that -- the page is erroneously too wide, and shouldn't have a horizontal scrollbar at all widths.

Issue #11023, opened 2020-03-26
Bazel CI: Buildtools is broken on Windows with Bazel@HEAD


meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1435#cb1c36f4-8ab9-4375-acc0-8a36a3b8558a
```
//extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest      FAILED in 3 out of 3 in 36.8s
```
Error message:
```
Executing tests from //extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest
-----------------------------------------------------------------------------
/usr/bin/bash: warning: shell level (1000) too high, resetting to 1
LAUNCHER ERROR: Cannot launch process: "C:/tools/msys64/usr/bin/bash.exe
Reason: (error: 8): Not enough memory resources are available to process this command.
```
Culprit: d899a036bd043fcde7db8a7d568c102f50411c79

From the error message, we can see there is probably a dead loop of calling the bash repeatedly.
Here is how it happened:
1. In d899a036bd043fcde7db8a7d568c102f50411c79, we made Bazel find the test binary under the runfiles tree when runfiles tree is enabled. 
2. However, for `//extra_actions_base_proto:extra_actions_base.gen.pb.go_checkshtest`, the path length of the test binary under runfiles tree is longer than 260. 
3. To deal with that, our windows test wrapper (tw.cc) converts it into a short Windows path (8dot3 style short path) before launching the test via CreateProcessW. 
4. However, the Windows bash launcher (bash_launcher.cc) relies on the first argument of the command to find the corresponding bash script by removing the `.exe` suffix. Eg, `C:/fooooo/baaaaar/bin.exe` to `C:/fooooo/baaaaar/bin`
5. Because the name is shortened, eg. `C:/f~1/b~1/b~1.exe`. Instead find the shell script(`C:/fooooo/baaaaar/bin`), the bash launcher will find `C:/f~1/b~1/b~1`, which Bash will interpret it as `C:/f~1/b~1/b~1.exe` when `C:/f~1/b~1/b~1` doesn't exist!
6. Therefore, the native launcher ends up calling itself and eventually ran out of memory.

To fix:
d899a036bd043fcde7db8a7d568c102f50411c79 actually exposed a bug in our Windows native launcher, that when it is called via a shortened Windows path (8dot3 style short path), it cannot launch the binary properly. We should fix this issue.

[...]

PR #11047, opened 2020-03-31
Windows native launcher: support launching with 8dot3 Windows path

meteorcloudy:
Windows native launcher finds the shell/python script to launch based on the first command line argument (the path name of itself). But it could be a shortened 8dot3 windows path. Eg. when `bazel test` runs a test, if the test binary path is too long (> 260), Bazel will convert it to a short path. The native launcher should convert it back to its full path before using it to find the script to launch.

Fixes https://github.com/bazelbuild/bazel/issues/11023


meteorcloudy:
Thank you!

PR #11048, opened 2020-03-31
Fix grammar mistake, add minor clarification

rytswd:
I noticed some minor grammar mistakes while reading up on the doc.
Also, suggested a minor addition to better clarify the mistake with indirect dependency.


dslomov:
I'll merge

PR #11055, opened 2020-04-02
Improve and expand macro documentation.

jin:
Addressing [user feedback](https://twitter.com/AutomatedTester/status/1245695247801372672).

### DIFF
--- a/site/docs/skylark/macros.md
+++ b/site/docs/skylark/macros.md
@@ -10,15 +10,15 @@ title: Macros
 ## Introduction
 
 A macro is a function called from the BUILD file that can instantiate rules.
-Macros are mainly used for encapsulation and code reuse of existing rules 
-and other macros. By the end of the [loading phase](concepts.md#evaluation-model), 
+Macros are mainly used for encapsulation and code reuse of existing rules
+and other macros. By the end of the [loading phase](concepts.md#evaluation-model),
 macros don't exist anymore, and Bazel sees only the concrete set of instantiated rules.
 
 ## Usage
 
-The typical use-case for a macro is when you want to reuse a rule. 
+The typical use-case for a macro is when you want to reuse a rule.
 
-For example, we have a genrule in a BUILD file that generates a file 
+For example, we have a genrule in a BUILD file that generates a file
 using `//test:generator` with a `some_arg` argument hardcoded in the command:
 
 ```python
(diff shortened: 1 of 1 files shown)


site/docs/skylark/macros.md
```diff
@@ -7,12 +7,131 @@ title: Macros
...
+  )
+```
+
+You can also use macros to chain rules together. This example shows a generated
+text file undergoing conversions to markdown and rich text format:
+
+```python
+def txt_md_rtf_generator(name, visibility=None):
+  # Generates a txt file
+  native.genrule(
```
aiuto, line 87: Why bother with this. How about taking the text file name as an arg to the macro. That makes the example shorter without diminishing the point you are making.

site/docs/skylark/macros.md
```diff
@@ -7,12 +7,131 @@ title: Macros
...
+  )
+  
+  # Converts the txt to markdown
+  native.genrule(
+    name = name + "-md",
+    srcs = [name + ".txt"],
+    outs = [name + ".md"],
+    cmd = "$(location //test:txt-to-md) $< > $@",
+    tools = ["//test:txt-to-md"],
+    visibility = visibility,
```
aiuto, line 102: You could illustrate another point here. What if the macro was  txt_to_rdf, and did not mention md.
You could give private visibility here to emphasize that the macro author only want the end result visible.

site/docs/skylark/macros.md
```diff
@@ -7,12 +7,131 @@ title: Macros
...
+file_generator(
+    name = "file-three",
+    arg = "some_arg_three",
+)
+```
+
+Here, we are loading the `file_generator` symbol from a `.bzl` file located in the 
+`//path` package. By putting macro function definitions in a separate `.bzl` file, 
+we can keep our BUILD files clean and declarative,  The `.bzl` file can be loaded 
+from any package in the workspace, as long as `visibility` conditions are met.
```
laurentlb, line 63: No, visibility doesn't apply.

site/docs/skylark/macros.md
```diff
@@ -7,12 +7,131 @@ title: Macros
...
+
+> Tip: To make a `.bzl` file visible from a package, add [`exports_files(["<filename>.bzl"])`](../be/functions.html#exports_files) into the `BUILD` file in the same directory as the `.bzl` file.
```
laurentlb, line 65: exports_files is not recommended

aiuto: Are you recommending dropping the paragraph or replacing it with a pointer to bzl_library?

jin: Not recommended at all? Also, it seems that we don't need it -- just an empty BUILD file will do.

jin: Thanks. Removed.

site/docs/skylark/macros.md
jin: Removed.

site/docs/skylark/macros.md
jin: The name is the arg to the macro. I simplified the example.

site/docs/skylark/macros.md
jin: Done.

PR #11080, opened 2020-04-06
Use @bazel_tools//tools/jdk:jni.

benjaminp:
This eliminates annoying analysis warnings from the cc rules.


jin:
cc @alandonovan

Issue #10874, opened 2020-02-29
Documentation issue: External Dependencies


dmadisetti:
Documentation URL: https://docs.bazel.build/versions/2.0.0/external.html#best-practices

Is using `http_archive` still best practice over using `git_repository`? The justification given is rather dated, since then shallow_since has been added (https://github.com/bazelbuild/bazel/commit/6496b1d1d25025f33406b1eaf9949cab126f08bd) in addition `git_repository` may have diverged from the original skylark implementation.

If there is a good justification maybe we can link this answer :)

jin:
Solely based on the Starlark git rules' dependency on system `git(1)`, I would still recommend using `http_archive` over `git_repository` whenever possible.

laurentlb:
* I think `http_repository` is faster for GitHub repositories (need to double-check).
* It's easier to provide mirrors.
* As jin said, `git_repository` depends on the git binary. It's not always installed (for example, on many Windows machines).

For using projects on GitHub, `http_repository` seems better to me.

Issue #5116, opened 2018-04-30
Should git_repository use repository cache?


ittaiz:
### Description of the problem / feature request:

have `new_git_repository` rules use the `repository_cache`

### Feature requests: what underlying problem are you trying to solve with this feature?

I'd like to be able to reuse external git dependencies like one can reuse new_http_archive

aehlig:
I think, in general it is a good idea, especially as git basically is a content-addressable storage, so even  individual shared blobs could be avoided being downloaded again.

Note that we're moving from the native `git_repository` rules to the Skylark version. So, assuming you're given a directory that you can use as cache, what would be the appropriate `git` commands to make best use of it?

aehlig:
As mentioned in #5086,  one possible way forward might be telling the Skylark `git_repository` rule (e.g., via an environment variable) about the location of a git repository on the local machine that might be used as a cache. But still, I'm not sure what the best option is.

aoded:
Hi @aehlig, we would appreciate some help with this :)

We tried to work on this feature, and while modifying this method -  https://github.com/bazelbuild/bazel/blob/ca2733c3ec4d0d5ebfe6b3c4b8dcee3c8855cf6b/tools/build_defs/repo/git.bzl#L18 we wanted to get the path to the cache folder either from environment or command line arguments.
We tried using ctx.os.environ, but adding a new environment variable didn't affect its value
We tired looking for a way to get a command line parameter value via ctx and found no such way

Thanks!
Oded

aehlig:
> we wanted to get the path to the cache folder either from environment or command line arguments. We tried using ctx.os.environ, but adding a new environment variable didn't affect its value

`ctx.os.environ` does refelect the invocation environment; however, a repository is only invalidated based on the declared environment variables&mdash;but you can always use `bazel sync`. (Note that the cache location should _not_ invalidate the repository, as its content do not depend on the location of the cache.)

sluongng:
It's worth to note that https://github.com/bazelbuild/bazel/pull/7424 has almost finished an implementation of git_repository cache but was abandoned as the author has no capacity to support it. Whoever want to pick up this issue should consider using it as reference.

I just want to note that in upstream Git core, Microsoft's Scalar is being integrated into `contrib` dir and is planned to eventually integrate to be part of Git itself. If you are not familiar with Scalar, it's a high level 'git repository manager' command where it helps:
1. Set a sane default git configs
2. Help cache all git repositories it manage
[...]

PR #11088, opened 2020-04-07
Add rationale for preferring http_archive over git_repository.

jin:
Fixes https://github.com/bazelbuild/bazel/issues/10874

### DIFF
--- a/site/docs/external.md
+++ b/site/docs/external.md
@@ -307,7 +307,10 @@ bootstrap test.
 ### Repository rules
 
 Prefer [`http_archive`](repo/http.html#http_archive) to `git_repository` and
-`new_git_repository`.
+`new_git_repository`. Rationale: Git repository rules depend on system `git(1)` whereas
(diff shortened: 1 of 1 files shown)


site/docs/external.md
```diff
@@ -307,7 +307,10 @@ bootstrap test.
 ### Repository rules
 
 Prefer [`http_archive`](repo/http.html#http_archive) to `git_repository` and
-`new_git_repository`.
+`new_git_repository`. Rationale: Git repository rules depend on system `git(1)` whereas
+the HTTP downloader is built into Bazel and has no system dependencies. `http_archive` 
+also supports a list of `urls` as mirrors, and `git_repository` supports only a single
+`remote`.
```
kastiglione, line 313: Another reason to use `http_archive` is that it makes use of the repository cache: https://github.com/bazelbuild/bazel/issues/5116

jin: Thanks! Added.

PR #11092, opened 2020-04-08
Add rel="noopener" to links with target="_blank" attribute

jin:
See https://web.dev/external-anchors-use-rel-noopener/

Also remove one instance of `target="_blank"` on the link to stardoc, since none of the other links in the page uses `_blank`.

### DIFF
--- a/site/_layouts/documentation.html
+++ b/site/_layouts/documentation.html
@@ -331,7 +331,7 @@ nav: docs
                 <ul class="collapse sidebar-nav sidebar-submenu" id="starlark-practices">
                   <li><a href="/versions/{{ current_version }}/skylark/bzl-style.html">.bzl style guide</a></li>
                   <li><a href="/versions/{{ current_version }}/skylark/testing.html">Testing extensions</a></li>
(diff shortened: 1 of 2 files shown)


site/_layouts/documentation.html
```diff
@@ -331,7 +331,7 @@ <h3>Extending Bazel</h3>
                 <ul class="collapse sidebar-nav sidebar-submenu" id="starlark-practices">
                   <li><a href="/versions/{{ current_version }}/skylark/bzl-style.html">.bzl style guide</a></li>
                   <li><a href="/versions/{{ current_version }}/skylark/testing.html">Testing extensions</a></li>
-                  <li><a href="https://github.com/bazelbuild/stardoc" target="_blank">Documenting rules with Stardoc</a></li>
+                  <li><a href="https://github.com/bazelbuild/stardoc">Documenting rules with Stardoc</a></li>
```
aiuto, line 334: Did you want rel=noopener here too?

jin: No, I removed the `target="_blank"` here, since none of the other links on the page uses it too.

PR #11093, opened 2020-04-08
Add //site:site_latest to presubmit build

jin:
This adds a smoke test to build the site on presubmits to ensure that the documentation generator works e2e.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -53,6 +53,7 @@ tasks:
       - "//src:bazel"
       - "//src:bazel_jdk_minimal"
       - "//src:test_repos"
+      - "//site:site_latest"
     test_flags:
       - "--sandbox_default_allow_network=false"
       - "--sandbox_writable_path=$HOME/bazeltest"
@@ -88,6 +89,7 @@ tasks:
       - "//src:bazel"
       - "//src:bazel_jdk_minimal"
       - "//src:test_repos"
+      - "//site:site_latest"
     test_flags:
       - "--sandbox_default_allow_network=false"
       - "--sandbox_writable_path=$HOME/bazeltest"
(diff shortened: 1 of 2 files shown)


jin:
@philwo / @gregestren what do you think about building a minimal site on presubmit to increase visibility into site build errors?

Looks like we need Jekyll and gems in https://github.com/bazelbuild/bazel/blob/master/scripts/docs/Gemfile on the CI machines.

gregestren:
Absolutely. Doc generation is a legit part of the pipeline. There's no reason to make it second class w.r.t. test coverage.

How long does this test take?

jin:
Using a 4-thread Mac, the ~upper bound for a clean sandboxed build took:

```
  bazel-bin/src/main/java/com/google/devtools/build/lib/build-encyclopedia.zip
  bazel-bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html
  bazel-bin/src/main/java/com/google/devtools/build/lib/skylark-library.zip
INFO: Elapsed time: 503.555s, Critical Path: 160.92s
INFO: 951 processes: 643 darwin-sandbox, 308 worker.
INFO: Build completed successfully, 971 total actions
```

Incrementally, running jekyll on the above inputs to generate the site took:

```
Target //site:site_latest up-to-date:
  bazel-bin/site/site_latest.tar
  bazel-bin/site/site_latest
INFO: Elapsed time: 31.972s, Critical Path: 29.60s
INFO: 28 processes: 26 darwin-sandbox, 2 worker.
INFO: Build completed successfully, 54 total actions
```

aiuto:
I think we should get doc-generation out of the bazelbuild/bazel pipeline
and move it into the bazelbuild/bazel-website.
Doing that would make most docs github first, and could make changes easier
to make with lower latency.
We would still want tests and a CI for that pipeline. It's just a
question of adding them before the move or after.

jin:
@aiuto Yeah, in fact we only really need the CLI reference, BE and Skylark generated zip files from this repo. The rest of the documentation can live elsewhere.

aiuto:
#11462   We should get all the jeykel out of the main bazel repo.
The presubmit here should be to see only if auto-doc generation works
Then the postsubmit should auto-generate and push to the storage bucket for the stream on bazel-website to pick it up.

meisterT:
What's the state of this PR?

Issue #11095, opened 2020-04-08
//src/main/java/com/google/devtools/build/lib:gen_command-line-reference is broken


jin:
Bisected to 09924273efc7050a2e342cd89b823a619ca073f8 (cc @anakanemison)

```
$ bazel build //src/main/java/com/google/devtools/build/lib:gen_command-line-reference --verbose_failures
...

INFO: Analyzed target //src/main/java/com/google/devtools/build/lib:gen_command-line-reference (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /Users/jingwen/bazels/github/src/main/java/com/google/devtools/build/lib/BUILD:756:1: Executing genrule //src/main/java/com/google/devtools/build/lib:gen_command-line-reference failed (Exit 37): bash failed: error executing command

...

  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cat site/command-line-reference-prefix.html > bazel-out/darwin-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html && TMP=`mktemp -d /tmp/tmp.XXXXXXXXXX` && bazel-out/host/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer --jvm_flag=-Dio.bazel.EnableJni=0 --batch --install_base=${TMP} --output_base=${TMP}/output/ --output_user_root=${TMP} help everything-as-html >> bazel-out/darwin-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html 2>/dev/null && cat site/command-line-reference-suffix.html >> bazel-out/darwin-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html')
Execution platform: //:default_host_platform

Use --sandbox_debug to see verbose messages from the sandbox bash failed: error executing command
[...]

PR #11096, opened 2020-04-08
Fix //src/main/java/com/google/devtools/build/lib:gen_command-line-reference and add tests.

jin:
09924273efc7050a2e342cd89b823a619ca073f8 adds a new required --failure_detail_out flag, which broke the docgen genrule because its invoking the Bazel server directly without the flag.

We should revisit this direct usage of the server, and look at using the Bazel launcher directly. https://github.com/bazelbuild/bazel/issues/11095#issuecomment-611063162

For now, manually add the flag to get our documentation pipeline working again.

Fixes https://github.com/bazelbuild/bazel/issues/11095


jin:
cc @anakanemison (can't add you as a reviewer for some reason, maybe you're not in the bazelbuild GitHub org?

src/test/java/com/google/devtools/build/docgen/generated_docs_build_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# A smoke file test for generated doc files.
+
+set -euo pipefail
+
+# --- begin runfiles.bash initialization v2 ---
+# Copy-pasted from the Bazel Bash runfiles library v2.
+set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
```
aiuto, line 9: This is obviously unneeded, but it also makes sense to copy over a tricky preamble like this without changes. So I'm calling it out as OK, in spite of it looking funny.

src/test/java/com/google/devtools/build/docgen/generated_docs_build_test.sh
```diff
@@ -0,0 +1,28 @@
...
+
+for out in $EXPECTED_ZIPS; do
+    filepath="$(rlocation io_bazel/src/main/java/com/google/devtools/build/lib/$out)"
+    # Test that the file has a corrected ZIP structure.
+    zipinfo "$filepath" 1>/dev/null
+done
+
+COMMAND_LINE_REF_FILE="$(rlocation io_bazel/src/main/java/com/google/devtools/build/lib/command-line-reference.html)"
+# Test that the file contains (at least) the title string.
+grep "Command-Line Reference" "$COMMAND_LINE_REF_FILE" 1>/dev/null
```
aiuto, line 28: grep -q  ....
will just set an exit status without doing any output.

jin: Nice, done.

PR #11104, opened 2020-04-08
proto_lang_toolchain: Add test to verify that blacklisted_protos contains transitive (original) source

Yannic:
This adds a test for 01df1e5c3aaf709e9388293514ae978e59ae2cbb, which
didn't come with a (OSS) test.


Yannic:
@cushon ping

Issue #11119, opened 2020-04-13
Allow remote-caching for Symlink and SolibSymlink actions


sohaibiftikhar:
### Description of the problem / feature request:
Bazel does not cache Symlink and SolibSymlink actions. Since the remote-cache API allows caching of symlinks this should be possible to implement.

### Feature requests: what underlying problem are you trying to solve with this feature?
When used on a clean output base but with the entire build/test cached remotely this leads to unnecessary downloads or action outputs even when the actual build/test action is already cached. Coupled with the `--remote_download_minimal` this leads to repeated downloads for a fully cached build without even shutting down the bezel server. For larger binaries/libraries this leads to significant time wastage.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
1. Create a file main.cpp
```
#include <stdio.h>
#include <thread>
#include <chrono>

template<int I> void f() { printf("%d\n", I); }

template<int N, int I>
struct C {
  void operator()() {
    f<I>();
    C<N/2, I+N>()();
    C<N/2, I-N>()();
  }
};

template<int I>
struct C<0, I> {
  void operator()() {}
};

int main() {
  // This is just to create a sufficiently sized binary
  C<65535, 0>()();
  printf("Sleeping\n");
  std::this_thread::sleep_for(std::chrono::seconds(5));
  printf("done\n");
}
```
And create a build file 
```
load("@rules_cc//cc:defs.bzl", "cc_binary")

cc_binary(
    name = "big",
    srcs = ["main.cpp"]
)

sh_test(
    name = "big_test",
    size = "small",
    srcs = ["big"],
)
```
Now run the command:
```
bazel test --remote_cache=grpc://mycache:8980" --remote_download_minimal  //:big_test
```
Everything should be sent to the remote cache now. Now clean your output base to simulate another system running the same build. Then re run the build.
```
bazel clean && bazel shutdown
bazel test --remote_cache="--remote_cache=grpc//mycache:8980" --remote_download_minimal  //:big_test
```
Bazel will download the remote binaries again on repeated invocations (remote_download_minimal ensures that we don't even need a clean anymore).

### What operating system are you running Bazel on?
Linux

### What's the output of `bazel info release`?
3.0.0

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

NA

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

NA

[...]

PR #11120, opened 2020-04-13
Allow remote caching for Symlink Actions

sohaibiftikhar:
Attempts fixing #11119

The change was made for 1.2.1 and adapted for master.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkHelper.java
@@ -1,0 +1,95 @@
+// Copyright 2014 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkHelper.java
```diff
@@ -0,0 +1,95 @@
...
+      String relativeLinkName,
+      String absoluteTarget)
+      throws ActionExecutionException {
+    try {
+      CommandLine symlinkCommandLine = getSymlinkCommandLine(relativeLinkName, absoluteTarget);
+      return new SimpleSpawn(
+          action,
+          ImmutableList.copyOf(symlinkCommandLine.arguments()),
+          actionExecutionContext.getClientEnv(),
+          ImmutableMap.of(ExecutionRequirements.NO_REMOTE_EXEC, ""),
```
sohaibiftikhar, line 67: This can be removed once we allow [`remote_download_minimal`](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java#L556) to also work for symlink outputs.

sohaibiftikhar: Or maybe it's better to extend the --remote_download_minimal to expand to 
```
--strategy=ExecutableSymlink=sandboxed,local
--strategy=Symlink=sandboxed,local
--strategy=SolibSymlink=sandboxed,local
```
Open to suggestions.

ulfjack:
I posted some questions on the bug. I am not sure caching Symlink and SolibSymlink actions is the right solution.

meisterT:
What's the state of this PR?

coeuvre:
@sohaibiftikhar Since we are looking for another way to solve this problem, can we close this for now?

sohaibiftikhar:
Closing after discussion on #11119

PR #11127, opened 2020-04-14
Update install-windows.md

Smokrow:
Added a little helper text to highlight the use of Msys64 with CMD or Powershell

### DIFF
--- a/site/docs/install-windows.md
+++ b/site/docs/install-windows.md
@@ -95,6 +95,7 @@ We recommend installing:
     ```
     pacman -S zip unzip patch diffutils git
     ```
+    Make sure to add the `MSYS2_INSTALL_PATH/usr/bin` to your Path Environmentvariable in case you want to use CMD or Powershell. 
 
 *   [Build Tools for Visual Studio 2019](https://aka.ms/buildtools)
 
--- a/site/docs/install-windows.md
+++ b/site/docs/install-windows.md
@@ -95,7 +95,7 @@ We recommend installing:
     ```
     pacman -S zip unzip patch diffutils git
     ```
-    Optional: If you want to use Bazel from CMD or Powershell make sure to add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH` Environmentvariable. 
+    Optional: If you want to use Bazel from CMD or Powershell and still be able to use Bash tools, make sure to add MSYS2_INSTALL_PATH/usr/bin to your PATH Environmentvariable.
 
 *   [Build Tools for Visual Studio 2019](https://aka.ms/buildtools)
 
--- a/site/docs/install-windows.md
+++ b/site/docs/install-windows.md
@@ -88,7 +88,7 @@ We recommend installing:
 *   Common MSYS2 packages
 
     You will likely need these to build and run targets that depend on Bash.  MSYS2 does not install
-    these tools by default, so you need to install them manually. Projects that that depend heavily on these tools are for example Tensorflow.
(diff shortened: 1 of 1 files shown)


aiuto:
Is there anyone who can validate this? I don't have access to a windows machine while WFH?

Smokrow:
I basically copied it from the Tensorflow Documentation:
https://www.tensorflow.org/install/source_windows#install_msys2

site/docs/install-windows.md
```diff
@@ -95,6 +95,7 @@ We recommend installing:
     ```
     pacman -S zip unzip patch diffutils git
     ```
+    Make sure to add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH` Environmentvariable in case you want to use CMD or Powershell.
```
jin, line 98: "Optional: If you want to use these tools from CMD or Powershell, add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH` Environment Variable."

Bazel already has access to these binaries if you set `BAZEL_SH` to `C:\msys64\usr\bin\bash.exe`, so let's mark this step as optional.

@meteorcloudy please verify :)

meteorcloudy: @jin is absolutely correct. MSYS is no longer a mandatory requirement for Bazel on Windows. TensorFlow needs it because they use a lot unix tools in their build. But that might not be true for other projects. So if we want to add this, it should be optional.

aiuto: And, of course, the comment should say that it is less optional if you are a TensorFlow user.

site/docs/install-windows.md
```diff
@@ -88,13 +88,14 @@ We recommend installing:
...
     You will likely need these to build and run targets that depend on Bash.  MSYS2 does not install
-    these tools by default, so you need to install them manually.
+    these tools by default, so you need to install them manually. Projects that that depend heavily on these tools are for example Tensorflow.
 
     Open the MSYS2 terminal and run this command:
 
     ```
     pacman -S zip unzip patch diffutils git
     ```
+    Optional: If you want to use Bazel from CMD or Powershell make sure to add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH` Environmentvariable.
```
meteorcloudy, line 98: I would remove the above line and add the following at the end of line 98.
```
Projects that depend on Bash tools in `PATH` need this step, for example TensorFlow.
```

Smokrow: It actually works if you run Bazel in the MSYS2 Terminal. You just need to add the PATH Variable if you want to use pacman Packages downloaded with MSYS2 in CMD and Powershell

Smokrow:
ready for review

site/docs/install-windows.md
meteorcloudy: This line still sounds a bit strange to me. It's noted as optional, but says make sure to add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH`.

Can we change this to:

Optional: If you want to use Bazel from CMD or Powershell and still be able to use Bash tools, make sure to add `MSYS2_INSTALL_PATH/usr/bin` to your `PATH` Environmentvariable.

meisterT:
ping @meteorcloudy

meteorcloudy:
@Smokrow Can you address the comment here: https://github.com/bazelbuild/bazel/pull/11127#discussion_r414368631?

Smokrow:
@meteorcloudy sorry. Missed this one. 
Done!

meteorcloudy:
@Smokrow Thanks!

PR #11140, opened 2020-04-17
Guard parseActionResultMetadata with bulk wrapper

werkt:
ActionResult Orphaned Output Directories must be wrapped in
BulkTransferExceptions in order to be eligible for re-execution. This
change prevents an unavoidable build failure introduced with #10029 in
the event that an output directory tree specified in an action result
is missing from the CAS, and adds testing that would detect such a
regression.


ulfjack:
@meisterT @laurentlb

laurentlb:
I'm doing the cherrypick.

I agree with Ulf, please prefer smaller changes in the future. This is especially true for cherrypick requests (less risk of having an error; less risk of conflicting).

werkt:
Completely agree, just I really didn't want to leave this without a test in place, and duplicating the resource name logic and the unnecessarily complex exception handling inside of the read routine felt irresponsible. Thanks for the quick merge!

PR #11143, opened 2020-04-17
Improve profiling support for repository_ctx.execute

buchgr:
This change improves upon #11131.
 * The first eighty characters of the command line are included
   in the profile. This should make it easier to correlate execution
   times with repository_ctx.execute callsides.
 * The different phases of a remote execution request are highlighted
   in the profile just as they are for remote action execution. In
   particular this exposes information about cache hits and execution
   times.

Example Profile

Screenshot from 2020-04-17 15-20-46

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
@@ -495,11 +495,11 @@ public class SkylarkRepositoryContext
   /** Returns the command line arguments as a string for display in the profiler. */
   private static String profileArgsDesc(String method, List<String> args) {
     StringBuilder b = new StringBuilder();
-    b.append(method).append(": ");
+    b.append(method).append(":");
 
-    String sep = "";
+    final String sep = " ";
     for (String arg : args) {
-      int appendLen = arg.length() + sep.length();
+      int appendLen = sep.length() + arg.length();
       int remainingLen = MAX_PROFILE_ARGS_LEN - b.length();
 
       if (appendLen <= remainingLen) {
@@ -511,8 +511,6 @@ public class SkylarkRepositoryContext
         b.append("...");
         break;
       }
-
-      sep = " ";
     }
 
     return b.toString();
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
```diff
@@ -487,6 +492,32 @@ private void validateExecuteArguments(Sequence<?> arguments) throws EvalExceptio
...
   }
 
+  /** Returns the command line arguments as a string for display in the profiler. */
+  private static String profileArgsDesc(String method, List<String> args) {
+    StringBuilder b = new StringBuilder();
+    b.append(method).append(": ");
+
+    String sep = "";
+    for (String arg : args) {
+      int appendLen = arg.length() + sep.length();
```
meisterT, line 502: super-nit: can you please swap the `arg` and `sep` here to have the same order as the appending order?

src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
```diff
@@ -487,6 +492,32 @@ private void validateExecuteArguments(Sequence<?> arguments) throws EvalExceptio
...
+        b.append(sep);
+        b.append(arg);
+      } else {
+        String shortenedArg = (sep + arg).substring(0, remainingLen);
+        b.append(shortenedArg);
+        b.append("...");
+        break;
+      }
+
+      sep = " ";
```
meisterT, line 515: instead of modifying `sep` in the loop, just remove the space at the end of the string in line 498?

buchgr: done

src/main/java/com/google/devtools/build/lib/bazel/repository/skylark/SkylarkRepositoryContext.java
buchgr: done

PR #11144, opened 2020-04-17
Allow bazel run on windows to handle Ctrl+C signals

tomdegoede:
Currently bazel run on windows prevents the process it's running from receiving Ctrl+C events.

This is because the windows implementation (unlike posix execv) runs the target as a subprocess of the bazel cli maintaining it's Ctrl+C interceptor which tries to interrupt the (no longer running) server instead of the spawned process (three times and then kills).

Not specifying CREATE_NEW_PROCESS_GROUP during CreateProcessW for this case resolves that issue. Potentially a de-register of the bazel cli Ctrl+C handler is desired since it still prints "Bazel Ctrl+C handler; shutting down.".


tomdegoede:
Any feedback?

jin:
Sorry, I forgot to place an assignee for team-Windows.

meteorcloudy:
@tomdegoede Thanks for the contribution! 
The change looks reasonable, can you explain a bit more about how did this affect `bazel run`? I'm not very familiar with this code. The engineer does know this code is no longer available.

tomdegoede:
@meteorcloudy Thanks for your response!
Previously a long running process (e.g. a web server) could not be gracefully terminated using Ctrl+C on Windows during bazel run. Only pressing Ctrl+C three times in a row could force kill the process, causing it's termination logic to be skipped. With this change the Ctrl+C signal is propagated to the running process allowing it to gracefully terminate.

src/main/native/windows/process.cc
```diff
@@ -71,15 +71,15 @@ bool WaitableProcess::Create(const std::wstring& argv0,
                              LARGE_INTEGER* opt_out_start_time,
                              std::wstring* error) {
   return Create(argv0, argv_rest, env, wcwd, stdin_process, stdout_process,
-                stderr_process, opt_out_start_time, false, error);
+                stderr_process, opt_out_start_time, false, false, error);
```
meteorcloudy, line 74: Do you know why changing this method affects `bazel run` command? I'm just a bit worried that this will cause some other place that we actually want to create a new process group to have a wrong behavior.

tomdegoede: This is one of two public Create overloads for WaitableProcess, and the method's comment mentions that it wants to inherit console. This specific overload is only called from blaze_util_windows.cc ExecuteProgram which in it's turn is called by ExecuteRunRequest, and ExecuteServerJvm.

ExecuteRunRequest is the method which we are interested in, so we could pass the boolean from there instead. However looking at blaze_util_posix.cc I think it's safe to say that ExecuteServerJvm is also expected to just "take-over" the current process by calling execv there.

meteorcloudy: Sounds good, that's what I want to know! Thanks so much!

PR #11149, opened 2020-04-18
Point rule reference link to correct subsection

amalkurup89:
Description

Correcting an href link to point to correct subsection on the child html page

Motive

To help current and future users of this repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Build and test software of any size, quickly and reliably.
 ## Documentation
 
   * [Bazel command line](https://docs.bazel.build/user-manual.html)
-  * [Rule reference](https://docs.bazel.build/be/overview.html)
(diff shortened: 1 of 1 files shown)


amalkurup89:
@googlebot I signed it!

amalkurup89:
@googlebot I signed it!

gregestren:
I could see an argument either way.

It's true this change jumps straight to the rule list itself. But the whole page is a "rule reference" that includes common functionality that's still relevant to using rules.

What do you think?

philwo:
What's the status of this PR?

gregestren:
There was some recent debate about whether we should even have this page. @floriographygoth is best qualified to further triage.

Thanks, @amalkurup89 , for your interest and patience in this.

aiuto:
Closing because it is 9 months stale.

PR #11192, opened 2020-04-22
Make it possible to bootstrap Bazel from original source

meteorcloudy:

When providing required `protoc` compiler and relevant tools and libraries, Bazel should be able to bootstrap itself from the original source.

This change fixes the bootstrap scripts by

- Adding missing proto compilation.
- Search for protobuf jars under /usr/share/java if the derived directory doesn't exist. (Should be available by `apt-get install libprotobuf-java`)
- Don't use the bootstrap toolchain if not bootstrapping from the dist archive. Because using that toolchain will make Bazel depend on things under derived directory, which doesn't exist if we bootstrap from Bazel original source and it's not necessary in this case.

### DIFF
--- a/scripts/bootstrap/compile.sh
+++ b/scripts/bootstrap/compile.sh
@@ -16,8 +16,13 @@
 
 # Script for building bazel from scratch without bazel
 
-PROTO_FILES=$(ls src/main/protobuf/*.proto src/main/java/com/google/devtools/build/lib/buildeventstream/proto/*.proto src/main/java/com/google/devtools/build/skyframe/*.proto src/main/java/com/google/devtools/build/lib/skyframe/proto/*.proto)
-LIBRARY_JARS=$(find derived/jars third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
+if [ -d derived/jars ] then
+  PROTOBUF_JARS=derived/jars
+else
+  PROTOBUF_JARS="/usr/share/java/protobuf-3.11.4.jar /usr/share/java/protobuf-java-util-3.11.4.jar"
+fi
+PROTO_FILES=$(find third_party/remoteapis third_party/googleapis third_party/pprof src/main/protobuf src/main/java/com/google/devtools/build/lib/buildeventstream/proto src/main/java/com/google/devtools/build/skyframe src/main/java/com/google/devtools/build/lib/skyframe/proto src/main/java/com/google/devtools/build/lib/bazel/debug src/main/java/com/google/devtools/build/lib/skylarkdebug/proto -name "*.proto")
+LIBRARY_JARS=$(find $PROTOBUF_JARS third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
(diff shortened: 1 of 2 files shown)


scripts/bootstrap/compile.sh
```diff
@@ -16,8 +16,13 @@
 
 # Script for building bazel from scratch without bazel
 
-PROTO_FILES=$(ls src/main/protobuf/*.proto src/main/java/com/google/devtools/build/lib/buildeventstream/proto/*.proto src/main/java/com/google/devtools/build/skyframe/*.proto src/main/java/com/google/devtools/build/lib/skyframe/proto/*.proto)
-LIBRARY_JARS=$(find derived/jars third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
+if [ -d derived/jars ] then
+  PROTOBUF_JARS=derived/jars
+else
+  PROTOBUF_JARS="/usr/share/java/protobuf-3.11.4.jar /usr/share/java/protobuf-java-util-3.11.4.jar"
```
meisterT, line 22: nit: can you extract the version into a variable?

meteorcloudy: Sounds good!

meteorcloudy:
FYI @olekw

scripts/bootstrap/compile.sh
```diff
@@ -16,8 +16,14 @@
 
 # Script for building bazel from scratch without bazel
 
-PROTO_FILES=$(ls src/main/protobuf/*.proto src/main/java/com/google/devtools/build/lib/buildeventstream/proto/*.proto src/main/java/com/google/devtools/build/skyframe/*.proto src/main/java/com/google/devtools/build/lib/skyframe/proto/*.proto)
-LIBRARY_JARS=$(find derived/jars third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
+if [ -d derived/jars ]; then
+  PROTOBUF_JARS=derived/jars
+else
+  PROTOBUF_JAVA_VERSION=3.11.4
+  PROTOBUF_JARS="/usr/share/java/protobuf-$PROTOBUF_JAVA_VERSION.jar /usr/share/java/protobuf-java-util-$PROTOBUF_JAVA_VERSION.jar"
```
olekw, line 23: Is protobuf-java's behavior likely to change enough from one release to the next that the versioning is necessary? Note that this will need to be updated whenever the local system's version of these packages changes, and this is likely to not be in sync with Bazel updates/releases. I'd recommend using the non-versioned name as long as version changes are unlikely to break the functionality needed by Bazel. If protobuf changes **are** likely to break Bazel then default assignment ( :=) should be used instead of normal assignment (=) to allow users to easily override this value.

ulfjack: I'm not sure if they have improved, but in the v2 release series we saw breaking changes on minor releases.

meteorcloudy: Maybe we can specify what protobuf version or range or version Bazel could depend on in the Debian package? So that here, we can just use the non-versioned name?

olekw:
That would work as well. However, we shouldn't limit this just to Debian. Having the full list of versioned dependencies easily accessible should allow build on any system.

meteorcloudy:
OK, I'll just use the non-versioned artifacts.

PR #11197, opened 2020-04-22
Perform Bulk Wait in InputFetcher for downloads

werkt:
Use the bulk wait mechanism to block on input fetcher downloads, and
annotate only-CNF caused BulkTransferExceptions with the CAS eviction
advisory.


werkt:
Worth noting here, #10029 broke the presentation of the CAS eviction message as a result of wrapping due to exception inference of IOException.

werkt:
@meisterT this seems relevant to your interests

PR #11201, opened 2020-04-23
Convert to Python 3

olekw:
Python 2 has reached end-of-life.
This change was imported from Debian packaging.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
@@ -208,7 +208,7 @@ public class BazelPythonSemantics implements PythonSemantics {
      * logic will extract the zip's runfiles into a temporary directory.
      *
      * The stub script has a shebang pointing to a first-stage Python interpreter (as of this
-     * writing "#!/usr/bin/env python"). When a zip file is built on unix, this shebang is also
+     * writing "#!/usr/bin/env python3"). When a zip file is built on unix, this shebang is also
      * prepended to the final zip artifact. On Windows shebangs are ignored, and the launcher
      * runs the first stage with an interpreter whose path is passed in as LaunchInfo.
      */
@@ -240,7 +240,7 @@ public class BazelPythonSemantics implements PythonSemantics {
         PathFragment shExecutable = ShToolchain.getPathOrError(ruleContext);
         // TODO(#8685): Remove this special-case handling as part of making the proper shebang a
         // property of the Python toolchain configuration.
-        String pythonExecutableName = OS.getCurrent() == OS.OPENBSD ? "python3" : "python";
+        String pythonExecutableName = "python3";
         // NOTE: keep the following line intact to support nix builds
         String pythonShebang = "#!/usr/bin/env " + pythonExecutableName;
         ruleContext.registerAction(
--- a/src/test/py/bazel/launcher_test.py
+++ b/src/test/py/bazel/launcher_test.py
@@ -355,7 +355,7 @@ class LauncherTest(test_base.TestBase):
         'helloworld(', '  name = "hello",', '  out = "hello.txt",', ')'
     ])
     foo_py = self.ScratchFile('foo/foo.py', [
-        '#!/usr/bin/env python',
+        '#!/usr/bin/env python3',
         'import sys',
         'if len(sys.argv) == 2:',
         '  with open(sys.argv[1], "w") as f:',
@@ -364,7 +364,7 @@ class LauncherTest(test_base.TestBase):
         '  print("Hello World!")',
     ])
     test_py = self.ScratchFile('foo/test.py', [
-        '#!/usr/bin/env python',
+        '#!/usr/bin/env python3',
         'import unittest',
         'class MyTest(unittest.TestCase):',
         '  def test_dummy(self):',
--- a/third_party/py/mock/setup.py
+++ b/third_party/py/mock/setup.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
 # Copyright (C) 2007-2012 Michael Foord & the mock team
 # E-mail: fuzzyman AT voidspace DOT org DOT uk


aiuto:
I am dubious about updating third_party except to pull down fresh releases or cherrypicks from from upstream. Unless this is scriptable from an update, it is going to introduce a maintenance burden.

We would be better off eliminating the dependency on gflags.  A very quick grep on the source tree shows no use of it. I wonder if this package is vestigial from before conversion to absl flags?

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
```diff
@@ -237,7 +237,7 @@ public void createExecutable(
         PathFragment shExecutable = ShToolchain.getPathOrError(ruleContext);
         // TODO(#8685): Remove this special-case handling as part of making the proper shebang a
         // property of the Python toolchain configuration.
-        String pythonExecutableName = OS.getCurrent() == OS.OPENBSD ? "python3" : "python";
+        String pythonExecutableName = "python3";
```
aiuto, line 240: Won't this break anyone who is still using Python3?

cc: @brandjon

olekw:
> I am dubious about updating third_party except to pull down fresh releases or cherrypicks from from upstream. Unless this is scriptable from an update, it is going to introduce a maintenance burden.

That's a fair point. Regarding scripting, with the exception of BazelPythonSemantics.java, all the other changes were made by using sed (obviously, double-checked afterwards!) I used the whole "#! /usr/bin/env python" string to make sure I didn't accidentally catch something extraneous. That would probably be fairly trivial to script if you wanted to go that route. I searched and (at the time at least) did not find any other references for python vs python3.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
olekw: @aiuto assuming you mean Python 2 then, yes. If a user has *only* Python 2 installed and not Python 3 then this would fail for them. However, I'm not sure how many users, even those who use Python 2 as default, don't have Python 3 installed at all. (Or couldn't easily install it) Debian ships it even in our oldoldstable (which is really obsolete).

keith: FYI macOS before Catalina only had python2 installed by default, so there is no global /usr/bin/python3, _some_ users might have that installed through homebrew at /usr/local/bin/python3, but it's not a requirement today and arguably relying on brewed python is an issue since users could have any sub-version there depending on when they installed it

brandjon: We can't require that Python3 be installed on all target platforms where Python 2 targets run.

The proper solution is probably to review #11434 and make the autodetecting toolchain for Python 3 emit a `python3` shebang. The only case I can think of where this would be a breaking change is where there's no `python3` command in the environment, but there is a `python` command that resolve to a Python 3 interpreter. In which case I'm tempted to just say add an explicit toolchain to your build.

philwo: This code path only affects the disabled-by-default "launchable Python zips" in the deprecated native Python rules of Bazel. For example, if you build something like this:

```
philwo@philwo-macbookpro2:src/pytest $ cat BUILD 
py_binary(
    name = "test",
    srcs = ["test.py"],
)

philwo@philwo-macbookpro2:src/pytest $ cat test.py 
print("Hello world!")

philwo@philwo-macbookpro2:src/pytest $ bazel build --build_python_zip :test
Target //:test up-to-date:
  bazel-bin/test
  bazel-bin/test.zip
```

Then the `bazel-bin/test` would be a ZIP file with the shebang prepended, so you can just run it as an executable, but only if you have a `python3` executable in your path. Without this change you can only run it if you have a `python` executable in your path.

The `python` executable on the path means "Python 2" on almost all distros (only Arch Linux was bold enough to migrate that to Python 3 :)). So currently we require everyone to install an outdated, EOL'd version of Python to run these binaries, even if their own code is Python 3, as it should be.

With the change, we no longer require Python 2.x installed, but Python 3. IMHO this is fine, especially considering https://python3statement.org/.

philwo: @brandjon's suggestion is obviously the cleanest one, but IMHO (just speaking for myself here) any time invested into supporting Python 2 is wasted at this point.

@keith You have a good point there, but looking forward, Apple also said that [you should not rely](https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_release_notes) on any Python version being installed by default on macOS as they might remove them in the future. So this will all become moot in a few months if they actually go ahead and remove pre-installed Python or make it unusable.

brandjon: I think our comments crossed, so interesting that we have opposite conclusions. :)

> deprecated native Python rules of Bazel

Deprecated in the sense that we want people to use stubs instead? Sure, but the rules themselves aren't deprecated, they just haven't been migrated to Starlark yet.

The Bazel logo isn't on the python3statement.org; we still support building code that uses Python 2, and we'll need to make a conscious decision to drop such support before we can change the stub in that way. In the meantime I *think* allowing configuring the shebang should suffice.

You may be right about wasted effort, but I suspect the shebang thing is useful even outside of just selecting the interpreter version.

philwo: > Deprecated in the sense that we want people to use stubs instead?

Sorry, I mixed this up - I thought our `rules_python` were already a new implementation independent of the native rules.

> The Bazel logo isn't on the python3statement.org

Yes, but we made an [announcement for Bazel 2.0](https://groups.google.com/forum/#!msg/bazel-discuss/isTkOF-av_A/iV3UIfB0AQAJ) referencing that page and saying that we commit to it. Admittedly we wrote there that we'd still support building Python 2.x code. :/

What would it take to make that conscious decision to drop Python 2.x support? I would really like to see us do so, of course following all required steps for a change like that.

philwo: > You may be right about wasted effort, but I suspect the shebang thing is useful even outside of just selecting the interpreter version.

Yes, that sounds like a nice way. We could let it default to the current `python` one and introduce an incompatible flag that flips it to `python3`, WDYT?

brandjon: > Admittedly we wrote there that we'd still support building Python 2.x code. :/

Which was a very intentional policy point, not an oversight. Google isn't even free of Python 2 code, so we can't expect all our users to be. (Of course, we're talking about the ability to build and run Python 2 targets, which isn't the same thing as requiring Python 3 to also be present.)

What would dropping Python 2 support look like in your ideal vision? Would we no longer have a `python_version` attribute that chose between PY2 and PY3? To drop support of the whole concept of running PY2 code from the Python rules, I'd want to see virtually no users having any PY2 code in their builds. Given that only last year we changed the default and it broke people, I think we have some time before that happens. Perhaps enough that we'll first migrate the rules entirely to Starlark and it'll be a different calculus (e.g. if the version selection system gets more granularity).

> We could let it default to the current python one and introduce an incompatible flag that flips it to python3, WDYT?

If people are not setting up an explicit Python toolchain in their build, then they're using the autodetecting toolchain, which will look for Python 2 or Python 3 depending on the configuration of the target. It's easy for these toolchains to specify custom shebangs that use `python` and `python3` respectively. (Of course, the `python` command on Arch will yield a Python 3 interpreter, but the stub script doesn't care about the version so long as it's at least present.)

If people are setting up a custom toolchain, they can modify their toolchain definition to explicitly specify a shebang.

For the default value, of a toolchain that is not the autodetecting one, we can probably use an incompatible flag to nudge people along. But this would be for convenience; it doesn't block anything.

aiuto:
I think we should drop this PR and replace it with 3 independent ones.

- Drop the dependency on third_party/py/gflags.  That is a large part of this PR and it is a bigger win
- Convert out nanopb use to py3
- Convert mock to py3

Circle back on which python gets invoked in a few months, after we see more migration to rules_python.

aiuto:
Hi Olek:  #11564 removed //third_party/gflags, so much of this can go away.

werkt:
https://github.com/bazelbuild/rules_docker/blob/faaa10a72fa9abde070e2a20d6046e9f9b849e9a/container/BUILD#L47 (current master) is still referencing this flag - this breaks rules_docker for bazel master.

philwo:
What's the status of this PR?

olekw:
> What's the status of this PR?

Hi @philwo. I'm happy to refactor this in light of the other changes. As I mentioned in the meeting last week, I've been heavily occupied with things much less fun than working on FOSS. But if the consensus is that you'd like to merge (an updated version of) this PR then I can certainly fix it up in the next couple days.

lberki:
@brandjon , I'm inclined to merge this change as it is. WDYT? (then it'll need to be rebased, though)

aiuto:
Is this still a work in progress? 
third_party/py/gflags is no longer part of Bazel, so half of this PR can be deleted.

aiuto:
Friendly ping.  This is very stale now. Much of it refers to deleted code.
@olekw Are you going to refresh or close this?

oquenchil:
@olekw If you update this PR, we can merge.

olekw:
Thanks for the pings, @aiuto and @oquenchil! Day job has been keeping me busy but I can definitely update this PR. Will try to do that in the next couple days.

olekw:
Ok, went through line by line and this (much smaller) PR is updated for HEAD.

I did notice that there is now a %shebang% in [python_stub_template.txt](https://github.com/bazelbuild/bazel/blob/763dd0ce6e1644bf895231432f616427a11d385a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt#L1). From what I can tell, this new code appears to *Do The Right Thing* but I'll leave it to whoever implemented that functionality to verify. :)

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
olekw: Ah, just looked at #11434 and I can now understand the %shebang% functionality better. Is there a consensus here to change the [default](https://github.com/beasleyr-vmw/bazel/blob/6e433a29e3e814dc487df69882e911421b0b2a40/src/main/java/com/google/devtools/build/lib/skylarkbuildapi/python/PyRuntimeInfoApi.java#L46) to Python 3 in that mechanism as well? I would personally recommend that but I'll hold off on adding that to this PR lacking said consensus.

oquenchil:
@fweikert Hi Florian, IIRC you have experience with Py2 to Py3 migration. Could you please answer olekw's question and take over merging this PR?

fweikert:
Happy to merge this PR once the discussion is resolved. However, I'm, not familiar with the particular problem.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
philwo: @olekw Yes, let's do that and change the default there to python3.

philwo (approved):

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
philwo: I'll merge this and send it for internal review, then we can do a follow-up change on that line.

PR #11203, opened 2020-04-23
Introduce new option experimental_use_starlark_python.

abergmeier:
Can be used to test Starlark implementation of Python rules.
Currently would expect rules_python to disable loading symbols from native.
Also in the future this should lead to a broken Java implementation to
find errors in transition fast.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/StarlarkSemanticsOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/StarlarkSemanticsOptions.java
@@ -291,6 +291,19 @@ public class StarlarkSemanticsOptions extends OptionsBase implements Serializabl
   public boolean experimentalAllowTagsPropagation;
 
   @Option(
+      name = "experimental_use_starlark_python",
+      defaultValue = "false",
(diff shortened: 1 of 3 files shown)


abergmeier:
Are the failing tests flaky?

meisterT:
@abergmeier I have no clue about these tests nor Python rules but it seems unlikely that the same tests fails multiple times on multiple platforms. Perhaps trying to rebase and trigger buildkite again.

meisterT:
What's the state of this PR?

brandjon:
In the current state of the PR I only see the flag being added, but no logic to make it have any effect of prohibiting access to the native rules. Not sure why there would be CI failures.

The source has also shifted underneath this PR a bit in the meantime: The flag machinery has been updated, see for instance `BuildLanguageOptions.java`.

But in any case, we already have a flag `--incompatible_load_python_rules_from_bzl`, which disables direct access to the native rules. I'm not sure if this flag was intended to work differently or if it's redundant. We haven't flipped the the load-from-bzl flag because our current strategy is to focus on porting native rules to Starlark without necessarily debundling them at the same step.

PR #11198, opened 2020-04-22
Clean up remote worker tests

katre:

PR #11208, opened 2020-04-23
Remote tests use worker via singlejar

katre:
This is needed to prevent failures after the change to break up build-base.
This depends on #11198


ulfjack:
The title seems off - this is the change to extract the start/stop into a helper file and to use runfiles properly. Otherwise looks good.

katre:
The actual changes with the new remote_util are from https://github.com/bazelbuild/bazel/pull/11198. GitHub doesn't do a good job displaying stacked PRs, sadly.

Issue #11105, opened 2020-04-09
Duplicate entries in `/proc/meminfo` make bazel server crash


DarkDimius:
### Description of the problem / feature request:

running `bazel build //:hello_world` leads to

```
$ bazel build //:hello_world
Starting local Bazel server and connecting to it...

Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/home/dark/.cache/bazel/_bazel_dark/37eeb5c8b6e4daaae7f4d5edfdef7735/server/jvm.out')
```

jvm.out:
```
java.lang.IllegalArgumentException: Multiple entries with same key: Writeback=0 and Writeback=0
        at com.google.common.collect.ImmutableMap.conflictException(ImmutableMap.java:215)
        at com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:209)
        at com.google.common.collect.RegularImmutableMap.checkNoConflictInKeyBucket(RegularImmutableMap.java:147)
        at com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:110)
        at com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:393)
        at com.google.devtools.build.lib.unix.ProcMeminfoParser.<init>(ProcMeminfoParser.java:62)
        at com.google.devtools.build.lib.actions.LocalHostResourceManagerLinux.getMemoryInMbHelper(LocalHostResourceManagerLinux.java:86)
        at com.google.devtools.build.lib.actions.LocalHostResourceManagerLinux.getMemoryInMb(LocalHostResourceManagerLinux.java:40)
        at com.google.devtools.build.lib.actions.LocalHostResourceManagerLinux.getLocalHostResources(LocalHostResourceManagerLinux.java:46)
        at com.google.devtools.build.lib.actions.LocalHostCapacity.getNewLocalHostCapacity(LocalHostCapacity.java:46)
        at com.google.devtools.build.lib.actions.LocalHostCapacity.getLocalHostCapacity(LocalHostCapacity.java:34)
        at com.google.devtools.build.lib.runtime.LoadingPhaseThreadsOption$LoadingPhaseThreadCountConverter.lambda$new$0(LoadingPhaseThreadsOption.java:54)
        at com.google.devtools.build.lib.util.ResourceConverter.applyOperator(ResourceConverter.java:136)
        at com.google.devtools.build.lib.util.ResourceConverter.convert(ResourceConverter.java:122)
[...]

PR #11213, opened 2020-04-24
handle duplicate entries in `/proc/meminfo` see #11105

Links2004:
see #11105 for details

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
+++ b/src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
@@ -45,7 +45,6 @@ public class ProcMeminfoParser {
   @VisibleForTesting
   public ProcMeminfoParser(String fileName) throws IOException {
     List<String> lines = Files.readLines(new File(fileName), Charset.defaultCharset());
-    ImmutableMap.Builder<String, Long> builder = ImmutableMap.builder();
     Map<String, Long> newMemInfo = new HashMap<>();
     for (String line : lines) {
       int colon = line.indexOf(':');
@@ -63,8 +62,7 @@ public class ProcMeminfoParser {
         // Ignore: we'll fail later if somebody tries to capture this value.
       }
     }
-    builder.putAll(newMemInfo);
-    memInfo = builder.build();
+    memInfo = ImmutableMap.copyOf(newMemInfo);
   }
 
   /** Gets a named field in KB. */
--- a/src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
+++ b/src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
@@ -27,6 +27,8 @@ import java.util.HashMap;
 
 /**
  * Parse and return information from /proc/meminfo.
+ * In case of duplicate entries the first one is used
+ * and other values are skipped.
  */
 public class ProcMeminfoParser {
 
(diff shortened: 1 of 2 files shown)


Links2004:
@googlebot I signed it!

src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
```diff
@@ -54,11 +56,14 @@ public ProcMeminfoParser(String fileName) throws IOException {
...
         long val =  Long.parseLong(CharMatcher.inRange('0', '9').retainFrom(valString));
-        builder.put(keyword, val);
+        if (!newMemInfo.containsKey(keyword)) {
+          newMemInfo.put(keyword, val);
+        }
       } catch (NumberFormatException e) {
         // Ignore: we'll fail later if somebody tries to capture this value.
       }
     }
+    builder.putAll(newMemInfo);
```
ulfjack, line 66: Just change this into `memInfo = ImmutableMap.copyOf(newMemInfo);`, and drop the builder variable above. No need to create three maps if two are enough.

src/test/java/com/google/devtools/build/lib/unix/ProcMeminfoParserTest.java
```diff
@@ -75,6 +76,7 @@ public void memInfo() throws Exception {
     assertThat(memInfo.getFreeRamKb()).isEqualTo(14717640);
     assertThat(memInfo.getRamKb("Cached")).isEqualTo(509940);
     assertThat(memInfo.getTotalKb()).isEqualTo(3091732);
+    assertThat(memInfo.getRamKb("Writeback")).isEqualTo(123);
```
ulfjack, line 79: The code implements first wins, but this asserts last wins. Did you actually run the test?

src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
Links2004: good point, changed

src/test/java/com/google/devtools/build/lib/unix/ProcMeminfoParserTest.java
Links2004: No, since I where confused by the documentation, 
what are the values I need to use for the `...`?
in this command ```bazel test //src/... //third_party/ijar/...```

src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
```diff
@@ -23,6 +23,7 @@
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
+import java.util.HashMap;
 
 /**
  * Parse and return information from /proc/meminfo.
```
ulfjack, line 29: Can you add documentation here what happens if there are duplicate entries?

src/test/java/com/google/devtools/build/lib/unix/ProcMeminfoParserTest.java
ulfjack: `bazel test //src/test/java/com/google/devtools/build/lib:unix_test` is probably good enough for local testing in this case. :-)

src/main/java/com/google/devtools/build/lib/unix/ProcMeminfoParser.java
Links2004: updated

Issue #11218, opened 2020-04-24
Remove pkg_rpm from Bazel


aiuto:
//tools/build_defs/pkg:pkg_rpm has moved to https://github.com/bazelbuild/rules_pkg/tree/master/pkg and is no longer maintained as part of Bazel nor used in its build or tests.  Users should migrate to the new versions.

Migration:

1. Add WORKSPACE stanza as specified in a rule_pkg release
https://github.com/bazelbuild/rules_pkg/releases
2. Replace instances of
```
load("//tools/build_defs/pkg:pkg.bzl", "pkg_rpm")
```
with
```
load("@rules_pkg//:pkg.bzl", "pkg_rpm")
```
One line solution (for some users)
```
find . -name 'BUILD' -o -name '*.bzl' | xargs sed -i -e '/^load/s#//tools/build_defs/pkg:#@rules_pkg//:#'
```

Expected timeline:
- Flag flip at next major Bazel release
- Code removal by next minor release after that.

siddharthab:
This flag is not yet available in bazel. Can we remove the `migration-3.y` labels from the issue because bazelisk uses it to check for migration readiness?

Issue #11217, opened 2020-04-24
//tools/build_defs/pkg:incompatible_no_build_defs_pkg_deb Remove pkg_deb from Bazel


aiuto:
//tools/build_defs/pkg:pkg_deb has moved to https://github.com/bazelbuild/rules_pkg/tree/master/pkg and is no longer maintained as part of Bazel nor used in its build or tests.  Users should migrate to the new versions.

Migration:

1. Add WORKSPACE stanza as specified in a rule_pkg release
https://github.com/bazelbuild/rules_pkg/releases
2. Replace instances of
```
load("//tools/build_defs/pkg:pkg.bzl", "pkg_deb")
```
with
```
load("@rules_pkg//:pkg.bzl", "pkg_deb")
```
One line solution (for some users)
```
find . -name 'BUILD' -o -name '*.bzl' | xargs sed -i -e '/^load/s#//tools/build_defs/pkg:#@rules_pkg//:#'
```

Expected timeline:
- Flag flip at next major Bazel release
- Code removal by next minor release after that.

siddharthab:
This flag is not yet available in bazel. Can we remove the `migration-3.y` labels from the issue because bazelisk uses it to check for migration readiness?

aiuto:
Done. I'm not sure what happened there.

siddharthab:
Thanks. Can you also remove the labels from #11218?

aiuto:
Status update: Planned removal in March 2021

aiuto:
Isn't this a "defeature request"? :-)

aiuto:
Status update:
- Removal pkg_deb planned for March 1
- Latest presubmit: https://buildkite.com/bazel/google-bazel-presubmit/builds/43661
- Thanks to @dannysullivan for migrating Google to rules_pkg

PR #11219, opened 2020-04-24
Create incompatible flags to remove pkg_deb and pkg_rpm

aiuto:
Create incompatible flags to remove pkg_deb and pkg_rpm on a different schedule from pkg_tar.
   //tools/build_defs/pkg:incompatible_no_build_defs_pkg_deb
   //tools/build_defs/pkg:incompatible_no_build_defs_pkg_rpm

The process of removing pkg_tar is blocked on switching from a dummy internal rules_python to the external rules_python, so that may not happen in time for the next major Bazel release.

Splitting removal of  deb and rpm packaging rules out let's us flip that flag earlier.

Track removal in #11217 and #11218


aiuto:
Ping

aiuto:
friendly ping

philwo (approved):

philwo:
Don't see anything problematic with this, so from my side please go ahead.

aiuto:
I don't think it will help.   From looking at the CI scripts, I think this kind of flag will never be used.
bazelisk --migration might, but CI won't detect anything that does not begin with the prefix --incompatible.

philwo:
@aiuto Can we close this?

aiuto:
Yes. I think the best path forward is to simply delete the code without a flag. The version in rules_pkg is still backwards compatible so people can easily migrate to that.

Issue #11223, opened 2020-04-24
Duplicate symbols from apple_static_library with Swift


keith:
When using the `ios_static_framework` rule from rules_apple, which has an underlying `apple_static_library` target. If you have a dependency tree like this:

```
ios_static_framework
swift_library
objc_library
cc_library
```

The symbols from the `cc_library` target end up duplicated in the final binary. Our assumption is that it has something to do with this comment https://github.com/bazelbuild/bazel/blob/1645eb8ac6a1df5f7f41599039113bba9f081c89/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcProvider.java#L760-L765 since the `cc_library` targets get added to the ObjcProvider's `library` field by rules_swift https://github.com/bazelbuild/rules_swift/blob/2ebfc403ffdb428befd38e25e236ae181aa1a49b/swift/internal/compiling.bzl#L1486-L1501 which is required for normal cases, but appears to cause this duplication issue when fully linking the binary.

Here's a small sample project that repros this: [duplicatesymbols.zip](https://github.com/bazelbuild/bazel/files/4531969/duplicatesymbols.zip)

With this project run `bazel build ios_framework`, you will then see this warning:

```
/Applications/Xcode-11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: warning same member name (ccfoo_8ef19ba5cbaeb2c2d18053f5cec4195e.o) in output file used for input files: bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-fastbuild-ST-96e8508176c304a494afa69d0efb2bff98cc46f1bcec8e953485b6ad02335d21/bin/libcclib.a(ccfoo_8ef19ba5cbaeb2c2d18053f5cec4195e.o) and: bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-fastbuild-ST-96e8508176c304a494afa69d0efb2bff98cc46f1bcec8e953485b6ad02335d21/bin/libcclib.a(ccfoo_8ef19ba5cbaeb2c2d18053f5cec4195e.o) due to use of basename, truncation and blank paddingf
```

and if you unzip the produced framework and run `nm` on it you can see the duplication:

```
% nm swiftlib.framework/swiftlib

swiftlib.framework/swiftlib(swiftfoo.swift_942f052b5affefc31fa39383ae5d7835.o):
0000000000000000 T _$s8swiftlib3fooyyF
[...]

PR #11225, opened 2020-04-25
Filter duplicate libraries from objc fully link

keith:
Fixes https://github.com/bazelbuild/bazel/issues/11223

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcVariablesExtension.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcVariablesExtension.java
@@ -192,6 +192,10 @@ class ObjcVariablesExtension implements VariablesExtension {
     builder.addStringVariable(
         FULLY_LINKED_ARCHIVE_PATH_VARIABLE_NAME, fullyLinkArchive.getExecPathString());
 
+    // ObjcProvider.getObjcLibraries contains both libraries from objc providers
+    // as well as those from CcInfo. ObjcProvider.getCcLibraries only contains
+    // those from CcInfo. We have to split these lists to make sure duplicate
+    // libraries are not included in the fully linked archive.
     ImmutableSet<Artifact> ccLibs = ImmutableSet.copyOf(objcProvider.getCcLibraries());
     Predicate<Artifact> isNotCcLib = library -> !ccLibs.contains(library);
     Iterable<Artifact> objcLibraries = Iterables.filter(objcProvider.getObjcLibraries(), isNotCcLib);
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcVariablesExtension.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcVariablesExtension.java
@@ -18,8 +18,10 @@ import static com.google.devtools.build.lib.rules.objc.ObjcProvider.IMPORTED_LIB
 import static com.google.devtools.build.lib.rules.objc.ObjcProvider.LINKOPT;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.analysis.RuleContext;
 import com.google.devtools.build.lib.analysis.config.BuildConfiguration;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/rules/objc/ObjcVariablesExtension.java
```diff
@@ -189,9 +191,18 @@ private void addArchiveVariables(CcToolchainVariables.Builder builder) {
   private void addFullyLinkArchiveVariables(CcToolchainVariables.Builder builder) {
     builder.addStringVariable(
         FULLY_LINKED_ARCHIVE_PATH_VARIABLE_NAME, fullyLinkArchive.getExecPathString());
+
+    ImmutableSet<Artifact> ccLibs =
+        new ImmutableSet.Builder()
+        .addAll(objcProvider.getCcLibraries())
+        .build();
```
styurin, line 198:
```suggestion
    ImmutableSet<Artifact> ccLibs = ImmutableSet.copyOf(objcProvider.getCcLibraries());
```

keith: Thanks! I was trying to find the nicer invocation here

styurin: `copyOf` is most likely more efficient. Depending on the type of `getCcLibraries` it can allocate a set of right size and copy elements from it without resizing. A builder allocates a buffer of default size which is fine for small collections, but with large collections it would thrown away right after `addAll` call.

jin:
@jmmv do you know who would be the best person to approve changes native objc rules?

ulfjack:
You touch it, you own it. ;-D

I'm only concerned that this might increase memory consumption otherwise it seems fine to merge. Maybe try merging it and roll it back if the TGP doesn't go through?

keith:
Any idea how we can get this merged?

laurentlb:
ping @lberki

keith:
bump

tetromino (commented):
The problem is real, and the PR will fix it. But IMHO, the change reads a bit like a hack on top of the existing rules_swift hack in https://github.com/bazelbuild/rules_swift/blob/2ebfc403ffdb428befd38e25e236ae181aa1a49b/swift/internal/compiling.bzl#L1486-L1501

At the minimum, we need a TODO comment in addFullyLinkArchiveVariables() referencing an issue explaining why getObjcLibraries() can end up with duplicates from getCcLibraries(). Because after rules_swift has been updated, we'd want to remove the duplicate search.

(Also, is this really the *only* place where getObjcLibraries() is used where the duplicates need to be removed? Why?)

But ideally, we want to change apple_binary so that the rules_swift hack and this hack are unnecessary. @allevato, what would it take to fix apple_binary?

allevato:
@googlewalt can probably speak to this better than I can (and correct anything wrong that I say below).

I believe the issue is simply the disconnect between how `apple_binary` (and `ObjcProvider`) treat `cc_library` dependencies differently from `objc_library` dependencies. For reasons (possibly historical) that I'm not 100% clear on, `cc_library` dependencies (and the code looks *specifically* for `cc_library` rules, which is even worse) are put into a separate provider field from other static libraries, and `apple_binary` later on combines those two sets when it provides the inputs to the linker action. But it ignores libraries from `CcInfo.linking_context` entirely, so the Swift hack exists because there's no other way to get libraries from rules that only propagate a `CcInfo` (like `cc_library`) up to `apple_binary` for linking on Darwin platforms.

To fix this, I think we'd need to fix up the `apple_binary` code to only consume linker inputs from `CcInfo`. I see code in `ObjcLibrary.java` that looks like it's creating a `CcLinkingContext` for the target (though I can't see whether it's merging them from its deps) and then it propagates a `CcInfo`, but the code that `apple_binary` eventually calls (`CompilationSupport.registerLinkActions`) doesn't do anything with the `CcInfo` libraries; it only pulls them from `ObjcProvider`.

keith:
I've updated the original change with a comment. I'm happy to investigate other paths if folks would prefer.

lberki (approved):

lberki:
Let me put this through out internal submit queue.

keith:
Thanks!!

lberki:
You're welcome!

tetromino:
After merging this PR, our performance benchmark flagged a significant heap size regression when linking large apps; we fixed the regression via 28c5332cff1c4d2cdfebda0db0680dc6bd9e5529

keith:
Oof thanks!

PR #11235, opened 2020-04-27
remote: don't fail in the remote package if an in-memory output wasn'…

buchgr:
…t created

Instead ignore it. The ActionExecutionFunction will ensure that all
declared outputs have been created and fail with a more detailed
error than is possible in the remote package.

This behaviour already exists for normal outputs. It's been overlooked
for in-memory outputs. In-memory outputs are typically .d files in C++
compilation.


buchgr:
@ulfjack

ulfjack:
Looks good.

PR #11236, opened 2020-04-27
Anchor input fetches to source action id

werkt:
Provide the specific action id via RequestMetadata which provided an
action input artifact when using remote_download_minimal. This replaces
the unattributable "fetch-remote-inputs" identifier populated for each
input via a nested context.


src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -82,6 +82,8 @@
   // TODO(ulfjack): Throw an exception if it's not a file.
   public abstract long getSize();
 
+  public abstract String getActionId();
```
meisterT, line 85: Consider moving the default implementation (`return "";`) here.

meisterT:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
werkt: Done and commented.

meisterT:
This seems to fail on import: https://buildkite.com/bazel/google-bazel-presubmit/builds/33753#_

I'll have a closer look tomorrow

werkt:
> This seems to fail on import: https://buildkite.com/bazel/google-bazel-presubmit/builds/33753#_

Just needs a merge update. Forthcoming.

werkt:
Ping on this, should be clean for merge.

meisterT:
Sorry George, this needs some changes to internal code. I'll try to fix it next week.

coeuvre:
@werkt We are considering removing `action_id` from remote metadata. Can you elaborate more why would you like to track input prefetches with source `action_id`?

According to the REAPI spec, `action_id` is

>   An identifier that ties multiple requests to the same action. For example, multiple requests to the CAS, Action Cache, and Execution API are used in order to compile foo.cc.

The input prefetches are requested by local actions that depend on outputs of remote executed action (which is identified by the `action_id`). These requests are not part of generating action hence shouldn't be bound with the `action_id`.

cc @tjgq

Issue #11210, opened 2020-04-23
--worker_sandboxing fails to create output directories


jongerrish:
### Description of the problem / feature request:

When performing builds of a android_kt_library and using sandboxed workers (on mac and linux) build fails as unable to create output directories. Bazel fails as it attempts to create child directories before parents. I see this with a Kotlin library but I guess this is an issue with any rules that pass outputs with directories more than one deep.

### Feature requests: what underlying problem are you trying to solve with this feature?

Not blowing up

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

bazel build --worker_sandboxing //some/kotlin/library

### What operating system are you running Bazel on?

mac + linux fail the same way

### What's the output of `bazel info release`?

2020/04/23 12:39:22 Using unreleased version at commit 27

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

Version 3.X with some local patches.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

git@github.sc-corp.net:XXXXXX/bazel.git
0c72f0e2d07b125c4c73dd7733b522f24003c4a3
0c72f0e2d07b125c4c73dd7733b522f24003c4a3

###  Have you found anything relevant by searching the web?

No

### Any other information, logs, or outputs that you want to share?

This can simply be fixed by moving this line

dirsToCreate.addAll(outputs.dirs());

from:-

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java#L103

to after the loop to build the ancestors right below as ancestor directories should be created before children.

I can send a PR for this.

ulfjack:
That code is definitely not correct, and I'm wondering how this could be passing even basic testing. I'm not sure how best to fix it, though. What are you proposing?

ulfjack:
@jmmv @philwo

jongerrish:
I have a fix in our local fork, I'll upstream a PR for this today.

PR #11252, opened 2020-04-29
Create output child directories before their parents

jongerrish:
Fixes: https://github.com/bazelbuild/bazel/issues/11210

Parent directories now created before children.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java
@@ -63,7 +63,7 @@ final class WorkerExecRoot extends SymlinkedSandboxedSpawn {
     // First compute all the inputs and directories that we need. This is based only on
     // `workerFiles`, `inputs` and `outputs` and won't do any I/O.
     Set<PathFragment> inputsToCreate = new LinkedHashSet<>();
-    Set<PathFragment> dirsToCreate = new LinkedHashSet<>();
+    LinkedHashSet<PathFragment> dirsToCreate = new LinkedHashSet<>();
     populateInputsAndDirsToCreate(inputsToCreate, dirsToCreate);
 
     // Then do a full traversal of the `workDir`. This will use what we computed above, delete
@@ -80,7 +80,7 @@ final class WorkerExecRoot extends SymlinkedSandboxedSpawn {
 
   /** Populates the provided sets with the inputs and directories than need to be created. */
   private void populateInputsAndDirsToCreate(
-      Set<PathFragment> inputsToCreate, Set<PathFragment> dirsToCreate) {
+      Set<PathFragment> inputsToCreate, LinkedHashSet<PathFragment> dirsToCreate) {
     // Add all worker files and the ancestor directories.
     for (PathFragment path : workerFiles) {
       inputsToCreate.add(path);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java
```diff
@@ -99,16 +99,16 @@ private void populateInputsAndDirsToCreate(
...
     // And all ancestor directories of outputs. Note that we don't add the files themselves -- any
     // pre-existing files that have the same path as an output should get deleted.
     for (PathFragment path : Iterables.concat(outputs.files(), outputs.dirs())) {
       for (int i = 0; i < path.segmentCount(); i++) {
         dirsToCreate.add(path.subFragment(0, i));
       }
     }
+
+    // Add all ouput directories.
+    dirsToCreate.addAll(outputs.dirs());
```
ulfjack, line 111: I think there should be a comment here somewhere about the order of set additions mattering, and maybe (I know some people hate it) change the method signature to say LinkedHashSet. I certainly had to look up the caller to double-check that it's a LinkedHashSet. Otherwise, this wouldn't work.

jongerrish: Good point. Done.

ulfjack:
@philwo @meisterT

src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java
jmmv: I understand the desire to change the type of the variable to clarify intent, but this change in the variable declaration is a no-op: a future reader of this code (or maybe even a reformatter!) won't understand why this variable looks different than others and may undo the "fix".

I think I suggested this in the original review for this code and I think the author agreed to it... but then I'm not sure what happened with the suggestion. What about changing this (and the other variables) to a `ImmutableSortedSet.Builder`, passing the builder around to all the functions that append to it, and then constructing the final set here?

This would clearly show that this is passed to functions as an in/out parameter, and would also show that order is important (vs. being a hashset, which is only an implementation detail of an ordered container).

philwo:
@tetromino @jmmv I assume you'll handle this one, right?

src/main/java/com/google/devtools/build/lib/worker/WorkerExecRoot.java
tetromino: @jmmv pointed out that because `ImmutableSortedSet` extends `ImmutableSet`, using `ImmutableSortedSet.Builder` may not necessarily fix the potential future refactoring breakage problem. I suppose we should merge the PR in its current state, and think of how to improve readability in such cases as a followup.

jongerrish: @tetromino @jmmv ya - let me know how you want me to proceed, I'm happy to make changes. I noticed that ImmutableSortedSet.builder() is deprecated and didn't see an API that would preserve insertion order though. Maybe I missed something?

tetromino: I think your original idea to use a `LinkedHashSet` was the right way to go.

An `ImmutableSortedSet.Builder` sorts entries via a comparator (the default natural order comparator, if you use the deprecated `ImmutableSortedSet.builder()` or the new `ImmutableSortedSet.naturalOrder()`), not by insertion order.

For directories that would incidentally work, since a subdirectory would be sorted after its parent directory, but a data structure that preserves insertion order would more explicitly fit your intent and in any case would be more efficient.

Unfortunately, I don't see an immutable insertion order preserving set in Guava.

PR #11253, opened 2020-04-29
Add flag to disable output library.ap_

jongerrish:
android_library `ResourceLinkAction` exports generated library.ap_ files. These files grow exponentially in size as their contents are merged in to depending libraries, but library.ap_ files are in the format consumable by devices only. Since android_binary generated the final library.ap_ file these intermediates are not needed as outputs.

For a build the size of ours this saves around 2GB of unnecessary outputs.


src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
```diff
@@ -881,6 +881,14 @@ public static AndroidManifestMerger fromString(String value) {
         help = "Use R.txt from the merging action, instead of from the validation action.")
     public boolean useRTxtFromMergedResources;
 
+    @Option(
+            name = "output_library_linked_resources",
+            defaultValue = "true",
+            documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
+            effectTags = {OptionEffectTag.UNKNOWN},
+            help = "If disabled, does not provide library.ap_ outputs for library targets")
```
jin, line 889: Could you also include the PR description here? The PR description has great context, e.g. reducing download overhead and devices not requiring these files.

jongerrish: Done - thanks for the review!

jin:
Update: looking into failing tests internally with blaze.

jongerrish:
Thanks Jin!

jongerrish:
I think https://github.com/bazelbuild/rules_android/issues/10 is a better solution

artem-zinnatullin:
>I think bazelbuild/rules_android#10 is a better solution

But do you think there is any use case in `library.ap_ ` files in general?

I agree that disabling R class merging is generally what we should aim for but wouldn't it still be producing useless `library.ap_` files, just smaller?

jongerrish:
I can't see a usecase for library.ap_ - I think it was just a side affect (-o is a mandatory parameter) from calling aapt2 link which is used to generate the R.java + R.txt. However, `AndroidResourceMerge` now generates an R.class + R.txt directly from the intermediate compilation outputs, so really linking is no longer necessary at all. That's why I prefer bazelbuild/rules_android#10 Also both is behind a flag, just in case there is a strange usecase I haven't thought of.

ulfjack:
@jin, can you provide an update here, please? This is part of a batch of changes that are necessary to get Bazel's Android builds on par with Buck wrt. build performance. This is a blocker for some people for migrating from Buck to Bazel.

jin:
I'm currently adding tests to add coverage for our decoupled `@android_tools` infra, so @jongerrish's PRs can be properly tested on our CI and not run into failures internally.

ulfjack:
Cool, glad to hear it. Have you been able to make progress? Anything we can help with?

jin:
Blocking on internal review.

jin:
Could you rebase your PRs onto  https://github.com/bazelbuild/bazel/commit/f3d1683fcb3c4a25c4c4b8251d00ddfa66f958ba please?

jongerrish:
@jin Just resolved conflicts is that enough?

src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
```diff
@@ -1309,6 +1323,10 @@ boolean useRTxtFromMergedResources() {
     return useRTxtFromMergedResources;
   }
 
+
+  boolean outputLibraryLinkedResources() {
+    return outputLibraryLinkedResources;
+
```
jin, line 1329: You have a missing closing brace here

aiuto:
Friendly  ping.  The tests are failing. Are you going to fix or drop this.

Bencodes:
@jin any updates on getting this merged?

nkoroste:
Any updates on this?

oquenchil:
@jin It looks like this PR was approved and it doesn't need any more changes. Can it be merged? Or does it not make sense any more and it should be closed?

ajanuar:
Any update on this?

Bencodes:
This is still something that would be really nice to merge so I opened the PR back up with the merge conflicts resolved. Mind taking another quick look @jin @aiuto?

https://github.com/bazelbuild/bazel/pull/14786

aiuto:
I tried to get a better reviewer than myself. Let's follow up in the other
PR.

Bencodes:
Awesome thanks @aiuto!

Bencodes:
This PR was reopened over here with the conflicts resolved.

https://github.com/bazelbuild/bazel/pull/14786

ted-xie:
Oh, thanks Ben! Closing this PR then.

PR #11257, opened 2020-04-29
Remove jdk langtools from third_party

meteorcloudy:
We don't actually depend on anything under this directory except the
LICENSE file, which should be fetched from the remote java tools.


meteorcloudy:
Turns out we do need this for src/test/shell/bazel/bazel_java_tools_test.sh

PR #11260, opened 2020-04-30
Generate grpc using built-in shell environment

gabrfarina:
This commit updates the implementation of `cc_grpc_library` to use the built-in shell environment. It aligns the implementation with the `proto_gen` rule for protocol buffers.


ulfjack:
The change is certainly correct. I'm not sure this is the right place for it though - we're pulling grpc from upstream, so I think the change would have to be made there.

gabrfarina:
Ah, that's a good point, I didn't realize it was just a mirror of upstream grpc. I'll open a PR on GRPC instead then :-)

Issue #11265, opened 2020-04-30
python_stub_template: `AttributeError: 'module' object has no attribute 'OrderedDict'` on CentOS 6


beasleyr-vmw:
### Description of the problem / feature request:

Recently upgraded to Bazel 3.1.0. Discovered that CentOS 6 users (don't ask) are now broken due to https://github.com/bazelbuild/bazel/commit/d5012a7ed19c1e75e413373fa22c5f1b08a9ee29.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Get yourself a Linux environment where `#!/usr/bin/env python` invokes Python 2.6.6.
1. Run any `py_binary`, such as rules_pkg's `build_tar`.

### What operating system are you running Bazel on?

CentOS 6.6

### What's the output of `bazel info release`?

`release 3.1.0-vmware`

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

n/a

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

n/a

###  Have you found anything relevant by searching the web?

- https://github.com/bazelbuild/rules_python/blob/master/proposals/2018-11-08-customizing-the-python-stub-template.md acknowledges non-hermeticity.
- https://github.com/bazelbuild/bazel/issues/137 discusses support for customizable templates.

### Any other information, logs, or outputs that you want to share?

Non-hermeticity is the key problem for me. We have Python toolchains to decouple our builds from tools provided by the user's OS, but the stub's shebang bypasses these.

Workarounds I'm considering:
- Reverting d5012a7ed19c1e75e413373fa22c5f1b08a9ee29.
- Reimplementing in terms of Python 2.6.
- Tweaking the stub to begin with `%python_binary%` instead of `#!/usr/bin/env python`.

beasleyr-vmw:
Closing since #11269 was merged. (Test [failure](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/ba8e8f6f-6aa7-4ba4-88ae-15b4adda44a5/src/test/java/com/google/devtools/build/lib/skyframe/SkyframeTests/attempt_1.log) appears to be unrelated.)

PR #11269, opened 2020-04-30
Reintroduce support for Python 2.6 to python_stub_template

beasleyr-vmw:
Commit d5012a7ed19c1e75e413373fa22c5f1b08a9ee29 introduced a Python
2.7 dependency into python_stub_template.txt.  Unfortunately this stub
is non-hermetic, so even if the user has configured Python 2.7 and 3
`py_runtime`s, their builds may fail if the system-provided Python is ancient
(ex: CentOS 6.6 provides Python 2.6.6).

Accommodate ancient Python by reworking the path deduplication in terms
of a `set` and a generator instead of `collections.OrderedDict`.

Workaround for #11265.

Testing Done:
- `bazelisk test //src/test/shell/integration:python_stub_test`
- In a CentOS 6.6 env:
```console
$ /usr/bin/env python -V
Python 2.6.6
$ cat >test.py <<EOF
import sys
print(sys.executable)
EOF
$ cat >BUILD <<EOF
py_binary(name = "test", srcs = ["test.py"])
EOF
$ bazel run :test
/path/to/my/hermetic/python3
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -18,7 +18,6 @@ import shutil
 import subprocess
 import tempfile
 import zipfile
-import collections
 
 # Return True if running on Windows
 def IsWindows():
@@ -264,7 +263,11 @@ https://github.com/bazelbuild/bazel/issues/7899 for more information.
 
 def Deduplicate(items):
   """Efficiently filter out duplicates, keeping the first element only."""
-  return list(collections.OrderedDict((it, None) for it in items).keys())
+  seen = set()
+  for it in items:
+      if it not in seen:
+          seen.add(it)
+          yield it
 
 def Main():
   args = sys.argv[1:]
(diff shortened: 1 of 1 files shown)


aiuto:
@brandjon Since it is already the weekend for Jin, I'll cover here. If you do the approve, I've got the patch CL lined up.

brandjon:
I'm a little reluctant to do this, because it sets a precedent that we want to support Python 2.6 (released about 12 years ago) in the stub script. If that's going to be a requirement then we should arguably include a Python 2.6 interpreter in our test environment and run at least one `py_binary` in it.

It also may be unnecessary if we instead add the ability to select an alternate stub script shebang in the Python toolchain. See #8685. But I appreciate that in the absence of a shebang-selecting feature in the toolchain, this PR is useful and is anyway harmless.

How about we accept this patch and just add a comment reminding to keep compatibility with Python 2.6 until #8685 is implemented.

brandjon:
(Added comment clarifying this stance in the internal merge.)

beasleyr-vmw:
Great, thanks!

PR #11270, opened 2020-04-30
update pkg_tar to 0.3.x

aiuto:
Needed to unblock #8857


aiuto:
Blocked on https://github.com/bazelbuild/rules_pkg/issues/171
and subsequently rules_pkg 2.6

meisterT:
It seems that the blockers are resolved now?

aiuto:
But not it is so stale I would go to a newer release.  So still alive, still WIP

philwo:
@aiuto friendly ping

aiuto:
Going to skip 0.2.5 and  go right to 0.2.7 when that comes out.

meisterT:
What's the state of this PR?

aiuto:
At this point I'm going to wait until rules_pkg 0.3.0, which should be out soon.

limdor:
@aiuto can you give an update of this PR if there is any blocker? I saw that 0.3.0 ruels are out and I also saw that Bazel still uses 2.4.0 https://github.com/bazelbuild/bazel/blob/master/distdir_deps.bzl

limdor:
@aiuto I created a PR to see the results of the tests https://github.com/bazelbuild/bazel/pull/12739

limdor:
I see, not so straightforward 😥 
```
(11:54:53) ERROR: Traceback (most recent call last):
--
  | File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/rules_pkg/BUILD", line 117, column 14, in <toplevel>
  | exports_files(["WORKSPACE"], visibility = ["//visibility:public"])
  | Error in exports_files: visibility for exported file 'WORKSPACE' declared twice
```

limdor:
The problem if I understood right is that there is this patch to add the visibility public for the workspace file
```
http_archive(
    name = "rules_pkg",
    patch_cmds = EXPORT_WORKSPACE_IN_BUILD_FILE,
    patch_cmds_win = EXPORT_WORKSPACE_IN_BUILD_FILE_WIN,
    sha256 = DIST_DEPS["rules_pkg"]["sha256"],
    urls = DIST_DEPS["rules_pkg"]["urls"],
)
```
But the visibility for WORKSPACE file is already set in the BUILD file: https://github.com/bazelbuild/rules_pkg/blob/main/pkg/BUILD#L27
This was not an issue in the current version because WORKSPACE file is not exported.
@aiuto what would be the way to proceed? Should the patch be adapted that if it is already there it should be replaced?

aiuto:
Status update

The current blocker is annoying
- rules_pkg requires rules_python
- bazel uses a mocked_up rules_python.
- bazel tests rely on rules_pkg, but they currently don't include a usable python

One solution is to add rules_python into the WORKSPACES users for tests. I'm not doing that. It just makes the complexity worse.
A different path is to change the tests so they don't dependon rules pkg.
- we could fall back to @bazel_tools/build_defs/rules_pkg.
- we could provide a mini tar maker just for testing

The first approach is quick and easy. It just defers the technical debt payback until some future, hidden date.
The second approach pays back all the debt right now, but there are more pressing issues than this.

I'm going with approach 1.

meteorcloudy (approved):

limdor:
@aiuto I wanted to ask if we could cherry pick this for Bazel 4 LTS, I do not directly needed but somehow I think that when I realize that is needed might be too late. What is your opinion?

meisterT:
@limdor if you do want to request a cherry pick I suggest to do it here: https://github.com/bazelbuild/bazel/issues/13099

limdor:
that's the plan, but before writing too much there I wanted to sync here first if it makes sense or not. If you think that we should discuss it directly there I will put it right away

aiuto:
This would have to be hand backported instead of direct changed.  The 4.x branch does not have distdir_deps.bzl, so you would need the changes directly to WORKSPACE and maybe other places in the code tree.

It is also of low value. This only impacts the version of pkg_tar we use to build Bazel, not the versions that users would use.

PR #11274, opened 2020-05-01
Update Bazel to use the 3.1.0 toolchains.

katre:
Can someone (@philwo ?) help me populate mirror.bazel.build with this?


philwo:
@katre I tried to mirror the file now, but it was already on the mirror. So I think this should be fine. :)

PR #11277, opened 2020-05-02
deltazulu@ add //third_party:dagger to bazel build

dx404:
- Allows Android desugar tool to use DI framework


philwo:
I'm not a fan of adding more third-party dependencies to Bazel. :/

Also note that this doesn't add just Dagger, but also a 2.7 MB "javac-shaded-9-dev-r4023-3.jar", part of "googlejavaformat".

meteorcloudy:
Is there any plan to move android tools out of Bazel?

philwo:
I just asked Lukacs and he said, it's mostly historical that these tools are still in our repository. They're not actually used directly by Bazel, instead we use them to build a release archive that we publish as an external repository and then Bazel downloads it. So they could be extracted to a separate repo. Apparently there are no real technical blockers, it just hasn't been done yet.

meteorcloudy:
I see.
@dx404 Can you confirm that Bazel binary won't depend on those targets added?
If so, this PR should not add any trouble to https://github.com/bazelbuild/bazel/issues/9408

But same as @philwo , I would really like to avoid adding new dependencies to third_party/ if that's possible.

philwo:
@dx404 How urgent is this? Maybe we can do the extraction first and then add your new dependencies in the new repo? I'm really worried about our repo size, too. If we add 3 MB now, it will forever be in the history that we `git clone` on all our machines etc.

aiuto:
Is seems like the first step should be to move the android stuff out of bazel, so that this PR become moot.

dx404:
1. I am using it only for Android's build tools, the desugar. I don't think it is integrated into the bazel binary, i.e. no such configuration flag as bazel build --config=android_java8_libs for bazel.

2.I am OK with moving Android stuff separately, as long as the third party jars are easy to update (in sync with g3). 

3. Not a hard requirement, though it would be nice to have it by this quarter.  

-

meisterT:
It's currently unused in this PR, right?

Can you please show us how the follow up change would look like?

meisterT:
Closing for lack of activity, please reply if you want to pick up the work again.

PR #11281, opened 2020-05-04
Provide a way to build against Debian system installed java libraries

meteorcloudy:
This change provides a way to replace some java dependencies in third_party with Debian system installed java libraries.

- Add a new local repository which points to `/usr/share/java`
- Define targets for needed system java libraries in `debian_java.BUILD"
- Add a new config_setting to enable the Debian build
- Introduced a new distrib_java_import rule to help redirect targets to the ones defined in debian_java.BUILD

With the above work, we can replace a jar dependency by two steps:

- Define the target in debian_java.BUILD
- Add `enable_distributions = ["debian"]` attribute for the target we want to replace.

To enable building against Debian system libs, add `--define=distribution=debian` as a build option when building Bazel.

To add a new distribution support for Bazel, we only need to

- Add a new config_setting definition for the distribution
- Implement the logic for redirecting targets for the distribution in `tools/distributions/distribution_rules.bzl`
- Add the distribution name to enable_distributions attribute for the targets we want to replace.

Note that this is for Bazel's java dependencies only, we'll implement the similar procedure for our C++ dependencies in later changes.


meteorcloudy:
FYI @olekw, you should be able to delete a jar file under third_party in the Debian source package 
once it's replaced.

tools/distributions/distribution_rules.bzl
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_java//java:defs.bzl", _java_import = "java_import")
+
+
+def distrib_java_import(name, enable_distributions = [], **kwargs):
+    """A marco for java_import rule to support distributions build (eg. Debian)"""
```
philwo, line 19: s/marco/macro/

third_party/BUILD
```diff
@@ -1,4 +1,5 @@
-load("@rules_java//java:defs.bzl", "java_import", "java_library", "java_plugin")
+load("@rules_java//java:defs.bzl", "java_library", "java_plugin")
+load("//tools/distributions:distribution_rules.bzl", java_import = "distrib_java_import")
```
meisterT, line 2: Overwriting a well known symbol with another one seems hacky. Why not make it explicit where you need it?

tools/distributions/distribution_rules.bzl
meteorcloudy: Thanks, done!

tools/distributions/debian/debian_java.BUILD
```diff
@@ -0,0 +1,39 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
```
meisterT, line 1: General question: should things like these live in our source tree or be part of debian packaging?

Probably @philwo has an opinion on this.

third_party/BUILD
meteorcloudy: I try to be as less intrusive for Bazel developer as possible. Because most Bazel developer don't have to know what distrib_java_import is. But I don't have a strong opinion on this. @philwo WDYT?

tools/distributions/debian/debian_java.BUILD
meteorcloudy: I think it's just easier to maintain if we keep it in Bazel repo?

third_party/BUILD
philwo: I agree with Tobi - it's probably more confusing if a well-known rule like `java_import` suddenly acts differently. I'd make it explicit and just call it `distrib_java_import`.

tools/distributions/debian/debian_java.BUILD
philwo: Yeah, I'd say let's keep it in our repo for now - Debian / Ubuntu is a distribution we officially support and distributions try to keep their patches for upstream packages minimal. If we have it in our repo, we can at least test that we don't accidentally break it.

meisterT: Sounds reasonable.

third_party/BUILD
meteorcloudy: Cool, it's done!

meteorcloudy:
Thanks for the review, I'll split this change into two so that we can merge it.

olekw:
I love this concept! This is a much cleaner and more-sustainable way of addressing this problem than the hacks we were using before! Plus, I love the extensibility to other distributions!

olekw:
PS Planning to cherry-pick this today into our working 3.0.0 tree. This is great!

meteorcloudy:
Merged by 
https://github.com/bazelbuild/bazel/pull/11282
https://github.com/bazelbuild/bazel/pull/11283

PR #11283, opened 2020-05-04
Debian build: Add enable_distributions attribute for targets we want to replace

meteorcloudy:
Third_party part of https://github.com/bazelbuild/bazel/pull/11281


meteorcloudy:
Merged at https://github.com/bazelbuild/bazel/commit/c393ef2d3c587dabd20a6221b7f63acbfdf149d6

PR #11286, opened 2020-05-05
Fix Debian netty target

meteorcloudy:
The netty-all.jar doesn't contain any actual class file, it depends on
all other netty jars. Therefore, we should list all netty jars in
Debian's netty target.


meteorcloudy:
FYI @olekw

olekw:
We'll cherry-pick this for now in Debian packaging but I've also [requested that the Debian netty package include all classes in netty-all.jar](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959766) so that it can be compatible with upstream and reasonable expectations.

PR #11287, opened 2020-05-05
Add zsh completion to debian package

siddharthab:
The zsh completion script was missing from the debian package. It is already packaged in the installer, and is installed by Homebrew, so might as well do the same thing for the debian package.

Also updated the documentation to be more current.


aiuto:
Passing off to meteorcloudy, since they are doing a lot of debian fixes right now.

PR #11288, opened 2020-05-05
Try to remove instrumentation api

meteorcloudy:
The maven page shows instrumentation-api is from
https://github.com/google/instrumentation-java,
but that repo has become
https://github.com/census-instrumentation/opencensus-java,
which is another dependency we have (opencensus-api)

instrumentation-api looks obsolete and should be removed.


meteorcloudy:
I am not 100% sure we can remove this dependency, but let's see if it will break anything

meteorcloudy:
FYI @olekw

meisterT:
I checked and while it was necessary back then, it doesn't appear to be necessary anymore, since the file `tools/grpcz/census.proto` doesn't exist any more.

meteorcloudy:
Nice, thanks!

meteorcloudy:
> since the file tools/grpcz/census.proto doesn't exist any more.

Do we still need opencensus-api as a runtime dependency?

meisterT:
> Do we still need opencensus-api as a runtime dependency?

Sorry, I don't know.

meteorcloudy:
No problem, I can check that

meteorcloudy:
Merged at https://github.com/bazelbuild/bazel/commit/13effb9bd881b89822a7d8f6e9b26aa4e13e2b90

meteorcloudy:
We cannot remove opencensus-api as a runtime dependency, check https://github.com/bazelbuild/bazel/pull/11289

PR #11294, opened 2020-05-05
Resource linking for android_library targets is unnecessary.

jongerrish:
Use R.txt from AndroidCompiledResourceMerger
Don't include R.java in srcjar output
Avoid AndroidResourceLink action
Avoid large library.ap_ + resource.srcjar outputs
Enable with the following flags:-

--nolink_library_resources --experimental_use_rtxt_from_merged_resources

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidRuleClasses.java
@@ -374,7 +374,6 @@ public final class AndroidRuleClasses {
 
           if (AndroidResources.definesAndroidResources(attributes)) {
             implicitOutputs.add(
-                AndroidRuleClasses.ANDROID_JAVA_SOURCE_JAR,
                 AndroidRuleClasses.ANDROID_R_TXT,
                 AndroidRuleClasses.ANDROID_RESOURCES_CLASS_JAR);
           }
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
@@ -113,7 +113,8 @@ public class AarImport implements RuleConfiguredTargetFactory {
                 dataContext,
                 manifest,
                 DataBinding.contextFrom(ruleContext, dataContext.getAndroidConfig()),
-                neverlink);
+                neverlink,
+                dataContext.getAndroidConfig().linkLibraryResources());
 
     MergedAndroidAssets mergedAssets =
         AndroidAssets.forAarImport(assets)
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidCommon.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidCommon.java
@@ -409,21 +409,17 @@ public class AndroidCommon {
     return ruleContext.getHostPrerequisiteArtifact("debug_key");
   }
 
-  private void compileResources(
-      JavaSemantics javaSemantics,
+  private void addResourceClassJarToClassPath(
       Artifact resourceJavaClassJar,
-      Artifact resourceJavaSrcJar,
       JavaCompilationArtifacts.Builder artifactsBuilder,
       JavaTargetAttributes.Builder attributes,
       NestedSetBuilder<Artifact> filesBuilder)
-      throws InterruptedException, RuleErrorException {
+      throws InterruptedException {
 
     // The resource class JAR should already have been generated.
     Preconditions.checkArgument(
         resourceJavaClassJar.equals(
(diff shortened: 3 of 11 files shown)


jongerrish:
@jin @cgruber Thoughts?

jongerrish:
Goes some way to addressing https://github.com/bazelbuild/rules_android/issues/10

jin:
cc @donaldchai @timpeut @djwhang

donaldchai:
Thanks for reminding me about `--experimental_use_rtxt_from_merged_resources`.  @jin: Do we need to go through the "incompatible change" process to unconditionally switch to that?  (I don't expect any downsides to this.)

Jon: overall I like the idea of pruning things from the build, but I think this disables a useful validation step.  In particular the "aapt2 link" verifies that resources referenced from XML are defined.  So trying to build `java/com/foo:foo` should fail:

```
-- java/com/foo/BUILD --
android_library(
  name = "foo",
  manifest = ...
  resource_files = glob(["res/**"]),
  deps = ["//java/com/bar"],
)

-- java/com/foo/res/values/values.xml --
<resources>
  <string name="asdf">@string/missing</string>
</resources>

-- java/com/bar/BUILD --
android_library(
  name = "bar",
  srcs = "Bar.java",
  # no resources
)
```

Is that still true after this change?  I was planning on writing a new validator (potentially emitting a warning if a referenced resource is not in a *direct* dep) but haven't gotten to it.

jin:
> Thanks for reminding me about --experimental_use_rtxt_from_merged_resources. @jin: Do we need to go through the "incompatible change" process to unconditionally switch to that? (I don't expect any downsides to this.)

No, I don't think it's required for this.

ulfjack:
@donaldchai, since this is behind a flag which defaults to building linking the resources, that seems fine?

donaldchai:
> @donaldchai, since this is behind a flag which defaults to building linking the resources, that seems fine?

Sure, but this decision is ultimately about how code is merged and how flags interact.  I have no authority here.  All of my own changes require approval by someone from the Bazel team, or others CCed to this PR.

ulfjack:
@jin can you merge this?

meisterT:
@jin any objections?

jongerrish:
@donaldchai - with this flag enabled you won't get any validation from linking as its disabled.

Probably this validation should happen in `AndroidCompiledResourceMergingAction` actually, it already supports a flag `--throwOnResourceConflict` (it looks like this defaults to `true` unless the rules is whitelisted which I don't think happens in open source Bazel (i.e: resource conflicts are always allowed) but it seems that this is the action that should validate the resources (does a referenced resource exist? is a resource duplicated? etc) as it does the merge (generates merged R class).

jongerrish:
Hi @jin - the PR has been outstanding for a while now, is there any concern with it? I'm happy to discuss and address any issue you may have.

src/main/java/com/google/devtools/build/lib/rules/android/AndroidRuleClasses.java
```diff
@@ -374,7 +374,6 @@ public static boolean hasProguardSpecs(AttributeMap rule) {
 
           if (AndroidResources.definesAndroidResources(attributes)) {
             implicitOutputs.add(
-                AndroidRuleClasses.ANDROID_JAVA_SOURCE_JAR,
```
ahumesky, line 377: Removing this will break any current usage of the .srcjar output as a label, whether the flag is on or off, so it looks like we'd need to go through the incompatible change process. Unfortunately there's no way to conditionally register implicit outputs based on configuration. Glancing through the code, it should be technically possible to enable that, but we'd need to think through that separately. The next best thing is to leave this here, and when --link_library_resources is false, register a FailAction for that output. This means that the failure won't be surfaced until execution time, unfortunately.

donaldchai:
> Probably this validation should happen in `AndroidCompiledResourceMergingAction` actually, it already supports a flag `--throwOnResourceConflict`

I feel the opposite way, i.e. `AndroidCompiledResourceMergingAction` shouldn't be doing any of these checks since it's on the critical path of the build, and the checks are not needed to generate R classes.

As as a side note, detection and reporting of "resource conflicts" on `android_library` should be overhauled so that they can be made errors instead of warnings (i.e. build log spam).  The current override semantics leave a lot to be desired.

aiuto:
Time for a decision.  Are we going to merge it or not?

jin:
@ahumesky could you make a call here about merging the PRs modifying Android native rules?

nkoroste:
Any updates on this?

aiuto:
Friendly ping to all.

meisterT:
Given the Starlarkification of the rules, this PR probably should be reopened against https://github.com/bazelbuild/rules_android at one point.

I am going to close the PR, feel free to re-open if you disagree.

PR #11297, opened 2020-05-05
Fix C++ string include

keith:
This file uses functions like `strrchr` that are provided by string.h,
it seems like somehow this works implicitly on macOS but this seems to
be more correct.


keith:
This test failure appears to be unrelated

ulfjack:
LGTM for this small cleanup change. I also double-checked `man strrchr`.

PR #11300, opened 2020-05-06
Allow replace zlib with Debian system libs

meteorcloudy:

### DIFF
--- a/third_party/zlib/BUILD
+++ b/third_party/zlib/BUILD
@@ -1,3 +1,5 @@
+load("//tools/distributions:distribution_rules.bzl", "distrib_cc_library")
+
 licenses(["notice"])  #  BSD/MIT-like license
 
 filegroup(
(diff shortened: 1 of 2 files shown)


meteorcloudy:
FYI @olekw

meteorcloudy:
It turned out the zlib sources are packaged in the embedded tools. I cannot ship a BUILD file that uses the `distribution_rules.bzl", so I created a BUILD.tools file for zlib. PTAL.

meteorcloudy:
I have to rename the BUILD.tools to BUILD.bazel for the embedded source, PTAL

olekw:
Quick question on this: does Bazel need to link zlib statically? If so, is it possible to link dynamically at least in the Debian variant? Debian is rather averse to static linking if it can possibly be avoided. (Security considerations) Thanks!

meisterT:
cc @oquenchil for the question on static vs dynamic linking

oquenchil:
I have verified that meteorcloudy's proposed solution of removing the dependency from zlib_client and adding linkopts = ["-lz"] to the zipper cc_binary works. So, yes this is possible.

meteorcloudy:
Cool, I'll send a change to update the debian_cc.BUILD file

meteorcloudy:
Merged at https://github.com/bazelbuild/bazel/commit/2a07773baa0f64bb5b467df77088130486ffb3e9

PR #11302, opened 2020-05-06
Use a custom LICENSE file for the Debian build

meteorcloudy:
The `//src/main/java/com/google/devtools/build/lib:merge_licenses` target causes the Bazel binary to depend on `//third_party:srcs`, which introduced many unnecessary third party dependencies.

This change excluded that target for Debian build and point users to a standard location for license file in Debian.

With this change, we can get rid of all deps in [Bazel dependencies](https://docs.google.com/spreadsheets/d/1yqbzbR0PntWAxt4AxwyuXGn-5CjeKaGEEfCJ91n6VPA/edit#gid=0) marked as "Exclude" for the Debian build.

### DIFF
--- a/tools/distributions/debian/bazel_debian_license
+++ b/tools/distributions/debian/bazel_debian_license
@@ -1,0 +1,2 @@
+Please find the license file of Bazel at /usr/share/doc/bazel/copyright
+
--- a/src/main/java/com/google/devtools/build/lib/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/BUILD
@@ -450,7 +450,17 @@ merge_licenses(
         "//:LICENSE",
         "//third_party:srcs",
     ],
-    out = "runtime/commands/LICENSE",
+    out = "LICENSE.merged",
+)
+
+genrule(
+    name = "bazel_license",
+    srcs = select({
+        "//src/conditions:debian_build": ["//tools/distributions/debian:bazel_debian_license"],
+        "//conditions:default": [":merge_licenses"],
+    }),
+    outs = ["runtime/commands/LICENSE"],
+    cmd = "cp $< $@",
 )
 
 java_library(
--- a/src/main/java/com/google/devtools/build/lib/bazel/commands/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/commands/BUILD
@@ -16,7 +16,7 @@ java_library(
     resources = [
         "fetch.txt",
         "sync.txt",
-        "//src/main/java/com/google/devtools/build/lib:merge_licenses",  # license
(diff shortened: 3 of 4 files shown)


meteorcloudy:
FYI @olekw, can you also help review this?
Is `/usr/share/doc/bazel/copyright` the correct place for Bazel's license file in Debian?

olekw:
Removing the unnecessary deps is great and I'm very excited about that! It will definitely streamline the build process!

Yes, that is the correct location for the Debian copyright file. Note that the one for Bazel is woefully incomplete at present, I'll work on it once we're closer to understanding what will be in the final package. It should ultimately look more [like this](https://salsa.debian.org/games-team/cegui-mk2/-/blob/master/debian/copyright).

olekw:
Although, note that for the bootstrap package it will be in `bazel-bootstrap` rather than `bazel` but that's a tweak we can make in Debian. I think the focus of efforts here should be (as it is) on the eventual full Bazel package in Debian, not this temporary bootstrap version.

tools/distributions/debian/bazel_debian_license
```diff
@@ -0,0 +1,2 @@
+Please find the license file of Bazel at /usr/share/doc/bazel/copyright
```
meisterT, line 1: How is `/usr/share/doc/bazel/copyright` created?

olekw: By hand, unfortunately. Although we do have [some tools](https://packages.debian.org/sid/licensecheck) to make it (slightly) less painful.

olekw: When I started working on Debian I had no idea how much copyright law was involved! ;) Or how important it is...

meteorcloudy:
@aiuto , the LICENSE file in Debian should be the same as Bazel's, except it will be manually generated. I don't see that as a problem. 

Can you add compliance people to the review? I'm not sure how to do that.

meteorcloudy:
And to clarify, this change only affects the content printed by `bazel license` command.

meteorcloudy:
@olekw , to avoid potential license issue mentioned by @aiuto , can we store a simple patch in the Debian side instead?
This patch will do: https://github.com/meteorcloudy/bazel/commit/51ca9268d731a1fecab2ac7d70963c806614d687

olekw:
> @olekw , to avoid potential license issue mentioned by @aiuto , can we store a simple patch in the Debian side instead?
> This patch will do: [meteorcloudy@51ca926](https://github.com/meteorcloudy/bazel/commit/51ca9268d731a1fecab2ac7d70963c806614d687)

Absolutely! That's a tiny patch and I think it's perfectly reasonable for us to implement it downstream.

meteorcloudy:
Awesome, then I'll close this PR!

PR #11317, opened 2020-05-06
Add javax.annotations to Debian distribution build

olekw:
Extending the Debian integration that @meteorcloudy started based on [clarification](https://lists.debian.org/debian-devel/2020/05/msg00055.html) and current [package location](https://packages.debian.org/sid/all/libgeronimo-annotation-1.3-spec-java/filelist).

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -424,11 +424,12 @@ filegroup(
 
 # javax.annotation.Generated is not included in the default root modules in 9,
 # see: http://openjdk.java.net/jeps/320.
-java_import(
+distrib_java_import(
     name = "javax_annotations",
     jars = ["javax_annotations/javax.annotation-api-1.3.2.jar"],
     neverlink = 1,  # @Generated is source-retention
     srcjar = "javax_annotations/javax.annotation-api-1.3.2-sources.jar",
+    enable_distributions = ["debian"],
 )
 
 java_import(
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Nice!
Did you test the build with `--define=distribution=debian`?

olekw:
> Did you test the build with `--define=distribution=debian`?

Unfortunately not because I can't get a successful Debian build until all the build-dependencies are in place. However, I have added that flag to the packaging. I'm hoping to push a semi-stable `olek1` branch today.

meteorcloudy:
It's actually very simple to test if those targets already available for replacing work or not.
First, install all packages listed in debian_java.BUILD and debian_cc.BUILD
Second, run `bazel build //src:bazel_nojdk --define=distribution=debian`

meteorcloudy:
To be sure, it's using the system library, you can even delete the related jar files from third_party directory.

olekw:
Actually, I've already removed *all* of the jar files from the source package which means I'm not going to be able to get a good build until I replace them. ;) But, at least I'm starting from a good clean source package. Changing your source package after you've already done packaging on top of it is very painful...

olekw:
You may want to actually just pull the patches from the Debian packaging I've done instead of using this PR. I think I've added 20+ more jar files to that patch...

meteorcloudy:
Oh, yes, this is already covered by https://github.com/bazelbuild/bazel/pull/11353/files#diff-109b1868455a98f6f45066f8a24e6fd6R155

PR #11324, opened 2020-05-07
Fix headers checking mode for cc_common.compile

loshca:
Looks like Bazel should always use STRICT headersCheckingMode.

cc_binary & cc_library get this mode from semantic, but for cc_common.compile LOOSE mode used as a default value in CcCompilationHelper.

It leads to extra looseIncludeDirs and some headers checking are missed.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
@@ -1743,7 +1743,8 @@ public abstract class CcModule
             .addAdditionalCompilationInputs(
                 Sequence.cast(additionalInputs, Artifact.class, "additional_inputs"))
             .addAditionalIncludeScanningRoots(headersForClifDoNotUseThisParam)
-            .setPurpose(common.getPurpose(getSemantics()));
+            .setPurpose(common.getPurpose(getSemantics()))
+            .setHeadersCheckingMode(getSemantics().determineHeadersCheckingMode(actions.getRuleContext()));
     if (disallowNopicOutputs) {
       helper.setGenerateNoPicAction(false);
     }


loshca:
@oquenchil 
Can I do something to keep these changes or rework it somehow to solve the initial problem?

oquenchil:
It needs an incompatible flag and I'm working on it. It should be checked in next week. Thanks!

oquenchil:
Can you please give me an example of how you triggered this? I'm trying to write a test for the incompatible flag but I can't reproduce the error. It always complains about the missing header.

oquenchil:
As far as I can see we never enter the part in CppCompileAction that would discover undeclared headers when the checking mode is LOOSE because we return [here](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java;drc=b77bf3c7f8d2872ce7ed33f063ccf1cfc1ab22aa;l=559).

That's always false in Bazel: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java;drc=1bc736e9e49b2dbea0cc36f682a370503dc28c1b;l=73

Can you show me an example of how you triggered this? I'm writing BUILD targets but it always fails even with LOOSE.

loshca:
I've just tried to reproduce it with a small example outside my project environment, and couldn't. But I still can reproduce it with the same example inside my project workspace. The project on windows and with a custom msvc toolchain if it helps.
I am investigating the difference.

oquenchil:
I have figured out the difference just now. It's whether sandboxing is enabled. You can only triggered this when it's disabled. I will continue working on the change introducing the incompatible flag. Thanks for reporting this!

c-mita:
I've created #11677 for this problem and added the flag `--incompatible_force_strict_header_check_from_starlark` in 1b22a24. Thanks for finding this.

PR #11334, opened 2020-05-08
Fix windows --watchfs

### LINKED ISSUE #1931: Windows: support --watchfs
Everything is in the title :) /cc @dslomov

tomdegoede:
As discussed in #1931 --watchfs was made a no-op on windows in d04d7baa1605c8c3a637255bd61c00dba73ad800. The reason for this is that it will hold locks for the (parent-directories of) directories it is watching. Since we recursively register the watcher this leads to locks on folders we do not want to be locked.

The issue is known and tracked at https://bugs.openjdk.java.net/browse/JDK-6972833 but is marked as a Won't Fix. The suggested fix is to instead use ExtendedWatchEventModifier.FILE_TREE which allows a single WatchKey to recursively watch a folder on windows.

This pull request implements the FILE_TREE flag for the root folder that is being watched, and disables the registration of the recursive folders. I have tried to keep the change as small as possible, but I can imagine that we wish to separate this logic to dedicated windows class similar to the OSX variant.

Fixes #1931

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java
@@ -52,8 +52,18 @@ public abstract class LocalDiffAwareness implements DiffAwareness {
         help =
             "On Linux/macOS: If true, %{product} tries to use the operating system's file watch "
                 + "service for local changes instead of scanning every file for a change. On "
-                + "Windows: this flag is a non-op.")
+                + "Windows: this flag currently is a non-op but can be enabled in conjunction "
+                + "with --experimental_windows_watchfs.")
     public boolean watchFS;
+    @Option(
+        name = "experimental_windows_watchfs",
+        defaultValue = "false",
+        documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
+        effectTags = {OptionEffectTag.UNKNOWN},
+        help =
+            "If true, experimental Windows support for --watchfs is enabled. Otherwise --watchfs"
+                + "is a non-op on Windows. Make sure to also enable --watchfs.")
+    public boolean windowsWatchFS;
   }
 
   /** Factory for creating {@link LocalDiffAwareness} instances. */
--- a/src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java
@@ -55,6 +55,7 @@ public abstract class LocalDiffAwareness implements DiffAwareness {
                 + "Windows: this flag currently is a non-op but can be enabled in conjunction "
                 + "with --experimental_windows_watchfs.")
     public boolean watchFS;
+    
     @Option(
         name = "experimental_windows_watchfs",
         defaultValue = "false",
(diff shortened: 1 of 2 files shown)


ulfjack:
Also needs documentation.

tomdegoede:
@ulfjack definitely agree. I'm not that familiar with the architecture yet but can look into propagating a flag next to --watchfs. Would you agree with --experimental_windows_watchfs?

ulfjack:
That sounds fine. Let me know if you need help with the flag.

tomdegoede:
@ulfjack I've added the flag and chosen for it to only work in conjunction with --watchfs so the experimental flag can safely be removed in the future. Is there any other place that requires documentation to be updated or does the flag help suffice?

src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java
```diff
@@ -52,8 +52,18 @@
         help =
             "On Linux/macOS: If true, %{product} tries to use the operating system's file watch "
                 + "service for local changes instead of scanning every file for a change. On "
-                + "Windows: this flag is a non-op.")
+                + "Windows: this flag currently is a non-op but can be enabled in conjunction "
+                + "with --experimental_windows_watchfs.")
     public boolean watchFS;
+    @Option(
```
ulfjack, line 58: style nit: add empty line between `watchFS` and `@Option`

src/main/java/com/google/devtools/build/lib/skyframe/WatchServiceDiffAwareness.java
```diff
@@ -115,7 +116,23 @@ public View getCurrentView(OptionsProvider options) throws BrokenDiffAwarenessEx
     Set<Path> modifiedAbsolutePaths;
     if (isFirstCall()) {
       try {
-        registerSubDirectoriesAndReturnContents(watchRootPath);
+        // Due to a known issue nested watches may result in errors on windows: https://bugs.openjdk.java.net/browse/JDK-6972833
+        // Therefore on windows we register using the special ExtendedWatchEventModifier.FILE_TREE
+        // This watches a folder recursively so there is no need to apply this ourselves
+        if(OS.getCurrent() == OS.WINDOWS) {
```
ulfjack, line 122: It would be good to call OS.getCurrent() once in the constructor, store it in a field, and then use that here.

meteorcloudy:
Thanks for fixing `--watchfs` on Windows. I wonder if we can add a simple test for this?

tomdegoede:
Maybe replicating the MacOSXFsEventsDiffAwarenessTest is a good option. The test cases that it contains are ignored due to flakyness however. I cannot find any test scenarios for the generic WatchServiceDiffAwareness. Do you think we should introduce these or leave it untested as on other platforms?

I'm quite skeptic for introducing flakyness again due to relying on the OS to handle watching. Having used the flag for some time now I occasionally get the overflow message (I assume due to too many file changes) which bazel handles gracefully by assuming EVERYTHING_MODIFIED.

meteorcloudy:
OK. Since this feature is guarded by an experimental flag, I think it's probably fine.

meteorcloudy:
I getting this build error internally:
```
src/main/java/com/google/devtools/build/lib/skyframe/WatchServiceDiffAwareness.java:131: warning: [rawtypes] found raw type: Kind
                  new WatchEvent.Kind[] {
                                ^
  missing type arguments for generic class Kind<T>
  where T is a type-variable:
    T extends Object declared in interface Kind
error: warnings found and -Werror specified
```
Can you help fix?

meteorcloudy:
Thanks!

tomdegoede:
YW! Not a Java expert but I think this could work :)

PR #11335, opened 2020-05-08
Upgrade some headings so they appear in the navbar

gregestren:


gregestren:
IMO these are worthy of being able to find directly in a navbar search.

PR #11336, opened 2020-05-08
Rename UnloadedToolchainContextKey to ToolchainContextKey.

katre:
Cleanup leading to toolchain transitions, #10523.


src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -485,11 +485,11 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
             configuration.getFragmentsMap().keySet(),
             BuildOptions.diffForReconstruction(defaultBuildOptions, toolchainOptions));
 
-    Map<String, UnloadedToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
+    Map<String, ToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
```
juliexxia, line 488: Should these relevant var names be changed to, e.g., toolchainContextKeys? Does UnloadedToolchainContext still exist? Sorry if I'm missing context!

src/test/java/com/google/devtools/build/lib/analysis/util/BuildViewForTesting.java
```diff
@@ -501,19 +501,19 @@ public RuleContext getRuleContextForTesting(
     SkyFunctionEnvironmentForTesting skyfunctionEnvironment =
         skyframeExecutor.getSkyFunctionEnvironmentForTesting(eventHandler);
 
-    Map<String, UnloadedToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
+    Map<String, ToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
```
juliexxia, line 504: same renaming comment here and below

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
katre: UnloadedToolchainContext still exists, but yes, these should be renamed.

src/test/java/com/google/devtools/build/lib/analysis/util/BuildViewForTesting.java
katre: Done.

PR #11337, opened 2020-05-08
Add key() method to ToolchainContext.

katre:
Based on #11336 

Cleanup leading to toolchain transitions, #10523.


src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -485,11 +485,11 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
             configuration.getFragmentsMap().keySet(),
             BuildOptions.diffForReconstruction(defaultBuildOptions, toolchainOptions));
 
-    Map<String, UnloadedToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
+    Map<String, ToolchainContextKey> unloadedToolchainContextKeys = new HashMap<>();
```
juliexxia, line 488: or maybe this PR is the right time to change these var names? (:

katre: Fixed in #11336.

Issue #11339, opened 2020-05-09
Bazel doesn't cleanup partial input downloads for remote_download_minimal


sohaibiftikhar:
### Description of the problem / feature request:

The build without the bytes feature downloads remotely located output files locally only if needed to execute an action locally. This is done as follows:
1. It downloads the file at the exact output path where the actual action would have generated it hence making it transparent to the local action where the previous action was executed. 
2. It [marks the permissions](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java;l=189;drc=f17c403b2365ff0268b0ecae2e00e95fcad9d36b?q=RemoteActionInputFetcher&ss=bazel) of the local file to 0755.
3. On completion of the action, the input files are [removed from the local disk](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java;drc=f17c403b2365ff0268b0ecae2e00e95fcad9d36b;l=600?q=RemoteModule&ss=bazel) to make the local action cache in sync with the local output base.

However, since Bazel executes actions in parallel if any other build action fails this cancels all other concurrent actions using the thread executor service (I assume). This can happen at multiple points:
<a name="scenario1"></a>
1. During/Before (1) above. This leads to an incomplete download. On the next run bazel sees that the digest of the output does not match the remotely expected digest and hence reruns the action. This is okay.
<a name="scenario2"></a>
2. After (1) above. Since the file is completely downloaded and the digest does not take permissions into account, Bazel assumes the file is completely downloaded and attempts to execute the action using the file as an input. **However, this is not okay.** Some actions like `ExecutableSymlink` expect the file to be executable before the action is executed on it and can fail.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
<a name="affected_versions"></a>
The bug is very hard to reproduce. There are a few relatively easier ways to reproduce it.
1. Take bazel_version < 1.2.0.
2. Revert the [context propagation to the Grpc Server](https://github.com/bazelbuild/bazel/commit/fdfd88d3e9885f4f28d1ce8f2cee4461d6cb0ef4) in bazel version after 1.2.0.
3. Take a commit from the master after [this commit](https://github.com/bazelbuild/bazel/commit/3ef8fb98e7becaac160003f5c24f35914b764bb9). Because it seems to have broken context propagation once again.
[...]

PR #11340, opened 2020-05-09
Change RemoteActionInputFetcher to use temp files before staging

sohaibiftikhar:
Attempts fixing #11339

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
@@ -33,6 +33,7 @@ import com.google.devtools.build.lib.profiler.ProfilerTask;
 import com.google.devtools.build.lib.profiler.SilentCloseable;
 import com.google.devtools.build.lib.remote.common.CacheNotFoundException;
 import com.google.devtools.build.lib.remote.util.DigestUtil;
+import com.google.devtools.build.lib.vfs.FileSystemUtils;
 import com.google.devtools.build.lib.remote.util.TracingMetadataUtils;
 import com.google.devtools.build.lib.sandbox.SandboxHelpers;
 import com.google.devtools.build.lib.vfs.Path;
@@ -155,6 +156,12 @@ class RemoteActionInputFetcher implements ActionInputPrefetcher {
     }
   }
 
+  private static Path toTmpDownloadPath(Path actualPath) {
+    return actualPath.getParentDirectory().getRelative(
+      actualPath.getBaseName() + "."  + actualPath.hashCode()
+    );
+  }
+
   private ListenableFuture<Void> downloadFileAsync(Path path, FileArtifactValue metadata)
       throws IOException {
     synchronized (lock) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
```diff
@@ -158,6 +159,12 @@ void downloadFile(Path path, FileArtifactValue metadata)
     }
   }
 
+  private static Path toTmpDownloadPath(Path actualPath) {
+    return actualPath.getParentDirectory().getRelative(
+      actualPath.getBaseName() + "."  + actualPath.hashCode()
```
sohaibiftikhar, line 164: This should give a unique + deterministic enough name so that we don't need to [sort all downloads before doing this](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java;l=406;drc=f17c403b2365ff0268b0ecae2e00e95fcad9d36b?q=RemoteCache&ss=bazel%2Fbazel).

src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
```diff
@@ -173,23 +180,30 @@ void downloadFile(Path path, FileArtifactValue metadata)
...
                     logger.atWarning().withCause(e).log("Failed to chmod 755 on %s", path);
                   }
+                  try {
+                    FileSystemUtils.moveFile(tmpPath, path);
+                  } catch (IOException e) {
+                    logger.atWarning().withCause(e).log(
+                      "Failed to move to final location %s. -> %s", tmpPath, path
+                    );
+                  }
+                  synchronized (lock) {
```
sohaibiftikhar, line 203: Moved the synchronize block below. A file is considered to be downloaded only after its permissions are set and it has been moved to its final location.

meisterT:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
meisterT: I think this is wrong. You are doing this even when the above code fails with an IOException

src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java
ulfjack: Why not use the same approach as RemoteCache?

sohaibiftikhar: Because RemoteCache collects all outputs before doing the [moveToFinalLocation](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java;l=372?q=RemoteCache&ss=bazel). This class was designed a bit differently in that it can call downloadFileAsync from two different methods ([downloadFile](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java;l=149?q=RemoteActionInputFetcher&ss=bazel) and [prefetchFiles](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java;l=89?q=RemoteActionInputFetcher&ss=bazel)) and the shared variables `downloadsInProgress` and `downloadedPaths` are controlled per download here.

So I would have to move the `synchronized` blocks to those two methods and that looked like a lot more refactoring to me. I can make that change if you think that is the way to go. But this seems cleaner to me.

sohaibiftikhar:
> Also, did you consider rm-ing the file after download instead of downloading to a temporary location.

There is code written in the `[onFailure](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteActionInputFetcher.java;l=201?q=RemoteActionInputFetcher&ss=bazel)` here block that does the delete. The problem is that this never gets called in case the thread gets terminated from the outside. To be honest I don't quite know why. I couldn't find the place in the code where the termination happens.

Maybe someone with more knowledge of that part could point that out.

jmmv:
I went through various iterations to make the "download as temp" feature work (for the internal version of this feature).

1. Put the temp file next to the output, with a `.tmp` suffix. This ended up causing trouble with tree artifacts (which would see "extra" files on disk before they were deleted, and failed validation).

1. Put the temp files in the actions temp dir, using a file name derived from the original execroot-relative path but translating all path separators into underscores. This ended up running into path limits (on macOS).

1. Put the temp files in the actions temp dir, using a counter for their file names. This has been working well so far.

So essentially, we do this:

```java
Path tempDestFile =
    actionTempsDir.getChild("output-" + lastTempOutput.getAndAdd(1) + ".tmp");
```

More details here: https://jmmv.dev/2019/12/bazel-dynamic-execution-download-times.html

Would be great if the external logic of this code did the same.

philwo:
@sohaibiftikhar Are you still working on this?

sohaibiftikhar:
I was involved with some other work that put this on a sidetrack. I can finish this by aligning with what @jmmv proposes here.

meisterT:
What's the current state?

meisterT:
cc @coeuvre

coeuvre:
#12453 fixes #11339 by deleting the partial downloaded files. However, changing how `RemoteCache` handles the temp download files to what @jmmv suggested would be great.

jin:
Closing this PR as as the referenced issue has been fixed.

Issue #10690, opened 2020-01-31
Crash for --compilation_mode in transition


moroten:
### Description of the problem

`//command_line_option:compilation_mode` exposes a Java class when used as input in a transition, not a string. It is possible to set it to a string as output, but not to read the input.

The same problem also applies for `//command_line_option:run_under` and possibly more.

I would prefer a quick fix for `compilation_mode` whereas `run_under` is not such a big problem for me (just annoying).

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
touch WORKSPACE

cat <<EOF > BUILD.bazel
load(":rules.bzl", "rule_with_transition")

rule_with_transition(
    name = "my_target",
)
EOF

cat <<EOF > rules2.bzl
def _my_transition_impl(settings, attr):
    print(settings)
    return dict(settings)

inputs_and_outputs = [
    "//command_line_option:cpu",
    "//command_line_option:compilation_mode",
    "//command_line_option:platform_suffix",
]

my_transition = transition(
    implementation = _my_transition_impl,
    inputs = inputs_and_outputs,
    outputs = inputs_and_outputs,
)

def _rule_impl(ctx):
    pass

rule_with_transition = rule(
    implementation = _rule_impl,
    cfg = my_transition,
    attrs = {
        "_whitelist_function_transition": attr.label(default = "@bazel_tools//tools/whitelists/function_transition_whitelist"),
    },
)
EOF

bazel build :my_target
```

### What operating system are you running Bazel on?

Linux (Ubuntu 17.04)

### What's the output of `bazel info release`?

`release 2.0.0`

###  Have you found anything relevant by searching the web?

GitHub issues: https://github.com/bazelbuild/bazel/issues
- Nothing found.

### Any other information, logs, or outputs that you want to share?

```
Internal error thrown during build. Printing stack trace: java.lang.IllegalArgumentException: cannot expose internal type to Starlark: class com.google.devtools.build.lib.analysis.config.CompilationMode
	at com.google.devtools.build.lib.syntax.Starlark.fromJava(Starlark.java:122)
	at com.google.devtools.build.lib.syntax.Dict.putAllUnsafe(Dict.java:350)
	at com.google.devtools.build.lib.syntax.Dict.plus(Dict.java:531)
	at com.google.devtools.build.lib.syntax.MethodLibrary.dict(MethodLibrary.java:584)
[...]

Issue #10875, opened 2020-02-29
Reuse input map and merkle tree for tools with lots of inputs


JaredNeil:
### Description of the feature request:
Create a separate input map and merkle tree for each tool used in an action, then combine those to calculate the final action digest. Each tool's input map and merkle tree should be reused between actions that depend on the same tool.

### Feature requests: what underlying problem are you trying to solve with this feature?
[These two function calls](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnCache.java#L119-L121) cause remote caching to be really slow for actions with lots of inputs. Having a way to re-use the parts that don't change between multiple actions that use the same tool could make these significantly faster.
Tools written in JavaScript will often have a high number of runfiles (>10,000) because they depend on the `node_modules` folder. This is also sometimes true of Python tools.

###  Have you found anything relevant by searching the web?

- #4839 from 2018 is a different option to potentially speed up remote caching, but looks like it was decided that it wouldn't offer much benefit.
- This [bazel-discuss thread](https://groups.google.com/forum/#!topic/bazel-discuss/nF2fTkHiWy8) is what prompted me to look into the code, and see if/how this could be improved.

### Any other information, logs, or outputs that you want to share?

In our specific case, we have an action per source file that uses a `nodejs_binary` tool. So we have 15,000 actions with the same ~10,000 runfiles for the tool and 1 unique input file. Each of these takes ~200ms to calculate the cache key, so that's 50 CPU-minutes of work just to check if the actions are cached.

![Screenshot of profile showing Merkle Tree build times](https://user-images.githubusercontent.com/1205857/75600823-79d49280-5a71-11ea-896e-2a8f31219b66.png)

moroten:
I've been profiling our Bazel and remote execution when building C/C++ code. We have experimented with a packaged C/C++ toolchain, but 10000 extra input files made `MerkleTree.build` very slow. The numbers below are with a system installed toolchain.

Bazel 2.2.0 gives me average self time of (listed numbers above 1 ms):
```
Remote.download                       35 ms (network latency)
check cache hit                       17 ms (network latency)
MerkleTree.build                      11 ms (cpu)
AbstractSpawnStrategy.getInputMapping  3 ms (cpu)
```
`14ms/action * 100k actions / 4 CPU cores = 6 minutes` for a fully cached build.

[...]

Issue #7477, opened 2019-02-20
Adjustable `outputDirectoryName`


excitoon:
### Description of the problem / feature request:

Hi. I wonder if one can make `outputDirectoryName` (these `{cpu}-{compilation_mode}` parts of paths) for the build adjustable from command line. It would help to integrate Bazel into other pipelines because it is necessary to know where output files are. I can run `bazel info`/`bazel cquery`, but that will involve extracting, loading and/or configuring, I guess, and therefore it looks like extra overhead.

###  Have you found anything relevant by searching the web?

https://stackoverflow.com/questions/51292429/bazel-how-to-get-path-to-output-binary
https://stackoverflow.com/questions/47859615/bazel-how-do-you-get-the-path-to-a-generated-file

gregestren:
You can use [`--platform_suffix`](https://docs.bazel.build/versions/master/command-line-reference.html#flag--platform_suffix) to add a customizable string to the *end* of that. But you can't replace `{cpu}-{compilation_mode}` with, say, a string literal because Bazel *has* to support building the same output for different CPUs in the same build, and each needs a distinct path.

Depending on exactly which outputs you need, can you just look at `blaze-bin`?

Also see https://github.com/bazelbuild/bazel/issues/6526 re: plans to change this naming completely. But that's a longer term effort.

excitoon:
Thank you for pointing to the issue, @gregestren . Will that cover scenario when we gather artifacts from different platforms and place it to some kind of universal archive (in almost cpu-agnostic way)?

gregestren:
Essentially yes.

The basic problem is Bazel's logic predates concern for multi-platform builds, so it just doesn't address that problem very well. The issue I mentioned, as well as other work, recognizes we need better interfaces for these sorts of problems. So that'll happen - but they''ll need to be properly designed.

Out of curiosity, what kind of multi-platform builds are you interested in? We're doing a lot of work to improve multi-platform support and some cool new features are coming down the pipeline even now.

excitoon:
@gregestren Thank you for asking. In most easiest case we will need to build some C++ target for `x64` and `x86` and use both versions of it say in C++ program for `x64`. Would that be possible?

gregestren:
Essentially yes. You may also want to review https://bazel.build/roadmaps/configuration.html, which explores new ways to support multiple CPUs in the same binaries.

excitoon:
I leave this hack here as a quick workaround:
```
--platform_suffix=/..
[...]

Issue #6526, opened 2018-10-25
Multiplatform output paths are safe, correct, and efficient


gregestren:
Tracking issue on [Bazel Configurability Roadmap](https://bazel.build/roadmaps/configuration.html)

By "multiplatform" I mean any scenario where two different rules in the same build build with different settings. This also includes non-platform settings like *app version*, but "multiplatform" is a concise term to capture the essence.

Long-story short is `bazel-out/$(cpu)-$compilation_mode)/...` doesn't work well for multiplatform builds:

* Unrelated actions can inadvertently write to the same output path (correctness issue)
* `cpu` is redundant for cpu-agnostic actions (efficiency issue: switching up the CPU shouldn't require re-executing these actions)
* Actions that depend on flags that *aren't* CPU or compilation mode write to the same path when those flags change (correctness issue) 
* All the above can destroy remote execution efficiency

This issue tracks the long and complicated effort of making a better output path syntax. Expect the next deliverable on this to be a design doc.

gregestren:
**April '19 update:**

Detailed plans at [Experimental Content-Based Output Paths](https://docs.google.com/document/d/17snvmic26-QdGuwVw55Gl0oOufw9sCVuOAvHqGZJFr4/edit) (please comment!).

Goal is to get an `--experimental` prototype available this summer that automatically caches multiplatform Java compilation.

gregestren:
P1 issue review: still relevant, still very much intend to explore this but I simply am unable to put time into it at the moment. Hoping I can pick this up next quarter.

gregestren:
This is being up-prioritized with about ~1 dev's full-time commitment over the next 3 months.

michaelmartak:
Is there an escalation path at Google (i.e., someone we could reach out to) that could help align on business priorities?

gregestren:
Write to me as a technical contact (gregce@bazel.build) explaining your needs as best you can.  I'm happy to chat technical concerns and CC in folks who can help with business priorities.

gregestren:
To elaborate on https://github.com/bazelbuild/bazel/issues/6526#issuecomment-658927627,

I believe the generic solution described in this issue and https://github.com/bazelbuild/bazel/issues/6526#issuecomment-488103473 is nuanced enough that it'd have to go through a long experimental phase before we could consider productionizing parts. 

I'd still like to get to that phase because it would still let interested folks opt in, explore, and help evolve its path. 

[...]

PR #11347, opened 2020-05-10
Add --compilation_mode support in transitions

moroten:
Add support for converting //command_line_option:compilation_mode from
Java value to Starlark value so that it can be read in a transition.

Fixes --compilation_mode in #10690, but not the general case.

This patch does not contain any tests. What kind of test is expected and where should it be implemented?

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java
+++ b/src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java
@@ -518,6 +518,92 @@ public class StarlarkAttrTransitionProviderTest extends BuildViewTestCase {
     assertThat(getConfiguration(Iterables.getOnlyElement(dep)).getCpu()).isEqualTo("armeabi-v7a");
   }
 
+  private void writeReadAndPassthroughOptionsTestFiles() throws Exception {
+    setBuildLanguageOptions("--experimental_starlark_config_transitions=true");
+    writeAllowlistFile();
+
+    scratch.file(
+        "test/skylark/my_rule.bzl",
+        "load('//myinfo:myinfo.bzl', 'MyInfo')",
+        "settings_under_test = {",
+        "  '//command_line_option:cpu': 'armeabi-v7a',",
+        "  '//command_line_option:compilation_mode': 'dbg',",
+        "  '//command_line_option:crosstool_top': '//android/crosstool:everything',",
+        "  '//command_line_option:platform_suffix': 'my-platform-suffix',",
+        "}",
+        "def set_options_transition_func(settings, attr):",
+        "  return settings_under_test",
+        "def passthrough_transition_func(settings, attr):",
+        // All values in this test should be possible to copy within Starlark.
+        "  ret = dict(settings)",
+        // All values in this test should be possible to read within Starlark.
+        // This does not mean that it is possible to set a string value for all settings,
+        // e.g. //command_line_option:test_arg should be set to a list of strings.
+        "  for key, expected_value in settings_under_test.items():",
+        "    if str(ret[key]) != expected_value:",
+        "      fail('%s did not pass through, got %r expected %r' %",
+        "        (key, str(ret[key]), expected_value))",
+        "  ret['//command_line_option:test_arg'] = ['ok']",
+        "  return ret",
+        "my_set_options_transition = transition(",
+        "  implementation = set_options_transition_func,",
+        "  inputs = [],",
+        "  outputs = settings_under_test.keys())",
+        "my_passthrough_transition = transition(",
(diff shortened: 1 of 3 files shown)


aiuto:
I would look at the tests in src/test/j.c.g/devtools/build/lib to find examples of how we test for this.

@alandonovan You had comments in #10690

juliexxia:
I think StarlarkAttrTransitionProviderTest would be a good place to write a test where the set up would be a transition that transitions on //command_line_option:compilation_mode and asserts that the transitioned value is as expected

Here's a specific test case that is somewhat close to that (no need to make it a split transition) https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java;l=120

sdtwigg:
The broader issue here has come up in conversation before where our current Starlark-Java interface for these commandline options in transitions easily allows Java objects without a well-defined Starlark API to leak into the Starlark interpreter context.

I'm hesitant to approve piecemeal fixes since they could interfere with future broad fixes.

The (rough and currently not being worked on) plan for this is to add another interface that options parsing Converters can inherit, tentatively called StarlarkConvert. It would contain toStarlark (T -> Object) and fromStarlark (Object -> T) methods. The transitions infrastructure would be amended to use these methods when passing values to and from Starlark code (options not inheriting that interface cannot be used in transitions). There are a bunch of commonly used Converters that, if amended, would immediately apply the feature to a majority of options.

moroten:
Rebased and added a test.

I agree with @sdtwigg that a broader solution would be welcome, but the time plan for it is unfortunately suffering.

In this PR, I only address `//command_line_option:compilation_mode`. Being able to read `cpu`, `compilation_mode` and `platform_suffix` helps when setting a manual output directory name (see #7477), which can improve cache hit ratio before #6526 has been solved.

gregestren:
@moroten which cache miss scenarios are you trying to address?

moroten:
The cache miss scenario is when building the following:
```
bazel test --platforms=//:my_armv8_platform --run_under=qemu //...
bazel build //:release_package
```
The `//:release_package` does a split transition to put binaries for many platforms in a `.tar.gz` file, including documentation. By setting `//command_line_option:output directory name` manually in the transition, the second command can reuse the compiled object files and libraries from the first command. (Ideally, we would like to add test and coverage report in the package as well.)

Annoyingly, the analysis cache is discarded between these commands so checking for cache hits remotely takes some time (see #10875). Running both steps in a single command is not possible due to the `//...` target pattern and because you can't depend (and apply a transition) on test and coverage results.

moroten:
By enabling `//command_line_option:compilation_mode` as an input, I can allow the user to build the release package using either `--compilation_mode=opt` or `--compilation_mode=fastbuild` and still produce the matching `//command_line_option:output directory name`.

gregestren:
I see, thanks.

@sdtwigg would this PR actually interfere with a broader solution? As ad hoc changes go it's not terribly hackish, and I don't see this opening floodgate exceptions for all the other options. If it's this vs. nothing maybe it's worthwhile.

sdtwigg:
After the weekend mental refresh, I am less concerned that this would interfere with the broader solution.

That said, can you add a TODO (you can link back to this bug or specifically my comment before) that other places should _strongly avoid_ replicating this fix and look in to implementing the proper fix for other desired flags.

meisterT:
What's the state of this PR?

moroten:
I am sorry I missed @sdtwigg's last comment. I have now added the missing TODO in the code.

sdtwigg:
Looking into this.

src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java
```diff
@@ -514,6 +514,92 @@ public void testOptionConversionCpu() throws Exception {
     assertThat(getConfiguration(Iterables.getOnlyElement(dep)).getCpu()).isEqualTo("armeabi-v7a");
   }
 
+  private void writeReadAndPassthroughOptionsTestFiles() throws Exception {
+    setStarlarkSemanticsOptions("--experimental_starlark_config_transitions=true");
+    writeWhitelistFile();
```
sdtwigg, line 519: I think this line specifically out of date. Need to amend this so the tests still pass and I can import this into Bazel.

moroten: I've rebased and Buildkite run the tests.

gregestren: Thanks for rebasing, Fredrik. @sdtwigg ready to merge?

PR #11348, opened 2020-05-10
Add --output_base_suffix as startup option

moroten:
Simplify how to specify an output_base parallel to the default one or the output_base specified in the user .bazelrc file. Use case include using different `--output_base_suffix` depending on configuration to avoid discarding the analysis cache or to avoid interrupting a long running test.

I know this patch does not contain any tests, which of course is needed if this should be merged at all. 

To start with, is this a reasonable feature to add?

### DIFF
--- a/src/main/cpp/startup_options.h
+++ b/src/main/cpp/startup_options.h
@@ -142,6 +142,11 @@ class StartupOptions {
   // Blaze's output base.  Everything is relative to this.  See
   // the BlazeDirectories Java class for details.
   blaze_util::Path output_base;
+  std::string output_base_option;
+
+  // Suffix to append to Blaze's output base, e.g. `-extra` or `/../no-hash`.
+  // See the BlazeDirectories Java class for details.
+  std::string output_base_suffix;
 
   // Installation base for a specific release installation.
   std::string install_base;
--- a/src/main/cpp/bazel_startup_options.cc
+++ b/src/main/cpp/bazel_startup_options.cc
@@ -82,21 +82,36 @@ void BazelStartupOptions::MaybeLogStartupOptionWarnings() const {
   }
   bool output_user_root_has_space =
       output_user_root.find_first_of(' ') != std::string::npos;
+  bool output_base_has_space =
+      output_base_option.find_first_of(' ') != std::string::npos;
+  bool output_base_suffix_has_space =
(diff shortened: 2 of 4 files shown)


aiuto:
Please start with an issue describing the feature and we can debate a design there.

moroten:
@aiuto Thank you for your reply. I've created #11435, so please close this PR if you prefer to keep the list of open PRs short.

aiuto:
Thanks for creating the issue. We can leave this alive. I'll mark it WIP so we don't keep trying to assign a reviewer right now.

philwo:
@moroten I think this is a useful change with localized impact - would you mind adding a test, then I'd go ahead and merge this. :)

src/main/cpp/startup_options.h
```diff
@@ -142,6 +142,11 @@ class StartupOptions {
   // Blaze's output base.  Everything is relative to this.  See
   // the BlazeDirectories Java class for details.
   blaze_util::Path output_base;
+  std::string output_base_option;
+
+  // Suffix to append to Blaze's output base, e.g. `-extra` or `/../no-hash`.
```
philwo, line 147: Could you change "append" to "path-append" to make it clear the resulting string will be post-processed into a valid normalized and absolute path?

philwo (changes requested):
Could you please add the startup flag to the Java code, too, similar to this one :https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/runtime/BlazeServerStartupOptions.java;l=132?q=%5C%22output_user_root&ss=bazel

That ensures it's documented and shows up in the help :)

src/main/cpp/startup_options.h
philwo: Oh, and we should absolutely test for these `/../no-hash` up-references in the tests for this 😄

moroten:
I will eventually work on this, but it might take some weeks until I find time as this feature is no high priority for me.

philwo:
We're replacing the "WIP" label with the native GitHub "draft PRs". I thus converted this PR to a draft.

Please mark this as ready for review when it's ready, or close it if you no longer intend to work on it. Thanks! 👍

moroten:
Will not have time to do it in any near future. Therefore, closing.

PR #11364, opened 2020-05-11
Allow reading of MANIFEST file over MAX_PATH

ozio85:
Not using the wide API made it impossible to read the
MANIFEST file on a path longer than 260 chars. This affects
the test wrapper (tw.exe).

Note that this wrunfiles use Bazel internal logic, and therefore
much of the runfiles logic is copied from cpp/runfiles.

Change-Id: I238d644430653aea0b74a892b4441383b95942d1

### DIFF
--- a/tools/test/BUILD
+++ b/tools/test/BUILD
@@ -57,11 +57,17 @@ cc_binary(
 cc_library(
     name = "tw_lib",
     srcs = select({
-        "@bazel_tools//src/conditions:windows": ["windows/tw.cc"],
+        "@bazel_tools//src/conditions:windows": [
+            "windows/tw.cc",
+            "windows/wrunfiles_src.cc",
+        ],
         "//conditions:default": ["empty_test_wrapper.cc"],
     }),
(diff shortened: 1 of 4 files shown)


meteorcloudy:
@ozio85 Thanks for the contribution! But I have two questiones:
1. Can you provide an example where "the MANIFEST file on a path longer than 260 chars". It will be useful to verify this PR actually solves the issue. Ideally we should add a test for that.
2. Can we fix [tools/cpp/runfiles](https://github.com/bazelbuild/bazel/tree/master/tools/cpp/runfiles) instead of duplicating the code?

ozio85:
@meteorcloudy Ah my bad, i thought runfiles_src.cc was created to be standalone .. but i browsed the code available in bazel_tools and all seems to be available.

philwo:
What's the state of this pull request?

meisterT:
Closing this PR for lack of inactivity. Feel free to reply if you want to open it again.

PR #11370, opened 2020-05-12
Explicit description about reproducible rules.

glukasiknuro:
Added small clarification when workspace rule may return None, to signify that the rule is reproducible given the specified parameters, and not that it's always reproducible.


glukasiknuro:
@googlebot I signed it!

Issue #8988, opened 2019-07-25
Make the timeout to start the Bazel server configurable


beasleyr-vmw:
### Description of the problem / feature request:

When launched on a machine under heavy load, the Bazel client may fail to connect to the server after the hardcoded 120s timeout.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

- Give your machine a **lot** of jobs to chew on.
- Launch Bazel.
- Observe the Bazel client terminate with a `FATAL: couldn't connect to server (<pid>) after 120 seconds` message.

### What operating system are you running Bazel on?

CentOS 7

### What's the output of `bazel info release`?

release 0.26.0+vmware

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

na

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

na

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/blob/0.28.1/src/main/cpp/blaze.cc#L729-L732

I'm aware of the `--batch` flag, but it's deprecated.

### Any other information, logs, or outputs that you want to share?

I'm sporadically observing this on VMs in a build farm with ~30 vCPUs which, for dumb reasons I can't get into, need to service multiple build jobs simultaneously. Applications aren't starved for memory---it's a function of the kernel's runqueue; Bazel is just occasionally drawing the short straw.

I would appreciate it if this timeout were made user configurable or user disable-able. We'll likely patch our distribution in the meantime and are open to submitting a pull request.

jmmv:
Huh, so we do have a `--connect_timeout_secs` but it doesn't apply to this hardcoded value, which still exists.

I'm sure we'd take a PR on this if you have one already around.

janakdr:
Can this be closed?

beasleyr-vmw:
Yes, I believe so. Thanks!

PR #11391, opened 2020-05-14
Allow users to override Bazel's 120s connect timeout

beasleyr-vmw:
Bazel constrains starting up and connecting to a local server to 120s.  This
occasionally causes problems for us on heavily loaded, high numcpu machines,
because the Bazel client+server may end up starved out via simple CPU
contention.  (Where we observed startup timeouts, machines had plenty of RAM,
I/O looked normal---procs weren't stuck in disk wait---but the run queues were
60+.)

Mitigate this by introducing a new startup option, `--local_startup_timeout_secs`,
which allows users to specify their own timeout values.  (Note:  I primarily
used `connect_timeout_secs` as a reference.)

TODO: Consult Bazel team to add test case per comments in
[bazel_startup_options_test.cc].

Resolves [#8988].

[bazel_startup_options_test.cc]: https://github.com/bazelbuild/bazel/blob/8075057af6108ebc23c146f18eecec911d4b8c00/src/test/cpp/bazel_startup_options_test.cc#L79-L81
[#8988]: https://github.com/bazelbuild/bazel/issues/8988

Testing Done:
- Induced high load on my MBP, then
```console
$ vbazel --local_startup_timeout_secs=1 info release
FATAL: couldn't connect to server (16290) after 1 seconds.
```

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -778,7 +778,8 @@ static void ConnectOrDie(const OptionProcessor &option_processor,
   // Give the server two minutes to start up. That's enough to connect with a
   // debugger.
   const auto start_time = std::chrono::system_clock::now();
-  const auto try_until_time = start_time + std::chrono::seconds(120);
+  const auto try_until_time = start_time +
+     std::chrono::seconds(startup_options.local_startup_timeout_secs);
   // Print an update at most once every 10 seconds if we are still trying to
   // connect.
   const auto min_message_interval = std::chrono::seconds(10);
@@ -820,7 +821,8 @@ static void ConnectOrDie(const OptionProcessor &option_processor,
     }
   }
   BAZEL_DIE(blaze_exit_code::INTERNAL_ERROR)
-      << "couldn't connect to server (" << server_pid << ") after 120 seconds.";
+      << "couldn't connect to server (" << server_pid << ") after "
+      << startup_options.local_startup_timeout_secs << " seconds.";
 }
 
 // Ensures that any server previously associated with `server_dir` is no longer
@@ -1048,7 +1050,7 @@ static bool IsVolatileArg(const string &arg) {
   // server command line difference logic can be simplified then.
   static const std::set<string> volatile_startup_options = {
       "--option_sources=", "--max_idle_secs=", "--connect_timeout_secs=",
-      "--client_debug="};
+      "--local_startup_timeout_secs=", "--client_debug="};
 
   // Split arg based on the first "=" if one exists in arg.
   const string::size_type eq_pos = arg.find_first_of('=');
--- a/src/main/cpp/startup_options.cc
+++ b/src/main/cpp/startup_options.cc
@@ -82,6 +82,7 @@ StartupOptions::StartupOptions(const string &product_name,
       fatal_event_bus_exceptions(false),
       command_port(0),
       connect_timeout_secs(30),
+      local_startup_timeout_secs(120),
       have_invocation_policy_(false),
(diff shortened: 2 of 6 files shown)


nikola-sh:
Do we really need this timeout? It is a connection between processes on one machine, without unstable transport etc. I think it should be infinite.

src/main/cpp/blaze.cc
```diff
@@ -821,7 +822,8 @@ static void ConnectOrDie(const OptionProcessor &option_processor,
     }
   }
   BAZEL_DIE(blaze_exit_code::INTERNAL_ERROR)
-      << "couldn't connect to server (" << server_pid << ") after 120 seconds.";
+      << "couldn't connect to server (" << server_pid << ") after "
+      << startup_options.local_startup_timeout_secs << " seconds.";
```
janakdr, line 826: why not use try_until_time?

meisterT:
What's the state of this PR?

philwo:
@beasleyr-vmw Friendly ping.

src/main/cpp/blaze.cc
beasleyr-vmw: Ack. Lost this in my inbox.

Hm, I don't understand the suggestion. `try_until_time` is a point in time, but we're displaying an error to the user indicating the timeout _duration_. Am I missing something?

beasleyr-vmw:
Just rebased to master, added the one-liner test suggested by @janakdr and also cobbled together an integration test case.

src/main/cpp/blaze.cc
janakdr: I have no idea now what I was thinking. Sorry!

janakdr (approved):

philwo:
@beasleyr-vmw I prepared your change for merging, but ran into some issues with the newly added test. It seems like the JVM does something weird when we pass `--host_jvm_debug` and it fails to open the listening socket on our remote execution platform (although local networking is allowed there):

```
ERROR: transport error 202: No sockets to listen to: Permission denied
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [src/jdk.jdwp.agent/share/native/libjdwp/debugInit.c:735]
```

It's possible that it's trying to open the socket on all available interfaces instead of just localhost, so I applied this diff to the Bazel invocation in the test, hope that's fine:

```
- --host_jvm_debug
+ --host_jvm_args="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=localhost:41687"
```

Now it passes. :)

beasleyr-vmw:
> It's possible that it's trying to open the socket on all available interfaces instead of just localhost, so I applied this diff to the Bazel invocation in the test, hope that's fine:
> 
> ```
> - --host_jvm_debug
> + --host_jvm_args="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=localhost:41687"
> ```
> 
> Now it passes. :)

Aha! Sorry about that, and good find. Though I suppose there could be other issues if the listening port clashes with something else on the same machine. Not sure if it's rare enough that we should wait until someone complains, find some other mechanism to delay the server, or if we should delete/disable the test.

philwo:
> Though I suppose there could be other issues if the listening port clashes with something else on the same machine.

Yes, you have a good point there. I checked, but `--host_jvm_debug` actually also just uses a hard-coded port number (5005), so the chance of that clashing is probably similar. Let's see how this goes. 😊

PR #11398, opened 2020-05-14
Split ConfigurationTransitionDependency into a new class

katre:
and break relation to the existing Dependency.


katre:
I still don't like the name ConfigurationTransitionDependency. Can't think of anything better, though.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -589,12 +590,12 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
       BuildOptions defaultBuildOptions)
       throws DependencyEvaluationException, ConfiguredTargetFunctionException,
           AspectCreationException, InterruptedException {
-    // Create the map from attributes to set of (target, configuration) pairs.
-    OrderedSetMultimap<DependencyKind, Dependency> depValueNames;
```
gregestren, line 593: I've lost track of what kind of "dependency" descriptor can get created and passed through what parts of the logic.

Originally I thought depValueNames. could get assigned to both configuration transition dependencies and direct ones such as HOST and NULL. is that no longer the case?

To be clear, that's all the more reason for this work you're doing. My main interest in the final outcome is really getting naming that clarifies the current reality as clearly as possible, so I'm excited about this.

katre: I also thought that was the case, but it turns out that it's not true. Every time the Dependency type was used before this change, it was either a ConfigurationTransitionDependency, or one of the NullConfigurationDependency or ExplicitConfigurationDependency. This is why the hasExplicitConfiguration() method is entirely removed: it's not needed.

One of the major motivators of this change is clarifying what data exists at what points during dependency resolution.

katre:
Any further comments?

src/main/java/com/google/devtools/build/lib/analysis/Dependency.java
```diff
@@ -144,14 +144,6 @@ public static Dependency withConfiguredAspects(
         ImmutableList.of());
   }
```
gregestren, line 145: I don't know why I can't add a comment at that line, but let's update the javadocs at line 32 on.

src/main/java/com/google/devtools/build/lib/analysis/ConfigurationTransitionDependency.java
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+package com.google.devtools.build.lib.analysis;
+
+import com.google.auto.value.AutoValue;
+import com.google.devtools.build.lib.analysis.config.transitions.ConfigurationTransition;
+import com.google.devtools.build.lib.cmdline.Label;
+
+/** Implementation of a dependency with a given configuration transition. */
+@AutoValue
+public abstract class ConfigurationTransitionDependency {
```
gregestren, line 22: What about DependencyInfo or DependencyKey? I support, as part of distinguishing this, distinguishing this a clearly as possible in the naming too. As long as it ends in *Dependency" is still reads to me like it's an example of a `Dependency`.

I'd personally be happy with either of the above suggestions. But I also want to acknowledge DependencyInfo could get a bit confused with the existing DependencyKind  and DependencyKey implies the concept of Sky keys. 

But even if not perfect, let's give it a name that doesn't end in "Dependency".

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: Wow, you're right! Can't believe how much this logic has moved past me over time.

src/main/java/com/google/devtools/build/lib/analysis/Dependency.java
katre: Good point, done.

src/main/java/com/google/devtools/build/lib/analysis/ConfigurationTransitionDependency.java
katre: Changed to DependencyKey, let me know how this reads.

src/main/java/com/google/devtools/build/lib/analysis/AnalysisUtils.java
```diff
@@ -203,17 +203,16 @@ public static TopLevelTargetsAndConfigsResult getTargetsWithConfigs(
     // configurations
     // for deps including transitions. So to satisfy its API we resolve transitions and repackage
     // each target as a Dependency (with a NONE transition if necessary).
```
gregestren, line 205: nit: should we update the `repackage each target as a Dependency` comment? Maybe we can just eliminate this whole sentence now, since the explicit typing makes the requirements clearer?

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -589,12 +590,12 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
...
+    // Create the map from attributes to set of (target, transition) pairs.
+    OrderedSetMultimap<DependencyKind, DependencyKey> initialDependencies;
```
gregestren, line 594: Open question (but not for this PR): I wonder if we can even fold DependencyKind into DependencyKey?

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigurationResolver.java
```diff
@@ -118,7 +118,7 @@
   public static OrderedSetMultimap<DependencyKind, Dependency> resolveConfigurations(
       SkyFunction.Environment env,
       TargetAndConfiguration ctgValue,
-      OrderedSetMultimap<DependencyKind, Dependency> originalDeps,
+      OrderedSetMultimap<DependencyKind, DependencyKey> originalDeps,
```
gregestren, line 121: NIt: maybe rename this variable now too?

src/main/java/com/google/devtools/build/lib/analysis/AnalysisUtils.java
katre: Done.

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigurationResolver.java
katre: Done. Also in sortResolvedDeps.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
katre: I don't think it makes sense: the DependencyKind is effectively an enum: it distinguishes different types of dependency (toolchain, visibility, output_group, or different attributes). Whereas the DependencyKey is a single dependency: it's a label, a config transition, and maybe a bit more data, but it doesn't care what the key was (except in a few cases where we are trying to optimize).

I have more work on ConfigurationResolver, so we can revisit this when those changes are ready for review.

Issue #11382, opened 2020-05-13
Java Builder Stub Action is by default using jdk11 as javabin, not the one specified in toolchain


ThomasCJY:
### Description of the problem / feature request:
While creating java stub action, bazel is using the default java executable (jdk11, on mac that is `remotejdk11_macos`) to replace the `%javabin%` placeholder in `java_stub_template` file. [Code Link](https://github.com/bazelbuild/bazel/blob/77502091f06b84054921f4eeee3f1d093dd8427d/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BazelJavaSemantics.java#L288-L291) 

Ideally, bazel should respect the java version specified in toolchain and align the stub javabin version with it. I know it's only a stub but in reality it's been used a lot in java-related build systems, for example, [rules_kotlin](https://github.com/bazelbuild/rules_kotlin/) is silently using this stub action on all compilation tasks, which partly contributed to [the issue](https://github.com/bazelbuild/rules_kotlin/issues/188) where kotlin target is always being built by jdk11 without respecting the java version defined in toolchain.

### What operating system are you running Bazel on?

macos 10.15.4

### What's the output of `bazel info release`?

2.2.0

cushon:
what's the simplest way to reproduce the issue you're seeing? can you provide a minimal example?

ThomasCJY:
@cushon simply build any kotlin targets can reproduce this issue. I added example code here https://github.com/ThomasCJY/bazel_example/pull/1

cushon:
Isn't this a `rules_kotlin` bug and not a Bazel bug, then? It looks like it's accessing the template directly and doing its own substitution: https://github.com/bazelbuild/rules_kotlin/blob/484f116c1949984697556bcd371ef8eecacc08cb/kotlin/internal/jvm/impl.bzl#L67

ThomasCJY:
@cushon well the [code you linked](https://github.com/bazelbuild/rules_kotlin/blob/484f116c1949984697556bcd371ef8eecacc08cb/kotlin/internal/jvm/impl.bzl#L67) doesn't apply to `kotlin_jvm_lib` rule, it is only applied for the binary target. I believe `kotlin_jvm_library` rule is following the default path to create a builder file.

lberki:
@comius , WDYT? It looks like this is something for which a fix may fall out from your toolchain work?

comius:
@ThomasCJY I downloaded your reproduction and tried it out.

[...]

PR #11402, opened 2020-05-15
[RFC] Fix Java Builder Stub Action always using jdk11

ThomasCJY:
### Background
Issue: https://github.com/bazelbuild/bazel/issues/11382

As stated in the issue above, in certain cases we want the stub action to take into account of the current JAVA_HOME. Example: https://github.com/bazelbuild/rules_kotlin/pull/330 

### Change
If JAVA_HOME is set, use it to locate javabin

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
@@ -228,7 +228,12 @@ else
 fi
 
 # Set JAVABIN to the path to the JVM launcher.
-%javabin%
+JAVABIN=""
+if [[ -z "$JAVA_HOME" ]]; then
+    %javabin%
+else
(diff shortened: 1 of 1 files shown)


jin:
I have a feeling this is configuring the java toolchain at the wrong level. You shouldn't have to tweak the stub template, since `%javabin%` is populated with information from the [various Java toolchain flags.](https://docs.bazel.build/versions/master/bazel-and-java.html)

ThomasCJY:
TBH, I'm a bit confused while digging into this, which `JAVABIN` are we supposed to use here? The one specified in `--javabase` or the one we specified in `--java_toolchain`?

comius (commented):
Cannot accept this PR. 

%javabin% substitution ensures code is executed using the JDK it was built for. There can be multiple JDKs on the machine. Changing this to JAVA_HOME when JAVA_HOME is set would result in breakages both internally and for other clients. 

More details explained in https://github.com/bazelbuild/bazel/issues/11382

Issue #11243, opened 2020-04-29
Clarify macOS installation instructions and de-duplicate homebrew formulas


jin:
There exist two Homebrew formulas for Bazel:

Default tap: https://github.com/Homebrew/homebrew-core/blob/master/Formula/bazel.rb
bazelbuild tap: https://github.com/bazelbuild/homebrew-tap/blob/master/Formula/bazel.rb

Note that the default formula has a dependency on Python 3.8 and Java 1.8:

https://github.com/Homebrew/homebrew-core/blob/ac3ab999a2be2871e256819138a0fabb8f1c6787/Formula/bazel.rb#L14-L16

```
  depends_on "python@3.8" => :build
  depends_on :java => "1.8"
  depends_on :macos => :yosemite
```

whereas the bazelbuild tap does not depend on Java nor Python (which makes sense, given that we have an embedded JDK to run Bazel, and bring in a remote JDK to run Java tools).

Our [official macOS instructions](https://docs.bazel.build/versions/master/install-os-x.html#install-on-mac-os-x-homebrew) recommend using the bazelbuild tap. But given that most macOS Homebrew users would instinctively type "brew install bazel" without reading the installation instructions, they would end up using the default tap, bringing a system dependency on Java 8 and Python, and not using the brew tap the Bazel team maintains at all. 

Quote from a [user](https://twitter.com/bernaferrari/status/1255340067440254977?s=20): "I never actually read (the installation instructions). When I needed Bazel I just did brew install bazel (because that's how 98% of things work) and it worked.. until I updated Java and formula broke."

Can we merge the bazelbuild tap into the default tap, and drop the dependency on Java 1.8? That'll allow us to stop maintaining a custom tap during the release process as well, and actually release Bazel on the "blessed" path.

tetromino:
Homebrew by policy [doesn't allow pre-built binary blobs](https://github.com/Homebrew/brew/blob/master/docs/Acceptable-Formulae.md#we-dont-like-binary-formulae) in their core tap (similar to Debian and other distros with a strongly free/libre philosophy). This means building Bazel from source, which means a build-time dependency on Python and a build- and run-time dependency on some version of the JDK. Therefore, I think the best fix we could do would be along the following lines:

* Send a PR to update the core tap to use a Homebrew-managed OpenJDK 11 (so users won't accidentally uninstall a systemwide Java version outside of Homebrew's control accidentally breaking Bazel).
[...]

PR #11403, opened 2020-05-15
Install the Bazel Homebrew package directly

NashMiao:
I try to install bazel in the way provided by documents over and over again, but it always failed :

```bash
$ brew install bazelbuild/tap/bazel
Updating Homebrew...
==> Installing bazel from bazelbuild/tap
==> Downloading https://releases.bazel.build/3.1.0/release/bazel-3.1.0-installer-darwin-x86_64.sh
-=O=-            #     #     #     #                                          
curl: (7) Failed to connect to releases.bazel.build port 443: Operation timed out
Error: An exception occurred within a child process:
  DownloadError: Failed to download resource "bazel"
Download failed: https://releases.bazel.build/3.1.0/release/bazel-3.1.0-installer-darwin-x86_64.sh
```

image

However, I found we can install bazel directly with Homebrew successfully:

```bash
brew install bazel
```

image

Here is the information:

```bash
$ brew info bazel
bazel: stable 3.1.0 (bottled)
Google's own build tool
https://bazel.build/
Not installed
From: https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git/Formula/bazel.rb
==> Dependencies
Build: python@3.8 ✔
Required: openjdk@11 ✘
==> Requirements
Required: macOS >= 10.10 ✔
==> Analytics
install: 4,734 (30 days), 17,069 (90 days), 70,323 (365 days)
install-on-request: 4,434 (30 days), 15,988 (90 days), 65,300 (365 days)
build-error: 0 (30 days)
```
image

image

### DIFF
--- a/site/docs/install-os-x.md
+++ b/site/docs/install-os-x.md
@@ -110,8 +110,7 @@ recommend using the `bazelbuild` tap, which is maintained by the Bazel team._
 Install the Bazel package via Homebrew as follows:
 
 ```bash
-brew tap bazelbuild/tap
-brew install bazelbuild/tap/bazel
+brew install bazel
 ```
 
 All set! You can confirm Bazel is installed successfully by running the following command:
@@ -123,7 +122,7 @@ bazel --version
 Once installed, you can upgrade to a newer version of Bazel using the following command:
 
 ```bash
-brew upgrade bazelbuild/tap/bazel
+brew upgrade bazel
 ```
 
 <h2 id="install-on-mac-os-x-bazelisk">Install using Bazelisk</h2>


NashMiao:
@googlebot I signed it!

tetromino:
The actual problem appears to be that https://releases.bazel.build is not accessible from China.

The two formulas (bazelbuild/tap/bazel vs. the default bazel) are actually different - `bazelbuild/tap/bazel` installs our official binary, the default `bazel` formula installs a binary built by the Homebrew project from source. I _think_ for now we want to encourage the `bazelbuild/tap` one (since we can better control that our users get a good experience) - see #11243.

But obviously we need to fix connectivity from China.

NashMiao:
Thank you for your answer. After reading this [#11243](https://github.com/bazelbuild/bazel/issues/11243), I fully agree with you. Now, maybe the server's network needs to be improved.

Issue #1787, opened 2016-09-19
Missing files from embedded_tools


davidzchen:
This issue was discovered in tensorflow/tensorflow#4285

The non-GPU TF CI builds with that PR are failing consistently with messages similar to the following:

```
ERROR: package contains errors: bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/install/2ed8c6c819c55073ad03a9309500f9b2/_embedded_binaries/embedded_tools/src/main/protobuf.
ERROR: error loading package 'bazel-ci_build-cache/.cache/bazel/_bazel_jenkins/install/2ed8c6c819c55073ad03a9309500f9b2/_embedded_binaries/embedded_tools/src/main/protobuf': Extension file not found. Unable to load package for '//tools/build_rules:genproto.bzl': BUILD file not found on package path.
```

The error occurs when running `bazel fetch //...` after running `bazel clean --expunge`.

It appears that this error is caused by files, namely `.bzl` files under `//tools/build_rules` that are missing in `embedded_tools`. This can be reproduced by running the following:

```
❯❯❯ bazel fetch @bazel_tools//...
Extracting Bazel installation...
..
ERROR: package contains errors: third_party/jformatstring.
ERROR: error loading package '@bazel_tools//third_party/jformatstring': Extension file not found. Unable to load file '@bazel_tools//tools/build_rules:java_rules_skylark.bzl': file doesn't exist or isn't a file.
```

damienmg:
We should stub out the BUILD file for third_party so bazel query @bazel_tools//... works

laszlocsomor:
This looks like a serious issue. I'm starting to investigate.

laszlocsomor:
Can repro at https://github.com/bazelbuild/bazel/commit/36395c417899bd4009efc3a6cda5ea838f3d98ef:

```
$ bazel fetch @bazel_tools//...
ERROR: package contains errors: third_party/jformatstring.
ERROR: error loading package '@bazel_tools//third_party/jformatstring': Extension file not found. Unable to load file '@bazel_tools//tools/build_rules:java_rules_skylark.bzl': file doesn't exist or isn't a file.

$ cat third_party/jformatstring/BUILD | grep '^load'
load("//tools/build_rules:java_rules_skylark.bzl", "bootstrap_java_library")

$ cat tools/build_rules/java_rules_skylark.bzl | grep 'bootstrap_java_library'
bootstrap_java_library = rule(
```

[...]

PR #11404, opened 2020-05-15
Make it possible to do `bazel query  @bazel_tools//...` without errors.

meisterT:
Fixes https://github.com/bazelbuild/bazel/issues/1787


philwo:
Nice, thanks!

meisterT:
This fails with RBE, do you have idea why this would behave differently?
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/401449e2-a170-446c-a47a-887fd0b3cbdc/src/test/shell/bazel/bazel_tools_test/attempt_3.log

PR #11413, opened 2020-05-15
Correct link to patch utility functions.

adam-azarchs:


gregestren:
Nice catch, thank you!

Issue #10930, opened 2020-03-09
Tracking issue for refactoring a Bazel BUILD files to more granular structure


kkress:
This is a tracking bug for work on Bazel's own BUILD files.  The goal of this work is to establish 1 BUILD file per Java package in Bazel's own source code.

This has a number of advantages for performance, especially when run with a remote executor. 

In general the focus is on targets in //src/main/java/com/google/devtools/build/lib:*

kkress:
Just wanted to drop a comment here indicating this is in progress.  There have been a steady stream of BUILD file changes coming through.  I'm currently cleaning up src/test/java/... which makes it significantly easier to reorganize the src/main code when complete.

PR #11417, opened 2020-05-16
Break cyclic dependency between lib.runtime and lib.runtime.commands

Yannic:
Pre-work for moving `.../lib/runtime/commands` into its own package.

This change
  - Renames `runtime/BuiltinCommandModule.java` to `runtime/commands/BuiltinCommandModule.java`
  - Renames `runtime/commands/InfoItem.java` to `runtime/InfoItem.java`
  - Moves inner classes specific to `{bazel,blaze} info` from `runtime/InfoItem.java` to `runtime/commands/info`

Updates #10930


Yannic:
@katre ptal, thanks!

/cc @kkress

PR #11430, opened 2020-05-18
Don't retry UNAUTHENTICATED gRPC responses

illicitonion:
These failures will not be resolved by simply retrying, they require
performing some kind of authentication (or attaching some form of
authorization), which the RemoteRetrier will not automatically do.


ulfjack:
@meisterT

PR #11434, opened 2020-05-18
Add `stub_shebang` to `py_runtime`

### LINKED ISSUE #8685: Allow customizing Python stub script's shebang in the Python toolchain

beasleyr-vmw:
Added a `stub_shebang` to `py_runtime` to allow users to specify a custom
"shebang" expression used by the Python stub script.  Motivation is to
support environments where `#!/usr/bin/env python` isn't valid (ex:
no `/usr/bin/python`, but there's a `/usr/bin/python3`).

Closes #8685.

### Guided tour
- Added field to `PyRuntimeInfoApi` inc. default value.
- Added field to `PyRuntimeInfo` provider inc. tests.
- Added field to `py_runtime` Starlark rule inc tests.
- Replaced static `#!/usr/bin/env python` in the stub w/ `%shebang%`.
- There are a few redundancies w/r/t declaring defaults and documentation.
  This is because there are a few separate public APIs (ex:
  `PyRuntimeInfo(...)`, `py_runtime(...)`), and I want to make sure defaults
  appear in the generated docs.

Testing Done:
- `bazelisk test src/test/java/com/google/devtools/build/lib/bazel/rules/python/...`
  `bazelisk test src/test/java/com/google/devtools/build/lib/rules/python/...`

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
@@ -165,6 +165,9 @@ public class BazelPythonSemantics implements PythonSemantics {
             STUB_TEMPLATE,
             ImmutableList.of(
                 Substitution.of(
+                    "%shebang%", getStubShebang(ruleContext, common)
+                ),
+                Substitution.of(
                     "%main%", common.determineMainExecutableSource(/*withWorkspaceName=*/ true)),
                 Substitution.of("%python_binary%", pythonBinary),
                 Substitution.of("%imports%", Joiner.on(":").join(common.getImports().toList())),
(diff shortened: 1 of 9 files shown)


aiuto:
@brandjon Sorry to hand this to you, but you're the best person to route it if it's not yours to look at.

lberki:
/cc @brandjon @comius 

This looks pretty mergeable to me, although I haven't touched our Python rules since a good while. I'll send out an internal code review hoping that it is. If it's not, I'll just abandon it.

lberki (approved):

brandjon:
Reviewed, looks good to me. lberki will merge.

PR #11443, opened 2020-05-19
[Profiler] Add target label of action owner to profiler trace files

sbalabanov:
While analyzing trace files, it is useful to have target label associated with the action which was created by that target. This change adds target label under trace events which have owner, mostly Action-type event. Target label is saved as a custom entry to "args" map attached to each Action event, as per trace format definition.
The change is guarded by new flag `--experimental_include_target_label`

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
+++ b/src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
@@ -165,15 +165,18 @@ public final class Profiler {
 
   private static final class ActionTaskData extends TaskData {
     final String primaryOutputPath;
+    final String targetLabel;
 
     ActionTaskData(
         int id,
         long startTimeNanos,
         ProfilerTask eventType,
         String description,
-        String primaryOutputPath) {
+        String primaryOutputPath,
+        String targetLabel) {
       super(id, startTimeNanos, eventType, description);
       this.primaryOutputPath = primaryOutputPath;
+      this.targetLabel = targetLabel;
     }
   }
 
@@ -370,7 +373,8 @@ public final class Profiler {
       long execStartTimeNanos,
       boolean enabledCpuUsageProfiling,
       boolean slimProfile,
-      boolean includePrimaryOutput)
+      boolean includePrimaryOutput,
+      boolean includeTargetLabel)
       throws IOException {
     Preconditions.checkState(!isActive(), "Profiler already active");
     initHistograms();
@@ -401,7 +405,8 @@ public final class Profiler {
                   slimProfile,
                   outputBase,
                   buildID,
-                  includePrimaryOutput);
+                  includePrimaryOutput,
+                  includeTargetLabel);
           break;
         case JSON_TRACE_FILE_COMPRESSED_FORMAT:
           writer =
@@ -411,7 +416,8 @@ public final class Profiler {
                   slimProfile,
                   outputBase,
                   buildID,
-                  includePrimaryOutput);
+                  includePrimaryOutput,
+                  includeTargetLabel);
       }
       writer.start();
     }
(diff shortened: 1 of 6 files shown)


ulfjack:
Do you have an idea of how much this increases the profile size?

@meisterT

sbalabanov:
@ulfjack Depends on many things, like the amount of actions in the profile, pretty hard to calculate a reasonable average. But in any case we are talking about a 10-20% increase at worst.
I do not see it may become a problem anyways because this is guarded by a flag and it's off by default (same as primary output)

src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
```diff
@@ -932,6 +941,12 @@ private void writeTask(JsonWriter writer, TaskData data) throws IOException {
       if (includePrimaryOutput && data instanceof ActionTaskData) {
         writer.name("out").value(((ActionTaskData) data).primaryOutputPath);
       }
+      if (includeTargetLabel && data instanceof ActionTaskData) {
+        writer.name("args");
```
meisterT, line 945: I assume you placed this in args to have it visible in `chrome://tracing`?

cc @joeleba we may want to consider the same for the primary output

src/main/java/com/google/devtools/build/lib/profiler/TraceEvent.java
```diff
@@ -57,6 +63,8 @@ public static TraceEvent create(
   // Only applicable to action-related TraceEvents.
   @Nullable
   public abstract String primaryOutputPath();
+  @Nullable
+  public abstract Map<String, Object> args();
```
meisterT, line 67: Instead of storing and exposing the full map here, I would only expose what we know of, in this case the action owner.
So parsing the full map below is fine and then extracting the data that we need.

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
```diff
@@ -288,6 +288,14 @@ public String getTypeDescription() {
               + "to the action's primary output.")
   public boolean includePrimaryOutput;
 
+  @Option(
+          name = "experimental_include_target_label",
```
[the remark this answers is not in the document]
sbalabanov: I can change to that for 'experimental_include_target_label', not sure you want to change a flag for primary output as this is potentially a breaking change

src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
sbalabanov: Yes, this makes it visible in Chrome. I see primary output to be in args as well, but may be too late to change it.
Trace file has pretty strict format about JSON fields allowed in it, however `chrome://tracing` parser is not very strict so it just skips unknown fields.
I guess it is better to stick with the format just for the sake of consistency if not visibility.

src/main/java/com/google/devtools/build/lib/profiler/TraceEvent.java
sbalabanov: My understanding of that object was that it closely mimics the structure of the trace file record, but I can change to proposed no problem.

sbalabanov:
updated; mind another look please?

src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
meisterT: My only concern would be file size. But since it is guarded behind a flag and after gzipping the difference is probably negligible your reasoning sounds valid.

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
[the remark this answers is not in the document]
meisterT: I'll do that separately, thanks!

sbalabanov:
thank you for the prompt review and merge!

PR #11444, opened 2020-05-19
[WIP] Demonstrate the error when the host configuration is trimmed.

katre:
In this case, the trimmed "host" configuration has the same checksum as
the real host configuration, but the fragments are very different, and
it's possible that if other flags were set, the checksums would diverge.


katre:
See the log at https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/830d9718-a9ae-4807-b72a-d289eac6f4e0/src/test/java/com/google/devtools/build/lib/analysis/AnalysisTests/attempt_1.log

```
1) testGetDirectPrerequisites_hostTransition(com.google.devtools.build.lib.analysis.BuildViewTest$WithTrimmedConfigurations)
expected: BuildConfigurationValue.Key[3dc93380092d4adfa5cb0afc8468e2ed45f557aa2378bc2716d71dfcc54c1c3d]
but was : (non-equal instance of same class with same string representation)
	at com.google.devtools.build.lib.analysis.BuildViewTest.testGetDirectPrerequisites_hostTransition(BuildViewTest.java:561)
```

Issue #5088, opened 2018-04-24
Support completion for fish shell


tmandry:
### Description of the problem / feature request:

Bazel generates a bash completion script.

It would be nice if completions were also generated for [fish, a fairly popular shell](https://github.com/fish-shell/fish-shell).

### Feature requests: what underlying problem are you trying to solve with this feature?

Complete bazel commands and package paths in fish.

### What operating system are you running Bazel on?

Linux

###  Have you found anything relevant by searching the web?

Fish has its own completion syntax, [described in the documentation](https://fishshell.com/docs/current/index.html#completion). It has a lot of [built-in completions](https://github.com/fish-shell/fish-shell/tree/master/share/completions) that can be used as examples.

Fish also supports running a script or process to get the completions; see this [gcloud completion script for fish in python](https://gist.github.com/rjohnsondev/4069b87c9a3f7adf4379662eaba3c3d7) for an example.

Not expecting this anytime soon, but putting this on the radar with links in case anyone wants to pick it up.

akirabaruah:
I've got basic completions working in fish. Definitely not a fish expert, so there are probably many improvements to be made to the script, but it works reasonably well so far. Hope it helps!

![Peek 2020-05-20 02-08](https://user-images.githubusercontent.com/6751517/82429304-8220e600-9a40-11ea-86f7-a33ac45957c8.gif)

meisterT:
Thanks, that looks nice!

I'm not a fish user myself so I would appreciate if this is also looked at by one who is.

aiuto:
That's what worries me about the PR. It adds a feature that no on on the Bazel team can maintain.
It's a pity we don't have a keyword auto-triage in github so that the mention of fish automatically assigns it to @tmandry. Something like that would reduce the cost of taking it in.

meisterT:
aiuto@ this is one of the most upvoted issues and the code is not entangled with Bazel itself, so I am willing to merge it and give it a try.

aiuto:
Be my guest.
My secondary concern is why are contributions not needed to build bazel or by all users in the bazel repo rather than a contrib repo. Right now, I can only get these command completion scripts with the debian distro. They should be their own package.

But... we can always fork them off another day.

akirabaruah:
Are there already plans to create a separate contrib / completions repo? Happy to help fork out the completions if needed.

aiuto:
[...]

PR #11450, opened 2020-05-20
Add fish shell completion for bazel

akirabaruah:
Addresses issue #5088 

Test: Follow steps in README.md and use bazel in fish shell.

### DIFF
--- a/scripts/fish/completions/bazel.fish
+++ b/scripts/fish/completions/bazel.fish
@@ -1,0 +1,141 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# fish completion for bazel
+
+set __bazel_command "bazel"
+set __bazel_completion_text ($__bazel_command help completion 2>/dev/null)
+set __bazel_help_text ($__bazel_command help 2>/dev/null)
+set __bazel_startup_options_text ($__bazel_command help startup_options 2>/dev/null)
+
+function __bazel_get_completion_variable \
+    -d 'Print contents of a completion helper variable from `bazel help completion`'
+    set -l var $argv[1]
+    set -l regex (string join '' "$var=\"\\([^\"]*\\)\"")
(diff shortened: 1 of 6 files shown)


scripts/fish/completions/bazel.fish
```diff
@@ -0,0 +1,141 @@
...
+end
+
+function __bazel_get_subcommand_args \
+    -d 'Print an argument string for subcommand completion'
+    set -l subcommand $argv[1]
+    set -l arg_type (__bazel_get_subcommand_arg_type $subcommand)
+
+    switch $arg_type
+        case "label"
+            echo "($__bazel_command query '//...' 2>/dev/null)"
```
meisterT, line 94: I assume we should add `-k` to all these queries to return results even when you can't do `bazel query ...`.

akirabaruah: Yep, thanks for pointing that out. Done!

akirabaruah:
> Please include the default license header.

Done

> Also, you probably want to do the equivalent of this in the installer?
> [74c274a](https://github.com/bazelbuild/bazel/commit/74c274a4ed8f97ab32724641bb011cee2e057a76)

Done

> Even after the change that I proposed it doesn't complete for me in the Bazel repository for `bazel test //src/test/<TAB><TAB>`.
> 
> Can you check what's going on there?

Hmm, just to sanity check:
1. Does `bazel <tab><tab>` complete subcommands for you like so?
Peek 2020-05-28 22-41

2. Does `bazel query -k 'tests(//...)' 2>/dev/null` list targets as expected?

`bazel test //src/test/<tab><tab>` works for me locally, but it does take a few seconds the first time (as demonstrated in the screencap above). The subcommand completions are derived by running `bazel help`, which is why it takes so long. Could it be that your shell was hanging on running bazel in the background?

If needed, I think we can speed it up by building fish.bazel using templates, similar to bash completion.

meisterT:
I just tried again, it works now. Perhaps I was too impatient.

I think there's still room for improvement, e.g. if one does starts the completion with the prefix `//foo/bar/zoo` instead of querying `//...` you could query `//foo/bar/...`.
Another idea: if it's possible to display extra information then it would be nice to query with `--output=label_kind` and display the full info.

But this is a great start and I am going to merge it as is. Thanks for your contribution!

akirabaruah:
Awesome. Glad it works! I agree that there's much to be improved, so I'm happy to follow up with some more PRs to address your suggestions.

gorakhargosh:
An improvement for this would be if I could use https://github.com/junegunn/fzf to select the target. Consider installing fzf and run the following command.

`bazel test (bazel query '...' --output label | fzf)`

*mind blown*

meisterT:
@gorakhargosh That's unrelated to fish and works in any shell, right?

gorakhargosh:
@meisterT sure, `fzf` can be used with any shell. The command I've used uses fish shell expansion however. It may make a good addition to the autocompletion feature set. The equivalent in bash would be:

`bazel test $(bazel query '...' --output label | fzf)`

PR #11452, opened 2020-05-20
Remove lib.runtime.commands from lib.runtime

Yannic:
Pre-work for splitting `.../lib:runtime`.

This change
  - Renames `runtime/commands/ProjectFileSupport.java` to `runtime/ProjectFileSupport.java` (which was an oversight in #11417)
  - Moves `runtime.commands.CleanCommand.CleanStartingEvent` into `runtime/commands/events` to avoid having `.../lib/worker` depend on all of `.../lib/runtime/commands`.
  - Renames `runtime/commands/GotProjectFileEvent.java` to `runtime/events/GotProjectFileEvent.java` for similar reasons as above.

Updates #10930


Yannic:
@katre ptal, thanks!

/cc @kkress

katre:
Working on importing this. The test error is a known flake.

PR #11460, opened 2020-05-21
Simplify Dependency and make it use AutoValue.

katre:


katre:
There's no good reason to have two subclasses, and the equals methods were hiding actual differences between instances.

Issue #11356, opened 2020-05-11
Allow more flexibility for C++ file extensions given as input to rules or Starlark API


oquenchil:
Currently, extensions are allowed for rules and the Starlark C++ API are hard-coded. This leads to users getting errors unnecessarily and we see a lot of PRs to add a new extension to the hard-coded list.

We should evaluate how to be more flexible with extensions in a safe way without breaking existing code.

loshca:
How about to start with
1) removing extension validation for public headers and private headers in CcModule.java. 
2) removing extension validation in addPrivateHeader (CcCompilationHelper.java)?

This will not break existing code: cc_binary && cc_library still will use the hardcoded list to separate sources and private headers, but we can use any headers (including headers with no extension) in custom cc rules using cc_common.compile.
It looks cheap, but will help a lot!

oquenchil:
Yes, that sounds excellent! If you want to send a PR we will review happily.

Issue #164, opened 2015-04-27
Allow exporting hdrs without filename extensions


kalakris:
I was trying to package a library which contains header files without extensions, and I wanted to export just a subset of them. However, when I listed these headers in a hdrs attribute of a cc_library...
```
hdrs = ["my_stl/vector", "my_stl/array"]
```
...I got the following error:
```
ERROR: /home/mrinal/bazel_test/my_stl/BUILD:64:12: in hdrs attribute of cc_library rule //my_stl:my_stl: file '//my_stl:my_stl/vector' is misplaced here (expected .h, .hh, .hpp, .hxx or .inc).
ERROR: /home/mrinal/bazel_test/my_stl/BUILD:64:12: in hdrs attribute of cc_library rule //my_stl:my_stl: '//my_stl:my_stl/vector' does not produce any cc_library hdrs files (expected .h, .hh, .hpp, .hxx or .inc).
```

It would be nice to either (1) allow header files without extensions, since this is a fairly common case encountered in the wild, or (2) allow customizing the list of allowed extensions per file type in the cc_library rule.

ulfjack:
We actually allow any file in hdrs internally. We didn't want to do that from the beginning, but random extensions are not that uncommon, so it seems like we need to revisit that decision.

fzakaria:
Just thought I'd open this:
You made the change to 'hdrs' which is only for "public"

I've come across some repositories (i.e. gflags ) that use different extensions for internal-only header files.
Perhaps this needs to be more lax in general. I'm not sure why bazel makes assumptions on the extension since we are providing the file format to the properties.

lberki:
I think you're actually right, but IIRC we have a bunch of places where some extensions are hard-coded into Bazel, or at least that's a few references to CppFilesTypes.CPP_SOURCE would suggest. @ulfjack  ?

ulfjack:
Yeah, for files in srcs we need to know whether they're actually sources or headers. We could consider re-designing cc_* to have another attribute for internal headers, so we don't need to switch on extension, but it's unfortunately a bigger change.

ddwolf:
I am running this problem now.
I am converting an existing project which is built by make. this project contains some header files whose names end with `.map`. So is there a solution for this?

finn-ball:
+1 having a similar issue

jiemingwang:
It will be nice to support such requiremment. Many large projects include header files that don't have extensions. e.g., QT project uses class names as the header file names. Can this be done with command line options?

PR #11469, opened 2020-05-22
Allow any header extensions in the cc_common.compile

loshca:
#11356

Allow any extension in the cc_common.compile, but still use a hardcoded list of extensions in CcCommon to separate sources and headers from "srcs" attribute for cc_binary and cc_library.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java
@@ -420,10 +420,8 @@ public final class CcCompilationHelper {
   }
 
   private CcCompilationHelper addPrivateHeader(Artifact privateHeader, Label label) {
-    boolean isHeader = CppFileTypes.CPP_HEADER.matches(privateHeader.getExecPath());
     boolean isTextualInclude =
         CppFileTypes.CPP_TEXTUAL_INCLUDE.matches(privateHeader.getExecPath());
-    Preconditions.checkState(isHeader || isTextualInclude);
 
     if (ccToolchain.shouldProcessHeaders(featureConfiguration, cppConfiguration)
         && !isTextualInclude) {
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
@@ -1695,16 +1695,6 @@ public abstract class CcModule
         sources,
         CppFileTypes.ALL_C_CLASS_SOURCE.including(CppFileTypes.ASSEMBLER),
         FileTypeSet.of(CppFileTypes.CPP_SOURCE, CppFileTypes.C_SOURCE, CppFileTypes.ASSEMBLER));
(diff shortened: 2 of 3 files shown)


loshca:
I need some clarifications:

1. If we have the source file to cxxopt apply, then all headers used for the source file compilation also will use this option, don't we? Does this line have any sense here? https://github.com/bazelbuild/bazel/blob/ff402874afda761effab6bf230fa1919a1dfcc96/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java#L1478
2. "srcs" attribute has validation on the rule level, but "hdrs" attribute in cc_library doesn't. This PR allows any header extension there. Should I add restrictions for "hdrs" attribute to keep old behaviour?
https://github.com/bazelbuild/bazel/blob/74d84d43b0b810e7eb2c8f866694fa2dcf209baa/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppRuleClasses.java#L278

chancila:
@oquenchil any updates here? I've ran into this issue attempting to compile libc++ with a custom rule using cc_common...

oquenchil:
> 1. If we have the source file to cxxopt apply, then all headers used for the source file compilation also will use this option, don't they? Does this line have any sense here?

Can you please re-formulate this question? I don't understand it.

> 2\. "srcs" attribute has validation on the rule level, but "hdrs" attribute in cc_library doesn't. This PR allows any header extension there. Should I add restrictions for "hdrs" attribute to keep old behaviour?

Precisely because of this I don't like this change. It's easy to change it for cc_library to keep old behavior but there are many other Starlark rules out there which you won't be able to change.

Ideally, you submit a separate PR with the specific standard header extension that is not yet in the list of extensions (not something like *.foo or *.unknown). This is also very inconvenient but a better option until we figure out the problem of hard-coded extensions.

I will close this PR.

loshca:
@oquenchil 

> Can you please re-formulate this question? I don't understand it.

I don’t realize the case when compilation action is created with a CPP_HEADER as a source, probably we can remove this check.

> Precisely because of this I don't like this change. It's easy to change it for cc_library to keep old behavior but there are many other Starlark rules out there which you won't be able to change.

There was some kind of misunderstanding. It seems to me that this PR does exactly what I described here https://github.com/bazelbuild/bazel/issues/11356#issuecomment-627352877

> Ideally, you submit a separate PR with the specific standard header extension that is not yet in the list of extensions (not something like *.foo or *.unknown). This is also very inconvenient but a better option until we figure out the problem of hard-coded extensions.

We decided to use a custom build with this patch for now:
- expanding the existing list will not solve our major problem - supporting headers **without** extensions
- we also can not rely on the fact that when we need a new extension, we will be able to add it fast enough (not to mention the fact that such PRs can be rejected because the extension is not so common or even be company internal)

Perhaps is it worth tweaking these changes so that they can be used with the incompatible flag?

The problem of headers without extensions has been around since 2015: https://github.com/bazelbuild/bazel/issues/164. There is still no even workaround.
Maybe it's time to think not about a perfect solution, but about some solution.

PR #11476, opened 2020-05-25
Bootstrap with system jar files in Debian build

meteorcloudy:
In the compile.sh script, we search all jar files under third_party directory and provide them to classpath. This change allows us to use system jar files as a replacement in a distribution build.

Working towards: #9408

### DIFF
--- a/scripts/bootstrap/compile.sh
+++ b/scripts/bootstrap/compile.sh
@@ -20,10 +20,10 @@ if [ -d derived/jars ]; then
   ADDITIONAL_JARS=derived/jars
 else
   DISTRIBUTION=${DISTRIBUTION:-debian}
-  ADDITIONAL_JARS="$(grep -o "\".*\.jar\"" tools/distributions/${DISTRIBUTION}/${DISTRIBUTION}_java.BUILD | sed "s/\"//g" | sed "s/^/\/usr\/share\/java\//g")"
+  ADDITIONAL_JARS="$(grep -o '".*\.jar"' tools/distributions/${DISTRIBUTION}/${DISTRIBUTION}_java.BUILD | sed 's/"//g' | sed 's|^|/usr/share/java/|g')"
 fi
 PROTO_FILES=$(find third_party/remoteapis third_party/googleapis third_party/pprof src/main/protobuf src/main/java/com/google/devtools/build/lib/buildeventstream/proto src/main/java/com/google/devtools/build/skyframe src/main/java/com/google/devtools/build/lib/skyframe/proto src/main/java/com/google/devtools/build/lib/bazel/debug src/main/java/com/google/devtools/build/lib/starlarkdebug/proto -name "*.proto")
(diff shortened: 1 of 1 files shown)


meteorcloudy:
FYI @olekw

scripts/bootstrap/compile.sh
```diff
@@ -17,12 +17,13 @@
...
-  PROTOBUF_JARS=derived/jars
+  ADDITIONAL_JARS=derived/jars
 else
-  PROTOBUF_JARS="/usr/share/java/protobuf.jar /usr/share/java/protobuf-java-util.jar"
+  DISTRIBUTION=${DISTRIBUTION:-debian}
+  ADDITIONAL_JARS="$(grep -o "\".*\.jar\"" tools/distributions/${DISTRIBUTION}/${DISTRIBUTION}_java.BUILD | sed "s/\"//g" | sed "s/^/\/usr\/share\/java\//g")"
 fi
 PROTO_FILES=$(find third_party/remoteapis third_party/googleapis third_party/pprof src/main/protobuf src/main/java/com/google/devtools/build/lib/buildeventstream/proto src/main/java/com/google/devtools/build/skyframe src/main/java/com/google/devtools/build/lib/skyframe/proto src/main/java/com/google/devtools/build/lib/bazel/debug src/main/java/com/google/devtools/build/lib/starlarkdebug/proto -name "*.proto")
-LIBRARY_JARS=$(find $PROTOBUF_JARS third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
+LIBRARY_JARS=$(find $ADDITIONAL_JARS third_party -name '*.jar' | grep -Fv JavaBuilder | grep -Fv third_party/guava | grep -Fv third_party/guava | grep -ve 'third_party/grpc/grpc.*jar' | tr "\n" " ")
```
meisterT, line 26: Can you add a comment here why we are excluding some jars?

Also see the duplicate guava that I mentioned in the other commit

scripts/bootstrap/compile.sh
```diff
@@ -17,12 +17,13 @@
 # Script for building bazel from scratch without bazel
 
 if [ -d derived/jars ]; then
-  PROTOBUF_JARS=derived/jars
+  ADDITIONAL_JARS=derived/jars
 else
-  PROTOBUF_JARS="/usr/share/java/protobuf.jar /usr/share/java/protobuf-java-util.jar"
+  DISTRIBUTION=${DISTRIBUTION:-debian}
+  ADDITIONAL_JARS="$(grep -o "\".*\.jar\"" tools/distributions/${DISTRIBUTION}/${DISTRIBUTION}_java.BUILD | sed "s/\"//g" | sed "s/^/\/usr\/share\/java\//g")"
```
meisterT, line 23: Please rewrite this like I proposed on the other PR.

meteorcloudy: Done.

scripts/bootstrap/compile.sh
meteorcloudy: Actually, looks like we don't have to exclude them. JavaBuilder is no longer  checked into third_party. And we have only single version of grpc and guava checked in. But I'll fix them later we I try to replace them with system jars.

The duplicated guava is also fixed in previous PR, I'll make sure to rebase correctly.

PR #11477, opened 2020-05-25
Exclude java sources for test while bootstrapping

meteorcloudy:
These sources are only for testing and are not part of Bazel binary. However, they depends on libraries like `truth`, excluding them means we can have less dependencies while bootstrapping Bazel.

Working towards: #9408

### DIFF
--- a/scripts/bootstrap/compile.sh
+++ b/scripts/bootstrap/compile.sh
@@ -50,10 +50,10 @@ if [ "$ERROR_PRONE_INDEX" -lt "$GUAVA_INDEX" ]; then
 fi
 
 DIRS=$(echo src/{java_tools/singlejar/java/com/google/devtools/build/zip,main/java} tools/java/runfiles ${OUTPUT_DIR}/src)
-EXCLUDE_FILES="src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java src/java_tools/buildjar/java/com/google/devtools/build/buildjar/javac/testing/*"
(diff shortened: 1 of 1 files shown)


meteorcloudy:
FYI @olekw

scripts/bootstrap/compile.sh
```diff
@@ -50,10 +50,10 @@ if [ "$ERROR_PRONE_INDEX" -lt "$GUAVA_INDEX" ]; then
 fi
 
 DIRS=$(echo src/{java_tools/singlejar/java/com/google/devtools/build/zip,main/java} tools/java/runfiles ${OUTPUT_DIR}/src)
-EXCLUDE_FILES="src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java src/java_tools/buildjar/java/com/google/devtools/build/buildjar/javac/testing/*"
+EXCLUDE_FILES="src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java src/java_tools/buildjar/java/com/google/devtools/build/buildjar/javac/testing/* src/main/java/com/google/devtools/build/lib/collect/nestedset/NestedSetCodecTestUtils.java"
```
meisterT, line 53: Why are we excluding these files?

meteorcloudy: Sorry, I should make it more clear in the PR description.

These sources are only for testing and are not part of the Bazel binary. If you do `bazel query "deps(//src:bazel)"`, they will not show up. However, they requires dependencies like `truth`.

meisterT: Ok, that makes sense. Please add this as a comment here.

PR #11478, opened 2020-05-25
Always use //src/java_tools/buildjar:bootstrap_toolchain

meteorcloudy:
To build Bazel without downloading remote java tools, we'll have to use the bootstrap toolchain.

This is basically a part of rollback of https://github.com/bazelbuild/bazel/pull/11192

Working towards: #9408


meteorcloudy:
FYI @olekw

PR #11479, opened 2020-05-25
Use system_repo rule in Debian build

meteorcloudy:
Previously, we use `new_local_repository` to make binaries and headers under `/usr/bin`, `/usr/include` available to the Debian build. Because `new_local_repository` symlinks everything under the directory, it has the following problems:

- headers under `/usr/include` will get leaked from the external repo. Eg, `limits.h` gets found via `external/debian_proto_deps/limits.h`. Bazel will throw an error for this while doing header checking because it's an undeclared dependency.
- `/usr/bin` may contain infinite symlink loop. Eg. `/usr/bin/X11 -> /usr/bin` ([why](https://askubuntu.com/questions/191654/why-are-there-infinitely-many-x11-subdirectories-in-usr-bin-x11)). Bazel doesn't like this and throws an error when it's detected.

In this PR, we introduce a new repository rule called `system_repo`. Instead of symlinking everything under a directory, it only creates given symlink paths. This will not only avoid the above problems, but also increase the performance by a bit.

Working towards: #9408

### DIFF
--- a/tools/distributions/debian/deps.bzl
+++ b/tools/distributions/debian/deps.bzl
@@ -14,6 +14,8 @@
 
 """Macros for defining external repositories for Debian system installed libraries."""
 
+load("//tools/distributions:system_repo.bzl", "system_repo")
+
 def debian_deps():
     debian_java_deps()
     debian_cc_deps()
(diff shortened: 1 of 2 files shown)


meteorcloudy:
FYI @olekw

meteorcloudy:
> Oh, wow. That's a nice solution to this problem!

Thanks! ;D

PR #11480, opened 2020-05-25
Bootstrap JavaBuilder with Debain system jars (third_party)

meteorcloudy:
The bootstrap_toolchain depends on jar files directly. We also need to replace them with system libraries for the Debian build.

Working towards: #9408


meteorcloudy:
FYI @olekw

PR #11487, opened 2020-05-26
Replace guava and jacoco in Debian build

meteorcloudy:
Working towards: #9408

### DIFF
--- a/tools/distributions/debian/debian_java.BUILD
+++ b/tools/distributions/debian/debian_java.BUILD
@@ -211,3 +211,73 @@ java_import(
         "com.android.tools.layoutlib.layoutlib-api.jar",
     ],
 )
+
+# libguava-java
+java_import(
+    name = "guava",
+    jars = ["guava.jar"],
+    exports = [
+        "@//third_party:error_prone_annotations",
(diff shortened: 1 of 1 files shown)


tools/distributions/debian/debian_java.BUILD
```diff
@@ -211,3 +211,73 @@ java_import(
...
+JACOCOVERSION = "0.8.3"
+java_import(
+    name = "agent",
+    jars = ["org.jacoco.agent-%s.jar" % JACOCOVERSION],
+)
+
+java_import(
+    name = "agent-%s" % JACOCOVERSION,
+    jars = ["org.jacoco.agent-%s.jar" % JACOCOVERSION],
+)
```
meisterT, line 242: why do we need these duplicates?

meteorcloudy: Because we do this in https://github.com/bazelbuild/bazel/blob/master/third_party/java/jacoco/BUILD, some places we use the versioned target and some places we use the non-versioned ones. I agree this needs some clean up, but looks like it's also related to how we handle jacoco in third_party/java/jacoco in Google3.

PR #11489, opened 2020-05-26
Minor bugfixes to generate Gcov/Lcov branch coverage

MichaelKlemm:
When using bazel to generate C++ branch coverage `BA` entries are created instead of `BRDA` lines. As far as these `BA` lines could not be used by `genhtml`, I changed it be processed as `BRDA`. Doing so the program throws an exception because of the wrong condition, trying to convert a '-' to a number. Inverted this condition to have the correct behavior of '-' which means `wasExecuted = false` and `executionCount = 0`.

### DIFF
--- a/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/BranchCoverage.java
+++ b/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/BranchCoverage.java
@@ -22,7 +22,7 @@ abstract class BranchCoverage {
 
   static BranchCoverage create(int lineNumber, int nrOfExecutions) {
     return new AutoValue_BranchCoverage(
-        lineNumber, /*blockNumber=*/ "", /*branchNumber=*/ "", nrOfExecutions);
+        lineNumber, /*blockNumber=*/ "0", /*branchNumber=*/ "0", nrOfExecutions);
   }
 
   static BranchCoverage createWithBlockAndBranch(
--- a/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java
+++ b/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java
@@ -280,7 +280,7 @@ class LcovParser {
 
       boolean wasExecuted = false;
       int executionCount = 0;
-      if (taken.equals(TAKEN)) {
+      if (!taken.equals(TAKEN)) {
         executionCount = Integer.parseInt(taken);
         wasExecuted = true;
       }
--- a/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java
+++ b/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java
@@ -14,7 +14,6 @@
 
 package com.google.devtools.coverageoutputgenerator;
 
-import static com.google.devtools.coverageoutputgenerator.Constants.BA_MARKER;
 import static com.google.devtools.coverageoutputgenerator.Constants.BRDA_MARKER;
 import static com.google.devtools.coverageoutputgenerator.Constants.BRF_MARKER;
 import static com.google.devtools.coverageoutputgenerator.Constants.BRH_MARKER;
@@ -114,9 +113,6 @@ class LcovParser {
     }
     if (line.startsWith(BRDA_MARKER)) {
       return parseBRDALine(line);
-    }
-    if (line.startsWith(BA_MARKER)) {
-      return parseBALine(line);
     }
     if (line.startsWith(BRF_MARKER)) {
       return parseBRFLine(line);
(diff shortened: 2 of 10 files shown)


MichaelKlemm:
@googlebot I signed it!

MichaelKlemm:
recheck

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovPrinter.java
```diff
@@ -132,16 +131,12 @@ private void printFNHLine(SourceFileCoverage sourceFile) throws IOException {
...
     for (BranchCoverage branch : sourceFile.getAllBranches()) {
-      if (branch.blockNumber().isEmpty() || branch.branchNumber().isEmpty()) {
-        // We skip printing this as a BRDA line and print it later as a BA line.
-        continue;
-      }
       bufferedWriter.write(Constants.BRDA_MARKER);
       bufferedWriter.write(Integer.toString(branch.lineNumber()));
       bufferedWriter.write(Constants.DELIMITER);
-      bufferedWriter.write(branch.blockNumber());
+      bufferedWriter.write(branch.blockNumber().isEmpty() ? "0" : branch.blockNumber());
```
ulfjack, line 137: is the printBALines method dead code after this change?

MichaelKlemm: Exactly. Everything about BA-lines is now obsolete. Do you know where the BA-lines came from? From Gcov/Lcov/GenInfo side I was not able to find something about, even not in the history.

MichaelKlemm:
> It might be better to split this into two changes, and also add tests for each of the bugfixes.

I know, one is the parsing and the other one the printing. Implementing tests will strongly combine these two, so I would suggest to keep it together.

ulfjack:
Unfortunately, I don't know the history of the BA lines, but I wouldn't be surprised if the tool just changed silently at some minor version. The coverage tools seem to have a habit of doing that. :-/

The change looks reasonable to me, but I didn't do a careful review of the tests, and I also didn't research the BA / BRDA thing.

@meisterT can you take a look as well, or find someone to take a look?

meisterT:
Sorry, I have no clue about this.
cc @lberki

MichaelKlemm:
Are there any other tools being used besides the lcov project (`genhtml`, `geninfo`, ...) using the *.dat output? I don't want to destroy compatibility but as far as I know there is only the one specification at the LCOV project and this is not using BA at all. So BRDA seems to be the way to go and it's also working quite well with my local C++ environment using branch coverage.

How to proceed here?

ulfjack:
I think a couple tools use lcov internal output for further processing. Regardless, if lcov HEAD accepts this and doesn't accept or document BA, I think it's fine to change it to BRDA unconditionally.

ulfjack:
So, the `BA` lines date back to lcov 1.9 (https://chromium.googlesource.com/chromium/src/third_party/lcov/+/dd41bfc42b684e08eb9205301932bae3056ff8e4%5E%21/). `BRDA` was introduced in 1.10 (http://ltp.sourceforge.net/coverage/lcov/changes.php):
```
commit b5c1bdddd1380be3ad12952ed2747df3744e227e 
Author: Peter Oberparleiter <oberpapr@users.sourceforge.net>
Date:   Wed Oct 10 08:20:21 2012 +0000

    lcov: finalizing release 1.10
```

I think it's fine not to support an 8-year-old release at this point.

ulfjack:
@lberki doesn't seem to be very responsive. Lots of work to do?

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java
```diff
@@ -280,7 +248,7 @@ private boolean parseBRDALine(String line) {
 
       boolean wasExecuted = false;
```
oquenchil, line 249: This line can be deleted.

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/BranchCoverage.java
```diff
@@ -22,7 +22,7 @@
 
   static BranchCoverage create(int lineNumber, int nrOfExecutions) {
     return new AutoValue_BranchCoverage(
-        lineNumber, /*blockNumber=*/ "", /*branchNumber=*/ "", nrOfExecutions);
+        lineNumber, /*blockNumber=*/ "0", /*branchNumber=*/ "0", nrOfExecutions);
```
oquenchil, line 25: Are the two numbers here valid GCC IDs?

Below in this file it says it should be -1 for non-gcc emitted coverage. I'm guessing that if we don't have the values present, it should be -1 and not 0?

oquenchil:
Thank you for the change! I have a couple of nits.

oquenchil:
Actually hold on addressing my comments. A colleague has just told me that BA may be important internally. I will ping this again once I know more.

ulfjack:
@oquenchil any updates? It's been a week...

oquenchil:
The problem that we are facing with coverage is that we have quite the knowledge gap right now and not enough people. I'm trying to fill that gap but I have been very busy with other issues.

The previous expert on coverage in the team told me that the BA line is important internally but they didn't remember why. I wrote an email today to reach out to 2 other people in the company that may have more context.

I will ping this once they reply.

ulfjack:
Would it help if this change added support for BRDA lines without converting to or from BA lines? Would that be an option, @MichaelKlemm ?

oquenchil:
I asked this as well because from the reply I got yesterday it wasn't clear. If I cannot tell you for certain tomorrow, I will try importing this PR with the BA deletion anyway and then we might find out later in which way it was important.

Meanwhile, @MichaelKlemm could you please address the 2 nits?

oquenchil:
Ok, I have more information now. We definitely have internal tools still reporting BA lines. But we will change that. Meanwhile, if you can change this PR to be able to parse BA lines and treat them as BRDA, then that should be enough not to break anything as far as I can tell.

Thank you!

meisterT:
What is the status of this PR?

oquenchil:
> Ok, I have more information now. We definitely have internal tools still reporting BA lines. But we will change that. Meanwhile, if you can change this PR to be able to parse BA lines and treat them as BRDA, then that should be enough not to break anything as far as I can tell.
> 
> Thank you!

It doesn't look like the internal tools will be changed any time soon, so this PR should keep the BA lines.

c-mita:
We're currently in the process of fixing issues around the handling of lcov; fixes to the parser are part of that, as is changing the generators to not output "BA" lines.

Unfortunately we cannot remove support for BA lines yet as they are still used internally at Google.

Issue #11468, opened 2020-05-22
StandaloneTestStrategy NPE/RuntimeException on shutdown after abort handled.


werkt:
Encountered the following runtime exception seemingly *after* a build has begun handling an abort signal. Despite shutting down, this should not be printing noisy NPEs that might distract from real discovery of a build failure reason.

Version was 3.1.0, running on linux. Abort message is uninteresting, but was definitely printed *before* this NPE trace.  I've included it at the end of the issue for reference.

```
Internal error thrown during build. Printing stack trace: java.lang.RuntimeException: Unrecoverable error while evaluating node 'ActionLookupData{actionLookupKey=<target> BuildConfigurationValue.Key[dda694ec6988aadc6904a86cb5a2ead6a2c7e63ef4697d3e33bf2f01b333cb51] false, actionIndex=8}' (requested by nodes 'TestCompletionKey{configuredTargetKey=<target> BuildConfigurationValue.Key[dda694ec6988aadc6904a86cb5a2ead6a2c7e63ef4697d3e33bf2f01b333cb51] false, topLevelArtifactContext=com.google.devtools.build.lib.analysis.TopLevelArtifactContext@500abbf6, exclusiveTesting=false}')
  at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:515)
  at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:399)
  at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
  at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
  at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.NullPointerException
  at com.google.devtools.build.lib.exec.StandaloneTestStrategy.processTestAttempt(StandaloneTestStrategy.java:252)
  at com.google.devtools.build.lib.exec.StandaloneTestStrategy.finalizeTest(StandaloneTestStrategy.java:205)
  at com.google.devtools.build.lib.exec.StandaloneTestStrategy.access$300(StandaloneTestStrategy.java:68)
  at com.google.devtools.build.lib.exec.StandaloneTestStrategy$StandaloneTestRunnerSpawn.finalizeTest(StandaloneTestStrategy.java:461)
[...]

PR #11494, opened 2020-05-26
Prevent NPE on missing TEST_LOG due to cancel

werkt:
Tests added that exhibit this failure reported in #11468 without the
correction.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
@@ -14,6 +14,7 @@
 
 package com.google.devtools.build.lib.exec;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
```diff
@@ -240,20 +241,26 @@ private StandaloneFailedAttemptResult processTestAttempt(
...
       if (TestFileNameConstants.TEST_LOG.equals(pair.getFirst())) {
+        Preconditions.checkState(renamedTestLog == null, "multiple test_log matches");
         renamedTestLog = pair.getSecond();
+        status = dataBuilder.getStatus();
       }
     }
 
-    TestResultData.Builder dataBuilder = result.testResultDataBuilder();
-    if (dataBuilder.getStatus() == BlazeTestStatus.PASSED) {
+    if (status == BlazeTestStatus.PASSED) {
```
ulfjack, line 257: I would do this slightly differently. I think it might be nicer to check explicitly for `renamedTestLog == null` here as the first case, and then override the status in dataBuilder.

However, I'm also not sure how this can happen in the first place - why doesn't the code realize that the action got cancelled?

werkt: That's fine on the rework, I considered that but thought adding that check seemed a bit off - felt more consistent to have a status that we observed associated with the found TEST_LOG

I'm curious about that as well, but my guess is that some cancellation of a test strategy internal or post process is what triggered this after test cancellation detection took place. See the associated issue for proof of the capability to hit this itself.

ulfjack:
Thanks, I think this is better.

@meisterT

PR #11504, opened 2020-05-27
[MSVC] Fix Bazel unable to find tools.

Niram7777:


Niram7777:
@googlebot I signed it!

Issue #11506, opened 2020-05-27
clean command errors if Starlark-defined flags are provided


Yannic:
### Description of the problem / feature request:

I have set a Starlark-defined flag in my `.bazelrc` file:
```
build --@io_bazel_rules_go//go/config:incompatible_package_conflict_is_error=True
```

Since `bazel clean` inherits the command-line flags from `build`, running `bazel clean` fails with the following error message:
```
INFO: Options provided by the client:
  [...]
INFO: Reading rc options for 'clean' from <redacted>/.bazelrc:
  Inherited 'common' options: [...]
INFO: Reading rc options for 'clean' from <redacted>/.bazelrc:
  Inherited 'build' options: --@io_bazel_rules_go//go/config:incompatible_package_conflict_is_error=True  [...]
ERROR: Unrecognized arguments: --@io_bazel_rules_go//go/config:incompatible_package_conflict_is_error=True
```

It's unclear to me why `bazel clean` looks at any flags defined for `bazel build` (AFAICT, clean unconditionally deletes everything, so it probably shouldn't even look at `.bazelrc` `common` flags). At the very least, `bazel clean` should ignore Starlark-defined flags.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

clone `https://github.com/bazelbuild/rules_go`
run `bazel clean --@io_bazel_rules_go//go/config:incompatible_package_conflict_is_error=True`

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

3.1.0 / 3.2.0rc1

katre:
@juliexxia, is this fixed by your recent work on InfoCommand, or does that change (I think 2ec58f60ae1732f4c1d09330c4583050ac8d1f46) need to be repeated for CleanCommand?

Yannic:
@katre Thanks, I remember seeing that commit some time ago (but didn't remember enough about it to find it again). I also see it with `last_green` (`4d19c0095b791f0c4ccccfaea5f279e373469b24`), so it probably needs to be repeated for `bazel clean`.

juliexxia:
FWIW, you're totally right about clean not needing build commands and this is know technical debt for the bazel team (though not one that is prioritized right now): https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/runtime/commands/CleanCommand.java;l=58?q=CleanCommand&ss=bazel

[...]

PR #11508, opened 2020-05-27
update clean command to allow residue and ignore starlark options

juliexxia:
work towards #11506

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/CleanCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/CleanCommand.java
@@ -51,7 +51,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.logging.LogManager;
 
-/** Implements 'blaze clean'. */
+/**
+ * Implements 'blaze clean'.
+ */
 @Command(
     name = "clean",
     builds = true, // Does not, but people expect build options to be there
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/runtime/commands/CleanCommand.java
```diff
@@ -49,14 +53,15 @@
...
-  builds = true, // Does not, but people expect build options to be there
-  writeCommandLog = false, // Do not create a command.log, otherwise we couldn't delete it.
-  options = {CleanCommand.Options.class},
-  help = "resource:clean.txt",
-  shortDescription = "Removes output files and optionally stops the server.",
-  // TODO(bazel-team): Remove this - we inherit a huge number of unused options.
-  inherits = {BuildCommand.class}
+    name = "clean",
+    builds = true, // Does not, but people expect build options to be there
+    allowResidue = true, // Does not, but need to allow so we can ignore Starlark options.
```
juliexxia, line 58: not sure what git diff thinks is happening, I only added line 58

katre: The indentation changed (because the formatter considers this one statement to be reformatted when any part of it changes).

juliexxia:
I'm realizing this is going to be a little more complicated because the way we implemented for info doesn't properly ignore starlark flags from other repos. I think we should still submit the previous change as is and I'll make a second update.

PR #11510, opened 2020-05-27
Update sorting out of starlark flags to also identify external repo starlark flags

### LINKED ISSUE #11506: clean command errors if Starlark-defined flags are provided

juliexxia:
Fixes #11506

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
@@ -228,7 +228,7 @@ public class StarlarkOptionsParser {
     ImmutableList.Builder<String> keep = ImmutableList.builder();
     ImmutableList.Builder<String> remove = ImmutableList.builder();
     for (String name : list) {
-      ((name.startsWith("--//") || name.startsWith("--no//")) ? remove : keep).add(name);
+      (name.matches("--@?.*//.*") ? remove : keep).add(name);
     }
     return Pair.of(remove.build(), keep.build());
   }
--- a/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
@@ -23,6 +23,8 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
+import com.google.devtools.build.lib.cmdline.LabelValidator;
+import com.google.devtools.build.lib.cmdline.LabelValidator.BadLabelException;
 import com.google.devtools.build.lib.cmdline.TargetParsingException;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
 import com.google.devtools.build.lib.events.Reporter;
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/devtools/build/lib/skylark/StarlarkOptionsParsingTest.java
```diff
@@ -314,4 +319,16 @@ public void testOptionsAreParsedWithBuildTestsOnly() throws Exception {
 
     assertThat(result.getStarlarkOptions().get("//test:my_int_setting")).isEqualTo(15);
   }
+
+  @Test
+  public void testRemoveStarlarkOptionsWorks() throws Exception {
```
jin, line 324: Could you also add a test for the `--@//external/starlark/option` main repo case please?

src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
```diff
@@ -228,7 +228,7 @@ private Target loadBuildSetting(
     ImmutableList.Builder<String> keep = ImmutableList.builder();
     ImmutableList.Builder<String> remove = ImmutableList.builder();
     for (String name : list) {
-      ((name.startsWith("--//") || name.startsWith("--no//")) ? remove : keep).add(name);
+      (name.matches("--(no)?(@.*)?//.*") ? remove : keep).add(name);
```
katre, line 231: Regex-based parsing is going to fall into difficulties. Can this just strip off the -- and the no (if present), and call [LabelValidator.validateAbsoluteLabel](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/cmdline/LabelValidator.java;drc=61d663a503710c8cb0d03d18f2eb776981b49e17;l=227), and check for exceptions being thrown?

juliexxia: So much better. Done.

src/test/java/com/google/devtools/build/lib/skylark/StarlarkOptionsParsingTest.java
juliexxia: Yep! added

src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
```diff
@@ -228,7 +230,21 @@ private Target loadBuildSetting(
     ImmutableList.Builder<String> keep = ImmutableList.builder();
     ImmutableList.Builder<String> remove = ImmutableList.builder();
     for (String name : list) {
-      ((name.startsWith("--//") || name.startsWith("--no//")) ? remove : keep).add(name);
+      if (!name.startsWith("--")) {
```
katre, line 233: This logic is great, but needs comments. More comments.

juliexxia: Done! added a description to the javadoc as well

PR #11516, opened 2020-05-28
Remove javac.jar as a dependency of java tools

meteorcloudy:
We no longer need the javac.jar for building java tools if we're building with JDK11.

This helps remove javac.jar as a dependency of the Debian build, which
is a part of OpenJDK and won't be able to be packaged in Debian.

However, the javac jar will still exist in third_party because some tests depend on it.

Working towards: #9408

### DIFF
--- a/src/java_tools/buildjar/BUILD
+++ b/src/java_tools/buildjar/BUILD
@@ -61,7 +61,6 @@ java_toolchain(
     genclass = ["bootstrap_genclass_deploy.jar"],
     ijar = ["//third_party/ijar"],
     javabuilder = ["bootstrap_VanillaJavaBuilder_deploy.jar"],
-    javac = ["//third_party/java/jdk/langtools:javac_jar"],
     jvm_opts = [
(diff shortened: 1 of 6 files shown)


meteorcloudy:
@cushon Looks like we are able to remove javac.jar if we build the java tools with JDK 11.

I have to add extra `--javabase` option to the build script to make sure we use JDK 11, which is not what is said at https://docs.bazel.build/versions/master/bazel-and-java.html#configuring-the-jdk:
> Bazel’s default flags essentially look like this:
> ...
> `    --javabase=@bazel_tools//tools/jdk:remote_jdk11 \`
> ...

meteorcloudy:
OK.. it broken many targets if we have openJDK 8 installed..

meteorcloudy:
@olekw I think this will stay a patch on the Debian side until we totally give up JDK 8 on Bazel CI.

olekw:
Ok, I'll add this to my queue for Debian changes. Thanks!

PR #11521, opened 2020-05-29
Add documentation for running Bazel in an airgapped environment using distribution directories

jin:


site/docs/guide.md
```diff
@@ -327,20 +327,77 @@ be determined, for example to manually clean up the cache. The cache is never
...
+bazel build @additional_distfiles//:archives.tar
+```
+
+Export this tarball to a directory that can be copied into your airgapped
+environment. Note the `--strip-components` flag, because `--distdir` can be
+quite finicky with the directory nesting level:
+
+
+```
+tar xvf bazel-bin/external/additional_distfiles/archives.tar -C
```
aiuto, line 390: Perhaps split on two lines with a \\ at the end so it won't horizontal scroll and is still pastable.

Also, if you move --strip-compontns before -C NEWDIRECTORY, then someone can copy-paste the entire preamble and just type the distdir path.

jin: Resolving in the internal CL: cr/313751874

aiuto:
Thanks for that write-up.  ISTM that what we could use is a bazel mode where you can run a build and get a simple plain text file of each file we downloaded and the checksum.
Then someone could wget/curl each to their own distdir without needed bazel sources - which is just sort of ridiculous.

jin:
@aiuto We have that in some capacity with `bazel sync` and workspace resolved files.

site/docs/guide.md
```diff
@@ -327,20 +327,77 @@ be determined, for example to manually clean up the cache. The cache is never
...
+"$NEW_DIRECTORY" --strip-components=3
+```
+
+Finally, when you use Bazel in your airgapped environment, pass the `--distdir`
+flag pointing to the directory. For convenice, you can add it as an `.bazelrc`
```
gregestren, line 395: typo nit

aiuto:
> @aiuto We have that in some capacity with `bazel sync` and workspace resolved files.

Yes. But the implementation is inconvenient and hard to use. But I don't want to derail this PR with that discussion.

jin:
No worries, and I agree with your general idea. It would be useful for
audit purposes.

PR #11524, opened 2020-05-31
Add option rename dll on Windows

vnvo2409:
@meteorcloudy 
This PR will add an option rename dll on windows.
It won't change any name in the cc_binary and cc_library. It will only change the name when we copy the DLL to the binary directory and the name in the def file
https://github.com/bazelbuild/bazel/blob/1c03c8c4fac1e1028fcb5bb342da3a7fdfc88327/src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java#L1054

It is just a draft and it can not handle correctly the file name extension ( only dll for now )


src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -903,7 +918,23 @@ private static Artifact createEmptyDefFileAction(RuleContext ruleContext) {
...
+    if(cppConfiguration.isRenameDLL()) {
+      String dllName = ruleContext.getLabel().getName();
+      String dllPath = ruleContext.getRepository() 
+          + "/" 
+          + ruleContext.getPackageDirectory();
+      String dllHash = Hashing.sha256()
+          .hashString(dllPath, Charsets.UTF_8)
+          .toString();
+      String dllNameWithoutExtension = dllName
+          .replaceAll(".dll", "");
```
vnvo2409, line 932: We have to think about the situation where the extension is hardcorded (i.e name = "utils.so"). Maybe hash_name.dll is better than name_hash.dll

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
```diff
@@ -1046,13 +1048,29 @@ private static Packager createIntermediateDwpPackagers(
...
       RuleContext ruleContext, Iterable<Artifact> dynamicLibrariesForRuntime) {
     NestedSetBuilder<Artifact> result = NestedSetBuilder.stableOrder();
+    CppConfiguration cppConfiguration = ruleContext.getFragment(CppConfiguration.class);
     for (Artifact target : dynamicLibrariesForRuntime) {
       if (!ruleContext
           .getLabel()
           .getPackageIdentifier()
-          .equals(target.getOwner().getPackageIdentifier())) {
+          .equals(target.getOwner().getPackageIdentifier())
+          || cppConfiguration.isRenameDLL()) {
```
vnvo2409, line 1057: We have to check `isRenameDLL()` here since we wrote `name_hash.dll` instead of `name.dll` to the DEF file in the first place. It will rename the dll to `name_hash.dll` even if the dll is in the same directory. It causes a problem when there are many rules depend on the same dll but we could check if `name_hash.dll` is existed before rename.
```
cc_binary(
  name = "a",
  deps = [":c"],
)

cc_binary(
  name = "b",
  deps = [":c"],
)

cc_library(
  name = "c",
)
```
If we want to use the truly benefit of shared linked library (i.e one dll for many binary/library), we may want to create a folder to store dll and `set PATH=c:\bazel\MyDLLFolder;%PATH%`. `set` applies to the current command prompt only so we will be fine

src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
```diff
@@ -244,6 +244,20 @@ public String getTypeDescription() {
...
   public DynamicMode dynamicMode;
 
+  @Option(
+    name = "rename_dll",
+    defaultValue = "false",
+    documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+    effectTags = {OptionEffectTag.LOADING_AND_ANALYSIS, OptionEffectTag.AFFECTS_OUTPUTS},
+    help =
+        "If enabled, all C++ dynamic linked libraries (DLLs) on Windows will be renamed "
+            + "to name_hash.dll where hash is caculated based on the RepositoryName and "
```
meteorcloudy, line 254: Nit: name_{hash}.dll

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
meteorcloudy: I think it's better to do the renaming in CcLibrary when we generate the DLL than in CcBinary when we copy the DLLs. Because

- `dynamicLibrariesForRuntime` may contain not only DLLs generated by CcLibrary but also DLL imported by ` cc_import` rule. If we rename the DLL, it won't work.
- Ww put the renamed DLL name in DEF file in CcLibrary, which will generate a LIB file (interface library) that is not consistent with the DLL generated by the same CcLibrary rule. I think that's not a good thing.

In CcLibrary, you can probably add a `linkedArtifactNameSuffix` pretty easily here:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java;l=629?q=CcLibrary.java&ss=bazel%2Fbazel

To make sure `--rename_dll` flag only affects Windows build, you can always check `featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)`.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
meteorcloudy: Yes. `createEmptyDefFileAction` is used by both CcBinary and CcLibrary. We don't actually want to change anything for CcBinary. You'll have to distinguish them. Maybe add an optional argument to specify the DLL name?

And to generate the hash, can you do something similar to:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java;l=288?q=getTmpDirName&ss=bazel%2Fbazel

Using the Fingerprint class seems to be a nicer approach. And to avoid [the long path issues](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation) on Windows, can you truncate the hash to 16 or maybe 8 characters?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcBinary.java
vnvo2409: Thank you for your reviews. And how about the idea creating a folder to store dll and `set PATH=c:\bazel\MyDLLFolder;%PATH%`. If it is good, I will implement it in this PR too and we could change the flag to something like `--better_dll_support`

vnvo2409:
@meteorcloudy 
Thank you for your reviews. And how about the idea creating a folder to store dll and `set PATH=c:\bazel\MyDLLFolder;%PATH%`. If it is good, I will implement it in this PR too and we could change the flag to something like `--better_dll_support`

meteorcloudy:
> And how about the idea creating a folder to store dll and set PATH=c:\bazel\MyDLLFolder;%PATH%. 

How's this gonna work? I've considered this approach but there is no way to set the `PATH` for the users during runtime, right?

For example, after we build a binary via `bazel build //foo/bar:bin` and try to run `./bazel-bin/foo/bar/bin.exe`. how do you set the `PATH` env automatically?

vnvo2409:
@meteorcloudy 
Sorry I only think about Tensorflow when everything get packed into one. How about a static folder for all projects ? We could have a flag to enable this feature. `set` affects only the current command prompt but `setx` will affect all command prompt.

meteorcloudy:
> How about a static folder for all projects ?

What if there are DLLs with the same name from different projects?

vnvo2409:
the hash will be different since the hash is calculated by repository name

meteorcloudy:
I see, so you still want to keep the renaming approach. But there are still few problems:

- Different projects can actual have the same repository name or depends on the same repository with different versions. We'll have to create such directory for each project.
- `setx` might require admin right?
- If a binary is built dynamically and used as a tool in the build. When Bazel run the action with this tool, Bazel might prune the environment variables, including `PATH`. This will cause the binary to crash at runtime.
- If you try to deploy the binary to other machines, it's not obvious to figure out what DLLs the binary actually depends on.

In general, I'm not so willing to change user's environment variable because it's hard to control. Maybe we can first make the renaming approach work, and discuss this later?

vnvo2409:
>Maybe we can first make the renaming approach work, and discuss this later?

i am thinking the same as you.

vnvo2409:
@meteorcloudy 
A new implementation. In this implementation, I change the name of DLL generated by `cc_library` to `name_{hash}.dll` in the first place and keep the names untouched when call `SymlinkAction`

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
```diff
@@ -620,13 +630,20 @@ public static boolean appearsToHaveObjectFiles(AttributeMap rule) {
 
     if (!ruleContext.attributes().get("linkstatic", Type.BOOLEAN)
         && !ccCompilationOutputs.isEmpty()) {
+      String linkedArtifactNameSuffix = "";
+      if(cppConfiguration.isRenameDLL()) {
```
meteorcloudy, line 634: The code for calculating the dll suffix repeated twice in the change. Can you extract it to a function? Maybe in CppHelper?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -117,6 +117,7 @@ public CcLinkingOutputs getCcLinkingOutputs() {
   private final CcToolchainProvider ccToolchain;
   private final FdoContext fdoContext;
   private String linkedArtifactNameSuffix = "";
+  private String linkedDLLNameSuffix = "";
```
meteorcloudy, line 120: Can we just set the `linkedArtifactNameSuffix` in CcLibrary instead of introducing a new variable?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -857,6 +871,10 @@ private CppLinkActionBuilder newLinkActionBuilder(
    */
   private Artifact getLinkedArtifact(LinkTargetType linkTargetType) throws RuleErrorException {
       String maybePicName = label.getName() + linkedArtifactNameSuffix;
+      if(featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)
```
meteorcloudy, line 874: This should not be necessary if we set `linkedArtifactNameSuffix` in CcLibrary

src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
```diff
@@ -244,6 +244,20 @@ public String getTypeDescription() {
   )
   public DynamicMode dynamicMode;
 
+  @Option(
+    name = "rename_dll",
```
meteorcloudy, line 248: After some discussion with our C++ rules owner. We want to enable this behaviour by default in future. However, it should be an incompatible change for Bazel, so it has to follow our [incompatible change policy](https://docs.bazel.build/versions/master/backward-compatibility.html#incompatible-changes-and-migration-recipes)
Can you rename the flag to maybe `--incompatible_avoid_conflict_dlls`? And please file another issue to track this incompatible change. Thanks!

src/main/java/com/google/devtools/build/lib/rules/cpp/CppConfiguration.java
```diff
@@ -324,6 +325,11 @@ public DynamicMode getDynamicModeFlag() {
     return cppOptions.dynamicMode;
   }
 
+  public boolean isRenameDLL() {
+    return cppOptions.renameDLL && (cppOptions.dynamicMode == DynamicMode.FULLY)
+        && (OS.getCurrent() == OS.WINDOWS);
```
meteorcloudy, line 330: I think we shouldn't check `cppOptions.dynamicMode == DynamicMode.FULLY` because even if the dynamic mode is auto, cc_library can still be built dynamically.

And if you want to check the platform we're building for, it's better to use `featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)` if possible. Because in the future, Bazel may runs on Linux but builds for Windows.

vnvo2409: So I will check `cppOptions.dynamicMode != DynamicMode.OFF`

vnvo2409:
@meteorcloudy 
- The problem with `linkedArtifactNameSuffix` is deeper than we thought. 
If we only add `linkedArtifactNameSuffix` to https://github.com/bazelbuild/bazel/blob/ca41a9fd49b485a4219fc3227031624ed3af3bb1/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java#L640-L646
We have a problem with https://github.com/bazelbuild/bazel/blob/eb048fd3bb74d05d6f65edeb6f29161bc0da5c97/src/main/java/com/google/devtools/build/lib/analysis/CachingAnalysisEnvironment.java#L289-L295
Because there is no action associated with the `name_{hash}.dll`. `name_{hash}.dll` should be created by the action `CppLinkingAction`. But this action is associated with `name.dll` because 
`ccCompilationOutputs` only contains information about the rule (i.e the original name of the library)
https://github.com/bazelbuild/bazel/blob/ca41a9fd49b485a4219fc3227031624ed3af3bb1/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java#L348
If we use this function, it will change the name of some orther artifacts (.lib, .pdb, ...) which we don't want. So I have too add a new variable to control what is add to the name of the dll. https://github.com/bazelbuild/bazel/blob/ca41a9fd49b485a4219fc3227031624ed3af3bb1/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java#L311
If you want, I will sent you the screenshot of the error if we dont use a new variable.

- Where should I put my test, I really don't know I should create a new file or put into existing file ?

meteorcloudy:
Thanks for the explanation! 

> If we use this function, it will change the name of some orther artifacts (.lib, .pdb, ...)

Maybe it's not a bad idea to also add the same suffix for corresponding .lib. pdb files? @oquenchil WDYT?

> Where should I put my test, I really don't know I should create a new file or put into existing file ?

You can add integration test in `bazel_windows_cpp_test.py`

vnvo2409:
@meteorcloudy 
If we have a cc_library name utils. The artifacts which are renamed are:
- utils_hash.dll
- utils_hash.dll-2.params
- utils_hash.if.exp
- utils_hash.if.lib
- utils_hash.pdb

utils.obj and .def aren't renamed.

I don't know how to get the `featureConfiguration` directly so I have to pass it as an argument.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
```diff
@@ -615,7 +618,7 @@ public static boolean appearsToHaveObjectFiles(AttributeMap rule) {
               ccToolchain,
               configuration,
               Link.LinkTargetType.STATIC_LIBRARY,
-              /* linkedArtifactNameSuffix= */ ""));
+              linkedArtifactNameSuffix));
```
meteorcloudy, line 621: Ah... we should not add suffix to static library. But I looked at the code and found no good way to only apply   linkedArtifactNameSuffix to dynamic library in CcLinkingHelper. Looks like we still have to have an additional attribute for this. They we can apply the DLL suffix based on the `linkTargetType`. Sorry about the back and forth.

vnvo2409:
@meteorcloudy 
I will write the test tonight. I have tested with my stand-alone tests before commiting so don't worry.
So we will revert to the previous implementation where we add a variable to `CcLinkingHelper` ?

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -312,6 +313,19 @@ public CcLinkingHelper setLinkedArtifactNameSuffix(String suffix) {
...
   }
 
+  /*
+   * Adds a suffix (_{hash}) for path of linked DLL. This will be enabled only when rename_dll is enabled
+   * and the linked artifact's LinkTargetType is NODEPS_DYNAMIC_LIBRARY (i.e DLL generated by cc_library).
+   * We have to add the suffix to make sure the CppLinkAction link against the renamed DLL. If not, CppLinkAction
+   * will link against the DLL whose name is the same as the name of cc_library.
+   */
+  public CcLinkingHelper setLinkedDLLNameSuffix(String suffix) {
+    if(this.cppConfiguration.isRenameDLL(featureConfiguration)) {
```
meteorcloudy, line 323: Why do we need to check it again after we already check if there should be a suffix in CppHelper?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppConfiguration.java
```diff
@@ -324,6 +325,11 @@ public DynamicMode getDynamicModeFlag() {
     return cppOptions.dynamicMode;
   }
 
+  public boolean isRenameDLL(FeatureConfiguration featureConfiguration) {
```
meteorcloudy, line 328: I think this function maybe not necessary?

src/test/py/bazel/bazel_windows_cpp_test.py
```diff
@@ -278,19 +280,26 @@ def testBuildDifferentCcBinariesDependOnConflictDLLs(self):
             '  linkstatic = 0,'
             ')',
         ])
+    bazel_bin = self.getBazelInfo('bazel-bin')
     self.ScratchFile('main/other_main.cc', ['int main() {return 0;}'])
 
     # Building //main:main should succeed
-    exit_code, _, stderr = self.RunBazel(['build', '//main:main'])
+    # We should check the A_{hash}.dll exist.
```
meteorcloudy, line 287: Maybe check if there is a file name as a pattern like "A_*.dll"? We don't have to calculate the actual hash.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
```diff
@@ -857,6 +871,9 @@ private CppLinkActionBuilder newLinkActionBuilder(
    */
   private Artifact getLinkedArtifact(LinkTargetType linkTargetType) throws RuleErrorException {
       String maybePicName = label.getName() + linkedArtifactNameSuffix;
+      if(linkTargetType == LinkTargetType.NODEPS_DYNAMIC_LIBRARY) {
+        maybePicName += linkedDLLNameSuffix;
+      }
```
meteorcloudy, line 876: Can we move this to line 882? Like
```
String linkedName = maybePicName + (linkTargetType == LinkTargetType.NODEPS_DYNAMIC_LIBRARY ? linkedDLLNameSuffix : "");
linkedName =
    CppHelper.getArtifactNameForCategory(
              ruleErrorConsumer, ccToolchain, linkTargetType.getLinkerOutput(), linkedName);
```

It might make the code easier to understand.

vnvo2409:
@meteorcloudy 
I have removed some unused part as you suggest. And I have added tests as well. I think this PR is ready to merge. I will create an issue to track incompatible change after this PR is merged

PR #11535, opened 2020-06-02
Introduce `pythonShebang` variable for nix packages collection

avdv:
Building on Nix / NixOS we need to avoid using `/usr/bin/env` to determine the python interpreter (or other executables). 

This change makes it easier for us to replace the shebang line when building the derivation of bazel with a fixed path to a python3 executable.

See NixOS/nixpkgs#87726 for details.


avdv:
@googlebot I signed it!

avdv:
> As far as I can see, this is a no-op change that makes the nix diff cleaner.

Yes, exactly.

jin:
cc @brandjon

laurentlb:
Should we add a comment in the code about it? Otherwise any refactoring might affect you.

avdv:
> Should we add a comment in the code about it? Otherwise any refactoring might affect you.

Good point! I'll add one.

avdv:
> Should we add a comment in the code about it? Otherwise any refactoring might affect you.

Done.

I am not sure, should I squash both commits or what do you prefer?

jin:
No need to squash -- the import process takes care of it. Thanks!

Issue #10902, opened 2020-03-04
Bazel query silently thrashes the analysis cache


kastiglione:
### Description of the problem / feature request:

`bazel query` invalidates the analysis cache for subsequent builds, and neither the output of `bazel build` nor `bazel query` indicate the problem.

The best solution would be to call `bazel query` in a way that uses and preserves the current analysis cache. A secondary solution is to have one or both commands log about the problem.

`bazel build` does output a message to say that the analysis cache is being discarded – when build flags are changed. However `bazel query` is silently causing discarding the analysis cache, so we didn't know this was happening.

Chrome traces were showing 10-15s of overhead at the start of each build, and we didn't know why. Removing our use of `bazel query` between builds solved this problem.

### Feature requests: what underlying problem are you trying to solve with this feature?

Allow `bazel query` to reuse and preserve the analysis cache from the recent `bazel build`, if applicable.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Do a build
2. Change a file
3. Do the same build again

In the second build, the output should say:

```
INFO: Analyzed target //package:target (0 packages loaded, 0 targets configured).
```

Now, to see the bug:

1. Do a build
2. **Do a bazel query**
3. Change a file
4. Do the same build again

In the second build, the output will show that bazel has had to load >0 packages and configure >0 targets.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 2.1.0

haxorz:
@kastiglione Please provide more information. Your certainly on the right track that `bazel` flags are the culprit, but haven't told us anything about your flag usage. I think bazel's INFO log suffices to fully debug this. If you can't upload the full file for sensitivity reasons, please read over it and copy/paste the relevant parts. Someone on the Bazel team can help you decipher the log if you need help.

@irengrig Please find a different assignee. I don't have time to look at this.

irengrig:
Removing P1, we should indeed get more information.

/cc @lberki can you help investigate the problem?

gregestren:
@kastiglione - is it possible that you're setting flags in a `.bazelrc` that apply to `build` but not `query`?

kastiglione:
@gregestren yes, that's a possible cause. After posting this, we learned about #10961. In particular, we `common` settings were the cause of thrashing with `bazel run`. https://github.com/bazelbuild/bazel/issues/10961#issuecomment-598844709

[...]

PR #11536, opened 2020-06-02
Fetching symlink inputs for top-level output

### LINKED ISSUE #11532: remote_download_outputs=toplevel doesn't download libraries in `cc_import` needed to run.

I have a `cc_binary` target which depends on a DSO in the following way.

```
cc_binary(
  name = "my_binary",
  deps = [":stuff", ":import_lib"],
)

cc_library(
  name = "import_lib",
  deps = [":libfoo_lib", ":another_thing"]
)

cc_import(
  name = "libfoo_lib",
  shared_library = ":libfoo.so",
)

cc_binary(
  name = "libfoo.so",
  srcs = ["blah.cpp"],
  linkshared = 1,
  linkstatic = 1,
)
```

When a machine does `bazel run :my_binary`, it's not downloading `libfoo.so` and then producing an error like:
```
error while loading shared libraries: libfoo.so: cannot open shared object file: No such file or directory
```

It seems that it doesn't recognise that `libfoo.so` is necessary for execution.

Ubuntu 16.04

release 1.2.1

ulfjack:
Top-level output files can be symlinks to non-top-level output files, such as
in the case of a cc_binary linking against a .so built in the same build.

I reuse the existing mayInsensitivelyPropagateInputs() method, which was
originally introduced for action rewinding, but seems to have exactly the
intended semantics here as well.

Unfortunately, this requires a small change to the BlazeModule API to pass in
the full analysis result (so we can access the action graph, so we can read
the aformentioned flag).

I considered using execution-phase information on whether an output is a
symlink. However, at that point it's too late! The current design only allows
pulling an output file *when its generating action runs*. It would be better if
this was changed to pull output files independently of action execution. That
would avoid a lot of problems with the current design, such as #10902.

Fixes #11532.

Change-Id: Iaf1e48895311fcf52d9e1802d53598288788a921


ulfjack:
@buchgr (who wrote the code originally)
@mschaller (because I'm abusing a flag in Action that is used by rewinding)

benjaminp:
itym @anakanemison

ulfjack:
@anakanemison is not linked to Google. Might want to do that?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -511,6 +513,17 @@ public void afterAnalysis(
...
       }
+      // Also fetch all inputs for actions that propagate their inputs (symlink actions).
+      for (ActionInput actionInput : filesToDownload.build()) {
+        if (!(actionInput instanceof Artifact)) {
+          continue;
+        }
+        ActionExecutionMetadata action =
+            (ActionExecutionMetadata) analysisResult.getActionGraph().getGeneratingAction((Artifact) actionInput);
+        if (action.mayInsensitivelyPropagateInputs()) {
+          filesToDownload.addAll(action.getInputs().toList());
```
anakanemison, line 524: Is it possible for there to be more than one symlink-ish action chained together? My concern is that this may be insufficient if the inputs to this action are outputs of other symlink actions. Rewinding recurses in such a case. Perhaps that would be overkill here.

src/main/java/com/google/devtools/build/lib/runtime/BlazeModule.java
```diff
@@ -282,13 +283,13 @@ public BuildOptions getDefaultBuildOptions(BlazeRuntime runtime) {
    * @param env the command environment
    * @param request the build request
    * @param buildOptions the build's top-level options
-   * @param configuredTargets the build's requested top-level targets as {@link ConfiguredTarget}s
+   * @param analysisResult the build's requested top-level targets as {@link ConfiguredTarget}s
```
anakanemison, line 286: this description deserves some rewording, given that the parameter's type changed.

anakanemison:
Sorry for the delay. I'm not used to GitHub! I think I'm in the Google org on GitHub already.  
But I wasn't in bazelbuild, until Tony extended me an invite tonight!

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
ulfjack: Good catch.

src/main/java/com/google/devtools/build/lib/runtime/BlazeModule.java
ulfjack: Thanks, reworded.

ulfjack:
No worries, and thanks for the review. GitHub seems to think that @anakanemison is not in the Google org. For example, you can compare your profile page with Jakob's: @buchgr vs. @anakanemison 

Did you add your @google.com address to your GitHub profile?

anakanemison:
Thanks for pointing out my empty profile! I entered some information there. (Looks like I could say anything I wanted there..)

LGTM. I'll do the internal merging thing and respond here when the change is in.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -496,22 +501,47 @@ public void afterAnalysis(
...
+      // Only fetch recursively if we don't have the file to avoid visiting nodes multiple times.
+      if (builder.add(input)) {
+        fetchSymlinkDependenciesRecursively(actionGraph, builder, input);
+      }
+    }
+  }
+
+  private static void fetchSymlinkDependenciesRecursively(
+      ActionGraph actionGraph, Set<ActionInput> builder, Artifact artifact) {
+    ActionExecutionMetadata action = (ActionExecutionMetadata) actionGraph.getGeneratingAction(artifact);
```
janakdr, line 541: what happens if a top-level artifact is a tree artifact generated by a template action? I think that can happen?

If so, then it should be ok to ignore non-ActionExecutionMetadata instances, since templates shouldn't insensitively propagate their inputs (although I guess theoretically there could be a template action that just symlinked its inputs?). Could pull #mayInsensitivelyPropagateInputs() up to ActionAnalysisMetadata too.

ulfjack: Good catch. Right now that leads to a Bazel crash. Fixed!

The main issue right now are SymlinkAction and SymlinkTreeAction, but in general, any remotely run action can return a symlink in the open-source remote execution API. Therefore, this will eventually need to be rewritten to follow symlinks at execution time. The current design mixes up action execution and output fetching, and separating them should actually result in a nicer implementation, because we can then remove the code that injects remote execution flags into Skyframe.

So yeah, this is a short-term fix to get cc_binary to work properly, and there's more work needed.

ulfjack:
Added a test for top-level template actions and fixed the resulting ClassCastException. PTAL.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
janakdr: Would you mind adding a TODO for that somewhere appropriate?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -538,6 +538,12 @@ private static void fetchSymlinkDependenciesRecursively(
 
   private static void fetchSymlinkDependenciesRecursively(
       ActionGraph actionGraph, Set<ActionInput> builder, Artifact artifact) {
+    if (!(actionGraph.getGeneratingAction(artifact) instanceof ActionExecutionMetadata)) {
+      // The top-level artifact could be a tree artifact, in which case the generating action may
+      // be an ActionTemplate, which does not implement ActionExecutionMetadata. We don't handle
+      // this case right now, so exit.
```
janakdr, line 544: Is the concern here that an ActionTemplate might create symlinks? I think that currently ActionTemplate is only implemented by CppCompileActionTemplate and SpawnActionTemplate. SpawnActionTemplate has no callers in Bazel, so I think there's no deficiency here as it stands. If we ever allow Starlark actions to create ActionTemplates, then I think there will be a gap here.

ulfjack: Yes. Agreed that current templates don't have this problem, but we expect to see it in the future when / if Starlark is allowed to create them or if we add more templates.

ulfjack:
Added a TODO as requested. Sorry for the delay - the last month was pretty hectic.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
janakdr: Sorry, now that I understand this better, could you add a test exhibiting this behavior? I assume it would be a Starlark action that created a symlink on the remote worker?

The behavior of creating symlinks on the remote worker seems very problematic to me. What happens if the action creates an intermediate file and then its final output is a symlink to that file? We'll just download a dangling symlink? Is there any chance of getting rid of this symlinking feature altogether?

ulfjack:
Bazel currently errors out if the remote executor returns a dangling symlink, and also if it returns a symlink when build-without-the-bytes is enabled.

ulfjack:
Not quite: it returns an error if an action generates a symlink and --remote_download_minimal is active, but not if --remote_download_toplevel is active. I added a test for this.

ulfjack:
Note that genrules can also generate symlinks on the remote machine.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -515,26 +520,63 @@ public void afterAnalysis(
...
+      ActionGraph actionGraph, Set<ActionInput> builder, Artifact artifact) {
+    if (!(actionGraph.getGeneratingAction(artifact) instanceof ActionExecutionMetadata)) {
+      // The top-level artifact could be a tree artifact, in which case the generating action may
+      // be an ActionTemplate, which does not implement ActionExecutionMetadata. We don't handle
+      // this case right now, so exit.
+      return;
+    }
+    ActionExecutionMetadata action = (ActionExecutionMetadata) actionGraph.getGeneratingAction(artifact);
+    if (action.mayInsensitivelyPropagateInputs()) {
+      List<Artifact> inputs = action.getInputs().toList();
```
janakdr, line 578: Would you mind adding a warning-level logline here if the list is >5 elements or so? I don't think we expect such actions to insensitively propagate inputs, and if there's a bug, we could slow down significantly due to nested set expansion.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -496,22 +501,60 @@ public void afterAnalysis(
...
       BuildOptions buildOptions,
-      Iterable<ConfiguredTarget> configuredTargets) {
+      AnalysisResult analysisResult) {
     if (remoteOutputsMode != null && remoteOutputsMode.downloadToplevelOutputsOnly()) {
       Preconditions.checkState(actionContextProvider != null, "actionContextProvider was null");
       boolean isTestCommand = env.getCommandName().equals("test");
       TopLevelArtifactContext artifactContext = request.getTopLevelArtifactContext();
-      ImmutableSet.Builder<ActionInput> filesToDownload = ImmutableSet.builder();
-      for (ConfiguredTarget configuredTarget : configuredTargets) {
+      Set<ActionInput> filesToDownload = new HashSet<>();
```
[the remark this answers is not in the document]
ulfjack: This includes test outputs which are not declared as artifacts, so no, not easily.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
ulfjack: Done.

PR #11537, opened 2020-06-02
Fix `--run_under` Windows escaping bug

### LINKED ISSUE #10993: Windows: `--run_under` invokes bash incorrectly

When using the `--run_under` flag, Bazel tries and fails to invoke bash (hypothesis: _iff_ the path the Bash executable contains a space, i.e. the default `c:/tools/msys64/usr/bin/bash.exe` works as expected, the path `C:\Program Files\Git\usr\bin\bash.exe` does not).
 
The path to Bash is set to: `"C:\Program Files\Git\usr\bin\bash.exe"` using `BAZEL_SH`, Bazel and then constructs and executes the path:
```
FATAL: ExecuteProgram(c:\users\robinn~1\_bazel~1\w477h2zl\execroo
t\__main__\bazel-~1\x64_wi~1\bin\main\hello-~2.run\"C:\Program Fi
les\Git\usr\bin\bash.exe")
```

Changing the path to a relative path e.g. to `bash`, appear to exhibit the same behavior, e.g: 
```
c:\users\robinn~1\_bazel~1\w477h2zl\execroo
t\__main__\bazel-~1\x64_wi~1\bin\main\hello-~2.run\bash
```

A brief look at the source code seems to point to https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java;l=237;drc=d094e94ac6ba905483e6bc4a5afe6ad5ec8097b5;bpv=1;bpt=1
doing something incorrect. Specifically, it _looks_ like the the path creation doesn't recognize the fact that the path is absolute already. Withing the context of `bazel test` there is a mechanism to detect whether or not bash is required, (e.g. [`needsShell`](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/test/TestTargetExecutionSettings.java;l=190?q=needsShell&sq=&ss=bazel)) but this does not appear to happen for `bazel run`.

More digging:
 - The Path to Bazel is normalized using: https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/vfs/WindowsOsPathPolicy.java;l=111;drc=a6de3825bbe67ea916024ae91f5674121fd7c36e;bpv=1;bpt=1	

It may also be a bug that Bazel invokes Bash at all since Bazel 2.2.0 removed the test and run bash wrapping mechanism.

Run any binary with `--run_under` on Windows.

1. Checkout the Bazel [examples](https://github.com/bazelbuild/examples/tree/master/cpp-tutorial/stage3) repo.
2. `cd cpp-tutorial/stage3`
3. `bazel run main:hello-world --run_under=cmd`
4. Error occurs, e.g.
```
FATAL: ExecuteProgram(c:\users\robinn~1\_bazel~1\w477h2zl\execroo
t\__main__\bazel-~1\x64_wi~1\bin\main\hello-~2.run\"C:\Program Fi
les\Git\usr\bin\bash.exe") failed: ERROR: src/main/native/windows
/process.cc(199): CreateProcessW("c:\users\robinn~1\_bazel~1\w477
h2zl\execroot\__main__\bazel-~1\x64_wi~1\bin\main\hello-~2.run\"C
:\Program Files\Git\usr\bin\bash.exe""  -c "cmd C:/users/robinnab
el/_bazel_robinnabel/w477h2zl/execroot/__main__/bazel-out/x64_win
dows-fastbuild/bin/main/hello-world.exe "): Access is denied.
 (error: 5)
```

Windows 10.

2.2.0

Related to #8229

RNabel:
Fixes #10993.

The argument escaping that happens in [line 517](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java#L516) violates the invariant assumed by [`AsWindowsPath()`](https://github.com/bazelbuild/bazel/blob/master/src/main/cpp/util/path_windows.cc#L245) on the client, which is called when ensuring that the binary path is absolute ([`AsAbsoluteWindowsPathImpl`](https://github.com/bazelbuild/bazel/blob/master/src/main/cpp/util/path_windows.cc#L323)).

`AsWindowsPathImpl` clearly assumes that the passed string is not quoted (i.e. it is either empty, `/` or contains a drive letter: `C:\`, etc.)

PR verification: Follow the steps outlined in the repro case for #10993, but use `--run_under=":;"` which is a valid prefix for a windows binary run from within bash (the original repro case used --run_under=cmd, which is not a valid way to execuate a binary if I understand `cmd`'s cryptic error message correctly).

Testing: The PR is missing a test case - any pointers as to how to add a unit test for this bug would be much appreciated.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
@@ -511,9 +511,15 @@ public class RunCommand implements BlazeCommand  {
             ByteString.copyFrom(workingDir.getPathString(), StandardCharsets.ISO_8859_1));
 
     if (OS.getCurrent() == OS.WINDOWS) {
+      boolean isBinary = true;
       for (String arg : cmdLine) {
-        execDescription.addArgv(
-            ByteString.copyFrom(ShellUtils.windowsEscapeArg(arg), StandardCharsets.ISO_8859_1));
+        if (!isBinary) {
+          // All but the first element in `cmdLine` have to be escaped. The first element is the
+          // binary, which must not be escaped.
+          arg = ShellUtils.windowsEscapeArg(arg);
+        }
+        execDescription.addArgv(ByteString.copyFrom(arg, StandardCharsets.ISO_8859_1));
+        isBinary = false;
       }
     } else {
       PathFragment shExecutable = ShToolchain.getPath(configuration);
--- a/src/test/shell/bazel/run_test.sh
+++ b/src/test/shell/bazel/run_test.sh
@@ -94,3 +94,27 @@ eof
   bazel run --enable_runfiles=no $pkg:x >&$TEST_log || fail "expected success"
   bazel run --enable_runfiles=yes $pkg:x >&$TEST_log || fail "expected success"
 }
+
+function test_windows_argument_escaping() {
+  if ! "$is_windows"; then
+    return # Run test only on Windows.
+  fi
+
+  local -r pkg="pkg${LINENO}"
+  mkdir $pkg
+  cat >$pkg/BUILD <<eof
+sh_binary(
+  name = "a",
+  srcs = [":a.sh"],
+)
+eof
+  cat >$pkg/a.sh <<eof
+echo Hello World
+eof
+  disable_errexit
+  output="$(BAZEL_SH="C:/first_part second_part" bazel run --run_under=":;" $pkg:a 2>&1)"
+  enable_errexit
+  echo "$output" | grep --fixed-strings 'ExecuteProgram(C:\first_part second_part)' || fail "Expected error message to contain unquoted path"
+}
+
+run_suite "run_under_tests"
--- a/src/test/shell/bazel/run_test.sh
+++ b/src/test/shell/bazel/run_test.sh
@@ -112,6 +112,16 @@ eof
 echo Hello World
 eof
   disable_errexit
+  # This test uses the content of the Bazel error message to test that Bazel correctly handles
+  # paths to Bash which contain spaces - which is needed when using --run_under as it
+  # unconditionally results in Bazel wrapping the executable in Bash.
+  #
+  # The expected error message starts with:
+  # ```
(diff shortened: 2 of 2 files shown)


RNabel:
Hey @katre, since you recently touched this file, can you point me to the code owner of this file, so I can add them?

katre:
I can review it.

src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
```diff
@@ -511,9 +511,12 @@ public BlazeCommandResult exec(CommandEnvironment env, OptionsParsingResult opti
             ByteString.copyFrom(workingDir.getPathString(), StandardCharsets.ISO_8859_1));
 
     if (OS.getCurrent() == OS.WINDOWS) {
+      boolean isBinary = true; // The first element in `cmdLine` is the binary, which must not be escaped.
```
katre, line 514: I'd put the comment right before the call to windowsEscapeArg, and also split up the ternary operator into an actual if/else statement.

RNabel: Done + altered the comment to reflect its new location.

src/test/shell/bazel/run_test.sh
```diff
@@ -94,3 +94,27 @@ eof
...
+sh_binary(
+  name = "a",
+  srcs = [":a.sh"],
+)
+eof
+  cat >$pkg/a.sh <<eof
+echo Hello World
+eof
+  disable_errexit
+  output="$(BAZEL_SH="C:/first_part second_part" bazel run --run_under=":;" $pkg:a 2>&1)"
```
katre, line 115: Add a comment explaining why this run_under value was chosen and what the expected result of this command is.

The details in the PR are useful, but should be in the actual test for future readers.

RNabel: Done.

RNabel: The value of `--run_under` turns out to be irrelevant, let me know if you still want this bit of information in the comment.

Issue #9406, opened 2019-09-19
C++ Coverage on Linux is broken


ruifangChen:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:
When I execute the bazel coverage command with version 0.24.1, I can generate a valid coverage.dat file, but when using a version higher than 0.24.1, the coverage.dat file is empty.
This problem exists only when generating coverage for the C++ language.

$ ll bazel-testlogs/id_test/
total 3136
drwxr-xr-x 2 admin admin    4096 Sep 19 15:29 ./
drwxr-xr-x 3 admin admin    4096 Sep 19 15:07 ../
-r-xr-xr-x 1 admin admin  581600 Sep 19 15:07 baseline_coverage.dat*
-r-xr-xr-x 1 admin admin   68477 Sep 19 15:29 coverage.dat*
-r-xr-xr-x 1 admin admin     179 Sep 19 15:29 test.cache_status*
-r-xr-xr-x 1 admin admin 2539566 Sep 19 15:29 test.log*
-rw-r--r-- 1 admin admin     904 Sep 19 15:28 test.xml
$ bazel version
Build label: 0.24.1
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Apr 2 16:29:26 2019 (1554222566)
Build timestamp: 1554222566
Build timestamp as int: 1554222566

$ ll bazel-testlogs/id_test/
total 4108
drwxr-xr-x 2 admin admin    4096 Sep 19 15:25 ./
drwxr-xr-x 3 admin admin    4096 Sep 19 15:07 ../
-r-xr-xr-x 1 admin admin  581600 Sep 19 15:07 baseline_coverage.dat*
-r-xr-xr-x 1 admin admin       0 Sep 19 15:25 coverage.dat*
-r-xr-xr-x 1 admin admin     176 Sep 19 15:25 test.cache_status*
-r-xr-xr-x 1 admin admin 3600923 Sep 19 15:25 test.log*
-rw-r--r-- 1 admin admin     904 Sep 19 15:25 test.xml
$ bazel version
Build label: 0.25.0
[...]

PR #11538, opened 2020-06-02
Fix GCC 9 Code coverage by adding its JSON format

helaan:
GCC 9 changed the intermediate format to a JSON based format and with it
changed the meaning of the `-i` flag. Because of these changes it was not
possible to generate code coverage with GCC 9. This patch addresses that
by adding its format next to the existing GCov parser.

Addresses: #9406

### DIFF
--- a/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/GcovJsonParser.java
+++ b/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/GcovJsonParser.java
@@ -1,0 +1,120 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.coverageoutputgenerator;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.zip.GZIPInputStream;
+import com.google.gson.Gson;
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * A {@link Parser} for gcov intermediate json format introduced in GCC 9.1. See the flag {@code --intermediate-format} in <a
+ * href="https://gcc.gnu.org/onlinedocs/gcc-9.3.0/gcc/Invoking-Gcov.html">gcov documentation</a>.
+ */
+public class GcovJsonParser {
+  private static final Logger logger = Logger.getLogger(GcovJsonParser.class.getName());
+  private final InputStream inputStream;
+
+  private GcovJsonParser(InputStream inputStream) {
+    this.inputStream = inputStream;
+  }
+
+  public static List<SourceFileCoverage> parse(InputStream inputStream) throws IOException {
+    return new GcovJsonParser(inputStream).parse();
+  }
+
+  private List<SourceFileCoverage> parse() throws IOException {
+    ArrayList<SourceFileCoverage> allSourceFiles = new ArrayList<>();
+    try (InputStream gzipStream = new GZIPInputStream(inputStream)) {
+      ByteArrayOutputStream contents = new ByteArrayOutputStream();
+      byte[] buffer = new byte[1024];
+      int length;
+      while ((length = gzipStream.read(buffer)) != -1) {
+        contents.write(buffer, 0, length);
+      }
+      Gson gson = new Gson();
+      GcovJsonFormat document = gson.fromJson(contents.toString(), GcovJsonFormat.class);
+      if (!document.format_version.equals("1")) {
+        logger.log(Level.WARNING, "Expect GCov JSON format version 1, got format version " + document.format_version);
+      }
+      for (GcovJsonFile file : document.files) {
+        SourceFileCoverage currentFileCoverage = new SourceFileCoverage(file.file);
+        for (GcovJsonFunction function : file.functions) {
+          currentFileCoverage.addLineNumber(function.name, function.start_line);
+          currentFileCoverage.addFunctionExecution(function.name, function.execution_count);
+        }
+        for (GcovJsonLine line : file.lines) {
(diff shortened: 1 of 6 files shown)


tools/test/collect_cc_coverage.sh
```diff
@@ -125,10 +125,20 @@ function gcov_coverage() {
           # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84879).
           "${GCOV}" -i $COVERAGE_GCOV_OPTIONS -o "$(dirname ${gcda})" "${gcda}"
 
-          # Append all .gcov files in the current directory to the output file.
-          cat *.gcov >> "$output_file"
-          # Delete the .gcov files.
-          rm *.gcov
+          gcov_version=$("${GCOV}" --version | sed -n -r -e 's/^.*\s([0-9]\.[0-9]\.[0-9])\s?.*$/\1/p')
+
+          # Check the gcov version so we can process the data correctly
```
ulfjack, line 130: I think it would be good to explain here how this check works. It took me a minute to figure out.

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/Main.java
```diff
@@ -78,8 +79,13 @@ static int runWithArgs(String... args) throws ExecutionException, InterruptedExc
                 getTracefiles(flags, filesInCoverageDir),
                 LcovParser::parse,
                 flags.parseParallelism()),
-            parseFiles(
-                getGcovInfoFiles(filesInCoverageDir), GcovParser::parse, flags.parseParallelism()));
+	    Coverage.merge(
```
ulfjack, line 82: I'd only use a single call to `Coverage.merge` here and pass all parsed files to that call. If `merge` only takes two arguments, then change that to take more than two. I think it's strictly better to pass everything in at once, so `Coverage.merge` has the full list - there may be optimizations it can do in that case.

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/GcovJsonParser.java
```diff
@@ -0,0 +1,120 @@
...
+      }
+    }
+
+    return allSourceFiles;
+  }
+}
+
+// Classes for the Gson data mapper representing the structure of the GCov JSON format
+
+class GcovJsonFormat {
```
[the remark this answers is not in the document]
ulfjack: Also, at least provide a *link* to the documentation for the JSON file format.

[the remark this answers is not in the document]
helaan: Would it be acceptable to make them inner classes? I think that separating these 5 small classes that do not have associated methods and can't really be reused in other parts of the codebase into their own file would be close to littering.

Furthermore I'm not following the styleguide w.r.t. field naming as well because the elements in the JSON file use this naming, is that an issue as well?

The GCov manpage contains a thorough explanation of the fomat, however I'm having issues finding a versioned permalink other than their git source: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/doc/gcov.texi;h=dcdd7831ff063483d43e5347af0b67083c85ecc4;hb=4212a6a3e44f870412d9025eeb323fd4f50a61da#l184 which is only in texinfo format

tools/test/collect_cc_coverage.sh
```diff
@@ -125,10 +125,20 @@ function gcov_coverage() {
           # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84879).
           "${GCOV}" -i $COVERAGE_GCOV_OPTIONS -o "$(dirname ${gcda})" "${gcda}"
 
-          # Append all .gcov files in the current directory to the output file.
-          cat *.gcov >> "$output_file"
-          # Delete the .gcov files.
-          rm *.gcov
+          gcov_version=$("${GCOV}" --version | sed -n -r -e 's/^.*\s([0-9]\.[0-9]\.[0-9])\s?.*$/\1/p')
```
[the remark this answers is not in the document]
helaan: I don't think there is a way to deliver this fix without making changes in both parts. Using a combination (updated script but old collector or vice versa) shouldn't be hurtful, you'll just get missing coverage as you did before the patch.

Is it possible to release an updated Coverage Collector when the bazel update containing this fix is released?

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/GcovJsonParser.java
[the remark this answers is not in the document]
ulfjack: Inner classes are fine. I don't know how strict the style check is on the names. Please add a comment saying that these have to match the upstream definitions, so they don't get 'auto-corrected' on import.

tools/test/collect_cc_coverage.sh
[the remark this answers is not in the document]
ulfjack: The coverage collector release is separate from the bazel release, but someone at Google will have to make the push.

ulfjack:
@meisterT

tools/test/collect_cc_coverage.sh
```diff
@@ -125,10 +125,24 @@ function gcov_coverage() {
...
-          rm *.gcov
+          # Extract Gcov's version: the output of `gcov --version` contains the
+          # version as a set of major-minor-patch numbers, of which we extract
+          # the major version.
+          gcov_major_version=$("${GCOV}" --version | sed -n -E -e 's/^.*\s([0-9]+)\.[0-9]+\.[0-9]+\s?.*$/\1/p')
+
+          # Check the gcov version so we can process the data correctly, GCC 9
+          # contains the required changes
+          if [[ $gcov_major_version -ge 9 ]]; then
+              # GCOV 9.1.0 or higher generate a JSON based coverage format
```
meisterT, line 136: there is no GCOV 9.0.0?

helaan: GCC's versioning is a bit weird: 9.0.0 are development releases and the first actual release is 9.1.0. So there are only GCC 9.0.0 development snapshots, which no one (still) uses and even if they are in use, they'll probably already contain this JSON code. 

I probably should have updated the comment to say GCOV 9 instead

Source: https://gcc.gnu.org/develop.html#num_scheme

helaan: I've clarified the comment a few days ago, is that better?

meisterT: Sorry for the delay. It's stuck in internal review for a while now. Let me see if I can do something about it.

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/GcovJsonParser.java
```diff
@@ -0,0 +1,123 @@
...
+    int blocks_executed;
+    int execution_count;
+    String demangled_name;
+    int start_column;
+    int end_line;
+  }
+
+  class GcovJsonLine {
+    GcovJsonBranch[] branches;
+    int count;
```
xingxinghuo1000, line 111: The type of count could be long?   
see : https://github.com/bazelbuild/bazel/issues/9406#issuecomment-664211209

PR #11541, opened 2020-06-03
URL encode "%" due to rules_jvm_external paths

justhecuke-zz:
Fixes a problem where the manifest classpath is incorrectly written due to not URL-encoding symbols.

The paths provided in the manifest when the command line is too long is supposed to be a URL, which means we have to URL-encode special characters. `rules_jvm_external` generates a bunch of artifacts with paths that are themselves a URL-encoded version of their maven server host so we have a bunch of "%40" strings in the path due to usage of "@" for user@host http auth.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
@@ -312,8 +312,7 @@ function create_and_run_classpath_jar() {
       path="file:${path}" # e.g. "file:/usr/local/foo.jar"
     fi
 
-    path=$(sed "s/%/%25/g" <<< "${path}")
-    path=$(sed "s/ /%20/g" <<< "${path}")
+    path=$(sed "s/ /%20/g; s/%/%25/g" <<< "${path}")
     MANIFEST_CLASSPATH+=("${path}")
   done
   IFS="$OLDIFS"
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
@@ -312,6 +312,7 @@ function create_and_run_classpath_jar() {
       path="file:${path}" # e.g. "file:/usr/local/foo.jar"
     fi
 
+    path=$(sed "s/%/%25/g" <<< "${path}")
     path=$(sed "s/ /%20/g" <<< "${path}")
     MANIFEST_CLASSPATH+=("${path}")
   done
--- a/src/tools/launcher/java_launcher.cc
+++ b/src/tools/launcher/java_launcher.cc
@@ -152,6 +152,9 @@ static void WriteJarClasspath(const wstring& jar_path,
       if (x == L' ') {
         *manifest_classpath << L"%20";
       }
+      if (x == L'%') {
+        *manifest_classpath << L"%25";
+      }
       if (x == L'\\') {
         *manifest_classpath << L"/";
       } else {


jin:
I think this is the incorrect place to do it. Changing the stub template for a single user like rules_jvm_external usually means that we should fix the encoding at the rules_jvm_external side, and I think that's where it should encode it (or remove the host from the generated path)

justhecuke:
Honestly, I think the real answer is to put an actual robust URL encode
logic into the stub since the manifest classpath will always have a chance
of being malformed without that.

src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
```diff
@@ -312,6 +312,7 @@ function create_and_run_classpath_jar() {
       path="file:${path}" # e.g. "file:/usr/local/foo.jar"
     fi
 
+    path=$(sed "s/%/%25/g" <<< "${path}")
     path=$(sed "s/ /%20/g" <<< "${path}")
```
jin, line 316: This could be combined into a single sed invocation: `sed "s/ /%20/g; s/%/%25/g"`

jin:
Hmm, I just realized that this could cause problems with Windows environment variables in the form of `%VAR%`. Should the sed call be in the linux/mac call instead?

cc @meteorcloudy

meteorcloudy:
We don't use this bash script for launching Java binary on Windows anymore.
You probably want to also change
https://github.com/bazelbuild/bazel/blob/7c4f3ff7391862881056005db02c4be0d2bc147a/src/tools/launcher/java_launcher.cc#L153 ?

justhecuke:
Sorry for the delay, I just noticed the suggestion and added it in.

meteorcloudy:
Thanks, looks good to me!

justhecuke:
@jin any other thoughts/concerns on this?

jin:
From internal review: @justhecuke could you add some tests, please?

(cc @lberki)

src/tools/launcher/java_launcher.cc
```diff
@@ -152,6 +152,9 @@ static void WriteJarClasspath(const wstring& jar_path,
       if (x == L' ') {
         *manifest_classpath << L"%20";
       }
+      if (x == L'%') {
```
comius, line 155: 1. This code seem to skip jar_paths that have '%' in them and no spaces or back-slash. (because of first if) Is this intended?
2. If '%' occurs in the jar_path, then '%25%' seems to be the result in manifest_classpath. (because there is no else between ifs). Is this intended behaviour?

src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
```diff
@@ -312,7 +312,7 @@ function create_and_run_classpath_jar() {
       path="file:${path}" # e.g. "file:/usr/local/foo.jar"
     fi
 
-    path=$(sed "s/ /%20/g" <<< "${path}")
+    path=$(sed "s/ /%20/g; s/%/%25/g" <<< "${path}")
```
comius, line 315: I'm not expert in sed. Do you care to explain how '%' and ' ' are suppose to be handled?
For example from space and percentage ' %', is the desired result '%2520%25' or '%20%25'. On first sight it seems that you'll get the first one. I've seen cases in the past where this is actually desired, but they were rare.

Is is possible by any chance to use urlencode and urldecode functions?

src/tools/launcher/java_launcher.cc
justhecuke: You're right, the ordering of this is wrong. I'll need to change this here and elsewhere.

src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
justhecuke: As stated in the other comment, this is a bug and I'd prefer the second example rather than the first.

I'd love to use `urlencode` but it isn't available on `macos` which presents a problem. If you know of an alternative tool I'd be happy to use that instead since rolling our own url encoder, besides being a waste of time, would be error-prone.

src/tools/launcher/java_launcher.cc
lberki: @justhecuke-zz , do I understand correctly that this PR is still waiting on the resolution of this item?

aiuto:
I'm trying to clean up old PRs. This has several problems.
1. There is no underlying bug report first. That is required to show a reproducible case, and discuss the merits of each solution.
2. There are no tests added for this.

@justhecuke-zz  Are you going address those issues, or should we close this as obsolete?

Bencodes:
@aiuto the problem that this PR fixes was addressed in another PR that's already merged.

aiuto:
Sounds good. I'm just doing general triage, so I'm not going to hunt down the refs. Unless I hear from anyone else I'll just close this as stale next week.

PR #11542, opened 2020-06-03
Fix minor code issues flagged by error prone

davido:


aiuto:
I'll try to merge this. If anyone wants to look for subtle side effects, LMK soon.

alandonovan:
For the record: this PR would have introduced a number of behavior changes and even bugs (mostly due to issues of encoding). The linter is not to be blindly trusted, and many of its suggestions are actively harmful to readability.

If anyone is tempted to send PRs to apply lint cleanups in future, please address en masse only the findings that have zero false positives. If the findings require scrutiny due to potential behavior changes (such as those related to string encoding), please send a single PR that fixes only those issues, with (as always, for behavior changes) justification.

aiuto:
Sorry for the delay.  I lost track of it.

I reverted any changes that added UTF-8 converters.  It is not so simple with Bazel because strings are note really Unicode. They are a funny representation. If you do a UTF-8 encoding you risk getting double encoded characters.

I'm running the presubmit right now.

davido:
@aiuto Thanks for the review and thanks for clarifying.

aiuto:
The other bit I ended up reverting is anything which potentially increased the length of an error message (like including the originating exception).  This can have a bad impact on logging.

PR #11562, opened 2020-06-08
WIP: Ignore absolute path inputs from dotd header parsing.

jin:


meisterT:
What's the state of this PR?

PR #11565, opened 2020-06-08
Restore macOS 10.10 version compatibility to Bazel

nikhilm:
This is part of the fixes for #11442. (Not a complete fix because it does not address the bazel 2 branch).

It makes the macOS deployment target 10.10. Then it wraps the calls to the macOS logging infrastructure so they only execute on macOS 10.12+.

I've tested this change and a similar cherry-pick on tag 2.2.0 and confirmed this results in a functioning Bazel (at least for the targets we build) on macOS 10.10.

I have questions around what I can do better:
1. I have just added a universal .bazelrc entry. Is there some way to gate this to just macOS in the Bazel infrastructure? Is that even necessary?
2. The Bazel tagged versions are just tags and not branches, which means I cannot submit a PR for a backport to Bazel 2.2. How would you like to handle this?

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -39,5 +39,7 @@ build:ubuntu1604_java8 --config=remote_shared
 # Alias
 build:remote --config=ubuntu1604_java8
 
+build --macos_minimum_os=10.10
+
 # User-specific .bazelrc
 try-import user.bazelrc
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,5 @@
+common --enable_platform_specific_config
+
 # Shared configuration flags to build and test Bazel on RBE.
 build:remote_shared --define=EXECUTOR=remote
 build:remote_shared --remote_instance_name=projects/bazel-untrusted/instances/default_instance
@@ -39,7 +41,7 @@ build:ubuntu1604_java8 --config=remote_shared
 # Alias
 build:remote --config=ubuntu1604_java8
 
-build --macos_minimum_os=10.10
+build:macos --macos_minimum_os=10.10
 
 # User-specific .bazelrc
 try-import user.bazelrc
--- a/src/main/native/unix_jni_darwin.cc
+++ b/src/main/native/unix_jni_darwin.cc
@@ -144,15 +144,32 @@ static dispatch_queue_t JniDispatchQueue() {
 // Log used for all of our anomaly logging.
 // Logging can be traced using:
 // `log stream -level debug --predicate '(subsystem == "build.bazel")'`
+//
+// This may return NULL if `os_log_create` is not supported on this version of macOS.
+// Use `log_if_possible` to log when supported.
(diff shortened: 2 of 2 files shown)


nikhilm:
@googlebot I signed it! (as `nikhilm@dropbox.com`)

nikhilm:
@googlebot I signed it!

nikhilm:
perhaps @tetromino, @aiuto  or @philwo could take a look at this?

I don't know why cla/google is still failing. possibly needs manual intervention.

philwo:
Thank you and sorry for the delay in reviewing it! It looks good to me.

The CLA bot told me that it's confused because the PR was created by your gmail.com address (although the commits show your Dropbox address) and because it doesn't know that they belong to the same person, it doesn't like it. Just to verify - have you added your dropbox.com e-mail to your GitHub account as a secondary e-mail address? I think it should work then.

nikhilm:
ah! that is what was missing. let's try again.

@googlebot I signed it!

nikhilm:
@philwo what about 

> 
>     I have just added a universal .bazelrc entry. Is there some way to gate this to just macOS in the Bazel infrastructure? Is that even necessary?
>     The Bazel tagged versions are just tags and not branches, which means I cannot submit a PR for a backport to Bazel 2.2. How would you like to handle this?
>

philwo:
@nikhilm Nice, thanks! :)

Regarding the bazelrc change - if CI passed, it apparently doesn't break the build on Linux and Windows. 😄

OTOH maybe this is a good time to add `build --enable_platform_specific_config` to our `.blazerc` and then change your line to `build:macos --macos_minimum_os=10.10` - this would ensure that the flag only gets set when you're building on macOS. Here's the [documentation](https://docs.bazel.build/versions/master/command-line-reference.html#flag--enable_platform_specific_config) for that flag.

philwo:
Sorry for the delay. I'm importing this now.

philwo:
I got this compiler error when trying to submit the PR internally:

```
third_party/bazel/src/main/native/unix_jni_darwin.cc:156:7: error: comparison of function 'os_log_create' not equal to a null pointer is always true [-Werror,-Wtautological-pointer-compare]
  if (os_log_create != NULL) {
      ^~~~~~~~~~~~~    ~~~~
third_party/bazel/src/main/native/unix_jni_darwin.cc:156:7: note: prefix with the address-of operator to silence this warning
  if (os_log_create != NULL) {
      ^
      &
1 error generated.
```

Any idea how to fix this? Is it just a matter of prefixing `os_log_create` with `&`?

nikhilm:
Ah! good catch. I have updated the PR to have `&`.

philwo:
Perfect, thanks! All presubmit tests pass - it's now out for a quick review by a colleague and should then be auto-pushed to GitHub.

PR #11566, opened 2020-06-08
When re-using a ToolchainContextKey, make sure to remove the actual

katre:
toolchains to prevent dependency loops.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
@@ -538,6 +538,14 @@ public final class ConfiguredTargetFunction implements SkyFunction {
           (UnloadedToolchainContext) values.get(unloadedToolchainContextKey.getValue()).get();
       if (!valuesMissing) {
         String execGroup = unloadedToolchainContextKey.getKey();
+        if (parentToolchainContextKey != null) {
+          // Since we inherited the toolchain context from the parent of the dependency, the current
+          // target may also be in the resolved toolchains list. We need to clear it out.
+          // TODO(configurability): When updating this for config_setting, only remove the current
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -538,6 +538,14 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
           (UnloadedToolchainContext) values.get(unloadedToolchainContextKey.getValue()).get();
       if (!valuesMissing) {
         String execGroup = unloadedToolchainContextKey.getKey();
+        if (parentToolchainContextKey != null) {
+          // Since we inherited the toolchain context from the parent of the dependency, the current
+          // target may also be in the resolved toolchains list. We need to clear it out.
```
juliexxia, line 543: I don't think I have all the context on this - but does the issue arise if the current target's parent has the current target in the toolchain list? Won't that always happen for the dependency link between target (parent) and it's toolchain (current target)? Sorry if this phrasing is confusing!

katre: Right, that's exactly the problem this is dealing with. When toolchain resolution is enabled for cc rules, a `cc_library` target will have a dependency on the correct `cc_toolchain`. Currently, that is a host transition, which is incorrect in many cases.

The goal of toolchain transitions is to change the host transition into a target transition, but to include extra information for the toolchain to know what execution platform to use. Unfortunately, this naively adds a dependency from the `cc_toolchain` back to itself.

The bug fix here is to clear out all of the resolved toolchains in the child (`cc_toolchain`) configured target. Toolchains themselves shouldn't depend on further toolchains, so this should be fine. Another version would be to just remove the loop, and leave other toolchain dependencies in place, but this leads to extra unused dependencies.

PR #11567, opened 2020-06-08
Fix typo in "Build without Bytes" error message

utsav-dbx:
`--experimental_remote_outputs` flag does not exist. The right flag is ` --remote_download_outputs`


utsav-dbx:
@googlebot I signed it!

ulfjack:
The correct name is actually `remote_download_outputs` (although the experimental name also still works). Can you update the patch?

ulfjack:
@meisterT

PR #11576, opened 2020-06-10
Add -D flag to libtool for the Apple crosstool

michaeleisel:
libtool's man page states, `-D  When building a static library, set archive contents' user ids, group ids, dates, and file modes to reasonable defaults. This allows libraries created with identical input to be identical to each other, regardless of time of day, user, group, umask, and other aspects of the environment.`. So, this flag makes things deterministic for libtool. In fact, it seems to be on by default for xcodebuild as of Xcode 11.4.1. Without it, I find libtool unable to be reproducible. Also of note is `ZERO_AR_DATE`, but it seems to cover a subset of what `-D` does.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -1803,6 +1803,7 @@ def _impl(ctx):
                         flag_group(
                             flags = [
                                 "-no_warning_for_no_symbols",
+                                "-D",
                                 "-static",
                                 "-filelist",
                                 "%{obj_list_path}",
@@ -1834,6 +1835,7 @@ def _impl(ctx):
                         flag_group(
                             flags = [
                                 "-no_warning_for_no_symbols",
+                                "-D",
                                 "-static",
                                 "-filelist",
                                 "%{obj_list_path}",
@@ -1866,6 +1868,7 @@ def _impl(ctx):
                         flag_group(
                             flags = [
                                 "-no_warning_for_no_symbols",
+                                "-D",
                                 "-static",
                                 "-filelist",
                                 "%{obj_list_path}",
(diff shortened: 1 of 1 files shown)


michaeleisel:
would there be a good api change to opt-in to this, e.g. a flag that can be set? when are we planning to drop xcode 10 support?

keith:
I'm not sure how these have been handled in the past. Theoretically you could add a new feature and folks could opt in with `--features=foo` or maybe that could even be done in javaland by checking against the current Xcode version. Someone from the bazel team will have to chime in. Also I think the goal is to support the last 3 major versions of Xcode but I might be mis-remembering

thomasvl:
@trybka do you know who could advise?

trybka:
You can derive the value for the current xcode version in starlark at crosstool creation.

Something like this:

This would be added as an attribute on the cc_toolchain_config_rule
```    
       "_xcode_config": attr.label(default = configuration_field(
            fragment = "apple",
            name = "xcode_config_label",
        )),
```

Then when adding flags or what-have-you, you can check it like this:
```
def compare_versions(dv1, v2):
    """Return value is <0, 0, >0 depending on DottedVersion dv1 comparison to string v2."""
    return dv1.compare_to(apple_common.dotted_version(v2))

supports_deterministic_mode = "11.0"  # or whatever

def can_use_deterministic_libtool(ctx):
    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
    xcode_version = xcode_config[apple_common.XcodeVersionConfig].xcode_version()
    if compare_versions(xcode_version, supports_deterministic_mode) >= 0:
            return True
    else:
            return False
```

Something like that. I just typed this into the comment box, so some syntax may be slightly off, or I might have the < or > inverted.

Does that help?

michaeleisel:
closing, as it looks like https://github.com/bazelbuild/bazel/pull/11840 is on it

Issue #11579, opened 2020-06-11
HttpConnectorMultiplexer catches Throwable instead of Exception


ehkloaj:
### Description of the problem:

In the class [HttpConnectorMultiplexer](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpConnectorMultiplexer.java) on line 288 catch Throwable is used instead of catch Exception.

```
catch (Throwable e) {
      // Oh no the connector failed for some reason. We won't let that interfere with our plans.
      synchronized (context) {
            context.errors.add(e);
            continue;
      } 
}
```

This means that Throwables that are also of the type Error are being caught which can lead to unexpected behavior and it is not recommended to catch Errors according to the [Java documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Error.html). 

Is there a reason for using Throwable instead of exception?
If so I think this should be stated in a comment near the usage, if not I think it should be changed.

Changing this would prevent Throwables to be added to the list errors in MutexConditionSharedMemory but it might be a good idea to change all occurences of Throwable to Exception in the file. 

```
private static class MutexConditionSharedMemory {
  @GuardedBy("this") @Nullable HttpStream result;
  @GuardedBy("this") final List<Thread> threads = new ArrayList<>();
  @GuardedBy("this") final Deque<WorkItem> jobs = new LinkedList<>();
  @GuardedBy("this") final List<Throwable> errors = new ArrayList<>();
}
```

ehkloaj:
This was solved in #11812

PR #11580, opened 2020-06-11
Catch Exception instead of Throwable

ehkloaj:
This fixes the problem in #11579 but it might be a good idea to also change all occurences of Throwable to Exception in the class as discussed in the issue.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpConnectorMultiplexer.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpConnectorMultiplexer.java
@@ -285,7 +285,7 @@ final class HttpConnectorMultiplexer {
             tellParentThreadWeAreDone();
             return;
           }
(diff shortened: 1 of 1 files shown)


ehkloaj:
@googlebot I signed it!

ehkloaj:
@googlebot I signed it!

ehkloaj:
@googlebot I signed it!

laurentlb:
The commit is associated with the address `ehkloaj@seroiuts01372.sero.gic.ericsson.se`.
Can you make sure you signed the CLA for this exact address? Or you can try to update the address in the commit.

ehkloaj:
Thanks for looking at this @laurentlb. I also found that this was the problem, the address I signed the CLA with is `kalle.johansson@ericsson.com` and I've changed this in my later pull requests. The problem is that while troubleshooting I deleted the fork that this commit was made from and I don't think there is a way to recover from that.
Do you want me to close this PR and create a new commit with the right email?

PR #11591, opened 2020-06-15
RELNOTES[NEW]: bazel info: Allow to specify multiple keys.

meisterT:
If you specify more than one key, then the value will prefixed with the
key.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/InfoCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/InfoCommand.java
@@ -18,6 +18,7 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.devtools.build.lib.analysis.NoBuildEvent;
 import com.google.devtools.build.lib.analysis.config.BuildConfiguration;
 import com.google.devtools.build.lib.analysis.config.InvalidConfigurationException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/runtime/commands/InfoCommand.java
```diff
@@ -213,37 +213,27 @@ public BlazeCommandResult exec(
...
+          if (items.containsKey(key)) {
+            try (SilentCloseable c = Profiler.instance().profile(key + ".infoItem")) {
+              value = items.get(key).get(configurationSupplier, env);
+              if (residue.size() > 1) {
+                outErr.getOutputStream().write((key + ": ").getBytes(StandardCharsets.UTF_8));
+              }
+              outErr.getOutputStream().write(value);
+            }
+          } else {
+            String message = "unknown key: '" + key + "'";
```
ulfjack, line 230: Is it intentional that you error out on the first unknown key? This may not be easy to change when people have started relying on it. Maybe first check all key names for validity before printing?

meisterT: Done in a slightly different way: I collect all unknown keys and report them in the end.

src/main/java/com/google/devtools/build/lib/runtime/commands/InfoCommand.java
```diff
@@ -216,6 +218,7 @@ public BlazeCommandResult exec(
 
       env.getEventBus().post(new NoBuildEvent());
       if (residue.size() > 0) {
+        ImmutableSet.Builder<String> unknownKeysBuilder = ImmutableSet.builder();
```
ulfjack, line 221: This will crash if I specify the same unknown key twice in a row.

meisterT: Nope, it doesn't, see the new test below. Note that this is an ImmutableSet not an ImmutableMap.

meisterT:
> Needs tests?

Done.

ulfjack:
Thanks! I thought ImmutableSet throws just like ImmutableMap, but hey, why be consistent?

Issue #11585, opened 2020-06-11
Bazel not using latest version of CoverageOutputGenerator


paymog:
### Description of the problem / feature request:

It seems like bazel is not using the latest version of CoverageOutputGenerator. When I run `bazel coverage --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //...` I see the following lines of output:

```
Jun 11, 2020 2:14:20 PM com.google.devtools.coverageoutputgenerator.Main parseFiles
SEVERE: Parsing file bazel-out/darwin-fastbuild/testlogs/examples/bundles/nginx/go_default_test/coverage.dat
```

Note that this is a `SEVERE` log but that [this commit](https://github.com/bazelbuild/bazel/commit/21f633ca58d45b4d270668ddb147f680931efef3#diff-28066f583dd7408ef66042f7403845ea) changes that to an `INFO` log.

The commit exists on versions of bazel back to 2.2 and I'm on version 3.1. I've tried running `bazel clean` but that didn't help.

### Feature requests: what underlying problem are you trying to solve with this feature?

Not a feature request

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I don't have a minimal repro.

### What operating system are you running Bazel on?

osx 10.15.5

### What's the output of `bazel info release`?

`release 3.1.0`

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

NA

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

I don't think this is relevant because I'm not building bazel

###  Have you found anything relevant by searching the web?

Yes, I've found the following:
* https://stackoverflow.com/questions/60751912/output-coverage-coverage-report-dat-was-not-created-when-using-bazel-coverag
* https://github.com/bazelbuild/bazel/issues/6450

### Any other information, logs, or outputs that you want to share?

I think because an old version of the coverage generator is being used, coverage fails to be combined:

```
ERROR: output '_coverage/_coverage_report.dat' was not created
ERROR: not all outputs were created or valid
```

ulfjack:
Someone will need to create a new release (although note that will only affect future releases of Bazel).

paymog:
A new release of the CovergeOutputGenerator?

ulfjack:
[...]

PR #11597, opened 2020-06-16
Update coverage output generator to include 3b75bc7d7aee7bcdbb51a1522…

meisterT:
…df8efb3121bb4e5.

RELNOTES[NEW]: Support code coverage with GCC 9.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -149,7 +149,7 @@ distdir_tar(
         "java_tools_javac11_linux-v8.0.zip",
         "java_tools_javac11_windows-v8.0.zip",
         "java_tools_javac11_darwin-v8.0.zip",
-        "coverage_output_generator-v2.1.zip",
+        "coverage_output_generator-v2.3.zip",
         # bazelbuid/stardoc
         "1ef781ced3b1443dca3ed05dec1989eca1a4e1cd.tar.gz",
         # rules_sass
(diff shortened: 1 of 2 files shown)


meisterT:
cc @helaan

helaan:
In case you were CC'ing me because the build is failing: I see it is trying to download a file from `https://mirror.bazel.build/bazel_coverage_output_generator/releases/coverage_output_generator-v2.2.zip`, which it can't find. Is the new release uploaded properly or does it just need some time to propagate?

meisterT:
No, I was just cc'ing you to keep you in the loop.

The file is now in the right place and I triggered a rebuild. So let's hope that it'll come back clean.

meisterT:
Well, no luck, some coverage tests are broken.

I don't have time to look into them today, so you may want to do so. Easiest is to patching in my PR and then run `bazel test //src/test/shell/bazel:bazel_coverage_java_test`.

helaan:
Hmm, looked into it for a bit, I can reproduce the issue on my machine when I perform the steps myself, is it possible that a change different to the one we are integrating is responsible for the breakage? As spotted in #11585 the previous release of COG did not include commit 21f633ca, nor the 6 other commits that have been made since then.

meisterT:
955a4fd seems to be the culprit, but I haven't yet looked why

meisterT:
This is the fix: https://github.com/bazelbuild/bazel/pull/11605

PR #11611, opened 2020-06-18
Fix export of @bazel_tools//src/main/protobuf

### LINKED ISSUE #8738: Unable to reference @bazel_tools//src/main/protobuf targets

I am unable to use Bazel's protobuf targets and sources---specifically extra_action_base_java_proto---directly from `@bazel_tools//src/main/protobuf` because of an unsatisfied dependency on `@bazel_tools//tools/build_rules:genproto.bzl`.

As a workaround, in order to use extra actions, my project's WORKSPACE needs to refer to and clone Bazel's source via git_repository. Cloning the full repository is slow, and needing to rev the SHA1 each time we upgrade Bazel is a pain.

```console
$ cat BUILD
java_binary(
    name = "dummy",
    srcs = ["src/main/java/Dummy.java"],
    main_class = "Dummy",
    deps = [
        "@bazel_tools//src/main/protobuf:extra_actions_base_proto",
    ],
)
$ bazel build :dummy
...
ERROR: .../dummy/BUILD:1:1: error loading package '@bazel_tools//src/main/protobuf': Unable to load file '@bazel_tools//tools/build_rules:genproto.bzl': file doesn't exist and referenced by '//dummy:dummy'
ERROR: Analysis of target '//dummy:dummy' failed; build aborted: error loading package '@bazel_tools//src/main/protobuf': Unable to load file '@bazel_tools//tools/build_rules:genproto.bzl': file doesn't exist
```

- Linux (CentOS 6.x, 7.x)
- macOS High Sierra and Mojave (10.13 and 10.14)

release 0.26.0+vmware

- #1188

beasleyr-vmw:
`src/main/protobuf/BUILD` referenced other parts of the Bazel source tree
which aren't available when repackaged into the embedded tools distribution.
This commit adds a parallel `BUILD.tools` which is stripped of any such
references.

Resolves #8738.

Testing Done:
- Referred to `@bazel_tools//src/main/protobuf:worker_protocol.proto` in
  an internal rule w/o encountering #8738.

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -126,7 +126,7 @@ JAVA_TOOLS = [
                "//src/tools/launcher:srcs",
                "//src/main/cpp/util:embedded_tools",
                "//src/main/native:embedded_tools",
-               "//src/main/protobuf:srcs",
+               "//src/main/protobuf:embedded_tools",
                "//third_party/def_parser:srcs",
                "//third_party/zlib:embedded_tools",
            ] + select({
--- a/src/main/protobuf/BUILD
+++ b/src/main/protobuf/BUILD
@@ -276,6 +276,15 @@ filegroup(
 )
 
 filegroup(
+    name = "embedded_tools",
+    srcs = glob([
+        "BUILD.tools",
+        "*.proto",
+    ]),
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
     name = "dist_jars",
     srcs = [s + "_java_proto_srcs" for s in FILES] + [
         ":analysis_java_proto_srcs",
--- a/src/main/protobuf/BUILD.tools
+++ b/src/main/protobuf/BUILD.tools
@@ -1,0 +1,140 @@
+load("@rules_java//java:defs.bzl", "java_proto_library")
+load("@rules_cc//cc:defs.bzl", "cc_proto_library")
+load("@rules_proto//proto:defs.bzl", "proto_library")
+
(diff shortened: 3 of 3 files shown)


beasleyr-vmw:
Not sure what the scope of any test case(s) should be. My best guess would be something similar to `//src/test/shell/bazel:embedded_tools_deps_test`.

aiuto:
I assigned reviewers based on who had the TODO and who the blame layer points to.

philwo:
@meisterT Friendly ping :P

philwo:
@beasleyr-vmw Could you just copy that protocol buffer file to your repository instead? Protobufs are supposed to be backwards compatible, so even if you don't update your version for some time, it should be fine until you want to use newer fields. It's also easier for tools like Code Search (if you use that) if the protobuf is in the repo, instead of an embedded repo inside the Bazel binary.

I tried to import it, but I'm not sure if the overhead of maintaining yet another parallel BUILD file that might get out of date in our code is worth it.

beasleyr-vmw:
@philwo If that's the official position, I'm fine with it. I'm unfamiliar w/ protobuf best practices, but my gut feeling was to stick to canonical locations instead of privately snapshotting the .proto files. LMK and I can withdraw this.

meisterT:
@beasleyr-vmw thanks for your understanding - I agree with Philipp that the separate BUILD file is brittle (it's untested) and overhead.

PR #11613, opened 2020-06-18
Fix documentation for createLinkerInput owner

deeglaze:
The existing text is a clear copy/paste error. My proposed replacement is the most sensible explanation I can give for this field, given my lack of context for this component.

Suggested reviewer: plf@

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skylarkbuildapi/cpp/CcModuleApi.java
+++ b/src/main/java/com/google/devtools/build/lib/skylarkbuildapi/cpp/CcModuleApi.java
@@ -633,7 +633,7 @@ public interface CcModuleApi<
       parameters = {
         @Param(
             name = "owner",
-            doc = "List of <code>LibraryToLink</code>.",
+            doc = "The label of the target that produced all files used in this input.",
             positional = false,
             named = true,
             type = Label.class),


src/main/java/com/google/devtools/build/lib/skylarkbuildapi/cpp/CcModuleApi.java
```diff
@@ -633,7 +633,7 @@ LibraryToLinkT createLibraryLinkerInput(
       parameters = {
         @Param(
             name = "owner",
-            doc = "List of <code>LibraryToLink</code>.",
+            doc = "The label of the target that produced any files used in this input.",
```
ulfjack, line 636: I think it should probably be "all" rather than "any"

ulfjack:
@oquenchil

ulfjack:
@aiuto It's odd that this hasn't been looked at for almost a week.

src/main/java/com/google/devtools/build/lib/skylarkbuildapi/cpp/CcModuleApi.java
deeglaze: Done.

PR #11617, opened 2020-06-19
Add flag --incompatible_blacklisted_protos_requires_proto_info

Yannic:
This change adds a new incompatible flag that makes ProtoInfo mandatory
for skipping code-gen for some proto sources (e.g. WKPs that are already
provided by the proto runtime). Flipping this flag will allow us to get rid of
ProtoInfo#getOriginal{Direct,Transitive}ProtoSources().

https://docs.google.com/document/d/1UmF_P4NDVl6Nw4JsHoHYp74iCdEQNh1O3-SVAX1CdjU/edit#

RELNOTES: Adds --incompatible_blacklisted_protos_requires_proto_info to indicate whether proto_lang_toolchain.blacklisted_protos requires ProtoInfo.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
@@ -42,8 +42,10 @@ public class ProtoLangToolchain implements RuleConfiguredTargetFactory {
     for (TransitiveInfoCollection protos :
         ruleContext.getPrerequisites("blacklisted_protos", TARGET)) {
       ProtoInfo protoInfo = protos.get(ProtoInfo.PROVIDER);
-      if (ruleContext.getFragment(ProtoConfiguration.class).blacklistedProtosRequiresProtoInfo()
-          && protoInfo == null) {
+      if (protoInfo == null
+          && ruleContext
+              .getFragment(ProtoConfiguration.class)
+              .blacklistedProtosRequiresProtoInfo()) {
         ruleContext.ruleError(
             "'"
                 + ruleContext.getLabel().toString()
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
```diff
@@ -42,7 +42,13 @@ public ConfiguredTarget create(RuleContext ruleContext)
     for (TransitiveInfoCollection protos :
         ruleContext.getPrerequisites("blacklisted_protos", TARGET)) {
       ProtoInfo protoInfo = protos.get(ProtoInfo.PROVIDER);
-      // TODO(cushon): it would be nice to make this mandatory and stop adding files to build too
+      if (ruleContext.getFragment(ProtoConfiguration.class).blacklistedProtosRequiresProtoInfo()
```
comius, line 45: Can you swap the conjuncts, i.e (protoInfo == null && ...). This way in case protoInfo is not null, configuration fragment is not required.

Probably it would also be best adding requiredConfigurationFragment to the RuleConfiguredTargetBuilder below.

Yannic: Done. That said, configuration fragment is always read in `ProtoCommon.checkRuleHasValidMigrationTag(ruleContext);` above.

I may be missing something, but aren't configuration fragments automatically added to `RuleConfiguredTargetBuilder`? https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/RuleConfiguredTargetBuilder.java;l=118;drc=8b856f5484f0117b2aebc302f849c2a15f273310

Yannic:
@comius I addressed your comments. I'm hoping to get this into Bazel 3.4, can you take another look?

PR #11623, opened 2020-06-22
Add linkopts to the Starlark implementation of cc_import

agluszak:
This change allows users to pass custom linking options when using Starlark version of cc_import rule (currently hidden behind '--experimental_starlark_cc_import' flag). 

Co-authored-by: @oquenchil


agluszak:
@googlebot I fixed it.

PR #11627, opened 2020-06-23
Added Makefile

TheMindVirus:
This Makefile makes it simpler to bootstrap Bazel from ARM Linux.
It follows the instructions already available from the building from source page of your website.
Users will download the distribution zip, unzip it inside a directory called "bazel" and enter "make" in the terminal from inside the "bazel" folder.
Users may also enter "make clean" to remove the generated output directory to start the build again from scratch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,77 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-# This script bootstraps building a Bazel binary without Bazel then
-# use this compiled Bazel to bootstrap Bazel itself. It can also
-# be provided with a previous version of Bazel to bootstrap Bazel itself.
-#
+# This script aids the bootstrapping of building a Bazel binary without Bazel
+# using GNU Makefile. Simply use "make" in the current directory to start.
 # The resulting binary can be found at output/bazel.
 
 all:
-	# Make sure dependencies for bootstrapping bazel are installed
 	sudo apt-get install build-essential openjdk-11-jdk python zip unzip
-	env EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk"
-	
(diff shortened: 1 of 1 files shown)


philwo:
Hi @TheMindVirus,

Thanks for sending your PR. I think in general the idea of having a `Makefile` for Bazel that can be used to bootstrap it is interesting. However, in this case, it seems like it would be more helpful to improve the existing `compile.sh` script to
- check that the required `apt-get` packages are installed and print a helpful warning if that's not the case and
- that the `--host_javabase=@local_jdk//:jdk` flag is passed on systems where Bazel doesn't already know how to fetch a good OpenJDK version.

Would you be interested in looking into this?

TheMindVirus:
I initially had reservations about adding the contents of compile.sh to Makefile considering differences in how they work.
I then tried a Makefile that runs modified steps from compile.sh but it confirmed my reservations and became more complex.
This aims to simplify the process by giving users one simple command that won't be likely to change: "make".
It then allows you to change what actually happens in the backend in compile.sh, buildenv.sh and bootstrap.sh.
To test it the Makefile has to be added to a download of the bazel-dist.zip release instead of the bazel dev source tree.

meisterT:
What's the state of this PR?

jin:
Perhaps we want to make the instructions for compile.sh clearer? Adding yet another entry point to compile Bazel (and dependency on "make" in whichever environment you're bootstrapping bazel in) seems relatively heavy to updating compile.sh itself.

aiuto:
+1 to @jin\'s comment.

aiuto:
I'm going to close this on the grounds that we don't want a Makefile, period. If we want better docs, then let's do a different PR for that.

PR #11630, opened 2020-06-23
Add Ubuntu 20.04 platform, but do not test it in CI yet

mostynb:
This seems like it might be a prerequisite for adding a ubuntu 20.04 platform to bazelci, based on this CI failure:
https://github.com/bazelbuild/continuous-integration/pull/988

### DIFF
--- a/.bazelci/build_bazel_binaries.yml
+++ b/.bazelci/build_bazel_binaries.yml
@@ -18,6 +18,12 @@ platforms:
     build_flags:
       - "-c"
       - "opt"
+  ubuntu2004:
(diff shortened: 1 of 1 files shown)


philwo:
I will need to add support for that platform to `bazelci.py` first, then we can merge this 👍

mostynb:
> I will need to add support for that platform to `bazelci.py` first, then we can merge this 👍

I tried that previously: https://github.com/bazelbuild/continuous-integration/pull/988

IIRC it looked like this PR was needed before the bazelci.py change.

PR #11642, opened 2020-06-25
Stop `--apple_generate_dsym` from applying to host targets

keith:
As part of d7d0e972c75a2e83fc5ae19927be38feda006ac9 these flags started
started applying to host targets. This causes unexpected rebuilds of
host tools that you will likely never consume the dSYM from.


keith:
Related convo https://github.com/bazelbuild/bazel/commit/d7d0e972c75a2e83fc5ae19927be38feda006ac9#r40171545

PR #11647, opened 2020-06-26
Add the 'includes' param for the Starlark version of cc_import

agluszak:
This change allows users to add a list of include dirs when using Starlark version of cc_import rule (currently hidden behind '--experimental_starlark_cc_import' flag).

### DIFF
--- a/tools/build_defs/cc/tests/cc_import_test.bzl
+++ b/tools/build_defs/cc/tests/cc_import_test.bzl
@@ -55,10 +55,7 @@ def _cc_import_linkopts_test_impl(ctx):
     found = False
     for action in actions:
         if action.mnemonic == "CppLink":
-            for arg in action.argv:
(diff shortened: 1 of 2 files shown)


tools/build_defs/cc/tests/cc_import_test.bzl
```diff
@@ -32,41 +59,39 @@ def _cc_import_linkopts_test_impl(ctx):
...
+    includes = target_under_test[CcInfo].compilation_context.includes.to_list()
 
-    cc_import_linkopts_test(
-        name = "cc_import_linkopts_test",
-        target_under_test = ":cc_import_linkopts_test_binary",
-        tags = TAGS,
-    )
+    found = False
+    for include in includes:
+        if include == "testinclude":
```
c-mita, line 76: Replace the loop with `found = "testinclude" in includes`?

PR #11652, opened 2020-06-27
Update build.proto docs for column numbers.

benjaminp:
Starlark locations now include columns, but this was not reflected in the comments for various location fields.


alandonovan:
Thanks!

Issue #11656, opened 2020-06-28
Android desugaring is not resilient to pre-optimized bytecode.


restingbull:
### Description of the problem / feature request:

Building an android application that contains library jars that have pre-optimized byte code throws illegal state exceptions.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Add the following maven dependencies to an android build:
-   'com.amplifyframework:core:1.0.0'
-   'com.amplifyframework:aws-storage-s3:1.0.0'
-   'com.amplifyframework:aws-auth-cognito:1.0.0'

```
Alternatively, run the following commands in a bazelbuild workspace with ANDROID_HOME defined (change version as necessary, of course):

wget https://maven.google.com/androidx/annotation/annotation/1.1.0/annotation-1.1.0.jar
wget https://maven.google.com/androidx/multidex/multidex/2.0.0/multidex-2.0.0.aar &&\
 unzip -d multidex-2.0.0.classes.jar multidex-2.0.0.aar classes.jar
wget https://repo1.maven.org/maven2/com/jakewharton/threetenabp/threetenabp/1.2.4/threetenabp-1.2.4.aar &&\
 unzip -d  threetenabp-1.2.4.classes.jar threetenabp-1.2.4.aar classes.jar
wget https://maven.google.com/androidx/legacy/legacy-support-v4/1.0.0/legacy-support-v4-1.0.0.aar &&\
 unzip -d legacy-support-v4-1.0.0.classes.jar legacy-support-v4-1.0.0.aar classes.jar
wget https://repo1.maven.org/maven2/com/amplifyframework/core/1.0.0/core-1.0.0.aar &&\
 unzip core-1.0.0.aar classes.jar

bazel run //src/tools/android/java/com/google/devtools/build/android/desugar:Desugar --\
 -i classes.jar \
 --output /tmp/not_gonna_happen.jar\
 --bootclasspath_entry $ANDROID_HOME/platforms/android-29/android.jar\
 --classpath_entry  threetenabp-1.2.4.classes.jar\
 --classpath_entry  legacy-support-v4-1.0.0.classes.jar\
 --classpath_entry annotation-1.1.0.jar\
 --classpath_entry multidex-2.0.0.classes.jar

```
### What operating system are you running Bazel on?
Catalina 10.15.5

### What's the output of `bazel info release`?

release 3.2.0. Persists in 3.3.x as well.

restingbull:
[...]

PR #11657, opened 2020-06-28
Modify LambdaDesugar to be more tolerant of operand stack optimizations.

restingbull:
New Behaviour:
* More resilient instruction tracing when attempting to avoid lambda factory methods. Previously, any instruction set that deviated from javac 8 by optimizing stack calls would violate preconditions. Now, tracing considers the simple case of DUP and POP, as well as handling all methods of pushing instructions onto the stack
* Added a proguard limited optimized binary to verify desugaring against, as well as a new integration test based on the DesugarFunctionalTest
* Precondition violations will now print out the offending instruction set.

Fixes: https://github.com/bazelbuild/bazel/issues/11656


src/tools/android/java/com/google/devtools/build/android/desugar/LambdaDesugaring.java
```diff
@@ -44,14 +41,25 @@
...
 import org.objectweb.asm.tree.MethodNode;
 import org.objectweb.asm.tree.TypeInsnNode;
+import org.objectweb.asm.tree.analysis.Analyzer;
+import org.objectweb.asm.tree.analysis.BasicInterpreter;
+import org.objectweb.asm.tree.analysis.BasicValue;
+import org.objectweb.asm.util.Printer;
+import org.objectweb.asm.util.Textifier;
+import org.objectweb.asm.util.TraceMethodVisitor;
+
+import static com.google.common.base.Preconditions.checkArgument;
```
cgruber, line 51: I think the square formatter got to you. I think these need to go at the top. Does bazel code use google-java-format?

restingbull: Ayup. Fixed.

WORKSPACE
```diff
@@ -78,8 +78,8 @@ list_source_repository(name = "local_bazel_source_list")
 #   2. Set the $ANDROID_HOME and $ANDROID_NDK_HOME environment variables
 #   3. Uncomment the two lines below
 #
-# android_sdk_repository(name = "androidsdk")
-# android_ndk_repository(name = "androidndk")
+android_sdk_repository(name = "androidsdk")
```
jin, line 81: Revert this please.

restingbull: Apologies, fixed.

restingbull:
Ping?

src/tools/android/java/com/google/devtools/build/android/desugar/LambdaDesugaring.java
```diff
@@ -494,62 +555,106 @@ public void visitInvokeDynamicInsn(String name, String desc, Handle bsm, Object.
...
-          insn = insn.getPrevious();
-
-          while (insn.getOpcode() == Opcodes.GETFIELD) {
-            // Nested inner classes can cause a cascade of getfields from the outermost one inwards
+
+      // Walk backwards past push instructions corresponding to constructor arguments to find the
+      // instruction after which we need to insert our NEW/DUP pair
+
+      AbstractInsnNode insn = findInstruction(instructions.getLast());
+      int ignorePushes = 0;
```
kevin1e100, line 563: how about "skipPushes"?

src/tools/android/java/com/google/devtools/build/android/desugar/LambdaDesugaring.java
```diff
@@ -473,7 +513,28 @@ public void visitInvokeDynamicInsn(String name, String desc, Handle bsm, Object.
...
+          n.accept(visitor);
+        }
+        StringWriter b = new StringWriter();
+        p.print(new PrintWriter(b));
+        throw new IllegalStateException("Unexpected instructions in :\n" + b.toString(), e);
+      }
+    }
+
+    /** If node is not an instruction, scans backwards until it finds one that is. */
+    private AbstractInsnNode findInstruction(AbstractInsnNode node) {
```
kevin1e100, line 531: is skipping over labels really a good idea?  seems like if there's branches then rewriting code before and after an incoming jump destination could be problematic, not to mention that we may not even be looking at code that will ever executed in the order that it appears (in the case of unconditional jumps)

sgowroji:
Hello @restingbull, Could you please respond to the above review comments. Thanks!

src/tools/android/java/com/google/devtools/build/android/desugar/LambdaDesugaring.java
```diff
@@ -366,12 +373,45 @@ private static String packageName(String internalClassName) {
...
+    public static MethodReferenceBridgeInfo bridge(
+        Handle methodReference, Executable referenced, Handle bridgeMethod) {
+      checkArgument(!bridgeMethod.equals(methodReference));
+      return new AutoValue_LambdaDesugaring_MethodReferenceBridgeInfo(
+          methodReference, checkNotNull(referenced), bridgeMethod);
+    }
+
+    public abstract Handle methodReference();
+
+    /** Returns {@code null} iff {@link #bridgeMethod} equals {@link #methodReference}. */
```
inez, line 399:
```suggestion
    /** Returns {@code null} if {@link #bridgeMethod} equals {@link #methodReference}. */
```

PR #11664, opened 2020-06-29
Add python module six as dependency to bazel_external_repository_test

### LINKED ISSUE #11663: bazel_external_repository_test: ModuleNotFoundError: No module named 'six'
When running **bazel_external_repository_test** we're encountering the following issue:
`fastbuild/bin/src/test/py/bazel/bazel_external_repository_test.runfiles/io_bazel/src/test/py/bazel/bazel_external_repository_test.py", line 19, in <module>
from six.moves import SimpleHTTPServer
ModuleNotFoundError: No module named 'six'`  

.. where the failing line in the tests looks like this: `from six.moves import SimpleHTTPServer
`

Reproducing this issue is fairly hard since we're not 100% sure about what's causing this issue in our specific environment.

Red Hat Enterprise Linux 7 (RHEL7)

No

When looking in the relevant BUILD-file we find that the BUILD initially looks like:

`py_test(
    name = "bazel_external_repository_test",
    size = "medium",
    srcs = ["bazel_external_repository_test.py"],
    data = glob(["testdata/bazel_external_repository_test/**"]),
    tags = [
        "requires-network",
    ],
    deps = [
        ":test_base"
    ],
)`

Which results in the issue stated as above. 

But by adding: `//third_party/py/six` in the list of dependencies, resulting in 
`py_test(
    name = "bazel_external_repository_test",
    size = "medium",
    srcs = ["bazel_external_repository_test.py"],
    data = glob(["testdata/bazel_external_repository_test/**"]),
    tags = [
        "requires-network",
    ],
    deps = [
        ":test_base",
        "//third_party/py/six",
    ],
)`

This issue is fixed and no more issues are envoked when running the test. 

So now to the question: We do not understand how this test can pass in the orignial Bazel repo without this dependency added to the list. We're still not sure if this might is to be seen as a bug in the system, or if it is something related to our own environment? Thoughts on this?

ehkloaj:
In the test bazel_external_repository_test the python module six was not added as a dependency which caused the test to fail in some envoirnments.

This closes #11663

### DIFF
--- a/src/test/py/bazel/BUILD
+++ b/src/test/py/bazel/BUILD
@@ -76,7 +76,10 @@ py_test(
     tags = [
         "requires-network",
     ],
-    deps = [":test_base"],
+    deps = [
+        ":test_base",
+        "//third_party/py/six",
+    ],
 )
 
 py_test(


ehkloaj:
@googlebot I signed it!

philwo:
For some reason, I get an "unknown pull request" error when trying to import this change, but I'll merge this manually and will credit you. 👍

PR #11665, opened 2020-06-29
Bump javac11 java_tools to v9.0

philwo:


philwo:
> Should we also consider to bump javac12 java_tools as well? I don't need it, and would even consider to remove javac12 altogether.

Mhm. I'm leaning towards removing it. Java 12 is already end-of-life. @meisterT WDYT?

meisterT:
Let's not do Java 12 unless someone asks for it.

davido:
Yay, I can see new `java_tools` released, thanks. Tension mounts: how many downstream project on Buildkite CI would break...

meteorcloudy:
Hmm.. this upgrade is breaking some projects in downstream:
https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/199

philwo:
It seems like they are failing due to (new?) ErrorProne checks, e.g.

```
external/io_grpc_grpc_java/core/src/main/java/io/grpc/internal/RetriableStream.java:702: error: [GuardedBy] This access should be guarded by 'this.scheduledHedging.lock'; instead found: 'this.lock'
--
  | futureToBeCancelled = scheduledHedging.markCancelled();
  | ^
  | (see https://errorprone.info/bugpattern/GuardedBy)
```

This is probably WAI, but we should document how to disable newly introduced warnings - is there a command-line flag that we could set in their presubmit.yml?

davido:
> Tension mounts: how many downstream project on Buildkite CI would break...

>> Hmm.. this upgrade is breaking some projects in downstream:
https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/199

//CC @cushon 

We should really try hard not to revert the EP upgrade, as it was done during last attempt to bump EP version.
The question is: are these errors flagged by EP justified and could be easily fixed? If they are false positive, error should be files against EP project.

If all other options fail, there is `javacopt` to disable every single EP bug pattern. See this real example here: [1].

[1] https://gerrit-review.googlesource.com/c/gerrit/+/237183

meteorcloudy:
I think they are probably all valid bug pattern report, but I'm more inclined to fix the project itself than adding build options to their presubmit.yml.

philwo:
> I'm more inclined to fixing the project itself

That's the spirit 😁 Yeah, let's file issues with them and try to fix them!

meteorcloudy:
OK, issues filed and PRs sent.

cushon:
`GuardedBy` is an existing check, but there have been some bugs fixed in it that could cause it to report additional diagnostics.

Thanks for putting together those fixes. If that ends up being too disruptive, an alternative to abandoning the EP upgrade is to explicitly disable the problematic check on Bazel's side, i.e. by adding `-Xep:GuardedBy:WARN` (or `OFF`) here: https://github.com/bazelbuild/bazel/blob/1b22a2421d3ba2dba1d74feabf843940d0cd7907/tools/jdk/default_java_toolchain.bzl#L45

meteorcloudy:
I see, thanks for the backup solution!

meteorcloudy:
OK, I feel it's probably better to add `-Xep:GuardedBy:WARN` as a default java option. Some projects are slow to respond, and it will save some trouble for our users.

davido:
> OK, I feel it's probably better to add -Xep:GuardedBy:WARN as a default java option.

*Sigh*, but OK. We should add a big TODO comment, that this check demotion is only temporary and should be removed once the known Bazel projects (reference the pending issues?) are fixed.

meteorcloudy:
It's just I really want our CI to be green for our cut of next Bazel release. But yeah, we can also temporarily disable them in downstream.

meteorcloudy:
How important is this check? `-Xep:GuardedBy:WARN` will turn it into a warning, so users will still be able to notice the bug pattern, is that OK?

davido:
> -Xep:GuardedBy:WARN will turn it into a warning, so users will still be able to notice the bug pattern, is that OK?

Yes, totally OK. Add this demotin to the Release Notes and mention that it could be removed in next Bazel releases?

meteorcloudy:
> My concern is, that such exception in toolchain definition would clutter the code and complicate addition of new toolchains, for JDK 14, 15 and 17.

I see. Then I'll give more time for bazel-buildfarm and googlecloudrobotics to migrate. Hopefully we don't have to add this flag.

meisterT:
I would want to cut the new release (3.4) by tomorrow, see #11687. For that we need to get this resolved in one way or another.
Obviously I can also cut before this PR has landed but I would prefer not to.

meteorcloudy:
OK, I'll disable the two projects to give you a green downstream. Please make sure to cut after 2ed398021122fab00b0c4e2eb2bc7756c3f18209, which is required for Bazel Debian build.

meisterT:
Well, it gives me the illusion of a green downstream pipeline :-D
How annoying are are those fixes and how likely is it that they will land in time for the release?

meteorcloudy:
> Well, it gives me the illusion of a green downstream pipeline :-D

Yeah, but we always disable projects that fail to migrate in time. For example, Envoy, rules_k8s are already disabled. We cannot always wait for them.

> How annoying are are those fixes and how likely is it that they will land in time for the release?

That's the problem, I'm not sure. But I don't see the fixes will be merged before tomorrow.

meisterT:
Can you please link me to an example fixes?

davido:
> Can you please link me to an example fixes?

Fix build failure with Bazel@HEAD bazelbuild/bazel-buildfarm#441
Fix build failure with Bazel@HEAD googlecloudrobotics/core#49

meteorcloudy:
Nice, the fix for buildfarm is just merged.

davido:
> Nice, the fix for buildfarm is just merged.

Oh, great, re-enable it in downstream project on Bazel CI then?

meteorcloudy:
Will do!

PR #11666, opened 2020-06-29
Remove Java 12 support

philwo:
Java 12 is end-of-life and there's no point in investing work to maintain it.

Note that this doesn't actively remove any kind of support - people can still use `java_tools_javac12` in their WORKSPACE file and use their own OpenJDK 12 version as the host_javabase and it will continue to work just as before.

It just means we no longer test changes against it, release new java_tools versions for it and no longer provide bundled `http_archive` definitions for OpenJDK 12 (which is the right thing to do, as we don't want to point our users to unmaintained OpenJDK versions that might eventually be found to contain security issues).


philwo:
FYI @davido

davido:
Thanks, @philwo that's long overdue.

PR #11668, opened 2020-06-29
Fixes build issue with JDK headers location on ppc64le.

### LINKED ISSUE #11643: bazel-3.3.0 fails to bootstrap on ppc64le

While bootstrapping bazel-3.3.0 on ppc64le:

```
In file included from src/main/java/com/google/devtools/build/lib/syntax/cpu_profiler_posix.cc:20:0:
bazel-out/ppc-opt/bin/external/bazel_tools/tools/jdk/include/jni.h:45:10: fatal error: jni_md.h: No such file or directory
 #include "jni_md.h"
          ^~~~~~~~~~
compilation terminated.
Target //src:bazel_nojdk failed to build
INFO: Elapsed time: 173.915s, Critical Path: 19.09s
INFO: 735 processes: 735 local.
```

This is probably because `jni_md.h` is in `jdk/include/linux`, not `jdk/include`. Looking at the JDK installation path on Ubuntu 18.04:
```
ubuntu@horus:~$ sudo find / -name jni_md.h
/usr/lib/jvm/java-8-openjdk-ppc64el/include/linux/jni_md.h
/usr/lib/jvm/java-11-openjdk-ppc64el/include/linux/jni_md.h
```

Just try a bootstrap build on ppc64le.

Ubuntu 18.04.3

ceseo:
Fixes #11643.

### DIFF
--- a/tools/jdk/BUILD
+++ b/tools/jdk/BUILD
@@ -135,7 +135,7 @@ cc_library(
     name = "jni",
     hdrs = [":jni_header"] + select({
         "//src/conditions:linux_aarch64": [":jni_md_header-linux"],
-        "//src/conditions:linux_ppc": [":jni_md_header-linux"],
+        "//src/conditions:linux_ppc64le": [":jni_md_header-linux"],
         "//src/conditions:linux_s390x": [":jni_md_header-linux"],
         "//src/conditions:linux_x86_64": [":jni_md_header-linux"],
         "//src/conditions:darwin": [":jni_md_header-darwin"],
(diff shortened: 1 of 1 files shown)


philwo:
Thanks for the patch @ceseo! Bazel (officially) doesn't support or test on PowerPC at all currently, so we rely on the community to test and send us patches that improve support for that platform. I think this looks good, so let's merge it.

ceseo:
@philwo what would be required for a ppc64le CI? I could help with hardware access if necessary. In Golang we have a bunch of systems hosted at Oregon State University. Does that interest you?

PR #11671, opened 2020-06-29
Add oso_prefix_is_pwd macOS feature

keith:
With Xcode 11 Apple's linker introduced a new oso_prefix option which
removes the given prefix from the path in debug info. This improves the
goal of producing reproducible builds.

More info: https://milen.me/writings/apple-linker-ld64-deterministic-builds-oso-prefix

RELNOTES: Add opt in 'oso_prefix_is_pwd' feature for Apple builds

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5451,6 +5451,17 @@ def _impl(ctx):
         requires = [feature_set(features = ["opt"])],
     )
 
+    oso_prefix_feature = feature(
+        name = "oso_prefix_is_pwd",
+        flag_sets = [
+            flag_set(
+                actions = all_link_actions +
+                          ["objc-executable", "objc++-executable"],
+                flag_groups = [flag_group(flags = ["OSO_PREFIX_MAP_PWD"])],
+            ),
+        ],
+    )
+
     generate_dsym_file_feature = feature(
         name = "generate_dsym_file",
         flag_sets = [
@@ -5782,6 +5793,7 @@ def _impl(ctx):
             default_compile_flags_feature,
             debug_prefix_map_pwd_is_dot_feature,
             generate_dsym_file_feature,
+            oso_prefix_feature,
             contains_objc_source_feature,
             objc_actions_feature,
             strip_debug_symbols_feature,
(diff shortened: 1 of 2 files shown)


keith:
If there's a better way to do this type of feature that requires a specific Xcode version I'm happy to change this, but I figured it being opt in is probably "good enough" (although it's possible we want to join it with `-fdebug-prefix-map`'s logic

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5451,6 +5451,17 @@ def _impl(ctx):
         requires = [feature_set(features = ["opt"])],
     )
 
+    oso_prefix_feature = feature(
+        name = "oso_prefix_is_pwd",
+        flag_sets = [
+            flag_set(
+                actions = all_link_actions +
+                          ["objc-executable", "objc++-executable"],
```
michaeleisel, line 5459: will this flag apply to executables without any objc as well?

michaeleisel:
what's the status of debug-prefix-map in bazel? since it requires that each compile invocation change based on the path, as opposed to just the link invocation, does it invalidate any bazel caches for that reason?

otherwise, i haven't tested this linker flag myself but this look sound overall

tools/osx/crosstool/cc_toolchain_config.bzl
keith: Should https://github.com/keith/bazel/blob/ks/add-oso_prefix_is_pwd-macos-feature/tools/osx/crosstool/cc_toolchain_config.bzl#L246-L250

keith:
Both debug-prefix-map and this one are "hidden" in the clang wrapper, so the command line is stable, and the wrapper resolves and puts the arbitrary path in the actually invocation (rules_swift does something similar too) so they don't invalidate any caches.

michaeleisel:
👍

jin:
+ @allevato this seems like a useful thing to have, and I don't see it set up in internal toolchains.

allevato:
@googlewalt @trybka

trybka:
> +@allevato this seems like a useful thing to have, and I don't see it set up in internal toolchains.

We are tracking this internally.

keith:
Even if this was fixed internally that wouldn't apply to the public crosstool correct?

trybka:
> Even if this was fixed internally that wouldn't apply to the public crosstool correct?

Correct. Not at this time. This change seems fine.

keith:
Is this aligned with how it would work with the internal crosstool? It doesn't sound like that's happening anymore but if there's a small change I could make to be more in line with that I'd be happy to!

trybka:
Yes, I think this is how it would be handled. wrapped_clang needs to set this because we don't know pwd until the compile/link is executing on the machine.

jin:
@trybka I'll import this for your review.

keith:
Thanks!!

PR #11682, opened 2020-06-30
Replace the remaining dependencies for Bazel Debian build (third_party)

meteorcloudy:
Working towards https://github.com/bazelbuild/bazel/issues/9408

Thanks to @olekw, the remaining jar dependencies are all packaged for Debian. We'll verify if they work as soon as they are available in unstable distribution.


meteorcloudy:
FYI @olekw

PR #11688, opened 2020-07-01
Tool paths are relative to cc_toolchain not suite

illicitonion:
If your cc_toolchain_suite references cc_toolchains in different
packages, currently path computation is incorrect because it attempts to
append relative paths to the package containing the cc_toolchain_suite
not the cc_toolchain.

This commit makes the paths be looked up relative to the cc_toolchain.

Where these happen to be the same package (which must always have been
the case before, beacuse this code requires it), the behaviour will be
identical. But if separate packages are used, that will not work too.


lberki:
@oquenchil I'm assuming you'll take a look at this eventually; let me know if you are overbooked.

illicitonion:
@oquenchil @lberki Would it be possible to take a look at this? I'm running into issues because of not having it :) Thanks!

PR #11696, opened 2020-07-02
Add support for object files in the Starlark implementation of cc_import

agluszak:
This change allows users to import object files when using Starlark version of cc_import rule (currently hidden behind '--experimental_starlark_cc_import' flag). Previously in order to import object files users were required to put them in the srcs field of cc_library.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
@@ -394,6 +394,8 @@ public abstract class CcModule
    * @param alwayslink boolean
    * @param dynamicLibraryPath String
    * @param interfaceLibraryPath String
+   * @param picObjectFiles Sequence<Artifact>
+   * @param nopicObjectFiles Sequence<Artifact>
    * @return
    * @throws EvalException
    * @throws InterruptedException
@@ -407,6 +409,8 @@ public abstract class CcModule
       Object picStaticLibraryObject,
       Object dynamicLibraryObject,
       Object interfaceLibraryObject,
+      Object picObjectFiles, // Sequence<Artifact> expected
+      Object nopicObjectFiles, // Sequence<Artifact> expected
       boolean alwayslink,
       String dynamicLibraryPath,
       String interfaceLibraryPath,
@@ -422,6 +426,10 @@ public abstract class CcModule
     Artifact picStaticLibrary = nullIfNone(picStaticLibraryObject, Artifact.class);
     Artifact dynamicLibrary = nullIfNone(dynamicLibraryObject, Artifact.class);
     Artifact interfaceLibrary = nullIfNone(interfaceLibraryObject, Artifact.class);
+    ImmutableList<Artifact> picObjects =
+        nullIfNoneOrEmptySequence(picObjectFiles, Artifact.class, "PIC object files");
+    ImmutableList<Artifact> nopicObjects =
+        nullIfNoneOrEmptySequence(nopicObjectFiles, Artifact.class, "No-PIC object files");
 
     StringBuilder extensionErrorsBuilder = new StringBuilder();
     String extensionErrorMessage = "does not have any of the allowed extensions";
@@ -580,6 +588,8 @@ public abstract class CcModule
         .setResolvedSymlinkDynamicLibrary(resolvedSymlinkDynamicLibrary)
         .setInterfaceLibrary(interfaceLibrary)
         .setResolvedSymlinkInterfaceLibrary(resolvedSymlinkInterfaceLibrary)
+        .setObjectFiles(nopicObjects)
+        .setPicObjectFiles(picObjects)
         .setAlwayslink(alwayslink)
         .build();
   }
@@ -1536,6 +1546,13 @@ public abstract class CcModule
   @Nullable
   private static <T> T nullIfNone(Object object, Class<T> type) {
     return object != Starlark.NONE ? type.cast(object) : null;
+  }
+
+  @Nullable
+  private static <T> ImmutableList<T> nullIfNoneOrEmptySequence(
+      Object object, Class<T> type, String what) throws EvalException {
+    Sequence<T> seq = Sequence.noneableCast(object, type, what);
+    return seq.isEmpty() ? null : seq.getImmutableList();
   }
 
   @Override
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
@@ -395,7 +395,7 @@ public abstract class CcModule
    * @param dynamicLibraryPath String
    * @param interfaceLibraryPath String
    * @param picObjectFiles Sequence<Artifact>
-   * @param nopicObjectFiles Sequence<Artifact>
+   * @param objectFiles Sequence<Artifact>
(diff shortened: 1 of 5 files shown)


tools/build_defs/cc/cc_import.bzl
```diff
@@ -71,6 +75,60 @@ def _perform_error_checks(
         not _is_shared_library_extension_valid(shared_library_artifact.basename)):
         fail("'shared_library' does not produce any cc_import shared_library files (expected .so, .dylib or .dll)")
 
+    if static_library != None and (bool(pic_object_files) or bool(nopic_object_files)):
+        fail("'static_library' and object files should not be provided at the same time")
+
+    if bool(pic_object_files) and bool(nopic_object_files):
+        fail("Only one type of object files (PIC or no-PIC) should be provided")
```
agluszak, line 82: Do these checks make sense?

agluszak: If yes, I should probably write tests for them...

tools/build_defs/cc/tests/cc_import_test.bzl
```diff
@@ -76,15 +79,84 @@ def _cc_import_includes_test_impl(ctx):
...
+    env = analysistest.begin(ctx)
+    _assert_linkopts_present(env, "bazel-out/k8-fastbuild/bin/" + TESTS_PATH + "libcc_import_pic_object_files_no_start_end_lib_test_import.pic.a")
+    return analysistest.end(env)
+
+cc_import_pic_object_files_no_start_end_lib_test = analysistest.make(_cc_import_pic_object_files_no_start_end_lib_test_impl)
+
+def _cc_import_nopic_object_files_no_start_end_lib_test_impl(ctx):
+    env = analysistest.begin(ctx)
+    for a in analysistest.target_actions(env):
+        print(a.argv)
```
agluszak, line 106: Oops, forgot to remove that

src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
```diff
@@ -1538,6 +1548,13 @@ private static void getLegacyArtifactNamePatterns(
     return object != Starlark.NONE ? type.cast(object) : null;
   }
 
+  @Nullable
+  private static <T> ImmutableList<T> nullIfNoneOrEmptySequence(
+      Object object, Class<T> type, String what) throws EvalException {
+    Sequence<T> seq = Sequence.noneableCast(object, type, what);
+    return seq.isEmpty() ? null : seq.getImmutableList();
+  }
+
```
c-mita, line 1557: This feels weird; returning null instead of an empty list when the input is empty. I would normally expect the opposite (null being transformed into an empty list); is there a reason for this?

tools/build_defs/cc/cc_import.bzl
oquenchil: The first one makes sense. The second one doesn't, because it's up to the target later downstream to decide if it wants pic or no-pic, so we should make both available.

src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java
agluszak: Yup, in LibraryToLink there is a check for null [](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/LibraryToLink.java;l=383?q=librarytolink&ss=bazel%2Fbazel)
but maybe instead of replacing an empty list here I should check inside cc_import.bzl if it's empty and simply pass null there [](https://github.com/bazelbuild/bazel/pull/11696/files#diff-de10b4edef58e36c067b24bf39170475R189). AFAIK when a bazel rule has a label_list attribute then this attribute cannot have a default None value, it must be an empty list [](https://github.com/bazelbuild/bazel/pull/11696/files#diff-de10b4edef58e36c067b24bf39170475R224)

tools/build_defs/cc/cc_import.bzl
```diff
@@ -128,10 +205,10 @@ def _cc_import_impl(ctx):
         name = ctx.label.name,
     )
 
-    return [CcInfo(
-        compilation_context = compilation_context,
-        linking_context = linking_context,
-    )]
+    if output_file:
+        _create_archive(ctx, feature_configuration, cc_toolchain, output_file, object_files)
```
c-mita, line 203: Part of me feels that `_create_archive` should be inlined; that declaring and defining an action is too important to bury behind this call. But I don't feel like I can insist.

tools/build_defs/cc/cc_import.bzl
```diff
@@ -97,12 +155,29 @@ def _cc_import_impl(ctx):
...
+        ctx.files.nopic_object_files,
     )
 
     (no_pic_static_library, pic_static_library) = _get_no_pic_and_pic_static_library(
         ctx.file.static_library,
     )
 
+    output_file = None
+
+    if bool(ctx.files.pic_object_files):
```
c-mita: I don't think you need these calls to `bool`.

tools/build_defs/cc/cc_import.bzl
meisterT: what about using a name like `_create_archive_action` to make this more obvious?

konste:
@agluszak I see you added these checks:
```
    if (nopicObjects != null && staticLibrary == null) {
      throw Starlark.errorf("If you pass 'objects' you must also pass a 'static_library'");
    }
    if (picObjects != null && picStaticLibrary == null) {
      throw Starlark.errorf("If you pass 'pic_objects' you must also pass a 'pic_static_library'");
    }
```
And wonder why this limitation?

We have a custom rule similar to cc_library which compiles sources and produces a collection of object files as the output. We don't want those object files to be bundled into a static library. So that the custom rule at the end does this:
```
    library_to_link = cc_common.create_library_to_link(
        actions = ctx.actions, feature_configuration = feature_configuration, cc_toolchain = cc_toolchain,
        pic_objects = compilation_outputs.pic_objects,
        objects = compilation_outputs.objects,
    )
    linker_input = cc_common.create_linker_input(
        owner = ctx.label,
        libraries = depset([library_to_link]),
    )
    linking_context = cc_common.create_linking_context(linker_inputs = depset([linker_input])) 
```
and unfortunately it does not work because it hits your checks.

So I wonder if the checks can be removed or how can we work around it.

oquenchil:
The checks cannot be removed.

The reason why we are forcing a static library to be passed here is because we can't rely on a custom rule at the end doing this. This is not possible through the API and it wasn't possible with cc_library before which implicitly created a static library if you had object files in srcs.

If you have a custom rule at the end, then the workaround is to pass any file in the static library parameter and simply ignore that file later.

konste:
There is misunderstanding about "at the end" part. What I meant is that our custom rule cc_compile implementation function has this code snippet at the end of it's code. Not that we have another custom rule at the end of the build graph. The plan is to make cc_compile so that native cc_binary can depend on it.
So you are saying that if cc_compile would create a library_to link with objects, but no static lib file then other cc_rules depending on it would choke?
But then the same would happen if we pass any random file in the static library parameter - depending native rules may try to use it and blow up.
I am thinking about passing not just any random file there, but specially crafted "empty library" file (which must be platform specific), but I am not sure how depending native rules would interpret that combo - library_to_link with some objects and empty static lib.

PR #11701, opened 2020-07-03
Sort repo directories in python_stub_template

glukasiknuro:
The listdir does not guarantee any order, and lack of the ordering looks like potential issue of non-reproducible runs.


lberki (approved):
Looks good! (after a somewhat embarrassing amount of time). I'll import this.

Issue #7085, opened 2019-01-10
Use keepAliveTime on grpc channels


werkt:
### Description of the problem / feature request:

bazel as a grpc client is unconfigurable with respect to keepAliveTime as available with the channel transport. I propose adding configuration to allow a transport keepAliveTime to be specified, possibly per service.

### Feature requests: what underlying problem are you trying to solve with this feature?

Presenting a grpc endpoint through network transports (i.e. commodity load balancers) that insist on non-idle connections to maintain connectivity results - particularly with remote execution and waiting for operation responses - unrecognized connection loss. This will stall executions for a transport-implementation dependent/indeterminate amount of time until the client realizes the connection has dropped.

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

release 0.21.0

buchgr:
IIRC the current behavior should be that connection failures are retried automatically by the gRPC library. Is that not the case? Is the idea behind a keepAliveTime to fail faster?

werkt:
Realistically, yes, fail faster is the expected outcome of this, since I've seen connection states where (java-)grpc is ignorant of the connection loss induced by the load balancer. The keepalive will cause terminated connections in this state to yield the connection loss reaction by grpc.

v2046:
Yes

v2046:
P2

v2046:
Yes

v2046:
Vincy

ulfjack:
This was fixed in https://github.com/bazelbuild/bazel/commit/761c16265139c2e5f5db50158d39c6717094ec49.

PR #11702, opened 2020-07-03
Allow keepalives to be enabled for REv2 gRPC channels

EdSchouten:
Bazel currently doesn't make use of gRPC's keepalive options, meaning
that builds may get stuck unnecessarily when connected to a server that
disappears off the network. It may well be the case that reconnecting to
an alternative server allows the build to proceed.

Considering that Bazel's RPC rate is already relatively high, there is
generally little harm in enabling keepalives from a performance point of
view, especially if we don't send them when idle (i.e., the default).

This change introduces two new configuration options,
--remote_keepalive_time (the interval) and --remote_keepalive_timeout.

Fixes: https://github.com/bazelbuild/bazel/issues/7085

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -34,10 +34,12 @@ import com.google.devtools.common.options.OptionMetadataTag;
 import com.google.devtools.common.options.OptionsBase;
 import com.google.protobuf.TextFormat;
 import com.google.protobuf.TextFormat.ParseException;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.SortedMap;
+import javax.annotation.Nullable;
 
 /** Options for remote execution and distributed caching. */
 public final class RemoteOptions extends OptionsBase {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -433,6 +435,33 @@ public RemoteOutputsStrategyConverter() {
...
+      name = "remote_keepalive_time",
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      help =
+          "Sets the time without read activity before sending a keepalive ping. Clients must "
+              + "receive permission from the service owner before enabling this option. Keepalives "
+              + "can increase the load on services and are commonly \"invisible\" making it hard "
+              + "to notice when they are causing excessive load. Clients are strongly encouraged "
+              + "to use only as small of a value as necessary.")
```
ulfjack, line 448: I think this last sentence is confusing. It would be clearer to say that "larger values for this option generate less traffic, so only use a smaller value if necessary". More explicit guidance would be even better, e.g., "don't set this to less than one minute".

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -433,6 +435,33 @@ public RemoteOutputsStrategyConverter() {
...
+
+  @Option(
+      name = "remote_keepalive_timeout",
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      help =
+          "Sets the time waiting for read activity after sending a keepalive ping. If the time "
+              + "expires without any read activity on the connection, the connection is considered "
+              + "dead. This value should be at least multiple times the RTT to allow for lost "
```
ulfjack, line 460: Do not use abbreviations - they may not be universally clear to all readers.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
EdSchouten: Yeah, that sounds a lot better. I took these descriptions from the documentation of the gRPC Java library and already redacted them slightly. This seems like a good improvement.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
EdSchouten: Agreed!

buchgr:
Hi @EdSchouten :wave: 

One aspect of gRPC keep alives to keep in mind is that the client stops sending PING frames after having sent 2 [1] consecutive PING frames without any DATA or HEADER frames in between. This does not only happen when the client is idle but also when it waits for an Execute call. So if you set the keep alive time to 1 minute it will not detect/repair a broken connection if Execute takes longer than two minutes. So while helpful gRPC keep alives might not be what we want in Bazel.

An alternative idea to make Execute more resilient was for Bazel to regularly close and reopen the Execute call. That being said, keep alives should still be a win for other calls.

[1] Default value of GRPC_ARG_HTTP2_MAX_PINGS_WITHOUT_DATA: https://github.com/grpc/grpc/blob/master/doc/keepalive.md#defaults-values

jin:
Is still still being worked on? Assigning back to PR author.

ulfjack:
Ouch! I'm sorry, Ed. I got https://github.com/bazelbuild/bazel/commit/761c16265139c2e5f5db50158d39c6717094ec49 merged. I apparently totally forgot about this pending PR. I think this can be closed now.

EdSchouten:
> Ouch! I'm sorry, Ed. I got [761c162](https://github.com/bazelbuild/bazel/commit/761c16265139c2e5f5db50158d39c6717094ec49) merged. I apparently totally forgot about this pending PR. I think this can be closed now.

No need to feel sorry! Awesome we got this fixed! I was too busy to push this through myself.

PR #11703, opened 2020-07-03
REv2: Allow downloads to be prevented by emitting symlinks

EdSchouten:
By default, Bazel's REv2 client downloads all output files of a build
action, which is slow. This can be disabled by using flags like
--remote_download_minimal, but this causes bazel-out/ to be mostly
empty, making the user experience suboptimal.

This change introduces a middle ground: it allows people to develop
virtual file systems (FUSE, NFS, 9p, etc. etc. etc.) that lazily expose
the CAS onto people's systems. It does this by adding a new flag,
--remote_download_symlink_template, that causes Bazel to generate
symlinks that conform to a given template.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
@@ -445,6 +445,18 @@ public class RemoteCache implements AutoCloseable {
       return COMPLETED_SUCCESS;
     }
 
+    if (!options.remoteDownloadSymlinkTemplate.isEmpty()) {
+      // Don't actually download files from the CAS. Instead, create a
+      // symbolic link that points to a location where CAS objects may
+      // be found. This could, for example, be a FUSE file system.
+      path.createSymbolicLink(
+          path.getRelative(
+              options.remoteDownloadSymlinkTemplate
+                  .replace("{hash}", digest.getHash())
+                  .replace("{size_bytes}", String.valueOf(digest.getSizeBytes()))));
+      return COMPLETED_SUCCESS;
+    }
+
     OutputStream out = new LazyFileOutputStream(path);
     SettableFuture<Void> outerF = SettableFuture.create();
     ListenableFuture<Void> f = cacheProtocol.downloadBlob(digest, out);
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -460,6 +460,21 @@ public final class RemoteOptions extends OptionsBase {
               + " discard the remotely cached values if they don't match the expected value.")
   public boolean remoteVerifyDownloads;
 
+  @Option(
+      name = "remote_download_symlink_template",
+      defaultValue = "",
+      category = "remote",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
+      help =
+          "Instead of downloading remote build outputs to the local machine, create symbolic "
+              + "links. The target of the symbolic links can be specified in the form of a "
+              + "template string. This template string may contain {hash} and {size_bytes} that "
+              + "expand to the hash of the object and the size in bytes, respectively. "
+              + "These symbolic links may, for example, point to a FUSE file system "
+              + "that loads objects from the CAS on demand.")
+  public String remoteDownloadSymlinkTemplate;
+
   // The below options are not configurable by users, only tests.
   // This is part of the effort to reduce the overall number of flags.
 
--- a/src/test/java/com/google/devtools/build/lib/remote/RemoteCacheTests.java
+++ b/src/test/java/com/google/devtools/build/lib/remote/RemoteCacheTests.java
@@ -1207,6 +1207,33 @@ public class RemoteCacheTests {
   }
 
   @Test
+  public void testDownloadFileWithSymlinkTemplate() throws Exception {
+    // Test that when a symlink template is provided, we don't actually download files to disk.
+    // Instead, a symbolic link should be created that points to a location where the file may
+    // actually be found. That location could, for example, be backed by a FUSE file system that
+    // exposes the Content Addressable Storage.
+
+    // arrange
+    final ConcurrentMap<Digest, byte[]> cas = new ConcurrentHashMap<>();
+
+    Digest helloDigest = digestUtil.computeAsUtf8("hello-contents");
+    cas.put(helloDigest, "hello-contents".getBytes(Charsets.UTF_8));
+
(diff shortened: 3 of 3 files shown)


ulfjack:
I am not convinced this feature carries its weight. If you're going to write a FUSE filesystem, why not write one that shows the actual file system layout? Why not have the remote execution system return the symlinks directly by rewriting the responses?

At the same time, the template system here isn't powerful enough even for simple cases - on Linux ext2-4, having a single directory with all the cas entries is too slow. And, it's losing performance because symlinks are slow.

I hope and expect that someone (maybe me) will write a FUSE filesystem that shows the actual file layout and that fully integrates with Bazel, and then this feature will become irrelevant, but it will be difficult to remove again.

EdSchouten:
Hey Ulf!

I'm going to reply to your message 

> I am not convinced this feature carries its weight. If you're going to write a FUSE filesystem, why not write one that shows the actual file system layout? Why not have the remote execution system return the symlinks directly by rewriting the responses?

Those are good questions. To answer the second question, I think the answer to that would be that it then requires that all clients have the FUSE file system mounted, and potentially have it mounted at the same location on the system. Or you somehow have to make it possible for clients to specify a template to the remote execution system. That may work for ActionResult messages, but does become a bit hairy when the ActionResult message points to Tree objects.

With regards to the first question: the idea behind this approach was to keep the responsibility of tracking state where it is now: in bazel-out/. The FUSE daemon that exposes the CAS would remain a stateless process. It could be a single per-user (or even system-wide?) daemon that caches and serves blobs for all Bazel workspaces/checkouts that a person may be working on.

> At the same time, the template system here isn't powerful enough even for simple cases - on Linux ext2-4, having a single directory with all the cas entries is too slow. And, it's losing performance because symlinks are slow.

That's a fair point. In my case I'm using a FUSE file system that exposes a directory that is not iterable. Hoping that the kernel is smart enough to use dirhashing, it should remain fairly fast.

> I hope and expect that someone (maybe me) will write a FUSE filesystem that shows the actual file layout and that fully integrates with Bazel, and then this feature will become irrelevant, but it will be difficult to remove again.

Oh, man. I would honestly donate one of my kidneys to see a feature like that appear. I mainly took this approach, because I already had a lot of good FUSE code lying around that was not written in Java, and because I'm not familiar enough with Bazel's innards to make a change like that.

buchgr:
I really like the approach that you took Ed. The code is very simple and easy to maintain, but gives users the option to use a FUSE filesystem as an alternative to --remote_download_minimal.

I have a few questions:
- IIRC Bazel will follow symlinks in the output base (unless they are declared an output symlink) and compute their digest by reading the file contents and hashing them. I assume this is not happening here. Is that so and do you know why?
- Wouldn't this feature already be a sufficient protocol to put the output base behind a fuse filesystem (as ulf suggests)? While Bazel would do a symlink() syscall, the fuse filesystem could expose a file instead. Shouldn't that work?

> At the same time, the template system here isn't powerful enough even for simple cases - on Linux ext2-4, having a single directory with all the cas entries is too slow. And, it's losing performance because symlinks are slow.

I think this problem could easily be addressed by introducing a third template variable {delim_hash}. It would be equal to the first few bytes of {hash} and allow you to construct paths like /cas/{delim_hash}/{hash}-{size_bytes}. Wouldn't that work? If say the upper limit of the filesystem is 2^32 files a reasonable length for {delim_hash} would be 2^16, so the first 4 characters of {hash}. It's safe to assume that {hash} is truly random and so even in the case of 4 billion output files we should get no more than 2^16 files in one directory.

buchgr:
> IIRC Bazel will follow symlinks in the output base (unless they are declared an output symlink) and compute their digest by reading the file contents and hashing them. I assume this is not happening here. Is that so and do you know why?

I am not trying to suggest that this can't be made work, I am just surprised that it works :-). I would expect that in addition to creating the symlink you will also need to inject the digest into the metadatahandler, like builds without the bytes does.

EdSchouten:
Hey Jakob! \o/

> I really like the approach that you took Ed. The code is very simple and easy to maintain, but gives users the option to use a FUSE filesystem as an alternative to --remote_download_minimal.
> 
> I have a few questions:
> 
> * IIRC Bazel will follow symlinks in the output base (unless they are declared an output symlink) and compute their digest by reading the file contents and hashing them. I assume this is not happening here. Is that so and do you know why?

Exactly! I'm using this change in combination with PR #11662, so that Bazel reads extended attributes in case it needs to reobtain the digest.

> * Wouldn't this feature already be a sufficient protocol to put the output base behind a fuse filesystem (as ulf suggests)? While Bazel would do a symlink() syscall, the fuse filesystem could expose a file instead. Shouldn't that work?

That could work as well, but has the downside that the FUSE file system needs to be stateful. It would need to keep track of the actual layout that was constructed by Bazel.

> > At the same time, the template system here isn't powerful enough even for simple cases - on Linux ext2-4, having a single directory with all the cas entries is too slow. And, it's losing performance because symlinks are slow.
> 
> I think this problem could easily be addressed by introducing a third template variablxe {delim_hash}. It would be equal to the first few bytes of {hash} and allow you to construct paths like /cas/{delim_hash}/{hash}-{size_bytes}. Wouldn't that work? If say the upper limit of the filesystem is 2^32 files a reasonable length for {delim_hash} would be 2^16, so the first 4 characters of {hash}. It's safe to assume that {hash} is truly random and so even in the case of 4 billion output files we should get no more than 2^16 files in one directory.

Sure! I was thinking that maybe we could use some kind of printf()-like mechanism, where people could do expressions like these:

```
--remote_download_symlink_template=/some/path/%.4(hash)s/%(hash)s-%(sizeBytes)d
```

Is there any convention for this in Java/Bazel land?

buchgr:
> Sure! I was thinking that maybe we could use some kind of printf()-like mechanism, where people could do expressions like these:

Why do you think this would carry its weight? In my mind hard-coding 4 characters seems sufficient.

> Is there any convention for this in Java/Bazel land?

Not that I am aware of.

buchgr:
Overall the change looks good to me. @philwo

EdSchouten:
> > Sure! I was thinking that maybe we could use some kind of printf()-like mechanism, where people could do expressions like these:
> 
> Why do you think this would carry its weight? In my mind hard-coding 4 characters seems sufficient.

That's a fair point. Using full string formatting for this is a bit overkill.

For my use case I don't think I need to shard objects across directories. In fact, I even prefer to keep it a flat namespace to not waste inodes. Would it be all right if we just stick to the patch as is, keeping the option open to add a special tag for 4 character prefixes? I'm more than happy to add it right now, but I suppose it only makes sense to add such a feature if people observe it to be a problem.

EdSchouten:
Friendly ping. :-)

EdSchouten:
Hey @coeuvre! I've just rebased this change on top of latest master. Is there anything left you want me to take care of?

coeuvre:
Merged! Thanks for your PR.

PR #11707, opened 2020-07-04
Fixes a typo in the JDK URL that would cause Bazel to warn.

gorakhargosh:
Fixes a typo in the JDK URL that would cause Bazel to issue warnings like the following:

```
WARNING: Download from https://mirror.bazel.com/openjdk/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.7%2B10/OpenJDK11U-jdk_ppc64le_linux_hotspot_11.0.7_10.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Unknown host: mirror.bazel.com
```


gorakhargosh:
@googlebot I consent

Issue #11278, opened 2020-05-03
Support disabling bitcode mode for iOS apps bundling a watchOS app


thii:
### Description of the problem / feature request:

Our iOS app bundles a watchOS app. We're currently not building the iOS app with bitcode, but embedding bitcode to the watchOS app binary is mandatory if you want to submit the app to the App Store. There's currently no way to disable generating bitcode for iOS app only.

Feature request: Add support for only generating bitcode for watchOS app.

### What underlying problem are you trying to solve with this feature?

Unable to submit an iOS app without bitcode to the App Store if it has a watchOS app.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Build an iOS app with a bundled watchOS app with `--apple_bitcode=embedded` flag produces an app that has bitcode embedded for both iOS app and watchOS app.

### What operating system are you running Bazel on?

macOS Catalina 10.15.4 (19E287)
### What's the output of `bazel info release`?

release 3.1.0

###  Have you found anything relevant by searching the web?

Only found https://github.com/bazelbuild/bazel/pull/7356 that added support bitcode for Apple platforms, but it forces bitcode for all platforms that you build for.

susinmotion:
@trybka, is this in your wheelhouse?

trybka:
@susinmotion sure, I can take a look. If this needs corresponding bazel changes, I'll ping it back, but I suspect this can be done in the crosstool alone.

susinmotion:
Wonderful, thank you!

thii:
@trybka Gently ping. Or maybe could you give some hints on this should be addressed?

trybka:
Can you see https://bazel-review.googlesource.com/c/bazel/+/135960 ? 

Try patching it in and building with `--features=omit_bitcode_for_ios_arm64` (replace ios_arm64 with whichever CPU(s) you're targeting, you can set it multiple times)

thii:
I got an error `MyFramework/MyFramework.__internal__.apple_binary.bcsymbolmap' was not created` while trying to build an app that has a dynamic framework.

thii:
`.bcsymbolmap` files are still declared as action outputs. I'm guessing this might need corresponding changes from Bazel, and probably rules_swift too.

trybka:
OK, let's take a step back then, and reconsider how this should work.

How is your project set up--rules / dependencies, etc.?

Do you just have an ios_application with a watchos_application set? 

[...]

PR #11710, opened 2020-07-04
Add transition support for Apple bitcode

thii:
Embedding bitcode to the watchOS app binary is mandatory if you want to
submit the app to the App Store. Previously it wasn't possible to enable
bitcode just for the watchOS app but not for the iOS app that bundles
it.

This patch adds a new `--watchos_force_bitcode` build option --- when
enabled, it will set the bitcode mode to "embedded" for the produced
watchOS binaries, regardless of the value of the global
`--apple_bitcode` option.

Closes https://github.com/bazelbuild/bazel/issues/11278.

RELNOTES: Apple bitcode for watchOS binaries can now be forced embedded
with the new flag `--watchos_force_bitcode`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
@@ -366,7 +366,18 @@ public class AppleCommandLineOptions extends FragmentOptions {
             + "Values: 'none', 'embedded_markers', 'embedded'."
   )
   public AppleBitcodeMode appleBitcodeMode;
-  
+
+  @Option(
+    name = "watchos_force_bitcode",
+    defaultValue = "false",
+    documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+    effectTags = {OptionEffectTag.LOSES_INCREMENTAL_STATE},
+    help =
+        "If true, use 'embedded' bitcode mode for watchOS binaries' compilation steps."
+            + "Enable this option if you only need bitcode for watchOS binaries but not "
+            + "for the iOS companion app, e.g. for App Store submission.")
+  public boolean watchosForceBitcode;
+
   /** Returns whether the minimum OS version is explicitly set for the current platform. */
   public DottedVersion getMinimumOsVersion() {
     DottedVersion.Option option;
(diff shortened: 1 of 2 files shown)


thii:
@trybka Could you take a look to see if this is the right fix?

trybka:
This seems about right. I would probably remove the CPU check, honestly. The fewer places that need to be updated when they rev their chips, the better. Unless you know of a case where it's harmful to do this for a simulator build?

thii:
Removed the CPU check. Thanks.

thii:
@trybka Could you help import this?

trybka:
Yup, lemme create the import change and get it reviewed.

jmmv:
This feels like a clutch but I know very little about this to tell.

Is watchOS the only place where we need this? Does the flag have to hardcode the name or it could be made more general? Wondering if we'll end up having to add extra flags for iOS, tvOS, etc. in the future.

Also, why are we making this conditional on `-c opt`? In other words, couldn't this be `--watchos_force_bitcode` and just do the thing no matter what the compilation mode is, relying on the user to only pass this for release builds? Or would that not work?

thii:
AFAIK watchOS is the only place where we need this--Apple requires bitcode for watchOS apps just for the ease of transition to 64-bit watchOS devices, and watchOS is the only platform that can have apps embedded into other platform (iOS)'s apps. If you build a watch app alone, you can already control this using the `--apple_bitcode` flag. 

I don't have a strong opinion on making this conditional on `-c opt` or not--I can remove this check. I added it because I find that it has little to no use-case of using this for non-optimized builds.

trybka:
If it's not harmful to do in fastbuild/dbg then might as well remove the check as jmmv suggests.

thii:
Removed the compilation mode check. Please take a look.

thii:
@jmmv Addressed your feedback. Could you review this again? Thanks.

thii:
Ping everyone. Any feedback would be greatly appreciated!

allevato:
We've gone a different direction in https://github.com/bazelbuild/bazel/commit/a4acc1b2e44b747572a1b174591c6c5d60844693 by letting the `--apple_bitcode` flag take an optional platform name to restrict the platforms to which Bitcode is applied. With that commit, doing a watchOS-only Bitcode build is achieved by passing `--apple_bitcode=watchos=embedded`.

thii:
Thanks a ton!

thii:
Looks like the test function names are mismatch:

https://github.com/bazelbuild/bazel/blob/052e9b704266c28f801c7cdb25de88c0414fbea0/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcLibraryTest.java#L740

and 

https://github.com/bazelbuild/bazel/blob/052e9b704266c28f801c7cdb25de88c0414fbea0/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcLibraryTest.java#L759

PR #11711, opened 2020-07-04
Remove always failing include path searching

keith:
Since this was run as part of the workspace setup, DEVELOPER_DIR and
SDKROOT are not set, which means wrapped_clang, which was being used as
the cc for this, will always fail. Whether or not the invocation fails
isn't taken into account for the result, and the expected output was
never produced, so this always returned nothing.

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -25,7 +25,6 @@ load(
     "@bazel_tools//tools/cpp:unix_cc_configure.bzl",
     "configure_unix_toolchain",
     "get_env",
-    "get_escaped_cxx_inc_directories",
 )
 
 def _get_escaped_xcode_cxx_inc_directories(repository_ctx, cc, xcode_toolchains):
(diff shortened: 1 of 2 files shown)


keith:
@c-parsons question for you about the TODO here, which actions were subject to this issue? In theory we could also fix this and just pass `clang` for macOS instead, but I agree with what it was getting at that this shouldn't matter given everything should be running with xcrun or through one of the wrappers that forces specific tools versions

keith:
Also note, given that this isn't run with DEVELOPER_DIR, the selection of paths here if we did fix this to call /usr/bin/clang directly, would be a bit misleading and probably shouldn't be included in acceptable include paths since they could be from a different installed Xcode version that your build doesn't specify

gregestren:
I'm not sure who has the best expertise on this code. I CC'ed the C++ rules folks (see labels) but please ping if you need me to look for a more precise contact after some time period.

jin:
Gong to pop this over to team-Apple, since it touches the macOS CC configuration rule.

keith:
bump

philwo:
@c-mita Could you please review this?

c-mita (approved):
As far as I can see, this is fine.

keith:
Thanks!

Issue #7224, opened 2019-01-23
use units in --remote_timeout=


buchgr:
--remote_timeout=INT currently expects seconds. this can be quite error prone as  https://github.com/bazelbuild/bazel/issues/5440 showed. We could instead implement parsing of time units i.e. `--remote_timeout=10s` or `--remote_timeout=1m`.

cc @nicolov

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

brentleyjones:
Can we reopen this one?

brentleyjones:
Sorry, @sgowroji, I was mistaken, it looks like this was completed it https://github.com/ehkloaj/bazel/commit/88201b061283fb48c6f8364edbe57b17ba80e203. You can close this as completed.

brentleyjones:
Looks like it was completed here: https://github.com/bazelbuild/bazel/commit/62475e738f1206a06dbda0d2638015c2c2fb1c88

PR #11713, opened 2020-07-05
Add units to --remote_timeout

bnczk:
Implements #7224. 

Maintains backwards compatibility by allowing unit-less values, defaulting to seconds.

Open issues are these Int casts [here](https://github.com/bazelbuild/bazel/compare/master...bnczk:remote-timeout-units?expand=1#diff-2f15328ea07078b2109edd1d4e288c82R111) and [here](https://github.com/bazelbuild/bazel/compare/master...bnczk:remote-timeout-units?expand=1#diff-2f15328ea07078b2109edd1d4e288c82R123).
As the netty channel in `HttpCacheClient` only accepts ints as millisecond timeout, this cast is necessary as `Duration` returns a long.
We could either 
- don't use `Duration`, only `int` directly (still exposes an overflow as the current implementation of `HttpCacheClient` multiplies the seconds by 1000 to get milliseconds)
- let the converter check if the `long` to `int` conversion throws an exception and throw an explicit `OptionsParsingException`
- allow it to fail during the cast

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -4,7 +4,7 @@ common --enable_platform_specific_config
 build:remote_shared --define=EXECUTOR=remote
 build:remote_shared --remote_instance_name=projects/bazel-untrusted/instances/default_instance
 build:remote_shared --remote_executor=grpcs://remotebuildexecution.googleapis.com
-build:remote_shared --remote_timeout=600s
+build:remote_shared --remote_timeout=600
 build:remote_shared --google_default_credentials
 build:remote_shared --jobs=100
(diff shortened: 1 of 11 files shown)


gregestren:
I'm not sure who has the current best expertise to do this review. I did some initial routing with the labels above. If no one steps up after some period of time please feel free to re-ping the thread. Either way thanks for this contribution!

meisterT:
cc @ulfjack

ulfjack:
Nice!

PR #11722, opened 2020-07-07
Let --remote_{cache,executor} support gRPC over UNIX sockets

EdSchouten:
All of the necessary code to construct gRPC clients that communicate
over UNIX sockets is there. It's just that we currently restrict its use
to the creation of proxies using --remote_proxy.

Let's shuffle some code around in GoogleAuthUtils to support the unix:
URL scheme directly. Move all of that code out of
newNettyChannelBuilder() into newUnixNettyChannelBuilder(), so that the
former can use it for both target and proxy channel creation.

UNIX sockets are useful, as they allow peers to authenticate using
light-weight features such as getsockopt(SO_PEERCRED).


buchgr:
LGTM

EdSchouten:
> LGTM. Add a test maybe?

I've added a test case for it to `src/test/shell/bazel/remote/remote_execution_test.sh`, inspired by the existing one called `test_remote_grpc_via_unix_socket()`.

buchgr:
Great. Merging it now.

PR #11723, opened 2020-07-07
Allow extra env vars on OSX toolchain config

illicitonion:
Just like we allow `tool_paths_overrides`.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -4404,7 +4404,7 @@ def _impl(ctx):
                         key = "APPLE_SDK_PLATFORM",
                         value = "%{apple_sdk_platform_value}",
                     ),
-                ],
+                ] + [env_entry(key = key, value = value) for key, value in ctx.attr.extra_env.items()],
             ),
         ],
     )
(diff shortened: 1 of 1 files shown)


oquenchil:
Hey Tony, do you think this is a good idea?

keith:
What's the use case for this?

illicitonion:
I'm using this to set `DEVELOPER_DIR` and `SDKROOT` for a remote environment with Xcode installed.

thii:
This doesn't seem right for that use-case. There's no guarantee that the same Xcode version is installed at the same location on the remote machine (though technically you can control that with your own infrastructure).

I would suggest to use an execution wrapper, and figure out the right DEVELOPER_DIR and SDKROOT for each action via the wrapper (there's already enough info to do that).

ulfjack:
Sorry for the plug: our remote execution service has built-in support to set `DEVELOPER_DIR` and `SDKROOT` from the `XCODE_VERSION_OVERRIDE`, `APPLE_SDK_VERSION_OVERRIDE`, and `APPLE_SDK_PLATFORM` env variables. I'm not sure that's the right way to go either, but it is probably faster than a wrapper and more resilient to different fs layouts than setting the env variables on the client?

ulfjack:
(To be clear, I'm not against this change, but it should be clear that setting env variables with absolute paths can go wrong depending on the service setup.)

PR #11725, opened 2020-07-07
Add toolchain_type and toolchain definitions for Android SDK.

katre:


tools/android/BUILD.tools
```diff
@@ -3,6 +3,9 @@ load("//tools/python:private/version.bzl", "PY_BINARY_VERSION")
 
 package(default_visibility = ["//visibility:public"])
 
+# The toolchain type used to distinguish Android SDK toolchains.
+toolchain_type(name = "sdk_toolchain_type")
```
juliexxia, line 7: Why not use the regular toolchain_type naming convention? Maybe worth leaving a note explaining so this isn't copied without understanding? 

Is it because this file isn't clearly for just sdk logic so the toolchain type purpose isn't obvious from the target path? That would make sense to me. i guess probably not worth creating an sdk subdir for it?

tools/android/android_sdk_repository_template.bzl
```diff
@@ -159,6 +160,16 @@ def create_android_sdk_rules(
...
         )
 
+        native.toolchain(
+            name = "sdk-%d-toolchain" % api_level,
+            toolchain_type = "@bazel_tools//tools/android:sdk_toolchain_type",
+            exec_compatible_with = HOST_CONSTRAINTS,
+            target_compatible_with = [
+                "@bazel_tools//platforms:android",
+            ],
+            toolchain = ":sdk-%d" % api_level,
```
juliexxia, line 170: Will the toolchains also live in that same BUILD.tools ?

tools/android/BUILD.tools
katre: I wanted to clarify that this is specifically for the SDK. I don't think we'll need an NDK toolchain type (that's really just a special case of the C++ toolchain type), but there's also various other things in Android that might be toolchains.

tools/android/android_sdk_repository_template.bzl
katre: No, this file is loaded from the BUILD file written by `android_sdk_repository`, so the toolchains are part of that repo.

src/test/java/com/google/devtools/build/lib/bazel/rules/android/AndroidSdkRepositoryTest.java
```diff
@@ -38,6 +38,7 @@ public void setup() throws Exception {
     scratch.file(
         "bazel_tools_workspace/tools/android/android_sdk_repository_template.bzl",
         ResourceLoader.readFromResources("tools/android/android_sdk_repository_template.bzl"));
+    scratch.appendFile("WORKSPACE", "local_config_platform(name='local_config_platform')");
```
juliexxia, line 41: Why is this change needed?

tools/android/BUILD.tools
juliexxia: Yeah that makes sense to me. I'd say maybe leave a comment that this isn't a conventional toolchain type name (re: https://docs.bazel.build/versions/master/toolchains.html#writing-rules-that-use-toolchains) but up to you

katre: I don't think the comment is helpful, there are plenty of other examples to follow.

src/test/java/com/google/devtools/build/lib/bazel/rules/android/AndroidSdkRepositoryTest.java
katre: The androidsdk repo reads constraints.bzl from `@local_config_platform` (to determine the host constraints for the `exec_compatible_with`), so it needs to be defined.

PR #11728, opened 2020-07-08
Allow native repository rules to declare toolchains they want

katre:
registered.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
@@ -141,11 +141,15 @@ public class RuleClass {
 
   @AutoCodec
   static final Function<? super Rule, Map<String, Label>> NO_EXTERNAL_BINDINGS =
-      Functions.<Map<String, Label>>constant(ImmutableMap.<String, Label>of());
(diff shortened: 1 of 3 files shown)


juliexxia:
Do we expect different rule classes to want to declare toolchains with different functionality?

katre:
Only repository rules will want to register toolchains. The `register_toolchains` function just takes a string (which should be a valid target pattern, but that's checked later). There's no actual functionality possible in this stage.

juliexxia:
Right right that makes sense to me, I was just wondering why we need to pass a function to RuleClass instead of the toolchains themselves if we expect what we want to pull out of RuleClass to always be the same - if that makes sense?

katre:
Oh, the function is what allows the rule to specify the actual name. using the repository name.

See https://github.com/bazelbuild/bazel/commit/6ac6bef8a0da746c0a711f165e1e3170e7d95d30#diff-688f3bda2aede4d7ff4b19e9e61bb840R55 for the example of this being used.

juliexxia:
Oh oh I understand, it's android_respository_rule related functionality so we don't want to embed it into ruleclass. gotcha!

katre:
Any further comments?

Issue #3705, opened 2017-09-08
--test_tag_filters do not allow to select tests that contains flag A and B


damienmg:
In issue #3704, we were executing too many tests because of confusion:

`-a,-b` is going to look for all tests that have neither `a` or `b` tag whereas `a,b` will match all tests that have either `a` or `b` tag. There is no way to select all tests that matches both `a` and `b`.

The [documentation](https://docs.bazel.build/versions/master/user-manual.html#flag--test_tag_filters) talks about positive flag as "required tags" which leads to believe that we apply a AND whereas it is a OR. The preceding '+' is also confusing... If the '+' sign is not used anywhere, maybe we should make it a way to say always required?

sladyn98:
@jin @damienmg I guess this is a simple enough issue to begin with, any sort of pointers to begin
CC : @aiuto

aiuto:
You would being the same way as any new feature.
- propose the design for exactly how the semantics of test_tag_filters would change
- get some feedback and iterate until people agree it makes sense
- come up with a PR that implements it (with tests)

It sounds like the simple plan is to actually make + mean required, so +a,+b means both a and b must be present.

ehkloaj:
Here is a pull request that solves this, #11733.

This behavior is already supported by bazel when using the rule [test_suite](https://docs.bazel.build/versions/master/be/general.html#test_suite) which means that it was just a boolean that needed to be flipped. 

We decided to not use the syntax where +a,+b means that both _a_ and _b_ must be present, as suggested by @aiuto, since it would conflict with the way users are used to how --test_tag_filters is working. Also, there is not really any point in specifying tags both with and without + since the once without + wouldn't have any affect. For example a,+b would mean that only _b_ must be present and it does not matter if _a_ is present or not.  

Instead we decided to use a syntax where you add _&&_ as an extra tag since this allowed for an easy implementation.

**Example of usage**:
`bazel test //... --test_tag_filters="&&,tag1,tag2"`
This will only run the tests that have tag1 and tag2.

What do you think about this solution? Do you have a suggestion for a more elegant syntax?

alandonovan:
Please let's not add yet more alien notations to Blaze. Can we reuse Starlark, restricted to the following expressions:

  expr = IDENT   |   '(' expr ')'   |   expr AND expr   |   expr OR expr   |  NOT expr

Tokens and precedence are as defined by the spec. We can detect the old syntax by looking for a comma and continue to support it for a grace period.

[...]

PR #11733, opened 2020-07-09
AND support for --test_tag_filters and --build_tag_filters

ehkloaj:
AND logic for required tags is now supported for --test_tag_filters by
specifying '&&' as a tag.

Example:
bazel test //... --test_tag_filters="&&,tag1,tag2"

This will only run tests that have tag1 and tag2.

This is related to #3705

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
@@ -355,6 +355,11 @@ public final class TargetUtils {
         TestTargetUtils.sortTagsBySense(tagFilterList);
     final Collection<String> requiredTags = tagLists.first;
     final Collection<String> excludedTags = tagLists.second;
+
+    // If "&&" is provided as a tag we use AND logic.
+    // We have to remove "&&" from requiredTags since it should not be regarded as a regular tag.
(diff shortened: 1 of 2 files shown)


gregestren:
Thanks for putting up this PR. I'm giving this some attention now, and I'd like to have a bit of time to think through the consequences.

My main concern is trying to understand what the broadest amount of flexibility we practically want is (you can imagine others arbitrary boolean combinations) and make sure we don't evolve an inconsistent, piecemeal syntax over time.

We should also do our best to ensure consistency. This positive/negative selection UI exists in other places too (`--build_tag_filters` and even specifying raw targets at the command line).  We don't want to complicate Bazel as a whole by having different features support this in different ways.

So let me think this through some before iterating more. Thanks for starting this out concretely.

ehkloaj:
Thanks for looking at this @gregestren. 

My colleagues and I also discussed supporting arbitrary boolean combinations but decided not to implement that since we didn't have a use case where it was needed but I get that it can be useful for others and that it might be a better solution in general. 

This commit also works for `--build_tag_filters`, I should have been clear about that, sorry.

ehkloaj:
This functionality has been implemented in https://github.com/bazelbuild/bazel/pull/11900 so I will close this one now.

PR #11735, opened 2020-07-09
Add missing tests for object files support and a separate param for PIC static library in cc_import

agluszak:
A follow-up to #11696


tools/build_defs/cc/cc_import.bzl
```diff
@@ -200,16 +184,14 @@ def _cc_import_impl(ctx):
...
-
     return [CcInfo(compilation_context = compilation_context, linking_context = linking_context)]
 
 cc_import = rule(
     implementation = _cc_import_impl,
     attrs = {
         "hdrs": attr.label_list(allow_files = [".h"]),
-        "static_library": attr.label(allow_single_file = [".a", ".lib", ".pic.a"]),
+        "static_library": attr.label(allow_single_file = [".a", ".lib"]),
+        "pic_static_library": attr.label(allow_single_file = [".a", ".lib", ".pic.a"]),
```
oquenchil, line 194: These should be .pic.a and .pic.lib:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java;drc=8b856f5484f0117b2aebc302f849c2a15f273310;l=459

tools/build_defs/cc/tests/cc_import_test.bzl
```diff
@@ -104,10 +104,36 @@ def _cc_import_objects_archive_action_test_impl(ctx):
 
     asserts.true(env, found, "Archive creation action should be present if object files were given")
 
+    linker_inputs = target_under_test[CcInfo].linking_context.linker_inputs.to_list()
```
oquenchil, line 107: Can you add a space between each block of 2 lines with assignment + assert? I think that will make it easier to read.

Issue #11741, opened 2020-07-10
Cannot build Bazel from bazel-distfile.tar


katre:
Error is:
```
ERROR: /tmp/dist/third_party/BUILD:568:8: Executing genrule //third_party:filter_netty_dynamic_libs failed (Exit 15): bash failed: error executing command 
  (cd /tmp/bazel_Gs7t3gE3/out/execroot/io_bazel && \
  exec env - \
    PATH=/usr/local/google/home/jcater/.local/bin:/usr/local/google/home/jcater/bin:/usr/local/google/home/jcater/.local/bin:/usr/local/google/home/jcater/bin:/usr/local/google/home/jcater/.local/bin:/usr/local/google/home/jcater/bin:/usr/lib/google-golang/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cp third_party/netty_tcnative/netty-tcnative-boringssl-static-2.0.24.Final.jar bazel-out/k8-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar && if [ -n '\''*.jnilib *.dll'\'' ]; then zip -qd bazel-out/k8-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar *.jnilib *.dll; fi')
Execution platform: //:default_host_platform
zip I/O error: Permission denied
zip error: Could not create output file (bazel-out/k8-opt/bin/third_party/netty_tcnative/netty-tcnative-filtered.jar)
```

PR #11742, opened 2020-07-10
Fix filter_netty_dynamic_libs to handle the case when the original

katre:
source isn't writable.

This happens, for example, with the output from bazel-distfile.tar.

Part of #11741.


katre:
Because I haven't merged https://github.com/bazelbuild/bazel/pull/11743 yet, so this code isn't tested.

meteorcloudy:
I see, thanks!

PR #11752, opened 2020-07-11
Handle transition exceptions inside AspectFunction

segiddins:
Otherwise, if there are errors over a starlark transitition’d dependency, an illegal cast exception would be raised when attempting to cast to an InvalidConfigurationException

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
@@ -41,6 +41,7 @@ import com.google.devtools.build.lib.analysis.config.ConfigMatchingProvider;
 import com.google.devtools.build.lib.analysis.config.DependencyEvaluationException;
 import com.google.devtools.build.lib.analysis.config.InvalidConfigurationException;
 import com.google.devtools.build.lib.analysis.configuredtargets.MergedConfiguredTarget;
+import com.google.devtools.build.lib.analysis.starlark.StarlarkTransition.TransitionException;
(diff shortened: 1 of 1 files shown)


segiddins:
Looks like something's moved around since 3.2.0 (which is what I based the patch on), let me try to hunt that down...

segiddins:
It looks like the one failure on CI is due to the bazel server not starting, I don’t think it’s due to this change

segiddins:
This is passing again, ping on a review?

keith:
@jin do you know who the right person is for this?

jin:
Seeing as it's related to transitions, it's probably configurability.

gregestren:
Happy to review.

This overall looks good, but can you clarify the precise connection between aspects, configured targets, deps, and transitions that triggered this error? Was this a dependency of the aspect?

segiddins:
Unfortunately, I can't find the commit of my project that repro'd the issue :/

It was definitely something to do with a dependency of the aspect that was behind a transition was erroring in some way, and the error message once I rescued here made it clear that I had misconfigured.... something.

I know on July 10th I was running into:

```
java.lang.ClassCastException: class com.google.devtools.build.lib.analysis.skylark.StarlarkTransition$TransitionException cannot be cast to class com.google.devtools.build.lib.analysis.config.InvalidConfigurationException
```

but unfortunately I don't seem to have saved any other context, beyond this patch.

gregestren:
No worries. Just want to clarify the sometimes subtle connections between these various components. That basic description makes sense.

Issue #11473, opened 2020-05-24
Build event uploader is not robust to upload failures and does not respect no-cache tag


scele:
### Description of the problem / feature request:

The BEP code that uploads missing files to remote cache has problematic behavior in two cases:

1. Remote cache uploads that happen during bazel build are robust to upload failures: if a CAS blob upload fails even after retries, bazel will print a warning message but continue the build. However, the BEP uploader stage will also try to upload any missing blobs referenced in build events, and if the upload fails there, the whole build fails.
2. While `no-cache` tag works correctly in that an AC entry never gets written to the remote cache, the BEP uploader's "upload-referenced-blobs" behavior means that `no-cache` cannot be used to prevent writing the action outputs (= CAS blobs) to the remote cache. In particular, we would like to use `no-cache` tags for actions that generate large artifacts that change often, and therefore not upload them to remote cache.

`--nobuild_event_json_file_path_conversion` fixes these issues, but it has the downside that **none** of the entries in the BES report reference the files in the remote cache, even if they were uploaded there as part of the build process.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run a remote cache at grpc://localhost:8980 (I used bb-storage).

Create two targets that output a large file, and add a `no-cache` tag to the other one.
```
$ cat BUILD 
genrule(
   name = "bigfile",
   outs = ["bigfile.img"],
   cmd = "fallocate -l 1G \"$@\"",
)
genrule(
   name = "bigfile_nocache",
   outs = ["bigfile_nocache.img"],
   cmd = "fallocate -l 1G \"$@\"",
   tags = ["no-cache"],
)
```
We simulate upload failures by disabling retries and using a small timeout value. Without `--build_event_json_file` the build of `//:bigfile` passes with warnings:
```
$ bazel build //:bigfile --remote_cache=grpc://localhost:8980 --remote_instance_name=default --remote_retries=0 --remote_timeout=3
INFO: Invocation ID: 5a9ad001-0658-4bc6-8167-bdb9be6ecc3a
INFO: Analyzed target //:bigfile (5 packages loaded, 7 targets configured).
INFO: Found 1 target...
WARNING: Writing to Remote Cache:
BulkTransferException
Target //:bigfile up-to-date:
  bazel-bin/bigfile.img
INFO: Elapsed time: 6.079s, Critical Path: 5.93s
INFO: 1 process: 1 linux-sandbox.
[...]

PR #11771, opened 2020-07-14
Update `--remote_http_cache` to `--remote_cache`

angerson:
`bazel`'s help text refers only to `--remote_cache`, and this is the only instance of `--remote_http_cache` on this page. I tried a build earlier with `--remote_http_cache`, and nothing happened. Now I'm double-checking with `--remote_cache` but it certainly seems like this section is out-of-date.

### DIFF
--- a/site/docs/remote-caching.md
+++ b/site/docs/remote-caching.md
@@ -154,7 +154,7 @@ to/from your GCS bucket.
 
 4. Connect to Cloud Storage by adding the following flags to your Bazel command:
    * Pass the following URL to Bazel by using the flag:
-       `--remote_http_cache=https://storage.googleapis.com/bucket-name`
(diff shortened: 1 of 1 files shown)


meisterT:
I am surprised that this is not working since it is still defined as old flag name (https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java;l=80;drc=3e58cca5a72aa72ecfcaed6cee2393e59e284325) but it's a good change anyway.

angerson:
I think the error I was seeing is actually #11473 (I didn't notice because `WARNING: Writing to Remote Cache:
BulkTransferException` doesn't break anything, it just doesn't upload).

Issue #11756, opened 2020-07-13
Grpc update breaks with Bazel 3.4.0 in docker container


katre:
Error log:
https://buildkite.com/bazel/google-bazel-presubmit/builds/36688#7500893b-351a-4549-86ae-19803c73cbd1

katre:
Looks like something changed so bazel frontend is unable to communicate with the server inside the docker container for RBE autoconfig.

keith:
We see this too with our projects and RBE and 3.4.0:

```

Starting local Bazel server and connecting to it...
--
  | ... still trying to connect to local Bazel server after 10 seconds ...
  | ... still trying to connect to local Bazel server after 20 seconds ...
  | ... still trying to connect to local Bazel server after 30 seconds ...
  | ... still trying to connect to local Bazel server after 40 seconds ...
  | ... still trying to connect to local Bazel server after 50 seconds ...
  | ... still trying to connect to local Bazel server after 60 seconds ...
  | ... still trying to connect to local Bazel server after 70 seconds ...
  | ... still trying to connect to local Bazel server after 80 seconds ...
  | ... still trying to connect to local Bazel server after 90 seconds ...
  | ... still trying to connect to local Bazel server after 100 seconds ...
  | ... still trying to connect to local Bazel server after 110 seconds ...
  | FATAL: couldn't connect to server (2502) after 120 seconds.

```

katre:
Yes, we're trying to diagnose and get out a patch as fast as we can.

nathanhleung:
Same issue here in a Docker runner on CircleCI:

```yml
jobs:
  build:
    docker:
      - image: circleci/node:12.16.1
        environment:
          NODE_OPTIONS: --max_old_space_size=4096
    resource_class: medium+
```

```
Extracting Bazel installation...
 
Starting local Bazel server and connecting to it...
 
... still trying to connect to local Bazel server after 10 seconds ...
 
... still trying to connect to local Bazel server after 20 seconds ...
 
... still trying to connect to local Bazel server after 30 seconds ...
 
... still trying to connect to local Bazel server after 40 seconds ...
 
... still trying to connect to local Bazel server after 50 seconds ...
 
... still trying to connect to local Bazel server after 60 seconds ...
 
... still trying to connect to local Bazel server after 70 seconds ...
 
... still trying to connect to local Bazel server after 80 seconds ...
 
... still trying to connect to local Bazel server after 90 seconds ...
 
[...]

PR #11776, opened 2020-07-15
Check isIpV4StackPreferred before binding with ipv6 address

meteorcloudy:
This is a workaround for a netty bug https://github.com/netty/netty/issues/10402 that caused the rollback of upgrading grpc-java to 1.26.0 (https://github.com/bazelbuild/bazel/issues/11756)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
+++ b/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
@@ -453,7 +453,7 @@ public class GrpcServerImpl extends CommandServerGrpc.CommandServerImplBase impl
     InetSocketAddress address = new InetSocketAddress("[::1]", port);
     try {
       // TODO(bazel-team): Remove the following check after upgrading netty to a version with a fix
-      //  for https://github.com/netty/netty/issues/10402
+      //   for https://github.com/netty/netty/issues/10402
       if (NetUtil.isIpV4StackPreferred()) {
         throw new IOException("ipv4 is preferred on the system.");
       }
--- a/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
+++ b/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
@@ -60,6 +60,7 @@ import io.grpc.StatusRuntimeException;
 import io.grpc.netty.NettyServerBuilder;
 import io.grpc.stub.ServerCallStreamObserver;
 import io.grpc.stub.StreamObserver;
+import io.netty.util.NetUtil;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
@@ -451,6 +452,11 @@ public class GrpcServerImpl extends CommandServerGrpc.CommandServerImplBase impl
     // and if that fails, try again with IPv4.
     InetSocketAddress address = new InetSocketAddress("[::1]", port);
     try {
+      // TODO(bazel-team): Remove the following check after upgrading netty to a version with a fix
+      //  for https://github.com/netty/netty/issues/10402
+      if (NetUtil.isIpV4StackPreferred()) {
+        throw new IOException("ipv4 is preferred on the system.");
+      }
       server =
           NettyServerBuilder.forAddress(address).addService(this).directExecutor().build().start();
     } catch (IOException ipv6Exception) {
--- a/src/main/java/com/google/devtools/build/lib/server/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/server/BUILD
@@ -47,6 +47,7 @@ java_library(
         "//third_party:flogger",
         "//third_party:guava",
         "//third_party:jsr305",
+        "//third_party:netty",
         "//third_party/grpc:grpc-jar",
         "//third_party/protobuf:protobuf_java",
     ],


meteorcloudy:
I manually verified this workaround works for grpc-java 1.26.0 in a docker container, but I have no idea how to add a test with ipv6 disabled.. Any idea?

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
```diff
@@ -451,6 +452,11 @@ public void serve() throws AbruptExitException {
     // and if that fails, try again with IPv4.
     InetSocketAddress address = new InetSocketAddress("[::1]", port);
     try {
+      // TODO(bazel-team): Remove the following check after upgrading netty to a version with a fix
+      //  for https://github.com/netty/netty/issues/10402
+      if (NetUtil.isIpV4StackPreferred()) {
+        throw new IOException("ipv4 is preferred on the system.");
```
meisterT, line 458: Throwing an exception is a little bit awkward.

Consider rearranging the code, something like:
```
InetSocketAddress address = null;

try {
  if (!ipv4preferred) {
    address = new InetSocketAddress("[::1]", port);
    server = ....
  }
} catch (IOException e) {
  address = null;
}

if (address == null) {
  // insert the code from the catch block here
}
```

meteorcloudy: Hmm, that will cause the `ipv6Exception` to null when ipv4preferred is true, which is required for constructing the error message. 

Considering we probably could revert this when the bug is fixed in netty, can we keep the code simple like this?

meisterT: Fair point.

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
```diff
@@ -451,6 +452,11 @@ public void serve() throws AbruptExitException {
     // and if that fails, try again with IPv4.
     InetSocketAddress address = new InetSocketAddress("[::1]", port);
     try {
+      // TODO(bazel-team): Remove the following check after upgrading netty to a version with a fix
+      //  for https://github.com/netty/netty/issues/10402
```
meisterT, line 456: super-nit: double space

meteorcloudy: Done

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
meteorcloudy: Thanks!

benjaminp:
> I manually verified this workaround works for grpc-java 1.26.0 in a docker container, but I have no idea how to add a test with ipv6 disabled.. Any idea?

`sysctl net.ipv6.conf.all.disable_ipv6=0`

This is probably already set in your docker container, though, since docker's default network configuration is triggering the problem.

meteorcloudy:
Do you mean `sysctl net.ipv6.conf.all.disable_ipv6=1`? 
I'm not sure how to implement this in an integration test without affecting other tests.

We run docker with `--network host` for normal tests on CI. That's why this bug was only caught when some project tries to run RBE build and the bazel-toolchain tries to do a remote configuration in a docker container we don't have any control. The bazel-toolchain ignores all dev version of Bazel and falls back to an old release version of Bazel in downstream pipeline, therefore we only detected the issue after the release.

meteorcloudy:
Bad news, I might be doing something wrong previously. When I was verifying https://github.com/bazelbuild/bazel/pull/11778, but `NetUtil.isIpV4StackPreferred()` didn't return true in my docker container with ipv6 disabled.

benjaminp:
Oops, I think `!io.netty.channel.unix.Socket.isIPv6Preferred()` is the right one.

meteorcloudy:
Hmm, it failed with
```
Exception in thread "main" java.lang.UnsatisfiedLinkError: io.netty.channel.unix.LimitsStaticallyReferencedJniMethods.udsSunPathSize()I
        at io.netty.channel.unix.LimitsStaticallyReferencedJniMethods.udsSunPathSize(Native Method)
        at io.netty.channel.unix.Socket.<clinit>(Socket.java:49)
        at com.example.grpc.server.MyGrpcServer.main(MyGrpcServer.java:42)
```

Looks like `io.netty.channel.unix.Socket.isIPv6Preferred()` needs some JNI library which doesn't work in the Bazel build.

meteorcloudy:
Oh, It seems I have to load the JNI library via Epoll.isAvailable first.

But, unfortunately, `io.netty.channel.unix.Socket.isIPv6Preferred()` still returned `true`...

benjaminp:
You may also have to call `Socket.initialize()`?

benjaminp:
(I think the vendored netty is also pretty old, so maybe it should be updated.)

meteorcloudy:
Thanks, Benjamin! 
That was my mistake, `io.netty.channel.unix.Socket.isIPv6Preferred()` does return correct result (even without `Socket.initialize()`), I was confused it's result with `Epoll.isAvailable()`..

meteorcloudy:
And indeed, I have to upgrade the netty version first.

PR #11777, opened 2020-07-15
Close VanillaJavaBuilder before writing output, in persistent worker mode

### LINKED ISSUE #11796: Race condition in VanillaJavaBuilder
Under sufficiently high load when using VanillaJavaBuilder, we intermittently see this exception crashing the worker:

---8<---8<--- Start of log, file at …/bazel-workers/worker-1-Javac.log ---8<---8<---
java.nio.file.NoSuchFileException: bazel-out/k8-fastbuild/bin/build-plugin-showcase/testing-build-plugin-generated.srcjar
\tat sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
\tat sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
\tat sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
\tat sun.nio.fs.UnixPath.toRealPath(UnixPath.java:837)
\tat com.sun.nio.zipfs.ZipFileSystemProvider.removeFileSystem(ZipFileSystemProvider.java:322)
\tat com.sun.nio.zipfs.ZipFileSystem.close(ZipFileSystem.java:305)
\tat com.google.devtools.build.buildjar.VanillaJavaBuilder.close(VanillaJavaBuilder.java:348)
\tat com.google.devtools.build.buildjar.VanillaJavaBuilder.runPersistentWorker(VanillaJavaBuilder.java:114)
\tat com.google.devtools.build.buildjar.VanillaJavaBuilder.main(VanillaJavaBuilder.java:89)
---8<---8<--- End of log ---8<---8<---

Some debugging later has tracked this down to a race condition between VanillaJavaBuilder's cleanup functions and bazel's; the semantics are that once a worker has sent back the response to the bazel server, it can't assume any input files still exist, because bazel will start deleting them.

https://github.com/bazelbuild/bazel/pull/11777 fixes this

asuffield:
The VanillaJavaBuilder.close() function turns out to depend on the
input files still existing. In persistent worker mode, as soon as we
write the output, bazel will begin cleaning up the working
directory. We must therefore wait to write the output until after
calling .close()

Fixes #11796

### DIFF
--- a/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/VanillaJavaBuilder.java
+++ b/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/VanillaJavaBuilder.java
@@ -103,15 +103,20 @@ public class VanillaJavaBuilder implements Closeable {
         if (request == null) {
           break;
         }
+        WorkResponse response;
         try (VanillaJavaBuilder builder = new VanillaJavaBuilder()) {
           VanillaJavaBuilderResult result = builder.run(request.getArgumentsList());
-          WorkResponse response =
+          response =
               WorkResponse.newBuilder()
                   .setOutput(result.output())
                   .setExitCode(result.ok() ? 0 : 1)
                   .build();
-          response.writeDelimitedTo(System.out);
         }
+        /* As soon as we write the response, bazel will start cleaning
+         * up the working tree. The VanillaJavaBuilder must be fully
+         * closed at this point.
+         */
+        response.writeDelimitedTo(System.out);
         System.out.flush();
       } catch (IOException e) {
         e.printStackTrace();


larsrc-google:
Are there any issues connected with this change?

asuffield:
I haven't filed one - I can if you want.

larsrc-google:
That would be great - helps with future software archaeology.

asuffield:
There, debugging stashed in an issue

larsrc-google:
An update: I'm currently running benchmarks to see if this slows things down too much.

src/java_tools/buildjar/java/com/google/devtools/build/buildjar/VanillaJavaBuilder.java
```diff
@@ -103,15 +103,20 @@ private static int runPersistentWorker() {
         if (request == null) {
           break;
         }
+        WorkResponse response;
         try (VanillaJavaBuilder builder = new VanillaJavaBuilder()) {
           VanillaJavaBuilderResult result = builder.run(request.getArgumentsList());
-          WorkResponse response =
+          response =
```
jmmv, line 109: This seems a bit fragile / confusing to me. Wouldn't something like this also work and make it clearer that the try block is specifically for the VanillaJavaBuilder?

```java
VanillaJavaBuilderResult result;
try (...) {
  result = ...
}
WorkResponse response = ...;
```

asuffield: Sure, but I think you're commenting on the original code rather than the change.

Do you want to include this here or make it a follow-up?

larsrc-google: Let's make that a follow-up, get the actual bug fixed first.

jmmv: I'm not sure what you mean about the original code. If I click on "Files changed" in this PR, I'm seeing the final state of the change, and I commented on that. But... the GitHub UI always confuses me, so what am I missing?

Issue #11187, opened 2020-04-22
http_archive gets 406 error, can download using browser


Dig-Doug:
# Description of the problem:

When trying to download an archive using `http_archive` from a project on gitlab, I get a `406` error. However, I can download the archive using my browser.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I was able to reproduce with the code posted [here](https://github.com/fenollp/gitlab_bazel_406_on_download_and_extract).

```bash
cd /tmp
git clone https://github.com/fenollp/gitlab_bazel_406_on_download_and_extract.git
cd gitlab_bazel_406_on_download_and_extract 
bazel sync
INFO: Repository radamsa instantiated at:
  no stack (--record_rule_instantiation_callstack not enabled)
Repository rule http_archive defined at:
  /home/doug/.cache/bazel/_bazel_doug/220a71cbe7341f96c28e272f1720771c/external/bazel_tools/tools/build_defs/repo/http.bzl:336:16: in <toplevel>
WARNING: Download from https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2 failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 406 Not Acceptable
ERROR: An error occurred during the fetch of repository 'radamsa':
   java.io.IOException: Error downloading [https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2] to /home/doug/.cache/bazel/_bazel_doug/220a71cbe7341f96c28e272f1720771c/external/radamsa/radamsa-v0.6.tar.bz2: GET returned 406 Not Acceptable
ERROR: java.io.IOException: Error downloading [https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2] to /home/doug/.cache/bazel/_bazel_doug/220a71cbe7341f96c28e272f1720771c/external/radamsa/radamsa-v0.6.tar.bz2: GET returned 406 Not Acceptable
Loading: loading...
```

### What operating system are you running Bazel on?

Ubuntu 19.10

### What's the output of `bazel info release`?

```
INFO: Invocation ID: 389ae0d8-4b39-4f14-8503-6725a9892926
release 3.1.0
```

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```
[...]

PR #11789, opened 2020-07-16
Generate a valid Accept header in the request

### LINKED ISSUE #11788: Bazels HTTP Accept header doesn't follow RFC7231

Fetching resources using HTTP fail with servers strictly checking request headers. The HttpURLConnection of OpenJDK has a [known issue](https://bugs.openjdk.java.net/browse/JDK-8163921) and produces an Accept header that is rejected by some servers.

Linux and Windows

We first encountered this with 0.28.1 and 1.2 but the downloader/HttpConnector.java file seems unchanged since then.

zecke:
The OpenJDK generates an accept header that does not follow the BNF of the HTTP
RFC. Generate a conforming default header that can be overridden by the usual request
headers.

Fixes #11788.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpConnector.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpConnector.java
@@ -107,6 +107,8 @@ class HttpConnector {
       try {
         connection = (HttpURLConnection)
             url.openConnection(proxyHelper.createProxyIfNeeded(url));
+        // TODO(zecke): Revise once https://bugs.openjdk.java.net/browse/JDK-8163921 is fixed.
+        connection.addRequestProperty("Accept", "text/html, image/gif, image/jpeg, */*");
         boolean isAlreadyCompressed =
             COMPRESSED_EXTENSIONS.contains(HttpUtils.getExtension(url.getPath()))
                 || COMPRESSED_EXTENSIONS.contains(HttpUtils.getExtension(originalUrl.getPath()));


meisterT:
What request header was Bazel sending before?
Do we signal now that we accept more than before?

zecke:
What 

> What request header was Bazel sending before?

[JDK code](https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L256):
```
"text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2";
```

> Do we signal now that we accept more than before?

Good question. We don't accept more but signal different quality factors. For maximum compatibility but conforming to the HTTP RFC BNF we could use:

```
"text/html, image/gif, image/jpeg, */*; q=.2";
```

meisterT:
Let's play safe and use that then?

gyroninja:
This should also fix #11187 which is preventing people from downloading archives from GitLab.

I believe this issue is caused by failing a check in https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/hotlinking_detector.rb which parses and inspects the Accept header. I independently wrote a patch that set Accept to */* and GitLab permitted me to download the archive instead of returning a 406 error. I just tried setting it to the proposed "text/html, image/gif, image/jpeg, */*; q=.2" and GitLab gave me a 406. It would be great if we could also solve this issue with pulling down archives from GitLab while we are touching this part of the code base.

zecke:
> This should also fix #11187 which is preventing people from downloading archives from GitLab.
> 
> I believe this issue is caused by failing a check in https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/hotlinking_detector.rb which parses and inspects the Accept header. I independently wrote a patch that set Accept to `*/*` and GitLab permitted me to download the archive instead of returning a 406 error. I just tried setting it to the proposed `text/html, image/gif, image/jpeg, */*; q=.2` and GitLab gave me a 406. It would be great if we could also solve this issue with pulling down archives from GitLab while we are touching this part of the code base.

Would you have a URL to a public repository to try this? What about re-ordering and putting */*; q=.2 first? My issue seems to be fixed with the proposed accept header.

gyroninja:
Inside of #11187 there is setup for reproducing the issue.  The link it uses is https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2

I switched over to your branch and here are my results of setting the Accept header to different values.

`Accept: */*`: This works.

`Accept: text/html, image/gif, image/jpeg, */*; q=.2`: 406

`Accept: q=.2; text/html, image/gif, image/jpeg, */*`: 406

zecke:
> Inside of #11187 there is setup for reproducing the issue. The link it uses is https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2
> 
> I switched over to your branch and here are my results of setting the Accept header to different values.
> 
> `Accept: */*`: This works.
> 
> `Accept: text/html, image/gif, image/jpeg, */*; q=.2`: 406
> 
> `Accept: q=.2; text/html, image/gif, image/jpeg, */*`: 406

The last header is not correct. It should be `Accept: */*; q=.2...` but that triggers a 406 as well. For the GitLab case we need the Accept: */*.


```
$ curl -H "Accept: */*; q=.2, text/html, image/gif, image/jpeg" -w "%{http_code}\n" -s -o /dev/null https://gitlab.com/akihe/radamsa/-/archive/v0.6/radamsa-v0.6.tar.bz2          
406
```

gyroninja:
What about `Accept: text/html, image/gif, image/jpeg, */*`?

I could be wrong but isn't setting the quality factor factor redundant? Since it the least specific shouldn't it have the least priority anyways?

>   Media ranges can be overridden by more specific media ranges or
>   specific media types.  If more than one media range applies to a
>   given type, the most specific reference has precedence.

https://tools.ietf.org/html/rfc7231#page-39

ulfjack:
Yuk! Let's get this merged, @meisterT!

emacski:
Just as an FYI as to why the header was still failing with the "q=.2"; the "qvalue" needs to include "0" or "1" depending on the values after the decimal based on the http1.1 spec: https://tools.ietf.org/html/rfc7231#section-5.3.1

so should be: "q=0.2"

gyroninja:
@emacski 
Good catch. In regards to your suggestions in my personal patch for #11187 I did exactly that as to unify the Accept headers that Bazel sends. When using a remote cache Bazel currently uses `Accept: */*`.
 https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java#L167
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/http/HttpUploadHandler.java#L135

Another approach I thought of was changing it to use Chrome's default Accept header as to look more like a browser.

PR #11793, opened 2020-07-16
Add distrib_cc_binary rule

meteorcloudy:
This is for using Debian's `/usr/bin/grpc_java_plugin` in Bazel Debian build.

Working towards https://github.com/bazelbuild/bazel/issues/9408


meteorcloudy:
FYI @olekw

PR #11799, opened 2020-07-17
Remove old incompatible option --incompatible_symlinked_sandbox_expands_tree_artifacts_in_runfiles_tree.

benjaminp:
This one was flipped more than a year ago.

RELNOTES: Remove old incompatible flag --incompatible_symlinked_sandbox_expands_tree_artifacts_in_runfiles_tree.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/sandbox/DarwinSandboxedSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/DarwinSandboxedSpawnRunner.java
@@ -232,11 +232,7 @@ final class DarwinSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
 
     SandboxInputs inputs =
         helpers.processInputFiles(
-            context.getInputMapping(
-                getSandboxOptions().symlinkedSandboxExpandsTreeArtifactsInRunfilesTree),
-            spawn,
-            context.getArtifactExpander(),
-            execRoot);
+            context.getInputMapping(true), spawn, context.getArtifactExpander(), execRoot);
     SandboxOutputs outputs = helpers.getOutputs(spawn);
 
     final Path sandboxConfigPath = sandboxPath.getRelative("sandbox.sb");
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/devtools/build/lib/sandbox/DarwinSandboxedSpawnRunner.java
```diff
@@ -232,11 +232,7 @@ protected SandboxedSpawn prepareSpawn(Spawn spawn, SpawnExecutionContext context
 
     SandboxInputs inputs =
         helpers.processInputFiles(
-            context.getInputMapping(
-                getSandboxOptions().symlinkedSandboxExpandsTreeArtifactsInRunfilesTree),
-            spawn,
-            context.getArtifactExpander(),
-            execRoot);
+            context.getInputMapping(true), spawn, context.getArtifactExpander(), execRoot);
```
meisterT, line 235: Do we call `getInputMapping` with `false` anywhere?

benjaminp: Not as far as I can see in Bazel. I was hoping to follow up with a CL removing that parameter if someone could assure me that Google is not passing `false`.

meisterT: Ok, let's do that in a separate CL then.

meisterT: https://bazel-review.googlesource.com/c/bazel/+/137717

benjaminp: Thanks—I also raced and submitted #11809.

meisterT: Sorry for the duplicate work, I thought it's easier for me to do with the internal changes that were necessary.

benjaminp: No problem; it was almost completely mechanical anyway.

PR #11800, opened 2020-07-17
make curl fail loudly for non-2xx downloads

ashi009:
`curl` fails silently if the status code of the response is not 2xx. This CL makes it fail loudly, and break the pipe.

### DIFF
--- a/scripts/docs/Dockerfile
+++ b/scripts/docs/Dockerfile
@@ -7,7 +7,7 @@
 # $ docker run -it --rm -p 8000:8000 --entrypoint /bin/bash bazel-jekyll
 # $ git clone https://bazel.googlesource.com/bazel
 # $ cd bazel
-# $ curl -o bazel https://releases.bazel.build/0.19.0/release/bazel-0.19.0-linux-x86_64
+# $ curl -f -o bazel https://releases.bazel.build/0.19.0/release/bazel-0.19.0-linux-x86_64
 # $ chmod +x bazel
 # $ ./bazel build //site
 # $ cd bazel-bin/site/site-build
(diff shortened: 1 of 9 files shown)


meisterT:
Let's use the short option form as we do with the other options.

ashi009:
> Let's use the short option form as we do with the other options.

Done.

ashi009:
DO NOT SUBMIT, I'll do another sweep over the codebase.

ashi009:
> DO NOT SUBMIT, I'll do another sweep over the codebase.

PTAL, if this goes too far.

scripts/docs/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN apt-get -qqy update && \
 
 RUN ln -fs /usr/bin/python3.7 /usr/bin/python
 
-RUN curl -Lo /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v1.3.0/bazelisk-linux-amd64 && \
+RUN curl -f -Lo /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/download/v1.3.0/bazelisk-linux-amd64 && \
```
meisterT, line 28: in general you can combine these short binary options e.g. `curl -fLo /usr/local/bin/bazel ...`

src/test/shell/bazel/remote_helpers.sh
```diff
@@ -185,7 +185,7 @@ function serve_timeout() {
 # connections, which causes flakes.
 function wait_for_server_startup() {
   touch some-file
-  while ! curl http://localhost:$fileserver_port/some-file > /dev/null; do
+  while ! curl -f http://localhost:$fileserver_port/some-file > /dev/null; do
```
meisterT, line 188: I don't think we want it here.

scripts/docs/Dockerfile
ashi009: Done.

src/test/shell/bazel/remote_helpers.sh
ashi009: Done.

PR #11805, opened 2020-07-18
Close fewer tags.

w-hat:


laurentlb:
As I learned recently, `</li>` tags are not needed.

https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags

meisterT:
See also https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#examples

vladmos:
Even if `<li>` tags shouldn't be closed, there are wrong `</p>` tags without an opening counterpart.

PR #11809, opened 2020-07-20
Make SpawnInputExpander always expand tree artifacts.

benjaminp:


benjaminp:
https://cr.bazel.build/137717

Issue #11817, opened 2020-07-22
Absolute paths embedded in dependency files


thii:
### Description of the problem / feature request:

When compiling `objc_library` targets, we're seeing absolute paths of generated files being embedded in dependency files (`.d` files).

### Feature requests: what underlying problem are you trying to solve with this feature?

We're trying to make the dependency files cacheable by removing or normalizing the absolute paths that are embedded in the dependency files.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please check the attached project: [nondeterministic-depfile.zip](https://github.com/bazelbuild/bazel/files/4957578/nondeterministic-depfile.zip).

* Run `bazel build //:B`
* Check the `B.d` file contents. It contains an absolute path to a file that is generated during the build (`  /private/var/tmp/_bazel_user/b4422045cf7215be2d3fa4e439c7376a/sandbox/darwin-sandbox/214/execroot/__main__/bazel-out/apl-darwin_x86_64-fastbuild/bin/AModule-module.modulemap`)

The attached project uses some custom rules but the issue can be reproduced with an `objc_library` (that has modules enabled) depending on a `swift_library` and importing it via a Clang module.

### What operating system are you running Bazel on?

macOS Catalina 10.15.5 (19F101)

### What's the output of `bazel info release`?

release 3.4.1

###  Have you found anything relevant by searching the web?

Found a related issue with system headers: https://github.com/bazelbuild/bazel/issues/7772. We found an additional issue with generated files during a build.

Found two issues with dependency files: https://github.com/bazelbuild/bazel/issues/7769, https://github.com/bazelbuild/bazel/pull/7771. The `dependency_file` feature can't be disabled because it's necessary to keep the C++ builds correct.

keith:
Does using `--experimental_inmemory_dotd_files` work around this?

thii:
Tried but it didn’t change anything—the produced .d files are still part of the outputs and are still being cached as is. 

From the documentation of the option, it looks like it’s specific for Remote BuildS w/o the Bytes feature.

https://github.com/bazelbuild/bazel/blob/82e6b6a3b9589b93f4abcf3eb0f14cd1d50008ff/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java#L689-L690

trybka:
I think non-determinism in .d files is harmless? 
[...]

PR #11818, opened 2020-07-22
Make dependency files deterministic

thii:
The dependency file generated by Clang can sometimes contain absolute
paths to system headers, system modules and files generated by Bazel
during the build.  These paths can defer between machines and between
Bazel instances, for example:

```
bazel-out/apl-darwin_x86_64-fastbuild/bin/_objs/B/arc/B.o: B.m \
  /private/var/tmp/_bazel_user/b4422045cf7215be2d3fa4e439c7376a/sandbox/darwin-sandbox/214/execroot/__main__/bazel-out/apl-darwin_x86_64-fastbuild/bin/A.modulemap \
  /Applications/Xcode-11.6.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/Foundation.framework/Modules/module.modulemap
```

This patch adds a post-process to wrapped_clang to normalize those paths
by:

* converting absolute paths to relative paths for paths to files under
the execution root
* standardizing the paths to system headers, system modules to always
use the standard location of Xcode (`/Applications/Xcode.app`).  The
user doesn't need to have Xcode installed to this location as Bazel are
already mapping them to the correct location locally.

This makes the produced `.d` files deterministic and can reliably be
shared with the remote cache.

Fixes https://github.com/bazelbuild/bazel/issues/11817.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5572,7 +5572,11 @@ def _impl(ctx):
                 ],
                 flag_groups = [
                     flag_group(
-                        flags = ["-MD", "-MF", "%{dependency_file}"],
+                        flags = [
+                            "-MD",
+                            "-MF",
+                            "DEPENDENCY_FILE=%{dependency_file}",
+                        ],
                         expand_if_available = "dependency_file",
                     ),
                 ],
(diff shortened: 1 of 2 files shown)


keith:
So it doesn't matter when downloading these files if these paths are valid? Maybe we should remove the Xcode prefix entirely?

thii:
Removing the Xcode prefix breaks the builds. It seems Bazel needs it for remapping.

thii:
@trybka Hi. Could you take a look or request a review for this?

jin:
Hi @trybka, @googlewalt and/or @bbarenblat, could you take a brief look and share if this change makes sense?

trybka:
See comment here: https://github.com/bazelbuild/bazel/issues/11817#issuecomment-716592082

Notably, if these are harmless, it seems wasteful to clean them up in this manner.

PR #11821, opened 2020-07-22
Update desugar_jdk_libs

menny:
taken latest from https://github.com/google/desugar_jdk_libs


donaldchai:
This is new to me, but based on a964125af6fea06a5b3414a978d2af4186c4e357 it seems like you need to add any new classes to `desugar.sh`?  https://github.com/google/desugar_jdk_libs/commit/ab1edbaa978a88a3f6c1a83cb37a3a3ab840f472 desugars a new class not already listed.

menny:
oh.. this is a hard one to complete. I took a look at what Android Studio source code is specifying but it seems pretty much what you have right now here at HEAD.

menny:
I looked at https://android.googlesource.com/platform/tools/base/+/studio-master-dev/lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/LibraryDesugaring.kt

menny:
I [compared](https://github.com/google/desugar_jdk_libs/compare/e0b0291b2c51fbe5a7cfa14473a1ae850f94f021...master) the commit currently in bazel to HEAD at the desugar repo  and looking at the [retarget](https://github.com/bazelbuild/bazel/blob/master/src/tools/android/java/com/google/devtools/build/android/desugar/retarget/retarget_config.textproto) details, and it seems that there aren't any new things that need to be added explicitly to the shell script.

aiuto:
ping.
Also, if you wait about a week, I'll have completed an overhaul to how we specify deps like this, so the PR will be entirely different, but much smaller (look at //distdir_deps.bzl)

menny:
@aiuto I don't mind waiting, but since this is a pretty small PR anyway I think it's okay as is. Your call.

aiuto:
I'm on triage duty this week, so I'll merge it. But first the tests have to pass.

aiuto:
See #12811.   That does this in the new style.  The tests should work now too, since I updated the mirror.

aiuto:
Sorry this took so long.

PR #11824, opened 2020-07-23
Add deps attribute to Starlark version of cc_import

### LINKED ISSUE #5456: Add a "deps" attribute to cc_import

This is a feature request to add a `deps` attribute to `cc_import`.

Some prebuilt libraries have public dependencies on other libraries, by using these libraries' public headers in their own public headers. There currently does not seem to be a way to express this with `cc_import` which is the preferred way to include prebuilts.

There's a way to work around this by declaring a `cc_library` which both depends on the `cc_import` and on its public dependencies, and direct clients to depend on the `cc_library`.

agluszak:
Fixes #5456

### DIFF
--- a/tools/build_defs/cc/cc_import.bzl
+++ b/tools/build_defs/cc/cc_import.bzl
@@ -191,7 +191,7 @@ def _cc_import_impl(ctx):
         cc_infos.append(dep[CcInfo])
     merged_cc_info = cc_common.merge_cc_infos(cc_infos = cc_infos)
 
-    return merged_cc_info
+    return [merged_cc_info]
 
 cc_import = rule(
     implementation = _cc_import_impl,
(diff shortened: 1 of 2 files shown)


tools/build_defs/cc/cc_import.bzl
```diff
@@ -184,7 +184,14 @@ def _cc_import_impl(ctx):
...
 
-    return [CcInfo(compilation_context = compilation_context, linking_context = linking_context)]
+    this_cc_info = CcInfo(compilation_context = compilation_context, linking_context = linking_context)
+    cc_infos = [this_cc_info]
+
+    for dep in ctx.attr.deps:
+        cc_infos.append(dep[CcInfo])
+    merged_cc_info = cc_common.merge_cc_infos(cc_infos = cc_infos)
+
+    return merged_cc_info
```
c-mita, line 194: I'm not sure if it's required, but we tend to return single providers as a list (so `[merged_cc_info]`).

Issue #11831, opened 2020-07-24
Slow performance when resolving symbolic links


ehrtro:
### Description of the problem / change request:
When building with Bazel, our code can currently be efficiently built with bazel caches only if we're getting cache hits in our workspace cache and have "warm" bazel Java processes running. Otherwise (e.g. if the cache hit is from disk-cache or remote-http-cache, or if running in batch mode) the build is very slow, especially if it involves many .so files. This is due to that we're creating unnecessarily many symlinks to the same files. As mentioned in https://github.com/bazelbuild/bazel/commit/fca63525158321e5a5f0ff67a8796992a7347c73, the usage of --experimental_multi_threaded_digest enhances the performance somewhat, but we're still experiencing unnecessarily long build times. We suspect there is something else problematic with SolibSymlink target. Our local patch resolves the symlinks before hashing the files in order to shorten the build times:

Our changes to [these lines in manuallyComputeDigest](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/actions/cache/DigestUtils.java#L244-L253) is to be found below 

```
    // Attempt a cache lookup if the cache is enabled.
    Cache<CacheKey, byte[]> cache = globalCache;
    CacheKey key = null;
    if (cache != null) {
      Path canonicalPath = path.resolveSymbolicLinks();
      key = new CacheKey(canonicalPath, canonicalPath.stat());
      digest = cache.getIfPresent(key);
      if (digest != null) {
        return digest;
      }
    }
```
What's your opinion regarding this patch in order to get a shorter build time? When using this patch in our local environment we've experienced a performance enhancement; A build that previously took 17 seconds to finish is now done in 8 seconds. Using this patch and  --experimental_multi_threaded_digest at the same time gives us the best result. 

### What operating system are you running Bazel on?
Red Hat Enterprise 7 (RHEL7) Linux kernel 3.10.0-1127.8.2.el7.x86_64

ehkloaj:
Here is a pull request for this issue, #11832. 
Do you think that this functionality should be introduced together with a --experimental_... flag?

meisterT:
Should this be closed now that the PR is merged?

ehrtro:
Yes, we're satisfied with the solution! Thanks for fast response!

joeleba:
Due to https://github.com/bazelbuild/bazel/issues/12225, we'll roll back 81245fd to minimize the impact and will perhaps get back to it later.

PR #11832, opened 2020-07-24
Reduce number of digest calculations

ehkloaj:
Only calculate digest once per file, even if there are multiple symbolic links pointing to the same file.
This reduce build times when the digests are not already available in memory, e.g. if the bazel server process has been restarted.

Related to #11831

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/actions/DigestUtilsTest.java
+++ b/src/test/java/com/google/devtools/build/lib/actions/DigestUtilsTest.java
@@ -214,6 +214,50 @@ public class DigestUtilsTest {
   }
 
   @Test
+  public void testCacheWithSymlinks_cacheHit() throws Exception {
+    final AtomicInteger getFastDigestCounter = new AtomicInteger(0);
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/devtools/build/lib/actions/DigestUtilsTest.java
```diff
@@ -213,6 +213,50 @@ public void testCache() throws Exception {
     assertThat(digest3).isEqualTo(digest1);
   }
 
+  @Test
+  public void testCacheWithSymlinks() throws Exception {
```
joeleba, line 217: nit: Test names should include the condition being tested and the expected outcome.

So I'd suggest something like `testCacheWithSymlinks_cacheHit`.

ehkloaj: Thanks for looking at this @joeleba . I'll change this right away.

joeleba:
I'm performing some internal checks. Please allow some time for this to be merged.

Issue #10886, opened 2020-03-02
macOS wrapped_clang should export ZERO_AR_DATE=1 to make linking hermetic


brentleyjones:
### Description of the problem / feature request:

`wrapped_clang`, unlike `bazel/libtool`, doesn't export `ZERO_AR_DATE=1`. This causes timestamps to be embedded in linked binaries.

### Feature requests: what underlying problem are you trying to solve with this feature?

`wrapped_clang` should export `ZERO_AR_DATE=1`.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

On macOS, using the Xcode toolchain:

```
bazel clean
bazel build :some_objc_target
cp bazel-bin/some_objc_target /tmp/some_objc_target1
bazel clean
bazel build :some_objc_target
cp bazel-bin/some_objc_target /tmp/some_objc_target2
cmp /tmp/some_objc_target1 /tmp/some_objc_target2
```

The binaries will report differences because of the embedded timestamps.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 2.1.0

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/rules_swift/pull/136
https://github.com/bazelbuild/rules_swift/pull/120
https://github.com/bazelbuild/bazel/blob/0e1100f4a31c07ddf5efe3d4820032e0512cb91e/tools/objc/libtool.sh#L115

### Any other information, logs, or outputs that you want to share?

http://blog.llvm.org/2019/11/deterministic-builds-with-clang-and-lld.html

The above blog mentions setting this value to remove the timestamps.

brentleyjones:
A workaround for now is to call the build with `--action_env=ZERO_AR_DATE=1`

thii:
`ZERO_AR_DATE=1` is a subset of `-D`, so it looks like passing `-D` is the preferred way now. There's a PR for it: https://github.com/bazelbuild/bazel/pull/11576.

`ZERO_AR_DATE=1` also has [a bug](https://source.chromium.org/chromium/chromium/src/+/62848c8d298690e086e49a9832278ff56b6976b5?originalUrl=https%2F%2F:%2F%2F%2F%2Fcs.chromium.org%2F) with timezone.

> The problem with this flag is that it resets the file mtime on the file to epoch=0, e.g. 1970-1-1 or 1969-12-31 depending on timezone.

brentleyjones:
[...]

PR #11840, opened 2020-07-27
Support deterministic mode when using a version of libtool that supports it.

thii:
Xcode 10.2 adds a new flag `-D` to libtool that sets archive contents'
user ids, group ids, dates, and file modes to reasonable defaults.

This patch adds this new flag to all libtool actions to when they're
built with a version of libtool that supports this mode, to ensure
hemertic results for linking actions. Note that this new flag covers the
functionality of the `ZERO_AR_DATE` environment variable, but this
variable is still being left in the libtool wrapper for
backward-compatibility with older versions.

Fixes https://github.com/bazelbuild/bazel/issues/10886.

RENOTES: None.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -30,6 +30,30 @@ load(
 )
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 
+# The Xcode version from which that has support for deterministic mode
+_SUPPORTS_DETERMINISTIC_MODE = "10.2"
+
+def _compare_versions(dv1, v2):
+    """Return value is <0, 0, >0 depending on DottedVersion dv1 comparison to string v2."""
+    return dv1.compare_to(apple_common.dotted_version(v2))
+
+def _can_use_deterministic_libtool(ctx):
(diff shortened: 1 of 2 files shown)


keith:
Related https://github.com/bazelbuild/bazel/pull/11576

thii:
Thanks @keith. This actually uses the suggestion from @trybka from that PR.

jin:
Assigning @oquenchil, who self-assigned in the #10886.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -30,6 +30,30 @@ load(
 )
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 
+# The Xcode version from which that has support for deterministic mode
+_SUPPORTS_DETERMINISTIC_MODE = "10.2"
+
+def _compare_versions(dv1, v2):
+    """Returns value is <0, 0, >0 depending on DottedVersion dv1 comparison to string v2."""
```
trybka, line 37: nit: "Return value is..." or "Returns ..."

thii: Fixed. Thanks.

PR #11849, opened 2020-07-28
Update RPBB prebuilt to include cl/323460004

donaldchai:


Wyverald:
FYI this broke rules_jvm_external on RBE Ubuntu 16.04: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1585#4987f73c-08a6-4830-be8c-377ed90bcc3e

donaldchai:
Looks like the presubmit tests didn't catch the fact that this JAR needs to be compiled against JDK8.  Sent https://github.com/bazelbuild/bazel/pull/11888 to rollforward.

PR #11851, opened 2020-07-28
fix int overflow

xingxinghuo1000:
When a UT test case run very long time. Lines that covered a lot of times. Int 32 could overflow. And in my real project, error occured

Recorded in:
https://github.com/bazelbuild/bazel/issues/9406#issuecomment-664211209


xingxinghuo1000:
@googlebot I signed it!

xingxinghuo1000:
@googlebot I signed it!

PR #11866, opened 2020-07-29
Allow string unsafe test to pass on s390x

ruixin-bao:
Hello,

This test fails on s390x (a big endian platform) due to comparison to a `StandardCharsets.UTF_16LE` encoded string. Changing that to `StandardCharsets.UTF_16BE` on s390x could let the test pass. This test now passes on both intel and s390x.

Thanks.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/unsafe/StringUnsafeTest.java
+++ b/src/test/java/com/google/devtools/build/lib/unsafe/StringUnsafeTest.java
@@ -20,6 +20,7 @@ import java.nio.charset.StandardCharsets;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import java.nio.ByteOrder;
 
 /** Tests for {@link StringUnsafe}. */
 @RunWith(JUnit4.class)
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/devtools/build/lib/unsafe/StringUnsafeTest.java
```diff
@@ -44,8 +44,13 @@ public void testGetBytes() {
     StringUnsafe stringUnsafe = StringUnsafe.getInstance();
     assertThat(ByteBuffer.wrap(stringUnsafe.getByteArray("hello")))
         .isEqualTo(StandardCharsets.ISO_8859_1.encode("hello"));
-    assertThat(ByteBuffer.wrap(stringUnsafe.getByteArray("lambda λ")))
-        .isEqualTo(StandardCharsets.UTF_16LE.encode("lambda λ"));
+
+    if (System.getProperty("os.arch").equals("s390x"))
```
philwo, line 48: Instead of hardcoding the s390x arch here, could we instead use e.g.

```java
import java.nio.ByteOrder;

if (ByteOrder.nativeOrder().equals(ByteOrder.BIG_ENDIAN)) {
  ...
}
```

?

philwo (changes requested):

src/test/java/com/google/devtools/build/lib/unsafe/StringUnsafeTest.java
ruixin-bao: Thanks for reviewing. Yes, I will try this change and upload a new patch once I have results on s390x.

ruixin-bao: Updated this PR and commit message, this test still passes on s390x.

ruixin-bao:
Thank you for merging!

PR #11872, opened 2020-07-30
NDK 21 support

### LINKED ISSUE #10811: Add support for Android NDK r21

Add support for Android NDK r21, which is the latest stable NDK.
https://developer.android.com/ndk/downloads

Among other things, this allow to use the Khronos unified Vulkan validation layer, which ships with NDK r21.

Try to compile any Android project with NDK r21, you get the following warning:

```
WARNING: The major revision of the Android NDK referenced by
android_ndk_repository rule 'androidndk' is 21. The major revisions
supported by Bazel are [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20].
Bazel will attempt to treat the NDK as if it was r20. This may cause
compilation and linkage problems. Please download a supported
NDK version.
```

Linux

release 2.0.0

Given that Android NDK r21 support is nowhere to be seen on https://blog.bazel.build/2020/02/06/bazel-2.1.html, I guess it is not supported yet.

### LINKED ISSUE #11084: Support for NDK 21, API level 29

Bazel does not support NDK 21 or, more importantly, API level 29 with the NDK.

I want to build stuff that uses API level 29...

Linux, Fedora Silverblue 32.

release 3.0.0

No.

I'm guessing that [this would get an extra key added](https://github.com/bazelbuild/bazel/blob/3.0.0/src/main/java/com/google/devtools/build/lib/bazel/rules/android/ndkcrosstools/r19/ApiLevelR19.java#L55)? Also, it would be nice to have API 29 in 3.1...

steeve:
This change includes support for Android NDK r21, largely based on the
support of NDK r19.

Closes #10811
Closes #11084

Signed-off-by: Steeve Morin <steeve@zen.ly>

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/ndkcrosstools/AndroidNdkCrosstools.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/ndkcrosstools/AndroidNdkCrosstools.java
@@ -53,6 +53,7 @@ public final class AndroidNdkCrosstools {
           .put(18, new NdkMajorRevisionR18("7.0.2"))
           .put(19, new NdkMajorRevisionR19("8.0.2"))
           .put(20, new NdkMajorRevisionR19("8.0.7")) // no changes relevant to Bazel
+          .put(21, new NdkMajorRevisionR19("9.0.8")) // no changes relevant to Bazel
           .build();
 
   public static final Map.Entry<Integer, NdkMajorRevision> LATEST_KNOWN_REVISION =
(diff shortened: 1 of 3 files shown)


ahumesky:
Thank you Steeve!

ahumesky:
I've imported the PR internally, should have it submitted in a day or two.

steeve:
Great! While we're here, would you have more info about rules_android by chance? Thanks!

EwoutH:
Hi! When NDK 21 was added, it only added support up to API Level 29. Could API Level 30 also be supported? It's mandatory for new apps on Google Play starting August 1, 2021 and currently NDK 21 is still the latest LTS release.

I've opened an issue here: #13421

Issue #9189, opened 2019-08-16
remote_jdk11 doesn't include required security policy configs


AngusDavis:
### Description of the problem / feature request:

When running with --config=docker-sandbox or --config=remote and using a --javabase/--host_javabase of @bazel_tools//tools/jdk:remote_jdk11, attempting to initialize SSL (e.g., from tests) fails with the following:
```
Time: 0.043
There was 1 failure:
1) testSslInit(TestSsl)
java.lang.ExceptionInInitializerError
        at java.base/javax.crypto.JceSecurityManager.<clinit>(JceSecurityManager.java:65)
        at java.base/javax.crypto.Cipher.getConfiguredPermission(Cipher.java:2624)
        at java.base/javax.crypto.Cipher.getMaxAllowedKeyLength(Cipher.java:2646)
        at java.base/sun.security.ssl.SSLCipher.isUnlimited(SSLCipher.java:540)
        at java.base/sun.security.ssl.SSLCipher.<init>(SSLCipher.java:472)
        at java.base/sun.security.ssl.SSLCipher.<clinit>(SSLCipher.java:174)
        at java.base/sun.security.ssl.CipherSuite.<clinit>(CipherSuite.java:67)
        at java.base/sun.security.ssl.SSLContextImpl.getApplicableSupportedCipherSuites(SSLContextImpl.java:348)
        at java.base/sun.security.ssl.SSLContextImpl$AbstractTLSContext.<clinit>(SSLContextImpl.java:579)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:315)
        at java.base/java.security.Provider$Service.getImplClass(Provider.java:1848)
        at java.base/java.security.Provider$Service.newInstance(Provider.java:1824)
        at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:236)
        at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:164)
        at java.base/javax.net.ssl.SSLContext.getInstance(SSLContext.java:168)
        at java.base/javax.net.ssl.SSLContext.getDefault(SSLContext.java:99)
        at TestSsl.testSslInit(TestSsl.java:11)

[ snip ]
Caused by: java.lang.SecurityException: Can not initialize cryptographic mechanism
        at java.base/javax.crypto.JceSecurity.<clinit>(JceSecurity.java:120)
        ... 41 more
[...]

PR #11873, opened 2020-07-30
Expose conf folder in the JDK repository

menny:
the conf folder holds the security policy files.

### DIFF
--- a/tools/jdk/jdk.BUILD
+++ b/tools/jdk/jdk.BUILD
@@ -1,6 +1,6 @@
-package(default_visibility = ["//visibility:public"])
+load("@rules_java//java:defs.bzl", "java_import", "java_runtime")
 
-load("@rules_java//java:defs.bzl", "java_runtime", "java_import")
+package(default_visibility = ["//visibility:public"])
 
 exports_files(["BUILD.bazel"])
 
@@ -181,6 +181,12 @@ filegroup(
     ),
 )
 
+#This folder holds security policies
+filegroup(
+    name = "jdk-conf",
+    srcs = glob(["conf/**"]),
+)
+
 filegroup(
     name = "jdk-include",
     srcs = glob(["include/**"]),
@@ -201,6 +207,7 @@ java_runtime(
     name = "jdk",
     srcs = [
         ":jdk-bin",
+        ":jdk-conf",
         ":jdk-include",
         ":jdk-lib",
         ":jre-default",


src/main/tools/jdk.BUILD
```diff
@@ -1,6 +1,6 @@
-package(default_visibility = ["//visibility:public"])
+load("@rules_java//java:defs.bzl", "java_import", "java_runtime")
 
-load("@rules_java//java:defs.bzl", "java_runtime", "java_import")
+package(default_visibility = ["//visibility:public"])
```
menny, line 3: These two lines will make `buildifier` happy.

src/main/tools/jdk.BUILD
```diff
@@ -181,6 +181,12 @@ filegroup(
     ),
 )
 
+#This folder holds security policies
+filegroup(
+    name = "jdk-conf",
+    srcs = glob(["conf/**"]),
```
menny, line 187: The `conf` folder holds security policy files (`limited` and `unlimited`). We use a Java mock-server that in tests that uses an untrusted cert, so we rely on `unlimited` policy to allow this to work.

src/main/tools/jdk.BUILD
menny, line 3: These two lines will make buildifier happy.

aiuto:
Liam: Can you take a look at this?

src/main/tools/jdk.BUILD
menny: we use Corretto JDK, and it has about 20 files under that folder (and its sub-folders) around 100kb in size.

philwo:
I looked into this now (sorry for the delay 😞 ), but I think most targets in this BUILD file are deprecated.

Unfortunately I don't understand how to use the recommended replacement `@bazel_tools//tools/jdk:current_java_runtime` instead of these filegroups.

@cushon @comius Can you help?

comius (approved):

comius:
@philwo: the changes look fine to me, the particular targets (jdk, java-includes) in jdk.BUILD are used, and I don't think you  can extend them in tools/jdk/BUILD.

I checked JDK11 and JDK14, and they have conf directory, also with other configuration files, not just policy. JDK8 does not yet have a conf folder. Also our remote JDKs have the conf directory.

@menny Can you rebase and resolve the conflicts?

comius:
closes https://github.com/bazelbuild/bazel/issues/9189

menny:
@comius done. Thank you.

PR #11887, opened 2020-08-03
Remove deprecated usage from ubuntu instructions.

rpavlik:
apt-key add is deprecated, using `/etc/apt/trusted.gpg.d` is the recommended approach now. De-armoring the GPG key before putting it there since some apt versions don't recognize armored keys, see https://github.com/openSUSE/software-o-o/pull/812


gregestren:
@philwo I think you're the best point of contact for this?

PR #11889, opened 2020-08-03
Add useDefaultShellEnvironment() in two places within the Java rules

grossag:
Two files in the Java tools were not using useDefaultShellEnvironment()
when using the SpawnAction.Builder class. This causes the PATH to be dropped
when using these rules, even if the user passed --action_env=PATH. This is
problematic for me because I depend on the PATH to help Bazel find the MSVCRT
dlls.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/java/DeployArchiveBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/DeployArchiveBuilder.java
@@ -375,6 +375,7 @@ public class DeployArchiveBuilder {
     if (!usingNativeSinglejar) {
       ruleContext.registerAction(
           new SpawnAction.Builder()
+              .useDefaultShellEnvironment()
               .addTransitiveInputs(inputs.build())
               .addTransitiveInputs(JavaRuntimeInfo.forHost(ruleContext).javaBaseInputsMiddleman())
               .addOutput(outputJar)
@@ -390,6 +391,7 @@ public class DeployArchiveBuilder {
     } else {
       ruleContext.registerAction(
           new SpawnAction.Builder()
+              .useDefaultShellEnvironment()
               .addTransitiveInputs(inputs.build())
               .addOutput(outputJar)
               .setResources(DEPLOY_ACTION_RESOURCE_SET)
--- a/src/main/java/com/google/devtools/build/lib/rules/java/ResourceJarActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/ResourceJarActionBuilder.java
@@ -127,6 +127,7 @@ public class ResourceJarActionBuilder {
     }
     ruleContext.registerAction(
         builder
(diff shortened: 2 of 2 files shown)


meteorcloudy:
@grossag 

On Windows, we always ensure the following directories are in PATH:
```
<SYSTEMROOT>
<SYSTEMROOT>\System32
<SYSTEMROOT>\System32\WindowsPowerShell\v1.0
```
(<SYSTEMROOT> is usually C:\Windows)
See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java;l=488

And msvcrt.dll should be available under `C:\Windows\System32`, can you explain a bit more about why do you need a custom PATH for providing msvcrt.dll?

meteorcloudy:
Using `useDefaultShellEnvironmen()` will leak many env vars to the action, which could damage Bazel's remote caching ability for those actions.

ulfjack:
@meteorcloudy, I don't think so. Last I checked, `useDefaultShellEnvironment()` was a misnomer, and the only thing it did was to propagate the env variables set in the configuration (through `--action_env` and `--test_env`). If I'm remembering correctly, then this should be a reasonable change.

grossag:
> @meteorcloudy, I don't think so. Last I checked, `useDefaultShellEnvironment()` was a misnomer, and the only thing it did was to propagate the env variables set in the configuration (through `--action_env` and `--test_env`). If I'm remembering correctly, then this should be a reasonable change.

That's right. Without `useDefaultShellEnvironment()`, these actions actually don't get any PATH. So for example, if you run a Windows Java Tools build without this patch, you will see an action invoking simplejar_local.exe that is:

1. cd <some_directory>
2. simplejar_local.exe <params>

That is, you don't see any step saying "set PATH=<default>". Every action is supposed to call `useDefaultShellEnvironment()` so the action sets the PATH as requested (either a default one containing some msys directories and C:\windows\system32) or passes through the entirety of the PATH if `--action_env=PATH` is provided.

For what it's worth, some of our builds pass `--action_env=PATH` because the build machines don't have the MSVC runtime installed. Instead, it is picked up from a separate directory. We may change that assumption for our consumers, but it is nice anyway for all actions to have compatibility with `--action_env=PATH`, which is what this action tries to fix.

meteorcloudy:
> @meteorcloudy, I don't think so. Last I checked, useDefaultShellEnvironment() was a misnomer, and the only thing it did was to propagate the env variables set in the configuration (through --action_env and --test_env). If I'm remembering correctly, then this should be a reasonable change.

@ulfjack Thanks for pointing out that! I was indeed confused by the function name. But to be more precise, using `useDefaultShellEnvironment()` will pass the client `PATH` env var to the action no matter using `--action_env=PATH` or not. But of course, you can override it by `--action=PATH=<a specific value>` so that you still get a hermetic build.

@grossag Then this change looks good to me! Thanks for the explanation. ;)

PR #11896, opened 2020-08-05
Account for internal actions in stats summary.

meisterT:
RELNOTES: None


meisterT:
cc @ulfjack

ulfjack:
Did you run some tests with this change?

meisterT:
I tested it locally with some builds and checked with the BEP whether it made sense.

If you are referring to the buildkite run: I didn't update the tests yet, wanted to get your opinion first.

ulfjack:
I meant manual tests to see if the numbers make sense. The general approach seems fine, but why treat internal actions as a separate category?

meisterT:
I see that my more verbose commit message (that I added after the fact) got lost.

What annoys me and triggered this change is this:
```
INFO: 2162 processes: 1445 linux-sandbox, 1 local, 716 worker.
INFO: Build completed successfully, 2178 total actions
```

That just doesn't add up and users are confused by this frequently.

I used internal since this is stuff like SymlinkTree which clearly is something else than local.

ulfjack:
I don't have any good counter-propopsals, so sure, go ahead.

PR #11898, opened 2020-08-06
Fix ISE w/--remote_download_toplevel + local exec

### LINKED ISSUE #11825: bazel crashes with IllegalStateException when "remote_download_toplevel" is passed to local build
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

`bazel` crashes with `java.lang.IllegalStateException` when `--remote_download_toplevel` is passed to a local build. Often we run remote builds and local builds via a wrapper and keep the same arguments. If the build is not remote (i.e. no --remote_executor) is provided, bazel crashes. 

Consistent behaviour with incompatible flags. If the flag is not compatible, we should errorout. Ideally this flag should be a noop for local builds. 

```sh
workspace=$(mktemp -d)
cd $workspace/
touch WORKSPACE BUILD.bazel
bazel build //... --remote_download_toplevel
```

Linux

release 3.4.1

ulfjack:
This silently ignores --remote_download_toplevel if remote execution is
disabled; some users have the flag enabled by default, but sometimes
run things locally for debugging.

Fixes #11825.

Change-Id: I0419f01ba5a123428e6d2fd0addfc80437da9525

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -497,8 +497,9 @@ public final class RemoteModule extends BlazeModule {
       BuildRequest request,
       BuildOptions buildOptions,
       Iterable<ConfiguredTarget> configuredTargets) {
-    if (remoteOutputsMode != null && remoteOutputsMode.downloadToplevelOutputsOnly()) {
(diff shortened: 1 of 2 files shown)


coeuvre:
LGTM. Merging.

PR #11899, opened 2020-08-06
Fix parsing of BRDA lines in LcovParser

### LINKED ISSUE #11051: Incorrect parsing of BRDA lines in LcovParser.java

This snippet of code: https://github.com/bazelbuild/bazel/blob/master/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/LcovParser.java#L283

Is trying to parse BRDA lines which have the following format: BRDA:<line number>,<block number>,<branch number>,<taken>

<taken> is either "-" if branch block was never executed, or a number indicating how many times the branch was executed.

In case that LCOV file contains a line like this one: 

BRDA:52,0,1,-

Piece of code on line 283 will exception:

if (taken.equals(TAKEN)) {
        executionCount = Integer.parseInt(taken);
        wasExecuted = true;
}

An exception is logged (WARNING: Tracefile contains an invalid number BRDA line BRDA:52,0,1,-) and processing continues. Ultimately, this will lead to incorrect coverage numbers.

Seems like the line should say if (**!**taken.equals(TAKEN)){ ...

Example of a file where coverage will be incorrectly calculated:
[example.info.txt](https://github.com/bazelbuild/bazel/files/4417186/example.info.txt)

c-mita:
Fixes the interpretation of the "-" in a BRDA line (meaning the branch
was never executed, as opposed to executed but never taken) and fixes
the internal SourceFileCoverage abstraction to account for the fact that
multiple BRDA statements will occur for the same line.

Based on the PR https://github.com/bazelbuild/bazel/pull/10708 but does
not include any changes for the Java code coverage generator.

Fixes #11051


c-mita:
Issues with the BRDA parser have come up in several pull requests that also target individual lcov generators (e.g. https://github.com/bazelbuild/bazel/pull/10708, https://github.com/bazelbuild/bazel/pull/11489).

There is still some confusion with how `BranchCoverage` objects get mapped to "BA" or "BRDA" in the Lcov output; which can be worked on separately.

tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/Constants.java
```diff
@@ -36,7 +36,7 @@
   static final String LH_MARKER = "LH:";
   static final String LF_MARKER = "LF:";
   static final String END_OF_RECORD_MARKER = "end_of_record";
-  static final String TAKEN = "-";
+  static final String NOT_TAKEN = "-";
```
c-mita, line 39: Maybe this is better as "NEVER_EXECUTED", since it means the branches in a block were never executed, not just taken.

There's a bit of inconsistency with the usages of the words "taken" and "executed" when compared to the docs here: http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php

```
  BRDA:<line number>,<block number>,<branch number>,<taken>`

       Block  number  and  branch  number are gcc internal IDs for the branch.
       Taken is either '-' if the basic block containing the branch was  never
       executed or a number indicating how often that branch was taken.
```
We seem to have reversed the words "taken" and "execution"...

PR #11900, opened 2020-08-06
Use boolean formulas to express filters in --build_tag_filters and --test_tag_filters

ehkloaj:
Starlark can now be used to express filters instead of a list of tags in --test_tag_filters and --build_tag_filters.

Example of usage:
bazel test ... --test_tag_filters="tag1 and (tag2 or not tag3)"

This is a draft of the functionality requested in #3705. 

### Summary of commit
I modified the method tagFilter which takes a list of tags as input and assumed if this list only contains one required tag and none excluded tag that this is a Starlark expression (there is a problem with this, see Concerns). I implemented the function evalExpression as described by @alandonovan in https://github.com/bazelbuild/bazel/issues/3705#issuecomment-661106688, and it works well. In the case where a user provides an invalid Starlark expression (that fails to parse or uses forbidden operators) a new sub type of RuntimeException is thrown (this is a temporary solution, see Concerns). 

### Concerns
I have a few concerns about the implementation that I need the community's help with.

1. Since tags can be any valid string (meaning it can include spaces) it is possible that we have a situation where we have defined a tag "tag1 and tag2". If we then run --test_tag_filters="tag1 and tag2" there is no way of knowing if this is meant to be a Starlark expression or not.
In my implementation this would also be a problem if we had a tag "tag 1 and 2 " and ran --test_tag_filters="tag 1 and 2" because this would be assumed to be a Starlark expression and then the parsing would fail.

2. It would be better to throw an OptionsParsingException than a sub type of RuntimeException because then this could be caught in an appropriate way and communicated to the user instead of the user getting an internal error. The reason that I did not decide to use OptionsParsingException was that then I would have to add "throws OptionsParsingException" to a lot of methods before finally reaching one where it can be handled appropriately. This was considered unnecessary work at the moment. 

### Proposed solution
I propose that we either add a new flag --build_tag_filter_expression (and same for test) or a flag that specifies if the input given to --build_tag_filters is a Starlark expression or not. 
This way we avoid the problems in Concerns because there is no way that a tag will be misinterpreted as a Starlark expression and we can do the parsing earlier and therefore handle invalid input in a correct manner.
What do you think about this?

### DIFF
--- a/site/docs/user-manual.html
+++ b/site/docs/user-manual.html
@@ -1194,13 +1194,37 @@ memory-saving mode</a>.
   setting is high enough.
 </p>
 
-<h4 id='flag--build_tag_filters'><code class='flag'>--build_tag_filters=<var>tag[,tag]*</var></code></h4>
+<h4 id='flag--build_tag_filters'><code class='flag'>--build_tag_filters <var>tag[,tag]*</var> | <var>boolean expression</var></code></h4>
 <p>
-  If specified, Bazel will build only targets that have at least one required tag
-  (if any of them are specified) and does not have any excluded tags. Build tag
-  filter is specified as comma delimited list of tag keywords, optionally
-  preceded with '-' sign used to denote excluded tags. Required tags may also
-  have a preceding '+' sign.
+  If specified, Bazel will build only targets with matching tags. The
+  value is either a comma-separated list, or a boolean expression.
+</p>
+<p>
+  A comma-separated list matches if the target has any of the tags,
+  and none of the excluded tags. Excluded tags are preceded with '-'
+  signs. Required tags may also have a preceding '+' sign.
+</p>
+<p>
+  For example,
+</p>
+<pre>
+  % bazel build --build_tag_filters=product,stub,-prototype //myproject:all
+</pre>
+  will build targets that are tagged with either <code>product</code> or <code>stub</code> but are
+  <b>not</b> tagged with the <code>prototype</code> tag.
+</p>
+<p>
+  Boolean expressions have the form
+  <code>expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'.</code>
+  <code>IDENT</code> is defined as a Starlark identifier, with the same reserved words.
+<p>
+<pre>
+  % bazel build --build_tag_filters='(product or stub) and not prototype' //myproject:all
+</pre>
+  is equivalent to
+<pre>
+  % bazel build --build_tag_filters=product,stub,-prototype //myproject:all
+</pre>
 </p>
 <p>
   When running tests, Bazel ignores <code class='flag'>--build_tag_filters</code> for test targets,
@@ -1244,26 +1268,41 @@ memory-saving mode</a>.
 </p>
 
 
-<h4 id='flag--test_tag_filters'><code class='flag'>--test_tag_filters=<var>tag[,tag]*</var></code></h4>
+<h4 id='flag--test_tag_filters'><code class='flag'>--test_tag_filters <var>tag[,tag]</var> | <var>boolean expression</var></code></h4>
 <p>
   If specified, Bazel will test (or build if <code class='flag'>--build_tests_only</code>
-  is also specified) only test targets that have at least one required tag
-  (if any of them are specified) and does not have any excluded tags. Test tag
-  filter is specified as comma delimited list of tag keywords, optionally
-  preceded with '-' sign used to denote excluded tags. Required tags may also
-  have a preceding '+' sign.
+  is also specified) only targets with matching tags. The value is either a comma-separated
+  list, or a boolean expression.
+</p>
+<p>
+  A comma-separated list matches if the target has any of the tags,
+  and none of the excluded tags. Excluded tags are preceded with '-'
+  signs. Required tags may also have a preceding '+' sign.
 </p>
 <p>
   For example,
+</p>
 <pre>
-  % bazel test --test_tag_filters=performance,stress,-flaky //myproject:all
+  % bazel test --test_tag_filters=release,smoke,-flaky //myproject:all
 </pre>
-<p>
-  will test targets that are tagged with either <code>performance</code> or
-  <code>stress</code> tag but are <b>not</b> tagged with the <code>flaky</code>
-  tag.
+  will test targets that are tagged with either <code>release</code> or <code>smoke</code> but are
+  <b>not</b> tagged with the <code>flaky</code> tag.
 </p>
 <p>
+  Boolean expressions have the form
+  <code>expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'.</code>
+  <code>IDENT</code> is defined as a Starlark identifier, with the same reserved words.
+<p>
+<pre>
+  % bazel test --test_tag_filters='(release or smoke) and not flaky' //myproject:all
+</pre>
+  is equivalent to
+<pre>
+  % bazel test --test_tag_filters=release,smoke,-flaky //myproject:all
+</pre>
+</p>
+<p>
+  By default, test tag filtering is not applied.
   By default, test tag filtering is not applied. Note that you can also filter
   on test's <code>size</code> and <code>local</code> tags in
   this manner.
--- a/src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
@@ -14,9 +14,6 @@
 
 package com.google.devtools.build.lib.packages;
 
-import static com.google.devtools.build.lib.packages.BuildType.TRISTATE;
-import static com.google.devtools.build.lib.packages.Type.BOOLEAN;
-
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
@@ -24,16 +21,24 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.devtools.build.lib.actions.ExecutionRequirements;
 import com.google.devtools.build.lib.cmdline.Label;
-import com.google.devtools.build.lib.util.Pair;
+import net.starlark.java.eval.Dict;
+import net.starlark.java.eval.EvalException;
+import net.starlark.java.syntax.BinaryOperatorExpression;
+import net.starlark.java.syntax.Expression;
+import net.starlark.java.syntax.Identifier;
+import net.starlark.java.syntax.Location;
+import net.starlark.java.syntax.TokenKind;
+import net.starlark.java.syntax.UnaryOperatorExpression;
+
+import javax.annotation.Nullable;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nullable;
-import net.starlark.java.eval.Dict;
-import net.starlark.java.eval.EvalException;
-import net.starlark.java.syntax.Location;
+import java.util.Set;
+
+import static com.google.devtools.build.lib.packages.BuildType.TRISTATE;
+import static com.google.devtools.build.lib.packages.Type.BOOLEAN;
 
 /**
  * Utility functions over Targets that don't really belong in the base {@link
@@ -348,29 +353,11 @@ public final class TargetUtils {
   }
 
   /**
-   * Returns a predicate to be used for test tag filtering, i.e., that only accepts tests that match
-   * all of the required tags and none of the excluded tags.
+   * Returns a predicate to be used for test tag filtering based on an expression.
    */
-  public static Predicate<Target> tagFilter(List<String> tagFilterList) {
-    Pair<Collection<String>, Collection<String>> tagLists =
-        TestTargetUtils.sortTagsBySense(tagFilterList);
-    final Collection<String> requiredTags = tagLists.first;
-    final Collection<String> excludedTags = tagLists.second;
-    return input -> {
-      if (requiredTags.isEmpty() && excludedTags.isEmpty()) {
-        return true;
-      }
-
-      if (!(input instanceof Rule)) {
-        return requiredTags.isEmpty();
-      }
-      // Note that test_tags are those originating from the XX_test rule, whereas the requiredTags
-      // and excludedTags originate from the command line or test_suite rule.
-      // TODO(ulfjack): getRuleTags is inconsistent with TestFunction and other places that use
-      // tags + size, but consistent with TestSuite.
-      return TestTargetUtils.testMatchesFilters(
-          ((Rule) input).getRuleTags(), requiredTags, excludedTags, false);
-    };
+  public static Predicate<Target> tagFilter(@Nullable Expression tagFilter) {
+    return input -> tagFilter == null ||
+        input instanceof Rule && evalExpression(tagFilter, ((Rule) input).getRuleTags());
   }
 
   /**
@@ -422,5 +409,40 @@ public final class TargetUtils {
   public static String formatMissingEdge(
       @Nullable Target target, Label label, NoSuchThingException e) {
     return formatMissingEdge(target, label, e, null);
+  }
+
+  /**
+   * Evaluates a boolean formula given by this grammar:
+   * expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'
+   * Identifiers follow the Starlark definition (and exclude Starlark reserved words).
+   * If the given expression is invalid an IllegalStateException is thrown.
+   * This should never happen since the input has been controlled in ExpressionConverter.
+   */
+  private static boolean evalExpression(Expression expression, Set<String> tags) {
+    switch (expression.kind()) {
+      case IDENTIFIER:
+        Identifier ident = (Identifier) expression;
+        return tags.contains(ident.getName());
+      case BINARY_OPERATOR:
+        BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
+        switch (boe.getOperator()) {
+          case OR:
+            return evalExpression(boe.getX(), tags) || evalExpression(boe.getY(), tags);
+          case AND:
+            return evalExpression(boe.getX(), tags) && evalExpression(boe.getY(), tags);
+          default:
+            throw new IllegalStateException(
+                String.format("Got %s but expected 'and' or 'or'", boe.getOperator()));
+        }
+      case UNARY_OPERATOR:
+        UnaryOperatorExpression uoe = (UnaryOperatorExpression) expression;
+        if (uoe.getOperator() == TokenKind.NOT) {
+          return !evalExpression(uoe.getX(), tags);
+        } else {
+          throw new IllegalStateException(String.format("Got %s but expected 'not'", uoe.getOperator()));
+        }
+      default:
+        throw new IllegalStateException(String.format("Unsupported expression kind: %s", expression.kind()));
+    }
   }
 }
--- a/src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
@@ -15,11 +15,21 @@ package com.google.devtools.build.lib.pkgcache;
 
 import com.google.devtools.build.lib.packages.TestSize;
 import com.google.devtools.build.lib.packages.TestTimeout;
+import com.google.devtools.common.options.Converter;
 import com.google.devtools.common.options.Converters.CommaSeparatedOptionListConverter;
 import com.google.devtools.common.options.Option;
 import com.google.devtools.common.options.OptionDocumentationCategory;
 import com.google.devtools.common.options.OptionEffectTag;
 import com.google.devtools.common.options.OptionsBase;
+import com.google.devtools.common.options.OptionsParsingException;
+import net.starlark.java.syntax.BinaryOperatorExpression;
+import net.starlark.java.syntax.Expression;
+import net.starlark.java.syntax.ParserInput;
+import net.starlark.java.syntax.SyntaxError;
+import net.starlark.java.syntax.TokenKind;
+import net.starlark.java.syntax.UnaryOperatorExpression;
+
+import javax.annotation.Nullable;
 import java.util.List;
 import java.util.Set;
 
@@ -56,32 +66,52 @@ public class LoadingOptions extends OptionsBase {
 
   @Option(
     name = "build_tag_filters",
-    converter = CommaSeparatedOptionListConverter.class,
+    converter = TagFilterConverter.class,
     defaultValue = "",
     documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
-        "Specifies a comma-separated list of tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those targets will be built that "
-            + "contain at least one included tag and do not contain any excluded tags. This option "
+        "A boolean formula over rule tags that selects the targets to build. "
+            + "Example: '(server or client) and not experimental'. "
+            + "See --test_tag_filters for details. This option "
             + "does not affect the set of tests executed with the 'test' command; those are be "
             + "governed by the test filtering options, for example '--test_tag_filters'"
   )
-  public List<String> buildTagFilterList;
+  public TagFilter buildTagFilter;
 
   @Option(
     name = "test_tag_filters",
-    converter = CommaSeparatedOptionListConverter.class,
+    converter = TagFilterConverter.class,
     defaultValue = "",
     documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
-        "Specifies a comma-separated list of test tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those test targets will be "
-            + "found that contain at least one included tag and do not contain any excluded "
-            + "tags. This option affects --build_tests_only behavior and the test command."
+        "A boolean formula over test tags that selects the tests to run. "
+            + "Example: '(release or smoke) and not flaky'. Formulas have the form: "
+            + "expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'. "
+            + "IDENT is defined as a Starlark identifier, with the same reserved words. "
+            + "Comma-separated list of tags are also supported: tag[,tag]*. Each tag can be optionally "
+            + "preceded with '-' to specify excluded tags."
   )
-  public List<String> testTagFilterList;
+  public TagFilter testTagFilter;
+
+  /**
+   * TagFilterConverter parses a tag filter, which can either be a boolean expression or a comma-separated list.
+   *
+   * Declared public to make it accessible to reflection.
+   */
+  public static class TagFilterConverter implements Converter<TagFilter> {
+    @Override
+    @Nullable
+    public TagFilter convert(String input) throws OptionsParsingException {
+      return new TagFilter(input);
+    }
+
+    @Override
+    public String getTypeDescription() {
+      return "A boolean formula over test tags (e.g. '(release or smoke) and not flaky')";
+    }
+  }
 
   @Option(
     name = "test_size_filters",
@@ -163,4 +193,73 @@ public class LoadingOptions extends OptionsBase {
         + "then they can analyze test_suite targets."
   )
   public boolean expandTestSuites;
+
+  /** BooleanFormulaConverter parses a boolean formula to an Expression.
+   * Declared public to make it accessible to reflection. */
+  public static class BooleanFormulaConverter implements Converter<Expression> {
+    @Override
+    @Nullable public Expression convert(String input) throws OptionsParsingException {
+      if (input.isEmpty()) {
+        return null;
+      }
+      if (isLegacySyntax(input)) {
+        input = convertLegacySyntaxToBooleanFormula(input);
+      }
+      try {
+        Expression result = Expression.parse(ParserInput.fromLines(input));
+        validate(result);
+        return result;
+      } catch (SyntaxError.Exception e) {
+        throw new OptionsParsingException("Failed to parse expression: " + e.getMessage() + " input: " + input, e);
+      }
+    }
+
+    @Override
+    public String getTypeDescription() {
+      return "A Boolean formula over test tags (e.g. '(release or smoke) and not flaky')";
+    }
+
+    private static boolean isLegacySyntax(String input) {
+      return input.contains(",") || input.contains("-");
+    }
+
+    /**
+     * Converts a string containing tags separated by comma to a boolean formula.
+     */
+    private String convertLegacySyntaxToBooleanFormula(String input) throws OptionsParsingException {
+      return input.replaceAll(" *, *-", " and not ")
+              .replaceAll(" *, *", " or ")
+              .replace("-", " not ").trim();
+    }
+
+    /**
+     * Throws an OptionsParsingException when the given boolean formula does not follow the grammar:
+     * expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'
+     */
+    private void validate(Expression expression) throws OptionsParsingException {
+      switch (expression.kind()) {
+        case IDENTIFIER:
+          break;
+        case BINARY_OPERATOR:
+          BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
+          if (boe.getOperator() != TokenKind.OR && boe.getOperator() != TokenKind.AND) {
+            throw new OptionsParsingException(String.format("invalid Boolean operator: %s (want 'and' or 'or')",
+                boe.getOperator()));
+          }
+          validate(boe.getX());
+          validate(boe.getY());
+          break;
+        case UNARY_OPERATOR:
+          UnaryOperatorExpression uoe = (UnaryOperatorExpression) expression;
+          if (uoe.getOperator() != TokenKind.NOT) {
+            throw new OptionsParsingException(String.format("invalid Boolean operator: %s (want 'not')",
+                uoe.getOperator()));
+          }
+          validate(uoe.getX());
+          break;
+        default:
+          throw new OptionsParsingException("invalid Boolean operator");
+      }
+    }
+  }
 }
--- a/src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
+++ b/src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
@@ -23,20 +23,20 @@ import com.google.devtools.build.lib.packages.Target;
 import com.google.devtools.build.lib.packages.TargetUtils;
 import com.google.devtools.build.lib.packages.TestSize;
 import com.google.devtools.build.lib.packages.TestTimeout;
-import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
+import net.starlark.java.syntax.Expression;
+
+import javax.annotation.Nullable;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.Predicate;
-import javax.annotation.Nullable;
 
 /**
  * Predicate that implements test filtering using the command-line options in {@link
  * LoadingOptions}. Implements {@link #hashCode} and {@link #equals} so it can be used as a Skyframe
  * key.
  */
-@AutoCodec
 public final class TestFilter implements com.google.common.base.Predicate<Target> {
   private static final Predicate<Target> ALWAYS_TRUE = (t) -> true;
 
@@ -49,25 +49,24 @@ public final class TestFilter implements com.google.common.base.Predicate<Target
     return new TestFilter(
         ImmutableSet.copyOf(options.testSizeFilterSet),
         ImmutableSet.copyOf(options.testTimeoutFilterSet),
-        ImmutableList.copyOf(options.testTagFilterList),
+        options.testTagFilter.getExpression(),
         ImmutableList.copyOf(options.testLangFilterList));
   }
 
   private final ImmutableSet<TestSize> testSizeFilterSet;
   private final ImmutableSet<TestTimeout> testTimeoutFilterSet;
-  private final ImmutableList<String> testTagFilterList;
+  @Nullable private final Expression testTagFilter;
   private final ImmutableList<String> testLangFilterList;
   private final Predicate<Target> impl;
 
-  @AutoCodec.VisibleForSerialization
   TestFilter(
       ImmutableSet<TestSize> testSizeFilterSet,
       ImmutableSet<TestTimeout> testTimeoutFilterSet,
-      ImmutableList<String> testTagFilterList,
+      @Nullable Expression testTagFilter,
       ImmutableList<String> testLangFilterList) {
     this.testSizeFilterSet = testSizeFilterSet;
     this.testTimeoutFilterSet = testTimeoutFilterSet;
-    this.testTagFilterList = testTagFilterList;
+    this.testTagFilter = testTagFilter;
     this.testLangFilterList = testLangFilterList;
     Predicate<Target> testFilter = ALWAYS_TRUE;
     if (!testSizeFilterSet.isEmpty()) {
@@ -76,8 +75,8 @@ public final class TestFilter implements com.google.common.base.Predicate<Target
     if (!testTimeoutFilterSet.isEmpty()) {
       testFilter = testFilter.and(testTimeoutFilter(testTimeoutFilterSet));
     }
-    if (!testTagFilterList.isEmpty()) {
-      testFilter = testFilter.and(TargetUtils.tagFilter(testTagFilterList));
+    if (testTagFilter != null) {
+      testFilter = testFilter.and(TargetUtils.tagFilter(testTagFilter));
     }
     if (!testLangFilterList.isEmpty()) {
       testFilter = testFilter.and(testLangFilter(testLangFilterList));
@@ -92,7 +91,7 @@ public final class TestFilter implements com.google.common.base.Predicate<Target
 
   @Override
   public int hashCode() {
-    return Objects.hash(testSizeFilterSet, testTimeoutFilterSet, testTagFilterList,
+    return Objects.hash(testSizeFilterSet, testTimeoutFilterSet, testTagFilter,
         testLangFilterList);
   }
 
@@ -107,7 +106,7 @@ public final class TestFilter implements com.google.common.base.Predicate<Target
     TestFilter f = (TestFilter) o;
     return f.testSizeFilterSet.equals(testSizeFilterSet)
         && f.testTimeoutFilterSet.equals(testTimeoutFilterSet)
-        && f.testTagFilterList.equals(testTagFilterList)
+        && Objects.equals(f.testTagFilter, testTagFilter)
         && f.testLangFilterList.equals(testLangFilterList);
   }
 
--- a/src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
@@ -33,10 +33,12 @@ import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.build.skyframe.SkyFunctionName;
 import com.google.devtools.build.skyframe.SkyKey;
 import com.google.devtools.build.skyframe.SkyValue;
+import net.starlark.java.syntax.Expression;
+
+import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Objects;
-import javax.annotation.Nullable;
 
 /**
  * A value referring to a computed set of resolved targets. This is used for the results of target
@@ -162,7 +164,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
       boolean compileOneDependency,
       boolean buildTestsOnly,
       boolean determineTests,
-      ImmutableList<String> buildTargetFilter,
+      @Nullable Expression buildTargetFilter,
       boolean buildManualTests,
       boolean expandTestSuites,
       @Nullable TestFilter testFilter) {
@@ -189,7 +191,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
   public static SkyKey keyWithoutFilters(
       ImmutableList<String> targetPatterns, PathFragment offset) {
     return new TargetPatternPhaseKey(
-        targetPatterns, offset, false, false, false, ImmutableList.of(), false, false, null);
+        targetPatterns, offset, false, false, false, null, false, false, null);
   }
 
   /** The configuration needed to run the target pattern evaluation phase. */
@@ -202,7 +204,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
     private final boolean compileOneDependency;
     private final boolean buildTestsOnly;
     private final boolean determineTests;
-    private final ImmutableList<String> buildTargetFilter;
+    @Nullable private final Expression buildTargetFilter;
     private final boolean buildManualTests;
     private final boolean expandTestSuites;
     @Nullable private final TestFilter testFilter;
@@ -213,7 +215,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
         boolean compileOneDependency,
         boolean buildTestsOnly,
         boolean determineTests,
-        ImmutableList<String> buildTargetFilter,
+        @Nullable Expression buildTargetFilter,
         boolean buildManualTests,
         boolean expandTestSuites,
         @Nullable TestFilter testFilter) {
@@ -222,7 +224,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
       this.compileOneDependency = compileOneDependency;
       this.buildTestsOnly = buildTestsOnly;
       this.determineTests = determineTests;
-      this.buildTargetFilter = Preconditions.checkNotNull(buildTargetFilter);
+      this.buildTargetFilter = buildTargetFilter;
       this.buildManualTests = buildManualTests;
       this.expandTestSuites = expandTestSuites;
       this.testFilter = testFilter;
@@ -256,7 +258,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
       return determineTests;
     }
 
-    public ImmutableList<String> getBuildTargetFilter() {
+    @Nullable public Expression getBuildTargetFilter() {
       return buildTargetFilter;
     }
 
@@ -314,7 +316,7 @@ public final class TargetPatternPhaseValue implements SkyValue {
           && other.compileOneDependency == compileOneDependency
           && other.buildTestsOnly == buildTestsOnly
           && other.determineTests == determineTests
-          && other.buildTargetFilter.equals(buildTargetFilter)
+          && Objects.equals(other.buildTargetFilter, buildTargetFilter)
           && other.buildManualTests == buildManualTests
           && other.expandTestSuites == expandTestSuites
           && Objects.equals(other.testFilter, testFilter);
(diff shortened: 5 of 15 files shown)


alandonovan:
> Since tags can be any valid string (meaning it can include spaces) it is possible that we have a situation where we have defined a tag "tag1 and tag2". If we then run --test_tag_filters="tag1 and tag2" there is no way of knowing if this is meant to be a Starlark expression or not.

This sounds like a problem in theory but not in practice. The vast majority of tags are identifiers; users who use arbitrary string tags won't be able to use them in filter expressions. If that matters to them, they will be motivated to change their tags.

If it really becomes necessary to support arbitrary string tags (and I am not convinced that it will), then the obvious fix is to allow Starlark string literals too, so you can say "foo and bar and 'tag with spaces'".  But let's not add that yet.

> I propose that we either add a new flag --build_tag_filter_expression (and same for test) or a flag that specifies if the input given to --build_tag_filters is a Starlark expression or not.

What was wrong with my simpler suggestion in https://github.com/bazelbuild/bazel/issues/3705#issuecomment-661106688, that we assume (and check) that the string is a valid expression unless it contains a comma, in which case it should be interpreted as a legacy form "tag1,...,tagN" (an implicit conjunction)?

We don't need another flag.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -414,4 +445,42 @@ public static String formatMissingEdge(
...
+            return evalExpression(boe.getX(), trueIdentifiers) && evalExpression(boe.getY(), trueIdentifiers);
+          default:
+            throw restrictedOperatorException;
+        }
+      case UNARY_OPERATOR:
+        UnaryOperatorExpression uoe = (UnaryOperatorExpression) expression;
+        if (uoe.getOperator() == TokenKind.NOT) {
+          return !evalExpression(uoe.getX(), trueIdentifiers);
+        } else {
+          throw restrictedOperatorException;
```
alandonovan, line 480: "syntax error: got %s, want 'not'"

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -414,4 +445,42 @@ public static String formatMissingEdge(
...
+        }
+      default:
+        throw restrictedOperatorException;
```
alandonovan, line 483: "syntax error"

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -339,14 +346,38 @@ private static boolean isExplicitDependency(Rule rule, Label label) {
...
   public static Predicate<Target> tagFilter(List<String> tagFilterList) {
     Pair<Collection<String>, Collection<String>> tagLists =
         TestTargetUtils.sortTagsBySense(tagFilterList);
     final Collection<String> requiredTags = tagLists.first;
     final Collection<String> excludedTags = tagLists.second;
+
+    // If there is only one required tag and none excluded tags we can be sure that it is safe to handle this as a
+    // Starlark expression.
+    // If the comparison fails we will handle it as a list of tags.
+    if (requiredTags.size() == 1 && excludedTags.size() == 0) {
```
alandonovan, line 366: It would be much clearer to have the caller provide a string (not a list of strings already split on commas) and parse it here. Then this condition becomes simply !str.contains(',').  If the string does contain a comma, it's trivial to split it and return an evaluator for an array of implicitly conjoined tags, or emit an error message "invalid boolean formula: a,b,c (did you mean 'a and b and c'?)".

ehkloaj:
Thanks for your review @alandonovan. I have updated most of your comments.

The reason why tagFilter is accepting a list as input is because the value of --test_tag_filter and --build_tag_filter is stored as a list. 
https://github.com/bazelbuild/bazel/blob/1f684e1b87cd8881a0a4b33e86ba66743e32d674/src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java#L70 
Also it is not enough to check for commas since you could have a case where you want to exclude "tag1" using the legacy syntax and run --test_tag_filter="-tag1". This would then be interpreted as an expression instead of legacy and we would get unexpected behavior.  

I am assuming that throwing a RuntimeException when an invalid expression is given (as it is implemented now) is not preferred seeing as though this would result in an internal error and a stack trace. When you wrote "syntax error" in your comments did you mean that you wanted the exception to be of type `com.google.devtools.build.lib.syntax.SyntaxError`? If so do you have any suggestion on where this can be caught and communicated to the user?

alandonovan:
> The reason why tagFilter is accepting a list as input is because the value of --test_tag_filter and --build_tag_filter is stored as a list.

Right. So let's change the type of the LoadingOptions field to syntax.Expression, and change its 'converter' to a new class that will parse the string (whether old or new) into a boolean formula. It should report an invalid formula by throwing an OptionsParsingException. The "syntax error..." comments were suggestions for the actual message that should be displayed to the user.

> Also it is not enough to check for commas since you could have a case where you want to exclude "tag1" using the legacy syntax and run --test_tag_filter="-tag1".

Ah, I forgot that the legacy syntax permits negation too. So, let's use refine the predicate for legacy mode: contains(',') or contains('-').

ehkloaj:
Thanks @alandonovan! I have implemented your suggestions. Let me know what you think.

laurentlb:
fyi, Alan is on vacation for a few days.
Since the issue was tagged with 'team-configurability', I'm adding Greg.

The new syntax is much nicer, IMO!

gregestren:
FYI I'm still taking a back seat to @alandonovan here since you both have already had good review back & forth.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+        checkForForbiddenOperators(result);
+        return result;
+      } catch (SyntaxError.Exception e) {
+        throw new OptionsParsingException("Failed to parse expression: " + e.getMessage() + " input: " + input, e);
+      }
+    }
+
+    @Override
+    public String getTypeDescription() {
+      return "Converts to an Expression.";
```
alandonovan, line 659: This string should be a noun-phrase that concisely describes what the user is expected to type. For example:

"A Boolean formula over test tags (e.g. '(release or smoke) and not flaky')".

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+    }
+
+    private boolean isLegacySyntax(String input) {
```
alandonovan, line 662: static 

Or just inline it with a comment.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+      return input.contains(",") || input.contains("-");
+    }
+
+    /**
+     * Converts a string containing tags separated by comma to a boolean expression.
+     */
+    private String convertLegacySyntaxToBooleanExpression(String input) throws OptionsParsingException {
+      CommaSeparatedOptionListConverter legacyConverter = new CommaSeparatedOptionListConverter();
+      List<String> tags = legacyConverter.convert(input);
```
alandonovan, line 671: Simpler: input.split(",")

Better still, just replace "," with " or " and "-" with "not " and don't bother with lists.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+      return (" " + String.join(" or ", tags)).replace(" -", " not ").substring(1);
```
alandonovan, line 672: Print a warning showing the expression in the new syntax. At some point we'll make this an error. Let's say Jan 1 2021.

src/test/java/com/google/devtools/common/options/ExpressionConverterTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.devtools.build.lib.syntax.Expression;
+import com.google.devtools.build.lib.syntax.ParserInput;
+import com.google.devtools.build.lib.syntax.SyntaxError;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** A test for {@link ExpressionConverter}. */
+@RunWith(JUnit4.class)
+public class ExpressionConverterTest {
```
alandonovan, line 16: Don't test ExpressionConverter: it is not a public API. Instead, test its effects through the blaze test command, as that is the interface that actually matters.

src/main/java/com/google/devtools/build/lib/packages/TargetUtils.java
```diff
@@ -414,4 +402,39 @@ public static String formatMissingEdge(
...
+        return tags.contains(ident.getName());
+      case BINARY_OPERATOR:
+        BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
+        switch (boe.getOperator()) {
+          case OR:
+            return evalExpression(boe.getX(), tags) || evalExpression(boe.getY(), tags);
+          case AND:
+            return evalExpression(boe.getX(), tags) && evalExpression(boe.getY(), tags);
+          default:
+            throw new FilterEvaluationException(
```
alandonovan, line 426: Use IllegalArgumentException. We've already verified the expression is valid.

src/main/java/com/google/devtools/build/lib/packages/FilterEvaluationException.java
```diff
@@ -0,0 +1,16 @@
+package com.google.devtools.build.lib.packages;
+
+/**
+ *  Exception indicating a failed attempt to use an expression in --test_tag_filter or --build_tag_filter.
+ *  The exception is thrown when the expression could not be parsed or when a forbidden operator is used.
+ */
+public class FilterEvaluationException extends RuntimeException {
```
alandonovan, line 7: Delete. This exception is never thrown.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -56,32 +58,38 @@
...
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
-        "Specifies a comma-separated list of tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those targets will be built that "
-            + "contain at least one included tag and do not contain any excluded tags. This option "
-            + "does not affect the set of tests executed with the 'test' command; those are be "
-            + "governed by the test filtering options, for example '--test_tag_filters'"
+        "Specifies a boolean expression or a list of tags separated by comma. A boolean expression must follow the "
+            + "syntax: expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'. "
+            + "When specifying a list of tags a tag can be optionally preceded with '-' to specify excluded tags. "
```
alandonovan, line 68: This sentence is about the legacy syntax. Make clear that it is going away after Jan 1, 2021.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -56,32 +58,38 @@
...
     defaultValue = "",
     documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
-        "Specifies a comma-separated list of tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those targets will be built that "
-            + "contain at least one included tag and do not contain any excluded tags. This option "
-            + "does not affect the set of tests executed with the 'test' command; those are be "
-            + "governed by the test filtering options, for example '--test_tag_filters'"
+        "Specifies a boolean expression or a list of tags separated by comma. A boolean expression must follow the "
```
alandonovan, line 66: Boolean "formula", not expression.

Delete "or a list of tags separated by comma": that's the legacy syntax. Add a note at the end explaining that it is deprecated.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -56,32 +58,38 @@
...
+    converter = ExpressionConverter.class,
     defaultValue = "",
     documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
-        "Specifies a comma-separated list of test tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those test targets will be "
-            + "found that contain at least one included tag and do not contain any excluded "
-            + "tags. This option affects --build_tests_only behavior and the test command."
+        "Specifies a boolean expression or a list of tags separated by comma. A boolean expression must follow the "
```
alandonovan, line 84: Don't duplicate the comment. Instead, direct the reader to --build_tag_filters. 

(Perhaps better the other way around: I suspect tags are used more for tests than builds.)

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+    }
+
+    /**
+     * Throws an OptionsParsingException when the given expression does not follow the grammar:
+     * expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'
+     */
+    private void checkForForbiddenOperators(Expression expression) throws OptionsParsingException {
```
alandonovan, line 679: Rename to 'validate'.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+      switch (expression.kind()) {
+        case IDENTIFIER:
+          break;
+        case BINARY_OPERATOR:
+          BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
+          if (boe.getOperator() == OR || boe.getOperator() == AND) {
+              checkForForbiddenOperators(boe.getX());
+              checkForForbiddenOperators(boe.getY());
+          } else {
+            throw new OptionsParsingException(String.format("Syntax Error: Got %s but expected 'and' or 'or'",
```
alandonovan, line 689: Use lowercase in error messages. e.g.

"invalid Boolean operator: %s (want 'and' or 'or')"

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
...
+          UnaryOperatorExpression uoe = (UnaryOperatorExpression) expression;
+          if (uoe.getOperator() == TokenKind.NOT) {
+            checkForForbiddenOperators(uoe.getX());
+          } else {
+            throw new OptionsParsingException(String.format("Syntax Error: Got %s but expected 'not'",
+                uoe.getOperator()));
+          }
+          break;
+        default:
+          throw new OptionsParsingException("Syntax Error");
```
alandonovan, line 703: "invalid Boolean formula"

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +635,73 @@ public String getTypeDescription() {
       return "Converts to a CacheBuilderSpec, or null if the input is empty";
     }
   }
+
+  public static class ExpressionConverter implements Converter<Expression> {
```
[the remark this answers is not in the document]
ehkloaj: Do you mean that ExpressionConverter should be moved to [LoadingOptions](https://github.com/bazelbuild/bazel/blob/1fe12e26a4b3d898326385a05b15b9a4408d5454/src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java)? 
The previous solution used CommaSeparatedOptionListConverter which is also a part of [Converters.java](https://github.com/bazelbuild/bazel/blob/1fe12e26a4b3d898326385a05b15b9a4408d5454/src/main/java/com/google/devtools/common/options/Converters.java#L292) so I do not see why this new converter would not belong there aswell.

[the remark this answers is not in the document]
alandonovan: > Do you mean that ExpressionConverter should be moved to LoadingOptions?

Yes, exactly.

> The previous solution used CommaSeparatedOptionListConverter which is also a part of Converters.java so I do not see why this new converter would not belong there as well.

The reason is that Converters is for commonly used converters, but this one is very esoteric and has only a single user (and is likely to have only a single user), so the appropriate place for the complexity is adds is with that user, not with Converters.

src/main/java/com/google/devtools/build/lib/syntax/EmptyExpression.java
```diff
@@ -0,0 +1,27 @@
+package com.google.devtools.build.lib.syntax;
+
+/** An expression without content. */
```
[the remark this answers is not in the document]
ehkloaj: The reason why I chose to use a null object instead of just passing null is that it is a precondition that buildTagFilter is not null in [TargetPatternPhaseValue](https://github.com/bazelbuild/bazel/blob/1fe12e26a4b3d898326385a05b15b9a4408d5454/src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java#L225). Using null would invalidate this. When removing the precondition this results in many test failures. 
Is there any chance you might reconsider? If so do you have a suggestion for the name if you are not pleased with empty?

src/test/java/com/google/devtools/common/options/ExpressionConverterTest.java
ehkloaj: Should I do this by adding a shell script test in [src/test/shell/integration](https://github.com/bazelbuild/bazel/tree/master/src/test/shell/integration)?

src/main/java/com/google/devtools/common/options/Converters.java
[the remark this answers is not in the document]
ehkloaj: Making this class private will not work since the methods **convert** and **getTypeDescription** must be accessible outside of **LoadingOptions**. Do you still want it to be moved but stay public or is it better if it stays in **Converters**?

src/test/java/com/google/devtools/common/options/ExpressionConverterTest.java
ehkloaj: > Making this class private will not work since the methods **convert** and **getTypeDescription** must be accessible outside of **LoadingOptions**. Do you still want it to be moved but stay public or is it better if it stays in **Converters**?

If this class is public should we keep the tests as they are?

ehkloaj:
Thanks for the review @alandonovan!

I have implemented some of your suggestions. I am waiting with the rest until you have had a look at my questions.

src/main/java/com/google/devtools/build/lib/syntax/EmptyExpression.java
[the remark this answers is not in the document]
alandonovan: We absolutely cannot add a new expression type to the language for this: it would be adding complexity to a widely used lower-level package just to support a single esoteric use in a higher-level package, and that is nearly always a mistake. Every existing client of the lower-level package would be immediately broken because their switch(stmt.kind()) statements would become non-exhaustive. So they would have add a case for EmptyExpression, but what should it do? What does empty expression even mean? Where does it come from? Why isn't it mentioned in the language spec? And so on.

(One could avoid null by using some trivially true expression such as "x or not x", but that's still a hack, so let's not do that.)

The obvious solution is simply to remove the non-null precondition check, as it no longer applies. Make clear that this is intentional by annotating the relevant parameters, fields, and accessor methods with @Nullable. Follow the flow of values as far as it goes so that we don't pass null to a function that requires non-null. This particular value doesn't flow very far.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +632,71 @@ public String getTypeDescription() {
       return "Converts to a CacheBuilderSpec, or null if the input is empty";
     }
   }
+
+  public static class BooleanFormulaConverter implements Converter<Expression> {
```
alandonovan, line 636: (I don't know why GitHub's code review tool won't let me continue the previous discussion. Is it because the relevant declaration moved? In any case, not helpful.)

You wrote:
> Making this class private will not work since the methods convert and getTypeDescription must be accessible outside of LoadingOptions. Do you still want it to be moved but stay public or is it better if it stays in Converters?

The class can be private, but its methods must be public because they override public methods defined in a superclass. This is quite normal practice in Java.

Please do move it out of Converters, for the reason I gave previously.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -625,4 +632,71 @@ public String getTypeDescription() {
...
+     * Throws an OptionsParsingException when the given boolean formula does not follow the grammar:
+     * expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'
+     */
+    private void validate(Expression expression) throws OptionsParsingException {
+      switch (expression.kind()) {
+        case IDENTIFIER:
+          break;
+        case BINARY_OPERATOR:
+          BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
+          if (boe.getOperator() == TokenKind.OR || boe.getOperator() == TokenKind.AND) {
```
alandonovan, line 680: Tip: the control structures are slightly easier to read if you reject the error first:

```
   if (op != OR && op != AND) {
            throw ...
    }
    validate(X)
    validate(Y)
    break
```

Ditto for the unary case.

src/main/java/com/google/devtools/common/options/Converters.java
ehkloaj: I think I have misunderstood what you meant then. 
I thought you wanted me to move the class **BooleanFormulaConverter** to **LoadingOptions** like this:
```
public class LoadingOptions extends OptionsBase {
  ...

  @Option(
    name = "build_tag_filters",
    converter = BooleanFormulaConverter.class,
    defaultValue = "",
    documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
    help =
        "See '--test_tag_filters' for instructions. "
            + "This option does not affect the set of tests executed with the 'test' command; "
            + "those are to be governed by the test filtering options, for example '--test_tag_filters'"
  )
  public Expression buildTagFilter;

  @Option(
    name = "test_tag_filters",
    converter = BooleanFormulaConverter.class,
    defaultValue = "",
    documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
    help =
        "Specifies a boolean formula that follows the syntax: "
            + "expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'. "
            + "Only those targets will be built that contain at least one included tag and do not contain any "
            + "excluded tags. This option affects --build_tests_only behavior and the test command."
            + "Example of boolean formula: tag1 and (tag2 or not tag3). "
            + "NOTE: To specify a list of tags separated by comma is legacy syntax and will not be supported "
            + "after Jan 1, 2021."
  )
  public Expression testTagFilter;

  ...

  private static class BooleanFormulaConverter implements Converter<Expression> {
    @Override
    public Expression convert(String input) throws OptionsParsingException {
      if (input.isEmpty()) {
        return Expression.empty();
      }
      if (isLegacySyntax(input)) {
        input = convertLegacySyntaxToBooleanFormula(input);
      }
      try {
        Expression result = Expression.parse(ParserInput.fromLines(input));
        validate(result);
        return result;
      } catch (SyntaxError.Exception e) {
        throw new OptionsParsingException("Failed to parse expression: " + e.getMessage() + " input: " + input, e);
      }
    }

    @Override
    public String getTypeDescription() {
      return "A Boolean formula over test tags (e.g. '(release or smoke) and not flaky')";
    }

    private static boolean isLegacySyntax(String input) {
      return input.contains(",") || input.contains("-");
    }

    /**
     * Converts a string containing tags separated by comma to a boolean expression.
     */
    private String convertLegacySyntaxToBooleanFormula(String input) throws OptionsParsingException {
      return input.replaceAll(" *, *", " or ").replace("-", " not ").trim();
    }

    /**
     * Throws an OptionsParsingException when the given boolean formula does not follow the grammar:
     * expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'
     */
    private void validate(Expression expression) throws OptionsParsingException {
      switch (expression.kind()) {
        case IDENTIFIER:
          break;
        case BINARY_OPERATOR:
          BinaryOperatorExpression boe = (BinaryOperatorExpression) expression;
          if (boe.getOperator() == TokenKind.OR || boe.getOperator() == TokenKind.AND) {
              validate(boe.getX());
              validate(boe.getY());
          } else {
            throw new OptionsParsingException(String.format("invalid Boolean operator: %s (want 'and' or 'or')",
                boe.getOperator()));
          }
          break;
        case UNARY_OPERATOR:
          UnaryOperatorExpression uoe = (UnaryOperatorExpression) expression;
          if (uoe.getOperator() == TokenKind.NOT) {
            validate(uoe.getX());
          } else {
            throw new OptionsParsingException(String.format("invalid Boolean operator: %s (want 'not')",
                uoe.getOperator()));
          }
          break;
        default:
          throw new OptionsParsingException("invalid Boolean operator");
      }
    }
  }
}
```

This leads to a build failure which does not occur when the visibility of **BooleanFormulaConverter** is set to public.

The error message is:
```
ERROR: /repo/ehkloaj/bbi_bazel/bazel/src/main/java/com/google/devtools/build/lib/BUILD:498:1: Executing genrule //src/main/java/com/google/devtools/build/lib:gen_command-line-reference failed (Exit 37): bash failed: error executing command 
  (cd /repo/ehkloaj/bazel_build_root/584b087bddb32230268955ea080f433e/sandbox/processwrapper-sandbox/3904/execroot/io_bazel && \
  exec env - \
    PATH=/app/vbuild/RHEL7-x86_64/python/3.8.0/bin:/app/vbuild/RHEL6-x86_64/jdk/1.8.0_162-1/bin:/opt/tools/tools/vulcanbase/0.74/bin:/app/vbuild/SLED11-x86_64/openjdk/12.0.2/bin:/app/vbuild/RHEL7-x86_64/idea/2019.3.3/bin:/app/vbuild/RHEL7-x86_64/git/2.9.3-1/bin:/app/vbuild/RHEL7-x86_64/git/2.9.3-1/libexec/git-core:/env/sero/bin:/home/ehkloaj/.afs/0/ibin:/usr/afsws/bin:/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/usr/sbin:/sbin:/app/arc/0/bin \
    python3_dir=/app/vbuild/RHEL7-x86_64/python/3.8.0/bin \
  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cat site/command-line-reference-prefix.html > bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html && TMP=`mktemp -d /tmp/tmp.XXXXXXXXXX` && export JARBIN=external/remotejdk11_linux/bin/jar && bazel-out/host/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer --jvm_flag=-Dio.bazel.EnableJni=0 --batch --install_base=${TMP} --output_base=${TMP}/output/ --output_user_root=${TMP} --failure_detail_out=${TMP}/output/failure_detail.rawproto help everything-as-html >> bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html 2>/dev/null && cat site/command-line-reference-suffix.html >> bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html')
Execution platform: //:default_host_platform

Use --sandbox_debug to see verbose messages from the sandbox bash failed: error executing command 
  (cd /repo/ehkloaj/bazel_build_root/584b087bddb32230268955ea080f433e/sandbox/processwrapper-sandbox/3904/execroot/io_bazel && \
  exec env - \
    PATH=/app/vbuild/RHEL7-x86_64/python/3.8.0/bin:/app/vbuild/RHEL6-x86_64/jdk/1.8.0_162-1/bin:/opt/tools/tools/vulcanbase/0.74/bin:/app/vbuild/SLED11-x86_64/openjdk/12.0.2/bin:/app/vbuild/RHEL7-x86_64/idea/2019.3.3/bin:/app/vbuild/RHEL7-x86_64/git/2.9.3-1/bin:/app/vbuild/RHEL7-x86_64/git/2.9.3-1/libexec/git-core:/env/sero/bin:/home/ehkloaj/.afs/0/ibin:/usr/afsws/bin:/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/usr/sbin:/sbin:/app/arc/0/bin \
    python3_dir=/app/vbuild/RHEL7-x86_64/python/3.8.0/bin \
  /bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cat site/command-line-reference-prefix.html > bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html && TMP=`mktemp -d /tmp/tmp.XXXXXXXXXX` && export JARBIN=external/remotejdk11_linux/bin/jar && bazel-out/host/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer --jvm_flag=-Dio.bazel.EnableJni=0 --batch --install_base=${TMP} --output_base=${TMP}/output/ --output_user_root=${TMP} --failure_detail_out=${TMP}/output/failure_detail.rawproto help everything-as-html >> bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html 2>/dev/null && cat site/command-line-reference-suffix.html >> bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/command-line-reference.html')
Execution platform: //:default_host_platform

INFO: Elapsed time: 37.507s, Critical Path: 22.59s
INFO: 478 processes: 106 processwrapper-sandbox, 372 worker.
```

I thought that this was because **BooleanFormulaConverter** is not accessible when the command line reference is generated due to its private visibility but I might be mistaken. 

Also the method **convert** will need to be called outside of the class **LoadingOptions** and I think this will fail if **BooleanFormulaConverter** is private or am I mistaken here as well?

Could you please explain what you meant in more detail, thanks.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -163,4 +175,72 @@
...
+
+  public static class BooleanFormulaConverter implements Converter<Expression> {
+    @Override
+    public Expression convert(String input) throws OptionsParsingException {
+      if (input.isEmpty()) {
+        return null;
+      }
+      if (isLegacySyntax(input)) {
+        input = convertLegacySyntaxToBooleanFormula(input);
+        GoogleLogger.forEnclosingClass().at(Level.WARNING).log(
```
ehkloaj, line 187: I do not see this logging in my environment. Is this the proper way to display warnings? If not what should I use instead?

ehkloaj:
Thanks again @alandonovan.
I have implemented the changes you asked for.

What is left to do is that I need your help with is:
- Add a test for BooleanFormulaConverter through the bazel command. Where should I write this test? https://github.com/bazelbuild/bazel/pull/11900#discussion_r478302168
- Make BooleanFormulaConverter private https://github.com/bazelbuild/bazel/pull/11900#discussion_r478928713
- Display warning to user https://github.com/bazelbuild/bazel/pull/11900#discussion_r479269864

src/main/java/com/google/devtools/common/options/Converters.java
alandonovan: Well, that is a truly terrible error message: it has suppressed the part the really matters. (Does running with --sandbox_debug show you a Java stack trace?)

I tried it locally and it appears that the class must be declared public for it to be accessible to reflection; the "bazel help everything-as-html" command that failed above uses reflection to print all the flags.

So, please keep the class inside LoadingOptions, but declare it public, and precede it by a comment such as this:

/** BooleanFormulaConverter parses a boolean formula to an Expression. Declared public to make it accessible to reflection. */

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
alandonovan: Just use System.err.println.

alandonovan:
The integration test of the new feature can go in third_party/bazel/src/test/shell/bazel/bazel_test_test.sh. All the tests have the form: write some files, run a blaze command, make assertions about the exit code/logs/file system.

src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
```diff
@@ -49,25 +51,25 @@ public static TestFilter forOptions(
...
         ImmutableSet.copyOf(options.testTimeoutFilterSet),
-        ImmutableList.copyOf(options.testTagFilterList),
+        options.testTagFilter,
         ImmutableList.copyOf(options.testLangFilterList));
   }
 
   private final ImmutableSet<TestSize> testSizeFilterSet;
   private final ImmutableSet<TestTimeout> testTimeoutFilterSet;
-  private final ImmutableList<String> testTagFilterList;
+  private final Expression testTagFilter;
```
alandonovan, line 60: `@Nullable private`

src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
```diff
@@ -49,25 +51,25 @@ public static TestFilter forOptions(
...
   private final ImmutableList<String> testLangFilterList;
   private final Predicate<Target> impl;
 
   @AutoCodec.VisibleForSerialization
   TestFilter(
       ImmutableSet<TestSize> testSizeFilterSet,
       ImmutableSet<TestTimeout> testTimeoutFilterSet,
-      ImmutableList<String> testTagFilterList,
+      Expression testTagFilter,
```
alandonovan, line 68: `@Nullable public`

src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
```diff
@@ -107,7 +109,8 @@ public boolean equals(Object o) {
     TestFilter f = (TestFilter) o;
     return f.testSizeFilterSet.equals(testSizeFilterSet)
         && f.testTimeoutFilterSet.equals(testTimeoutFilterSet)
-        && f.testTagFilterList.equals(testTagFilterList)
+        && (f.testTagFilter == null && testTagFilter == null
```
alandonovan, line 112: Objects.equal(f.testTagFilter, testTagFilter)

(it handles null)

src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
```diff
@@ -161,7 +163,7 @@ public static SkyKey key(
       boolean compileOneDependency,
       boolean buildTestsOnly,
       boolean determineTests,
-      ImmutableList<String> buildTargetFilter,
+      Expression buildTargetFilter,
```
alandonovan, line 166: `@Nullable Expression`

src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
```diff
@@ -200,7 +202,7 @@ public static SkyKey keyWithoutFilters(ImmutableList<String> targetPatterns, Str
     private final boolean compileOneDependency;
     private final boolean buildTestsOnly;
     private final boolean determineTests;
-    private final ImmutableList<String> buildTargetFilter;
+    private final Expression buildTargetFilter;
```
alandonovan, line 205: `@Nullable private`

src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
```diff
@@ -254,7 +256,7 @@ public boolean getDetermineTests() {
       return determineTests;
     }
 
-    public ImmutableList<String> getBuildTargetFilter() {
+    public Expression getBuildTargetFilter() {
```
alandonovan, line 259: `@Nullable public`

src/main/java/com/google/devtools/build/lib/skyframe/TargetPatternPhaseValue.java
```diff
@@ -312,7 +314,8 @@ public boolean equals(Object obj) {
           && other.compileOneDependency == compileOneDependency
           && other.buildTestsOnly == buildTestsOnly
           && other.determineTests == determineTests
-          && other.buildTargetFilter.equals(buildTargetFilter)
+          && (other.buildTargetFilter == null && buildTargetFilter == null
```
alandonovan, line 317: Objects.equal

src/main/java/com/google/devtools/common/options/BUILD
```diff
@@ -37,6 +37,7 @@ java_library(
         ],
     ),
     deps = [
+        "//src/main/java/com/google/devtools/build/lib/syntax:frontend",
```
alandonovan, line 40: Revert this.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -20,6 +20,13 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import com.google.devtools.build.lib.syntax.BinaryOperatorExpression;
```
alandonovan, line 23: (Move this.)

src/test/java/com/google/devtools/common/options/BUILD
```diff
@@ -41,6 +41,7 @@ java_test(
     ),
     deps = [
         ":testutils",
+        "//src/main/java/com/google/devtools/build/lib/syntax:frontend",
```
alandonovan, line 44: (Revert this.)

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -163,4 +175,72 @@
...
+
+    private static boolean isLegacySyntax(String input) {
+      return input.contains(",") || input.contains("-");
+    }
+
+    /**
+     * Converts a string containing tags separated by comma to a boolean formula.
+     */
+    private String convertLegacySyntaxToBooleanFormula(String input) throws OptionsParsingException {
+      return input.replaceAll(" *, *", " or ").replace("-", " not ").trim();
```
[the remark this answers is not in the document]
ehkloaj: Yes but I thought it would look nicer in the warning to remove extra spaces or maybe that does not matter?

src/main/java/com/google/devtools/common/options/Converters.java
ehkloaj: > Well, that is a truly terrible error message: it has suppressed the part the really matters. (Does running with --sandbox_debug show you a Java stack trace?)
No, same message.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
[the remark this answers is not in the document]
alandonovan: I see. I don't think it matters. I'll leave it up to you.

(It still boggles my mind that Java's Strings depend on regular expression---and in the innocuously named 'replaceAll', no less.)

alandonovan:
> same message

Would you mind reporting a Bazel bug about the lack of error detail in this
failure? Thanks.

src/test/java/com/google/devtools/common/options/BUILD
ehkloaj: This one is needed.
```
src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java:78: error: [strict] Using type com.google.devtools.build.lib.syntax.Expression from an indirect dependency (TOOL_INFO: "//src/main/java/com/google/devtools/build/lib/syntax:frontend"). See command below **
  public Expression buildTagFilter;
         ^
 ** Please add the following dependencies: 
  //src/main/java/com/google/devtools/build/lib/syntax:frontend to //src/main/java/com/google/devtools/build/lib/pkgcache
```

ehkloaj:
I have fixed all your comments now @alandonovan. I hope it looks alright.
Thank you for all the help!

> Would you mind reporting a Bazel bug about the lack of error detail in this failure? Thanks.

Yes, absolutely.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -56,32 +66,32 @@
 
   @Option(
     name = "build_tag_filters",
```
haxorz, line 68: Both of these flags are documented in the Bazel User Manual (https://docs.bazel.build/versions/master/user-manual.html => bazel/site/docs/user-manual.html)

So, if you want to change the behavior of these flags, please also change the docs in the same commit. Thank you!

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -56,32 +66,32 @@
...
-        "Specifies a comma-separated list of test tags. Each tag can be optionally "
-            + "preceded with '-' to specify excluded tags. Only those test targets will be "
-            + "found that contain at least one included tag and do not contain any excluded "
-            + "tags. This option affects --build_tests_only behavior and the test command."
+        "A Boolean formula over test tags that selects the tests to run. "
+            + "Example: '(release or smoke) and not flaky'. Formulas have the form: "
+            + "expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'. "
+            + "IDENT is defined as a Starlark identifier, with the same reserved words. "
+            + "The legacy syntax, a comma-separated list of tags, implicitly joined as if by 'or', "
+            + "each optionally preceded by '-' for negation, will be removed after Jan 1, 2021."
```
haxorz, line 92: > ... The legacy syntax...  will be removed after Jan 1, 2021.

Sorry, I saw this PR in passing and haven't been following the discussion or motivation. Was this removal discussed and agreed upon? If so, can you please link me? Also, is there a tracking Github issue for the removal? I ask because some things internally at Google use the legacy syntax and they will need to be updated. Thanks!

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
emusand: Thanks for the comment! I pushed a new commit that updates the documentation of these options in bazel/site/docs/user-manual.html.

(@ehkloaj worked as a summer intern at our company, but has now gone back to college and is trying to leave his wild bazel days behind him. I will continue working with this pull request.)

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
emusand: @alandonovan suggested to drop the legacy syntax in a review comment. I don't think it has been discussed anywhere else. Personally, I don't have any strong opinion. For me it would be fine to keep the old syntax as well.

I suppose that the process for incompatible changes needs to be followed here? Will you write the tracking Github issue, or should I do it?

haxorz: Thanks.

I can file the tracking issue once this PR lands.

meisterT:
I see that this PR needs rebasing. Is there anything else that needs to be done before merging?

emusand:
I think that rebasing is the only thing that remains. All review comments have been fixed as far as I know, and Alan has approved the change. @haxorz promised to write a tracking issue that the old value formats for --build_tag_filters and --test_tag_filters soon will be removed (see comments above). Should I rebase, or will you do it?

meisterT:
Please rebase it and I'll import it and send it to Alan for final confirmation.

jin:
Ping to author: please rebase on your end.

emusand:
Thank you for the reminder. Now I have rebased this PR. I am sorry that it took a few days. I first had to find my keyboard, trim my eyebrows and herd my cats. You know how it is.

jin:
Thanks! @meisterT, would you like to import this?

emusand:
Any news about when this PR will be merged? Are there any problems that I can help out with?

meisterT:
@emusand sorry for the delay, there was a problem with our PR import and I lost track of it.

But now I discovered a different problem:
```sh
function test_test_tag_filters_with_legacy_syntax_negative() {
  mkdir -p dir

  touch dir/test.sh
  chmod u+x dir/test.sh
  cat <<'EOF' > dir/BUILD
sh_test(
    name = 'test1',
    srcs = ['test.sh'],
    tags = ['tag1']
)
sh_test(
    name = 'test2',
    srcs = ['test.sh'],
    tags = ['tag1','tag2']
)
sh_test(
    name = 'test3',
    srcs = ['test.sh'],
    tags = ['tag3']
)
EOF
  bazel test //dir/... --test_tag_filters="tag1,-tag2" >& $TEST_log
  expect_log //dir:test1
  expect_not_log //dir:test2
  expect_not_log //dir:test3
}
```
This test passes before your change and fails after it.

emusand:
Aha, thanks for finding this problem @meisterT! I will fix it. (I thought that CI ran this test, but now I see that it has been filtered out.)

meisterT:
>  I thought that CI ran this test, but now I see that it has been filtered out.

Note that the test is not a copy of your test but a slight modification.

emusand:
Now I have fixed the test failure. Thanks again for finding it!

site/docs/user-manual.html
```diff
@@ -1119,13 +1119,36 @@ <h4 id='flag--save_temps'><code class='flag'>--save_temps</code></h4>
...
+  list of tags, optionally preceded with '-' sign used to denote excluded tags. Required tags
+  may also have a preceding '+' sign.
+<pre>
+  % bazel build --test_tag_filters=product,stub,-prototype //myproject:all
+</pre>
+  is equivalent to
+<pre>
+  % bazel test --test_tag_filters='(product or stub) and not prototype' //myproject:all
+</pre>
+  This legacy format will be removed after Jan 1, 2021.
```
meisterT, line 1151: This needs to be removed as per the comment above.

site/docs/user-manual.html
```diff
@@ -1169,29 +1192,40 @@ <h4 id='flag--test_timeout_filters'><code class='flag'>--test_timeout_filters <v
...
+<p>
+  Expressions have the form:
+  <code>expr = IDENT | expr 'or' expr | expr 'and' expr | 'not' expr | '(' expr ')'.</code>
+  <code>IDENT</code> is defined as a Starlark identifier, with the same reserved words.
+</p>
 <p>
-  will test targets that are tagged with either <code>performance</code> or
-  <code>stress</code> tag but are <b>not</b> tagged with the <code>flaky</code>
-  tag.
+  There is also a legacy value format: <code>tag[,tag]*</code>. This expression matches if
```
meisterT, line 1214: This needs to be removed as per the comment above.

site/docs/user-manual.html
```diff
@@ -1169,29 +1192,40 @@ <h4 id='flag--test_timeout_filters'><code class='flag'>--test_timeout_filters <v
...
+  list of tags, optionally preceded with '-' sign used to denote excluded tags. Required tags
+  may also have a preceding '+' sign.
+<pre>
+  % bazel test --test_tag_filters=release,smoke,-flaky //myproject:all
+</pre>
+  is equivalent to
+<pre>
+  % bazel test --test_tag_filters='(release or smoke) and not flaky' //myproject:all
+</pre>
+  This legacy format will be removed after Jan 1, 2021.
```
meisterT, line 1225: This needs to be removed as per the comment above.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
meisterT, line 92: This needs to be removed as per the comment above.

src/main/java/com/google/devtools/build/lib/pkgcache/LoadingOptions.java
```diff
@@ -163,4 +173,75 @@
...
+  public static class BooleanFormulaConverter implements Converter<Expression> {
+    @Override
+    @Nullable public Expression convert(String input) throws OptionsParsingException {
+      if (input.isEmpty()) {
+        return null;
+      }
+      if (isLegacySyntax(input)) {
+        input = convertLegacySyntaxToBooleanFormula(input);
+        System.err.println("Specifying a list of tags separated by comma is legacy syntax and will not be supported " +
+            "after Jan 1, 2021. Consider changing to: " + input);
```
meisterT, line 188: This needs to be removed as per the comment above.

src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
```diff
@@ -107,7 +109,7 @@ public boolean equals(Object o) {
     TestFilter f = (TestFilter) o;
     return f.testSizeFilterSet.equals(testSizeFilterSet)
         && f.testTimeoutFilterSet.equals(testTimeoutFilterSet)
-        && f.testTagFilterList.equals(testTagFilterList)
+        && Objects.equals(f.testTagFilter, testTagFilter)
```
meisterT, line 112: Since Starlark expressions don't have defined equality, can you plumb the original String through instead and use that for comparison?

Then in this class, you can parse the String into a /transient/ Expression. Also you can remove the AutoCodec annotations.

To be able to surface expression errors early, you would need to parse twice but that shouldn't be an issue.

meisterT (commented):
Sorry, during internal review a couple of new comments have come up which I added here.

Also, please update the commit description wrt. to error handling in expression parsing.

emusand:
Thanks for the comments! I have fixed them, but changed one comment in a slightly different way than what you suggested (see above).

src/main/java/com/google/devtools/build/lib/pkgcache/TestFilter.java
alandonovan: Expression strings are not precisely defined, and they abbreviate the full syntax tree. (At a guess, this feature was added so that stack traces could print approximate source syntax of the call stack; that's no longer needed now that we read and print the actual source.) So Expression.toString isn't suitable for this purpose: two different expressions may have the same string.

I suggest using the original expression provided by the user, parsing it once for validation, and again when needed.

I'm sorry the hidden constraints imposed by Google-internal code have made this PR such an ordeal for you. Thanks for your patience.

emusand: Aha, thanks for the clarification! Then I will change to use the original string instead.

emusand:
Now I have fixed the review comment about avoid comparing the Expression objects in TestFilter. Instead of passing the original --build_tag_filters and --test_tag_filters value strings to TestFilter and parse them twice, I found it cleaner to convert these options to a new object, TagFilter. Tell me what you think.

I am sorry that I am slow with pushing new commits. My current work situation only allows me to work with this PR on my spare time.

aiuto:
Friendly ping to reviewers.

elklein:
This PR came to my attention, and it would be really nice to have merged and available. It would help solve some issues that we're having. Looking forward to it!

meisterT:
Sorry, this PR slipped through the cracks, I rebased the PR now, and could move forward once the CLA issue is removed. @ehkloaj can you please have a look?

Kal-Elx:
I'm @ehkloaj, it was the GitHub account I used during my internship last summer. I don't have access to the account anymore and this CLA issue is probably due to my company account being deactivated.

I think @emusand or someone else in the team I worked in can take over here. Either changing ownership of the commit or adding my old email back to the group of authorized contributors.

emusand:
@Kal-Elx, you are correct - your company account has been deleted. Unfortunately I cannot restore your account again.

Is it possible to move the ownership of this pull request to me, without access to the account of the current owner?

meisterT:
@Kal-Elx I think the cleanest way to recover from this is that you rewrite the commit, change the email address of the git author to one that you control and with which you can sign the CLA and then force push to the branch in ehkloaj/bazel. That way your authorship is retained.

Kal-Elx:
I think the easiest way would be for @emusand to just amend the commit with his email and do the push since he has access to my fork and an email with a signed CLA.

emusand:
Ok, I will amend the commit with my email address and push it tomorrow.

emusand:
@googlebot I signed it!

emusand:
Hmm, it looks like the deleted account has confused the google bot. I changed the author of the commit to me, I have signed the CLA in the same name and email address, but googlebot still thinks that the CLA is not signed.

Any ideas on how to proceed?

meisterT:
The bot still wants @ehkloaj to sign the CLA since they send the PR. Probably the easiest is to create a new PR.

emusand:
OK, then I will create a new PR. Thanks for the help!

emusand:
Now I have created a new PR that passes the CLA check: https://github.com/bazelbuild/bazel/pull/13400.

Issue #11842, opened 2020-07-27
Allow safe stripping of Bazel binary


olekw:
### Description of the problem / feature request:

Please consider implementing a means of safely stripping the Bazel binary. I see two possible means by which this might be achieved:

1. Document [strip command](https://manpages.debian.org/unstable/binutils-common/strip.1.en.html) options that would allow debug symbols to be removed without also removing the embedded zip file.

2. Provide (or document, if it exists already) a means by which a "skinny" Bazel binary could be built that would not contain the embedded zip and could therefore be safely stripped. This is the preferred solution from a Debian perspective since we are already packaging Bazel with an extracted system-level installation base that we point to upon execution with `--install_base`. 

### Feature requests: what underlying problem are you trying to solve with this feature?

Stripping release binaries is common practice and even required in many cases for both speed and storage efficiency. For example, this is a [requirement in Debian](https://www.debian.org/doc/debian-policy/ch-files.html#binaries) and produces a [packaging error](https://lintian.debian.org/tags/unstripped-binary-or-object.html) if skipped.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Strip the Bazel binary with `strip --strip-unneeded` or `strip --strip-debug`.

    $ bazel version
    Opening zip "/proc/self/exe": open(): No such file or directory
    FATAL: Failed to open '/proc/self/exe' as a zip file: (error: 2): No such file or directory

### What operating system are you running Bazel on?

Debian Unstable

### What's the output of `bazel info release`?

Same as `bazel version` above.

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

Built as part of Debian package for #9408.

###  Have you found anything relevant by searching the web?

Yes. The stripping problem seem to be a known issue as shown in #600.

olekw:
Following up on Solution 2 above: a simple way of achieving that might be if the C++ launcher skipped looked for the embedded zip when the `--install_base` option is present. That would allow the binary to be stripped, and the embedded zip removed, without impacting functionality or requiring any extensive changes in Bazel.

olekw:
So perhaps a check [like this](https://github.com/bazelbuild/bazel/blob/master/src/main/cpp/blaze.cc#L961) could be extended to the other locations (I believe all in the blaze.cc file) where Bazel tries to read from the embedded archive?

olekw:
[...]

PR #11905, opened 2020-08-07
WIP: Add --trust_install_base option

meteorcloudy:
This will allow Bazel client to run individually without the appended
zip archive, which is a requirement for Debian installation.

The following changes are made:
  - Add --trust_install_base option. When this is enabled. Bazel client
    trusts the install base provided by --install_base option and skip
    the step of verifying the install directory. We need it because the
    Bazel client alone doesn't contain information to verify the install
    directory.
  - Reimplement --version as a actual Bazel startup flag. The version
    info is read from the zip archive, but when --trust_install_base is
    enabled it should be read from the given install base.

Related: https://github.com/bazelbuild/bazel/issues/9408
Fixes: https://github.com/bazelbuild/bazel/issues/11842


olekw:
I can confirm that this PR allows the Bazel binary to be safely stripped and will resolve #11842. I tested with a build and install in the current Debian unstable.

michajlo:
Folding `--version` into regular option handling is something I had wanted to do when first adding it, but I didn't want to go too deep into figuring out/defining how it would get along with the option parsing semantics and the things that can go wrong between the 3rd line of Main and options parsing. Do you have any thoughts on those?

Also, documentation will need to be updated to reflect that `--version` is more lenient now, and we should add at least one test to show how it interacts with other options.

meteorcloudy:
> Folding --version into regular option handling is something I had wanted to do when first adding it, but I didn't want to go too deep into figuring out/defining how it would get along with the option parsing semantics and the things that can go wrong between the 3rd line of Main and options parsing. Do you have any thoughts on those?

So with the implementation in this PR. If the option parsing succeeded and `--version` is passed, Bazel will just print the version info, ignore other flags and possible bazel command, and exits. Comparing to the current behaviour that Bazel complains `--version` is not a recognized flag when other flags are passed, I think it is better.

> Also, documentation will need to be updated to reflect that --version is more lenient now, and we should add at least one test to show how it interacts with other options.

Adding documentation and test sounds a good idea. Is there any original doc or test for `--version` available somewhere?

src/main/cpp/blaze.cc
```diff
@@ -1007,6 +1007,11 @@ static DurationMillis ExtractData(const string &self_path,
           << "install base directory '" << install_base
           << "' could not be created. It exists but is not a directory.";
     }
+
+    if (startup_options.trust_install_base) {
```
meisterT, line 1011: I am a little surprised to see this down here.

What's the expected behavior of this option when no install base exists? I would not assume that it should create one (but it does above).

mcarpenter:
Meanwhile... workaround for debian/ubuntu users whose bazel gets zapped by the prelinker:

```
# echo '-b /usr/bin/bazel-real' >> /etc/prelink.conf
```

(`prelink(8)` is invoked from `/etc/cron.daily/prelink` on Ubuntu 20.04).

meisterT:
Yun, what's the state of this PR?

meteorcloudy:
Sorry, I haven't found time for this recently. But it's on my to-do list. Will take on this ASAP!

PR #11909, opened 2020-08-10
Fixing Issue 11241

### LINKED ISSUE #11241: PATH variable does not contain current directory element (".") on Windows

When running at least the `sh_test` rule on Windows, we get a different PATH variable structure that does not contain `.` for the current directory. On Linux we do have `.` as a leading PATH element so we are able to do thinks in an `sh_test` executed script like the following:

```
#!/bin/bash

cd somedir

foo.sh
```

instead of having to use `./foo.sh`. Other than that sort of trivial example, there could be other use more useful use cases for wanting the current directory on the PATH. The main point of the issue is the fact that the execution environment is different on Linux and Windows is surprising.

See a minimal repro here: https://github.com/greenhouse-org/bazel-issue-repro/tree/481218c7a30560f7c758379ff771cbd2de79081c/sh_test_env_path (note that the repro does take into account a strict action environment and also pulling the PATH from the user's environment, in both cases it seems Bazel is behaving differently between Windows and Linux)

Windows and Linux

`release 3.1.0`

- See reproduction linked above
- We had to work around this in Envoy to support Windows and Linux/macOS here: https://github.com/envoyproxy/envoy/pull/10977
- So far messing around with bash on Linux and MSYS2, I don't have evidence this is a behavior difference between the two, bash on Linux doesn't seem to ever add `.` to the PATH itself but I can't immediately find where that would be happening in Bazel

mai12:
Update tools/test/windows/tw.cc by adding "." to PATH. @meteorcloudy 

Fixes #11241

### DIFF
--- a/src/test/py/bazel/test_wrapper_test.py
+++ b/src/test/py/bazel/test_wrapper_test.py
@@ -111,6 +111,10 @@ class TestWrapperTest(test_base.TestBase):
         '    name = "xml2_test",',
         '    srcs = ["xml2_test.py"],',
         ')',
+		'py_test(',
+		'	name = "add_cur_dir_to_path_test",',
+		'   srcs = ["add_cur_dir_to_path_test.py"],',
+		')'
     ])
 
     self.CopyFile(
@@ -211,6 +215,15 @@ class TestWrapperTest(test_base.TestBase):
             'import os',
             'with open(os.environ.get("XML_OUTPUT_FILE"), "wt") as f:',
             '  f.write("leave this")'
+        ],
+        executable=True)
+		
+    self.ScratchFile(
+        'foo/add_cur_dir_to_path_test.py', [
+            'import os',
+            'path = os.getenv("PATH")',
+            'if ".;" not in path:',
+            '  exit(1)'
         ],
         executable=True)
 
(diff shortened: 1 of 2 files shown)


mai12:
@googlebot I signed it!

meteorcloudy:
Great! Can you also add an integration test for this?
[test_wrapper_test.py](https://github.com/bazelbuild/bazel/blob/b5499b6dc73c63610e302f0110c07c72c0d6962e/src/test/py/bazel/test_wrapper_test.py) may be a good place ;)

src/test/py/bazel/test_wrapper_test.py
```diff
@@ -111,6 +111,10 @@ def _CreateMockWorkspace(self):
         '    name = "xml2_test",',
         '    srcs = ["xml2_test.py"],',
         ')',
+		'py_test(',
```
meteorcloudy: Can you adjust the indentation to be consistent as above?

tools/test/windows/tw.cc
```diff
@@ -409,6 +409,14 @@ bool UnsetEnv(const wchar_t* name) {
   }
 }
 
+bool AddCurrentDirectoryToPATH() {
+  std::wstring PATH;
```
meteorcloudy, line 413: Oh, I overlooked this. Variable name should be lower case.

tools/test/windows/tw.cc
```diff
@@ -409,6 +409,14 @@ bool UnsetEnv(const wchar_t* name) {
   }
 }
 
+bool AddCurrentDirectoryToPATH() {
+  std::wstring path;
+  if (!GetEnv(L"PATH", &path) || !SetEnv(L"PATH", L".;" + path)) {
```
meteorcloudy, line 414: From internal reviewer: 
this is probably more readable as simply "return GetEnv(...) && SetEnv(...);"

PR #11912, opened 2020-08-10
Add action_names filegroup

keith:
This seems to exist internally at google and was imported by rules_swift https://github.com/bazelbuild/rules_swift/pull/461


keith:
Ideally this would probably be a `bzl_library`, but my assumption is we're trying to avoid some sort of cycle there?

meisterT (approved):

keith:
Thanks!

meisterT:
Thank you for the continued contributions!

PR #11914, opened 2020-08-10
Remove apple_binary extension_safe

keith:
This has been marked as deprecated for 3+ years and not doing anything


keith:
Maybe this does something in google? Relevant rules change https://github.com/bazelbuild/rules_apple/pull/865

jin:
This is still pretty widely referenced in the Google codebase, and unfortunately will require a cleanup before this can be merged.

keith:
Ha, ok, nbd, thanks!

PR #11915, opened 2020-08-11
Use definition location for StarlarkRuleFunction.export errors.

benjaminp:
StarlarkRuleFunction.getLocation() always returns BUILTIN.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
@@ -705,7 +705,7 @@ public class StarlarkRuleClassFunctions implements StarlarkRuleFunctionsApi<Arti
         if (attr.hasAnalysisTestTransition()) {
           if (!builder.isAnalysisTest()) {
             throw new EvalException(
-                getLocation(),
+                definitionLocation,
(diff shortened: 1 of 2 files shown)


benjaminp:
@gregestren, @juliexxia this PR makes it easier to figure out where the problem is with transition allowlist errors occur.

juliexxia:
Awesome thank you!

laurentlb:
Have you tried using `Starlark.errorf`?
In many cases, it's no longer needed to set the location explicitly (cc @alandonovan)

benjaminp:
This change is in a `StarlarkExportable.export` implementation, which bubbles into the closure in `BzlLoadFunction.execAndExport` rather than the Starlark interpreter. Probably that interface could be fixed up, but that's a separate change.

alandonovan:
This change seems fine. I plan to abolish the Location parameter to EvalException eventually, but it requires a series of changes. The Location is used only when the exception is thrown and caught in Java code without passing through Starlark calls; in other words, parts of Bazel (ab)use EvalException as a String+Location pair for local error reporting. All such places should either (a) pass an EventHandler down from the (current) catch to the throw, or (b) locally use a private exception that has Location+String.

Unfortunately the changes will require the introduction of several `catch(EvalException){throw LocationError}` statements because so many low-level operators (e.g. Sequence.cast) report their failures as EvalException (sans stack). The challenge is to find the natural seams (or restructure) to minimize the need for them.

PR #11918, opened 2020-08-11
Update windows.md

mai12:
Adding enable 8.3 filename support @meteorcloudy

### DIFF
--- a/site/docs/windows.md
+++ b/site/docs/windows.md
@@ -27,6 +27,12 @@ For example, add the following line to your bazelrc file:
 startup --output_user_root=C:/tmp
 ```
 
+### Enable 8.3 Filename Support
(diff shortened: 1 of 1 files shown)


sluongng:
Circling back to this thread, I wonder if we should still recommend "8.3 filename". Looking at https://github.com/actions/runner-images/pull/9862#issue-2298494856

> 8.3 filename creation is known to hurt file creation performance, especially when there are a large number of similarly named files in the same folder:
> 
> - https://learn.microsoft.com/en-us/archive/blogs/josebda/windows-server-2012-file-server-tip-disable-8-3-naming-and-strip-those-short-names-too
> - https://www.linkedin.com/pulse/dont-forget-disable-short-filenames-83-servers-folders-wes-brown
> - https://deep.data.blog/2013/06/19/debugging-story-slowness-due-to-ntfs-short-file-8-3-name-generation/

meteorcloudy:
Hmm, thanks for pointing this out, I'm fine with removing this recommendation, can you send a PR?

PR #11919, opened 2020-08-11
Fix JacocoLCOVFormatter to emit BRDA lines for branches.

c-mita:
We have to ignore the execution counts parameter because we do not get
that information from Jacoco. The same is true of the DA lines; we
simply record an execution count of "1" if a source line is hit.


c-mita:
Ok, so this is why the unit tests are failing:

This PR depends on fixes made to `coverageoutputgenerator`, but that hasn't been "released" yet. The integration tests use the released version of coverageoutputgenerator, but `java_tools` at HEAD. This will have to wait until the changes made to the LcovParser are in effect.

Issue #11801, opened 2020-07-17
Remote concurrency limited by gRPC connections


EricBurnett:
### Description of the problem / feature request:

For remote RPCs, bazel only opens one connection per server (resolved IP). This constrains the maximum parallelism of a build - gRPC limits the number of in-flight requests per connection to MAX_CONCURRENT_STREAMS, which is a per-server HTTP2 setting, often set to 100 or 128.

In short, this means that when talking to a server resolved to a single IP, bazel can only have ~100 remote executions in flight, regardless of the setting of --jobs. If the server resolves to more IPs, this may be linearly increased, but rarely above 500. Most remote builds can exceed 100x parallelism, and many (esp. builds with long-duration actions) could exceed 500.

Note that due to bazel _displaying_ full parallelism even if actions are actually queued waiting for gRPC connections, it's hard to know just how many users this affects - more than realize it, I'm quite sure. But I've had at least 3 independent projects root cause this reason for limited parallelism at this point, so it seems to be fairly frequent in reality (at least amongst RBE users).

The necessary fix in bazel is to replace the netty 'round_robin' load balancer with one that can open more than 1 connection (subchannel) per resolved address. Maybe by dynamically adding more when bottlenecked, but feeding in a target number of connections as e.g. `max(2, jobs/50)` would probably do just fine.

### Feature requests: what underlying problem are you trying to solve with this feature?

Enable higher build parallelism for remote builds

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run a remote build with --jobs=1000 and observe bazel showing '1000 remote'. Then check server perspective and realize bazel never had more than X00 RPCs in flight. 

### What operating system are you running Bazel on?

Linux (but applies to all)

### What's the output of `bazel info release`?

3.4.1

###  Have you found anything relevant by searching the web?

 - gRPC are not currently planning on offering this feature natively: https://github.com/grpc/grpc/issues/21386
 - Usually resolved in clients by writing a custom load balancer, as is likely the necessary fix for bazel. Example: https://github.com/GoogleCloudPlatform/grpc-gcp-go/tree/master/grpcgcp

cc @buchgr @bergsieker @gkousik @ulfjack

gkousik:
Thanks for filing the bug @EricBurnett !

A sample repo through which you can reproduce the issue from the client side: https://github.com/gkousik/bazel-parallelism-test

[...]

PR #11922, opened 2020-08-12
Display remote action as running only if it is actually executing.

coeuvre:
From #11801 

> due to bazel displaying full parallelism even if actions are actually queued waiting for gRPC connections

This PR makes Bazel display running actions only if it is actually executing remotely as a first step to fix parallelism limitation.

With `--jobs=500`, the progress bar is displayed as:
```
[1 / 501] 500 actions running
```

However, the number of actual running actions by remote workers is not 500 since some actions are pending because of the limitation. With this change, the progress bar now become:
```
[1 / 501] 500 actions, 200 running
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -87,6 +87,14 @@ class GrpcRemoteExecutor {
     return null;
   }
 
+  public ExecuteResponse executeRemotely(ExecuteRequest request) throws IOException, InterruptedException {
+    return executeRemotely(request, null);
+  }
+
+  interface ExecuteOperationNotifier {
+    void onNextOperation(Operation o);
+  }
+
   /* Execute has two components: the Execute call and (optionally) the WaitExecution call.
    * This is the simple flow without any errors:
    *
@@ -105,7 +113,7 @@ class GrpcRemoteExecutor {
    *   are completed and failed; however, some of these errors may be retriable. These errors should
    *   trigger a retry of the Execute call, resulting in a new Operation.
    * */
-  public ExecuteResponse executeRemotely(ExecuteRequest request)
+  public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationNotifier notifier)
       throws IOException, InterruptedException {
     // Execute has two components: the Execute call and (optionally) the WaitExecution call.
     // This is the simple flow without any errors:
@@ -150,6 +158,9 @@ class GrpcRemoteExecutor {
               try {
                 while (replies.hasNext()) {
                   Operation o = replies.next();
+                  if (notifier != null) {
+                    notifier.onNextOperation(o);
+                  }
                   operation.set(o);
                   waitExecution.set(!operation.get().getDone());
                   ExecuteResponse r = getOperationResponse(o);
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -163,14 +163,16 @@ class GrpcRemoteExecutor {
 
                   // Update execution progress to the caller.
                   //
-                  // After called `execute` above, the action is actually waiting for available
+                  // After called `execute` above, the action is actually waiting for an available
                   // gRPC connection to be sent. Once we get a reply from server, we know the
-                  // connection is up and indicate caller the fact by forwarding the `operation`.
+                  // connection is up and indicate to the caller the fact by forwarding the
+                  // `operation`.
                   //
-                  // The accurate execution status of the action relies on server implementation:
+                  // The accurate execution status of the action relies on the server
+                  // implementation:
                   //   1. Server can reply the accurate status in `operation.metadata.stage`;
                   //   2. Server may send a reply without metadata. In this case, we assume the
-                  //      action is accepted by server and will be executed ASAP;
+                  //      action is accepted by the server and will be executed ASAP;
                   //   3. Server may execute the action silently and send a reply once it is done.
                   if (receiver != null) {
                     receiver.onNextOperation(o);
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -92,7 +92,7 @@ class GrpcRemoteExecutor {
   }
 
   interface ExecuteOperationUpdateReceiver {
-    void onNextOperation(Operation o);
+    void onNextOperation(Operation o) throws IOException;
   }
 
   /* Execute has two components: the Execute call and (optionally) the WaitExecution call.
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -158,11 +158,11 @@ class GrpcRemoteExecutor {
               try {
                 while (replies.hasNext()) {
                   Operation o = replies.next();
+                  operation.set(o);
+                  waitExecution.set(!operation.get().getDone());
                   if (receiver != null) {
                     receiver.onNextOperation(o);
                   }
-                  operation.set(o);
-                  waitExecution.set(!operation.get().getDone());
                   ExecuteResponse r = getOperationResponse(o);
                   if (r != null) {
                     return r;
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
@@ -91,7 +91,7 @@ class GrpcRemoteExecutor {
     return executeRemotely(request, null);
   }
 
-  interface ExecuteOperationNotifier {
+  interface ExecuteOperationUpdateReceiver {
     void onNextOperation(Operation o);
   }
 
@@ -113,7 +113,7 @@ class GrpcRemoteExecutor {
    *   are completed and failed; however, some of these errors may be retriable. These errors should
    *   trigger a retry of the Execute call, resulting in a new Operation.
    * */
-  public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationNotifier notifier)
+  public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationUpdateReceiver receiver)
       throws IOException, InterruptedException {
     // Execute has two components: the Execute call and (optionally) the WaitExecution call.
     // This is the simple flow without any errors:
@@ -158,8 +158,8 @@ class GrpcRemoteExecutor {
               try {
                 while (replies.hasNext()) {
                   Operation o = replies.next();
-                  if (notifier != null) {
-                    notifier.onNextOperation(o);
+                  if (receiver != null) {
+                    receiver.onNextOperation(o);
                   }
                   operation.set(o);
                   waitExecution.set(!operation.get().getDone());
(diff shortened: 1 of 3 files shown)


coeuvre:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -152,6 +160,9 @@ public ExecuteResponse executeRemotely(ExecuteRequest request)
                   Operation o = replies.next();
                   operation.set(o);
                   waitExecution.set(!operation.get().getDone());
+                  if (receiver != null) {
+                    receiver.onNextOperation(o);
```
ulfjack, line 164: This is my biggest concern with the change: The service may send replies even when the action is still in the queue. I'm saying that because our service does, and I don't see a constraint on this in the documentation.

(All other comments are optional.)

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -322,9 +321,20 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
                 spawnMetrics.setUploadTime(
                     uploadTime.elapsed().minus(networkTime.getDuration().minus(networkTimeStart)));
               }
+              final AtomicBoolean reportedExecuting =
```
ulfjack, line 324: Put this on one line, please. Technically, the final modifier is unnecessary as of Java support for closures (Java 9?) which automatically capture effectively-final fields.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -322,9 +321,20 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
...
+                  new AtomicBoolean(false);
               ExecuteResponse reply;
               try (SilentCloseable c = prof.profile(REMOTE_EXECUTION, "execute remotely")) {
-                reply = remoteExecutor.executeRemotely(request);
+                reply = remoteExecutor.executeRemotely(request, (operation) -> {
+                  if (!reportedExecuting.get()) {
```
ulfjack, line 329: This is technically a check-then-act race. It is probably not a problem here because the caller is single-threaded (although then, why use AtomicBoolean), but IMO you should get in a habit of using compareAndSet rather than separate get and set calls.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -322,9 +321,20 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
                 spawnMetrics.setUploadTime(
                     uploadTime.elapsed().minus(networkTime.getDuration().minus(networkTimeStart)));
               }
+              final AtomicBoolean reportedExecuting =
+                  new AtomicBoolean(false);
               ExecuteResponse reply;
               try (SilentCloseable c = prof.profile(REMOTE_EXECUTION, "execute remotely")) {
-                reply = remoteExecutor.executeRemotely(request);
+                reply = remoteExecutor.executeRemotely(request, (operation) -> {
```
ulfjack, line 328: My personal preference would be to not use a closure here but an anonymous inner class and move the AtomicBoolean as a field into the class. Feel free to ignore this.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -207,7 +207,6 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
     boolean uploadLocalResults = remoteOptions.remoteUploadLocalResults && spawnCacheableRemotely;
     boolean acceptCachedResult = remoteOptions.remoteAcceptCached && spawnCacheableRemotely;
 
-    context.report(ProgressStatus.EXECUTING, getName());
```
ulfjack, line 210: Maybe change this to PREPARING or similar so that the action shows up in the UI?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -152,6 +160,9 @@ public ExecuteResponse executeRemotely(ExecuteRequest request)
                   Operation o = replies.next();
                   operation.set(o);
                   waitExecution.set(!operation.get().getDone());
+                  if (receiver != null) {
```
ulfjack, line 163: Please add a comment here why you're calling this from here.

coeuvre:
> How will the UI look like if we run into the limitation?

Take the parallelism test [repo](https://github.com/gkousik/bazel-parallelism-test) as an example. With `--jobs=500`, the progress bar is: 
```
[1 / 501] 500 actions running
```
However, the number of actual running actions by remote workers is not `500` since some are pending because of the limitation. It shows 500 because `RemoteSpawnRunner` [report](https://github.com/bazelbuild/bazel/blob/c4dfb91dfe0adc98200e0dcceb617e29eedf65f1/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java#L210) `EXECUTING` status at the beginning of `exec`.

We want to report the `EXEUCTING` status when the remote worker actually executing the action. With this change, the progress bar now become:
```
[1 / 501] 500 actions, 200 running
```

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
coeuvre: Basically, I want to have a way to update caller the remote execution status. So call `receiver.onNextOperation` here after getting a reply from server.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
coeuvre: You are right. I was assuming the action is executing as long as we get a reply from service.

How about use `metadata.stage` to check execution status if the reply from server has [`metadata`](https://github.com/bazelbuild/bazel/blob/4b6ad34dbba15dacebfb6cbf76fa741649cdb007/third_party/remoteapis/build/bazel/remote/execution/v2/remote_execution.proto#L1219) field, and fallback to above strategy otherwise? Since `metadata` is not returned by some services (e.g. RBE), we can't know the accurate execution status in that case. But once we get a reply from service, we know the connection is up and the service should execute the action as soon as possible.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
coeuvre: Good point. I will try this.

coeuvre:
@ulfjack Thanks for you comments! I will update the code soon.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
ulfjack: That sounds like a fine plan!

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
ulfjack: Sorry for being unclear. I would like to see a comment explaining the problem and how this is solving it. As a reader of this code, I'd wonder why this is here rather than before the execute call, and that's specifically to avoid showing an action as executing when it's still queued locally.

coeuvre: > rather than before the execute call

This is the behavior Bazel currently have and causing the problem (showing an action as executing when it's still queued locally).

If we report EXECUTING before the [execute](https://github.com/bazelbuild/bazel/blob/52c8773c3ccbcc9051427d6f86a9fe23217f3977/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java#L148) call, the action might be queued locally waiting for the connection.

By reporting after got a reply from service, we can assume that the connection is up and service will execute the action as soon as possible. However, this assumption might cause other problems if service:
1. only send a reply after the action is done which will leave Bazel showing action as preparing when it's is executing;
2. or send a reply and queue the action. (Can use metadata to tell the status as explained above to avoid this issue)

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
coeuvre: New commit changes this to SCHEDULING so actions waiting for execution are displayed as `[Sched] ...` rather than `[Prepa]`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
meisterT: Ulf means that you should a comment to the code so that in the future people understand why the code is written in this way.

coeuvre: Ops.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -160,9 +160,22 @@ public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationU
                   Operation o = replies.next();
                   operation.set(o);
                   waitExecution.set(!operation.get().getDone());
+
+                  // Update execution progress to the caller.
+                  //
+                  // After called `execute` above, the action is actually waiting for available
```
meisterT, line 166: for _an_ available

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -160,9 +160,22 @@ public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationU
...
+                  // gRPC connection to be sent. Once we get a reply from server, we know the
+                  // connection is up and indicate caller the fact by forwarding the `operation`.
```
meisterT, line 168: indicate _to the_ caller

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -160,9 +160,22 @@ public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationU
...
+                  //
+                  // The accurate execution status of the action relies on server implementation:
```
meisterT, line 170: relies on _the_ server implementation

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutor.java
```diff
@@ -160,9 +160,22 @@ public ExecuteResponse executeRemotely(ExecuteRequest request, ExecuteOperationU
...
+                  //   1. Server can reply the accurate status in `operation.metadata.stage`;
+                  //   2. Server may send a reply without metadata. In this case, we assume the
+                  //      action is accepted by server and will be executed ASAP;
```
meisterT, line 173: accepted by _the_ server

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -199,6 +203,42 @@ public String getName() {
...
+    private SpawnExecutionContext context;
+
+    ExecutingStatusReporter(SpawnExecutionContext context) {
+      this.context = context;
+    }
+
+    @Override
+    public void onNextOperation(Operation o) throws IOException {
+      if (!reportedExecuting) {
+        if (o.hasMetadata()) {
```
ulfjack, line 217: I'd recommend `o.getMetadata().is(ExecuteOperationMetadata.class)` instead or Bazel will crash if the server sends something that isn't this proto.

PR #11937, opened 2020-08-13
Add channel pool for remote execution to overcome gRPC connections limitation.

### LINKED ISSUE #11801: Remote concurrency limited by gRPC connections

coeuvre:
This PR add a `ReferenceCountedChannelPool` which will create `poolSize` number of channels and round-robin across them for gRPC requests.

The `poolSize` is calculated as `jobs / 100`.

Fixes #11801.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
@@ -1,0 +1,101 @@
+package com.google.devtools.build.lib.remote;
+
+import com.google.common.collect.ImmutableList;
+import io.grpc.CallOptions;
+import io.grpc.ClientCall;
+import io.grpc.ManagedChannel;
+import io.grpc.MethodDescriptor;
+import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.ReferenceCounted;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * <p>See {@link ReferenceCounted} for more information about reference counting.
+ */
+public class ReferenceCountedChannelPool extends ReferenceCountedChannel {
+
+  private final AtomicInteger indexTicker = new AtomicInteger();
+  private final ImmutableList<ManagedChannel> channels;
+
+  public ReferenceCountedChannelPool(ImmutableList<ManagedChannel> channels) {
+    super(channels.get(0), new AbstractReferenceCounted() {
+      @Override
+      protected void deallocate() {
+        for (ManagedChannel channel : channels) {
+          channel.shutdown();
+        }
+      }
+
+      @Override
+      public ReferenceCounted touch(Object o) {
+        return null;
+      }
+    });
+    this.channels = channels;
+  }
+
+  @Override
+  public boolean isShutdown() {
+    for (ManagedChannel channel : channels) {
+      if (!channel.isShutdown()) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @Override
+  public boolean isTerminated() {
+    for (ManagedChannel channel : channels) {
+      if (!channel.isTerminated()) {
+        return false;
+      }
+    }
+    return true;  }
+
+  @Override
+  public boolean awaitTermination(long l, TimeUnit timeUnit) throws InterruptedException {
+    long endTimeNanos = System.nanoTime() + timeUnit.toNanos(l);
+    for (ManagedChannel channel : channels) {
+      long awaitTimeNanos = endTimeNanos - System.nanoTime();
+      if (awaitTimeNanos <= 0) {
+        break;
+      }
+      channel.awaitTermination(awaitTimeNanos, TimeUnit.NANOSECONDS);
+    }
+    return isTerminated();
+  }
+
+  @Override
+  public <RequestT, ResponseT> ClientCall<RequestT, ResponseT> newCall(
+      MethodDescriptor<RequestT, ResponseT> methodDescriptor, CallOptions callOptions) {
+    return getNextChannel().newCall(methodDescriptor, callOptions);
+  }
+
+  @Override
+  public String authority() {
+    // Assume all channels have the same authority.
+    return channels.get(0).authority();
+  }
+
+  /**
+   * Performs a simple round robin on the list of {@link ManagedChannel}s in the {@code channels}
+   * list.
+   *
+   * @return A {@link ManagedChannel} that can be used for a single RPC call.
+   */
+  private ManagedChannel getNextChannel() {
+    return getChannel(indexTicker.getAndIncrement());
+  }
+
+  private ManagedChannel getChannel(int affinity) {
+    int index = affinity % channels.size();
+    index = Math.abs(index);
+    // If index is the most negative int, abs(index) is still negative.
+    if (index < 0) {
+      index = 0;
+    }
+    return channels.get(index);
+  }
+}
--- a/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
@@ -70,11 +70,12 @@ public class ReferenceCountedChannelPool extends ReferenceCountedChannel {
         return false;
       }
     }
-    return true;  }
+    return true;
+  }
 
   @Override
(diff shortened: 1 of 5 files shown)


coeuvre:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -255,6 +256,13 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
     ReferenceCountedChannel execChannel = null;
     ReferenceCountedChannel cacheChannel = null;
     ReferenceCountedChannel downloaderChannel = null;
+
+    int poolSize = 1;
+    BuildRequestOptions buildRequestOptions = env.getOptions().getOptions(BuildRequestOptions.class);
+    if (buildRequestOptions != null) {
+      // TODO(chiwang): Should we use a more intelligent way to calculate poolSize?
```
meisterT, line 263: yeah, this is my main question: How did you come up with this number?

gkousik: I believe its based on the recommendation here - https://github.com/bazelbuild/bazel/issues/11801#issuecomment-672973245

meisterT: That makes sense indeed. Chi, can you please document this in the code?

src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
```diff
@@ -0,0 +1,101 @@
+package com.google.devtools.build.lib.remote;
```
meisterT, line 1: Please add the default copyright header here

src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannelPool.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public String authority() {
+    // Assume all channels have the same authority.
+    return channels.get(0).authority();
+  }
+
+  /**
+   * Performs a simple round robin on the list of {@link ManagedChannel}s in the {@code channels}
+   * list.
```
meisterT, line 84: Reference https://github.com/grpc/grpc/issues/21386#issuecomment-564742173 in the comment?

meisterT:
@coeuvre ISTR that Bazel has a RBE pool with 100 executors (I might be mistaken), but I would like you to test this before submission. Perhaps @EricBurnett can help you out with a larger pool.

gkousik:
I just tested it using repro I had previously here - https://github.com/gkousik/bazel-parallelism-test. It simply runs 490 actions in parallel where each action sleeps for 10 seconds. 

With this change (whole build runs in 12s on a 500 worker pool):
```
➜  bazel-parallelism-test git:(master) ✗ ./runbuild.sh
INFO: Invocation ID: eb16917c-3c40-4407-8e58-f2c3973e9e5c
ERROR: /usr/local/google/home/kousikk/.cache/bazel/_bazel_kousikk/07c64333bc485db1757df85f8da29b91/external/bazel_gazelle_go_repository_config/WORKSPACE:3:1: name 'go_repository' is not defined (did you mean 'local_repository'?)
ERROR: /usr/local/google/home/kousikk/.cache/bazel/_bazel_kousikk/07c64333bc485db1757df85f8da29b91/external/bazel_gazelle_go_repository_config/WORKSPACE:8:1: name 'go_repository' is not defined (did you mean 'local_repository'?)
DEBUG: /usr/local/google/home/kousikk/.cache/bazel/_bazel_kousikk/07c64333bc485db1757df85f8da29b91/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:59:14:
Current running Bazel is not a release version and one was not defined explicitly in rbe_autoconfig target. Falling back to '3.2.0'
INFO: Build option --action_env has changed, discarding analysis cache.
DEBUG: /usr/local/google/home/kousikk/.cache/bazel/_bazel_kousikk/07c64333bc485db1757df85f8da29b91/external/bazel_toolchains/rules/rbe_repo.bzl:491:10: Bazel 3.2.0 is used in rbe_default.
INFO: Analyzed 489 targets (6 packages loaded, 500 targets configured).
INFO: Found 489 targets...
INFO: Elapsed time: 12.484s, Critical Path: 11.87s
INFO: 489 processes: 489 remote.
INFO: Build completed successfully, 490 total actions
```

Without this change (whole build takes 35s on a 500 worker pool):
```
➜  bazel-parallelism-test git:(master) ✗ ./runbuild.sh
INFO: Invocation ID: 5e7bc63c-3e9f-42c5-95c4-300a444c4ddb
DEBUG: /usr/local/google/home/kousikk/.cache/bazel/_bazel_kousikk/07c64333bc485db1757df85f8da29b91/external/bazel_toolchains/rules/rbe_repo.bzl:491:10: Bazel 3.2.0 is used in rbe_default.
INFO: Analyzed 489 targets (7 packages loaded, 500 targets configured).
INFO: Found 489 targets...
INFO: Elapsed time: 53.713s, Critical Path: 35.42s
INFO: 489 processes: 489 remote.
INFO: Build completed successfully, 490 total actions
```

coeuvre:
@meisterT  Yes, I have tested before create this PR and got similar result. Thanks for your comments!

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
EricBurnett: Note that the recommendation there is to either take into account the number of IPs (the number of connections, aka subchannels, per channel), or to assume 2 IPs per host which would be jobs/100 "Channels" in this pool.

Looking at this code, just assuming 2 makes more sense to me than having to also reimplement the subchannel->channel interface, which I think would be required to count the IPs. So I vote for simply changing this constant to 100 and documenting clearly why it is what it is.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -255,6 +256,22 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+      // The following calculation is based on the suggestion from comment
+      // https://github.com/bazelbuild/bazel/issues/11801#issuecomment-672973245
+      //
+      // The number of concurrent requests for one connection to a gRPC server is limited by
+      // MAX_CONCURRENT_STREAMS which is normally being 100+. We assume 50 concurrent requests for
+      // each connection should be fairly well. The number of connections opened by one channel is
+      // based on the resolved IPs of that server. We assume servers normally have 2 IPs. So the
+      // number of required channel is calculated as: ceil(jobs / 100).
+      //
+      // TODO(chiwang): Should we use a more intelligent way to calculate poolSize?
```
meisterT, line 272: I think you can remove this TODO.

PR #11938, opened 2020-08-13
Adding support for compiling 32-bit binaries

mai12:
@meteorcloudy

### DIFF
--- a/tools/cpp/BUILD.windows.tpl
+++ b/tools/cpp/BUILD.windows.tpl
@@ -49,11 +49,13 @@ cc_toolchain_suite(
     toolchains = {
         "armeabi-v7a|compiler": ":cc-compiler-armeabi-v7a",
         "x64_windows|msvc-cl": ":cc-compiler-x64_windows",
+        "x64_x86_windows|msvc-cl": ":cc-compiler-x64_x86_windows",
         "x64_windows|msys-gcc": ":cc-compiler-x64_windows_msys",
         "x64_windows|mingw-gcc": ":cc-compiler-x64_windows_mingw",
         "x64_windows|clang-cl": ":cc-compiler-x64_windows-clang-cl",
         "x64_windows_msys": ":cc-compiler-x64_windows_msys",
         "x64_windows": ":cc-compiler-x64_windows",
+        "x64_x86_windows": ":cc-compiler-x64_x86_windows",
         "armeabi-v7a": ":cc-compiler-armeabi-v7a",
     },
 )
@@ -85,8 +87,8 @@ cc_toolchain_config(
     cxx_builtin_include_directories = [%{cxx_builtin_include_directories}],
     tool_paths = {%{tool_paths}},
     tool_bin_path = "%{tool_bin_path}",
-    dbg_mode_debug_flag = "%{dbg_mode_debug_flag}",
-    fastbuild_mode_debug_flag = "%{fastbuild_mode_debug_flag}",
+    dbg_mode_debug_flag = "%{dbg_mode_debug_flag_x64}",
+    fastbuild_mode_debug_flag = "%{fastbuild_mode_debug_flag_x64}",
 )
 
 toolchain(
@@ -131,8 +133,8 @@ cc_toolchain_config(
     tool_bin_path = "%{mingw_tool_bin_path}",
     cxx_builtin_include_directories = [%{mingw_cxx_builtin_include_directories}],
     tool_paths = {%{mingw_tool_paths}},
-    dbg_mode_debug_flag = "%{dbg_mode_debug_flag}",
-    fastbuild_mode_debug_flag = "%{fastbuild_mode_debug_flag}",
+    dbg_mode_debug_flag = "%{dbg_mode_debug_flag_x64}",
+    fastbuild_mode_debug_flag = "%{fastbuild_mode_debug_flag_x64}",
 )
 
 toolchain(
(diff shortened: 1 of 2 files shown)


meteorcloudy:
Thanks for working on this!  A few suggestions first:

- Can you add a brief description about this PR? What it does, how should user use this new feature?
- When creating a new commit, it's better to also have a more detailed message as the title instead of only "Support compiling 32-bit binaries". It will make it easier to review ;)

tools/cpp/windows_cc_configure.bzl
```diff
@@ -415,17 +431,24 @@ def find_msvc_tool(repository_ctx, vc_path, tool):
...
-
+    
+    tools = _get_target_tools(target_arch)
+    for tool_name in tools:
+        if not find_msvc_tool(repository_ctx, vc_path, tools[tool_name], target_arch):
+            missing_tools.append(tools[tool_name])
     return missing_tools
+    
+def _get_target_tools(target):
+    """Return list of required tools names and their filenames for a certain target."""
```
meteorcloudy, line 447: Nit: "Return a list of ..."

meteorcloudy:
https://github.com/bazelbuild/bazel/blob/40e763682ec2b03d3330fd0ffbd0099febbd4140/tools/cpp/windows_cc_toolchain_config.bzl#L501
Another `/MACHINE:X64` is hard coded in windows_cc_toolchain_config.bzl, you probably need to refactor it as an attribute like the `default_link_flags`, otherwise the build will fail with:
```
bazel-out\x64_x86_windows-fastbuild\bin\examples\cpp\_objs\hello-lib\hello-lib.obj : fatal error LNK1112: module machine type 'x86' conflicts with target machine type 
'x64'
Target //examples/cpp:hello-world failed to build
```

tools/cpp/BUILD.windows.tpl
```diff
@@ -49,11 +49,13 @@ cc_toolchain_suite(
     toolchains = {
         "armeabi-v7a|compiler": ":cc-compiler-armeabi-v7a",
         "x64_windows|msvc-cl": ":cc-compiler-x64_windows",
+        "x64_x86_windows|msvc-cl": ":cc-compiler-x64_x86_windows",
```
coeuvre, line 52: Should this be `x86_64_windows`?

meteorcloudy: I think we intend to name the host arch as the first and the target arch as second, which is consistent with the vcvarsall.bat's arch argument.
https://sites.google.com/site/windowsdocumentation/visual-stud/visual-studio-communi/manual-pages/vcvarsall

tools/cpp/windows_cc_configure.bzl
```diff
@@ -725,11 +746,12 @@ def configure_windows_toolchain(repository_ctx):
...
     template_vars = dict()
-    msvc_vars = _get_msvc_vars(repository_ctx, paths)
-    template_vars.update(msvc_vars)
-    template_vars.update(_get_clang_cl_vars(repository_ctx, paths, msvc_vars))
+    msvc_vars_x64 = _get_msvc_vars(repository_ctx, paths, "x64")
+    template_vars.update(msvc_vars_x64)
+    template_vars.update(_get_clang_cl_vars(repository_ctx, paths, msvc_vars_x64))
     template_vars.update(_get_msys_mingw_vars(repository_ctx))
-
+    template_vars.update(_get_msvc_vars(repository_ctx, paths, "x86"))
```
meteorcloudy, line 753: @mai12 Have you noticed a significant longer time when first running Bazel on Windows. It's because we have to run the vcvarsall.bat script twice now, which is quite slow. But from https://github.com/meteorcloudy/my_tests/commit/b6d3bc43b96b635d9045e18d7efcaca048bbf6dd#diff-be9303cb6d3994a5092fee62598e3df8, you can see, the `INCLUDE` env var doesn't change, `PATH` env var only has one more x86 directory at the beginning, `LIB` changes x64 to x86. Since they are quite predictable, can we avoid running the vcvarsall.bat twice to save some time for our users?

meteorcloudy:
@mai12 It looks like you committed with a different account? Can you switch back to the same account?

meteorcloudy:
I think you can do a force push to override the commits pushed by another account?

mai12:
@googlebot I fixed it.

mai12:
@googlebot I fixed it.

tools/cpp/windows_cc_configure.bzl
```diff
@@ -297,7 +299,22 @@ def _is_support_winsdk_selection(repository_ctx, vc_path):
         if vc_common_ide.get_child(tool).exists:
             return True
     return False
+    
+def _get_vc_env_vars(repository_ctx, vc_path, msvc_vars_x64, target_arch):
```
meteorcloudy, line 303: This is a nice solution, can you put a document here to explain what does this function do and why do we need it?

src/test/py/bazel/bazel_windows_cpp_test.py
```diff
@@ -854,7 +854,24 @@ def testCcCompileWithTreeArtifactAsSource(self):
...
+    ])
+    self.ScratchFile('main.cc', [
+        'int main() {',
+        '  return 0;',
+        '}',
+    ])
+    exit_code, _, stderr = self.RunBazel([
+        'build', '-s', '--cpu=x64_x86_windows', '//:main'
+    ])
+    self.AssertExitCode(exit_code, 0, stderr)
```
meteorcloudy, line 874: `-s` will print the command Bazel executed, you can get the content in stderr. Can you check here we are actually using the compiler for 32 bit?  Something like line 773.

tools/cpp/windows_cc_configure.bzl
```diff
@@ -398,34 +416,49 @@ def _get_winsdk_full_version(repository_ctx):
...
-            missing_tools.append(tool)
-
+    
+    tools = _get_target_tools(target_arch)
+    for tool_name in tools:
+        if not find_msvc_tool(repository_ctx, vc_path, tools[tool_name], target_arch):
+            missing_tools.append(tools[tool_name])
     return missing_tools
+    
+def _get_target_tools(target):
```
meteorcloudy, line 456: What if the `target` given is not `x64` and `x86`? Can you throw an error saying the target arch is not recognized?

Also maybe consider refactor the if statement to using:
```
{
    "x64": {"CL": "cl.exe", "LINK": "link.exe", "LIB": "lib.exe", "ML": "ml64.exe"},
    "x86": {"CL": "cl.exe", "LINK": "link.exe", "LIB": "lib.exe", "ML": "ml.exe"},
}
```
?

PR #11939, opened 2020-08-13
Change disk cache to create sub folders for entries

exoson:
Having all of disk cache in a single folder can cause problems with huge
projects on some filesystems. Splitting the cache entries to a folder
part and a file part to fix this.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java
@@ -36,7 +36,8 @@ import javax.annotation.Nullable;
 /** A on-disk store for the remote action cache. */
 public class DiskCacheClient implements RemoteCacheClient {
 
-  private static final String ACTION_KEY_PREFIX = "ac_";
+  private static final String AC_DIRECTORY = "ac";
+  private static final String CAS_DIRECTORY = "cas";
 
   private final Path root;
   private final boolean verifyDownloads;
@@ -60,6 +61,7 @@ public class DiskCacheClient implements RemoteCacheClient {
 
   public void captureFile(Path src, Digest digest, boolean isActionCache) throws IOException {
     Path target = toPath(digest.getHash(), isActionCache);
+    target.getParentDirectory().createDirectoryAndParents();
     src.renameTo(target);
   }
 
@@ -109,7 +111,7 @@ public class DiskCacheClient implements RemoteCacheClient {
   public void uploadActionResult(ActionKey actionKey, ActionResult actionResult)
       throws IOException {
     try (InputStream data = actionResult.toByteString().newInput()) {
-      saveFile(getDiskKey(actionKey.getDigest().getHash(), /* actionResult= */ true), data);
+      saveFile(actionKey.getDigest().getHash(), data, /* actionResult= */ true);
     }
   }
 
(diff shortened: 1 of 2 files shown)


coeuvre:
The code LGTM. But I'm not sure about the strategy for naming the subfolder (take the last 3 characters of digest). @philwo What do you think?

ulfjack:
I've seen this strategy used in other places before, so it's not that unusual. Personally, I prefer using the first few characters rather than the last few.

exoson:
The thing is that all the ac entries start with `ac_` if using 3 first letters that'd mean all the actions will be in the same subfolder. I felt it would make sense to split the ac entries as well and taking last 3 seemed the simplest way of doing it.

coeuvre:
How about using a strategy similar to [bazel-remote](https://github.com/buchgr/bazel-remote/blob/d5141b921ad442627435316f79ccaf227ed2456e/cache/disk/disk.go#L660)?

Basically, it has two folders `ac` and `cas` under cache folder for ac and other entries respectively. Using the first two characters of digest to create subfolders and leave entries inside them (with complete digest as file name).

exoson:
> How about using a strategy similar to [bazel-remote](https://github.com/buchgr/bazel-remote/blob/d5141b921ad442627435316f79ccaf227ed2456e/cache/disk/disk.go#L660)?
> 
> Basically, it has two folders `ac` and `cas` under cache folder for ac and other entries respectively. Using the first two characters of digest to create subfolders and leave entries inside them (with complete digest as file name).

Sounds good. Made changes accordingly.

src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java
```diff
@@ -143,22 +145,31 @@ public void close() {}
     return Futures.immediateFuture(ImmutableSet.copyOf(digests));
   }
 
+  protected Path toPathNoSplit(String key) {
+    return root.getChild(key);
+  }
+
   protected Path toPath(String key, boolean actionResult) {
-    return root.getChild(getDiskKey(key, actionResult));
+    String cacheFolder = actionResult ? "ac" : "cas";
```
coeuvre, line 153: Should use `AC_DIRECTORY` and `CAS_DIRECTORY` respectively.

coeuvre (approved):
LGTM. Thanks!

src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java
exoson: Done

src/main/java/com/google/devtools/build/lib/remote/disk/DiskCacheClient.java
```diff
@@ -60,6 +61,7 @@ public boolean containsActionResult(ActionKey actionKey) {
 
   public void captureFile(Path src, Digest digest, boolean isActionCache) throws IOException {
     Path target = toPath(digest.getHash(), isActionCache);
+    ensureParentExists(target);
```
ulfjack, line 64: Path has a method called createDirectoryAndParents. Use that instead?

exoson: Somehow had missed it. Changed to using that.

PR #11838, opened 2020-07-25
Allow hyphen char in workspace name

davido:
Closes #11837.

Issue #11885, opened 2020-08-03
Release 3.5 - August 2020


meisterT:
Tracking bug for the Bazel 3.5 release. Instructions are at https://github.com/bazelbuild/continuous-integration/blob/master/docs/release-playbook.md

aiuto:
Status update:  Branch cut slightly delayed because of power outages after storms in NYC area. Probably will cut on morning of 2020-08-07

davido:
> Probably will cut on morning of 2020-08-07

This regression seems to be a release blocker: https://github.com/bazelbuild/bazel/issues/11837?

aiuto:
Well. You win. I still have not cut the release because of extenuating
power problems. So the fix in
https://github.com/bazelbuild/bazel/pull/11838 will
be included.

aiuto:
# Status of Bazel 3.5

- Target baseline: 2020-08-03
- Expected release date: 2020-08-18
 [List of release blockers](https://github.com/bazelbuild/bazel/labels/Release%20blocker)

To report a release-blocking bug, please file a bug using the `Release blocker` label, and cc me.

Task list:

- [ ] Update GitHub issues for incompatible changes
- [x] Pick release baseline: f674007557dc73c50a19063d671c662ffa0f423e
- [x ] Create release candidate: 
- [ ] Check downstream projects:
- [ ] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [ ] Send for review the release announcement PR:
- [ ] Push the release, notify package maintainers:
- [ ] Update the documentation
- [ ] Push the blog post
- [ ] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

aiuto:
Currently broken and investgating:
rules_python: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1594#97a7e8ea-c457-4456-a566-84fa4466586c
rules_typescript: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1594#51e3cf05-9c63-4af5-b0ad-b80c2019df80
flogger: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1594#eda26e49-1c79-4ac6-b9a9-284715b51ac6
[...]

PR #11941, opened 2020-08-13
Allow dot char in workspace name

or-shachar:
Continuing https://github.com/bazelbuild/bazel/pull/11838 and adresses https://github.com/bazelbuild/bazel/pull/11838#issuecomment-668435204

> The local_repository rule uses this regex, but says it allows a '.' character, which is currently does not:
```local_repository rule //external:shared.name's name field must be a legal workspace name; workspace names may contain only A-Z, a-z, 0-9, '-', '_' and '.'```

cc: @Gromba


or-shachar:
closing due to https://github.com/bazelbuild/bazel/issues/11885#issuecomment-673549376

PR #11949, opened 2020-08-14
Check for errors from config conditions before checking for missing

katre:
values.

The resolveConfiguredTargetDependencies method currently tracks both
missing values (which will require a skyframe restart) and errors from
dependencies (which should be re-thrown). Unfortunately, it checks this
in the wrong order, which means that when a value is missing and there
is an error, the error will be swallowed and not reported.


katre:
Leaving as a draft while I attempt to write a test.

Issue #11846, opened 2020-07-27
The OSX libtool wrapper increases incremental build times


michaeleisel:
`tools/objc/libtool.sh` increases the incremental build time for a single-target app with 700 implementation files by ~3 seconds. It appears that /usr/bin/libtool takes only a fraction of a second, but the work being done beforehand with symlinks and hashing takes longer. This work is to prevent giving libtool object files with the same basename. A few things:

- libtool seems to just throw a warning, rather than error out. Is there really a problem here?
- If so, which parts of this  can we get rid of? E.g., reproducible hashing, symlinks for all files rather than just the duplicates, etc.

michaeleisel:
@c-parsons can you provide some context on this?

keith:
I would definitely be curious to know what the bug mentioned here is (or was):

https://github.com/bazelbuild/bazel/blob/6326ff803765b08fb09727835827f9d3a73d7d07/tools/objc/libtool.sh#L61-L64

kastiglione:
Last I checked, basenames needs to be unique in Swift modules.

c-parsons:
This change was from several years ago, so take my recollection with a grain of salt.
If I recall correctly, these libtool warnings for same-basename files was problematic for Xcode integration when compiling objective C sources. Given your findings on build performance, we should maybe revisit this.

michaeleisel:
So we could either:
- detect if everything is unique, in which case take a fast path
- only for those that are not unique do this work
- try to optimize the specific operations being done
 
I don't like working in bash, so if I'm implementing it, I'm inclined towards the first option, especially because it couldn't have been working for people in Xcode anyways if they had duplicate basenames. But I'm curious what people would like.

allevato:
cc @trybka who may have some thoughts on improvements to this script.

michaeleisel:
Another alternative that @keith brought up is to provide some ability to just disable this whole thing and run /usr/bin/libtool directly

michaeleisel:
@trybka ping. i'm happy to rewrite the current rule in c++, as i believe that will solve the performance issues here, if you're fine with that as well

trybka:
The warning isn't spurious: It picks one and omits the other, which is ... problematic.

If you use xcodebuild, then it'll name outputs foo.o, foo0.o, etc. which does avoid the problem.

In Bazel, we had to resort to this ... workaround.

The current functionality (hashing/symlinking) is actually in Python: tools/objc/make_hashed_objlist.py
[...]

PR #11958, opened 2020-08-17
Run libtool without symlinking if possible

michaeleisel:
If there are no duplicate .o basenames, then libtool can be invoked directly without symlinking. From https://github.com/bazelbuild/bazel/issues/11846


michaeleisel:
@trybka could you take a look? Thanks!

tools/objc/libtool_check_unique.cc
```diff
@@ -0,0 +1,56 @@
...
+#include <regex>
+#include <fstream>
+#include <unordered_set>
+
+using namespace std;
+
+static unordered_set<string> sBasenames;
+
+string getBasename(string path) {
+    auto idx = path.find_last_of("/");
```
michaeleisel, line 11: not portable to windows, but is this libtool wrapper only used for macos?

tools/objc/libtool.sh
```diff
@@ -111,11 +122,4 @@ while [[ $# -gt 0 ]]; do
...
 done
 
-# Ensure 0 timestamping for hermetic results.
-export ZERO_AR_DATE=1
-
 "${WRAPPER}" libtool "${ARGS[@]}"
-
-# Prevents a pre-Xcode-8 bug in which passing zero-date archive files to ld
-# would cause ld to error.
-touch "$OUTPUTFILE"
```
michaeleisel, line 121: looks like an old fix we would've had to move around, so i just deleted it

michaeleisel:
@trybka is there a way that libtool is currently tested? i didn't see any test rules in that dir

tools/objc/BUILD
```diff
@@ -16,6 +16,14 @@ sh_binary(
     data = [
         ":make_hashed_objlist.py",
         ":xcrunwrapper",
+        ":libtool_check_unique",
+    ],
+)
+
+cc_binary(
+    name = "libtool_check_unique",
```
keith, line 24: I think you need to set this up in the toolchain setup for this to work

tools/objc/libtool.sh
```diff
@@ -34,6 +34,17 @@ fi
...
+# Ensure 0 timestamping for hermetic results.
+export ZERO_AR_DATE=1
+
+if $(/usr/bin/dirname "$0")/libtool_check_unique "$@"; then
+  # If there are no duplicate .o basenames, libtool can be invoked with the original arguments
+  "${WRAPPER}" libtool "$@"
+  exit
+fi
+
+echo "Duplicate .o basenames detected, taking slow path"
```
keith, line 46: Probably want to remove this or this will always get logged for folks who are hitting this, and likely it will just be noise

tools/objc/libtool_check_unique.cc
trybka: I believe this is only used on macOS, but I don't necessarily have visibility into what all other folks do with this. :-)

michaeleisel: i see. what would you suggest here? e.g., is there a way to detect if it's running on apple/linux, and only in that case apply this optimization?

tools/objc/BUILD
michaeleisel: yeah i was wondering if some setup was needed. can you elaborate?

tools/objc/libtool_check_unique.cc
trybka: I honestly don't think it's worth worrying deeply about. Apple's libtool != GNU libtool and I doubt (though I could be wrong, our users are very clever) that anyone is cross-compiling their libtool onto Windows.

Plus there's lots of other stuff that the shell script already invoked that likely doesn't exist (e.g. xcrun).

I would add a comment here and move on.

If you want to check for platform and throw a helpful warning, it should be easy enough to do in the shell script, but I wouldn't bother here in the C++.

michaeleisel: good point about xcrun, yeah i'll just add a comment here

tools/objc/BUILD
keith: well if this works I guess it's fine, but tools like this are normally setup here https://github.com/bazelbuild/bazel/blob/39bc97eab295bddb35b38bfc4a2ff3d2b15d034e/tools/cpp/osx_cc_configure.bzl#L52

michaeleisel:
@keith @kastiglione @trybka responded, curious about toolchain setup as per keith's comment and test practices

trybka:
re: testing

I think your best bet would be to check out the rules_apple repo and run their integration tests:
https://github.com/bazelbuild/rules_apple/tree/master/test

They are quite robust, and would likely identify any issues in both toolchain setup / libtool usage.

re: toolchain setup

I suspect you will have to do something like we do for wrapped_clang.cc (see the link @keith mentioned).
Otherwise you have a dependency loop: you can't use the toolchain you're trying to build prerequisites for to build the prerequisites.

michaeleisel:
@keith that line is added now for the crosstool. how can libtool.sh get the path to the executable for the c++? (currently it just uses dirname of the script itself and appends the executable name for testing purposes)

keith:
If something like that works that is ok, other scripts rely on something similar

michaeleisel:
it does not work afaict :P it just helped me test, but if there's no standard way to do this i can dig a bit

michaeleisel:
@keith it seems like it would work if it was a script, like the python script, but not since it's c++. i see the wrapped_clang test uses `$(rlocation ...)` but that doesn't seem to work, so i guess make variable substitution isn't done here? i'm not sure how to get the location of the binary. i also see that there's mention of wrapped_clang in the rules_cc repo, so i'm curious if i have to make changes there as well.

michaeleisel:
got it now, seems to work. @trybka i tried the rules_apple tests, but i get 52 of 289 tests failed on master as well as with this pr. they fail with `could not get the user's cache directory: $HOME is not defined`

trybka:
@allevato have you seen that failure-mode before?

allevato:
The only other instance I can find of it is here, but it's not clear what resolved it: https://github.com/bazelbuild/tulsi/pull/106

If they fail on master as well as this PR, then there might be an unrelated problem on the Buildkite CI; @philwo may be able to assist or pull in someone who can.

trybka:
OK, all the tests are passing for me locally, I'll patch your commit in and try to test it as well.

michaeleisel:
Awesome, thank you! I'll make an issue for this local test issue

trybka:
err... once I figure out how to patch your changes in. >_> I think this is a problem for tomorrow @trybka

tools/objc/libtool_check_unique.cc
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return path.substr(idx + 1);
+}
+
+// Returns 0 if there are no duplicate basenames in the object files (both via -filelist as well as shell args),
+// 1 otherwise
+int main(int argc, const char * argv[]) {
+    unordered_set<string> basenames;
+    const regex libRegex = regex(".*\\.a$");
+    const regex noArgFlags = regex("-static|-s|-a|-c|-L|-T|-no_warning_for_no_symbols");
```
thii, line 22: `-D` was added in the other PR, so maybe handle that as well.

```suggestion
    const regex noArgFlags = regex("-static|-s|-a|-c|-L|-T|-D|-no_warning_for_no_symbols");
```

michaeleisel: added

michaeleisel:
thanks to @keith i can run rules_apple tests now. they succeed on the latest commit, ad7a8bd

tools/cpp/osx_cc_configure.bzl
```diff
@@ -108,6 +109,34 @@ def configure_osx_toolchain(repository_ctx, overriden_tools):
             paths["@bazel_tools//tools/osx/crosstool:cc_toolchain_config.bzl"],
             "cc_toolchain_config.bzl",
         )
+        libtool_check_unique_src_path = str(repository_ctx.path(
+            paths["@bazel_tools//tools/objc:libtool_check_unique.cc"],
+        ))
+        xcrun_result = repository_ctx.execute([
+            "env",
```
trybka, line 116: pull these into a common copts list or something so they can be shared with wrapped_clang?

michaeleisel: 👍

michaeleisel:
@trybka updated

tools/objc/libtool.sh
```diff
@@ -34,6 +34,15 @@ fi
 
 WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
 
+# Ensure 0 timestamping for hermetic results.
+export ZERO_AR_DATE=1
+
+if $(/usr/bin/dirname "$0")/libtool_check_unique "$@"; then
```
trybka, line 40: Sorry I missed this before--is it possible to refer to this relative to $MY_LOCATION?

Given that this script may be invoked in multiple contexts (according to the logic above), it might be prudent to refer to this in a canonical way.

michaeleisel:
for a 700-file swift_library target, libtool_check_unique takes about 1ms. the overall time seems to be about the same between calling /usr/bin/libtool directly and doing this

tools/objc/libtool.sh
michaeleisel: 👍

michaeleisel:
@trybka responded

trybka:
> for a 700-file swift_library target, libtool_check_unique takes about 1ms. the overall time seems to be about the same between calling /usr/bin/libtool directly and doing this

Cool. How much slower was it before? (with the always hash-symlinking?)

trybka:
> > for a 700-file swift_library target, libtool_check_unique takes about 1ms. the overall time seems to be about the same between calling /usr/bin/libtool directly and doing this
> 
> Cool. How much slower was it before? (with the always hash-symlinking?)

Nevermind, it's in the linked issue.

trybka:
Yeah, I think that should be good. Re-running the import script now.

Issue #6477, opened 2018-10-23
API for Skylark rules and genrules to declare ResourceSets


beasleyr-vmw:
### Description of the problem / feature request:

My project uses multithreaded code emitters written in Java. They may eat CPU and RAM, and when running concurrently may overwhelm the build host. It'd be great if I could somehow signal to Bazel's scheduler that these processes are heavier so that it could adjust accordingly.

### Feature requests: what underlying problem are you trying to solve with this feature?

When porting a project from Maven to Bazel, I observe longer build times with Bazel.  Looking at the critical path, I see that a few actions--most notably one of the code emitters--takes significantly longer under Bazel (23s) than when run from the command line (15s) or Maven.

If I understand the code correctly (GenRuleAction.java), Bazel treats all genrules the same.  They're assumed to consume ~300M of RAM and a single CPU.  My tasks, however, typically occupy 2-4 CPUs, so I'd like Bazel to take this into account when scheduling actions running on my build host.

I'm still too new to Bazel to suggest a concrete API, but maybe something like

```python
genrule(
    ...,
    resources=(512, 3.0, 0.0),  # Corresponds to ResourceSet: 512M, 3 CPUs, 0/default for I/O
)

# or

genrule(
    ...,
    resources_ram=512,
    resources_cpu=3.0,
)    
```

would be feasible? (Ditto for plain `rule`s, too!)

### What operating system are you running Bazel on?

Linux (CentOS 6.6)

### What's the output of `bazel info release`?

```console
$ bazel info release
release 0.16.1- (@non-git)
```

(This is 0.18.0rc8.)

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

n/a. This is just a feature request.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

n/a. This is just a feature request.

###  Have you found anything relevant by searching the web?

#988 touches on a similar topic.

jin:
cc @ulfjack @jmmv

jmmv:
I think this makes a lot of sense. (The need for this came up as well during the Bazel hackathon but I cannot remember the details right now.)

I don't like the first proposal of using a tuple: unnamed parameters are very hard to read. I prefer either a dictionary with named quantities, the separate attributes, or maybe better, a higher-level data type to represent the resource set with named fields (a la `collections.namedtuple`).

[...]

PR #11963, opened 2020-08-18
Implement extra resources for test actions

exoson:
Adds support for specifying extra resources available for the bazel
invocation and requirement for test actions. Using
--local_extra_resources=\<resource name\>=\<amount of resource\> one can
specify the resources available for tests. Specifying a tag
"resources:\<resource name\>:\<amount of resources\>" for a test action will
tell bazel how much of the said resource this test will require. Bazel
will then limit concurrently running test actions using the same
resource if there isn't enough of said resource available.

This can be used to arbitrate access to additional resources such as GPUs or embedded developer boards.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
@@ -25,9 +25,13 @@ import com.google.devtools.build.lib.unix.ProcMeminfoParser;
 import com.google.devtools.build.lib.util.OS;
 import com.google.devtools.build.lib.util.Pair;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 
 /**
@@ -137,6 +141,10 @@ public class ResourceManager {
   // definition in the ResourceSet class.
   private double usedRam;
 
+  // Used amount of extra resources. Corresponds to the extra resource
+  // definition in the ResourceSet class.
+  private Map<String, Float> usedExtraResources;
+
   // Used local test count. Corresponds to the local test count definition in the ResourceSet class.
   private int usedLocalTestCount;
 
@@ -145,6 +153,7 @@ public class ResourceManager {
 
   private ResourceManager() {
     requestList = new LinkedList<>();
+    usedExtraResources = new HashMap<>();
   }
 
   @VisibleForTesting public static ResourceManager instanceForTestingOnly() {
@@ -158,6 +167,7 @@ public class ResourceManager {
   public synchronized void resetResourceUsage() {
     usedCpu = 0;
     usedRam = 0;
+    usedExtraResources = new HashMap<>();
     usedLocalTestCount = 0;
     for (Pair<ResourceSet, CountDownLatch> request : requestList) {
       // CountDownLatch can be set only to 0 or 1.
@@ -177,6 +187,7 @@ public class ResourceManager {
         ResourceSet.create(
             staticResources.getMemoryMb(),
             staticResources.getCpuUsage(),
+            staticResources.getExtraResourceUsage(),
             staticResources.getLocalTestCount());
     processWaitingThreads();
   }
@@ -264,6 +275,16 @@ public class ResourceManager {
   private void incrementResources(ResourceSet resources) {
     usedCpu += resources.getCpuUsage();
     usedRam += resources.getMemoryMb();
+
+    for (Map.Entry<String, Float> resource : resources.getExtraResourceUsage().entrySet()) {
+      String key = (String)resource.getKey();
+      float value = resource.getValue();
+      if (usedExtraResources.containsKey(key)) {
+        value += (float)usedExtraResources.get(key);
+      }
+      usedExtraResources.put(key, value);
+    }
+
     usedLocalTestCount += resources.getLocalTestCount();
   }
 
@@ -271,7 +292,7 @@ public class ResourceManager {
    * Return true if any resources have been claimed through this manager.
    */
   public synchronized boolean inUse() {
-    return usedCpu != 0.0 || usedRam != 0.0 || usedLocalTestCount != 0 || !requestList.isEmpty();
+    return usedCpu != 0.0 || usedRam != 0.0 || !usedExtraResources.isEmpty() || usedLocalTestCount != 0 || !requestList.isEmpty();
   }
 
 
@@ -322,6 +343,13 @@ public class ResourceManager {
   private synchronized boolean release(ResourceSet resources) {
     usedCpu -= resources.getCpuUsage();
     usedRam -= resources.getMemoryMb();
+
+    for (Map.Entry<String, Float> resource : resources.getExtraResourceUsage().entrySet()) {
+      String key = (String)resource.getKey();
+      float value = (float)usedExtraResources.get(key) - resource.getValue();
+      usedExtraResources.put(key, value);
+    }
+
     usedLocalTestCount -= resources.getLocalTestCount();
 
     // TODO(bazel-team): (2010) rounding error can accumulate and value below can end up being
@@ -333,6 +361,19 @@ public class ResourceManager {
     if (usedRam < epsilon) {
       usedRam = 0;
     }
+
+    Set<String> toRemove = new HashSet<>();
+    for (Map.Entry<String, Float> resource : usedExtraResources.entrySet()) {
+      String key = (String)resource.getKey();
+      float value = (float)usedExtraResources.get(key);
+      if (value < epsilon) {
+        toRemove.add(key);
+      }
+    }
+    for (String key : toRemove) {
+      usedExtraResources.remove(key);
+    }
+
     if (!requestList.isEmpty()) {
       processWaitingThreads();
       return true;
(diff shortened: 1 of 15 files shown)


philwo:
I think this is pretty cool. :)

EricCousineau-TRI:
Does this relate #6477 at all?

exoson:
> Does this relate #6477 at all?
> EDIT: Ah, prolly not. That issue's for genrule things.

It seems to me that these are quite similar in the things they achieve. Or at least some of the proposed things.

gregestren:
This PR has been sitting around for a while and deserves proper attention, apologies. 

I don't have enough expertise to feel I know how to review this well. But I should be able to route it to the right place. I'm going to chat with people to get this moving. Stay tuned.

src/main/java/com/google/devtools/build/lib/exec/ExecutionOptions.java
```diff
@@ -351,6 +351,22 @@ public boolean shouldMaterializeParamFiles() {
...
   public boolean localMemoryEstimate;
 
+  @Option(
+      name = "local_extra_resources",
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      allowMultiple = true,
+      help =
+          "Explicitly set the number of extra recourses available to Bazel. "
```
meisterT, line 361: nit: typo in resources

src/main/java/com/google/devtools/build/lib/exec/ExecutionOptions.java
```diff
@@ -351,6 +351,22 @@ public boolean shouldMaterializeParamFiles() {
...
+            + "Takes in a string-float pair. Can be used multiple times to specify multiple "
+            + "types of extra resources. Bazel will limit concurrently running test actions "
+            + "based on the available extra resources and the extra resources required "
+            + "by the test actions. "
+            + "Note: This is a no-op if --local_resources is set.",
```
meisterT, line 366: I wonder whether Bazel should fail when both are specified.

src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
```diff
@@ -322,6 +343,13 @@ private synchronized CountDownLatch acquire(ResourceSet resources) {
   private synchronized boolean release(ResourceSet resources) {
     usedCpu -= resources.getCpuUsage();
     usedRam -= resources.getMemoryMb();
+
+    for (Map.Entry<String, Float> resource : resources.getExtraResourceUsage().entrySet()) {
```
larsrc-google, line 347: This would look nicer as a forEach().

src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
```diff
@@ -333,6 +361,19 @@ private synchronized boolean release(ResourceSet resources) {
     if (usedRam < epsilon) {
       usedRam = 0;
     }
+
+    Set<String> toRemove = new HashSet<>();
+    for (Map.Entry<String, Float> resource : usedExtraResources.entrySet()) {
```
larsrc-google, line 366: You can also use forEach here.

src/main/java/com/google/devtools/build/lib/actions/ResourceSet.java
```diff
@@ -41,12 +44,24 @@
...
 
   private ResourceSet(double memoryMb, double cpuUsage, int localTestCount) {
+    this(memoryMb, cpuUsage, null, localTestCount);
+  }
+
+  private ResourceSet(double memoryMb, double cpuUsage, Map<String, Float> extraResourceUsage, int localTestCount) {
     this.memoryMb = memoryMb;
     this.cpuUsage = cpuUsage;
+    if (extraResourceUsage == null) {
+        this.extraResourceUsage = new HashMap<>();
```
larsrc-google, line 61: Use Collections.emptyMap() instead.

src/main/java/com/google/devtools/build/lib/actions/ResourceSet.java
```diff
@@ -103,16 +118,25 @@ public double getCpuUsage() {
...
+  }
+
   /** Returns the local test count used. */
   public int getLocalTestCount() {
     return localTestCount;
   }
 
   @Override
   public String toString() {
+    StringBuilder sb = new StringBuilder();
```
larsrc-google, line 132: You can use Joiner.MapJoiner for a more readable version.

src/main/java/com/google/devtools/build/lib/actions/ResourceSet.java
```diff
@@ -135,7 +159,7 @@ public ResourceSet convert(String input) throws OptionsParsingException {
         if (memoryMb <= 0.0 || cpuUsage <= 0.0) {
           throw new OptionsParsingException("All resource values must be positive");
         }
-        return create(memoryMb, cpuUsage, Integer.MAX_VALUE);
+        return create(memoryMb, cpuUsage, null, Integer.MAX_VALUE);
```
larsrc-google, line 162: This could be expanded to also take comma-separated key-value pairs of extra resources. The downside would be that typos aren't easy to check, leaving an opening for confusion.

src/main/java/com/google/devtools/build/lib/analysis/test/TestTargetProperties.java
```diff
@@ -150,23 +151,21 @@ public ResourceSet getLocalResourceUsage(Label label, boolean usingLocalTestJobs
     ResourceSet testResourcesFromSize = TestTargetProperties.getResourceSetFromSize(size);
 
     // Tests can override their CPU reservation with a "cpus:<n>" tag.
```
larsrc-google, line 153: These comments are a sign that the relevant portions would prefer to live in separate functions.

src/main/java/com/google/devtools/common/options/Converters.java
```diff
@@ -457,6 +457,31 @@ public String getTypeDescription() {
...
+  /**
+   * A converter for variable assignments from the parameter list of a blaze command invocation.
+   * Assignments are expected to have the form "name=value", where names are defined to
+   * be as permissive as possible and the values should be floats.
+   */
+  public static class StringToFloatAssignmentConverter implements Converter<Map.Entry<String, Float>> {
+
+    @Override
+    public Map.Entry<String, Float> convert(String input) throws OptionsParsingException {
+      int pos = input.indexOf("=");
```
larsrc-google, line 469: Just use a static Splitter instead of hand-coding. Alternatively, make this a subclass of AssignmentConverter. Or even better, make AssignmentConverter a generic method with a converter lambda.

src/main/java/com/google/devtools/build/lib/exec/ExecutionOptions.java
larsrc-google: Doesn't matter, `--local_resources` has been removed since the revision this is based on.

src/main/java/com/google/devtools/build/lib/exec/ExecutionOptions.java
larsrc-google, line 361: You may want to mention that CPU, RAM, and tests cannot be specified here.

src/test/java/com/google/devtools/build/lib/dynamic/DynamicSpawnStrategyTest.java
```diff
@@ -419,7 +419,7 @@ private Spawn newCustomSpawn(String mnemonic, ImmutableMap<String, String> execu
         executionInfo,
         EmptyRunfilesSupplier.INSTANCE,
         action,
-        ResourceSet.create(1, 0, 0));
+        ResourceSet.create(1, 0, null, 0));
```
larsrc-google, line 422: A number of these tests could use the two-parameter overload.

katre:
Is this still under review? What's the status?

src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
scele: Done.

src/main/java/com/google/devtools/build/lib/actions/ResourceManager.java
scele: Done.

src/main/java/com/google/devtools/build/lib/actions/ResourceSet.java
scele: In the latest version I converted the extraResourceUsage from Map to an ImmutableMap, and using `ImmutableMap.of()` here.

src/main/java/com/google/devtools/build/lib/actions/ResourceSet.java
scele: Done.

src/main/java/com/google/devtools/build/lib/analysis/test/TestTargetProperties.java
scele: Done.

src/test/java/com/google/devtools/build/lib/dynamic/DynamicSpawnStrategyTest.java
scele: Looks like many actually can't, because they also specify a nonzero localTestCount. In fact, looks like this was the only place where I could use `ResourceSet.createWithRamCpu` instead.

scele (commented):
@exoson is no longer working with us, I'm picking this up.

A new PR is here: https://github.com/bazelbuild/bazel/pull/13996 - I think this one could be closed.
I've tried to address all comments from @larsrc-google in the new PR.

sgowroji:
We are closing this PR with respect to the comment mentioned over here https://github.com/bazelbuild/bazel/pull/11963#pullrequestreview-751294743 and also the new reference PR (https://github.com/bazelbuild/bazel/pull/13996) . Please feel free to reach to reopen or for any further update. Thanks!

PR #11973, opened 2020-08-19
Support building C++ binaries for ARM and ARM64 architectures on Windows with MSVC toolchain

mai12:
This PR will allow Bazel to compile arm and arm64 c++ build targets on windows. New cc toolchains are added to support compiling these binaries using MSVC compiler. These toolchains are `cc-compiler-x64_arm_windows` and `cc-compiler-x64_arm64_windows` for building arm and arm64 c++ targets respectively.

To use these toolchains, the --cpu option in the build command needs to be set to their corresponding CPU value.

- For arm binaries: --cpu=x64_arm_windows
- For arm64 binaries: --cpu=x64_arm64_windows

@meteorcloudy


meteorcloudy:
@mai12 Hi Mai. The arm and arm64 toolchains should be available on CI now. I took a look at the test failures, I think even we don't have the toolchains available, it shouldn't be so many broken tests. Can you take a closer look at what's going on?

tools/cpp/BUILD.windows.tpl
```diff
@@ -215,6 +222,205 @@ toolchain(
...
+    dwp_files = ":empty",
+    linker_files = ":empty",
+    objcopy_files = ":empty",
+    strip_files = ":empty",
+    supports_param_files = 1,
+)
+
+cc_toolchain_config(
+    name = "msvc_x64_x86",
+    cpu = "x64_windows",
```
meteorcloudy, line 242: I think this should be "x64_x86_windows" ?

tools/cpp/BUILD.windows.tpl
```diff
@@ -215,6 +222,205 @@ toolchain(
...
+    dwp_files = ":empty",
+    linker_files = ":empty",
+    objcopy_files = ":empty",
+    strip_files = ":empty",
+    supports_param_files = 1,
+)
+
+cc_toolchain_config(
+    name = "msvc_x64_arm",
+    cpu = "x64_windows",
```
meteorcloudy, line 308: "x64_arm_windows"?

tools/cpp/BUILD.windows.tpl
```diff
@@ -215,6 +222,205 @@ toolchain(
...
+    dwp_files = ":empty",
+    linker_files = ":empty",
+    objcopy_files = ":empty",
+    strip_files = ":empty",
+    supports_param_files = 1,
+)
+
+cc_toolchain_config(
+    name = "msvc_x64_arm64",
+    cpu = "x64_windows",
```
meteorcloudy, line 374: "x64_arm64_windows"?

tools/cpp/windows_cc_configure.bzl
```diff
@@ -557,80 +610,80 @@ def _get_msvc_vars(repository_ctx, paths):
         write_builtin_include_directory_paths(repository_ctx, "msvc", [], file_suffix = "_msvc")
         msvc_vars = {
             "%{msvc_env_tmp}": "msvc_not_found",
-            "%{msvc_env_path}": "msvc_not_found",
             "%{msvc_env_include}": "msvc_not_found",
```
meteorcloudy, line 613: `"%{msvc_env_include_" + target_arch + "}"`?

meteorcloudy: Also `msvc_env_tmp`, `msvc_cxx_builtin_include_directories` should not be override if arm or arm64 toolchains are not found.

mai12: Will it be better if I added `_{target_arch}` at the end of these variables' names so they won't be overridden by missing toolchains?

meteorcloudy: Yes, I think that makes sense.

tools/cpp/BUILD.windows.tpl
mai12: I will need to change this method `_use_msvc_toolchain` in `windows_cc_toolchain_config.bzl`, to use MSVC toolchains for the new cpu values right?

meteorcloudy: Oh, I see. This is actually not affecting the toolchains selection, but only used in `windows_cc_toolchain_config.bzl`. Since other config like `msys_x64` also uses `x64_windows`, let's just keep it the same here.

tools/cpp/BUILD.windows.tpl
meteorcloudy: No need to change

meteorcloudy:
This is great!

One point I think we can improve is when user try to build with --cpu=x64_arm_windows or --cpu=x64_arm64_windows without the arm toolchains installed. It will fail with the following error:
```
Execution platform: //:default_host_platform

The target you are compiling requires Visual C++ build tools.
Bazel couldn't find a valid Visual C++ build tools installation on your machine.

Visual C++ build tools seems to be installed at C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC
But Bazel can't find the following tools:
    cl.exe, link.exe, lib.exe

Please check your installation following https://docs.bazel.build/versions/master/windows.html#using

Target //examples/cpp:hello-world failed to build
```

It will be a little confusing for users already have x64 toolchain installed, can we mention the exact toolchain we need in the error message?

Also, we probably want to document those new toolchains at https://docs.bazel.build/versions/master/windows.html#using

PR #11977, opened 2020-08-19
Draft: Fix #11424

### LINKED ISSUE #11424: cc_library incorrectly errors out when given empty filegroup

cc_library correctly accepts **srcs** attribute being **None** or set to empty list, as it can be headers only library. Unfortunately when given an empty filegroup it gives the same error as when it sees the file type it does not understand. 

[Here is the repro.](https://github.com/Bazel-snippets/cc_library_repro_1) It shows three cases when cc_library behaves correctly and one when it gives the incorrect error.

Why this error is bad: it does not only happen with the empty filegroup - this is just the easiest repro. When cc_library sources are generated by other rules the list of source files generated can be empty and it is fine. Unfortunately cc_library does not accept it and we have to resort to the ugly workaround of giving it meaningless header file instead of the empty list just to prevent it from stopping the build.

Linux, Mac, Windows

3.1.0

agluszak:


src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
```diff
@@ -2218,8 +2218,7 @@ private void validateDirectPrerequisiteFileTypes(
         // It's not a label or label_list attribute.
         return;
       }
-      if (allowedFileTypes == FileTypeSet.ANY_FILE && !attribute.isNonEmpty()
-          && !attribute.isSingleArtifact()) {
+      if (!attribute.isNonEmpty() && !attribute.isSingleArtifact()) {
```
c-mita, line 2221: I think this is only valid if the set of prerequisite artefacts is _empty_.

agluszak:
#11424 has been closed

PR #11982, opened 2020-08-20
Windows: Remove read-only attribute for directory before deleting

meteorcloudy:
If some directory is set to read-only on Windows, DeletePath will fail
to delete that directory even when no one else is using it.

Previously, we only remove the read-only attribute for files before
deleting, we should also do the same for deleting directory.

Fixes https://github.com/bazelbuild/continuous-integration/issues/1012


laurentlb (approved):
lgtm, thanks

PR #11987, opened 2020-08-21
Make processing of paths with tilde consistent between client and server

chancila:
PathFragmentConverter in Java expands ~ to the user home directory.

Startup arguments that contain paths with `~` processed in client behave strangely compared to arguments processed in Java:

`startup --output_user_root=~/test` actually expands to `WORKSPACE_ROOT/~/test`.

Change AbsolutePathFromFlag to expand `~` suffixes to the user's home directory.


chancila:
friendly ping

Issue #9447, opened 2019-09-27
Missing data_runfiles then use cc_import output


maximyurchuk:
### Description of the problem / feature request:
`cc_import` doesn't expose `data_runfiles` while importing sharing library. So, I can't get needed runtime dependencies for my custom packaging rules.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

See [here](https://github.com/maximyurchuk/bazel-issues/tree/master/use_runfiles_from_cc_import). This example shows, that cc_library have `data_runfiles` with shared library, but cc_import doesn't. 

### What operating system are you running Bazel on?

Windows 10. But originally problem appeared on osx (I just wrote and run repro case on Windows).

### What's the output of `bazel info release`?

c:\git\bazel-issues\use_runfiles_from_cc_import>bazel-latest info release
release 0.28.1

maximyurchuk:
Moreover even explicit `data` parameter doesn't work:

```
cc_import(
	name = "import_library",
	shared_library = "1.dll",
	data = ["1.dll"],
) 
```

So, no idea how to workaround that situation.

hlopko:
I actually don't know what is the difference between normal runfiles and data runfiles. Are shared libraries present in normal runfiles?

cc_library has the data runfiles set for the legacy reasons that nobody in the team remembers, and I never had the time and motivation to dig into that, unfortunately.

maximyurchuk:
@hlopko
> Are shared libraries present in normal runfiles?

I have updated [example](https://github.com/maximyurchuk/bazel-issues/tree/master/use_runfiles_from_cc_import), there are no `runfiles` field for cc_import and cc_library. Seems that this is contradiction with [docs](https://docs.bazel.build/versions/1.0.0/skylark/lib/DefaultInfo.html):
```
runfiles

set of files acting as both the data_runfiles and default_runfiles.
```

> cc_library has the data runfiles set for the legacy reasons

Maybe, its needed because `cc_binary` must be able collect all shared library from `cc_library` dependencies (which can be transitive) while creating binary with `linkstatic = False`. All collected shared libraries are exposed as runfiles for cc_binary. But its just my thoughts, i'm not checked this.

oquenchil:
This was partially rolled back for the native implementation of cc_import:
https://cs.opensource.google/bazel/bazel/+/55436a7e53380466d515836e61cdfee4a4a9dfb5

ozio85:
[...]

PR #11990, opened 2020-08-21
Propagate data runfiles in cc_import

### LINKED ISSUE #11425: cc_library and cc_import do not propagate runfiles from their input to the output

Contrary to reasonable expectations the rules such as **cc_library** and **cc_import** do not propagate the runfiles from their inputs to the output. 

I expected runfiles from the transitive dependencies to propagate through the build graph to dependent targets. For instance, I expected cc_test to collect runfiles from all its direct and transitive dependencies. For some reason I observe that cc_import and cc_library only include runfiles specified in their own **data** attribute, but runfiles which come from **srcs** or **shared_library** are dropped. If this behavior is by design it needs to be documented. If it is not by design I can provide a repro.

Linux, Mac, Windows.

3.1.0

agluszak:
This PR adds data runfiles propagation to both Java and Starlark versions of cc_import.

Fixes #11425, #9447


c-mita (approved):
LGTM

Issue #11967, opened 2020-08-18
Improve configurability documentation


gregestren:
Spawned from https://github.com/bazelbuild/bazel/issues/5574#issuecomment-667299335.

Specific ideas:

* Update [configurable build attributes](https://docs.bazel.build/versions/3.4.0/configurable-attributes.html#custom-keys) docs to reflect Starlark flag support
* Minimize uses of `--define` in doc examples
* Platforms example: why is everything [specified twice](https://docs.bazel.build/versions/master/configurable-attributes.html#platforms)? (from https://github.com/bazelbuild/bazel/issues/5574#issuecomment-667670328)
* Make the difference between `constraint_setting`, `config_setting`, `build_setting`, `platform`, etc. clearer
* State current reality and best practices vs. "this will be possible in the future"
* Clearer, more actionable error messaging

gregestren:
Also review references docs:

- [ ] https://docs.bazel.build/versions/master/be/functions.html#select
- [ ] https://docs.bazel.build/versions/master/be/common-definitions.html#configurable-attributes
- [ ] https://docs.bazel.build/versions/master/be/general.html#config_setting

gregestren:
Also:

- [ ] clarify difference between `cfg = "exec"` and `cfg = "host"`

gregestren:
Also: [beautify error messages](https://github.com/bazelbuild/bazel/issues/11984).

PR #11991, opened 2020-08-21
Update configuration docs to discuss the exec and host configurations…

katre:
…, and how they are different and similar.

Part of improving configurability documentation, #11967.


katre:
Sorry for the delays, I've added a bit on what flags are used for host/exec transitions, and linked directly to the platforms Overview, rather than the entire page.

We might want to add glossary entries for these, but I think that should be done separately.

Issue #11901, opened 2020-08-06
Remote Execution: Bazel should make multiple GetCapabilities requests when using multiple endpoints


ColdrickSotK:
### Description of the problem / feature request:

When using Remote Execution with Bazel, it is possible to specify separate endpoints for the cache and executor. However, Bazel expects the executor's response to `GetCapabilities` requests to contain both `ExecutionCapabilities` and `CacheCapabilities`.

Logically I'd expect Bazel to look for `ExecutionCapabilities` at the executor endpoint, and `CacheCapabilities` at the cache endpoint instead.

This was previously brought up in #7476, which led to bazelbuild/remote-apis#76, clarifying the expected behaviour in the REAPI spec (after some more discussion in bazelbuild/remote-apis#61). It seems like a followup to update Bazel's behaviour slipped through the cracks, since I'm still hitting the original issue when using a remote execution service which doesn't advertise all capabilities in the executor endpoint.

Some extra context on what this looks like is in https://gitlab.com/BuildGrid/buildgrid/-/issues/174.

### What operating system are you running Bazel on?

Ubuntu 19.10

### What's the output of `bazel info release`?

release 3.1.0

PR #11998, opened 2020-08-24
Set --remote_cache to --remote_executor if it is empty.

### LINKED ISSUE #11913: Remote downloader cannot be used with --remote-executor

When trying to use the `--experimental_remote_downloader` flag with remote execution, bazel reports the following error:

```
ERROR: The remote downloader can only be used in combination with gRPC caching
```

`bazel build //... --remote-executor="grpc://localhost:8980" --experimental_remote_downloader="grpc://localhost:8981"`

### What version of bazel are you running?

Release 3.3.1 

Linux

A workaround for this is to also provide --remote_cache="grpc://localhost:8980" assuming your remote execution endpoint is also your cache endpoint.

coeuvre:
Fixes #11913

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -181,8 +181,13 @@ public final class RemoteModule extends BlazeModule {
 
     boolean enableDiskCache = RemoteCacheClientFactory.isDiskCache(remoteOptions);
     boolean enableHttpCache = RemoteCacheClientFactory.isHttpCache(remoteOptions);
-    boolean enableGrpcCache = GrpcCacheClient.isRemoteCacheOptions(remoteOptions);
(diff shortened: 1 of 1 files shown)


ulfjack:
I'm not sure that this is the right fix. Conceptually, combining --remote_cache and --remote_executor would seem to imply that you're using different cache and executor, which is generally probably not what users want. I think it would make more sense to me to disallow having both set at the same time rather than doing this. (Now, that doesn't answer the question of how to fix the linked issue.)

coeuvre:
Pointing `--remote_cache` and `--remote_executor` to different endpoints is allowed currently and I'm not sure whether users are using this feature or not. This PR tries to fix the linked issue without breaking current behavior. How about create another issue describing the problem you mentioned?

coeuvre:
It seems that users are depended on combining `--remote_cache` and `--remote_executor` in some cases. See #11901 and some discussion in bazelbuild/remote-apis#61.

philwo (approved):

PR #11999, opened 2020-08-24
Add new target architectures for c++ binaries built on Windows with MSVC

mai12:
Adding documentation for the newly supported target architectures for c++ binaries built on Windows with MSVC. Bazel now supports building binaries for x86, ARM and ARM64 architectures besides the default x64.

@meteorcloudy

### DIFF
--- a/site/docs/windows.md
+++ b/site/docs/windows.md
@@ -198,7 +198,7 @@ C:\projects\bazel> bazel-bin\examples\cpp\hello-world.exe
 ```
 
 By default the built binaries are targeted to x64 architecture. However, you can specify a different target architecture by setting the ```--cpu``` build option:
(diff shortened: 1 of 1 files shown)


site/docs/windows.md
```diff
@@ -197,6 +197,12 @@ C:\projects\bazel> bazel build //examples/cpp:hello-world
 C:\projects\bazel> bazel-bin\examples\cpp\hello-world.exe
 ```
 
+By default the built binaries are targeted to x64 architecture. However, you can specify a different target architecture by setting the ```--cpu``` build option:
+*   Without ```--cpu``` or with ```--cpu=cpu=x64_windows``` targets are built for x64 architecture.
```
meteorcloudy, line 201: `--cpu=x64_windows`

meteorcloudy (changes requested):
Thanks!

PR #12001, opened 2020-08-24
fix UNC path for Windows/mingw case

jgehw:
(There are cases where not a native python installation on Windows can be used but a mingw-python port inside MSYS2 for collaboration with other tools. mingw-python on MSYS2 returns os.name == 'nt' but doesn't support this UNC path pattern.)


jgehw:
@googlebot I signed it!

jgehw:
@googlebot I fixed it.

jgehw:
@googlebot giving up and doing a new PR.

PR #12002, opened 2020-08-24
fix UNC path for Windows/mingw case

jgehw:
(There are cases where not a native python installation on Windows can be used but a mingw-python port inside MSYS2 for collaboration with other tools. mingw-python on MSYS2 returns os.name == 'nt' but doesn't support this UNC path pattern.)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -22,6 +22,7 @@ import os
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


jgehw:
One use case for this fix is to make tensorflow available via vcpkg.

jgehw:
@lberki do you need further info for reviewing?

lberki:
Erm, sorry. This fell off my radar. Let's merge this.

jgehw:
@lberki Who's going to merge this?  (I can't, insufficient permissions.)

lberki:
Erm, sorry, this got stuck in my queue. Merged now.

PR #12008, opened 2020-08-26
RemoteModule only check required capabilities for a given endpoint

### LINKED ISSUE #11901: Remote Execution: Bazel should make multiple GetCapabilities requests when using multiple endpoints

coeuvre:
According to changes in bazelbuild/remote-apis#76 (and discussion in bazelbuild/remote-apis#61):
> Each endpoint should implement the Capabilities service, and populate the fields relevant to the services available at that endpoint. Clients may choose to ignore irrelevant capabilities if the client does not plan to use a given service on the specified endpoint.

This PR:
1. Refactor `RemoteModule` to use a `ChannelFactory` when creating channels so that tests can mock into channels connected to a fake server.
2. Add tests for verifying server capabilities.
3. Only check required capabilities for a given endpoint.
    - If `--remote_executor` and `--remote_cache point` to the same endpoint, we require that endpoint has both execution and cache capabilities.
    - If they point to different endpoints, we check the endpoint with execution or cache capabilities respectively.

Fixes #11901.


coeuvre:
cc @buchgr @EricBurnett @ulfjack

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -135,7 +145,9 @@ private void verifyServerCapabilities(
       CallCredentials credentials,
       RemoteRetrier retrier,
       CommandEnvironment env,
-      DigestUtil digestUtil)
+      DigestUtil digestUtil,
+      boolean shouldCheckExecutionCapabilities,
+      boolean shouldCheckCacheCapabilities)
```
philwo, line 150: What about replacing this with an enum:

```java
private enum Capabilities {CACHE_ONLY, EXECUTION_ONLY, CACHE_AND_EXECUTION }
```

Then the code below with the various true/false permutations to `verifyServerCapabilities` becomes easier to read.

philwo (approved):

Issue #9153, opened 2019-08-13
objc_generate_linkmap doesn't produce linkmap


keith:
Currently if you `bazel build --objc_generate_linkmap some_target` a link map is not created, and bazel fails because the output is missing.

This appears to be because the `generate_linkmap` feature is not handled in the current macOS crosstool. If this change lands https://bazel-review.googlesource.com/c/bazel/+/104770/5/tools/osx/crosstool/apple_cc_toolchain_config.bzl#1263 I assume it will work correctly.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 0.28.0

tetromino:
@trybka - your upcoming crosstool changes should fix this, right?

trybka:
> @trybka - your upcoming crosstool changes should fix this, right?

In theory,  but I think it's delayed on reviews,  and possibly incompatible change resolution.

keith:
Let me know if there's anything I can do to expedite merging the new crosstool!

jmmv:
@trybka Was the new crosstool ever merged? What's missing to make this happen?

keith:
It did not, that work was dropped

trybka:
Time. I lost my reviewers and the CL has bitrotted.

thii:
@trybka Would you accept a patch to add a `generate_linkmap` feature to the current crosstool?

trybka:
Sure. Should be pretty straightforward, if you have time to add some tests that would be great.

I think this is the crosstool piece:
```
        # Bazel requests this for OBJC actions, when --objc_generate_linkmap is specified.
        feature(name = "generate_linkmap"),

        ...

        flag_set(
            actions = [
                ACTION_NAMES.objc_executable,
                ACTION_NAMES.objcpp_executable,
            ],
            flags = [
                "-Xlinker",
                "-map",
                "-Xlinker",
                "%{linkmap_exec_path}",
            ],
            features = [
                "generate_linkmap",
            ],
        )
```

thii:
Could you give me some pointers on where I should look for to add tests?

trybka:
I would guess some integration test in rules_apple.

Actually, does this work? https://github.com/bazelbuild/rules_apple/blob/master/test/starlark_tests/rules/linkmap_test.bzl

(i.e. does it fail today, and can it be made to succeed with your change?)

thii:
[...]

PR #12016, opened 2020-08-28
Add generate_linkmap feature to Apple CROSSTOOL.

thii:
This adds `generate_linkmap` feature to Apple CROSSTOOL. To generate a
link map, add the `--objc_generate_linkmap` flag to a `bazel build`. By
default, only the top level linkmap file is built. If you want the
linkmap file of the top level target dependencies, add the
`--output_groups=+linkmaps` flag.

Fixes https://github.com/bazelbuild/bazel/issues/9153.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5792,6 +5792,28 @@ def _impl(ctx):
         bitcode_embedded_markers_feature = feature(name = "bitcode_embedded_markers")
         bitcode_embedded_feature = feature(name = "bitcode_embedded")
 
+    generate_linkmap_feature = feature(
+        name = "generate_linkmap",
+        flag_sets = [
+            flag_set(
+                actions = [
+                    ACTION_NAMES.objc_executable,
+                    ACTION_NAMES.objcpp_executable,
+                ],
+                flag_groups = [
+                    flag_group(
+                        flags = [
+                            "-Xlinker",
+                            "-map",
+                            "-Xlinker",
+                            "%{linkmap_exec_path}",
+                        ],
+                    ),
+                ],
+            ),
+        ],
+    )
+
     if (ctx.attr.cpu == "ios_arm64" or
         ctx.attr.cpu == "ios_arm64e" or
         ctx.attr.cpu == "ios_armv7" or
@@ -5816,6 +5838,7 @@ def _impl(ctx):
             default_compile_flags_feature,
             debug_prefix_map_pwd_is_dot_feature,
             generate_dsym_file_feature,
+            generate_linkmap_feature,
             oso_prefix_feature,
             contains_objc_source_feature,
             objc_actions_feature,
@@ -5888,6 +5911,7 @@ def _impl(ctx):
             default_compile_flags_feature,
             debug_prefix_map_pwd_is_dot_feature,
             generate_dsym_file_feature,
+            generate_linkmap_feature,
             oso_prefix_feature,
             contains_objc_source_feature,
             objc_actions_feature,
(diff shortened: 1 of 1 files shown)


thii:
@trybka Can you take a look? Tests were reënabled in https://github.com/bazelbuild/rules_apple/pull/886. The tests are passed now without this change seems to be because they doesn’t really do a `bazel build`. I can add shell tests if you think those aren’t enough.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5792,6 +5792,44 @@ def _impl(ctx):
...
+                        ACTION_NAMES.objc_executable,
+                        ACTION_NAMES.objcpp_executable,
+                    ],
+                    flag_groups = [
+                        flag_group(
+                            flags = [
+                                "-Xlinker",
+                                "-map",
+                                "-Xlinker",
+                                "LINKMAP_EXEC_PATH=%{linkmap_exec_path}",
```
keith, line 5822: Can you pass this path directly instead of having the wrapper expand it?

trybka: This path comes directly from bazel as a build var-- that is, it will already be expanded once it hits the wrapper.

I think you need bazel to tell you the path as it inserts the artifact into the correct output group?

thii: Ah right. Why am I adding a no-op to the wrapper...

Updated. PTAL.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5792,6 +5792,43 @@ def _impl(ctx):
...
+        ctx.attr.cpu == "ios_arm64e" or
+        ctx.attr.cpu == "ios_armv7" or
+        ctx.attr.cpu == "ios_i386" or
+        ctx.attr.cpu == "ios_x86_64" or
+        ctx.attr.cpu == "tvos_arm64" or
+        ctx.attr.cpu == "tvos_x86_64" or
+        ctx.attr.cpu == "watchos_arm64_32" or
+        ctx.attr.cpu == "watchos_armv7k" or
+        ctx.attr.cpu == "watchos_i386" or
+        ctx.attr.cpu == "watchos_x86_64"):
```
keith, line 5807: it looks like this is all cpus? so maybe it shouldn't be in an if?

thii: I wasn’t sure about this. Would it be safe to assume that future added cpus support this feature?

keith: I think so

keith: there are other features not wrapped in ifs, so that seems safe

trybka: It's really more a function of the linker. Unless Apple ships a new CPU that requires a different linker, it should be safe. 

If they do, then you'll have to update more than this stanza.

thii: Thanks. Updated.

thii:
@trybka Could you help import this? You can test this with the integration tests from https://github.com/bazelbuild/rules_apple/pull/886. (Though I’m not sure if you need to import that one since rules_apple seems to prefer analysis tests over integration tests now)

trybka:
Yup, I'll kick this off. I think the import process for rules_apple is sufficiently different that I will not import them at this time.

thii:
Friendly ping - any update? Happy to answer any questions from internal reviews.

I'm hoping this can make it into 3.6 release which is planned to be cut on Monday.

thii:
Thanks a ton!

trybka:
Sorry for the delay! I thought it was submitted, but it was stuck on a merge conflict and I was in meetings for the last 2 days.

PR #12018, opened 2020-08-28
Windows: Make runfiles symlink tree actions depend on the runfiles artifacts.

benjaminp:
Windows distinguishes between symlinks to files and symlinks to directories. The
code to create symlink trees on Windows thus inspects the targets of the links
to learn what kind of symlink to make. Unfortunately, the symlinking code did
not depend on the link targets actually being created. This race could lead to
non-deterministic and non-functional symlink trees.

Fix this problem by depending on runfiles artifacts in the SymtreeTreeAction
when the host platform is Windows.

It's certainly possible to avoid this os-dependent input dependency—by using
in-memory Artifact state for the in-process implementation and extending the
runfiles manifest format to indicate target type for build-runfiles-windows—but
this commit is the most straightforward change that remediates the serious
correctness issue represented by the previous state.

This topic has a long history. When Windows runfiles trees were actually copies
of artifacts, the "symlink" action obviously had to depend on the origin
artifacts (41f4456ac2348bef66739194853a1ddadcbb887e). That code was removed in
an interregnum period where runfiles trees weren't supported at all on Windows
(0885abd851b17d19661dfbd5459a5b91feb45620). However, the dependency was not
added back when Windows symlinked runfiles trees support was finally added
(b592dbd46d5aae2977b11426850eecb89d94a6cb).

Fixes https://github.com/bazelbuild/bazel/issues/12033.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkTreeAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkTreeAction.java
@@ -25,11 +25,12 @@ import com.google.devtools.build.lib.actions.ActionResult;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.analysis.Runfiles;
 import com.google.devtools.build.lib.analysis.config.BuildConfiguration;
+import com.google.devtools.build.lib.collect.nestedset.NestedSet;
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@benjaminp Thanks for fixing this!

nikhilm:
Thanks @benjaminp !

Would appreciate if this could be picked for 3.5.1 (unless 3.5 is magically still a possibility).

PR #12025, opened 2020-08-31
XCode DottedVersion ignore descriptive versions like 10.10.testing

wenyu:


src/test/java/com/google/devtools/build/lib/rules/apple/DottedVersionTest.java
```diff
@@ -46,6 +46,7 @@ public void testCompareTo() throws Exception {
         .addEqualityGroup(DottedVersion.fromString("10.0.0.10A255"))
         .addEqualityGroup(DottedVersion.fromString("10.2"))
         .addEqualityGroup(DottedVersion.fromString("10.2.0.10P99q"))
+        .addEqualityGroup(DottedVersion.fromString("10.10.simulator.internal"), DottedVersion.fromString("10.10"))
```
DavidGoldman, line 49: Would you want to support a build version still?

`10.10.simulator.internal.10A255`

src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
```diff
@@ -129,8 +129,11 @@ public static Option option(DottedVersion version) {
   private static final Splitter DOT_SPLITTER = Splitter.on('.');
   private static final Pattern COMPONENT_PATTERN =
       Pattern.compile("(\\d+)([a-z0-9]*?)?(\\d+)?", Pattern.CASE_INSENSITIVE);
+  private static final Pattern COMMENTIVE_COMPONENT_PATTERN =
```
philwo, line 132: Can we change "commentive" to "descriptive"?

src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
```diff
@@ -129,8 +129,11 @@ public static Option option(DottedVersion version) {
...
+      Pattern.compile("([a-z]\\w*)", Pattern.CASE_INSENSITIVE);
```
philwo, line 133: I'm wondering if we want be more restrictive here by making a list of common patterns, like "experimental", "testing", "internal". My concern is that we'll have some legit version that doesn't start with a number in one of the dotted components and things will break quietly.
If we know that the dotted version components will always start with a number, this is fine as is.

src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
```diff
@@ -129,8 +129,11 @@ public static Option option(DottedVersion version) {
...
   private static final String ILLEGAL_VERSION =
-      "Dotted version components must all be of the form \\d+([a-z0-9]*?)?(\\d+)? but got '%s'";
+      "Dotted version components must all be starting of the form \\d+([a-z0-9]*?)?(\\d+)? " +
```
philwo, line 135: nit: "start with the form"

philwo (changes requested):
I'm relaying @susinmotion's comments from the internal review thread here. :)

src/test/java/com/google/devtools/build/lib/rules/apple/DottedVersionTest.java
wenyu: The descriptive segments will stay by the end of the version number.

src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
wenyu: AFAIK there is no version like that for now. This rule should work.

wenyu:
PTAL. @philwo

wenyu:
And I have Java readability. LOL.

src/main/java/com/google/devtools/build/lib/rules/apple/DottedVersion.java
wenyu: Done

PR #12028, opened 2020-08-31
Add Bazel man page

olekw:


bazel.1
```diff
@@ -0,0 +1,99 @@
+.TH BAZEL 1 "August 2020"
+.SH NAME
+bazel, bazel-3 \- tool to automate software builds and tests
+.SH SYNOPSIS
+.B bazel
+\fI\,<command> <options> \/\fR...
+.SH DESCRIPTION
+.IP
+[bazel release 3.4.0]
```
jin, line 9: This should be autogenerated, right?

bazel.1
```diff
@@ -0,0 +1,99 @@
...
+.IP
+Explains the syntax for specifying targets.
+.IP
+bazel help info\-keys
+.IP
+Displays a list of keys used by the info command.
+.SH "SEE ALSO"
+The full documentation for
+.B bazel
+is maintained as a Texinfo manual.  If the
```
jin, line 91: Where is this?

bazel.1
olekw: It would be nice if it were but I'm not sure how you would do that. We could also put something else here for a description. This was just generic. I was going for something simple that could be expanded upon later. :)

Could use "Build and test software of any size, quickly and reliably" instead.

bazel.1
olekw: Good catch! Nowhere.... Removed and updated the Debian package. Thanks!

bazel.1
meisterT: @olekw will you work on this?

meteorcloudy:
Closing due to inactivity.

Issue #12030, opened 2020-09-01
_rbe_autoconfig failure


agluszak:
### Description of the problem / feature request:

I'm trying to set up RBE with a docker container according to this tutorial https://docs.bazel.build/versions/master/remote-execution-sandbox.html#troubleshooting-in-a-docker-container, but when I try to build any target, I get this error message

```
ERROR: Traceback (most recent call last):
	File "/usr/local/google/home/agluszak/.cache/bazel/_bazel_agluszak/3225c6a145aa13894125b491ab92d176/external/bazel_toolchains/rules/rbe_repo.bzl", line 650, column 34, in <toplevel>
		_rbe_autoconfig = repository_rule(
Error in repository_rule: There is already a built-in attribute 'exec_properties' which cannot be overridden
ERROR: error loading package '': Extension file 'rules/rbe_repo.bzl' has errors
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Here's the repro https://github.com/agluszak/bazel-repros/tree/master/rbe_setup_error (run.sh is the command I use)

Apparently `--experimental_repo_remote_exec` is the culprit

### What operating system are you running Bazel on?

gLinux

### What's the output of `bazel info release`?

version 3.4.1, but it also happens with bazel built from HEAD

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/pull/11029

agluszak:
Fixed by https://github.com/bazelbuild/bazel-toolchains/pull/912

PR #12031, opened 2020-09-01
Make --experimental_repo_remote_exec compatible with rbe_autoconfig

agluszak:
Based on https://github.com/bazelbuild/bazel/pull/11029 by @buchgr

Closes #11029, #12030

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
@@ -27,6 +27,7 @@ import static com.google.devtools.build.lib.packages.Type.STRING_DICT;
 import static com.google.devtools.build.lib.packages.Type.STRING_LIST;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -49,6 +50,8 @@ import com.google.devtools.build.lib.packages.Type;
 import com.google.devtools.build.lib.packages.Type.ConversionException;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
 import com.google.devtools.build.lib.util.FileTypeSet;
+
+import java.util.Map;
 
 /**
  * Rule class definitions used by (almost) every rule.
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
```diff
@@ -343,10 +346,23 @@ public Metadata getMetadata() {
     return builder.add(attr("name", STRING).nonconfigurable("Rule name"));
   }
 
-  public static RuleClass.Builder execPropertiesAttribute(RuleClass.Builder builder)
+  /**
+   * Adds an {@code exec_properties} attribute of type {@code STRING_DICT} if the rule does not
+   * already have one.
+   */
+  public static void addOrOverrideExecPropertiesAttribute(RuleClass.Builder builder)
```
katre, line 353: I think the previous name is fine. It's low-enough use that reading the source or doc to find out what happens when the attribute already exists is fine.

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryModule.java
```diff
@@ -98,6 +97,13 @@ public StarlarkCallable repositoryRule(
         builder.addAttribute(attrDescriptor.build(attrName));
       }
     }
+    if (thread.getSemantics().experimentalRepoRemoteExec()) {
+      // Add the 'exec_properties' attribute last. That way it's backwards compatible with
+      // repository rules that already declare 'exec_properties'. In particular
+      // https://github.com/bazelbuild/bazel-toolchains
+      // TODO(buchgr): Remove this backwards compatiblity hack once the feature is marked stable.
```
katre, line 104: I don't think buchgr is the right person for this anymore. Probably should be me or @juliexxia

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryModule.java
```diff
@@ -79,7 +79,6 @@ public StarlarkCallable repositoryRule(
     builder.addAttribute(attr("$configure", BOOLEAN).defaultValue(configure).build());
     if (thread.getSemantics().experimentalRepoRemoteExec()) {
       builder.addAttribute(attr("$remotable", BOOLEAN).defaultValue(remotable).build());
-      BaseRuleClasses.execPropertiesAttribute(builder);
```
katre, line 82: @juliexxia Is there a reason why repository rules have `exec_properties`? They can't be run remotely and are outside the normal action system, so it seems like it'd be unused. Might be simpler to just remove it from repo rules entirely rather than adding this conditionally.

buchgr:
Hi @agluszak, thanks for picking this up. I think my solution wasn't ideal. I think a better solution is to send a patch to bazel-toolchains and rename their `exec_properties` attribute [1] to `_exec_properties` or similar. It won't be a breaking change for bazel-toolchains because the rbe_autoconfig rule is wrapped in a macro. You'll only rename the internal attribute name.

[1] https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo.bzl#L785

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryModule.java
juliexxia: It looks like @buchgr added this in  https://github.com/bazelbuild/bazel/commit/755e29dfb21d7abfc2bab8dc30813fd5e340d930 with the note "This change is in preparation for adding remote execution capabilities
to repository_context.execute()." Maybe @buchgr has more context?

agluszak:
Fixed by https://github.com/bazelbuild/bazel-toolchains/pull/912

PR #12039, opened 2020-09-02
Update Python rules to use the toolchain transition

katre:
Builds on #12038, must be submitted after.

This also requires a full downstream test and probably some internal work as well.


katre:
Ran downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1624

Failures are the same as the latest nightly: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1623

Issue #12040, opened 2020-09-02
Release 3.6 - September 2020


laurentlb:
# Status of Bazel 3.6

- Target baseline: 2020-09-10
- Expected release date: 2020-10-06
- [List of release blockers](https://github.com/bazelbuild/bazel/labels/Release%20blocker)

To report a release-blocking bug, please file a bug using the `Release blocker` label, and cc me.

Task list:

- [x] Update GitHub issues for incompatible changes
- [x] Pick release baseline: aa0d97c0bfc4c09ec6f45303aa80052ba28afbd9
- [x] Create release candidate: https://releases.bazel.build/3.6.0/rc2/index.html
- [x] Check downstream projects: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1681
- [x] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [x] Send for review the release announcement PR: https://github.com/bazelbuild/bazel-blog/pull/239
- [x] Push the release, notify package maintainers:
- [x] Update the documentation
- [x] Push the blog post
- [x] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

keith:
If possible I think we should cherry pick https://github.com/bazelbuild/bazel/pull/12046 once it lands. Without it in sandboxed builds on darwin you get log spew for every static archive action:

```
external/local_config_cc/libtool: line 40: /private/var/tmp/_bazel_iosci/4988be46b65c576ca5906aacc315f576/sandbox/darwin-sandbox/333/execroot/lyftios/external/local_config_cc/libtool_check_unique: No such file or directory
external/local_config_cc/libtool: line 40: /private/var/tmp/_bazel_iosci/4988be46b65c576ca5906aacc315f576/sandbox/darwin-sandbox/375/execroot/lyftios/external/local_config_cc/libtool_check_unique: No such file or directory
external/local_config_cc/libtool: line 40: /private/var/tmp/_bazel_iosci/4988be46b65c576ca5906aacc315f576/sandbox/darwin-sandbox/376/execroot/lyftios/external/local_config_cc/libtool_check_unique: No such file or directory
[...]

PR #12046, opened 2020-09-04
Fix issue where libtool_check_unique isn't found for sandbox builds

michaeleisel:
It seems that for sandbox builds, the executable isn't copied over. libtool still succeeds, but it prints out an error message about the missing executable. This change, which I just made to mimic wrapped_clang, appears to fix it. I have no idea if there are other changes that should also be made for this, e.g. in tools/osx/crosstool/cc_toolchain_config.bzl wrapped_clang is referenced a lot but libtool_check_unique is not.


michaeleisel:
@trybka @keith FYI

once it looks good, i can run tests with rules_apple

keith (approved):

keith:
I'm also not sure if there are other places this is needed. Probably worth grepping or doing a rdeps query on libtool itself, but this seems like a solid fix and likely the only other place

keith:
Maybe worth editing the script to error if this is missing vs treating that as the same type of failure?

michaeleisel:
in a way, i'm glad the script kept working without it. but if we feel confident enough in this change, or at least that doing a simple build with + without the sandbox is enough to test it properly, i can have it fail early

trybka:
There's also https://github.com/bazelbuild/bazel/blob/master/tools/objc/BUILD#L14 ? 

(That is referenced in java code here: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/objc/ObjcRuleClasses.java#L531)

Alternatively, unless we're certain that we have all the uses necessary, we could also make the script warn and continue gracefully if it is not present--considering it is just an optimization.

keith (approved):

tools/objc/BUILD
```diff
@@ -16,9 +16,15 @@ sh_binary(
     data = [
         ":make_hashed_objlist.py",
         ":xcrunwrapper",
+        ":libtool_check_unique",
     ],
 )
 
+filegroup(
```
keith, line 23:
```suggestion
cc_library(
```

I think we actually need the binary here?

michaeleisel:
@trybka @keith updated, does this look good? it passes rules_apple tests for me. i'm not sure if i did the new rule right

michaeleisel:
to be clear, i don't know when that missing rule in tools/objc/BUILD would cause a problem anyways

tools/osx/crosstool/BUILD.tpl
```diff
@@ -50,6 +50,7 @@ cc_toolchain_suite(
             ":builtin_include_directory_paths",
             ":cc_wrapper",
             ":libtool",
+            ":libtool_check_unique",
```
keith, line 53: +1 on this one, seems like this label references the binary in the toolchain setup

tools/objc/BUILD
```diff
@@ -16,9 +16,15 @@ sh_binary(
     data = [
         ":make_hashed_objlist.py",
         ":xcrunwrapper",
+        ":libtool_check_unique",
     ],
 )
 
+cc_binary(
+    name = "libtool_check_unique",
+    srcs = ["libtool_check_unique.cc"],
```
keith, line 25: I can't figure out how this stuff is actually used, so I can't tell if it works or not.

keith (approved):

tools/objc/BUILD
trybka: Doing some archaeology, I think this was a remnant of the pre-crosstool ObjcCompilation rule implementations. Which is presumably why you can't find it. 

I bet this can actually be removed, but I am not knowledgeable enough about the Bazel CI to know if we'd catch any issues.

keith: related https://github.com/bazelbuild/bazel/pull/12073/

I do think some of it looks more used in j2objc stuff tho

trybka: It looks like they pass it along in the J2ObjcAspect, but I can't find any references to it being used after that. Did you catch something else? (We can also take this to your other PR)

keith: sounds like we can pretty safely ignore this piece for this use case and fix it if anything comes up, so I think this PR should be good to go!

michaeleisel: so then i can leave it as-is and we can merge? or does it need to be deleted?

tools/objc/libtool.sh
```diff
@@ -37,7 +37,9 @@ WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
 # Ensure 0 timestamping for hermetic results.
 export ZERO_AR_DATE=1
 
-if "${MY_LOCATION}"/libtool_check_unique "$@"; then
+if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
```
michaeleisel, line 40: my bash is not fantastic, but this at least works for me

tools/objc/BUILD
trybka: I would delete it for now unless we can prove it is needed otherwise.

keith: Up to whoever can merge, but unless mine merges first I do think this should work, so we could leave it to try and get this in 3.6, and I can remove it if it's really unused.

tetromino:
@michaeleisel, thanks for the PR. We'd really like to have a test for what exactly this PR fixes to make sure that it stays fixed. Would you mind either writing a test or providing a small case to reproduce the problem so that we can add a test based on that?

michaeleisel:
@tetromino sure, here's a test case. if we have a folder with the following three files:
BUILD.bazel
```
objc_library(
    name = "a",
    srcs = ["a.c"],
)
```
a.c
```
void a() {
}
```
WORKSPACE (empty)
```
```
if we run `bazel build a` without this fix, the resulting output for a clean build will include an error message along the lines of
```
INFO: From Linking liba.a:
external/local_config_cc/libtool: line 40: /private/var/tmp/_bazel_meisel/1054548cf645506d169069fe6cc12477/sandbox/darwin-sandbox/8/execroot/__main__/external/local_config_cc/libtool_check_unique: No such file or directory
```
e.g., a test could check for `libtool_check_unique: No such file or directory` not being present

michaeleisel:
@keith to resolve the merge conflict i just undid any changes for that file

keith (approved):

keith:
a problem with a test case is it doesn't fail, it just follows the slow path

michaeleisel:
@tetromino curious what your thoughts are

tetromino:
@michaeleisel, I'm really sorry for the delayed response. Unfortunately, I cannot reproduce the problem (the error message you are seeing) using bazel-3.5.0 on my Mac (macOS 10.15.6, Xcode 11.6) and the example you gave. I see the following:

```
$ cat BUILD.bazel
objc_library(
    name = "a",
    srcs = ["a.c"],
)
$ cat WORKSPACE
$ cat a.c
void a() {
}
$ bazelisk clean --expunge
INFO: Starting clean.
$ bazelisk info
bazel-bin: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/execroot/__main__/bazel-out/darwin-fastbuild/bin
bazel-genfiles: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/execroot/__main__/bazel-out/darwin-fastbuild/bin
bazel-testlogs: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/execroot/__main__/bazel-out/darwin-fastbuild/testlogs
character-encoding: file.encoding = ISO-8859-1, defaultCharset = ISO-8859-1
command_log: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/command.log
committed-heap-size: 100MB
execution_root: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/execroot/__main__
gc-count: 7
gc-time: 167ms
install_base: /var/tmp/_bazel_arostovtsev/install/23356377d522b84e85cc6f4126877e6b
java-home: /private/var/tmp/_bazel_arostovtsev/install/23356377d522b84e85cc6f4126877e6b/embedded_tools/jdk
java-runtime: OpenJDK Runtime Environment (build 11.0.6+10-LTS) by Azul Systems, Inc.
java-vm: OpenJDK 64-Bit Server VM (build 11.0.6+10-LTS, mixed mode) by Azul Systems, Inc.
max-heap-size: 32178MB
output_base: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6
output_path: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/execroot/__main__/bazel-out
package_path: %workspace%
release: release 3.5.0
repository_cache: /var/tmp/_bazel_arostovtsev/cache/repos/v1
server_log: /private/var/tmp/_bazel_arostovtsev/7a6b4a7b43eb9194232ce1fb3493a8c6/java.log.arostovtsev-imacpro.arostovtsev.log.java.20200918-142934.33009
server_pid: 33009
used-heap-size: 38MB
workspace: /Users/arostovtsev/bzl/pr12046
$ bazelisk build --genrule_strategy=sandboxed --spawn_strategy=sandboxed a
INFO: Analyzed target //:a (19 packages loaded, 173 targets configured).
INFO: Found 1 target...
Target //:a up-to-date:
  bazel-out/apl-darwin_x86_64-fastbuild/bin/liba.a
INFO: Elapsed time: 5.554s, Critical Path: 0.18s
INFO: 2 processes: 2 darwin-sandbox.
INFO: Build completed successfully, 4 total actions
$ grep libtool_check_unique $(~/bin/bazelisk info server_log) || echo "nothing in the log"
nothing in the log
```

What version of Bazel are you seeing this in? What build flags are you using (if not `--genrule_strategy=sandboxed --spawn_strategy=sandboxed`)?

tetromino:
I see - the problem only occurs in Bazel built after d8723b48bf6d36d31444ff352bb15592399cc4d0. I should have read more carefully...

tetromino:
I wrote a small test for this which I'm not entirely happy with (would like to get it reviewed). But for this sort of highly user visible error message, I'm ok with accepting the PR without a test.

Thank you for the test case!

meteorcloudy:
Looks like this CL is breaking rules_cc in downstream:
https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/283

Please take a look

michaeleisel:
@meteorcloudy discussion is at https://github.com/bazelbuild/rules_cc/issues/80

michaeleisel:
based on that discussion, there will be some syncing of rules_cc with this, but no changes are needed in bazel. rules_cc just has to catch up (which will be worked on shortly)

laurentlb:
Can you explain more why this should be cherrypicked in Bazel 3.6?
I'm slightly reluctant because it seems to break rules_cc (which hasn't been updated yet). Can this wait one more month?

keith:
The rules_cc fix is here https://github.com/bazelbuild/rules_cc/issues/80

The tl;dr of this change is that the introduction of this optimization here: https://github.com/bazelbuild/bazel/pull/11958 didn't work for sandboxed builds, that wasn't noticed because builds still pass, they just take the non-optimized path (test for that here https://github.com/bazelbuild/bazel/pull/12136) and without this change it leads to significant log spam for every static archive action:

```
INFO: From Linking liba.a:
external/local_config_cc/libtool: line 40: /private/var/tmp/_bazel_meisel/1054548cf645506d169069fe6cc12477/sandbox/darwin-sandbox/8/execroot/__main__/external/local_config_cc/libtool_check_unique: No such file or directory
```

The reason rules_cc broke is because there's no automated path (or even process AFAICT) for updating when there are breaking changes with bazel's crosstool, which seems like something that should also be fixed.

IMO this is worth the cherry pick since it's a very localized change that without will discourage people from using sandboxing even more than they already are because of the performance implications

laurentlb:
@keith @oquenchil What's the status of this?

laurentlb:
Thanks for the precision!
I've made the cherrypick and created the rc, but I'd like this issue to be resolved before I can complete the release.

keith:
I don't see this cherry pick in the list of commits here: https://github.com/bazelbuild/bazel/issues/12040#issuecomment-702237058

laurentlb:
Oops, thanks for noticing (I copied the previous list). I've created a new rc.

PR #12048, opened 2020-09-04
Update String documentation: they are not iterable

TimNN:
According to the Starlark Language Design, "strings are not iterable".

Source: https://github.com/bazelbuild/starlark/blob/master/design.md#strings-are-not-iterable


laurentlb (approved):
lgtm

Thanks!

PR #12053, opened 2020-09-05
Add description about installation support from Bazel team or community members on the installation page.

jin:
I think this is something useful to explicitly call out.

### DIFF
--- a/site/docs/install.md
+++ b/site/docs/install.md
@@ -5,13 +5,16 @@ title: Installing Bazel
 
 # Installing Bazel
 
-See the instructions for installing Bazel on:
+The Bazel team officially supports Bazel on the following platforms. See the instructions for installing Bazel on:
 
 *   [Ubuntu Linux](install-ubuntu.md)
 *   [macOS](install-os-x.md)
 *   [Windows](install-windows.md)
 
 ## Community-supported packages
+
+These packages are maintained by Bazel community members and are not officially supported by the Bazel Team. Please 
+reach out to the respective package maintainers for support and questions.
 
 *   [Arch Linux](https://www.archlinux.org/packages/community/x86_64/bazel/)
 *   [Fedora 25, 26, 27, 28, and CentOS 7](install-redhat.md)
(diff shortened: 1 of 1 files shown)


site/docs/install.md
```diff
@@ -5,14 +5,17 @@ title: Installing Bazel
 
 # Installing Bazel
 
-See the instructions for installing Bazel on:
+The Bazel team officially supports Bazel on the following platforms. See the instructions for installing Bazel on:
```
gregestren, line 8: Maybe "the following" -> "these" to shorten?

gregestren, line 8: The original sentence reads a bit strange to me now that it's not alone. Do we even need it now?

site/docs/install.md
```diff
@@ -5,14 +5,17 @@ title: Installing Bazel
...
 
 *   [Ubuntu Linux](install-ubuntu.md)
 *   [macOS](install-os-x.md)
 *   [Windows](install-windows.md)
 
 ## Community-supported packages
 
+These packages are maintained by Bazel community members and are not officially supported by the Bazel Team. Please 
+reach out to the respective package maintainers for support and questions.
```
gregestren, line 17: Non-passive voice idea: "Bazel community members maintain these packages. The Bazel team does not officially support them" (this would also be consistent with the wording in the previous section).

site/docs/install.md
aiuto: And maybe "see the instructions for installing Bazel on" could go entirely.

site/docs/install.md
```diff
@@ -25,7 +28,7 @@ See the instructions for installing Bazel on:
 *   [Scoop](https://github.com/scoopinstaller/scoop-main/blob/master/bucket/bazel.json)
 
 ## Community-supported architectures
-*   [ppc64le](https://oplab9.parqtec.unicamp.br/pub/ppc64el/bazel)
 
+*   [ppc64le](https://oplab9.parqtec.unicamp.br/pub/ppc64el/bazel)
```
aiuto, line 32: I don't know what is correct here, but one use is ppc64**le** and the other is ppc64**el**. That seems off.

site/docs/install.md
jin: Done.

site/docs/install.md
jin: Removed.

site/docs/install.md
jin: Fixed to `el`.

gregestren (approved):

Issue #11619, opened 2020-06-19
--experimental_windows_watchfs fails when --symlink_prefix is specified


konste:
### Description of the problem / feature request:

Settings `--symlink_prefix=.build/` and `--experimental_windows_watchfs` together cause infinite symlink loop.

### Bugs: what's the simplest, easiest way to reproduce this bug?

Bazel user guide recommends setting `--symlink_prefix=.bazel/` to reduce clutter and that is what we do. With the arrival of Bazel 3.3.0 we tried new setting `common --experimental_windows_watchfs` in .bazelrc file and as the result we observe this in the build output: ```WARNING: Error encountered with local file system watcher java.nio.file.FileSystemException: D:\dev\bazel\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build\product_root\.build: The name of the file cannot be resolved by the system.
... temporarily falling back to manually checking files for changes```

Apparently this is caused by the folder `.build` (specified by --symlink_prefix) where Bazel creates symlink `product_root` (root of the workspace). This somehow confuses file system watcher on Windows. 

### What operating system are you running Bazel on?

Windows 10 1806.

### What's the output of `bazel info release`?

3.3.0

###  Have you found anything relevant by searching the web?

Relevant pull request: https://github.com/bazelbuild/bazel/pull/11334

### Any other information, logs, or outputs that you want to share?

Author of the pull request: @tomdegoede

tomdegoede:
Hi Konste, thanks for letting me know!

It seems that the default `bazel-*` symlinks are not symlinked in the execroot but when specifying `--symlink_prefix=.build/` the second invocation (after taking a long time) symlinks the .build folder in the execroot causing the infinite symlink loop.

I'm wondering how other OS scenarios are not hurt by this.

tomdegoede:
Two more observations.
1. When I shutdown bazel after the .build symlink is created inside execroot, and reboot bazel again it seems to work properly.
2. I never get the exception you listed. Maybe that's due to my permissive path length settings?

I have a feeling that the watcher incorrectly registers the .build folder to be watched when it's created.

konste:
Tom, thank you for looking into this! 
What setting do you mean when you said `my permissive path length settings`?

tomdegoede:
[...]

PR #12054, opened 2020-09-05
Take windows junctions into account with --watchfs

tomdegoede:
Related PR: #11334 I enabled --watchfs for windows under --experimental_windows_watchfs. 

Related issues: #11619:
1. (infinite) recursion occurs on the convenience symlinks when they are created while watch is active. This also causes a delay on the first bazel invocation after the symlinks are created.
2. quite often an overflow warning occurs `WARNING: Overflow when watching local filesystem for changes...`. I think this is because the watcher also watches the bazel-* symlinks with many file changes due to being output directories.

The reason that this behavior is different on windows seems to be that the convenience symlinks are actually created as junctions. I have tried changed this logic to creating actually directory symlinks which seemed to resolve the issues. This however could be a breaking change on some systems because creating symlinks requires more privileges than creating junctions.

~~Therefore I propose the following changes:~~
~~1. Watch the root directory non-recursively. As it's done on non windows systems.~~
~~2. Watch each direct sub-directory recursively. Excluding junctions.~~

~~The downside of this approach is that the convenience symlinks must be at the root of the workspace. When using --symlink_prefix with a sub-directory (which is the case for in the linked issue) both issues would still occur.~~

I have updated the PR to reflect my comment below because this also resolves the above mentioned downside and is a simpler solution. I have tested several scenarios with regards to the folder locking but was able to delete workspace folders from explorer, ide and cli every time.


tomdegoede:
Some other thought: I'm uncertain if the initial issues that caused windows --watchfs to become a non-op as discussed in #1931 and angular/angular#29541 were only related to output directory file handles. With the knowledge that excluding junctions from the tree walk excludes watching bazel-out etc. we may be able to just watch each individual directory again allowing to check for junctions on all sub-directory levels.

What do you think @meteorcloudy ?

meteorcloudy:
@tomdegoede Thanks for looking into this!

Do I understand correctly, on other platforms, bazel-* folders are also skipped because they are symlinks? If so, this solution looks good to me.

Just to make sure we are not relying on checking the bazel-* folder. Can you verify after enabling --watchfs on Windows, the following still work?
```
bazel build examples/cpp:hello-world
rm -f bazel-bin/examples/cpp/hello-world
touch bazel-bin/examples/cpp/hello-world
bazel build examples/cpp:hello-world -s  # The link action should be re-executed
```

tomdegoede:
Jep thats correct @meteorcloudy. The Files.walkFileTree does skip symlinks by default but was lacking junctions.

I've validated that the test binary is build again. This is because output files are checked via the `SequencedSkyframeExecutor::handleDiffsWithMissingDiffInformation` when checkOutputFiles is true.

meteorcloudy:
Great, thanks!

meteorcloudy (approved):

Issue #10159, opened 2019-11-04
Remote-cached build fails with javax.net.ssl.SSLException: handshake timed out


scele:
### Description of the problem / feature request:

We are running bazel with:

```
--remote_cache=https://our-aws-cloudfront-distribution.com/backed/by/s3/bucket
--disk_cache=~/path/on/local/disk
--remote_download_minimal
--remote_timeout=300
--remote_retries=2
```
and the build/test parallelism is high: `--jobs=auto` (nproc == 56), `--local_test_jobs=40`.

We are seeing intermittent errors that seem to be related to the use of https as the remote cache protocol.

Example 1:
```
(06:43:31) [3,407 / 3,521] 19 / 872 tests; Testing //some:target; 73s local ... (87 actions running)
(06:43:35) ERROR: /ssd/home/gitlab-runner/builds/ZQ9e7vGx/0/some/path/BUILD:40:1:  failed due to unexpected I/O exception: handshake timed out
javax.net.ssl.SSLException: handshake timed out
	at io.netty.handler.ssl.SslHandler.handshake(...)(Unknown Source)
(06:43:35) INFO: Elapsed time: 82.648s, Critical Path: 77.62s
(06:43:35) INFO: 1313 processes: 1216 remote cache hit, 97 local.
(06:43:35) FAILED: Build did NOT complete successfully
```

Example 2:
```
(06:21:37) WARNING: Reading from Remote Cache:
Symlinks in action outputs are not yet supported by --experimental_remote_download_outputs=minimal
(06:21:52) [2,129 / 3,349] 3 / 759 tests; Executing genrule @local_config_cuda//cuda:cuda-include; 19s local, remote-cache ... (78 actions running)
(06:21:58) WARNING: Writing to Remote Cache:
ClosedChannelException
(06:22:16) [2,600 / 3,768] 72 / 759 tests; Executing genrule @local_config_cuda//cuda:cuda-include; 43s local, remote-cache ... (60 actions, 51 running)
(06:22:22) WARNING: Writing to Remote Cache:
ExtendedClosedChannelException
(06:22:44) [2,615 / 3,782] 77 / 759 tests; Executing genrule @local_config_cuda//cuda:cuda-include; 70s local, remote-cache ... (70 actions, 68 running)
(06:23:15) [2,920 / 4,087] 144 / 759 tests; Executing genrule @local_config_cuda//cuda:cuda-include; 101s local, remote-cache ... (85 actions running)
[...]

PR #12055, opened 2020-09-05
Fix races in the HTTP upload and download handlers

ulfjack:
The user promise has a callback that returns the connection to the pool.
If the server returns a 'connection: close' HTTP header, then this can
currently happen before the connection is closed, in which case the client
attempts to reuse the connection, which - of course - fails.

This changes the ordering to close the connection *before* completing the
user promise.

This is at least a partial fix for the linked issue. It is unclear if this
is the root cause for all the reported failure modes.

Progress on #10159.

Change-Id: I2897e55c6edda592a6fb5755ddcccd1a89cde528


ulfjack:
@coeuvre

coeuvre (approved):
Thanks for the fix. LGTM, merging.

Issue #2964, opened 2017-05-08
remote/performance: If remote cache is inaccessible, fall back to building without the cache, rather than failing the build


mmorearty:
### Description of the problem / feature request / question:

Feature request:

I was able to get caching to work with `--spawn_strategy=remote --rest_cache_url=...`. It works well; but if the cache is inaccessible for any reason (e.g. I have gone offline and am working while commuting, or the server has gone down), then my builds fail.

Of course, I can change the options I'm using to launch Bazel; but that isn't always a good option. For one thing, my company has quite a lot of developers, and I would prefer that they not all have to learn this workaround. Secondly, in our automated Jenkins builds, launching with different command-line arguments isn't an option.

What I have hacked together for our own use is some changes to Bazel so that:

* Each time an error occurs trying to read or write the remote cache, it displays a short warning message, but continues the build. (`get` operations pretend the item was not found in the cache; `put` operations pretend the operation succeeded.)
* After ten consecutive such errors with no intervening successful cache accesses, Bazel displays a message that says, "Cache encountered multiple consecutive errors; disabling cache for 5 minutes."

My code for this was pretty quick-and-dirty, so it's not really in a shareable state, but it was pretty easy to write.

It would make sense for this to work regardless of what remote caching protocol is used -- REST API, gRPC, etc.

### Environment info
* Bazel version (output of `bazel info release`):

0.4.5

softprops:
This is going to be important for my organization. We are rolling out a replacement of sbt in favor of a cached bazel build. We do not want our cache server's downtime to block development.

softprops:
Just leaving this here for documentation.  For others that want a way to work around with using a release there is a [documented set of error codes](https://bazel.build/versions/master/docs/bazel-user-manual.html#what-exit-code-will-i-get). The error code bazel returns for this is 37. 

```
37 - Unhandled Exception / Internal Bazel Error.
```

To work around cache downtime ( or offline builds ), I'm falling back on a build that doesn't use remote cache if the return code is 37.

Ideally bazel would gracefully fallback on local cache if a remote cache could not be reached printing an informative warning that the cache is not available atm

The stack trace looks something like this 

```
$ bazel bulid ...
Caused by: org.apache.http.conn.HttpHostConnectException: Connection to https://bogus.cache.com refused
	at org.apache.http.impl.conn.DefaultClientConnectionOperator.openConnection(DefaultClientConnectionOperator.java:190)
[...]

PR #12064, opened 2020-09-09
Fall back to building without remote if remote server is inaccessible before build.

### LINKED ISSUE #11989: If grpc remote cache is inaccessible, fall back to building without the cache

It seems bazel is falling back to building without remote cache only when using a http remote cache. If a grpc cache is specified, the build fails with the following error:
```
INFO: Invocation ID: 6c2d307e-e24d-46b3-bce6-6fb116de093e
ERROR: java.io.IOException: io.grpc.StatusRuntimeException: UNAVAILABLE: Unable to resolve host fakeurl.net
	at com.google.devtools.build.lib.remote.RemoteServerCapabilities.get(RemoteServerCapabilities.java:82)
	at com.google.devtools.build.lib.remote.RemoteModule.verifyServerCapabilities(RemoteModule.java:148)
	at com.google.devtools.build.lib.remote.RemoteModule.beforeCommand(RemoteModule.java:349)
	at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.execExclusively(BlazeCommandDispatcher.java:390)
	at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.exec(BlazeCommandDispatcher.java:236)
	at com.google.devtools.build.lib.server.GrpcServerImpl.executeCommand(GrpcServerImpl.java:654)
	at com.google.devtools.build.lib.server.GrpcServerImpl.lambda$run$2(GrpcServerImpl.java:719)
	at io.grpc.Context$1.run(Context.java:595)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
Caused by: io.grpc.StatusRuntimeException: UNAVAILABLE: Unable to resolve host fakeurl.net
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:233)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:214)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:139)
	at build.bazel.remote.execution.v2.CapabilitiesGrpc$CapabilitiesBlockingStub.getCapabilities(CapabilitiesGrpc.java:221)
	at com.google.devtools.build.lib.remote.RemoteServerCapabilities.lambda$get$0(RemoteServerCapabilities.java:77)
	at com.google.devtools.build.lib.remote.Retrier.execute(Retrier.java:237)
	at com.google.devtools.build.lib.remote.RemoteRetrier.execute(RemoteRetrier.java:115)
	at com.google.devtools.build.lib.remote.RemoteServerCapabilities.get(RemoteServerCapabilities.java:77)
	... 10 more
Caused by: java.lang.RuntimeException: java.net.UnknownHostException: fakeurl.net
	at io.grpc.internal.DnsNameResolver.resolveAll(DnsNameResolver.java:414)
	at io.grpc.internal.DnsNameResolver$Resolve.resolveInternal(DnsNameResolver.java:252)
	at io.grpc.internal.DnsNameResolver$Resolve.run(DnsNameResolver.java:214)
	... 3 more
Caused by: java.net.UnknownHostException: fakeurl.net
	at java.base/java.net.InetAddress$CachedAddresses.get(Unknown Source)
	at java.base/java.net.InetAddress.getAllByName0(Unknown Source)
	at java.base/java.net.InetAddress.getAllByName(Unknown Source)
	at java.base/java.net.InetAddress.getAllByName(Unknown Source)
	at io.grpc.internal.DnsNameResolver$JdkAddressResolver.resolveAddress(DnsNameResolver.java:634)
	at io.grpc.internal.DnsNameResolver.resolveAll(DnsNameResolver.java:382)
	... 5 more
```

Allow developers to build offline or when our grpc remote cache service goes down. Maybe I'm missing a special flag to allow the build to continue or have a custom timeout if the grpc service is unreachable?

Pass `--remote_cache=grpc://any-grpc-service:8980` to any build and if the service is not reachable, the build will fail.

macOS 10.15

`release 3.4.1`

Similar issue here, but it seems like it was related to the HTTP cache instead: https://github.com/bazelbuild/bazel/issues/2964

coeuvre:
In the case where gRPC remote mode is enabled (gRPC remote cache or remote execution), Bazel will check the server capabilities [before a build](https://github.com/bazelbuild/bazel/blob/0cb411b525cba2d9d5ac854e679e5b115151035c/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java#L426). If the server is inaccessible at that time, the build fails. (Bazel won't check the connection to the server at that time for HTTP remote cache.)

There were some discussion at #2964 and some fixes are landed already but are for the case where remote server is inaccessible during the build.

This PR addresses above issue by falling back to building without remote cache/execution if `--remote_local_fallback` is set.

Fixes #11989.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -47,6 +47,7 @@ import com.google.devtools.build.lib.buildtool.BuildRequestOptions;
 import com.google.devtools.build.lib.collect.nestedset.NestedSet;
 import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.Reporter;
+import com.google.devtools.build.lib.exec.ExecutionOptions;
 import com.google.devtools.build.lib.exec.ExecutorBuilder;
 import com.google.devtools.build.lib.exec.ModuleActionContextRegistry;
 import com.google.devtools.build.lib.exec.SpawnStrategyRegistry;
@@ -154,7 +155,7 @@ public final class RemoteModule extends BlazeModule {
       CommandEnvironment env,
       DigestUtil digestUtil,
       ServerCapabilitiesRequirement requirement)
-      throws AbruptExitException {
+      throws AbruptExitException, IOException {
     RemoteServerCapabilities rsc =
         new RemoteServerCapabilities(
             remoteOptions.remoteInstanceName,
@@ -165,12 +166,6 @@ public final class RemoteModule extends BlazeModule {
     ServerCapabilities capabilities = null;
     try {
       capabilities = rsc.get(env.getCommandId().toString(), env.getBuildRequestId());
-    } catch (IOException e) {
-      env.getReporter().handle(Event.error(Throwables.getStackTraceAsString(e)));
-      throw createExitException(
-          "Failed to query remote execution capabilities: " + Utils.grpcAwareErrorMessage(e),
-          ExitCode.REMOTE_ERROR,
-          Code.CAPABILITIES_QUERY_FAILURE);
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
       return;
@@ -200,6 +195,10 @@ public final class RemoteModule extends BlazeModule {
     AuthAndTLSOptions authAndTlsOptions = env.getOptions().getOptions(AuthAndTLSOptions.class);
     DigestHashFunction hashFn = env.getRuntime().getFileSystem().getDigestFunction();
     DigestUtil digestUtil = new DigestUtil(hashFn);
+
+    ExecutionOptions executionOptions = Preconditions
+        .checkNotNull(env.getOptions().getOptions(ExecutionOptions.class));
+    boolean verboseFailures = executionOptions.verboseFailures;
 
     boolean enableDiskCache = RemoteCacheClientFactory.isDiskCache(remoteOptions);
     boolean enableHttpCache = RemoteCacheClientFactory.isHttpCache(remoteOptions);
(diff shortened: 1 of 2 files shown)


coeuvre:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -441,13 +433,21 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
-      } catch (IOException e) {
-        if (verboseFailures) {
-          env.getReporter().handle(Event.warn(Throwables.getStackTraceAsString(e)));
-        }
-        env.getReporter().handle(Event
-            .warn("Failed to query remote cache capabilities: " + Utils.grpcAwareErrorMessage(e)));
+      }
+    } catch (IOException e) {
+      if (verboseFailures) {
+        env.getReporter().handle(Event.error(Throwables.getStackTraceAsString(e)));
```
ulfjack, line 439: As a general note, I'd prefer a single warning / error message for a failure rather than multiple.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -441,13 +433,21 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+      }
+      String errorMessage =
+          "Failed to query remote execution capabilities: " + Utils.grpcAwareErrorMessage(e);
```
ulfjack, line 442: You might want to also say how much time Bazel waited for the server to respond.

ulfjack (commented):
I think the main scenario here is that the capabilities request times out, which means you need to think about what Bazel shows during that time and what Bazel shows to the user afterwards.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
coeuvre: Done

coeuvre:
> I think the main scenario here is that the capabilities request times out, which means you need to think about what Bazel shows during that time and what Bazel shows to the user afterwards.

We makes the capabilities request with a retrier which means, if the request time out or encounter other IO errors it will retry `--remote_retries` times silently and print out the error/warn message once afterwards if all retries failed.

Do you mean we should print a warning message if a retry happened? Since all other operations behind a retrier are retried silently, I'm not sure whether it is too verbose for users by doing this.

> You might want to also say how much time Bazel waited for the server to respond.

Do you think it is enough to print total time waited along with retried times or we should print at each retry level?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
coeuvre: Please see comment below.

ulfjack:
I don't have a strong opinion about what to show exactly.

Also, note that this part of the code is executed during early Bazel server startup. IIRC, the reporter is connected to an in-memory buffer, and nothing you print from here is immediately visible in the terminal. This is something we might want to fix separately.

Feel free to go ahead and check this in. We can still iterate on the details.

Issue #8583, opened 2019-06-07
Provide ability to directly select() on constraint_value


gregestren:
### Description of the problem / feature request:

A common and annoyingly redundant pattern is:

```
constraint_value(
    name = "my_value",
    constraint_setting = "my_setting")

config_setting(
    name = "my_value_selectable",
    constraint_values = [":my_value"])

some_rule(
    ...,
    attr = select({
        ":my_value_selectable": ...
    })
```

It would be nice to be able to just write:

```
some_rule(
    ...,
    attr = select({
        ":my_value": ...
    })

```

### Feature requests: what underlying problem are you trying to solve with this feature?

Unnecessary bloat for situations where `config_setting` is just abstraction overhead.

###  Have you found anything relevant by searching the web?

Examples:
* https://github.com/bazelbuild/rules_rust/blob/5ec0b832d6a7a7ad183cfb9fb99ff2cf91429caa/rust/platform/platform.bzl#L59-L67
* https://github.com/bazelbuild/rules_dotnet/blob/07ee724341d5bd70e41c6980ee202a282ccb3462/dotnet/platform/list.bzl#L77-L82

Also see https://github.com/bazelbuild/bazel/issues/7668 and https://github.com/bazelbuild/bazel-skylib/pull/89#issuecomment-499257552.

gregestren:
@katre I tried mocking up an implementation of this since it's conceptually at least easy (and `select` was always designed to support rules beside `config_setting`).

Some challenges that arose that I appreciate your perspective on:

1. `config_setting` implements https://github.com/bazelbuild/bazel/blob/af53a11104b549cca8b593a1f8cac224a21727da/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java#L1389-L1393 which is necessary for it to always invoke toolchain resolution so there are actual target `constraint_values` to compare against. The `constraint_value` rule doesn't do that though. What options do we have here?

1. What configuration do `constraint_value` rules evaluate in today? If in a `select` would they properly use the same configuration of the rule with the `select` in it?
1. If this change involves evaluating `constraint_value` in more configurations, would that bloat the build graph? Or maybe just do this only in the context of a `select`? Maybe some secret alias rule that secretly sets a `requires platforms just for this call` option in `PlatformOptions` that `constraint_value` can specially consume?

gregestren:
@hlopko

katre:
[...]

Issue #11984, opened 2020-08-20
Beautify Bazel configuration errors


gregestren:
**tl;dr:** apply [beautiful error](https://bazel.build/designs/2016/05/23/beautiful-error-messages.html)  principles to build failures from analysis, `select()`, configurations, user-defined flags, transitions, and so on. 

---
We have enough experience to identify where current errors fall short. There are straightforward ways we make them clearer, more actionable, and more closely connected to relevant documentation.

I'm starting this for configuration-related errors since this is the Bazel area I know best. More broadly I'd like to extend these principles to Bazel error culture as a whole in the interest of making Bazel a more accessible build tool.

Crowd-sourced input on confusing messages welcome!

gregestren:
@jin

mobileink:
If you're collecting error messages that need fixing, here's another one:

 ```
ERROR: <root>/src/BUILD.bazel:596:13: output 'src/refList.o' was not created
ERROR: <root>/src/BUILD.bazel:596:13: not all outputs were created or valid
```

This is not much better than "Your build failed".  It will be totally opaque to users who do not know that "output" has special meaning for Bazel rules.  It does not mean the compile failed; it might have failed, but it might have succeeded yet failed to put the output in the right place.  Worst of all is "[not] created or valid" - doesn't Bazel know which?

Best,

Gregg

**Comment by gregestren** [comment]
Thanks, Gregg. Excellent and subtle point on the "A or B" phrasing. I'll dig in when I get a chance to see what options we have for those errors.

**Comment by mobileink** [comment]
Here's another useless error message:

```
ERROR: Failed to load Starlark extension '@bazel_skylib//lib:paths.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @bazel_skylib
This could either mean you have to add the '@bazel_skylib' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
```

I have `@bazel_skylib//lib:paths.bzl` in a dozen files; which is it?  Surely Bazel knows.  I also have multiple workspaces. Bazel should be able to tell me where the problem is.

Gregg

gregestren:
[...]

PR #12071, opened 2020-09-09
Support constraint_value directly in select()

gregestren:
No more need for a redundant config_setting. See #8583 for an example.

This was a bit subtle. constraint_value can't directly export a ConfigMatchingProvider because it needs to know the
platform to determine if it's a match. But platforms are built out of constraint_values, so the platform isn't available
yet. So the parent target with the select() provides this detail.

Also beautifies "invalid select() key" errors in support of https://github.com/bazelbuild/bazel/issues/11984.

Fixes https://github.com/bazelbuild/bazel/issues/8583.

RELNOTES[NEW]: select() directly supports constraint_value (no need for an intermediate config_setting).

### DIFF
--- a/site/docs/configurable-attributes.md
+++ b/site/docs/configurable-attributes.md
@@ -144,9 +144,14 @@ build parameters, which include `--cpu=arm`. The `tools` attribute changes
 ## Configuration conditions
 
 Each key in a configurable attribute is a label reference to a
-[`config_setting`](be/general.html#config_setting). This is just a collection of
+[`config_setting`](be/general.html#config_setting) or
+[`constraint_value`](be/platform.html#constraint_value).
+
+`config_setting` is just a collection of
 expected command line flag settings. By encapsulating these in a target, it's
 easy to maintain "standard" conditions users can reference from multiple places.
+
+`constraint_value` provides support for [multi-platform behavior](#platforms).
 
 
 ### Built-in flags
@@ -263,7 +268,7 @@ While the ability to specify multiple flags on the command line provides
 flexibility, it can also be burdensome to individually set each one every time
 you want to build a target.
    [Platforms](platforms.html)
-allow you to consolidate these into simple bundles.
+let you consolidate these into simple bundles.
 
 ```python
 # myapp/BUILD
@@ -338,8 +343,26 @@ Without platforms, this might look something like
 bazel build //my_app:my_rocks --define color=white --define texture=smooth --define type=metamorphic
 ```
 
-Platforms are still under development. See the [documentation](platforms.html)
-and [roadmap](https://bazel.build/roadmaps/platforms.html) for details.
+`select()` can also directly read `constraint_value`s:
+
+```python
+constraint_setting(name = "type")
+constraint_value(name = "igneous", constraint_setting = "type")
+constraint_value(name = "metamorphic", constraint_setting = "type")
+sh_binary(
+    name = "my_rocks",
+    srcs = select({
+        ":igneous": ["igneous.sh"],
+        ":metamorphic" ["metamorphic.sh"],
+    }),
+)
+```
+
+This saves the need for boilerplate `config_setting`s when you only need to
+check against single values.
+
+Platforms are still under development. See the
+[documentation](platforms-intro.html) for details.
 
 ## Combining `select()`s
 
--- a/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java
@@ -14,6 +14,10 @@
 
 package com.google.devtools.build.lib.analysis.platform;
 
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.devtools.build.lib.analysis.config.ConfigMatchingProvider;
 import com.google.devtools.build.lib.cmdline.Label;
 import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
 import com.google.devtools.build.lib.packages.BuiltinProvider;
@@ -56,6 +60,31 @@ public class ConstraintValueInfo extends NativeInfo implements ConstraintValueIn
   @Override
   public Label label() {
     return label;
+  }
+
+  /**
+   * Returns a {@link ConfigMatchingProvider} that matches if the owning target's platform includes
+   * this constraint.
+   *
+   * <p>The {@link com.google.devtools.build.lib.rules.platform.ConstraintValue;ConstraintValue}
+   * rule can't directly return a {@link ConfigMatchingProvider} because, as part of a platform's
+   * definition, it doesn't have access to the platform during its analysis.
+   *
+   * <p>Instead, a target with a <code>select()</code> on a {@link
+   * com.google.devtools.build.lib.rules.platform.ConstraintValue} passes its platform info to this
+   * method.
+   */
+  public ConfigMatchingProvider configMatchingProvider(PlatformInfo platformInfo) {
+    return new ConfigMatchingProvider(
+        label,
+        ImmutableMultimap.of(),
+        ImmutableMap.of(),
+        // Technically a select() on a constraint_value requires PlatformConfiguration, since that's
+        // used to build the platform this checks against. But platformInfo's existence implies
+        // the owning target already depends on PlatformConfiguration. And we can't reference
+        // PlatformConfiguration.class here without a build dependency cycle.
+        ImmutableSet.of(),
+        platformInfo.constraints().hasConstraintValue(this));
   }
 
   @Override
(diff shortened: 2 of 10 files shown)


gregestren:
@philsc @AustinSchuh  FYI

site/docs/configurable-attributes.md
```diff
@@ -338,8 +343,26 @@ Without platforms, this might look something like
...
+    name = "my_rocks",
+    srcs = select({
+        ":igneous": ["igneous.sh"],
+        ":metamorphic" ["metamorphic.sh"],
+    }),
+)
+```
+
+This saves the need for boilerplate `config_setting`s when you only need to
+check against single values.
```
katre, line 362: Should we give an example of using config_setting for multiple constraint values?

src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java
```diff
@@ -58,6 +62,31 @@ public Label label() {
     return label;
   }
 
+  /**
+   * Returns a {@link ConfigMatchingProvider} that matches if the owning target's platform includes
+   * this constraint.
+   *
+   * <p>The {@link com.google.devtools.build.lib.rules.platform.ConstraintValue;ConstraintValue}
```
katre, line 69: The `@link` syntax looks wrong here, what are you intending to do?

src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java
```diff
@@ -58,6 +62,31 @@ public Label label() {
...
+   * rule can't directly return a {@link ConfigMatchingProvider} because, as part of a platform's
+   * definition, it doesn't have access to the platform during its analysis.
+   *
+   * <p>Instead, a target with a <code>select()</code> on a {@link
+   * com.google.devtools.build.lib.rules.platform.ConstraintValue} passes its platform info to this
+   * method.
```
katre, line 75: That is really clever. :)

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -267,13 +269,27 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
...
       // Get the configuration targets that trigger this rule's configurable attributes.
       ImmutableMap<Label, ConfigMatchingProvider> configConditions =
           getConfigConditions(
-              ctgValue.getTarget(),
               env,
               ctgValue,
               transitivePackagesForPackageRootResolution,
+              unloadedToolchainContexts == null
+                  ? null
+                  : unloadedToolchainContexts.getDefaultToolchainContext().targetPlatform(),
```
katre, line 292: All toolchain contexts for the same configured target, regardless of exec group, should have the same target platform (right @juliexxia ?)

We should consider enshrining that in the API by adding a unloadedToolchainContexts.targetPlatform() method.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -789,16 +794,28 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
...
+      ConfigMatchingProvider matchingProvider = value.getProvider(ConfigMatchingProvider.class);
+      ConstraintValueInfo constraintValueInfo = value.get(ConstraintValueInfo.PROVIDER);
+
+      if (matchingProvider != null) {
+        configConditions.put(entry.getLabel(), matchingProvider);
+      } else if (constraintValueInfo != null && platformInfo != null) {
+        // If platformInfo == null, that means the owning target doesn't invoke toolchain
+        // resolution, in which case depending on a constraint_value is non-sensical.
+        configConditions.put(
+            entry.getLabel(), constraintValueInfo.configMatchingProvider(platformInfo));
```
katre, line 808: Could we add a ConfigMatchingProvider.configMatchingProvider(???) method, so that we'd only have one branch here? We'd have to do some weird generic tricks to get the type right, and it might not be worth it, but if we ever add a third, we should consider a way to collapse the choices.

katre (approved):
This is really clever and I like it.

Most of my comments are not actionable, just ways I want to think about the API. I will probably come back tomorrow and turn some of them into standalone issues, sorry for adding my stream-of-consciousness to your PR review.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
juliexxia: Yeah that sounds right and looks right according to https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java;l=531?q=ConfiguredTargetFunction&ss=bazel

@katre I like that idea. Do you think there's anything else to discuss there? If not I can mail something.

katre: I don't think so, it makes sense. Go ahead if you have time, I was going to file an issue and then deal with it later :)

gregestren:
> This is really clever and I like it.
> 

I assure you my first pass on this looked 10,000 times worse! I was surprised and happy to reduce this to something so non-invasive.

> Most of my comments are not actionable, just ways I want to think about the API. I will probably come back tomorrow and turn some of them into standalone issues, sorry for adding my stream-of-consciousness to your PR review.

No worries, thanks for the comments. I'll follow up properly on them as soon as I can.

site/docs/configurable-attributes.md
gregestren: The example before here already does this, so I think we have good coverage overall: https://docs.bazel.build/versions/3.5.0/configurable-attributes.html#platforms.

src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java
gregestren: Right. Just removing the part after the semi-colon works. I fully qualified it because otherwise an import on its package adds a dependency which messes with the BUILD dependency structure. Which isn't worth it just for a code link (I believe that's our actual style guidance too: to not add new dependencies just for doc links).

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: To be clear, this suggestion is for a followup PR?

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: Good point. Given the frequency of rule expansion (very infrequent), I vote for not supporting that much extra abstraction yet. But I added a note with your suggestion if this ever goes further.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
katre: Yes, nothing to do in this PR.

katre (approved):

pcjanzen:
Is this supposed to work in a filegroup?

```
filegroup(
    name = "cv_in_filegroup",
    srcs = select({
        "@platforms//os:windows": ["dummy"],
        "//conditions:default": ["dummy"],
    }),
)

$ bazel-4.0.0-darwin-x86_64 build //:cv_in_filegroup
ERROR: BUILD:84:10: @platforms//os:windows is not a valid select() condition for //:cv_in_filegroup.
To inspect the select(), run: bazel query --output=build //:cv_in_filegroup.
For more help, see https://docs.bazel.build/be/functions.html#select.
```
but the corresponding version with a config_setting works correctly, as does the select clause in (for example) a sh_binary.

cc @UebelAndre

gregestren:
Interesting catch!

My first instinct was to say it shouldn't matter what rule uses the `select` - that's not where the logic kicks in. But lo and behold I replicated your exact error, then changed the `filegroup` to a `cc_library` and suddenly it worked. This calls for more digging...

UebelAndre:
One other thing I noticed is the [ctx.attr.bool](https://docs.bazel.build/versions/master/skylark/lib/attr.html#bool) does not accept select statements. Not sure if that's directly related to this 😅

gregestren:
Found the problem! https://github.com/bazelbuild/bazel/blob/63e03929ddde6310fedc25c5ca782c124d639e8e/src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java#L814

In this example `platformInfo` is null.

Matching the `constraint_value` with the platform requires knowing what the platform is. `select()` logic gets all its info from the target it's attached to. Most rules trigger platform resolution, but some don't. `filegroup` is one of the exceptions.

I think the idea is that `filegroup` is such a simple rule it doesn't matter what the platform is, since there's no `filegroup` toolchain to match it to. So Bazel avoids the extra work of platform resolution. As a side effect you get this bug. 

i believe `filegroup` is one of the rare rules with this problem. CC @katre if there's any obvious fix to this blip.

`ctx.attr.bool` sounds like a separate issue. Can you file a distinct bug for it?

katre:
Filegroup was explicitly added to the set of rules that don't participate in toolchain resolution (and that thus don't have a target platform) back during the early development. As I recall, it led to a cycle in evaluation, because filegroups were needed at a low-enough level that we couldn't process loading platforms without them.

I'll take another look at this and see if we can turn it back on.

gregestren:
If that doesn't make sense conceptually, I think we could conditionally load the platform when we notice a `constraint_value` within the `select()` logic.

PR #12073, opened 2020-09-09
Remove dead objc setup code

keith:
This appears to be unused


keith:
@trybka continuing from https://github.com/bazelbuild/bazel/pull/12046#discussion_r485973823

if no one references `LIBTOOL_ATTRIBUTE` does that mean this is unreferenced entirely? I assumed that the registration of the `LibtoolRule` might mean it could be used implicitly somehow?

trybka:
Good find on `LibtoolRule`. My read of this is that any caller which would want to invoke *this* libtool would need to grab it via the attribute (or calling the getter), and since I can't find those references, I _think_ the whole mess is unused? Curious how far on the thread this can be pulled out.

keith:
I will push a second commit with that assumption after I validate I can build our app without that attribute

keith:
if all it was doing was adding that implicit attribute, I would assume that someone would have to reference `$libtool` (even if not with the constant) to use it? similar to https://github.com/bazelbuild/bazel/blob/24a082de6abc9bcaec5bf04035a7c55680bd614f/src/main/java/com/google/devtools/build/lib/analysis/test/TestActionBuilder.java#L245-L246

trybka:
> if all it was doing was adding that implicit attribute, I would assume that someone would have to reference `$libtool` (even if not with the constant) to use it? 

That's my understanding, yes.

trybka:
+cc @googlewalt as an FYI

tools/objc/gcov_stub
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
```
keith, line 1: hrm based on the code in rules_apple, this might work inside google? https://github.com/bazelbuild/rules_apple/blob/0a1cbce9b3d169269313aba1113d2aa0a8c3c514/apple/internal/testing/apple_test_rule_support.bzl#L124-L128

src/main/java/com/google/devtools/build/lib/rules/objc/ObjcCommandLineOptions.java
```diff
@@ -225,29 +225,6 @@
...
-      converter = LabelConverter.class,
-      documentationCategory = OptionDocumentationCategory.TOOLCHAIN,
-      effectTags = {OptionEffectTag.CHANGES_INPUTS},
-      help =
-          "Location of tool to scan Objective-C code for inclusions and output a .headers_list "
-              + "file.")
-  public Label objcHeaderScannerTool;
-
-  @Option(
-    name = "apple_sdk",
```
trybka, line 240: Not sure if we can nuke this one just yet.

trybka (commented):
Let's put the apple_sdk flag back for now, there are more references to it internally, and we'd need to clean them up. 

You can probably still remove the getters and whatnot

keith:
added that flag back

keith:
Thanks a ton!~

Issue #12056, opened 2020-09-06
3.5.0 doesn't compile with gcc 10


svenstaro:
### Description of the problem / feature request:

Compiling bazel fails with
```
google/devtools/build/v1/build_events.proto:20:1: warning: Import google/api/annotations.proto is unused.
ERROR: /tmp/bazel_2EvbRwLJ/out/external/upb/BUILD:57:11: C++ compilation of rule '@upb//:upb' failed (Exit 1): gcc failed: error executing command 
  (cd /tmp/bazel_2EvbRwLJ/out/execroot/io_bazel && \
  exec env - \
    PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl \
    PWD=/proc/self/cwd \
  /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections -fdata-sections -MD -MF bazel-out/k8-opt/bin/external/upb/_objs/upb/upb.d '-frandom-seed=bazel-out/k8-opt/bin/external/upb/_objs/upb/upb.o' -iquote external/upb -iquote bazel-out/k8-opt/bin/external/upb -Werror -Wno-long-long -pedantic -Wstrict-prototypes -fno-canonical-system-headers -Wno-builtin-macro-redefined '-D__DATE__="redacted"' '-D__TIMESTAMP__="redacted"' '-D__TIME__="redacted"' -c external/upb/upb/upb.c -o bazel-out/k8-opt/bin/external/upb/_objs/upb/upb.o)
Execution platform: //:default_host_platform
In file included from /usr/include/string.h:519,
                 from external/upb/upb/upb.h:16,
                 from external/upb/upb/upb.c:2:
In function 'strncpy',
    inlined from 'upb_status_seterrmsg' at external/upb/upb/upb.c:40:3:
[...]

PR #12077, opened 2020-09-10
Update upb to fix compilation error with gcc 10

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel/issues/12056

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -178,7 +178,7 @@ distdir_tar(
         # c-ares/c-ares
         "e982924acee7f7313b4baa4ee5ec000c5e373c30.tar.gz",
         # protocolbuffers/upb
-        "9effcbcb27f0a665f9f345030188c0b291e32482.tar.gz",
+        "382d5afc60e05470c23e8de19b19fc5ad231e732.tar.gz",
     ],
     dirname = "derived/distdir",
     sha256 = {
@@ -215,7 +215,7 @@ distdir_tar(
         # c-ares/c-ares
         "e982924acee7f7313b4baa4ee5ec000c5e373c30.tar.gz": "e8c2751ddc70fed9dc6f999acd92e232d5846f009ee1674f8aee81f19b2b915a",
         # protocolbuffers/upb
-        "9effcbcb27f0a665f9f345030188c0b291e32482.tar.gz": "61d0417abd60e65ed589c9deee7c124fe76a4106831f6ad39464e1525cef1454",
+        "382d5afc60e05470c23e8de19b19fc5ad231e732.tar.gz": "7992217989f3156f8109931c1fc6db3434b7414957cb82371552377beaeb9d6c",
     },
     urls = {
         "e0b0291b2c51fbe5a7cfa14473a1ae850f94f021.zip": [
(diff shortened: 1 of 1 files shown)


meteorcloudy:
@vbatts Can you verify if this PR fix the compilation error for you?

aiuto:
The change looks fine, but the distfile tests are still failing in various ways.

aiuto:
https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/6893#23242da4-4c34-4fd8-bf2e-f391ee10d809

It seems to be downloading 9effcbcb27f0a665f9f345030188c0b291e32482 instead of the newer version.
Did this miss the instance at WORKSPACE@641?

meteorcloudy:
@aiuto Nice catch, fixed!

meteorcloudy:
Hmm, looks like we have to explicitly override the upb on WORKSPACE file.

aiuto:
Now it fails in the non_determinism test. Could it be related to //third_party/grpc:grpc_1.26.0.patch?

aiuto:
Or maybe it is the grpc patch doing it.

meteorcloudy:
> Now it fails in the non_determinism test. Could it be related to //third_party/grpc:grpc_1.26.0.patch?

No, the patch is for something else. The upb is actually introduced in grpc_deps(), I tried to override it by declaring it directly in Bazel's WORKSPACE file. But it doesn't seem to work because the new upb version doesn't work with the old grpc version.. This is not looking good ...

aiuto:
Yeah. I still see the same error with the update patch.  But, if you decide you want it. 
```
diff --git a/third_party/grpc/grpc_1.26.0.patch b/third_party/grpc/grpc_1.26.0.patch
index 03b4aed96a..25d0631773 100644
--- a/third_party/grpc/grpc_1.26.0.patch
+++ b/third_party/grpc/grpc_1.26.0.patch
@@ -70,12 +70,14 @@ index 09fcad95a2..9b737e5deb 100644
      if "com_google_absl" not in native.existing_rules():
 @@ -209,7 +212,10 @@ def grpc_deps():
              name = "upb",
-             sha256 = "61d0417abd60e65ed589c9deee7c124fe76a4106831f6ad39464e1525cef1454",
-             strip_prefix = "upb-9effcbcb27f0a665f9f345030188c0b291e32482",
+-            sha256 = "61d0417abd60e65ed589c9deee7c124fe76a4106831f6ad39464e1525cef1454",
+-            strip_prefix = "upb-9effcbcb27f0a665f9f345030188c0b291e32482",
 -            url = "https://github.com/protocolbuffers/upb/archive/9effcbcb27f0a665f9f345030188c0b291e32482.tar.gz",
++            sha256 = "7992217989f3156f8109931c1fc6db3434b7414957cb82371552377beaeb9d6c",
++            strip_prefix = "upb-382d5afc60e05470c23e8de19b19fc5ad231e732",
 +            urls = [
-+                "https://mirror.bazel.build/github.com/protocolbuffers/upb/archive/9effcbcb27f0a665f9f345030188c0b291e32482.tar.gz",
-+                "https://github.com/protocolbuffers/upb/archive/9effcbcb27f0a665f9f345030188c0b291e32482.tar.gz",
++                "https://mirror.bazel.build/github.com/protocolbuffers/upb/archive/382d5afc60e05470c23e8de19b19fc5ad231e732.tar.gz",
++                "https://github.com/protocolbuffers/upb/archive/382d5afc60e05470c23e8de19b19fc5ad231e732.tar.gz",
 +            ],
          )
      if "envoy_api" not in native.existing_rules():
```

meteorcloudy:
Thanks for coming up with the patch, unfortunately the newer version isn't compatible with grpc. The best I can come up with is to patch the original upb version, see https://github.com/bazelbuild/bazel/pull/12083

aiuto:
So, is this PR moot, and we should do 3.5.1 with a cherrpick of #12083?

meteorcloudy:
Yes, looks like a regression we should fix.

PR #12079, opened 2020-09-10
add proxy for external http archive url

yancl:
Hi, Every time we build our projects by Bazel from a clean environment(ci for example), it will cost a lot of time to download the external artifacts(sometimes it even breaks because of network jitters). So I think it will be better for Bazel to support something like the [GOPROXY](https://proxy.golang.org/) so that we can maintain our own mirrored external artifacts and speed up the building process. 

I know we can make it currently by specify multi urls like [this](https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl#L97), but it needs to change every project's WORKSPACE. So maybe make Bazel supporting it will be better:)


philwo:
@yancl I really like this idea and wanted to have a feature like this for quite some time.

I think using an environment variable is not the right approach here, though. Could you instead add a flag to `RepositoryOptions` here: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/repository/RepositoryOptions.java;l=34?

The `download` method then looks like a good place to implement the prefixing logic: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java;l=61;drc=4fd17d3ec846c126dc157987311e377dcc5838b7;bpv=1;bpt=1 because this already iterates over the list of mirrors and the prefixed one would just be a special case.

Issue #11699, opened 2020-07-02
Bazel crashes when http_archive downloads an archive which contains an archive of the same name


carpenterjc:
# Description of the problem / feature request:

http_archive downloads then expands the http_archive in place. This at best means the archive extraction is corrupted. 
I have seen a few different errors: `Unexpected end of ZLIB input stream` and/or ERROR: An error occurred during the fetch of repository 'archive_in_archive':
   java.io.IOException: Error extracting C:/dev/_bazel/3laerure/external/archive_in_archive/archive_in_archive.zip to C:/dev/_bazel/3laerure/external/archive_in_archive: The file 'archive_in_archive.zip' is nis not a correctly formatted zip file: Expected a Fileot a correctly formatted zip file: Expected a File Header at file offset 239, but was not present.
ERROR: java.io.IOException: Error extracting C:/dev/_bazel/3laerure/external/archive_in_archive/archive_in_archive.zip to C:/dev/_bazel/3laerure/external/archive_in_archive: The file 'archive_in_archive.zip' 
is not a correctly formatted zip file: Expected a File Header at file offset 239, but was not present.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

> git clone https://github.com/carpenterjc/bazel_experiments.git
> cd bazel_experiments
Works:
> bazel fetch @archive_in_archive_new_name//...
Fails:
> bazel fetch @archive_in_archive//...

### What operating system are you running Bazel on?

Windows
### What's the output of `bazel info release`?

> INFO: Invocation ID: 14b242a8-8ca7-4d28-aa71-a559ed10bfec
> release 2.2.0

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

> https://github.com/carpenterjc/bazel_experiments.git
108b8703d0fa3403de852deaa27bca8447895c2a
108b8703d0fa3403de852deaa27bca8447895c2a

###  Have you found anything relevant by searching the web?

Not with the same underlying cause. 

### Any other information, logs, or outputs that you want to share?

> Having looked into this the archive is extracted into the relevant external directory, it then starts expanding the archive directly into that archive. Which will trample over the archive being read. 

Suggest the archive is dowloaded to a different directory then expanded to the external directory.

meisterT:
Note that this also fails on Linux with the same error message.

[...]

PR #12082, opened 2020-09-10
Download to tmp directory in ctx.download_and_extract

mai93:
Modify `ctx.download_and_extract` to download the archive file to a tmp directory inside the given output path then delete the directory after extraction. This solves the problem that happens when the archive contains an archive with the same name.

Fixes: #11699

@philwo

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
@@ -81,7 +81,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
+import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -914,13 +917,13 @@ public StructImpl downloadAndExtract(
 
     StructImpl downloadResult = calculateDownloadResult(checksum, downloadedPath);
     try {
-      if (downloadedPath.exists()) {
-        downloadedPath.delete();
+      if (downloadDirectory.exists()) {
+        downloadDirectory.deleteTree();
```
philwo, line 921: In case multiple of these `downloadAndExtract` operation happen in parallel, one thread might try to delete the `tmp` directory, while the other is still busy extracting, which can cause an error.

Could we instead of using a hardcoded `tmp` directory to put the ZIP file in use a securely generated random temporary file name to download the ZIP file to and then in a finally block ensure that we delete it?

We can probably use Java I/O here to create a temporary file:

```
Path tempfile = fileSystem.getPath(File.createTempFile(prefix, suffix, outputPath).getAbsolutePath());
```

Prefix can be anything you like, suffix has to be the actual suffix of the archive (e.g. `.tar.gz` or `.zip`) so that the detection logic for which decompressor to use stays working.

mai93: Yes, I think this will be better. But just to make sure, I think I will use `Files.createTempDirectory()` to create the `tmp` directory and pass this to the download manager to download the archive in it. The downloaded file is created by the download manager using VFS API.

mai93: Never mind, I can see now that we can pass the required file path to `downloadManager.download()`.

mai93: I worked on creating a temporary file using Java I/O but I found that there is some logic used to determine the suffix of the archive from the main url implemented in the DownloadManager. I am not sure if I should do a similar logic to create the temp file or just create a temp directory and pass it to the download manager.

philwo: This is a good point. It is probably easier to just create a temporary directory, put the file into it with the original filename and then later delete the directory again. I also cannot think of a drawback with that approach, so let's do that. :)

philwo (approved):

philwo:
Looks good to me, can you import this and send the CL to me for review? :)

PR #12093, opened 2020-09-11
Set a default value for --host_crosstool_top

katre:
Previously it was unset and defaulted to --crosstool_top in whatever configuration was being used, but that
doesn't make sense for android/apple/etc builds.

Instead we use getNormalized to set --host_crosstool_top in the target configuration.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
@@ -560,21 +560,35 @@ public class CppOptions extends FragmentOptions {
               + "except bar.o.")
   public List<PerLabelOptions> perFileLtoBackendOpts;
 
+  /** See {@link #hostCrosstoolTop} documentation. * */
+  private static final String HOST_CROSSTOOL_TOP_NOT_YET_SET = "HOST CROSSTOOL TOP NOT YET SET";
+
+  /**
+   * The value of "--crosstool_top" to use for building tools.
+   *
+   * <p>We want to make sure this stays bound to the top-level configuration when not explicitly set
+   * (as opposed to a configuration that comes out of a transition). Otherwise we risk using the
+   * wrong crosstool (i.e., trying to build tools with an Android-specific crosstool).
+   *
+   * <p>To accomplish this, we initialize this to a special value that means "I haven't been set
+   * yet" and use {@link #getNormalized} to rewrite it to {@link #hostCrosstoolTop} <b>only</b> from
+   * that default. Blaze always evaluates top-level configurations first, so they'll trigger this.
+   * But no followup transitions can.
+   */
   @Option(
-    name = "host_crosstool_top",
-    defaultValue = "null",
-    converter = LabelConverter.class,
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
```diff
@@ -1004,10 +1018,20 @@ public Label getFdoPrefetchHintsLabel() {
...
+    CppOptions newOptions = (CppOptions) this.clone();
+    boolean changed = false;
     if (targetLibcTopLabel != null
         && targetLibcTopLabel.getName().equals(TARGET_LIBC_TOP_NOT_YET_SET)) {
-      CppOptions newOptions = (CppOptions) this.clone();
       newOptions.targetLibcTopLabel = libcTopLabel;
+      changed = true;
+    }
+    if (hostCrosstoolTop != null
+        && hostCrosstoolTop.getName().equals(HOST_CROSSTOOL_TOP_NOT_YET_SET)) {
```
gregestren, line 1029: Do we actually need a special marker value or does null still work here? As the code stands now, what case does the `!= null` check catch?

It took me a lot of time to follow the logic from https://github.com/bazelbuild/bazel/commit/6720ff31ba45b6225993cf7a0ccf50ad2001f0e9, which set this basic pattern. Even with the crucial comments at https://github.com/bazelbuild/bazel/blob/6720ff31ba45b6225993cf7a0ccf50ad2001f0e9/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java#L653-L656.

My recollection now is that null default doesn't work for `targetLibcTopLabel` because the similarly named `libcTopLabel` does default to null. So if you build in a default target configuration but with [`--android_grte_top`](https://github.com/bazelbuild/bazel/blob/18ee6bacdb079432d71bbcbefeb5da8586dbd05f/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java#L235)=`<something non-null>`, the top-level version of `getNormalized` would set `targetLibcTopLabel = libcTopLabel = null`, and then the Android version would re-trigger because it thought `null` meant "never set".

`host_crosstool_top` isn't quite the same, is it? Because the equivalent to `libcTopLabel` is `crosstoolTop`, which itself doesn't default to null? 

Whatever logic we want to use, can we document these obscure subtleties super-precisely?

gregestren:
Also, did a particular use case catch this issue?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
katre: You're right, using "null" is fine for host_crosstool_top. Let me know if the current documentation is insufficient.

katre:
This was caught when I was testing the change to use toolchain transitions for Java toolchains. Specifically, building an android binary (which transitions the crosstool_top to an android-specific crosstool) stopped working, because Bazel was using the android crosstool to try and build tools.

This wasn't a problem previously because the single host transition was created before the android transition had happened, and so saw the original crosstool_top.

gregestren (approved):
That's indeed a large test! But looks good, and I think simpler than it was before.

PR #12097, opened 2020-09-12
Fix using Virtual File System for Git

loshca:
see #12098


meteorcloudy:
Great, thanks for the fix!

PR #12106, opened 2020-09-15
Add CallCredentialsProvider and refresh credentials in ByteStreamUploader

coeuvre:
Users may get authentication error in the mid of a long remote build due to credentials timeout. This PR:
1. Add `CallCredentialsProvider` which can be used by gRPC clients to refresh credentials.
2. Update `ByteStreamUploader.java` to use the provider and refresh credentials on authentication error.

The next step would be updating other places where using `CallCrendentials` currently (e.g. `RemoteCacheClient.java`) to use this provider and refresh credentials when necessary.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthCallCredentialsProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthCallCredentialsProvider.java
@@ -1,0 +1,48 @@
+package com.google.devtools.build.lib.authandtls;
+
+import com.google.auth.Credentials;
+import com.google.common.base.Preconditions;
+import com.google.devtools.build.lib.concurrent.ThreadSafety.ThreadSafe;
+import io.grpc.CallCredentials;
+import io.grpc.auth.MoreCallCredentials;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+@ThreadSafe
+public class GoogleAuthCallCredentialsProvider implements CallCredentialsProvider {
+
+  private final Credentials credentials;
+
+  private CallCredentials callCredentials;
+  private long lastRefreshTime;
+
+  public GoogleAuthCallCredentialsProvider(Credentials credentials) {
+    Preconditions.checkNotNull(credentials, "credentials");
+    this.credentials = credentials;
+
+    callCredentials = MoreCallCredentials.from(credentials);
+  }
+
+  @Override
+  public CallCredentials getCallCredentials() {
+    synchronized (this) {
+      return callCredentials;
+    }
+  }
+
+  @Override
+  public void refresh() throws IOException {
+    synchronized (this) {
+      long now = System.currentTimeMillis();
+      // Call credentials.refresh() at most once per second. The one second was arbitrarily chosen,
+      // as a small enough value that we don't expect to interfere with actual token lifetimes, but
+      // it should just make sure that potentially hundreds of threads don't call this method
+      // at the same time.
+      if ((now - lastRefreshTime) > TimeUnit.SECONDS.toMillis(1)) {
+        lastRefreshTime = now;
+        credentials.refresh();
+        callCredentials = MoreCallCredentials.from(this.credentials);
+      }
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


coeuvre:
cc @buchgr @ulfjack

coeuvre:
When authentication with google credentials, we use `GoogleAuthLibraryCallCredentials` which is a wrapper around `OAuth2Credentials`. The `OAuth2Credentials` itself will refresh tokens when gRPC clients constructing a new call using its `getRequestMetadata` method. So we typically doesn't see authentication error due to token expired.

However, for a long remote build, if a gRPC call is constructed at `time1` and haven't be sent to the server (e.g. waiting for available connections), the token is about to be expired at `time2`, so the refresh isn't triggered at `time1`. The request is sent to server at `time3` and will get an unauthenticated error.

By using a `CallCredentialsProvider` instead of using `CallCredentials` directly, we can:
1. Manually refresh credentials when necessary
2. Don't rely on implementation details of `CallCredentials` for credentials refresh

ulfjack (commented):
The change generally looks good, but I'm not sure why only the ByteStreamUploader calls refresh - shouldn't all of the classes that make calls potentially call refresh?

coeuvre:
Yes, all the classes that use `CallCredentials` should be replaced with `CallCredentialsProvider`. I want to get feedbacks first and keep the PR small. Will update other classes after this PR approved.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -350,15 +349,26 @@ public ReferenceCounted touch(Object o) {
...
       AtomicLong committedOffset = new AtomicLong(0);
+
+      ListenableFuture<Void> callFuture = callAndQueryOnFailureWithRetrier(ctx, committedOffset, progressiveBackoff);
+
+      callFuture = Futures.catchingAsync(
+          callFuture,
+          Exception.class,
+          (e) -> {
+            Status status = Status.fromThrowable(e);
+            if (status != null && status.getCode() == Code.UNAUTHENTICATED) {
```
ulfjack, line 360: I'd also check for Code.PERMISSION_DENIED here.

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthCallCredentialsProvider.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public CallCredentials getCallCredentials() {
+    return callCredentials;
+  }
+
+  @Override
+  public void refresh() throws IOException {
+    synchronized (this) {
```
ulfjack, line 32: Upstream is already thread-safe, so this isn't needed.

ulfjack (commented):
I'm a bit concerned that someone in the future might forget to call refresh. Consider calling refreshIfExpired() in getCallCredentials() ?

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
ulfjack: I double-checked the documentation, and I'm not sure in which cases a service will return UNAUTHENTICATED vs. PERMISSION_DENIED. It seems like it won't hurt to also check PERMISSION_DENIED here, just in case?

coeuvre:
In the case of `GoogleAuthCallCredentialsProvider`, the returned `CallCredential` will be refreshed automatically since it [use](https://github.com/grpc/grpc-java/blob/efc14bd299f83850fa2f2a34623c86c7b10825a7/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java#L110) `Credentials` from `com.google.auth` whose all implementations call `refreshIfExpired()` internally if it supports token refresh (e.g. [OAuth2Credentials](https://cs.opensource.google/googleapis/google-auth-library-java/+/master:oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java;l=144?q=OAuth2Credentials)). For other implementations of `CallCredentialsProvider`, we definitely want to call `refreshIfExpired()`.

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthCallCredentialsProvider.java
coeuvre: IIUC, we need this guard for `lastRefreshTime`.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
coeuvre: Done.

coeuvre:
I have submitted internally for another PR depending on this. Feel free to leave comments if there are other concerns.

PR #12112, opened 2020-09-16
ByteStreamUploader: accept ALREADY_EXISTS

### LINKED ISSUE #12111: Bazel should accept ALREADY_EXISTS for byte stream uploads

`com.google.devtools.build.lib.remote.ByteStreamUploader` only treats uploads as successful when they return an OK status. Naively, this seems fine - Bazel first checks if the file exists remotely and then uploads it. However, this is a check-then-act race, where either another thread in the same Bazel process or another Bazel process may finish an upload of the same file between the check and the upload (or even while the upload is ongoing). In that case, the CAS service might return `ALREADY_EXISTS` for this file, which Bazel currently treats as a non-retryable error.

While the documentation does not clearly indicate what the CAS service should do if a file already exists, the `ALREADY_EXISTS` return code seems appropriate.

Build a CAS service that returns `ALREADY_EXISTS` and run a build where multiple actions have the same input. Bazel currently does not deduplicate uploads.

```
ERROR: /home/ulfjack/Google/bazel2/foo/BUILD:15:9: Executing genrule //foo:bar-0 failed (Exit 34): com.google.devtools.build.lib.remote.BulkTransferException
	at com.google.devtools.build.lib.remote.RemoteCache.waitForBulkTransfer(RemoteCache.java:225)
...
	Suppressed: java.io.IOException: io.grpc.StatusRuntimeException: ALREADY_EXISTS: Digest ca6e692227cd35f862d56f0ef4cb0a3ba06cc86674cff63d8b34a5a35d4999a2/159 already exists
		at com.google.devtools.build.lib.remote.ByteStreamUploader.lambda$uploadBlobAsync$1(ByteStreamUploader.java:247)
		at com.google.common.util.concurrent.AbstractCatchingFuture$AsyncCatchingFuture.doFallback(AbstractCatchingFuture.java:175)
		at com.google.common.util.concurrent.AbstractCatchingFuture$AsyncCatchingFuture.doFallback(AbstractCatchingFuture.java:162)
		at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:107)
...
		at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:584)
		at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
		at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
		... 3 more
	Caused by: io.grpc.StatusRuntimeException: ALREADY_EXISTS: Digest ca6e692227cd35f862d56f0ef4cb0a3ba06cc86674cff63d8b34a5a35d4999a2/159 already exists
		at io.grpc.Status.asRuntimeException(Status.java:523)
		... 23 more
```

Linux.

ulfjack:
If the service returns an ALREADY_EXISTS error, then we assume that the
proper file is present remotely.

Fixes #12111.

Change-Id: I7f5af0cca2e5efea32067b092f619c6593af0aac

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
@@ -350,27 +350,14 @@ class ByteStreamUploader extends AbstractReferenceCounted {
       Context ctx = Context.current();
       ProgressiveBackoff progressiveBackoff = new ProgressiveBackoff(retrier::newBackoff);
       AtomicLong committedOffset = new AtomicLong(0);
-      return Futures.transformAsync(
-          retrier.executeAsync(
-              () -> {
-                if (committedOffset.get() < chunker.getSize()) {
-                  return ctx.call(() -> callAndQueryOnFailure(committedOffset, progressiveBackoff));
-                }
-                return Futures.immediateFuture(null);
-              },
-              progressiveBackoff),
-          (result) -> {
-            long committedSize = committedOffset.get();
-            long expected = chunker.getSize();
-            if (committedSize != expected) {
-              String message =
-                  format(
-                      "write incomplete: committed_size %d for %d total", committedSize, expected);
-              return Futures.immediateFailedFuture(new IOException(message));
+      return retrier.executeAsync(
+          () -> {
+            if (committedOffset.get() < chunker.getSize()) {
+              return ctx.call(() -> callAndQueryOnFailure(committedOffset, progressiveBackoff));
             }
             return Futures.immediateFuture(null);
           },
-          MoreExecutors.directExecutor());
+          progressiveBackoff);
     }
 
     private ByteStreamFutureStub bsFutureStub() {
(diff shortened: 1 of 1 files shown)


ulfjack:
@EricBurnett to double-check my reasoning here.

EricBurnett:
SGTM on the change to ignore ALREADY_EXISTS - no harm in it.

Unconvinced on servers returning it though - do we need to document anything more explicitly in the REAPI?  I think semantically ALREADY_EXISTS means what it says on the tin, but I suspect most clients of the REAPI ByteStream will not special-case it today, as you're seeing here. So while I think it's completely reasonable for _clients_ to tolerate it, I wouldn't recommend any _servers_ return it today, not without something stronger in the API text (_"SHOULD treat ALREADY_EXISTS the same as a successful upload..."_). Which would take some discussion in the community.

You can get around all this by returning OK instead of ALREADY_EXISTS, including without having first read all the incoming bytes - gRPC allows this (see https://github.com/grpc/grpc/issues/7032).

And so that's my recommendation in general - anywhere you're looking to return ALREADY_EXISTS you'd probably be better off returning a preemptive 'OK' instead. Most clients (including this SDK) should work out of the box with that, and I'm not aware of any semantic reason to need to return ALREADY_EXISTS (clients don't actually care, they just want to know the file is uploaded).

ulfjack:
We are doing that right now. The reason I was looking at the code is that we have tested one closed-source client which seems to misbehave when we prematurely return OK, although we're not sure why.

coeuvre:
@ulfjack Can you rebase and fix the tests?

PR #12115, opened 2020-09-16
HttpCacheClient: make upload async

### LINKED ISSUE #6091: Jobs generating large amounts of files drastically slow down building with remote cache.

Building the go standard library with remote caching enabled is considerably slower than building locally.

IRL discussion with @buchgr resulted in the conclusion that the contents of the standard library were uploaded serially within Bazel, which lead to major performance decreases. 
It was noted that some concurrency is present for the rest of the build, since Bazel executes multiple jobs at once, resulting in pseudo-concurrent results; but since the go standard library counts as a single job, the upload is then performed serially.

Decrease build time when populating empty remote caches by allowing concurrent uploads of files in individual actions.

Build any go project using an empty remote cache; on normal internet connections, this can take a _long_ time as the go standard library files are uploaded one by one.

### Example timings, running bazel test on one of our projects with a clean local cache:
With remote caching:
```
bazel test --remote_http_cache=https://storage.googleapis.com/<our cache>  
0.07s user 0.15s system 0% cpu 14:07.77 total
```
No remote caching:
```
bazel test
0.05s user 0.08s system 0% cpu 2:46.71 total
```

ulfjack:
This matches the behavior for download, and improves performance for actions
with a lot of outputs.

The code was already using futures in the API and in the implementation,
but the glue code blocked on the internal future and returned an immediate
future to the caller, rather than doing it async.

Fixes #6091.

Change-Id: I3151aa96b879323e0000d3209f6b9bc8be0066d4

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
@@ -616,8 +616,9 @@ public final class HttpCacheClient implements RemoteCacheClient {
                               } catch (IOException e) {
                                 result.setException(e);
                               }
+                            } else {
+                              result.setException(cause);
                             }
-                            result.setException(cause);
                           }
                         });
               });
--- a/src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
@@ -16,9 +16,11 @@ package com.google.devtools.build.lib.remote.http;
 import build.bazel.remote.execution.v2.ActionResult;
 import build.bazel.remote.execution.v2.Digest;
 import com.google.auth.Credentials;
+import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.hash.HashingOutputStream;
+import com.google.common.io.Closeables;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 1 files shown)


ulfjack:
@coeuvre

coeuvre:
Thanks for the PR! LGTM. But some tests failed, any idea with that?

ulfjack:
It was a bug in the change which I just fixed.

ulfjack:
All tests pass!

coeuvre (approved):

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
```diff
@@ -285,64 +287,58 @@ public void channelCreated(Channel ch) {
...
+            .addListener(
+                    (Future<Channel> channelAcquired) -> {
+                      if (!channelAcquired.isSuccess()) {
+                        channelReady.setFailure(channelAcquired.cause());
+                        return;
+                      }
+
+                      try {
+                        Channel ch = channelAcquired.getNow();
+                        ChannelPipeline p = ch.pipeline();
```
meisterT, line 303: nit: can you please change `p` to `pipeline` (or similar)?

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
```diff
@@ -588,85 +583,97 @@ public void close() {
...
-        }
-        in.close();
-        if (ch != null) {
-          releaseUploadChannel(ch);
-        }
-      }
+      SettableFuture<Void> result = SettableFuture.create();
+      acquireUploadChannel()
+          .addListener(
+              (Future<Channel> chP) -> {
```
meisterT, line 590: nit: `channelPromise`?

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
```diff
@@ -588,85 +583,97 @@ public void close() {
...
+            });
+  }
+
   @Override
   public ListenableFuture<Void> uploadFile(Digest digest, Path file) {
-    try (InputStream in = file.getInputStream()) {
-      uploadBlocking(digest.getHash(), digest.getSizeBytes(), in, /* casUpload= */ true);
-    } catch (IOException | InterruptedException e) {
+    try {
+      return uploadAsync(digest.getHash(), digest.getSizeBytes(), file.getInputStream(), true);
```
meisterT, line 660: nit: please re-add `/* casUpload= */`
(same a few lines below and line 695)

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
```diff
@@ -588,85 +583,97 @@ public void close() {
...
+                        (f) ->{
+                          releaseUploadChannel(ch);
+                          if (f.isSuccess()) {
+                            result.set(null);
+                          } else {
+                            Throwable cause = f.cause();
+                            if (cause instanceof HttpException) {
+                              HttpResponse response = ((HttpException) cause).response();
+                              try {
+                                // If the error is due to an expired auth token and we can reset the input stream,
```
meisterT, line 608: nit: line too long

meisterT (commented):
Some minor nits

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
ulfjack: Fixed. I changed the download side as well.

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
ulfjack: Changed. I didn't change the download side.

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
ulfjack: Sorry about that. Not sure how that got lost.

src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java
ulfjack: Fixed. Also a couple more of these.

ulfjack:
PTAL

meisterT (approved):

PR #12119, opened 2020-09-17
docfix: Replace Skydoc reference with Stardoc, as former is deprecated

thundergolfer:
I just followed the Skydoc link from https://docs.bazel.build/versions/master/skylark/deploying.html#documentation and see that Skydoc is "heavily deprecated" in favour of Stardoc, so I'm not sure if Skydoc should still be referenced in the latest Bazel  documentation.


gregestren:
Paging @c-parsons for domain expertise.

gregestren:
CC @laurentlb

c-parsons (approved):

Issue #12113, opened 2020-09-16
Duplicate file uploads over gRPC


ulfjack:
### Description of the problem / feature request:

Bazel does not deduplicate file uploads across actions, so it may attempt multiple uploads of the same content in parallel. If these are large files, then this causes significant network overhead.

For remote builds of TensorFlow over the public internet, this makes the build take a very long time.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
[genrule(
  name = "bar-%s" % i,
  srcs = ["input.txt"],
  outs = ["bar-%s.txt" % i],
  cmd = "cp $< $@",
) for i in range(0, 100)]
```

Enable a remote gRPC cache or executor.

### What operating system are you running Bazel on?

Linux.

###  Have you found anything relevant by searching the web?

This was previously also reported here, but I could not find a matching bug report:
https://github.com/bazelbuild/remote-apis/pull/131#issuecomment-615312235

@EdSchouten

ulfjack:
I have a patch to deduplicate uploads, which has shown significant reductions in build times across a number of different builds and network configurations.

ulfjack:
My patch is here: https://github.com/ulfjack/bazel/tree/blob-existence-cache

I am currently in the process of cleaning it up and splitting it into smaller PRs.

moroten:
I have been using (a rebased variant of) https://github.com/moroten/bazel/commits/optimize-remote-execution for a while now. With these patches, I usually run into limited CPU capacity on my laptop, with no network or remote execution limitations.

One part of the patch chain is the in memory caches, which I've put command line options to adjust the size of. That seems like a clumsy solution, it would be better to automatically scale the size according to available memory. @ulfjack What are your thoughts about the memory management? I see that you use `ConcurrentHashMap` on your branch as well.

ulfjack:
We haven't seen any memory issues so far, but you're right @moroten, it's probably a good idea to limit memory use. I don't have a good proposal at this time.

moroten:
We keep the Bazel server running "forever" on our CI servers, so we will eventually fill the memory. One idea we have is to port our C++ rules to Starlark and thereby avoid loading .d files. That will mean that I can turn off the CAS and the memory footprint of the AC can increase.

Is one option to have a command line flag saying either specify number of bytes to use or, if specified with % suffix, a percentage of available memory when the Bazel server starts? How is the memory management handled more generally in Bazel?

[...]

PR #12120, opened 2020-09-17
RemoteExecutionCache: extract single-digest method

ulfjack:
This is in preparation for adding a cache to this class that deduplicates
uploads.

Progress on #12113.

Change-Id: Ic39347a7a7a8dc7cfd463d78f0a80e0d26a970bc

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
@@ -66,10 +66,6 @@ public class RemoteExecutionCache extends RemoteCache {
 
     List<ListenableFuture<Void>> uploadFutures = new ArrayList<>();
     for (Digest missingDigest : missingDigests) {
-      if (!missingDigests.contains(missingDigest)) {
-        continue;
-      }
-
       uploadFutures.add(uploadBlob(missingDigest, merkleTree, additionalInputs));
     }
 
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
@@ -20,7 +20,9 @@ import build.bazel.remote.execution.v2.Digest;
 import build.bazel.remote.execution.v2.Directory;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
 import com.google.devtools.build.lib.remote.common.RemoteCacheClient;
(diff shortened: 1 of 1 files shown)


ulfjack:
@coeuvre

ulfjack:
@coeuvre

ulfjack:
Whoops, sorry for adding you twice. That wasn't intentional.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
```diff
@@ -58,36 +63,43 @@ public void ensureInputsPresent(MerkleTree merkleTree, Map<Digest, Message> addi
...
     ImmutableSet<Digest> missingDigests =
         getFromFuture(cacheProtocol.findMissingDigests(allDigests));
-    List<ListenableFuture<?>> uploadFutures = new ArrayList<>();
+
+    List<ListenableFuture<Void>> uploadFutures = new ArrayList<>();
     for (Digest missingDigest : missingDigests) {
-      Directory node = merkleTree.getDirectoryByDigest(missingDigest);
-      if (node != null) {
-        uploadFutures.add(cacheProtocol.uploadBlob(missingDigest, node.toByteString()));
+      if (!missingDigests.contains(missingDigest)) {
```
coeuvre, line 69: IIUC, `missingDigest` is an element of `missingDigests`, so this statement is always `false`. My guess is your following PRs will share `missingDigests` between actions but can we remove this check for this PR since I am not sure whether it will trigger internal checks and prevent us from submitting.

coeuvre (approved):
Thanks!

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
ulfjack: Fixed! I'm not sure how that happened.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
```diff
@@ -29,11 +31,14 @@
...
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** A {@link RemoteCache} with additional functionality needed for remote execution. */
 public class RemoteExecutionCache extends RemoteCache {
+  private final ConcurrentHashMap<Digest, ListenableFuture<Void>> uploadFutures = new ConcurrentHashMap<>();
```
meisterT, line 41: I think this can be removed.

ulfjack: Maybe I shouldn't be sending Bazel patches in the middle of the night. :-/

meisterT (approved):

Issue #4008, opened 2017-11-02
--action_env not forwarded through "tools" dependency


didzis:
### Description of the problem:

Variables set by `--action_env` are not forwarded through `tools` dependency of a rule. To generate test example, paste following lines into terminal:

```shell
mkdir test
cd test
touch WORKSPACE
touch tool.sh
chmod +x tool.sh
cat > BUILD << EOF
genrule(
    name = "rule1",
    srcs = [],
    outs = ["out1.c"],
    cmd = "bash -c 'echo RULE1:TESTVAR=\$\$TESTVAR'; exit 1",
)

sh_binary(
    name = "tool",
    srcs = ["tool.sh"],
    # srcs = [":rule1"],
    data = [":rule1"],
)

genrule(
    name = "rule2",
    srcs = [],
    outs = ["out2.c"],
    cmd = "",
    tools = [":tool"],
)

genrule(
    name = "rule3",
    srcs = [],
    outs = ["out3.c"],
    cmd = "",
    tools = [":rule1"],
)

genrule(
    name = "rule4",
    srcs = [":rule1"],
    outs = ["out4.c"],
    cmd = "",
)

cc_binary(
    name = "main",
    srcs = [
        # ":rule1",
        # ":rule2",
        # ":rule3",
        # ":rule4",
    ],
    data = [
        # ":rule1",
        ":rule2",
        # ":rule3",
        # ":rule4",
        # ":tool",
    ],
)
EOF
bazel build :main --action_env TESTVAR=hello
```
outputs
```
RULE1:TESTVAR=
```
however depending on what is uncommented, i.e., `":rule1",`, `":rule2",`, `":rule3",`, `":rule4",`, `":tool",` at the end of BUILD, bazel may output also
```
RULE1:TESTVAR=hello
```
`rule1` gives `RULE1:TESTVAR=hello` - rule1 is directly referenced
`rule2` and `rule3` gives `RULE1:TESTVAR=` - rule1 is referenced via tools option
`rule4` and `tool` again gives `RULE1:TESTVAR=hello` - rule1 is referenced via middleman

If the `rule1` is referenced from the "_tool_" and from another rule as a dependency, it may be randomly executed with different environment set for each run. Is this the expected behaviour?

A consistent behaviour is obtained when exporting the environment variable into bazel environment, i.e.,
```
export TESTVAR="hello world!"
bazel shutdown
bazel build :main
```

[...]

PR #12122, opened 2020-09-17
Add --host_action_env option

mai93:
This PR adds `--host_action_env` option to specify a set of environment variables to be added to the host configuration. The variables specified by this option are only added to host configuration while those specified by `--action_env` are only included in the target configuration.

Fixes: #4008

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -358,9 +358,9 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
           documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
           effectTags = {OptionEffectTag.ACTION_COMMAND_LINES},
           help =
-                  "Specifies the set of environment variables available to actions with host configuration. "
-                          + "Variables can be either specified by name, in which case the value will be "
-                          + "taken from the invocation environment, or by the name=value pair which sets "
+                  "Specifies the set of environment variables available to actions with host or execution "
+                          + "configurations. Variables can be either specified by name, in which case the value "
+                          + "will be taken from the invocation environment, or by the name=value pair which sets "
                           + "the value independent of the invocation environment. This option can be used "
                           + "multiple times; for options given for the same variable, the latest wins, options "
                           + "for different variables accumulate.")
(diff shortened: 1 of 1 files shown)


mai93:
cc @philwo @meteorcloudy

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -342,14 +342,30 @@ public String getTypeDescription() {
       documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
       effectTags = {OptionEffectTag.ACTION_COMMAND_LINES},
       help =
-          "Specifies the set of environment variables available to actions. "
+          "Specifies the set of environment variables available to actions performed using target configuration. "
```
meteorcloudy, line 345: Maybe "with target configuration" sounds a bit simper? Same below.

meteorcloudy (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
mai93: To be "available to actions with target configuration." ?

meteorcloudy: Yes, that's what I mean

gregestren:
I'm going to punt this to @katre. Eventually I think we should move away from needing explicit `--foo` / `--host_foo` flags toward a real API for specifying which flags go to the host and which don't. @katre had some ideas about evolving these themes.

I also acknowledge we won't have that tomorrow. So the above isn't necessarily an objection to this approach.

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -342,14 +342,30 @@ public String getTypeDescription() {
...
 
+  @Option(
+          name = "host_action_env",
+          converter = Converters.OptionalAssignmentConverter.class,
+          allowMultiple = true,
+          defaultValue = "null",
+          documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+          effectTags = {OptionEffectTag.ACTION_COMMAND_LINES},
+          help =
+                  "Specifies the set of environment variables available to actions with host configuration. "
```
katre, line 361: "with host or execution configurations".

Ideally we'd move away from target/host to non-tool/tool, but we're not there now so no need to burden your PR.

katre (approved):
This looks fine with a small fix to the doc comment.

katre:
Currently, host.actionEnvironment is left unset, correct? Then no need to start propogating the target actionEnvironment to the host configuration.

meteorcloudy:
Thanks for the clarification! Mai, can you fix the documentation? Then you can import this change.

katre (approved):

PR #12130, opened 2020-09-18
Add .netrc support for remote build

coeuvre:
This PR enables .netrc support for remote build by default. Existing authentication methods take higher priority than .netrc if they are enabled by flags.

Bazel search for .netrc in the following order:
1. If environment variable `$NETRC` exists, use it as the path to the .netrc file
2. Fallback to `$HOME/.netrc`

If .netrc doesn't exists, Bazel simply go forward without authentication.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
@@ -83,5 +83,15 @@ public class Netrc {
     abstract String login();
     abstract String password();
     abstract String account();
+
+    @Override
+    public String toString() {
+      return "Credential{"
+          + "machine=" + machine() + ", "
+          + "login=" + login() + ", "
+          + "password=<password>" + ", "
+          + "account=" + account()
+          + "}";
+    }
   }
 }
--- a/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
@@ -1,0 +1,85 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.google.devtools.build.lib.authandtls;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.io.InputStream;
+import javax.annotation.Nullable;
+
+public class Netrc {
+
+  @Nullable
+  private final Credential defaultCredential;
+  private final ImmutableMap<String, Credential> credentials;
+
+  public static Netrc fromStream(InputStream inputStream) throws IOException  {
+    return NetrcParser.parseAndClose(inputStream);
+  }
+
+  public Netrc(@Nullable Credential defaultCredential, ImmutableMap<String, Credential> credentials) {
+    this.defaultCredential = defaultCredential;
+    this.credentials = credentials;
+  }
+
+  @Nullable
+  public Credential getCredential(String machine) {
+    return credentials.getOrDefault(machine, defaultCredential);
+  }
+
+  @VisibleForTesting
+  @Nullable
+  Credential getDefaultCredential() {
+    return defaultCredential;
+  }
+
+  @VisibleForTesting
+  ImmutableMap<String, Credential> getCredentials() {
+    return credentials;
+  }
+
+  @AutoValue
+  public static abstract class Credential {
+    public static Builder builder(String machine) {
+      return new AutoValue_Netrc_Credential.Builder().setMachine(machine);
+    }
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+      public abstract String machine();
+      public abstract Builder setMachine(String machine);
+
+      public abstract String login();
+      public abstract Builder setLogin(String value);
+
+      public abstract String password();
+      public abstract Builder setPassword(String value);
+
+      public abstract String account();
+      public abstract Builder setAccount(String value);
+
+      public abstract Credential build();
+    }
+
+    abstract String machine();
+    @Nullable
+    abstract String login();
+    @Nullable
+    abstract String password();
+    @Nullable
+    abstract String account();
+  }
+}
--- a/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
@@ -20,8 +20,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import javax.annotation.Nullable;
 
+/** The content of a .netrc file. */
 public class Netrc {
-
   @Nullable
   private final Credential defaultCredential;
   private final ImmutableMap<String, Credential> credentials;
--- a/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
@@ -53,8 +53,13 @@ public class Netrc {
 
   @AutoValue
   public static abstract class Credential {
+
     public static Builder builder(String machine) {
-      return new AutoValue_Netrc_Credential.Builder().setMachine(machine);
+      return new AutoValue_Netrc_Credential.Builder()
+          .setMachine(machine)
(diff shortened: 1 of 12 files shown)


coeuvre:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/authandtls/NetrcParser.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.ArrayDeque;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+/** A parser used to parse .netrc content */
```
ulfjack, line 34: Is there any documentation on the format? If so, please add a link to it.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public boolean hasRequestMetadataOnly() {
+    return true;
+  }
+
+  @Override
+  public void refresh() throws IOException {
+
```
ulfjack, line 54: unnecessary empty line

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class NetrcCredentials extends Credentials {
+
```
ulfjack, line 14: unnecessary empty line

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.devtools.build.lib.authandtls.Netrc.Credential;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class NetrcCredentials extends Credentials {
```
ulfjack, line 13: missing Javadoc

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
```diff
@@ -0,0 +1,56 @@
...
+    this.netrc = netrc;
+  }
+
+  @Override
+  public String getAuthenticationType() {
+    return "netrc";
+  }
+
+  @Override
+  public Map<String, List<String>> getRequestMetadata(URI uri) throws IOException {
```
ulfjack, line 27: There should be tests for this method.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -881,4 +902,30 @@ void setChannelFactory(ChannelFactory channelFactory) {
   RemoteActionContextProvider getActionContextProvider() {
     return actionContextProvider;
   }
+
+  private Credentials newCredentialsFromNetrc(CommandEnvironment env) throws IOException {
```
ulfjack, line 906: Ideally, this method would be static and have tests. Consider changing the interface to take a client env map and FileSystem, which might be easier to test.

src/test/java/com/google/devtools/build/lib/authandtls/NetrcParserTest.java
```diff
@@ -0,0 +1,235 @@
...
+        fooCredential
+    );
+    assertThat(netrc.getDefaultCredential()).isEqualTo(defaultCredential);
+    assertThat(netrc.getCredentials()).isEqualTo(expectedCredentials);
+    assertThat(netrc.getCredential(fooMachine)).isEqualTo(fooCredential);
+  }
+
+  @Test
+  public void shouldFailWithBadStartingContent() {
+    String content = "this is not netrc syntax";
```
ulfjack, line 197: What do other tools do for bad .netrc files? Do they error out or do they silently ignore it?

ulfjack (commented):
The integration could do with more tests.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcParser.java
eed3si9n: Here's an example of netrc file documentation - https://man.cx/netrc(4)

A useful source in terms of documentation might be https://docs.python.org/3.8/library/netrc.html

src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
```diff
@@ -0,0 +1,87 @@
...
+    return defaultCredential;
+  }
+
+  @VisibleForTesting
+  ImmutableMap<String, Credential> getCredentials() {
+    return credentials;
+  }
+
+  @AutoValue
+  public static abstract class Credential {
```
eed3si9n, line 55: Does `@AutoValue` produce `toString` that would leak the password?
If so there should be an override for it that would display as `<password>` or something.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcParser.java
coeuvre: Added.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
coeuvre: Removed.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
coeuvre: Removed.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
coeuvre: Added.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcCredentials.java
coeuvre: Added.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcParser.java
gibfahn: Maybe link to the existing Starlark netrc parser in this repo: https://github.com/bazelbuild/bazel/blob/4c0febc02fab5e9fda3f40532a62854fb1c45631/tools/build_defs/repo/utils.bzl#L203-L204

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
coeuvre: Changed and added few tests.

src/test/java/com/google/devtools/build/lib/authandtls/NetrcParserTest.java
coeuvre: curl just silently ignore it. Should we also ignore the bad .netrc files silently?

philwo (approved):

src/main/java/com/google/devtools/build/lib/authandtls/Netrc.java
coeuvre: Updated.

src/main/java/com/google/devtools/build/lib/authandtls/NetrcParser.java
coeuvre: Added.

src/test/java/com/google/devtools/build/lib/authandtls/NetrcParserTest.java
wisechengyi: We've ran into issues python `requests` module ignores the error silently as well, not my favorite behavior, but I think it's fine to silently ignore here to keep it consistent.

Although just to be ideal, if there's a flag that can change the behavior, that'll be even better. (not blocking here)

wisechengyi (commented):
Thanks, @coeuvre!

src/test/java/com/google/devtools/build/lib/authandtls/NetrcParserTest.java
gibfahn: Would be good to respect `#`-style comments, as they're pretty common in netrcs.

Otherwise we should probably do whatever the starlark netrc parser does.

coeuvre: > but I think it's fine to silently ignore here to keep it consistent.

Done.

> Would be good to respect #-style comments, as they're pretty common in netrcs.

Done.

coeuvre:
This is committed as e379e942e2c4440f9c9288dcd397fa5c1d04ed57, 0077eaa64db1dffeccd309b7ecb0078e983b9550 and ae28ab7d911460839c1db65c0c325717fdeaa13b with some changes made internally.

Issue #11748, opened 2020-07-10
Bazel can build different tools on different execution platforms in the same build


gregestren:
For example, building on a Mac where some tools execute on the Mac and others execute on a remote Linux machine.

This tracks a [Blaze configurability roadmap](https://bazel.build/roadmaps/configuration.html) effort.

cpsauer:
Came across this looking to build and bundle together binaries, one for each platform, some of which would need to get built remotely. This would be awesome for that!

Could I ask what use case originally inspired this feature request? 

Also, @gregestren: I was just building on some of your great examples around transitions. Thank you for those!

katre:
We have mixed-backend builds we'd like to do, mostly around developers who are working on Mac laptops but want to use the Linux remote build backends, so we'd like to get that moving. There's a lot of cruft to clean up first, though, so I'm not sure when we'll be moving forward.

Mixed binaries like you want can be done with transitions, however, so I am glad you found the examples for those helpful.

ulfjack:
We have successfully run some builds where Bazel runs on a Mac and the build runs on Linux, although that required a patch to the bazel-toolchains repository which currently looks at the local host to determine the remote host (which is obviously incorrect in this case).

cpsauer:
Thanks, both of you, for replying. 

@katre, just to confirm, I *can't* currently use transitions to have parts of the graph build remotely, right? That's what it seemed like from my experience playing around with transitions, and this seemed like the issue that would lead to the change from reading the design doc ([link if anyone else is looking](https://docs.google.com/document/d/1U9HzdDmtRnm244CaRM6JV-q2408mbNODAMewcGjnnbM/edit#heading=h.y890cog3hszc)). But your comment seemed to say otherwise? Just to make it concrete: imagine building on a Mac, where you can locally (cross)compile the binary for macOS, iOS, Android (using transitions), but want to compile for Windows, Linux remotely. Or equivalently compiling from Linux, able to build for Linux, Android locally, but needing to compile for iOS, macOS, Windows remotely.

ulfjack:
@cpsauer, even if you could, the code still has other problems. In particular, it looks at the OS that Bazel runs on to select things like path style (windows with backslashes vs. linux/mac with forward slashes). Right now, even running Windows actions remotely from a Linux/Mac machine (and vice versa) is not working.

katre:
[...]

PR #12132, opened 2020-09-18
Add the OS as a field to the configuration

ulfjack:
This helps move Bazel towards being able to perform cross-platform builds
with a remote execution system that's running a different OS. This is in
preparation for updating the rule implementations to use the configuration
setting instead of OS.current() directly.

I'm not sure how this interacts with platforms, but this change seems safe
enough.

Progress on #11748.

Change-Id: I20dff1fb332d4567b621d65e43b05a00905f75b0

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
@@ -126,8 +126,23 @@ public final class CommandHelper {
    * [1] https://msdn.microsoft.com/en-us/library/ms682425(VS.85).aspx
    * [2] https://support.microsoft.com/en-us/kb/830473.
    */
+  // TODO(ulfjack): Remove this field (this may be referenced by non-oss code).
   @VisibleForTesting
-  public static int maxCommandLength = OS.getCurrent() == OS.WINDOWS ? 8000 : 64000;
+  public static int maxCommandLength = getMaxCommandLength(OS.getCurrent());
+
+  /**
+   * Maximum total command-line length, in bytes, not counting "/bin/bash -c ".
+   * If the command is very long, then we write the command to a script file,
+   * to avoid overflowing any limits on command-line length.
+   * For short commands, we just use /bin/bash -c command.
+   *
+   * Maximum command line length on Windows is 32767[1], but for cmd.exe it is 8192[2].
+   * [1] https://msdn.microsoft.com/en-us/library/ms682425(VS.85).aspx
+   * [2] https://support.microsoft.com/en-us/kb/830473.
+   */
+  private static int getMaxCommandLength(OS os) {
+    return os == OS.WINDOWS ? 8000 : 64000;
+  }
 
   /** {@link RunfilesSupplier}s for tools used by this rule. */
   private final Sequence<RunfilesSupplier> toolsRunfilesSuppliers;
@@ -282,7 +297,7 @@ public final class CommandHelper {
       RuleContext ruleContext, String command, CommandConstructor constructor) {
     List<String> argv;
     Artifact scriptFileArtifact = null;
-    if (command.length() <= maxCommandLength) {
+    if (command.length() <= getMaxCommandLength(ruleContext.getHostConfiguration().getOS())) {
       argv = constructor.asExecArgv(command);
     } else {
       // Use script file.
@@ -302,18 +317,18 @@ public final class CommandHelper {
    */
   @Nullable
   public static Artifact commandHelperScriptMaybe(
-      RuleContext ruleCtx, String command, CommandConstructor constructor) {
-    if (command.length() <= maxCommandLength) {
+      RuleContext ruleContext, String command, CommandConstructor constructor) {
+    if (command.length() <= getMaxCommandLength(ruleContext.getHostConfiguration().getOS())) {
       return null;
     } else {
-      return constructor.commandAsScript(ruleCtx, command);
+      return constructor.commandAsScript(ruleContext, command);
     }
   }
 
   /**
    * Builds the set of command-line arguments using the specified shell path. Creates a bash script
-   * if the command line is longer than the allowed maximum {@link #maxCommandLength}. Fixes up the
-   * input artifact list with the created bash script when required.
+   * if the command line is longer than some allowed maximum. Fixes up the input artifact list with
+   * the created bash script when required.
    */
   public List<String> buildCommandLine(
       String command, NestedSetBuilder<Artifact> inputs, CommandConstructor constructor) {
@@ -327,8 +342,8 @@ public final class CommandHelper {
 
   /**
    * Builds the set of command-line arguments. Creates a bash script if the command line is longer
-   * than the allowed maximum {@link #maxCommandLength}. Fixes up the input artifact list with the
-   * created bash script when required.
+   * than some allowed maximum. Fixes up the input artifact list with the created bash script when
+   * required.
    */
   public List<String> buildCommandLine(
       String command, List<Artifact> inputs, CommandConstructor constructor) {
--- a/src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java
@@ -165,10 +165,7 @@ public abstract class TestStrategy implements TestActionContext {
   public static ImmutableList<String> expandedArgsFromAction(TestRunnerAction testAction)
       throws CommandLineExpansionException {
     List<String> args = Lists.newArrayList();
-    // TODO(ulfjack): `executedOnWindows` is incorrect for remote execution, where we need to
-    // consider the target configuration, not the machine Bazel happens to run on. Change this to
-    // something like: testAction.getConfiguration().getTargetOS() == OS.WINDOWS
-    final boolean executedOnWindows = (OS.getCurrent() == OS.WINDOWS);
+    final boolean executedOnWindows = testAction.getConfiguration().getOS() == OS.WINDOWS;
 
     Artifact testSetup = testAction.getTestSetupScript();
     args.add(testSetup.getExecPath().getCallablePathString());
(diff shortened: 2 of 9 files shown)


ulfjack:
@katre

src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
```diff
@@ -282,7 +297,7 @@ public String expandLabelsHeuristically(String expr) {
       RuleContext ruleContext, String command, CommandConstructor constructor) {
     List<String> argv;
     Artifact scriptFileArtifact = null;
-    if (command.length() <= maxCommandLength) {
+    if (command.length() <= getMaxCommandLength(ruleContext.getHostConfiguration().getOS())) {
```
katre, line 300: Most of the uses you have seem to be using it with `ruleContext.getHostConfiguration().getOS()`. This is a bit inflexible given the presence of [execution groups](https://docs.bazel.build/versions/3.5.0/exec-groups.html), but we do have a way to tell what platform is being used as the execution platform:
```
ruleContext.getToolchainContext(execGroupName).executionPlatform();
```

You can then call `executionPlatform.constraints().hasConstraintValue()` to see whether the current execution platform has the Windows OS constraint (you'll need an instance of `ConstraintValueInfo`, or we could add an overload that just takes a Label and ignores the configuration).

This could probably all be packaged up more hygenically, but the code is in place.

See the docs for [`RuleContext.getToolchainContext()`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java;bpv=1;bpt=1;l=1208?ss=bazel&q=RuleContext&gsn=getToolchainContext&gs=kythe%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel%3Flang%3Djava%3Fpath%3Dcom.google.devtools.build.lib.analysis.RuleContext%23b5d9b148eb20de8be0d1443d570c851cc8d3eae5961072a9ff0cf471ed53d612) (and the various related methods), and for [`ToolchainContext.executionPlatform()`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/ToolchainContext.java;drc=49c9475aa7db252c19157ee059c40862a412df5d;bpv=1;bpt=1;l=28?gsn=executionPlatform&gs=kythe%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel%3Flang%3Djava%3Fpath%3Dcom.google.devtools.build.lib.analysis.ToolchainContext%232330a5d89a3cbbc320688c988fe74955ca37910f80c897258b25e514ae69d181).

src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java
```diff
@@ -165,10 +165,7 @@ public final boolean isTestKeepGoing() {
   public static ImmutableList<String> expandedArgsFromAction(TestRunnerAction testAction)
       throws CommandLineExpansionException {
     List<String> args = Lists.newArrayList();
-    // TODO(ulfjack): `executedOnWindows` is incorrect for remote execution, where we need to
-    // consider the target configuration, not the machine Bazel happens to run on. Change this to
-    // something like: testAction.getConfiguration().getTargetOS() == OS.WINDOWS
-    final boolean executedOnWindows = (OS.getCurrent() == OS.WINDOWS);
+    final boolean executedOnWindows = testAction.getConfiguration().getOS() == OS.WINDOWS;
```
katre, line 168: Actions also have access to their execution platform via [`ActionAnalysisMetadata.getExecutionPlatform()`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/actions/ActionAnalysisMetadata.java;drc=49c9475aa7db252c19157ee059c40862a412df5d;bpv=1;bpt=1;l=233?gsn=getExecutionPlatform&gs=kythe%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel%3Flang%3Djava%3Fpath%3Dcom.google.devtools.build.lib.actions.ActionAnalysisMetadata%23c6439d067f3aa5d6f841e899332022befabb9829f2f85fb28d571c7b2402034f).

src/main/java/com/google/devtools/build/lib/util/DependencySet.java
```diff
@@ -72,10 +73,18 @@ public void setOutputFileName(String outputFileName) {
   /**
    * Constructs a new empty DependencySet instance.
    */
-  public DependencySet(Path root) {
+  public DependencySet(OS os, Path root) {
```
katre, line 76: (Argh the ordering of methods in this file).

Every DependencySet seems to be created from an Action (either NinjaAction or CppCompileAction), so we can get the execution platform from that.

src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
```diff
@@ -126,8 +126,23 @@ public CommandHelper build() {
    * [1] https://msdn.microsoft.com/en-us/library/ms682425(VS.85).aspx
    * [2] https://support.microsoft.com/en-us/kb/830473.
    */
+  // TODO(ulfjack): Remove this field (this may be referenced by non-oss code).
```
katre, line 129: There is one non-OSS reference, in a test, that would be easily removed.

katre (commented):
I like the idea of removing host OS dependence from the code, but I think the existing mechanisms are there, just not wired up, and I'd like to someday remove the OS enum entirely. See my comments for notes on where the information can be obtained now.

src/main/java/com/google/devtools/build/lib/analysis/CommandHelper.java
ulfjack: What's the relationship between constraints and execution properties?

Instead of using OS, I could check the `OSFamily` setting:
https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/platform.md

That would require adding that to the execution properties by default - otherwise local execution wouldn't work, I think.

If using the execution properties isn't desirable, any suggestions for how to get a constraint value? It would have to be a universal label that all users of Bazel agree upon.

ulfjack: How do you feel about adding a method to Platform that returns an OS instance, as a temporary measure?

katre: Constraints and execution properties are entirely separate: ideally Bazel never even looks inside the execution properties, just hands them off to remote execution. Currently only remote platforms set them, and also the properties defined by the RE API are only valid for platforms intended to be used for remote execution: when we start using platforms to define the properties for sandboxing, for example, they wouldn't make much sense (or we'd be using remote execution API for non-remote execution purposes).

We have several useful constraints defined in [`@bazel_tools//platforms`](https://cs.opensource.google/bazel/bazel/+/master:tools/platforms/BUILD.tools), including OS and CPU constraints. These are actually just aliases for the values in https://github.com/bazelbuild/platforms, which are available to all builds (or else the aliases wouldn't work), so those would be fine to use.

As a shortcut, I'm okay with adding a PlatformInfo.getOS method, but how would we set it? Just add code in PlatformInfo that knows which constraints mean windows, darwin, etc? Is this actually easier than having code that cares about Windows ask if the PlatformInfo has the windows constraint?

ulfjack: I don't see exec properties as being purely for remote execution. For example, exec properties could also be used to configure the sandbox whether it should allow networking or not, or which local paths to map into the sandbox (currently this is only possible through per-build settings, which is not ideal).

For `OSFamily` specifically, I think the constraints and `OSFamily` setting should be consistent, and, ideally, this would be ~enforced by Bazel (for some value of enforced).

Regardless, I'm a bit at a loss as to how I can get the right object into - say - the `TestStrategy` to check for a Windows constraint.

katre: Execution properties as a general concept make sense for all platforms. The specific definitions of the keys and values are dependent on the specific execution strategy being used. Especially since Bazel's remote execution and Google's internal remote execution use different keys, I don't want to tie Bazel into using one specific set.

The constraints matching the execution properties has to be managed by whoever defines the platforms: if Bazel checks at all, that should be inside the remote execution strategy (which is what defines the keys and values), and not generic to all execution strategies (because they could all use different keys and values).

The [`ConstraintCollection`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java) class currently defines `get(ConstraintSettingInfo)` and `hasConstraintValue(ConstraintValueInfo)` methods which take providers, but those providers are fairly lightweight. The [`ConstraintSettingInfo`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintSettingInfo.java) is a wrapper around a Label defining the setting (and a default constraint that could be used). [`ConstraintValueInfo`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintValueInfo.java) is just a `ConstraintSettingInfo` and a label defining the value.

The way we get these for toolchain resolution is to actually fetch the ConfiguredTarget and extract the providers, which is heavyweight for your purposes. Since the OS constraint doesn't provide a default, I am fine with adding an overload of `ConstraintCollection.hasConstraintValue` which takes a Label of a constraint value, and checks to see if any directly declared constraint value matches. This will will any constraint values that are present due to being the default, but this will work for your purposes. Then you can call like this:
```
boolean windows = getExecutionPlatform().constraints().hasConstraintValue(Label.parseAbsoluteUnchecked("@platforms//os:windows");
```

Would that work for you? Adding the overload should be straightforward.

meisterT:
What's the state of this PR?

aiuto:
Any updates?

Issue #6578, opened 2018-11-01
Native CUDA support


cinfra-nvidia:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, ask it on StackOverflow instead: https://stackoverflow.com/questions/tagged/bazel
> - if this is a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

Add native toolchain support for CUDA.

### Feature requests: what underlying problem are you trying to solve with this feature?

Reduce overhead to adopting bazel for CUDA software projects.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Compiling the saxpy example from https://devblogs.nvidia.com/easy-introduction-cuda-c-and-c/ without duplicating all the toolchain work done for something like TensorFlow.

If you name the file with a "*.cu" suffix as is typical for CUDA programs, I do not believe the TensorFlow toolchain works.  Also note that CUDA headers are often suffixed with a "*.cuh"

### What operating system are you running Bazel on?

Linux, Windows

### What's the output of `bazel info release`?

0.17.2

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

https://groups.google.com/forum/#!topic/bazel-discuss/JF8BUqH-hzE
https://stackoverflow.com/questions/46354467/bazel-cuda-add-executable-equivalent
https://github.com/tensorflow/tensorflow/blob/master/third_party/gpus/cuda_configure.bzl
https://github.com/Hibbert-pku/bazel_nvcc

### Any other information, logs, or outputs that you want to share?

N/A

Artem-B:
This is a problem that straddles multiple domains. While it's relatively easy to solve for any given project, solving it once and for all is a bit more complicated. It's probably doable, but will take some work and may need to wait a bit until bazel configurability features mature. In order to make things work for the end user, someone will need to own, develop and maintain something very similar to what's been done in tensorflow. Chances are that that something would end up being more complicated than it's tensorflow's counterpart as it would potentially need to handle more diverse use cases.

Off the top of my head bazel needs to deal with following:

[...]

PR #12134, opened 2020-09-18
Recognize CUDA header file type (.cuh)

liuliu:
This PR mirrors the fix in
https://github.com/bazelbuild/bazel/commit/8804128f52a674b9f8faf3b9534ae156cbad00a1

.cuh is widely used as suffix for CUDA. It is made more prominent since
"cub" (https://nvlabs.github.io/cub/) recently distributed with CUDA
ToolKit 11, hence, now .cuh headers are in every CUDA ToolKit
distribution.

I would like to continue with this PR and fix any breakages that may encounter.


liuliu:
Just noticed @chsigg has to rollback this change for some internal breakages. Would love to know more to close this one out.

oquenchil (approved):
LGTM

liuliu:
@oquenchil do you want to merge it? Fixed the merge conflict earlier today.

c-mita:
This had to be rolledback because it caused internal build failures. We don't yet know why.

chsigg:
> We don't yet know why.

It's because Google internal builds try to build CUDA headers into header modules (a Google internal variant of C++ modules) and fail spectacularly. We can land this in bazel, but not in blaze.

See the comment here: https://github.com/bazelbuild/bazel/issues/6578#issuecomment-586636686

oquenchil:
Ideally, we keep the codebase between Bazel and the internal version as close as possible. Can we control whether we build a type of header into modules through the toolchain? IMO that's the only thing that should be different instead of introducing yet another flag.

PR #12136, opened 2020-09-18
Add a test for sandboxed cc_library builds on Mac.

tetromino:
This is a regression test for the issue in https://github.com/bazelbuild/bazel/pull/12046 - ensure that in sandboxed mode, we don't get error messages about missing tools from our toolchain.

Should be merged after https://github.com/bazelbuild/bazel/pull/12046 is merged (until then, the new test is expected to fail).

### DIFF
--- a/src/test/shell/bazel/cpp_darwin_integration_test.sh
+++ b/src/test/shell/bazel/cpp_darwin_integration_test.sh
@@ -96,5 +96,32 @@ EOF
       fail "Stripping failed, debug symbols still found in the stripped binary"
 }
 
+# Regression test for https://github.com/bazelbuild/bazel/pull/12046
+function test_osx_sandboxed_cc_library_build() {
+  mkdir -p cpp/osx_sandboxed_cc_library_build
+  cat > cpp/osx_sandboxed_cc_library_build/BUILD <<EOF
+cc_library(
+    name = "a",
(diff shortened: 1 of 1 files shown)


trybka:
OK, #12046 is in.

keith:
Awesome thanks for adding this!

src/test/shell/bazel/cpp_darwin_integration_test.sh
```diff
@@ -96,5 +96,22 @@ EOF
...
+  cat > cpp/osx_sandboxed_cc_library_build/BUILD <<EOF
+cc_library(
+    name = "a",
+    srcs = ["a.cc"],
+)
+EOF
+  cat > cpp/osx_sandboxed_cc_library_build/a.cc <<EOF
+void a() { }
+EOF
+  assert_build --spawn_strategy=sandboxed //cpp/osx_sandboxed_cc_library_build:a
```
jmmv, line 111: Is it possible to verify here that the sandbox was really used for this action?

I'm concerned about --spawn_strategy being overridden by some other --strategy flag and not having the right effect for this test.

tetromino: I don't know how to check what strategy is used for the action, but we can check in BEP that no other strategy flag was used.

PR #12154, opened 2020-09-22
Upgrade remote_java_tools versions (javac 11 to v10.0, javac 14 to v2.0)

comius:
Upgrade remote_java_tools versions (javac 11 to v10.0, javac 14 to v2.0)

Fix coverage tests after upgrade of java_tools.
The change in LCOV was made in https://github.com/bazelbuild/bazel/pull/11919, however some of the tests were testing only LCOV formatter packed with java_tools.


comius:
closes https://github.com/bazelbuild/java_tools/issues/32

PR #12156, opened 2020-09-23
Use CallCredentialsProvider and refresh credentials when needed for GrpcCacheClient and GrpcRemoteExecutor

coeuvre:
Follow up on #12106, this PR make changes to `GrpcCacheClient` and `GrpcRemoteExecutor` to use `CallCredentialsProvider` and refresh credentials when needed to avoid unauthenticated error during a long remote build.


coeuvre:
cc @ulfjack

ulfjack:
Looks reasonable. I like that it actually simplifies the code in parts, and also that it now properly retries after credential refresh. I think another improvement might be to use a Retrier implementation for the credential refresh. I think that would simplify the call sites. (Feel free to ignore or defer this last proposal.)

coeuvre:
> I think another improvement might be to use a Retrier implementation for the credential refresh.

Thanks for your suggestion, I like this idea! Will refactor the code in another PR soon.

katre:
Is this likely to submit before October 1? It was flagged to be in the 3.7 release (#12188), and I'd like to figure out if I need to delay cutting RC1 or plan to make an RC2.

philwo:
This is already in internal code review (see cl/334286851). Please wait with cutting rc1 until it's submitted, but I think we can get this in tomorrow.

PR #12170, opened 2020-09-24
Add a mechanism to configure the downloader

shs96c:
This adds a `--experimental_remote_downloader_config` flag to the
`build` command that accepts the path to a configuration file. This
file has a line-based format, with each line starting with a directive
and then the action to take. These directives are:

* `allow hostName` Will allow access to the given host and subdomains
* `block hostName` Will block access to the given host and subdomains
* `rewrite pattern pattern` Rewrite a URL using the given
  pattern. Back references are numbered from `$1`

The directives are applied in the order `rewrite, allow, block'. An
example config may look like:

```
block mvnrepository.com
block maven-central.storage.googleapis.com
block gitblit.github.io
rewrite repo.maven.apache.org/maven2/(.*) artifacts.mycorp.com/libs-release/$1

block github.com
rewrite github.com/([^/]+)/([^/]+)/releases/download/([^/]+)/(.*) artifacts.mycorp.com/github-releases/$1/$2/releases/download/$3/$4
rewrite github.com/([^/]+)/([^/]+)/archive/(.+).(tar.gz|zip) artifacts.mycorp.com/api/vcs/downloadRelease/github.com/$1/$2/$3?ext=$4
```

In addition, you can block all hosts using the `*` wildcard.

Comments within the config file are allowed, and must be on their own
line preceded by a `#`


shs96c:
One useful use-case for this is companies that want to prevent access to the wider internet from their CI systems, but mirror useful packages on their own servers in predictable locations. This allows builds to use third party extensions and rule sets without needing to rewrite them.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -105,6 +105,20 @@
...
+    documentationCategory = OptionDocumentationCategory.REMOTE,
+    effectTags = {OptionEffectTag.UNKNOWN},
+    help =
+        "Specify a file to configure the remote downloader with. This file consists of lines, "
+            + "each of which starts with a directive (`allow`, `block` or `rewrite`) followed "
+            + "by either a host name (for `allow` and `block`) or two patterns, one to match "
+            + "against, and one to use as a substitute URL, with back-references starting from "
+            + "`$1`. It is possible for multiple `rewrite` directives for the same URL to be "
+            + "give, and in this case multiple URLs will be returned.")
+  public String remoteDownloaderConfig;
```
coeuvre, line 120: `RemoteOptions` is used for remote cache/execution. Since your changes are focusing on repository downloading, maybe move the flag to `RepositoryOptions`?

shs96c: I grouped this with the other `remote_*` options, including `remote_downloader_header` and `remote_cache_header`. Should they all be moved?

coeuvre: `remote_download_header` is used for `GrpcRemoteDownloader` and `remote_cache_header` is used for `GrpcCacheClient`. I think we should keep them there since they are both part of Remote Execution API.

shs96c: I've moved the field.

shs96c:
@coeuvre I've moved things around, as requested. I think this is ready for review again. Thank you!

laurentlb:
@coeuvre Could you take another look at this, please?

shs96c:
A gentle ping, @coeuvre. A review or a merge would be very much appreciated :)

meteorcloudy:
Interesting, have you considered using proxy to achieve the same thing? 
https://docs.bazel.build/versions/0.29.0/external.html#using-proxies

We are designing a new way to handle external dependencies in Bazel, we would also like to provide similar functionalities in the new tool for users to redirect the URLs or have access control. I wonder if a proxy is enough.

meteorcloudy:
BTW, @coeuvre is on vacation, he will be back next week (maybe this Saturday because it's a working day in China.)

shs96c:
I did consider using a proxy, and that's why I wrote 665c6659f344b3a3e3e209294fd57f03bfc354fe

I've put this behind an `experimental` flag, so hopefully while we wait for this new way of handling external dependencies to land this could be usable.

shs96c:
While the proxy would have worked nicely in environments we control, a config option can be set in a `.bazelrc` option, and will work no matter who has checked out the code, and whether they have access to the proxy server, and is therefore more robust. It also removes a potential single-point-of-failure from the build.

src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
```diff
@@ -281,6 +283,14 @@ public void beforeCommand(CommandEnvironment env) {
         }
       }
 
+      RepositoryOptions repositoryOptions = env.getOptions().getOptions(RepositoryOptions.class);
```
coeuvre, line 286: Should use the `repoOptions` declared above.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/UrlRewriterConfig.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.IOException;
+import java.io.Reader;
+import java.io.UncheckedIOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+class UrlRewriterConfig {
```
coeuvre, line 30: Please add javadoc for this class and it's public methods.

src/main/java/com/google/devtools/build/lib/runtime/commands/BuildCommand.java
```diff
@@ -52,7 +53,8 @@
       LoadingOptions.class,
       KeepGoingOption.class,
       LoadingPhaseThreadsOption.class,
-      BuildEventProtocolOptions.class
+      BuildEventProtocolOptions.class,
+      RemoteOptions.class,
```
coeuvre, line 57: Since we have moved the `RemoteDownloaderConfig` from `RemoteOptions` to `RepositoryOptions`, can you check whether adding `RemoteOptions.class` here and `//src/main/java/com/google/devtools/build/lib/remote/options` in BUILD files are necessary?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/UrlRewriter.java
```diff
@@ -0,0 +1,168 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
```
coeuvre, line 1: nit: Copyright 2020

src/main/java/com/google/devtools/build/lib/bazel/repository/RepositoryOptions.java
```diff
@@ -133,6 +133,20 @@
...
+    documentationCategory = OptionDocumentationCategory.REMOTE,
+    effectTags = {OptionEffectTag.UNKNOWN},
+    help =
+      "Specify a file to configure the remote downloader with. This file consists of lines, "
+        + "each of which starts with a directive (`allow`, `block` or `rewrite`) followed "
+        + "by either a host name (for `allow` and `block`) or two patterns, one to match "
+        + "against, and one to use as a substitute URL, with back-references starting from "
+        + "`$1`. It is possible for multiple `rewrite` directives for the same URL to be "
+        + "give, and in this case multiple URLs will be returned.")
+  public String remoteDownloaderConfig;
```
coeuvre, line 148: Can we remove the `remote` prefix? This is confused with `RemoteDownloader` in remote module.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/UrlRewriterConfig.java
```diff
@@ -0,0 +1,106 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
```
coeuvre, line 1: nit: Copyright 2020

coeuvre (commented):
Sorry for the delay, I was OOO.

coeuvre:
Thanks! Merging.

Wyverald:
I have two concerns.

The first is that this config only applies to `repository_ctx.download`. There is nothing preventing repo rules from simply using `repository_ctx.execute` to call out to a random executable which performs HTTP downloads. In fact, this is what rules_jvm_external does, so IIUC the lines about blocking/rewriting maven URLs in your example are essentially no-op.

Technically we could set up a sandbox / proxy server (similar to mod_proxy for Apache) that enforces this rewrite config, which would solve the first problem. That brings us to the second concern -- this URL rewriter script has a very ad-hoc, home-grown format. What proxy server solutions are out there, and what config language do they use? Can we pick an existing solution, or better, decouple the choice from Bazel itself?

I don't want to block this PR if it solves a pain point and especially since it's behind an experimental flag. But if we do one day want to de-experimental-ize this, I'd like to see those concerns addressed.

jin:
> In fact, this is what rules_jvm_external does, so IIUC the lines about blocking/rewriting maven URLs in your example are essentially no-op.

Not exactly. In `pinned_maven_install` mode, these are [downloaded through `http_file`](https://github.com/bazelbuild/rules_jvm_external/blob/6f72e13cf95c729ac0b8587e2fb9724fd873d271/coursier.bzl#L337-L359), which use `repository_ctx.download` underneath. Your points still stand though.

shs96c:
I'd also like to see a globally applicable solution that works in all cases. The nice thing with this approach is that we don't require anyone to install anything else on their machine, and the same config is shared between (eg) CI and developer machines. It gets us at least part-way towards a better solution.

As you point out, this does rely on rules being well-behaved and using bazel's own mechanisms for downloads. As @jin points out, once dependencies are pinned in `rules_jvm_external`, they're downloaded via `http_file`. IME, this seems to be true of the rulesets that we currently use: pinning doesn't use bazel, but once dependencies are pinned, they do.

Someone shelling out to an arbitrary program for downloading is, indeed, a problem. A previous PR of mine allowed the downloader to honour the regular Java HTTP proxy system properties, so it should be easy enough to expose those to other rules, and to use a full-fledged third party proxy.

coeuvre:
Merged with some changes to pass internal checks.

PR #12172, opened 2020-09-25
Added UserAgent to downloader HttpConnector

rdelfin:
The issue is outlined in #11187, but it seems like some servers expect a `User-Agent` to allow fetching archives, notably gitlab. This prevents bazel from fetching archives sourced there via `http_archive()`. This pull should fix the issue


rdelfin:
@googlebot i signed it!

meisterT (approved):

PR #12175, opened 2020-09-25
Add feature to remap Xcode path in debug info

keith:
By default the path to Xcode and the current SDK root are embedded in
binary's debug info. This information is mostly unused, and otherwise
has other workarounds for including for debugging. This feature scrubs
that path to produce more hermetic binaries.

https://forums.swift.org/t/lldb-usage-of-dw-at-llvm-isysroot/40575

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5357,6 +5357,29 @@ def _impl(ctx):
         ],
     )
 
+    remap_xcode_path_feature = feature(
+        name = "remap_xcode_path",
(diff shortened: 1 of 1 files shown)


tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5357,6 +5357,29 @@ def _impl(ctx):
         ],
     )
 
+    remap_xcode_path_feature = feature(
+        name = "remap_xcode_path",
```
keith, line 5361: happy to use different names here too

keith:
Thanks!

PR #12179, opened 2020-09-28
add --http_mirror to accelerate http downloads on external repositories

yancl:
Every time we build our projects by Bazel from a clean environment(ci for example), it will cost a lot of time to download the external artifacts(sometimes it even breaks because of network jitters). So I think it will be better for Bazel to support something like the [GOPROXY](https://proxy.golang.org/) so that we can maintain our own mirrored external artifacts and speed up the building process.

According to @philwo's great advice, it is better to use repository options to implement it.

Thank you for your help, @philwo, PTAL:)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java
@@ -29,6 +29,7 @@ import com.google.devtools.build.lib.vfs.Path;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.io.OutputStream;
+import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URL;
@@ -50,11 +51,16 @@ public class HttpDownloader implements Downloader {
   private static final Semaphore semaphore = new Semaphore(MAX_PARALLEL_DOWNLOADS, true);
 
   private float timeoutScaling = 1.0f;
+  private URL mirror = null;
 
   public HttpDownloader() {}
 
   public void setTimeoutScaling(float timeoutScaling) {
     this.timeoutScaling = timeoutScaling;
+  }
+
+  public void setMirror(URL mirror) {
+    this.mirror = mirror;
   }
 
   @Override
@@ -85,6 +91,14 @@ public class HttpDownloader implements Downloader {
     boolean success = false;
 
     List<IOException> ioExceptions = ImmutableList.of();
+
+    // Prefix all urls with the mirror to send requests to the mirror server first
+    // if mirrored urls failed, then retry the origin urls
+    if (mirror != null) {
+      List<URL> mirroredUrls = getMirroredUrls(urls);
+      mirroredUrls.addAll(urls);
+      urls = mirroredUrls;
+    }
 
     for (URL url : urls) {
       semaphore.acquire();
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java
```diff
@@ -321,6 +323,123 @@ public void downloadFrom2UrlsBothSocketTimeoutDuringBodyRead()
...
+                        "HTTP/1.1 200 OK",
+                        "Date: Fri, 31 Dec 1999 23:59:59 GMT",
+                        "Connection: close",
+                        "Content-Type: text/plain",
+                        "",
+                        "hello");
+
+                    // Never close the socket to cause SocketTimeoutException during body read on client
+                    // side.
+                    return null;
```
philwo, line 395: Can we change this test instead to immediately return an HTTP 404, as I think this is the more common case for "the mirror does not have my file yet". :)

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java
```diff
@@ -136,4 +150,21 @@ public void download(
...
   }
+
+  public List<URL> getMirroredUrls(List<URL> urls) throws IOException {
+    List<URL> mirroredUrls = new ArrayList<>(urls.size());
+    for (URL url : urls) {
+      String portStr = "";
+      if ( url.getPort() != -1 ) {
+        portStr = String.format(":%d", url.getPort());
+      }
+      String mirroredUrlStr = mirror.toString().replaceAll("/*$","") + "/" + url.getHost() + portStr + "/" + url.getPath().replaceAll("^/*","");
```
philwo, line 161: Could you extract this URL transformation into a method and add a simple unit test for it that ensures it does what we think it does? ;)

philwo (changes requested):
Very nice, thank you! I'm already looking forward to removing all the duplicate URLs in our WORKSPACE file... :)

src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java
yancl: Yes, you are right.  HTTP 404 is much better in this case:)

In our case we have a customized mirror server that will fetch artifacts from the original server automatically when it can't find the requested files locally. Then we don't need to upload the added/updated artifacts to our mirror server manually when WORKSPACE reviewed.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java
yancl: Thanks for your review, updated, PTAL:)

philwo (approved):
Thank you! The code looks good to me :)

I feel bad that I only now noticed this, but my colleagues told me that we'll need a little design document for this new feature - just a paragraph what this is about and the transformation algorithm from "URL => mirrored URL" with some examples would be fine. Then we could post this on the https://github.com/bazelbuild/proposals repo, see if we get any further feedback on bazel-dev@ and then merge this PR.

The reason is that this is a new flag that affects how Bazel communicates with remote systems and we want to ensure that we get this absolutely right and don't forget about any edge case (otherwise if we change this transformation again in the future, it would break all people who are already relying on this version of the method and have structured the URLs on their mirror in that way).

Are you interested in writing up a tiny little proposal doc for this? If you're busy, I totally understand that, just let me know and I'd be happy to give it a try myself in the next days. Sorry for not having stated this in the beginning.

yancl:
Thank you @philwo, very very appreciate for what you have done for me. I'm very happy to give a design doc on this new flag in the next few days:)

philwo:
@yancl Do I understand correctly that we're instead going ahead with https://github.com/bazelbuild/bazel/pull/12170 and can close this? :)

yancl:
Yes, thanks  :)

PR #12168, opened 2020-09-24
Add support for toolchain java 15

Jonpez2:
Closes https://github.com/bazelbuild/bazel/issues/11871

I modelled this on @davido's addition of a java 14 toolchain, here:[https://github.com/bazelbuild/bazel/pull/11514](url).

The test plan has the same form: 

Create java_tools from this commit:
`  $ bazel build src:java_tools_java11.zip`

Switch to using the java_tools from the above step in WORKSPACE file:

```
  load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
  http_archive(
    name = "remote_java_tools_linux",
    urls = [
        "file:///<path to the java_tools_java11.zip>",
    ],
  )

```
Add Zulu OpenJDK15 to use as javabase in WORKSPACE file:
```
  http_archive(
    name = "openjdk15_linux_archive",
    build_file_content = """
java_runtime(name = 'runtime', srcs =  glob(['**']), visibility = ['//visibility:public'])
exports_files(["WORKSPACE"], visibility = ["//visibility:public"])
""",
    strip_prefix = "zulu15.27.17-ca-jdk15.0.0-linux_x64",
    urls = ["https://cdn.azul.com/zulu/bin/zulu15.27.17-ca-jdk15.0.0-linux_x64.tar.gz"],
  )
```

Activate custom java toolchain and javabase in .bazelrc file:
```
  build --java_toolchain=@remote_java_tools_linux//:toolchain_jdk_15
  build --host_java_toolchain=@remote_java_tools_linux//:toolchain_jdk_15
  build --javabase=@openjdk15_linux_archive//:runtime
  build --host_javabase=@openjdk15_linux_archive//:runtime

```
Create Java 15 example class file:
```
public class Javac15Example {
  static String textBlock = """
              Hello,
              World
              """;
  public static void main(String[] args) {
    System.out.println(textBlock);
  }
}

```
Add Bazel file to build Java 15 syntax class:
```
  java_binary(
    name = "Javac15Example",
    srcs = ["Javac15Example.java"],
    main_class = "Javac15Example",
  )

```
Test that it works as expected:
```
  $ bazel run java:Javac15Example
INFO: Analyzed target //java:Javac15Example (1 packages loaded, 2 targets configured).
INFO: Found 1 target...
INFO: From Building java/Javac15Example.jar (1 source file):
Target //java:Javac15Example up-to-date:
  bazel-bin/java/Javac15Example.jar
  bazel-bin/java/Javac15Example
[...]

PR #12180, opened 2020-09-28
Remove java_tools_javac14 (releases and dependencies) from Bazel.

comius:
Java_tools_javac14 is a binary equivalent to java_tools_javac11. It was split from java_tools_javac 11 in order to be able to provide specific support for Java 14.

As a result a release of java_tools_javac* we now need 6 binaries and fix of dependencies and their hashes in Bazel on cca. 18 different places. Adding another Java version (e.g. Java 15) will add additional 3 binaries and 9 places to fix (total 9/27).

Also JDK 14 has reached EOL.

It makes sense to have only one release until we actually have some specific support for Java 14 or newer.

This is a partial revert of commits fe291f7dcadc20eb85cff07b2cd4d899d79fba71 and 693c88a8f897b6764b98c7d9da922ffc96c6d890:
- Add support for toolchain java 14
- Add unit test against released java_tools_javac14


comius:
cc @davido, @Jonpez2

comius:
related to https://github.com/bazelbuild/bazel/pull/12168

davido:
SGTM.

Though, I would rather change the justificatin in the commit message to just: "JDK 14 is reached EOL", instead of bringing obscure math counting places needed to change in Bazel code base to add supprot for new `java_toolchain_jdk_xyz`.

philwo (approved):
Thanks!

PR #12182, opened 2020-09-28
Fix two toolchain types resolved to the same label

comius:
Fixing toolchain resolution to allow two different toolchain types resolve to the same label.


katre (approved):
Can you add a test (in [src/test/shell/bazel/toolchain_test.sh](https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/toolchain_test.sh)) for this case?

comius:
cc @lberki, @c-mita, @oquenchil

katre (approved):

PR #12189, opened 2020-09-29
Keep client from using cookie files from the old server

loshca:


lberki:
Looks reasonable, but I'd like to understand what's happening first. Let's discuss on #12190 .

meisterT:
Can we merge this?

loshca:
> Can we merge this?

Sorry, not yet. It's not enough to fix the race from the issue. I will add the second part next week.
Converted to draft for now.

lberki (approved):

lberki (changes requested):
...and it's now that I caught up with the thread. Let me know when it's ready for review again!

sgowroji:
Hello @loshca, Do we have any update on this PR for further review ? Thanks!

PR #12192, opened 2020-09-29
Add default target constraint

comius:
Changes for proposal https://github.com/bazelbuild/proposals/blob/master/designs/2020-09-29-new-default-target-constraint.md


comius:
The change would make target and host platform different when running on the same machine. Looking for other solutions.

PR #12194, opened 2020-09-29
[doc] correct bazel's ancestor

wisechengyi:


wisechengyi:
I am assuming this is what the text meant, but feel free to close this if incorrect.

jin (approved):

jin:
(this is unfortunately not the right fix, we need to fix our internal copybara string transformation)

wisechengyi:
ack. thanks for checking!

Issue #7713, opened 2019-03-13
Minimize output of --toolchain_resolution_debug


EricBurnett:
### Description of the problem / feature request:

--toolchain_resolution_debug contains useful information for understanding why a toolchain was selected, but it's also very verbose - in an example case involving the Go rules, 1600 lines of output mostly of the form "considering X; constraint doesn't match; rejected X", plus a large number of repeated lines when unspecified actions choose a toolchain (_"Selected execution platform //:builder_platform, type @io_bazel_rules_go//go:toolchain -> toolchain @go_sdk//:go_linux_amd64-impl"_ 800 times)

### Feature requests: what underlying problem are you trying to solve with this feature?

It'd be nice if this could be more compacted, and easier to scan. E.g. by collecting up the resolution details first, then producing a compact structured output second - 

    Toolchain resolution debug info:
        @io_bazel_rules_go//go:toolchain for @bazel_tools//platforms:target_platform
            Matched execution platforms and toolchains: {//:builder_platform -> @go_sdk//:go_linux_amd64-impl, @bazel_tools//platforms:host_platform -> @go_sdk//:go_linux_amd64-impl}
            Rejected toolchains @go_sdk//:go_android_386-impl, @go_sdk//:go_android_amd64-impl, and 20 others for not matching constraint @bazel_tools//platforms:os , value @bazel_tools//platforms:linux
        @io_bazel_rules_go//go:toolchain for @bazel_tools//platforms:host_platform
            Matched execution platforms and toolchains: {//:builder_platform -> @go_sdk//:go_linux_amd64-impl, @bazel_tools//platforms:host_platform -> @go_sdk//:go_linux_amd64-impl}
            Rejected toolchains @go_sdk//:go_android_386-impl, @go_sdk//:go_android_amd64-impl, and 20 others for not matching constraint @bazel_tools//platforms:os , value @bazel_tools//platforms:linux
        @bazel_tools//tools/cpp:toolchain_type for @bazel_tools//platforms:target_platform
        ...

For matching of toolchains against actions themselves, it seems it'd be better to either omit those lines, or to emit the target as well - 800 lines saying the go toolchain was selected but without context don't quite solve that well.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Add --toolchain_resolution_debug to any build including Go.

cushon:
see also b/126375095

katre:
[...]

PR #12196, opened 2020-09-30
Improve toolchain resolution debug

comius:
When multiple toolchain types are resolved, skyframe functions run in parallel and the debug output is often hard to follow. Adding type of toolchain to each message makes it possible to analyse the results better.

Also after a execution transition, target changes. Added target label to the message, to be able to tell for which target toolchain was selected.

Removing initial messages "Looking for toolchain of type...", "Considering toolchain...". With additional fields displayed they become obsolete.

Condensing 'rejection explanation', i.e. which values are mismatched, into a single line.

#7713, b/126375095: there is still some duplications in ToolchainResolutionFunction:125, which I didn't remove because some tests depend on it (when specifying extra_toolchains)

b/117279695 is fixed with this

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
@@ -132,29 +132,26 @@ public class SingleToolchainResolutionFunction implements SkyFunction {
     ImmutableMap.Builder<ConfiguredTargetKey, Label> builder = ImmutableMap.builder();
     ToolchainTypeInfo toolchainType = null;
 
-    debugMessage(eventHandler, "%s: Looking for toolchain for target %s", toolchainTypeLabel,
-        targetPlatform.label());
     for (DeclaredToolchainInfo toolchain : toolchains) {
       // Make sure the type matches.
       if (!toolchain.toolchainType().typeLabel().equals(toolchainTypeLabel)) {
         continue;
       }
-      debugMessage(eventHandler, "%s: %s: Considering toolchain.", toolchainTypeLabel,
-          toolchain.toolchainLabel());
 
       // Make sure the target platform matches.
       if (!checkConstraints(
           eventHandler, toolchain.targetConstraints(), "target", targetPlatform, toolchainTypeLabel,
           toolchain.toolchainLabel())) {
-        debugMessage(
-            eventHandler,
-            "%s: %s: Rejected toolchain, because of target platform mismatch",
-            toolchainTypeLabel, toolchain.toolchainLabel());
         continue;
       }
 
       // Find the matching execution platforms.
       for (ConfiguredTargetKey executionPlatformKey : availableExecutionPlatformKeys) {
+        // Only check the toolchains if this is a new platform.
+        if (platformKeysSeen.contains(executionPlatformKey)) {
+          continue;
+        }
+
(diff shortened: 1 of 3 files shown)


comius:
cc @lberki @c-mita @oquenchil

src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
```diff
@@ -210,41 +205,50 @@ private static boolean checkConstraints(
       @Nullable EventHandler eventHandler,
       ConstraintCollection toolchainConstraints,
       String platformType,
-      PlatformInfo platform) {
+      PlatformInfo platform,
+      Label toolchainTypeLabel, Label toolchainLabel) {
```
katre, line 209: Please put toolchainLabel on a new line for consistency,

src/test/shell/bazel/toolchain_test.sh
```diff
@@ -533,9 +533,8 @@ EOF
     --toolchain_resolution_debug \
     --incompatible_auto_configure_host_platform \
     //demo:use &> $TEST_log || fail "Build failed"
-  expect_log 'ToolchainResolution: Looking for toolchain of type //toolchain:test_toolchain'
-  expect_log 'ToolchainResolution:   For toolchain type //toolchain:test_toolchain, possible execution platforms and toolchains: {@local_config_platform//:host -> //:test_toolchain_impl_1}'
-  expect_log 'ToolchainResolution: Selected execution platform @local_config_platform//:host, type //toolchain:test_toolchain -> toolchain //:test_toolchain_impl_1'
+  expect_log 'ToolchainResolution:   Type //toolchain:test_toolchain: target @local_config_platform//.*: execution @local_config_platform//:host: Selected toolchain //:test_toolchain_impl_1'
```
katre, line 536: Instead of "target @local_config_platform", which looks like it is the target being built, specify that this is "target platform @local_config_platform".

katre (changes requested):
Does this help mitigate #7713? Also see http://b/117279695 and http://b/126375095.

src/test/shell/bazel/toolchain_test.sh
comius: > Does this help mitigate #7713? Also see http://b/117279695 and http://b/126375095.

#7713, b/126375095: there is still some duplications in ToolchainResolutionFunction:125, which I didn't remove because some tests depend on it (when specifying extra_toolchains)

b/117279695 is fixed with this

What would be most useful for me is to have two levels: displaying selected toolchains, displaying also why they were rejected.

src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
comius: Done.

comius:
> Does this help mitigate #7713? Also see http://b/117279695 and http://b/126375095.

#7713, b/126375095: there is still some duplications in ToolchainResolutionFunction:125, which I didn't remove because some tests depend on it (when specifying extra_toolchains)

b/117279695 is fixed with this

What would be most useful for me is to have two levels: displaying selected toolchains, displaying also why they were rejected.

src/test/shell/bazel/toolchain_test.sh
comius: Done.

katre (approved):

PR #12205, opened 2020-10-01
Add Java constraint settings.

comius:
The constraint setting are copied from rules_java, with added new versions of JDKs and added constraint for JDK location (local, absolute, remote) - with which we can 'tag' the toolchains more precisely - i.e. allowing the user to select arbitrary java toolchain that is defined by default.

Duplication of constraints with rules_java is generally not a good idea. However the rules_java has been last updated in bazel at commit https://github.com/bazelbuild/rules_java/commit/7cf3cefd652008d0a64a419c34c13bdca6c8f178.

Comparison https://github.com/bazelbuild/rules_java/compare/7cf3cefd652008d0a64a419c34c13bdca6c8f178...HEAD
shows that since then dependency to bazel_federation was added, and also references to remote_java releases (which are stale). At this point of time release of rules_java would potentially cause more problems than duplication of those constraints (which are not yet used in Bazel).

### DIFF
--- a/tools/jdk/BUILD
+++ b/tools/jdk/BUILD
@@ -443,6 +443,7 @@ filegroup(
         "proguard_whitelister_test_input.pgcfg",
         "remote_java_tools_aliases.bzl",
         "toolchain_utils.bzl",
+        "//tools/jdk/constraints:srcs",
     ],
 )
 
--- a/tools/jdk/constraints/BUILD
+++ b/tools/jdk/constraints/BUILD
@@ -1,0 +1,116 @@
+# Standard constraint_settings and constraint_values for Java.
+
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])
+
+filegroup(
+    name = "srcs",
+    srcs = ["BUILD"],
+)
+
+constraint_setting(name = "runtime")
+
+constraint_value(
+    name = "jdk8",
+    constraint_setting = ":runtime",
+)
+
+constraint_value(
+    name = "jdk9",
+    constraint_setting = ":runtime",
+)
+
+constraint_value(
+    name = "jdk10",
+    constraint_setting = ":runtime",
+)
+
+constraint_value(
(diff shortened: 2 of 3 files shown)


katre:
Internally, we're moving away from explicit constraints for JDK and Java language level, and towards using a Starlark flag to control those. One of the reasons was because it became too confusing to create different platform definitions for each value: users' dev machines, for example, might have multiple JDKs (each capable of multiple language levels), and defining multiple platforms to refer to the same physical machine was cumbersome. Also, users became confused when asked to change the `--platforms` flag when building for the same machine, to change the language level available.

The local/remote/absolute constraints do look more useful, however.

comius:
> Internally, we're moving away from explicit constraints for JDK and Java language level, and towards using a Starlark flag to control those.

This seems like a good idea for the language level. I think I still need JDK constraints and I see them used internally. Is there a design document or some other reference for them being removed?

lberki:
@katre : how is the appropriate JVM/JDK selected then? Probably not through the toolchain selection mechanism then since Starlark flags cannot affect that.

katre:
I've shared the design document with you both. The summary is:

```
# Also create build_setting rules.
config_setting(
    name = "runtime_version_8",
    values = {"java_runtime_version": "8"},
)

config_setting(
    name = "runtime_version_11",
    values = {"java_runtime_version": "11"},
)

toolchain(
    name = "java-runtime-toolchain-k8",
    target_compatible_with = [
        "@platforms//cpu:k8",
        "@platforms//os:linux",
    ],
    toolchain = select({
      ":runtime_version_8": ":jdk8-k8",
      ":runtime_version_11": ":jdk11-k8",
    }),
    toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
)
```

So toolchain resolution handles finding the right `toolchain` entry for the CPU and OS, and then the select handles finding the right JDK for the given language level, based on the Starlark flag.

PR #12211, opened 2020-10-02
Add undeclared ninja file path in exception message

torgil:
When a ninja file includes another ninja file it's needs to be declared
to the ninja_graph rule, otherwise an exeception is thrown.

This patch adds information of which file were not declared to the
exception error message which could be handy for trouble-shooting.


jin (approved):

jin:
Thank you!

PR #12220, opened 2020-10-05
Change host transitions to exec transition in android sdk.

comius:


katre (approved):

comius:
Downstream tests at https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1686

PR #12224, opened 2020-10-06
Move default_java_toolchain definitions into java_tools repository

comius:
default_java_toolchain macro is handy to define toolchains in //tools/jvm/BUILD without making a release of java_tools.

However default_java_toolchain heavily relies on selects by operating system to pick the right parts from the right java_tools. This indirection is removed by putting definition directly into java_tools.

This is the first part. After java_tools release, default_java_toolchain and remote_java_tools_java_import can be removed.

### DIFF
--- a/tools/jdk/BUILD.java_tools
+++ b/tools/jdk/BUILD.java_tools
@@ -19,7 +19,6 @@ SUPRESSED_WARNINGS = select({
 
 java_toolchain_default(
     name = "toolchain",
-    jacocorunner = ":jacoco_coverage_runner_filegroup",
     jvm_opts = [
         # In JDK9 we have seen a ~30% slow down in JavaBuilder performance when using
         # G1 collector and having compact strings enabled.
@@ -97,7 +96,6 @@ RELEASES = (8, 9, 10, 11)
 # A toolchain that targets java 11.
 java_toolchain_default(
     name = "toolchain_jdk_11",
-    jacocorunner = ":jacoco_coverage_runner_filegroup",
     jvm_opts = [
         # In JDK9 we have seen a ~30% slow down in JavaBuilder performance when using
         # G1 collector and having compact strings enabled.
@@ -111,7 +109,6 @@ java_toolchain_default(
 # A toolchain that targets java 14.
 java_toolchain_nojavac(
     name = "toolchain_jdk_14",
-    jacocorunner = ":jacoco_coverage_runner_filegroup",
     source_version = "14",
     target_version = "14",
 )
@@ -119,7 +116,6 @@ java_toolchain_nojavac(
 # A toolchain that targets java 15.
 java_toolchain_nojavac(
     name = "toolchain_jdk_15",
-    jacocorunner = ":jacoco_coverage_runner_filegroup",
     source_version = "15",
     target_version = "15",
 )
@@ -132,7 +128,6 @@ java_toolchain_nojavac(
 java_toolchain_default(
     name = "prebuilt_toolchain",
     ijar = [":ijar_prebuilt_binary"],
-    jacocorunner = ":jacoco_coverage_runner_filegroup",
     jvm_opts = [
         # In JDK9 we have seen a ~30% slow down in JavaBuilder performance when using
         # G1 collector and having compact strings enabled.
--- a/tools/jdk/BUILD.java_tools
+++ b/tools/jdk/BUILD.java_tools
@@ -97,6 +97,14 @@ java_toolchain_nojavac(
     target_version = "14",
 )
 
+# A toolchain that targets java 15.
+java_toolchain_nojavac(
+    name = "toolchain_jdk_15",
+    jacocorunner = ":jacoco_coverage_runner_filegroup",
+    source_version = "15",
+    target_version = "15",
+)
+
 # The new toolchain is using all the pre-built tools, including
 # singlejar and ijar, even on remote execution. This toolchain
 # should be used only when host and execution platform are the
--- a/tools/jdk/BUILD.java_tools
+++ b/tools/jdk/BUILD.java_tools
@@ -35,8 +35,18 @@ java_toolchain_default(
     target_version = "8",
 )
 
+# Default to the Java 8 language level.
+# TODO(cushon): consider if/when we should increment this?
 java_toolchain_default(
     name = "legacy_toolchain",
+    source_version = "8",
+    target_version = "8",
+)
+
+# Needed for openbsd
+java_toolchain_default(
+    name = "legacy_toolchain_jvm8",
+    jvm_opts = JDK8_JVM_OPTS,
     source_version = "8",
     target_version = "8",
 )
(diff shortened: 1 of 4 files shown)


comius:
cc @lberki, @c-mita, @katre

tools/jdk/BUILD.java_tools
```diff
@@ -15,203 +17,124 @@ SUPRESSED_WARNINGS = select({
...
     target_version = "8",
-    tools = [
-        ":java_compiler_jar",
-        ":jdk_compiler_jar",
-    ],
 )
 
+# The 'vanilla' toolchain is an unsupported alternative to the default.
+#
+# It does not provider any of the following features:
```
katre, line 55: Nit: "It should not provide any ..."

tools/jdk/BUILD.java_tools
```diff
@@ -15,203 +17,124 @@ SUPRESSED_WARNINGS = select({
...
+#   * Error Prone
+#   * Strict Java Deps
+#   * Header Compilation
+#   * Reduced Classpath Optimization
+#
+# It uses the version of javac from the `--host_javabase` instead of the
+# embedded javac, which may not be source- or bug-compatible with the embedded
```
katre, line 62: Do we actually mean "embedded" here, or "local" instead?

tools/jdk/java_toolchain_default.bzl
```diff
@@ -0,0 +1,123 @@
...
+]
+
+DEFAULT_JAVACOPTS = [
+    "-XDskipDuplicateBridges=true",
+    "-g",
+    "-parameters",
+]
+
+DEFAULT_TOOLCHAIN_CONFIGURATION = {
+    "forcibly_disable_header_compilation": 0,
```
katre, line 50: Instead of 0 (and 1, for java_cupports_workers below), can this be False?

tools/jdk/java_toolchain_default.bzl
```diff
@@ -0,0 +1,123 @@
...
+    "javac": [":javac_jar"],
+    "tools": [
+        ":java_compiler_jar",
+        ":jdk_compiler_jar",
+    ],
+    "javac_supports_workers": 1,
+    "jvm_opts": JDK9_JVM_OPTS,
+    "misc": DEFAULT_JAVACOPTS,
+    "singlejar": [":singlejar"],
+    "bootclasspath": ["@bazel_tools//tools/jdk:platformclasspath"],
```
katre, line 66: Why is this specifically from @bazel_tools, when nothing else is? If there's a good reason, it deserves a comment.

tools/jdk/java_toolchain_default.bzl
```diff
@@ -0,0 +1,123 @@
...
+
+    toolchain_args = dict(DEFAULT_TOOLCHAIN_CONFIGURATION)
+    toolchain_args.update(kwargs)
+    native.java_toolchain(
+        name = name,
+        **toolchain_args
+    )
+
+JDK_NOJAVAC_JVM_OPTS = [
+    # Allow JavaBuilder to access internal javac APIs.
```
katre, line 82: Can these be shared with JDK9_JVM_OPTS from above?

tools/jdk/java_toolchain_default.bzl
```diff
@@ -0,0 +1,123 @@
...
+    "--add-opens=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED",
+
+    # quiet warnings from com.google.protobuf.UnsafeUtil,
+    # see: https://github.com/google/protobuf/issues/3781
+    # and: https://github.com/bazelbuild/bazel/issues/5599
+    "--add-opens=java.base/java.nio=ALL-UNNAMED",
+    "--add-opens=java.base/java.lang=ALL-UNNAMED",
+]
+
+NOJAVAC_TOOLCHAIN_CONFIGURATION = {
```
katre, line 99: It would be really nice to express this as DEFAULT_TOOLCHAIN_CONFIGURATION with some changes, rather than just another straight set of properties. I can't easily tell what is different between the configurations.

tools/jdk/BUILD.java_tools
comius: Fixed to "does not provide".

tools/jdk/BUILD.java_tools
comius: This is referring to javac which is usually provided with java_tools. I reworded it.

tools/jdk/java_toolchain_default.bzl
comius: Yes, it can. Thanks

tools/jdk/java_toolchain_default.bzl
comius: It involves two java_runtimes. No chance to put it here.

tools/jdk/java_toolchain_default.bzl
comius: Done.

tools/jdk/java_toolchain_default.bzl
comius: Done.

tools/jdk/java_toolchain_default.bzl
```diff
@@ -29,44 +36,49 @@ JDK9_JVM_OPTS = [
...
     "--add-opens=java.base/java.lang=ALL-UNNAMED",
 ]
 
-DEFAULT_JAVACOPTS = [
-    "-XDskipDuplicateBridges=true",
-    "-g",
-    "-parameters",
+# java_toolchain parameters without specifying javac, java.compiler,
+# and jdk.compiler module
+BASE_TOOLCHAIN_CONFIGURATION = dict(
```
katre, line 48: Nit: Any reason to use `dict(foo = "bar")` syntax over `{foo = "bar"}`?

katre (approved):
LGTM but please wait for @cushon to also review.

tools/jdk/java_toolchain_default.bzl
comius: The second one was actually `{"foo": "bar"}`. I find the first one nicer: it is closer to actually calling rules with parameters and doesn't use so many strings.

tools/jdk/java_toolchain_default.bzl
```diff
@@ -0,0 +1,103 @@
...
+
+    toolchain_args = dict(_DEFAULT_TOOLCHAIN_CONFIGURATION)
+    toolchain_args.update(kwargs)
+    native.java_toolchain(
+        name = name,
+        **toolchain_args
+    )
+
+def java_toolchain_nojavac(name, **kwargs):
+    """Defines a java_toolchain without overriding javac for Bazel."""
```
cushon, line 93: Would it be possible to change `java_toolchain_default` to have this behaviour, and then explicitly pass the additional flags to override javac at the use sites that still need them, and avoid the separate macro here?

It seems clearer to have one macro that provides defaults, and then override them where necessary. IIUC this is an opportunity to make breaking changes / do cleanup, since we're making a copy of this macro at a new location anyways.

cushon (approved):
LG modulo the question about `java_toolchain_nojavac`.

PR #12229, opened 2020-10-08
Add executable_use_pic to Starlark CcLinkingOutputs

torgil:
This makes it possible for Starlark code to know if an executable is
linked using pic objects or not.

Needed for instance to choose correct gcno-files in a clean way for custom
code coverare rules.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
@@ -771,6 +771,7 @@ public final class CcLinkingHelper {
     CppLinkAction dynamicLinkAction = dynamicLinkActionBuilder.build();
     if (dynamicLinkType.isExecutable()) {
       ccLinkingOutputs.setExecutable(linkerOutput);
+      ccLinkingOutputs.setExecutableUsePic(usePic);
     }
     actionConstructionContext.registerAction(dynamicLinkAction);
 
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingOutputs.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingOutputs.java
@@ -30,6 +30,7 @@ public class CcLinkingOutputs implements CcLinkingOutputsApi<Artifact> {
 
   @Nullable private final LibraryToLink libraryToLink;
   @Nullable private final Artifact executable;
+  @Nullable private final Boolean executableUsePic;
 
   private final ImmutableList<LtoBackendArtifacts> allLtoArtifacts;
   private final ImmutableList<Artifact> linkActionInputs;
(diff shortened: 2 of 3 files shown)


oquenchil:
Can you please show me a small sample project where you need to figure out whether the binary is pic or not? I believe it should already be possible without adding this to the API.

torgil:
> Can you please show me a small sample project where you need to figure out whether the binary is pic or not? I believe it should already be possible without adding this to the API.

Sure. This works for me using this patch. I want the gcno file matching the gcda file generated if the executable is run.

```
$ bazel-with-gcno-and-pic-patches build //:main_gcno
INFO: Analyzed target //:main_gcno (28 packages loaded, 118 targets configured).
INFO: Found 1 target...
Target //:main_gcno up-to-date:
  bazel-bin/_objs/main_gcno.o/main_gcno.gcno
INFO: Elapsed time: 2.468s, Critical Path: 0.23s
INFO: 3 processes: 2 internal, 1 processwrapper-sandbox.
INFO: Build completed successfully, 3 total actions
```

BUILD.bazel
```
load(":gcno.bzl", "generate_gcno")

generate_gcno(
    name = "main_gcno",
    contents = "int main(void) { return 0; }",
)
```

gcno.bzl
```
def _generate_gcno_impl(ctx):
    src = ctx.actions.declare_file(ctx.label.name + ".c")
    ctx.actions.write(output = src, content = ctx.attr.contents)
    cc_toolchain = ctx.toolchains["@bazel_tools//tools/cpp:toolchain_type"]
    feature_configuration = cc_common.configure_features(
        ctx = ctx,
        cc_toolchain = cc_toolchain,
        requested_features = ["coverage"],
    )
    cc_context, cc_outputs = cc_common.compile(
        name = ctx.label.name + ".o",
        actions = ctx.actions,
        cc_toolchain = cc_toolchain,
        srcs = [src],
    )
    linking_outputs = cc_common.link(
        name = ctx.label.name,
        actions = ctx.actions,
        cc_toolchain = cc_toolchain,
        compilation_outputs = cc_outputs,
    )
    if linking_outputs.executable_use_pic:
        gcno_files = cc_outputs.pic_gcno_files
    else:
        gcno_files = cc_outputs.gcno_files
    return DefaultInfo(files=depset(gcno_files))

generate_gcno = rule(
    fragments = ["cpp"],
    implementation = _generate_gcno_impl,
    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
    attrs = {
        "contents": attr.string(mandatory = True),
    },
)
```

oquenchil:
Right, well the way to figure out whether pic is used should be to look at [this condition](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java;l=558).

From that condition, the following:
- cppConfiguration.forcePic()
- cppConfiguration.getCompilationMode()

are not exposed to Starlark. So the right way would be to expose those 2 methods to Starlark, then re-creating the condition in your code to figure that out.

If you want you can send a PR for each of forcePic and getCompilationMode separately, each with their own test.

torgil:
Does this mean that you recommend me to maintain a replica of an internal java function and make tests to be sure that they're equal?  Or are you saying that these conditions will never change?

I'd prefer to get this kind of information directly from the source to avoid maintenance and possible problems with deviating definition.

oquenchil:
That's an internal Java function for now. But we are in the process of re-writing the rules to Starlark, so eventually that same function would be in a file with other utility function for C++ rules.

The existing native rules call that method, they don't query a field from CcLinkingOutputs. So my previous comment would be the right way to go.

I don't see why that condition would change in the coming years and if it ever changes I'm hoping that we are done with re-writing the rules to Starlark by then.

I'm going to close this PR now.

torgil:
Then I'll await the refactored rules and see if my issue with cc_common is solved in the process. In the meantime i'll maintain this patch.

Thank you for your support. Keep up the good work!

Issue #10843, opened 2020-02-22
Please implement output format 'graph' for cquery and aquery


konste:
### Description of the problem / feature request:

Bazel 'query' command supports output format 'graph' which is very useful for visualization of dependencies. Unfortunately cquery and aquery commands introduced later do not support 'graph' output format and it would be great to have it implemented.

### Feature requests: what underlying problem are you trying to solve with this feature?

Graph visualization with 'query' gets very complex for multi-platform and multi-configuration builds. It can be made ten-fold less busy and more readable if produced by cquery for particular platform and configuration, which would make it substantially more usable.

### What's the output of `bazel info release`?

2.1.1

gregestren:
Hi @konste,

For `cquery`, We recognize this as a valuable feature request. The `cquery` core devs currently have the resources to do probably one or two cquery features per quarter. We're seriously considering this one for next quarter (starting April). We'll update as we're able to prioritize.

FWIW, I don't *think* this is technically too hard. If anyone's willing to volunteer to make the change we're happy to consult and review.

I can't speak for `aquery`, which I'm not as familiar with.

gregestren:
We've (Bazel configurability) officially prioritized this for Q2.

gregestren:
@meisterT - any thoughts on aquery?

konste:
@gregestren I am not a Java developer and cannot contribute directly, but as a stop-gap solution for this problem I have a tiny Python script which converts cquery jsonproto output to .dot graph. You can find it at the end of this thread: https://groups.google.com/forum/#!msg/bazel-discuss/NMoRLSthQHY/quk9E-ACCAAJ  It works and confirms that graph generation from cquery is not technically challenging.

meisterT:
I have a similar script like @konste somewhere lying around. I can brush that up and make it work with the v2 proto output of aquery and then publish it. However it turned out to be less useful (at least internally) than I expected since even subsets of action graphs are so large that they don't render with graphviz anymore.

konste:
For the actual visualization of .dot I don't use Graphviz. Instead, I apply [dot2dgml](https://github.com/timepp/dot2dgml) and use Visual Studio to render resulting .dgml. It can render bigger graphs and adds much needed interactivity to help research it.

mihaimaruseac:
TensorFlow would greatly benefit from this. I'll see if I can use the scripts mentioned in recent commits, though wfh makes this somewhat harder :)

konste:
@gregestren 
[...]

PR #12248, opened 2020-10-09
Implement cquery --output=graph

gregestren:
Thankfully, `query` already has most of the infrastructure necessary to make this
easy.

query implements graph output (in `GraphOutputFormatter`) over a
`Digraph<Target>`, which is a generic graph data structure with `Target` payloads.
All output logic then runs over this data structure. To opt query in, all we have
to do is create an equivalent `Digraph<ConfiguredTarget>`, which is a simple
transformation from the backing graph.

This change creates a new generic class for that common logic:
`GraphOutputWriter`. query's `GraphOutputFormatter` then becomes a simple wrapper
over that, and the new `GraphOutputFormatterCallback` is cquery's equivalent.

A few differences:

 - cquery output is always fully ordered (`--order_output=full`). We could match
   this with query's controllable version, but I don't see a reason to make this
   yet another bit to configured.
 - query output annotates edges with select() conditions. cquery doesn't do this
   because select()s are resolved and removed from the graph after analysis. I
   think we could annotate edges with the *chosen* condition if there was
   demand, but that'd be a followup effort.

Fixes https://github.com/bazelbuild/bazel/issues/10843 (for `cquery`, not `aquery`)


gregestren:
@meisterT  re: potential aquery integration.

gregestren:
TODO:

- [ ] Verify streamed mode doesn't apply here (i.e. `partialResults` in the output formatters aren't actually partial results)
- [ ] Add tests
- [ ] Add docs

joeleba (commented):
Generally LGTM. Re: aquery integration: conceptually it makes sense, but in actuality aquery outputs are humongous and I'm not sure a visual representation of that is useful. Maybe with very specific scopes specified by the filters.

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
```diff
@@ -0,0 +1,310 @@
+// Copyright 2014 The Bazel Authors. All rights reserved.
```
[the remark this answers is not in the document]
gregestren: I never know what threshold requires a change here.

GitHub's differ doesn't show this swell, but `GraphOutputWriter` is mostly forked out of `GraphOutputFormatter` with a little extra refactoring sprinkled in. So it's mostly the same code moved to a new file. Does that call for a new year?

src/main/java/com/google/devtools/build/lib/query2/common/CommonQueryOptions.java
```diff
@@ -253,4 +270,30 @@ public AspectResolutionModeConverter() {
               + "precise mode is not completely precise: the decision whether to compute an aspect "
               + "is decided in the analysis phase, which is not run during 'bazel query'.")
   public AspectResolver.Mode aspectDeps;
+
+  ///////////////////////////////////////////////////////////
+  // GRAPH OUTPUT FORMATTER OPTIONS                        //
+  ///////////////////////////////////////////////////////////
+
+  @Option(
+      name = "graph:node_limit",
```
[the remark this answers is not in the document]
gregestren: This carries over `query`'s `--graph:node_limit` flag into a common location. However clear or not the flag is, I'd prefer to keep the existing API for the context of this change.

gregestren:
> Generally LGTM. Re: aquery integration: conceptually it makes sense, but in actuality aquery outputs are humongous and I'm not sure a visual representation of that is useful. Maybe with very specific scopes specified by the filters.

I figured as much. Thanks for clarifying. I'm personally happy just that you're aware this is a thing now.

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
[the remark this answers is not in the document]
joeleba: I generally update the year every time a new file is created :) That being said, I don't know the actual team convention, or if there's one at all, hence the `super` nit.

gregestren:
Added a long disclaimer to `PostAnalysisQueryBuildTool` in support of @haxorz 's suggestions to guarantee non-streaming mode.

src/main/java/com/google/devtools/build/lib/query2/common/CommonQueryOptions.java
```diff
@@ -253,4 +270,30 @@ public AspectResolutionModeConverter() {
               + "precise mode is not completely precise: the decision whether to compute an aspect "
               + "is decided in the analysis phase, which is not run during 'bazel query'.")
   public AspectResolver.Mode aspectDeps;
+
+  ///////////////////////////////////////////////////////////
+  // GRAPH OUTPUT FORMATTER OPTIONS                        //
```
juliexxia, line 275: Are the proto:<foo> options above also graph output formatter options?

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
```diff
@@ -0,0 +1,115 @@
...
+import com.google.devtools.build.lib.graph.Digraph;
+import com.google.devtools.build.lib.graph.Node;
+import com.google.devtools.build.lib.query2.engine.QueryEnvironment.TargetAccessor;
+import com.google.devtools.build.lib.query2.query.output.GraphOutputWriter;
+import com.google.devtools.build.lib.query2.query.output.GraphOutputWriter.NodeReader;
+import com.google.devtools.build.lib.skyframe.SkyframeExecutor;
+import java.io.OutputStream;
+import java.util.Comparator;
+
+/** cquery output formatter that prints the result as factored graph in AT&amp;T GraphViz format. */
```
juliexxia, line 30: nit: just AT&T?

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
```diff
@@ -0,0 +1,115 @@
...
+  public interface DepsRetriever {
+    Iterable<ConfiguredTarget> getDirectDeps(ConfiguredTarget taget) throws InterruptedException;
+  }
+
+  private final DepsRetriever depsRetriever;
+
+  private final GraphOutputWriter.NodeReader<ConfiguredTarget> nodeReader =
+      new NodeReader<ConfiguredTarget>() {
+
+        private final Comparator<ConfiguredTarget> configuredTargetOrdering =
```
juliexxia, line 48: For my own edification - reason to initialize this outside of the comparator method?

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
```diff
@@ -0,0 +1,115 @@
...
+  public void processOutput(Iterable<ConfiguredTarget> partialResult) throws InterruptedException {
+    // Transform the cquery-backed graph into a Digraph to make it suitable for GraphOutputWriter.
+    // Note that this involves an extra iteration over the entire query result subgraph. We could
+    // conceptually merge transformation and output writing into the same iteration if needed.
+    Digraph<ConfiguredTarget> graph = new Digraph<>();
+    ImmutableSet<ConfiguredTarget> allNodes = ImmutableSet.copyOf(partialResult);
+    for (ConfiguredTarget configuredTarget : partialResult) {
+      Node<ConfiguredTarget> node = graph.createNode(configuredTarget);
+      for (ConfiguredTarget dep : depsRetriever.getDirectDeps(configuredTarget)) {
+        if (allNodes.contains(dep)) {
```
juliexxia, line 96: Assuming this is for a situation like --noimplicit_deps or the like?

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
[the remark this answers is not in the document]
juliexxia: ^ I do the same! Update whenever a new file is created/branched

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
```diff
@@ -0,0 +1,311 @@
...
+  private void outputUnfactored(
+      Digraph<T> graph,
+      @Nullable ConditionalEdges conditionalEdges,
+      PrintWriter out) {
+    graph.visitNodesBeforeEdges(
+        new DotOutputVisitor<T>(out, nodeReader::getLabel) {
+          @Override
+          public void beginVisit() {
+            super.beginVisit();
+            // TODO(bazel-team): (2009) make this the default in Digraph.
```
juliexxia, line 129: laughing at the date of this TODO

juliexxia (commented):
Overall LGTM! Just minor nits/qs. Excited this is happening!

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
```diff
@@ -252,6 +252,13 @@ private static BuildConfiguration mergeEqualBuildConfiguration(
             OutputType.JSON),
         new BuildOutputFormatterCallback(
             eventHandler, cqueryOptions, out, skyframeExecutor, accessor),
+        new GraphOutputFormatterCallback(
```
haxorz, line 255: Don't forget to update the documentation. I think you'll want to remove the bullet point at https://github.com/bazelbuild/bazel/blob/74bd30243515055f41c1e6cf3fc2ce87ae644138/site/docs/cquery.html#L569 and then also add a section to `#output-formats` on the new support for `--output=graph`.

haxorz (approved):

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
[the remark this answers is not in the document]
gregestren: Consensus it is! Updated.

src/main/java/com/google/devtools/build/lib/buildtool/PostAnalysisQueryBuildTool.java
```diff
@@ -149,11 +152,32 @@ private void doPostAnalysisQuery(
                       NamedThreadSafeOutputFormatterCallback.callbackNames(callbacks))));
       return;
     }
+
+    // A certain subset of output formatters support "streaming" results - the formatter is called
+    // multiple times where each call has only a some of the full query results (see
+    // StreamedOutputFormatter for details). Post-analysis queries don't do this. But the reason is
```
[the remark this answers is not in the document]
gregestren: Added.

src/main/java/com/google/devtools/build/lib/query2/common/CommonQueryOptions.java
gregestren: Not to my knowledge? As far as I can see those only apply to `--output=proto`

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
gregestren: This inherits the same comment as in the `query` version, which I believe follows Javadoc's "Comments are written in HTML" guidance: https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#blockandinlinetags

Key snippet:

> entities for the less-than (<) and greater-than (>) symbols should be written &lt; and &gt;. Likewise, the ampersand (&) should be written \&amp;

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
gregestren: My original reasoning is that the `comparator` method can be called multiple times whereas the actual comparator logic only needs to be defined once. So it theoretically saves unnecessary extra instantiation.

In practice I don't think that'd make a huge difference (and I'd hope the JDK could optimize that). So I don't have strong feelings on the subject.

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
gregestren: The reasoning here is that the query output may only contain a subset of all deps.

So if A depends on B, C, and D and some magical query expression filtered the results down to just A and C, `depsRetriever.getDirectDeps` returns all of A's deps (B, C, and D). We only want to include the ones that are part of th query result (any target in `partialResult`).

src/main/java/com/google/devtools/build/lib/query2/query/output/GraphOutputWriter.java
gregestren: I definitely noticed that too. :p

src/main/java/com/google/devtools/build/lib/buildtool/PostAnalysisQueryBuildTool.java
```diff
@@ -149,11 +152,32 @@ private void doPostAnalysisQuery(
...
+    // subtle and hard to follow. Post-analysis output formatters inherit from Callback, which
+    // declares "void process(Iterable<T> partialResult)". Its javadoc says that the subinterface
+    // BatchCallback may stream partial results. But post-analysis callbacks don't inherit
+    // BatchCallback!
+    //
+    // To protect against accidental feature regression (like implementing a callback that
+    // accidentally inherits BatchCallback), we explicitly disable streaming here. The
+    // aggregating callback collects the entire query's results (even if they're streamed). Once the
```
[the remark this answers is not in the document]
gregestren: Got it - comments integrated.

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
gregestren: Done.

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
```diff
@@ -0,0 +1,115 @@
...
+
+        @Override
+        public String getLabel(Node<ConfiguredTarget> node) {
+          // Node payloads are ConfiguredTargets. Output node labels are target labels + config
+          // hashes.
+          ConfiguredTarget ct = node.getLabel();
+          return String.format(
+              "%s (%s)",
+              ct.getLabel(),
+              shortId(skyframeExecutor.getConfiguration(eventHandler, ct.getConfigurationKey())));
```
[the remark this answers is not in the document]
gregestren: This is a great suggestion, and I think it has teeth. This call is going to happen with egregious redundancy and I wouldn't be surprised if it had impact. So I'd like to explore this but do so as a followup focused change. I added a TODO to that effect that I'm happy to look at immediately after this one.

joeleba:
> > Generally LGTM. Re: aquery integration: conceptually it makes sense, but in actuality aquery outputs are humongous and I'm not sure a visual representation of that is useful. Maybe with very specific scopes specified by the filters.
> 
> I figured as much. Thanks for clarifying. I'm personally happy just that you're aware this is a thing now.

Related: https://www.youtube.com/watch?v=GDbaBOCDwrQ

gregestren:
> > > Generally LGTM. Re: aquery integration: conceptually it makes sense, but in actuality aquery outputs are humongous and I'm not sure a visual representation of that is useful. Maybe with very specific scopes specified by the filters.
> > 
> > 
> > I figured as much. Thanks for clarifying. I'm personally happy just that you're aware this is a thing now.
> 
> Related: https://www.youtube.com/watch?v=GDbaBOCDwrQ

That graph visualization is super-cool.

src/main/java/com/google/devtools/build/lib/query2/common/CommonQueryOptions.java
juliexxia: Ah, I didn't quite read this as --output=graph output formatter options. makes sense.

juliexxia (approved):

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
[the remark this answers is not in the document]
gregestren: https://github.com/bazelbuild/bazel/pull/12309

Issue #10756, opened 2020-02-11
BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN causes error in presence of --incompatible_enable_cc_toolchain_resolution


konste:
### Description of the problem / feature request:

We have --incompatible_enable_cc_toolchain_resolution set and have custom toolchains for our platforms. When we try to set BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 to make sure detected toolchain is never picked up we get the error `ERROR: C:/builds/y77h45xt/external/local_config_cc/BUILD:28:1: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'x64_windows'`

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

With --incompatible_enable_cc_toolchain_resolution **cc_toolchain_suite** should not be relevant at all. BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 causes generation of dummy cc_toolchain_suite ```cc_toolchain_suite(
    name = "toolchain",
    toolchains = {
        "local|local": ":local",
        "local": ":local",
    },
) ``` which does not contain toolchains for any cpu, which causes bogus error `cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu...` for any cpu. 

### What operating system are you running Bazel on?

Windows 10, but the problem does not seem OS specific.

### What's the output of `bazel info release`?

2.1

###  Have you found anything relevant by searching the web?

https://www.bountysource.com/issues/69193507-incompatible_enable_cc_toolchain_resolution-turn-on-toolchain-resolution-for-cc-rules - indicates the same problem and https://github.com/bazelbuild/bazel/pull/8459 does not fix it.

katre:
This is internal to the CC rules implementation and doesn't involve configurability.

oquenchil:
Fixed by chancila's patch.

PR #12252, opened 2020-10-11
Ensure BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN does not break builds

chancila:
The workspace prefix calls `cc_configure` which causes the toolchain in `@local_config_cc` to be evaluated.

Currently the evaluation of `@local_config_cc//:toolchain` with toolchain detection turned off causes an error in CcToolchainSuite due to none of the cpu labels matching the local host cpu.

This patch fixes this issue by templating the cpu label to the host cpu in `cc_autoconf` when `BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN` is true. 

Addresses https://github.com/bazelbuild/bazel/issues/10756

### DIFF
--- a/tools/cpp/BUILD.empty.tpl
+++ b/tools/cpp/BUILD.empty.tpl
@@ -15,6 +15,7 @@
 package(default_visibility = ["//visibility:public"])
 
 load(":cc_toolchain_config.bzl", "cc_toolchain_config")
+load("@rules_cc//cc:defs.bzl", "cc_toolchain_suite", "cc_toolchain")
 
 cc_library(
     name = "malloc",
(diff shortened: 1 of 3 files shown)


tools/cpp/BUILD.empty.tpl
```diff
@@ -28,8 +29,8 @@ filegroup(
 cc_toolchain_suite(
     name = "toolchain",
     toolchains = {
-        "local|local": ":local",
-        "local": ":local",
```
chancila, line 32: it's unclear to me if the 'local' tag is suppose to have any meaning anymore, I wasn't able to dig into the history, was this the default name of the host configuration before platforms were a thing?

oquenchil (approved):
LGTM

PR #12258, opened 2020-10-13
Export `.bzl` files in `@bazel_tools//tools/...`

achew22:
With this change it is now possible to create a `bzl_library` entry that has a `deps` entry for all its inputs.

### DIFF
--- a/tools/build_defs/repo/BUILD
+++ b/tools/build_defs/repo/BUILD
@@ -4,6 +4,8 @@ load("@rules_pkg//:pkg.bzl", "pkg_tar")
 
 licenses(["notice"])  # Apache 2.0
 
+exports_files(glob(["*.bzl"]))
+
 filegroup(
     name = "srcs",
     srcs = glob(["**"]),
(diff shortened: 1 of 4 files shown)


achew22:
@c-parsons friendly ping

achew22:
@c-parsons friendly ping on this

jin:
This seems to be blocking rules_go from working nicely with Stardoc according to https://github.com/bazelbuild/bazel-gazelle/pull/760#issuecomment-712961346. @c-parsons, could you take a look please?

jin:
cc @alandonovan and @brandjon as the owners of Starlark docgen tooling

achew22:
As a note, I think that something like this needs to be done for every `.bzl` file in `@bazel_tools` and related repos since stardoc requires all files that were transitively loaded to be in the `data` dependency for parsing. Unfortunately I haven't been able to come up with a clever way to test this and ensure the transitive set of loads is visible with just this change because I'm blocked on a more general error in [Stardoc](https://github.com/bazelbuild/stardoc/issues/76)

achew22:
Friendly ping

achew22:
Even friendlier ping

jin:
@brandjon could you take a look at this please?

achew22:
Friendly ping

PR #12261, opened 2020-10-13
Add a test which verifies that all URLs mentioned in WORKSPACE actually

katre:
exist.

Helps prevent #12262 from happening again.

### DIFF
--- a/src/test/shell/bazel/BUILD
+++ b/src/test/shell/bazel/BUILD
@@ -1352,4 +1352,8 @@ sh_test(
         "//:workspace-file",
         "@bazel_tools//tools/bash/runfiles",
     ],
+    tags = [
+        "no_windows",
+        "requires-network",
+    ],
 )
--- a/src/test/shell/bazel/BUILD
+++ b/src/test/shell/bazel/BUILD
@@ -1343,3 +1343,13 @@ sh_test(
     tags = ["no_windows"],
     deps = ["@bazel_tools//tools/bash/runfiles"],
 )
+
+sh_test(
+    name = "verify_workspace",
+    srcs = ["verify_workspace.sh"],
+    data = [
+        ":test-deps",
+        "//:workspace-file",
+        "@bazel_tools//tools/bash/runfiles",
+    ],
+)
--- a/src/test/shell/bazel/verify_workspace.sh
+++ b/src/test/shell/bazel/verify_workspace.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+#
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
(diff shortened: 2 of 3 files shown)


katre:
Looks like network access is blocked (sensibly) on most workers.

Should the test only veryify mirror.bazel.build urls?

src/test/shell/bazel/verify_workspace.sh
```diff
@@ -0,0 +1,66 @@
...
+      invalid_urls+=("${url}")
+    fi
+  done
+
+  if [[ ${#invalid_urls[@]} > 0 ]]; then
+    fail "Invalid urls: ${invalid_urls[@]}"
+  fi
+}
+
+run_suite "verofy_workspace"
```
philwo, line 66: Typo: verofy?

src/test/shell/bazel/verify_workspace.sh
```diff
@@ -0,0 +1,66 @@
...
+  # Find url-shaped lines, skipping jekyll-tree (which isn't a valid URL), and
+  # skipping comments.
+  invalid_urls=()
+  for url in $(grep -E '"https://|http://' "${WORKSPACE_FILE}" | \
+    grep -v 'jekyll-tree' | \
+    grep -v '^ *#' | \
+    perl -pe 's#.*"(https?://[^"]+)".*#$1#g' | \
+    sort -u); do
+    #echo "Checking ${url}"
+    if ! curl --head -silent --fail --output /dev/null "${url}"; then
```
philwo, line 55: I wonder if we should add `--retry 3` here to ignore occasional HTTP errors.

src/test/shell/bazel/BUILD
```diff
@@ -1343,3 +1343,13 @@ sh_test(
     tags = ["no_windows"],
     deps = ["@bazel_tools//tools/bash/runfiles"],
 )
+
+sh_test(
+    name = "verify_workspace",
+    srcs = ["verify_workspace.sh"],
+    data = [
```
philwo, line 1350: You can add the `"requires-network"` tag to enable network access for this test.

philwo (commented):
That's a great idea! Thank you for coming up with this!

src/test/shell/bazel/verify_workspace.sh
katre: Done.

src/test/shell/bazel/BUILD
katre: Thanks, added that and "no-windows", since there's not really any benefit to running this on every platform.

src/test/shell/bazel/verify_workspace.sh
katre: Done.

Issue #7687, opened 2019-03-11
incompatible_do_not_split_linking_cmdline: Order of C++ link flags affected


oquenchil:
**Flag**: --incompatible_do_not_split_linking_cmdline
**Available since**: 0.25 (March 2019 release)
**Will be flipped in**: 0.26 (April 2019 release)
**Tracking issue**: #7670

This changes how flags are passed to the linker. If you don't have your own custom CROSSTOOL, then you don't need to migrate anything, but you should still test with this flag to see that nothing breaks. If it does, please let us know.

**If you have your own custom CROSSTOOL:**
If the CROSSTOOL doesn't support linker parameter files, then you don't have to migrate anything. If it supports linker parameter file but you do not specify the feature 'linker_param_file' in  your CROSSTOOL, then you don't have to migrate anything.

If you specify the feature 'linker_param_file' and the linker is using gcc-style linker flags, then you should modify the feature so that instead of having a parameter file specified like this:
`      flag: '-Wl,@%{linker_param_file}'`
it is like this:
`      flag: '@%{linker_param_file}'`

In other words, the parameter file should not be passed directly to the linker with `-Wl,`. In addition to this you should specify the feature 'do_not_split_linking_cmdline' in your CROSSTOOL, whether it's enabled or disabled it does not matter, we only check its presence. This will tell Bazel to use the new behavior with this toolchain. After the flip of `incompatible_disable_legacy_cc_provider` this feature will do nothing, it can safely be deleted.

This change will put every flag inside the parameter file whereas before only flags starting with `-l` were put there, i.e.: all libraries. This might cause the order of link flags to change with respect to what you originally had in your CROSSTOOL, if it does you will have to re-order the corresponding flags. If you have to make any changes to the CROSSTOOL, make sure that you add the feature 'do_not_split_linking_cmdline' too. Example of change:

**Old:**
```
feature {
  name: 'linker_param_file',
  flag_set {
         action: 'c++-link-executable'
         action: 'c++-link-dynamic-library'
         action: 'c++-link-nodeps-dynamic-library'
         flag_group {
             expand_if_all_available: 'linker_param_file'
             flag: '-Wl,@%{linker_param_file}'
         }
}
```

**New:**
```
feature {
  name: 'do_not_split_linking_cmdline'
}
feature {
  name: 'linker_param_file',
  flag_set {
         action: 'c++-link-executable'
         action: 'c++-link-dynamic-library'
[...]

PR #12265, opened 2020-10-14
Add support for params files for darwin

keith:
Clang has supported params files for a while now. This updates the cc
toolchain for darwin to use them.

The logic for processing response files is mostly copied from
rules_swift where similar processing is done.


tools/cpp/BUILD.tpl
```diff
@@ -66,7 +66,7 @@ cc_toolchain(
     linker_files = ":compiler_deps",
     objcopy_files = ":empty",
     strip_files = ":empty",
-    supports_param_files = %{supports_param_files},
```
keith, line 69: while this substitution is always true here now, we may want to keep this if it's considered public API?

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5194,20 +5194,12 @@ def _impl(ctx):
     else:
         unfiltered_compile_flags_feature = None
 
+    do_not_split_linking_cmdline = feature(name = "do_not_split_linking_cmdline")
```
keith, line 5197: https://github.com/bazelbuild/bazel/issues/7687

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5194,20 +5194,12 @@ def _impl(ctx):
...
-                flag_groups = [
-                    flag_group(
-                        flags = ["-Wl,@%{linker_param_file}"],
-                        expand_if_available = "linker_param_file",
-                    ),
-                ],
-            ),
-            flag_set(
-                actions = [ACTION_NAMES.cpp_link_static_library],
+                actions = all_link_actions + [ACTION_NAMES.cpp_link_static_library],
```
keith, line 5202: We might want to expand this to the objc linking actions as well? (`all_link_actions` does not contain those)

segiddins (approved):

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5194,20 +5194,18 @@ def _impl(ctx):
...
-                    flag_group(
-                        flags = ["-Wl,@%{linker_param_file}"],
-                        expand_if_available = "linker_param_file",
-                    ),
+                actions = all_link_actions + [
+                    ACTION_NAMES.cpp_link_static_library,
+                    "objc-archive",
+                    "objc-fully-link",
+                    "objc-executable",
+                    "objc++-executable",
```
thii, line 5207: These all have predefined constants.
https://github.com/bazelbuild/bazel/blob/83f62a4f9bb094c4c30b843d7d44bff6c6f5abd8/tools/build_defs/cc/action_names.bzl#L115-L120

keith:
@oquenchil I would love this to make the 3.8 cut if possible

tools/osx/crosstool/cc_toolchain_config.bzl
oquenchil: I'm confused, why do you need to add the feature? That flag was flipped

tools/cpp/BUILD.tpl
oquenchil: I think documentation should show that the default is true, but I think it's fine to leave it too.

tools/osx/crosstool/cc_toolchain_config.bzl
keith: I found without this this didn't work at all and no arguments were passed to the linker.

tools/cpp/BUILD.tpl
keith: great going to remove it then for clarity

keith:
rebased and added a test for the new behavior

tools/osx/crosstool/cc_toolchain_config.bzl
oquenchil: The reason for that must be this: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java;drc=a5a4f47295ccb6cd81cf8732772c849b1a8b9d58;l=1500

Can you make sure you don't have the flag `incompatible_do_not_split_linking_cmdline` anywhere? Bazelrc perhaps?

This feature and the flag should eventually go away.

keith: I definitely don't have that one set. I can try it in a more isolated environment though.

keith: My assumption is when that flag went away the checks that required this would also go away and we could delete this then

keith: ok I'm not sure what was happening in my case before, but you're right this isn't needed.

keith:
Looks like I actually have to make some more fixes here and add the objc link actions to here:

https://github.com/bazelbuild/bazel/blob/a3a00f222b9b9c5a651e2176c39dd707182226ac/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java#L515-L540

Right off the bat it looks like that breaks when passing `-rpath @loader_path` because something in bazel assumes `@loader_path` is a params file. I'll have to debug this more next week.

src/main/java/com/google/devtools/build/lib/rules/cpp/LinkCommandLine.java
```diff
@@ -296,39 +296,43 @@ CommandLine paramCmdLine() {
         doNotSplitLinkingCmdLine);
   }
 
-  public static void extractArgumentsForStaticLinkParamFile(
+  private static boolean isKnownAtArgument(String arg) {
```
keith, line 299: This feels like a huge hack, I would love to know if anyone else has ideas here. Theoretically we could force all users to pass `--linkopt=-Wl,-rpath,@executable_path` instead of `--linkopt=-rpath --linkopt=@executable_path`, but short of potentially breaking existing users, I don't see another path for this. In rules_swift we're able to check if the argument exists as a params file, but I don't think that would work in this context since the files wouldn't exist yet?

keith: cc @googlewalt I was told you might have some idea here

googlewalt: I discussed with @trybka and others internally.  bazel is trying to get out of the business of understanding command lines, so we'd rather not to add more semantics there.

Suggested way to proceed:

* There is one instance of `-Xlinker -rpath -Xlinker @loader_path` in native code for AppleBinary -- I can fix that.
* Make sure that the existing external toolchain does not have any uses of prefix `@executable_path`, `@loader_path`, and `@rpath` arguments.  I don't believe it does?
* Officially, param files will not work with `-Xlinker -rpath -Xlinker @loader_path/@executable_path`.  The few (if any) affected people can either fix those flags, or turn off param files.

Then we can drop this patch.

keith: If you're happy with that I'm up for making those changes! I noticed those cases in bazel as well and they do seem trivially replacable. There are also some in the rules that we can fix https://github.com/bazelbuild/rules_apple/pull/941

keith: Updated, lmk what you think!

keith:
There's an issue where currently the params files conflict if you run a build for multiple architectures at once. I would have expected this to have been solved for C++ even. Looking into it

trybka:
That's surprising--I would have thought they would have been differentiated by the output configuration dir?

keith:
Same, it's very easy to repro tho:

```bzl
% cat BUILD
apple_binary(
    name = "bin",
    minimum_os_version = "10.0",
    platform_type = "ios",
    deps = [":lib"],
)

objc_library(
    name = "lib",
    srcs = ["main.m"],
)
```

```
$ bazel build :bin --ios_multi_cpus=i386,x86_64
ERROR: file 'bin_bin-2.params' is generated by these conflicting actions:
Label: //:bin
RuleClass: apple_binary rule
Configuration: ed93f65d05ded6845d66048ac3ade22d7d88b5de42c13efee80d216689eeaa76
Mnemonic: FileWrite
Action key: 447c15515b5910cd244d50bb7f2bd0e703d82814e00b3099e9e802e6e0a36f68, c7d79bd92b4d69d8111e4e57bc225fa82e4fd84fd65dfeeba124fa50f90107ab
Progress message: Writing file bin_bin-2.params
PrimaryInput: (null)
PrimaryOutput: File:[[<execution_root>]bazel-out/apl-darwin_x86_64-fastbuild/bin]bin_bin-2.params
Owner information: ConfiguredTargetKey{label=//:bin, config=BuildConfigurationValue.Key[ed93f65d05ded6845d66048ac3ade22d7d88b5de42c13efee80d216689eeaa76]}
MandatoryInputs: are equal
Outputs: are equal
ERROR: Analysis of target '//:bin' failed; build aborted: for bin_bin-2.params, previous action: action 'Writing file bin_bin-2.params', attempted action: action 'Writing file bin_bin-2.params'
```

keith:
I think it's because the configuration of the params files is the same, since it's the exec configuration not the target configuration?

keith:
So I discovered that in some case a different configuration was being used for the link actions, which fixes this params file issue. I had to make the change for fully linked static libraries and for multi arch executables. The big question now is what other things might this break?

keith:
I'll have to check the windows tests, but everything for darwin passes now. But I would love some assurance from someone who knows this logic better whether or not the configuration changes make any sense.

src/main/java/com/google/devtools/build/lib/rules/objc/AppleBinary.java
```diff
@@ -291,7 +291,7 @@ private static ExtraLinkArgs getRequiredLinkopts(RuleContext ruleContext)
     switch (binaryType) {
       case LOADABLE_BUNDLE:
         extraLinkArgs.add("-bundle");
-        extraLinkArgs.add("-Xlinker", "-rpath", "-Xlinker", "@loader_path/Frameworks");
+        extraLinkArgs.add("-Wl,-rpath,@loader_path/Frameworks");
```
googlewalt, line 294: FYI I prepared an identical patch last night and sanity checked that it didn't break anything.  It should be landing shortly.

keith: sounds great, I'll revert once that is pushed out

keith:
The windows test failures appear to have been a fluke. Fixed after a rebase. So this might be good to go

oquenchil (approved):
LGTM

keith:
Thanks for the review! Btw, is this being imported? (I never know if that's implicit with the +1)

lberki:
It should be. Generally, it's the "sponsor" of the pull request who does it, who is @oquenchil in this case.

brentleyjones:
How do we know who the sponsor is. I will admit that the current process confuses me as well, it's not clear to me when it's being imported. Can we get a label or something?

keith:
Thanks everyone!

oquenchil:
Unfortunately, this had to be rolled back because it broke tests internally.

keith:
Can you provide any details I can use to fix the issue?

oquenchil:
I won't have time today and I can't share the internal log here. I will take a look tomorrow and share the details, maybe I can fix it myself.

oquenchil:
Apologies, this wasn't internal at all:
Before rollback: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1743
After rollback: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1744

src/main/java/com/google/devtools/build/lib/rules/objc/AppleBinary.java
```diff
@@ -291,7 +291,7 @@ private static ExtraLinkArgs getRequiredLinkopts(RuleContext ruleContext)
     switch (binaryType) {
       case LOADABLE_BUNDLE:
         extraLinkArgs.add("-bundle");
-        extraLinkArgs.add("-Wl,-rpath,@loader_path/Frameworks");
+        extraLinkArgs.add("-Xlinker", "-rpath", "-Xlinker", "@loader_path/Frameworks");
```
keith, line 294: this being reverted was 1 bug

keith:
Is there any way I can run that ahead of time? The comment I made above was 1 issue that was caused by conflicting changes. This also depends on https://github.com/bazelbuild/rules_apple/pull/973 which I was hoping we could merge after this (naive in retrospect), but the big question here is how many other repos in this test matrix will require being pinned to the new version after this rules change

keith:
Here's an update for the examples repo https://github.com/bazelbuild/examples/pull/186

keith:
Here's the update for rules_go https://github.com/bazelbuild/rules_go/pull/2699

keith:
I think once those 2 issues are resolved we should be good to give it another go

oquenchil:
I merged the first one, please ping this thread when the rules_go PR is merged, then I will submit this again.

Regarding testing with downstream, the information is here:
https://github.com/bazelbuild/continuous-integration/blob/master/docs/downstream-testing.md

You have to request access but they will definitely grant it to you.

Yannic:
Downstream: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1747#ad818f41-8ad0-4dad-8ced-485006558e9d

keith:
Ok rules_go is merged, the fix for rules_cc is here https://github.com/bazelbuild/rules_cc/pull/89 because that repo duplicates the wrappers I changed here

keith:
I re-ran a downstream build after the rules_go fix merged https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1755#_ going to look at tensorflow once more

keith:
Ok I figured it out, tensorflow is building with Xcode 10.3, which doesn't support params files to libtool. AFAIK 10.3 isn't _really_ considered supported at this point so I've submitted this change attempting to bump the Xcode version it uses: https://github.com/bazelbuild/continuous-integration/pull/1057

keith:
@oquenchil this should be ready for you again! Tensorflow passed now that the Xcode version has been bumped https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1757

So the last thing is this rules_cc change! https://github.com/bazelbuild/rules_cc/pull/89

oquenchil (approved):
LGTM

keith:
Rebased to resolve the conflict!

keith:
Rebased to resolve my other conflict 😛

keith:
I started another downstream build because I assume we have some time here https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1760#7612b6a6-0422-4580-b2a3-e78791c31297

keith:
Seems like `incompatible_restrict_string_escapes` and some other changes broke a lot of downstream projects. Nothing stands out as related to this change anymore.

keith:
Thanks for all the help!

PR #12249, opened 2020-10-10
Generate fish shell completion at build time

akirabaruah:
Fish shell completion for bazel previously relied on parsing bazel help
text at run time, leading to latency of multiple seconds for first-time
completion. This change adds a script that instead parses bazel help
text and generates an appropriate fish completion script at build time,
greatly reducing completion latency for the user.

Fixes #12206
Fixes #12207
Fixes #12208
Fixes #12209
Fixes #12210

PR #12267, opened 2020-10-14
Add fish shell completion hint to installer

akirabaruah:
See original request in
https://github.com/bazelbuild/bazel/pull/12249#discussion_r504056101

### DIFF
--- a/scripts/packages/template_bin.sh
+++ b/scripts/packages/template_bin.sh
@@ -178,9 +178,14 @@ cat <<EOF
 
 Bazel is now installed!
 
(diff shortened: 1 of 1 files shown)


scripts/packages/template_bin.sh
```diff
@@ -182,6 +182,10 @@ Make sure you have "${bin}" in your path. You can also activate bash
 completion by adding the following line to your ${BASHRC}:
```
meisterT, line 182: Can you start a new paragraph for the bash completion as well?

meisterT (approved):

PR #12268, opened 2020-10-14
Flush err after completing standalone jar build request.

benjaminp:
Otherwise, diagnostics are eaten.

### DIFF
--- a/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/BazelJavaBuilder.java
+++ b/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/BazelJavaBuilder.java
@@ -43,10 +43,15 @@ public class BazelJavaBuilder {
       WorkRequestHandler workerHandler = new WorkRequestHandler(builder::parseAndBuild);
       System.exit(workerHandler.processRequests(System.in, System.out, System.err));
     } else {
-      System.exit(
-          builder.parseAndBuild(
-              Arrays.asList(args),
-              new PrintWriter(new OutputStreamWriter(System.err, Charset.defaultCharset()))));
+      PrintWriter pw =
+          new PrintWriter(new OutputStreamWriter(System.err, Charset.defaultCharset()));
+      int returnCode;
+      try {
+        returnCode = builder.parseAndBuild(Arrays.asList(args), pw);
+      } finally {
+        pw.flush();
+      }
+      System.exit(returnCode);
     }
   }
 


src/java_tools/buildjar/java/com/google/devtools/build/buildjar/BazelJavaBuilder.java
```diff
@@ -43,10 +43,15 @@ public static void main(String[] args) {
       WorkRequestHandler workerHandler = new WorkRequestHandler(builder::parseAndBuild);
       System.exit(workerHandler.processRequests(System.in, System.out, System.err));
     } else {
-      System.exit(
-          builder.parseAndBuild(
-              Arrays.asList(args),
-              new PrintWriter(new OutputStreamWriter(System.err, Charset.defaultCharset()))));
+      PrintWriter pw =
```
comius, line 46: Can you use try-with-resources?

benjaminp: The goal is to call `flush()` in the finally not `close()` (which would impolitely close the underlying stderr).

comius: Doesn't `close` also call `flush`? And don't you just call `System.exit()` right after? So it might actually be polite to close stderr before :)

comius: Sorry, you're right, System.exit is too abrupt to be handled by try-with-resources.

benjaminp: Yes, currently it doesn't matter much because the program promptly exits, but it is a refactoring hazard.

Anyway, I don't feel strongly about this, so I'll just autoclose.

comius (approved):

src/java_tools/buildjar/java/com/google/devtools/build/buildjar/BazelJavaBuilder.java
cushon: I agree we don't want to close `System.err`, otherwise if something goes wrong between closing it and exiting we'll get a context-free failure. Calling `flush` is the right thing to do.

comius: Sorry @benjaminp again. I stand corrected.

comius (approved):

PR #12269, opened 2020-10-14
Compile macOS toolchain tools with optimizations

keith:
These are invoked for every action so any performance improvement we can
make is worth it.


segiddins (approved):

oquenchil (approved):
LGTM

Issue #11945, opened 2020-08-13
target visibility attribute pointing to its direct consumer causes cycle in dependency graph  


konste:
### Description of the problem / feature request:

When target A depends on target B and target B has visibility attribute pointing to target A Bazel sees it as the dependency loop and errors out.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Repro is [here](https://github.com/Bazel-snippets/repro_visibility_loop).
Short snippet:
```
# In one package:
cc_binary(
    name = "hello-world",
    srcs = ["hello-world.cc"],
    deps = [
        "//hello-greet",
    ],
)
# In another package:
cc_library(
    name = "hello-greet",
    srcs = ["hello-greet.cc"],
    hdrs = ["hello-greet.h"],
    visibility = ["//hello-world"],
)
```
Bazel considers visibility as a dependency, while it is normally directed towards consumers, i.e. against the actual dependency direction.
```
bazel build //hello-world
ERROR: D:/dev/bazel/repro/repro_visibility_loop/hello-world/BUILD:3:10: in cc_binary rule //hello-world:hello-world: cycle in dependency graph:
    //hello-world:hello-world
    //hello-greet:hello-greet
.-> //hello-world:hello-world
|   //hello-greet:hello-greet
`-- //hello-world:hello-world
This cycle occurred because of a configuration option
```

### What operating system are you running Bazel on?

Mac, Linux, Windows

### What's the output of `bazel info release`?

3.5.0RC2

ulfjack:
That is working as intended. There are only five things you are allowed to put in visibility attributes:
1. `//visibility:public`
2. `//visibility:private`
3. `//foo/bar:__pkg__`
4. `//foo/bar:__subpackages__`
5. `//foo/bar:baz` which must be a `package_group` rule

See documentation here: https://docs.bazel.build/versions/master/visibility.html

Pointing it at an individual rule does precisely nothing in terms of visibility, and returns an expected dependency cycle error.

konste:
`Pointing it at an individual rule does precisely nothing in terms of visibility` What I was trying to express is that particular target in one package is only supposed to be used by exactly one specific target in another package and should not be visible to anything else. So it is not possible to express that? There is no way to specify visibility with more precise granularity than the package?

konste:
[...]

PR #12272, opened 2020-10-15
Restore target cycle errors to their former grandeur.

benjaminp:
In recent years, a frustrating error for Bazel users is the one caused by putting a reverse dependency rather than a package group into a target's visibility attribute. Such a mistake generally resulted in an error like this:
```
BUILD:7:10: in filegroup rule //pkg1:actual_target: cycle in dependency graph:
    //pkg1:actual_target
    //pkg1:dep_with_visibility
.-> //pkg1:actual_target
|   //pkg1:dep_with_visibility
`-- //pkg1:actual_target
This cycle occurred because of a configuration option
```
This was confusing primarily because users don't think of visibility as causing a dependency—see, for instance, https://github.com/bazelbuild/bazel/issues/11945. Additionally, the "This cycle occurred because of a configuration option" hint was not usually accurate or helpful.

That the visibility mistake will be reported as a cycle is a fundamental property of how analysis is mapped onto Skyframe today, but a better hint was certainly possible. In fact, deep within Bazel's bowels there was already code to produce a better message, but it became mostly dormant during the analysis-loading interleaving work of 2017.

This CL removes the "This cycle occurred because of a configuration option" hint completely and restores the old visibility hint when appropriate. The primary code change is to combine ConfiguredTargetCycleReporter and TransitiveTargetCycleReporter to form TargetCycleReporter.

Now, a visibility cycle looks like this:
```
BUILD:2:10: in filegroup rule //pkg1:actual_target: cycle in dependency graph:
    //pkg1:actual_target
    //pkg1:dep_with_visibility
.-> //pkg1:actual_target
|   //pkg1:dep_with_visibility
`-- //pkg1:actual_target
The cycle is caused by a visibility edge from //pkg1:dep_with_visibility to the non-package_group target //pkg1:actual_target. Note that visibility labels are supposed to be package_group targets, which prevents cycles of this form.
```

I humbly include this change in the https://github.com/bazelbuild/bazel/issues/11984 program.


meisterT:
Cc @ulfjack

src/main/java/com/google/devtools/build/lib/skyframe/TargetCycleReporter.java
```diff
@@ -13,49 +13,73 @@
...
+    return SUPPORTED_SKY_KEY.apply(topLevelKey)
+        && cycleInfo.getPathToCycle().stream().allMatch(SUPPORTED_SKY_KEY)
+        && cycleInfo.getCycle().stream().allMatch(SUPPORTED_SKY_KEY);
   }
 
   @Override
-  protected Label getLabel(SkyKey key) {
-    return ((TransitiveTargetKey) key).getLabel();
+  public String prettyPrint(SkyKey key) {
+    if (SkyFunctions.isSkyFunction(SkyFunctions.CONFIGURED_TARGET).apply(key)) {
```
ulfjack, line 63: Why not check if instanceof? I think that would be clearer here, unless there's a good reason not to.

src/main/java/com/google/devtools/build/lib/skyframe/TargetCycleReporter.java
```diff
@@ -13,49 +13,73 @@
...
+      return ((AspectValueKey) key).getDescription();
+    } else {
+      return getLabel(key).toString();
+    }
   }
 
   @Override
-  protected boolean shouldSkipOnPathToCycle(SkyKey key) {
-    return SkyFunctions.PREPARE_ANALYSIS_PHASE.equals(key.functionName());
+  public Label getLabel(SkyKey key) {
```
ulfjack, line 75: It's been a while since I looked at this code. Why's this a different set of key types than prettyPrint?

ulfjack (commented):
I like it! Change looks straightforward to me, but I don't have time to investigate more deeply right now. If you want an approval for me, that'll have to wait until next week, I'm afraid.

src/main/java/com/google/devtools/build/lib/skyframe/TargetCycleReporter.java
benjaminp: good idea

src/main/java/com/google/devtools/build/lib/skyframe/TargetCycleReporter.java
benjaminp: Both the CT and aspect keys implement `ActionLookupValue`, which can be handled generically here but want to be pretty-printed differently.

meisterT:
cc @janakdr

janakdr (approved):

PR #12280, opened 2020-10-15
Delete old guava jars.

benjaminp:


benjaminp:
should work after #12281

meteorcloudy (approved):

PR #12284, opened 2020-10-15
Load rules_cc from cc_configure.WORKSPACE.

katre:
rules_cc is referenced from @bazel_tools//tools/cpp, and was previously only working by accident (because it is also loaded from jdk.WORKSPACE).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -503,6 +503,8 @@ http_archive(
     ],
 )
 
+# This must be kept in sync with src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE.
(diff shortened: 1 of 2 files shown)


lberki (commented):
Why do this? We gave up on `rules_cc` for the time being.

katre:
Right, but until it's removed from use, it still needs to be loaded. Right now, it's loaded on accident because jdk.WORKSPACE happens to be processed before cc_configure.WORKSPACE. With this change, we're explicitly declaring the dependency instead of relying on rule loading order.

When we stop actually using rules_cc, we can remove this from everywhere it's referenced.

lberki:
Argh. I missed the "Is referenced from `@bazel_tools//tools/cpp` part. Move along.

lberki (approved):

PR #12285, opened 2020-10-15
Add relative_ast_path_feature

keith:
This feature applies for darwin builds to disable absolutizing
swiftmodule paths to link commands. Previously these paths were
absolutized to support debugging. This is not required if users change
the pwd of their debugger so that the relative paths are valid. Since we
don't have a way to force that on users, absolutizing them is still the
default, but users who configure that can enable this feature to improve
hermeticity of their builds.

RELNOTES: Add `relative_ast_path` feature for darwin builds to relativize swiftmodule paths for debugging

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5218,6 +5218,24 @@ def _impl(ctx):
         ],
     )
 
+    relative_ast_path_feature = feature(
+        name = "relative_ast_path",
+        env_sets = [
+            env_set(
+                actions = all_link_actions + [
+                    "objc-executable",
+                    "objc++-executable",
+                ],
+                env_entries = [
+                    env_entry(
+                        key = "RELATIVE_AST_PATH",
+                        value = "true",
+                    ),
+                ],
+            ),
+        ],
+    )
+
     archiver_flags_feature = feature(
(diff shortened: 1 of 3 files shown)


keith:
tbh I'm not sure a feature / env var is the best way to handle this, but I can't think of another way to have a "global" setting like this

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -5218,6 +5218,24 @@ def _impl(ctx):
         ],
     )
 
+    relative_ast_path_feature = feature(
+        name = "relative_ast_path",
+        env_sets = [
+            env_set(
+                actions = all_link_actions + [
+                    "objc-executable",
+                    "objc++-executable",
```
thii, line 5227:
```suggestion
                    ACTION_NAMES.objc_executable,
                    ACTION_NAMES.objcpp_executable,
```

keith:
@thii can you accept the CLA

thii:
@googlebot I fixed it.

keith:
@oquenchil I would love this to make the 3.8 cut if possible

oquenchil (approved):
LGTM

keith:
Thanks! Can you help land this or do we need to cc someone?

keith:
Thanks!

PR #12286, opened 2020-10-15
starlark: dedent continued function param lines

alexeagle:
This allows multi-line parameter documentation under Args: without extra indents.
Fixes https://github.com/bazelbuild/stardoc/issues/74

### DIFF
--- a/src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
+++ b/src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
@@ -479,23 +479,25 @@ public final class DocstringUtils {
     /** Parses additional lines that can come after "param: foo" in an 'Args' section. */
     private void parseContinuedParamDescription(
         int baselineIndentation, StringBuilder description) {
-      // continuations are de-dented to the minimum value seen
+      // Two iterations: first buffer lines and find the minimal indent, then trim to the min
+      List<String> buffer = new ArrayList();
       int continuationIndentation = Integer.MAX_VALUE;
       while (nextLine()) {
-        if (line.isEmpty()) {
-          description.append('\n');
-          continue;
+        if (!line.isEmpty()) {
+          if (getIndentation(line) <= baselineIndentation) {
+            break;
+          }
+          continuationIndentation = Math.min(getIndentation(line), continuationIndentation);
         }
-        int lineIndentation = getIndentation(line);
-        if (lineIndentation <= baselineIndentation) {
-          break;
-        }
-        if (lineIndentation < continuationIndentation) {
-          continuationIndentation = lineIndentation;
-        }
-        String trimmedLine = line.substring(continuationIndentation);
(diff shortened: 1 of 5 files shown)


alexeagle:
ping @c-parsons are you the right person to merge?

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
```diff
@@ -479,15 +479,21 @@ private String checkForNonStandardDeprecation(String line) {
...
           description.append('\n');
           continue;
         }
-        if (getIndentation(line) <= baselineIndentation) {
+        int lineIndentation = getIndentation(line);
+        if (lineIndentation <= baselineIndentation) {
           break;
         }
-        String trimmedLine = line.substring(baselineIndentation);
+        if (lineIndentation < continuationIndentation) {
```
c-parsons, line 493: We're not truly finding the minimum indentation value here, as we're still processing as we go.
Consider the difference in behavior for:

```
  Line one
          Line two
                    Line three
```

versus

```
               Line one
        Line two
   Line three
```

c-parsons (commented):
Could you modify `pure_markdown_test` as well to demonstrate how this change affects markdown tables?

c-parsons:
FYI @alandonovan @brandjon

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
alexeagle: correct, but the failure mode is that you get some extra whitespace if the continuation is accidentally not consistent. For your example I think

```
Line one
        Line two
                  Line three
```
and
```
Line one
Line two
Line three
```
are the best outcome we'll get without scanning ahead for the minimal intent. Do you think it's worth adding the complexity to scan ahead?

alexeagle:
Added the change to `pure_markdown_template_test` as requested :)

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
c-parsons: Yes, I don't think it's much more complex to scan ahead (we're not dealing with gigantic strings here, so it should be fine to store all the lines in a list, take the min indent, and then iterate over)

I foresee later user confusion / frustration with the current behavior, if we don't fix this right the first time.

Thanks!

c-parsons:
FYI @alandonovan

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
alexeagle: Okay, done PTAL

alexeagle:
ping @c-parsons mergey merge-merge?

c-parsons:
Merge in review. Thanks for your patience.

PR #12292, opened 2020-10-16
ijar: Avoid running past the end of a static buffer.

benjaminp:
```
INFO: From Compiling third_party/ijar/zip_main.cc:
In file included from /usr/include/string.h:495,
                 from third_party/ijar/zip_main.cc:28:
In function 'char* strncpy(char*, const char*, size_t)',
    inlined from 'int devtools_ijar::extract(char*, char*, char**, bool, bool, bool)' at third_party/ijar/zip_main.cc:159:12:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:106:34: warning: 'char* __builtin_strncpy(char*, const char*, long unsigned int)' specified bound 4096 equals destination size [-Wstringop-truncation]
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/third_party/ijar/zip_main.cc
+++ b/third_party/ijar/zip_main.cc
@@ -80,21 +80,26 @@ class UnzipProcessor : public ZipExtractorProcessor {
 };
 
 // Concatene 2 path, path1 and path2, using / as a directory separator and
-// puting the result in "out". "size" specify the size of the output buffer
-void concat_path(char* out, const size_t size,
-                 const char *path1, const char *path2) {
+// putting the result in "out". "size" specify the size of the output buffer. If
+// the result would overflow the output buffer, print an error message and
+// return false.
+bool concat_path(char *out, const size_t size, const char *path1,
+                 const char *path2) {
   int len1 = strlen(path1);
   size_t l = len1;
   strncpy(out, path1, size - 1);
-  out[size-1] = 0;
+  out[size - 1] = 0;
   if (l < size - 1 && path1[len1] != '/' && path2[0] != '/') {
     out[l] = '/';
     l++;
     out[l] = 0;
   }
-  if (l < size - 1) {
-    strncat(out, path2, size - 1 - l);
+  if (l >= size - 1) {
+    fprintf(stderr, "paths too long to concat: %s + %s", path1, path2);
+    return false;
   }
+  strncat(out, path2, size - 1 - l);
+  return true;
 }
 
 void UnzipProcessor::Process(const char* filename, const u4 attr,
@@ -123,8 +128,8 @@ void UnzipProcessor::Process(const char* filename, const u4 attr,
   }
   if (extract_) {
     char path[PATH_MAX];
-    concat_path(path, PATH_MAX, output_root_, output_file_name);
-    if (!make_dirs(path, perm) ||
+    if (!concat_path(path, sizeof(path), output_root_, output_file_name) ||
+        !make_dirs(path, perm) ||
         (!isdir && !write_file(path, perm, data, size))) {
       abort();
     }
@@ -140,8 +145,8 @@ void basename(const char *path, char *output, size_t output_size) {
   } else {
     pointer++;  // Skip the leading slash.
   }
-  strncpy(output, pointer, output_size);
-  output[output_size-1] = 0;
+  strncpy(output, pointer, output_size - 1);
+  output[output_size - 1] = 0;
 }
 
 // Execute the extraction (or just listing if just v is provided)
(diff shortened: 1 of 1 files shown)


aiuto:
Is there an upstream change that this is reflecting? 
I worry that if we commit this, we'll just wipe it out next time we update

See:  https://github.com/bazelbuild/bazel/issues/12299

benjaminp:
I was under the impression that that open-source SoT for ijar is `third_party` in this repository—ijar is a Google thing. I've submitted PRs against it before (5a405eb93c092aa9bfd1c69a3fcbdb0997e3c55b).

aiuto:
Yeah. This seem strange. Being in third_party presumes the SoT is elsewhere, but I can not find an externl repo anywhere.
I'll let the EngProd team figure out who can import and merge this PR. My concern is why we have this strangeness of seemingly owned code that is also listed as third_party

meisterT:
cc @cushon @comius

third_party/ijar/zip_main.cc
```diff
@@ -152,11 +152,14 @@ int extract(char *zipfile, char *exdir, char **files, bool verbose,
     return -1;
   }
 
-  char output_root[PATH_MAX];
+  char output_root[PATH_MAX + 1];
   if (exdir != NULL) {
-    concat_path(output_root, PATH_MAX, cwd.c_str(), exdir);
+    concat_path(output_root, sizeof(output_root), cwd.c_str(), exdir);
```
comius, line 157: Shouldn't we also check if cwd.length()+strlen(exdir) >= sizeof(output_root)?

comius (changes requested):

comius:
@aiuto I checked and it seems that changes to third_party/ijar are "correctly" imported internally.

third_party/ijar/zip_main.cc
comius: Or alternatively always properly truncate, which is then `strncpy(..., PATH_MAX-1)`

aiuto: Using PATH_MAX-1 is probably not right, since it is unspecified if PATH_MAX has room for the nul or not on all systems.  Increasing the buffer by 1 always leaves room.

aiuto (approved):

aiuto:
> @aiuto I checked and it seems that changes to third_party/ijar are "correctly" imported internally.
Great.  Since you see where it imports, I'll count on you to merge this.

third_party/ijar/zip_main.cc
comius: @aiuto then you're ok that the value is truncated when `exdir` is set and error returned when `cwd` is too long? but not error returned when `exdir+cwd` is too long?

benjaminp: I agree `concat_path` will not behave correctly if the paths are too long, but my goal was to fix possible memory corruption rather than that.

I will make `concat_path` return success boolean.

comius: Ok, since this is an improvement with respect to the previous code I'll approve and import it. Thanks @benjaminp.

comius (approved):

third_party/ijar/zip_main.cc
```diff
@@ -152,11 +157,16 @@ int extract(char *zipfile, char *exdir, char **files, bool verbose,
...
-    concat_path(output_root, PATH_MAX, cwd.c_str(), exdir);
+    if (!concat_path(output_root, sizeof(output_root), cwd.c_str(), exdir)) {
+      return -1;
+    }
+  } else if (cwd.length() >= sizeof(output_root)) {
+    fprintf(stderr, "current working directory path too long");
+    return -1;
   } else {
-    strncpy(output_root, cwd.c_str(), PATH_MAX);
+    strcpy(output_root, cwd.c_str());
```
comius, line 169: Linter is not happy with `strcpy`, can you put back `strncpy`?

benjaminp: Hum, is this linter why there are so many `strncpy` calls with off-by-one errors in this file? :)

third_party/ijar/zip_main.cc
aiuto: Yes. I was OK with the truncation because fixing that in a useful way (reporting the error and having bazel act on it) is a much larger PR.

Issue #12296, opened 2020-10-18
Canonicalize flags after reading configuration files? (to enable composing tag filters)


EricCousineau-TRI:
### Description of the problem / feature request:

I'd like `canonicalize-flags` (or some related command) to output canonical flags *after* parsing RC files.
At present, they do not seem to:

```
  % bazel canonicalize-flags -- --config=any_name --test_tag_filters="-lint"
  --config=any_name
  --test_tag_filters=-lint
```

### Feature requests: what underlying problem are you trying to solve with this feature?

I'd like to compose tag filters (w/ custom logic), and would love to not create add'l `--config` entries in an RC file.

Ultimately, this is poor-man's workaround to #6477 - we've tinkered w/ Bazel's scheduling on a single machine w/ multiple GPUs, and haven't found a Bazel-expressible way to prevent bottlenecks unless we explicitly separate our `bazel test` statements.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

N/A

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

```
$ bazel info release 
INFO: Invocation ID: 77860539-db2e-43c1-9916-b3f9271a31da
release 3.5.0
```

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

Closest thing was #10213, but that looked like a bug.

### Any other information, logs, or outputs that you want to share?

N/A

EricCousineau-TRI:
FTR The workaround for this is to just re-parse the `.rc` files, or use a dummy test target and use `--announce_rc`, and just intercept the output and hope for stability in this formatting.

juliexxia:
Agree that this would be a nice feature to have and thanks for the docs updates. A potential complication would be breaking anyone that does expect --configs to stay unexpanded. 

We're not prioritizing this as a feature request right now, but I'll leave open in case we prioritize in the future/so others can chime in as well.

EricCousineau-TRI:
Sounds great, thanks!

Perhaps that kind of thing could be made explicit, with a direct flag, like
`bazel canonicalize-flags --expand_config -- --config blerg`
or something indirect / more encompassing
`bazel canonicalize-flags --flatten -- --config blerg`

PR #12297, opened 2020-10-18
DOCS: Add example for canonicalize-flags

EricCousineau-TRI:
Would've loved to have seen this while investigating #12296

### DIFF
--- a/site/docs/user-manual.html
+++ b/site/docs/user-manual.html
@@ -2667,7 +2667,7 @@ adb -s deadbeef install ...
 <h3 id='canonicalize'><code>canonicalize-flags</code></h3>
 
 <p>
-  The <code>canonicalize-flags</code> command, which takes a list of options for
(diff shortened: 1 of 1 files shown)


site/docs/user-manual.html
```diff
@@ -2681,6 +2681,15 @@ <h3 id='canonicalize'><code>canonicalize-flags</code></h3>
   Note that a small number of options cannot be reordered, because Bazel cannot
   ensure that the effect is identical.
 </p>
+<p>
+  As an example:
+</p>
+<pre>
+  % bazel canonicalize-flags -- --config=any_name --test_tag_filters="-lint"
```
jin, line 2688: It's worth explicitly calling out that the command doesn't expand `config` flags, even though you have an example here already.

EricCousineau-TRI: Done. Thanks!

jin (approved):

jin:
Thanks, importing.

PR #12300, opened 2020-10-19
Remove bazel_java_test_local_java_tools_jdkXX test from upload_all_java_tools.sh…

comius:
The tests don't pass on Windows, because of an outdated CI setup, see https://github.com/bazelbuild/bazel/issues/12244, because of Windows path limit. Removing the tests to be able make a release. Similar tests are already executed on the main branch.

### DIFF
--- a/src/upload_all_java_tools.sh
+++ b/src/upload_all_java_tools.sh
@@ -61,6 +61,14 @@ for java_version in 11; do
         file_url="file://${zip_path}"
     fi
 
+    # Skip for now, as the test is broken on Windows.
+    # See https://github.com/bazelbuild/bazel/issues/12244 for details
+    if not "$is_windows"; then
+        bazel test --verbose_failures --test_output=all --nocache_test_results \
+            //src/test/shell/bazel:bazel_java_test_local_java_tools_jdk${java_version} \
+            --define=LOCAL_JAVA_TOOLS_ZIP_URL="${file_url}"
+    fi
+
     bazel run //src:upload_java_tools_java${java_version} -- \
         --java_tools_zip src/java_tools_java${java_version}.zip \
         --commit_hash ${commit_hash} \
--- a/src/upload_all_java_tools.sh
+++ b/src/upload_all_java_tools.sh
@@ -60,9 +60,6 @@ for java_version in 11; do
         # Non-Windows needs "file:///foo/bar".
         file_url="file://${zip_path}"
     fi
-    bazel test --verbose_failures --test_output=all --nocache_test_results \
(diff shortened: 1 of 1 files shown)


src/upload_all_java_tools.sh
```diff
@@ -60,9 +60,6 @@ for java_version in 11; do
         # Non-Windows needs "file:///foo/bar".
         file_url="file://${zip_path}"
     fi
-    bazel test --verbose_failures --test_output=all --nocache_test_results \
-      //src/test/shell/bazel:bazel_java_test_local_java_tools_jdk${java_version} \
-      --define=LOCAL_JAVA_TOOLS_ZIP_URL="${file_url}"
```
davido, line 65: Instead of removing this test, may be comment it out, in case @meteorcloudy will find a way to fix too long command line on Windows?

davido (approved):

davido:
Thanks, @comius for your efforts to release `java_tools` ASAP with new and shiny Java 15 support.

src/upload_all_java_tools.sh
meteorcloudy: That's a good idea. You probably want to also enable it while doing the manual release?

davido: {quote}
You probably want to also enable it while doing the manual release?
{quote}

Right. This test is passing in sane platforms: Linux and Mac Os X. And shoud be executed there and only skipped on Windows, so may be even:

```bash
# Skip for now, as the test is broken on Windows.
# See https://github.com/bazelbuild/bazel/issues/12244 for details
if not "$is_windows"; then
    bazel test --verbose_failures --test_output=all --nocache_test_results \
          //src/test/shell/bazel:bazel_java_test_local_java_tools_jdk${java_version} \
          --define=LOCAL_JAVA_TOOLS_ZIP_URL="${file_url}"
fi
```

comius: Following @davido suggestion, so we don't have commented out code.

I don't have a setup to do manual releases. Also there is a comment in the script advising against doing them.

davido (approved):

katre (approved):

meteorcloudy:
OK, let's see if https://github.com/bazelbuild/bazel/pull/12301 could work around this issue for Windows, testing at https://buildkite.com/bazel-trusted/java-tools-binaries-java/builds/148#2578de91-7d4b-4f75-b992-c1c1db833114

meteorcloudy:
#12301 doesn't seem to work, I have to somehow shorten the output root for the inner Bazel in the test. Let's go with this PR first!

meteorcloudy (approved):

Issue #7415, opened 2019-02-13
MacOS: cc_library: Patch "library name" to mangled library name


aherrmann:
### Feature request:

On MacOS, when building a dynamic library from a `cc_library` target, Bazel should modify the "library name" of the resulting dynamic library to the mangled name relative to `@rpath`. E.g.

```
@rpath/libtests_Sindirect-link_Slibcbits.so
```

instead of the current

```
bazel-out/darwin-fastbuild/bin/tests/indirect-link/libcbits.so
```

### Feature requests: what underlying problem are you trying to solve with this feature?

When building dynamic `cc_library` targets with Bazel on MacOS, then Bazel distinguishes between the unmangled library (e.g. `bazel-out/darwin-fastbuild/bin/tests/indirect-link/libcbits.so`), and the mangled library (e.g. `bazel-out/darwin-fastbuild/bin/_solib_darwin/libtests_Sindirect-link_Slibcbits.so`), which is a symbolic link to the former.

The "library name" of the library refers to the unmangled library:
```
$ otool -D bazel-out/darwin-fastbuild/bin/_solib_darwin/libtests_Sindirect-link_Slibcbits.so
bazel-out/darwin-fastbuild/bin/_solib_darwin/libtests_Sindirect-link_Slibcbits.so:
bazel-out/darwin-fastbuild/bin/tests/indirect-link/libcbits.so
```

If another target links against this library dynamically, then the linker will refer to that "library name" in the load command of the new binary. E.g.
```
          cmd LC_LOAD_DYLIB
         name bazel-out/darwin-fastbuild/bin/tests/indirect-link/libcbits.so ...
```

But, it should refer to the mangled library instead, because at runtime only the mangled library will be available. Ideally, using an `@rpath` relative path. I.e. it should be
```
          cmd LC_LOAD_DYLIB
         name @rpath/libtests_Sindirect-link_Slibcbits.so ...
```

[...]

PR #12304, opened 2020-10-19
Add -install_name when linking shared libraries on macOS

Yannic:
This change exposes the name of the solib symlink available at runtime
to link actions and uses that name to set the `-install_name` of shared
libraries on macOS. Exposing the runtime solib name to link actions
allows us to remove the need for the cc_wrapper script that patched the
binary on macOS (which relied on the fact that the runtime solibs are
symlinks locally, which is not true in all cases).

Updates #7415


keith (approved):
this is awesome, removing these wrappers is a huge win!

Yannic:
Downstream: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1712

Most projects are passing, except for `rules_go`. It seems like `cgo` relies on the default value for `-install_name` (which seems to be that the value of `-o` is used), so we may have to roll this out as incompatible change.

Yannic:
@oquenchil Could we please get an early (high-level) review on this?

oquenchil:
This looks like a very good clean up. Thank you very much. If rules_go indicates that it can break people, then I would say put it behind a flag.

Apart from that the code looks good to me. I'm a bit hesitant about the name runtime_solib_name, does the way the build variable is used make sense only for macOS or you see this potentially being used for unix too? If not, perhaps the link build variable name should indicate that with an osx_ prefix. Up to you.

As a more general thought but nothing for you to take care of, I wish we could come up with a more flexible way of adding build variables that doesn't involve hardcoding them in the Java code.

tools/cpp/osx_cc_wrapper.sh.tpl
```diff
@@ -13,94 +13,13 @@
...
-# the binary. It parses the command line to behave as rpath is supposed
-# to work.
-#
-# See https://blogs.oracle.com/dipol/entry/dynamic_libraries_rpath_and_mac
-# on how to set those paths for Mach-O binaries.
-#
-set -eu
 
-INSTALL_NAME_TOOL="/usr/bin/install_name_tool"
+# TODO(yannic): Remove this wrapper.
```
keith, line 17: FWIW when we remove these I think we should point these actions at wrapped_clang https://github.com/bazelbuild/bazel/pull/12356

Yannic: +1

Yannic:
@oquenchil PTAL, thanks!

> This looks like a very good clean up. Thank you very much. If rules_go indicates that it can break people, then I would say put it behind a flag.

Done

> Apart from that the code looks good to me. I'm a bit hesitant about the name runtime_solib_name, does the way the build variable is used make sense only for macOS or you see this potentially being used for unix too? If not, perhaps the link build variable name should indicate that with an osx_ prefix. Up to you.

The runtime solibs are not specific to macOS, so I don't think the variable should have an osx_ prefix. I can imagine the value being used on Unix/Windows as well.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -3842,108 +3842,153 @@ def _impl(ctx):
...
+                actions = [
+                    ACTION_NAMES.cpp_link_executable,
+                    "objc-executable",
+                    "objc++-executable",
+                ],
+                flag_groups = [flag_group(flags = ["-undefined", "dynamic_lookup"])],
+                with_features = [with_feature_set(features = ["dynamic_linking_mode"])],
+            ),
+        ]
+        if ctx.fragments.cpp.do_not_use_macos_set_install_name:
```
keith, line 3872: instead of these conditions could we only inject the `runtime_solib_name` variable with the feature is enabled?

Yannic:
@oquenchil Any chance you can take a look and we get this into 4.0?

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -6228,6 +6250,7 @@ def _impl(ctx):
             supports_dynamic_linker_feature,
             objcopy_embed_flags_feature,
             dynamic_linking_mode_feature,
+            set_install_name,
```
keith, line 6253: shouldn't this feature apply to dylibs built for iOS as well?

Yannic: Probably we should, done. PTAL

oquenchil (approved):
LGTM

oquenchil:
I'm seeing this test failure:
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/1b58eec1-10d5-4843-8af4-324e41f79556/src/test/shell/bazel/cpp_darwin_integration_test/attempt_1.log

Yannic:
Looks like this broke because of https://github.com/bazelbuild/bazel/pull/12265. Fixed by using `-Wl,-install_name,@rpath/...`-

keith:
Are we going to have to wait until 5.0 to flip this and cleanup those wrappers?

oquenchil:
For flipping and cleaning up those wrappers you can do so right after 4.0 is cut (very soon, not sure if today or in 1 day). The next minor releases will have 4.0 as baseline so the behavior won't be there by default without using the flag.

I was told you can even flip before 4.0, but since this is an incompatible change, it might be too rushed, getting everything in downstream to pass etc,... If you are concerned about is not having to use the flag, then you can try that. If you are just concerned about cleaning up the code you will be able to do that soon.

PR #12305, opened 2020-10-19
Add package_metrics.proto to bootstrap protos list

divanorama:
Without it non-dist non-bazel bootstrapping fails with messages like
following:
```
src/main/java/com/google/devtools/build/lib/packages/metrics/CompletePackageMetricsRecorder.java:30: error: cannot find symbol
  private final HashMap<PackageIdentifier, PackageMetrics> metrics = new HashMap<>();
                                           ^
  symbol:   class PackageMetrics
  location: class CompletePackageMetricsRecorder
```

Another option could be to move it under src/main/protobuf.
Note that there are more protos currently under src/main/java, but
it seems like not all of them are needed for bootstrap compilation.
$ find src/main/java/ -name "*.proto"
src/main/java/com/google/devtools/build/skyframe/graph_inconsistency.proto
src/main/java/com/google/devtools/build/skydoc/rendering/proto/stardoc_output.proto
src/main/java/com/google/devtools/build/lib/skyframe/proto/action_rewind_event.proto
src/main/java/com/google/devtools/build/lib/starlarkdebug/proto/starlark_debugging.proto
src/main/java/com/google/devtools/build/lib/packages/metrics/package_metrics.proto
src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
src/main/java/com/google/devtools/build/lib/bazel/debug/workspace_log.proto

### DIFF
--- a/scripts/bootstrap/compile.sh
+++ b/scripts/bootstrap/compile.sh
@@ -25,7 +25,7 @@ fi
 
 # Parse third_party/googleapis/BUILD.bazel to find the proto files we need to compile from googleapis
 GOOGLE_API_PROTOS="$(grep -o '".*\.proto"' third_party/googleapis/BUILD.bazel | sed 's/"//g' | sed 's|^|third_party/googleapis/|g')"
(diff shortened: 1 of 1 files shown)


lberki:
Thanks for the change! I'll merge this.

lberki (approved):

PR #12309, opened 2020-10-19
Improve cquery speed by caching Skyframe calls

gregestren:
Calling Skyframe adds a performance cost even when hitting the Skyframe
cache. This change adds a local cache for retrieving configurations.

Experiments show about a 7% reduction in query time on `--output=graph` (see code comments).

Followup to
https://github.com/bazelbuild/bazel/pull/12248/files/939e4cadf7a49a316017d207bd4504ac0c10373c#r503552367.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/CqueryThreadsafeCallback.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/CqueryThreadsafeCallback.java
@@ -20,6 +20,7 @@ import com.google.devtools.build.lib.analysis.config.BuildConfiguration;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
 import com.google.devtools.build.lib.query2.NamedThreadSafeOutputFormatterCallback;
 import com.google.devtools.build.lib.query2.engine.QueryEnvironment.TargetAccessor;
+import com.google.devtools.build.lib.skyframe.BuildConfigurationValue;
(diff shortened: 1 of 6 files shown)


juliexxia (approved):

src/main/java/com/google/devtools/build/lib/query2/cquery/CqueryThreadsafeCallback.java
```diff
@@ -88,6 +94,14 @@ public void close(boolean failFast) throws InterruptedException, IOException {
     }
   }
 
+  protected BuildConfiguration getConfiguration(BuildConfigurationValue.Key configKey) {
+    // Experiments querying:
```
haxorz, line 98: Use ConcurrentHashMap?

This class is named CqueryThreadsafeCallback and it extends ThreadSafeOutputFormatterCallback, so it's best to either (A) make this method thread safe (even if it's the case that cquery output formatting *currently* occurs on a single thread) or (B) make the property in the previous parenthetical explicit.

haxorz (changes requested):
Cool! I'm happy the suggestion helped!

src/main/java/com/google/devtools/build/lib/query2/cquery/CqueryThreadsafeCallback.java
gregestren: Good point, thanks.

Issue #12303, opened 2020-10-19
IgnoredPackagePrefixesFunction loses errors


katre:
Here's a short reproduction:
```
$ mkdir repo
$ touch repo/WORKSPACE
$ echo 'toolchain_type(name = "toolchain_type")' > repo/BUILD.bazel
$ mkdir main
$ echo 'local_repository(name = "repo", path = "../repo")' >> main/WORKSPACE
$ touch main/BUILD.bazel
$ cd main/
$ bazel cquery --output=starlark --starlark:expr 'target' @repo//:toolchain_type
...
FAILED: Build did NOT complete successfully
Internal error thrown during build. Printing stack trace: java.lang.NullPointerException
	at com.google.devtools.build.lib.query2.PostAnalysisQueryEnvironment.getIgnoredPackagePrefixesPathFragments(PostAnalysisQueryEnvironment.java:235)
	at com.google.devtools.build.lib.query2.cquery.ConfiguredTargetQueryEnvironment.getTargetsMatchingPattern(ConfiguredTargetQueryEnvironment.java:293)
	at com.google.devtools.build.lib.query2.engine.TargetLiteral.eval(TargetLiteral.java:73)
	at com.google.devtools.build.lib.query2.engine.AbstractQueryEnvironment.eval(AbstractQueryEnvironment.java:129)
	at com.google.devtools.build.lib.query2.common.AbstractBlazeQueryEnvironment.evalTopLevelInternal(AbstractBlazeQueryEnvironment.java:113)
	at com.google.devtools.build.lib.query2.common.AbstractBlazeQueryEnvironment.evaluateQuery(AbstractBlazeQueryEnvironment.java:154)
...
```

[...]

PR #12315, opened 2020-10-20
Improve error reporting from repository functions

katre:
Use more specific exception types, and don't blindly rethrow other
SkyFunctions' exception types.

Part of #12303.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
@@ -187,7 +187,7 @@ public final class RepositoryDelegatorFunction implements SkyFunction {
 
   @Override
   public SkyValue compute(SkyKey skyKey, Environment env)
-      throws SkyFunctionException, InterruptedException {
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
```diff
@@ -207,12 +207,12 @@ public SkyValue compute(SkyKey skyKey, Environment env)
     Rule rule;
     try {
       rule = getRepository(repositoryName, env);
-    } catch (ExternalRuleNotFoundException e) {
+      if (rule == null) {
```
lberki, line 210: nit: why move this check here?

lberki (approved):

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
katre: It felt better to me to have it closer to the actual call, than outside the try/catch block. It's not a huge change.

PR #12317, opened 2020-10-20
Improve error handling within IgnoredPackagePrefixesFunction

katre:
Specifically added checks for exceptions thrown by requesting RepositoryDirectoryValue, and rethrowing them correctly.

This does not update _callers_ of IgnoredPackagePrefixesFunction to properly handle errors.

Part of fixing #12303.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
@@ -21,6 +21,8 @@ import com.google.devtools.build.lib.actions.InconsistentFilesystemException;
 import com.google.devtools.build.lib.cmdline.RepositoryName;
 import com.google.devtools.build.lib.pkgcache.PathPackageLocator;
 import com.google.devtools.build.lib.rules.repository.RepositoryDirectoryValue;
+import com.google.devtools.build.lib.rules.repository.RepositoryFunction;
+import com.google.devtools.build.lib.rules.repository.RepositoryFunction.RepositoryFunctionException;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.build.lib.vfs.Root;
 import com.google.devtools.build.lib.vfs.RootedPath;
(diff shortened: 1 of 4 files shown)


meteorcloudy (approved):
Looks good. Can you add a brief description about how the error handle was improved?

joeleba:
@katre As promised, I'll look into this issue first thing tomorrow!

src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
```diff
@@ -91,10 +93,19 @@ public SkyValue compute(SkyKey key, Environment env)
...
-        RepositoryDirectoryValue repositoryValue =
-            (RepositoryDirectoryValue) env.getValue(RepositoryDirectoryValue.key(repositoryName));
-        if (repositoryValue == null) {
-          return null;
+        RepositoryDirectoryValue repositoryValue;
+        try {
+          repositoryValue =
+              (RepositoryDirectoryValue) env
+                  .getValueOrThrow(RepositoryDirectoryValue.key(repositoryName),
+                      RepositoryFunctionException.class, IOException.class);
```
joeleba, line 101: I don't think this case can happen? When we evaluate a SKyKey with `getValueOrThrow`, the thrown exception is `ErrorInfo.getCause()` [1], which is basically unwrapping the `SkyFunctionException` to get the root cause exception [2]. So what you're looking for here should be one of the wrapped exceptions that could be thrown when evaluating the skykey with `RepositoryFunction` [3].

[1] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/skyframe/SkyFunctionEnvironment.java;l=603;drc=b26a1a5be07be1f2200daed8bba94ac955518c6a 
[2] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/skyframe/ErrorInfo.java;l=39;drc=a5a4f47295ccb6cd81cf8732772c849b1a8b9d58
[3] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java;l=98-119;drc=0617f2c7b3f442a00f7359d216b8fab14d2dd762

joeleba (changes requested):

src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
katre: Hmm, I think you're right, I'll update it on the internal change (since I've already imported it).

katre:
Dropping this PR in favor of @joeleba's changes.

PR #12319, opened 2020-10-20
Add support for AArch64 release artifacts

koleini:
kokoro scripts are updated to support:

1. bazelisk v1.7.2
2. AArch64 release artifacts

### DIFF
--- a/scripts/kokoro/release/build_linux.sh
+++ b/scripts/kokoro/release/build_linux.sh
@@ -18,11 +18,18 @@ set -e
 set -x
 
 RELEASE_NAME=${RELEASE_NAME:-unknown}
+ARCHITECTURE=`uname -m`
+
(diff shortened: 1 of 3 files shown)


koleini:
@googlebot I signed it!

philwo:
Hi @koleini,

I'm very sorry that this wasn't merged yet. I must have overlooked it during multiple passes of reviewing my backlog :(

We can probably merge this, but I wonder - is there anything in particular you think would be achieved by this PR? The scripts that you patch are currently unused by our release infrastructure. They were created to prototype building on a new release infra, but we haven't gotten around to do the switch. So, while we can certainly merge this, it would not change anything visibly yet.

Would that be OK for you (maybe you're using these scripts for something internally)? Or do you need something else, too?

Cheers,
Philipp

philwo (approved):

PR #12323, opened 2020-10-21
Update guide.md

bkmgit:
Fix typo from .html to .md to generate correct link. It is also possible to use the link
https://docs.bazel.build/versions/master/skylark/performance.html#performance-profiling

### DIFF
--- a/site/docs/guide.md
+++ b/site/docs/guide.md
@@ -1075,5 +1075,5 @@ these messages.
 
 ## Troubleshooting performance by profiling
 
-See the [Performance Profiling](performance.html#performance-profiling) section.
+See the [Performance Profiling](performance.md#performance-profiling) section.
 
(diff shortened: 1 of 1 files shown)


gregestren:
This is a really naive question, but why does changing to `.md` work if the linked page is under `/skylark/` and the original isn't?

bkmgit:
Changing to .md goes to the markdown page in GitHub. Using a full html 
path goes to the page on docs.bazel.build  - not clear to me what the 
original intention for the link was.

It would be nice if html links are autogenerated during the build process.

gregestren:
I still get a 404 if I click on the changed file (in GitHub). For example, here from your PR: https://github.com/bazelbuild/bazel/blob/cf4706d5830cd998f3fe4e493fdf8e200cf3f7a9/site/docs/guide.md

If I change it from `site/docs/performance.md` to `site/docs/skylark/performance.md` it links.

You're seeing a successful link from your side?

bkmgit:
@gregestren Thanks for the fix, it does need skylark in the path.

gregestren:
That makes sense, thanks!

gregestren (approved):

PR #12324, opened 2020-10-21
Reduce platform dependant selects in //tools/jdk/BUILD

comius:
Macro `default_java_toolchain` used to create a `java_toolchain` rule with each attribute set using a platform dependent select. For example `ijar` attribute was set to `//tools/jdk:ijar`, which was a macro `remote_java_tools_filegroup`. That macro contains a `select` choosing the right `@remote_java_tools` repository.

With Java toolchainisation this become a fragile mechanism (selects need to match the toolchain...), which would be hard to maintain.

In previous commits `java_toolchain_default` was implemented, which is part of each `java_tools` repository and does not use any `selects`.

For backward compatibility `default_java_toolchain` was reimplemented to invoke `java_toolchain_default`. Implementation is not perfect, because it load 3 java_tools repositories (linux,darwin,windows). Not to invoke this misfeature, some macros were moved into `utils.bzl`. The use of `default_java_toolchain` should be discouraged.

`remote_java_tools_filegroup/import` targets in `//tools/jdk/BUILD` that were previous used only by `default_java_toolchain` were removed.

`default_java_toolchain` targets in `//tools/jdk/BUILD` were replaced with select aliases. Those will be removed when Java toolchainsation is complete.

`legacy_toolchain` was removed as a fall-back - it was an awkard one - going from Java 11 to Java 8 language level.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/BazelAndroidLocalTestRule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/BazelAndroidLocalTestRule.java
@@ -42,7 +42,7 @@ import com.google.devtools.build.lib.rules.java.JavaSemantics;
 /** Rule definition for Bazel android_local_test */
 public class BazelAndroidLocalTestRule implements RuleDefinition {
 
-  protected static final String JUNIT_TESTRUNNER = "//tools/jdk:TestRunner_deploy.jar";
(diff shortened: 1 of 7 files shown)


comius:
Downstream tests at https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1774

PR #12325, opened 2020-10-21
Allow Jacoco Coverage to work with classpath jar.

### LINKED ISSUE #11847: Coverage doesn't work with long classpaths

If the classpath for a test target exceeds the limit (specified at https://github.com/bazelbuild/bazel/blob/fd5d29aa4ef907a15ec6db0e359e86df1acf64fc/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L358) the classpath is written to separate file, and coverage output is empty.

Clone bazel repo and run
```
$ bazel coverage //examples/java-native/src/test/java/com/example/myproject:custom
...
//examples/java-native/src/test/java/com/example/myproject:custom        PASSED in 0.7s
  /home/ubuntu/.cache/bazel/_bazel_ubuntu/5bb6c1fa5da68b7db1883687715859ce/execroot/io_bazel/bazel-out/k8-fastbuild/testlogs/examples/java-native/src/test/java/com/example/myproject/custom/coverage.dat

$ bazel coverage --action_env=CLASSPATH_LIMIT=100 //examples/java-native/src/test/java/com/example/myproject:custom
...
//examples/java-native/src/test/java/com/example/myproject:custom        PASSED in 0.6s
# No .dat file mentioned
$ ls -lh /home/ubuntu/.cache/bazel/_bazel_ubuntu/5bb6c1fa5da68b7db1883687715859ce/execroot/io_bazel/bazel-out/k8-fastbuild/testlogs/examples/java-native/src/test/java/com/example/myproject/custom/coverage.dat
-r-xr-xr-x 1 ubuntu ubuntu 0 Jul 27 16:27 /home/ubuntu/.cache/bazel/_bazel_ubuntu/5bb6c1fa5da68b7db1883687715859ce/execroot/io_bazel/bazel-out/k8-fastbuild/testlogs/examples/java-native/src/test/java/com/example/myproject/custom/coverage.dat
```

Linux

release 3.4.1

No :(

Workaround/Pointer to fix would be appreciated. I briefly looked at JacocoCoverageRunner.java but don't have enough context with this code to make progress quickly.

c-mita:
If the classpath for a java binary exceeds the limits specified, a
separate jar is created at runtime to contain the classpath and is the
target of execution. The JacocoRunner could not handle this case, only
inspecting the wrapping jar.

This change adds code to extract the objects in the original classpath
and pass them to the original handling code.

Fixes #11847

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
@@ -349,21 +349,20 @@ public class JacocoCoverageRunner {
   private static URL[] getUrls(ClassLoader classLoader, boolean wasWrappedJar) {
     URL[] urls = getClassLoaderUrls(classLoader);
     // If the classpath was too long then a temporary top-level jar is created containing nothing but a manifest with
-    // the original classpath. For the moment, we just assume it's made up of URLs to other jars and extract them.
-    if (urls != null) {
-      if (wasWrappedJar && urls.length == 1) {
-        try {
-          String jarClassPath =
-              new JarInputStream(urls[0].openStream()).getManifest().getMainAttributes().getValue("Class-Path");
-          String[] urlStrings = jarClassPath.split(" ");
-          URL[] newUrls = new URL[urlStrings.length];
-          for (int i = 0; i < urlStrings.length; i++) {
-            newUrls[i] = new URL(urlStrings[i]);
-          }
-          return newUrls;
-        } catch (Exception e) {
-         return null;
+    // the original classpath. Those are the URLs we are looking for.
+    if (wasWrappedJar && urls != null && urls.length == 1) {
+      try {
+        String jarClassPath =
+            new JarInputStream(urls[0].openStream()).getManifest().getMainAttributes().getValue("Class-Path");
+        String[] urlStrings = jarClassPath.split(" ");
+        URL[] newUrls = new URL[urlStrings.length];
+        for (int i = 0; i < urlStrings.length; i++) {
+          newUrls[i] = new URL(urlStrings[i]);
         }
+        return newUrls;
+      } catch (Exception e) {
+        e.printStackTrace();
+        return null;
       }
     }
     return urls;
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
@@ -46,6 +46,7 @@ import java.util.TreeMap;
 import java.util.jar.Attributes;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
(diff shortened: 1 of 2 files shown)


src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
```diff
@@ -345,7 +346,30 @@ private static String getUniquePath(String pathTemplate, String suffix) throws I
     return convertedMetadataFiles.build();
   }
 
-  private static URL[] getUrls(ClassLoader classLoader) {
+  private static URL[] getUrls(ClassLoader classLoader, boolean wasWrappedJar) {
+    URL[] urls = getClassLoaderUrls(classLoader);
+    // If the classpath was too long then a temporary top-level jar is created containing nothing but a manifest with
+    // the original classpath. For the moment, we just assume it's made up of URLs to other jars and extract them.
+    if (urls != null) {
+      if (wasWrappedJar && urls.length == 1) {
```
comius, line 354: nit: Can you merge those two together: `if (wasWrappedJar && urls != null && urls.length ==1)`. This will reduce for one block ident and have urls parameters together, which is easier to read.

src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
```diff
@@ -345,7 +346,30 @@ private static String getUniquePath(String pathTemplate, String suffix) throws I
...
+          String jarClassPath =
+              new JarInputStream(urls[0].openStream()).getManifest().getMainAttributes().getValue("Class-Path");
+          String[] urlStrings = jarClassPath.split(" ");
+          URL[] newUrls = new URL[urlStrings.length];
+          for (int i = 0; i < urlStrings.length; i++) {
+            newUrls[i] = new URL(urlStrings[i]);
+          }
+          return newUrls;
+        } catch (Exception e) {
+         return null;
```
comius, line 365: There's not much error handling in the surrounding code, but there is a printStackTrace, which you might want to have.

comius (commented):
It took me a while to figure out if there is something better that you could do than this. Short answer is: no, I don't think so.

So if classpath is too long, you can't use a variable or a parameter. You need a file. You could write a custom file, but you already need .jar file with a metafile, because it also has to be loaded. 

You could write a second file with classpath and parse it - that would replace both getUrls and getClassLoaderUrls with one parsing function. But then you would have two files and just another custom parsing function.

A small simplification would be to just always generate .jar even if classpath is short. But this might cause a regression on small packages.

src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
c-mita: The error handling in this entire region can be described succinctly with the statement "Hope nothing goes wrong".

comius (approved):

PR #12328, opened 2020-10-21
Add libraries to allowed macOS include directories

keith:
This adds macOS `/Library` and `/Users/$USER/Library` to allowed include
paths. In macOS workflows most system headers are included in the Xcode
toolchains, which were already covered in this logic. The only exception
is custom `xctoolchains`, which can be used for custom Swift compilers
(among other things). These toolchains are either installed globally in
`/Library`, or without root they can be installed in the user's
`~/Library`. This adds those paths since they can safely be referenced
by passing `--action_env=TOOLCHAINS=toolchain_id` which will be picked
up by invocations of `xcrun`.

### DIFF
--- a/tools/cpp/cc_configure.bzl
+++ b/tools/cpp/cc_configure.bzl
@@ -169,6 +169,7 @@ cc_autoconf = repository_rule(
         "GCOV",
         "HOMEBREW_RUBY_PATH",
         "SYSTEMROOT",
+        "USER",
     ] + MSVC_ENVVARS,
     implementation = cc_autoconf_impl,
     configure = True,
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -38,12 +38,20 @@ def _get_escaped_xcode_cxx_inc_directories(repository_ctx, cc, xcode_toolchains)
       include_paths: A list of builtin include paths.
     """
 
-    include_dirs = []
+    # Assume that everything is managed by Xcode / toolchain installations
+    include_dirs = [
+        "/Applications/",
+        "/Library/",
+    ]
+
+    user = repository_ctx.os.environ.get("USER")
+    if user:
+        include_dirs.append("/Users/{}/Library/".format(user))
+
+    # Include extra Xcode paths in case they're installed on other volumes
     for toolchain in xcode_toolchains:
         include_dirs.append(escape_string(toolchain.developer_dir))
 
-    # Assume that all paths that point to /Applications/ are built in include paths
-    include_dirs.append("/Applications/")
     return include_dirs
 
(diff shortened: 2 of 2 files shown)


jin:
CI failures look related. Could you take a look?

```
ERROR: /private/var/tmp/_bazel_buildkite/06004920da606e381d79e0e126a5665b/sandbox/darwin-sandbox/3716/execroot/io_bazel/_tmp/3a63f12eb3b00bf266b2cf45eae7c788/workspace/BUILD:7:10: Compiling bad.cc failed: (Exit 1): wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' -iquote . -iquote ... (remaining 28 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' -iquote . -iquote ... (remaining 28 argument(s) skipped)

bad.cc:2:10: fatal error: 'foo/v1/foo.h' file not found
#include "foo/v1/foo.h"
```

keith:
Ok I think `/` was too aggressive. I've updated this change to include only `Library` directories instead.

tools/cpp/cc_configure.bzl
```diff
@@ -169,6 +169,7 @@ cc_autoconf = repository_rule(
         "GCOV",
         "HOMEBREW_RUBY_PATH",
         "SYSTEMROOT",
+        "USER",
```
keith, line 172: I'm not sure `USER` has to be here, as far as I could tell changing `USER` would force this even without adding it here, so maybe it supersedes this list, but it seems like semantically this is still the right thing to do.

gregestren:
@keith I got the message you have a few PRs that need routing and attention. I'm on triage rotation now. I'll categorize / route them soon (although it's running late today so I'm going to aim for Monday if that's okay).

keith:
No problem, thanks!!

gregestren:
Okay, I've assigned your PRs to @aiuto, who should be a good contact for further handling.

Feel free to ping me if you need more updates.

aiuto:
Actually, I'm not a great contact on them. So many touch the edges of C++.
I'll reroute.

tools/cpp/cc_configure.bzl
oquenchil: Let's still add it then.

oquenchil (approved):
LGTM

PR #12337, opened 2020-10-23
Allow remapping of __BAZEL_EXECUTION_ROOT__ to execution root in wrapped_clang

thii:
Some clang flags requires the file path passed to it to be an absolute
path (for instance, the `-fuse-ld` flag).

Currently, in order to use those flags, we would need to either:
- Expand the location of the file before it is passed to wrapped_clang,
  which breaks reproducibility, since the absolute path would go into a
  cache key.
- Place the file at a global path (for example, `/tmp/my-linker`), and
  make that available to the build, which breaks hermeticity.

This change adds a new shared secret, `__BAZEL_EXECUTION_ROOT__`, so
that rule authors can use it as a prefix to paths that they want to be
expanded properly.

### DIFF
--- a/tools/osx/crosstool/wrapped_clang.cc
+++ b/tools/osx/crosstool/wrapped_clang.cc
@@ -223,6 +223,7 @@ int main(int argc, char *argv[]) {
     if (arg.compare("OSO_PREFIX_MAP_PWD") == 0) {
       arg = "-Wl,-oso_prefix," + std::string(cwd.get()) + "/";
     }
+    FindAndReplace("__BAZEL_EXECUTION_ROOT__", std::string(cwd.get()), &arg);
     FindAndReplace("__BAZEL_XCODE_DEVELOPER_DIR__", developer_dir, &arg);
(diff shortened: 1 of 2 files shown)


keith:
Note for `-fuse-ld` specifically because of this issue there is an upcoming `--ld-path` flag that accepts relative paths. https://reviews.llvm.org/D83015

keith:
General thought if this lands, we can rework the debug info and oso prefix flags to use this instead.

thii:
I still think landing this has the benefits of it:
- The new `--ld-path` is still not yet available in Xcode 12.2 beta 3. At the earliest, if it can make it into 12.3, that would still probably be a few months to go, while Bazel has a faster release cycle of one release per month. Also, I think many teams still on Xcode 11.x.
- This provides the flexibility in trying new flags that require remapping without having to patch wrapped_clang every time.

keith (approved):

thii:
The `-fuse-ld` case can be mitigated with clang's tools search paths so this doesn't seem to be needed anymore.

thii:
Abandoning this.

PR #12336, opened 2020-10-23
Prepare to make bazel-toolchains compatible with Incompatible Target Skipping

philsc:
The upcoming Incompatible Target Skipping patch (#10945) introduces a
new `target_compatible_with` attribute which clashes with one of the
custom attributes in the bazel-toolchains repo. This patch adds a .patch
file in `third_party` so that we can fix the clash in a followup patch.

Once the clash is fixed we can merge the Incompatible Target
Skipping patch.

This is essentially a backport of bazelbuild/bazel-toolchains#913.

PR #12341, opened 2020-10-23
Make bazel-toolchains compatible with Incompatible Target Skipping (part 2)

philsc:
This is a follow-up to #12336. It makes use of the .patch file to make
the 3.1 version of bazel-toolchains compatible with Incompatible
Target Skipping.


philsc:
@katre , here is the WORKSPACE-only patch.

katre:
Thanks, once the tests pass I'll start the internal merge process.

PR #12343, opened 2020-10-23
remote: Include file size in hash-mismatch message

Yannic:


jin:
cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/util/DigestOutputStream.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hasher;
+import java.io.FilterOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * An {@link OutputStream} that maintains a {@link Digest} of the data written to it.
+ *
+ * <p>Based on Guava's {@link com.google.common.hash.HashingOutputStream}.
```
ulfjack, line 30: s/Based on/Similar to/ and also add that it returns a Digest object that's specific to the remote execution protocol.

src/main/java/com/google/devtools/build/lib/remote/util/Utils.java
```diff
@@ -173,13 +173,13 @@ public static String grpcAwareErrorMessage(IOException e) {
...
+  public static void verifyBlobContents(Digest expected, Digest actual) throws IOException {
+    if (!expected.equals(actual)) {
       String msg =
           String.format(
-              "An output download failed, because the expected hash"
-                  + "'%s' did not match the received hash '%s'.",
-              expectedHash, actualHash);
+              "Output download failed: Expected to download a blob of %d bytes with hash '%s', "
+                  + "got %d bytes with '%s'.",
+              expected.getSizeBytes(), expected.getHash(), actual.getSizeBytes(), actual.getHash());
```
ulfjack, line 182: Personally, I'd prefer to format it like digests elsewhere as `<hash>/<size>`. WDYT?

ulfjack:
LGTM! Looking forward to seeing this merged. \o/

src/main/java/com/google/devtools/build/lib/remote/util/Utils.java
Yannic: SGTM, done

src/main/java/com/google/devtools/build/lib/remote/util/DigestOutputStream.java
Yannic: done

coeuvre (approved):
LGTM. Thanks!

PR #12345, opened 2020-10-23
Enable debug_prefix_map_pwd_is_dot feature by default on macOS

keith:
This feature improves hermeticity of binaries, and shouldn't have impact
on debugging.

RELNOTES: Enable debug_prefix_map_pwd_is_dot feature by default on macOS, this passes `-fdebug-prefix-map=$PWD=.` for every compile to remove absolute paths from debug info.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5356,6 +5356,7 @@ def _impl(ctx):
 
     debug_prefix_map_pwd_is_dot_feature = feature(
         name = "debug_prefix_map_pwd_is_dot",
+        enabled = True,
         flag_sets = [
             flag_set(
                 actions = [


jin:
Are there any docs behind this feature (`DEBUG_PREFIX_MAP_PWD`)? Could you link it in the PR description and/or RELNOTES?

jin:
cc @trybka

trybka:
Corresponds to https://cs.opensource.google/bazel/bazel/+/364a867df255c57c8edc4a8aae8f78cb54900a54:tools/osx/crosstool/wrapped_clang.cc;l=221 which maps to `-fdebug-prefix-map=`; This is late-bound because it needs the value of PWD when clang is invoked, not when the action is created.

Docs for that are https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fdebug-prefix-map

keith:
Updated the relnotes to explain it too

PR #12346, opened 2020-10-23
script: improve bazel upgrade experience

chenrui333:
After this change, the current directory would be preserved, no need to manually `cd` anymore.

Otherwise, the dir would be `~/.bazel/bin`

### DIFF
--- a/scripts/packages/bazel.sh
+++ b/scripts/packages/bazel.sh
@@ -203,11 +203,11 @@ if [[ ! -x $BAZEL_REAL ]]; then
     if [[ -x $(command -v curl) && -w $wrapper_dir ]]; then
       (echo ""
       echo "You can download the required version directly using this command:"
-      echo "  (cd \"${wrapper_dir}\" && curl -fLO https://releases.bazel.build/${bazel_version}/release/${long_binary_name} && chmod +x ${long_binary_name})") 2>&1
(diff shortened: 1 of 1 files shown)


scripts/packages/bazel.sh
```diff
@@ -203,11 +203,11 @@ if [[ ! -x $BAZEL_REAL ]]; then
     if [[ -x $(command -v curl) && -w $wrapper_dir ]]; then
       (echo ""
       echo "You can download the required version directly using this command:"
-      echo "  (cd \"${wrapper_dir}\" && curl -fLO https://releases.bazel.build/${bazel_version}/release/${long_binary_name} && chmod +x ${long_binary_name})") 2>&1
+      echo "  pushd `pwd` && (cd \"${wrapper_dir}\" && curl -fLO https://releases.bazel.build/${bazel_version}/release/${long_binary_name} && chmod +x ${long_binary_name}) && popd") 2>&1
```
meisterT, line 206: The surrounding `(` `)` should already make sure that this is executed in a subshell and you end up where you started.
Isn't that the case for you?

philwo: Indeed this was the intent behind this. Does it not work?

chenrui333: yeah, it does not work for me, I am using Mac and bash shell.

chenrui333: I did manage cd before, surprising that no one had the similar issue.

meteorcloudy:
Do we still need this?

PR #12347, opened 2020-10-23
Implement com.google.devtools.build.lib.rules.apple.DottedVersion$Option toString()

segiddins:
This way, the string value shows up when running bazel config, instead of values such as `com.google.devtools.build.lib.rules.apple.DottedVersion$Option@1703e2`


keith (approved):

lberki (approved):
Thanks! Makes Bazel a bit nicer. I'll merge this (or at least try then report back and see what breaks...)

PR #12356, opened 2020-10-27
Remove C++ linking from oso_prefix feature for now

keith:
Currently in the macOS toolchain C++ linking goes through
`cc_wrapper.sh`, not `wrapped_clang`. I didn't add the handling to
`cc_wrapper` for this argument. These wrappers will likely go away soon,
when they do we should probably make them route to wrapped_clang
instead, in which case we could revert this.

https://github.com/bazelbuild/bazel/pull/12304

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -5721,8 +5721,7 @@ def _impl(ctx):
         name = "oso_prefix_is_pwd",
         flag_sets = [
             flag_set(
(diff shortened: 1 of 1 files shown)


aiuto:
Is this blocked on #12304 or should it happen first?

keith:
Not blocked. We can potentially revert this after that lands, but it depends if that works, and since it sounds like it requires a flag it will take long enough that I think this is worth merging.

oquenchil (approved):
LGTM

PR #12360, opened 2020-10-27
java: allow deps=[] without srcs

dmivankov:
deps=[] and not setting deps should behave same.
There's no default value for deps attribute, but even if there was one emptiness check seems more correct anyway.


lberki (approved):
Thanks for the patch! I'll run this through our internal test battery and merge if all goes well (it probably should)

PR #12362, opened 2020-10-27
Report error when fetching repository in RepositoryDelegatorFunction

### LINKED ISSUE #12303: IgnoredPackagePrefixesFunction loses errors

joeleba:
Printing the error in `RepositoryDelegatorFunction` should provide enough context for the users, in case of an error.

What it looks like now:

```
[14:54:32] leba@leba:~/src/error_demo$ /tmp/bazel build --keep_going //:sample --extra_toolchains=@broken//...
ERROR: Error fetching repository: java.io.IOException: broken_repo rule //external:broken must create a directory
WARNING: errors encountered while analyzing target '//:sample': it will not be built
INFO: Analyzed target //:sample (0 packages loaded, 0 targets configured).
INFO: Found 0 targets...
ERROR: command succeeded, but not all targets were analyzed
INFO: Elapsed time: 0.168s, Critical Path: 0.01s
INFO: 1 process: 1 internal.

[15:38:18] leba@leba:~/src/error_demo$ /tmp/bazel build --nokeep_going //:sample --extra_toolchains=@broken//...
ERROR: Error fetching repository: java.io.IOException: broken_repo rule //external:broken must create a directory
ERROR: Analysis of target '//:sample' failed; build aborted: broken_repo rule //external:broken must create a directory
INFO: Elapsed time: 0.104s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

Fixes #12303

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
@@ -350,6 +350,7 @@ public final class RepositoryDelegatorFunction implements SkyFunction {
       // Upon an exceptional exit, the fetching of that repository is over as well.
       env.getListener().post(new RepositoryFetching(repositoryName, true));
       env.getListener().post(new RepositoryFailedEvent(repositoryName, e.getMessage()));
+      env.getListener().handle(Event.error("Error fetching repository: " + e.getMessage()));
       throw e;
     }
 
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
@@ -34,6 +34,7 @@ import com.google.devtools.build.lib.repository.ExternalPackageException;
 import com.google.devtools.build.lib.repository.ExternalPackageHelper;
 import com.google.devtools.build.lib.repository.ExternalRuleNotFoundException;
 import com.google.devtools.build.lib.repository.RepositoryFailedEvent;
+import com.google.devtools.build.lib.rules.repository.RepositoryFunction.AlreadyReportedRepositoryAccessException;
 import com.google.devtools.build.lib.rules.repository.RepositoryFunction.RepositoryFunctionException;
 import com.google.devtools.build.lib.skyframe.ManagedDirectoriesKnowledge;
 import com.google.devtools.build.lib.skyframe.PrecomputedValue.Precomputed;
@@ -351,7 +352,11 @@ public final class RepositoryDelegatorFunction implements SkyFunction {
       env.getListener().post(new RepositoryFetching(repositoryName, true));
       env.getListener().post(new RepositoryFailedEvent(repositoryName, e.getMessage()));
       env.getListener().handle(Event.error("Error fetching repository: " + e.getMessage()));
-      throw e;
+
+      // Rewrap the underlying exception to signal callers not to re-report this error.
+      throw new RepositoryFunctionException(
+          new AlreadyReportedRepositoryAccessException(
+              e.getCause()), e.isTransient() ? Transience.TRANSIENT : Transience.PERSISTENT);
     }
 
     if (env.valuesMissing()) {
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
@@ -14,6 +14,8 @@
 
 package com.google.devtools.build.lib.rules.repository;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
@@ -35,6 +37,7 @@ import com.google.devtools.build.lib.repository.ExternalPackageException;
 import com.google.devtools.build.lib.repository.ExternalPackageHelper;
 import com.google.devtools.build.lib.repository.ExternalRuleNotFoundException;
 import com.google.devtools.build.lib.skyframe.ActionEnvironmentFunction;
+import com.google.devtools.build.lib.skyframe.AlreadyReportedException;
 import com.google.devtools.build.lib.skyframe.PackageLookupFunction;
 import com.google.devtools.build.lib.skyframe.PackageLookupValue;
 import com.google.devtools.build.lib.skyframe.PrecomputedValue;
@@ -114,8 +117,26 @@ public abstract class RepositoryFunction {
       super(cause, transience);
     }
 
+    public RepositoryFunctionException(
+        AlreadyReportedRepositoryAccessException cause, Transience transience) {
+      super(cause, transience);
+    }
+
     public RepositoryFunctionException(ExternalPackageException e) {
       super(e.getCause(), e.isTransient() ? Transience.TRANSIENT : Transience.PERSISTENT);
+    }
+  }
+
+  /**
+   * Encapsulates the exceptions that arise when accessing a repository.
+   * Error reporting should ONLY be handled in {@link RepositoryDelegatorFunction#fetchRepository}.
+   */
+  public static class AlreadyReportedRepositoryAccessException extends AlreadyReportedException {
+    public AlreadyReportedRepositoryAccessException(Exception e) {
+      super(e.getMessage(), e.getCause());
+      checkState(e instanceof NoSuchPackageException
+          || e instanceof IOException || e instanceof EvalException
+          || e instanceof ExternalPackageException);
     }
   }
 
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
@@ -136,7 +136,7 @@ public abstract class RepositoryFunction {
       super(e.getMessage(), e.getCause());
       checkState(e instanceof NoSuchPackageException
           || e instanceof IOException || e instanceof EvalException
-          || e instanceof ExternalPackageException);
+          || e instanceof ExternalPackageException, e);
     }
   }
 
--- a/src/main/java/com/google/devtools/build/lib/skyframe/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/BUILD
@@ -1742,6 +1742,7 @@ java_library(
     name = "package_lookup_function",
     srcs = ["PackageLookupFunction.java"],
     deps = [
+        ":already_reported_exception",
         ":file_symlink_exception",
         ":ignored_package_prefixes_value",
         ":local_repository_lookup_value",
@@ -2800,4 +2801,10 @@ java_library(
         "//src/main/java/com/google/devtools/build/skyframe:skyframe-objects",
         "//third_party:jsr305",
     ],
+)
+
+java_library(
+    name = "already_reported_exception",
+    srcs = ["AlreadyReportedException.java"],
+    deps = [],
 )
--- a/src/main/java/com/google/devtools/build/lib/skyframe/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/BUILD
@@ -2806,5 +2806,4 @@ java_library(
 java_library(
     name = "already_reported_exception",
     srcs = ["AlreadyReportedException.java"],
-    deps = [],
 )
--- a/src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
@@ -330,14 +330,14 @@ public class PackageLookupFunction implements SkyFunction {
     RepositoryValue repositoryValue;
     try {
       repositoryValue = (RepositoryValue) env.getValueOrThrow(
-          repositoryKey, NoSuchPackageException.class, IOException.class, EvalException.class);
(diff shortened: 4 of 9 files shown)


src/test/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunctionTest.java
```diff
@@ -0,0 +1,55 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
```
joeleba, line 1: Credit to @katre for the majority of this file.

katre (approved):

src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
```diff
@@ -142,5 +157,23 @@ public String extractTag(SkyKey skyKey) {
...
+    }
+  }
+
+  /**
+   * Repackages the Exceptions that could arise from fetching an external repository.
+   */
+  public static class RepositoryFetchingException extends Exception {
+    public RepositoryFetchingException(Exception e) {
+      super(e.getMessage(), e.getCause() == null ? e : e.getCause());
+      Preconditions.checkArgument(
```
janakdr, line 172: nit: I think internal style now suggests statically importing checkArgument. Also, please add the exception as a debugging message.

src/main/java/com/google/devtools/build/lib/skyframe/IgnoredPackagePrefixesFunction.java
```diff
@@ -142,5 +157,23 @@ public String extractTag(SkyKey skyKey) {
...
+      super(e, Transience.TRANSIENT);
+    }
+  }
+
+  /**
+   * Repackages the Exceptions that could arise from fetching an external repository.
+   */
+  public static class RepositoryFetchingException extends Exception {
+    public RepositoryFetchingException(Exception e) {
+      super(e.getMessage(), e.getCause() == null ? e : e.getCause());
```
janakdr, line 171: using the cause seems like pretty delicate logic that could be broken by changes to RepositoryDelegatorFunction.

janakdr (commented):
It looks like essentially all callers of RepositoryDelegatorFunction are not processing the error. Would it make more sense to handle the errors directly in that function, rather than changing all the callers? I don't see that IgnoredPackagePrefixesFunction is enriching the error at all, so I don't see why it should be responsible for reporting.

joeleba:
@janakdr that makes sense. Moved the error reporting to RepositoryDelegatorFunction.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
```diff
@@ -350,6 +350,7 @@ private RepositoryFunction getHandler(Rule rule) {
       // Upon an exceptional exit, the fetching of that repository is over as well.
       env.getListener().post(new RepositoryFetching(repositoryName, true));
       env.getListener().post(new RepositoryFailedEvent(repositoryName, e.getMessage()));
+      env.getListener().handle(Event.error("Error fetching repository: " + e.getMessage()));
       throw e;
```
janakdr, line 354: Your change doesn't make the situation worse, but it's going to cause further problems that this SkyFunction is throwing random exceptions underneath RepositoryFunctionException (which is unwrapped anyway). It would make more sense to me if the underlying exception were wrapped in a RepositoryException or similar. That would let callers know that they shouldn't try to handle or show errors coming from here. As it stands, some callers might try to do something with the IOException.

joeleba: I'm generally of the opinion that low level exceptions (like IOException) should not be known to reverse deps and should instead be wrapped in some local exception. This would allow a cleaner interface between SkyFunctions.

I thought about doing that for `RepositoryFunctionException` but was discouraged by the fact that a few callers of `RepositoryDelegatorFunction` are handling `IOException`. We now have a mix of both styles: some functions like `ActionExecutionFunction` repackages `IOException` as `ActionExecutionException`, while some like `RepositoryFunctionException` basically expects callers to know all the underlying exceptions that could be thrown.

> That would let callers know that they shouldn't try to handle or show errors coming from here. As it stands, some callers might try to do something with the IOException.

Sure, as long as we document somewhere that one shouldn't catch and re-report `RepositoryException`.

janakdr: I don't see any place where RepositoryDirectoryValue#key is passed to env.getValueOrThrow. Do you just mean that those SkyFunctions separately handle IOExceptions? That makes sense: if you have to directly declare a dep on a file, you should get an IOException.

As I said before, I think the principle is that you should enrich your exception as much as you are able to. With RepositoryDelegatorFunction, it seems to be fully enriched, so we should just report and encapsulate it as a specific type.

Documenting that you shouldn't catch RepositoryException in RepositoryException's javadoc seems straightforward.

joeleba: >  but was discouraged by the fact that a few callers of RepositoryDelegatorFunction are handling IOException.

Sorry, my memory from last Friday failed me... It's not that "a few callers of RepositoryDelegatorFunction are handling IOException", but rather there are 6 implementations of [`RepositoryFunction#fetch`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java;l=132;drc=0617f2c7b3f442a00f7359d216b8fab14d2dd762) (which originates the `RepositoryFunctionException`), each throwing `RepositoryFunctionException` with a different underlying exception. 

As a result, it's hairier to encapsulate those in a `RepositoryException` before wrapping that in `RepositoryFunctionException`, but it's not infeasible. Was that what you had in mind?

joeleba: Changes:
- Wrapped the low level exceptions (IOException, ...) into an AlreadyReportedRepositoryAccessException before wrapping that in RepositoryFunctionException.
- Fixed the failing tests (mostly because of the failFastHandler).

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
```diff
@@ -350,7 +351,12 @@ private RepositoryFunction getHandler(Rule rule) {
       // Upon an exceptional exit, the fetching of that repository is over as well.
       env.getListener().post(new RepositoryFetching(repositoryName, true));
       env.getListener().post(new RepositoryFailedEvent(repositoryName, e.getMessage()));
-      throw e;
+      env.getListener().handle(Event.error("Error fetching repository: " + e.getMessage()));
+
+      // Rewrap the underlying exception to signal callers not to re-report this error.
+      throw new RepositoryFunctionException(
```
janakdr, line 357: nit: I don't see why this is reusing RepositoryFunctionException. The convention is for each SkyFunction to have its own wrapper, that can narrow the exceptions it takes. Then RepositoryFunction doesn't need to know about any of this.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
```diff
@@ -114,11 +117,29 @@ public RepositoryFunctionException(EvalException cause, Transience transience) {
...
+  /**
+   * Encapsulates the exceptions that arise when accessing a repository.
+   * Error reporting should ONLY be handled in {@link RepositoryDelegatorFunction#fetchRepository}.
+   */
+  public static class AlreadyReportedRepositoryAccessException extends AlreadyReportedException {
+    public AlreadyReportedRepositoryAccessException(Exception e) {
+      super(e.getMessage(), e.getCause());
+      checkState(e instanceof NoSuchPackageException
+          || e instanceof IOException || e instanceof EvalException
+          || e instanceof ExternalPackageException);
```
janakdr, line 139: nit: add e as a debugging parameter

src/main/java/com/google/devtools/build/lib/skyframe/BUILD
```diff
@@ -2801,3 +2802,9 @@ java_library(
         "//third_party:jsr305",
     ],
 )
+
+java_library(
+    name = "already_reported_exception",
+    srcs = ["AlreadyReportedException.java"],
+    deps = [],
```
janakdr, line 2809: nit: can we just leave out an empty deps attribute?

src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
```diff
@@ -330,14 +330,14 @@ private PackageLookupValue computeExternalPackageLookupValue(
     RepositoryValue repositoryValue;
     try {
       repositoryValue = (RepositoryValue) env.getValueOrThrow(
-          repositoryKey, NoSuchPackageException.class, IOException.class, EvalException.class);
+          repositoryKey, NoSuchPackageException.class, IOException.class, EvalException.class, AlreadyReportedException.class);
```
janakdr, line 333: is this actually needed? I don't think it's ever thrown, right? Here and below

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
joeleba: I agree with your point. However, from reading [RepositoryDelegatorFunction](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java;l=183;drc=d4d9e28140dedbcc84e3486d06b373fbc2fbc51f), it seems to me that `RepositoryFunctionException` is being used as if it's `RepositoryDelegatorFunctionException`. We can have a separate PR for that refactoring.

src/main/java/com/google/devtools/build/lib/skyframe/PackageLookupFunction.java
joeleba: It is, actually. The path is PackageLookupFunction -> RepositoryLoaderFunction -> RepositoryDelegatorFunction. Caught this via a failing test on BazelCI.

src/main/java/com/google/devtools/build/lib/skyframe/BUILD
joeleba: Done.

src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryFunction.java
joeleba: Done.

janakdr (approved):

PR #12368, opened 2020-10-28
Fix incompatible target skipping for skylib's analysistest

### LINKED ISSUE #12366: `target_compatible_with` failing with `analysis_test`
Using the new [`target_compatible_with` feature](https://github.com/bazelbuild/bazel/pull/10945) with an [`analysis_test`](https://github.com/bazelbuild/bazel-skylib/blob/182046f090d6ecae264c7e5d520da096370287e2/rules/analysis_test.bzl) fails with this error on incompatible platforms:

```
ERROR: /workdir/test/BUILD:13:29: in coverage_xcode_prefix_map_test rule //test:coverage_settings_xcode_prefix_map: rules with analysis_test=true must return an instance of AnalysisTestResultInfo
ERROR: Analysis of target '//test:coverage_settings_xcode_prefix_map' failed; build aborted: Analysis of target '//test:coverage_settings_xcode_prefix_map' failed
```

https://buildkite.com/bazel/rules-swift-swift/builds/1980#af8b4f17-01b9-4e2d-99d0-93ef5abc6203

Example failure here for a test that is only compatible with macOS, and therefore fails with the above error on Linux https://github.com/bazelbuild/rules_swift/pull/514

Flipping the `target_compatible_with` condition to only support Linux also makes it fail the same way on macOS.

macOS / Linux

```
2020/10/27 17:39:36 Using unreleased version at commit f3604a6c9c433c42f3adfa2bc1551150482aa73e
development version
```

`USE_BAZEL_VERSION=last_green` with bazelisk

philsc:
In #12366, Keith Smiley (@keith) reported that he was having trouble
using the new `target_compatible_with` attribute in `rules_swift`.
Specifically, setting it on an `analysistest` test target was giving
him the following error:

    ERROR: /workdir/test/BUILD:13:29: in coverage_xcode_prefix_map_test rule //test:coverage_settings_xcode_prefix_map: rules with analysis_test=true must return an instance of AnalysisTestResultInfo
    ERROR: Analysis of target '//test:coverage_settings_xcode_prefix_map' failed; build aborted: Analysis of target '//test:coverage_settings_xcode_prefix_map' failed

This was caused by the fact that for incompatible targets we create a
dummy `ConfiguredTarget` that only provides the bare minimum to make
the rest of bazel happy before it gets skipped. It turns out that
`analysistest` rules have additional checks that need to be satisfied.
This patch aims to satisfy those additional checks without impacting
functionality.

Fixes #12366

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
@@ -43,6 +43,7 @@ import com.google.devtools.build.lib.analysis.constraints.EnvironmentCollection.
 import com.google.devtools.build.lib.analysis.constraints.SupportedEnvironmentsProvider.RemovedEnvironmentCulprit;
 import com.google.devtools.build.lib.analysis.platform.ConstraintValueInfo;
 import com.google.devtools.build.lib.analysis.platform.PlatformProviderUtils;
+import com.google.devtools.build.lib.analysis.test.AnalysisTestResultInfo;
 import com.google.devtools.build.lib.cmdline.Label;
 import com.google.devtools.build.lib.collect.nestedset.NestedSet;
 import com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder;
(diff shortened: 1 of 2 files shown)


philsc:
@keith , when you have a chance, could you do me a favor and give this a try?

I realized just now that the test case I used is slightly different from your use case, but I'm hoping it'll still work.

If it doesn't work, I'll take another look tomorrow or Thursday.

philsc:
@gregestren , does this look reasonable? Using `addNativeDeclaredProvider` feels a bit invasive, but I couldn't find anything better.

gregestren:
It looks reasonable to me given the context. What about it do you find invasive?

philsc:
"invasive" was definitely the wrong word to use, apologies.
It just seemed a bit too low-level, "reachy".

Looking at it with fresh eyes, it's not looking that bad. Sorry for the confusion.

keith:
I validated this now correctly skips the test:

```
//test:coverage_settings_xcode_prefix_map                               SKIPPED
Executed 42 out of 43 tests: 42 tests pass and 1 was skipped.
```

gregestren (approved):

keith:
@gregestren can you help land this?

gregestren:
I started merging on Friday but ran into some CI silliness. Stay tuned...

PR #12391, opened 2020-10-30
Add --experimental_remote_execution_keepalive option

coeuvre:
When it's on, Bazel will make remote execution calls keepalive using GrpcRemoteExecutorKeepalived. https://github.com/bazelbuild/proposals/pull/204.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
@@ -167,12 +167,6 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
         // If lastOperation is not null, we know the execution request is accepted by the server. In
         // this case, we will fallback to WaitExecution() loop when the stream is broken.
         if (lastOperation != null) {
-          // At this point we know that the waitExecutionBackoff is just reset, so call
-          // nextDelayMillis() here to ensure the expected retry times(--remote_retry + 1) for
-          // waitExecutionBackoff: once you reset it, the next call to nextDelayMillis() will not
-          // increase the counter.
-          waitExecutionBackoff.nextDelayMillis();
-
           // By returning null, we are going to fallback to WaitExecution() loop.
           return null;
         }
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
@@ -84,8 +84,8 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
     private final CallCredentialsProvider callCredentialsProvider;
     // Count retry times for Execute() calls and can't be reset.
     private final Backoff executeBackoff;
-    // Count retry times for WaitExecution() calls and will be reset once we received response from
-    // server without errors.
+    // Count retry times for WaitExecution() calls and is reset when we receive any response from
+    // the server that is not an error.
     private final ProgressiveBackoff waitExecutionBackoff;
     private final Supplier<ExecutionBlockingStub> executionBlockingStubSupplier;
 
@@ -162,17 +162,14 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
 
       try {
         Iterator<Operation> operationStream = executionBlockingStubSupplier.get().execute(request);
-
-        // We don't want to reset backoff for Execute() since if there is an error:
-        //   1. If happened before we received a first response, we want to ensure the retry counter
-        //      is increased and call Execute() again.
-        //   2. Otherwise, we will fallback to WaitExecution() loop.
-        //
-        // This also prevent us from being stuck at a infinite loop:
-        //   Execute() -> WaitExecution() -> Execute()
-        return handleOperationStream(operationStream, /* resetWaitExecutionBackoff */ false);
+        return handleOperationStream(operationStream);
       } catch (StatusRuntimeException e) {
-        if (lastOperation != null) {
+        // If lastOperation is not null, we know the execution request is accepted by the server. In
+        // this case, we will fallback to WaitExecution() loop when the stream is broken.
+        //
+        // However, we only do so if waitExecutionBackoff should retry. Also increase the retry
+        // counter at the same time (done by nextDelayMillis()).
+        if (lastOperation != null && waitExecutionBackoff.nextDelayMillis() >= 0) {
           // By returning null, we are going to fallback to WaitExecution() loop.
           return null;
         }
@@ -190,15 +187,12 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
       try {
         Iterator<Operation> operationStream = executionBlockingStubSupplier.get()
             .waitExecution(request);
-
-        // We want to reset backoff for WaitExecution() so we can "infinitely" wait for the
-        // execution to complete as long as they are making progress (by returning response).
-        return handleOperationStream(operationStream, /* resetWaitExecutionBackoff */ true);
+        return handleOperationStream(operationStream);
       } catch (StatusRuntimeException e) {
         // A NOT_FOUND error means Operation was lost on the server, retry Execute().
         //
         // However, we only retry Execute() if executeBackoff should retry. Also increase the retry
-        // counter at the same time since we will technically retry (done by nextDelayMillis()).
+        // counter at the same time (done by nextDelayMillis()).
         if (e.getStatus().getCode() == Code.NOT_FOUND && executeBackoff.nextDelayMillis() >= 0) {
           lastOperation = null;
           return null;
@@ -208,19 +202,27 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
     }
 
     /** Process a stream of operations from Execute() or WaitExecution(). */
-    ExecuteResponse handleOperationStream(Iterator<Operation> operationStream,
-        boolean resetWaitExecutionBackoff) throws IOException {
+    ExecuteResponse handleOperationStream(Iterator<Operation> operationStream) throws IOException {
       try {
         while (operationStream.hasNext()) {
           Operation operation = operationStream.next();
-          ExecuteResponse response = handleOperation(operation);
+          ExecuteResponse response = extractResponseOrThrowIfError(operation);
 
-          // At this point, we received the response without error.
+          // At this point, we successfully received a response that is not an error.
           lastOperation = operation;
 
-          if (resetWaitExecutionBackoff) {
-            waitExecutionBackoff.reset();
-          }
+          // We don't want to reset executeBackoff since if there is an error:
+          //   1. If happened before we received a first response, we want to ensure the retry counter
+          //      is increased and call Execute() again.
+          //   2. Otherwise, we will fallback to WaitExecution() loop.
+          //
+          // This also prevent us from being stuck at a infinite loop:
+          //   Execute() -> WaitExecution() -> Execute()
+          //
+          // However, we do want to reset waitExecutionBackoff so we can "infinitely" wait
+          // for the execution to complete as long as they are making progress (by returning a
+          // response that is not an error).
+          waitExecutionBackoff.reset();
 
           // Update execution progress to the observer.
           //
@@ -242,7 +244,7 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
         }
 
         // The operation completed successfully but without a result.
-        throw new IOException("Remote server error: execution terminated with no result.");
+        throw new IOException("Remote server error: execution terminated with no result");
       } finally {
         close(operationStream);
       }
@@ -271,7 +273,7 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
     }
 
     @Nullable
-    static ExecuteResponse handleOperation(Operation operation)
+    static ExecuteResponse extractResponseOrThrowIfError(Operation operation)
         throws IOException {
       if (operation.getResultCase() == Operation.ResultCase.ERROR) {
         throwIfError(operation.getError(), null);
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
@@ -28,6 +28,7 @@ import com.google.devtools.build.lib.remote.options.RemoteOptions;
 import com.google.devtools.build.lib.remote.util.TracingMetadataUtils;
 import com.google.devtools.build.lib.remote.util.Utils;
 import com.google.longrunning.Operation;
+import com.google.longrunning.Operation.ResultCase;
 import com.google.rpc.Status;
 import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
@@ -83,6 +84,7 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
     private final ProgressiveBackoff backoff;
     private final Supplier<ExecutionBlockingStub> executionBlockingStubSupplier;
 
+    // Last response (without error) we received from server.
     private Operation lastOperation;
 
     Execution(ExecuteRequest request,
@@ -120,11 +122,18 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
       Preconditions.checkState(lastOperation == null);
 
       ExecuteResponse response = null;
+      // Exit the loop as long as we get a response from either Execute() or WaitExecution().
       while (response == null) {
+        // We use refreshIfUnauthenticated inside retry block. If use it outside, retrier will stop
+        // retrying when received a unauthenticated error, and propagate to refreshIfUnauthenticated
+        // which will then call retrier again. It will reset the retry time counter so we could
+        // retry more than --remote_retry times which is not expected.
         response = retrier
             .execute(() -> Utils.refreshIfUnauthenticated(this::execute, callCredentialsProvider),
                 backoff);
 
+        // If no response from Execute(), use WaitExecution() in a "loop" which is implicated inside
+        // the retry block.
         if (response == null) {
           response = retrier.execute(
               () -> Utils.refreshIfUnauthenticated(this::waitExecution, callCredentialsProvider),
@@ -140,11 +149,16 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
       Preconditions.checkState(lastOperation == null);
 
       try {
-        Iterator<Operation> operations = executionBlockingStubSupplier.get().execute(request);
-        return handleStreamOperations(operations, /* resetBackoff */ false);
+        Iterator<Operation> streamOperations = executionBlockingStubSupplier.get().execute(request);
+
+        // We don't want to reset backoff for Execute() since if there is an error:
+        //   1. If happened before we received a first response, we want to ensure the retry counter
+        //      is increased and call Execute() again.
+        //   2. Otherwise, we will fallback to WaitExecution() loop.
+        return handleStreamOperations(streamOperations, /* resetBackoff */ false);
       } catch (StatusRuntimeException e) {
         if (lastOperation != null) {
-          // By returning null, we are going to call WaitExecution in a loop.
+          // By returning null, we are going to fallback to WaitExecution() loop.
           return null;
         }
         throw new IOException(e);
@@ -159,9 +173,11 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
           .setName(lastOperation.getName())
           .build();
       try {
-        Iterator<Operation> operations = executionBlockingStubSupplier.get()
+        Iterator<Operation> streamOperations = executionBlockingStubSupplier.get()
             .waitExecution(request);
-        return handleStreamOperations(operations, /* resetBackoff */ true);
+        // We want to reset backoff for WaitExecution() so we can "infinitely" wait for the
+        // execution to complete as long as they are making progress (by returning response).
+        return handleStreamOperations(streamOperations, /* resetBackoff */ true);
       } catch (StatusRuntimeException e) {
         if (e.getStatus().getCode() == Code.NOT_FOUND) {
           // Operation was lost on the server. Retry Execute.
@@ -172,42 +188,49 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
       }
     }
 
-    ExecuteResponse handleStreamOperations(Iterator<Operation> operations, Boolean resetBackoff)
-        throws IOException {
+    /**
+     * Process a stream of operations from Execute() or WaitExecution().
+     *
+     * @param resetBackoff reset backoff if true once received a response successfully.
+     */
+    ExecuteResponse handleStreamOperations(Iterator<Operation> streamOperations,
+        boolean resetBackoff) throws IOException {
       try {
-        while (operations.hasNext()) {
-          Operation operation = operations.next();
+        while (streamOperations.hasNext()) {
+          Operation operation = streamOperations.next();
 
           if (resetBackoff) {
-            // Assuming the server has made progress since we received the response. Reset the backoff
-            // so that this request has a full deck of retries
+            // Assuming the server has made progress since we received the response. Reset the
+            // backoff so that this request has a full deck of retries.
             backoff.reset();
           }
 
           ExecuteResponse response = handleOperation(operation, observer);
+
+          // At this point, we received the response without error
+          lastOperation = operation;
+
           if (response != null) {
             return response;
           }
-
-          lastOperation = operation;
         }
 
         // The operation completed successfully but without a result.
         throw new IOException("Remote server error: execution terminated with no result.");
       } finally {
-        close(operations);
+        close(streamOperations);
       }
     }
 
-    void close(Iterator<Operation> operations) {
+    void close(Iterator<Operation> streamOperations) {
       // The blocking streaming call closes correctly only when trailers and a Status are received
       // from the server so that onClose() is called on this call's CallListener. Under normal
       // circumstances (no cancel/errors), these are guaranteed to be sent by the server only if
-      // operations.hasNext() has been called after all replies from the stream have been
+      // streamOperations.hasNext() has been called after all replies from the stream have been
       // consumed.
       try {
-        while (operations.hasNext()) {
-          operations.next();
+        while (streamOperations.hasNext()) {
+          streamOperations.next();
         }
       } catch (StatusRuntimeException e) {
         // Cleanup: ignore exceptions, because the meaningful errors have already been propagated.
@@ -243,13 +266,23 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
       }
 
       if (operation.getDone()) {
+        if (operation.getResultCase() == ResultCase.RESULT_NOT_SET) {
+          throw new ExecutionStatusException(Status.newBuilder()
+              .setCode(com.google.rpc.Code.DATA_LOSS_VALUE)
+              .setMessage("Unexpected result of remote execution: no result")
+              .build(), null);
+        }
         Preconditions.checkState(operation.getResultCase() != Operation.ResultCase.RESULT_NOT_SET);
         ExecuteResponse response = operation.getResponse().unpack(ExecuteResponse.class);
         if (response.hasStatus()) {
           throwIfError(response.getStatus(), response);
         }
-        Preconditions.checkState(
-            response.hasResult(), "Unexpected result of remote execution: no result");
+        if (!response.hasResult()) {
+          throw new ExecutionStatusException(Status.newBuilder()
+              .setCode(com.google.rpc.Code.DATA_LOSS_VALUE)
+              .setMessage("Unexpected result of remote execution: no result")
+              .build(), response);
+        }
         return response;
       }
 
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
@@ -82,7 +82,10 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
     private final OperationObserver observer;
     private final RemoteRetrier retrier;
     private final CallCredentialsProvider callCredentialsProvider;
+    // Count retry times for Execute() calls and can't be reset.
     private final Backoff executeBackoff;
+    // Count retry times for WaitExecution() calls and will be reset once we received response from
+    // server without errors.
     private final ProgressiveBackoff waitExecutionBackoff;
     private final Supplier<ExecutionBlockingStub> executionBlockingStubSupplier;
 
@@ -135,8 +138,14 @@ public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
             .execute(() -> Utils.refreshIfUnauthenticated(this::execute, callCredentialsProvider),
                 executeBackoff);
 
-        // If no response from Execute(), use WaitExecution() in a "loop" which is implicated inside
-        // the retry block.
+        // If no response from Execute(), use WaitExecution() in a "loop" which is implemented
+        // inside the retry block.
+        //
+        // The cases to exit the loop:
+        //   1. Received the final response.
+        //   2. Received a un-retriable gRPC error.
+        //   3. Received NOT_FOUND error where we will retry Execute() (by returning null).
+        //   4. Received consecutive retriable gRPC errors (up to max retry times).
         if (response == null) {
           response = retrier.execute(
               () -> Utils.refreshIfUnauthenticated(this::waitExecution, callCredentialsProvider),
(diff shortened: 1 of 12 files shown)


coeuvre:
cc @ulfjack @buchgr @EricBurnett

coeuvre:
test cases in coming commits next week.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+
+  private static class Execution {
+    private final ExecuteRequest request;
+    private final OperationObserver observer;
+    private final RemoteRetrier retrier;
+    private final CallCredentialsProvider callCredentialsProvider;
+    private final ProgressiveBackoff backoff;
+    private final Supplier<ExecutionBlockingStub> executionBlockingStubSupplier;
+
+    private Operation operation;
```
lberki: nit: currentOperation? (some Javadoc would be useful so that I know what this field does)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+      //      accepted by the server and will be executed ASAP;
+      //   3. Server may execute the action silently and send a reply once it is done.
+      observer.onNext(operation);
+
+      if (operation.getResultCase() == Operation.ResultCase.ERROR) {
+        throwIfError(operation.getError(), null);
+      }
+
+      if (operation.getDone()) {
+        Preconditions.checkState(operation.getResultCase() != Operation.ResultCase.RESULT_NOT_SET);
```
lberki, line 246: If this precondition check fails, it causes Bazel to crash. Looks like this come from the server, which would mean that the server can give inputs to Bazel that cause it to crash?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+        ExecuteResponse response = operation.getResponse().unpack(ExecuteResponse.class);
+        if (response.hasStatus()) {
+          throwIfError(response.getStatus(), response);
+        }
+        Preconditions.checkState(
```
lberki, line 251: same here: a buggy server shouldn't crash Bazel

lberki (commented):
...and this is how far I got today. You should probably also ask someone from the RBE side to review this.

lberki:
@agoulti , mind taking a look? @philwo and @meisterT are out of office today so I am trying to provide Bazel knowledge, but the RBE interface isn't exactly my forte.

agoulti:
> @agoulti , mind taking a look? @philwo and @meisterT are out of office today so I am trying to provide Bazel knowledge, but the RBE interface isn't exactly my forte.

Can this wait until they are back?
I'm quite constrained for time lately and I'm not off the cuff familiar with this portion of the code.

I'm willing to "drop everything to help" for the underlying issue (https://github.com/bazelbuild/bazel/issues/11782) but my understanding is that this is a very limited mitigation? (I think in the vast majority of cases it's not the Execute call that hangs?)

lberki:
@agoulti : This is a P0 on our side, but at the same time, it's just something hidden behind an experimental flag, so if we don't get it right the first time, it doesn't cause any damage other than delaying the fix for that issue. I'll try to pretend I'm a competent reviewer then.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+        if (e.getStatus().getCode() == Code.NOT_FOUND) {
+          // Operation was lost on the server. Retry Execute.
+          operation = null;
+          return null;
+        }
+        throw new IOException(e);
+      }
+    }
+
+    ExecuteResponse handleStreamOperations(Iterator<Operation> operations, Boolean resetBackoff)
```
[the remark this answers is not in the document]
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+ * A remote work executor that uses gRPC for communicating the work, inputs and outputs.
+ *
+ * <p>It differs from {@link GrpcRemoteExecutor} by setting timeout on each execution calls to
+ * ensure we never be stuck due to network issues.
+ *
+ * @see <a href="https://docs.google.com/document/d/1NgDPsCIwprDdqC1zj0qQrh5KGK2hQTSTux1DAvi4rSc">
+ * Keepalived Remote Execution</a>
+ */
+@ThreadSafe
+public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
```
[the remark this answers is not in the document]
coeuvre: 1. The details are different between them that simply adding a timeout option to GrpcRemoteExecutor is not enough.
2. Using new class allow me to explore new ideas without breaking users :)

I will remove GrpcRemoteExecutor once this is stable.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+      //   That means the Operation was lost on the server, and we will retry to Execute.
+      // - Any call can return an Operation object with an error status in the result. Such
+      //   Operations are completed and failed; however, some of these errors may be retriable.
+      //   These errors should trigger a retry of the Execute call, resulting in a new Operation.
+      Preconditions.checkState(operation == null);
+
+      ExecuteResponse response = null;
+      while (response == null) {
+        response = retrier
+            .execute(() -> Utils.refreshIfUnauthenticated(this::execute, callCredentialsProvider),
```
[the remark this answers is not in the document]
coeuvre: In GrpcRemoteExecutor, we use refreshIfUnauthenticated outside of retry block: if there is a unauthenticated error, retrier will stop retrying and propagate the error to refreshIfUnauthenticated which will call retrier again. It will reset the retry time counter so we could retry more than `--remote_retry` times which is not expected.

I will review all the refreshIfUnauthenticated call sites and refactor the code as per https://github.com/bazelbuild/bazel/pull/12156#issuecomment-701104900.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+      //   In that case, we retry the WaitExecution call on the same operation object.
+      // - A WaitExecution call may fail with a NOT_FOUND error (raise a StatusRuntimeException).
+      //   That means the Operation was lost on the server, and we will retry to Execute.
+      // - Any call can return an Operation object with an error status in the result. Such
+      //   Operations are completed and failed; however, some of these errors may be retriable.
+      //   These errors should trigger a retry of the Execute call, resulting in a new Operation.
+      Preconditions.checkState(operation == null);
+
+      ExecuteResponse response = null;
+      while (response == null) {
```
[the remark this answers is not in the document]
coeuvre: Exit the loop as long as we get a response from either Execute() or WaitExecution().

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+    ExecuteResponse waitExecution() throws IOException {
+      Preconditions.checkState(operation != null);
+
+      WaitExecutionRequest request = WaitExecutionRequest.newBuilder()
+          .setName(operation.getName())
+          .build();
+      try {
+        Iterator<Operation> operations = executionBlockingStubSupplier.get()
+            .waitExecution(request);
+        return handleStreamOperations(operations, /* resetBackoff */ true);
```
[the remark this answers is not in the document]
coeuvre: By reseting backoff, we tell retrier keep retrying in case of an error.

We don't want to reset backoff for Execute() since if there is an error:
1. It happens before we received a first response, we want to ensure the retry counter is increased and call Execute() again.
2. Otherwise, we will fallback to WaitExecution() loop.

However, we want to reset backoff for WaitExecution() so we can "infinitely" wait for the execution to complete as long as they are making progress (by returning response with WaitExecution()).

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,272 @@
...
+        this::executionBlockingStub);
+    return execution.start();
+  }
+
+  @Override
+  public void close() {
+    if (closed.getAndSet(true)) {
+      return;
+    }
+    channel.release();
```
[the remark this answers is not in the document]
coeuvre: If the reference counter reaches zero, `channel.shutdown` will be called which

> Initiates an orderly shutdown in which preexisting calls continue but new calls are immediately cancelled.

as described [here](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannel.html#shutdown--)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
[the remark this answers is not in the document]
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Good catch! Changed to throw ExecutionStatusException (which will be wrapped in an IOException).

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

coeuvre:
@lberki Thanks for reviewing! Updated.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,308 @@
...
+        throw new IOException(e);
+      }
+    }
+
+    /**
+     * Process a stream of operations from Execute() or WaitExecution().
+     *
+     * @param resetBackoff reset backoff if true once received a response successfully.
+     */
+    ExecuteResponse handleStreamOperations(Iterator<Operation> streamOperations,
```
lberki, line 196: nit: not that this is very important, but this method is still called `handleStreamOperations` and not `streamOperations` and you said that you renamed this. How about calling the method `streamOperations` and the argument `operations` or `operationStream`?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,308 @@
...
+
+      return response;
+    }
+
+    @Nullable
+    ExecuteResponse execute() throws IOException {
+      Preconditions.checkState(lastOperation == null);
+
+      try {
+        Iterator<Operation> streamOperations = executionBlockingStubSupplier.get().execute(request);
```
lberki, line 152: nit: call the variables `operations` or `operationStream`? (the way I parse "streamOperations" is that it "streams Operations", that is, an activity which is better suited as a name for a method and "operationStream" sounds like "a stream of Operations" to me, which is a noun and is thus a better variable name)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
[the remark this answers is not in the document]
lberki: Thanks! I think it's actually better this way, but I didn't know if this was intentional or not and I was not quite sure what it does. I also appreciate the new comment!

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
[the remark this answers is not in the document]
lberki: Nevermind; I got confused by the fact that `waitExecute` is passed as a method reference in a call to `execute()`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Process a stream of operations from Execute() or WaitExecution().
+     *
+     * @param resetBackoff reset backoff if true once received a response successfully.
+     */
+    ExecuteResponse handleStreamOperations(Iterator<Operation> streamOperations,
+        boolean resetBackoff) throws IOException {
+      try {
+        while (streamOperations.hasNext()) {
```
lberki, line 199: What is the difference between getting a new `Operation` on the stream returned by `execute()` and the stream finishing and requiring us to call `waitExecution()` again?

My theory: this is so that we can have actions that run longer than the maximum allowed time for an RPC but then why stream `Operation` messages at all instead of just returning one and requiring to call `waitExecution()` in a loop?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,308 @@
...
+          Operation operation = streamOperations.next();
+
+          if (resetBackoff) {
+            // Assuming the server has made progress since we received the response. Reset the
+            // backoff so that this request has a full deck of retries.
+            backoff.reset();
```
lberki, line 205: It looks like the intent is that the backoff is reset when the server thinks that progress was made. The comment in line 210 indicates that that's the point where the client is certain that there were no errors in the last `Operation`. Why not unconditionally reset the backoff there?

lberki (commented):
I have to run some errands, so I haven't had time to look at the new test infrastructure. I hope we can still get another review round in today.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Sorry, I thought you were talking about renaming the parameter `operations` to `streamOperations`. How about rename to `handleOperationStream`? Since there is a corresponding method called `handleOperation` below for processing a single operation.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: According to the [specification](https://source.corp.google.com/piper///depot/google3/third_party/bazel_remoteapis/build/bazel/remote/execution/v2/remote_execution.proto), server can return multiple `Operation`s using the stream but is not required. After the first response, the stream we get from Execute() is the same as what we can get from WaitExecution().

If server implementation send `Operation`s over the stream periodically, we don't need `WaitExecution` in normal cases as you said.

However, stream can finish due to network issues or simply caused by timeout. In this case, we have to use WaitExecution().

So why Execute() return a stream of `Operation` instead of one? I am not sure but this is defined by REAPI.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Moved the code for resting after client verified no errors in the `Operation`. However, we still need to distinct between Execute() and WaitExecution() to avoid infinite loop caused by a buggy server. For example:
1. Call Execute(), server returns the first response, and then timeout when waiting for the next. Fallback to WaitExecution().
2. Call WaitExecution(), but returned NOT_FOUND (server may already finished the execution and delete the reference). Retry with Execute().

See test case `executeRemotely_notFoundLoop_reportError`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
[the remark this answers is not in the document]
coeuvre: > do you have a list of things in which they are different?

From the users perspective, they should be the same except the timeout. However, this requires carefully design of the control flow and handle the edge cases correctly.

> What in this code do you think is a breaking change?

Add timeout to detect network issues so Bazel will not hang forever but also support arbitrary long execution.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
lberki: If the spec says so, the spec says so... it sure looks weird and I was hoping you could tell me why this strange pattern exists for my education. From what I see, it's not at all common anywhere else (but exists in a few places). In any case, this is not relevant for this code review.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,316 @@
...
+      while (response == null) {
+        // We use refreshIfUnauthenticated inside retry block. If use it outside, retrier will stop
+        // retrying when received a unauthenticated error, and propagate to refreshIfUnauthenticated
+        // which will then call retrier again. It will reset the retry time counter so we could
+        // retry more than --remote_retry times which is not expected.
+        response = retrier
+            .execute(() -> Utils.refreshIfUnauthenticated(this::execute, callCredentialsProvider),
+                executeBackoff);
+
+        // If no response from Execute(), use WaitExecution() in a "loop" which is implicated inside
```
lberki, line 138: nit: s/implicated/implemented/?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,316 @@
...
+            .waitExecution(request);
+
+        // We want to reset backoff for WaitExecution() so we can "infinitely" wait for the
+        // execution to complete as long as they are making progress (by returning response).
+        return handleOperationStream(operationStream, (operation) -> {
+          // Assuming the server has made progress since we received the response. Reset the
+          // backoff so that this request has a full deck of retries.
+          waitExecutionBackoff.reset();
+
+          lastOperation = operation;
```
lberki, line 192: These two lines seem to be common between the two lambdas passed in the second arg of `handleOperationStream()`. Move them to where the lambda is called.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
lberki: FWIW, if this is necessary, I think it's nicer to do it with a Boolean flag and not a lambda (lambdas make the code flow harder to follow visually). It looks like the `lastOperation = operation; observer.onNext(operation);`

So you want to protect against the case where the server "forgets" about the existence of a request between an  `Execute()` call and the subsequent `WaitExecution()` one? That makes sense, but are we then not afraid of this scenario:

1. Client sends `Execute()`, then the server breaks the stream
2. Client sends `WaitExecution()`, server returns a few `Operation` messages, then stops responding
3. GOTO 1. The backoff is reset, so this is an infinite loop.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: > 3. GOTO 1. The backoff is reset, so this is an infinite loop.

If server stops responding, we won't GOTO 1. The cases to exit the `WaitExecution()` loop:
1. Received the final response.
2. Received a noretriable gRPC error.
3. Received NOT_FOUND error where we will retry Execute().
4. Received consecutive retriable gRPC errors (up to `--remote_retry`).

Then the scenario could be:
1. Client sends `Execute()`, server return a response, and then stream is broken
2. Client sends `WaitExecution()`, server returns a few `Operation` messages, and then return a `NOT_FOUND` error
3. GOTO 1.

However, this is fixed by having two types of backoff:
1. `executeBackoff`: used for count the retry times for `Execute()` calls and can't be reset.
2. waitExecutionBackoff: used for count retry times for `WaitExecution()` calls and will be reset once we received a GOOD response from server.

Test case `executeRemotely_notFoundLoop_reportError` cover this case.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,320 @@
...
+  }
+
+  private static class Execution {
+    private final ExecuteRequest request;
+    private final OperationObserver observer;
+    private final RemoteRetrier retrier;
+    private final CallCredentialsProvider callCredentialsProvider;
+    // Count retry times for Execute() calls and can't be reset.
+    private final Backoff executeBackoff;
+    // Count retry times for WaitExecution() calls and will be reset once we received response from
```
lberki, line 87: nit: "and is reset when we receive any response from the server that is not in error" is probably a more precise wording (although English is not my first language...)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,320 @@
...
+          //   3. Server may execute the action silently and send a reply once it is done.
+          observer.onNext(operation);
+
+          if (response != null) {
+            return response;
+          }
+        }
+
+        // The operation completed successfully but without a result.
+        throw new IOException("Remote server error: execution terminated with no result.");
```
lberki, line 245: nit: remove dot from the end of the exception message (IIRC that's the convention around here)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,320 @@
...
+        boolean resetWaitExecutionBackoff) throws IOException {
+      try {
+        while (operationStream.hasNext()) {
+          Operation operation = operationStream.next();
+          ExecuteResponse response = handleOperation(operation);
+
+          // At this point, we received the response without error.
+          lastOperation = operation;
+
+          if (resetWaitExecutionBackoff) {
```
lberki, line 221: What harm would it do to unconditionally reset this backoff? When called from Execute(), the Execute() backoff is bumped, so we don't have an infinite loop.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,320 @@
...
+        // If no response from Execute(), use WaitExecution() in a "loop" which is implemented
+        // inside the retry block.
+        //
+        // The cases to exit the loop:
+        //   1. Received the final response.
+        //   2. Received a un-retriable gRPC error.
+        //   3. Received NOT_FOUND error where we will retry Execute() (by returning null).
+        //   4. Received consecutive retriable gRPC errors (up to max retry times).
+        if (response == null) {
+          response = retrier.execute(
```
lberki, line 150: Do I understand correctly that if the WaitExecution() backoff runs out, we'll raise an exception? 

The scenario I'm looking to avoid is that we first wait for the WaitExecution() backoff to run out, then that results in bumping the Execute() backoff, and then this would signal an error after a N*M tries when one backoff running out causes the other backoff just to be bumped by one)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Good catch! I cannot come up with a case where this could be an issue. Changed.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Yes. It will rase an exception.

See test case `executeRemotely_executeAndRetryWait_failForConsecutiveErrors`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -162,17 +162,14 @@ ExecuteResponse execute() throws IOException {
...
-        return handleOperationStream(operationStream, /* resetWaitExecutionBackoff */ false);
+        return handleOperationStream(operationStream);
       } catch (StatusRuntimeException e) {
-        if (lastOperation != null) {
+        // If lastOperation is not null, we know the execution request is accepted by the server. In
+        // this case, we will fallback to WaitExecution() loop when the stream is broken.
+        //
+        // However, we only do so if waitExecutionBackoff should retry. Also increase the retry
+        // counter at the same time (done by nextDelayMillis()).
+        if (lastOperation != null && waitExecutionBackoff.nextDelayMillis() >= 0) {
```
lberki, line 172: Why do you increase the delay here?

coeuvre: This is internal behavior of ProgressiveBackoff: once you reset it, the next call to `nextDelayMillis()` will not increase the counter. If `lastOperation != null`, we know that the `waitExecutionBackoff` is just reset, so call `nextDelayMillis()` here to ensure the expected retry times(--remote_retry + 1).

coeuvre: (`nextDelayMillis()` is used by retrier to determine whether it should retry and if so, how long to wait before try again.)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,325 @@
...
+        return handleOperationStream(operationStream);
+      } catch (StatusRuntimeException e) {
+        // If lastOperation is not null, we know the execution request is accepted by the server. In
+        // this case, we will fallback to WaitExecution() loop when the stream is broken.
+        if (lastOperation != null) {
+          // At this point we know that the waitExecutionBackoff is just reset, so call
+          // nextDelayMillis() here to ensure the expected retry times(--remote_retry + 1) for
+          // waitExecutionBackoff: once you reset it, the next call to nextDelayMillis() will not
+          // increase the counter.
+          waitExecutionBackoff.nextDelayMillis();
```
lberki: opt: since this is only so that you get the exact number of retries, consider removing this for the sake of simplicity (note the opt tag -- I don't feel very strongly about this)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,325 @@
...
+      try {
+        Iterator<Operation> operationStream = executionBlockingStubSupplier.get()
+            .waitExecution(request);
+        return handleOperationStream(operationStream);
+      } catch (StatusRuntimeException e) {
+        // A NOT_FOUND error means Operation was lost on the server, retry Execute().
+        //
+        // However, we only retry Execute() if executeBackoff should retry. Also increase the retry
+        // counter at the same time (done by nextDelayMillis()).
+        if (e.getStatus().getCode() == Code.NOT_FOUND && executeBackoff.nextDelayMillis() >= 0) {
```
lberki, line 193: (hiding a side effect in the nextDelayMillis() call is not great, but I think this is the best possible way to express what you want)

src/test/java/com/google/devtools/build/lib/remote/FakeExecutionService.java
```diff
@@ -0,0 +1,176 @@
...
+public class FakeExecutionService extends ExecutionImplBase {
+
+  private final OperationProvider executeOperationProvider = new OperationProvider();
+  private final OperationProvider waitExecutionOperationProvider = new OperationProvider();
+  private int execTimes;
+  private int waitTimes;
+
+  public static class OperationProvider {
+
+    private final Map<String, Deque<ImmutableList<Supplier<Operation>>>> operationProvider =
```
lberki, line 43: nit: add a comment as to what this data structure does

AFAIU it's a map from the request to the list of operations to be returned for each instance of that request, with `Supplier` used for either throwing an exception or returning an `Operation`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
```diff
@@ -0,0 +1,325 @@
...
+ * A remote work executor that uses gRPC for communicating the work, inputs and outputs.
+ *
+ * <p>It differs from {@link GrpcRemoteExecutor} by setting timeout on each execution calls to
+ * ensure we never be stuck due to network issues.
+ *
+ * @see <a href="https://docs.google.com/document/d/1NgDPsCIwprDdqC1zj0qQrh5KGK2hQTSTux1DAvi4rSc">
+ * Keepalived Remote Execution</a>
+ */
+@ThreadSafe
+public class GrpcRemoteExecutorKeepalived implements RemoteExecutionClient {
```
lberki, line 53: nit: Maybe call this `GrpcRemoteExecutor2` or `ExperimentalGrpcRemoteExecutor`? After all, it has differences from the current one other than keepalives, doesn' it?

lberki (approved):
Only nits left. Let's import this, it's been quite a few days already.

src/test/java/com/google/devtools/build/lib/remote/FakeExecutionService.java
coeuvre: Done by copying your words :)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Yes. I will probably refactor the Backoff and Retrier in another PR.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Renamed to `ExperimentalGrpcRemoteExecutor`.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteExecutorKeepalived.java
coeuvre: Removed.

werkt:
Hyrum's law strikes again.

This deviates from the existing GrpcRemoteExecutor's behavior, where RETRIABLE_GRPC_ERRORS inspire a reexecution. Without another call to execute, the same operation status is returned over multiple calls, all with the same retriable error. The language of the remote api seems to indicate that statuses returned in the operation error field with 'should retry' as its response mean 'should retry the entire execution', not just the [waitExecution] request.

So a server which always returns a particular RETRIABLE_GRPC_ERRORS member as an operation error will negate retry benefits.

I'm putting together a change with adjustments to this to match GrpcRemoteExecutor's behavior, and a test suite to exercise it as well based on your ExperimentalGrpcRemoteExecutorTest. Hopefully we can put this to rest, as I need the experimental to limit execute to a timeframe and try an outbound request to convince grpc that a socket has broken.

PR #12378, opened 2020-10-28
Update new target_compatible_with attribute to not exist on repository_rules

katre:
…y rules.

Fixes #12374.

PR #12396, opened 2020-11-02
Remove superfluous bazel_toolchains patch

philsc:
This is effectively a revert of
8946cc1746f4092d86380a23b0ac8b2d8699d3fc.

I originally needed to rename an attribute in `bazel_toolchains` for
22b4dbcaf05756d506de346728db3846da56b775. The `target_compatible_with`
attribute used in one of the repository rules clashed with the
attribute I added in the Incompatible Target Skipping patch. Once we
merged 22b4dbcaf05756d506de346728db3846da56b775 we realized that we
broke a bunch of projects because they pulled in unpatched versions of
`bazel_toolchains`.

John Cater (@katre) noticed that we can fix everything by making the
new attribute not apply to repository rules. He submitted
775a979bde21a71ddf04f67b865e8c43a29f702a to fix this.

That means that the original patch I added is actually not needed
anymore.

A follow-up patch will delete the .patch file from the source tree.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -432,10 +432,8 @@ http_file(
 
 http_archive(
     name = "bazel_toolchains",
-    patch_args = ["-p1"],
     patch_cmds = EXPORT_WORKSPACE_IN_BUILD_FILE,
     patch_cmds_win = EXPORT_WORKSPACE_IN_BUILD_FILE_WIN,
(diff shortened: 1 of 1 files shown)


philsc:
@katre , this is the first revert as per #12378.

katre (approved):

katre:
Thanks, I'll merge this today

PR #12399, opened 2020-11-02
Retrieve ijar from java_toolchain instead of using remote_java_tools.

comius:
Android tools depend on using ijar directly, which is provided in tools/jdk/BUILD using remote_java_tools macro.
Remote_java_tools uses selects to produce an archive name (java_lang_tools_linux, ..) and grab a tool out of it. The selected archive name needs to match the one toolchains will resolve to. 

Instead of ijar is already among things provided by java_toolchain and we can obtain it directly from the toolchain. The right way to do it is using already provided java_common.run_ijar.


comius:
cc @oquenchil @lberki

comius:
ping

Issue #12397, opened 2020-11-02
Bazel's shell tests on Windows are noisy unless you use "bazel shutdown"


philsc:
### Description of the problem / feature request:
When running a shell test, you get lots of noise about not being able to remove directories at the end of a test.
I was able to fix this by running `bazel shutdown` in the `tear_down()` helper.

@katre  suggested that there's likely a good way to fix this for all tests.

Here's the output without `bazel shutdown':
```console
$ bazel test src/test/shell/integration:target_compatible_with_external_repo_test --test_output=streamed
WARNING: Streamed test output requested. All tests will be run locally, without sharding, one at a time
Loading:
Loading: 0 packages loaded
DEBUG: C:/msys64/home/gamer/_bazel_gamer/sawo4taj/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14:
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: C:/msys64/home/gamer/_bazel_gamer/sawo4taj/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1804_java11 not using checked in configs as detect_java_home was set to True
DEBUG: C:/msys64/home/gamer/_bazel_gamer/sawo4taj/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14:
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: C:/msys64/home/gamer/_bazel_gamer/sawo4taj/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1604_java8 not using checked in configs as detect_java_home was set to True
Analyzing: target //src/test/shell/integration:target_compatible_with_external_repo_test (0 packages loaded, 0 targets configured)
INFO: Analyzed target //src/test/shell/integration:target_compatible_with_external_repo_test (0 packages loaded, 2 targets configured).
INFO: Found 1 test target...
[0 / 4] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[5 / 6] Testing //src/test/shell/integration:target_compatible_with_external_repo_test; 1s local
INFO[target_compatible_with_external_repo_test 2020-11-02 01:50:17 (+0000)] bazel binary is at /home/gamer/bazel/src/test/shell/bin
[...]

PR #12400, opened 2020-11-02
[WIP] Shutdown bazel after every test case

katre:
This reduces log spam on Windows, see #12397.

### DIFF
--- a/src/test/shell/testenv.sh
+++ b/src/test/shell/testenv.sh
@@ -654,6 +654,11 @@ function cleanup_workspace() {
 }
 
 function testenv_tear_down() {
+  if [ -d "${WORKSPACE_DIR:-}" ]; then
+    # Try to shutdown Bazel at the end to prevent log spam.
+    cd "${WORKSPACE_DIR}"
+    try_with_timeout bazel shutdown || true
+  fi
   cleanup_workspace
 }
 


src/test/shell/testenv.sh
```diff
@@ -654,6 +654,11 @@ function cleanup_workspace() {
 }
 
 function testenv_tear_down() {
+  if [ -d "${WORKSPACE_DIR:-}" ]; then
+    # Try to shutdown Bazel at the end to prevent log spam.
+    cd "${WORKSPACE_DIR}"
+    try_with_timeout bazel shutdown || true
```
meisterT, line 660: Note that @philwo carefully removed a lot of `bazel shutdown`s a while ago to speed up our tests by "a lot". This would revert this speedup.

philwo: Yes. We can’t do this. This would completely kill our test performance.

katre: I wasn't proposing it seriously, I assumed there are performance issues. Is there another way we could address #12397?

meisterT: Perhaps @meteorcloudy has an idea?

meteorcloudy: You cannot delete a directory on Windows until its file handle is released by other processes. `try_with_timeout` was [introduced](https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/testenv.sh;l=246?q=try_with_timeout&ss=bazel) to solve this problem. In #12397, the error message was generated by `try_with_timeout rm -fr target_skipping`, to reduce the noise, maybe changing `try_with_timeout` to redirect the error message to `/dev/null`. See https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/testenv.sh;l=248?q=try_with_timeout&ss=bazel

PR #12410, opened 2020-11-03
Use selects based on constraint_values in java_tools.

comius:
This makes them consistent with platformization which works on OS and CPU constraints from @platforms.

Flag --cpu=darwin maps on Bazel to macos+x86_64, internally --cpu=darwin_x86_64 is used. With platforms, those become a single config_setting.

I didn't combine together other darwin config_settings (dropping cpu), because of possible combination macos+ppc.


comius:
cc @plf,@lberki,@katre

katre (approved):

PR #12416, opened 2020-11-04
ensure the gRPC LoggingInterceptor does not throw exceptions where they cannot be handled

kevingessner:
This fixes a hang that I've struggled with for ages: when using `--build_event_json_file` to write a BEP file, enabling `--experimental_remote_grpc_log` causes the build to hang indefinitely after completion with a message of "Waiting for build events upload: JsonFormatFileTransport".

The default `--build_event_json_file_path_conversion` ends up doing an RPC call when closing the BEP file. Unfortunately, this call happens after the gRPC log file has already been closed. That exception in the LoggingInterceptor is uncatchable -- it [causes the RPC call to hang indefinitely](https://github.com/grpc/grpc-java/pull/6107). Therefore, Bazel needs to handle and consume the exception so `onClose` always succeeds.


coeuvre (approved):
Hi Kevin, thanks for your catch, pointers and PR!

This also remind me that we maybe handle onClose incorrectly at other places causing #11782.

PR #12422, opened 2020-11-05
Ensure NetworkTime never throw exceptions inside onClose

coeuvre:
Failed doing so will cause gRPC hanging forever. This could be one of causes that leads to #11782.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
@@ -101,7 +101,6 @@ public class NetworkTime {
               // already stopped.
               try {
                 networkTime.stop();
-                throw new IllegalStateException();
               } catch (RuntimeException e) {
                 logger.atWarning().withCause(e).log("Failed to stop networkTime");
               }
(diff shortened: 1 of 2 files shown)


coeuvre:
See #12416.

src/main/java/com/google/devtools/build/lib/remote/util/NetworkTime.java
```diff
@@ -87,9 +90,21 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
+             */
             @Override
             public void onClose(Status status, Metadata trailers) {
-              networkTime.stop();
+              // There is a risk that networkTime.stop() would throw a IllegalStateException: if
+              // networkTime.outstanding is overflowed, wallTime.stop() will be called even it's
+              // already stopped.
+              try {
+                networkTime.stop();
+                throw new IllegalStateException();
```
lberki, line 104: Is this line debugging code left in? (otherwise how could this code end up not throwing an exception?)

coeuvre: Ops, my bad, it was used for debugging. ( I told myself remember to delete that line)

lberki:
Yeah, this looks much better :)

lberki (approved):

PR #12425, opened 2020-11-06
Add target settings attribute to toolchain rule

comius:
This is the implementation of proposal https://docs.google.com/document/d/13yXhOQeMZQmlW3obyZ7fJP3sm3eVLbK3Nuk2LZjcZiU/edit.

The added `targetSetting` attribute, can reference `config_setting` rules. The attribute is resolved in target configuration and the returned `ConfigMatchingProvider` already has the value if this toolchain matches.

Difference to the proposal is that `targetSettings` are a `LABEL_LIST` where all the settings need to be satisfied. This might increase a bit usability. It makes the implementation a bit nicer, because instead of special handling when the value is not present, a list can uniformly handled.


comius:
cc @lberki, @oquenchil, @c-mita

src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
```diff
@@ -139,7 +140,8 @@ private static SingleToolchainResolutionValue resolveConstraints(
       }
 
       // Make sure the target platform matches.
-      if (!checkConstraints(
+      if (!toolchain.targetSettings().stream().allMatch(ConfigMatchingProvider::matches)
```
katre, line 143: We should probably enhance the output of --toolchain_resolution_debug to clarify when a toolchain is rejected for settings instead of target/exec constraints.

src/main/java/com/google/devtools/build/lib/rules/platform/ToolchainRule.java
```diff
@@ -78,6 +81,14 @@ public RuleClass build(RuleClass.Builder builder, RuleDefinitionEnvironment env)
                 .mandatoryProviders(ConstraintValueInfo.PROVIDER.id())
                 .allowedFileTypes()
                 .nonconfigurable("part of toolchain configuration"))
+        /* <!-- #BLAZE_RULE(toolchain).ATTRIBUTE(target_settings) -->
+        A list of <code>config_setting</code>s that must be satisfied by the target
+        build configuration in order for this toolchain to be selected for a target building.
```
katre, line 86: How about this wording?

```
A list of <code>config_setting</code>s that must be satisfied by the target configuration in order for this toolchain to be selected during toolchain resolution.
```

katre (changes requested):
Is the design review for this finished? I think we've discussed my concerns (and I see you're only implementing target_setting, not exec_setting), but has @cushon reviewed the design yet?

src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java
comius: Done.

src/main/java/com/google/devtools/build/lib/rules/platform/ToolchainRule.java
comius: Done.

comius:
> Is the design review for this finished? I think we've discussed my concerns (and I see you're only implementing target_setting, not exec_setting), but has @cushon reviewed the design yet?

No, still waiting for @cushon. Don't want to miss Bazel release though, because this could prolong everything for another month....

katre (approved):
LGTM, acknowledging that we may need to change details after the review finishes.

PR #12426, opened 2020-11-06
Ensure SettableFutures in remote module are all set.

coeuvre:
Failed doing so, will cause Bazel hanging forever. This could be one of causes for #11782.


lberki:
Reassigning to @philwo since he should be back this week

philwo (approved):
Thanks!

PR #12427, opened 2020-11-06
Use constraints in global conditions instead of configuration flags.

comius:
I changed configuration settings in //src/conditions to use constraints instead of flags and then fixed some places where this is used.

I fixed uses of darwin and darwin_x86_64, which using flags I believe is the same thing and using constraints actually becomes different.

Added config_setting for linux. Used it where it makes sense. It cannot be used in tools/jdk/..., because it somehow gets new @bazel_tools and old @bazel_tools/constraints.

This might break users of remote execution or cross compiling. The users working on a single platform should be safe.

### DIFF
--- a/src/main/java/net/starlark/java/eval/BUILD
+++ b/src/main/java/net/starlark/java/eval/BUILD
@@ -95,13 +95,7 @@ cc_binary(
     name = "libcpu_profiler.so",
     srcs = select({
         "//src/conditions:darwin": ["cpu_profiler_posix.cc"],
-        # There is no config_setting for Linux.
-        # See https://github.com/bazelbuild/bazel/issues/11107
-        "//src/conditions:linux_aarch64": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_arm": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_ppc": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_s390x": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_x86_64": ["cpu_profiler_posix.cc"],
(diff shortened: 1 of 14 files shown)


comius:
Downstream test at https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1751 (previously failing gerrit, cloudrobotics, flatbuffers)

katre (approved):
Nice to see the reduced duplication.

comius:
> Downstream test at https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1751 (previously failing gerrit, cloudrobotics, flatbuffers)

rules_nodejs fix: https://github.com/bazelbuild/rules_nodejs/pull/227
rules_docker fix: https://github.com/bazelbuild/rules_docker/pull/1674

src/main/java/net/starlark/java/eval/BUILD
```diff
@@ -95,13 +95,7 @@ cc_binary(
...
     srcs = select({
         "//src/conditions:darwin": ["cpu_profiler_posix.cc"],
-        # There is no config_setting for Linux.
-        # See https://github.com/bazelbuild/bazel/issues/11107
-        "//src/conditions:linux_aarch64": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_arm": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_ppc": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_s390x": ["cpu_profiler_posix.cc"],
-        "//src/conditions:linux_x86_64": ["cpu_profiler_posix.cc"],
+        "//src/conditions:linux": ["cpu_profiler_posix.cc"],
```
alandonovan, line 98: Nice.

Issue #11779, opened 2020-07-15
dummy_link for Windows runfiles can causes rare build failures


tomrenn:
### Description of the problem / feature request:

In rare instances we have seen the following error in builds.

```
build-runfiles error: CreateSymbolicLinkW failed: (error: 183): Cannot create a file when that file already exists.
```

Further investigation found a `dummy_link` file was under a host config tool under `$EXECUTION_ROOT/bazel-out/host/...`. Our builds use dynamic scheduling and `--windows_enable_symlink`, although both of these seem irrelevant to cause the problem. 

Until we remove `dummy_link` all builds on our CI failed with the CreateSymbolicLinkW error. 

Since we use `--windows_enable_symlink`, we guarantee our Windows host is up to date and running with Developer Mode enabled. It should be possible then to avoid the creation of `dummy_link` for determining if symlinking is supported.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

We do not have a clear repro case

### What operating system are you running Bazel on?

Windows Server 2019.

### What's the output of `bazel info release`?

[modified 1.1.0] release

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

We have built a custom version of Bazel 1.1.0 with the following changes cherry-picked:

- https://github.com/bazelbuild/bazel/commit/cffd417a4c4ec4d082f815f57a771ad1010972dd
- https://github.com/bazelbuild/bazel/commit/08a15e795107ec76d574579223485a3efb9da274

tomdegoede:
After upgrading to bazel 3.6 we are encountering this issue as well (at least three times now). I do not recall seeing it before. @meteorcloudy out ci machines are running with just the Create Symlink privilege enabled, not developer mode. I do see the dummy_link file existing on disk.

PR #12428, opened 2020-11-06
Make runfiles-windows.cc not crash on dangling dummy_link

tomdegoede:
Since upgrading to bazel 3.6 we were experiencing similar issues to #11779. Occasionally a dangling dummy_link would exist (I think when the create runfiles process was terminated early) causing the next invocation to crash. This change handles the already exists error and continues execution in case the dummy_link is recognized as a symlink.

### DIFF
--- a/src/main/tools/build-runfiles-windows.cc
+++ b/src/main/tools/build-runfiles-windows.cc
@@ -255,19 +255,20 @@ class RunfilesCreator {
     }
     wstring dummy_link = runfiles_base_dir + L"\\dummy_link";
     wstring dummy_target = runfiles_base_dir + L"\\dummy_target";
-
+    
     // Try creating symlink with admin privilege
-    if (!CreateSymbolicLinkW(dummy_link.c_str(), dummy_target.c_str(), 0)) {
-      // on a rare occasion the dummy_link may exist from a previous run
-      if (GetLastError() == ERROR_ALREADY_EXISTS) {
-        wstring target, werror;
-        // only die when we cannot ensure this is a symlink
-        if (!ReadSymlink(dummy_link, &target, &werror)) {
-          die(L"The dummy_link exists but is not a symlink: %hs", werror.c_str());
-        }
-      }
-      // If we couldn't create symlink, print out an error message and exit.
-      else if (GetLastError() == ERROR_PRIVILEGE_NOT_HELD) {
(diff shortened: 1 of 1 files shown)


src/main/tools/build-runfiles-windows.cc
```diff
@@ -257,25 +257,31 @@ class RunfilesCreator {
...
-          "    2. The system version is Windows 10 Creators Update (1703) or "
-          "later and "
-          "developer mode is enabled.",
-          GetLastErrorString().c_str());
-    } else {
-      die(L"CreateSymbolicLinkW failed: %hs", GetLastErrorString().c_str());
+    if (!CreateSymbolicLinkW(dummy_link.c_str(), dummy_target.c_str(), 0)) {
+      // on a rare occasion the dummy_link may exist from a previous run
+      if (GetLastError() == ERROR_ALREADY_EXISTS) {
+        wstring target, werror;
```
meteorcloudy, line 263: I think the problem happens when we run the same binary multiple times at once. All runs will try to create and delete this dummy_link file. This approach could still fail if another run deleted the `dummy_link` between receiving `ERROR_ALREADY_EXISTS` and `ReadSymlink`. Is there any better way to avoid the race condition? Like adding a random suffix (maybe the [pid](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocessid)?)

tomdegoede: Interesting idea, but why would bazel run the process multiple times? In my experience the dummy_link is actually left behind in the runfiles root. Adding the pid (or any random number) will greatly decrease the risk of encountering such a collision.

meteorcloudy: Oh sorry, I think you are right. Bazel supposes to only run this tool under each runfiles root once in each build. Then this isn't caused by race condition.

Maybe the file got left behind because Bazel somehow exited and terminated the process of build-runfiles-windows before it can delete the dummy_link file? In that case, we can just try to delete the dummy_link file before creating it. Because the logic in this PR actually checks the file created by last run, which might have admin right, but the current run may not have the admin right. WDYT?

tomdegoede: A yes that's another scenario I did not think of. Ill update the PR to reflect your suggestion.

meteorcloudy: Thanks!

meteorcloudy (approved):

PR #12430, opened 2020-11-07
Use exec-transitions in proto rules

Yannic:


Yannic:
@katre PTAL, thanks!

katre:
LGTM but I will defer to @lberki as the owner of proto rules.

lberki:
Let's import this!

lberki (approved):

PR #12435, opened 2020-11-09
Use constraints in global conditions instead of configuration flags (third_party/BUILD).

comius:
This is part of https://github.com/bazelbuild/bazel/pull/12427, that
changes third_party/BUILD file.


philwo:
Merged in https://github.com/bazelbuild/bazel/commit/bf00c3a63c7cffbe90757f580f2b5cf997972b3c

PR #12440, opened 2020-11-10
Remove -no_dead_strip_inits_and_terms ld64 option

keith:
This option is explicitly mentioned as obsolete in the man page. As far
back as Xcode 9 it was a alias for `-dead_strip`:

https://github.com/keith/ld64/blob/40383cc964a1c9993000c2fdd50ecf25802df668/src/ld/Options.cpp#L3090-L3095


allevato (approved):
Indeed, it looks like this flag... has been dead-stripped 😎

PR #12441, opened 2020-11-10
Add predefined configurations to java_toolchain_default

comius:
In order for java_toolchain_default to be more usable, predefined configurations are supplied: `JVM8_TOOLCHAIN_CONFIGURATION`, `JAVABUILDER_TOOLCHAIN_CONFIGURATION`, `VANILLA_TOOLCHAIN_CONFIGURATION`, `PREBUILT_TOOLCHAIN_CONFIGURATION`.

Those configurations are going to be needed once `java_runtime` is attached to `java_toolchain`, because predefined toolchain in `BUILD.java_tools` (e.g. vanilla_toolchain) will need to be removed (as they don't have a default choice of `java_runtime`).

After that the user can use `java_toolchain_default(name="toolchain_vanilla", configuration = VANILLA_TOOLCHAIN_CONFIGURATION, java_runtime = ...)`.
 
Additionally to this, `jvm_opts` parameter is automatically converted to absolute labels. For example from `--patch-module=java.compiler=$(location //:java_compiler_jar)` to `--patch-module=java.compiler=$(location @bazel_java_tools_linux//:java_compiler_jar)`.

Also `java_toolchain_default.bzl` is renamed to `java_toolchain_default.bzl.java_tools` to make it accessible only from @java_tools repository and not @bazel_tools. In the latter it is was never supposed to work.


comius:
cc @lberki @oquenchil @cushon

c-mita (approved):
LGTM

PR #12442, opened 2020-11-10
Retrieve singlejar from java_toolchain instead of using remote_java_tools.

comius:
Android tools depend on using singlejar directly, which is provided in tools/jdk/BUILD via remote_java_tools macro.
Remote_java_tools uses selects to produce an archive name (java_lang_tools_linux, ..) and grab a tool out of it. The selected archive name needs to match the one toolchains will resolve to.

Instead of singlejar is already among things provided by java_toolchain and we can obtain it directly from the toolchain.


comius:
cc @lberki, @c-mita, @oquenchil

PR #12443, opened 2020-11-10
Fix macOS warnings for zstd-jni

keith:
The option used here was gcc specific. This adds the relevant clang
option.

Fixes https://github.com/bazelbuild/bazel/issues/12461

### DIFF
--- a/third_party/zstd-jni/BUILD.bazel
+++ b/third_party/zstd-jni/BUILD.bazel
@@ -10,6 +10,13 @@ filegroup(
     visibility = ["//third_party:__pkg__"],
 )
 
+config_setting(
+    name = "macos",
+    constraint_values = [
+        "@platforms//os:macos",
+    ],
+)
+
 # Below two rules are copied from src/main/native/BUILD.
 # Move to some common place?
 genrule(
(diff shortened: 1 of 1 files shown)


keith:
@philwo

keith:
Ideally this would be conditional on clang instead I guess?

philwo:
Thank you! Happy to merge this when presubmits pass.

Unfortunately I don't know how to use compiler specific cflags. :/ @oquenchil Any idea?

keith:
Envoy uses:

```bzl
config_setting(
    name = "clang_build",
    flag_values = {
        "@bazel_tools//tools/cpp:compiler": "clang",
    },
)
```

But testing this quickly on macOS this doesn't seem to work

oquenchil:
Where is the error log? It looks like all presubmits are passing.

That compiler target is [here](https://cs.opensource.google/bazel/bazel/+/master:tools/cpp/BUILD.tools;drc=7ca5d7d273025882cab03232e10e6bb5dc9c4110;l=469) . By looking at the Starlark implementation of the rule, it looks like it takes the compiler string as is, it might be that on macOS is not just clang but something slightly different.

keith:
There isn't an error when I use that config it just doesn't match on macOS and the warnings are still produced in the console

third_party/zstd-jni/BUILD.bazel
```diff
@@ -10,6 +10,13 @@ filegroup(
     visibility = ["//third_party:__pkg__"],
 )
 
+config_setting(
+    name = "macos",
+    constraint_values = [
+        "@bazel_tools//platforms:osx",
```
Yannic, line 16:
```suggestion
        "@platforms//os:macos",
```

Yannic (approved):

keith:
https://github.com/bazelbuild/bazel/pull/12463

meisterT:
This becomes obsolete after the revert.

Issue #12176, opened 2020-09-25
`--instrument_test_targets` doesn't apply if source rule and test rule are split


keith:
When using `--instrument_test_targets=false` (the default), sources from test target's single dep are not excluded. This affects test targets which do not accept `srcs`.

For example if you have this setup:

```bzl
objc_library(
    name = "lib",
    testonly = True,
    srcs = ...
)

ios_unit_test(
...
    deps = ["lib"]
)
```

The sources from `lib` will still be instrumented even though they are actually test sources.

Ideally there would be some way to exclude these source files from instrumentation (outside of changing the `--instrumentation_filter` which isn't always ergonomic to do) for cases where source rules and test rules are split.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 3.5.0

oquenchil:
I don't see how Bazel could ever know that's the behavior you want without introducing yet another flag. Since the instrumentation filter flag should work for that I will close this bug for now. Feel free to re-open if you come up with a way that wouldn't add much more complexity.

ulfjack:
I think the existing `--instrument_test_targets` flag should *also* consider the `testonly` flag. That seems like it may be sufficiently close to current semantics without too much of a migration period.

keith:
+1 to that option, alternatively a tag could work for this.

I find that working in these detached source and test targets actually need this knowledge quite a bit, for example to compile a bit differently including paths to test dependencies you don't also want to leak into production builds. Currently in rules_apple / rules_swift we do use `testonly` for this.

> Since the instrumentation filter flag should work for that

I was getting at the issue with this above, but what is the solution to having hundreds of modules you want to filter this way? For example a source tree with:

```
a/BUILD
a/srcs
a/tests
b/BUILD
b/srcs
b/tests
```

Where you want to test `a` and `b` would require a target entry for each module, which doesn't scale well. I can see this would be more doable if you had:

```
srcs/a/BUILD
srcs/b/BUILD
tests/a/BUILD
tests/b/BUILD
```

But not all projects will make sense to lay out this way.

keith:
I wonder if another option for this info is to use a `features = ["test_target"]`? The interesting thing about `testonly` is you could have `testonly` targets that you still want coverage info for. In our case this is running the unit tests against a "test helpers" type module. The "test helpers" module is `testonly = True,` but we still want to see coverage for it when running its own unit tests. In this case we don't want to change the semantics by flipping it to `testonly = False`.

[...]

PR #12446, opened 2020-11-10
Remove unused isTestRule for ObjC builds

keith:
This was likely left over from before rules_apple was split out. This
was never set to true.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -176,16 +176,6 @@ public class CompilationSupport {
   private static final String NO_ENABLE_MODULES_FEATURE_NAME = "no_enable_modules";
   private static final String DEAD_STRIP_FEATURE_NAME = "dead_strip";
 
-  /**
-   * Enabled if this target's rule is not a test rule. Binary stripping should not be applied in the
-   * link step. TODO(b/36562173): Replace this behavior with a condition on bundle creation.
-   *
-   * <p>Note that the crosstool does not support feature negation in FlagSet.with_feature, which is
-   * the mechanism used to condition linker arguments here. Therefore, we expose
-   * "is_not_test_target" instead of the more intuitive "is_test_target".
-   */
-  private static final String IS_NOT_TEST_TARGET_FEATURE_NAME = "is_not_test_target";
-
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
```diff
@@ -799,12 +782,6 @@ public Builder doNotUsePch() {
       return this;
     }
 
-    /** Indicates that this CompilationSupport is for use in a test rule. */
-    public Builder setIsTestRule() {
```
keith, line 803: The key to all this being dead was this not being called

lberki:
Hey, this is nice! I'll let @allevato do the actual review, though, because I'm at best a recurrent visitor in our Objective-C rules.

src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
```diff
@@ -894,7 +870,7 @@ public InstrumentedFilesInfo getInstrumentedFilesProvider(ImmutableList<Artifact
         NestedSetBuilder.<Artifact>emptySet(Order.STABLE_ORDER),
         // The COVERAGE_GCOV_PATH environment variable is added in TestSupport#getExtraProviders()
         NestedSetBuilder.<Pair<String, String>>emptySet(Order.COMPILE_ORDER),
-        !isTestRule,
+        true,
```
allevato, line 873: Can you precede this argument with a comment like the one below that indicates the formal argument name?

allevato (approved):
This looks fine if it's not being used, but we should have @googlewalt or @trybka take a look at how the removal of this feature would affect our version of crosstool before it gets pulled in.

src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
keith: Done!

googlewalt:
> This looks fine if it's not being used, but we should have @googlewalt or @trybka take a look at how the removal of this feature would affect our version of crosstool before it gets pulled in.

I filed a bug internally to look into this.  Our internal crosstool still references this, so we need to determine whether the functionality is still useful but broken, or obsolete.

keith:
Potentially related: https://github.com/bazelbuild/bazel/issues/12176

googlewalt:
LGTM.  I will try to merge this.

googlewalt (approved):

Issue #11600, opened 2020-06-16
incompatible_avoid_conflict_dlls: Rename DLL to avoids conflict on Windows


vnvo2409:
Because DLL search paths on Windows are very limited. We choose to copy the DLLs to the binary directory. This cause DLL name conflicts when there are 2 `cc_library` with the same name but different packages (e.g `foo1:bar` and `foo2:bar`). This flag renames the DLLs that are generated by `cc_library` (`NODEPS_DYNAMIC_LIBRARY`) to avoid those conflicts. This flag will be turned on only on Windows and DynamicMode isn't off.

**Flag:** `--incompatible_avoid_conflict_dlls`

Specifically, the flag does the following:
- It calculates a hash based on repository name and package path directory.
- It generated the following file with a hash as suffix:
  + name_{hash}.dll
  + name_{hash}.dll-2.params
  + name_{hash}.pdb (generated by MSVC)
- Other files will be keep untouched
- It also changes the LIBRARY name in the .def file if the feature `windows_export_all_symbols` is turned on.

**Expected Timeline**
Will be flipped in 4.0

vnvo2409:
cc @meteorcloudy

vnvo2409:
@meteorcloudy I will have to make another PR for user provided .def file. I think we shouldn't rename the DLL in that case. what do you think ?

meteorcloudy:
Oh,, you are right. If the DEF file is provided by user then it doesn't have the renamed DLL name.
Nice catch!

meteorcloudy:
You can check Bazel's release plan by https://github.com/bazelbuild/bazel/issues?q=is%3Aopen+is%3Aissue+label%3Arelease

PR #12465, opened 2020-11-12
Disable rename DLL when using win_def_file

vnvo2409:
We should not rename the `*.dll` in `cc_library` when `win_def_file` is specified. Part of #11600 https://github.com/bazelbuild/bazel/issues/11600#issuecomment-645234548 

cc @meteorcloudy

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
@@ -327,11 +327,7 @@ public abstract class CcLibrary implements RuleConfiguredTargetFactory {
     if (ruleContext.getRule().getImplicitOutputsFunction() != ImplicitOutputsFunction.NONE
         || !ccCompilationOutputs.isEmpty()) {
       if (featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
-        Artifact customDefFile = common.getWinDefFile();
-        String dllNameSuffix =
-            customDefFile != null
-                ? CppHelper.getDLLHashSuffix(ruleContext, featureConfiguration)
-                : "";
+        String dllNameSuffix = CppHelper.getDLLHashSuffix(ruleContext, featureConfiguration);
         linkingHelper.setLinkedDLLNameSuffix(dllNameSuffix);
         Artifact generatedDefFile = null;
 
@@ -355,7 +351,7 @@ public abstract class CcLibrary implements RuleConfiguredTargetFactory {
         }
         linkingHelper.setDefFile(
             CppHelper.getWindowsDefFileForLinking(
-                ruleContext, customDefFile, generatedDefFile, featureConfiguration));
+                ruleContext, common.getWinDefFile(), generatedDefFile, featureConfiguration));
       }
       ccLinkingOutputs = linkingHelper.link(ccCompilationOutputs);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
```diff
@@ -327,7 +327,11 @@ public static void init(
     if (ruleContext.getRule().getImplicitOutputsFunction() != ImplicitOutputsFunction.NONE
         || !ccCompilationOutputs.isEmpty()) {
       if (featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
-        String dllNameSuffix = CppHelper.getDLLHashSuffix(ruleContext, featureConfiguration);
+        Artifact customDefFile = common.getWinDefFile();
+        String dllNameSuffix =
+            customDefFile != null
+                ? CppHelper.getDLLHashSuffix(ruleContext, featureConfiguration)
```
meteorcloudy, line 333: Reversed logic? And also, can you put this logic in `getDLLHashSuffix` function? It's used at other places as well.

meteorcloudy (changes requested):
Nice catch! Thanks

src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
vnvo2409: For some reasons, I did not realize the reversed logic. Thank you for pointing it out.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -949,7 +949,8 @@ public static String getDLLHashSuffix(
             ruleContext.getConfiguration().getOptions().get(CppOptions.class));
     if (cppOptions.renameDLL
         && cppOptions.dynamicMode != DynamicMode.OFF
-        && featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
+        && featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)
+        && ruleContext.isAttrDefined("win_def_file", LABEL)) {
```
vnvo2409, line 953: Is this enough for guaranteeing that `ruleContext.getPrerequisiteArtifact("win_def_file")` will not return `null`.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -949,7 +949,8 @@ public static String getDLLHashSuffix(
             ruleContext.getConfiguration().getOptions().get(CppOptions.class));
     if (cppOptions.renameDLL
         && cppOptions.dynamicMode != DynamicMode.OFF
-        && featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)) {
+        && featureConfiguration.isEnabled(CppRuleClasses.TARGETS_WINDOWS)
+        && !ruleContext.isAttrDefined("win_def_file", LABEL)) {
```
vnvo2409, line 953: Is this enough for guaranteeing that `ruleContext.getPrerequisiteArtifact("win_def_file")` will not return `null` ?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
meteorcloudy: I don't think that's enough, you have to check
`&& (!ruleContext.isAttrDefined("win_def_file", LABEL) || ruleContext.getPrerequisiteArtifact("win_def_file") == null)`

Please also put a comment here to explain why we need it.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
meteorcloudy: I don't think that's enough, you have to check
`&& (!ruleContext.isAttrDefined("win_def_file", LABEL) || ruleContext.getPrerequisiteArtifact("win_def_file") == null)`

Please also put a comment here to explain why we need it.

meteorcloudy (changes requested):

meteorcloudy (approved):
Thanks!

PR #12468, opened 2020-11-12
docs: remove stray code fence causing incorrect formatting

mattem:
Removes a stray code fence that causes the block to render incorrectly in https://docs.bazel.build/versions/master/configurable-attributes.html#and-chaining


aiuto (approved):

aiuto:
Thanks.

PR #12478, opened 2020-11-13
Fix action cache checking for missing constant-metadata artifacts.

benjaminp:
There were two bugs:
- getMetadataOrConstant always returned a non-null metadata for constant-metadata artifacts. But, it should return null if the artifact is missing.
- ConstantMetadataValue was indistinguishable from null in a digest map because ConstantMetadataValue's digest was a 0-length array.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
@@ -54,9 +54,6 @@ import javax.annotation.Nullable;
  * otherwise lightweight, and should be constructed anew and discarded for each build request.
  */
 public class ActionCacheChecker {
-  private static final byte[] EMPTY_DIGEST = new byte[0];
-  private static final FileArtifactValue CONSTANT_METADATA = new ConstantMetadataValue();
-
   private final ActionCache actionCache;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -361,11 +358,10 @@ private boolean mustExecute(
...
   private static FileArtifactValue getMetadataOrConstant(
       MetadataHandler metadataHandler, Artifact artifact) throws IOException {
-    if (artifact.isConstantMetadata()) {
-      return CONSTANT_METADATA;
-    } else {
-      return metadataHandler.getMetadata(artifact);
-    }
+    FileArtifactValue metadata = metadataHandler.getMetadata(artifact);
+    return (metadata != null && artifact.isConstantMetadata())
+        ? ConstantMetadataValue.INSTANACE
```
meisterT, line 363: s/INSTANACE/INSTANCE/

benjaminp: thank you—that's pretty embarrasing

meisterT (approved):

PR #9000, opened 2019-07-28
Expose proto fragment fields necessary for writing proto_toolchain to Starlark

Yannic:
This change exposes some of the Protobuf-related command-line flags to Starlark.

There is a risk of adding these fields, so we gave them a scary name and documented how users are supposed to access them. Hopefully, this will be enough to fight Hyrum's Law.

Blocking #10006

Working towards #10005 

Design doc:
https://docs.google.com/document/d/1u95vlQ1lWeQNR4bUw5T4cMeHTGJla2_e1dHHx7v4Dvg/edit#

PR #12484, opened 2020-11-14
Create proto_toolchain rule

Yannic:
This change adds a new `proto_toolchain` rule which we will use later to resolve
tools and options required to compile Protocol Buffers, thus eventually replacing
some of the command-line options from `ProtoConfiguration`.

To preclude inconsistencies in the migration period, we will start with a
`proto_toolchain` rule that doesn't have any arguments and populate
its provider from the existing command-line options. After all consumers
have been migrated to resolve the tools from the toolchain instead of
reading the command-line, we will replace all proto-specific flags with
attributes on `proto_toolchain` (or similar, language-specific toolchains
for flags that only affect a specific language).

Updates #10005


Yannic:
@comius or @lberki PTAL, thanks!

comius:
Since we have a long term plan to write everything in Starlark, it might make more sense to implement this already in Starlark in the rules_proto repository. See for example rules Scala which alread uses Starlark https://github.com/bazelbuild/rules_scala/blob/master/scala/scala_toolchain.bzl

Yannic:
I did choose to implement this as a native rule for two reasons:
- It avoids exposing the porto-related command-line flags to Starlark that would tie the `proto_toolchain` rule to specific versions of Bazel that support these fields and that we do not want anyone to depend on, and
- We most likely need at least a new native Provider to migrate the native protobuf rules to use the toolchain.

I absolutely do plan to rewrite this new rule in Starlark as soon as the command-line flags are removed.

comius:
> I did choose to implement this as a native rule for two reasons:
> 
> * It avoids exposing the porto-related command-line flags to Starlark that would tie the `proto_toolchain` rule to specific versions of Bazel that support these fields and that we do not want anyone to depend on, and

Can you be a bit more specific. Are you talking about ProtoConfiguration flags?

I'll go forward with the review, but I would still like to know about this.

> * We most likely need at least a new native Provider to migrate the native protobuf rules to use the toolchain.

I also thought so and I would accept a new native Provider, however if you take a look at scala toolchain, there is a case where they don't need a new provider and can extend it with arbitrary attributes. 

@lberki, @katre is Scala using ToolchainInfo provider (https://github.com/bazelbuild/rules_scala/blob/master/scala/scala_toolchain.bzl#L51) in a good way by ad hoc extending it or is this going to create another technical debt?

> I absolutely do plan to rewrite this new rule in Starlark as soon as the command-line flags are removed.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchainRule.java
```diff
@@ -0,0 +1,67 @@
...
+                    Attribute.LabelLateBoundDefault.fromTargetConfiguration(
+                        ProtoConfiguration.class,
+                        defaultProtoCompiler,
+                        (rule, attributes, protoConfig) ->
+                            protoConfig.protoCompiler() != null
+                                ? protoConfig.protoCompiler()
+                                : defaultProtoCompiler)))
+        .requiresConfigurationFragments(ProtoConfiguration.class)
+        .advertiseProvider(ProtoToolchainInfo.class)
+        .build();
```
comius, line 56: Why don't you add Label defaultProtoCompiler as an attribute to the rule and remove it from the constructor? You can specify default value on the attribute or perhaps even better leave it blank and set it in the BUILD file.

It looks like you invested some effort to provide a design where ProtoToolchainRule can be modified by other, that is creating an abstract BazelProtoToolchainRule, however I think it is better to keep it simple without an additional abstract class. Toolchain rules usually just need to pass the data thru, and don't have much logic that would need changing. (This is different to BazelProtoLibraryRule)

comius (changes requested):
Nicely done, thanks. But I still think, the whole PR can already be done in Starlark.

Yannic:
> > I did choose to implement this as a native rule for two reasons:
> > 
> > * It avoids exposing the porto-related command-line flags to Starlark that would tie the `proto_toolchain` rule to specific versions of Bazel that support these fields and that we do not want anyone to depend on, and
> 
> Can you be a bit more specific. Are you talking about ProtoConfiguration flags?

Yes, I'm talking about ProtoConfiguration options. I did experiment with exposing the fields on the fragment and writing the rule in Starlark in https://github.com/bazelbuild/bazel/pull/9000, but it wasn't simpler than implementing it as native rule (and I never came to try out how exactly to use the Starlark toolchain from native rules). I think having this as a native rule for now is the better option.

> 
> I'll go forward with the review, but I would still like to know about this.
> 
> > * We most likely need at least a new native Provider to migrate the native protobuf rules to use the toolchain.
> 
> I also thought so and I would accept a new native Provider, however if you take a look at scala toolchain, there is a case where they don't need a new provider and can extend it with arbitrary attributes.
> 
> @lberki, @katre is Scala using ToolchainInfo provider (https://github.com/bazelbuild/rules_scala/blob/master/scala/scala_toolchain.bzl#L51) in a good way by ad hoc extending it or is this going to create another technical debt?

AFAIK, ToolchainInfo is the recommended way to create toolchains in Starlark and would work if we only read the values from Starlark. However, there are native rules that would also need to access the fields of the toolchain. It's still possible, but it doesn't feel right. Once everything is migrated to Starlark, we can switch to ToolchainInfo.

> 
> > I absolutely do plan to rewrite this new rule in Starlark as soon as the command-line flags are removed.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchainRule.java
Yannic: The abstraction exists so that Google can create an internal `proto_toolchain` rule that uses a different label for `protoc`. If you can achieve the same thing with Copybara transformations, I'm happy to remove the abstraction :).

comius: Would that be the case if the label was only specified on an attribute? I expect there to be a different BUILD file at Google internally than the new WORKSPACE suffix.

Yannic: Yes, there needs to be an internal version of what is in `tools/proto/BUILD.tools` and the new toolchain must be registered in Google's WORKSPACE file.

Note that there are currently no settable attributes on `proto_toolchain`, everything is read from the configuration fragment. Thinking on this a bit more, it's probably really is fine to just have 1 `proto_toolchain` that is identical for Bazel and Blaze. I'll update the PR.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchainInfo.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
+import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
+import com.google.devtools.build.lib.starlarkbuildapi.proto.ProtoToolchainInfoApi;
+
+/**
+ * Information about the tools used by the <code>proto_*</code> and <code>LANG_proto_*</code> rules.
+ */
+@Immutable
+@AutoCodec
+public class ProtoToolchainInfo extends ToolchainInfo implements ProtoToolchainInfoApi<FilesToRunProvider> {
```
katre, line 30: What we've learned is that extending ToolchainInfo isn't a good idea: Starlark doesn't have the concept of inheritance and so using it for native providers leads to a mismatch.

Instead, this class should just be "ProtoToolchainInfo extends NativeInfo", and in the actual ProtoToolchain rule implementation, create a new ToolchainInfo that contains the ProtoToolchainInfo:
```
ProtoToolchainInfo protoToolchainInfo = new ProtoToolchainInfo(...);
ToolchainInfo toolchainInfo = new ToolchainInfo(ImmutableMap.of("protoToolchain", protoToolchainInf);

...

ruleBuilder
    .addProvider(protoToolchainInfo)
    .addProvider(toolchainInfo)
    .build()
```

Then, rules that require the toolchain can get the ToolchainInfo and find the ProtoToolchainInfo by calling:
```
ToolchainInfo toolchainInfo = ruleContext.getToolchainContext().forToolchainType(toolchainType);
ProtoToolchainInfo protoToolchainInfo = (ProtoToolchainInfo) toolchainInfo.get("protoToolchain");
```

(I'm translating this from the Starlark version, so my syntax may not be 100% correct).

Yannic:
Rewrote this CL to expose the fragment fields that are necessary for writing `proto_toolchain` in Starlark.

I also have some tests to verify that this is working as intended, but they depend on bazelbuild/rules_proto#82, so I'd like to wait until thet PR lands and we roll `@rules_proto` if that's ok

@comius PTAL

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/proto/ProtoCommonApi.java
```diff
@@ -31,4 +33,16 @@
...
+      name = "available_configuration_fields",
+      doc =
+          "Indicates that this version exposes fields on this configuration fragment. Do not use "
+              + "this field, its only purpose is to help with migration.",
+      documented = false,
+      structField = true)
+  default ImmutableList<String> availableConfigurationFields() {
+    return ImmutableList.of("protoc", "protoc_options");
+  }
+}
```
comius, line 48: Please see my comments in https://github.com/bazelbuild/rules_proto/pull/82

comius (changes requested):
Everything else seems fine.

PR #12486, opened 2020-11-14
Accept proto_library as //tools/objc:protobuf_well_known_types

Yannic:
Part of the migration of Protobuf rules/aspects to `ProtoSource`.
A future change will make `ProtoInfo` mandatory (similar to what was
done for `proto_lang_toolchain`).

Note that this is for Obj-C Protobuf rules which don't exist in Bazel.

Updates #10005


lberki (approved):

lberki:
Let's see what the presubmit checks say!

PR #12491, opened 2020-11-16
Remove jarjar from java_tools.

comius:
Jarjar is an advanced processor of jar files. It can rename classes consistently within a .jar file. It is used in Jacoco deployment, renaming package of 'asm' dependency in the Jacoco deployment jar. This is done so that 'asm' does not clash with another version of asm that might be used in the target being "code coveraged".

After the deployment of java_tools, jarjar is not used by Bazel itself. There might be downstream projects using jarjar from java_tools. However it is a much better idea to remove jarjar from java_tools interface, and that downstream projects depend on it directly.


comius:
Downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1766

comius:
cc @lberki @oquenchil

PR #12492, opened 2020-11-16
Reenabled jacoco agent test.

comius:


comius:
cc @oquenchil @lberki

PR #12494, opened 2020-11-16
Stop profilers in clean command test.

meisterT:
This became necessary after d5f96eb44a1fd70048a6bfb42b3e0f308b2dcb88
changed how we handle crashes.

```
bazel test --test_output=streamed --test_filter=CleanCommandRecommendsAsyncTest //src/test/java/com/google/devtools/build/lib:runtime-tests
```
passes now.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/BlazeCommandDispatcher.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/BlazeCommandDispatcher.java
@@ -44,6 +44,7 @@ import com.google.devtools.build.lib.events.ExtendedEventHandler.Postable;
 import com.google.devtools.build.lib.events.PrintingEventHandler;
 import com.google.devtools.build.lib.events.Reporter;
 import com.google.devtools.build.lib.events.StoredEventHandler;
+import com.google.devtools.build.lib.profiler.MemoryProfiler;
 import com.google.devtools.build.lib.profiler.Profiler;
 import com.google.devtools.build.lib.profiler.SilentCloseable;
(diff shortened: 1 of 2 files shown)


justinhorvitz (approved):
Had to do this in BlazeCommandDispatcherTest too, which I am assuming you saw because the comment matches.

src/test/java/com/google/devtools/build/lib/runtime/commands/CleanCommandRecommendsAsyncTest.java
```diff
@@ -55,6 +58,13 @@ public CleanCommandRecommendsAsyncTest(List<String> commandLine, OS os, boolean
     this.expectSuggestion = expectSuggestion;
   }
 
+  @After
+  public void stopProfilers() throws Exception {
```
justinhorvitz, line 62: Actually, now that I look closer, this test is not simulating crashes. It just happens to be failing with a crash, which then poisons subsequent test cases. I suspect that if you fix or comment out the already failing case, then the other cases past.

A better solution to prevent this might be to move https://github.com/bazelbuild/bazel/blob/e6751ed25615e0ce6e5ecee68dd6010b2de2225c/src/main/java/com/google/devtools/build/lib/runtime/BlazeRuntime.java#L680-L685 to https://github.com/bazelbuild/bazel/blob/e6751ed25615e0ce6e5ecee68dd6010b2de2225c/src/main/java/com/google/devtools/build/lib/runtime/BlazeCommandDispatcher.java#L624 (also then need to add it to BlazeRuntimeWrapper).

meisterT:
Thanks for that suggestion. Done. I'll look into the crash in the test after this is merged.

PR #12500, opened 2020-11-17
Fix packing jacoco sources.

comius:


philwo:
Merged in https://github.com/bazelbuild/bazel/commit/a893086c6845f819a5313cca2058c98a67fedbcd.

PR #12507, opened 2020-11-18
Report digest of failed uploads

Yannic:
This will make the error message more useful because otherwise, there
was no way of telling which upload failed or timed out (other than
running with `-j 1`)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
@@ -84,7 +84,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
 
   /** Contains the hash codes of already uploaded blobs. **/
   @GuardedBy("lock")
-  private final Set<Digest> uploadedBlobs = new HashSet<>();
+  private final Set<HashCode> uploadedBlobs = new HashSet<>();
 
   @GuardedBy("lock")
   private final Map<Digest, ListenableFuture<Void>> uploadsInProgress = new HashMap<>();
@@ -201,6 +201,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
   }
 
   @Deprecated
+  @VisibleForTesting
   public ListenableFuture<Void> uploadBlobAsync(
       HashCode hash, Chunker chunker, boolean forceUpload) {
     Digest digest =
@@ -229,7 +230,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
     synchronized (lock) {
       checkState(!isShutdown, "Must not call uploadBlobs after shutdown.");
 
-      if (!forceUpload && uploadedBlobs.contains(digest)) {
+      if (!forceUpload && uploadedBlobs.contains(HashCode.fromString(digest.getHash()))) {
         return Futures.immediateFuture(null);
       }
 
@@ -243,7 +244,7 @@ class ByteStreamUploader extends AbstractReferenceCounted {
               startAsyncUpload(digest, chunker),
               (v) -> {
                 synchronized (lock) {
-                  uploadedBlobs.add(digest);
+                  uploadedBlobs.add(HashCode.fromString(digest.getHash()));
                 }
                 return null;
               },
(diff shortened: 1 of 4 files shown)


Yannic:
@coeuvre PTAL

/cc @ulfjack

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -83,10 +84,10 @@
 
   /** Contains the hash codes of already uploaded blobs. **/
   @GuardedBy("lock")
-  private final Set<HashCode> uploadedBlobs = new HashSet<>();
+  private final Set<Digest> uploadedBlobs = new HashSet<>();
```
ulfjack, line 87: This may have used HashCode because it's more compact. I think you are changing this to get the size as well. The downside is that this could increase memory consumption by 2x.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -244,14 +253,20 @@ void shutdown() {
           Futures.catchingAsync(
               uploadResult,
               StatusRuntimeException.class,
-              (sre) -> Futures.immediateFailedFuture(new IOException(sre)),
+              (sre) -> Futures.immediateFailedFuture(
+                  new IOException(
+                      String.format(
+                          "Error while uploading artifact with digest '%s/%s'",
```
ulfjack, line 259: It's an improvement, but it would be even better if it gave a filename.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
```diff
@@ -268,24 +283,25 @@ boolean uploadsInProgress() {
...
+  private ListenableFuture<Void> startAsyncUpload(Digest digest, Chunker chunker) {
     try {
       chunker.reset();
     } catch (IOException e) {
       return Futures.immediateFailedFuture(e);
     }
 
     UUID uploadId = UUID.randomUUID();
-    String resourceName = buildUploadResourceName(instanceName, uploadId, hash, chunker.getSize());
+    String resourceName = buildUploadResourceName(instanceName, uploadId, digest);
```
coeuvre, line 304: How about check that whether `digest.getSizeBytes()` is the same as `chunker.getSize()` at the beginning?

coeuvre (approved):
Thanks for the improvement! Please take a look at comments and I will import afterwards.

BTW, I am working on refactoring remote module with RxJava to improve readability and maintainability. [`RxByteStreamUploader`](https://github.com/coeuvre/bazel/blob/remote-with-rx/src/main/java/com/google/devtools/build/lib/remote/RxByteStreamUploader.java) is the first step and is almost done.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
Yannic: Absolutely! However, we won't get a filename here because the blob may not be a named file (e.g., `std{out,err}`). We would need to do that further downstream.

Let's start with the hash and I'll send a followup if I can figure out how to create an error message that contains the name.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
Yannic: Switched `uploadedBlobs´ back to `HashCode`.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
Yannic: Done

coeuvre:
Thanks! Importing.

src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
coeuvre: Maybe catch the exception and add filename at the call site. e.g. https://github.com/coeuvre/bazel/blob/remote-with-rx/src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java#L396

PR #12509, opened 2020-11-18
Update Proguard from 5.3.3 -> 6.2.2

Bencodes:
The baked in version of proguard hasn't been update in quite some time.

https://www.guardsquare.com/en/products/proguard/manual/versions


Bencodes:
@jin mind taking a look at this?

Bencodes:
@ahumesky any thoughts here?

ahumesky (commented):
Thanks!
Could you add a `README.bazel.md` to third_party/java/proguard that says where the jar comes from? It looks like it's from https://search.maven.org/artifact/net.sf.proguard/proguard-base/6.2.2/jar

ahumesky:
Could you also add into that README.bazel.md a link to where the source came from (is it https://github.com/Guardsquare/proguard/commit/ebe9000c9baa36a53dc63902b96956c9525ab052 ?)

Bencodes:
Added a quick README with some links

Bencodes:
@ahumesky do those changes look good?

ahumesky (approved):
Looks good, thanks! I'll merge this in now.

ahumesky:
Hmm, so after looking at merging this, it appears that this is going to be tricky to submit. Due to how our code syncing works across multiple repos, we can't merge pull requests that contain changes to the third_party directory and other directories. Usually this just means having to separate the changes to third_party in their own change, submit that first, then do the other changes. But doing that here would break the tests because the switch to 6.2.2 happens in `third_party/java/proguard/BUILD`. And we can't make the change to the build file along with the tests, because then we're back to the original problem....

So, the best thing to do here is probably to modify the tests to accept either 5.3.3 or 6.2.2 (and either "GPL.html" or "GPL.md") in one change, then submit all the changes to third_party by themselves in a separate change, and then a final change to have the tests accept only 6.2.2.

Also, I don't have the permissions to merge changes into third_party, so we'll need to get @philwo or someone to merge in those changes.

Bencodes:
The first two PRs are here
- https://github.com/bazelbuild/bazel/pull/12909
- https://github.com/bazelbuild/bazel/pull/12910

Bencodes:
@ahumesky everything should be ready to go

PR #12514, opened 2020-11-18
Take no action to prefetch empty artifacts.

benjaminp:
Previously, Bazel would attempt to overwrite /dev/null.

PR #12516, opened 2020-11-19
Don't ever claim /dev/null is an execpath.

benjaminp:
SpawnInputExpander.EMPTY_FILE claimed an execpath of /dev/null, which could lead to clients trying to do mutating filesystem operations in /dev.

In this change, SpawnInputExpander.EMPTY_FILE is replaced with an EmptyVirtualInput with no execpath. This way errant callers will explode before trying filesystem operations on it. Such a EmptyVirtualInput doesn't really meet the contract of being an ActionInput but that was really already true.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/cache/VirtualActionInput.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/cache/VirtualActionInput.java
@@ -13,7 +13,6 @@
 // limitations under the License.
 package com.google.devtools.build.lib.actions.cache;
 
-import com.google.common.base.Preconditions;
 import com.google.devtools.build.lib.actions.ActionInput;
 import com.google.devtools.build.lib.actions.FileArtifactValue;
 import com.google.devtools.build.lib.util.StreamWriter;
@@ -27,6 +26,11 @@ import java.io.OutputStream;
  * OutputStream.
  */
 public interface VirtualActionInput extends ActionInput, StreamWriter {
+  /**
+   * An empty virtual artifact <b>without</b> an execpath. This is used to denote empty files in
+   * runfiles and filesets.
+   */
+  public static final VirtualActionInput EMPTY_MARKER = new EmptyActionInput();
 
   /**
    * Gets a {@link ByteString} representation of the fake file. Used to avoid copying if the fake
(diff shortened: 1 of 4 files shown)


benjaminp:
This is a cleanup related to #12514 but split out because it might break blaze due to fileset things I can't see.

src/main/java/com/google/devtools/build/lib/actions/cache/VirtualActionInput.java
```diff
@@ -48,7 +54,11 @@ default FileArtifactValue getMetadata() throws IOException {
    * use instances of this class to represent those files.
    */
   final class EmptyActionInput implements VirtualActionInput {
-    private final PathFragment execPath;
+    @Nullable private final PathFragment execPath;
+
+    private EmptyActionInput() {
+      execPath = null;
```
justinhorvitz, line 60: Instead of setting the field to null and handling that case in all of the methods, it would be easier to just have a private class for the empty marker singleton.

justinhorvitz (changes requested):

src/main/java/com/google/devtools/build/lib/actions/cache/VirtualActionInput.java
benjaminp: Yes, that sounds better. I didn't know if this was the only use of `EmptyActionInput` or if some Google code makes real execpath empty inputs.

justinhorvitz: There is one internal use of `EmptyActionInput` that also passes `/dev/null`. I suggest keeping this class as it was for now and just create a new `VirtualActionInput` implementation class for the "null" case. I will separately see if the internal use case can be migrated similarly and if so follow up with a change to delete this class.

benjaminp: Okay, reverting to the original version of the PR now...

justinhorvitz (approved):

aiuto:
@justinhorvitz sending you the import CL. I'm on triage duty this week and trying to clean the backlog of stale PRs.

aiuto:
Had to roll this back.  I'm working on relanding it.

benjaminp:
Thank you for sorting that out.

aiuto:
You're welcome.  I forgot, however, to munge the re-landing to give the attribution to you. Sorry about that. It's a pity the tooling does not do this automatically.

PR #12523, opened 2020-11-19
Add rxjava3 to third_party

coeuvre:
We are going to refactor remote module with RxJava. This PR add rxjava3 to third_party so we can use it later.


meteorcloudy:
/cc @olekw
Hey Olek, we are planning to use rxjava3 to improve Bazel's remote execution code. It's not yet packaged for Debian, but it has almost no additional dependencies. Can you please verify if it's ok to package this for future Bazel version in Debian?
https://github.com/ReactiveX/RxJava
https://mvnrepository.com/artifact/io.reactivex.rxjava3/rxjava/3.0.7

olekw:
> Can you please verify if it's ok to package this for future Bazel version in Debian?

Ugh, a Gradle build... But, from a quick look, it does look like it only has one real dependency (already in Debian) and I should be able to just delete most of the extraneous Gradle build stuff. I'll try to work on that in the next couple days, @meteorcloudy. :)

olekw:
Ok, packaged and in the NEW queue.

philwo:
Thank you @olekw!! 😊

meteorcloudy:
Amazing, thank you, Olek!

olekw:
Looks like a nice piece of software, glad I could help bring it to Debian. :)

Issue #12524, opened 2020-11-19
Provide the ability to interrupt certain bazel commands


daveyc123:
### Description of the problem / feature request:

The bazel server has a single lock that ensures that multiple commands do not run concurrently. When a command is run it obtains the lock and holds on to the lock until it completes (see BlazeCommandDispatcher#exec). If a second command is run and the first command is running, the second command will wait until the first command completes and the lock is released before it runs - "Another command (".... ") is running. Waiting for it to complete on the server..." There are a few GitHub issues raised related to this issue (such as https://github.com/bazelbuild/bazel/issues/532 and https://github.com/bazelbuild/bazel/issues/1541). 

Unfortunately, this model poses problems for tools who wish to present information about a user's project, but not interfere with their traditional workflows.

Consider the case where a visual tool provides information regarding the available targets for a given bazel project. When the tool opens, it executes bazel query to find the list of targets so that it can present them to the user. For large projects and those with many large externals, this first query can take a long time. Unfortunately, during this time, if the user were to do a bazel build, they'll receive a message indicating a bazel command is already running. It's not obvious to them what is happening under the hood.

Ideally, I'd like the ability to indicate that a bazel invocation is "interruptible". If another command is subquently run, the interruptible command would stop and the subsequent command would run. This would allow tools who leverage bazel "under the hood" to mark their commands as "interruptible" so that they don't interfere with other workflows. They could then re queue their bazel invocations.

### What's the output of `bazel info release`?
release 3.5.0

daveyc123:
This is my first foray into the bazel codebase. I'd like to contribute a PR for this issue, and have started to put some code together that can accomplish this by updating CommandManager#RunningCommand to have an "interruptible" flag. GrpcServerImpl#executeCommand would then ask the command manager to interrupt all interruptible commands when a new command is run. That part is fairly straight forward.

My question is how to plumb in the --interruptible flag. Currently, GrpcServerImpl#executeCommand doesn't parse out the command options (it only has the startup options). Looking at the other start up options, I don't think --interruptible makes sense there.

The BlazeCommandDispatcher is where the command options are parsed. At the moment though it doesn't have access to RunningCommand. It could in some fashion but that seems leaky.

@ulfjack - your name seems to come up in a lot of the blame lines in the relevant code. Any thoughts on the above?

michajlo:
[...]

PR #12526, opened 2020-11-20
Add ability to interrupt commands

daveyc123:
Add the ability to mark an invocation of a command as interruptible. If another command is executed while this command is running, the command will be interrupted.

### DIFF
--- a/src/main/cpp/startup_options.h
+++ b/src/main/cpp/startup_options.h
@@ -245,6 +245,9 @@ class StartupOptions {
   // Whether to output addition debugging information in the client.
   bool client_debug;
 
+  // Whether the resulting command will be preempted if a subsequent command is run.
+  bool preemptible;
+
   // Value of the java.util.logging.FileHandler.formatter Java property.
   std::string java_logging_formatter;
 
--- a/src/main/java/com/google/devtools/build/lib/server/CommandManager.java
+++ b/src/main/java/com/google/devtools/build/lib/server/CommandManager.java
@@ -43,6 +43,24 @@ class CommandManager {
     idle();
   }
 
+  void preemptEligibleCommands() {
+    synchronized (runningCommandsMap) {
+      ImmutableSet.Builder<String> commandsToInterruptBuilder = new ImmutableSet.Builder<>();
+
+      for (RunningCommand command : runningCommandsMap.values()) {
+        if (command.isPreemptible()) {
+          command.thread.interrupt();
+          commandsToInterruptBuilder.add(command.id);
+        }
+      }
+
+      ImmutableSet<String> commandsToInterrupt = commandsToInterruptBuilder.build();
+      if (!commandsToInterrupt.isEmpty()) {
+        startSlowInterruptWatcher(commandsToInterrupt);
+      }
+    }
+  }
+
   void interruptInflightCommands() {
     synchronized (runningCommandsMap) {
       for (RunningCommand command : runningCommandsMap.values()) {
@@ -54,7 +72,7 @@ class CommandManager {
   }
 
   void doCancel(CancelRequest request) {
-    try (RunningCommand cancelCommand = create()) {
+    try (RunningCommand cancelCommand = createCommand()) {
       synchronized (runningCommandsMap) {
         RunningCommand pendingCommand = runningCommandsMap.get(request.getCommandId());
         if (pendingCommand != null) {
@@ -88,8 +106,19 @@ class CommandManager {
     }
   }
 
-  RunningCommand create() {
-    RunningCommand command = new RunningCommand();
+  RunningCommand createPreemptibleCommand() {
+    RunningCommand command = new RunningCommand(true);
+    registerCommand(command);
+    return command;
+  }
+
+  RunningCommand createCommand() {
+    RunningCommand command = new RunningCommand(false);
+    registerCommand(command);
+    return command;
+  }
+
+  private void registerCommand(RunningCommand command) {
     synchronized (runningCommandsMap) {
       if (runningCommandsMap.isEmpty()) {
         busy();
@@ -98,7 +127,6 @@ class CommandManager {
       runningCommandsMap.notify();
     }
     logger.atInfo().log("Starting command %s on thread %s", command.id, command.thread.getName());
-    return command;
   }
 
   private void idle() {
@@ -148,10 +176,12 @@ class CommandManager {
   class RunningCommand implements AutoCloseable {
     private final Thread thread;
     private final String id;
+    private final boolean preemptible;
 
-    private RunningCommand() {
+    private RunningCommand(boolean preemptible) {
       thread = Thread.currentThread();
       id = UUID.randomUUID().toString();
+      this.preemptible = preemptible;
     }
 
     @Override
@@ -169,6 +199,10 @@ class CommandManager {
 
     String getId() {
       return id;
+    }
+
+    boolean isPreemptible() {
+      return this.preemptible;
     }
   }
 }
--- a/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
+++ b/src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
@@ -512,7 +512,9 @@ public class GrpcServerImpl extends CommandServerGrpc.CommandServerImplBase impl
           option.getOption().toString(StandardCharsets.ISO_8859_1)));
     }
 
-    try (RunningCommand command = commandManager.create()) {
+    commandManager.preemptEligibleCommands();
+
+    try (RunningCommand command = request.getPreemptible() ? commandManager.createPreemptibleCommand() : commandManager.createCommand()) {
       commandId = command.getId();
 
       try {
@@ -608,7 +610,7 @@ public class GrpcServerImpl extends CommandServerGrpc.CommandServerImplBase impl
 
   @Override
   public void ping(PingRequest pingRequest, StreamObserver<PingResponse> streamObserver) {
-    try (RunningCommand command = commandManager.create()) {
+    try (RunningCommand command = commandManager.createCommand()) {
       PingResponse.Builder response = PingResponse.newBuilder();
       if (pingRequest.getCookie().equals(requestCookie)) {
         response.setCookie(responseCookie);
--- a/src/main/protobuf/command_server.proto
+++ b/src/main/protobuf/command_server.proto
@@ -58,6 +58,10 @@ message RunRequest {
   // came from. These options have already been parsed and already have had
   // their effect. This information should only be used for logging.
   repeated StartupOption startup_options = 6;
+
+  // Whether the resulting command can be preempted if additional commands
+  // are received.
+  bool preemptible = 7;
 }
 
 // Contains the a startup option with its source file. Uses bytes to preserve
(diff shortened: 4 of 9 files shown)


daveyc123:
@googlebot I signed it!

src/main/java/com/google/devtools/build/lib/server/CommandManager.java
```diff
@@ -88,8 +90,8 @@ void waitForChange(long timeout) throws InterruptedException {
     }
   }
 
-  RunningCommand create() {
-    RunningCommand command = new RunningCommand();
+  RunningCommand create(boolean interuptable) {
```
michajlo, line 93: Same as above wrt boolean params.

michajlo (commented):
Some preliminary comments. LMK when you finish up tests and all.

daveyc123:
Thanks for the feedback @michajlo. I've updated accordingly and added a server test.

src/main/cpp/startup_options.h
```diff
@@ -245,6 +245,9 @@ class StartupOptions {
   // Whether to output addition debugging information in the client.
   bool client_debug;
 
+  // Whether the resulting command will be preempted if a subsequent command is run
+  bool preemptible;
```
michajlo, line 249: nit: s/will be/can be/ - being noncommittal lets us tweak implementation if we need. Same for other places it's documented.

Also, add a period at the end of the comment for consistency (at least with the other comments i see in this diff)

src/main/java/com/google/devtools/build/lib/server/CommandManager.java
```diff
@@ -43,6 +45,23 @@
     idle();
   }
 
+  void preemptEligibleCommands() {
+    synchronized (runningCommandsMap) {
+      Set<String> commandsToInterrupt = new HashSet<>();
```
michajlo, line 50: Use ImmutableSet.Builder instead - no need for the intermediate hashset

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
```diff
@@ -512,7 +512,9 @@ private void executeCommand(RunRequest request, BlockingStreamObserver<RunRespon
           option.getOption().toString(StandardCharsets.ISO_8859_1)));
     }
 
-    try (RunningCommand command = commandManager.create()) {
+    commandManager.preemptEligibleCommands();
+
+    try (RunningCommand command = request.getPreemptible() ? commandManager.createPreemptibleCommand() : commandManager.createCommand()) {
```
michajlo, line 517: ok, i guess i can see how the two methods thing i mentioned can be awkward, but i think it's still a net win (no action required here btw)

src/main/protobuf/command_server.proto
```diff
@@ -58,6 +58,10 @@ message RunRequest {
   // came from. These options have already been parsed and already have had
   // their effect. This information should only be used for logging.
   repeated StartupOption startup_options = 6;
+
+  // Whether the resulting command should be preempted if additional commands
```
michajlo, line 62: s/should/can/ - calling this one out explicitly since it's in the grpc interface. Also, don't forget '.' at the end of the sentence.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,101 @@ public void onCompleted() {
...
+              throws InterruptedException {
+            if (args.contains(firstCommandArg)) {
+              OutputStream out = outErr.getOutputStream();
+              try {
+                while (true) {
+                  if (Thread.interrupted()) {
+                    return BlazeCommandResult.failureDetail(
+                        FailureDetail.newBuilder()
+                            .setInterrupted(
+                                Interrupted.newBuilder().setCode(Code.INTERRUPTED_UNKNOWN))
```
michajlo, line 501: use the regular old INTERRUPTED code (without _UNKNOWN)

michajlo (commented):
Can we document and test for the expected behavior when more than one command is present?

src/main/java/com/google/devtools/build/lib/server/GrpcServerImpl.java
daveyc123: I agree. It's definitely more explicit, even at the cost of a bit of duplicated code. These sort of scenarios make me appreciate keyword args in other languages.

daveyc123:
> Can we document and test for the expected behavior when more than one command is present?
Sure. Just to confirm though, which case are you referring to for docs? Where there are multiple commands with --preemptible set? Or the current behaviour when --preemptible is not set? Not sure if the later should be doc'ed in the --preemptible flag.

Either way, I'll add tests for both scenarios.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+              LockingMode lockingMode,
+              String clientDescription,
+              long firstContactTimeMillis,
+              Optional<List<Pair<String, String>>> startupOptionsTaggedWithBazelRc)
+              throws InterruptedException {
+            if (args.contains(firstCommandArg)) {
+              OutputStream out = outErr.getOutputStream();
+              try {
+                while (true) {
+                  if (Thread.interrupted()) {
```
michajlo, line 497: Here and others - can you just sleep and catch InterruptedException instead of busy waiting/writing? I don't think the `out.write`s are necessary for the test. Should simplify the code and give potentially under powered CI machines a break.

```
while (true) {
    try {
      Thread.sleep(TestUtils.WAIT_TIMEOUT_WHATEVER);
    } catch (InterruptedException e) {
       return BlazeCommandResult...
    }
}
```

I

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+          public BlazeCommandResult exec(
+              InvocationPolicy invocationPolicy,
+              List<String> args,
+              OutErr outErr,
+              LockingMode lockingMode,
+              String clientDescription,
+              long firstContactTimeMillis,
+              Optional<List<Pair<String, String>>> startupOptionsTaggedWithBazelRc)
+              throws InterruptedException {
+            if (args.contains(firstCommandArg)) {
```
michajlo, line 693: i'm probably missing something, but the if and else blocks look awfully similar, can they be combined?

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+    // Wait for the first command to startup
+    gotFoo.await();
+
+    stub.run(
+        createRequest(secondCommandArg),
+        new StreamObserver<RunResponse>() {
+          @Override
+          public void onNext(RunResponse value) {
+            if (value.getFinished()) {
+              finishedBar.countDown();
```
michajlo, line 763: should this go in onCompleted or onError? (see previous comment about this)

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+
+          @Override
+          public void onError(Throwable t) {}
+
+          @Override
+          public void onCompleted() {}
+        });
+
+    gotBar.await();
+    server.shutdown();
```
michajlo, line 558: instead of relying on the side-effects of shutdown would it be cleaner to wait on the commands finishing (via latches in onError/onCompleted). I *think* this will be easier to follow than relying on the side-effect of shutdown to be an orderly shutdown.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+    assertThat(lastFooResponse.get().getFinished()).isTrue();
+    assertThat(lastFooResponse.get().getExitCode()).isEqualTo(8);
+    assertThat(lastFooResponse.get().hasFailureDetail()).isTrue();
+    assertThat(lastFooResponse.get().getFailureDetail().hasInterrupted()).isTrue();
+    assertThat(lastFooResponse.get().getFailureDetail().getInterrupted().getCode())
+        .isEqualTo(Code.INTERRUPTED);
+  }
+
+  /**
+   * Ensure that if a command is marked as preemptible, running a second command
```
michajlo, line 572: rm extra "command"

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
```diff
@@ -460,6 +469,320 @@ public void onCompleted() {
...
+              OutErr outErr,
+              LockingMode lockingMode,
+              String clientDescription,
+              long firstContactTimeMillis,
+              Optional<List<Pair<String, String>>> startupOptionsTaggedWithBazelRc)
+              throws InterruptedException {
+            if (args.contains(firstCommandArg)) {
+              OutputStream out = outErr.getOutputStream();
+              try {
+                while (barsReceived.get() < 3) {
```
michajlo, line 696: Is the intention with the `3` here to generate some overlap? It took me a bit to figure this out, so I'd like to recommend instead gating it with latches to make it more explicit.

```
  CountDownLatch fooBlocked, fooProceed, barBlocked, BarProceed...

  ... exec(...) {
    if (foo) {
      fooBlocked.countDown(1);
      fooProceed.await();
    } else {
      barBlocked.countDown(1);
      barProceed.await();
    }
  }
  ...
  stub.run(fooRequest, ...);
  fooBlocked.await();

  stub.run(barRequest, ...);
  barBlocked.await();
  // At this point both commands should be blocked on proceed latch, carry on...
  fooProceed.countDown(1);
  barProceed.countDown(1);

  // do verifications that prove they weren't interrupted.
```

michajlo (commented):
Largely test clarity comments .

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
daveyc123: Definitely. I was using the logic from the interrupt stream tests, but it's not really necessary here.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
daveyc123: The shutdown mechanics are used throughout the other tests. Usually I'd try to keep things consistent. What's your philosophy on cleaning up the other tests as part of this change? Happy to fix them as well.

daveyc123 (commented):
Thanks for the further feedback @michajlo. I'll get these fixed up and post a new PR

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
michajlo: I think it's fine to leave it as it is for consistency.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
daveyc123: Yup. All removed now.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
daveyc123: Removed now.

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
daveyc123: Much better this way. Thanks

daveyc123:
#12524

src/test/java/com/google/devtools/build/lib/server/GrpcServerTest.java
michajlo: make the change here too?

michajlo (changes requested):

daveyc123:
gah. Yes, sorry missed that one. New patch up

PR #12529, opened 2020-11-20
Support Scala3 .tasty files

TimothyKlim:
**ijar** tool support Kotlin modules by [not stripped out](https://github.com/bazelbuild/bazel/commit/088d8de6584864adaec82712c8ace601404afd0b)  files in `META-INF/*.kotlin_module`.
This PR add support for new Scala 3 (and Scala 2.13.4+) [TASTy](https://dotty.epfl.ch/docs/reference/metaprogramming/toc.html) format to have ability to use Scala 3 modules.

### DIFF
--- a/third_party/ijar/ijar.cc
+++ b/third_party/ijar/ijar.cc
@@ -37,6 +37,8 @@ const char *CLASS_EXTENSION = ".class";
 const size_t CLASS_EXTENSION_LENGTH = strlen(CLASS_EXTENSION);
 const char *KOTLIN_MODULE_EXTENSION = ".kotlin_module";
 const size_t KOTLIN_MODULE_EXTENSION_LENGTH = strlen(KOTLIN_MODULE_EXTENSION);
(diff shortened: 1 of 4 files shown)


TimothyKlim:
cc @philwo

Jacke:
Hey guys @philwo  You should check this MR, it's really important

TimothyKlim:
cc @alandonovan

TimothyKlim:
> Thank you! This definitely looks like a tasty PR. 
> 
> I think there are no issues with merging it, the only question is, is this important enough to be cherry-picked into Bazel 4.0 LTS (= will the lack of it be an issue for Scala users going forward?).

I hope that 4.0 will include this patch and I'll continue to work on 3.0 support for rules_scala.

alandonovan:
Code change looks fine to me. I didn't realize we still used ijar. I wrote parts of that in grad school nearly 20 years ago!

Issue #11975, opened 2020-08-19
cmd_bat quoted args escaping


codeschlosser:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

genrules cmd_bat argument quoting results in incorrect syntax/"The system cannot find the file specified"

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

create a random file so that it's full path contains a space, then

<pre>genrule(
    name = "test",
    outs = ["test.txt"],
    cmd_bat = "copy \"C:\\New folder\\test.txt\" $(@D)\\test.txt",
)</pre>

OR

<pre>genrule(
    name = "test",
    outs = ["test.txt"],
    cmd_bat = """copy "C:\\New folder\\test.txt" $(@D)\\test.txt""",
)</pre>

### What operating system are you running Bazel on?

Windows 10 x64

### What's the output of `bazel info release`?

release 3.4.1

###  Have you found anything relevant by searching the web?

No

### Any other information, logs, or outputs that you want to share?

Dumping the command by adding echo in front of copy:
<pre>genrule(
    name = "test",
    outs = ["test.txt"],
    cmd_bat = """echo copy "C:\\New folder\\test.txt" > $(@D)\\test.txt""",
)</pre>
shows that cmd.exe actually gets
<pre>copy \"C:\New folder\test.txt\"</pre>

which is not accepted by cmd.exe

PR #12530, opened 2020-11-20
Handle escaping command arguments in cmd_bat

mai93:
This pull request stops arguments escaping for commands run using cmd.exe since the result of `ShellUtils.windowsEscapeArg()` method is not accepted by cmd.exe.

Fixes: #11975


meteorcloudy (approved):
Nice, thanks for fixing this!

Issue #10062, opened 2019-10-18
git_repository strip_prefix breaks WORKSPACE `load` statements


beasleyr-vmw:
### Description of the problem / feature request:

We mirror GitHub repositories internally, so we have to replace http_archive rules with git_repository equivalents.

When cloning [bazelbuild/rules_pkg](https://github.com/bazelbuild/rules_pkg/) via `git_repository` with `strip_prefix`, I encounter the following error:

```console
$ bazel build ...
ERROR: Failed to load Starlark extension '@rules_pkg//:deps.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @rules_pkg
This could either mean you have to add the '@rules_pkg' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
INFO: Call stack for the definition of repository 'rules_pkg' which is a git_repository (rule definition at /home/beasleyr/.cache/bazel/_bazel_beasleyr/86146646415a94b3b4b4be4de036080d/external/bazel_tools/tools/build_defs/repo/git.bzl:181:18):
 - /work/git/git-strip-prefix/WORKSPACE:3:1
ERROR: cycles detected during target parsing
INFO: Elapsed time: 1.789s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

WORKSPACE:
```python
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "rules_pkg",
    remote = "https://github.com/bazelbuild/rules_pkg.git",
    tag = "0.2.2",
    strip_prefix = "pkg",
)

load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")

rules_pkg_dependencies()
```

BUILD:

```python
load("@rules_pkg//:pkg.bzl", "pkg_tar")
```

If I create a local Git repository rooted at rules_pkg's `pkg` directory, allowing me to remove the `strip_prefix` attr, then the WORKSPACE `load`+`rules_pkg_dependencies()` calls work just fine.

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

release 0.28.1

[...]

PR #12539, opened 2020-11-23
Fix using load statement on git_repository with strip_prefix

meteorcloudy:
Prevent git_repository from creating directory outside of its repository root while implementing strip_prefix.

Fixes https://github.com/bazelbuild/bazel/issues/10062


meteorcloudy:
I'll add a test for this

PR #12540, opened 2020-11-23
[Bazel] Fix mobile-install for python2

ThomasCJY:
**Background**
Fix issue introduced by https://github.com/bazelbuild/bazel/commit/1049fe89f12cfcae00902fbae6f6a5f1bc0b3a33
It turns out the queue module name is inconsistent across different python versions. We found that:
* On some macos system:
  - python2 contains both queue and Queue module. All python2 contains Queue module
  - python3 only contains queue module
* On some Linux system
  - python2 contains only Queue module. 
  - python3 only contains queue module 

Therefore, some developers are seeing `ImportError: No module named queue` errors locally on linux machine after using mobile-install. 

**Change**
Import correct Queue module instead

**Test**
Local test pass

### DIFF
--- a/tools/android/build_incremental_dexmanifest.py
+++ b/tools/android/build_incremental_dexmanifest.py
@@ -32,7 +32,7 @@ or
 
 import hashlib
 import os
-from queue import Queue
+import Queue
 import shutil
 import sys
 import tempfile
--- a/tools/android/build_incremental_dexmanifest.py
+++ b/tools/android/build_incremental_dexmanifest.py
@@ -33,9 +33,11 @@ or
 import hashlib
 import os
 try:
-    from Queue import Queue
+  # python2 without compatibility package
+  from Queue import Queue
 except ImportError:
-    import queue as Queue
+  # python3
+  from queue import Queue
 import shutil
 import sys
 import tempfile
--- a/tools/android/build_incremental_dexmanifest.py
+++ b/tools/android/build_incremental_dexmanifest.py
@@ -32,7 +32,10 @@ or
 
 import hashlib
 import os
-from Queue import Queue
+try:
+    from Queue import Queue
+except ImportError:
+    import queue as Queue
 import shutil
 import sys
 import tempfile
--- a/tools/android/build_incremental_dexmanifest.py
+++ b/tools/android/build_incremental_dexmanifest.py
@@ -32,7 +32,7 @@ or
 
 import hashlib
 import os
-import Queue
+from Queue import Queue
 import shutil
 import sys
 import tempfile


tools/android/build_incremental_dexmanifest.py
```diff
@@ -32,7 +32,7 @@
 
 import hashlib
 import os
-from queue import Queue
+import Queue
```
ThomasCJY, line 35: Should we use
```
try:
    import Queue
except ImportError:
    import queue as Queue 
```
here to support both python 2 and 3?

tools/android/build_incremental_dexmanifest.py
```diff
@@ -32,7 +32,7 @@
 
 import hashlib
 import os
-from queue import Queue
+from Queue import Queue
```
ThomasCJY, line 35: Should we use

try:
    import Queue from Queue
except ImportError:
    import queue as Queue 
here to support both python 2 and 3?

arunkumar9t2:
```python
try:
    from Queue import Queue
except ImportError:
    import queue as Queue 
```

This is the syntax that worked when python 2 is used.

nkoroste:
Any updates on this? pretty trivial change...

tools/android/build_incremental_dexmanifest.py
odracirnumira: @ThomasCJY can you change the code to the alternative suggested by @arunkumar9t2 ? That'll make this work in both Pyton2 and 3.

ThomasCJY: updated

odracirnumira: The change looks reasonable to me. Unfortunately I lack the necessary permissions to stamp the diff. I think @ahumesky is the right person to loop in here?

ahumesky:
I've imported this change internally, but our linters are pretty unhappy about imports "not at top of file". We seem to be doing this in other places so I think I can work around it.

ThomasCJY:
@ahumesky yeah you are right. It's a bit confusing but according to the python doc:

python 2 uses `Queue` module name and the creation of object is `Queue.Queue()`
https://docs.python.org/2.7/library/queue.html

python 3 uses `queue` module name and the creation of object is `queue.Queue()`
https://docs.python.org/3.8/library/queue.html

so the import syntax should be
```
try:
  # python2 without compatibility package
  from Queue import Queue
except ImportError:
  # python3
  from queue import Queue
```

ThomasCJY:
@ahumesky is this good to go?

PR #12541, opened 2020-11-24
Detect ar executable on Linux.

Lalaland:
This commit makes it possible for the default linux toolchain to detect the ar executable.
This is necessary for rules_foreign_cc as some of their rules depend on ar.

Tested: Manually


Lalaland:
@oquenchil Sorry for the trouble, but would it be possible for you to give this a quick review? It's a relatively small change and helps improve my workflow significantly.

oquenchil (approved):
LGTM

Issue #12451, opened 2020-11-11
No way exists to ignore a host JDK


AustinSchuh:
We have developers who accidentally (or on purpose) install different JDK versions.  I want to be able to tell Bazel to never ever look at those.

In `src/main/cpp/blaze_util_linux.cc`

```
string GetSystemJavabase() {
  // if JAVA_HOME is defined, then use it as default.
  string javahome = GetPathEnv("JAVA_HOME");
  if (!javahome.empty()) {
    string javac = blaze_util::JoinPath(javahome, "bin/javac");
    if (access(javac.c_str(), X_OK) == 0) {
      return javahome;
    }   
    BAZEL_LOG(WARNING)
        << "Ignoring JAVA_HOME, because it must point to a JDK, not a JRE.";
  }

  // which javac
  string javac_dir = Which("javac");
  if (javac_dir.empty()) {
    return ""; 
  }

  // Resolve all symlinks.
  char resolved_path[PATH_MAX];
  if (realpath(javac_dir.c_str(), resolved_path) == NULL) {
    return ""; 
  }
  javac_dir = resolved_path;

  // dirname dirname
  return blaze_util::Dirname(blaze_util::Dirname(javac_dir));
}
```

This tries pretty hard to find a JDK.  I'd like an environmental variable, reserved value (/dev/null?) or flag or something which I can pass in to return `""`, ie no JDK found.  (I'm also happy to implement it if there is consensus.)

comius:
If you prevent one JDK selection, which JDK should be selected instead?

If you know absolute path to the preferred JDK, you can put `local_java_repository(name = 'local_jdk', java_home = '...')` into the workspace.

You can also use JDK from a remote repository for execution, by setting `build --javabase=//tools/jdk:remote_jdk11` in `.bazelrc`.

AustinSchuh:
We have the following in our `WORKSPACE` file.

```
# Java9 JDK.
http_archive_architecture(
    name = "openjdk_linux_archive",
    build_file_content = """
java_runtime(
    name = 'jdk',
    srcs = glob(['**']),
    visibility = ['//visibility:public'],
)
""",
    sha256 = "ddb0fd4526089cf1ce2db36282c282263f587a9e8be373fa02f511a12923cc48",
    strip_prefix = "zulu11.31.11-ca-jdk11.0.3-linux_x64",
    urls = [
[...]

PR #12542, opened 2020-11-24
Add --{no,}autodetect_server_javabase.

AustinSchuh:
We want bazel to fail to start instead of falling back to a host
JRE/JDK.  We are using a hermetic JDK and the embedded JRE, so there
should be no need to use anything from the host.  We've debugged enough
cases so far where the host installed JDK was buggy and causing random
crashes on specific machines.

Fixes: #12451

### DIFF
--- a/site/docs/user-manual.html
+++ b/site/docs/user-manual.html
@@ -2838,7 +2838,7 @@ adb -s deadbeef install ...
 <h4 id='flag--autodetect_server_javabase'><code class='flag'>--autodetect_server_javabase</code></h4>
 <p>
   This option causes Bazel to automatically search for an installed JDK on startup,
-  and to fall back to the installed JRE if the embedded JRE isn't available. 
+  and to fall back to the installed JRE if the embedded JRE isn't available.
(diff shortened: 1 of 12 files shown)


AustinSchuh:
It should be pretty easy to make the test pass, but can you confirm the approach looks sane?

comius:
> It should be pretty easy to make the test pass, but can you confirm the approach looks sane?

Yes, it's ok.

Mind also adding a test for it? This will ensure that somebody else won't break the feature later. There are integration tests in //src/test/shell/integration. statup_options or java_integration_test should get you started, but there is no integration test on //src:bazel_nojdk, so you'll need to set it up.

AustinSchuh:
> statup_options

Sure, done.  That was surprisingly easy.  Tested both normal and nojdk bazel with --autodetect_server_javabase and --noautodetect_server_javabase.

comius (approved):

comius:
Nice implementation, thanks for your effort! Sorry for the delay.

PR #12547, opened 2020-11-24
Change ConfiguredTargetQuery to use KeyedConfiguredTarget as a value.

### LINKED ISSUE #11993: cquery produces non-actionable warning related to toolchains

IF
1. there is a target Foo which depends on some toolchain;
2. `--incompatible_override_toolchain_transition` is in effect;
3 command is `bazel cquery deps(Foo)
THEN
`WARNING: Targets were missing from graph` is generated.

The small repro with Python toolchain is published [here](https://github.com/Bazel-snippets/cquery_targets_missing_repro)
We get exactly the same problem with C++ toolchain as well.

Windows, Linux, OSX.

3.4.1

katre:
This will allow us to keep the configured target key between lookup rounds.

Fixes #11993.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
@@ -56,7 +56,7 @@ import javax.annotation.Nullable;
  *
  * <p>Incomplete; we'll implement getVisibility when needed.
  */
-public class ConfiguredTargetAccessor implements TargetAccessor<ConfiguredTarget> {
+public class ConfiguredTargetAccessor implements TargetAccessor<KeyedConfiguredTarget> {
 
   private final WalkableGraph walkableGraph;
   private final ConfiguredTargetQueryEnvironment queryEnvironment;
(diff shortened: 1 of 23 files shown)


src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQueryEnvironment.java
```diff
@@ -517,18 +517,15 @@ protected BuildConfiguration getConfiguration(ConfiguredTarget target) {
   }
 
   @Override
-  protected ConfiguredTargetKey getSkyKey(ConfiguredTarget target) {
-    return ConfiguredTargetKey.builder()
-        .setConfiguredTarget(target)
-        .setConfiguration(getConfiguration(target))
-        .build();
+  protected ConfiguredTargetKey getSkyKey(KeyedConfiguredTarget target) {
```
juliexxia, line 520: I think this method's assumptions have caused other issues in the past so it's nice we can tighten it up here.

src/test/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQuerySemanticsTest.java
```diff
@@ -120,9 +119,9 @@ public void testLabelFunction_getCorrectlyConfiguredDeps() throws Exception {
     setUpLabelsFunctionTests();
 
     // Test that this retrieves the correctly configured version(s) of the dep(s).
-    ConfiguredTarget patchDep =
+    KeyedConfiguredTarget patchDep =
```
juliexxia, line 122: Can we add a test case for the situation that caused #11993 ?

juliexxia (commented):
Overall looks great!

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
```diff
@@ -180,14 +176,17 @@ public Target getTargetFromConfiguredTarget(ConfiguredTarget configuredTarget) {
   }
 
   /** Returns the rule that generates the given output file. */
-  public RuleConfiguredTarget getGeneratingConfiguredTarget(OutputFileConfiguredTarget oct)
+  RuleConfiguredTarget getGeneratingConfiguredTarget(KeyedConfiguredTarget kct)
```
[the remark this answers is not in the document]
katre: Done.

src/test/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetQuerySemanticsTest.java
katre: Good point. Added to //src/test/shell/bazel:toolchain_transition_test

src/test/shell/bazel/toolchain_transition_test.sh
```diff
@@ -355,4 +354,29 @@ EOF
   expect_log 'extra_lib: message: extra_lib foo, target_dep: target, tool_dep: exec-alpha'
 }
 
+function test_toolchain_transition_cquery() {
```
juliexxia, line 357: Do you mind adding a comment about what this seemingly innocuous test case is testing for?

juliexxia (approved):

src/test/shell/bazel/toolchain_transition_test.sh
katre: Done.

PR #12549, opened 2020-11-24
Clean up ConfiguredTargetValueAccessor and ConfiguredTargetAccessor

katre:
- Fix looking up a target from the walkable graph.
  - This prevents ConfiguredTargetValueAccessor from needing to look into
    ConfiguredTargetAccessor.
  - And removes duplicated code.
- Move CTVA into the correct package.

Part of #11993.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/engine/QueryEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/engine/QueryEnvironment.java
@@ -17,7 +17,9 @@ import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.devtools.build.lib.cmdline.Label;
 import com.google.devtools.build.lib.concurrent.ThreadSafety.ThreadSafe;
+import com.google.devtools.build.lib.packages.Target;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/devtools/build/lib/query2/engine/QueryEnvironment.java
```diff
@@ -159,6 +161,11 @@ public final FilteringQueryFunction asFilteringFunction() {
     public abstract int getExpressionToFilterIndex();
   }
 
+  @FunctionalInterface
+  interface TargetLookup {
+    Target getTarget(Label label) throws TargetNotFoundException, InterruptedException;
```
juliexxia, line 166: apologies I think I'm missing something - where does this actually get implemented?

katre: It's a functional interface, so in ActionGraphQueryEnvironment's constructor, the method reference ActionGraphQueryEnvironment::getTarget is used as a TargetLookup and passed into ConfiguredTargetValueAccessor.

juliexxia: Ah, and AGQE has an inherited getTarget method from PostAnalysisQE. Missed that bit. Thanks!

juliexxia (approved):

PR #12550, opened 2020-11-24
Move getConfigConditions into ConfiguredTarget.

katre:
Uses a default value for non-alias and non-rule CTs.

This alloows removing many instanceof/cast checks when getting config
conditions.

Part of the work on #11993.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BUILD
@@ -696,6 +696,7 @@ java_library(
     name = "configured_target",
     srcs = ["ConfiguredTarget.java"],
     deps = [
+        ":config/config_matching_provider",
         ":transitive_info_collection",
         "//src/main/java/com/google/devtools/build/lib/actions:artifacts",
         "//src/main/java/com/google/devtools/build/lib/cmdline",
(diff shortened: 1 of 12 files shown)


src/main/java/com/google/devtools/build/lib/analysis/BUILD
```diff
@@ -696,6 +696,7 @@ java_library(
     name = "configured_target",
     srcs = ["ConfiguredTarget.java"],
     deps = [
+        ":config/config_matching_provider",
```
juliexxia, line 699: This is totally orthogonal to your PR and just something I've never understood - why have a slash in a target name as opposed to just putting it under a directory called config?

juliexxia (approved):
Really lovely simplification

src/main/java/com/google/devtools/build/lib/analysis/BUILD
katre: Because [analysis_cluster](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/BUILD;l=164) contains files from inside the config subdirectory, config can't be its own package with its own BUILD file.

Ideally we'd further break up analysis_cluster and fix this, but it's a hard problem and I couldn't crack it last time I tried.

PR #12551, opened 2020-11-24
Fix error message from getPrerequisites to not print internal details.

katre:
Part of work on #11993.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
@@ -131,8 +131,8 @@ public class ConfiguredTargetAccessor implements TargetAccessor<ConfiguredTarget
       throw new QueryException(
           caller,
           String.format(
-              "%s %s of type %s does not have attribute '%s'",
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
```diff
@@ -131,8 +131,8 @@ public boolean isTestSuite(ConfiguredTarget target) {
       throw new QueryException(
           caller,
           String.format(
-              "%s %s of type %s does not have attribute '%s'",
-              errorMsgPrefix, actualConfiguredTarget, rule.getRuleClass(), attrName),
+              "%sconfigured target of type %s does not have attribute '%s'",
```
juliexxia, line 134: does this need a space between %s and configured ? It looks fine in the test but the error message above does seem to have a space after errorMsgPrefix.

juliexxia (approved):

src/main/java/com/google/devtools/build/lib/query2/cquery/ConfiguredTargetAccessor.java
katre: No, the passed-in prefix already has the space, and I didn't want to try and fix all the other uses to make it look nicer here.

PR #12556, opened 2020-11-25
Separate ijar sources from deployment zip.

comius:
This is a github only patch from PR https://github.com/bazelbuild/bazel/pull/12546. Needs to go in first.


philwo:
@comius Can you check out the test failures?

comius:
> @comius Can you check out the test failures?

Done, I love monorepos.

PR #12560, opened 2020-11-25
Fix a couple of bugs with Incompatible Target Skipping

### LINKED ISSUE #12553: Incompatible Target Skipping breaks rules with custom providers

Consider the following Starlark rule:
```python
DummyProvider = provider()

def _dummy_rule_impl(ctx):
    return [DummyProvider()]

dummy_rule = rule(
    implementation = _dummy_rule_impl,
    attrs = {
        "deps": attr.label_list(providers=[DummyProvider]),
    },
)
```

If you instantiate it like this:
```python
load(":rules.bzl", "dummy_rule")

dummy_rule(
    name = "dummy1",
)

dummy_rule(
    name = "dummy2",
    deps = [
        ":dummy1",
    ],
)
```

then building fails with:
```
ERROR: /home/phil/.cache/bazel/_bazel_phil/b4868b516be4e3bea4d221937a55ced4/sandbox/linux-sandbox/2572/execroot/io_bazel/_tmp/c749fd38ac9a4ad6bd41e9653bbceab5/workspace/target_skipping/BUILD:112:11: in deps attribute of dummy_rule rule //target_skipping:dummy2: '//target_skipping:dummy1' does not have mandatory providers: 'DummyProvider'
```

This occurs because the incompatible target is assigned an `IncompatiblePlatformProvider` and not a `DummyProvider`.

The same thing happens when using Python targets: 
```
ERROR: /home/phil/repos/971-Robot-Code/frc971/control_loops/python/BUILD:189:11: in deps attribute of py_library rule //frc971/control_loops/python:basic_window: '@python_gtk//:python_gtk' does not have mandatory providers: 'py' or 'PyInfo'
```

A quick hack to fix this is:
```patch
diff --git a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
index 78ee802789..e094b7b529 100644
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -43,6 +43,7 @@ import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.actions.Artifact.SpecialArtifact;
 import com.google.devtools.build.lib.actions.ArtifactRoot;
 import com.google.devtools.build.lib.analysis.AliasProvider.TargetMode;
+import com.google.devtools.build.lib.analysis.IncompatiblePlatformProvider;
 import com.google.devtools.build.lib.analysis.actions.ActionConstructionContext;
 import com.google.devtools.build.lib.analysis.buildinfo.BuildInfoKey;
 import com.google.devtools.build.lib.analysis.config.BuildConfiguration;
@@ -2292,6 +2293,10 @@ public final class RuleContext extends TargetContext
         return true;
       }

+      if (prerequisite.getConfiguredTarget().getProvider(IncompatiblePlatformProvider.class) != null) {
+        return true;
+      }
+
       unfulfilledRequirements.add(
           String.format(
               "'%s' does not have mandatory providers: %s",
```

x86_64 Linux

Running off latest master (d0efd7b9e5109ff5ac6d13c91f58c3fc4dc3afd8).

```console
$ bazel build -c opt //src:bazel
```

philsc:
While adding `target_compatible_with` attributes in the FRC team 971's
code base I came across bug #12553. It wasn't possible to make a
Python target depend on another incompatible Python target.

This patch fixes that issue by teaching `RuleContext` about
incompatible prerequisites. This also fixes an issue with the
validation of file extensions.

It's possible that `validateDirectPrerequisite()` skips a bit too much
validation. It was unfortunately the cleanest way I could think of.

I struggled a bit to come up with what ended up becoming
`RuleContextConstraintSemantics.IncompatibleCheckResult`. The purpose
of that class is to centralize the logic for checking for
`OutputFileConfiguredTarget` dependencies. Such dependencies need a
bit of special logic in order to find `IncompatiblePlatformProvider`
instances. When I implemented this patch, I realized that the
`TopLevelConstraintSemantics` logic had a very similar problem. It
didn't deal with the `OutputFileConfiguredTarget` problem at all. I
took the liberty of fixing that issue in this patch also because it
nicely re-used the same new helper.

I could not figure out a good way to avoid making `RuleContext` depend
on `RuleContextConstraintSemantics`.

Fixes #12553

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
@@ -54,6 +54,7 @@ import com.google.devtools.build.lib.analysis.config.FragmentCollection;
 import com.google.devtools.build.lib.analysis.config.transitions.ConfigurationTransition;
 import com.google.devtools.build.lib.analysis.config.transitions.NoTransition;
 import com.google.devtools.build.lib.analysis.constraints.ConstraintSemantics;
+import com.google.devtools.build.lib.analysis.constraints.RuleContextConstraintSemantics;
 import com.google.devtools.build.lib.analysis.platform.ConstraintValueInfo;
 import com.google.devtools.build.lib.analysis.platform.PlatformInfo;
 import com.google.devtools.build.lib.analysis.stringtemplate.TemplateContext;
@@ -2306,6 +2307,14 @@ public final class RuleContext extends TargetContext
 
     private void validateDirectPrerequisite(
         Attribute attribute, ConfiguredTargetAndData prerequisite) {
+      if (RuleContextConstraintSemantics.checkForIncompatibility(prerequisite.getConfiguredTarget())
+          .isIncompatible()) {
+        // If the prerequisite is incompatible (e.g. has an incompatible provider), we pretend that
+        // there is no further validation needed. Otherwise, it would be difficult to make the
+        // incompatible target satisfy things like required providers and file extensions.
+        return;
+      }
+
       validateDirectPrerequisiteType(prerequisite, attribute);
       validateDirectPrerequisiteFileTypes(prerequisite, attribute);
       if (attribute.performPrereqValidatorCheck()) {
--- a/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
@@ -17,6 +17,7 @@ package com.google.devtools.build.lib.analysis.constraints;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Streams.stream;
 
+import com.google.auto.value.AutoValue;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Verify;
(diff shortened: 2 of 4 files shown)


philsc:
Okay, this is ready for review.

I'm still not happy about making `RuleContext.java` import `RuleContextConstraintSemantics`, but I can't think of a better way.

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
```diff
@@ -2306,6 +2307,14 @@ private boolean checkRuleDependencyMandatoryProviders(
 
     private void validateDirectPrerequisite(
         Attribute attribute, ConfiguredTargetAndData prerequisite) {
+      if (RuleContextConstraintSemantics.checkForIncompatibility(prerequisite.getConfiguredTarget())
```
gregestren, line 2310: Is there any reason this couldn't be put directly in `checkRuleDependencyMandatoryProviders`? It looks like none of the other checks here (like checking valid rule type) would be affected by incompatible targets?

gregestren (commented):
This overall looks good to me. I also don't think it's that huge a deal with the new dependency from `RuleContext`, although I appreciate the general concern about that.

Just the code placing comment as my only nit.

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
philsc: I could move it in there, but then I need a duplicate check in `validateDirectPrerequisiteFileTypes` I think. Basically the code that checks for extensions also needs to be skipped/tweaked.

In other words, if I move this into `checkRuleDependencyMandatoryProviders`, then the `test_dependencies_with_extensions` test would fail.

That of course works for me too. It's not like I'd be copy-pasting tens of lines of code in two places.

gregestren: Ah, got it. No worries. Looks good as is.

gregestren (approved):

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
```diff
@@ -830,6 +830,42 @@ private static void addSelectValuesToSet(BuildType.Selector<?> select, final Set
...
+   * Provides information about a target's incompatibility.
+   *
+   * <p>After calling {@code checkForIncompatibility()}, the {@code isIncompatible} flag tells you
+   * whether the target is incompatible. If the target is incompatible, then {@code
+   * underlyingTarget} tells you which underlying target provided the incompatibility. For the vast
+   * majority of targets this is the same one passed to {@code checkForIncompatibility()}. In some
+   * instances like {@link OutputFileConfiguredTarget}, however, the {@code underlyingTarget} is the
+   * rule that generated the file.
+   */
+  public static final class IncompatibleCheckResult {
```
gregestren, line 843: Forwarding one more comment from @katre:

"Nit: This class will be difficult to debug because it has no toString method. Consider using AutoValue to create it and populate reasonable default toString/hashCode/equals methods."

There are other AutoValue classes in the code base that can show precedent.

philsc: Ah yes, good idea. I'll get on this tonight.

philsc: Done

PR #12562, opened 2020-11-26
Use write result for pid done signal

werkt:
Prevent warnings during build for a signal write to indicate that the
pid file has been created. This is a warning under -Wunused-result on
linux.

### DIFF
--- a/src/main/tools/daemonize.c
+++ b/src/main/tools/daemonize.c
@@ -39,6 +39,7 @@
 
 #include <assert.h>
 #include <err.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


benjaminp:
Probably this should have an `EINTR` loop.

I also wonder why `write` at all? Closing the pipe seems like it should be a sufficient signal.

werkt:
> Probably this should have an `EINTR` loop.

Good plan.

> I also wonder why `write` at all? Closing the pipe seems like it should be a sufficient signal.

The pipe must remain open by the forked process in order to successfully write into the fd - this is a mild guarantee that the subproc hasn't failed/closed the pipe receive side until it could be signaled, and a way to ensure that the signaled side perceives this failure. No such guarantee is provided with a close. This code will die with SIGPIPE/fail with EPIPE if the read side has been closed (by all processes).

aiuto:
@werkt Are you still working on this, or are you ready for a rewview.

werkt:
> @werkt Are you still working on this, or are you ready for a rewview.

All ready for review, not in draft.

aiuto (approved):
This looks correct to me. Of course, since we are not checking an error, my hunch is that something is going to complain.

PR #12563, opened 2020-11-26
Include the product version as an alias when generating xcode_version targets

segiddins:
This allows passing `--xcode_version=$XCODE_PRODUCT_BUILD_VERSION` in the context of an Xcode shell script build phase


keith (approved):

susinmotion (approved):

tetromino (changes requested):
This change is of course better than the broken check we had before, but it still doesn't match the comments and variable names; "at least as large" means version >= subversion, i.e. !(subversion > version), but your new check is for subversion < version.

Please change to

      BOOL atLeastAsLarge =
          ([subversionEntry.version compare:entry.version
                                    options:NSNumericSearch] != NSOrderedDescending);

segiddins:
@tetromino good catch, done

tetromino (approved):

Issue #4301, opened 2017-12-14
Doc for @bazel_tools


olafure:
Bazel documentation is missing an explanation the "@bazel_tools" built in repository.

panzhongxian:
What's the link of doc for @bazel_tools ? thx.

dslomov:
We did add documentation for `http_archive` etc; shall we add more?

laurentlb:
The documentation doesn't say what is `@bazel_tools`, so yes.
We should mention it's a magic repository that's part of Bazel binary.

beasleyr-vmw:
External contributors would benefit from guidelines for what (not) to do when adding to bazel_tools (see https://github.com/bazelbuild/bazel/issues/8738#issuecomment-581163441).

laurentlb:
I think it's better if we don't add anything to `@bazel_tools`.
Most files should live outside the Bazel binary.

jin:
+1 we should be removing targets from `@bazel_tools` as much as possible. 

cc @alandonovan who ran into issues with the JDK / Java rules side of the embedded repository recently.

alandonovan:
I recently added a "jni" target to @bazel_tools//java/jdk, and was struck that (a) this repo's public API is essentially undocumented, and that declarations added to it take a full Bazel release before they can be used, which imposes a surprisingly subtle obstacle to evolution.

I agree we should get rid of this repo. Although there may still be a need for a mechanism to expose built-in symbols to certain rule sets (Java, for example) owned by the Bazel team, exposing these symbols directly to the whole world is a liability.

blais:
I just encountered this tool and got surprised by its existence as well.

nickgeorge:
+1, as an internal google user, this was still very surprising to me that there is a magic @target, that is a requirement to know about for things as basic as http_archive. This needs to be documented well.

dzintars:
Not sure is it important there, but i came across this error/warning when run `ibazel`:
```sh
iBazel [3:29PM]: Querying for files to watch...
Loading: 0 packages loaded
[...]

PR #12567, opened 2020-11-26
Move realpath to tools/osx.

thii:
So that it will be made available to Starlark rules via
@bazel_tools/osx/realpath.


thii:
Just seen https://github.com/bazelbuild/bazel/issues/4301. It looks like we shouldn't add anything to `@bazel_tools`.

PR #12584, opened 2020-11-29
Migrate some uses of NestedSet#forEachElement to NestedSetVisitor

Yannic:
These were the only 2 uses of NestedSet#forEachElement in Bazel.

No functional change intended.


aiuto:
Friendly ping.  This seems like a good tech-debt cleanup if it means we can remove the unused method.

michaeledgar:
Approved. This looks like a solid clean-up. With this merged, and a followup internal cleanup, we can drop the method in question.

I'll import the patch and push it through our internal CI to merge it now.

Issue #11259, opened 2020-04-30
BAZEL_VS / BAZEL_VC don't account for custom install locations well.


laramiel:
### Bazel fails to find vc installation.

I'm trying to use bazel with a non-standard ms_buildtools install location, and it fails to find all the tools.
Specifically, I've installed ms_buildtools like this:

--locale en-US --installPath C:\BuildTools --config config.txt

config.txt:
{
  "version": "1.0",
  "components": [
    "Microsoft.VisualStudio.Component.Roslyn.Compiler",
    "Microsoft.Component.MSBuild",
    "Microsoft.VisualStudio.Component.CoreBuildTools",
    "Microsoft.VisualStudio.Workload.MSBuildTools",
    "Microsoft.VisualStudio.Component.Windows10SDK",
    "Microsoft.VisualStudio.Component.VC.CoreBuildTools",
    "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
    "Microsoft.VisualStudio.Component.VC.Redist.14.Latest",
    "Microsoft.VisualStudio.Component.Windows10SDK.18362",
    "Microsoft.VisualStudio.Component.VC.CMake.Project",
    "Microsoft.VisualStudio.Component.TestTools.BuildTools",
    "Microsoft.Net.Component.4.8.SDK",
    "Microsoft.Net.Component.4.6.1.TargetingPack",
    "Microsoft.VisualStudio.Component.VC.CLI.Support",
    "Microsoft.VisualStudio.Component.VC.ASAN",
    "Microsoft.VisualStudio.Workload.VCTools"
  ]
}

When running bazel build, it doesn't really matter what I set BAZEL_VC to, bazel fails to find the file VCVARSALL.BAT

in this configuration VCVARSALL.bat lives here:

C:\BuildTools\VC\Auxiliary\Build

while cl.exe, etc live here (among other places)

C:\BuildTools\VC\Tools\MSVC\14.25.28610\bin\Hostx64\x64

The windows_cc_configure.bzl file knows about this, but it looks at the path to determine the version. Since BuildTools does not include the magic string 2019, this fails.

It seems more reasonable to match against the file layout.

laszlocsomor:
Looks like `windows_cc_configure.bzl` needs to be fixed.

[...]

PR #12586, opened 2020-11-30
Update the visual studio version checking logic

mai93:
This PR updates the `_is_vs_2017_or_2019` function to specify the visual studio version based on the contents of the VC folder instead of depending on the parent directories of VC. 

This is helpful in cases where `BAZEL_VC` points to a custom location of VC.

Fixes: #11259

### DIFF
--- a/tools/cpp/windows_cc_configure.bzl
+++ b/tools/cpp/windows_cc_configure.bzl
@@ -261,8 +261,8 @@ def _is_vs_2017_or_2019(repository_ctx, vc_path):
     # In VS 2017 and 2019, it contains only three directories:
     # "Auxiliary", "Redist", "Tools"
 
-    vc_2017_or_2019_contents = ["Auxiliary", "Redist", "Tools"]
-    vc_path_contents = [d.basename for d in repository_ctx.path(vc_path).readdir()]
+    vc_2017_or_2019_contents = ["auxiliary", "redist", "tools"]
+    vc_path_contents = [d.basename.lower() for d in repository_ctx.path(vc_path).readdir()]
     vc_path_contents = sorted(vc_path_contents)
     return vc_path_contents == vc_2017_or_2019_contents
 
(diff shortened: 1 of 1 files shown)


mai93:
I am not sure if this is the best way to distinguish between the different VS versions based on VC directory. Other solutions that I considered:

- Since the contents of the VC folder are completely different between VS 2017/2019 and older versions, we can determine the VS version based on the existence of only one of VC subdirectories. For example, if VC contains 'Auxiliary' directory then it is VS 2017 or 2019 and if not then it is an older version. 

- Check for VCVARSALL.BAT script location and determine the VS version based on it, i.e. if {vc_path}\\Auxiliary\\Build\\VCVARSALL.BAT exists this means it is VS 2017 or 2019 while {vc_path}\\VCVARSALL.BAT means it is VS 2015 or older.

But I thought it is better to check that the VC subdirectories match the valid layout in VS 2017/2019.

tools/cpp/windows_cc_configure.bzl
```diff
@@ -254,14 +254,17 @@ def find_vc_path(repository_ctx):
     auto_configure_warning_maybe(repository_ctx, "Visual C++ build tools found at %s" % vc_dir)
     return vc_dir
 
-def _is_vs_2017_or_2019(vc_path):
-    """Check if the installed VS version is Visual Studio 2017."""
+def _is_vs_2017_or_2019(repository_ctx, vc_path):
```
meteorcloudy, line 257: Thanks! 

This check looks much better, according to https://devblogs.microsoft.com/cppblog/compiler-tools-layout-in-visual-studio-15/#where-the-compiler-toolset-moved-to-in-vs-2017, those three directory should always exist for 2017 and later version of Visual C++ Build Tools.

However, since file path is case-insensitive, can you convert all strings to lower case and then do the comparison? Just in case somehow the directory was created with different cases somehow.

meteorcloudy (changes requested):

meteorcloudy (approved):
Thank you!

Issue #4592, opened 2018-02-06
Convert Java rules to use Toolchain Resolution


katre:
Toolchain resolution is available in native rules. Java rules should be updated to use toolchain resolution and platforms, rather than the existing javabase and java_toolchain options in [JavaOptions](https://source.bazel.build/bazel/+/5be4dd62ee5d7bf3730ab2099829dc7f87ef2e94:src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java;l=43).

katre:
Some work has been done already to convert JavaToolchainProvider and JavaRuntimeInfo into ToolchainInfo instances. Future work will be to centralize access to those objects, and then convert them to use toolchain resolution instead of the current late-bound-attributes for lookup.

lberki:
to be clear: I thought you were working on this. In what areas do you require my help?

katre:
Mostly I need help in terms of code and design reviews. This isn't scheduled and I'll only be tackling it on a best-effort basis.

lberki:
Excellent, then we are on the same page. I'd love to see this happen, but I also can't commit more time than the occasional code/design review.

ittaiz:
@katre is this issue the same as https://github.com/bazelbuild/bazel/issues/5462? (which is now closed) If not can you please explain how they relate?
Thanks!

katre:
Yes, this is the same as #5462, which I closed as a dupe.

We have committment for work on this in 2019Q1, I will try and make sure the appropriate issue is tagged in the commits relating to the work, so that interested parties can follow along.

ittaiz:
Thanks!

meisterT:
@katre can you please update this issue?

katre:
All of the code is in place, but we haven't flipped the flag for Bazel because no one was available to write the toolchain rules for the Java toolchains. Once that is done we can test flipping the flag and see where downstream projects stand.

shs96c:
Is this something that the community can get involved in? Is there a guide or an design doc outlining what it means to "write the toolchain rules for the Java toolchains"? I'm guessing this is more than just adding a new toolchain type and a couple of targets.

katre:
The toolchain types exist. the big issues are still, I think, design issues: where do the toolchains live (currently it's some mix of @bazel_tools//tools/jdk, @rules_java, and several @java_tools-type repositories).

I'm not sure who is actually in charge of the state of the Java rules: @lberki, can you clarify who is the decider?

[...]

PR #12588, opened 2020-11-30
Add toolchain definitions for Java.

comius:
Design document: https://docs.google.com/document/d/1MVbBxbKVKRJJY7DnkptHpvz7ROhyAYy4a-TZ-n7Q0r4/edit

A new macro remote_java_repository is defined, which creates an additional
repository with BUILD file containing toolchain target and registers it.

An additional repository prevents download of all JDKs (even if not
used).

Similarly local_java_repository rule is extended to add toolchain
target. In case the local JDK is not present a fake toolchain is added in
order to provide a nicer error message (without it toolchain resolution
fails without giving a clear message what was wrong).

Default_java_toolchain macro is extended to define toolchain rule as well.

Local JDKs version is autodetected.

Addresses https://github.com/bazelbuild/bazel/issues/4592

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
@@ -612,7 +612,7 @@ public class JavaOptions extends FragmentOptions {
 
   @Option(
       name = "java_runtime_version",
-      defaultValue = "",
+      defaultValue = "local_jdk",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "The Java runtime version")
@@ -620,7 +620,7 @@ public class JavaOptions extends FragmentOptions {
 
   @Option(
       name = "tool_java_runtime_version",
-      defaultValue = "",
+      defaultValue = "remotejdk_11",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "The Java runtime version used to execute tools during the build")
--- a/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
@@ -628,7 +628,7 @@ public class JavaOptions extends FragmentOptions {
 
   @Option(
       name = "java_language_version",
-      defaultValue = "",
+      defaultValue = "8",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "The Java language version")
@@ -636,7 +636,7 @@ public class JavaOptions extends FragmentOptions {
 
   @Option(
       name = "tool_java_language_version",
-      defaultValue = "",
+      defaultValue = "8",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "The Java language version used to build tools that are executed during a build")
--- a/src/test/shell/bazel/bazel_java14_test.sh
+++ b/src/test/shell/bazel/bazel_java14_test.sh
@@ -125,7 +125,8 @@ public class Javac14Example {
   }
 }
 EOF
-  bazel run java/main:Javac14Example --test_output=all --verbose_failures &>"${TEST_log}"
+  # TODO(ilist): remove tool_java_runtime_version after java_runtime is attached to the toolchain
+  bazel run java/main:Javac14Example --java_language_version=14 --java_runtime_version=14 --tool_java_runtime_version=14 --test_output=all --verbose_failures &>"${TEST_log}"
   expect_log "0"
 }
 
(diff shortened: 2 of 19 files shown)


src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
```diff
@@ -612,31 +612,31 @@ public ImportDepsCheckingLevelConverter() {
...
+      defaultValue = "remotejdk_11",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "The Java runtime version used to execute tools during the build")
   public String hostJavaRuntimeVersion;
 
   @Option(
       name = "java_language_version",
-      defaultValue = "",
+      defaultValue = "8",
```
comius, line 631: Note: The plan is to update this to "11" when the machinery is a bit better and it will work on Java 8.
For this change not to cause problems internally it looks best option is to override it in .rc file.

comius:
cc @oquenchil

src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
katre: SGTM

src/test/shell/bazel/bazel_java14_test.sh
```diff
@@ -125,7 +125,8 @@ public class Javac14Example {
   }
 }
 EOF
-  bazel run java/main:Javac14Example --test_output=all --verbose_failures &>"${TEST_log}"
+  # TODO(ilist): remove tool_java_runtime_version after java_runtime is attached to the toolchain
```
katre, line 128: Is there an issue to attach to this?

tools/jdk/default_java_toolchain.bzl
```diff
@@ -146,6 +146,18 @@ def default_java_toolchain(name, configuration = DEFAULT_TOOLCHAIN_CONFIGURATION
         **toolchain_args
     )
 
+    native.config_setting(
+        name = name + "_version_setting",
+        values = {"java_language_version": toolchain_args["source_version"]},
+        visibility = ["//visibility:private"],
+    )
+    native.toolchain(
+        name = name + "_toolchain",
```
katre, line 155: The names this generates are kind of confusing, like `//:toolchain_toolchain` and //java/main:default_toolchain_toolchain`.

We still don't have good terminology around this, but would using `_definition` work? then we'd get `//:toolchain_definition` and `//java/main:default_toolchain_definition`, which is at least less redundant.

tools/jdk/local_java_repository.bzl
```diff
@@ -12,13 +12,25 @@
...
-Creates a local repository using jdk.BUILD file.
+def _detect_java_version(repository_ctx, java_bin):
+    properties_out = repository_ctx.execute([java_bin, "-XshowSettings:properties"]).stderr
 
-When Java executable is not present it creates a BUILD file with target "jdk"
-displaying an error message.
-"""
+    version_property = "java.version = "
+    versions = [
+        property.lstrip()[len(version_property):].rstrip()
```
katre, line 22: This is taking something of the form "java.version = 11.16.7" and parsing it to just "11.16.7", right? The lstrip and rstrip are a bit confusing, as is the multi-line list comprehension. I'd consider re-writing as a standard for loop (possibly in a separate function, but that might be overkill). But then you can just use `strip()` instead of `lstrip()` and `rstrip()`.

tools/jdk/local_java_repository.bzl
```diff
@@ -34,10 +46,39 @@ def _local_java_repository_impl(repository_ctx):
...
+    name = "name_setting",
+    values = {{"java_runtime_version": "{local_jdk}"}},
+    visibility = ["//visibility:private"],
+)
+config_setting(
+    name = "version_setting",
+    values = {{"java_runtime_version": "{version}"}},
+    visibility = ["//visibility:private"],
+)
+alias(
```
katre, line 67: We already have skylib in Bazel core, use [`selects.with_or`](https://github.com/bazelbuild/bazel-skylib/blob/master/docs/selects_doc.md#selectswith_or) instead of rolling your own alias?

katre (changes requested):
This looks good, most of the comments are minor.

src/test/shell/bazel/bazel_java14_test.sh
comius: No issues. I wanted to keep PR smaller, it is already quite big.

tools/jdk/default_java_toolchain.bzl
comius: Thanks. Appending `_definition` sounds much better.

tools/jdk/local_java_repository.bzl
comius: I didn't want to use `strip`, because it also removes inner spaces (https://docs.bazel.build/versions/3.7.0/skylark/lib/string.html#strip), but this is actually fine. Will re-write as a loop and use `strip`.

tools/jdk/local_java_repository.bzl
comius: `local_java_repository` is also used in `jdk.WORKSPACE` suffix. I would like to avoid importing skylib there.

tools/jdk/local_java_repository.bzl
comius: Actually for loop didn't look that nice - jumping out on first occurrence and handling case when version is not found looks worse. I opted for a more readable functional implementation.

katre (approved):
This property code is much more readable. Thanks!

PR #12590, opened 2020-12-01
Change test.xml from BasicActionInput to Artifact so it's metadata can be injected

### LINKED ISSUE #12554: Regenerate test.xml with generate-xml.sh when testing remotely with --remote_download_minimal

Normally, when running a test, Bazel expects the test to produce a `test.xml`. If it is not produced, Bazel will run a separate action using `generate-xml.sh` to generate that file.

When testing remotely with `--remote_download_minimal`, Bazel doesn't download the XML file, and then observes that it doesn't exist and so tries to regenerate it, regardless of whether the action actually produced it or not.

Bazel shouldn't regenerate the `test.xml` if the action actually produced it.

Java tests with junit produce `test.xml` automatically. You can verify by running a junit test locally (and Bazel won't run the `generate-xml.sh` action) and find the file under `bazel-testlogs` folder.

However, when running the test remotely with  `--remote_download_minimal`, you will notice that Bazel will run an additional `generate-xml.sh` action.

release 3.7.0

#6862
#11558

coeuvre:
This is achieved by adding test.xml to TestRunnerAction's outputs which means test.xml becomes a mandatory output of TestRunnerAction just like test.log.

test.xml should always be generated by either test actions or the separated spawn action enabled by --experimental_split_xml_generation.

Fixes #12554.


oquenchil (approved):
LGTM

aiuto:
Is this ready to go, or a WIP?

coeuvre:
Stil WIP

PR #12594, opened 2020-12-01
Patch grpc to fix cares selecting the wrong source when building for darwin_arm64 cpu.

thii:
This should use constraints instead of cpu flags, but that should be
fixed on the grpc repository. I just added new `config_setting`s
here to keep the patch size small.

This should allow building a Bazel binary for Apple Silicon hardware
from an Intel Mac with:

```
USE_BAZEL_VERSION=last_green bazel build -c opt --cpu=darwin_arm64 //src:bazel
```

(This needs last green Bazel because we need recent changes to
`@bazel_tools` which needs a Bazel release.)

### DIFF
--- a/third_party/grpc/grpc_1.32.0.patch
+++ b/third_party/grpc/grpc_1.32.0.patch
@@ -8,6 +8,8 @@ Date:   Wed Jun 3 15:35:31 2020 +0200
     libraries
     - Add bazel mirror URL for upb and cares
     - Redirect zlib to @//third_party/zlib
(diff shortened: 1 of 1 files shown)


thii:
(The bazel command above is bazelisk)

keith (approved):
Probably worth submitting upstream too.

thii:
Thanks, submitted it: https://github.com/grpc/grpc/pull/24872.

philwo (approved):

Issue #12497, opened 2020-11-17
Querying "//..." fails if a top-level target exists and any external dependency can't be fetched


jmillikin-stripe:
### Description of the problem / feature request:

If a workspace contains an external dependency that can't be fetched (upstream 404, shasum mismatch, whatever) then the presence of a build target in the top-level package will cause `bazel query` on `//...` to fail.

This happens regardless of whether the external dependency is in the transitive closure of the top-level target. As long as there is _any_ top-level target, and in the workspace there is _any_ target depending on the broken external dependency, then the query fails.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Given the following workspace:

```python
# WORKSPACE
workspace(name = "bazeltest_query_toplevel")
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "example_com_noexist",
    urls = ["https://example.com/noexist.tar.gz"],
    sha256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
)
```
```python
# BUILD
genrule(
    name = "gen_toplevel",
    outs = ["gen_toplevel.txt"],
    cmd = "touch $@",
)
```
```python
# subpkg/BUILD
genrule(
    name = "gen_subpkg",
    srcs = ["@example_com_noexist//:noexist.txt"],
    outs = ["gen_subpkg.txt"],
    cmd = "touch $@",
)
```

 `bazel query` fails:
```
$ bazel query 'kind("my_query_target", //...)'
[...]
Error in download_and_extract: java.io.IOException: Error downloading [https://example.com/noexist.tar.gz] to /private/var/tmp/_bazel_jmillikin/1d615cc9026acc39a3353897884f7725/external/example_com_noexist/temp2794112296295295544/noexist.tar.gz: GET returned 404 Not Found
ERROR: Evaluation of query "kind('my_query_target', //...)" failed due to BUILD file errors
Loading: 2 packages loaded
```

Removing the `gen_toplevel` target -- which has no relation to the failing dependency -- fixes the error, and lets `bazel query` successfully return an empty result set:
```
$ cat BUILD 
#genrule(
#    name = "gen_toplevel",
#    outs = ["gen_toplevel.txt"],
#    cmd = "touch $@",
#)
$ bazel query 'kind("my_query_target", //...)'
INFO: Empty results
Loading: 1 packages loaded
$
```

[...]

PR #12595, opened 2020-12-01
Fix the classic query package-loading cutoff optimization with external workspaces.

benjaminp:
A package's path fragment is not unambiguous when external workspaces are involved.

Fixes https://github.com/bazelbuild/bazel/issues/12497.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/query/BlazeQueryEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/query/BlazeQueryEnvironment.java
@@ -184,9 +184,9 @@ public class BlazeQueryEnvironment extends AbstractBlazeQueryEnvironment<Target>
       }
     }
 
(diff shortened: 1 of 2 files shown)


gregestren:
Reassigned to @meisterT for further triage since this is `query` (vs. `cquery`) logic.

meisterT (commented):
Can you add a simple test?

benjaminp:
Are there a lot of internal tests for query? `src/test/java/com/google/devtools/build/lib/query2/` seems pretty thin in this repository.

meisterT:
I assumed you would add the test here: https://github.com/bazelbuild/bazel/blob/master/src/test/shell/integration/bazel_query_test.sh

benjaminp:
I can do that. Normally, I make some effort to not write shell integration tests, though, since bash is not a very good programming language.

meisterT (approved):

PR #12603, opened 2020-12-02
Split tools/jdk/BUILD into part that's needed to produce @build_tools and the content of it.

comius:
There are two versions of Bazel tools available during compilation @bazel_tools//tools and @io_bazel//tools.

They are used in two different ways:
 - the actual targets are needed
 - source files are needed to create and embed a new version of @bazel_tools

Before the split //tools/jdk/BUILD file was used in both ways, even though it is only needed because of sources. Having everything in a single file caused evaluation by the compiling bazel. As a consequence it failed on targets/attributes that are not yet supported by it.

Splitting it into two files keeps BUILD.tools unevaluated and we can use unreleased features in it.


comius:
Note to reviewers: it's easier to review by commits, because there was a file moved.

meisterT (approved):

Issue #10949, opened 2020-03-11
Default `--instrumentation_filter` when 2 targets share a prefix is incorrect


keith:
The instrumentation_filter flag appears to to have a few different default cases:

1. If you're testing a single target like: `//Modules/ResourceSupport:ResourceSupportTests` it will default to `//Modules/ResourceSupport[/:]`
2. If you're testing 2 targets like: `//Modules/ResourceSupport:ResourceSupportTests` and
`//Modules/LyftKit:LyftKitTests` it will default to their common
prefix: `//Modules[/:]`
3. The unexpected case is if you test too targets like
`//Modules/TransitNearby:TransitNearbyTests` and
`//Modules/TransitNearbyFlow:TransitNearbyFlowTests` where they have
a common prefix, it defaults to `//Modules/TransitNearby[/:]` which
ends up excluding the `TransitNearbyFlow` files from the coverage manifest.

Am I correct in thinking that this is a bug and in that case it should either be `//Modules[/:]` or `//Modules/TransitNearby` without the `[/:]`

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 2.2.0

limdor:
I just came across this issue today. @lberki could we add the coverage label?
@caiyulun I saw that you created a PR some time ago, are you planning to continue it? I understand that the missing part are only tests right?

lberki:
sure can!

limdor:
@lberki I would like to fix this issue, taking the open PR that @caiyulun created. In the PR @caiyulun mentioned 4 months ago that was working on tests but I assume that something prevented to continue on the topic. However I have some question about the code base, would it be here, in the PR, or in bazel-dev the place to ask? For the moment I will put it here in case you can answer right a way and if not I can move it to some other channel.

What is missing in the PR is a test case but `InstrumentationFilterSupport` has directly no tests. I was checking if the test should derive from `BuildIntegrationTestCase` or `AnalysisTestCase` but I cannot get an answer. I thought that these instrumentation filters are computed during the analysis phase but the file `InstrumentationFilterSupport.java` is inside `buildtool` folder and not `analysis`. There is no test in `buildtool` that derive from `AnalysisTestCase`.

[...]

PR #12607, opened 2020-12-02
Fix common prefix for instrumentation filter

limdor:
In case that two path share a prefix, the instrumentation filter is
not computed proper.
https://github.com/bazelbuild/bazel/issues/10949
This PR takes as a base the one that was created by @caiyulun and
adds some tests.
Original PR: https://github.com/bazelbuild/bazel/issues/10949

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/buildtool/InstrumentationFilterSupportTest.java
+++ b/src/test/java/com/google/devtools/build/lib/buildtool/InstrumentationFilterSupportTest.java
@@ -1,0 +1,47 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/devtools/build/lib/buildtool/InstrumentationFilterSupportTest.java
```diff
@@ -0,0 +1,47 @@
...
+  @Test
+  public void testComputeInstrumentationFilter() throws Exception {
+    EventCollector events = new EventCollector(EventKind.INFO);
+    scratch.file("foo/BUILD", "sh_test(name='t', srcs=['t.sh'])");
+    scratch.file("foobar/BUILD", "sh_test(name='t', srcs=['t.sh'])");
+    List<Target> listOfTargets = new ArrayList<Target>();
+    listOfTargets.add(getTarget("//foo:t"));
+    listOfTargets.add(getTarget("//foobar:t"));
+    Collection<Target> targets = Collections.unmodifiableCollection(listOfTargets);
+    String expectedFilter = "^//foo[/:],^//foobar[/:]";
```
limdor, line 44: If prefered, I could split the PR in two commits to see more clear how the test would change with the changes.
Before the this PR the expected Filter would have been just ^//foo[/:]

c-mita: This is fine. We squash the PR into a single commit anyway, so there's limited value in splitting everything up incrementally in this case.

c-mita (approved):
LGTM.

limdor:
@c-mita some checks are failing but looks more like an http error connection.

philwo:
@c-mita and me are merging this right now and I'll make sure to get it into Bazel 4.0.0rc3! :)

PR #12617, opened 2020-12-03
fix build bugs in android sample

sheldonneuberger-sc:
- Add alwayslink otherwise the hello() C symbol is removed because it's not called by any C code.
- Add `-ldl` because libdl is no longer automatically linked.
- Bump target sdk to latest.
- Remove outdated android_sdk_repository instructions from README.
- Add workaround copt flag to fix ndk debugging on mac.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -45,3 +45,4 @@ build:macos --macos_minimum_os=10.10
 
 # User-specific .bazelrc
 try-import user.bazelrc
+build --copt="-fdebug-compilation-dir" --copt="/proc/self/cwd"
--- a/.bazelrc
+++ b/.bazelrc
@@ -38,6 +38,10 @@ build:ubuntu1604_java8 --host_platform=//:rbe_ubuntu1604_java8_platform
 build:ubuntu1604_java8 --platforms=//:rbe_ubuntu1604_java8_platform
 build:ubuntu1604_java8 --config=remote_shared
 
+# https://github.com/bazelbuild/intellij/issues/295#issuecomment-729656232
+# Configuration to fix ndk debugging on macos (uncomment this line on macos)
+# build --copt="-fdebug-compilation-dir" --copt="/proc/self/cwd"
+
 # Alias
 build:remote --config=ubuntu1604_java8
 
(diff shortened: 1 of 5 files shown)


.bazelrc
```diff
@@ -38,6 +38,10 @@ build:ubuntu1604_java8 --host_platform=//:rbe_ubuntu1604_java8_platform
 build:ubuntu1604_java8 --platforms=//:rbe_ubuntu1604_java8_platform
 build:ubuntu1604_java8 --config=remote_shared
 
+# https://github.com/bazelbuild/intellij/issues/295#issuecomment-729656232
+# Configuration to fix ndk debugging on macos (uncomment this line on macos)
+# build --copt="-fdebug-compilation-dir" --copt="/proc/self/cwd"
```
sheldonneuberger-sc, line 43: Couldn't find an easy way to enable a bazelrc line for only one platform, so I just commented it (https://github.com/bazelbuild/bazel/issues/5055).

ahumesky: one way to do this is to use a select with the target, i think it would look something like this:

```
cc_library(
    ....
    copts = select({
        "@bazel_tools//src/conditions:darwin": [
            "-fdebug-compilation-dir",
            "/proc/self/cwd",
        ],
        "//conditions:default": [],
    }),
    ....
```

sheldonneuberger-sc: Thanks that's similar to what I need, but can it be done in bazelrc? The user would want this to apply to all cc_library targets, without putting the copts into each cc_library.

bholmes-sc: You should be able to do this:

```
build:macos --copt="-fdebug-compilation-dir" --copt="/proc/self/cwd"
```

chancila: how would this even work on macos? macos has no proc?

sheldonneuberger-sc: > how would this even work on macos? macos has no proc? Is this strictly to have a consistent prefix for source files in the debug info?

The bazel android studio plugin is setting a source-map in lldb for /proc/self/cwd->[bazel workspace root]. See: https://github.com/bazelbuild/intellij/blob/edd4fb061dfcacb8a6d54ced39a147d11cfef811/aswb/src/com/google/idea/blaze/android/run/runner/BlazeAndroidDebuggerService.java#L80.

ahumesky: So would this work if you're not using the bazel android studio plugin? (e.g. just building from the command line for the first time)

PR #12622, opened 2020-12-04
Make gcov executable available in Starlark ToolchainInfo

torgil:
This is needed to use gcov in Starlark rules in custom coverage rules.


torgil:
ping @c-mita 
  
  Do you see any issues with this patch?

c-mita:
No "problems" per se. But we weren't really sure if we wanted to expose `gcov_executable` at this stage. Should we also do it for `llvm-cov`?
But I know it's needed if you want to write C++ coverage rules in Starlark...

Any thoughts @lberki or @oquenchil ?

oquenchil:
Charles and I discussed this offline and came to the conclusion that this is fine. We can expose the lcov one to Starlark when it's needed.

torgil:
I notice that GCC is provided as "compiler_executable", should this be named "coverage_executable" ?

oquenchil:
Yes, I agree that would be better.

Issue #12619, opened 2020-12-03
java coverage crash on 4.0.0rc2


alexeagle:
running `bazel coverage //...` in a big multi-lang monorepo.
Bazel version `37a429ad12b4c9e6a62dbae4881a1ff03b81ab40` (4.0.0-rc2)
rules_java at `d43b0aa2d3b2173fcc54d3c194e89fa66cdc6e1a` (current HEAD)

let me know if I can provide anything else short of a minimal repro

```

(21:13:45) FATAL: bazel crashed due to an internal error. Printing stack trace:
--
  | java.lang.RuntimeException: Unrecoverable error while evaluating node 'ConfiguredTargetKey{label=//test/java/com/rh/ledger:LedgerTest, config=BuildConfigurationValue.Key[f35c31b1baf1ab2bed1284160072656c995dfa7db8c9f4349087d4fbbb2e27de]}' (requested by nodes )
  | at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
  | at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
  | at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
  | Caused by: java.lang.NullPointerException
  | at com.google.devtools.build.lib.rules.java.JavaCompilationHelper.addCoverageSupport(JavaCompilationHelper.java:382)
  | at com.google.devtools.build.lib.bazel.rules.java.BazelJavaSemantics.addCoverageSupport(BazelJavaSemantics.java:613)
  | at com.google.devtools.build.lib.rules.java.JavaBinary.create(JavaBinary.java:199)
  | at com.google.devtools.build.lib.rules.java.JavaBinary.create(JavaBinary.java:65)
[...]

PR #12625, opened 2020-12-04
Fix NPE when coveragerunner is not set on the toolchain.

comius:
Fixes https://github.com/bazelbuild/bazel/issues/12619


comius:
I verified that this cleanly applies to 4.0.0 release baseline.

jin (approved):

PR #12626, opened 2020-12-04
Support symlinks for Android SDK repository

TimothyKlim:
For example, NixOS create a symlink for Android platform like:
```
ls -la /nix/store/azw6n236rgfkbmvfrbak4cnbcd4glbxn-androidsdk/libexec/android-sdk/platforms
total 0
dr-xr-xr-x 2 root root  24 Jan  1  1970 .
dr-xr-xr-x 7 root root 141 Jan  1  1970 ..
lrwxrwxrwx 1 root root  97 Jan  1  1970 android-29 -> /nix/store/2lll753i9y9j37b7mpsz5b81mspmpmnv-platforms-29/libexec/android-sdk/platforms/android-29
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/AndroidRepositoryFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/AndroidRepositoryFunction.java
@@ -83,12 +83,11 @@ abstract class AndroidRepositoryFunction extends RepositoryFunction {
   static final ImmutableSortedSet<Integer> getApiLevels(Dirents platformsDirectories) {
     ImmutableSortedSet.Builder<Integer> apiLevels = ImmutableSortedSet.reverseOrder();
     for (Dirent platformDirectory : platformsDirectories) {
-      if (platformDirectory.getType() != Dirent.Type.DIRECTORY) {
-        continue;
-      }
-      Matcher matcher = PLATFORMS_API_LEVEL_PATTERN.matcher(platformDirectory.getName());
-      if (matcher.matches()) {
-        apiLevels.add(Integer.parseInt(matcher.group(1)));
+      if (platformDirectory.getType() == Dirent.Type.DIRECTORY || platformDirectory.getType() == Dirent.Type.SYMLINK) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/bazel/rules/android/AndroidRepositoryFunction.java
```diff
@@ -83,7 +83,7 @@ final DirectoryListingValue getDirectoryListing(Path root, PathFragment dirPath,
   static final ImmutableSortedSet<Integer> getApiLevels(Dirents platformsDirectories) {
     ImmutableSortedSet.Builder<Integer> apiLevels = ImmutableSortedSet.reverseOrder();
     for (Dirent platformDirectory : platformsDirectories) {
-      if (platformDirectory.getType() != Dirent.Type.DIRECTORY) {
+      if (platformDirectory.getType() != Dirent.Type.DIRECTORY && platformDirectory.getType() != Dirent.Type.SYMLINK) {
```
ahumesky, line 86: So if the platform-* directories are all symlinks on some os's, wouldn't this skip all of the API levels and return nothing? I think android_sdk_repository will throw an error saying that no API levels are installed.

ahumesky: Looking at this again, I think I misread the code. This would be clearer I think if we rewrote it to this:

```java
for (Dirent platformDirectory : platformsDirectories) {
  if (platformDirectory.getType() == Dirent.Type.DIRECTORY || platformDirectory.getType() == Dirent.Type.SYMLINK) {
    Matcher matcher = PLATFORMS_API_LEVEL_PATTERN.matcher(platformDirectory.getName());
    if (matcher.matches()) {
      apiLevels.add(Integer.parseInt(matcher.group(1)));
    }
  }
}
```

TimothyKlim: fixed :+1:

ahumesky (approved):
thanks!

ahumesky:
i'll import this change internally now

PR #12628, opened 2020-12-04
Use release_archive to create java_tools' zips.

comius:
Github part of https://github.com/bazelbuild/bazel/pull/12615


philwo:
Merged in https://github.com/bazelbuild/bazel/commit/b04477fb97b697a09221832b3281e9647f616a0f

PR #12633, opened 2020-12-04
docs: clarify that urls is order sensitive

Aghassi:
According to https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java#L83-L84, `urls` is order sensitive. Noting it in the docs to make it clear to future users that order matters.

### DIFF
--- a/tools/build_defs/repo/http.bzl
+++ b/tools/build_defs/repo/http.bzl
@@ -219,7 +219,7 @@ Each entry must be a file, http or https URL. Redirections are followed.
 Authentication is not supported.
 
 URLs are tried in order until one succeeds, so you should list local mirrors first.
-If all downloads fail the rule will fail.""",
+If all downloads fail, the rule will fail.""",
     ),
     "sha256": attr.string(
         doc = """The expected SHA-256 of the file downloaded.
--- a/tools/build_defs/repo/http.bzl
+++ b/tools/build_defs/repo/http.bzl
@@ -216,7 +216,10 @@ to specify alternative URLs to fetch from.
             """A list of URLs to a file that will be made available to Bazel.
 
 Each entry must be a file, http or https URL. Redirections are followed.
-Authentication is not supported.""",
+Authentication is not supported.
+
+The resolution of this value is read top to bottom, and will fall through if a given
+url fails to fetch the artifact.""",
     ),
     "sha256": attr.string(
         doc = """The expected SHA-256 of the file downloaded.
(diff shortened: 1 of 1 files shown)


Aghassi:
@googlebot I signed it!

philwo (approved):

aiuto:
Now that it is approved, one of you should merge it.

epope18 (approved):

epope18 (approved):

epope18 (approved):

tools/build_defs/repo/http.bzl
```diff
@@ -216,7 +216,10 @@ to specify alternative URLs to fetch from.
             """A list of URLs to a file that will be made available to Bazel.
 
 Each entry must be a file, http or https URL. Redirections are followed.
-Authentication is not supported.""",
+Authentication is not supported.
+
+The resolution of this value is read top to bottom, and will fall through if a given
+url fails to fetch the artifact.""",
```
communikit, line 222: URL

tools/build_defs/repo/http.bzl
```diff
@@ -216,7 +216,10 @@ to specify alternative URLs to fetch from.
             """A list of URLs to a file that will be made available to Bazel.
 
 Each entry must be a file, http or https URL. Redirections are followed.
-Authentication is not supported.""",
+Authentication is not supported.
+
+The resolution of this value is read top to bottom, and will fall through if a given
```
communikit, line 221: Could this be reworded to something like "The list of URLs provided are evaluated in order, and an error could occur if a given URL fails to fetch the artifact."

Of course, feel free to add more emphasis or a better explanation if I didn't get the meaning quite right in the above phrase, but something like this would be a better explanatory sentence.

aiuto: The idea is that you do not get an error if one fails. You just try the next. You only get an error if ALL of them fail.
It could be better worded though, how about

URLs are tried in order until one succeeds, so you should list local mirrors first. If  all downloads fail the rule will fail.

communikit: +1, that sounds good.

Aghassi: Thanks, that's a clear way of putting it. Let me modify!

tools/build_defs/repo/http.bzl
```diff
@@ -218,8 +218,8 @@ to specify alternative URLs to fetch from.
 Each entry must be a file, http or https URL. Redirections are followed.
 Authentication is not supported.
 
-The resolution of this value is read top to bottom, and will fall through if a given
-url fails to fetch the artifact.""",
+URLs are tried in order until one succeeds, so you should list local mirrors first.
+If all downloads fail the rule will fail.""",
```
communikit, line 222: Nit: need to add a comma here. In general, looks good!

communikit (approved):

communikit (approved):

PR #12647, opened 2020-12-07
Update to java_tools javac11 release 10.5

comius:


comius:
I made java_tools_release-10.5 branch on bazelbuild, cherry-picked fix from cushon. Released java_tools 10.5.

This PR is now going into that branch in order to have everything reviews.

@philwo Perhaps it makes sense protecting back patch branch bazelbuild:java_tools_release-10.5

philwo:
@comius I have setup branch protection for `java_tools-release-*`, which should prevent accidental deletion and force pushes to these branches. Do I understand correctly that this is because we might need to do further backports to java_tools 10.x in the future?

comius:
> Do I understand correctly that this is because we might need to do further backports to java_tools 10.x in the future?
Yes. This is because of current backport and further ones, if there are any.

philwo (approved):

Issue #11301, opened 2020-05-06
'bazel info' reports 'unknown key' error for .bazelrc with Starlark flag


jayconrod:
### Description of the problem / feature request:

If the workspace's `.bazelrc` has a flag that sets a build setting defined in Starlark, `bazel info` fails with `ERROR: unknown key`. If `bazel info` is given an argument, it fails with `at most one key may be specified`.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

In the example workspace below, run `bazel info` or `bazel info output_base`.

```
-- WORKSPACE --
# empty file

-- BUILD.bazel --
label_flag(
    name = "example",
    build_setting_default = "WORKSPACE",  # default doesn't matter
)

-- .bazelrc --
build --//:example=//:WORKSPACE
```

```
$ bazel info
ERROR: unknown key: '--//:example=//:WORKSPACE'
$ bazel info output_base
ERROR: at most one key may be specified
```

### What operating system are you running Bazel on?

macOS / amd64 10.15.4

### What's the output of `bazel info release`?

```
ERROR: at most one key may be specified
```

If the .bazelrc file is removed or commented:

```
release 3.1.0
```

### Any other information, logs, or outputs that you want to share?

Originally reported in bazelbuild/rules_go#2475.

I switched rules_go's static analysis framework to using Starlark flags and configuration transitions. I'd like to recommend that users put the default static analysis binary in .bazelrc:

```
build --@io_bazel_rules_go//go/config:nogo=@io_bazel_rules_go//:tools_nogo
```

cc @katre @gregestren @juliexxia

gregestren:
@juliexxia is investigating. We believe there's a fast fix for this.

juliexxia:
It seems like bazel info doesn't need to know about build settings at all, so working on a change to ignore them for the info command.

PR #12648, opened 2020-12-07
Respect Starlark options with values in `removeStarlarkOptions()`

trustin:
Related: https://github.com/bazelbuild/bazel/issues/11301
(Partially fixed in https://github.com/bazelbuild/bazel/commit/2ec58f60ae1732f4c1d09330c4583050ac8d1f46)

Motivation:

`StarlarkOptionsParser.removeStarlarkOptions()` doesn't take the case
into account where the specified Starlark option has a value, e.g.
`--//my_rules/custom_flags:foo=bar`.

Modifications:

- Do not pass a Starlark flag value when validating the flag name.

Result:

`bazel info` does not fail anymore when `.bazelrc` contains a statement
like the following:

    build --//my_rules/custom_flags:foo=bar


trustin:
Gentle ping, Googlers :smile:

trustin:
Bump

trustin:
Thanks a lot @buchgr for helping this PR get merged. :bow:

Issue #10598, opened 2020-01-15
Binary targets on Windows use --run_under's runfiles manifest if the argument is a rule


gmishkin:
### Description of the problem / feature request:

On Windows only, when using `--run_under` with an argument that's a binary rule in the workspace (as opposed to some binary from PATH), the binary rule you're actually running with `bazel run` sees the run under rule's runfiles manifest and not its own.

Of course that manifest doesn't have much of anything in it and so rlocation of an actual data dep of the binary rule fails. It works fine on Mac and Linux.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

https://github.com/gmishkin/run_under_runfiles

`bazel run //:printer`

On Linux, it works fine. You see the contents of the data file from the "printer" rule and the "timer" run_under script prints the time it took.

On Windows it fails with `cat: '': No such file or directory` but you see the timer results from run_under. If you comment out the `--run_under` line in the .bazelrc you do see the contents of the data file, but not the elapsed time.

### What operating system are you running Bazel on?

Windows 10 version 1909

### What's the output of `bazel info release`?

release 2.0.0

###  Have you found anything relevant by searching the web?

Some stuff about runfiles on Windows but nothing about interaction with `--run_under`.

### Any other information, logs, or outputs that you want to share?

I'm using `--enable_runfiles=yes` on Windows but removing that has no effect one way or another.

meteorcloudy:
If --run_under points to a rule, it's better that the target itself make sure it doesn't pass its own RUNFILES_* env vars to the binary it invokes. In this case, 
```
diff --git a/timer.sh b/timer.sh
index df4efd1..7194607 100755
--- a/timer.sh
+++ b/timer.sh
@@ -2,4 +2,5 @@

 set -eu

+unset RUNFILES_MANIFEST_FILE
 time "$@"
```
will make it work.

PR #12650, opened 2020-12-07
Update find runfiles manifest file logic

mai93:
This pull request updates the `FindManifestFileImpl` function to look for the runfiles manifest file of the binary first. If it is not found, it can use the manifest file specified in the environment variables `RUNFILES_MANIFEST_FILE` if it exists. The same logic is applied in `runfiles_src.cc` to find runfiles manifest file / directory of cc binaries. 

This is done to avoid using the manifest file in `RUNFILES_MANIFEST_FILE` environment variable if the binary already has its own runfiles manifest file. While, at the same time, allow using the value in the environment variables for binaries run as data-dependency that have no runfiles manifest/directory and should use the main binary's.

Fixes: #10598


tools/cpp/runfiles/runfiles_src.cc
```diff
@@ -103,6 +103,12 @@ bool PathsFrom(const std::string& argv0, std::string runfiles_manifest_file,
                std::function<bool(const std::string&)> is_runfiles_directory,
                std::string* out_manifest, std::string* out_directory);
 
+bool PathsFrom(std::string runfiles_manifest_file,
```
meteorcloudy, line 106: Can you add some comments to explain what the PathsFrom functions do? I took me some time to parse them. I know we probably should have added them for the existing ones, but it's never too late to improve the readability ;)

meteorcloudy (changes requested):
Thanks for the fix!

meteorcloudy (approved):

PR #12653, opened 2020-12-07
Use configuration flags for //src/conditions:darwin_arm64

thii:
Partially reverts 6d637f4165445a0d6644f6e6f8a61b3f1de208d8, to allow
cross-compiling Bazel from darwin_x86_64 to darwin_arm64 with
the `--cpu=darwin_arm64` flag.

The only source that is different between darwin_x86_64 and darwin_arm64
now is the embedded JDK.

### DIFF
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -103,6 +103,10 @@ config_setting(
     values = {"cpu": "darwin_arm64"},
 )
 
+# Workaround for an issue where the dummy cc toolchain isn't being picked up
+# when cross-compile from darwin_x86_64 to darwin_arm64 cpu.
+# TODO(https://github.com/bazelbuild/bazel/issues/12655): Remove the flag based
+# select when the issue is resolved.
 selects.config_setting_group(
     name = "darwin_arm64",
     match_any = [
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -1,3 +1,5 @@
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
 filegroup(
     name = "srcs",
     srcs = glob(["**"]),
@@ -89,8 +91,24 @@ config_setting(
 )
 
 config_setting(
-    name = "darwin_arm64",
+    name = "darwin_arm64_constraint",
+    constraint_values = [
+        "@platforms//os:macos",
+        "@platforms//cpu:arm64",
+    ],
+)
+
+config_setting(
+    name = "darwin_arm64_flag",
     values = {"cpu": "darwin_arm64"},
+)
+
+selects.config_setting_group(
+    name = "darwin_arm64",
+    match_any = [
+        ":darwin_arm64_constraint",
+        ":darwin_arm64_flag",
+    ],
     visibility = ["//visibility:public"],
 )
 
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -90,10 +90,7 @@ config_setting(
 
 config_setting(
     name = "darwin_arm64",
-    constraint_values = [
-        "@platforms//os:macos",
-        "@platforms//cpu:arm64",
-    ],
+    values = {"cpu": "darwin_arm64"},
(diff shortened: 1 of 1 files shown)


thii:
@philwo @comius Could you take a look?

src/conditions/BUILD
```diff
@@ -90,10 +90,7 @@ config_setting(
 
 config_setting(
     name = "darwin_arm64",
-    constraint_values = [
-        "@platforms//os:macos",
-        "@platforms//cpu:arm64",
-    ],
+    values = {"cpu": "darwin_arm64"},
```
katre, line 93: Instead of removing the constraint version entirely, use https://github.com/bazelbuild/bazel-skylib/blob/master/docs/selects_doc.md#selectsconfig_setting_group to use either setting.

comius: Shouldn't whatever you're doing already work by setting both `--cpu` and `--platforms=//:target` with ```platform(name="target", constraint_values=[ "@platforms//os:macos", "@platforms//cpu:arm64"])```?

I don't know what state of CC crosscompilation platformization is. But specifying both platform and flags should get you to the same state as before conditions package was modified.

thii: Building with both `--cpu=darwin_arm64` and `--platforms=//:target` gets me this:

```
INFO: Invocation ID: 7514df6d-c786-4d10-8592-179657b0750b
DEBUG: /private/var/tmp/_bazel_admin/058bf0aeee6f9b65e93a9de41f4a4c2c/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1804_java11 not using checked in configs as detect_java_home was set to True
DEBUG: /private/var/tmp/_bazel_admin/058bf0aeee6f9b65e93a9de41f4a4c2c/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1604_java8 not using checked in configs as detect_java_home was set to True
INFO: Build option --platforms has changed, discarding analysis cache.
ERROR: While resolving toolchains for target //src/main/cpp:client: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
ERROR: Analysis of target '//src:bazel' failed; build aborted: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
INFO: Elapsed time: 0.475s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (47 packages loaded, 179 targets configured)
```

katre: Looks like the dummy cc toolchain isn't being picked up. Would you mind filing this as a separate issue so I can investigate? This PR looks fine to me with the recent change

katre (approved):
Looks fine to me, but I'll wait for @comius to comment.

thii:
Thanks for the review. I filed it here: https://github.com/bazelbuild/bazel/issues/12655.

src/conditions/BUILD
comius: I'd rather see the problem with dummy cc toolchain fixed than a workaroud.
What happens if you set `--incompatible_enable_cc_toolchain_resolution`?
Please at least open an issue for dummy cc toolchain and add a TODO here, that is
`TODO(issue): remove flag based select when the issue is resolved`

comius (approved):

src/conditions/BUILD
thii: I'm getting the same error even building with `--incompatible_enable_cc_toolchain_resolution`.

Issue #12081, opened 2020-09-10
The system for specifying Bazel's distribution dependencies is too complex


aiuto:
The scheme we use is needlessly complex, which makes it error prone. If you update a distribution dependency you might have to change a file name in 10 places.  Using *pkg_name* for the archive name, and *pkg_sha* for the sha256 of it, we have. 
```
distdir_tar(
  name = "additional_distfiles",
  archive = [...  pkg_name.tgz, ...]
  sha256 = { ..., pkg_name.tgz: pkg_sha },
  urls = {... pkg_name.tgz: [  http://mirror/....pkg_name.tgz, http://canonical_host/.../pkg_name.tgz ], ...},
)
```
Then, sometimes we repeat it again in `distdir_tar(name = "test_WORKSPACE_files", ...)`
or we use the name, sha256, and urls directly in an `http_archive` rule.

We could make this easier to update by consolidating the information into a single structure, something like

```
archives = {
   archive_1: { file_name: pkg_name.tgz, sha256: sha, url: [urls] }
   archive_2: ...
}
```

distdir_tar would change to take a list of keys from the archive rather than 3 aligned attributes (archive, sha26, urls)
instead of http_archive, we would use a wrapper method that just takes the archive name. 
 
Inspired by #12077

aiuto:
Current plan:

1. Make create a macro that does an http_archive for a repository from DIST_DEPS #12722
This reduces error-prone boilerplate from the top level WORKSPACE
2.  Generate embedded *.WORKSPACE files which must be in sync with dist_deps.  #12743 
     Example:  src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/cc_configure.WORKSPACE
    Add a rule that can create WORKSPACE stanzas from a template, using references to repositories in DIST_DEPS.
    - Replace the hand made *.WORKSPACE file 'srcs' for the package with the generated file.
    - Note that in the PRs that do this swap we must temporarily exclude the static file from srcs. See the TODO in src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BUILD in #12743
3. Iterate on dependencies until all elements //:additional_distfiles are moved to the DIST_DEPS list.

aiuto:
Status update:   I figured out the pattern for generating files like cc_configure.WORKSPACE. Now that that PR is in place, I can continue for Java.

aiuto:
Next round: #12779

aiuto:
Status update
[...]

PR #12654, opened 2020-12-07
Start making distribution dependencies easier to update.

aiuto:
This creates a mechanism to address #12081, but does not convert all use cases yet.

The basic idea is to change distdir_tar so that each dependency is defined in a block rather than split into distinct dictionaries for archive, sha256, and urls.  This means that updating a dependency can be isolated to 3 or 4 lines which appear together, rather than being sprinkled throughout WORKSPACE.

The next steps would be
1. do the same for the other dependencies. I want to see this land and stay that way before doing the rest.  Those PRs should be easy to review.
1. determine how to use DISTDIR_DEPS in the other embedded workspace files, such as src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/cc_configure.WORKSPACE
   - we need to splice distdir_deps.bzl into the dynamically created WORKSPACE files
   - or, export distdir_deps.bzl to the workspace and load it
   - or, since cc_configure.WORKSPACE is data, we generate it at build time.

These could be done in either order.

*other potential changes*

- streamline the http_archive and maybe instances to read directly from DISTDIR. This would eliminate the tedious pattern of
```
    sha256 = DIST_DEPS["rules_pkg"]["sha256"],
    urls = DIST_DEPS["rules_pkg"]["urls"],
```
- hide DISTDIR_DEPS from WORKSPACE.
   - pass just a list of package names (not archive names) to distfile_tar().
   - distdir.bzl loads the deps list directly and looks up data by package name


meteorcloudy (approved):
Thanks for the improvement!

aiuto:
I'll send a CL to import.
If there is no rollback, I will change the other repos to the new format later this week.

aiuto:
Closed in 337e717489d61418efaf957affdfdc96ba4c5372

PR #12658, opened 2020-12-08
[starlark] Force locale en_US in ScriptTest

Yannic:
Otherwise, the test can fail depending on the default locale of the system
(e.g., for me with locale `de`).

```
Executing tests from //src/test/java/net/starlark/java/eval:ScriptTest
./src/test/java/net/starlark/java/eval/testdata/float.star:53:10: called from <toplevel>
Error: assert_eq: "0,0000000000000000.0" != "0.0"
./src/test/java/net/starlark/java/eval/testdata/float.star:54:10: called from <toplevel>
Error: assert_eq: "0,0000000000000000.0" != "0.0"
./src/test/java/net/starlark/java/eval/testdata/float.star:58:10: called from <toplevel>
Error: assert_eq: "-1,5557538194652854e-61" != "-1.5557538194652854e-61"

...
```

### DIFF
--- a/src/test/java/net/starlark/java/eval/BUILD
+++ b/src/test/java/net/starlark/java/eval/BUILD
@@ -47,7 +47,10 @@ java_test(
     name = "ScriptTest",
     srcs = ["ScriptTest.java"],
     data = glob(["testdata/*.star"]),
-    jvm_flags = ["-Dfile.encoding=UTF8"],
+    jvm_flags = [
+        "-Duser.language=en_US",
+        "-Dfile.encoding=UTF8",
+    ],
     use_testrunner = False,
     deps = [
         "//src/main/java/net/starlark/java/annot",


Yannic:
@alandonovan PTAL

alandonovan:
Thanks for flagging this. The real problem is not that ScriptTests forgot to set the locale; it's that the behavior of Starlark's core operations such as number formatting depend on the locale. We need to make them all use the "C" locale so that behavior is deterministic and not a function of the user's environment, and make it clear in the spec that the locale is irrelevant.

I'm no expert, but my impression has long been that the POSIX locale mechanism is essentially broken. It is neither safe nor useful to change the behavior of number-formatting routines on a per-process basis. The correct display of a number depends on who is reading it, and for a server program, that varies from one request to another.

Issue #12579, opened 2020-11-27
generate-xml.sh fails to execute


Flamefire:
### Description of the problem / feature request:

I'm running some test of TensorFlow using bazel but on our multi-core POWER9 system it fails with e.g.

`ERROR: /dev/shm/s3248973-EasyBuild/TensorFlow/2.4.0/fosscuda-2019b-Python-3.7.4/TensorFlow/tensorflow-r2.4/tensorflow/core/platform/BUILD:1142:11:  failed (Exit 1): generate-xml.sh failed: error executing command `

I.e. there is no good error message, it simply failed to execute that script which comes from the Bazel installation. I verified that the executed command (`bazel -s`) runs correctly and the script hence also exists

I even modified that script in the Bazel sources to print something at the start but that doesn't show up. So it seems that script is not (yet?) created when Bazel tries to execute it. I hence expect a race condition or something but am unable to verify this.

Any hints, ideas, ...?

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Sorry, only thing I have is the command I use to test TF:
```
bazel --output_base=/dev/shm/s3248973-EasyBuild/TensorFlow/2.4.0/fosscuda-2019b-Python-3.7.4/tmptspeEg-bazel-tf/output_base --install_base=/dev/shm/s3248973-EasyBuild/TensorFlow/2.4.0/fosscuda-2019b-Python-3.7.4/tmptspeEg-bazel-tf/output_base/inst_base --output_user_root=/dev/shm/s3248973-EasyBuild/TensorFlow/2.4.0/fosscuda-2019b-Python-3.7.4/tmptspeEg-bazel-tf/output_user_root --host_jvm_args=-Xms512m --host_jvm_args=-Xmx4096m test --compilation_mode=opt --config=opt --subcommands --verbose_failures --config=noaws --jobs=64 --copt="-fPIC"  --distinct_host_configuration=false --test_output=errors --local_test_jobs=1 --build_tests_only --test_tag_filters='-no_gpu,-no_oss,-oss_serial,-benchmark-test,-no_oss_py37,-v1only' --build_tag_filters='-no_gpu,-no_oss,-oss_serial,-benchmark-test,-no_oss_py37,-v1only'  -- //tensorflow/core/... //tensorflow/cc/... //tensorflow/c/... -//tensorflow/core:example_java_proto -//tensorflow/core/example:example_protos_closure
```

### What operating system are you running Bazel on?

[...]

Issue #4137, opened 2017-11-20
ENV not propagated to process-wrapper


yliu120:
Please provide the following information. The more we know about your system and use case, the more easily and likely we can help.

### Description of the problem / feature request / question:
I have a problem using bazel to build the rules_scala on a custom platform. I am sure it is because the shell envs are not propagated correctly to a subprocess that is spawned by some actions of bazel. Please see the log information that I provide in the last section.

### Environment info
* Operating System:
CentOS 6.7

* Bazel version (output of `bazel info release`):
0.8.1

* If `bazel info release` returns "development version" or "(@non-git)", please tell us what source tree you compiled Bazel from; git commit hash is appreciated (`git rev-parse HEAD`):

###  Have you found anything relevant by searching the web?
(e.g. [StackOverflow answers](http://stackoverflow.com/questions/tagged/bazel),
[GitHub issues](https://github.com/bazelbuild/bazel/issues),
email threads on the [`bazel-discuss`](https://groups.google.com/forum/#!forum/bazel-discuss) Google group)

### Anything else, information or logs or outputs that would be helpful?
(If they are large, please upload as attachment or provide link).

The command I ran to build rules_scala is `bazel build -s --verbose_failures --sandbox_debug //src/...`

```
>>>>> # //src/scala/io/bazel/rules_scala/tut_support:tut_compiler [action 'scala //src/scala/io/bazel/rules_scala/tut_support:tut_compiler']
(cd /home-4/yliu120@jhu.edu/.cache/bazel/_bazel_yliu120@jhu.edu/12dd3863654b107695e643fa774ca856/execroot/io_bazel_rules_scala && \
  exec env - \
  /bin/bash -c '
rm -f bazel-out/local-fastbuild/bin/src/scala/io/bazel/rules_scala/tut_support/tut_compiler.jar
external/bazel_tools/tools/zip/zipper/zipper c bazel-out/local-fastbuild/bin/src/scala/io/bazel/rules_scala/tut_support/tut_compiler.jar @bazel-out/local-fastbuild/bin/src/scala/io/bazel/rules_scala/tut_support/bazel-out/local-fastbuild/bin/src/scala/io/bazel/rules_scala/tut_support/tut_compiler.jar_zipper_args
')
>>>>> # @scala//:scala-reflect [action 'Extracting interface @scala//:scala-reflect [for host]']
[...]

PR #12659, opened 2020-12-08
Propagate test envs to xml generation action

meteorcloudy:
Previously, we hardcode the envs of the xml generation action, which
caused problem for process-wrapper because it's dynamically linked to
some system libraries and the required PATH or LD_LIBRARY_PATH are not
set.

This change propagate the envs we set for the actual test action to the
xml file generation action to make sure the env vars are correctly set and
can also be controlled by --action_env and --test_env.

Fixes https://github.com/bazelbuild/bazel/issues/4137
Fixes https://github.com/bazelbuild/bazel/issues/12579

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
@@ -404,7 +404,7 @@ public class StandaloneTestStrategy extends TestStrategy {
     envBuilder.putAll(testEnv).put("TEST_NAME", action.getTestName());
     if (!action.isSharded()) {
       envBuilder.put("TEST_SHARD_INDEX", "0");
-      envBuilder.put("TEST_SHARD_INDEX", "0");
+      envBuilder.put("TEST_TOTAL_SHARDS", "0");
     }
     return new SimpleSpawn(
         action,
--- a/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
@@ -401,7 +401,12 @@ public class StandaloneTestStrategy extends TestStrategy {
             Long.toString(result.getWallTime().orElse(Duration.ZERO).getSeconds()),
             Integer.toString(result.exitCode()));
     ImmutableMap.Builder<String, String> envBuilder = ImmutableMap.builder();
+    // "PATH" and "TEST_BINARY" are also required, they should always be set in testEnv.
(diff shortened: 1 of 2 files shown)


meteorcloudy:
fyi @ulfjack

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
```diff
@@ -391,26 +391,25 @@ private static void writeOutFile(Path inFilePath, Path outFilePath) throws IOExc
...
+    envBuilder.putAll(testEnv).put("TEST_NAME", action.getTestName());
+    if (!action.isSharded()) {
+      envBuilder.put("TEST_SHARD_INDEX", "0");
+      envBuilder.put("TEST_TOTAL_SHARDS", "0");
+    }
     return new SimpleSpawn(
         action,
         args,
-        ImmutableMap.of(
-            "PATH", "/usr/bin:/bin",
```
meisterT, line 409: Are you sure we don't need the `PATH` and the `TEST_BINARY`?

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
```diff
@@ -391,26 +391,25 @@ private static void writeOutFile(Path inFilePath, Path outFilePath) throws IOExc
...
             action.getXmlOutputPath().getPathString(),
             Long.toString(result.getWallTime().orElse(Duration.ZERO).getSeconds()),
             Integer.toString(result.exitCode()));
-
-    String testBinaryName =
-        action.getExecutionSettings().getExecutable().getRootRelativePath().getCallablePathString();
+    ImmutableMap.Builder<String, String> envBuilder = ImmutableMap.builder();
+    envBuilder.putAll(testEnv).put("TEST_NAME", action.getTestName());
+    if (!action.isSharded()) {
+      envBuilder.put("TEST_SHARD_INDEX", "0");
```
meisterT, line 406: We do we write those two when the action is not sharded?

meteorcloudy: Np, we don't have those for the actual test action, only set for the xml generation action.

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
meteorcloudy: Those two are always set by the test action. Because are passing the test action envs to the xml generation action, we don't need to explicitly set them anymore.

meisterT (approved):

ulfjack:
Thanks!

PR #12661, opened 2020-12-08
Add windows_msvc back to conditions.

comius:
This is partial revert of 6d637f4165445a0d6644f6e6f8a61b3f1de208d8.

windows_msvc condition is used downstream by tensorflow via ruy.

The culprit line is in https://github.com/google/ruy/blob/master/ruy/build_defs.bzl#L60 (and #L67,#L77).

### DIFF
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -150,6 +150,12 @@ config_setting(
 )
 
 config_setting(
+    name = "windows_msvc",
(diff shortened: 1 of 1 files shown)


comius:
@katre, @gregestren can you please advise how "ruy" code should be modified to work with CC toolchain resolution instead of command line flags.

katre (approved):

PR #12681, opened 2020-12-10
Add the parse_netrc function

mrkkrp:
The existing `read_netrc` function can only be used from repository rules.
This makes it harder to e.g. test code that builds on top of `read_netrc`.
The new function is pure and can be easily tested.

### DIFF
--- a/tools/build_defs/repo/utils.bzl
+++ b/tools/build_defs/repo/utils.bzl
@@ -213,6 +213,19 @@ def read_netrc(ctx, filename):
       about them
     """
     contents = ctx.read(filename)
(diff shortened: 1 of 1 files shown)


aiuto:
I'm sending this on to the external deps team. The question is if we want to get out of the download business entirely and eventually not have read_netrc.

Wyverald (approved):
I think this change is well acceptable. Even though we might eventually want to get out of the netrc business, there's no harm in making testing a little easier right now. This PR is not making matters worse since `read_netrc` is already accessible by repo rules.

Issue #12685, opened 2020-12-11
jdeps_test fails just on RBE


meisterT:
The jdeps test should make sure that we don't accidentally add deps on new JDK modules breaking Bazel because we only embed a minimized JDK.

We are seeing this test failing in presubmit, but only on RBE claiming that `java.management` should be removed.
This blocks even docs only changes, like https://buildkite.com/bazel/google-bazel-presubmit/builds/42189#7a0e47e2-b9ca-47f2-b71d-04e19caa98f7

Example log: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/f1789ffa-fd21-400c-a8d7-5f9c7e262dc0/src/test/shell/bazel/jdeps_test/attempt_1.log

meisterT:
Apparently we don't run tests on RBE in postsubmit, why?

philwo:
Is this still an issue?

meisterT:
I think so, but https://github.com/bazelbuild/bazel/pull/12982 will tell us.

https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/8288 is the run on buildkite

PR #12687, opened 2020-12-11
Also run RBE testing in postsubmit and align the two configs.

meisterT:
The skew was discovered when looking at https://github.com/bazelbuild/bazel/issues/12685


meteorcloudy:
You can push a branch to the Bazel git repo, then start a new build with that branch at https://buildkite.com/bazel/bazel-bazel

meteorcloudy (approved):

Issue #12403, opened 2020-11-02
Conflicting actions when using --host_action_env


martinxsliu:
### Description of the problem / feature request:

When using the newly added `--host_action_env` flag, the analysis phase fails with lots of `file X is generated by these conflicting actions` type errors.

Example error log:

```
ERROR: file 'external/local_config_cc/watchos_i386.runfiles/MANIFEST' is generated by these conflicting actions:
Label: @local_config_cc//:watchos_i386
RuleClass: cc_toolchain_config rule
Configuration: cafef9f8b89de695d623b93423eb2b4d32ac34d4f9b36496f8139f880d1f4901, 9177372694e283f2d9a7ff337e520a3c2f33deeddac5d93aa911522a084518e2
Mnemonic: SymlinkTree
Action key: b5ecb575df0ad57da108b32678bdc98502e8f791db66d410fbe929282f7722a7, e1b483e3ef545df17ef13fe03742f0be50941c7e2835dc3560ae5b4c8878fe15
Progress message: Creating runfiles tree bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/local_config_cc/watchos_i386.runfiles
PrimaryInput: File:[[<execution_root>]bazel-out/darwin-opt-exec-2B5CBBC6/bin]external/local_config_cc/watchos_i386.runfiles_manifest
PrimaryOutput: File:[[<execution_root>]bazel-out/darwin-opt-exec-2B5CBBC6/bin]external/local_config_cc/watchos_i386.runfiles/MANIFEST
Owner information: ConfiguredTargetKey{label=@local_config_cc//:watchos_i386, config=BuildConfigurationValue.Key[cafef9f8b89de695d623b93423eb2b4d32ac34d4f9b36496f8139f880d1f4901]}, ConfiguredTargetKey{label=@local_config_cc//:watchos_i386, config=BuildConfigurationValue.Key[9177372694e283f2d9a7ff337e520a3c2f33deeddac5d93aa911522a084518e2]}
MandatoryInputs: are equal
Outputs: are equal
WARNING: errors encountered while analyzing target '@local_config_cc//:watchos_i386': it will not be built
```

Most failures are for `local_config_cc`, but there are also 2 for the protobuf library. They are not directly related to the targets I'm trying to build.

```
ERROR: file 'external/local_config_cc/watchos_armv7k.runfiles/MANIFEST' is generated by these conflicting actions:
[...]

PR #12694, opened 2020-12-14
Pass --host_action_env to host options hostActionEnvironment attribute

mai93:
This PR passes the value of `host_action_env` to `hostActionEnvironment` attribute of the host options in `CoreOptions.getHost`. 

Fixes: #12403

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
@@ -109,9 +109,8 @@ public class ExecutionTransitionFactory
 
     @Override
     public BuildOptions patch(BuildOptionsView options, EventHandler eventHandler) {
-      if (executionPlatform == null || options.get(CoreOptions.class).isExec) {
-        // If no execution platform is known or the input already comes from execution configuration,
-        // don't change anything and return the existing value.
+      if (executionPlatform == null) {
+        // No execution platform is known, so don't change anything.
         return options.underlying();
       }
       return cache.applyTransition(
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
@@ -109,8 +109,9 @@ public class ExecutionTransitionFactory
 
     @Override
     public BuildOptions patch(BuildOptionsView options, EventHandler eventHandler) {
-      if (executionPlatform == null) {
-        // No execution platform is known, so don't change anything.
+      if (executionPlatform == null || options.get(CoreOptions.class).isExec) {
+        // If no execution platform is known or the input already comes from execution configuration,
+        // don't change anything and return the existing value.
         return options.underlying();
       }
       return cache.applyTransition(
--- a/src/test/shell/integration/action_env_test.sh
+++ b/src/test/shell/integration/action_env_test.sh
@@ -44,6 +44,27 @@ sh_test(
     name = "test_env_foo",
     srcs = ["test_env_foo.sh"],
 )
+
+genrule(
+  name = "show_host_env_using_tools",
+  tools = ["env.txt"],
+  outs = ["tools_env.txt"],
+  cmd = "cp \$(location env.txt) \"\$@\""
+)
+
+genrule(
+  name = "show_host_env_using_exec_tools",
+  exec_tools = ["env.txt"],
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
```diff
@@ -109,8 +109,9 @@ public boolean isHostTransition() {
 
     @Override
     public BuildOptions patch(BuildOptionsView options, EventHandler eventHandler) {
-      if (executionPlatform == null) {
-        // No execution platform is known, so don't change anything.
+      if (executionPlatform == null || options.get(CoreOptions.class).isExec) {
```
katre, line 112: This isn't safe: it's entirely possible to need to build tools in order to build other tools. How does this lead to the error with action_env?

mai93: I thought that the error of conflicting actions may be caused by having two copies of the same execution transition. Adding this check fixed the --host_action_env conflicting actions issue, so I created this PR to just to make sure it does not break other tests.

katre: If the two configurations were exactly the same, they would have the same hash. Can you check what the difference between the two is? I do not think this change is safe to submit.

You can diff two configurations by using the "bazel config" command, with the same flags, after you've run the command to cause the error. The config subcommand checks for configurations present in skyframe's cache, so anything that invalidates the cache will make this fail.

mai93: The only difference between the 2 configurations is com.google.devtools.build.lib.analysis.config.CoreOptions.action_env where one config includes the env var passed with host_action_env and the other doesn't.

katre: I see that the value of `hostActionEnvironment` isn't being kept in `CoreOptions.getHost`. If you add the line, does that deal with the issue?

```
    // Pass host action environment variables
    host.actionEnvironment = hostActionEnvironment;
    host.hostActionEnvironment = hostActionEnvironment;
```

katre:
Can you add a test for the host_action_env issue? It looks like there aren't any tests for the flag at all, putting them in https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/integration/action_env_test.sh?ss=bazel&q=action_env would probably be appropriate.

.bazelci/postsubmit.yml
```diff
@@ -257,16 +257,52 @@ tasks:
       - test
   rbe_ubuntu1604:
     shell_commands:
-      - sed -i.bak -e 's/^# android_sdk_repository/android_sdk_repository/'
-        -e 's/^# android_ndk_repository/android_ndk_repository/'
-        -e 's/^# rbe_autoconfig/rbe_autoconfig/'
-        -e 's/^# load("@bazel_toolchains/load("@bazel_toolchains/' WORKSPACE
+      - sed -i.bak
```
katre, line 260: Why are there changes in the postsubmit and presubmit? We should get these reviewed by someone on who knows them better than I do.

mai93: I included these commits to fix the failing presubmit jdeps_test as mentioned in #12685

katre: They should be submitted (and reviewed) separately, since it's not directly part of fixing this issue.

mai93: They are already submitted, rebasing the pull request should fix this, right?

katre: Oh, I see. Yes, rebase onto master and that should be fine.

src/test/shell/integration/action_env_test.sh
```diff
@@ -218,4 +239,205 @@ function test_action_env_changes_honored {
...
+
+function test_host_env_using_successive_exec_tools_simple() {
+  export FOO=baz
+
+  bazel build --host_action_env=FOO=bar pkg:show_host_env_using_successive_exec_tools \
+      || fail "${PRODUCT_NAME} build show_host_env_using_successive_exec_tool failed"
+
+  cat `bazel info ${PRODUCT_NAME}-genfiles`/pkg/successive_exec_tools_env.txt > $TEST_log
+  expect_log "FOO=bar"
+
```
katre, line 440: Remove this extra blank line.

katre (approved):
One small nit, but otherwise this looks great. Thanks for all the work on this.

Issue #7849, opened 2019-03-27
incompatible_use_toolchain_resolution_for_java_rules: use toolchain resolution for Java rules


lberki:
**Flag:** `--incompatible_use_toolchain_resolution_for_java_rules`
**Available since:** 5.0.0
**Will be flipped in:** 5.0.0
**Tracking issue:** #4592

### **Motivation**

Currently, Java rules find their Java toolchain and JDK using the `--javabase` / `--java_toolchain` / `--host_javabase` / `--host_java_toolchain` command line options. This will be changed to use [platform-based toolchain resolution](https://docs.bazel.build/versions/master/toolchains.html) so as to be consistent with the rest of Bazel and to support multiple platforms more easily.

### **Migration notes**

For `--javabase` with old values: 
- `@local_jdk://jdk`
- `@remotejdk11_{linux,window,darwin}_{cpu}//:jdk`
- `@remotejdk14_{linux,window,darwin}//:jdk`
- `@remotejdk15_{linux,window,darwin}.*//:jdk`

Replace the flag with `--java_runtime_version={local_jdk,remotejdk_14,remotejdk_15}`. 

For `--java_toolchain` with old values: 
- `@bazel_tools//tools/jdk:toolchain`, 
- `@bazel_tools_hostjdk8`, 
- `@bazel_tools//jdk:legacy_toolchain`,
- `@bazel_tools//tools/jdk:remote_toolchain`,
- `@bazel_tools//tools/jdk:toolchain_java_{ver}`,
- `@remote_java_tools_xxx//:toolchain`,
- `@remote_java_tools_xxx//:toolchain_jdk_11`,
- `@remote_java_tools_xxx//:toolchain_jdk_14`,
- `@remote_java_tools_xxx//:toolchain_jdk_15`

Replace the flag with `--java_language_version={8,...,15}`

### **Migration of more advanced cases**

For `--javabase=@bazel_tools//tools/jdk:​absolute_javabase`, use `local_java_repositoy` in the `WORKSPACE` file.

For custom `--javabase` labels, do the following:
1. replace `http_archive` with `remote_java_repository`:

```py
remote_java_repository(
   name = ...
   sha256 = ...
   strip_prefix = ...
   urls = ...
   prefix = "myjdk",
   version = "11",
   exec_compatible_with = ["@platforms//cpu:arm", "@platforms//os:linux"]
)
```

[...]

PR #12695, opened 2020-12-14
Fix Java tests to use Java toolchain resolution.

comius:
The flag is flipped for tests, old configuration options `javabase` and `java_toolchain` are marked deprecated. All the tests are fixed to execute using new toolchain resolution.

Changed `local_java_repository` to generate `java_runtime` with `java_home` parameter set (instead of srcs). This behaves better, because `java_stub_template.txt` then does not need runfiles to execute produced `deploy.jar`. Kept the same version with `srcs` in the `jdk.WORKSPACE`.

Fixes https://github.com/bazelbuild/bazel/issues/4592.
Flips https://github.com/bazelbuild/bazel/issues/7849.


katre (commented):
With all of this code deleted, can users actually unset the flag? If someone needs to do "bazel build --noincompatible_use_toolchain_resolution_for_java_rules", will that work, or will it fail?

comius:
> With all of this code deleted, can users actually unset the flag? If someone needs to do "bazel build --noincompatible_use_toolchain_resolution_for_java_rules", will that work, or will it fail?

I will fail, I should remove the flag as well.

katre:
That's the opposite of what I would prefer: we should only flip the flag in this release, wait a bazel release, and **then** remove the code.

comius:
> That's the opposite of what I would prefer: we should only flip the flag in this release, wait a bazel release, and **then** remove the code.

Ok. I reverted the removal and kept everything else.

katre (approved):
I am very excited to see this!

comius:
I will delay flipping the flag a bit, because RBE would potentially break on downstream projects. I need to prepare couple of fixes in both bazel and bazel_toolchains.

PR #12704, opened 2020-12-15
Remove support for javabase and java_toolchain flags.

comius:
Removes JavaRuntimeAliasRule, JavaToolchainAliasRule and its targets. Makes javabase and java_toolchain flags a no-op. Removes backwards support from java_toolchain_alias, java_runtime_alias, java_runtime_version_alias.


comius:
Downstream tests at: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2142

src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
```diff
@@ -71,41 +71,34 @@ public ImportDepsCheckingLevelConverter() {
               + " instead; see #7849")
   public boolean disallowLegacyJavaToolchainFlags;
 
-  private static final String DEFAULT_JAVABASE = "@bazel_tools//tools/jdk:jdk";
-
   @Deprecated
   @Option(
       name = "javabase",
```
katre, line 76: Do you want to move these to [`BazelGraveyardOptions`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRulesModule.java;bpv=1;bpt=1;l=53?ss=bazel&q=graveyard&gsn=BuildGraveyardOptions&gs=kythe%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel%3Flang%3Djava%3Fpath%3Dcom.google.devtools.build.lib.bazel.rules.BazelRulesModule.BuildGraveyardOptions%23d805dab9a7881823fba3226a83075552e8e9a4e3c1b9e28d48de5691b32db6cc)?

katre (approved):
This is pretty exciting. LGTM. Pinging @cushon as an FYI.

src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
comius: I rather keep them here. We had some issues when an option was not copied to all of the *GraveyardOption.

katre: SGTM

lberki:
!!!! !!!!! !!!!

tools/jdk/java_toolchain_alias.bzl
```diff
@@ -123,13 +86,11 @@ _java_runtime_transition = transition(
 )
 
 java_runtime_version_alias = rule(
-    implementation = _java_runtime_version_alias,
+    implementation = _java_runtime_alias,
```
cushon, line 89: Maybe add some documentation here: IIUC this is now deprecated, and has the same behaviour as `_java_runtime_alias`, and anyone who wants to configure a specific Java major version should use a different approach?

cushon (approved):
🎉

tools/jdk/java_toolchain_alias.bzl
comius: Just the implementation is the same. You missed `cfg = _java_runtime_transition` in the definition. `java_runtime_version_alias` still selects a specific Java major version. `java_runtime_alias` doesn't have the configuration transition.

PR #12708, opened 2020-12-16
Make the navigation sidebar versioned

jin:
Fixes: #6937

This adds versioning support to the navigation sidebar in the default documentation page layout, conditionally loading a sidebar partial based on the current version's major version, or use the latest one in the case for rolling releases. This also means that the navigation sidebar list is [extracted out from the `documentation.html` layout file](https://github.com/bazelbuild/bazel/pull/12708/files#diff-9f04c21a8bdeaac5a1ca37b571ac0e343acbf1a8d50f5d3d11a4ffb7962198f8), into an include partial.

This gets rid of the many conditionals littered all over the navigation list items.

Currently, the biggest differences are between the groups of v0, v1-3, and master. This implementation allows v1 and v2's sidebar to just be a symlink to v3. For rolling releases, the sidebar is versioned as `master`.

```
% tree _includes/documentation-sidebar
_includes/documentation-sidebar
├── 0
│   └── sidebar.html
├── 1 -> 3
├── 2 -> 3
├── 3
│   └── sidebar.html
└── master
    └── sidebar.html
```

We can also allow finer-grained versioning with the full major.minor.patch, but I think major versions are a good enough start.


meisterT:
Thanks for working on this Jin! What's the expected workflow on updating this? Should we add an item to https://github.com/bazelbuild/continuous-integration/blob/master/docs/release-playbook.md?

jin:
Depending on whether we want sidebars versioned on the major version or the full version, the release manager would just need to create a new symlink or new directory in the versioned sidebar doc. 

Alternative, we have a generate_versioned_docs setup script, can probably add something in there..

floriographygoth (approved):
LGTM.

PR #12714, opened 2020-12-17
Coverage: Expose Jacoco Dependencies

sgammon:
This changeset exposes aliases for Jacoco and ASM dependencies, which can then be leveraged downstream in projects that extend Bazel and want to include integrated coverage support.

### DIFF
--- a/tools/jdk/BUILD.tools
+++ b/tools/jdk/BUILD.tools
@@ -285,6 +285,51 @@ alias(
     actual = "@remote_java_tools//:jacoco_coverage_runner",
 )
 
+alias(
+    name = "asm",
+    actual = "//third_party:asm",
+)
+
+alias(
+    name = "asm-analysis",
+    actual = "//third_party:asm-analysis",
+)
+
+alias(
+    name = "asm-commons",
+    actual = "//third_party:asm-commons",
+)
+
+alias(
+    name = "asm-tree",
+    actual = "//third_party:asm-tree",
+)
+
+alias(
+    name = "asm-util",
+    actual = "//third_party:asm-util",
+)
+
+alias(
+    name = "jacoco-agent",
+    actual = "//third_party/java/jacoco:agent",
+)
+
+alias(
+    name = "jacoco-core",
+    actual = "//third_party/java/jacoco:core",
+)
+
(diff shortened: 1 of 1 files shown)


meisterT:
Can this be closed?

sgammon:
@meisterT I actually do still need it downstream in [`rules_graalvm`](https://github.com/sgammon/rules_graalvm). It would be great if this could eventually be merged.

hvadehra:
cc @meteorcloudy who has a project to empty `third_party/`

meteorcloudy:
I don't think this is the right way to go. If other projects need those dependencies, you can fetch them via rules_jvm_external. We already migrated most of Bazel's java dependencies to rules_jvm_external and plan to do the same for the left overs.

c-mita:
What do you want to do with this?

Use of Jacoco is very much an implementation detail as far as the Java rules are concerned and we make no promises about forever continuing to use it.

sgammon:
In this (very specific) case, the [GraalVM rules](https://github.com/sgammon/rules_graalvm) may be [compiling a Java binary to a native image](https://www.graalvm.org/latest/reference-manual/native-image/), and it may notice via Bazel's [flags](https://docs.bazel.build/versions/4.0.0/skylark/lib/configuration.html#coverage_enabled) that coverage instrumentation has been requested.

We're working with JARs here which were created by Bazel, so they are instrumented with the version of JaCoCo used by Bazel internally.

To instrument the native image, an identical version of JaCoCo must be included as an agent during the `native-image` build. I can put together a reproducer which surfaces the specific exception.

For users wanting to use some arbitrary version of JaCoCo, they should be consuming that via `rules_jvm_external` et al. But in this case, we need to pin to Bazel's specific version of JaCoCo, since it is effectively being passed on transitively from `java_[binary|library]`.

It would be fine to expose this via the Java toolchain, or via some other internal target that is appropriate for use from rules only. Perhaps the [Java fragment](https://bazel.build/rules/lib/fragments/java) could be a candidate?

sgammon:
@fmeum Yes, that may work and would be a better approach. I hadn't seen that, I'll give it a try. Thanks for the pointer

PR #12431, opened 2020-11-07
Remove --incompatible_blacklisted_protos_requires_proto_info

Yannic:
Cleanup after flipping the flag in Bazel 4.0 (53705f6a014067a8f17cf0af4da0ad99fe1169be)

PR #12727, opened 2020-12-17
Remove {direct,transitive}OriginalSources field from ProtoInfo

Yannic:
The information is already part of another field.

Cleanup after #12431 and #10939

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
@@ -33,10 +33,10 @@ public class ProtoLangToolchain implements RuleConfiguredTargetFactory {
   @Override
   public ConfiguredTarget create(RuleContext ruleContext)
       throws InterruptedException, RuleErrorException, ActionConflictException {
-    NestedSetBuilder<Artifact> forbiddenProtos = NestedSetBuilder.stableOrder();
+    NestedSetBuilder<ProtoSource> providedProtoSources = NestedSetBuilder.stableOrder();
     for (ProtoInfo protoInfo :
         ruleContext.getPrerequisites("blacklisted_protos", ProtoInfo.PROVIDER)) {
-      forbiddenProtos.addTransitive(protoInfo.getOriginalTransitiveProtoSources());
+      providedProtoSources.addTransitive(protoInfo.getTransitiveSources());
     }
 
     return new RuleConfiguredTargetBuilder(ruleContext)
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchainProvider.java
```diff
@@ -40,15 +42,46 @@
...
+  @Deprecated
   public static ProtoLangToolchainProvider create(
       String commandLine,
       FilesToRunProvider pluginExecutable,
       TransitiveInfoCollection runtime,
       NestedSet<Artifact> blacklistedProtos) {
     return new AutoValue_ProtoLangToolchainProvider(
-        commandLine, pluginExecutable, runtime, blacklistedProtos);
+        commandLine, pluginExecutable, runtime, ImmutableList.of(), blacklistedProtos);
   }
```
comius, line 86: I checked internally and the method may be safely deleted. And then I believe you can use NestedSet for providedProtoSources.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchainProvider.java
```diff
@@ -40,15 +42,46 @@
...
   public abstract TransitiveInfoCollection runtime();
 
+  /**
+   * Returns a list of {@link ProtoSource}s that are already provided by the protobuf runtime (i.e.
+   * for which {@code <lang>_proto_library} should not generate bindings.
+   */
+  public abstract ImmutableList<ProtoSource> providedProtoSources();
+
+  @Deprecated
   public abstract NestedSet<Artifact> blacklistedProtos();
```
comius, line 53: Keep. Can't be safely removed.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
```diff
@@ -45,7 +45,14 @@ public ConfiguredTarget create(RuleContext ruleContext)
...
                 ruleContext.getPrerequisite("runtime"),
-                blacklistedProtos.build()))
+                // We intentionally flatten the NestedSet here.
+                //
+                // `providedProtoSources` are read during analysis, so flattening the set here once
+                // avoid flattening it in every `<lang>_proto_aspect` applied to `proto_library`
+                // targets. While this has the potential to use more memory than using a NestedSet,
+                // there are only a few `proto_lang_toolchain` targets in every build, so the impact
+                // on memory consumption should be neglectable.
+                providedProtoSources.build().toList()))
```
comius, line 55: Why do you need to flatten the NestedSet? If I understand this correctly, so that you can use overloaded ProtoLangToolchainProvider.create? I'd prefer if you remove the old implementation of create and just use NestedSet. I checked that there are no dependencies internally. If you're worried about other forks of Bazel, perhaps you could change the create method name?

comius (changes requested):
Nice work! Thanks!

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchain.java
comius: Friendly ping.

Yannic: No, this not for overloading `ProtoLangToolchainProvider#create()`. Today, every `<lang>_proto_library` has to flatten the set to figure out which of its `srcs` is in `blacklistedProtos()`, which does have performance overhead. We can avoid that by flattening it here once and propagate the excluded protos as list instead. We trade some (finite) memory for performance.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoLangToolchainProvider.java
Yannic: removed

Yannic (commented):
Sorry for the super long delay! PTAL

comius (approved):

PR #12729, opened 2020-12-18
Bail out quickly from _find_gold_linker_path for GCC.

benjaminp:
GCC can't make use of an absolute gold path, so keep parsing specific to clang. This avoids silly problems like https://stackoverflow.com/questions/65336037/how-do-i-fix-a-bazel-build-error-in-tensorflow.


aiuto:
@oquenchil friendly ping

oquenchil (approved):
LGTM

PR #12734, opened 2020-12-18
Simplify updating our dependency on rules_java.

aiuto:
More progress on #12081. With this change, updating our version of rules_java requires only editing contiguous lines in a single file.

- Migrate dependency on rules_java to distdir_dep.bzl.
- Generate jdk.WORKSPACE from the central list of dependencies.

Status: WIP. Incremental PR on top of #12743 and #12722.  When those merge, this will reduce in size.


aiuto:
cc: @jitsu-net  This is an example of some of the test brittleness I was talking about last week.  Before this change, if you wanted to update the version of rules_cc we used, you needed to edit multiple files. With this change, you do it in a single configuration file.

PR #12739, opened 2020-12-21
update pkg_tar to 0.3.0

limdor:


limdor:
Seems that it is not so straight forward: https://github.com/bazelbuild/bazel/pull/11270#issuecomment-749030973

aiuto:
Yes. It is not straightforward. Things break because of the fragility of the integration tests depending on this package.  My plan is what I described elsewhere

- create a small, internal for bazel building and testing alternative to pkg_tar. It should not be in @baze_tools
- do not include pkg_tar in Bazel WORKSPACE

That allows rules_pkg to take on any requisite dependencies without bloating the Bazel distro

limdor:
Closing this PR, @aiuto is preparing to use 0.3.0 on https://github.com/bazelbuild/bazel/pull/11270/files. This was just a try to see if it was only the merge conflict the problem or not.

Issue #11646, opened 2020-06-26
Starlark build  settings options on the command line does not take precedence over those in rc files.


liji-canva:
### Description of the problem / feature request:

Starlark build  settings options on the command line does not take precedence over those in rc files.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Use `https://github.com/bazelbuild/examples` to reproduce this bug.

```bash
cd rules/starlark_configurations/basic_build_setting
echo 'build --//starlark_configurations/basic_build_setting:coffee-temp=HOT' > bazelrc
bazel --bazelrc=./bazelrc build :today --//starlark_configurations/basic_build_setting:coffee-temp=ICED
```

As the option on command line should take precedence over those in rc files. The expected output should be something like
`DEBUG: repopath/examples/rules/starlark_configurations/basic_build_setting/defs.bzl:36:5: ICED`, but the actual output is `DEBUG: repopath/examples/rules/starlark_configurations/basic_build_setting/defs.bzl:36:5: HOT`

### What operating system are you running Bazel on?

macOS Catalina Version 10.15.5

### What's the output of `bazel info release`?

release 3.3.0

###  Have you found anything relevant by searching the web?

No

### Any other information, logs, or outputs that you want to share?

No

katre:
@juliexxia This is pretty concerning, can you take a look?

gregestren:
I'm pretty sure this is a result of https://github.com/bazelbuild/bazel/blob/ddb9b9a60c4df7defcac7c7141dff5eaf411b1f4/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java#L209-L217

So Bazel always parses explicit command-line options first, followed by bazelrc options. According to the "last one wins" principle, you'd think bazelrc options always take precedence, even for non-Starlark flags like `--cpu`.

But the reason non-Starlark falgs work is they also apply a priority ordering:

https://github.com/bazelbuild/bazel/blob/ddb9b9a60c4df7defcac7c7141dff5eaf411b1f4/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java#L212

vs.

https://github.com/bazelbuild/bazel/blob/ddb9b9a60c4df7defcac7c7141dff5eaf411b1f4/src/main/java/com/google/devtools/build/lib/runtime/BlazeOptionHandler.java#L188

[...]

PR #12740, opened 2020-12-21
Support multiple --bazelrc on command line

### LINKED ISSUE #7489: Add support for specifying --bazelrc multiple times
Today it is not possible to add multiple --bazelrc options to the command line.

We have some default options in a common bazelrc file that is used in many workspaces since the options are needed to build for our platform.
Today we need to add those options in all user .bazelrc WORKSPACE files.
If it would be possible to add multiple --bazelrc options on the command line we could have one rc file for the common parameters and still be able to use other bazelrc files for the user projects.

linux 

0.20.0

No

wisechengyi:
Address #7489

### Motivation

Multiple --bazelrc on CLI would be useful for us for various reasons:

Mostly importantly, we will have very long argument lists to pass to bazel query, so they have to be in a bazelrc file. `import/try import` in bazelrc would still work but very awkwardly. For example, if there are multiple bazelrc files to import, say `A` and `B`,

import `A` needs to be added into `$WORKSPACE/.bazelrc`
import `B` needs to be added into `A`
meaning the former bazelrc file needs to know what comes next.

Therefore allowing multiple --bazelrc would greatly improve the ergonomics, so the caller can create and append the new bazelrc without modifying the previous rc files.

### Note

Options passed on CLI will still overwrite any options specified in any bazelrcs.

### Result
`bazel --bazelrc=x.rc --bazelrc=y.rc ...` now works.
```
  --bazelrc (a string; default: see description)
    The location of the user .bazelrc file containing default values of Bazel 
    options. This option can also be chained together.
    E.g. `--bazelrc=x.rc --bazelrc=y.rc` so options in both RCs will be read.
    Note: `--bazelrc x.rc y.rc` is illegal, and each bazelrc file needs to be 
    accompanied by --bazelrc flag before it.
    If unspecified, Bazel uses the first .bazelrc file it finds in the 
    following two locations: the workspace directory, then the user's home 
    directory. Use /dev/null to disable the search for a user rc file, e.g. in 
    release builds.
```


src/main/cpp/option_processor.cc
```diff
@@ -201,11 +204,14 @@ std::set<std::string> GetOldRcPaths(
...
-      workspace);
-  if (!user_bazelrc_path.empty()) {
-    candidate_bazelrc_paths.push_back(user_bazelrc_path);
+  vector<const char *> cmd_line_rc_files = SearchNaryOption(startup_args, "--bazelrc");
+  for (int i=0; i<cmd_line_rc_files.size(); i++) {
+    string user_bazelrc_path = internal::FindLegacyUserBazelrc(
+        cmd_line_rc_files[i],
+        workspace);
+    if (!user_bazelrc_path.empty()) {
+      candidate_bazelrc_paths.push_back(user_bazelrc_path);
```
wisechengyi, line 213: Does "internal" mean google's internal usage, @philwo? just want to make sure this isn't breaking.

src/main/cpp/option_processor.cc
```diff
@@ -370,20 +376,18 @@ blaze_exit_code::ExitCode OptionProcessor::GetRcFiles(
...
-    // exist or is unreadable, since this path is explicitly passed, this is an
-    // error. Check this condition here.
-    if (!blaze_util::CanReadFile(absolute_cmd_line_rc)) {
-      BAZEL_LOG(ERROR) << "Error: Unable to read .bazelrc file '"
-                       << absolute_cmd_line_rc << "'.";
-      return blaze_exit_code::BAD_ARGV;
-    }
-    rc_files.push_back(absolute_cmd_line_rc);
+  vector<const char *> cmd_line_rc_files = SearchNaryOption(cmd_line->startup_args, "--bazelrc");
+  for (int i=0; i< cmd_line_rc_files.size(); i++) {
```
bierbaum, line 377: Same: `int` -> `size_t`

src/main/cpp/option_processor.cc
```diff
@@ -41,6 +42,8 @@ extern char **environ;
 
 namespace blaze {
 
+using std::cout;
```
bierbaum: Delete this.

src/main/cpp/option_processor.cc
```diff
@@ -41,6 +42,8 @@ extern char **environ;
...
+using std::endl;
```
bierbaum: Delete this, too.

src/main/cpp/option_processor.cc
```diff
@@ -201,11 +204,14 @@ std::set<std::string> GetOldRcPaths(
         internal::FindRcAlongsideBinary(cwd, path_to_binary);
     candidate_bazelrc_paths = {workspace_rc, binary_rc, system_bazelrc_path};
   }
-  string user_bazelrc_path = internal::FindLegacyUserBazelrc(
-      SearchUnaryOption(startup_args, "--bazelrc", /* warn_if_dupe */ true),
-      workspace);
-  if (!user_bazelrc_path.empty()) {
-    candidate_bazelrc_paths.push_back(user_bazelrc_path);
+  vector<const char *> cmd_line_rc_files = SearchNaryOption(startup_args, "--bazelrc");
+  for (int i=0; i<cmd_line_rc_files.size(); i++) {
```
bierbaum, line 205: `size_t` should be preferred for indexing.

bierbaum (commented):
Hi Yi, I left some style nits to get this change in line with what is going on in this codebase.

You also might want to add coverage in `src/test/cpp/startup_options_test.cc`.

src/main/cpp/option_processor.cc
```diff
@@ -201,11 +201,14 @@ std::set<std::string> GetOldRcPaths(
         internal::FindRcAlongsideBinary(cwd, path_to_binary);
     candidate_bazelrc_paths = {workspace_rc, binary_rc, system_bazelrc_path};
   }
-  string user_bazelrc_path = internal::FindLegacyUserBazelrc(
-      SearchUnaryOption(startup_args, "--bazelrc", /* warn_if_dupe */ true),
-      workspace);
-  if (!user_bazelrc_path.empty()) {
-    candidate_bazelrc_paths.push_back(user_bazelrc_path);
+  vector<const char *> cmd_line_rc_files = SearchNaryOption(startup_args, "--bazelrc");
+  for (vector<const char *>::size_type i=0; i<cmd_line_rc_files.size(); i++) {
```
eed3si9n, line 205: nit: range-based for?

```suggestion
  for (auto& rc_file : cmd_line_rc_files) {
```

wisechengyi:
> You also might want to add coverage in src/test/cpp/startup_options_test.cc.

In terms of `bazelrc`s chaining together, it was already supported, but just not from command line. Hence the scope of this change merely adds the `bazelrc`s passed via CLI to the final list of `bazelrc`s that are going to be parsed. I don't see anything specific about CLI related tests in `startup_options_test.cc`. Happy to add more if @michajlo thinks it is necessary and can kindly offer some guidance.

src/main/cpp/option_processor.cc
wisechengyi: Thanks. Adjusted.

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +74,36 @@ bool GetNullaryOption(const char *arg, const char *key) {
...
+      // If 'key' was not found, 'value' is nullptr and we can return that.
+      // If 'key' was found exactly once, then 'value' has the value and again
+      // we can return that.
+      // If 'key' was found more than once then we could not have reached this
+      // line, because we would have broken out of the loop when 'key' was found
+      // the second time.
+      return values;
+    }
+    const char* result = GetUnaryOption(args[i].c_str(),
+                                        args[i + 1].c_str(),
```
noxiouz, line 96: I could be wrong but following check
```
i < args.size()
```
does not protect from out of bound access via i + 1

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +74,36 @@ bool GetNullaryOption(const char *arg, const char *key) {
   return true;
 }
 
+std::vector<const char*> SearchNaryOption(const vector<string>& args,
```
noxiouz, line 77: It looks kind of dangerous to return vector of pointers as lifetime of provided buffers are bound by lifetime of **args**.
For example, if _SearchNaryOption_ is called with temporary argument for example (like you do in the tests)
returned vector will have pointers to freed memory.

I thing returning _vector<std::string>_ is better here. From memory copy perspective it would be almost the same. As
calling _FindLegacyUserBazelrc_ does implicit conversion to string from const char* now in code

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +74,36 @@ bool GetNullaryOption(const char *arg, const char *key) {
...
+      // we can return that.
+      // If 'key' was found more than once then we could not have reached this
+      // line, because we would have broken out of the loop when 'key' was found
+      // the second time.
+      return values;
+    }
+    const char* result = GetUnaryOption(args[i].c_str(),
+                                        args[i + 1].c_str(),
+                                        key);
+    if (result != NULL) {
```
noxiouz, line 98: better check for != nullptr

src/main/cpp/blaze_util.cc
wisechengyi: I agree it's a bit sketch, but it has something to do how `GetUnaryOption` is implemented: https://github.com/bazelbuild/bazel/blob/17e8b19297905abdc0c6ba715ef472e1f54d0b2b/src/main/cpp/blaze_util.cc#L47-L60 and I was hoping to minimize the surface area touched

wisechengyi: I can, however modify it to something like:
```
GetUnaryOption(args[i].c_str(),
                                        args[min(args.size() - 1) , i + 1].c_str(),
 ```

Comment (wisechengyi) [reply src/main/cpp/blaze_util.cc]: Adjusted.

File: src/main/cpp/blaze_util.cc (hunk 9 above)
Comment (wisechengyi) [reply src/main/cpp/blaze_util.cc]: adjusted.

File: src/main/cpp/blaze_util.cc (hunk 8 above)
Comment (wisechengyi) [reply src/main/cpp/blaze_util.cc]: Thanks, @noxiouz! Adjusted to use vector instead.

**Comment by aiuto** [comment]
cc: @juliexxia 
This should have been discussed as an issue first so we could discuss the merits. It's debatable if we want to add any new complexity to arg parsing, since a few of us have recently noticed that it is a mess in many ways and is ripe for a complete rebuilding.

**Comment by wisechengyi** [comment]
Hi @aiuto , the prior discussion is available at #7489

**Comment by wisechengyi** [comment]
Regarding arg parsing here, I was following the previous pattern to minimize the scope of the change and reduce the chances of breaking anything else. AFACIT these parsing functions are only used for `--bazelrc` even though they look more generic. I agree it's not in a great place, but a separate change to refactor them would be more fitting.

**Comment by philwo** [comment]
Thank you @wisechengyi! I'm just catching up with my mails after my vacation, I'll take a look at this in the next days :)

File: src/main/cpp/blaze_util.cc [hunk 10]
```diff
@@ -74,6 +76,32 @@ bool GetNullaryOption(const char *arg, const char *key) {
...
+                              const char *key) {
+  vector<std::string> values;
+  for (vector<string>::size_type i = 0; i < args.size(); ++i) {
+    if (args[i] == "--") {
+      // If the current argument is "--", all following args are target names.
+      // If 'key' was not found, 'value' is nullptr and we can return that.
+      // If 'key' was found exactly once, then 'value' has the value and again
+      // we can return that.
+      // If 'key' was found more than once then we could not have reached this
+      // line, because we would have broken out of the loop when 'key' was found
```
janakdr, line 89: where's the break?

src/main/cpp/option_processor.cc
```diff
@@ -201,11 +201,22 @@ std::set<std::string> GetOldRcPaths(
         internal::FindRcAlongsideBinary(cwd, path_to_binary);
     candidate_bazelrc_paths = {workspace_rc, binary_rc, system_bazelrc_path};
   }
-  string user_bazelrc_path = internal::FindLegacyUserBazelrc(
-      SearchUnaryOption(startup_args, "--bazelrc", /* warn_if_dupe */ true),
-      workspace);
-  if (!user_bazelrc_path.empty()) {
-    candidate_bazelrc_paths.push_back(user_bazelrc_path);
+  vector<std::string> cmd_line_rc_files = SearchNaryOption(startup_args, "--bazelrc");
+  // Divide the cases where the vector is empty vs not, as `FindLegacyUserBazelrc` has a case
```
janakdr, line 205: nit: is this line too long? Please reflow if so

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,32 @@ bool GetNullaryOption(const char *arg, const char *key) {
   return true;
 }
 
+std::vector<std::string> SearchNaryOption(const vector<string>& args,
+                              const char *key) {
+  vector<std::string> values;
+  for (vector<string>::size_type i = 0; i < args.size(); ++i) {
+    if (args[i] == "--") {
+      // If the current argument is "--", all following args are target names.
+      // If 'key' was not found, 'value' is nullptr and we can return that.
```
janakdr, line 85: nullptr or an empty vector?

janakdr (commented):
I'm not sure what it means to be assigned versus reviewer? I guess I'll review?

src/main/cpp/blaze_util.cc
wisechengyi: the `return` line below would be equivalent.

src/main/cpp/blaze_util.cc
wisechengyi: Thanks! Good catch.

src/main/cpp/option_processor.cc
wisechengyi: I wasn't able to find any auto formatter provided in the repo, so manually adjusted it. You have any pointers, that'd be great for future reference.

src/main/cpp/blaze_util.cc
wisechengyi: removed the comment here, only leaving
```
// If the current argument is "--", all following args are target names.
```
hope the rest should be self explanatory.

src/main/cpp/blaze_util.cc
wisechengyi: Corrected.

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,26 @@ bool GetNullaryOption(const char *arg, const char *key) {
   return true;
 }
 
+std::vector<std::string> SearchNaryOption(const vector<string>& args,
+                              const char *key) {
+  vector<std::string> values;
+  for (vector<string>::size_type i = 0; i < args.size(); ++i) {
+    if (args[i] == "--") {
```
janakdr, line 83: I'm not sure this is correct. Have you tried running commands like `bazel --host_jvm_args -- --host_jvm_args=-Xmx5g build --bes_keywords -- --keep_going :BUILD`? That works for me, so I think that `--` does *not* unconditionally end flag parsing.

janakdr (commented):
Could you please add an integration test? Check that passing the flag twice brings in both values of a list-valued flag, for instance, and in the proper order. Also test behavior when passing a nonexistent value, whether before or after an existing file. Thanks!

src/main/cpp/blaze_util.cc
wisechengyi: Thanks! will check. I'd guess it should still work because it has worked in the past for a single `--bazelrc`.

wisechengyi: Hi @janakdr, do you mean
```
bazel --host_jvm_args -- --host_jvm_args=-Xmx5g build --bes_keywords -- --keep_going :BUILD
```
or 
```
bazel --host_jvm_args=-Xmx5g build --bes_keywords -- --keep_going :BUILD
```
but in either case, the behavior is consistent before and after the change.

wisechengyi: With this change:
```
$ bazel-bin/src/bazel --host_jvm_args -- --host_jvm_args=-Xmx5g build --bes_keywords -- --keep_going :BUILD
WARNING: Running Bazel server needs to be killed, because the startup options are different.
Server crashed during startup. Now printing /private/var/tmp/_bazel_yic/5a517de0ca1edbc0f94cf017f53ee5a1/server/jvm.out
Unrecognized option: --
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
$ echo $?
37
```
Without this change:
```
$ bazel --host_jvm_args -- --host_jvm_args=-Xmx5g build --bes_keywords -- --keep_going :BUILD
Server crashed during startup. Now printing /private/var/tmp/_bazel_yic/5a517de0ca1edbc0f94cf017f53ee5a1/server/jvm.out
Unrecognized option: --
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
$ echo $?
37
```

wisechengyi:
Tests added. Please kindly let me know in case location needs adjustment.

src/main/cpp/blaze_util.cc
janakdr: My point is that `--` does not necessarily indicate the end of options parsing, whereas your code short-circuits if it ever encounters a `--`. It's not that your code changes current behavior. But it doesn't do what the user would expect in this case, although admittedly it's not a very common one.

wisechengyi: Thanks for clarifying, @janakdr. Yeah since I wasn't aware of all the possible scenarios here, I decided to mimic the existing logic in `SearchUnaryOption` in order to minimize the risk, so essentially there is no change how to treat `--`.

For the case described here, do you think it would be better to be addressed in an separate issue? I'd also like to understand what the expected outcome would be.

ulrfa:
At Ericsson, we have a need for syntax:

`--bazelrc file1 --bazelrc file2`

but no need for syntax:

`--bazelrc file1 file2`

Because we have a bazel wrapper script that is adding --bazelrc command line option, and we would like to do that without parsing the command line for already existing --bazelrc flags.

And accumulating values from multiple flags seems common. Bazel’s command line reference is frequently using the phrase _“multiple uses are accumulated”_.

src/main/cpp/blaze_util.cc
janakdr: Yes, it looks like the C++ code is fundamentally different in how it parses than the Java code is, which is unfortunate. You're correct that this isn't a new bug in this code. What I would expect is the Java behavior, which is able to distinguish between `--` as the argument to a flag or as the signal to end options parsing.

janakdr (approved):
> At Ericsson, we have a need for syntax:
> 
> `--bazelrc file1 --bazelrc file2`
> 
> but no need for syntax:
> 
> `--bazelrc file1 file2`

@wisechengyi would you mind documenting explicitly somewhere that --bazelrc file1 file2 is *not* supported? That behavior is often supported for multi-valued flags in the Java processor, so we should make sure nobody gets confused. Thanks!

wisechengyi:
> @wisechengyi would you mind documenting explicitly somewhere that --bazelrc file1 file2 is not supported? That behavior is often supported for multi-valued flags in the Java processor, so we should make sure nobody gets confused. Thanks!

Done. Thanks!
```
$ bazel-bin/src/bazel help startup_options --long | grep '\--bazelrc' -A 5
  --bazelrc (a string; default: see description)
    The location of the user .bazelrc file containing default values of Bazel 
    options. This option can also be chained together.
    E.g. `--bazelrc=x.rc --bazelrc=y.rc` so options in both RCs will be read.
    Note: `--bazelrc x.rc y.rc` is illegal, and each bazelrc file needs to be 
    accompanied by --bazelrc flag before it.
    If unspecified, Bazel uses the first .bazelrc file it finds in the 
    following two locations: the workspace directory, then the user's home 
    directory. Use /dev/null to disable the search for a user rc file, e.g. in 
    release builds.
      Tags: changes_inputs
```

michajlo:
Sorry, this must have gotten buried in the pre-holidays email pile. Thanks for picking it up Janak. LMK if you want me to add some comments.

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,26 @@ bool GetNullaryOption(const char *arg, const char *key) {
   return true;
 }
 
+std::vector<std::string> SearchNaryOption(const vector<string>& args,
+                              const char *key) {
+  vector<std::string> values;
+  for (vector<string>::size_type i = 0; i < args.size(); ++i) {
```
michajlo, line 82: I think you need to advance `i` extra if the flag value came from a `--bazelrc foo`. Adding a test for `--bazelrc --bazelrc foo` would help confirm/deny.

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,26 @@ bool GetNullaryOption(const char *arg, const char *key) {
   return true;
 }
 
+std::vector<std::string> SearchNaryOption(const vector<string>& args,
```
michajlo, line 79: I _think_ this'd be more clear if we named it `SearchUnaryOptions` or `GetAllUnaryOptionValues`. "Nary" implies the flag has multiple arguments, but what we really want to convey is that we want to find all values of the flag.

src/main/cpp/option_processor.cc
```diff
@@ -201,11 +201,22 @@ std::set<std::string> GetOldRcPaths(
...
+  vector<std::string> cmd_line_rc_files = SearchNaryOption(startup_args, "--bazelrc");
+  // Divide the cases where the vector is empty vs not, as `FindLegacyUserBazelrc`
+  // has a case for rc_file to be a nullptr.
+  if (cmd_line_rc_files.empty()){
+    string user_bazelrc_path = internal::FindLegacyUserBazelrc(nullptr, workspace);
+    if (!user_bazelrc_path.empty()) {
+      candidate_bazelrc_paths.push_back(user_bazelrc_path);
+    }
+  }
+  else {
```
michajlo, line 213: put else on same line as previous `}`

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
```diff
@@ -33,7 +33,11 @@
         valueHelp = "<path>",
         help =
             "The location of the user .bazelrc file containing default values of "
-                + "Bazel options. If unspecified, Bazel uses the first .bazelrc file it finds in "
+                + "Bazel options. This option can also be chained together.\n"
```
michajlo, line 36: I think you mean "can be specified multiple times"? "Chained together" might be the source of confusion in https://github.com/bazelbuild/bazel/pull/12740#issuecomment-762724352

src/main/cpp/blaze_util.cc
wisechengyi: Do you think this should be covered by this?
```
TEST_F(BlazeUtilTest, TestSearchNaryStartupOptionWithoutEquals2) {
  assert_equal_vector_char_pointer({"value1", "value2"}, SearchNaryOption({"bazel", "--flag", "value1", "--flag", "value2", "build", ":target"}, "--flag"));
}
```

or are you specifically mean `--bazelrc --bazelrc foo`? The current behavior is
```
$ bazel-bin/src/bazel --bazelrc --bazelrc 1.rc info
[ERROR 15:32:34.993 src/main/cpp/option_processor.cc:391] Error: Unable to read .bazelrc file '/Users/yic/workspace/bazel/--bazelrc'.
[FATAL 15:32:34.994 src/main/cpp/blaze.cc:1295]
```
I think this should be expected since every `--bazelrc` should expects a value?

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,26 @@ bool GetNullaryOption(const char *arg, const char *key) {
...
+                              const char *key) {
+  vector<std::string> values;
+  for (vector<string>::size_type i = 0; i < args.size(); ++i) {
+    if (args[i] == "--") {
+      // If the current argument is "--", all following args are target names.
+      return values;
+    }
+    const char* result = GetUnaryOption(args[i].c_str(),
+                                        args[std::min(i + 1, args.size() -1)].c_str(),
```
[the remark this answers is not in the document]
wisechengyi: Do you mean `char*` -> `char *`? `char*` seems to be the pattern in this file.

I wasn't able to find any auto formatter provided in the repo. You have any pointers, that'd be great for future reference.

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
wisechengyi: Thanks. Will clarify.

src/main/cpp/blaze_util.cc
wisechengyi: Sounds good.

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
wisechengyi: adjusted.

src/main/cpp/option_processor.cc
wisechengyi: adjusted.

src/main/cpp/blaze_util.cc
wisechengyi: Renamed.

src/main/cpp/blaze_util.cc
wisechengyi: Good point. Corrected.

aiuto (changes requested):
What I don't see is a definition of what happens if you do
`bazel --some_option --bazelrc foo build`

Do command line options override all bazelrc or will order matter, such that the second bazelrc can reset what what seen on the command line?
And we need a test for whatever behavior is chosen

aiuto:
Also see #11646

janakdr:
Tony, is your question new for this change? Seems like it's independent of the multiple bazelrc question. My impression is that explicit arguments on the command line always override what's set in the bazelrc, even if the bazelrc comes later, so there's no new problem here.

src/main/cpp/blaze_util.cc
michajlo: The latter case (`--bazelrc --bazelrc foo`). I think this currently returns `[--bazelrc, foo]`, when it should really return `[--bazelrc]`.

michajlo (changes requested):

src/main/cpp/blaze_util.cc
wisechengyi: With this change, the error 
```
$ bazel --bazelrc --bazelrc foo info
[ERROR 09:27:17.974 src/main/cpp/option_processor.cc:382] Error: Unable to read .bazelrc file '/Users/yic/workspace/bazel/--bazelrc'.
[FATAL 09:27:17.975 src/main/cpp/blaze.cc:1293] 
$ bazel --bazelrc --bazelrc foo 
[ERROR 09:27:20.867 src/main/cpp/option_processor.cc:382] Error: Unable to read .bazelrc file '/Users/yic/workspace/bazel/--bazelrc'.
[FATAL 09:27:20.867 src/main/cpp/blaze.cc:1293]
```
means it thinks `--bazelrc` is the value for key `--bazelrc`, and `foo` isn't in the picture, which is consistent with the behavior without the change:
```
$ bazel --bazelrc --bazelrc 1.rc info
[ERROR 09:26:19.832 src/main/cpp/option_processor.cc:382] Error: Unable to read .bazelrc file '/Users/yic/workspace/bazel/--bazelrc'.
[FATAL 09:26:19.833 src/main/cpp/blaze.cc:1293]
```

I will add a test case for this if I understand correctly here.

wisechengyi: oh ok. I see. Will adjust.

wisechengyi: Tests added. Thanks! That's a good catch!

wisechengyi:
@aiuto:
> What I don't see is a definition of what happens if you do
> `bazel --some_option --bazelrc foo build`
> 
> Do command line options override all bazelrc or will order matter, such that the second bazelrc can reset what what seen on the command line?
> And we need a test for whatever behavior is chosen

@michajlo:
> Tony, is your question new for this change? Seems like it's independent of the multiple bazelrc question. My impression is that explicit arguments on the command line always override what's set in the bazelrc, even if the bazelrc comes later, so there's no new problem here.

Correct. This change only adds more bazelrc to be processed, so it does not affect the existing priority between CLI options vs the ones in bazelrc.

michajlo (changes requested):

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
```diff
@@ -33,7 +33,11 @@
         valueHelp = "<path>",
         help =
             "The location of the user .bazelrc file containing default values of "
-                + "Bazel options. If unspecified, Bazel uses the first .bazelrc file it finds in "
+                + "Bazel options. This option can also be specified multiple times.\n"
+                + "E.g. `--bazelrc=x.rc --bazelrc=y.rc` so options in both RCs will be read.\n"
+                + "Note: `--bazelrc x.rc y.rc` is illegal, and each bazelrc file needs to be "
```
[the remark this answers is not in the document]
wisechengyi: Would you agree that there is still value for this change because this is the only place that documents `--bazelrc` can used multiple times?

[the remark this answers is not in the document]
wisechengyi: Happy to revise if the suggested way is still preferred. I was mostly thinking from other users' point of view, since this is a rather unique use of the options within bazel.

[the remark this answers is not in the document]
michajlo: I don't think any other flag support `--flag val1 val2` == `--flag=val1 --flag=val2`, which is why I don't think we should suggest such a thing could even be considered possible.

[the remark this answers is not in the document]
wisechengyi: Oh I see. Might be due to github's UI. IIUC I will remove 
```
                + "Note: `--bazelrc x.rc y.rc` is illegal, and each bazelrc file needs to be "
```
and leave the rest.

wisechengyi:
Hi @michajlo @aiuto, thanks again for the review! If all looks good, we can close the tap on this.

michajlo (approved):

src/test/shell/integration/startup_options_test.sh
```diff
@@ -82,4 +82,23 @@ function test_autodetect_server_javabase() {
   bazel --noautodetect_server_javabase version &> $TEST_log || fail "Should pass"
 }
 
+# Below are the regression tests for Issue #7489
+function test_multiple_bazelrc_later_overwrites_earlier() {
+  tmpdir1=$(cd $(mktemp -d) &> /dev/null && pwd -P)
+  tmpdir2=$(cd $(mktemp -d) &> /dev/null && pwd -P)
+  echo "common --repository_cache=$tmpdir1" > 1.rc
```
michajlo, line 89: Can you use a different flag? Running into problems with import where our internal version doesn't recognize `--respository_cache`. Same for other uses of `--repository_cache`. LMK if you want some suggestions (I'd have to look around the code some).

src/test/shell/integration/startup_options_test.sh
```diff
@@ -82,4 +82,23 @@ function test_autodetect_server_javabase() {
...
+  echo "common --repository_cache=$tmpdir2" > 2.rc
+  bazel --bazelrc=1.rc --bazelrc=2.rc info &> $TEST_log || fail "Should pass"
```
michajlo, line 91: Having some trouble with flag naming and internal test compatibility, using `"--${PRODUCT_NAME}rc=1.rc"` should fix it.

michajlo (changes requested):
Ran into some issues on import. Should be easy to resolve.

src/test/shell/integration/startup_options_test.sh
wisechengyi: `--repository_cache` was one of the few, if not the only `bazel info` key that will be read from bazelrc file, so had to be creative to test with `--color=yes/no` instead.

src/test/shell/integration/startup_options_test.sh
wisechengyi: Adjusted. Thanks for the pointer!

wisechengyi:
Thanks! have adjusted accordingly.

aiuto:
> Hi @michajlo @aiuto, thanks again for the review! If all looks good, it would be great to close the tap on this.

I think you should add a tiny bit to the docstring in the option definition that clarifies that command line always wins over blazerc, even if they appear in different orders.  If you don't, I would be that within a year, someone is will file a bug about order not working the way they expected, and we will have to fix the docs then.

wisechengyi:
HI @aiuto, just to clarify, do you mean
1. the doc for bazel? If so, I'm happy to do so, but had some issue in the past that github PR may not be able to do that: https://github.com/bazelbuild/bazel/pull/12194#issuecomment-702497443
2. adding to the description here:
```
                + "Bazel options. This option can also be specified multiple times.\n"
                + "E.g. `--bazelrc=x.rc --bazelrc=y.rc` so options in both RCs will be read.\n"
                + "If unspecified, Bazel uses the first .bazelrc file it finds in "
                + "the following two locations: the workspace directory, then the user's home "	                + "the following two locations: the workspace directory, then the user's home "
                + "directory. Use /dev/null to disable the search for a user rc file, e.g. in "	                + "directory. Use /dev/null to disable the search for a user rc file, e.g. in "
                + "release builds.")	                + "release builds.")
```

wisechengyi:
Also happy to follow up with a change suggested, so we can get this in first.

aiuto:
I mean the embedded doc in /BazelStartupOptionsModule.java

wisechengyi:
Thanks! Adjusted, so this is the help msg now:
```
  --bazelrc (a string; default: see description)
    The location of the user .bazelrc file containing default values of Bazel 
    options. This option can also be specified multiple times.
    E.g. `--bazelrc=x.rc --bazelrc=y.rc` so options in both RCs will be read.
    If unspecified, Bazel uses the first .bazelrc file it finds in the 
    following two locations: the workspace directory, then the user's home 
    directory. Use /dev/null to disable the search for a user rc file, e.g. in 
    release builds.
    Note: command line options will always supersede any option in bazelrc.
      Tags: changes_inputs
```

wisechengyi:
This should be ready now. Thank you!

aiuto (approved):

src/test/shell/integration/startup_options_test.sh
```diff
@@ -82,4 +82,30 @@ function test_autodetect_server_javabase() {
   bazel --noautodetect_server_javabase version &> $TEST_log || fail "Should pass"
 }
 
+# Below are the regression tests for Issue #7489
+function test_multiple_bazelrc_later_overwrites_earlier() {
+  plain_expected_output="INFO: Build completed successfully, 1 total action"
+
+  echo "common --color=yes" > 1.rc
+  echo "common --color=no" > 2.rc
+  bazel "--${PRODUCT_NAME}rc=1.rc" "--${PRODUCT_NAME}rc=2.rc" build &> $TEST_log || fail "Should pass"
```
michajlo, line 91: build is a heavyweight command and thus doesn't always work in CI, can you use `info`, `version`, or a similar "simpler" command

wisechengyi: Makes sense. Done.

wisechengyi:
For some reason "Invocation Id" doesn't always appear, so adjusted the test to use the simpler "help" command and check for help messages long vs short.

michajlo:
Thanks for the changes. Looks like I'm going to have to make some modifications to our internal code to get this to land cleanly, hence the extra delays.

wisechengyi:
Thank you, @michajlo! Appreciate it

wisechengyi:
Hello 👋  just checking to see if there's an ETA to land? Would be great for us to close the tap and not having to build a custom bazel with this change internally.

michajlo:
Sorry, haven't had a chance to work out the internal changes yet.

michajlo:
Good news is internal changes don't look so bad. Bad news is we overlooked the `Use /dev/null to disable the search for a user rc file, e.g. in release builds` bit of documentation. I think it's important to find a way to maintain this behavior, even if we haven't been great about having tests for it. It's been around for a while so there's inevitably someone relying on it. Perhaps ignoring all rc-files after the first `/dev/null` would work?

wisechengyi:
Thanks for checking @michajlo! Just to make sure I understand correctly:
Given
```
--bazelrc=x --bazelrc=y --bazelrc=/dev/null --bazelrc=z
```
x and y still need to be picked up, and z will be dropped, correct?

michajlo:
Correct. That seems like the best way to preserve the `/dev/null` behavior.

IIUC the `/dev/null`-disables-search behavior isn't explicitly implemented but works out in practice because we only respect the first `--bazelrc`. Now that we're tolerating multiple that convention is broken. Ignoring everything after the first `/dev/null` seems like the cleanest way to maintain backwards compatibility.

wisechengyi:
Hi @michajlo I have made the changes as discussed. Please kindly give it another go when you get a chance. Thanks!

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
```diff
@@ -33,10 +33,17 @@
         valueHelp = "<path>",
         help =
             "The location of the user .bazelrc file containing default values of "
-                + "Bazel options. If unspecified, Bazel uses the first .bazelrc file it finds in "
+                + "Bazel options. This option can also be specified multiple times.\n"
+                + "E.g. with `--bazelrc=x.rc --bazelrc=y.rc`, options in both RCs will be read.\n"
+                + "--bazelrc values that come after `/dev/null` will be ignored to maintain "
```
michajlo, line 38: Rephrase to avoid the "backward compatibility" bit - in a few weeks/months/years it won't make sense anymore. Also, can you combine the info on the special value of `/dev/null` with the explanation below? Something like...

`/dev/null` indicates that all further `--bazelrc`s should be ignored. This is useful for disabling the search for a ...`

src/test/shell/integration/startup_options_test.sh
```diff
@@ -82,4 +82,28 @@ function test_autodetect_server_javabase() {
...
+}
+
+function test_multiple_bazelrc_set_different_options() {
+  echo "common --verbose_failures" > 1.rc
+  echo "common --test_output=all" > 2.rc
+  bazel "--${PRODUCT_NAME}rc=1.rc" "--${PRODUCT_NAME}rc=2.rc" build --announce_rc &> $TEST_log || fail "Should pass"
+  expect_log "Inherited 'common' options: --verbose_failures"
+  expect_log "Inherited 'common' options: --test_output=all"
+}
+
```
michajlo, line 108: Add a test for the `/dev/null` case we added? This _should_ pass both before and after your changes, proving the compatibility.

src/main/cpp/blaze_util.cc
```diff
@@ -74,6 +76,49 @@ bool GetNullaryOption(const char *arg, const char *key) {
...
+    if (result == next_arg) {
+      i++;
+    }
+  }
+
+  if (ignore_after_value == nullptr) {
+    return values;
+  }
+  else {
+    vector<std::string> new_values;
```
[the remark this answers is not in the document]
wisechengyi: good point. will revise.

src/test/shell/integration/startup_options_test.sh
wisechengyi: will do

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
wisechengyi: Sounds good.

src/main/cpp/blaze_util.cc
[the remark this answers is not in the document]
wisechengyi: revised.

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
wisechengyi: revised.

src/test/shell/integration/startup_options_test.sh
wisechengyi: test added

wisechengyi:
Comment addressed. Thanks!

src/test/cpp/blaze_util_test.cc
```diff
@@ -234,4 +234,88 @@ TEST_F(BlazeUtilTest, TestSearchUnaryCommandOptionWarnsAboutDuplicates) {
...
+
+TEST_F(BlazeUtilTest, TestSearchAllUnaryStartupOptionWithRepeatingFlagOptionsOnly) {
+  assert_equal_vector_char_pointer({"--flag"}, GetAllUnaryOptionValues({"bazel", "--flag", "--flag", "value1"}, "--flag"));
+}
+
+TEST_F(BlazeUtilTest, TestSearchAllUnaryStartupOptionWithRepeatingFlagOptionsOnlyMulti) {
+  assert_equal_vector_char_pointer({"--flag", "value1"}, GetAllUnaryOptionValues({"bazel", "--flag", "--flag", "--flag", "value1"}, "--flag"));
+}
+
+TEST_F(BlazeUtilTest, TestSearchAllUnaryStartupOptionWithRepeatingFlagOptionsOnlyMultiWithEquals) {
```
michajlo, line 273: Lines should be <=80 cols. The import process auto-formats to get most lines there, but it's having a hard time with this one because of the length of the test name. Can you shorten it a bit? Just one or two cols over right now.

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
```diff
@@ -33,10 +33,17 @@
         valueHelp = "<path>",
         help =
             "The location of the user .bazelrc file containing default values of "
-                + "Bazel options. If unspecified, Bazel uses the first .bazelrc file it finds in "
+                + "Bazel options. This option can also be specified multiple times.\n"
+                + "E.g. with "
+                + "`--bazelrc=x.rc --bazelrc=y.rc --bazelrc=/dev/null --bazelrc=z.rc`,\n"
+                + "  1) x.rc and y.rc will be read.\n"
+                + "  2) /dev/null indicates that all further `--bazelrc`s will be ignored, "
```
michajlo, line 40: This should be explained up front, explaining it inside the example makes it hard to find...

"This option can be specified multiple times. `--bazelrc=/dev/null` indicates that all further `--bazelrc` should be ignored. This is useful for disabling the search for user rc files, such as in release builds. For example: `--bazelrc=...` will use ... and ignore ...`.

src/test/cpp/blaze_util_test.cc
wisechengyi: Manually corrected.

> The import process auto-formats

What do you mean by the import process? Is there an auto formatter I can run in this repo?

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
wisechengyi: Makes sense. Corrected.

src/test/cpp/blaze_util_test.cc
michajlo: The process of merging the change. I'm not sure on the formatter question.

src/main/java/com/google/devtools/build/lib/bazel/BazelStartupOptionsModule.java
```diff
@@ -33,10 +33,19 @@
         valueHelp = "<path>",
         help =
             "The location of the user .bazelrc file containing default values of "
-                + "Bazel options. If unspecified, Bazel uses the first .bazelrc file it finds in "
+                + "Bazel options. "
+                + "/dev/null indicates that all further `--bazelrc`s will be ignored, "
+                + "This is useful to disable the search for a user rc file, "
```
michajlo, line 38: "This" should be lower case since it follows a comma.

wisechengyi: replaced with "which" otherwise it has to be ", and this".

src/test/cpp/blaze_util_test.cc
wisechengyi: Thanks!

src/test/shell/integration/startup_options_test.sh
```diff
@@ -82,4 +82,40 @@ function test_autodetect_server_javabase() {
...
+  echo "help --help_verbosity=medium" > 1.rc
+  echo "help --help_verbosity=short" > 2.rc
+  bazel "--${PRODUCT_NAME}rc=1.rc" "--${PRODUCT_NAME}rc=2.rc" help startup_options &> $TEST_log || fail "Should pass"
+  expect_not_log "$help_message_in_description"
+}
+
+function test_multiple_bazelrc_set_different_options() {
+  echo "common --verbose_failures" > 1.rc
+  echo "common --test_output=all" > 2.rc
+  bazel "--${PRODUCT_NAME}rc=1.rc" "--${PRODUCT_NAME}rc=2.rc" build --announce_rc &> $TEST_log || fail "Should pass"
```
michajlo, line 104: Echoing earlier comments, build is a heavyweight command and thus doesn't always work in CI, can you use info, version, or a similar "simpler" command? I think I have the requisite internal changes done, but the extra overhead of build is causing problems. Same in the next test.

wisechengyi: Sorry about that. Corrected. Also appreciate the work on internal validation!

michajlo:
I had overlooked that there's some documentation that needs to be updated. Can you update https://github.com/bazelbuild/bazel/blob/master/site/docs/guide.md and https://github.com/bazelbuild/bazel/blob/master/site/docs/user-manual.html, and do a quick skim to see if there's anything else we missed?

wisechengyi:
Hi @michajlo, happy to do that. However logistically are we going to run into issues like https://github.com/bazelbuild/bazel/pull/12194 where we cannot directly update doc from github PR? If so, would you like me to send a separate PR, and then you can modify the internal repo accordingly?

michajlo:
I _think_ this instance is fine to change.

site/docs/guide.md
```diff
@@ -776,8 +776,17 @@ before the command (`build`, `test`, etc).
 4.  **The user-specified RC file**, if specified with
     <code>--bazelrc=<var>file</var></code>
 
-    This flag is optional. However, if the flag is specified, then the file must
-    exist.
```
wisechengyi, line 780: removed `if the flag is specified, then the file must exist` portion, because bazel would error clearly on non-existent bazelrc file.

wisechengyi (commented):
Doc updated. https://github.com/bazelbuild/bazel/blob/master/site/docs/user-manual.html should be still accurate on `--bazelrc`

wisechengyi:
Thank you @michajlo !

PR #12743, opened 2020-12-22
 Generate rules/cpp/cc_configure.WORKSPACE from distdir_deps.bzl rather than keeping it in sync by hand

aiuto:
Generate rules/cpp/cc_configure.WORKSPACE from distdir_deps.bzl rather than keeping it in sync by hand.

More progress on #12081.

Create a capability to generate WORKSPACE files for tests at build time, based on the repository definitions in distdir_deps.bzl.

This makes it possible for the versions of repositories used in tests to align with the versions used a Bazel build time without upddating files scattered throughout the source and test trees.

Demonstrated with rules_cc.


distdir_deps.bzl
```diff
@@ -35,3 +35,66 @@ DIST_DEPS = {
...
+    return [DefaultInfo(files = depset([ctx.outputs.out]))]
+
+gen_workspace_stanza = rule(
+    implementation = _gen_workspace_stanza_impl,
+    attrs = {
+        "repos": attr.string_list(doc = "Set of repos to inlcude"),
+        "out": attr.output(mandatory = True),
+        "preamble": attr.string(doc = "Preamble."),
+        "postamble": attr.string(doc = "setup rules to follow repos."),
+        "template": attr.label(
```
aiuto, line 93: Note to reviewers.  Template will be used in PR #12734.  Got it working there, but decided to split this PR down do only doing C++ to keep the review surface small.

distdir_deps.bzl
```diff
@@ -35,3 +35,66 @@ DIST_DEPS = {
...
+    http_archive,
+    "{repo}",
+    sha256 = "{sha256}",
+    strip_prefix = {strip_prefix},
+    urls = {urls},
+)
+"""
+    repo_stanzas = {}
+    for repo in ctx.attr.repos:
+        info = DIST_DEPS[repo]
```
meteorcloudy, line 54: Throw an error if repo is not in DIST_DEPS?

meteorcloudy (approved):

distdir_deps.bzl
aiuto: It will throw an an error with a call stack. That is the most useful thing because it pinpoints where I incorrectly used the name.

aiuto:
This PR is unexpectedly difficult to import.  It passes all CI from this side, but when imported to Google and reexported it fails to build a correct bazel. That is, the bazel runs, but is missing the cc_configure.WORKSPACE resource. But the strange part is that the generated file *is* in bazel-distfile.zip and in the deploy jar. And that makes sense, because we should not even be able to build bazel (with bazel) without having that file correctly generated.  So my suspicion is that we are building a bootstrap bazel directly from java sources, without bazel - and we are testing with that version, and that process is not actually using distfile.zip, but is instead relying on the github sources.    I have more to explore.

Issue #12565, opened 2020-11-26
Print a warning when using an oldName of a flag


philwo:
### Description of the problem / feature request:

In https://github.com/bazelbuild/bazel/issues/11421 a user was confused by the fact that `--experimental_enable_runfiles` still exists (undocumented) when `--enable_runfiles` also exists. They thought that these are two different flags, the experimental one maybe being a newer or different implementation.

In fact, they're the same flag, but `--experimental_enable_runfiles` was the `oldName` of the flag: https://cs.opensource.google/bazel/bazel/+/8eae6b34ec6e1e83010b17094d1225f045606eb2:src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java;l=722

### Feature requests: what underlying problem are you trying to solve with this feature?

We should print a warning if a flag is passed by its `oldName` to Bazel that tells the user that the name is deprecated and might go away in the future. The warning should also contain the new name of the flag.

philwo:
In fact, the JavaDoc says that this should already work: "If the old name is used, a warning will be printed indicating that the old name is deprecated and the new name should be used." 😏 

But it doesn't:

```
$ bazel shutdown; bazel build --experimental_enable_runfiles=yes //src:bazel
Starting local Bazel server and connecting to it...
DEBUG: /private/var/tmp/_bazel_philwo/7a01905b4627ca044e5e3f5ad5b14d26/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14: 
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: /private/var/tmp/_bazel_philwo/7a01905b4627ca044e5e3f5ad5b14d26/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1804_java11 not using checked in configs as detect_java_home was set to True 
DEBUG: /private/var/tmp/_bazel_philwo/7a01905b4627ca044e5e3f5ad5b14d26/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14: 
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
[...]

PR #12746, opened 2020-12-22
Report a warning when using an oldName of a flag

mai93:
This PR modifies the option parser to print a warning when the old name of an option is used. The warning indicates that the old name is deprecated and the option new name should be used.

Fixes: #12565

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -723,8 +723,7 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
       documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
       effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
       help =
-          "Enable runfiles symlink tree; By default, it's off on Windows, on on other platforms.",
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -723,7 +723,8 @@ public OutputPathsConverter() {
       documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
       effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
       help =
-          "Enable runfiles symlink tree; By default, it's off on Windows, on on other platforms.")
+          "Enable runfiles symlink tree; By default, it's off on Windows, on on other platforms.",
+      oldName = "experimental_enable_runfiles")
```
katre, line 727: This should be a separate PR, so that the history is easier to read.

katre (approved):
This change looks good. I'm a bit worried about adding more warning messages: we already have a lot and users frequently just tune it all out. That said, this is something I had already thought we were warning on, so I think it's fine.

Also, adding a flag to turn off flag warnings seems too recursive.

Please do split out the CoreOptions change, however. I'm happy to approve it as well.

katre (approved):

Issue #3778, opened 2017-09-20
aar_import should propagate proguard specs


aj-michael:
AAR files may contain a file called `proguard.txt` at the top level. These should be propagated up to the android_binary level and concatenated if a proguarded build is requested. This is especially important for the Android Support Libraries.

cpsauer:
Should this have been closed by #12749? 

(Happened across these while searching after running into the bug--fix still unreleased, I think, right?)

cpsauer:
^ tagging also @benjaminRomano and @philwo

benjaminRomano:
Yes, this should be resolved by https://github.com/bazelbuild/bazel/pull/12749. @cpsauer maybe try asking in https://github.com/bazelbuild/bazel/issues/13558 to get this cherrypicked in. My change is self-contained so it should be straightforward.

cpsauer:
Done. Thanks again, guys.

I don't have the power to close the issue either--could I ask someone who can to do so? @philwo?

PR #12749, opened 2020-12-23
Export proguard specs from aar_import

benjaminRomano:
**Background**
https://github.com/bazelbuild/bazel/issues/3778

proguard specs from the `aar_import` rule do not get bubbled up to `android_binary`. In this PR, I wire up a `ProguardSpecProvider` from this rule that exports the `proguard.txt` within an AAR if it exists and any transitive proguard specs from the `exports` attribute.

**Changes**
* Add an `aar_embedded_proguard_extractor` script to extract `proguard.txt` from an AAR if it exists otherwise generate an empty proguard specs file
* In AarImport, wire up the proguard extractor action and export results through a `ProguardSpecProvider`.

Once this lands, the android rules would need to be bumped.

**Test Plan**
* Added tests for the extraction python script
* Added tests for the `aar_import` rule changes

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
@@ -14,6 +14,7 @@
 package com.google.devtools.build.lib.rules.android;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.devtools.build.lib.actions.Action;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.actions.Artifact.SpecialArtifact;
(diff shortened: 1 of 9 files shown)


benjaminRomano:
I think the test failures are related to android_tools version need to be bumped?

nkoroste:
Any updates on this?

ahumesky (approved):
Thank you, sorry for the delay here. Looks good, I'll start importing this change today

ahumesky:
Also, could you clarify by what "the android rules would need to be bumped" means? If you're referring to the remote android tools (the dependencies automatically included by [android_remote_tools.WORKSPACE](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_remote_tools.WORKSPACE), right now those don't contain these python tools. These python tools are still embedded within bazel.

benjaminRomano:
This script: https://github.com/bazelbuild/bazel/blob/8346ea4cfdd9fbd170d51a528fee26f912dad2d5/tools/android/runtime_deps/upload_android_tools.sh

ahumesky:
Gotcha, so these python tools aren't part of the remote android tools. These get embedded in bazel. So these changes will be part of the next bazel release.

PR #12753, opened 2020-12-26
Add support for mips64el

zqzzq:
my host is mips64el:
```
uname -m
mips64el
```
with these changes I was able to compile bazel in my host.

### DIFF
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -31,6 +31,15 @@ config_setting(
 )
 
 config_setting(
+    name = "linux_mips64el",
+    constraint_values = [
+        "@platforms//os:linux",
+        "@platforms//cpu:mips64el",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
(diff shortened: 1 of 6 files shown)


src/conditions/BUILD
```diff
@@ -30,6 +30,15 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
+    name = "linux_mips64el",
+    constraint_values = [
+        "@platforms//os:linux",
+        "@platforms//cpu:mips64el",
```
aiuto, line 37: Is this defined?
I wonder about the need for mips64el specificially as the first case.
I'm not up on my mips naming. Is the el designation different from other mips64 instruction sets?

meteorcloudy: Only `mips64` is defined at https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD#L119

@zqzzq Can you double check this?

meteorcloudy (changes requested):

PR #12760, opened 2020-12-29
Look for WORKSPACE.bazel as well in Bazel wrapper script

GMNGeoffrey:
WORKSPACE.bazel is a valid alternative name for this file
(https://docs.bazel.build/versions/master/build-ref.html#workspace)

### DIFF
--- a/scripts/packages/bazel.sh
+++ b/scripts/packages/bazel.sh
@@ -24,7 +24,7 @@ set -eu
 # 1. Set $USE_BAZEL_VERSION to a version number
 #    (e.g. export USE_BAZEL_VERSION=1.0.0).
 # 2. Add a .bazelversion file that contains a version number next to your
-#    WORKSPACE file.
+#    WORKSPACE[.bazel] file.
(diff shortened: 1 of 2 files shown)


GMNGeoffrey:
@philwo PTAL

philwo (approved):

philwo:
LGTM, but please fix the failing tests.

GMNGeoffrey:
> LGTM, but please fix the failing tests.

Done (syntax error). And since this revealed where the tests were, I added a test for this case. How do you usually merge PRs? Would you like me to squash this into a single commit?

Also, since this is the thing that detects which version to use, I'm not sure just releasing it with the next release it would be a part of makes sense. Basically no one who relies on `.bazelversion` could safely rename their WORKSPACE file to WORKSPACE.bazel right now as long as there's any likelihood of someone using bazel <4. Maybe it would make sense to do a patch release for old versions?

jin:
Imported PR to internal review, @philwo please feel free to LGTM that if there are no actionables here.

Issue #10399, opened 2019-12-10
http_archive with tar.gz files containing symlinks between files does not work on windows


speedskater:
### Description of the problem / feature request:

When using the `http_archive` repository rule on windows pointing to a tar.gz file containing symbolic links between individual files, the build fails when referencing a file corresponding to such a symbolic link.

This results the build to fail with the following error:

> ...external/test/bar (No such file or directory)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Create a repro.tar.gz archive containing a file `prefix/foo` and a symlink `prefix/bar` pointing to foo
2. Host this file on a local http server using something like `python3 -m http.server 80`
3. Create a new bazel project
4. Place the following code into your `WORKSPACE` file:

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "repro",
    build_file_content = """filegroup(name = "bar-file", srcs = ["bar"], visibility = ["//visibility:public"]);filegroup(name = "foo-file", srcs = ["foo"], visibility = ["//visibility:public"])""",
    sha256 = "7c3f4d1f373d8d155d3accc6a127f42c77b4923a2c3c53e4b0125d4ad4dd67e1"
    strip_prefix = "repro",
    urls = ["http://localhost/repro.tar.gz"],
)
```

5. Execute a simple build:

> bazel build @repro//:all

6. You should get a similar output:

```
[repro.zip](https://github.com/bazelbuild/bazel/files/3946333/repro.zip)

ERROR: missing input file '@repro//:bar':<project-directory>/k52n5tiz/external/repro/bar (No such file or directory)
ERROR: <project-directory>/k52n5tiz/external/test/BUILD.bazel:1:1: @repro//:bar-file: missing input file 'LabelCause{label=@repro//:bar, msg=missing input file '@test//:bar':<project-directory>/k52n5tiz/external/repro/bar (No such file or directory)}'
ERROR: <project-directory>/k52n5tiz/external/repro/BUILD.bazel:1:1 1 input file(s) do not exist
```

### What operating system are you running Bazel on?

Windows 10.0.17763

### What's the output of `bazel info release`?

release 1.2.1

### Any other information, logs, or outputs that you want to share?

[repro.zip](https://github.com/bazelbuild/bazel/files/3946329/repro.zip)

laszlocsomor:
Thanks for reporting this bug!
What is the expected behavior?

PR #12763, opened 2020-12-31
Update Tar decompress function to handle symlinks after their targets

mai93:
This PR updates the Tar archive decompress function to handle symbolic links entries only after their target is created. 

If the symbolic link target does not exist on the time the link entry is being processed, we store the link entry so it will be processed once their target entry is extracted.

This is done to avoid a problem that happens on Windows where dangling junctions are created for file symlinks whose targets are part of the archive being decompressed but come after the link entry in the archive stream.

Fixes: #10399


meteorcloudy (changes requested):
Thanks for working on fixing this!

I wonder if we can delay the symlink creations after all other files are extracted. This way the code should be simpler. Like we did in https://github.com/bazelbuild/bazel/commit/053bedecdbf046c385b24a22bb708f095bebfd72

If you also want to support A -> B -> C, where A and B are symlinks and C is a regular file. You can do a topological sort on the symlink entries then create them in order (eg. first B -> C, then A -> B). But this should be a small usage which https://github.com/bazelbuild/bazel/commit/053bedecdbf046c385b24a22bb708f095bebfd72 didn't fix.

And we definitely test for verifying this ;)

mai93:
I updated this fix and I can add the topological sort part to both Zip and Tar decompression in a separate PR.

meteorcloudy (approved):
Awesome, thank you!

PR #12777, opened 2021-01-06
Add SHA-1 to subresource integrity format for download() checksums

alexeagle:
npm packages commonly still use SHA-1. While it may be discouraged for its poor security, Bazel cannot enforce what external ecosystems currently do.

I tested this locally against a feature we are working on in rules_nodejs.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
@@ -43,6 +43,11 @@ public class Checksum {
     byte[] hash = null;
     int expectedLength = 0;
 
+    if (integrity.startsWith("sha1-")) {
+      keyType = KeyType.SHA1;
+      expectedLength = 20;
+      hash = decoder.decode(integrity.substring(5));
+    }
     if (integrity.startsWith("sha256-")) {
       keyType = KeyType.SHA256;
       expectedLength = 32;
@@ -63,7 +68,7 @@ public class Checksum {
       throw new IllegalArgumentException(
           "Unsupported checksum algorithm: '"
               + integrity
-              + "' (expected SHA-256, SHA-384, or SHA-512)");
+              + "' (expected SHA-1, SHA-256, SHA-384, or SHA-512)");
     }
 
     if (hash.length != expectedLength) {


alexeagle:
Here's someone who works at NPM confirming that packages published with an older version of the tool always have SHA-1 integrity: https://npm.community/t/sha1-vs-sha512-integrity/3416/2
so there isn't a different fix for this.

alexeagle:
As for testing in this PR, I followed what was here for SHA-384 which is to say, no test case specifically for each format.

aiuto:
Aggh..... we explicitly got rid of SHA1 for the obvious insecurity reasons. Perhaps we can find a way to restrict the use to particular upstream respositories, or flag it so that you have to explicitly allow it.

alexeagle:
Hi Tony!

I think it's reasonable for Bazel to be opinionated about not *producing* any SHA-1 hashes (maybe requiring some flag to opt-in as you say) but I don't understand why it should be opinionated about *verifying* those produced in other systems.

WDYT?

aiuto:
I don't remember the specific objections to SHA1. ISTR someone getting a peer bonus for removing it.
@philwo  Do you remember the details and have an opinion?

philwo:
I also remember that someone complained about Bazel supporting SHA-1. This was the context: https://github.com/bazelbuild/bazel/issues/8880

In that issue, the team agreed that supporting SHA-1 is better than not verifying the hash at all and that we should warn about it in the docs. In this case, there's nothing to add to our docs, because we have to take what's coming from upstream sources anyway (e.g. npm would be the place where that notice has to go).

IMHO we should merge this.

alexeagle:
Ping! Need this for a rules_nodejs repository rule improvement

philwo:
I'll import this.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
```diff
@@ -43,6 +43,11 @@ public static Checksum fromSubresourceIntegrity(String integrity) {
     byte[] hash = null;
     int expectedLength = 0;
 
+    if (integrity.startsWith("sha1-")) {
+      keyType = KeyType.SHA1;
+      expectedLength = 20;
```
meisterT, line 48: isn't this 40 for sha1 hashes?

meisterT: Ignore this, this is about octet length

PR #12779, opened 2021-01-06
Move more distfile dependencies to the new format.

aiuto:
1. Move more distfile dependencies to the new format.

The goal for this PR is to add all of the dependencies which are needed for building Bazel, but are not packaged with it. Essentially, the easy ones.
- Thus, this excludes the JDKs and some other Java tools because I don't know enough to reason about them yet.
- This also excludes dependencies which are explicitly used in WORKSPACE files for tests, like rules_proto. Those should each be done as standalone PRs because they impact the build structure.

2.  Change the format of the dependency list, so that we no longer presume everything in DIST_DEPS is in the //:additional_distfiles.  Now all the uses are called out explicitly in a "used_in" field. This allows us to eventually use the table for dependencies which are needed for our build and test, but do not have to be part of any distfiles, such as `bazel_website`.

Next steps:
- do the Java dependency changes
- replace the explicit coding to dependency versions in black box java tests with references to dist_deps.
- replace the explicit coding to dependency versions in black box shell tests with references to dist_deps


meteorcloudy (approved):
Awesome, thanks!

meteorcloudy (approved):

aiuto:
Merge CL sent for review.

PR #12787, opened 2021-01-08
Generate rules_java http_archive() instances from distdir_deps.bzl.

aiuto:
This consolidates the specification of the rules_java we depend on into a single place. We generate `http_archive()` stanzas for the main WORKSPACE as well as test harnesses from that, rather than keeping 3 instances in sync by hand. The old practice was error prone because you had to know about all the locations, and an update to one might not break the others immediately, but could be a lurking alignment problem.

This continues the work on #12081.


aiuto:
I expect the import CL will be somewhat painful because of copybara oddity.

comius (approved):

comius:
Nice job!

meteorcloudy (approved):

PR #12789, opened 2021-01-08
Attach local JDK to JVM8_TOOLCHAIN_CONFIGURATION.

comius:
This is a partial revert of ec29e28a7f729c6a899ad0a75c770c4d174d78af.
The commit support rules_appengine usecase, which needs to supply additional parameters to a toolchain compiling with JDK8.
Previously the whole commit broke repos that are using bazel-toolchains.

Caveat: the downstream still breaks rules_appengine on Java 11 only system (Ubuntu 18.04). I believe this is correct behaviour, because rules_appengine seem to support only Java 8.

Addresses issue: https://github.com/bazelbuild/rules_appengine/issues/119


comius:
Downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1842

katre (approved):

Issue #10249, opened 2019-11-15
Compilation modes dbg and opt have no effect with compiler mingw-gcc on Windows


johnplate:
### Description of the problem:

The compilation modes dbg and opt don't add any compiler flag (like -g or -g0 -O2 -DNDEBUG) when using compiler mingw-gcc on Windows.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

git clone https://github.com/bazelbuild/examples
cd examples\cpp-tutorial\stage1
bazel build --compiler=mingw-gcc -s --noshow_progress //main:hello-world
bazel build --compiler=mingw-gcc -c dbg -s --noshow_progress //main:hello-world
bazel build --compiler=mingw-gcc -c opt -s --noshow_progress //main:hello-world
The subcommands are identical except the output paths.

### What operating system are you running Bazel on?

Windows

### What's the output of `bazel info release`?

1.1.0

### Any other information, logs, or outputs that you want to share?

The relevant features in lines 1198 and 1200 in tools/cpp/windows_cc_toolchain_config.bzl are empty ...

philwo:
@oquenchil Just curious - why is this P4? It seems like this is basic C++ rules functionality that should work on Windows, so maybe we'd be happy to take a patch from someone? :)

oquenchil:
I don't remember why I lowered the priority if I had to guess my thought process was meteorcloudy would work on it (since he's assigned) but not any time soon. I can unassign and make P3.

PR #12792, opened 2021-01-08
Handle dbg and opt modes for mingw-gcc compiler

mai93:
This PR adds the required compiler and linker flags to support building cpp targets in debug and optimization modes while using mingw-gcc compiler on Windows.

Fixes: #10249


meteorcloudy (approved):
Nice, thank you! 😊

PR #12807, opened 2021-01-11
Move skylib to the new dependencies format.

aiuto:
Move the bazel_skylib and skydoc declarations from WORKSPACE to distdir_deps.bzl.
Eliminate the use of bazel-skylib.


meteorcloudy (approved):
🎉🎉🎉

Issue #10620, opened 2020-01-20
Bazel creates non-compliant classpath jar


rivashin:
Paths in the `Class-Path` attribute of `MANIFEST.MF` should be URL-encoded.
Currently, bazel only encodes spaces: https://github.com/bazelbuild/bazel/blob/912e822cd14710b2ebe8c7484c682d627f2906cb/src/tools/launcher/java_launcher.cc#L152

It should encode all other characters too, e.g. `%` - which is a valid character for a filename - should be replaced with `%25`. If path is not encoded correctly then Java will not find the jar and will silently ignore it, resulting in a puzzling NoClassDefFoundError.

Coursier generates paths with `%` when username/password or port is specified as part of maven repository URL: `cache/v1/https/username%40server.com%3a8080/foo.jar`. As a result, any java project built with bazel & jvm_rules_external will experience this issue when the classpath gets long enough.

See [this small example](https://github.com/rivashin/jar-classpath-bug) reproducing the issue.

meteorcloudy:
Thanks for reporting this, I will look into this

rivashin:
Any progress on this?
And why is it tagged with windows-team? This affects all platforms.

meteorcloudy:
@rivashin Sorry, I currently don't have any bandwidth to look into this. But I wound very much like to review the change if you can contribute to a fix!

philwo:
@mai93 maybe you want to look into this. :)

Bencodes:
I just put up a PR that updates the other Java launcher path https://github.com/bazelbuild/bazel/pull/12946

PR #12809, opened 2021-01-11
Convert jars paths in class-path manifest to be url-encoded

mai93:
This PR converts the paths in the Class-Path attribute of MANIFEST.MF to be URL-encoded. If a special character such as '%' is not encoded, then Java will not find the jar and this results in NoClassDefFoundError.

Fixes: #10620

### DIFF
--- a/examples/java-native/src/main/java/com/example/myproject/BUILD
+++ b/examples/java-native/src/main/java/com/example/myproject/BUILD
@@ -28,18 +28,6 @@ java_library(
     resources = ["//examples/java-native/src/main/resources:greeting"],
 )
 
-java_binary(
-    name = "hello-special-resources",
-    main_class = "com.example.myproject.Greeter",
-    runtime_deps = [":special%greeting"],
-)
-
-java_library(
-    name = "special%greeting",
-    srcs = ["Greeter.java"],
-    resources = ["//examples/java-native/src/main/resources:greeting"],
-)
-
 java_library(
     name = "hello-error-prone",
     srcs = ["HelloErrorProne.java"],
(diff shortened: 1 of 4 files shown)


examples/java-native/src/main/java/com/example/myproject/BUILD
```diff
@@ -28,6 +28,18 @@ java_library(
...
 )
 
+java_binary(
+    name = "hello-special-resources",
+    main_class = "com.example.myproject.Greeter",
+    runtime_deps = [":special%greeting"],
+)
+
+java_library(
+    name = "special%greeting",
```
meteorcloudy, line 38: It's a bit awkward to put this special target in our examples. Can you instead move the test to the bazel_windows_test?

src/tools/launcher/java_launcher.cc
```diff
@@ -147,19 +147,18 @@ static wstring GetManifestJarDir(const wstring& binary_base_path) {
...
-        *manifest_classpath << x;
-      }
+  for (const auto& x : jar_path) {
+    if (isalnum(x) || x == L'.' || x == L'-' || x == L'_' || x == L'~') {
+      *manifest_classpath << x;
+    }
+    else if (x == L'\\') {
+      *manifest_classpath << L"/";
+    }
+    else {
```
meteorcloudy, line 157: Nit: please move "else" to the previous line.

src/tools/launcher/java_launcher.cc
```diff
@@ -147,19 +147,18 @@ static wstring GetManifestJarDir(const wstring& binary_base_path) {
...
+     char buffer[4];
+     sprintf(buffer, "%02X", x);
```
meteorcloudy, line 159: Using "%02X" is smart, can you add a comment to explain this? It took me some time to Google and understand this. 😅

meteorcloudy (approved):

meteorcloudy (approved):
Thanks!

PR #12810, opened 2021-01-12
Convert java tool dependencies to the new distdep style

aiuto:
Only done for java tools, because I want feedback before continuing.

-  We name the same repositories with several aliases.  For example:
    -  remote_java_tools_linux_for_testing == remote_java_tools_test_linux  
    - openjdk_linux_vanilla == remotejdk11_linux_for_testing  == openjdk11_linux_archive
   Is this by design? Is it needed?
- Most of the time we only allow the google mirror to be used when we bring in a repo in a testing context.
   - is that right? Do we even want the URL or do we want a file URL?
   - when we build jdk.WORKSPACE, we *do* keep the github urls for rules_java and rules_cc. Why the difference.


comius:
> Only done for java tools, because I want feedback before continuing.
> 
> * We name the same repositories with several aliases.  For example:
>   
>   * remote_java_tools_linux_for_testing == remote_java_tools_test_linux

I've noticed that. I don't know the reason for it, but if the cleanup doesn't break anything please go for it. 
Notice that "for_testing" repos are used in test_base.py.

>   * openjdk_linux_vanilla == remotejdk11_linux_for_testing  == openjdk11_linux_archive
>     Is this by design? Is it needed?

Don't know.

> * Most of the time we only allow the google mirror to be used when we bring in a repo in a testing context.
>   
>   * is that right? Do we even want the URL or do we want a file URL?

I'd say this is correct. The tests are running over mirrored files as well as freshly built java_tools.

>   * when we build jdk.WORKSPACE, we _do_ keep the github urls for rules_java and rules_cc. Why the difference.

jdk.WORKSPACE reaches end-user and we might want to have two URL there, for redundancy. For the purpose of testing I'd argue one url is better.

WORKSPACE
```diff
@@ -382,12 +360,7 @@ http_archive(
 distdir_tar(
     name = "test_WORKSPACE_files",
     archives = [
-        "java_tools-v11.0.zip",
-        "java_tools_linux-v11.0.zip",
-        "java_tools_windows-v11.0.zip",
-        "java_tools_darwin-v11.0.zip",
         "coverage_output_generator-v2.5.zip",
-        "zulu11.37.17-ca-jdk11.0.6-linux_x64.tar.gz",
```
comius, line 390: Would you mind perhaps just handling java_tools_* in this PR and JDKs in other PRs? I assume JDKs have more more changes than currently in this PR, that is here only 1 of 4 JDKs is handled.

WORKSPACE
```diff
@@ -754,15 +694,12 @@ http_archive(
 )
 
 # This must be kept in sync with src/test/shell/bazel/testdata/jdk_http_archives.
-http_archive(
+dist_http_archive(
```
comius, line 697: Handle JDKs in another PR.

distdir_deps.bzl
```diff
@@ -211,8 +211,112 @@ DIST_DEPS = {
...
+    #    "archive": "zulu11.37.17-ca-jdk11.0.6-win_x64.zip",
+    #    "sha256": "a9695617b8374bfa171f166951214965b1d1d08f43218db9a2a780b71c665c18",
+    #    "urls": [
+    #        "https://mirror.bazel.build/openjdk/azul-zulu11.37.17-ca-jdk11.0.6/zulu11.37.17-ca-jdk11.0.6-win_x64.zip",
+    #    ],
+    #    "used_in": [
+    #        "test_WORKSPACE_files",
+    #    ],
+    #    "strip_prefix": "zulu11.37.17-ca-jdk11.0.6-win_x64",
+    #},
```
comius, line 259: Handle JDKs in another PR.

distdir_deps.bzl
```diff
@@ -239,6 +343,7 @@ http_archive(
 
     repo_stanzas = {}
     for repo in ctx.attr.repos:
+        # info = DIST_DEPS.get(repo) or _ALIASES[repo]
```
comius, line 346: Remove

comius (changes requested):

WORKSPACE
aiuto: Ack. Removed.

distdir_deps.bzl
aiuto: Ack

distdir_deps.bzl
aiuto: Done

WORKSPACE
aiuto: Certainly.

aiuto:
I won't try to resolve the different names in this PR.  I think the choice of what is correct belongs to you or the EngProd team. I'm just trying to make the update process easier for people.

As far as JDK.workspace. Yes. That is embedded into Bazel, so it always needs both urls. The PR now reflects that.

aiuto (commented):
I limited the scope to the Java tools.

PR #12811, opened 2021-01-12
update desugar_jdk_libs to version as of 2020-08-01

aiuto:
Update desugar_jdk_libs to commit b102111a91ee5910e680bc9e32146da3e081055e of 2020-08-01

Based on https://github.com/bazelbuild/bazel/pull/11821, but in the new style.

Closes #11821


ahumesky (approved):
Thanks!

PR #12817, opened 2021-01-12
Generate src/test/shell/bazel/testdata/jdk_http_archive

aiuto:
Start generating src/test/shell/bazel/testdata/jdk_http_archive instead of it being static.

This allows us to update various Java dependencies without having to manually keep WORKSPACE, jdk_http_archive and ./src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE in sync.

Question for the reviewer:
As this change stands now, the remote tools will have a secondary URL from github. Before, there was only the bazel mirror. E.g.
```
<     urls = ["https://mirror.bazel.build/bazel_java_tools/releases/java/v11.0/java_tools-v11.0.zip", "https://github.com/bazelbuild/java_tools/releases/download/java_v11.0/java_tools-v11.0.zip"],
---
>     urls = [
>          "https://mirror.bazel.build/bazel_java_tools/releases/java/v11.0/java_tools-v11.0.zip",
>     ],
```
I do not know if this is a good change or a bad one.  The JDKs all have 2 URLs, one from mirror and one from azul.com.  Having remote tools only from the mirror seems insufficient.  It would be easy to add a feature filter URLs, but I don't want to build that if we don't really need it.


aiuto:
But we can go right to cl/351576942

Issue #12816, opened 2021-01-12
empty virtual artifact doesn't have an execpath


keith:
```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'ActionLookupData{actionLookupKey=ConfiguredTargetKey{label=//bazel:UnitTestMainPlist, config=BuildConfigurationValue.Key[0ed7ba617dcfb598fbbd28fa6e73ad21403f13f6512470dc85621f514e24636a]}, actionIndex=0}' (requested by nodes 'ArtifactNestedSetKey{rawChildren=[File:[[<execution_root>]bazel-out/darwin-dbg/bin]bazel/UnitTestMainPlist.generated.plist, File:[[<execution_root>]bazel-out/darwin-dbg/bin]external/build_bazel_rules_apple/apple/internal/environment_plist_ios.plist, File:[[<execution_root>]bazel-out/darwin-dbg/bin]Modules/DependencyInjection/DependencyInjectionTests.__internal__.__test_bundle-intermediates/Info.plist-root-control]}', 'ArtifactNestedSetKey{rawChildren=[File:[[<execution_root>]bazel-out/darwin-dbg/bin]bazel/UnitTestMainPlist.generated.plist, File:[[<execution_root>]bazel-out/darwin-dbg/bin]external/build_bazel_rules_apple/apple/internal/environment_plist_ios.plist, File:[[<execution_root>]bazel-out/darwin-dbg/bin]Modules/ResourceSupport/ResourceSupportTests.__internal__.__test_bundle-intermediates/Info.plist-root-control]}')
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.UnsupportedOperationException: empty virtual artifact doesn't have an execpath
	at com.google.devtools.build.lib.actions.cache.VirtualActionInput$EmptyActionInput.getExecPathString(VirtualActionInput.java:64)
[...]

PR #12819, opened 2021-01-13
Ignore empty virtual artifacts when spawn-logging inputs.

benjaminp:
Fixes https://github.com/bazelbuild/bazel/issues/12816.

### DIFF
--- a/src/test/shell/bazel/bazel_execlog_test.sh
+++ b/src/test/shell/bazel/bazel_execlog_test.sh
@@ -119,6 +119,26 @@ EOF
   wc output || fail "no output produced"
 }
 
+function test_empty_file_in_runfiles() {
+  mkdir d
+  touch d/main.py
+  cat > BUILD <<'EOF'
+py_binary(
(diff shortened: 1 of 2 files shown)


src/test/shell/bazel/bazel_execlog_test.sh
```diff
@@ -119,6 +119,26 @@ EOF
...
+)
+genrule(
+    name = "rule",
+    outs = ["out.txt"],
+    tools = [":py_tool"],
+    cmd = "echo hello > $(location out.txt)"
+)
+EOF
+  bazel build //:rule --experimental_execution_log_file output 2>&1 >> $TEST_log || fail "could not build"
+  [[ -e output ]] || "no output produced"
```
meisterT, line 139: nit: add `fail` after `||`

benjaminp: good catch

meisterT (approved):

Issue #5588, opened 2018-07-13
Support other output modes than 0555


wking:
### Description of the problem / feature request:

When building tarballs with [`pkg_tar`][1], the *output* tarball has a 0555 mode.  But tarballs are generally not going to be executable.  I think the default mode should be 0666 or, if you want to be very conservative 0444.  In both those cases, it would be nice to respect the user's umask, as long as it did not interfere with reproducible builds.  An alternative would be to provide an explicit `pkg_tar` parameter that functions similarly to the existing `mode` (although `mode` is for files packaged inside the tarball, not for the tarball itself, #2925).

### Feature requests: what underlying problem are you trying to solve with this feature?

Claiming that output tarballs are executable makes it harder to auto-complete sibling paths that are executable.  And if you attempt to execute the tarball, I think an EPERM is more obvious than "cannot execute binary file".

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```console
$ bazel version
Build label: 0.15.0- (@non-git)
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Jun 27 14:43:20 2018 (1530110600)
Build timestamp: 1530110600
Build timestamp as int: 1530110600
$ cat WORKSPACE 
workspace(name = "test")
$ cat BUILD.bazel 
load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")

pkg_tar(
    name = "test",
    srcs = glob(
        ["*.txt"],
        exclude_directories = 1,
    ),
    mode = "0666",
)
$ touch text.txt 
$ bazel build test
Starting local Bazel server and connecting to it...
........
INFO: Analysed target //:test (16 packages loaded).
INFO: Found 1 target...
Target //:test up-to-date:
  bazel-bin/test.tar
INFO: Elapsed time: 1.360s, Critical Path: 0.11s
INFO: 1 process: 1 processwrapper-sandbox.
INFO: Build completed successfully, 6 total actions
$ ls -l bazel-bin/test.tar 
-r-xr-xr-x. 1 wking wking 10240 Jul 12 21:49 bazel-bin/test.tar
```

### What operating system are you running Bazel on?

RHEL 7.5 (Linux).

### What's the output of `bazel info release`?

```console
$ bazel info release
release 0.15.0- (@non-git)
```

[...]

PR #12823, opened 2021-01-13
Remote Output Service: place bazel-out/ on a FUSE file system

EdSchouten:
Builds that yield large output files can generate lots of network
bandwidth when remote execution is used. To combat this, we have flags
such as --remote_download_minimal to disable downloading of output
files. Unfortunately, this makes it hard to perform ad hoc exploration
of build outputs.

In an attempt to solve this, this change adds an option to Bazel named
--remote_output_service. When enabled, Bazel effectively gives up the
responsibility of managing a bazel-out/ directory. Instead, it calls
into a gRPC service to request a directory and creates a symlink that
points to it.

Smart implementations of this gRPC service may use things like FUSE to
let this replacement bazel-out/ directory be lazy-loading, thereby
reducing network bandwidth significantly. In order to create
lazy-loading files and directories, Bazel can call into a BatchCreate()
RPC that takes a list of Output{File,Directory,Symlink} messages,
similar to REv2's ActionResult. This call is also used to create
runfiles directories by providing fictive instances of OutputSymlink.

To prevent Bazel from reading the contents of files stored in the FUSE
file system (which would cause network I/O), the protocol offers a
BatchStat() call that can return information such as the REv2 digest.
Though this is redundant with --unix_digest_hash_attribute_name, there
are a couple of reasons why I added this feature:

1. For non-Linux operating systems, it may make more sense to use NFSv4
   instead of FUSE (i.e., running a virtual NFS daemon on localhost).
   Even though RFC 8276 adds support for extended attributes to NFSv4,
   not all operating systems implement it.

2. It addresses the security/hermeticity concerns that people raised
   when this feature was added. There is no way to add extended
   attributes to files that can't be tampered with (as a non-root user),
   while using gRPC solves that problem inherently.

3. Callers of Bazel's BatchStat.batchStat() may generate many system
   calls successively. This causes a large number of context switches
   between Bazel and the FUSE daemon. Using gRPC seems to be cheaper.

By requiring that the output path returned by the gRPC service is
writable, no-remote actions can still run as before, both with
sandboxing enabled and disabled. The only difference is that it will use
space on the gRPC service side, as opposed to the user's home directory
(though the gRPC service may continue to store data in the user's home
directory.

I have a server implementation is written in Go on top of Buildbarn's
storage and file system layer. My plan is to release the code for this
service as soon as I've got a 'thumbs up' on the overall approach.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
@@ -1,0 +1,626 @@
+package com.google.devtools.build.lib.remote;
+
+import static com.google.common.collect.Iterators.filter;
+import static com.google.common.collect.Iterators.partition;
+import static java.lang.String.format;
+
+import build.bazel.remote.execution.v2.ActionResult;
+import build.bazel.remote.execution.v2.Digest;
+import build.bazel.remote.execution.v2.DigestFunction;
+import build.bazel.remote.execution.v2.OutputDirectory;
+import build.bazel.remote.execution.v2.OutputFile;
+import build.bazel.remote.execution.v2.OutputSymlink;
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.UnmodifiableIterator;
+import com.google.common.flogger.GoogleLogger;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.devtools.build.lib.actions.Action;
+import com.google.devtools.build.lib.actions.ActionInputMap;
+import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.actions.ArtifactPathResolver;
+import com.google.devtools.build.lib.actions.BuildFailedException;
+import com.google.devtools.build.lib.actions.EnvironmentalExecException;
+import com.google.devtools.build.lib.actions.ExecException;
+import com.google.devtools.build.lib.actions.FilesetOutputSymlink;
+import com.google.devtools.build.lib.actions.LostInputsActionExecutionException;
+import com.google.devtools.build.lib.actions.cache.OutputMetadataStore;
+import com.google.devtools.build.lib.events.EventHandler;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceGrpc.RemoteOutputServiceBlockingStub;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceGrpc.RemoteOutputServiceFutureStub;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.BatchCreateRequest;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.BatchStatRequest;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.BatchStatResponse;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.CleanRequest;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.FileStatus;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.FinalizeBuildRequest;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.InitialOutputPathContents;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.StartBuildRequest;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.StartBuildResponse;
+import com.google.devtools.build.lib.remote.RemoteOutputServiceProto.StatResponse;
+import com.google.devtools.build.lib.remote.util.DigestUtil;
+import com.google.devtools.build.lib.runtime.CommandEnvironment;
+import com.google.devtools.build.lib.util.AbruptExitException;
+import com.google.devtools.build.lib.vfs.BatchStat;
+import com.google.devtools.build.lib.vfs.DelegateFileSystem;
+import com.google.devtools.build.lib.vfs.DigestHashFunction;
+import com.google.devtools.build.lib.vfs.FileStatusWithDigest;
+import com.google.devtools.build.lib.vfs.FileSystem;
+import com.google.devtools.build.lib.vfs.ModifiedFileSet;
+import com.google.devtools.build.lib.vfs.OutputService;
+import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.build.lib.vfs.PathFragment;
+import com.google.devtools.build.lib.vfs.Root;
+import com.google.devtools.build.skyframe.SkyFunction.Environment;
+import io.grpc.Status.Code;
+import io.grpc.StatusRuntimeException;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+public class GrpcRemoteOutputService implements OutputService, ActionResultDownloader {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+
+  // State that needs to be retained across invocations of Bazel. The
+  // current build ID is placed inside, because it ensures that cached
+  // instances of Path that use GrpcFileSystem always generate RPCs to
+  // the Remote Output Service using the current build ID; not the one
+  // under which they were created.
+  public static class StateAcrossBuilds {
+    @Nullable private UUID currentBuildId;
+    @Nullable private UUID previousBuildId;
+  }
+
+  private final StateAcrossBuilds stateAcrossBuilds;
+  private final ReferenceCountedChannel channel;
+  private final String outputBaseId;
+  private final PathFragment outputPathPrefix;
+  private final String instanceName;
+  private final DigestFunction.Value digestFunction;
+  private final Digest emptyDigest;
+  private final RemoteRetrier retrier;
+  private final CommandEnvironment env;
+
+  private PathFragment currentRelativeOutputPath;
+  private boolean currentBuildSuccessful;
+
+  public GrpcRemoteOutputService(StateAcrossBuilds stateAcrossBuilds, ReferenceCountedChannel channel, String outputBaseId, PathFragment outputPathPrefix, String instanceName, DigestUtil digestUtil, RemoteRetrier retrier, CommandEnvironment env) {
+    this.stateAcrossBuilds = stateAcrossBuilds;
+    this.channel = channel;
+    this.outputBaseId = outputBaseId;
+    this.outputPathPrefix = outputPathPrefix;
+    this.instanceName = instanceName;
+    this.digestFunction = digestUtil.getDigestFunction();
+    this.emptyDigest = digestUtil.compute(new byte[] {});
+    this.retrier = retrier;
+    this.env = env;
+  }
+
+
+  @Override
+  public String getFilesSystemName() {
+    return "GrpcRemoteOutputService";
+  }
+
+  @Override
+  public ModifiedFileSet startBuild(
+      EventHandler eventHandler, UUID buildId, boolean finalizeActions)
+      throws BuildFailedException, AbruptExitException, InterruptedException {
+    // Notify the remote output service that the build is about to
+    // start. The remote output service will return the directory in
+    // which it wants us to let the build take place.
+    //
+    // Make the output service aware of the location of the output path
+    // from our perspective and any symbolic links that will point to
+    // it. This allows the service to properly resolve symbolic links
+    // containing absolute paths as part of BatchStat() calls.
+    StartBuildRequest.Builder builder = StartBuildRequest.newBuilder();
+    builder.setOutputBaseId(outputBaseId);
+    builder.setBuildId(buildId.toString());
+    builder.setInstanceName(instanceName);
+    builder.setDigestFunction(digestFunction);
+    builder.setOutputPathPrefix(outputPathPrefix.toString());
+    String relativeOutputPath = env.getDirectories().getRelativeOutputPath();
+    Path originalOutputPath = env.getExecRoot().getRelative(relativeOutputPath);
+    builder.putOutputPathAliases(originalOutputPath.toString(), ".");
+    StartBuildRequest request = builder.build();
+    StartBuildResponse response;
+    try {
+      response = retrier.execute(
+          () ->
+              channel.withChannelBlocking(
+                  channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                      .startBuild(request)));
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+
+    // Replace the output path with a symbolic link pointing to the
+    // directory managed by the remote output service.
+    PathFragment outputPath = outputPathPrefix.getRelative(response.getOutputPathSuffix());
+    try {
+      try {
+        originalOutputPath.deleteTree();
+      } catch (FileNotFoundException e) {}
+      originalOutputPath.createSymbolicLink(outputPath);
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+
+    stateAcrossBuilds.currentBuildId = buildId;
+    currentRelativeOutputPath = PathFragment.create(relativeOutputPath);
+
+    if (stateAcrossBuilds.previousBuildId == null || !response.hasInitialOutputPathContents()) {
+      // Either Bazel or the remote output service has performed no
+      // build before.
+      return ModifiedFileSet.EVERYTHING_MODIFIED;
+    }
+    InitialOutputPathContents initialContents = response.getInitialOutputPathContents();
+    if (!initialContents.getBuildId().equals(stateAcrossBuilds.previousBuildId.toString())) {
+      // Bazel and the remote output service disagree on the build ID of
+      // the previous build.
+      return ModifiedFileSet.EVERYTHING_MODIFIED;
+    }
+    // Bazel and the remote output service agree on the build ID of the
+    // previous build. Return the set of paths that have been modified.
+    //
+    // TODO: Do these paths need to be relative to the exec root or the
+    // output path? Assume the exec root for now.
+    return ModifiedFileSet.builder()
+        .modifyAll(
+            Iterables.transform(
+                initialContents.getModifiedPathsList(),
+                (p) -> currentRelativeOutputPath.getRelative(p)))
+        .build();
+  }
+
+  @Override
+  public void finalizeBuild(boolean buildSuccessful)
+      throws BuildFailedException, AbruptExitException, InterruptedException {
+    currentBuildSuccessful = buildSuccessful;
+  }
+
+  public void sendFinalizeBuild() {
+    if (stateAcrossBuilds.currentBuildId != null) {
+      FinalizeBuildRequest.Builder builder = FinalizeBuildRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setBuildSuccessful(currentBuildSuccessful);
+
+      try {
+        channel.withChannelBlocking(
+            channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                .finalizeBuild(builder.build()));
+        stateAcrossBuilds.previousBuildId = stateAcrossBuilds.currentBuildId;
+      } catch (InterruptedException e) {
+        stateAcrossBuilds.previousBuildId = null;
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      } finally {
+        stateAcrossBuilds.currentBuildId = null;
+        currentRelativeOutputPath = null;
+      }
+    }
+  }
+
+  @Override
+  public void finalizeAction(Action action, OutputMetadataStore outputMetadataStore)
+      throws IOException, EnvironmentalExecException, InterruptedException {
+    // TODO: Would this be the right place to call into the remote
+    // output service to check whether any I/O errors occurred? If so,
+    // we should likely let createActionFileSystem() call into the
+    // remote output service to start capturing I/O errors.
+  }
+
+  private String fixupExecRootPath(PathFragment path) {
+    return path.relativeTo(currentRelativeOutputPath).toString();
+  }
+
+  private static abstract class DumbFileStatus implements FileStatusWithDigest {
+    @Override
+    public long getSize() throws IOException {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public long getLastModifiedTime() throws IOException {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public long getLastChangeTime() throws IOException {
+      return -1;
+    }
+
+    @Override
+    public long getNodeId() throws IOException {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public byte[] getDigest() throws IOException {
+      throw new UnsupportedOperationException();
+    }
+  }
+
+  private static class RegularFileStatus extends DumbFileStatus {
+    private final long size;
+    private final byte[] digest;
+
+    RegularFileStatus(long size, byte[] digest) {
+      this.size = size;
+      this.digest = digest;
+    }
+
+    @Override
+    public boolean isFile() {
+      return true;
+    }
+
+    @Override
+    public boolean isDirectory() {
+      return false;
+    }
+
+    @Override
+    public boolean isSymbolicLink() {
+      return false;
+    }
+
+    @Override
+    public boolean isSpecialFile() {
+      return false;
+    }
+
+    @Override
+    public long getSize() throws IOException {
+      return size;
+    }
+
+    @Override
+    public byte[] getDigest() throws IOException {
+      return digest;
+    }
+  }
+
+  private static class DirectoryFileStatus extends DumbFileStatus {
+    private final long lastModifiedTime;
+
+    DirectoryFileStatus(long lastModifiedTime) {
+      this.lastModifiedTime = lastModifiedTime;
+    }
+
+    @Override
+    public boolean isFile() {
+      return false;
+    }
+
+    @Override
+    public boolean isDirectory() {
+      return true;
+    }
+
+    @Override
+    public boolean isSymbolicLink() {
+      return false;
+    }
+
+    @Override
+    public boolean isSpecialFile() {
+      return false;
+    }
+
+    @Override
+    public long getLastModifiedTime() throws IOException {
+      return this.lastModifiedTime;
+    }
+
+  }
+
+  private static class SymlinkFileStatus extends DumbFileStatus {
+    @Override
+    public boolean isFile() {
+      return false;
+    }
+
+    @Override
+    public boolean isDirectory() {
+      return false;
+    }
+
+    @Override
+    public boolean isSymbolicLink() {
+      return true;
+    }
+
+    @Override
+    public boolean isSpecialFile() {
+      return false;
+    }
+  }
+
+  private class GrpcBatchStat implements BatchStat {
+
+    public List<FileStatusWithDigest> batchStat(boolean includeDigest,
+                                                boolean includeLinks,
+                                                Iterable<PathFragment> paths)
+        throws IOException, InterruptedException {
+      // TODO: Do we need to partition the input, just like in
+      // createSymlinkTree(), or is input already guaranteed to be
+      // bounded in size?
+      BatchStatRequest.Builder builder = BatchStatRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setIncludeFileDigest(includeDigest);
+      for (PathFragment path : paths) {
+        builder.addPaths(fixupExecRootPath(path));
+      }
+
+      BatchStatResponse responses;
+      try {
+        responses = channel.withChannelBlocking(
+            channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                .batchStat(builder.build()));
+      } catch (StatusRuntimeException e) {
+        throw new IOException(e);
+      }
+      return Lists.newArrayList(
+          Iterables.transform(
+              responses.getResponsesList(),
+              (response) -> {
+                if (!response.hasFileStatus()) {
+                  // File not found.
+                  return null;
+                }
+                FileStatus fileStatus = response.getFileStatus();
+                if (fileStatus.hasFile()) {
+                  Digest digest = fileStatus.getFile().getDigest();
+                  return new RegularFileStatus(digest.getSizeBytes(), DigestUtil.toBinaryDigest(digest));
+                }
+                if (fileStatus.hasDirectory()) {
+                  return new DirectoryFileStatus(fileStatus.getDirectory().getLastModifiedTime().getSeconds());
+                }
+                if (fileStatus.hasSymlink()) {
+                  return new SymlinkFileStatus();
+                }
+                throw new UnsupportedOperationException();
+              }));
+    }
+  }
+
+  @Override
+  public BatchStat getBatchStatter() {
+    return new GrpcBatchStat();
+  }
+
+  @Override
+  public boolean canCreateSymlinkTree() {
+    return true;
+  }
+
+  @Override
+  public void createSymlinkTree(Map<PathFragment, PathFragment> symlinks, PathFragment symlinkTreeRoot)
+      throws ExecException, InterruptedException {
+    // The provided set of symbolic links may be too large to provide to
+    // the remote output service at once. Partition the symbolic links
+    // in groups of 1000, so that BatchCreateRequest messages remain
+    // small enough.
+    UnmodifiableIterator<List<Map.Entry<PathFragment, PathFragment>>> symlinkBatchIterator =
+        partition(symlinks.entrySet().iterator(), 1000);
+    boolean cleanPathPrefix = true;
+    while (symlinkBatchIterator.hasNext()) {
+      List<Map.Entry<PathFragment, PathFragment>> symlinksBatch = symlinkBatchIterator.next();
+      BatchCreateRequest.Builder builder = BatchCreateRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setPathPrefix(symlinkTreeRoot.relativeTo(currentRelativeOutputPath).toString());
+      // During the first iteration, we should ensure that any existing
+      // contents of the symlink tree directory are removed. This may be
+      // triggered by setting the 'clean_path_prefix' option.
+      builder.setCleanPathPrefix(cleanPathPrefix);
+      for (Map.Entry<PathFragment, PathFragment> symlink : symlinksBatch) {
+        PathFragment target = symlink.getValue();
+        if (target == null) {
+          // No target indicates an empty file needs to be created.
+          OutputFile.Builder fileBuilder = builder.addFilesBuilder();
+          fileBuilder.setPath(symlink.getKey().toString());
+          fileBuilder.setDigest(emptyDigest);
+          fileBuilder.setIsExecutable(true);
+        } else {
+          OutputSymlink.Builder symlinkBuilder = builder.addSymlinksBuilder();
+          symlinkBuilder.setPath(symlink.getKey().toString());
+          symlinkBuilder.setTarget(symlink.getValue().toString());
+        }
+      }
+      try {
+        channel.withChannelBlocking(
+            channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                .batchCreate(builder.build()));
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      }
+      cleanPathPrefix = false;
+    }
+  }
+
+  @Override
+  public void clean() throws ExecException, InterruptedException {
+    CleanRequest.Builder builder = CleanRequest.newBuilder();
+    builder.setOutputBaseId(outputBaseId);
+    try {
+      channel.withChannelBlocking(
+          channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+              .clean(builder.build()));
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  public ActionFileSystemType actionFileSystemType() {
+    return ActionFileSystemType.REMOTE_FILE_SYSTEM;
+  }
+
+  private class GrpcFileSystem extends DelegateFileSystem {
+    private final PathFragment originalOutputPath;
+
+    public GrpcFileSystem(FileSystem sourceDelegate, PathFragment originalOutputPath) {
+      super(sourceDelegate);
+      this.originalOutputPath = originalOutputPath;
+    }
+
+    @Override
+    protected byte[] getFastDigest(PathFragment path) throws IOException {
+      // Don't attempt to compute digests ourselves. Call into the
+      // remote output service to request the digest. The service may be
+      // able to return the digest from its bookkeeping, as opposed to
+      // reading the actual file contents.
+      BatchStatRequest.Builder builder = BatchStatRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setIncludeFileDigest(true);
+      builder.setFollowSymlinks(true);
+      PathFragment relativePath;
+      try {
+        relativePath = path.relativeTo(originalOutputPath);
+      } catch (IllegalArgumentException e) {
+        // Path is outside the output path. Send it to the regular file system.
+        return super.getFastDigest(path);
+      }
+      builder.addPaths(relativePath.toString());
+
+      BatchStatResponse responses;
+      try {
+        responses = channel.withChannelBlocking(
+            channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                .batchStat(builder.build()));
+      } catch (InterruptedException e) {
+        return null;
+      } catch (StatusRuntimeException e) {
+        if (e.getStatus().getCode() == Code.CANCELLED && Thread.currentThread().isInterrupted()) {
+          // The current thread is interrupted, meaning that gRPC calls
+          // into the remote output service are cancelled automatically.
+          // Return null, indicating that a fast digest is not
+          // available.
+          //
+          // Throwing an IOException in this case may cause other parts
+          // of Bazel to throw InconsistentFilesystemExceptions, which
+          // leads to spurious build error messages.
+          return null;
+        }
+        throw new IOException(e);
+      }
+      if (responses.getResponsesCount() != 1) {
+        throw new IOException("Remote output service returned an unexpected number of stat responses");
+      }
+      StatResponse response = responses.getResponses(0);
+      if (!response.hasFileStatus()) {
+        // Output service reports that the file does not exist.
+        throw new FileNotFoundException();
+      }
+      FileStatus fileStatus = response.getFileStatus();
+      if (fileStatus.hasFile()) {
+        FileStatus.File regularFileStatus = fileStatus.getFile();
+        return DigestUtil.toBinaryDigest(regularFileStatus.getDigest());
+      }
+      if (fileStatus.hasExternal()) {
+        // Path is a symbolic link that points to a location outside of
+        // the output path. The output service is unable to give a
+        // digest. We must compute it ourselves.
+        return super.getFastDigest(
+            originalOutputPath.getRelative(fileStatus.getExternal().getNextPath()));
+      }
+      throw new IOException(format("Remote output service did not return the status of file %s", path));
+    }
+  }
+
+  @Override
+  @Nullable
+  public FileSystem createActionFileSystem(
+      FileSystem sourceDelegate,
+      PathFragment execRootFragment,
+      String relativeOutputPath,
+      ImmutableList<Root> sourceRoots,
+      ActionInputMap inputArtifactData,
+      Iterable<Artifact> outputArtifacts,
+      boolean trackFailedRemoteReads) {
+    return new GrpcFileSystem(sourceDelegate, execRootFragment.getRelative(relativeOutputPath));
+  }
+
+  @Override
+  public void checkActionFileSystemForLostInputs(FileSystem actionFileSystem, Action action)
+      throws LostInputsActionExecutionException {
+    // There is no need to provide an implementation for this function,
+    // as StartBuild() on the server is supposed to call
+    // FindMissingBlobs() for all remotely tracked files. This ensures
+    // that they cannot disappear during the build.
+  }
+
+  public ListenableFuture<Void> downloadActionResult(ActionResult actionResult) {
+    // Request that all outputs of the action are created. Do make sure
+    // to remove the "bazel-out/" prefix from all paths, as that part of
+    // the path is not managed by the remote output service.
+    BatchCreateRequest.Builder builder = BatchCreateRequest.newBuilder();
+    builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+    for (OutputFile file : actionResult.getOutputFilesList()) {
+      // As there is no guarantee that permissions on files created by
+      // the remote output service can be modified, make sure that all
+      // downloaded files are marked executable.
+      // TODO: Do we only want to do this selectively?
+      builder.addFilesBuilder()
+          .mergeFrom(file)
+          .setPath(fixupExecRootPath(PathFragment.create(file.getPath())))
+          .setIsExecutable(true);
+    }
+    for (OutputDirectory directory : actionResult.getOutputDirectoriesList()) {
+      builder.addDirectoriesBuilder()
+          .mergeFrom(directory)
+          .setPath(fixupExecRootPath(PathFragment.create(directory.getPath())));
+    }
+    for (OutputSymlink symlink : Iterables.concat(actionResult.getOutputFileSymlinksList(), actionResult.getOutputDirectorySymlinksList())) {
+      builder.addSymlinksBuilder()
+          .mergeFrom(symlink)
+          .setPath(fixupExecRootPath(PathFragment.create(symlink.getPath())));
+    }
+
+    return Futures.catchingAsync(
+        Futures.transform(
+            channel.withChannelFuture(
+                channel ->
+                    RemoteOutputServiceGrpc.newFutureStub(channel)
+                        .batchCreate(builder.build())),
+            (result) -> null,
+            MoreExecutors.directExecutor()),
+        StatusRuntimeException.class,
+        (sre) -> Futures.immediateFailedFuture(new IOException(sre)),
+        MoreExecutors.directExecutor());
+  }
+
+  @Override
+  public boolean supportsPathResolverForArtifactValues() {
+    return true;
+  }
+
+  @Override
+  public ArtifactPathResolver createPathResolverForArtifactValues(
+      PathFragment execRoot,
+      String relativeOutputPath,
+      FileSystem fileSystem,
+      ImmutableList<Root> pathEntries,
+      ActionInputMap actionInputMap,
+      Map<Artifact, ImmutableCollection<? extends Artifact>> expandedArtifacts,
+      Map<Artifact, ImmutableList<FilesetOutputSymlink>> filesets) {
+    FileSystem remoteFileSystem =
+        new GrpcFileSystem(fileSystem, execRoot.getRelative(relativeOutputPath));
+    return ArtifactPathResolver.createPathResolver(remoteFileSystem, fileSystem.getPath(execRoot));
+  }
+}
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -14,6 +14,8 @@
 
 package com.google.devtools.build.lib.remote;
 
+import static com.google.common.hash.Hashing.md5;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 import build.bazel.remote.execution.v2.ActionCacheUpdateCapabilities;
@@ -100,6 +102,7 @@ import com.google.devtools.build.lib.vfs.FileSystem;
 import com.google.devtools.build.lib.vfs.OutputPermissions;
 import com.google.devtools.build.lib.vfs.OutputService;
 import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.common.options.OptionsBase;
 import com.google.devtools.common.options.OptionsParsingResult;
 import io.grpc.CallCredentials;
@@ -141,7 +144,10 @@ public final class RemoteModule extends BlazeModule {
   @Nullable private RemoteActionContextProvider actionContextProvider;
   @Nullable private RemoteActionInputFetcher actionInputFetcher;
   @Nullable private RemoteOptions remoteOptions;
-  @Nullable private RemoteOutputService remoteOutputService;
+  @Nullable private CommandEnvironment env;
+  @Nullable private OutputService remoteOutputService;
+  @Nullable private GrpcRemoteOutputService.StateAcrossBuilds remoteOutputServiceStateAcrossBuilds =
+      new GrpcRemoteOutputService.StateAcrossBuilds();
   @Nullable private TempPathGenerator tempPathGenerator;
   @Nullable private BlockWaitingModule blockWaitingModule;
   @Nullable private RemoteOutputChecker remoteOutputChecker;
@@ -285,7 +291,8 @@ public final class RemoteModule extends BlazeModule {
             remoteCache,
             /* retryScheduler= */ null,
             digestUtil,
-            remoteOutputChecker);
+            remoteOutputChecker,
+            /* actionResultDownloader= */ null);
   }
 
   @Override
@@ -313,6 +320,7 @@ public final class RemoteModule extends BlazeModule {
     }
 
     this.remoteOptions = remoteOptions;
+    this.env = env;
 
     AuthAndTLSOptions authAndTlsOptions = env.getOptions().getOptions(AuthAndTLSOptions.class);
     DigestHashFunction hashFn = env.getRuntime().getFileSystem().getDigestFunction();
@@ -631,6 +639,47 @@ public final class RemoteModule extends BlazeModule {
             cacheChannel.retain(), callCredentialsProvider, remoteOptions, retrier, digestUtil);
     cacheChannel.release();
 
+    ActionResultDownloader actionResultDownloader = null;
+    Preconditions.checkState(remoteOutputService == null, "remoteOutputService must be null");
+    if (remoteOptions.remoteOutputService != null) {
+      if (remoteOptions.remoteOutputsMode != RemoteOutputsMode.ALL) {
+        throw createOptionsExitException(
+            "If --remote_output_service is specified, --remote_download_outputs must be set to \"all\"",
+            FailureDetails.RemoteOptions.Code.OUTPUT_SERVICE_WITH_INCOMPATIBLE_REMOTE_OUTPUTS_MODE);
+      }
+      String outputBaseId = remoteOptions.remoteOutputServiceOutputBaseId;
+      if (Strings.isNullOrEmpty(outputBaseId)) {
+        outputBaseId = DigestUtil.hashCodeToString(md5().hashString(env.getWorkspace().toString(), UTF_8));
+      }
+      String outputPathPrefix = remoteOptions.remoteOutputServiceOutputPathPrefix;
+      if (Strings.isNullOrEmpty(outputPathPrefix)) {
+        throw createOptionsExitException(
+            "If --remote_output_service is specified, --remote_output_service_output_path_prefix must be set as well",
+            FailureDetails.RemoteOptions.Code.OUTPUT_SERVICE_WITHOUT_OUTPUT_PATH_REFIX);
+      }
+
+      ReferenceCountedChannel channel =
+          new ReferenceCountedChannel(
+              new GoogleChannelConnectionFactory(
+                  channelFactory,
+                  remoteOptions.remoteOutputService,
+                  remoteOptions.remoteProxy,
+                  authAndTlsOptions,
+                  ImmutableList.of(),
+                  maxConcurrencyPerConnection));
+      GrpcRemoteOutputService grpcRemoteOutputService = new GrpcRemoteOutputService(
+          remoteOutputServiceStateAcrossBuilds,
+          channel,
+          outputBaseId,
+          PathFragment.create(outputPathPrefix),
+          remoteOptions.remoteInstanceName,
+          digestUtil,
+          retrier,
+          env);
+      remoteOutputService = grpcRemoteOutputService;
+      actionResultDownloader = grpcRemoteOutputService;
+    }
+
     if (enableRemoteExecution) {
       if (enableDiskCache) {
         try {
@@ -686,7 +735,8 @@ public final class RemoteModule extends BlazeModule {
               retryScheduler,
               digestUtil,
               logDir,
-              remoteOutputChecker);
+              remoteOutputChecker,
+              actionResultDownloader);
       repositoryRemoteExecutorFactoryDelegate.init(
           new RemoteRepositoryRemoteExecutorFactory(
               remoteCache,
@@ -717,7 +767,13 @@ public final class RemoteModule extends BlazeModule {
               cacheCapabilities.getCacheCapabilities(), cacheClient, remoteOptions, digestUtil);
       actionContextProvider =
           RemoteActionContextProvider.createForRemoteCaching(
-              executorService, env, remoteCache, retryScheduler, digestUtil, remoteOutputChecker);
+              executorService,
+              env,
+              remoteCache,
+              retryScheduler,
+              digestUtil,
+              remoteOutputChecker,
+              actionResultDownloader);
     }
 
     buildEventArtifactUploaderFactoryDelegate.init(
@@ -914,7 +970,7 @@ public final class RemoteModule extends BlazeModule {
     actionContextProvider = null;
     actionInputFetcher = null;
     remoteOptions = null;
-    remoteOutputService = null;
+    env = null;
     tempPathGenerator = null;
     rpcLogFile = null;
     remoteOutputChecker = null;
@@ -947,6 +1003,16 @@ public final class RemoteModule extends BlazeModule {
             ExitCode.LOCAL_ENVIRONMENTAL_ERROR,
             Code.RPC_LOG_FAILURE);
       }
+    }
+  }
+
+  @Override
+  public void commandComplete() {
+    if (remoteOutputService != null) {
+      if (remoteOutputService instanceof GrpcRemoteOutputService) {
+        ((GrpcRemoteOutputService) remoteOutputService).sendFinalizeBuild();
+      }
+      remoteOutputService = null;
     }
   }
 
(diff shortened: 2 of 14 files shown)


EdSchouten:
Initially my plan was to try to get the protocol that Bazel will speak with the FUSE daemon be standardized as part of the remote-apis project. Mailing list discussion:

https://groups.google.com/g/remote-execution-apis/c/qOSWWwBLPzo

After discussions that took place during the monthly meeting, and after taking another look at some of the (remotely) similar protocols in this space, such as Buildgrid's local_casd protocol, I think it would make more sense to keep a protocol like this in the Bazel tree, so that the protocol remains minimal.

Cc @janakdr @bergsieker @ulfjack

janakdr:
cc @susinmotion @alexjski

rupertks:
cc @tetromino

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,475 @@
...
+    // as StartBuild() on the server is supposed to call
+    // FindMissingBlobs() for all remotely tracked files. This ensures
+    // that they cannot disappear during the build.
+  }
+
+  public ListenableFuture<Void> downloadActionResult(ActionResult actionResult) {
+    // Request that all outputs of the action are created. Do make sure
+    // to remove the "bazel-out/" prefix from all paths, as that part of
+    // the path is not managed by the remote output service.
+    BatchCreateRequest.Builder builder = BatchCreateRequest.newBuilder();
```
janakdr, line 447: Internally, we don't batch these requests. I think that part of the reason is that the internal workflow looks something like this: (1) optimistically ask the local daemon if the file with given digest is already known, and if it is, insert it at this file path; (2) if not known, issue a save call to the remote execution service for each such output; (3) on that save call's (asynchronous) completion do an insert. This avoids the expense of a remote RPC and save to backends if the file is already known locally, which is a fairly common case. I don't know if that complexity is needed for the initial implementation of the Bazel protocol.

The incentive to start with the more complex solution is for the proto definition here to essentially be the open-sourced internal version, allowing for more code sharing, but there may well be a lot of path dependence in how we got to where we are. Perhaps we can try to make something that shares as much as possible without burdening Bazel with premature complexity.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  rpc FinalizeBuild(FinalizeBuildRequest) returns (google.protobuf.Empty);
+}
+
+message CleanRequest {
+  // The output base identifier that was provided to
+  // StartBuildRequest.output_base_id whose data needs to be removed.
+  string output_base_id = 1;
+}
+
+message StartBuildRequest {
```
janakdr, line 81: Here is a redacted version of the internal message, for reference. If this seems compatible, we could try to open-source the relevant parts, although it may be a struggle to make copybara do what we need.

```
message StartBuildRequest {
  // The FUSE daemon workspace id for the workspace in which the build is performed.
  required string workspace = 1;
  // Root of the directory under which Bazel builds its entire output tree.
  required string blaze_exec_root = 2;

  // The uid and gid of the user performing the build.
  // The 'uid' must match the user making this request.
  required int32 uid = 3;
  required int32 gid = 4;

  // Root directory of the workspace in which Bazel is running
  optional string workspace_root = 7;

  // Build invocation ID in Bazel terminology
  optional string invocation_id = 9;
}

message StartBuildReply {
  // Build id will always be set if StartBuildRequest succeeds.
  required string build_id = 1;

  // Relative path of artifacts in this workspace that have been externally
  // modified since the last StartBuildReply for the same workspace (e.g.
  // "bazel-out/gcc-.../bin/some/file.o").
  optional StatusProto modified_path_status = 3;
  repeated string modified_path = 4;
}
```

janakdr (commented):
Looks cool!

meisterT:
cc @coeuvre

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,475 @@
...
+    currentBuildId = null;
+    currentOriginalOutputPath = null;
+    currentRelativeOutputPath = null;
+  }
+
+  @Override
+  public void finalizeAction(Action action, MetadataHandler metadataHandler)
+      throws IOException, EnvironmentalExecException, InterruptedException {
+    // TODO: Would this be the right place to call into the remote
+    // output service to check whether any I/O errors occurred? If so,
```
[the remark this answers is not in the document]
EdSchouten: The FUSE file system may be lazy loading, meaning that I/O errors, transient or permanent, may occur while a locally running build action attempts to access input files that were generated by a remote action. Right now these errors will get logged by the FUSE daemon, but the system call that triggered the error will return EIO.

Ideally, the FUSE daemon would record the errors instead of logging them, so that Bazel can fetch it after a locally running build action terminates. That way Bazel can display a rich error message or retry executing the action.

This is something that Buildbarn already implements at the worker level. I/O errors in the FUSE file system will get attached to the ExecuteResponse.

src/main/protobuf/remote_output_service.proto
EdSchouten: Nice! Thanks for sharing! A couple of questions I have:

- `StartBuildRequest.{uid,gid}`: This is likely sent across because the daemon that Google has developed internally is multi-user, right? In my case I'm building a single-user daemon, creating a cache directory and FUSE mount in the user's home directory. Both are valid approaches, I guess. I've had some good experiences using `SO_PEERCRED` (Linux ) / `LOCAL_PEERCRED` (BSD) in combination with grpc-go. This allows a gRPC server to obtain the PID, UID and GID of the calling process in an authenticated way. It may therefore be possible to implement a multi-user daemon without encoding this information explicitly, right?
- `StartBuildReply.modified_path`: (How) does this mechanism deal with FUSE daemon restarts/crashes/loss of data? For the protocol in this PR I made it so that the FUSE daemon also sends back the invocation of the previous build, which should allow Bazel to detect state inconsistencies.

I'm always a fan of open sourcing code, but I suppose it may be a good idea to let Bazel use its own protocol that is strongly based on REv2. Being able to support REv2 instance names and use the Digest{,Function}, Output{File,Directory,Symlink} messages seems nice. What are your thoughts?

jin:
@lberki if needed, please reassign to someone that could shepherd this PR into Google?

src/main/protobuf/remote_output_service.proto
janakdr: Re the modified path question, the "modified_path_status" field says whether the modified_path list can be trusted. If it says no, we assume ModifiedFileSet.EVERYTHING_MODIFIED and fall back to manual stats.

Agreed that it may make sense to have this protocol reflect Bazel's feature set. The code that could be shared is fairly minimal, but I think it'll be nice to keep things as parallel as possible, both for future readers and for any convergence efforts. I must admit that I'm not sure how layering these protos would work, so actual sharing may be more trouble than it's worth.

janakdr (commented):
@jin we may need to get a few more eyes to give general approval of the idea. When it comes time for a review, @alexjski should be one of the reviewers.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: For the FUSE daemon that I created, BatchCreate() is always a cheap call. It never loads any objects from the CAS. It just creates placeholder files in the file system that are backed by CAS objects. Only when opening the file and reading one or more bytes of data, it will fault in the file from the remote execution service and store it in the local system's cache. This means that it's 'valid' to create files that use a Digest that doesn't correspond to an actual CAS object. There is no need to call to the remote execution service to save the file. REv2 also doesn't provide any facilities for that, right?

That said, if someone wants to write an implementation that explicitly calls into the remote execution service to save files, then there is nothing that prevents that, right? It's just that it's the responsibility of the FUSE daemon. It can do that as part of its BatchCreate() implementation.

src/main/protobuf/remote_output_service.proto
EdSchouten: That part makes complete sense, but what I want to aim at specifically is that the protocol that I came up with also sends back the UUID of the previous build. In my opinion this should be a requirement to make this robust, right?

Google's development environment is pretty much streamlined: folks just run Blaze on their gLinux workstations where all of this stuff is readily deployed. In the case of Bazel, people are most likely going to do things that are a bit crazy: run the FUSE daemon in one container, and run Bazel in another one. Those containers may be permanent, they may be short lived, etc.

The negotiation process between the FUSE daemon and Bazel on whether to reuse existing output path state is incomplete if the inputs are:

- blaze_exec_root
- invocation_id

And the outputs are merely:

- modified_path_status
- modified_path

Because what will happen if someone accidentally runs two containerized copies of Bazel against a single FUSE daemon that happen to use the same (virtualized!) blaze_exec_root? The FUSE daemon could respond 'modified_path_status == OK', even though the previous build could have been triggered by another instance of Bazel. By letting the FUSE daemon return the previous invocation ID to Bazel, Bazel can detect these cases and ignore the FUSE daemon's state if needed.

EdSchouten: Furthermore, the protocol that you pasted above lets the FUSE daemon return a build_id, which I suppose is some UUID/identifier generated by the FUSE daemon that Bazel must attach to Google's internal equivalent of BatchStat() / BatchCreate(), right?

The protocol in this PR differs from that in that it doesn't let the FUSE daemon generate any identifier. The invocation ID provided by Bazel is reused for that. Do we think that's acceptable?

janakdr: I'm not super familiar with the build id returned, and I'll probably bow out and let @alexjski investigate the Google side soon, but from what I can see, the "build id" is actually a directory on the filesystem that the FUSE daemon controls, and so it's only used for setting up the symlink properly. I think that corresponds to "output_path_prefix" in this code.

I agree that in the scenario where multiple Bazels are writing to the same execroot, things can get crazy. I'm not sure, though, that anything is guaranteed in that scenario, since I *think* that means the two Bazel instances are sharing the same output base, which seems like it won't end well for anybody. Having the previous id in the protocol seems like a nice defensive measure, though, agreed.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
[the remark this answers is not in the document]
janakdr: That's a very interesting idea, thanks for explaining it.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
janakdr: I'm not advocating for the control flow that Blaze uses, but to clarify, it's Blaze's responsibility to issue that remote RPC to save the outputs: it is already in communication with the remote service, and at present, I'm not sure that the local insertion call has enough data for the FUSE daemon to issue the proper request. I know nothing about REv2, so if it has no facility to explicitly save outputs, then I agree that Bazel is unlikely to be pushed in the direction of individual inserts versus this batch approach.

EdSchouten:
I've just rebased this change on top of latest master, and made CI happy once again. Earlier today I released my implementation of the FUSE daemon on GitHub as well:

https://github.com/buildbarn/bb-clientd

:tada: :tada: :tada:

Be sure to let me know what I can do from my end to get this change polished up, upstreamed, etc. etc. etc.

janakdr:
@alexjski could you take the lead on this? Thanks!

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,475 @@
...
+  @Nullable
+  public FileSystem createActionFileSystem(
+      FileSystem sourceDelegate,
+      PathFragment execRootFragment,
+      String relativeOutputPath,
+      ImmutableList<Root> sourceRoots,
+      ActionInputMap inputArtifactData,
+      Iterable<Artifact> outputArtifacts,
+      boolean trackFailedRemoteReads) {
+    return new GrpcFileSystem(sourceDelegate, execRootFragment.getRelative(relativeOutputPath));
```
EdSchouten, line 431: This only takes care of the case where files are accessed as part of actions. Another case we need to tackle is BES uploads. Now I tried to add these functions:
```
diff --git a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
index 479e3b5a05..a4c67d45d0 100644
--- a/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
@@ -22,6 +22,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.devtools.build.lib.actions.Action;
 import com.google.devtools.build.lib.actions.ActionInputMap;
 import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.actions.ArtifactPathResolver;
 import com.google.devtools.build.lib.actions.BuildFailedException;
 import com.google.devtools.build.lib.actions.cache.MetadataHandler;
 import com.google.devtools.build.lib.actions.EnvironmentalExecException;
@@ -513,4 +514,23 @@ public class GrpcRemoteOutputService implements OutputService, ActionResultDownl
         (result) -> null,
         MoreExecutors.directExecutor());
   }
+
+  @Override
+  public boolean supportsPathResolverForArtifactValues() {
+    return true;
+  }
+
+  @Override
+  public ArtifactPathResolver createPathResolverForArtifactValues(
+      PathFragment execRoot,
+      String relativeOutputPath,
+      FileSystem fileSystem,
+      ImmutableList<Root> pathEntries,
+      ActionInputMap actionInputMap,
+      Map<Artifact, Collection<Artifact>> expandedArtifacts,
+      Map<Artifact, ImmutableList<FilesetOutputSymlink>> filesets) {
+    FileSystem remoteFileSystem =
+        new GrpcFileSystem(fileSystem, execRoot.getRelative(relativeOutputPath));
+    return ArtifactPathResolver.createPathResolver(remoteFileSystem, fileSystem.getPath(execRoot));
+  }
 }
```

But I've observed that this breaks in an interesting way: BES uploads may continue to run after OutputService.finalizeBuild() is called, meaning we can no longer call BatchStat().

Question: Is there some way to detect that the PathResolver returned by createPathResolverForArtifactValues() is released, or is there simply no such guarantee?

sohaibiftikhar:
This is amazing 🥳

lberki:
As a general comment, a design doc would be useful so that everyone can understand more easily what's going on. We even have a separate repository for those:

https://github.com/bazelbuild/proposals

Apart from that procedural note, my main worry is that this adds Yet Another Way for Bazel to handle the metadata pertaining to action outputs, of which we already have too many (objfsd, the internal equivalent of this FUSE service, ActionFS, the output tree under `bazel-out/`, the local action cache, the disk cache, etc. 

@philwo is probably the most competent person to comment on all these things. I personally don't feel confident saying anything about this change other than "pls no I don't need more abstractions", which is not exactly helpful. At the very least, it should be written down somewhere how this proposal interacts with these other metadata handling systems.

EdSchouten:
> As a general comment, a design doc would be useful so that everyone can understand more easily what's going on. We even have a separate repository for those:
> 
> https://github.com/bazelbuild/proposals

Done! https://github.com/bazelbuild/proposals/pull/211

lberki:
Thanks a lot :)

philwo:
My start into this year hasn't exactly been great unfortunately, so I have a backlog of things I have to go through first. I won't have time to check this out this week, but please ping me if I still haven't replied here by mid of next.

EdSchouten:
Friendly ping!

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // By default, Bazel sets this value to the MD5 sum of the absolute
+  // path of the workspace directory. This is generally sufficient,
+  // though a more complex scheme may necessary in case the file system
+  // namespace is virtualized.
+  string output_base_id = 1;
+
+  // A client-chosen value that uniquely identifies this build. This
+  // value must be provided to most other methods to ensure that
+  // operations are targeted against the right output path.
+  string build_id = 2;
```
glukasiknuro, line 100: Nit: maybe this could this be called invocation_id (probably more in sync with bazel terms).

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // The absolute path at which the remote output service exposes its
+  // output paths, as seen from the perspective of the client.
+  //
+  // This value needs to be provided by the client, because file system
+  // namespace virtualization may cause this directory to appear at a
+  // location that differs from the one used by the service.
+  //
+  // The purpose of this field is to ensure that the remote output
+  // service is capable of expanding symbolic links containing absolute
+  // paths.
```
glukasiknuro, line 122: Could you clarify whether it relates to `output_base_id` ? In particular, is there implicit assumption that all files created by ROS will be in output_path_prefix/output_base_id (or maybe, why requirement that output_base_id must be a valid filename?)

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // output_path_prefix, this option is used to ensure the remote output
+  // service is capable of expanding symbolic links.
+  //
+  // Map keys are absolute paths, while map values are paths that are
+  // relative to the output path.
+  map<string, string> output_path_aliases = 6;
+}
+
+message InitialOutputPathContents {
+  // The identifier of a finalized build whose results are stored in the
```
glukasiknuro, line 136: Nit: add "finalized previous build" - I was confused why this is needed here - like that is not the build_id from startrequest.

BTW. Could StartRequest contain previous_build_id? It could be more natural from client perspective to know which build was previous and craft return value in ROS on that (e.g. it could allow ROS to perform concurrent builds in a single workspace - not applicable for bazel, but maybe other clients, or perform clean if ROS stores just previous build and one from request does not match).

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+message CleanRequest {
+  // The output base identifier that was provided to
+  // StartBuildRequest.output_base_id whose data needs to be removed.
+  string output_base_id = 1;
+}
+
+message StartBuildRequest {
+  // A client-chosen value that uniquely identifies this output path.
+  // This value must be set to ensure that the remote output service is
+  // capable of managing builds concurrently.
```
glukasiknuro, line 84: ROS could manage builds concurrently even without it, based on build_id, right? I guess this is mostly optimization - so that subsequent builds for same output_base_id can be incremental?

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // changes made to the output path is too large to be expressed.
+  InitialOutputPathContents initial_output_path_contents = 1;
+
+  // A relative path that the client must append to
+  // StartBuildRequest.output_path_prefix to obtain the full path at
+  // which outputs of the build are stored.
+  //
+  // If the remote output service is incapable of storing the output of
+  // multiple builds, this string may be left empty.
+  string output_path_suffix = 2;
```
glukasiknuro, line 165: Nit: why not full path, like `output_path`? It will be clearer afterwards, e.g. in BatchCreateRequest that mentions output path.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+message InitialOutputPathContents {
+  // The identifier of a finalized build whose results are stored in the
+  // output path.
+  string build_id = 1;
+
+  // Paths that have been modified or removed since the build finalized.
+  //
+  // If the remote output service freezes the contents of the output
+  // path between builds, this field can be left empty.
+  repeated string modified_paths = 2;
```
glukasiknuro, line 144: Can this contain a directory? (And what does it mean in such case)?

src/main/protobuf/remote_output_service.proto
EdSchouten: This is explicitly called build_id, so that it is in sync with Bazel's terminology. Other protocols such as the Build Event Stream use build_id as well.

src/main/protobuf/remote_output_service.proto
EdSchouten: To be honest, I don't know. I'm hoping to get more clarifications on this from one of the Bazel developers. I'm literally passing these entries to Bazel's ModifiedFileSet. I don't even know whether Bazel expects these paths to be absolute, relative to the execroot, relative to the output base, etc.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -483,6 +483,28 @@ public RemoteOutputsStrategyConverter() {
...
+  public String remoteOutputService;
+
+  @Option(
+      name = "remote_output_service_output_path_prefix",
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      help =
+          "The path at which all output path directories are visible that are managed by the "
+              + "remote output service.")
```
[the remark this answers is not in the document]
EdSchouten: Responding bottom up:

The reason why this must be provided on the Bazel side is because this information may simply not be known on the Remote Output Service side. You may run bb_clientd in container A, and instances of Bazel in containers B to Z. The FUSE file system that is created by bb_clientd may not be placed at the same absolute path as seen from Bazel's perspective.

This flag is indeed required when `--remote_output_service` is provided. Let me put an explicit check for this in place.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // files from the file system manually.
+  rpc Clean(CleanRequest) returns (google.protobuf.Empty);
+
+  // Signal that a new build is about to start.
+  //
+  // The client uses this call to obtain a directory where outputs of
+  // the build may be stored, called the output path. Based on the
+  // parameters provided, the remote output service may provide an empty
+  // output path, or one that has contents from a previous build of the
+  // same project.
```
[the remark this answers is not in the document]
EdSchouten: s/project/workspace/

src/main/protobuf/remote_output_service.proto
EdSchouten: Letting this return a full path wouldn't buy us anything in terms of functionality, right?

By letting this be the suffix, it's more evident that the Remote Output Service will always return an output path that resides below the output_path_prefix. Letting this return an absolute path opens up the possibility that the Remote Output Service returns an output path that resides at a completely different location, which wouldn't be logical in my opinion.

src/main/protobuf/remote_output_service.proto
EdSchouten: Even though they both start with output_*, there is no correlation between the structure of `{output_base_id}`, and `{output_path_prefix, output_path_suffix}`. The former is just an identifier that Bazel can send to the Remote Output Service to identify the workspace being built, while the latter is a path returned by the Remote Output Service, telling Bazel where outputs may be stored.

A Remote Output Service may use `output_base_Id` as part of `output_path_suffix`, but there is no strong requirement for that.

src/main/protobuf/remote_output_service.proto
EdSchouten: Let me respond to this question, while also responding to a part of [this comment](https://github.com/bazelbuild/bazel/pull/12823#discussion_r588793838):

> it could allow ROS to perform concurrent builds in a single workspace - not applicable for bazel, but maybe other clients, or perform clean if ROS stores just previous build and one from request does not match.

When designing this protocol, I tried to make it so that implementations of the Remote Output Service may differ in terms of smartness. bb_clientd uses FUSE, so it could do all sorts of fancy virtualization tricks (snapshotting, copy-on-write, etc.). But it should still be possible to build a dumb daemon that does nothing more than offload the 'bottom half' of Bazel. With that I mean: it should be possible to develop a gRPC service that just manages a concrete (native) directory on your file system. The dumbest versions of such a daemon just download files when BatchCreate() is called, while others may use OS specific features to create placeholder files that 'trap' read requests, downloading the files on demand.

In situations like these, it isn't realistic to assume that you can do efficient snapshotting/cloning of build directories. It likely means you need to do recursive copying, which I don't think is acceptable. This is one of the reasons the `output_base_id` exists, and that at most one active build for an `output_base_id` may exist.

Eliminating `output_base_id` entirely also means that the Remote Output Service doesn't really know to which workspace build directories belong. It would make it hard (impossible?) to implement Clean(), right?

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,305 @@
...
+
+// The Remote Output Service may be used by users of the Remote
+// Execution API to construct a directory on the local system that
+// contains all output files of a build.
+//
+// Primitive implementations of this API may simply download files from
+// the Content Addressable Storage (CAS) and store them at their
+// designated location. Complex implementations may use a pseudo file
+// system (e.g., FUSE) to support deduplication, lazy loading and
+// snapshotting.
```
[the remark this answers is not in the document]
EdSchouten: Done! Linked to proposals and mailing list threads.

src/main/protobuf/remote_output_service.proto
EdSchouten: Regarding your first sentence: done!

Regarding the remainder: Right now the protocol provides no facilities for letting the client request what's in the output path. Bazel simply asks: "May I have the build directory for my workspace?" And the service responds, giving a description of what's in the output path right now.

I think what you're interested in is making it so that Bazel can place stronger demands on what's inside the output path: "I want you to initialize it with the results of build ${previous_build_id}". As explained below, I don't think we should aim for that just yet; we should still allow this to be a drop-in replacement for Bazel's existing bottom half.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -509,6 +513,32 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+    Preconditions.checkState(remoteOutputService == null, "remoteOutputService must be null");
+    if (remoteOptions.remoteOutputService != null) {
+      try {
+        ReferenceCountedChannel channel =
+            RemoteCacheClientFactory.createGrpcChannelPool(
+                channelFactory,
+                poolSize,
+                remoteOptions.remoteOutputService,
+                remoteOptions.remoteProxy,
+                authAndTlsOptions,
```
glukasiknuro, line 526: I believe this will mean the same auth and tls options will be used when contacting remote output service and remote execution service parts (like CAS?). Not sure whether this may break in some setups, like if someone will use tls to connect to CAS, then ROS will need to be deployed with proper server certificate as well so that client will trust ROS.

src/main/protobuf/remote_output_service.proto
EdSchouten: This would allow `output_base_id` to also have values `COM3`, `LPT1` and `AUX`, which are not valid filenames on Windows.

src/main/protobuf/remote_output_service.proto
EdSchouten: Probably, but I wasn't able to.

src/main/protobuf/remote_output_service.proto
EdSchouten: I know that extracting the UID using `{LOCAL,SO}_PEERCRED` works with grpc-go, both on Linux and macOS. I am using exactly this for some other project and it's working great.

If it's necessary to pass in authentication/identification information explicitly, I don't think there's really a need to extend the protocol to accommodate this. Couldn't users just use `--remote_header` for this, similar to how people can pass in authentication tokens?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
EdSchouten: At least for the bb_clientd use case this is exactly what's needed. bb_clientd's FUSE file system also needs to have credentials for the Remote Execution service, so that it can load files that are accessed through the file system. By letting Bazel send Remote Execution credentials to bb_clientd, it is able to reuse the credentials that are configured on the Bazel side.

(Furthermore, bb_clientd can act like a full REv2 proxy, so that a system-wide cache of blobs may be offered.)

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: It looks like I've managed to get integration with Build Event Streams working. In addition to afterCommand(), there is this commandComplete() method that we can use to safely call the FinalizeBuild() RPC.

src/main/protobuf/remote_output_service.proto
EdSchouten: Considering that I have first-hand experience using `LOCAL_PEERCRED` on macOS and `SO_PEERCRED` on Linux, and that `--remote_header` also exists, let me resolve this thread.

src/main/protobuf/remote_output_service.proto
EdSchouten: I've included parts of the phrasing that you proposed. I did not include the change to let output_base_id be "[a-zA-Z0-9]{1,64}", as that allows the use of names (e.g., AUX, COM1, LPT3) that Windows does not permit.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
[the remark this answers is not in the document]
EdSchouten: --remote_output_service without --remote_output_service_output_path_prefix will now fail.

EdSchouten:
Hi @alexjski, I've just rebased my PR and addressed most of the comments that have been provided so far. PTAL!

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+
+  // State that needs to be retained across invocations of Bazel. The
+  // current build ID is placed inside, because it ensures that cached
+  // instances of Path that use GrpcFileSystem always generate RPCs to
+  // the Remote Output Service using the current build ID; not the one
+  // under which they were created.
+  public static class StateAcrossBuilds {
+    private UUID currentBuildId;
+    private UUID previousBuildId;
```
alexjski, line 82: nit: nullable for both

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+
+  private final StateAcrossBuilds stateAcrossBuilds;
+  private final RemoteOutputServiceFutureStub futureStub;
+  private final RemoteOutputServiceBlockingStub blockingStub;
+  private final String outputBaseId;
+  private final PathFragment outputPathPrefix;
+  private final String instanceName;
+  private final DigestFunction.Value digestFunction;
+
+  private PathFragment currentOriginalOutputPath;
```
alexjski, line 93: This appears unused.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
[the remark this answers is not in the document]
alexjski: I am sorry, I think I may need a little more explanation on this one. To me, we are talking about 2 cases here:

1. Errors which happen at the time of running an action remotely when we inform the FUSE/NFS about incoming files. If that does anything which can fail, I would think that the API responses should be a place to report that.
2. Error which happen after the action is done and a dependency of it tries to read the files locally -- in this case, the regular FS-level errors should suffice (that I consider no different from user trying to run `less bazel-out/file` and that failing).

On the surface, that seems to me that we should never need this call for the purpose mentioned in the TODO, which is why I think I must be missing something. Would you mind adding a little more detail, please?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+        .modifyAll(
+            Iterables.transform(
+                initialContents.getModifiedPathsList(),
+                (p) -> currentRelativeOutputPath.getRelative(p)))
+        .build();
+  }
+
+  @Override
+  public void finalizeBuild(boolean buildSuccessful)
+      throws BuildFailedException, AbruptExitException, InterruptedException {
```
alexjski, line 178: nit: delete

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+  private class GrpcBatchStat implements BatchStat {
+    private abstract class DumbFileStatus implements FileStatusWithDigest {
+      @Override
+      public long getSize() throws IOException {
+        throw new RuntimeException("Not implemented");
+      }
+
+      @Override
+      public long getLastModifiedTime() throws IOException {
+        throw new RuntimeException("Not implemented");
```
alexjski, line 219: nit: UnsupportedOperationException

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+    // we should likely let createActionFileSystem() call into the
+    // remote output service to start capturing I/O errors.
+  }
+
+  private String fixupExecRootPath(PathFragment path) {
+    return path.relativeTo(currentRelativeOutputPath).toString();
+  }
+
+  private class GrpcBatchStat implements BatchStat {
+    private abstract class DumbFileStatus implements FileStatusWithDigest {
```
alexjski, line 211: That likely can (should) be static

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+      public boolean isDirectory() {
+        return false;
+      }
+
+      @Override
+      public boolean isSymbolicLink() {
+        return true;
+      }
+
+      @Override
```
alexjski, line 294: These can go to the base class. BTW you can consider giving the base class a boolean/enum indicating the type, so you don't need to override is subclasses.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+      public boolean isSpecialFile() {
+        return false;
+      }
+    }
+
+    public List<FileStatusWithDigest> batchStat(boolean includeDigest,
```
alexjski, line 300: This is a nice optimization. One thing I would suggest is to consider having that in a follow up PR, especially since there seem to be some TODOs -- this way we wouldn't need to block the main part of this change on getting all that done.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+        partition(
+            filter(symlinks.entrySet().iterator(), (symlink) -> symlink.getValue() != null),
+            1000);
+    boolean cleanPathPrefix = true;
+    while (symlinkBatchIterator.hasNext()) {
+      List<Map.Entry<PathFragment, PathFragment>> symlinksBatch = symlinkBatchIterator.next();
+      BatchCreateRequest.Builder builder = BatchCreateRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setPathPrefix(symlinkTreeRoot.toString());
+      builder.setCleanPathPrefix(cleanPathPrefix);
```
alexjski, line 374: What does `cleanPathPrefix` stand for? In here it is first batch, really.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+  private class GrpcFileSystem extends DelegateFileSystem {
+    private final PathFragment originalOutputPath;
+
+    public GrpcFileSystem(FileSystem sourceDelegate, PathFragment originalOutputPath) {
+      super(sourceDelegate);
+      this.originalOutputPath = originalOutputPath;
+    }
+
+    @Override
+    protected byte[] getFastDigest(PathFragment path) throws IOException {
```
alexjski, line 406: Is this necessary for the solution to work at all or is it just an optimization to avoid calls going through FUSE? I saw in the doc, you mentioned that xattrs may be problematic for certain NFS implementations -- I guess this may heavily depend on where we would use NFS for this and what the capabilities are. Would you mind expanding on this a little?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+      // Don't attempt to compute digests ourselves. Call into the
+      // remote output service to request the digest. The service may be
+      // able to return the digest from its bookkeeping, as opposed to
+      // reading the actual file contents.
+      BatchStatRequest.Builder builder = BatchStatRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
```
alexjski, line 412: Why is build id important for getting a digest from path?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
alexjski: I am also in the camp of not very knowledgeable about REv2. Does REv2 facilitate both permanent storage of outputs and execution? If so, how do we ensure that the file remains available in case I do `bazel build //something; bazel shutdown` and browse the FUSE with my favorite tool?

The optimistic insert Janak mentioned works as an optimization where if the output store already knows a digest, there is no need to store the same thing again -- we may as well reuse that. I guess if the FUSE/NFS merely registers the metadata on how to get the file from RE when user/local action requests it, then such optimization will not buy us anything here.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,527 @@
...
+    builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+    for (OutputFile file : actionResult.getOutputFilesList()) {
+      // As there is no guarantee that permissions on files created by
+      // the remote output service can be modified, make sure that all
+      // downloaded files are marked executable.
+      // TODO: Do we only want to do this selectively?
+      builder.addFilesBuilder()
+          .mergeFrom(file)
+          .setPath(fixupExecRootPath(PathFragment.create(file.getPath())))
+          .setIsExecutable(true);
```
alexjski, line 490: Do we ever insert a file which is not marked as executable? There is a request to have other modes supported, but we can maybe defer adding support for that until we get to #5588?

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -323,6 +325,32 @@ public void download(
       FileOutErr origOutErr,
       OutputFilesLocker outputFilesLocker)
       throws ExecException, IOException, InterruptedException {
+    if (actionResultDownloader != null) {
```
alexjski, line 328: Can you explain the purpose of that? I would think that eager downloading should not really interact with a lazy FUSE/NFS solution -- we should install all outputs there always and it's up to the user what they will choose to access (which would trigger a download).

src/main/java/com/google/devtools/build/lib/vfs/OutputService.java
```diff
@@ -141,9 +143,6 @@ void createSymlinkTree(Map<PathFragment, PathFragment> symlinks, PathFragment sy
    */
   void clean() throws ExecException, InterruptedException;
 
-  /** @return true iff the file actually lives on a remote server */
```
alexjski, line 144: Good call!

src/main/protobuf/remote_output_service.proto
alexjski: WatchServiceDiffAwareness [returns both](https://github.com/bazelbuild/bazel/blob/f6cce242420cad432797e0e2501475e821553af7/src/main/java/com/google/devtools/build/lib/skyframe/WatchServiceDiffAwareness.java#L249), we should probably do the same.

src/main/protobuf/remote_output_service.proto
alexjski: Sorry for how trivial this sounds, but I think an example value of `output_path_prefix` and `outptu_path_suffix` would be very helpful. Especially if that has something to do with where the FUSE/NFS will be mounted.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,314 @@
...
+  //
+  // If the remote output service is incapable of storing the output of
+  // multiple builds, this string may be left empty.
+  string output_path_suffix = 2;
+}
+
+message BatchCreateRequest {
+  // The identifier of the build. The remote output service uses this to
+  // determine which output path needs to be modified.
+  string build_id = 1;
```
alexjski, line 180: Is that indirection to get `output_base_id`?

alexjski (changes requested):
At a very high level, this looks pretty good. Couple action items here:

1. I left some questions here, some of them silly (sorry) -- if you could answer them, that would help me understand the idea better and make sure that we agree on the high-level design.
2. Once we agree that the high level design is right, I would suggest splitting that change to some smaller ones if possible, that would also make meeting request 3 easier.
3. I think this change deserves some more testing -- please don't spend time on that now, but I am just giving an early warning that once we fully agree on the protocol and the design, that is something I will request.

src/main/protobuf/remote_output_service.proto
alexjski: About the `build_id` in the response in the internal protocol -- that seems a little bit like abuse of the same term -- this has nothing to do with a Bazel build id, it is something that the output service manages and corresponds to a workspace. I agree with Janak that it is conceptually similar to the `output_path_prefix`.

moroten:
@EdSchouten, do you have a fresh rebase for this patch?

jgavris:
FWIW / if it is relevant, I am working on adding `setxaddr` support to the current `DiskCacheClient` for macOS / Linux native filesystems (using `unix_digest_hash_attribute_name`). 

I was wondering what the use cases were for `getxattr` without `setxattr`, I assume maybe that's covered somehow in the FUSE filesystem used by Google folks? I should wrap this up sometime in the next few days, but the WIP is here. 

https://github.com/bazelbuild/bazel/compare/master...ButterflyNetwork:setxattr?expand=1

alexjski:
> I was wondering what the use cases were for `getxattr` without `setxattr`, I assume maybe that's covered somehow in the FUSE filesystem used by Google folks? 

This is it exactly, the xattrs are a way for the FUSE system to communicate metadata about files with Bazel (example [here](https://github.com/bazelbuild/bazel/blob/47f491941b2137aa29f39a66c80327e6a7e4b1d7/src/main/java/com/google/devtools/build/lib/unix/UnixFileSystem.java#L419)). This is why we actually never need to set those -- what is you use for it?

jgavris:
@alexjski We don't use FUSE / sandboxfs currently, just APFS / EXT4. Could you point me to how bazel interaction with FUSE / sandboxfs knows in which `hashAttributeName` to store the digest? The only usages I could find for the startup option `unixDigestHashAttributeName` are for `getxattr`, which is why I started up this branch. Maybe that's internal / in blaze?

alexjski:
> @alexjski We don't use FUSE / sandboxfs currently, just APFS / EXT4. Could you point me to how bazel interaction with FUSE / sandboxfs knows in which `hashAttributeName` to store the digest? The only usages I could find for the startup option `unixDigestHashAttributeName` are for `getxattr`, which is why I started up this branch. Maybe that's internal / in blaze?

It is internal. It is a feature of our FUSE that it exposes this metadata as an xattr. In the example I showed, if the xattr is present, we would read it instead of digesting the file.

FUSE is really well positioned to expose information like that since it also knows about all of the writes. Bazel does not do that. In particular, setting xattrs by hand can lead to wrong digests if the file changes from unaccounted modifications:

```
$ echo hello /tmp/a.txt
$ setfattr -n user.digest -v 123 /tmp/a.txt
$ echo there > /tmp/a.txt
$ getfattr -n user.digest /tmp/a.txt
getfattr: Removing leading '/' from absolute path names
# file: tmp/a.txt
user.digest="123"
```

Ideally, we would at least clear/update such xattr on a change, but that is up to the FS to do that. I guess for Bazel execution, that should be less of a problem since we delete files before we run actions, therefore updates by locally executed actions (local/standalone strategy vs sandboxed) should be generally fine (other than ill-behaved ones e.g. writing over inputs, but those are problematic already since they potentially corrupt anything else reading the input). I guess that would put it at a similar risk level to running with `--noexperimental_check_output_files`.

jgavris:
@alexjski I see, thanks. So you have a particular attribute and configure your startup options there. Do you also assume SHA-256 for the hash? That's been the default for a while, but's still configurable in Bazel. Just curious about the coupling between Blaze and SandboxFS.

We'd assume that nobody touches the local disk cache (AC / CAS files) besides bazel, and we can run `--noexperimental_check_output_files` and get the full speed boost!

jgavris:
I'll see if I can get my PR up in the next day or two, I might ping you as a reviewer if that is recommended.

meisterT:
cc @coeuvre

brentleyjones:
@EdSchouten @coeuvre Is this something that we still want to land? Still seems important for BwtB.

leakingtapan:
Any updates to this PR? And which bazel version could potentially have this merged in?

For some context, I’ve been evaluating —remote_download_symlink_template to reduce disk consumption on the client machine that calls bazel. It looks from the design proposal in this PR the stated —remote_download_symlink_template has edge cases that isn’t solved by the flag (eg, dynamic linked binary) so is remote output service designed to cover those problems?

EdSchouten:
Eternal backlog; I don't have any time right now to get this polished up. What I did do, was update this branch/PR to be on top of Bazel 5.1.1.

Cc @joeljeske

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: As most of the of the TODOs have already been processed now, this functionality seems to work quite reliable now. Would you mind if immediately we aimed for getting it included?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: It causes the existing contents of the symlink tree directory to be removed. I have just added a comment for this.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: Sure thing! So there are multiple downsides to using extended attributes for this:

- Linux: As soon as your FUSE file system starts to announce the availability of the GETXATTR operation, it basically opens the floodgates for the kernel to call into this operation whenever it wants to do certain ACL-based permission checks. From a performance perspective, it's generally smarter to let GETXATTR return ENOSYS, and let Bazel send selective RPCs whenever it's desired to get the digest.

- macOS: Since this PR was initially published, bb\_clientd has gained support for NFSv4. The bb\_clientd virtual file system layer can now be exposed both over FUSE and NFSv4. Extended attributes are still problematic on NFSv4, as macOS only permits storing those in AppleDouble files (`._MyFile`) for the time being. The exact encoding of those files is not a public API, so we should not try to handcraft those.

- In general: when we use gRPC, we also have the ability to do fancy things like batching, which significantly reduces the number of context switches between Bazel and bb\_clientd. This is currently not implemented, but relying on extended attributes to convey this information already closes the door on that.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -814,14 +863,13 @@ public void executorInit(CommandEnvironment env, BuildRequest request, ExecutorB
               actionContextProvider.getRemoteCache(),
               env.getExecRoot());
       builder.setActionInputPrefetcher(actionInputFetcher);
-      remoteOutputService.setActionInputFetcher(actionInputFetcher);
+      ((RemoteOutputService) remoteOutputService).setActionInputFetcher(actionInputFetcher);
```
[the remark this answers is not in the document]
EdSchouten: Yes! This should be enforced now.

src/main/protobuf/remote_output_service.proto
EdSchouten: Yes. The build ID implies the output base ID. But a single output base ID may be associated with multiple build directories (due to snapshotting/copy-on-write semantics, if provided).

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
EdSchouten: > Can you explain the purpose of that? I would think that eager downloading should not really interact with a lazy FUSE/NFS solution -- we should install all outputs there always and it's up to the user what they will choose to access (which would trigger a download).

This is exactly what bb\_clientd does. So even though the thing is called `ActionResultDownloader` it doesn't actually trigger a download. Maybe a better name would be `ActionResultInstantiator` or something?

EdSchouten:
Hey! It's a miracle! I've finally found some time to rebase this change on top of `master`. I have also processed many of the comments left behind by @alexjski.

@alexjski Even though a lot of time has passed, could I persuade you to review it once more? Thanks!

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: I think this is actually necessary, right? The `buildSuccessful` value we capture gets passed to the `FinalizeBuild` RPC.

brentleyjones:
It would be _really nice_ if we could get this in for 6.0 🙏.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -499,6 +505,40 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
...
+      String outputBaseId = remoteOptions.remoteOutputServiceOutputBaseId;
+      if (Strings.isNullOrEmpty(outputBaseId)) {
+        outputBaseId = DigestUtil.hashCodeToString(md5().hashString(env.getWorkspace().toString(), UTF_8));
+      }
+      String outputPathPrefix = remoteOptions.remoteOutputServiceOutputPathPrefix;
+      if (Strings.isNullOrEmpty(outputPathPrefix)) {
+        throw createOptionsExitException(
+            "If --remote_output_service is specified, --remote_output_service_output_path_prefix must be set as well",
+            FailureDetails.RemoteOptions.Code.OUTPUT_SERVICE_WITHOUT_OUTPUT_PATH_REFIX);
+      }
```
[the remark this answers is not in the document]
EdSchouten: This should be fixed in https://github.com/bazelbuild/bazel/pull/16501

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: Moved to base class and made them static.

Considering that in the latest version of this PR, directories also have a property of their own (i.e., the last modified time), I think there is less of an incentive to add a boolean/enum to indicate the type. Feel free to reopen if I'm mistaken!

moroten:
What is the review status on this PR?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+  @Override
+  public void checkActionFileSystemForLostInputs(FileSystem actionFileSystem, Action action)
+      throws LostInputsActionExecutionException {
+    // There is no need to provide an implementation for this function,
+    // as StartBuild() on the server is supposed to call
+    // FindMissingBlobs() for all remotely tracked files. This ensures
+    // that they cannot disappear during the build.
+  }
+
+  public ListenableFuture<Void> downloadActionResult(ActionResult actionResult) {
```
alexjski, line 565: I think this name is either confusing or I am misunderstanding what is going on in here. In my head, I would think that this downloads nothing, but merely it informs the FUSE about the remote files to track, so I would consider naming it something like `injectRemote[Action]Outputs` or something if that's the case.

There is a separate call for prefetching of files for local execution, which I can imagine could internally mean a proactive download so that a later read from FUSE would not need any network traffic.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,320 @@
...
+  // which the build is being started.  This value must be set to ensure
+  // that the remote output service is capable of managing builds for
+  // distinct workspaces concurrently.
+  //
+  // This value must be a valid filename for the operating system on
+  // which the remote output service and client are being executed. This
+  // allows the remote output service to create one subdirectory per
+  // project that needs to be built.
+  //
+  // By default, Bazel sets this value to the MD5 sum of the absolute
```
alexjski, line 98: nit: Ignorant question: isn't MD5 out of fashion these days?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+                                                Iterable<PathFragment> paths)
+        throws IOException, InterruptedException {
+      // TODO: Do we need to partition the input, just like in
+      // createSymlinkTree(), or is input already guaranteed to be
+      // bounded in size?
+      BatchStatRequest.Builder builder = BatchStatRequest.newBuilder();
+      builder.setBuildId(stateAcrossBuilds.currentBuildId.toString());
+      builder.setIncludeFileDigest(includeDigest);
+      for (PathFragment path : paths) {
+        builder.addPaths(fixupExecRootPath(path));
```
alexjski, line 364: nit: I want to nitpick on that since we make the output service aware of the relativeOutputPath in startBuild, so it technically could know that already. There is an obvious statefulness of the API argument to be made, but I am still curious to your reasoning behind that.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,320 @@
...
+  string output_path_prefix = 5;
+
+  // A map of paths on the system that will become symbolic links
+  // pointing to locations inside the output path. Similar to
+  // output_path_prefix, this option is used to ensure the remote output
+  // service is capable of expanding symbolic links.
+  //
+  // Map keys are absolute paths, while map values are paths that are
+  // relative to the output path.
+  map<string, string> output_path_aliases = 6;
```
alexjski, line 142: What is the purpose of that the FUSE system? Especially given we fix up all the paths to not have the `bazel-out` in them? Why is that a map?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+  }
+
+  @Override
+  public void createSymlinkTree(Map<PathFragment, PathFragment> symlinks, PathFragment symlinkTreeRoot)
+      throws ExecException, InterruptedException {
+    // The provided set of symbolic links may be too large to provide to
+    // the remote output service at once. Partition the symbolic links
+    // in groups of 1000, so that BatchCreateRequest messages remain
+    // small enough.
+    UnmodifiableIterator<List<Map.Entry<PathFragment, PathFragment>>> symlinkBatchIterator =
```
alexjski, line 416: nit: You can use `Iterables::partition` to spare yourself dealing with the iterator and just use for-each.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
alexjski: Did you consider using a streaming API? In that case, cleanup on could would be implicit. Not sure what's easier, also not sure how large do those requests get.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+  @Override
+  public void clean() throws ExecException, InterruptedException {
+    CleanRequest.Builder builder = CleanRequest.newBuilder();
+    builder.setOutputBaseId(outputBaseId);
+    try {
+      channel.withChannelBlocking(
+          channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+              .clean(builder.build()));
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
```
alexjski, line 462: nit: That probably calls for some `ExecException`, does anything catch `UncheckedIOException` down the line?

src/main/protobuf/remote_output_service.proto
alexjski: This is really a question, not an ask to change it, but why would we not use `StartBuildResponse::output_path_suffix` here (that can also mean as part of the paths) instead?

I guess what I am really going after is that we make Bazel aware of these directories to later use the IDs anyway--I am wondering if we could possibly remove any of those concepts from the API.

src/main/protobuf/remote_output_service.proto
```diff
@@ -0,0 +1,320 @@
...
+
+  // Paths whose status needs to be obtained.
+  repeated string paths = 5;
+}
+
+message BatchStatResponse {
+  // The status response for each of the requested paths, using the same
+  // order as requested. This means that this list has the same length
+  // as BatchStatRequest.paths.
+  repeated StatResponse responses = 1;
```
alexjski, line 255: Why not make that FileStatus?

alexjski (changes requested):
Thank you for the updates to this change!

I added some comments, those are not really exhaustive, but hopefully touch on some important issues. I also added @coeuvre as a reviewer since his opinion is very important here.

1. I am happy to continue giving you general feedback on how to best architect this feature, however this significantly affects remote execution which I do not own. As such will require an approval from @coeuvre--I would recommend against spending a lot of time on this until we have a clear confirmation that this is something we want to support for remote execution in Bazel.
2. API discussion. Sorry for my ignorance/paging things out of my memory. I added some questions around how the API is structured, I suspect that @coeuvre will have more useful feedback there.
3. Tests. That will require adding tests, I found none in this change. Obviously, please don't do that until 1 and 2 are cleared, but I will definitely ask for that.
4. Code shepherding--I some amount of code hygiene comments to be made (didn't add them all). This is just a warning that this will come, I assume you were waiting to resolve bigger issues before we go there. Anyway, I would recommend going over the warnings e.g. Intellij adds automatically for this change at some point.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+    stateAcrossBuilds.currentBuildId = buildId;
+    currentRelativeOutputPath = PathFragment.create(relativeOutputPath);
+
+    if (stateAcrossBuilds.previousBuildId == null || !response.hasInitialOutputPathContents()) {
+      // Either Bazel or the remote output service has performed no
+      // build before.
+      return ModifiedFileSet.EVERYTHING_MODIFIED;
+    }
+    InitialOutputPathContents initialContents = response.getInitialOutputPathContents();
+    if (initialContents.getBuildId() != stateAcrossBuilds.previousBuildId.toString()) {
```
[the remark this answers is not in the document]
EdSchouten: No. This is just a Go programmer trying to write Java. Will change that to `.equals()`.

src/main/protobuf/remote_output_service.proto
EdSchouten: Yeah, MD5 is a bit old. But I guess that for a use case like this it's all right. We're just hashing short pathnames. SHA-256 would work, but would yield longer pathnames.

I'm merely using MD5 here to remain consistent with what Bazel already does for the output base directory.

src/main/protobuf/remote_output_service.proto
EdSchouten: The FileStatus that's embedded into the StatResponse is conditionally set based on whether the file exists. So if the file does not exists, the StatResponse message is empty.

As far as I know, repeated fields cannot contain null elements. So we need this additional level of indirection to distinguish between existing/non-existing files.

src/main/protobuf/remote_output_service.proto
EdSchouten: The `BatchStat()` call has a `follow_symlinks` option. This causes bb_clientd to expand symlinks, returning the stat information for the  file it targets. This is somewhat complex, as Bazel (or `no-sandbox` build actions that it spawns) may create symlinks inside the FUSE file system that have absolute targets.

- `output_path_prefix` allows Bazel to specify at which absolute path the FUSE file system is placed.
- `output_path_aliases` allows Bazel to specify additional absolute path prefixes that end up resolving to locations inside the FUSE file system.

With these two pieces of knowledge, bb_clientd is capable of expanding absolute symlinks, reliably answering whether the symlink points to a location inside/outside the FUSE file system.

`output_path_aliases` is a map, where:
- the key corresponds to an absolute path **outside** of the FUSE file system
- the value corresponds to a relative path **inside** the FUSE file system.

src/main/protobuf/remote_output_service.proto
EdSchouten: This is done for a somewhat defensive reason: it allows bb_clientd to detect accidental concurrent reuse of the same `output_path_suffix`. For example, in an environment where bb_clientd is run once, but multiple copies of Bazel are launched in separate containers, it would be harmful if each of those copies tried to start a build with the same `output_path_suffix`.

By letting all `BatchCreate()`, `BatchStat()` and `FinalizeBuild()` calls take the build ID, bb_clientd is capable of detecting this misuse. It can ensure that at most one build is permitted per `output_path_suffix`, rejecting calls that provide the wrong value.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: At least in the case of bb_clientd, some piece of code needs to convert the paths from absolute to relative to the output path. The reason being that bb_clientd's virtual file system doesn't store files under their absolute names.

I decided to let BatchStat() and BatchCreate() take relative paths, both to make requests smaller, and to reduce ambiguity in the API. These methods can only take paths that reside underneath the output base. For example, you can't call BatchCreate() to construct a file under "/etc" or something.

I think this is also a good thing to do, because the `build.bazel.remote.execution.v2.Output{File,Symlink,Directory}` messages that we embed also contain relative paths.

EdSchouten:
Hey @alexjski,

> 4. Code shepherding--I some amount of code hygiene comments to be made (didn't add them all). This is just a warning that this will come, I assume you were waiting to resolve bigger issues before we go there. Anyway, I would recommend going over the warnings e.g. Intellij adds automatically for this change at some point.

Yeah, that's to be expected. I'm not really a Java programmer, so I'm not really up to date on what the recommended set of tooling is. Out of curiosity, is there some kind of lint target/script in the Bazel repository that I can run to print such warnings, as opposed to installing IntelliJ?

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
EdSchouten: I have to confess I only added these UncheckedIOExceptions, because I wasn't sure what exception type ought to be returned here.

- It looks like createSymlinkTree() and clean() can already throw ExecException. Are you saying that ExecException is the right type to use here?
- startBuild() and sendFinalizeBuild() don't throw ExecException, but I guess that's also intended, because they don't run as part of executing a single action/target. Would BuildFailedException be the right exception type here?

alexjski: For `startBuild`, it throws `AbruptExitException` which is most likely what you want/what we generally use for such failures outside of execution. Probably the same would work for `sendFinalizeBuild`.

alexjski:
> Hey @alexjski,
> 
> > 4. Code shepherding--I some amount of code hygiene comments to be made (didn't add them all). This is just a warning that this will come, I assume you were waiting to resolve bigger issues before we go there. Anyway, I would recommend going over the warnings e.g. Intellij adds automatically for this change at some point.
> 
> Yeah, that's to be expected. I'm not really a Java programmer, so I'm not really up to date on what the recommended set of tooling is. Out of curiosity, is there some kind of lint target/script in the Bazel repository that I can run to print such warnings, as opposed to installing IntelliJ?

Great question. I assumed that you are already using Intellij hence my suggestion. Let me add @coeuvre who has more experience with Bazel tooling, maybe he knows.

src/main/protobuf/remote_output_service.proto
alexjski: Another option is to use Fingerprint for it (you can even just get a long from it). Anyway, just mentioning this option exists since MD5 is a known standard so may be a much better fit in a protocol like this.

src/main/protobuf/remote_output_service.proto
alexjski: This is a little surprising that the FS needs to know where it is mounted to stat, technically it could just delegate to `stat` (as opposed to 'lstat') and get the info. Is that an optimization?

src/main/protobuf/remote_output_service.proto
alexjski: Is the `path_prefix` kind of a per-blaze workspace? That could be a better way to name that maybe. Also, some example of all of these paths would be likely very welcome.

alexjski (changes requested):

jmmv:
I have observed a problem when using this patch on Bazel 6.0.0 and it goes like this:

1. Do a bazel build with bb-clientd enabled by passing `--remote_output_service_output_path_prefix`.
2. Restart the bazel server (this may not be involved at all but is part of the workflow).
3. Do a second bazel without bb-clientd but with BwtB via `--remote_download_minimal` or `--remote_download_toplevel`.
4. Observe via a network traffic monitor that Bazel downloads *all* outputs along the way, and for a sufficiently large build, gets stuck in "checking cached actions" for possibly many minutes.

I ran jstack and noticed that Bazel was digesting all files in the output tree by reading their contents, which explains the massive data transfers and slowness.

The problem is that Bazel is *not* resetting its knowledge of where the output tree lives. When switching from bb-clientd to a local output tree using BwtB, the output tree is left pointing at bb-clientd. So... Bazel ends up doing regular file operations on that tree to digest files, which then cause bb-clientd to download everything.

When *not* using BwtB, Bazel seems to properly reset the output tree location and things work as expected. I suspect that `bazel-out` is only updated on first write, and because BwtB do not perform writes until the very end, all intermediate operations are going to the wrong place.

Doing an `rm $(readlink bazel-out)` is sufficient to clear the problem.

I do not know if this is a bug in this feature or in BwtB.

src/main/java/com/google/devtools/build/lib/remote/GrpcRemoteOutputService.java
```diff
@@ -0,0 +1,623 @@
...
+
+      try {
+        channel.withChannelBlocking(
+            channel -> RemoteOutputServiceGrpc.newBlockingStub(channel)
+                .finalizeBuild(builder.build()));
+        stateAcrossBuilds.previousBuildId = stateAcrossBuilds.currentBuildId;
+      } catch (InterruptedException e) {
+        stateAcrossBuilds.previousBuildId = null;
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
```
jmmv, line 207: We have seen cases where the bb-clientd daemon dies for whatever reason and Bazel then crashes due to this unchecked exception. This could probably be handled in a less drastic way (by printing the error somehow to tell the user about the problem and exiting with an error code instead of a crash).

stagnation:
We really want this, so I will try to rebase and take the work forward.

Gormo:
@stagnation I've uploaded a rebase for an earlier pre-7 release that you could have as a base: https://github.com/sbjorkle-qualcomm/bazel/commit/4be736936f8d071adcafb474f46e907688181a1b

coeuvre:
After discussed with @tjgq and @meisterT internally, we propose the following procedure to make it work in a world where BwoB is enabled by default:

1. Review and merge the `remote_output_service.proto` definition first. 
2. We will implement the Bazel side from scratch (instead of rebasing the old PR) so we only have one implementation of `OutputService`.
3. We expect the community to implement and maintain the server part (local daemon) outside of the Bazel source tree.
4. Eventually, we expect `remote_output_service.proto` to also move out of the Bazel source tree and be maintained by the community, similar to the REAPI spec today.

The story for output mode in Bazel will eventually become:
1. Build without the bytes (i.e., `--remote_download_toplevel`) by default. We will have decent IDE support but the limitation is that end users (human or IDEs) won't have the ability to browse the output tree.
2. We provide two options for end users to browse the output tree:
   1. Download all outputs.
   2. Use FUSE through the proto proposed by this PR.

We're currently working on making BwoB the default in Bazel 7 and expect to finish this work by EoQ, so we won't have the bandwidth to engage with this PR until early Q3. In the meanwhile, we would like to keep this PR open for more discussions.

stagnation:
Thank you, that sounds great!

EdSchouten:
Great! Be sure to rope me in if you folks want to discuss the protocol, as I'd be happy to explain the design decisions/tradeoffs that were made.

stagnation:
Hi, do you want us to submit a PR with the proto file, as a central location for early work with the file? Or do you have that in your pipeline?

coeuvre:
> Hi, do you want us to submit a PR with the proto file, as a central location for early work with the file?

SG. Please do!

coeuvre:
Quick update: @tjgq and I had a in-depth review for the protocol and we want to make some changes. I am working on a proof of concept naive server implementation (which will also be used in the integration tests) to make sure our changes make sense before we share the feedbacks.

moroten:
Bazel builds with this PR are available at https://github.com/meroton/bazel/releases

coeuvre:
Prototype is done. Working on the design doc and will start checking in the code soon.

EdSchouten:
At BazelCon you folks mentioned you wanted to make some protocol changes. Any chance we can already discuss these ahead of things getting merged?

coeuvre:
Yes, I will include the changes in the design doc and have it reviewed by the community before merging the code.

coeuvre:
The design doc is out! Please check https://github.com/bazelbuild/bazel/discussions/20933.

PR #12825, opened 2021-01-13
Move rules_proto and protocolbuffers/ubp to distdir_deps.bzl

aiuto:
- generate @rules_proto workspace inclusions in shell tests
- generate @rules_proto workspace inclusions in jdk.WORKSPACE
- src/test/java/com/google/devtools/build/lib/blackbox/framework/BlackBoxTestEnvironment.java will be done in a future CL. That will be the first case where we need to generate Java code. It requires some design up front.

Context: #12081


meteorcloudy (approved):
Thanks!

Issue #11674, opened 2020-06-30
Upgrade Jacoco to 0.8.5+ to fix Scala 2.12 lambda code coverage


gergelyfabian:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

When building Scala 2.12 with https://github.com/bazelbuild/rules_scala coverage has a bug (Scala lambdas are not properly detected). This has been tracked down to a bug in Jacoco 0.8.3 + Scala 2.12 (Jacoco did not properly recognize Scala 2.12 lambdas in 0.8.3).
Jacoco 0.8.5 contains a bugfix for this problem.

To fix this issue Jacoco would need to be upgraded from 0.8.3 to 0.8.5 in Bazel.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Reproduction with Bazel:

1. Check out rules_scala repository on master branch.
2. Generate coverage data:
`bazel coverage  --extra_toolchains="//scala:code_coverage_toolchain" //test/coverage/...`
3. Generated coverage report:
`genhtml -o ${destdir} --ignore-errors source bazel-out/k8-fastbuild/testlogs/test/coverage/test-all/coverage.dat`
4. In coverage/D1.scala.gcov.html only the first line is shown as covered

Outside of Bazel:

Download http://search.maven.org/remotecontent?filepath=org/jacoco/jacoco/0.8.5/jacoco-0.8.5.zip and http://search.maven.org/remotecontent?filepath=org/jacoco/jacoco/0.8.3/jacoco-0.8.3.zip.

Take https://github.com/bazelbuild/rules_scala/blob/master/test/coverage/D1.scala and add a main method to it:

```
  def main(args: Array[String]) {
    D1.veryLongFunctionNameIsHereAaaaaaaaa()
  }
```

Steps:

```
# Install Scala 2.12 into ~/opt/scala-2.12.10
# Unpack jacoco archive (in this case 0.8.5).
unzip ~/Downloads/jacoco-0.8.5.zip
mkdir classes
# Added a main method to test/coverage/D1.scala and moved it to src/ folder.
scalac src/D1.scala -d classes
[...]

PR #12829, opened 2021-01-14
Expose JacocoRunner from JavaToolchain to Starlark.

comius:
Issue https://github.com/bazelbuild/bazel/issues/11674


c-mita (approved):
LGTM

PR #12832, opened 2021-01-14
Introduce Apache Commons Lang3 3.8.1

lindner:


jin:
(See internal thread to bazel-engprod, from PR author)

philwo:
I verified that this dependency and version is available in Debian stable, too: https://packages.debian.org/buster/libcommons-lang3-java

Happy to import it.

Don't we need a `java_import` target in `third_party/BUILD`, too, though?

lindner:
I think we can discard this now that the commons lang code is being modified to use Guava instead.

lindner:
It might be needed for an upgrade of Apache Velocity, but that's a larger discussion.

philwo:
Yup, thank you!

Issue #12406, opened 2020-11-03
Support for setting per-file host copt (e.g. `--host_per_file_copt`)


GMNGeoffrey:
### Description of the problem / feature request:

Ability to set host only per-file copt. Basically the combination of per_file_copt and host_copt

### Feature requests: what underlying problem are you trying to solve with this feature?

Our repository depends on other repositories. We use per_file_copt to turn of clang diagnostics for these other repositories and then set `-Werror`, but this doesn't work in host mode because there's no way to do things per-file.

### What operating system are you running Bazel on?

Linux Debian

### What's the output of `bazel info release`?

release 3.3.1

### What's the output of `git remote get-url origin ; git rev-parse main ; git rev-parse HEAD` ?

$ git remote get-url origin ; git rev-parse main ; git rev-parse HEAD
git@github.com:GMNGeoffrey/iree.git
d13da6a832c3b75ea58129cff1936d39d2005610
361087f36af996632800d1bf9c4bdfa754a645ba

###  Have you found anything relevant by searching the web?

No

GMNGeoffrey:
Just for future people who are perhaps interested in this feature for the same reason, we ended up just turning off all host build diagnostics. We build everything we care about as a target as well, so we'll see the diagnostics there.

https://github.com/google/iree/blob/a15d08f3adba62b9541824fc55e5501711da08f5/build_tools/bazel/iree.bazelrc#L84

keith:
https://github.com/bazelbuild/bazel/pull/16618

UebelAndre:
I've opened a cherry-pick for the 6.0 release per the guidance here: https://github.com/bazelbuild/bazel/issues/16159#issuecomment-1290400293
- https://github.com/bazelbuild/bazel/pull/16695

PR #12833, opened 2021-01-15
Add host_per_file_copt and host_per_file_ltobackendopt build options

bjacklyn:
Implementation for https://github.com/bazelbuild/bazel/issues/12406


jin:
cc @oquenchil

alexkaratarakis:
Is there an issue that prevents merging this? I was about to open a new issue about this but found this PR instead.

oquenchil:
You can use different toolchains for target and host. You can have features enabled in one but not the other. Maybe this PR fits into the theme of improving the extensibility of crosstools. Any solution to #12406 should probably tackle it from this direction. I will close this PR.

PR #12838, opened 2021-01-15
Add JAR for truth proto extension

justinhorvitz:
Needed to enable testing for BEP test suite expansions.


katre (commented):
The change looks good. Is there a bug or other context for why this is needed?

justinhorvitz:
I have a pending change that's going to use the proto truth library in tests (PiperOrigin-RevId: 352039550)

katre (approved):

Issue #12778, opened 2021-01-06
bazel test //... on Linux is downloading JDKs for Windows, Mac Os and Linux ppc64le


davido:
Trying to run `bazel test //...` for Gerrit@Head with Bazel@HEAD (cfebb18ad206162c1309962e8d396a5fd0ef234e) JDKs for ferign platforms are fetched:

```
  $ uname -a
  Linux localhost.localdomain 5.10.3-1

  $  java -fullversion
  openjdk full version "1.8.0_275-b01"

  $ git clone --recurse-submodules https://gerrit.googlesource.com/gerrit
  $ cd gerrit

  $ bazel test //...
  [....]

Analyzing: 183 targets (65 packages loaded, 2195 targets configured)
    Fetching @remotejdk11_linux_s390x; fetching 4s
    Fetching @remote_java_tools_darwin; fetching
    Fetching @remote_java_tools_windows; fetching
    Fetching @remotejdk11_macos; fetching
    Fetching @remotejdk11_win; fetching
    Fetching @remotejdk11_linux_aarch64; fetching
    Fetching @remotejdk11_linux_ppc64le; fetching
    Fetching ..._s390x; Extracting /home/davido/.cache/bazel/_bazel_davido/27a001f4182820ef315d8d2d4f1edafe/external/\


Analyzing: 183 targets (66 packages loaded, 2195 targets configured)
    Fetching @remote_java_tools_darwin; fetching 17s
    Fetching @remote_java_tools_windows; fetching 17s
    Fetching @remotejdk11_macos; fetching 17s
    Fetching @remotejdk11_win; fetching 17s
    Fetching @remotejdk11_linux_aarch64; fetching 17s
    Fetching @remotejdk11_linux_ppc64le; fetching 17s
    Fetching ...pc64le; Extracting /home/davido/.cache/bazel/_bazel_davido/27a001f4182820ef315d8d2d4f1edafe/external/\
remotejdk11_linux_ppc64le/temp14401437873842827563/OpenJDK11U-jdk_ppc64le_linux_hotspot_11.0.7_10.tar.gz


Analyzing: 183 targets (67 packages loaded, 2195 targets configured)
    Fetching @remote_java_tools_darwin; fetching 24s
    Fetching @remote_java_tools_windows; fetching 24s
    Fetching @remotejdk11_macos; fetching 24s
    Fetching @remotejdk11_win; fetching 24s
    Fetching @remotejdk11_linux_aarch64; fetching 24s


[...]

PR #12846, opened 2021-01-18
Only download JDK repositories that are needed

comius:
The problem were `select`s that pointed to remote repos, which causes all the remote repos to be downloaded. Inserting an alias in between prevents this download.

Fixes https://github.com/bazelbuild/bazel/issues/12778

### DIFF
--- a/tools/jdk/BUILD.tools
+++ b/tools/jdk/BUILD.tools
@@ -419,18 +419,45 @@ py_test(
     ],
 )
 
+# Aliases for JDKs, so that they are only downloaded when needed. Do not use,
+# the aliases are only temporary.
+_JDKS = [
+    "remotejdk11_macos",
+    "remotejdk11_win",
+    "remotejdk11_linux_aarch64",
+    "remotejdk11_win",
+    "remotejdk11_linux",
+    "remotejdk11_linux_ppc64le",
+    "remotejdk11_linux_s390x",
+    "remotejdk14_macos",
+    "remotejdk14_win",
+    "remotejdk14_linux",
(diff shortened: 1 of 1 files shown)


Wyverald:
Sorry for randomly jumping in, but this seems rather surprising to me. I would've expected that with `select("true":"@A//:thing", "false":"@B//:thing")`, the repo B should not be fetched at all. Is there a valid reason why B is fetched in this case? If not, can we fix this spurious loading instead?

comius:
> Sorry for randomly jumping in, but this seems rather surprising to me. I would've expected that with `select("true":"@A//:thing", "false":"@B//:thing")`, the repo B should not be fetched at all. Is there a valid reason why B is fetched in this case? If not, can we fix this spurious loading instead?

I can confirm this is the behaviour, but I don't know the reason.
@katre ?

katre:
I believe this behavior is due to how a target is loaded: all of the labels referenced in a target need to be loaded (unless the attribute is declared as LABEL_NODEP).

An example of why this is needed is that in the example you gave, if `@B//:thing` didn't exist, the error would never be reported.

I may be mistaken, however, so I'll tag in @gregestren to see if that matches his understanding.

Wyverald:
> An example of why this is needed is that in the example you gave, if `@B//:thing` didn't exist, the error would never be reported.

I'd argue that's WAI. In fact, for part of the external deps overhaul, I was banking on the fact that B is allowed to not exist in this case -- imagine that B is a repo produced by a Windows executable only, and is completely missing when the user is running on other platforms.

PR #12847, opened 2021-01-18
Remove JDK14.

comius:
It's already old :)


c-mita (approved):
Pity, we hardly got to know him. ;)

PR #12851, opened 2021-01-19
Enlarge allowed values range for --connect_timeout_secs

nikola-sh:
Max allowed value for `connect_timeout_secs` is 120.
It is not enough sometimes. For example on a heavily loaded build conveyor connection procedure can last up to 300 seconds.


jin:
Increasing the timeout limit globally doesn't seem like the right solution here. It shouldn't take more than a few seconds to boot a cold server and connect to it. Anything more than a minute is already too slow.

nikola-sh:
A similar problem with other timeout option was fixed in PR #11391

janakdr (approved):

PR #12864, opened 2021-01-20
Add the truth liteproto extension JAR

justinhorvitz:


justinhorvitz:
Apparently the truth proto extension JAR added in https://github.com/bazelbuild/bazel/commit/cb9b63012da485204be6e923f492019cc8aa46cf doesn't work on its own.

philwo (approved):

PR #12867, opened 2021-01-21
Extend Bazel Java documentation.

comius:


site/docs/bazel-and-java.md
```diff
@@ -47,17 +51,34 @@ The resulting binaries are compatible with locally installed JVM in system
 libraries, which means the resulting binaries depend on what is installed on the
 machine.
 
+To configure the JVM used for execution and testing use `--java_runtime_version`
+flag. It's default value is `local_jdk`.
```
katre, line 55: Remove "It's" and just use "The default value is ...".

katre (approved):

site/docs/bazel-and-java.md
comius: Done.

cushon (approved):

site/docs/bazel-and-java.md
```diff
@@ -47,17 +51,34 @@ The resulting binaries are compatible with locally installed JVM in system
...
+### Hermetic Testing and Compilation
+
 To create a hermetic compile, you can use command line flag
 `--java_runtime_version=remotejdk_11`. The code is compiled for, executed, and
 tested on the JVM downloaded from a remote repository. For more details, see
 [Java runtime version flag](user-manual.html#flag--java_runtime_version).
 
+### Configuring Compilation and Execution of Build tools in Java
+
 There is a second pair of JDK and JVM used to build and execute tools, which are
```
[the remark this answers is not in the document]
comius: You mean ijar,singlejar, jacoco? Only the default values are set. They can be overridden.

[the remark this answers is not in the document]
cushon: I mean that a tool used in the build process like a Java tool in a genrule would use `--tool_java_runtime_version`, but don't Java-based tools in `java_toolchain` like JavaBuilder use `java_toolchain.java_runtime`, which is separate from `--tool_java_runtime_version`? Or are `java_toolchain.java_runtime` and `--tool_java_runtime_version` still connected.

site/docs/bazel-and-java.md
```diff
@@ -179,6 +202,9 @@ a `BUILD` file and registering it either by adding `register_toolchain` rule to
 the `WORKSPACE` file or by using
 [`--extra_toolchains`](user-manual.html#flag--extra_toolchains) flag.
```
cushon, line 203: I realize this predates your change and this doesn't need to be resolved now, but I think it would be valuable to have a more complete example here (or linked from here) that shows defining and registering a custom toolchain. I tried to remember how to set up a 'vanilla' toolchain recently and couldn't immediately remember how to register it, so having complete documentation for that might be helpful for end users.

comius: This should be as simple now as:
```
default_java_toolchain(
  name = "vanilla",
  configuration = VANILLA_TOOLCHAIN_CONFIGURATION,
  java_runtime = "@local_jdk//:local"
)
```
and 
`--extra_toolchains=//:vanilla_definition`

I didn't add a specific example for this, because everything is already there. Just need to put it together. Perhaps I can add it in a new PR.

site/docs/bazel-and-java.md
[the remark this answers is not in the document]
comius: I get it now. Yes, java_toolchain's tools execute using `java_toolchain.java_runtime`. I added a comment about this in the internal review, next to `java_runtime` attribute. I don't want to call it out in the use case for 'build tools`, because probably most of the users won't be modifying `java_toolchain`'s tools.

site/docs/bazel-and-java.md
cushon: Is `:vanilla_definition` guaranteed to 'win' over the other built-in toolchains if they're both registered? I thought it might be necessary to add a constraints that isn't present on the built-in ones.

comius: Short answer, yes if --java_language_version matches the one in the toolchain (this is in the document). 
The definition is only constrained by the version.
For toolchains order of registration is important: first flags, then WORKSPACE, then jdk.WORKSPACE suffix. So user defined stuff always comes before built-in toolchains.

PR #12870, opened 2021-01-21
Clarify test_suite behaviour in the Platforms docs

philsc:
The current `test_suite` behaviour in regards to incompatible target
skipping is subtle and not necessarily obvious. This patch aims to
clear up some of the confusion.

### DIFF
--- a/site/docs/platforms.md
+++ b/site/docs/platforms.md
@@ -156,6 +156,13 @@ $ bazel build --platforms=//:myplatform //...`
 $ bazel build --platforms=//:myplatform //:all`
 ```
 
+Incompatible tests in a [`test_suite`](be/general.html#test_suite) are
+similarly skipped if the `test_suite` is specified on the command line with
+[`--expand_test_suites`](command-line-reference.html#flag--expand_test_suites).
+In other words, `test_suite` targets on the command line behave like `:all` and
+`...`. Using `--noexpand_test_suites` prevents expansion and causes
+`test_suite` targets with incompatible tests to also be incompatible.
+
 Explicitly specifying an incompatible target on the command line results in an
 error message and a failed build.
 


site/docs/platforms.md
```diff
@@ -156,6 +156,13 @@ $ bazel build --platforms=//:myplatform //...`
 $ bazel build --platforms=//:myplatform //:all`
 ```
 
+Note that incompatible tests in a [`test_suite`](be/general.html#test_suite)
```
gregestren, line 159: Minor pedantic grammatical feedback: apply or not at your discretion:

I've heard guidance to try to avoid "note that" since what it communicates is generally clear enough from the rest of the sentence. i.e. just having the sentence implies it's worth noting.

I admit sometimes I still find it valuable to add "note that", so consider as appropriate.

site/docs/platforms.md
```diff
@@ -156,6 +156,13 @@ $ bazel build --platforms=//:myplatform //...`
...
+are similarly skipped if the `test_suite` is specified on the command line and
+[`--expand_test_suites`](command-line-reference.html#flag--expand_test_suites)
+is used. In other words, `test_suite` targets on the command line behave like
```
gregestren, line 162: Simplification suggestion: "on the command line with --expand_test_suites"?

gregestren (commented):
Nice catch! This is a good, and I think relevant, clarification.

site/docs/platforms.md
philsc: That's totally fair. I very much appreciate your insight here!

I applied your suggestion.

site/docs/platforms.md
philsc: That's a good idea, thanks.
I applied your suggestion.

gregestren (approved):
Thanks for considering such nitty suggestions!

PR #12877, opened 2021-01-22
Enforce visibility on select() keys

### LINKED ISSUE #12669: config_setting visibility not enforced
### Description of the problem / bug:

It should not be allowed to access a **private** config_setting from select in other package. But bazel fails to enforce the visibility.

pkgA/BUILD:
```
load(
    "@bazel_skylib//rules:common_settings.bzl",
    "bool_flag",
)

bool_flag(
    name = "myFlag",
    build_setting_default = False,
    # This visibility is working, however the config_setting is in same package.
    visibility = ["//visibility:private"]
)

config_setting(
    name = "mySetting",
    flag_values = {"myFlag": "True"},
    # This can be accessed from pkgB, despite private visibility
    visibility = ["//visibility:private"]
)
```

pkgB/BUILD:
```
genrule(
    name = "foo",
    srcs = [],
    outs = ["foo.out"],
    # Can access mySetting from another package, despite private visibility. 
    cmd = select({
        "//pkgA:mySetting": "echo running with mySetting > $@",
        "//conditions:default": "echo running default > $@"}),
)
```

I expect both of the following commands to fail, due to pkgB accessing private config_setting in pkgA, but no error is detected:

```
bbi_bazel build //pkgB:foo --//pkgA:myFlag=True
bbi_bazel build //pkgB:foo --//pkgA:myFlag=False
```

Linux

3.7.0

I searched, but have not found anything.

gregestren:
Example:

```
my_rule(
  name = "buildme",
  deps = select({
    "//other/package:some_config": [":mydeps"]
  }))
```

Today, `//other/package:some_config` is exempt from visibility checking, even though it's technically a target dep of 
 `buildme`.

While this dep is "special" vs. other deps in various ways, there's no obvious reason why it needs to be special in this way. It adds an unclear corner case exception to visibility's API.

### Implementation:
select() keys are not "normal" dependencies and don't generally follow the same code path. Hence them not being automatically visibility checked like others.

In particular, normal dependencies are found in `ConfiguredTargetFunction` and validity-checked in `RuleContext.Builder`. select() keys' only purpose is to figure out which other normal dependencies should exist. There's generally no need to pass them to `RuleContext.Builder`. Instead, Blaze passes their `ConfigMatchingProvider`s, which remain useful for analysis phase attribute lookups.

`RuleContext.Builder` needs a `ConfiguredTargetAndData` to do validity-checking.  This patch propagates that information for select() keys too.

We could alternatively refactor the validity checking logic. But that's an even more invasive change. Or do ad hoc validity checking directly in `ConfiguredTargetFunction`. But that's duplicating logic we really want to keep consolidated.

### Backward compatibility:
This would break existing builds if `config_setting` defaulted to private visibility. So this change specially defaults `config_setting` to public visibility, with clarifying documentation. When ready we'll want to create an incompatible change to make `config_setting` the same as everything else.

Fixes #12669.

RELNOTES: `config_setting` now honors `visibility` attribute (and defaults to `//visibility:public`)

### DIFF
--- a/site/docs/visibility.md
+++ b/site/docs/visibility.md
@@ -62,9 +62,12 @@ specified in the [`package`](be/functions.html#package) statement of the
 target's BUILD file. If there is no such `default_visibility` declaration, the
 visibility is `//visibility:private`.
 
-`config_setting` targets default to `//visibility:public`. This is purely for
+`config_setting` targets default to `//visibility:public, regardless of how the
+package's [`default_visibility`](be/functions.html#package.default_visibility)
+is set. This is purely for
 legacy reasons. Best practice is to treat `config_setting` targets as if they
-also use the private default.
+use the private default: any `config_setting` intended for use by other packages
+should set its visibility explicitly.
 
 ### Example
 
--- a/site/docs/visibility.md
+++ b/site/docs/visibility.md
@@ -62,6 +62,10 @@ specified in the [`package`](be/functions.html#package) statement of the
 target's BUILD file. If there is no such `default_visibility` declaration, the
 visibility is `//visibility:private`.
 
+`config_setting` targets default to `//visibility:public`. This is purely for
+legacy reasons. Best practice is to treat `config_setting` targets as if they
+also use the private default.
+
 ### Example
 
 File `//frobber/bin/BUILD`:
--- a/site/docs/visibility.md
+++ b/site/docs/visibility.md
@@ -64,10 +64,10 @@ visibility is `//visibility:private`.
 
 `config_setting` targets default to `//visibility:public, regardless of how the
 package's [`default_visibility`](be/functions.html#package.default_visibility)
-is set. This is purely for
-legacy reasons. Best practice is to treat `config_setting` targets as if they
-use the private default: any `config_setting` intended for use by other packages
-should set its visibility explicitly.
+is set. This is purely for legacy reasons. Best practice is to treat
+`config_setting` targets as if they use the private default: any
+`config_setting` intended for use by other packages should set its visibility
+explicitly.
 
 ### Example
 
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
@@ -1,0 +1,81 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package com.google.devtools.build.lib.analysis.config;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
+import com.google.devtools.build.lib.analysis.ConfiguredTarget;
+import com.google.devtools.build.lib.analysis.platform.ConstraintValueInfo;
+import com.google.devtools.build.lib.analysis.platform.PlatformInfo;
+import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.skyframe.ConfiguredTargetAndData;
+
+/**
+ * Utility class for temporarily tracking {@code select()} keys' {@link ConfigMatchingProvider}s and
+ * {@link ConfiguredTarget}s.
+ *
+ * <p>This is a utility class because its only purpose is to maintain {@link ConfiguredTarget} long
+ * enough for {@link RuleContext.Builder} to do prerequisite validation on it (like visibility
+ * checking).
+ *
+ * <p>Once {@link RuleContext} is instantiated, it should only have access to {@link
+ * ConfigMatchingProvider}, on the principle that providers are the correct interfaces for storing
+ * and sharing target metadata. {@link ConfiguredTarget} isn't meant to persist that long.
+ */
+@AutoValue
+public abstract class ConfigConditions {
+  public abstract ImmutableMap<Label, ConfiguredTargetAndData> asConfiguredTargets();
+
+  public abstract ImmutableMap<Label, ConfigMatchingProvider> asProviders();
+
+  public static ConfigConditions create(
+      ImmutableMap<Label, ConfiguredTargetAndData> asConfiguredTargets,
+      ImmutableMap<Label, ConfigMatchingProvider> asProviders) {
+    return new AutoValue_ConfigConditions(asConfiguredTargets, asProviders);
+  }
+
+  public static final ConfigConditions EMPTY =
+      ConfigConditions.create(ImmutableMap.of(), ImmutableMap.of());
+
+  /** Exception for when a {@code select()} has an invalid key (like wrong target type). */
+  public static class InvalidConditionException extends Exception {}
+
+  /**
+   * Returns a {@link ConfigMatchingProvider} from the given configured target if appropriate, else
+   * triggers a {@link InvalidConditionException}.
+   *
+   * <p>This is the canonical place to extract {@link ConfigMatchingProvider}s from configured
+   * targets. It's not as simple as {@link ConfiguredTarget#getProvider}.
+   */
+  public static ConfigMatchingProvider fromConfiguredTarget(
+      ConfiguredTargetAndData selectKey, PlatformInfo targetPlatform)
+      throws InvalidConditionException {
+    ConfiguredTarget selectable = selectKey.getConfiguredTarget();
+    // The below handles config_setting (which natively provides ConfigMatchingProvider) and
+    // constraint_value (which needs a custom-built ConfigMatchingProvider).
+    ConfigMatchingProvider matchingProvider = selectable.getProvider(ConfigMatchingProvider.class);
+    ConstraintValueInfo constraintValueInfo = selectable.get(ConstraintValueInfo.PROVIDER);
+    if (matchingProvider != null) {
+      return matchingProvider;
+    }
+    if (constraintValueInfo != null && targetPlatform != null) {
+      // If platformInfo == null, that means the owning target doesn't invoke toolchain
+      // resolution, in which case depending on a constraint_value is nonsensical.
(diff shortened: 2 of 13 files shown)


gregestren:
This PR is proof-of-concepty: I haven't added tests and would like to resolve some philosophical questions before advancing:

1. Should we strive for all rules supporting visibility checking? I don't think `select()` keys are the only rules that skip it. @katre - what about toolchain deps? 
1. This change highlights that `select()` deps follow a different code path than normal deps. That's why it's a 10-file change with an entirely new data class to support it. And that much more Bazel code to support. For just this maybe it's worth it. But if we had to repeat this ad hocness for X other rule types, maybe that suggests deeper design brittleness.
1. We also have a workaround: route through aliases.

katre:
1. The actual `toolchain` targets that are identified with `register_toolchains` and `--extra_toolchains` probably skip visibility, but the toolchains themselves (the `cc_toolchain` or `java_toolchain` targets that are consumed by rules) are visibility checked.
    a.  Also, visibility is skipped for platforms that are loaded via `register_execution_platforms`, `--extra_execution_platforms`, `--platforms`, and `--host_platforms`.
2. I'm not sure how we'd restructure configurable attributes to be more neatly supported. I think at this point the change makes sense.
3. The workaround requires everyone who uses a select to follow it, which seems onerous.

This PR LGTM.

gregestren:
Thanks for all the feedback, @katre . Added tests and some documentation.

gregestren:
Thanks for all the feedback, @katre . Added tests and some documentation.

site/docs/visibility.md
```diff
@@ -62,6 +62,10 @@ specified in the [`package`](be/functions.html#package) statement of the
 target's BUILD file. If there is no such `default_visibility` declaration, the
 visibility is `//visibility:private`.
 
+`config_setting` targets default to `//visibility:public`. This is purely for
+legacy reasons. Best practice is to treat `config_setting` targets as if they
```
katre, line 66: Clarify what happens if the pacage has default visibility which isn't public.
Clarify that best practice is to explicitly set visibility on the package, and override for any targets that require something different.

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.devtools.build.lib.analysis.platform.PlatformInfo;
+import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.skyframe.ConfiguredTargetAndData;
+
+/**
+ * Utility class for temporarily tracking {@code select()} keys' {@link ConfigMatchingProvider}s and
+ * {@link ConfiguredTarget}s.
+ *
+ * <p>This is a utility class because its only purpose is to maintain {@link ConfiguredTarget} long
+ * enough for {@link RuleContext.Builder} to do prerequisite validation on it (like visibility
```
katre, line 29: Change "like visibility checking" to "for example, visibility checks".

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
```diff
@@ -0,0 +1,81 @@
...
+  public static ConfigConditions create(
+      ImmutableMap<Label, ConfiguredTargetAndData> asConfiguredTargets,
+      ImmutableMap<Label, ConfigMatchingProvider> asProviders) {
+    return new AutoValue_ConfigConditions(asConfiguredTargets, asProviders);
+  }
+
+  public static final ConfigConditions EMPTY =
+      ConfigConditions.create(ImmutableMap.of(), ImmutableMap.of());
+
+  /** Exception for when a {@code select()} has an invalid key (like wrong target type). */
```
katre, line 51: s/like/for example,/

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
```diff
@@ -0,0 +1,81 @@
...
+   * targets. It's not as simple as {@link ConfiguredTarget#getProvider}.
+   */
+  public static ConfigMatchingProvider fromConfiguredTarget(
+      ConfiguredTargetAndData selectKey, PlatformInfo targetPlatform)
+      throws InvalidConditionException {
+    ConfiguredTarget selectable = selectKey.getConfiguredTarget();
+    // The below handles config_setting (which natively provides ConfigMatchingProvider) and
+    // constraint_value (which needs a custom-built ConfigMatchingProvider).
+    ConfigMatchingProvider matchingProvider = selectable.getProvider(ConfigMatchingProvider.class);
+    ConstraintValueInfo constraintValueInfo = selectable.get(ConstraintValueInfo.PROVIDER);
```
katre, line 68: Move this down to just before it is used.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -795,33 +791,23 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
       throw e;
     }
 
-    Map<Label, ConfigMatchingProvider> configConditions = new LinkedHashMap<>();
+    Map<Label, ConfiguredTargetAndData> asConfiguredTargets = new LinkedHashMap<>();
```
katre, line 794: Ideally this entire method, or at least the part from here on, could be moved into ConfigConditions.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -795,33 +791,23 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
...
+    Map<Label, ConfigMatchingProvider> asConfigConditions = new LinkedHashMap<>();
```
katre, line 795: Why use a LinkedHashMap here, and then turn it into an ImmutableMap later on?

src/test/java/com/google/devtools/build/lib/analysis/ConfigurableAttributesTest.java
```diff
@@ -1279,4 +1279,71 @@ public void multipleMatchErrorWhenAliasResolvesToSameSetting() throws Exception
             + " //a:foo\n"
             + " //a:alias_to_foo");
   }
+
+  @Test
+  public void defaultVisibilityConfigSetting() throws Exception {
+    // Production builds default to private visibility, but BuildViewTestCase defaults to public.
+    setPackageOptions("--default_visibility=private");
+    scratch.file(
+        "c/BUILD", "config_setting(", "    name = 'foo',", "    define_values = { 'foo': '1' })");
```
katre, line 1288: Since the formatter removed your newlines, make this one string.

lberki:
Yay, this is great! I mean, great modulo the necessary depot cleanup required. Do you have a sense of how much work that is? 

I think this is about the best approach you can take. I was somewhat surprised that you have to plumb through two maps in `ConfigConditions` (I would have guessed that the `ConfiguredTarget` is enough), but meh.

Do we have many other places that are exempt from visibility checking? I was hoping that centralizing it into `RuleContext` makes it catch every case, but I was apparently mistaken in at least one case.

site/docs/visibility.md
gregestren: Is that best practice? I'm not suggesting it's not... it's just not an issue I've ever considered. For example, is it preferable to explicitly set a package's visibility to private, even though that's a noop?

katre: Good point, I'm not sure we have clear best practices. I guess just clarify what the default behavior is, that this is diverging from?

katre (approved):

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
gregestren: Done.

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
gregestren: Done.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: Good call. Changed to ImmutableMap.Builder.

src/test/java/com/google/devtools/build/lib/analysis/ConfigurableAttributesTest.java
gregestren: Done.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: I support this idea, and tried to make headway with it. But it's too complicated for this PR, in my view. The problem is that the method has various dependencies on ConfiguredTargetFunction that aren't trivial to extract out. Trivially moving it would result in ConfiguredTargetFunction and ConfigConditions calling back and forth to each other.

I do support keeping ConfiguredTargetFunction (which is a beast) simple in any ways we can. So over a longer term I'd love to make progress on things like this.

site/docs/visibility.md
gregestren: Clarified difference from package defaults. Note that the default behavior for non-config setting rules is described in the immediately preceding paragraph.

src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java
gregestren: Done.

gregestren:
> I think this is about the best approach you can take. I was somewhat surprised that you have to plumb through two maps in `ConfigConditions` (I would have guessed that the `ConfiguredTarget` is enough), but meh.

Yeah, I wasn't thrilled about that. But it's really necessary, without doing deeper Bazel refactoring.

About that refactoring... One possible alternative is to include `config_setting`s in `RuleContext.prerequisiteMap`. I avoided that approach because it technically changes the semantics and I can't say with confidence that wouldn't break some other invariant. But I am intrigued. I'd like to submit this PR to get the functionality in, then try to whip up a version around that and share with you all to see if it ultimately looks viable.

> 
> Do we have many other places that are exempt from visibility checking? I was hoping that centralizing it into `RuleContext` makes it catch every case, but I was apparently mistaken in at least one case.

I believe the answer is near universally no. It's really only dependencies that get evaluated directly in `ConfiguredTargetFunction` through their own code paths that have this issue. That should only be this and some toolchain deps as described at https://github.com/bazelbuild/bazel/pull/12877#issuecomment-765408003.

PR #12883, opened 2021-01-22
Add additional_outputs argument to cc_common.link

torgil:
This allows additional linker output files to be available to custom
build rules, for example to produce the map file as following:

    map_file = ctx.actions.declare_file(ctx.label.name + ".map")
    linking_outputs = cc_common.link(
        additional_outputs = [map_file],
        user_link_flags = ["-Xlinker", "-Map=%s" % map_file.path],
    ...

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCcModule.java
@@ -158,6 +158,7 @@ public class BazelCcModule extends CcModule
       boolean linkDepsStatically,
       StarlarkInt stamp,
       Sequence<?> additionalInputs, // <Artifact> expected
+      Sequence<?> additionalOutputs, // <Artifact> expected
       Object grepIncludes,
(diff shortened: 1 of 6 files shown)


oquenchil:
What a coincidence, I have a change that does this. I called the parameter `linker_outputs` but I think your name is actually better. My change also restricted usage but since you need it I will change that and make it public. I will let you know when it's in.

oquenchil:
This went in here:
https://github.com/bazelbuild/bazel/commit/e42a740353744bd76793bfb358fb12371ea5edae

DhanalakshmiDurairaj:
Hi
Greetings to all,
I am working on GHS toolchain. Ccarm compiler generates “per object debug info file” by providing the flag -G and the extension of the file is .dbo. Debug info file for the archive is .dba. GCC toolchain generates per object debug info file with the extension of .dwo and .dwp (for archive) by providing the flag -gsplit-dwarf. Moreover, I am trying to produce .map file also. How to get these outputs (.dbo,.dba and .map)? Please guide me, is there any way to achieve this without using custom rules?

PR #12890, opened 2021-01-24
Add support for LLD in unix_cc_configure

stolyaroleh:
Unix cc autoconfiguration would only enable LTO support with the `gold` linker.
This commit enables `thin_lto` when using LLD - it's a drop-in replacement for GNU linkers.


hlopko:
Looks good to me. I would actually put lld to be the first, and only if it wasn't found look for gold. Would you approve of that @oquenchil ?

oquenchil:
> Looks good to me. I would actually put lld to be the first, and only if it wasn't found look for gold. Would you approve of that @oquenchil ?

Sounds good.

oquenchil (approved):
LGTM

Issue #10162, opened 2019-11-04
ijar fails on empty jars


agrieve:
Here's a repro:

Tested with ijar from bazel 1.1.0.

```sh
$ python -c "import zipfile; zipfile.ZipFile('test.jar', 'w')"
$ ijar test.jar out.jar
size 108 > estimated size 22
Aborted
```

lberki:
Somewhat embarrassing.

cushon:
Out of curiosity, was what real world situation did you encounter this in? Typically `jar` files always have at least `META-INF/MANIFEST.MF` file, especially the ones created by Bazel.

agrieve:
We encountered this in Chrome's build system, which doesn't bother with META-INF files for library targets (just zips up .class files).

PR #12893, opened 2021-01-25
Fix ijar failing on empty jars.

comius:
The problem happened when all the files are filtered out. In such case a dummy file is created, but it was not accounted for in the size estimation.
Fixes https://github.com/bazelbuild/bazel/issues/10162

### DIFF
--- a/third_party/ijar/ijar.cc
+++ b/third_party/ijar/ijar.cc
@@ -371,8 +371,8 @@ static void OpenFilesAndProcessJar(const char *file_out, const char *file_in,
     abort();
   }
   u8 output_length = in->CalculateOutputLength();
-  if (output_length < 103ull + DUMMY_PATH_LENGTH) {
-    output_length = 103ull + DUMMY_PATH_LENGTH;
+  if (output_length < 98ull + 2 * DUMMY_PATH_LENGTH) {
+    output_length = 98ull + 2 * DUMMY_PATH_LENGTH;
   }
   output_length += EstimateManifestOutputSize(target_label, injecting_rule_kind);
 
--- a/third_party/ijar/ijar.cc
+++ b/third_party/ijar/ijar.cc
@@ -370,9 +370,12 @@ static void OpenFilesAndProcessJar(const char *file_out, const char *file_in,
             strerror(errno));
     abort();
   }
-  u8 output_length =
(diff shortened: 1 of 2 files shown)


third_party/ijar/ijar.cc
```diff
@@ -368,9 +370,12 @@ static void OpenFilesAndProcessJar(const char *file_out, const char *file_in,
             strerror(errno));
     abort();
   }
-  u8 output_length =
-      in->CalculateOutputLength() +
-      EstimateManifestOutputSize(target_label, injecting_rule_kind);
+  u8 output_length = in->CalculateOutputLength();
+  if (output_length < 103ull + DUMMY_PATH_LENGTH) {
```
cushon, line 374: Where does `103` come from?

third_party/ijar/ijar.cc
```diff
@@ -368,9 +370,12 @@ static void OpenFilesAndProcessJar(const char *file_out, const char *file_in,
             strerror(errno));
     abort();
   }
-  u8 output_length =
-      in->CalculateOutputLength() +
```
cushon, line 372: I wonder of `CalculateOutputLength` should be handling this instead of fixing up the size later?

cushon (approved):
Consider adding a test?

I also commented in the bug--it's not clear if this is something that happened in a real build, it should be rare to see completely empty jar files that don't even contain a manifest.

ahmadasiri-24 (commented):
third_party/ijar/ijar.cc

third_party/ijar/ijar.cc
comius: > Consider adding a test?

Done

> I also commented in the bug--it's not clear if this is something that happened in a real build, it should be rare to see completely empty jar files that don't even contain a manifest.

True that.

comius: `CalculateOutputLength` is part of `zip.cc` and has no knowledge that we're adding `dummy` file.

third_party/ijar/ijar.cc
comius: From measurement, error message tells 108 bytes are needed. "dummy" appears twice in the output, so 98+2*strlen("dummy") should work even if somebody changes "dummy". (I fixed 103).

Somehow I cannot get the same number by adding together numbers used EstimateManifestOutputSize. And I consider reverse engineering zip code and figuring out header sizes a bit too much effort for this fix.

PR #12896, opened 2021-01-26
Remove wrapped_clang params files after use

keith:
We use TempFile's destructor to remove the params file. Because we
previously exec'd a new process, this was never called. Now we run
them as subprocesses so we can cleanup afterwards.


keith:
@oquenchil can you review?

oquenchil (approved):
LGTM

keith:
@oquenchil thanks, can you help land this?

PR #12901, opened 2021-01-26
Enable toolchain resolution for filegroup targets.

### LINKED ISSUE #12899: Investigate whether filegroup can use platform-conditional selects
Branched from https://github.com/bazelbuild/bazel/pull/12071:

Currently, the `filegroup` rule doesn't participate in toolchain resolution (see https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/common/BazelFilegroupRule.java;l=78), which means the target platform doesn't exist and so platform-conditional selects don't work.

As I recall, it led to a cycle in evaluation, because filegroups were needed at a low-enough level that we couldn't process loading platforms without them.

I'll investigate whether this is still an issue and whether this can be re-enabled.

katre:
Fixes #12899.


katre:
Ran downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1886

The only failure is also at [HEAD](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1885).

gregestren (approved):

Issue #11628, opened 2020-06-23
Support macOS Big Sur and Apple Silicon hardware platform


philwo:
This is a tracking bug for adding support for macOS Big Sur and the Apple Silicon platform to Bazel. I'm currently trying to acquire a DTK to run some initial tests and will follow up here when I find out more.

We aim to have full support in time for their final release.

keith:
This may be out of the scope of this discussion, but I think it would be nice to break the assumptions in bazel that if you're compiling for an Apple CPU you have to be running on `darwin_x86_64` (previously, and now `darwin*`). This would unlock cross-compilation to Apple platforms from Linux and I think could be done without degrading the UX for macOS users. This would also likely make it easier to add more CPUs like this one.

keith:
I also have a DTK coming and would be happy to contribute some here as this is a pre-req for us long term as well.

kayasoze:
My DTK has arrived and I've hit this roadblock.  Happy to help.

dmaclach:
Just to be clear to what roadblock are you referring?

dmaclach:
FYI 

https://github.com/bazelbuild/bazel/commit/b235517a1f3b91561a513ea4381365dba30dd592#diff-1f4964536f397e8679625954c7139bd6

and

https://github.com/bazelbuild/bazel/commit/e4e06c0293bda20ad8c2b8db131ce821316b8d12#diff-1f4964536f397e8679625954c7139bd6

add support for `darwin_arm64` and `darwin_arm64e` respectively. 

`blaze build --config=darwin_arm64` should get you a basic Apple Silicon binary.

dmaclach:
I **think** that #12035 should unblock folks. Apologies on my earlier assertions as I was using a different crosstool and forgot that my changes there were not available externally. We will leave this bug open for the time being until we have bazel working natively on Apple Silicon, but for the time being you should be able to build on x86_64 for arm64(e) and build (running an emulated bazel) on arm64 for arm64(e).

keith:
Related https://github.com/bazelbuild/rules_apple/issues/980

Any general update on this from google folks? Is anyone actively looking at this? I could probably spend some time on this soon, I have a DTK

BalestraPatrick:
We are also blocked on being able to target arm64 and the iOS simulator. I also have a DTK but I'm unfamiliar with Bazel core, but happy to help in any way I can.

keith:
[...]

PR #12918, opened 2021-01-28
Revert default cpu value on x86 macOS to "darwin"

meteorcloudy:
Many downstream projects are still depending on the default cpu value
on x86 macOS being "darwin" instead of "darwin_x86_64"

https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/412

Related: #11628

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/AutoCpuConverter.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/AutoCpuConverter.java
@@ -31,12 +31,12 @@ public class AutoCpuConverter implements Converter<String> {
   public String convert(String input) throws OptionsParsingException {
     if (input.isEmpty()) {
       // TODO(philwo) - replace these deprecated names with more logical ones (e.g. k8 becomes
-      // linux-x86_64 ...).
+      // linux-x86_64, darwin includes the CPU architecture, ...).
       switch (OS.getCurrent()) {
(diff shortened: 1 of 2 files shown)


meteorcloudy:
FYI @keith

comius (approved):

meisterT (approved):

keith:
Do you all think we should ever attempt this change again? I don't know if having this legacy value has any impact on platforms or anything that would make it worth trying to do it again?

meteorcloudy:
Rolling this forward again will of course make the cpu value more clear, but it's a breaking change, many config_settings that have `cpu = "darwin"` will break. And unfortunately this is again an incompatible change that's hard to migrate because users cannot use `cpu = "darwin_x86_64"` with current Bazel release version.

keith:
Yea definitely. Can you think of anything else this will block in the future? Otherwise sounds like we should leave it indefinitely and workaround the issues that come up

meteorcloudy:
> Can you think of anything else this will block in the future?

I don't think so, because even if we change the name to `darwin_x86_64`, we won't use the original name (`darwin`) for anything else. So it shouldn't block us. 

But ideally, I still want this to be fixed, maybe when less people use `cpu = <name>` in config_setting, I think that would happen when we enable platform based C++ toolchain resolution.

PR #12922, opened 2021-01-28
Simplify config_setting visibility checking

gregestren:
Make select() keys normal attribute -> configured target prerequisite
dependencies. This automatically opts them into RuleContext.Builder's standard
visibility checking.

The main historic argument against this is interference with "manual
trimming". But that's an on-ice feature that needs fundamental reevaluation and
shouldn't block other code in the meantime.

This is a simplified alternative to
https://github.com/bazelbuild/bazel/pull/12877, for https://github.com/bazelbuild/bazel/issues/12669.

PiperOrigin-RevId: 354326088
Change-Id: Ifaafc563f22d401599540ead70e49db6a86a1995

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
@@ -578,14 +578,7 @@ public abstract class DependencyResolver {
     Label ruleLabel = rule.getLabel();
     for (AttributeDependencyKind dependencyKind : getAttributes(rule, aspects)) {
       Attribute attribute = dependencyKind.getAttribute();
-      if (!attribute.getCondition().apply(attributeMap)
-          // Not only is resolving CONFIG_SETTING_DEPS_ATTRIBUTE deps here wasteful, since the only
-          // place they're used is in ConfiguredTargetFunction.getConfigConditions, but it actually
-          // breaks trimming as shown by
-          // FeatureFlagManualTrimmingTest#featureFlagInUnusedSelectBranchButNotInTransitiveConfigs_DoesNotError
-          // because it resolves a dep that trimming (correctly) doesn't account for because it's
-          // part of an unchosen select() branch.
-          || attribute.getName().equals(RuleClass.CONFIG_SETTING_DEPS_ATTRIBUTE)) {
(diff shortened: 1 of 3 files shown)


gregestren:
Following up on https://github.com/bazelbuild/bazel/pull/12877#issuecomment-768609861:

I think the sole real reason not to make `config_setting`s standard dependencies just like any others in `RuleContext` was because of interference with manual trimming.

But manual trimming is a half-rotted  feature with somewhere between 0 and very little usage. As long as that usage isn't a problem, the right answer IMO is to sacrifice the trimming logic and revisit it from the ground up when or if it deserves priority.

If this PR works, I can basically revert all the extra complication from my last PR. I've already verified it passes the new `config_setting` visibility checks.

gregestren:
To be clear, if I submit this, I'd further patch this to undo most of the other PR.

lberki:
How much complexity do we have in the code base to support manual trimming? It has been on ice since a while, so my estimate would be that we would be better off removing that code, then adding them again if manual trimming turns out to be necessary after all.

src/main/java/com/google/devtools/build/lib/analysis/DependencyResolver.java
```diff
@@ -578,14 +578,7 @@ private void resolveAttributes(
     Label ruleLabel = rule.getLabel();
     for (AttributeDependencyKind dependencyKind : getAttributes(rule, aspects)) {
       Attribute attribute = dependencyKind.getAttribute();
-      if (!attribute.getCondition().apply(attributeMap)
```
lberki, line 581: Removing the special-casing of this attribute warms my heart

gregestren:
> How much complexity do we have in the code base to support manual trimming? It has been on ice since a while, so my estimate would be that we would be better off removing that code, then adding them again if manual trimming turns out to be necessary after all.

 I basically agree. The main caveat is there may be projects that depend on this stuff now. There are definitely projects that set the tags that manual trimming reads. So it's not something we can just do in a day.

My hope would be that @sdtwigg could ultimately evaluate this in the context of more current ideas for trimming. There remain some promising options for some kind of manual-like trimming. But if no one is focusing on trimming at the moment that may be moot.

PR #12935, opened 2021-01-30
Allow cquery to filter out incompatible targets

### LINKED ISSUE #12917: The query language lacks a way to query rules that are target_compatible_with

The query language lacks a way to express "I want to query all targets that are compatible with a certain platform. This is because even though you can use attr() with cquery to ask for values of the attribute, thats not good enough as the attribute is transitive and dependencies might be incompatible, causing the target to still be incompatible even though it has no attribute itself.

We had to work around this by manually running over the proto output of cquery and checking the target_compatible_with attribute of a rule and its dependencies manually in Go code.

The query language should have a function that allows one to query if a rule is compatible, something along the lines of `compatible_with(<rule>, <constraint>)`, with `compatible_with(<rule>)` defaulting to `@local_config_platform//:host` as constraint.

I'm trying to avoid needing to manually run over the analysis proto manually, which is slow, error prone, easy to get wrong and also restricts my ability to meaningfully use other operators alongside

This is a feature request, any use of `target_compatible_with` in a dependency and then trying to cquery compatible rules in any way will stumble on this.

We use bazel on MacOS and Linux.

`release 4.0.0`

N/A

N/A

I've discussed this on the bazel slack and was told I should likely log a feature request

philsc:
This patch exposes the `IncompatiblePlatformProvider` to Starlark just
enough that it can be used with `cquery`'s `platforms()` function. I
added an example of this to the documentation. The motivation here is
to let users filter out incompatible targets from queries that
provide things like the list of targets to build for CI.

This patch is minimal on purpose. It does not allow users to
instantiate an `IncompatiblePlatformProvider` in Starlark. This may be
added in a future patch to address a different use case.

Fixes #12917.


site/docs/cquery.html
```diff
@@ -600,6 +600,22 @@ <h4>Examples</h4>
...
+
+  def format(target):
+    if "IncompatiblePlatformProvider" not in providers(target):
+      return target.label
+    return ""
+
+
+  $ bazel cquery //... --output=starlark --starlark:file=example.cquery
+</pre>
+
```
philsc, line 618: I could also move this into the Platforms page instead of here. I couldn't convince myself one way over the other.

gregestren: I'd move this to the platforms page, since `cquery` is just the tool that happens to serve the need.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/platform/IncompatiblePlatformProviderApi.java
```diff
@@ -0,0 +1,11 @@
+package com.google.devtools.build.lib.starlarkbuildapi.platform;
+
+import com.google.devtools.build.docgen.annot.DocCategory;
+import net.starlark.java.annot.StarlarkBuiltin;
+import net.starlark.java.eval.StarlarkValue;
+
+@StarlarkBuiltin(
+    name = "IncompatiblePlatformProvider",
+    doc = "An interface for targets that are incompatible with the target platform.",
```
gregestren, line 9: Can we include a link to appropriate documentation to properly contextualize this?

gregestren:
Added @comius as Build API lead. My understanding is they should vet changes to the Starlark build API. FYI my personal opinion is this is a pretty good candidate for being part of the API.

site/docs/cquery.html
philsc: Sounds good to me. Done.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/platform/IncompatiblePlatformProviderApi.java
philsc: Good idea. I linked back to platforms.html in this case.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/platform/IncompatiblePlatformProviderApi.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.google.devtools.build.docgen.annot.DocCategory;
+import net.starlark.java.annot.StarlarkBuiltin;
+import net.starlark.java.eval.StarlarkValue;
+
+@StarlarkBuiltin(
+    name = "IncompatiblePlatformProvider",
+    doc = "An interface for targets that are incompatible with the target platform. See "
+            + "<a href='../../platforms.html#detecting-incompatible-targets-using-bazel-cquery'>"
+            + "Detecting incompatible targets using <code>bazel query</code></a> for more "
```
gregestren, line 11: query or cquery?

gregestren (approved):
Overall this looks great to me!

I would hold off merging before getting @comius 's input. Depending on his schedule, feel free to ping this issue again to keep momentum up.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/platform/IncompatiblePlatformProviderApi.java
philsc: Oh right. Thanks!

philsc:
@comius, are you interested in taking a look at this PR?

src/main/java/com/google/devtools/build/lib/analysis/BUILD
```diff
@@ -798,6 +798,9 @@ java_library(
         ":configured_target",
         ":transitive_info_provider",
         "//src/main/java/com/google/devtools/build/lib/analysis/platform",
+        "//src/main/java/com/google/devtools/build/lib/starlarkbuildapi/platform",
```
comius, line 801: ?

comius (approved):
LGTM

gregestren:
I'll merge...

Issue #12920, opened 2021-01-28
Bazel CI: RBE builds broken


meteorcloudy:
rules_docker: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1889#8859d532-0524-4468-a0ad-310325e56399
rules_go: 
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1889#30ce89e0-766c-47a4-8daf-9b4827421fbe
rules_nodjes:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1889#54f552d8-9843-47be-aad2-a1025d28c1b9

All failing with similar build error:
```
(11:49:29) ERROR: /var/lib/buildkite-agent/builds/bk-docker-6srk/bazel-downstream-projects/rules_docker/tests/container/BUILD:775:16: GZIP tests/container/new_alpine_linux_ppc64le_image_oci_go_join_layers-layer.tar.gz failed: (Exit 34): UNAVAILABLE: HTTP/2 error code: NO_ERROR
Received Goaway
load_shed
java.io.IOException: io.grpc.StatusRuntimeException: UNAVAILABLE: HTTP/2 error code: NO_ERROR
Received Goaway
load_shed
	at com.google.devtools.build.lib.remote.GrpcCacheClient.lambda$handleStatus$5(GrpcCacheClient.java:244)
	at com.google.common.util.concurrent.AbstractCatchingFuture$AsyncCatchingFuture.doFallback(AbstractCatchingFuture.java:192)
	at com.google.common.util.concurrent.AbstractCatchingFuture$AsyncCatchingFuture.doFallback(AbstractCatchingFuture.java:179)
	at com.google.common.util.concurrent.AbstractCatchingFuture.run(AbstractCatchingFuture.java:124)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1174)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:969)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:760)
	at io.grpc.stub.ClientCalls$GrpcFuture.setException(ClientCalls.java:563)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:533)
[...]

Issue #12363, opened 2020-10-27
Bazel should not crash when receiving gRPC Goaway load_shed


philwo:
In https://github.com/bazelbuild/continuous-integration/issues/1055 a few users reported that Bazel crashes with a stack trace when we receive a "Goaway" or "load_shed" reply from the remote execution service.

These replies are sent by load balancers when they're approaching or have reached their load limit and should cause Bazel to reconnect (?), but certainly not crash.

ulfjack:
I've previously seen cases kinda like this, and Bazel was properly retrying up to its retry limit. I'm not sure that there's a bug here.

bergsieker:
This has been root caused to bad server behavior sending GOAWAY frames in a way that was out of spec. No code change is required in Bazel.

I think the error in bazelbuild/continuous-integration#1055 was mischaracterized: that's just the exception logged via --verbose_failures. Bazel did not crash and correctly terminated the build with an error.

philwo:
I'm not sure that Bazel should crash with an exception though for cases that it handles "correctly"?

IMHO exceptions indicate an unexpected condition and should not occur in expected cases - a user-friendly error message would be more appropriate then.

coeuvre:
> I'm not sure that Bazel should crash with an exception though for cases that it handles "correctly"?

Bazel didn't crash in that case (failed to read remote cache due to GOAWAY), instead, it prints `WARNING: Reading from Remote Cache:` with exception details (enabled by `--verbose_failures`) and fallbacks to execute the action.

philwo:
@coeuvre Thanks for the explanation!

PR #12939, opened 2021-02-01
Add --experimental_remote_grpc_max_channels to allow users manually configure number of gRPC channels

coeuvre:
For remote RPCs, gRPC limits the number of in-flight requests per connection to MAX_CONCURRENT_STREAMS. Exceeding it will cause RPCs to be queued. The problem is described at #11801 and we added a channel pool in #11937 to overcome this limitation.

However, the assumption we made for calculating the size of channel pool is not accurate. Bazel uploads/downloads inputs/outputs concurrently, which could lead to substantially more than `2 * jobs` RPCs in flight at any given time. Number of inputs/outputs various across actions and Bazel doesn't know up front what the max number of inputs/outputs of an action in the build is, so there's really no way to choose the number of connections intelligently.

Failed opening enough connections for a build could reduce the throughput and lead to `GOAWAY/load_shed` error in some cases.

This PR is a short-term fix which adds an experimental flag `--experimental_remote_grpc_max_channels` allowing users manually configure the number of gRPC channels.

For a long-term fix, we probably need to build a client-side RPCs queue and increase the size of channel pool dynamically.

Related issues: #12920, #12363
Internal discussion: b/178710657

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -321,6 +321,10 @@ public final class RemoteModule extends BlazeModule {
       // number of required channels is calculated as: ceil(jobs / 100).
       poolSize = (int) Math.ceil((double) buildRequestOptions.jobs / 100.0);
     }
+    if (remoteOptions.remoteGrpcMaxChannels > 0) {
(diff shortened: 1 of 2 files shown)


coeuvre:
Closing since we have implemented the dynamic channel pool de8f69dd9ef2e06d2809a2ae3eedc97810e0d181.

PR #12940, opened 2021-02-01
Add --experimental_repository_disable_download to allow users disable download for external repos

coeuvre:
This PR adds a flag `--experimental_repository_disable_download` which when set will prevent Bazel from downloading external repositories.  However, `local_repository`, `new_local_repository` and external repositories already downloaded in the repository cache  would still work.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
@@ -46,6 +46,7 @@ public class DownloadManager {
   private List<Path> distdir = ImmutableList.of();
   private UrlRewriter rewriter;
   private final Downloader downloader;
+  private boolean disableDownload = false;
 
   public DownloadManager(RepositoryCache repositoryCache, Downloader downloader) {
     this.repositoryCache = repositoryCache;
@@ -58,6 +59,10 @@ public class DownloadManager {
 
   public void setUrlRewriter(UrlRewriter rewriter) {
     this.rewriter = rewriter;
+  }
+
+  public void setDisableDownload(boolean disableDownload) {
+    this.disableDownload = disableDownload;
   }
 
   /**
@@ -192,6 +197,10 @@ public class DownloadManager {
           }
         }
       }
+    }
+
+    if (disableDownload) {
+      throw new IOException(String.format("Failed to download repo %s: download is disabled.", repo));
     }
 
     try {
--- a/src/test/shell/bazel/starlark_repository_test.sh
+++ b/src/test/shell/bazel/starlark_repository_test.sh
@@ -2099,7 +2099,6 @@ function test_disable_download_should_allow_distdir() {
   echo 'Hello World' > x/file.txt
   tar cvf x.tar x
   sha256=$(sha256sum x.tar | head -c 64)
-  serve_file x.tar
 
   mkdir main
   cp x.tar main
@@ -2108,7 +2107,7 @@ function test_disable_download_should_allow_distdir() {
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
   name="ext",
-  url = "http://127.0.0.1:$nc_port/x.tar",
+  url = "http://127.0.0.1/x.tar",
   sha256="$sha256",
 )
 EOF
--- a/src/test/shell/bazel/starlark_repository_test.sh
+++ b/src/test/shell/bazel/starlark_repository_test.sh
@@ -2061,4 +2061,93 @@ EOF
       || fail "Expected success despite needing a file behind basic auth"
 }
 
+function test_disable_download_should_prevent_downloading() {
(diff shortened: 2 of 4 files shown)


coeuvre:
cc @lberki

meteorcloudy:
Maybe add a simple test for this? ;)

lberki:
@jin , it looks like this is enough, isn't it?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
```diff
@@ -194,6 +199,10 @@ public Path download(
       }
     }
 
+    if (disableDownload) {
```
lberki, line 202: Do I understand correctly that this works because we don't provide any built-in repositories anymore that do downloading and we delegate everything to Starlark code instead?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
```diff
@@ -194,6 +199,10 @@ public Path download(
...
+      throw new IOException(String.format("Failed to download repo %s: download is disabled.", repo));
```
lberki, line 203: Please add test cases to verify at least the following cases (`starlark_repository_test` looks like a logical location, but maybe there is something better):

1. When the flag is set, downloads don't work
2. When the flag is set, repositories in distdir are still accessible
3. Maybe: when the flag is unset, downloads work (although that case is probably covered by existing tests?)

I'd love if this flag to also affected `git_repository` and the like, but I guess that's not possible unless we put repository download actions into a sandbox which is probably too big a change for now :(

jin:
For repo rules, this covers the `repository_ctx.download` calls, but not arbitrary `repository_ctx.execute` commands that can make network calls.

lberki:
@jin , can't disagree with that, but that's not any different from regular actions doing network calls, is it?

jin:
Indeed, and that's the third area where Bazel can make network calls today AFAICT. Let's get this in to cover this case.

lberki:
Yeah. This change alone is probably good enough and if not, we can think further. It's definitely necessary, but probably also necessary and sufficient.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
coeuvre: Test cases added.

`git_repository` uses `repository_ctx.execute` under the hood so it's impossible for this change to have affect on it.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
coeuvre: This only works for repo rule which uses `repository_ctx.download`.

src/test/shell/bazel/starlark_repository_test.sh
```diff
@@ -2061,4 +2061,93 @@ EOF
...
+genrule(
+  name = "it",
+  srcs = ["@ext//x:file.txt"],
+  outs = ["it.txt"],
+  cmd = "cp $< $@",
+)
+EOF
+
+  bazel build --experimental_repository_disable_download //:it > "${TEST_log}" 2>&1 \
+      && fail "Expected failure" || :
```
lberki, line 2092: What does `|| :` do?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
lberki: Yeah, I don't think there is a lot we can do about that :(

src/test/shell/bazel/starlark_repository_test.sh
```diff
@@ -2061,4 +2061,93 @@ EOF
...
+  expect_log "Failed to download repo ext: download is disabled"
+}
+
+function test_disable_download_should_allow_distdir() {
+  mkdir x
+  echo 'exports_files(["file.txt"])' > x/BUILD
+  echo 'Hello World' > x/file.txt
+  tar cvf x.tar x
+  sha256=$(sha256sum x.tar | head -c 64)
+  serve_file x.tar
```
lberki, line 2102: If this uses distdir, why do you need this `serve_file` call?

src/test/shell/bazel/starlark_repository_test.sh
coeuvre: Since the first command failed (exit code is not 0), we need to use `:` (which is a symbol for `true`) to set the exit code to 0. Just a trick to ignore the errors. (Copied from other test cases in the same file)

src/test/shell/bazel/starlark_repository_test.sh
coeuvre: Removed.

lberki (approved):

Issue #12310, opened 2020-10-19
Bazel+RBE failure on Windows with target of length 89 or 90


veblush:
### Description of the problem:

I ran into a weird error when running a test through Bazel RBE on Windows. There is the [issue](https://github.com/grpc/grpc/issues/24178) describing it in detail. Its summary is that SymInitialize (Windows Debug API) fails only when the length of the target is 89 or 90. When the target name is shortened or lengthened, this issue is fixed. So something related to the length should be there.

This is a [PR](https://github.com/grpc/grpc/pull/24302) to reproduce this simply and I haven't seen this problem when running Bazel directly on Windows. The Bazel RBE team investigated this problem and this might be a result of combination of bazel and docker so I like to get help from the Bazel team.

I guess changing the threshold (from 260 to 258) to use shortened paths on Windows can fix this issue. [code](
https://github.com/bazelbuild/bazel/blob/61dade55b9963aea51d282abf8c40d1fcd61b51f/src/main/native/windows/util.cc#L228)

Internal bug with details: b/169902891

### Feature requests: what underlying problem are you trying to solve with this feature?

SymInitialize failed only when the length of path is 89 and 90

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

https://github.com/grpc/grpc/pull/24302

### What operating system are you running Bazel on?

Windows server 2016

### What's the output of `bazel info release`?

I cannot run bazel info but the version of bazel is 2.2.0.

jin:
Is this still a P1 bug?

veblush:
@jin Yep, this is still affecting gRPC debugging.

meteorcloudy:
@veblush Can you help verify if https://github.com/bazelbuild/bazel/pull/12941 fixes this problem? You can cherry-pick this change and build a custom Bazel binary based the Bazel version you were using. Let me know if I can help verify this!

veblush:
@meteorcloudy Thank for the fix! I don't know how to use a custom build Bazel, though. Have you tried to build my reproduction case described [here](https://github.com/grpc/grpc/pull/24302)?

meteorcloudy:
@veblush 

I couldn't reproduce because I don't have access to required RBE resources.

I prepared a branch with the fix based on Bazel 2.2.0:
https://github.com/bazelbuild/bazel/tree/gh-12310-fix-on-2.2.0

Can you fetch this branch and built a custom Bazel binary via `bazel build //src:bazel`?

[...]

PR #12941, opened 2021-02-01
Update threshold for long path shortening to be MAX_PATH - 4

mai93:
This PR applies the suggested fix for #12310. Although I could not reproduce it.

Fixes: #12310


meteorcloudy (approved):
Nit: We still don't have a reproduce case without RBE? Otherwise, adding a test would be a good idea.

mai93:
Yes, I could not reproduce it without RBE.

meteorcloudy (approved):
Thanks for confirming, let's import this!

PR #12969, opened 2021-02-05
 Convert special characters in classpath to hexadecimal representation in java_stub_template

mai93:
This PR fixes the url-encoding in `java_stub_template` to convert different special characters to their 2-digit hexadecimal representation.

Fixes: #10620


meteorcloudy (approved):
Thanks, can you refer to the github issue link this PR tries to solve?

PR #12981, opened 2021-02-09
cc_shared_library: Don't fail when same shared lib is depended multiple times

jlaxson:
When using `--experimental-cc-shared-library`, the duplicate symbol check fails if a shared lib is linked multiple times.  For example:

```
cc_library(name="X_static")
cc_shared_library(name="X_shared", roots=[":X_static"])

cc_shared_library(name="A", dynamic_deps=[":X_shared"])
cc_shared_library(name="B", dynamic_deps=[":X_shared"])

cc_binary(name="test", dynamic_deps=[":A", ":B"])
```

Will yield an error like:

> in dynamic_deps attribute of cc_binary rule //:test: Two shared libraries in dependencies link the same library statically. Both //:X_shared and //:X_shared link statically //:X_static

This fix checks to ensure it is truly different shared libraries exporting the same static lib.


oquenchil:
Thank you! I will import this into Bazel. Would you mind writing a test in rules_cc that checks this?

oquenchil (approved):
LGTM

PR #12982, opened 2021-02-09
Test whether #12685 magically resolved itself.

meisterT:


meisterT:
Well, it did resolve itself, the same way it started...

PR #12989, opened 2021-02-10
Report errors parsing rewriter config file

illicitonion:
Before this change, an invalid config file results in a silent server
crash.


coeuvre (approved):
Thanks!

PR #12990, opened 2021-02-10
Remove bitcode from Apple host transition

keith:
If you ship your apps with bitcode it doesn't matter if tools that were
used to build it were built with the same bitcode mode, only that the
final binaries have them. This reduces unnecessary churn of host tools.


lberki:
Looks reasonable to me and I like making configuration transitions simpler, but I'd rather Tony take a look since he knows way more about all things Apple.

allevato (approved):
LGTM. There's no reason that I can think of to ever build in the host config with Bitcode enabled.

PR #12994, opened 2021-02-11
Expose java stub template through @bazel_tools

illicitonion:
Currently rules_kotlin downloads this via raw.githubusercontent.com:

https://github.com/bazelbuild/rules_kotlin/blob/96c70d1d3cc788b03406f7db6a659c78d311c3d8/kotlin/internal/repositories/release_repositories.bzl#L55-L62

https://github.com/bazelbuild/rules_kotlin/blob/96c70d1d3cc788b03406f7db6a659c78d311c3d8/kotlin/internal/jvm/jvm.bzl#L135-L138

This will allow rules_kotlin to point at a file from Bazel proper,
rather than downloading it via the internet.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BUILD
@@ -13,6 +13,11 @@ filegroup(
     visibility = ["//src:__subpackages__"],
 )
 
+exports_files(
+    ["java_stub_template.txt"],
+    visibility = ["//tools/java:__pkg__"]
+)
+
 java_library(
     name = "java",
     srcs = glob(
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -49,6 +49,7 @@ filegroup(
         "//tools/cpp:embedded_tools",
         "//tools/genrule:srcs",
         "//tools/java:embedded_tools",
+        "//tools/java:java_stub_template.txt",
         "//tools/j2objc:srcs",
         "//tools/jdk:package-srcs",
         "//tools/jdk:srcs",
--- a/tools/java/BUILD
+++ b/tools/java/BUILD
@@ -9,6 +9,16 @@ filegroup(
     visibility = ["//tools:__pkg__"],
 )
 
+genrule(
+    name = "copy_java_stub_template",
+    srcs = [
+        "//src/main/java/com/google/devtools/build/lib/bazel/rules/java:java_stub_template.txt",
+    ],
+    outs = ["java_stub_template.txt"],
+    cmd = "cp $< $@",
+    visibility = ["//tools:__pkg__"],
+)
+
 filegroup(
     name = "embedded_tools",
     srcs = ["//tools/java/runfiles:embedded_tools"],


comius (changes requested):
I don't mind exposing the template, just the implementation of it.

Is it possible to move the java_stub_template.txt file into the tools/java/ directory and then depend on it from rules/java? This way there would be no need for a copy genrule.

illicitonion:
Absolutely - done! This is now exposed as `@bazel_tools//tools:java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt` without any copies - thanks for taking a look :)

comius:
> Absolutely - done! This is now exposed as `@bazel_tools//tools:java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt` without any copies - thanks for taking a look :)

Do you need the whole path? I thought you could put it to tools/jdk package directly.

illicitonion:
> > Absolutely - done! This is now exposed as `@bazel_tools//tools:java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt` without any copies - thanks for taking a look :)
> 
> Do you need the whole path? I thought you could put it to tools/jdk package directly.

If I do that, when I reference it as a `resources` it ends up in the root of the jar, rather than in the package, and that makes the `ResourceFileLoader` sad. I could side-step `ResourceFileLoader` to not do the package-name-prefixing when trying to read the file, if you'd prefer?

comius:
> > > Absolutely - done! This is now exposed as `@bazel_tools//tools:java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt` without any copies - thanks for taking a look :)
> > 
> > 
> > Do you need the whole path? I thought you could put it to tools/jdk package directly.
> 
> If I do that, when I reference it as a `resources` it ends up in the root of the jar, rather than in the package, and that makes the `ResourceFileLoader` sad. I could side-step `ResourceFileLoader` to not do the package-name-prefixing when trying to read the file, if you'd prefer?

No sidestepping `ResourceFileLoader` doesn't seem nice. Can `filegroup` rule help in this case (with `path`)? Or even with the previous version you had with the file in original location, i.e. replacing genrule with filegroup?

illicitonion:
Filegroup doesn't appear to have a `path` attribute? https://docs.bazel.build/versions/master/be/general.html#filegroup

comius:
> Filegroup doesn't appear to have a `path` attribute? https://docs.bazel.build/versions/master/be/general.html#filegroup

Sorry, you're right.

illicitonion:
Would you rather I fix up CI for this version, or go back to a genrule which does `cp`? (Or something else :))

comius:
> Would you rather I fix up CI for this version, or go back to a genrule which does `cp`? (Or something else :))

I'd like to have "pretty" code, but I don't have time to play around myself to figure out if there is something nicer.

I'll accept the genrule with 'cp'. This way you don't have to deal with peculiarities of the tests and other options like creating out of place directory structure or tinkering with ResourceFileLoader seem worse.

I don't know by heart if you could use a regular `filegroup` instead of  `cp genrule` in the old version of the code. If you are willing to try; otherwise it's fine too.

illicitonion:
I played around with `filegroups` for a while to no effect - reverted to the original `cp` approach

comius (approved):

comius:
> I played around with `filegroups` for a while to no effect - reverted to the original `cp` approach

Thanks!

PR #12996, opened 2021-02-11
docs: update java.bzl best practices to rm dead URL

cpu:
The rules for dealing with external java dependencies offer best practices in a comment that includes a note about using `maven.ibiblio.org` as a third `jar_urls` entry:

https://github.com/bazelbuild/bazel/blob/81a5a12b3566e0b341c3cfbc271a60e568e1ed41/tools/build_defs/repo/java.bzl#L47

At the time of writing this domain consistently [fails to resolve w/ status `NXDOMAIN`](https://unboundtest.com/m/A/maven.ibiblio.org/MFPMBKPS):

```bash
 $> dig +noall +comments +answer @8.8.8.8 maven.ibiblio.org
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 19083
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
```

This PR removes the `maven.ibiblio.org` URLs throughout. I'm not familiar enough with the Java packaging ecosystem in 2021 to suggest an alternative but leaving it as-is produces unnecessary warning logs when the dead mirror is selected from the `jar_urls` for a dep.

### DIFF
--- a/tools/build_defs/repo/java.bzl
+++ b/tools/build_defs/repo/java.bzl
@@ -44,8 +44,7 @@ The recommended best practices for downloading Maven jars are as follows:
 2. Permanently mirror all dependencies to GCS or S3 as the first URL
 3. Put the original URL in the GCS or S3 object name
 4. Make the second URL the original repo1.maven.org URL
-5. Make the third URL the maven.ibiblio.org mirror, if it isn't 404
-6. Always specify the sha256 checksum
+5. Always specify the sha256 checksum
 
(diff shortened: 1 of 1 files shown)


aiuto:
This looks good to me, but @jin knows more about maven than I do, and @meteorcloudy knows more about our external deps.

jin (approved):

Issue #12988, opened 2021-02-10
XcodeConfigInfo.macosMinimumOsVersion is incorrectly set for host configuration


brentleyjones:
### Description of the problem / feature request:

When building a target with the host configuration, the `macosMinimumOsVersion` is being set to the SDK version instead of what is set for `--macos_minimum_os`. The relevant lines are:

https://github.com/bazelbuild/bazel/blob/24c980b6d4adef456cf4b4f84ac3b8ec4580b172/src/main/java/com/google/devtools/build/lib/rules/apple/XcodeConfig.java#L149-L152

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Build a `genrule` that has a `swift_binary` `tool`, with `--macos_minimum_os` lower than the SDK version (say `10.15` when using Xcode 12.4.0, which has its SDK version at `11.1`). The tool will have an incorrect `-target` (`x86_64-apple-macosx11.1` for the example given).

### What operating system are you running Bazel on?

macOS 11.2.1

### What's the output of `bazel info release`?

release 4.0.0

###  Have you found anything relevant by searching the web?

Nothing.

keith:
I think the issue is that the minimum version isn't set for the host configuration here? https://github.com/bazelbuild/bazel/blob/660f5b2cf36fbd75908cf7c93cc59d98ee092eb2/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java#L456-L468

But maybe that's intentional

brentleyjones:
If it's intentional, could we get a `--host_minimum_os` flag? Hopefully it's not intentional though 😄.

brentleyjones:
I just realized this might be semi-intentional, since we only need to support a minimum OS of the host, not any lower. In my example though, by using Xcode 12.4, the `macOsSdkVersion` is `11.1`, while my OS version (in a previous test) was `10.15`. Ideally we wouldn't try targeting an OS version _higher_ than the current one?

keith:
> minimum OS of the host

To ensure caching across developers even this isn't particularly true is it? I get that you won't back deploy the binary directly, but it could result in a different output

brentleyjones:
>To ensure caching across developers even this isn't particularly true is it? I get that you won't back deploy the binary directly, but it could result in a different output

True. So then I would want to set a minimum OS version that might be lower than the OS version of the host.

keith:
[...]

PR #13001, opened 2021-02-11
Add --host_macos_minimum_os flag

keith:
This flag makes sure we set the minimum macOS version when compiling
host tools. Otherwise you can end up not sharing caches across
different OS versions.

Fixes https://github.com/bazelbuild/bazel/issues/12988

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
@@ -164,6 +164,17 @@ public class AppleCommandLineOptions extends FragmentOptions {
   public DottedVersion.Option macosMinimumOs;
 
   @Option(
+      name = "host_macos_minimum_os",
+      defaultValue = "null",
+      converter = DottedVersionConverter.class,
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.LOSES_INCREMENTAL_STATE},
+      help =
+          "Minimum compatible macOS version for host targets. "
+              + "If unspecified, uses 'macos_sdk_version'.")
+  public DottedVersion.Option hostMacosMinimumOs;
+
+  @Option(
       name = "experimental_prefer_mutual_xcode",
       defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.TOOLCHAIN,
@@ -488,6 +499,7 @@ public class AppleCommandLineOptions extends FragmentOptions {
     host.watchOsSdkVersion = watchOsSdkVersion;
     host.tvOsSdkVersion = tvOsSdkVersion;
     host.macOsSdkVersion = macOsSdkVersion;
+    host.macosMinimumOs = hostMacosMinimumOs;
     // The host apple platform type will always be MACOS, as no other apple platform type can
     // currently execute build actions. If that were the case, a host_apple_platform_type flag might
     // be needed.
--- a/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcBuildVariablesTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcBuildVariablesTest.java
@@ -197,6 +197,45 @@ public class ObjcBuildVariablesTest extends LinkBuildVariablesTestCase {
   }
 
   @Test
+  public void testAppleBuildVariablesMacosHost() throws Exception {
+    MockObjcSupport.setup(mockToolsConfig);
+    String dummyMinimumOsValue = "13.579";
+    useConfiguration(
+        "--crosstool_top=//tools/osx/crosstool",
+        "--cpu=darwin_x86_64",
+        "--host_cpu=darwin_x86_64",
+        "--macos_minimum_os=10.11",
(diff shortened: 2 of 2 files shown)


brentleyjones (approved):

thii (approved):
Related change previously: https://github.com/bazelbuild/bazel/pull/9403.

allevato (commented):
This seems reasonable, but I'm surprised the problem described in the linked issue hasn't been more widespread. I think I'd like @trybka to take a look at it too just to make sure this is the right way to tackle it.

trybka:
I think this is fine? I'm not sure I entirely understand the usecase, though I can understand usecases where it might help. 

This also feels like maybe not ideal, but until we have widespread execution platforms, I don't see a better alternative.

brentleyjones:
The cache sharing would be based on different Xcode versions, not OS versions.

I'm not entirely sure why targeting a newer-than-current OS version does, since it's clearly "working" as is, but it's less than ideal imo, and we should be able to target our actual minimum OS. Maybe it could affect deprecation warnings for example?

keith:
Did a bit more digging here. So it looks like it will happily create a binary for the SDK version you specify, and run it on an older version, but the downside is you get none of the protections around API usage, so there's the big risk. You won't find out until runtime on an older OS that your tools are actually incompatible in some way depending on what API they use.

allevato:
I spent some more time thinking about this because something about it still didn't feel quite right. Some random thoughts:

* Conceptually, minimum deployment version is a property of individual targets, not of an entire build.
* The existence of `--{ios,macos,tvos,watchos}_minimum_os` command line flags would seem to go against that, but their main purpose is to be set by a transition based on an attribute value on an Apple bundle rule so that they apply to the `{cc,objc,swift}_library` targets underneath. They can be set directly when building a `*_library` target from the command line, but we discourage that.
* But, the flags also end up applying to rules like `cc_binary` and `swift_binary`, which don't have a way to set the deployment version since they also support non-Apple platforms, so they don't use the same transitions. Or perhaps stated more accurately, they don't so much "apply to `{cc,swift}_binary`" as they _serve as a fallback value for anything in the build that isn't under a transition that sets it more specifically._

The original issue mentioned the use case being a `genrule` with a `swift_binary` tool. What if you had _two or more_ such tools in your build that were written with different minimum deployment targets in mind? I _think_ it's almost always fine to just go with the higher value, but if you want to build them as they were intended, you'd want to use different values for each one. In order to do that with rules like `{cc,swift}_binary`, you'd have to write a small rule + transition combo that sets `--macos_minimum_os` and then passes through the underlying binary. That's a lot of ceremony, unfortunately. But I'd like to make sure there's not a cleaner/more correct way out of this before we add a new command line flag (especially at a time when we're trying to reduce, rather than increase, the Apple API surface in Bazel core).

keith:
> I think it's almost always fine to just go with the higher value

I don't think you could ever jump to just 1 value for this given that in either case of going higher or lower you could fail the build depending on the warnings / errors that were introduced with different deployment targets.

Definitely open the other suggestions here, having folks create their own transitions to solve this case is a super high barrier of entry to solve this bug so I think we should land some solution

aiuto:
@gregestren @katre   Could this be done without a flag? It seems very similar to the Android case.

katre:
Given that this is adding a host version of an existing flag, I think it makes sense.

The new Starlark flag we've added for Android's minimum SDK version is only possible because that affects the NDK, which is really a `cc_toolchain`, and so we added lots of select code in there to read and use the new flag. I don't think the macOS rules are that far moved to Starlark yet, so trying to use a Starlark flag isn't really appropriate there yet.

Highaboveus2 (approved):

brentleyjones:
@aiuto: Given @katre's and @trybka's comments, can this be merged?

aiuto:
@allevato @trybka  Can one of you please do a review and actually approve. 
And, does anyone know who @highaboveus2 is? It looks like a random driveby.

trybka (approved):

aiuto:
OK. I'll do the merge.

aiuto:
Damm. It must have gone stale and tests don't pass any more.
https://buildkite.com/bazel/google-bazel-presubmit/builds/49852
  https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/02d97ba9-2756-40f0-b56a-70a3d405fda2/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcRulesTests/attempt_1.log

Keith: can you look into that?

brentleyjones:
Can this be imported now? 🙏 Thanks!

aiuto:
Sigh. I am trying to merge this, but it fails, and I don't know how to fix the newly added test.
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/cdc80700-682c-4e3e-90a3-53e0cd6c121c/src/test/java/com/google/devtools/build/lib/rules/objc/ObjcRulesTests/shard_13_of_20/test_attempts/attempt_1.log

keith:
Hrm, I thought that was the one I fixed last time, testing locally again

keith:
@aiuto which build did that failure come from? I don't see that locally anymore, did that include my more recent force push?

aiuto:
Oh,... silly me. I may have to re-sync. I'll check later tonight.

Issue #12915, opened 2021-01-27
cquery with --output=graph crashes


nikola-sh:
### Description of the problem / feature request:

`bazel cquery` with `--output=graph` crashes with message
```
Starting local Bazel server and connecting to it...
DEBUG: /home/nikola/.cache/bazel/_bazel_nikola/cdd5da0d482563cb2bf964e63e3b5128/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14:
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: /home/nikola/.cache/bazel/_bazel_nikola/cdd5da0d482563cb2bf964e63e3b5128/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1804_java11 not using checked in configs as detect_java_home was set to True
DEBUG: /home/nikola/.cache/bazel/_bazel_nikola/cdd5da0d482563cb2bf964e63e3b5128/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14:
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: /home/nikola/.cache/bazel/_bazel_nikola/cdd5da0d482563cb2bf964e63e3b5128/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1604_java8 not using checked in configs as detect_java_home was set to True
INFO: Analyzed target //src:bazel (329 packages loaded, 9069 targets configured).
INFO: Found 1 target...
WARNING: Targets were missing from graph: [ConfiguredTargetKey{label=@bazel_tools//tools/jdk:dummy_toolchain, config=BuildConfigurationValue.Key[86d23d2b5aedbeecfc6e6954ea9486ec39e9e40bcbc4b17cec46a06eed198d11]}]
WARNING: Targets were missing from graph: [ConfiguredTargetKey{label=@bazel_tools//tools/jdk:dummy_toolchain, config=BuildConfigurationValue.Key[d8d0eb07eb92791668ac973282be1523379b0025b22f0ade56b13d519f2feb2a]}]
[...]

PR #13002, opened 2021-02-12
Fix NullPointerException in cquery --output=graph.

gregestren:
Fixes https://github.com/bazelbuild/bazel/issues/12915.

The repro from the bug crashes because these two deps are compared:

`Target: @remote_java_tools_linux//:java_tools/src/tools/singlejar/singlejar_local`
`Config: d8d0eb07eb92791668ac973282be1523379b0025b22f0ade56b13d519f2feb2a`

`Target: @remote_java_tools_linux//:java_tools/src/tools/singlejar/singlejar_local`
`Config: null`

This is a `cc_binary`: https://github.com/bazelbuild/bazel/blob/0c1257ed4e1b83f8d0f6c79d641f6bfcf4d1cfc4/src/tools/singlejar/BUILD#L92-L93

I have no idea why the graph visitor tags one of its instances with a null configuration. Nevertheless, this code is still safer.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
@@ -25,6 +25,7 @@ import com.google.devtools.build.lib.query2.query.output.GraphOutputWriter.NodeR
 import com.google.devtools.build.lib.skyframe.SkyframeExecutor;
 import java.io.OutputStream;
 import java.util.Comparator;
+import java.util.Objects;
 
 /** cquery output formatter that prints the result as factored graph in AT&amp;T GraphViz format. */
 class GraphOutputFormatterCallback extends CqueryThreadsafeCallback {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
```diff
@@ -50,9 +51,20 @@ public String getName() {
...
               Label label2 = ct2.getLabel();
-              return label1.equals(label2)
-                  ? ct1.getConfigurationChecksum().compareTo(ct2.getConfigurationChecksum())
-                  : label1.compareTo(label2);
+              if (!label1.equals(label2)) {
+                return label1.compareTo(label2);
+              }
+              String checksum1 = ct1.getConfigurationChecksum();
+              String checksum2 = ct2.getConfigurationChecksum();
+              if (Objects.equals(checksum1, checksum2)) {
```
katre, line 59: This is the same as the compareTo below, and the checksum2 null check isn't needed, either. I'd simplify this as:

if (checksum1 == null) {
  return -1;
} else {
  return checksum1.compareTo(checksum2);
}

katre (changes requested):

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
gregestren: How's this? I left the double null check because of this comment from https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html:

> Note that null is not an instance of any class, and e.compareTo(null) should throw a NullPointerException even though e.equals(null) returns false.

gregestren: Or am I missing something about the null check?

gregestren:
Also note I don't know how to add a test since I don't know how to invoke a dep simultaneously in a null and non-null configuration.

It's probably worth exploring how the query logic decides a `cc_binary` appears in both configurations. Not sure I should worry about that for the scope of this PR (scope creep). But open to suggestions.

src/main/java/com/google/devtools/build/lib/query2/cquery/GraphOutputFormatterCallback.java
katre: Ah, I hadn't realized that. Okay, the second null check is necessary, then.

katre (approved):

PR #13006, opened 2021-02-12
Flip dontCollectSoArtifacts

comius:


comius:
Downstream test: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1911#b7a47544-7f5a-440b-ab15-c64c8b264bb8

Problems building: Bazel, Android Studio Plugin

PR #13027, opened 2021-02-14
Fix link text

dabrahams:
I realize the link starts with be/ but from the user experience POV, that link doesn't refer into the build encyclopedia, at least not according to the TOC in the left-hand column.  If you don't want to accept this change, perhaps the TOC should be changed to reflect the file structure of the site?


gregestren (approved):
I agree that this change is more precise *and* more future-proof (it doesn't matter what the broader document is called).

PR #13028, opened 2021-02-14
Imply shorthand applies at command line

dabrahams:
"in which it occurs" tends to imply that the text is used in a BUILD file in that repository, but IIUC this shorthand also works for targets specified at the command line.  I think this phrasing better suggests the shorthand works in both contexts.


communikit (approved):

gregestren:
Gentle ping @communikit : is anything blocking this from being merged?

communikit:
> Gentle ping @communikit : is anything blocking this from being merged?

Actually, looked into it, and it looks like the current wording has been updated already, reflecting what this commit seeks to change:

https://github.com/bazelbuild/bazel/blob/master/site/docs/build-ref.html#L173

@dabrahams , since you updated this section with https://github.com/bazelbuild/bazel/commit/804bcc1426421f8ac2ff2821e4da66c66b82a49c, maybe this PR is not needed?

dabrahams:
@gregestren The idea was that “from which it is used” imparts a different shade of meaning than “it is used in,” that is more appropriate to the command-line being part of the implication.  It may be too subtle a change to be worth making, or may not work that way for you, though.

gregestren:
I'm happy deferring to @communikit on this one.

PR #13031, opened 2021-02-14
Correct(?) package name

dabrahams:
If  `@myrepo//` and `//` are not part of the package name, then the label has at least 3 parts, so I'm assuming my change is correct.  If not, please correct me.

### DIFF
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -178,7 +178,8 @@ src/my/app/tests/test.cc
 
 <p>
 
-  Each label has two parts, a package name (<code>my/app/main</code>)
+  Each label has two parts, a package name
+  (<code>@myrepo//my/app/main</code> or <code>//my/app/main</code>)
   and a target name (<code>app_binary</code>). Every label uniquely
   identifies a target. Labels sometimes appear in other forms; when
   the colon is omitted, the target name is assumed to be the same as


dabrahams:
Ugh, now I see

> Short-form labels such as //my/app are not to be confused with package names. Labels start with //, but package names never do, thus my/app is the package containing //my/app.

So now I'm thinking my change is wrong, and the correct thing to say is: “labels have 3 parts: a repository (which may be shortened to `//` if it's local), a package name, and a target name (which can be empty when the target name is the same as the last component of the package name).”

gregestren:
Agreed. References docs:

- https://docs.bazel.build/versions/4.0.0/build-ref.html#packages
- https://docs.bazel.build/versions/4.0.0/glossary.html#package

dabrahams:
> * https://docs.bazel.build/versions/4.0.0/build-ref.html#packages

That's the document I'm editing. @gregestren?

gregestren:
> > * https://docs.bazel.build/versions/4.0.0/build-ref.html#packages
> 
> That's the document I'm editing. @gregestren?

I meant specifically the two definitions of "package".

dabrahams:
@gregestren a picture of the truth is starting to emerge; I tried to capture it with a more ambitious revision.

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
-  targets must always be referred to by their complete label, e.g.
-  <code>//my/app:generate.cc</code>.
+  Labels such as <code>//my/app</code> are not to
+  be confused with package names. Labels always start with a
+  repository identifier (often abbreviated <code>//</code>)
+  but package names never do. Thus, <code>my/app</code> is the
+  package containing <code>//my/app/lib</code> (a.k.a. 
+  <code>//my/app/lib:lib</code>).
+
+  (A common misconception is that <code>//my/app</code> refers
```
gregestren, line 233: Is it worth explaining what it *does* mean? Or position it in the text where that implication is clear?

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
-generate.cc
-testdata/input.txt
+//my/app/main:testdata/input.txt
 </pre>
 
 <p>
-  But from other packages, or from the command-line, these file
-  targets must always be referred to by their complete label, e.g.
-  <code>//my/app:generate.cc</code>.
+  Labels such as <code>//my/app</code> are not to
```
gregestren, line 226: Optional suggestion with stronger wording:

```suggestion
Don't confuse labels like <code>//my/app</code> with package names.
```

gregestren (commented):
Looks pretty good. Added lots of nits.

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
-  and a target name (<code>app_binary</code>). Every label uniquely
-  identifies a target. Labels sometimes appear in other forms; when
-  the colon is omitted, the target name is assumed to be the same as
-  the last component of the package name, so these two labels are
-  equivalent:
+  The second part of the label is the un-qualified package name
+  <code>my/app/main</code>, the path to the package
+  relative to the repository root.  When the label refers to the same
+  package it is used in, the package name (and optionally, the colon)
+  may be omitted.  So, inside <code>@myrepo//my/app/main</code>,
```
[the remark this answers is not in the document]
dabrahams: * In the first bullet, I would say “from that package's BUILD file,” even if labels can currently only appear within  rule targets.  Surely just putting the label in the text of some file target has no special meaning, right? Is there an additional case you mean to cover by saying “from a target?” 
* typos (bulid) aside, I don't understand the 2nd bullet; my shell won't `cd` into a bazel label.  So, yeah, bad directory, but I don't get what point you're trying to make.  Could you explain?
* What does “the core package” mean?
* if `subdir` isn't a distinct package, `subdir/myfile.cc` is just a target name (last part of the label), right?  At least that's consistent with

> The name of a file target in a subdirectory of the package is the file's path relative to the package root

which you seem to accept.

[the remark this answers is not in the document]
dabrahams: Anyway, the point of "inside" was that it *is* a little ambiguous, i.e. could mean in the BUILD file or on the command line, and I expect the reader to infer that it means both without thinking too hard about it.  We could explicitly spell out somewhere that any label syntax that is valid in a BUILD file is also valid on the command-line in that directory or any non-subpackage subdirectory.

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
+  this label may be written either of the following ways:
```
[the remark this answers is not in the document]
dabrahams: Sorry, is that an action item, or just a comment?  (N.B. I “don't do” Java 😉 )

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
-  Short-form labels such as <code>//my/app</code> are not to
-  be confused with package names. Labels start with <code>//</code>,
-  but package names never do, thus <code>my/app</code> is the
-  package containing <code>//my/app</code>.
-
-  (A common misconception is that <code>//my/app</code> refers
-  to a package, or to <em>all</em> the targets in a package; neither
-  is true.)
+  (It is a matter of convention that the colon is omitted for files,
+  but retained for rules, but it is not otherwise significant.)
```
dabrahams, line 197:
```suggestion
  It is a matter of convention that the colon is omitted for files,
  but retained for rules, but it is not otherwise significant.
```

dabrahams: Click and drag over the line numbers, then hit the little blue "+" tooltip, and you can make a multiline suggestion.

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
+//my/app/lib:lib
 </pre>
 
 <p>
-  (It is a matter of convention that the colon is omitted for files,
-  but retained for rules, but it is not otherwise significant.)
+  The name of a file target in a subdirectory of the
+  package is the file's path relative to the package
+  root (the directory containing the <code>BUILD</code> file),
+  so this file is in the <code>my/app/main/testdata</code>
```
dabrahams, line 217:
```suggestion
  root (the directory containing the <code>BUILD</code> file).
  So, this file is in the <code>my/app/main/testdata</code>
```

site/docs/build-ref.html
dabrahams: I wouldn't say it's more strongly worded.  But it is less high-handed and high-falutin', so it's all to the good 😉

site/docs/build-ref.html
```diff
@@ -167,82 +168,78 @@ <h3 id="labels">Labels</h3>
...
+  Labels such as <code>//my/app</code> are not to
+  be confused with package names. Labels always start with a
+  repository identifier (often abbreviated <code>//</code>)
+  but package names never do. Thus, <code>my/app</code> is the
+  package containing <code>//my/app/lib</code> (a.k.a. 
+  <code>//my/app/lib:lib</code>).
+
+  (A common misconception is that <code>//my/app</code> refers
+  to a package, or to <em>all</em> the targets in a package; neither
+  is true.)
```
dabrahams, line 235:
```suggestion
  is true.  Remember, it is equivalent to `//my/app:app`, so it names the
  `app` target in the `my/app` package of the current repository).
```

site/docs/build-ref.html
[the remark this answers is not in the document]
gregestren: That's all fine. Just food for thought.

My first bullet point was just to cover all possible interpretations of accessing a target. That one's most straightforward.

The point of the second bullet, as you pointed out, is that that interpretation of "inside" doesn't work if applied literally. You could do the slight variation `cd my/app/main && bazel build app_binary` and that better fulfills the assertion.

"Core" package was sloppy wording: that meant "the parent package of the subpackage, but in the case where the subpackage isn't actually a package". That's what I get for writing fast!

Anyway, I'm not trying to stir up the pot with all this. Just pointing out subtleties. I agree with your final assessment, and that there's only so far we have to ultimately walk the reader to be clear.

site/docs/build-ref.html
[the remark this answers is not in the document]
gregestren: Just an FYI for this PR's general record. I'll resolve.

gregestren (approved):
Enough pedantry on my part. Thanks for seeing this through.

site/docs/build-ref.html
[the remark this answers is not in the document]
dabrahams: > The point of the second bullet, as you pointed out, is that that interpretation of "inside" doesn't work if applied literally. You could do the slight variation cd my/app/main && bazel build app_binary and that better fulfills the assertion.

How about this?  But I only want to do it if you think it actually improves comprehension.  It is a bit wordy:

```suggestion
  may be omitted.  So, within the directory tree of <code>@myrepo//my/app/main</code>,
```

PR #13033, opened 2021-02-14
Correct statement about command-line

dabrahams:
Since 

```
bazel build :hello-greet.cc
```
and
```
bazel build hello-greet.cc
```
both work for me, I assume the statement about what's allowed at the command line is wrong.

### DIFF
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -234,7 +234,7 @@ testdata/input.txt
 </pre>
 
 <p>
-  But from other packages, or from the command-line, these file
+  But from other packages, these file
   targets must always be referred to by their complete label, e.g.
   <code>//my/app:generate.cc</code>.
 </p>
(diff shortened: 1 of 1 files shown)


gregestren (commented):
I agree with your point.

Grammar nitpick comment: I think we could trim "But" from this sentence (shorter and no meaning is lost).

site/docs/build-ref.html
```diff
@@ -234,7 +234,7 @@ <h3 id="labels">Labels</h3>
 </pre>
 
 <p>
-  But from other packages, or from the command-line, these file
+  But from other packages, these file
```
dabrahams, line 237:
```suggestion
  From other packages, these file
```

communikit:
@dabrahams : This PR may not be necessary or may need to be updated after the merge of https://github.com/bazelbuild/bazel/commit/804bcc1426421f8ac2ff2821e4da66c66b82a49c

Btw, thank you for all the contributions!

dabrahams:
Closing as obsolete.

PR #13039, opened 2021-02-14
Clarify that a label names a target.

dabrahams:
The text as written implies `//:foo` names a package.

### DIFF
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -349,8 +349,8 @@ testdata/input.txt
 
 <p>
   Although Bazel allows targets to be declared in the package at the build root
-  (e.g. <code>//:foo</code>), it is discouraged: it is better if every target
-  is associated with package having a descriptive name.
+  (e.g. <code>//:foo</code>), you are encouraged to leave that package empty:
+  it is better if every target is associated with package having a descriptive name.
 </p>
   
 <p>
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -348,7 +348,8 @@ testdata/input.txt
 </p>
 
 <p>
-  Although Bazel allows a package at the build root (e.g. <code>//:foo</code>), this
+  Although Bazel allows a package at the build root (e.g.
+  one that defines a target <code>//:foo</code>), this
   is not advised and projects should attempt to use more descriptively named
   packages.
 </p>
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -350,7 +350,7 @@ testdata/input.txt
 <p>
   Although Bazel allows targets to be declared in the package at the build root
   (e.g. <code>//:foo</code>), you are encouraged to leave that package empty:
-  it is better if every target is associated with package having a descriptive name.
+  it is better if every target is in a package having a descriptive name.
 </p>
   
 <p>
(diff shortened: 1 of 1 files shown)


site/docs/build-ref.html
```diff
@@ -348,7 +348,8 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
 </p>
 
 <p>
-  Although Bazel allows a package at the build root (e.g. <code>//:foo</code>), this
+  Although Bazel allows a package at the build root (e.g.
```
aiuto, line 351: It don't think this change is better.  It talks about packages, but the code block is a target.
... _package_ at the build route ...  defines a _target_ ...
It conflates the two.

aiuto (changes requested):

site/docs/build-ref.html
dabrahams: I disagree that it conflates anything but leaving that aside, I hope you can see that a standard English interpretation of what's there appears to be saying that `//:foo` is the name of a package.  

And I understand why the current text uses a target name; if you wanted to use the package name, you'd have to write

```suggestion
  Although Bazel allows a package at the build root (i.e. <code>//</code>), this
```

which is confusing in that `<code>//</code>` reads as though it refers to the build root, not the package, and the package in question has an empty name.  I'm guessing closer to the truth would be that the package at the build root always exists, we are discouraged from putting targets in it:

```suggestion
  Although Bazel allows targets at the build root (i.e. <code>//:foo</code>), it
  is discouraged: projects should attempt to only create targets in packages
  with descriptive names.
```

site/docs/build-ref.html
```diff
@@ -348,10 +348,11 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
 </p>
 
 <p>
-  Although Bazel allows a package at the build root (e.g. <code>//:foo</code>), this
-  is not advised and projects should attempt to use more descriptively named
-  packages.
+  Although Bazel allows targets to be declared in the package at the build root
+  (e.g. <code>//:foo</code>), you are encouraged to leave that package empty:
+  it is better if every target is in a package having a descriptive name.
```
gregestren, line 353: I think this clarification is fair but still wordy in this draft. So without changing the substance I'd suggest something like:

"Although Bazel supports targets in the workspace's root package (e.g. <code>//:foo</code>), it's best to leave that package empty so all meaningful packages have descriptive names".

Also note [Bazel's doc style guide](https://bazel.build/basics/docs-style.html) discourages Latin abbreviations ([parent doc reference](https://developers.google.com/style/abbreviations)). Maybe "you" is still better than "it's" as I suggest ([reference](https://bazel.build/basics/docs-style.html#subject)). Also try to use active voice throughout ([reference](https://bazel.build/basics/docs-style.html#tense), my suggestion isn't perfect with that).

gregestren: Please take my pedantic grammar comments as a sign that I support the main point of this change.

dabrahams: Love your suggestion, “omit needless words” can be challenging.  Next time use the "insert a suggestion" button and it will save substantial labor :-)

gregestren: Well I just learned about a new GitHub feature. :)

I'd like to give @aiuto another opportunity to respond. If there's no more update after a day or two, please re-ping and I'll get this approved and merged.

dabrahams: @aiuto it looks like this PR is waiting on you.

aiuto (approved):

Issue #11559, opened 2020-06-06
Documentation issue: repository_ctx.download omits "success" field in return struct.


psigen:
Documentation URL: https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html

The documentation for [`repository.ctx.download()`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download) and [`repository_ctx.download_and_extract()`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download_and_extract) 
omits that the return struct includes a `success` field that returns `true` if the download completed successfully, and `false` if it does not.

(At least, that's what I think it does-- I'm not entirely sure because it's not documented.)

I'm happy to make a PR with the above notes if someone can help me find the part of the repo where the source markdown for this part of the docs is located.

communikit:
Hi @psigen ,

You can make a PR updating the documentation in the doc fields for download and download_and_extract in this part of the Bazel repo: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/repository/StarlarkRepositoryContextApi.java

psigen:
Thanks @communikit .  Created https://github.com/bazelbuild/bazel/pull/13041 to address this.

PR #13041, opened 2021-02-14
Fix missing `success` field in `repository_ctx.download`

psigen:
This adds a description for the `success` field in `repository.ctx.download()` and
`repository_ctx.download_and_extract()` to the docstring for these functions.

Closes https://github.com/bazelbuild/bazel/issues/11559


src/main/java/com/google/devtools/build/lib/starlarkbuildapi/repository/StarlarkRepositoryContextApi.java
```diff
@@ -329,8 +329,9 @@ void patch(Object patchFile, StarlarkInt strip, StarlarkThread thread)
       name = "download",
       doc =
           "Downloads a file to the output path for the provided url and returns a struct"
-              + " containing a hash of the file with the fields <code>sha256</code> and"
-              + " <code>integrity</code>.",
+              + " containing <code>success</code>, a flag which is <code>true</code> if the"
```
aiuto, line 332: Is it a "flag" when it is the status indicator, or is it "containing success, which is true if the download completed ...

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/repository/StarlarkRepositoryContextApi.java
```diff
@@ -329,8 +329,9 @@ void patch(Object patchFile, StarlarkInt strip, StarlarkThread thread)
...
+              + " download completed successfully, and if successful, a hash of the file"
+              + " with the fields <code>sha256</code> and <code>integrity</code>.",
```
aiuto, line 334: "with the fields ..."  reads like the hash is a sub-struct.   {success: t/f,  "hash?": {sha256, integrity}}

Is it actually a single struct with up to 3 fields, where sha256 and integrity are not included if success is false?
I think so, but I do not know the answer offhand.

psigen: I also do not know the answer offhand.  I will try to research it, but I'm not a contributor to this part of the code so it will take some time for me.

psigen:
@sgowroji Yup, done.

meteorcloudy:
@aiuto Can you please follow up on this?

aiuto:
I'm passing this to @Wyverald to make a judgement on the wording. They added the feature, so they are best suited to know the intent.

Wyverald (approved):
welp I didn't create the feature, but the changes look good nonetheless. Thanks!

Issue #12927, opened 2021-01-29
Output download failed: digest mismatch


federicohml:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

After upgrading to Bazel 4.0.0, I get the following error

```
com.google.devtools.build.lib.remote.BulkTransferException
        at com.google.devtools.build.lib.remote.RemoteCache.waitForBulkTransfer(RemoteCache.java:227)
        at com.google.devtools.build.lib.remote.RemoteCache.download(RemoteCache.java:338)
        at com.google.devtools.build.lib.remote.RemoteSpawnRunner.downloadAndFinalizeSpawnResult(RemoteSpawnRunner.java:487)
        at com.google.devtools.build.lib.remote.RemoteSpawnRunner.exec(RemoteSpawnRunner.java:307)
        at com.google.devtools.build.lib.exec.SpawnRunner.execAsync(SpawnRunner.java:240)
        at com.google.devtools.build.lib.exec.AbstractSpawnStrategy.exec(AbstractSpawnStrategy.java:140)
        at com.google.devtools.build.lib.exec.AbstractSpawnStrategy.exec(AbstractSpawnStrategy.java:102)
        at com.google.devtools.build.lib.actions.SpawnStrategy.beginExecution(SpawnStrategy.java:47)
        at com.google.devtools.build.lib.exec.SpawnStrategyResolver.beginExecution(SpawnStrategyResolver.java:65)
        at com.google.devtools.build.lib.rules.cpp.CppLinkAction.beginExecution(CppLinkAction.java:306)
        at com.google.devtools.build.lib.actions.Action.execute(Action.java:127)
        at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$5.execute(SkyframeActionExecutor.java:855)
        at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$ActionRunner.continueAction(SkyframeActionExecutor.java:1016)
        at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$ActionRunner.run(SkyframeActionExecutor.java:975)
        at com.google.devtools.build.lib.skyframe.ActionExecutionState.runStateMachine(ActionExecutionState.java:129)
[...]

PR #12279, opened 2020-10-15
[1/3] Bump grpc to 1.32.x to fix a too_many_pings regression

dmivankov:
Part 1: add v1.32.x version to third_party/grpc
Note: partly switches to v1.32.x too as not all bits are versioned and
      some of unversioned bits are used from other third_party targets
Composed PR: https://github.com/bazelbuild/bazel/pull/12273

grpc-java transition from v1.26.0 to v1.31.1 enabled auto flow control
which  started failing in RBE with

io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Bandwidth exhausted
HTTP/2 error code: ENHANCE_YOUR_CALM
Received Goaway
too_many_pings

grpc-java v1.32.2 has a bugfix attempt on that
grpc v1.32.0 also has something new around keepalive pings

Hopefully version bump to those helps

https://github.com/bazelbuild/bazel/issues/12264

Note: also an attempt and disabling auto flow by default is made in
https://github.com/bazelbuild/bazel/pull/12266

PR #13055, opened 2021-02-17
Bump grpc to 1.33.1 to fix corruption when downloading CAS blobs

scele:
    

divanorama:
github says jars were renamed without change which is suspicious

scele:
@divanorama @coeuvre I originally tried to upgrade to 1.34.1, but that was causing issues since grpc 1.34.1 had upgraded many of its transitive dependencies like upb and abseil... So now I'm trying to upgrade to 1.33.1 instead.

divanorama:
There are further bugfixes in 1.34+, but shouldn't hurt to try 1.33.1 first imho
Also noticed that grpc-core has 1.33.2 https://github.com/grpc/grpc/releases/tag/v1.33.2

scele:
> Also noticed that grpc-core has 1.33.2 https://github.com/grpc/grpc/releases/tag/v1.33.2

@divanorama Good point! Diffing 1.33.1 and 1.33.2 it seems that the only material change is in the python client, so in practice it should not make any difference. I can certainly bump to 1.33.2 instead if you prefer?

coeuvre:
@scele Thanks for catching the bug and creating the fix!
@meteorcloudy Can you please import the changes?

meteorcloudy:
@scele Thanks for the PR! For reasons, we cannot merge a PR containing both third_party and non-third_party changes.
Could you please split this PR into separate changes?
An example is:
https://github.com/bazelbuild/bazel/pull/12279
https://github.com/bazelbuild/bazel/pull/12288
https://github.com/bazelbuild/bazel/pull/12289

scele:
@meteorcloudy I have created:
#13104 [1/3]
#13105 [2/3] (CI fails because this patch requires #13104)
#13106 [3/3] (This also includes the commit from #13104, since otherwise the patch does not apply cleanly on master)

Let me know if I can do more. I'm also 100% OK with you or other googlers modifying my commits/PRs as needed to get them in.

meteorcloudy:
Thanks! I'll start to merge them.

scele:
Merged by #13104, #13105, #13106 - closing. Thanks!

PR #13069, opened 2021-02-19
remove signal unsafe code from signal handlers

grimreaper:
Problem

Signal handlers must only call signal-safe code. In particular the
functions it calls must be reentrant or uninterruptible by a signal.

For more information see
https://www.freebsd.org/cgi/man.cgi?query=sigaction&sektion=2&apropos=0&manpath=FreeBSD+13.0
or
https://man7.org/linux/man-pages/man7/signal-safety.7.html

fprintf is not such a function but is called inside of ForwardSignal
(via PRINT_DEBUG)

Solution

Remove the call.

### DIFF
--- a/src/main/tools/linux-sandbox-pid1.cc
+++ b/src/main/tools/linux-sandbox-pid1.cc
@@ -355,7 +355,6 @@ static void EnterSandbox() {
 }
 
 static void ForwardSignal(int signum) {
-  PRINT_DEBUG("ForwardSignal(%d)", signum);
   kill(-global_child_pid, signum);
 }
 


grimreaper:
i also checked`OnGracefulSignal`, `OnAbruptSignal`, and `ForciblyKillEverything`

`OnTimeoutOrTerm` looks more suspicious but is bigger and requires more thought.

jin:
Not sure who owns this code now. @philwo, looks like you wrote most of it. Could you review this?

meisterT:
cc @larsrc-google

larsrc-google:
Context: The `PRINT_DEBUG` macro does nothing unless the `global_debug` variable in `logging.cc` is set to true. That value in turn is controlled by the `debug` flag, so it could be activated without knowing the implications. That makes me inclined to agree with this PR, but I'm not strong on the subtleties of signal handling.

aiuto:
Friendly ping to @larsrc-google and @philwo  
Are we going to merge this or close it?

larsrc-google:
Yeah, that should be merged.

aiuto:
Fine. Since I am on triage duty this week I will merge it.

aiuto (approved):

PR #13080, opened 2021-02-21
Fix Runfiles::Create doc when argv is unavailable

jerryct:


tools/cpp/runfiles/runfiles_src.h
```diff
@@ -39,7 +39,7 @@
 //         // Important:
 //         //   If this is a test, use Runfiles::CreateForTest(&error).
 //         //   Otherwise, if you don't have the value for argv[0] for whatever
-//         //   reason, then use Runfiles::Create(&error).
+//         //   reason, then use Runfiles::Create({}, &error).
```
jerryct, line 42: An alterative is that I add this:

```
Runfiles *Runfiles::Create(string *error) {
  return Runfiles::Create({}, error);
}

```

aiuto: the shore way is fine

PR #13085, opened 2021-02-22
Allow overriding the hostname and instance name in bytestream:// URIs

EdSchouten:
In cases where full network transparency doesn't exist, people may run
Bazel with custom values of --remote_executor and --remote_instance_name
to proxy gRPC traffic. Such proxies may do caching, tunneling and
authentication.

What's problematic about this is that the values of these command line
flags aren't just used to establish a gRPC connection to a remote
execution service. They also get logged by Bazel in build event streams
in the form of bytestream:// URIs. This means that a build event stream
generated on system A may contain bytestream:// URIs that are not valid
on system B.

Let's introduce a command line flag, --remote_bytestream_uri_prefix,
that can be used to force generation of bytestream:// URIs that are
canonical.


EdSchouten:
CI failure seems to be unrelated to the change at hand?

philwo (approved):

PR #13091, opened 2021-02-22
Add ZERO_AR_DATE to apple_env

keith:
This variable disables timestamps in archives and binary links in
Apple's toolchain. Previously this was set for archives but not the
links themselves.

Relevant logic in ld64: https://github.com/keith/ld64/blob/86955a06a7e8e0d18ab28e521aa1122bfecd05a9/src/ld/Options.cpp#L4420-L4422

Fixes the other half of https://github.com/bazelbuild/bazel/issues/10886


oquenchil (approved):
LGTM

Issue #348, opened 2015-07-29
Android target for a library aar


mrdomino:
One of my work projects is an Android library with a substantial JNI footprint, designed for inclusion in third-party Android apps. I'd love to be able to specify it as a single bazel target (say for an aar, which seems to be what the Android people want us to generate these days) that depends on my android_library and cc_library.

My current planned stopgap is to build an android_binary and then take the .jar and .so out of the target for redistribution. Improvements and suggestions welcome.

aj-michael:
`android_library` has an implicit AAR output called `name.aar`.

As an example,

```
$ cat java/com/foo/BUILD
cc_library(
    name = "cc_lib",
    srcs = ["foo.cc"],
)
android_library(
    name = "android_lib",
    srcs = ["Foo.java"],
    deps = [":cc_lib"],
)
$ bazel build //java/com/foo:android_lib.aar
INFO: Found 1 target...
Target //java/com/foo:android_lib.aar up-to-date:
  bazel-bin/java/com/foo/android_lib.aar
INFO: Elapsed time: 4.138s, Critical Path: 3.59s
$
```

aj-michael:
Augh, I should not have closed this. Per the documentation of android_library:

> name.aar: An android 'aar' bundle containing the java archive and resources of this target. It does not contain the transitive closure.

https://bazel.build/versions/master/docs/be/android.html#android_library

So the .aar output of android_library will not contain the .so from the cc_library.

I'm going to use this as the tracking bug for support transitive-closure-aars from android_library.

andrewharp:
This would also be useful for TF so that we can build a full Android distributable in a single file (using only Bazel) during our [nightly build process](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/builds/android_full.sh).

spinorx:
+1.  Any update of expected support for this?

ahumesky:
We're working on getting our testing for android apps in better shape, this is something we'll likely work on after that.

steeve:
hey guys, i'm having another issue: the `cc_library` on which my `android_library` depends is not built when building either `name` or `name.aar`

balazsbanyai:
+1

spinorx:
Use sample in https://github.com/spinorx/grpc_andr_ios to unblock urself till this is fixed.

balazsbanyai:
@spinorx thank you very much. I can build the app using the provided project, but the android_library won't build in this example either.

[...]

Issue #13092, opened 2021-02-23
Allowing cc_library to depend on android_library, like with objc_library


cpsauer:
Hello, awesome Bazel folks,

This one's a feature request: It would be awesome if a cc_library could depend on an android_library, just as a cc_library can depend on an objc_library.

The underlying problem I'm trying to solve (with Bazel in general, really) is creating a cross-platform library that's mostly shared native code. That native code needs to call some platform-specific code. This works quite well on, e.g., Apple platforms, since you can just select() the right objc_library to depend on for a given platform, and makes me very happy that I'm using Bazel.

Using Android, however, is a pain point for these cross-platform libraries. You often need to call into Java using the JNI, but, at least to the best of my knowledge, you can't really model those C++ -> Android (aar/java) dependencies in Bazel, because cc_libraries can't depend on android_libraries. So you're stuck having two separate dependency trees going down to the root that can get out of sync: one for java and one for C++. 

I think it's going to be possible hide the ugliness with a macro that constructs the two separate graphs in parallel, but that's more of summoning up a horrible hack than an elegant Bazel solution. Plus, I bet having the C++ depend on the .so and .h in the aar is going to be more of an adventure. I'm envisioning having to extract them and suppress linking, like in https://github.com/bazelbuild/bazel/blob/master/tools/android/aar_resources_extractor.py. If others want these hacks, happy to share em once I finish, just lmk.

Some links that shed more light on what's needed:
- [Android Studio/Gradle seem to be working on solving this problem with prefabs.](https://developer.android.com/studio/build/native-dependencies)
  - That illuminates some of the structure of the problem: Specifying dependencies in the graph, exposing headers, and doing the right linking checking and bundling with .so's. Obviously they aren't going for the cross-platform use case Bazel is.
- You can see fbjni grappling with similar issues [here](https://github.com/facebookincubator/fbjni/blob/master/docs/android_setup.md).
- As are tensorflow users...where the recommended fix is to ditch bazel for make...: https://github.com/tensorflow/tensorflow/issues/39148
- The golang folks are grappling with a similar issue, which they (delightfully) call the C++ sandwich :) #2176

It'd be great to solve this problem in Bazel, and model cross-platform build graphs elegantly!

Thanks so much for your consideration,
Chris

aiuto:
Note: I assigned this to team-Android, but team-Rules-CPP or configurability might find it interesting.

[...]

PR #13098, opened 2021-02-24
Support extracting aar files.

cpsauer:
Hey awesome Bazel team,

Would you consider allowing extract, download_and_extract, and http_archive to decompress .aar android archives? Figured I'd just toss up a PR to ask because it was such an easy addition; aars are just zip files under the hood, totally parallel to the jar/war cases already in the codebase.

The motivation is that you'll need this basic functionality to be able to depend on the native interface of an aar, which Android Studio & co are working on in [prefabs](https://developer.android.com/studio/build/native-dependencies). I ran into the lack of support while trying to hack myself support for https://github.com/bazelbuild/bazel/issues/13092.

Chris

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorValue.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorValue.java
@@ -98,7 +98,8 @@ public class DecompressorValue implements SkyValue {
   static Decompressor getDecompressor(Path archivePath)
       throws RepositoryFunctionException {
     String baseName = archivePath.getBaseName();
-    if (baseName.endsWith(".zip") || baseName.endsWith(".jar") || baseName.endsWith(".war")) {
+    if (baseName.endsWith(".zip")
+        || baseName.endsWith(".jar") || baseName.endsWith(".war") || baseName.endsWith(".aar")) {
       return ZipDecompressor.INSTANCE;
     } else if (baseName.endsWith(".tar")) {
       return TarFunction.INSTANCE;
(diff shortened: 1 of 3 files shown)


ahumesky:
I would say here "probably don't want to do this, probably want to use aar_import", but this might be what we want to do, pending more discussion on #13092

cpsauer:
Hey again, Alex! Happy to do whatever you want here. 

I do think this might be generally useful enough that it's worth decoupling from #13092, even if that was how I originally stumbled onto it. (I'll building my workaround with a shell script regardless to decouple things. Definitely not recommending it as a substitute to aar_import.)

This is handy more generally for rule writers or users looking to depend on subsets of the things in an aar archive. I assume this is why extract() & friends already support decompressing the other java archives, like jars and wars. When I see a function that handles a more general use case but unnecessarily prevents some of some of that usefulness, I get the urge to help it live up to its potential :) Sure it's handy in my workaround, but I PR'd it because I thought it would also serve other use cases.

Some examples of other uses: letting folks depend on just sub-resources of an AAR, supporting the new AAR format changes inside of aar_import, or creating a more robust rule to fix https://github.com/bazelbuild/bazel/issues/348. (Mini digression: Right now Tensorflow & MediaPipe are depending on the genrule workaround posted in that issue. But it doesn't copy transitive java dependencies into the aar, so they've having to filegroup all the Java files into one android_library rather than have a Java build graph. A rule implementing a fix for them would probably also need the ability to unpack APKs, in addition to AARs as I've done here. Happy to add APK extraction at the same time if you'd like.)

cpsauer:
(On tests, is the remote build suite just broken/flaky right now? I see doc-change only PRs failing those tests, but don't want to wrongly ignore them.)

cpsauer:
@ahumesky, any chance we could get this generalization in, or should I close it out/delete it?

cpsauer:
Ran into a couple more instances where this would have been handy while writing extensions. Any chance we could land it? It's free, useful generality.

PR #13101, opened 2021-02-24
Support contraint_values in select()s for alias rules

### LINKED ISSUE #13047: Not possible to select() on constraint_value in alias rules

According to #8583, it should be possible to directly use a `constraint_value` in a `select()` expression. This seems not be the case for `alias` rules though.

```
cc_binary(
    name = "appl",
    srcs = ["main.c"],
)   

constraint_value(
    name = "mycpu",
    constraint_setting = "@platforms//cpu:cpu",
)

alias(
    name = "foo",
    actual = select({
        ":mycpu" : ":appl",
    }),
)
```

When running `bazel build :foo`, the following error is produced:

```
ERROR: /tmp/bzl_alias/BUILD:12:6: //:mycpu is not a valid select() condition for //:foo.
To inspect the select(), run: bazel query --output=build //:foo.
For more help, see https://docs.bazel.build/be/functions.html#select.

ERROR: Analysis of target '//:foo' failed; build aborted: //:mycpu is not a valid select() condition for //:foo.
To inspect the select(), run: bazel query --output=build //:foo.
For more help, see https://docs.bazel.build/be/functions.html#select.

INFO: Elapsed time: 0.077s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 1 target configured)
FAILED: Build did NOT complete successfully (0 packages loaded, 1 target configured)
```

Linux

`release 3.7.0`, `release 4.0.0`

Tried hard, didn't find anything relevant.

I studied the relevant code a little bit, found a comment that might be interesting: https://github.com/bazelbuild/bazel/blob/79989f9becc2edefe8b35f7db687bf8de03e3580/src/main/java/com/google/devtools/build/lib/analysis/config/ConfigConditions.java#L73-L74 

If `targetPlatform` is `null` (which apparently is the case if using `alias`), there's some reason why this doesn't work which I don't fully understand. If this doesn't work by design, then it might be helpful to have this documented.

Workaround for me now is to use a `cc_library` with a `select()` on the `deps` attribute instead of an `alias` rule.

staalb:
Fixes #13047

Tested by running `bazel test //src/test/java/com/google/devtools/build/lib/...`

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/Alias.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/Alias.java
@@ -89,10 +89,6 @@ public class Alias implements RuleConfiguredTargetFactory {
                   .allowedRuleClasses(ANY_RULE)
                   .mandatory())
           .canHaveAnyProvider()
-          // Aliases themselves do not need toolchains or an execution platform, so this is fine.
-          // The actual target will resolve platforms and toolchains with no issues regardless of
-          // this setting.
-          .useToolchainResolution(false)
           .build();
     }
 
--- a/src/test/java/com/google/devtools/build/lib/analysis/ConfigurableAttributesTest.java
+++ b/src/test/java/com/google/devtools/build/lib/analysis/ConfigurableAttributesTest.java
@@ -1267,6 +1267,45 @@ public class ConfigurableAttributesTest extends BuildViewTestCase {
   }
 
   @Test
+  public void selectDirectlyOnConstraintsWithAlias() throws Exception {
+    // Tests select()ing directly on a constraint_value (with no intermediate config_setting).
+    scratch.file(
+        "conditions/BUILD",
+        "constraint_setting(name = 'fruit')",
+        "constraint_value(name = 'apple', constraint_setting = 'fruit')",
+        "constraint_value(name = 'banana', constraint_setting = 'fruit')",
+        "platform(",
+        "    name = 'apple_platform',",
+        "    constraint_values = [':apple'],",
+        ")",
+        "platform(",
+        "    name = 'banana_platform',",
+        "    constraint_values = [':banana'],",
+        ")");
+    scratch.file(
+        "check/BUILD",
+        "filegroup(name = 'adep', srcs = ['afile'])",
+        "filegroup(name = 'bdep', srcs = ['bfile'])",
+        "alias(name = 'hello',",
+        "    actual = select({",
+        "        '//conditions:apple': ':adep',",
+        "        '//conditions:banana': ':bdep',",
+        "    }))");
+    checkRule(
+        "//check:hello",
+        "srcs",
+        ImmutableList.of("--platforms=//conditions:apple_platform"),
+        /*expected:*/ ImmutableList.of("src check/afile"),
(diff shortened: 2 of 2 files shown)


staalb:
This is getting a bit complicated for me. It's hard to determine which tests to run on my system (couldn't find any docs on the subject other than CODEBASE.md). I cannot just run `bazel test //...` on my system, e.g. because I don't have an Android SDK installed.

There are a lot of build and test failures. Didn't have the time yet to go through all of them, but I found these so far:
* Test `//src/test/java/com/google/devtools/build/lib/analysis:AnalysisTests` [fails on BuidKite](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/9a574984-79e1-4c8c-86fb-c80039d06113/src/test/java/com/google/devtools/build/lib/analysis/AnalysisTests/test.log), while it runs just fine on my system. For a system integration test this wouldn't suprise me too much. However, this is a "unit test" which only runs on the analysis phase.
* Test `//src/test/shell/bazel:platform_mapping_test` [fails too](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/02544db0-6a7c-4ed2-b1a8-c92b958674cd/src/test/shell/bazel/platform_mapping_test/attempt_1.log). If I undo the change it passes again. I'll try to investigate this, but I'm afraid I have far too little knowledge of the Bazel codebase... Anyway, I give it a try.

katre:
Okay, took a look at your test failures. I'm actually surprised that this passes anywhere. The underlying problem is the reason why we first exempted `alias` from toolchain resolution: there are aliases used in defining core constraints.

Take a look at [the BUILD file for `@bazel_tools//platforms`](https://cs.opensource.google/bazel/bazel/+/master:tools/platforms/BUILD.tools). You'll see that these constraints are all actually aliases for canonical values from [the `@platforms` repo](https://github.com/bazelbuild/platforms). If `alias` targets required toolchain resolution, you'd see the following:

1. Bazel attempts to analyze a target, say `//java/package:foo`
2. Bazel needs to perform toolchain resolution for that target, so it begins analyzing the target platform. This then includes also analyzing the constraints for that platform.
    1.  Bazel analyzes the platform and constraints in the same configutation as `//java/package:foo`. There's not actually a good reason for this, but we have to analyze them in _some_ configuration, and this one is handy. We've discussed the concept of "null-configured targets" before, but didn't get very far with this because there are a lot of assumptions in Bazel about targets always having a configuration.
3.  When Bazel analyzes one of the constraints that happens to be an `alias`, with your change it now needs to perform toolchain resolution
4. We now have a dependency cycle, because in order to finish toolchain resolution for the original target we've re-invoked toolchain resolution with essentially the same arguments. In your tests this shows up as an unreported dependency cycle, because there's no special handling for this case.

I'm afraid that your change isn't going to be submittable as-is without a lot of underlying changes to how Bazel handles toolchain resolution. I'm going to copy this comment back to the original issue so we can continue discussing potential solutions there.

staalb:
Thanks for looking into the test results and your explanation. From the discusison in the bug report I understand a complete different approach is necessary for this problem, so I'll close this PR.

Issue #12009, opened 2020-08-26
FR: Add option to disable warnings from external C++ headers


devjgm:
I work on https://github.com/googleapis/google-cloud-cpp and we use CMake and bazel. Our project has several external dependencies defined in our [WORKSPACE](https://github.com/googleapis/google-cloud-cpp/blob/134572de299562d071db246d88412876661e3e7a/WORKSPACE#L19) file. This is all working, but we'd like to be able to compile our own code with a high-level of compiler warnings, like `-Wconversion`, and we don't want to be bothered by errors in our external dependencies (because we cannot fix those errors).

In our CMake builds, our deps get "installed", and we include them with angle-brackets and they're found by some `-isystem dir` flags. This works fine with CMake.

However, this doesn't work with Bazel, because bazel encourages all code to be compiled with the same flags, so there's no distinction between my project's code and the code of our external dependencies. The issue is that if I want to compile with `-Wconversion`, I'll be warned about all my deps' problems that I cannot fix.

*WANT*: I'd like bazel to provide some supported mechanism to let users compile their own code different warnings flags than their dependencies.

Can bazel already do this? If so, what's the right way to disable warnings from external deps with Bazel?

We have some ideas that seem to work (maybe) that rely on the `includes = [ "." ]` attribute. We think this might be working, but it also seems like maybe the wrong thing. Is there some other recommended good solution that we should be considering?

devjgm:
Tagging @hicksjoseph

ulfjack:
How do you identify which code is first-party and which is third-party?

devjgm:
I'm not sure who @ulfjack 's question is intended for, but I'll say that any dep that [bazel thinks is an external dep](https://docs.bazel.build/versions/master/external.html#external-packages) should likely be considered third-party.

devjgm:
Any update on this?

oquenchil:
@ulfjack would Greg's suggestion work?

Make anything starting with external be included as /external? Would that mean flattening the headers nested set in analysis?

Regarding when to tackle this Greg, it would help a lot if you can provide a Windows repro.  Also if you could try out replacing /I with /external yourself to verify that that it would work, I could take it from there and check it in the change.

ulfjack:
I think the proposal makes sense. I'm not sure how it would be implemented though. We did something like this in google3, but I'm not sure how it currently works.

derekmauro:
google3 basically says
```
--system-header-prefix=third_party/
[...]

PR #13107, opened 2021-02-25
Add external_include_paths C++ feature

erenon:
Projects with strict warnings level experience issues when they depend on third party libraries with less-strict warnings: The compiler will emit warnings for external sources. See #12009 for more.

This change optionally silences those warnings, by changing -I flags to -isystem flags and by adding -isystem flags for each -iquote flag, in case of external dependencies, i.e: those targets that come from a non-main workspace.

The new behavior can be enabled by --features=external_include_paths, otherwise there's no functional change.

The default flag_group in unix_cc_toolchain_config will silence warnings coming from -I and -iquote dirs in case of GCC, and -I for Clang. Clang will still produce warnings for -iquote dirs, therefore the Clang specific --system-header-prefix is recommended instead.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationContext.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationContext.java
@@ -313,6 +313,17 @@ public final class CcCompilationContext implements CcCompilationContextApi<Artif
   }
 
   /**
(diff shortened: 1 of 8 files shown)


oquenchil (approved):
Thank you!

LGTM

erenon:
Gentel reminder.

aiuto:
@oquenchil  Are you going to approve and merge?

PR #13110, opened 2021-02-25
Support execution constraints per exec group

quval:
When computing exec properties from the execution platform for an action, take into account only the properties that are relevant to the action's exec groups. In particular, allow setting exec properties for arbitrary exec groups on platforms. Previously, any such properties were rejected.

With this change, the following becomes possible:
```
cc_test(
    name = "my_test",
    ...,
    exec_properties = {
        "test.key": "value",
    },
)
```
This will apply `{"key": "value"}` for the test-runner action only (i.e., compilation and linkage won't be affected). The following also becomes possible:
```
platform(
    name = "test_platform",
    constraint_values = [":test_constraint"],
    exec_properties = {
        "test.key": "value",
    },
)

cc_test(
    name = "my_test",
    ...,
    exec_compatible_with = [":test_constraint"],
)
```
This achieves the same in a more succinct way.

For related discussion, see PR #12719 by @ulfjack.


katre:
I am rolling forward the test exec change separately (https://github.com/bazelbuild/bazel/pull/13119) so it can be separated from this. Once that is in place, can you rebase and update the commits to make the CLA bot happy?

quval:
Sure, will do (in fact, I believe this PR only includes those changes, but I originally based it on #12719).

quval:
@katre: rebased and double-checked everything's in order. Thanks!

katre:
Thanks! I've taken a quick look at this and it seems okay, but I want to think about it some more. One thing that would really help would be if you could add a test case to https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/toolchain_test.sh: this will a) test the code, and b) show how it's intended to be used.

It would also be useful to add tests to https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/analysis/StarlarkExecGroupTest.java

quval:
Great, will do. I've been wondering where would tests actually go. Thank you!

quval:
Added a couple of test cases. I think that just about covers it, but do let me know if you think any more could be helpful.

src/test/shell/bazel/platforms_test.sh
```diff
@@ -321,7 +321,7 @@ EOF
   bazel build --extra_execution_platforms=":my_platform" :a --execution_log_json_file out.txt || fail "Build failed"
   grep "platform_key" out.txt || fail "Did not find the platform key"
   grep "default_value" out.txt || fail "Did not find the default value"
-  grep "test_value" out.txt && fail "Used the test-action value when not testing"
+  (! grep "test_value" out.txt) || fail "Used the test-action value when not testing"
```
katre, line 324: I didn't know this syntax existed. I see the benefits, but it's very different than all the other tests, so can you please change this back to the previous style, and use the same style in your new tests? I think consistency is better in this case.

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
```diff
@@ -1390,19 +1411,18 @@ public boolean targetPlatformHasConstraint(ConstraintValueInfo constraintValue)
...
    * The raw map can have keys of two forms: (1) 'property' and (2) 'exec_group_name.property'. The
-   * former get parsed into the target's default exec group, the latter get parsed into their
-   * relevant exec groups.
+   * former get parsed into the default exec group, the latter get parsed into their relevant exec
+   * groups.
    */
-  private static ImmutableMap<String, ImmutableMap<String, String>> parseExecProperties(
-      Map<String, String> rawExecProperties, Set<String> execGroups)
-      throws InvalidExecGroupException {
+  private static Map<String, Map<String, String>> parseExecGroups(
```
katre, line 1424: Wow, I hate Map<?, Map<?, ?>> so much, but I see this predates your change so leave it. I will probably send a followup shortly to change this to a [Table](https://github.com/google/guava/wiki/NewCollectionTypesExplained#table).

katre (changes requested):
Looks good, just a few comments and then this will be ready to merge. Thanks for working with me to get this in shape!

src/test/shell/bazel/platforms_test.sh
quval: Done. I couldn't figure out why, but one of the tests failed with that style (and I missed the test case with expected failure). I ended up removing it, as there is a positive test anyway.

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
quval: Acknowledged. I didn't know about Table (never wrote Java at Google). I would be happy to fix this - please let me know if you'd like me to do this as part of this PR - but you probably have more context.

quval:
Thanks for the review! It's been a while since I made this change, and initially I intended it as a sketch - glad it can be merged.

katre (approved):

katre:
I'm merging this today. Thanks for your patience!

The internal reviewer did have one request: can you send a separate PR to update the docs for exec groups (at https://github.com/bazelbuild/bazel/blob/master/site/docs/exec-groups.md) to mention the new functionality?

quval:
Opened #13167 for this. Thank you for your help getting this in!

djmarcin:
Does this only work for test actions of native rules? I'm noticing that [rust_test rules](https://github.com/bazelbuild/rules_rust/blob/a0104fb64f7f8f5f63c2eb93b134cf33a5f8489a/rust/private/rust.bzl#L926-L939) from bazelbuild/rules_rust do not get `test.foo` attributes set on their executions, while native rules like `sh_test` seem to work correctly. Since test rules just return an executable, I'm not sure how a rule author would go about adding the test exec group to the test runner action.

katre:
Hmm, all test rules should get the "test" exec group added, but I'm not sure I've ever tested that with Starlark rules. Can you open a separate issue and ping me?

quval:
@katre @djmarcin This isn't instead of a new issue, and it's probably better if you take a look, but I've actually taken a look yesterday and thought I might share my findings - looks like [this line](https://github.com/bazelbuild/bazel/commit/0b3a5aad5818dd5bc1fd3c6b5c3ada77de85f328#diff-eaad93611c778efb29a5090f230db8be840c0351665e1bfbba41b4c95b2a8199R160) is missing [here](https://github.com/bazelbuild/bazel/blob/d5fd56e0172ceede128cc9f9604307ee16619727/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java#L183) (or possibly better, [here](https://github.com/bazelbuild/bazel/blob/d5fd56e0172ceede128cc9f9604307ee16619727/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java#L394-L404), to prevent breaking those who have already added a test exec group manually). I started preparing a PR, but for some reason couldn't get the tests to pass even though a manual run with the patched Bazel did work correctly.

katre:
This is a pull request, so I'd prefer to have an issue I can track and attach pull requests to. Unless there's another issue you think already fits this?

quval:
No, I was just sharing my thoughts in the meantime until @djmarcin opens an issue.

PR #13116, opened 2021-02-26
Change package name from bazel3 to bazel4

zregvart:


aiuto:
This was done in #13255

Issue #6761, opened 2018-11-26
incompatible_merge_genfiles_directory: unify bazel-bin and bazel-genfiles directories


laurentlb:
We plan to unify the `bazel-bin` and `bazel-genfiles` directories.

This means that `bazel-genfiles` will point to the same location as `bazel-bin`, since the distinction was not meaningful. This will allow us to make simplifications in Bazel and its API.

See discussion: https://groups.google.com/d/msg/bazel-dev/5iq7n1soenA/xGM_M1NOCAAJ

laurentlb:
Still blocked by Protobuf (https://github.com/protocolbuffers/protobuf/issues/5691).

meteorcloudy:
@laurentlb FYI, Bazelisk + Incompatible flags couldn't catch this breakage. Because building with this flag seems have an impact on subsequent builds.
See https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/18#d6d93bbf-ca8a-432f-bd99-23d93577ede5
Unless we do a `bazel clean` between the incompatible flags and no incompatible flags build, we still get the same error message.

jin:
This will cause all of these scripts / code that assume existence of `bazel-genfiles`, right? https://github.com/search?q=bazel-genfiles&type=Code

laurentlb:
When testing downstream projects, we noticed very few issues. But it's true that it tests only within Bazel. Scripts that call Bazel and then look at the specific directory might break.

Not sure what we can do, apart from advertising the change more loudly.
(at least, it shows why this should be done before Bazel 1.0)

laurentlb:
Projects that rely on `bazel-genfiles` can:
 - Specify the flag `--incompatible_merge_genfiles_directory=false`. This flag will be available for a long time (even after Bazel 1.0), to give users more time to migrate.
 - Create a symlink: `ln -s bazel-bin bazel-genfiles`

jin:
> When testing downstream projects, we noticed very few issues. But it's true that it tests only within Bazel. Scripts that call Bazel and then look at the specific directory might break.

The symlinks are, in a way, the only user facing filesystem level API that Bazel exposes. I also have no idea what else we can do. The Downstream pipeline has no visibility into the bazel-external build infra of our users :(

> advertising the change more loudly. (at least, it shows why this should be done before Bazel 1.0)

Completely agree.

PR #13126, opened 2021-02-28
Prevent --repo_env from triggering unnecessary fetches

philsc:
It looks like `--repo_env` entries get added to every repository
rule's marker data. That is regardless of whether the repository rule
declared that it uses the environment variable. This causes
unnecessary fetches of the external repositories.

This patch aims to fix that by making repository rules only depend on
`--repo_env` values that they specify in their `environ` attributes.

I also took this opportunity to fix up all instances of `bazel-genfiles`
in `starlark_repository_test.sh` to `bazel-bin`.

### DIFF
--- a/src/test/shell/bazel/starlark_repository_test.sh
+++ b/src/test/shell/bazel/starlark_repository_test.sh
@@ -120,7 +120,7 @@ EOF
   bazel build //zoo:ball-pit1 >& $TEST_log || fail "Failed to build"
   expect_log "bleh"
   expect_log "Tra-la!"  # Invalidation
-  cat bazel-genfiles/zoo/ball-pit1.txt >$TEST_log
+  cat bazel-bin/zoo/ball-pit1.txt >$TEST_log
   expect_log "Tra-la!"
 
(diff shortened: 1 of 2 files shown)


philsc:
@philwo , any thoughts on this?

src/test/shell/bazel/starlark_repository_test.sh
```diff
@@ -926,6 +926,49 @@ EOF
...
+  cmd = "cp $< $@",
+)
+EOF
+  cat > .bazelrc <<EOF
+build:foo --repo_env=FOO=foo
+build:bar --repo_env=FOO=bar
+EOF
+
+  bazel build --config=foo //:repoenv
+  cp `bazel info bazel-genfiles 2>/dev/null`/repoenv.txt repoenv1.txt
```
meteorcloudy, line 958: `bazel-genfiles` is deprecated, please just use `bazel-bin`. 
See https://github.com/bazelbuild/bazel/issues/6761

philsc: Good point.

Done

PR #13141, opened 2021-03-02
Remove double bazel_tools

illicitonion:


aiuto (approved):

aiuto:
I am unable to merge this.  The failure is in our pipeline that syncs github and google sources. The fix requires someone who understands our copybara config a little better.  I'm passing this off to @comius because it looks like they recently added what might be the culprit.

PR #13162, opened 2021-03-05
Use a dummy toolchain context for rules that don't have one.

### LINKED ISSUE #12610: Starlark build setting invoking go_context throws an exception

Bazel fails with `java.lang.IllegalStateException: method invocation returned null: toolchains()` when invoking `rules_go`'s `go_context` from a Starlark build setting rule.

I have created a minimal example as [this gist](https://gist.github.com/fmeum/86bd43948210ba2fc9fb64f6a3a3f6af). If the line `go_context(ctx)` is replaced with `pass`, the build succeeds without an exception. I suspect that `rules_go` is not really needed to trigger the exception here, but don't know enough about toolchains to minimize this example further.

I found this example while trying to create a config setting that would automatically generate code based on its current value.

Linux

`release 3.7.1`

Unfortunately not.

See https://gist.github.com/fmeum/86bd43948210ba2fc9fb64f6a3a3f6af#file-trace for the trace.

katre:
Fixes #12610.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
@@ -76,6 +76,7 @@ import com.google.devtools.build.lib.packages.semantics.BuildLanguageOptions;
 import com.google.devtools.build.lib.shell.ShellUtils;
 import com.google.devtools.build.lib.shell.ShellUtils.TokenizationException;
 import com.google.devtools.build.lib.starlarkbuildapi.StarlarkRuleContextApi;
+import com.google.devtools.build.lib.starlarkbuildapi.platform.ToolchainContextApi;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -701,9 +702,24 @@ public final class StarlarkRuleContext implements StarlarkRuleContextApi<Constra
   }
 
   @Override
-  public ResolvedToolchainContext toolchains() throws EvalException {
+  public ToolchainContextApi toolchains() throws EvalException {
     checkMutable("toolchains");
-    return ruleContext.getToolchainContext();
+    ResolvedToolchainContext toolchainContext = ruleContext.getToolchainContext();
+    if (toolchainContext == null) {
+      // Starlark rules are easier if this cannot be null, so return a no-op value instead.
+      return new ToolchainContextApi() {
+        @Override
+        public Object getIndex(StarlarkSemantics semantics, Object key) throws EvalException {
+          return Starlark.NONE;
+        }
+
+        @Override
+        public boolean containsKey(StarlarkSemantics semantics, Object key) throws EvalException {
+          return false;
+        }
+      };
+    }
+    return toolchainContext;
   }
 
   @Override
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
@@ -709,12 +709,12 @@ public final class StarlarkRuleContext implements StarlarkRuleContextApi<Constra
       // Starlark rules are easier if this cannot be null, so return a no-op value instead.
       return new ToolchainContextApi() {
         @Override
-        public Object getIndex(StarlarkSemantics semantics, Object key) throws EvalException {
+        public Object getIndex(StarlarkSemantics semantics, Object key) {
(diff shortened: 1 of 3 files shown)


src/test/shell/bazel/toolchain_test.sh
```diff
@@ -2034,6 +2034,35 @@ EOF
   expect_not_log "does not provide ToolchainTypeInfo"
 }
 
+function test_toolchain_from_no_toolchain_rule() {
+  # Create a build_setting that tries to load a toolchain.
+  # build_settings cnanot actually have toolchains because they
```
brandjon, line 2039: typo

src/test/shell/bazel/toolchain_test.sh
```diff
@@ -2034,6 +2034,35 @@ EOF
...
+toolchain_type(name = "toolchain_type")
+sample_setting(
+    name = "demo",
+    build_setting_default = True,
+)
+EOF
+  cat >> demo/rule.bzl <<EOF
+def _sample_impl(ctx):
+    info = ctx.toolchains["//:toolchain_type"]
+    if info:
```
brandjon, line 2054: `if info != None:`? To distinguish from other false-y values.

src/test/shell/bazel/toolchain_test.sh
```diff
@@ -2034,6 +2034,35 @@ EOF
   expect_not_log "does not provide ToolchainTypeInfo"
 }
 
+function test_toolchain_from_no_toolchain_rule() {
```
brandjon, line 2037: What's the purpose of testing this at both the unit test and shell test levels?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
```diff
@@ -701,9 +702,24 @@ public StarlarkAttributesCollection rule() throws EvalException {
...
-  public ResolvedToolchainContext toolchains() throws EvalException {
+  public ToolchainContextApi toolchains() throws EvalException {
     checkMutable("toolchains");
-    return ruleContext.getToolchainContext();
+    ResolvedToolchainContext toolchainContext = ruleContext.getToolchainContext();
+    if (toolchainContext == null) {
+      // Starlark rules are easier if this cannot be null, so return a no-op value instead.
+      return new ToolchainContextApi() {
+        @Override
+        public Object getIndex(StarlarkSemantics semantics, Object key) throws EvalException {
```
brandjon, line 712: Nit: You can omit the throws declaration since this override doesn't throw. Same below.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
```diff
@@ -701,9 +702,24 @@ public StarlarkAttributesCollection rule() throws EvalException {
...
 
   @Override
-  public ResolvedToolchainContext toolchains() throws EvalException {
+  public ToolchainContextApi toolchains() throws EvalException {
     checkMutable("toolchains");
-    return ruleContext.getToolchainContext();
+    ResolvedToolchainContext toolchainContext = ruleContext.getToolchainContext();
+    if (toolchainContext == null) {
+      // Starlark rules are easier if this cannot be null, so return a no-op value instead.
+      return new ToolchainContextApi() {
```
brandjon, line 710: Optionally make this a singleton, but I doubt it'll matter much.

brandjon (approved):

src/test/shell/bazel/toolchain_test.sh
katre: Removed entirely.

src/test/shell/bazel/toolchain_test.sh
katre: Fixed in the unit test.

src/test/shell/bazel/toolchain_test.sh
katre: Good point, I wrote this first then added the other to debug, but we only need one. Removed the shell test.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
katre: Fixed.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleContext.java
katre: I don't think there's enough benefit, this is fairly clear.

PR #13164, opened 2021-03-05
Clean up RuleContext to use a Table instead of a Map of Maps.

katre:


katre:
@quval: This is what I meant by using Table instead.

gregestren (approved):

Issue #12972, opened 2021-02-06
Bazel occasionally crashes after a cancelled task on dynamic builds


tomrenn:
### Description of the problem / feature request:

When running `bazel build`, Bazel crashes with the following stack trace

```
 FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'REDACTED'
at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.util.concurrent.CancellationException: Task was cancelled.
	at com.google.common.util.concurrent.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1386)
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:562)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:545)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:102)
	at com.google.devtools.build.lib.remote.util.Utils.getFromFuture(Utils.java:77)
	at com.google.devtools.build.lib.remote.RemoteCache.waitForBulkTransfer(RemoteCache.java:221)
	at com.google.devtools.build.lib.remote.RemoteActionInputFetcher.prefetchFiles(RemoteActionInputFetcher.java:121)
	at com.google.devtools.build.lib.exec.AbstractSpawnStrategy$SpawnExecutionContextImpl.prefetchInputs(AbstractSpawnStrategy.java:232)
	at com.google.devtools.build.lib.worker.WorkerSpawnRunner.execInWorker(WorkerSpawnRunner.java:412)
	at com.google.devtools.build.lib.worker.WorkerSpawnRunner.actuallyExec(WorkerSpawnRunner.java:227)
	at com.google.devtools.build.lib.worker.WorkerSpawnRunner.exec(WorkerSpawnRunner.java:144)
[...]

PR #13166, opened 2021-03-06
Add a blob existence cache to RemoteExecutionCache

ulfjack:
Bazel currently performs a full findMissingBlob call for every input of
every action that is not an immediate cache hit. This causes a
significant amount of network traffic, resulting in long build times on
low-bandwidth, high-latency connections (i.e., work from home).

In particular, for a build of TensorFlow w/ remote execution over a
slow-ish network, we see multi-minute delays on actions due to these
findMissingBlob calls, and the client is not able to saturate a
50-machine remote execution cluster.

This change carefully de-duplicates findMissingBlob calls as well as
file uploads to the remote execution cluster.

It is based on a previous simpler change that did not de-duplicate
*concurrent* calls. However, we have found that concurrent calls with
the same inputs are common - i.e., an action completes that unblocks a
large number of actions that have significant overlaps in their input
sets (e.g., protoc link completion unblocks all protoc compile
actions), and we were still seeing multi-minute delays on action
execution.

With this change, a TensorFlow build w/ remote execution is down to ~20
minutes on a 50-machine cluster, and is able to fully saturate the
cluster.

We have also seen significant improvements for remote Android app
builds.

Change-Id: Ic39347a7a7a8dc7cfd463d78f0a80e0d26a970bc

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
 import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext;
 import com.google.devtools.build.lib.remote.common.RemoteCacheClient;
 import com.google.devtools.build.lib.remote.merkletree.MerkleTree;
@@ -31,11 +32,15 @@ import com.google.devtools.build.lib.remote.util.DigestUtil;
 import com.google.protobuf.Message;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** A {@link RemoteCache} with additional functionality needed for remote execution. */
 public class RemoteExecutionCache extends RemoteCache {
+  private final ConcurrentHashMap<Digest, ListenableFuture<Void>> uploadFutures =
+      new ConcurrentHashMap<>();
 
   public RemoteExecutionCache(
       RemoteCacheClient protocolImpl, RemoteOptions options, DigestUtil digestUtil) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
```diff
@@ -53,23 +58,92 @@ public RemoteExecutionCache(
...
+    List<ListenableFuture<Void>> futures = new ArrayList<>();
+
+    if (checkAll || !options.experimentalRemoteDeduplicateUploads) {
+      ImmutableSet<Digest> missingDigests =
+          getFromFuture(cacheProtocol.findMissingDigests(context, allDigests));
+      for (Digest missingDigest : missingDigests) {
+        futures.add(uploadBlob(context, missingDigest, merkleTree, additionalInputs));
+      }
+    } else {
+      // This code deduplicates findMissingDigests calls as well as file uploads. It works by
```
coeuvre, line 82: Why do we need to distinguish owned and unowned futures?

I think we are doing too much here. Can we have a cache layer between `RemoteExecutionCache` and `RemoteCacheClient` which is used to deduplicate the uploads and owns the futures. All the futures are unowned from the perspective of `RemoteExecutionCache`. e.g.

```
class UploadCache {
    ListenableFuture<Void> upload(MerkleTree merkleTree, Map<Digest, Message> additionalInputs);
}
```

We can also hide the `findMissingDigests` call inside `UploadCache` and can potentially save the call if the requested uploads are already uploaded or are in progress.

[AsyncTaskCache](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/util/AsyncTaskCache.java;l=50;bpv=1;bpt=1?q=Async&ss=bazel%2Fbazel) which is added recently to fix #12972 can be used to implement the cache.

FYI, we have the deduplication logic at [ByteStreamUploader](https://cs.opensource.google/bazel/bazel/+/refs/tags/4.0.0:src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java;l=209?q=ByteStreamUploader&ss=bazel%2Fbazel) but isn't enabled since the `forceUpload` is set to `true` [here](https://cs.opensource.google/bazel/bazel/+/refs/tags/4.0.0:src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java;l=400). That being said, the deduplication is only available for gRPC cache client. This PR allows deduplication for all cache protocols.

src/main/java/com/google/devtools/build/lib/remote/RemoteRepositoryRemoteExecutor.java
```diff
@@ -132,7 +132,7 @@ public ExecutionResult execute(
         additionalInputs.put(actionDigest, action);
         additionalInputs.put(commandHash, command);
 
-        remoteCache.ensureInputsPresent(context, merkleTree, additionalInputs);
+        remoteCache.ensureInputsPresent(context, merkleTree, additionalInputs, true);
```
coeuvre, line 135: Why set `true` here?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -483,6 +483,17 @@ public RemoteOutputsStrategyConverter() {
...
+  @Option(
+          name = "experimental_remote_deduplicate_uploads",
+          defaultValue = "false",
+          category = "remote",
+          documentationCategory = OptionDocumentationCategory.REMOTE,
+          effectTags = {OptionEffectTag.EXECUTION},
+          help =
+              "If set to true, Bazel deduplicates calls to the remote service to find missing "
+                  + "files and also deduplicates uploads.")
+  public boolean experimentalRemoteDeduplicateUploads;
```
coeuvre, line 495: We have too much flags now. Are there concerns to introduce this flag and disable the feature by default? I think people always want this enabled.

coeuvre (commented):
Thanks for the PR!

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionCache.java
ulrfa: > FYI, we have the deduplication logic at [ByteStreamUploader](https://cs.opensource.google/bazel/bazel/+/refs/tags/4.0.0:src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java;l=209?q=ByteStreamUploader&ss=bazel%2Fbazel) but isn't enabled since the `forceUpload` is set to `true` [here](https://cs.opensource.google/bazel/bazel/+/refs/tags/4.0.0:src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java;l=400). ~That being said, the deduplication is only available for gRPC cache client. This PR allows deduplication for all cache protocols.~ (This is only available for remote execution which only supports gRPC protocol)

I’m prototyping remote execution support also with HTTP cache protocol. Because that allows using HTTP load balance infrastructure. I get performance on par with gRPC. Therefore **I very much appreciate that this PR is generic and works also for HTTP**.

ulfjack: I don't think AsyncTaskCache would be ideal here. The problem is that the `findMissingBlobs` calls are *bulk* calls: we have to deduplicate the individual entries, but we still want to perform bulk calls for performance. That requires passing a collection in and out of this class.

I think ByteStreamUploader is not a good place for a cache. The API to check for CAS entries is `findMissingBlobs`, which is a bulk API. ByteStreamUploader has a single file upload API - it can't really automatically turn a sequence of single-file uploads into a bulk check call. The caching has to be outside of ByteStreamUploader and itself needs a bulk API to be able to efficiently call `findMissingBlobs`. I also don't like that it unconditionally caches knowledge about an upload for the duration of the build - if the CAS loses entries faster than the build time, Bazel gets confused and fails for no good reason.

The distinction between owned and unowned futures is thus: there is any number of concurrent calls into this method, and we own some of the digests, but not all. We then make a bulk `findMissingBlobs` call for all owned digests, and then upload the ones that are missing remotely.

I'm happy to remove the flag and always enable this code.

coeuvre: Thanks for the clarification. I am now agree with you that ByteStreamUploader can not be used to deduplicate `findMissingBlobs` calls.

**Feel free to ignore**: AsyncTaskCache can be used to deduplicate uploads just like what you achieved with ConcurrentHashMap combined with deduplication logic.

> I'm happy to remove the flag and always enable this code.

Please remove the flag!

Can you please also add test cases for the deduplication?

PR #13167, opened 2021-03-06
Documentation for #13110

quval:
@katre

### DIFF
--- a/site/docs/exec-groups.md
+++ b/site/docs/exec-groups.md
@@ -132,6 +132,14 @@ All actions with `exec_group = "link"` would see the exec properties
 dictionary as `{"mem": "16g"}`. As you see here, execution-group-level
 settings override target-level settings.
 
+### Execution groups for native rules
+
+Currently, the following execution groups are available for actions defined by
+native rules:
+
+* `test`: Test runner actions.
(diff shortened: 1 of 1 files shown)


site/docs/exec-groups.md
```diff
@@ -162,3 +170,36 @@ my_rule = rule(
 #
 ```
 
+### Execution groups and platform execution properties
+
+It is possible to define `exec_properties` for arbitrary execution groups on
+platform targets (unlike `exec_properties` set directly on a target, where
+properties for unknown execution groups are rejected). Targets then inherit the
+platform's `exec_properties` that affect the default execution group and any
```
katre, line 178: "Targets then inherit the execution platform's `exec_proeprties`"

katre (changes requested):
Thanks! I left one comment, but in general can you please clarify that this is coming from execution platforms, not target platforms?

site/docs/exec-groups.md
quval: Done (here and in one other place) - not sure if it should be made clearer.

katre (approved):
I'll merge this today.

quval:
Thank you!

PR #13168, opened 2021-03-06
Cleanup: Replace NULL with nullptr

Vertexwahn:
At several places, NULL and nullptr were mixed. This commit cleans this up and switches NULL to nullptr where appropriate.

### DIFF
--- a/examples/windows/dll/hello_world-load-dll-at-runtime.cpp
+++ b/examples/windows/dll/hello_world-load-dll-at-runtime.cpp
@@ -12,11 +12,11 @@ int main() {
 
   hellolib = LoadLibrary(TEXT("hellolib.dll"));
 
-  if (hellolib != NULL) {
+  if (hellolib != nullptr) {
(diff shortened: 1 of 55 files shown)


brandjon:
Philipp, do you know who we can triage this to? Looks fairly simple but I'm not a C++ person or familiar with our C++ code.

philwo:
@brandjon I'll try to import this, maybe it'll just go through without issues. :)

philwo:
@Vertexwahn Thank you for this. The presubmit tests are running at the moment and when they pass, it'll be merged automatically. :)

Just FYI, while importing I noticed a few more lines in `process.cc` that still used NULL (e.g. https://cs.opensource.google/bazel/bazel/+/c789d78fa2a3fb760e99963bac5000f49747cc88:src/main/native/windows/process.cc;l=175-178) and I fixed them to use `nullptr` instead, too. Hope that was fine from your point of view.

PR #13170, opened 2021-03-07
Expand target_compatible_with to all non-workspace rules

### LINKED ISSUE #12745: Make target_compatible_with available also for cc_import

target_compatible_with is not available for cc_import

The same logic that one would like to apply to a cc_library applies to a cc_import

Build label: 4.0.0rc7

philsc:
This patch makes it so the `target_compatible_with` attribute can be
used with all non-workspace rules. The motivation here is to allow
rules like `filegroup` and `cc_import` to also make use of
incompatible target skipping. Even though these rules don't
inherently generate files that are incompatible with the target
platform, they can be used to provide incompatible files to other
targets. Before this patch the user had to do something like this to
create an incompatible pre-compiled library:

    cc_import(
        name = "foo_import",
        shared_library = "foo.so",
    )

    cc_library(
        name = "foo",
        deps = [":foo_import"],
        target_compatible_with = ["//:foo_platform"],
    )

Now users can directly declare compatibility on the `cc_import` rule.

    cc_import(
        name = "foo_import",
        shared_library = "foo.so",
        target_compatible_with = ["//:foo_platform"],
    )

Unfortunately, This does mean that some rules like `cc_toolchain` now
have a `target_compatible_with` that doesn't serve any purpose.

Before the changes in `BaseRulesClasses.java` the test would error out
with meesages like this:

    //target_skipping:some_precompiled_library: no such attribute 'target_compatible_with' in 'cc_import' rule
    //target_skipping:filegroup: no such attribute 'target_compatible_with' in 'filegroup' rule

I also took this opportunity to update the documentation a bit.

Fixes #12745

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
@@ -907,8 +907,11 @@ public class RuleContextConstraintSemantics implements ConstraintSemantics<RuleC
       RuleContext ruleContext,
       OrderedSetMultimap<DependencyKind, ConfiguredTargetAndData> prerequisiteMap)
       throws ActionConflictException, InterruptedException {
-    // The target (ruleContext) is incompatible if explicitly specified to be.
-    if (ruleContext.getRule().getRuleClassObject().useToolchainResolution()
+    // The target (ruleContext) is incompatible if explicitly specified to be. Any rule that has
+    // provides its own meaning for the "target_compatible_with" attribute has to be excluded here.
+    // For example, the "toolchain" rule uses "target_compatible_with" for bazel's toolchain
+    // resolution.
+    if (!ruleContext.getRule().getRuleClass().equals("toolchain")
         && ruleContext.attributes().has("target_compatible_with")) {
       ImmutableList<ConstraintValueInfo> invalidConstraintValues =
           PlatformProviderUtils.constraintValues(
--- a/src/test/shell/integration/target_compatible_with_test.sh
+++ b/src/test/shell/integration/target_compatible_with_test.sh
@@ -299,6 +299,59 @@ EOF
   expect_log 'Target //target_skipping:filegroup is incompatible and cannot be built'
 }
 
+# Validates that filegroups and other rules that don't inherit from
+# `NativeActionCreatingRule` can be marked with target_compatible_with. This is
+# a regression test for https://github.com/bazelbuild/bazel/issues/12745.
+function test_skipping_for_rules_that_dont_create_actions() {
+  # Create a fake shared library for cc_import.
+  echo > target_skipping/some_precompiled_library.so
+
+  cat >> target_skipping/BUILD <<EOF
+cc_import(
+    name = "some_precompiled_library",
+    shared_library = "some_precompiled_library.so",
+    target_compatible_with = [
+        ":foo3",
+    ],
+)
+
+cc_binary(
+    name = "some_binary",
+    deps = [
+        ":some_precompiled_library",
+    ],
+)
+
+filegroup(
+    name = "filegroup",
+    srcs = [
+        "some_precompiled_library.so",
+    ],
+    target_compatible_with = [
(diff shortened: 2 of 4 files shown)


philsc:
@gregestren , @katre , FYI

katre (approved):
This looks good to me, I'll wait for Greg to review.

katre:
Merging this now.

philsc:
Oh boy. Looks like I was a tad careless.
Thanks a lot for investigating!
I'll revisit this tomorrow most likely.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
```diff
@@ -908,7 +908,7 @@ public static ConfiguredTarget incompatibleConfiguredTarget(
       OrderedSetMultimap<DependencyKind, ConfiguredTargetAndData> prerequisiteMap)
       throws ActionConflictException, InterruptedException {
     // The target (ruleContext) is incompatible if explicitly specified to be.
-    if (ruleContext.getRule().getRuleClassObject().useToolchainResolution()
+    if (ruleContext.getRule().getRuleClass() != "toolchain"
```
philsc, line 911: WDYT about this, @katre ? It's not very elegant, but I can't find anything cleaner that expresses the same thing.

katre: I think this makes sense, but it'd be good to add a comment with the general definition of `target_compatible_with` that any rule that needs its own meaning for `target_compatible_with` needs to be added to this list.

Also, make sure you're checking String equality with `String.equals`, not `==`.

katre (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
philsc: Dang. Every time I feel like I'm getting the hang of Java there's something else that trips me up :P

I fixed the string comparison. Thanks!

By "a comment with the general definition of `target_compatible_with`" do you mean updating `attributes/common/target_compatible_with.html`?  For now I added a comment for this right here in the code.

katre: Sorry, I meant adding a comment to BaseRuleClasses.NativeBuildRule.

philsc: Gotcha, makes sense. Thanks!

Done

src/test/shell/integration/target_compatible_with_test.sh
```diff
@@ -299,6 +299,59 @@ EOF
   expect_log 'Target //target_skipping:filegroup is incompatible and cannot be built'
 }
 
+# Validates that filegroups and other rules that don't inherit from
+# `NativeActionCreatingRule` can be marked with target_compatible_with. This is
```
katre, line 303: The `target_compatible_with` attribute is now on NativeBuildRule, not NativeActionCreatingRule.

And as soon as I submit #13223, filegroup will again be part of toolchain resolution (see the discussion on that PR about why).

My conclusion from #13223 is that there are no rules that are actual build rules and disable toolchain resolution, so I can't think of a rule to use in this test.

philsc: I'm not really sure what you're looking for here.

Without this patch here, rules that don't inherit from `NativeActionCreatingRule` cannot use the `target_compatible_with` attribute. So this test will fail. With the patch, the desired functionality is implemented and the test starts passing.

katre: Sorry, I think I misunderstood (because I was still confused from the test I removed in #13223). I thought this test was relying on filegroup not doing toolchain resolution, but now I see that isn't the case. My bad!

katre (approved):

katre:
Trying to import this internally now.

PR #13171, opened 2021-03-07
Update documentation on //... and //:all

philsc:
The intent of this patch is to be a little more explicit about these
two patterns. It appears that there currently isn't a spot that
explicitly documents these patterns.

I also decided to update the examples for "Target names" and "Package
names" because the `...` syntax does have a special meaning. I thought
it best to avoid confusion for those cases.

### DIFF
--- a/site/docs/build-ref.html
+++ b/site/docs/build-ref.html
@@ -293,7 +293,7 @@ app_binary
   The precise details of allowed target names are below.
 </p>
 
-<h4 id="name">Target names, <code>//...:<b>target-name</b></code></h4>
+<h4 id="name">Target names, <code>//&lt;package name&gt;:<b>target-name</b></code></h4>
 
 <p><code>target-name</code> is the name of the target within the package.
   The name of a rule is the value of the <code>name</code>
@@ -331,7 +331,7 @@ app_binary
   certain rules must match their principal source file, which may
   reside in a subdirectory of the package.</p>
 
-<h4>Package names, <code>//<b>package-name</b>:...</code></h4>
+<h4>Package names, <code>//<b>package-name</b>:&lt;target name&gt;</code></h4>
 <p>
   The name of a package is the name of the directory containing its
 
--- a/site/docs/guide.md
+++ b/site/docs/guide.md
@@ -139,15 +139,15 @@ workspace.
 </tr>
 <tr>
   <td><code>//foo/bar:all</code></td>
-  <td>All rules in the package <code>foo/bar</code>.</td>
+  <td>All rule targets in the package <code>foo/bar</code>.</td>
 </tr>
 <tr>
   <td><code>//foo/...</code></td>
-  <td>All rules in all packages beneath the directory <code>foo</code>.</td>
+  <td>All rule targets in all packages beneath the directory <code>foo</code>.</td>
 </tr>
 <tr>
   <td><code>//foo/...:all</code></td>
-  <td>All rules in all packages beneath the directory <code>foo</code>.</td>
+  <td>All rule targets in all packages beneath the directory <code>foo</code>.</td>
 </tr>
 <tr>
   <td><code>//foo/...:*</code></td>
@@ -156,6 +156,16 @@ workspace.
 <tr>
   <td><code>//foo/...:all-targets</code></td>
   <td>All targets (rules and files) in all packages beneath the directory <code>foo</code>.</td>
+</tr>
+<tr>
+  <td><code>//...</code></td>
+  <td>All targets in packages in the workspace. This does not include targets 
+  from a href="external.html">external repositories</a>.</td>
+</tr>
+<tr>
+  <td><code>//:all</code></td>
+  <td>All targets in the top-level package, if there is a BUILD file at the 
+  root of the workspace.</td>
 </tr>
 </table>
 


site/docs/build-ref.html
```diff
@@ -291,7 +291,7 @@ <h3 id="lexi">Lexical specification of a label</h3>
   The precise details of allowed target names are below.
 </p>
 
-<h4 id="name">Target names, <code>//...:<b>target-name</b></code></h4>
+<h4 id="name">Target names, <code>//xxx:<b>target-name</b></code></h4>
```
gregestren, line 294: It took me some time to understand what's being communicated . But now I get it. This section says "ignore the package name, focus on target-name". The next section says "ignore the target name, focus on the package name". 

I agree with your basic point. I worry that an explicit string might blur the purpose, i.e. a reader might incorrectly think its naming is significant. But I can't think of anything better, and I agree what's there now is lacking.

Sensitivity suggestion: how about `xyz`, since `xxx` can have unfavorable associations?

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
   end with a slash.
 </p>
 
+<h4 id="pattern_expansion">Target pattern expansions, <code>//<b>...</b></code>
```
gregestren, line 360: https://docs.bazel.build/versions/4.0.0/guide.html#specifying-targets-to-build also mentions this, and also properly defines `target_pattern`. Link there when introducing the term?

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
...
+  and <code>//<b>:all</b></code></h4>
+<p>
+  The two <code>...</code>and <code>:all</code> notations are special when used
+  on the command line. These target patterns are expanded by bazel before the
+  build is started.
```
gregestren, line 365: Non-passive voice suggestion: "Bazel expands these target patterns before starting the build". Or possibly "Bazel expands these patterns into a list of targets before starting the build".

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
   end with a slash.
 </p>
 
+<h4 id="pattern_expansion">Target pattern expansions, <code>//<b>...</b></code>
+  and <code>//<b>:all</b></code></h4>
+<p>
+  The two <code>...</code>and <code>:all</code> notations are special when used
```
gregestren, line 363: What about just "... and :all are special when used..." Is that still clear?

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
...
+  on the command line. These target patterns are expanded by bazel before the
+  build is started.
+</p>
+<p>
+  Adding <code>...</code> after a package tells bazel that you're interested in
```
gregestren, line 368: Nit here and below:  I think we capitalize "Bazel" when we're not referring to the command line, which I think fits this context.

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
...
+  all targets of all packages below the specified directory. For example
```
gregestren, line 369: Nit: "after a package" vs. "below the specified directory" Could it be more consistent to just use "package" in both places?

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
...
+  <code>//...</code> means "all targets in all packages in the source tree".
+  On the other hand <code>//foo/bar/...</code> means "all targets in all
+  packages in the <code>//foo/bar</code> package and all packages underneath
+  the <code>//foo/bar</code> package.
```
gregestren, line 373: Agreed this is super helpful. I think it makes sense that https://docs.bazel.build/versions/4.0.0/guide.html#specifying-targets-to-build is the source of truth for the concept of target pattern, while this page provides an accessible summary of two major ones.

site/docs/build-ref.html
```diff
@@ -357,6 +357,29 @@ <h4>Package names, <code>//<b>package-name</b>:...</code></h4>
...
+</p>
+<p>
+  Adding <code>:all</code> after a package tells bazel that you're interested
+  in all targets within that package. For example <code>//:all</code> means
```
gregestren, line 377: There's some nuance here. See the ref in https://docs.bazel.build/versions/4.0.0/guide.html#specifying-targets-to-build. I don't think it's worth getting too deep in the weeds in this text. But technically `:all` doesn't cover all targets.

gregestren:
Pinging @jin and @floriographygoth since they both have opinions on how we organize communication of basic concepts.

Significantly, this demonstrates conceptual overlap between `build-ref.html` and https://docs.bazel.build/versions/4.0.0/guide.html#specifying-targets-to-build. I personally think this helps, since it highlights a meaningful concept ("target pattern") the page doesn't otherwise mention. But the source of truth for the concept is on another page.  Does this seem like a good balance to you?

gregestren:
Also the comment about wanting to add a section about this to https://bazel.build/faq.html. The FAQ as it stands isn't really technical: it looks focused on Bazel's general value proposition and organization. 

Is there a conceptual space for a technically-focused FAQ?

philsc:
Thanks @gregestren , really appreciate the feedback!

I'll incorporate all of that this weekend. If we end up deciding that this is a poor addition, that's totally fine too. Just trying to brainstorm some ways to make this information a little more accessible.

site/docs/build-ref.html
philsc: Fair enough. Switched to `xyz`.

site/docs/build-ref.html
philsc: After looking at the guide, I decided to remove this little section. I think that's a much better reference.

Instead, I added two entries in the guide to be more explicit about these two patterns. Let me know what you think.

Thanks!

philsc:
> Also the comment about wanting to add a section about this to https://bazel.build/faq.html. The FAQ as it stands isn't really technical: it looks focused on Bazel's general value proposition and organization.
> 
> Is there a conceptual space for a technically-focused FAQ?

Fair enough. I removed that from the summary.

gregestren:
Did your update intentionally revert most of your changes?

philsc:
> Did your update intentionally revert most of your changes?

Yes. After looking at the guide you linked (guide.html), I convinced myself that most of the additions I made are more of a distraction than a useful reference. So I decided to remove the section I added section.

Instead, I added two entries in guide.html to be more explicit about these two patterns. Let me know what you think.

I'm happy to bring back the original paragraph I added too, but it did seem superfluous.

site/docs/guide.md
```diff
@@ -157,6 +157,14 @@ workspace.
   <td><code>//foo/...:all-targets</code></td>
   <td>All targets (rules and files) in all packages beneath the directory <code>foo</code>.</td>
 </tr>
+<tr>
+  <td><code>//...</code></td>
+  <td>All rules in all non-<a href="external.html">external</a> packages.</td>
```
jin, line 162: `All targets in packages in the workspace. This does not include targets from <a href="external.html">external repositories.</a>`

jin, line 162: While you're here, I think `rules` in the other rows above should really mean `rule targets`. Do you mind changing those as well?

site/docs/guide.md
```diff
@@ -157,6 +157,14 @@ workspace.
...
+</tr>
+<tr>
+  <td><code>//:all</code></td>
+  <td>All rules in the top-level package.</td>
```
jin, line 166: All targets in the top-level package, if there is a BUILD file at the root of the workspace.

site/docs/build-ref.html
```diff
@@ -293,7 +293,7 @@ <h3 id="lexi">Lexical specification of a label</h3>
   The precise details of allowed target names are below.
 </p>
 
-<h4 id="name">Target names, <code>//...:<b>target-name</b></code></h4>
+<h4 id="name">Target names, <code>//xyz:<b>target-name</b></code></h4>
```
jin, line 296: I think the style is to use //<package name>:target-name or //path/to/package:target-name, instead of `xyz`.

site/docs/guide.md
philsc: Thanks!
Done

philsc: Good idea. Done.

site/docs/build-ref.html
philsc: Fair enough. Works for me.
Done.

jin (approved):

jin:
Will import after @gregestren takes a look.

gregestren:
> Will import after @gregestren takes a look.

Missed this last update, apologies.

gregestren (approved):

PR #13173, opened 2021-03-08
docs/install-ubuntu: fixing missing dependency

xinau:


meteorcloudy:
@xinau Can you explain why `apt-transport-https` is needed?

xinau:
@meteorcloudy Sry. The problem is that your example adds a source with HTTPS instead of HTTP as protocol. Using HTTPS is not supported by apt natively afaik, only through a transport method like apt-transport-https. But the apt-transport-https package providing said method isn't installed in every environment by default.

meteorcloudy:
OK, thanks!

meteorcloudy (approved):

PR #13182, opened 2021-03-09
Remove Java14 from testing

comius:
JDK14 has been removed, but some leftovers are still there.


comius:
Related to https://github.com/bazelbuild/java_tools/issues/45

Issue #13189, opened 2021-03-09
Starlark: ctx.actions.run creates wrong argv0 when executable is top-level tool


laszlocsomor:
### Description of the problem / feature request:

`ctx.actions.run(executable = x, ...)` won't work when `x` is a label in the root package (e.g. `//:bin.sh`).

The spawn will have `x` as argv0 (resolved against the PATH) instead of `./x`.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

<details>

<summary>Sources (click to expand):</summary>

`//:BUILD`:

```starlark
load("foo.bzl", "foo")

foo(
    name = "x",
    out = "x.out",
    tool = "bin.sh",
)

genrule(
    name = "y",
    outs = ["y.out"],
    cmd = "$(location bin.sh) $@",
    tools = ["bin.sh"],
)
```

`//:foo.bzl`:

```starlark
def _impl(ctx):
    ctx.actions.run(
        outputs = [ctx.outputs.out],
        executable = ctx.executable.tool,
        arguments = [ctx.outputs.out.path],
    )
    return [DefaultInfo(files = depset([ctx.outputs.out]))]

foo = rule(
    implementation = _impl,
    attrs = {
        "tool": attr.label(allow_single_file = True, executable = True, cfg = "host"),
        "out": attr.output(mandatory = True),
    },
)
```

`//:bin.sh`:
```sh
#!/bin/bash
echo hello $0 > $1
```

</details>

Building the Starlark rule fails, because `bin.sh` is not on the PATH (should be `./bin.sh`):

```
  $ bazel --ignore_all_rc_files build -s //:x
Starting local Bazel server and connecting to it...
INFO: Analyzed target //:x (4 packages loaded, 7 targets configured).
INFO: Found 1 target...
SUBCOMMAND: # //:x [action 'Action x.out', configuration: 86d23d2b5aedbeecfc6e6954ea9486ec39e9e40bcbc4b17cec46a06eed198d11, execution platform: @local_config_platform//:host]
(cd /home/laszlo/.cache/bazel/_bazel_laszlo/bc5487f033a0af65dff1ff0bd4000d75/execroot/__main__ && \
  exec env - \
  bin.sh bazel-out/k8-fastbuild/bin/x.out)
ERROR: /tmp/foo/BUILD:3:4: Action x.out failed: (Exit 1): bin.sh failed: error executing command bin.sh bazel-out/k8-fastbuild/bin/x.out

[...]

PR #13192, opened 2021-03-10
SpawnAction.Builder: use callable PathFragment

laszlocsomor:
If you pass an Artifact or FilesToRunProvider to
Builder.setExecutable, it will now use the
callable path, not just the execpath.

Fixes https://github.com/bazelbuild/bazel/issues/13189

Unfortunately, call sites that pass PathFragment
are susceptible to the bug, because PathFragment
is inherently ambiguous: if it's a bare file name,
the shell resolves it against PATH, otherwise
against the current directory.

RELNOTES: none

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/actions/CustomCommandLine.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/actions/CustomCommandLine.java
@@ -797,6 +797,16 @@ public final class CustomCommandLine extends CommandLine {
     }
 
     /**
+     * Adds an artifact by calling {@link PathFragment#getCallablePathString}.
+     *
+     * <p>Prefer this over manually calling {@link PathFragment#getCallablePathString}, as it avoids
+     * storing a copy of the path string.
+     */
+    public Builder addCallablePath(@Nullable PathFragment value) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/analysis/actions/CustomCommandLine.java
```diff
@@ -1320,4 +1330,18 @@ public void addToFingerprint(
       }
     }
   }
+
+  /** A {@link PathFragment} that is expanded with {@link PathFragment#getCallablePathString()}. */
+  private static final class CallablePathFragment {
```
lberki, line 1335: nit: I got confused for a while because I thought "Callable" means a "implementing the Java Callable interface". But then again, `PathFragment.getCallalbePathString()` is not new in this change, so meh.

lberki (approved):

src/main/java/com/google/devtools/build/lib/analysis/actions/CustomCommandLine.java
laszlocsomor: Shall I change the name?

PR #13194, opened 2021-03-10
Update the open source filegroup rule to disable toolchain resolution.

katre:
The internal version of this rule has toolchain resolution disabled, and
this has been widely documented, so it is easiest to update the external
version of the rule to match.


philsc:
@katre , as far as I can tell this patch undoes your work from here: https://github.com/bazelbuild/bazel/issues/12899

I.e. toolchain resolution allowed `select()` statements to use constraints.

katre:
@philsc Argh, you're right. Okay, I need a better solution to this.

Issue #12110, opened 2020-09-16
--toolchain_resolution_debug output not ordered


attilaolah:
### Description of the problem / feature request:

When running a build with `--toolchain_resolution_debug`, the output lines seem to come in random order, each time different, making it really hard to debug the underlying issue.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
$ bazel build //any/target --toolchain_resolution_debug
$ bazel clean
$ bazel build //any/target --toolchain_resolution_debug
$ bazel clean
…
``

### What operating system are you running Bazel on?

Linux (Debian).

### What's the output of `bazel info release`?

release 3.5.0

###  Have you found anything relevant by searching the web?

Nope.

### Any other information, logs, or outputs that you want to share?

A possibly separate issue is that the flag itself doesn't always seem to work. If I invoke the same command several times, sometimes there is no debug output at all. But usually changing the flags or running `bazel clean` fixes the issue for the next run.

katre:
This makes sense, but I'm not sure if I'll have time to tackle it.

The relevant code is in [`SingleToolchainResolutionFunction.resolveConstraints`](https://source.corp.google.com/piper///depot/google3/third_party/bazel/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java;rcl=282570615;l=135). I think the right approach is to sort the loop values (i.e., `toolchains`, `availableExecutionPlatformKeys`), and to do the same in [`checkConstraints`](https://source.corp.google.com/piper///depot/google3/third_party/bazel/src/main/java/com/google/devtools/build/lib/skyframe/SingleToolchainResolutionFunction.java;rcl=282570615;l=209) for `mismatchSettings`.

The one trick is that we'd want to only sort when outputting debug information, because otherwise we're slowing down all toolchain resolution.

If you want to take a crack at it, I'd be happy to review a PR, otherwise it'd be nice to do but isn't a priority.

dmitrig01:
Just ran into this. It makes this flag useless, as far as I can tell.

nfelt:
I would like to +1 that with the current output this flag honestly provides net negative value in debugging because the output is so confusing.  Here's a completely bare-bones example of the output I get on my gLinux workstation with Bazel 3.7.0 (which I realize is not latest, but it doesn't seem like much has changed other than converting the flag to take a regex filter).

Command:

```shell
[...]

PR #13207, opened 2021-03-11
Convert --toolchain_resolution_debug to take a regex of toolchain typ…

### LINKED ISSUE #7713: Minimize output of --toolchain_resolution_debug

katre:
…es to debug

Work towards #12110.
Fixes #7713.

RELNOTES: The flag `--toolchain_resolution_debug` now takes a regex argument, which is used to check which toolchain types should have debug info printed. You may use `.*` as an argument to keep the current behavior of debugging every toolchain type.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
@@ -24,6 +24,7 @@ import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.EventHandler;
 import com.google.devtools.build.lib.starlarkbuildapi.platform.PlatformConfigurationApi;
 import com.google.devtools.build.lib.util.RegexFilter;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
@@ -36,6 +37,7 @@ public class PlatformConfiguration extends Fragment implements PlatformConfigura
   private final Label targetPlatform;
   private final ImmutableList<String> extraToolchains;
   private final List<Map.Entry<RegexFilter, List<Label>>> targetFilterToAdditionalExecConstraints;
+  private final RegexFilter toolchainResolutionDebugRegexFilter;
 
   public PlatformConfiguration(BuildOptions buildOptions) {
     PlatformOptions platformOptions = buildOptions.get(PlatformOptions.class);
@@ -46,6 +48,7 @@ public class PlatformConfiguration extends Fragment implements PlatformConfigura
     this.extraToolchains = ImmutableList.copyOf(platformOptions.extraToolchains);
     this.targetFilterToAdditionalExecConstraints =
         platformOptions.targetFilterToAdditionalExecConstraints;
+    this.toolchainResolutionDebugRegexFilter = platformOptions.toolchainResolutionDebug;
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
```diff
@@ -111,4 +114,23 @@ public Label getTargetPlatform() {
...
+  public boolean debugToolchainResolution(Label toolchainType) {
+    return debugToolchainResolution(ImmutableList.of(toolchainType));
+  }
+
+  /**
+   * Returns true if toolchain resolution debug info should be printed for any of these toolchain
+   * types.
+   */
+  public boolean debugToolchainResolution(Collection<Label> toolchainTypes) {
+    if (toolchainTypes.isEmpty()) {
```
gregestren, line 128: If this is empty, is there any actual information to debug? i..e what gets printed when this returns true and the input is empty?

src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
```diff
@@ -111,4 +114,23 @@ public Label getTargetPlatform() {
...
+      // Check an empty string, in case the filter is .*
+      return this.toolchainResolutionDebugRegexFilter.test("");
```
gregestren, line 130: Does plain old `--toolchain_resolution_debug` still work? (as an alias for `.*`)?

src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
katre: Even targets with no toolchains will have an execution platform selected (and this is used in some tests).

Ideally we could instead do `--toolchain_resolution_debug=//some/configured:target`, but that would require ConfiguredTargetFunction to make the check and pass a debug flag, thus complicating the ToolchainContextKey even more.

src/main/java/com/google/devtools/build/lib/analysis/PlatformConfiguration.java
katre: No, the argument is required (and if it is skipped an error will be printed). I couldn't figure out a way to distinguish between no flag being passed, and a base `--toolchain_resolution_debug` with no argument in order to keep the current behavior unchanged.

gregestren (approved):

PR #13219, opened 2021-03-14
Fix Incompatible Target Skipping for test args

philsc:
Before the fix, the test in this patch would error out with the
following message:

    ERROR: ...: in args attribute of sh_test rule //target_skipping:foo_test: label '//target_skipping:some_foo3_target' in $(location) expression expands to no files

The problem was that the `RunfilesSupport` class for the
`RunfilesProvider` instance was computing the arguments as the
incompatible target was being constructed. That meant that the
arguments get evaluated and it would try to resolve things like
`$(location :incompatible_dep)`. That can't really succeed so bazel
would throw an error.

This patch makes it so arguments are ignored when constructing an
incompatible target.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
@@ -453,6 +453,21 @@ public final class RunfilesSupport {
         computeActionEnvironment(ruleContext));
   }
 
+  /**
+   * Creates and returns a {@link RunfilesSupport} object for the given rule and executable. This
+   * version discards all arguments. Only use this for
+   * <a href="https://docs.bazel.build/versions/master/platforms.html#skipping-incompatible-targets">Incompatible Target Skipping</a>.
+   */
+  public static RunfilesSupport withExecutableButNoArgs(
+      RuleContext ruleContext, Runfiles runfiles, Artifact executable) {
+    return RunfilesSupport.create(
+        ruleContext,
+        executable,
+        runfiles,
+        CommandLine.EMPTY,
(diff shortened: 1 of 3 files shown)


philsc:
@gregestren , FYI.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -453,6 +453,20 @@ public static RunfilesSupport withExecutable(
         computeActionEnvironment(ruleContext));
   }
 
+  /**
+   * Creates and returns a {@link RunfilesSupport} object for the given rule and executable. This
+   * version discards all arguments. Only use this for Incompatible Target Skipping.
```
gregestren, line 458: Can you link this more precisely to a specific file or source javadoc? 

This comment comes from the hard-learned lesson that even specific concepts can be fuzzy to devs who haven't focused specifically on them. Precise links are always great for that.

gregestren (commented):
Nice catch! Overall looks great. Just one quick comment request.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
philsc: I can't verify that the link renders correctly, but I added a link to https://docs.bazel.build/versions/master/platforms.html#skipping-incompatible-targets

Would you prefer a link to the `site/docs/platforms.md` file instead?

gregestren: That's fine, thanks. Just anything to connect the breadcrumbs.

gregestren (approved):

PR #13223, opened 2021-03-15
Bazel's filegroup should participate in toolchain resolution.

### LINKED ISSUE #12899: Investigate whether filegroup can use platform-conditional selects

katre:
This reverts dce861d7c1a13164851fc9070756b946969c5e41, which was a
mistake.

It also removes a test which is now useless. There are no more build
rules (which have the target_compatible_with attribute for target
skipping) which also have toolchain resolution disabled, so this cannot
be tested for.

Fixes #12899 and undoes #13194.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/common/BazelFilegroupRule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/common/BazelFilegroupRule.java
@@ -76,7 +76,6 @@ public final class BazelFilegroupRule implements RuleDefinition {
             attr("path", STRING)
                 .undocumented(
                     "only used to expose FilegroupPathProvider, which is not currently used"))
-        .useToolchainResolution(ToolchainResolutionMode.DISABLED)
         .build();
   }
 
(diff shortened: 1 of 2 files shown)


katre:
@philsc thanks for pointing out that this change was a mistake. I'm removing the test that caused my confusion.

katre:
Originally, #12899 was filed because selects in filegroup worked, _except_ for selects using constraint values. I fixed this by enabling toolchain resolution for filegroup.

Then, last week, I noticed that
-  We have tests that assume filegroup does not take part in toolchain resolution, and
-  The internal version of filegroup and the external version deviate on this.

I picked the fix the make them both not take part in toolchain resolution, forgetting about #12899. That, unfortunately, regressed that issue, so now I am reverting that change (and removing a useless test).

The test is useless because it was testing whether a build rule (therefore, one that can use `target_compatible_with` for target skipping) that had toolchain resolution disabled work work properly. However, with this change to filegroup, there are no remaining build rules that have toolchain resolution disabled, so this isn't a valid state.

gregestren:
Makes sense. Thanks so much for the clarification.

gregestren (approved):

PR #13235, opened 2021-03-17
Add docs for host_conlyopt and host_linkopt

aschleck:
Options were added in c026569 but seem to have never made it into the user manual


aiuto:
This PR raises a question about the existance of the option itself.  "host" should probably be exec. But maybe the option should go away in favor of something in the toolchain.

comius (approved):

PR #13237, opened 2021-03-17
Fix order of build request id and command id

ulfjack:
They should be in this order rather than the other way round.

See `RemoteServerCapabilities.get`.

Change-Id: I4be952cc2ad43ae35c7052bd26b153b1e887e562


ulfjack:
Looks like this broke in commit 52c8773c3ccbcc9051427d6f86a9fe23217f3977, Bazel 3.0.0.

ulfjack:
@coeuvre

coeuvre (approved):
Thanks!

Issue #12759, opened 2020-12-29
The --script_path flag on Windows does produce a correct bat file when target is the test_wrapper


KrauseStefan:
### Description of the problem:

When running something like `bazel run "--script_path=C:\test.bat" //services/Web-Client/src/services:test`
Where the target ends up running the `test_wrapper` the created `test.bat` file is invalid.

Inspecting the file reveals something like the below, I have removed the setup of environment variables as this is not the problem.

``` cmd
@echo off
cd C:/users/skk/_bazel_skk/tilbahl3/execroot/my_project

  external/bazel_tools/tools/test/tw.exe services/Web-Client/src/services/test_chromium-local.bat %*
```

The last line will always error with 
```
'external' is not recognized as an internal or external command, operable program or batch file.
```

I see this issue when this is used with [ibazel](https://github.com/bazelbuild/bazel-watcher) and the [karma_web_test_suite](https://docs.aspect.dev/rules_nodejs/Karma.html#karma_web_test_suite) test rules although the issue is not caused by them

There are multiple variants of that last line that could work on windows
``` cmd
  external\bazel_tools\tools\test\tw.exe services/Web-Client/src/services/test_chromium-local.bat %*
```
or
``` cmd
  C:/users/skk/_bazel_skk/tilbahl3/execroot/my_project/external/bazel_tools/tools/test/tw.exe services/Web-Client/src/services/test_chromium-local.bat %*
```
The last one is how the output is defined when used with the [concatjs_devserver](https://bazelbuild.github.io/rules_nodejs/Concatjs.html#concatjs_devserver) rule.
``` cmd
  C:/users/skk/_bazel_skk/tilbahl3/execroot/my_project/bazel-out/x64_windows-fastbuild/bin/services/Web-Client/src/devserver.exe %*
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Add the below to a `BUILD.bazel` file
``` skylark
sh_test(
    name = "foo_test",
    srcs = ["foo_test.sh"],
)
```

Then run ` bazel run "--script_path=C:\test_foo.bat" foo_test`

Try to run the `C:\test_foo.bat` file observe the error 

### What operating system are you running Bazel on?

Windows 10 1909

### What's the output of `bazel info release`?
`release 3.7.2`

I have also tested this with the `4.0.0 rc7` and as old as `1.2.1`

###  Have you found anything relevant by searching the web?

I found no related issues...

CC @alexeagle

PR #13242, opened 2021-03-18
Using back slashes in the binary path in the script_path output file on Windows

mai93:
On Windows, batch does not recognize the path of the binary written in the generated script if script_path is specified in Bazel run command. This PR fixes this by replacing the forward slashes ("/") in the binary path with back slashes ("\\").

fixes: #12759

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/util/CommandFailureUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/util/CommandFailureUtils.java
@@ -41,7 +41,14 @@ public class CommandFailureUtils {
     void describeCommandCwd(String cwd, StringBuilder message);
     void describeCommandEnvPrefix(StringBuilder message, boolean isolated);
     void describeCommandEnvVar(StringBuilder message, Map.Entry<String, String> entry);
-    void describeCommandElement(StringBuilder message, String commandElement);
+    /**
+     * Format the command element and add it to the message.
+     *
+     * @param message the message to modify
+     * @param commandElement the command element to be added to the message
+     * @param isBinary is true if the `commandElement` is the binary to be executed
+     */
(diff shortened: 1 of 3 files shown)


meteorcloudy:
Can we fix this in CommandFailureUtils and WindowsDescribeCommandImpl? This will also help `--subcomand` to print out correct command line on Windows.

src/main/java/com/google/devtools/build/lib/util/CommandFailureUtils.java
```diff
@@ -41,7 +41,14 @@
     void describeCommandCwd(String cwd, StringBuilder message);
     void describeCommandEnvPrefix(StringBuilder message, boolean isolated);
     void describeCommandEnvVar(StringBuilder message, Map.Entry<String, String> entry);
-    void describeCommandElement(StringBuilder message, String commandElement);
+    /**
+     * Format the command element and add it to the message.
+     *
+     * @param message the message to modify
+     * @param commandElement the command element to be added to the message
+     * @param isBinary is true if the `commandElement` is the binary of the command
```
meteorcloudy, line 49: Nit: Maybe "if the `commandElement` is the binary to be executed"?

src/main/java/com/google/devtools/build/lib/util/CommandFailureUtils.java
```diff
@@ -203,6 +211,7 @@ public static String describeCommand(
       }
     }
 
+    boolean isBinary = true;
```
meteorcloudy, line 214: Nit: maybe "isFirstArgument"?

meteorcloudy (approved):
Thanks!

meteorcloudy (approved):

Issue #13220, opened 2021-03-14
MSVC toolchain requiring `parse_showinclude` feature is not documented


limdor:
### Description of the problem:

When trying to build a cc_library using a custom msvc 2019 toolchain that is the same like the autodetected one, but without /showInclude options, it does not build and shows the following error:
`ERROR: C:/users/xavib/documents/development/quoniam/src/example/BUILD:3:11: Compiling src/example/library.cpp failed: error while parsing .d file: C:\users\xavib\_bazel_xavib\3r2w35xj\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\src\example\_objs\example\library.d (The system cannot find the file specified)
`
If I add /showInclude in the manually declared toolchain, then it works.

### Feature requests: what underlying problem are you trying to solve with this feature?

I want to have the same toolchain explicitly declared but I want to remove the /showInclude option

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create  toolchain like the autodetected one in Windows with Visual studio 2019, and remove the /showInclude options. Here is the one that I used: https://github.com/limdor/quoniam/commit/8605cd3368835624feb3422d01ea0e8d8e0d53de

### What operating system are you running Bazel on?

Windows

### What's the output of `bazel info release`?

4.0.0

###  Have you found anything relevant by searching the web?

I tried to check where this error about parsing .d files is comming from, and this seems to come from Bazel, and not from Visual Studio: https://github.com/bazelbuild/bazel/search?q=%22error+while+parsing+.d+file%22
The most interesting part is in this part of the code, it seems that the behavior is different if /showInclude is specified or not, maybe someone familiar with the code can say what is the reason for that.
https://github.com/bazelbuild/bazel/blob/2f00d4f5af575e03a6324eeb0e9ddbee2b0c6803/src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java#L1877

meteorcloudy:
Enabling the `/showincludes` feature is required for MSVC toolchain, because Bazel uses the output to check if the headers included are declared as input. So this failure is WAI.

limdor:
https://docs.bazel.build/versions/master/windows.html#build-c-with-msvc does not mention anything about it. Is it mentioned somewhere else? How is supposed the user know about it?
If that is intended, would it be ok if I create a feature request at least to show that this is missing?

meteorcloudy:
[...]

Issue #12767, opened 2021-01-03
Provide a way to create a toolchain from an autodetected one


limdor:
### Description of the problem / feature request:

Bazel has two ways to deal with toolchains, one is to use toolchains provided by the user, and the other one is to automatically provide toolchains depending on the tools that are installed in the computer where Bazel is executed.
The auto-detected toolchains are very useful for anyone starting with Bazel but soon people start running into the challenge of wanting to change some settings of the default toolchain (i.e: compiler flags).

The way to do it, is to create a toolchain like the auto-detected one (you can figure out the one used with `--toolchain_resolution_debug`) and change the desired flag, but this can be very tedious. Some of the auto-detected toolchains are more than 1.000 lines long and with autodetected paths of the tools to be used.

### Feature requests: what underlying problem are you trying to solve with this feature?

In the same way that there is `--toolchain_resolution_debug` to know what toolchain is selected, provide an option to print all the information of the selected toolchain that would be needed to create it manually.
https://docs.bazel.build/versions/master/be/platform.html#toolchain

Some of the information that should contain:

- Absolut path of the compiler/linker from the host being used
- Compiler flags being used
- Anything else that would be needed in order to create a toolchain following https://docs.bazel.build/versions/master/tutorial/cc-toolchain-config.html if you want to recreate the automatically detected one.

### What's the output of `bazel info release`?

4.0.0rc7

###  Have you found anything relevant by searching the web?

Some links showing why this might be relevant:
https://github.com/bazelbuild/bazel/issues/12737
https://github.com/bazelbuild/bazel/issues/2209
https://stackoverflow.com/questions/63204551/how-to-tell-bazel-which-compiler-to-use-without-configuring-a-new-toolchain
https://stackoverflow.com/questions/59092579/bazel-add-compile-flags-to-default-c-toolchain
https://stackoverflow.com/questions/63264512/bazel-windows-10-toolchain-configuration-with-mingw64-compiler

katre:
There is, unfortunately, no generic toolchain-discovery system. The CC, Java, Python, etc, all deal with this separately. You'll need to file this issue against the specific rules whose auto-discovered toolchain you want to specialize. It looks like CC rules from your description, so I'll start by re-assigning this to the language owners.

PR #13247, opened 2021-03-19
Add the requirement of /showIncludes for MSVC

limdor:
MSVC is not working in Bazel if showIncludes is not declared https://github.com/bazelbuild/bazel/issues/13220

### DIFF
--- a/site/docs/windows.md
+++ b/site/docs/windows.md
@@ -137,6 +137,8 @@ To build C++ targets with MSVC, you need:
 
 *   [The Visual C++ compiler](install-windows.html#install-vc).
 
+*   The compiler flag `/showIncludes` in case that you define your custom toolchain.
+
 *   (Optional) The `BAZEL_VC` and `BAZEL_VC_FULL_VERSION` environment variable.
 
     Bazel automatically detects the Visual C++ compiler on your system.


site/docs/windows.md
```diff
@@ -137,6 +137,8 @@ To build C++ targets with MSVC, you need:
 
 *   [The Visual C++ compiler](install-windows.html#install-vc).
 
+*   The compiler flag `/showIncludes` in case that you define your custom toolchain.
```
aiuto, line 140: This does not really help users. A pointer to how to set up a custom MSVC toolchain might be useful.  We don't want users to add /showIncludes, you want the toolchain author to set it up correctly. This hints that it is the user's problem.

meteorcloudy: Agree, normal users should only need to know this when they are trying to write their own custom MSVC toolchain. Adding this as a general requirement doesn't sound right.

limdor: > This does not really help users. A pointer to how to set up a custom MSVC toolchain might be useful. We don't want users to add /showIncludes, you want the toolchain author to set it up correctly. This hints that it is the user's problem.

@aiuto I understand your point, but everything depends on who you understand as a user. From your comment I get that this documentation is from your point of view intended for the end-user, the average software developer that do not have to create a toolchain. However in the same documentation page, in the section of clang is talking about the need of registering a platform, thing that I would also not expect an average user to have to do, but the person that sets up the Bazel. https://docs.bazel.build/versions/master/windows.html#build-c-with-clang

When you talk about a pointer to how to set up a custom MSVC toolchain, is there something already? Or you mean a pointer plus create the page explaining it?

> Agree, normal users should only need to know this when they are trying to write their own custom MSVC toolchain. Adding this as a general requirement doesn't sound right.

@meteorcloudy can we keep the ticket that I created https://github.com/bazelbuild/bazel/issues/13220 to improve the documentation or should I create a new one? Not being mentioned in the documentation and not having an open ticket for that, will only make people struggle on the same issue that I had for hours. I understand that for a Bazel developer is expected this behavior but other user (including toolchain writers) not involved with Bazel development might find this behavior hard to predict.

@meteorcloudy @aiuto in general the idea why I push for this transparency is because right now it is not easy to create a toolchain for MSVC similar to the autogenerated one https://github.com/bazelbuild/bazel/issues/12767 and when trying to do that I spent several hours trying to understand why it is failing when not specifying showIncludes but passing otherwise. I would like to prevent more people spending time like me on something that could be documented.

meteorcloudy: @limdor Thanks! I totally understand. I changed the original issue to reflect documentation issue. Please use that to track. I also suggested a place where this should be documented.

As for creating a toolchain for MSVC, maybe https://github.com/meteorcloudy/my_tests/tree/master/preconfigure_msvc_toolchain can help? Instead of writing one entirely from scratch, maybe you can start with the auto generated one?

aiuto:
Perhaps the two of you can pair up to turn that private test into a public
sample in the examples repository.

limdor:
@aiuto @meteorcloudy I will close the PR and continue the discussion on https://github.com/bazelbuild/bazel/issues/13220
Once we have it more clear we can create another PR with the changes.

PR #13248, opened 2021-03-19
Change `set -x` in coverage to be set by var

keith:
Using `set -x` in the coverage scripts causes a lot of noise, this changes it to be off by default, but something you can enable by passing `--test_env=VERBOSE_COVERAGE=1` for debugging.

### DIFF
--- a/tools/test/collect_cc_coverage.sh
+++ b/tools/test/collect_cc_coverage.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash
  # Copyright 2016 The Bazel Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


brentleyjones (approved):

finn-ball:
These scripts are quite broken in general. I found it very useful to leave these flags in when debugging what was going wrong with coverage. 

Out of curiosity, I see you also removed the `-x` flag from the `collect_cc_coverage.sh` - do you happen to use this with remote execution?

brentleyjones:
We plan on using code coverage with remote execution, yes. For context, we are doing Swift code coverage, not C++.

keith:
I definitely get the debugging use case, it's too bad we don't have an easy way to have both here. Maybe we can conditionally `set -x` based on some `--test_env` like `--test_env=DEBUG_COVERAGE=1`

keith:
@c-mita can you review?

c-mita (approved):

keith:
Thanks!

PR #13255, opened 2021-03-22
Update Fedora installation instructions in docs

JustusFT:
For [the Fedora installation instructions for Bazel 4](https://docs.bazel.build/versions/4.0.0/install-redhat.html), it says to do `dnf install bazel3`, but since the docs are for v4 I think it should be `dnf install bazel4`

### DIFF
--- a/site/docs/install-redhat.md
+++ b/site/docs/install-redhat.md
@@ -35,7 +35,7 @@ version of the Bazel package.
 
     ```bash
     dnf copr enable vbatts/bazel
-    dnf install bazel3
+    dnf install bazel4
     ```
 
 ## Installing on CentOS 7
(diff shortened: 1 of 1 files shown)


JustusFT:
@googlebot I signed it!

aiuto:
Did we push bazel4 to the appropriate repositories?

JustusFT:
> Did we push bazel4 to the appropriate repositories?
> More important. Does this actually work?

Yes, `bazel4` is a package from [vbatts/bazel](https://copr.fedorainfracloud.org/coprs/vbatts/bazel/packages/) (the instructions say to use this repository)
So `dnf install bazel4` works

aiuto (approved):

PR #13256, opened 2021-03-22
Remove makeLabel helper function and call Label.parseAbsoluteUnchecked

katre:
I'm breaking up https://github.com/bazelbuild/bazel/pull/13246 into smaller changes: this one is a pure inlining.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/analysis/platform/ConstraintSettingInfoTest.java
+++ b/src/test/java/com/google/devtools/build/lib/analysis/platform/ConstraintSettingInfoTest.java
@@ -17,6 +17,7 @@ package com.google.devtools.build.lib.analysis.platform;
 
 import com.google.common.testing.EqualsTester;
 import com.google.devtools.build.lib.analysis.util.BuildViewTestCase;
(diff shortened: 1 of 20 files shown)


brandjon:
I feel ambivalent about whether `makeLabel` has value as a shorthand for `Label.parseAbsoluteUnchecked`, given how common and verbose the latter is. If we were to keep it, we could make it call `parseAbsoluteUnchecked` directly rather than reimplementing its wrapping of `parseAbsolute`. I leave it up to you.

I dug up some history in case it's relevant to the decision: `parseAbsoluteUnchecked` was added in 2010. `makeLabel` was added in 2014, initially private, and made public in 2017 (by you). `makeLabel` has ~200 uses to `parseAbsoluteUnchecked`'s ~800.

brandjon (approved):

katre:
I added makeLabel because I didn't know parseAbsoluteUnchecked existed. If I had, makeLabel wouldn't.

Using parseAbsoluteUnchecked is clearer, I think, because it highlights that this is skipping several checks the normal parsing uses. We could rename the method to `parseStatic` or something similar.

PR #13258, opened 2021-03-22
Update ConfiguredTargetFunction.computeUnloadedToolchainContexts to

### LINKED ISSUE #13243: Toolchains that require toolchains don't get toolchains
Branched from https://github.com/bazelbuild/rules_rust/issues/635:

If a toolchain rule declares that it requires a toolchain type, that is currently dropped during analysis (see [`ConfiguredTargetFunction.computeUnloadedToolchainContexts`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java;drc=876deaa95da77c5d21c51693a8519f16e32b3f95;l=567)).

There are two issues here:
1.  Only the current target should be dropped (and only in the current configuration), in order to avoid a loop. The toolchain may still need any other toolchain types it requested.
2. If the toolchain declares toolchain types that the parent did not, those are ignored and not supplied. This may be a harder fix, but we can at least warn about it.

katre:
directly require the correct execution platform.

This fixes several issues around toolchains that depend on toolchains.

Fixes #13243.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
@@ -520,18 +520,31 @@ public final class ConfiguredTargetFunction implements SkyFunction {
 
     Map<String, ToolchainContextKey> toolchainContextKeys = new HashMap<>();
     String targetUnloadedToolchainContext = "target-unloaded-toolchain-context";
-    ToolchainContextKey toolchainContextKey;
+
+    ToolchainContextKey.Builder toolchainContextKeyBuilder =
+        ToolchainContextKey.key()
+            .configurationKey(toolchainConfig)
+            .requiredToolchainTypeLabels(requiredDefaultToolchains)
+            .execConstraintLabels(defaultExecConstraintLabels)
+            .shouldSanityCheckConfiguration(configuration.trimConfigurationsRetroactively());
+
     if (parentToolchainContextKey != null) {
-      toolchainContextKey = parentToolchainContextKey;
-    } else {
-      toolchainContextKey =
-          ToolchainContextKey.key()
-              .configurationKey(toolchainConfig)
-              .requiredToolchainTypeLabels(requiredDefaultToolchains)
-              .execConstraintLabels(defaultExecConstraintLabels)
-              .shouldSanityCheckConfiguration(configuration.trimConfigurationsRetroactively())
-              .build();
+      // Find out what execution platform the parent used, and force that.
+      // This key should always be present, but check just in case.
+      ToolchainContext parentToolchainContext =
+          (ToolchainContext)
+              env.getValueOrThrow(parentToolchainContextKey, ToolchainException.class);
+      if (env.valuesMissing()) {
+        return null;
+      }
+
+      Label execPlatform = parentToolchainContext.executionPlatform().label();
+      if (execPlatform != null) {
+        toolchainContextKeyBuilder.forceExecutionPlatform(execPlatform);
+      }
     }
+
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/devtools/build/lib/skyframe/ToolchainResolutionFunctionTest.java
```diff
@@ -431,4 +431,45 @@ public void resolve_noMatchingPlatform() throws Exception {
         .hasExceptionThat()
         .isInstanceOf(NoMatchingPlatformException.class);
   }
+
+  @Test
+  public void resolve_forceExecutionPlatform() throws Exception {
+    // This should select platform linux, toolchain extra_toolchain_linux, due to the forced
+    // platform,
+    // even though platform mac is registered first.
```
gregestren, line 439: What does the platform registration order mean if the build is explicitly setting the platform to `--platforms=//platforms:linux` anyway?

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -520,18 +520,31 @@ public SkyValue compute(SkyKey key, Environment env) throws ConfiguredTargetFunc
 
     Map<String, ToolchainContextKey> toolchainContextKeys = new HashMap<>();
     String targetUnloadedToolchainContext = "target-unloaded-toolchain-context";
-    ToolchainContextKey toolchainContextKey;
+
+    ToolchainContextKey.Builder toolchainContextKeyBuilder =
```
gregestren, line 524: I'm having trouble keeping track of all the interoperating pieces.

What precisely does the parent toolchain context mean w.r.t. the current one? That's the toolchain context of the configured target that may or may not exist, depending on which key instantiated ConfiguredTargetFunction?

What's the  full logic flow here? What exactly does dropping and loop avoidance do?

I just need a bit more of an overview to contextualize my understanding.

src/test/java/com/google/devtools/build/lib/skyframe/ToolchainResolutionFunctionTest.java
katre: Registration is for execution platform, and --platforms is for target. Updated the comment to hopefully be more clear.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
katre: When a configured target CT depends on a toolchain T, it also passed CT's ToolchainContextKey along. This allows T to make sure that it uses the same execution platform, to enable T's dependencies to use the exec transition and target the correct platform.

Previously, we were just throwing away T's toolchain context key, but that isn't correct: T may have its own required toolchain types. This change is adding a clearer way, where T can determine directly that CT's execution platform is, and then forcing toolchain resolution to use that (or throw an error, if it's not possible for some reason).

The specific use here is that rules_rust's rust_toolchain rule needs a cc_toolchain, and that was failing with the previous code.

katre:
Fixed the test failure and cleaned up ToolchainResolutionFunction a bit.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: Ah yes, I remember more of the general `ToolchainContextKey` semantics better, thanks. I don't think I'd considered before the idea of a toolchain requiring another toolchain - that's an interesting extension of the functionality.

gregestren (approved):

PR #13260, opened 2021-03-23
Update the verify_workspace test to also check distdir_deps.bzl.

katre:


katre:
The verify_workspace issue is being fixed by @alexjski .

src/test/shell/bazel/verify_workspace.sh
```diff
@@ -40,22 +40,24 @@ fi
...
-    #echo "Checking ${url}"
-    if ! curl --head -silent --fail --output /dev/null --retry 3 "${url}"; then
-      #fail "URL ${url} is invalid."
-      invalid_urls+=("${url}")
-    fi
+  for file in "${WORKSPACE_FILES[@]}"; do
+    for url in $(grep -E '"https://|http://' "${file}" | \
+      grep -v 'jekyll-tree' | \
+      grep -v '^ *#' | \
+      perl -pe 's#.*"(https?://[^"]+)".*#$1#g' | \
```
aiuto, line 53: Can we fix this to not use perl.  It can be done with sed.  I think
sed -e 's#.*"\(https\?://[^"]*\)".*$#\1#g'

Even better, a single command
sed -e '/jekyll-tree/d' -e '/^ *$/d' -e 's#.*"\(https\?://[^"]*\)".*$#\1#g'

aiuto (changes requested):

src/test/shell/bazel/verify_workspace.sh
katre: Done. I hate writing sed expressions.

aiuto (approved):

src/test/shell/bazel/verify_workspace.sh
aiuto: I have, on occasion, written sed *scripts* with conditional blocks and loops. It would almost always be quicker to write something in python. If only so you could have a comment block above the regex to say what you are doing.

Issue #10904, opened 2020-03-05
Use environment variables in .bazelrc via string interpolation


nathanhleung:
### Description of the problem / feature request:

I want to use environment variables in my .bazelrc via string interpolation

### Feature requests: what underlying problem are you trying to solve with this feature?

I'm running a remote Bazel cache secured with Basic Auth. I don't want to hardcode basic auth credentials into the `.bazelrc` checked into the repository.

Example `.bazelrc`:

```
build --remote_cache=https://username:password@mybazelcache.example.com
```

I would rather do something like this:

```
build --remote_cache=https://${CACHE_USERNAME}:${CACHE_PASSWORD}@mybazelcache.example.com
```

### What operating system are you running Bazel on?

macOS Catalina 10.15.3

### What's the output of `bazel info release`?

INFO: Invocation ID: 067b442a-efd2-431e-b39d-9ea6e2467bf3
release 2.0.0

###  Have you found anything relevant by searching the web?

Related issue comment: https://github.com/bazelbuild/bazel/issues/4635#issuecomment-365998683

booninite:
The official docs provide a workaround for this use case via `try-import` in your `.bazelrc`: https://docs.bazel.build/versions/master/best-practices.html#bazelrc

psigen:
This workaround isn't _exactly_ covering the use-case (I have a similar need).

Using a `try-import` with a `user.bazelrc` requires that the user rcfile contain the entirety of the option.  So in the situation where there is a small piece of user-specific information (i.e. *username*) that is combined with a big piece of information that should be source controlled (i.e. *full url of remote cache*), we are currently lacking a mechanism to combine those bits together.

nathanhleung:
@psigen Here was my workaround:

I needed this feature specifically for a CircleCI setup (where I set environment variables in "Project Settings"), so what worked for me was creating a script `generate_bazelrc.sh`:

```sh
#!/bin/bash

# Exit on error
set -e

echo "# Generated by generate_bazelrc.sh" > .bazelrc
echo "build --remote_cache=https://$BAZEL_CACHE_USER:$BAZEL_CACHE_PASSWORD@mybazelcache.example.com" >> .bazelrc
```

and then running `./generate_bazelrc.sh` before the start of every build.

psigen:
Thanks.  My workaround is also running a script.  But I'm using this for local development, so it's not as easy to make sure it's sourced when the environment is set up.

jheaff1:
[...]

PR #13266, opened 2021-03-25
Resolve env variables in bazelrc

anthpi:
Bazel will now try to resolve environment variables
provided in bazelrc files.
If environmetn variable is not found text will remain
unchanged.

Format to be used: ${FOO}

Example:
  import ${GIT_ROOT}/some/path/bazel.rc


anthpi:
This commit is related to the discussion in https://github.com/bazelbuild/bazel/issues/10904

aiuto:
We never resolve in #10904 if this is a good idea or not.
@lberki   I'm passing this off to you to make a judgment call on the feature in general.

anthpi:
@aiuto I think there are no opinions against the feature (a part from your comment about added complexity). Feature is optional to use and backwards compatible so it should not cause any issues for the existing users.

@lberki For some reason tests on CentOS are still failing and I was not able to identify why. I had similar issues with Windows platform tests and I was able to resolve them by using "blaze::GetEnv()" method that should take care of platform specific implementations. So my guess is that error probably related to the test case rather than to implementation. At the same time CentOS is a Linux platform and I was not able to find any special "ifdefs" that are CentOS specific. I locally run RHEL7 and tests pass. They even pass in Ubuntu environment while being tested by Buildkite. That's why I'm really puzzled about tests failing on CentOS.

@lberki Overall implementation of the feature is not very complex and located in one place. I hope that it should not be hard to maintain it. Feature is backwards compatible and optional to use. It will solve many issues that we have in our environment when we have customers running several workspaces in several repos. In https://github.com/bazelbuild/bazel/issues/10904 there were several people mentioning having a bazel wrapper just to be able to achieve same result. Today you either need to have a wrapper or provide parameters on the command line to be able to resolve env variables. Both of those things make environment complex and cumbersome to use for the users. With this feature in place one can remove the unnecessary wrappers and move over many (if not most) of the flags into the .bazelrc file thus greatly simplifying bazel commands that users need to type.

@lberki It's "by design" that format is not platform specific to allow portability between the platforms (if there is such a use case). Also, after discussing internally with my colleagues, we came to a conclusion that keeping values unresolved when required environment variables are not found is probably the best solution because this way users will get a clear error message containing actual name of the environment variable to be able to troubleshoot and fix it. Both of the things are debatable of course and I would like to get your feedback on this. By making these assumptions we were also able to simplify implementation.

anthpi:
@lberki Did you have time to start looking into this pull request?

lberki:
Erm, sorry for the delay. I answered on #10904  .

Issue #7313, opened 2019-01-31
Make the sandboxed file system more strict


emusand:
### Description of the problem / feature request:

Make it possible to configure the sandbox to whitelist local directories. The sandbox will have read access to only these directories (and its execroot). No other local directories will be available.

Today it is possible to blacklist directories with option `--sandbox_block_path=<directory>`. This feature request adds the possibility to whitelist directories instead.

### Feature requests: what underlying problem are you trying to solve with this feature?

The current sandbox has read permissions to its execroot and almost everything in /. If a rule reads a file with absolute path, bazel assumes it is a file provided by the operating system. Bazel will not rebuild the target if this file is updated.

My work group needs more hermetic builds. We have bad experience from a previous build system (IBM ClearCase) which did not track file accesses outside of the workspace (VOB). This is almost exactly the same limitation as in the current sandbox; rules can read any file on our distributed file systems with an absolute path, but the target will not be rebuilt if this file is updated. This limitation forced us to turn off the remote cache in ClearCase, and avoid using incremental builds in CI, since they were not reliable.

### Any other information, logs, or outputs that you want to share?

This has been discussed in the [bazel-discuss Google group](https://groups.google.com/forum/?nomobile=true#!topic/bazel-discuss/_tKOy-9Psyg).

[Design Document: Bazel Sandboxing 2.0](https://bazel.build/designs/2016/06/02/sandboxing.html) describes the current sandbox well, and the reason for allowing read access to everything in /.

My work group is willing to implement this feature.

philwo:
@benjaminp Could you share the link to your tweaked sandbox that can mount an image instead of the host `/` again? I remember that you posted it on an issue recently, but can't find it.

I wonder if that would already be enough to support this use case. We could spend some time to polish it and get it into Bazel mainline :)

benjaminp:
https://github.com/bazelbuild/bazel/issues/6994#issuecomment-457438551 has the details.

emusand:
I think that this feature request and mounting an image solve two different use cases.

As a first step we want to use parts the local system, but we want to make sure that no absolute paths to our distributed filesystems are available. This means that our builds will not be fully hermetical. That's OK for us. Our assumption is that the Linux distributions in our environment are similar enough, and that the differences in the local systems do not matter.

However, in the long run we want to migrate to fully hermetical builds. Then the possibility to mount a Docker image as / sounds very interesting.

benjaminp:
[...]

Issue #1525, opened 2016-07-14
Error with external dependencies: some change are not seen


chpatton013:
```
%> bazel version
Build label: 0.2.1-2016-04-08 (@447f7f3)
Build target: bazel-out/local_linux-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Apr 8 17:35:41 2016 (1460136941)
Build timestamp: 1460136941
Build timestamp as int: 1460136941
```

My project has an external dependency on a tar.gz hosted in s3 (`new_http_archive`). This dependency has several different build targets that I depend on in one of my packages (runtime dependency, so using the `data` attribute). I wanted to unify those targets into a single `filegroup` so that I could depend on them easily in multiple places. In the build file for my external dependency I created a `filegroup` that lists each of the relevant targets as both `srcs` and `data`. When I attempt to build the dependent package I receive an error about the new `filegroup` not existing:

```
ERROR: /path/to/workspace/dependent/BUILD:48:1: no such target '@dependency//:system': target 'system' not declared in package '' defined by /path/to/cache/dir/external/dependency/BUILD and referenced by '//dependent:runtime'.
```

The only way to get rid of this error is to expunge the build tree and rebuild. It seems that the state of the external dependency build targets are not reevaluated if the external dependency has not changed. I believe this to be an error in Bazel's handling of external dependencies because changes in the build file should be visible to other build targets without first destroying the state of the build tree.

damienmg:
0.2.1 is rather old release, can you update? We fixed a lot of invalidation issues since then.

Closing for now, feel free to ping if you have more issues.

marczych:
We're still seeing problems even on 0.3.0. For example:

```
INFO: Found 1 target...
ERROR: missing input file '@<redacted>//:<redacted>'.
ERROR: <redacted>/073241073e47b03b6e3590c37cd049f8/external/<redacted>/BUILD:3:1: @<redacted>//:<redacted>: missing input file '@<redacted>//:<redacted>'.
Target <redacted> failed to build
Use --verbose_failures to see the command lines of failed build steps.
ERROR: <redacted>73241073e47b03b6e3590c37cd049f8/external/<redacted>/BUILD:3:1 1 input file(s) do not exist.
INFO: Elapsed time: 21.521s, Critical Path: 4.59s
```

[...]

PR #13279, opened 2021-03-30
Implement Hermetic sandbox with support for hardlinks

frazze-jobb:
Adds linux-sandbox flag:
--experimental_use_hermetic_linux_sandbox - Configure linux-sandbox
 to run in a chroot environment to prevent access to files not
 mentioned in the bazel rules unless they can be found via
 explicitly whitelisted directories using --sandbox_add_mount_pair
 create hardlinks instead of symlinks, and fallback to copying.
 In case of writes to input files, the build will be aborted.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
@@ -20,27 +20,33 @@ import java.util.Objects;
 
 /**
  * In case we can't get a fast digest from the filesystem, we store this metadata as a proxy to the
- * file contents. Currently it is a pair of a relevant timestamp and a "node id". On Linux the
- * former is the ctime and the latter is the inode number. We might want to add the device number in
+ * file contents. Currently it is two timestamps and a "node id". On Linux we
+ * use both ctime and mtime and inode number. We might want to add the device number in
  * the future.
  *
- * <p>For a Linux example of why mtime alone is insufficient, note that 'mv' preserves timestamps.
+ * <p>For a Linux example of why mtime alone is insufficient, note that 'mv' preserves mtime.
  * So if files 'a' and 'b' initially have the same timestamp, then we would think 'b' is unchanged
  * after the user executes `mv a b` between two builds.
+ *
+ * <p>On Linux we also need mtime for hardlinking sandbox, since updating the inode reference counter
+ * preserves mtime, but updates ctime. isModified() call can be used to compare two FileContentsProxys
+ * of hardlinked files.
  */
 public final class FileContentsProxy {
   private final long ctime;
+  private final long mtime;
   private final long nodeId;
 
-  private FileContentsProxy(long ctime, long nodeId) {
+  public FileContentsProxy(long ctime, long mtime, long nodeId) {
     this.ctime = ctime;
+    this.mtime = mtime;
     this.nodeId = nodeId;
   }
 
   public static FileContentsProxy create(FileStatus stat) throws IOException {
     // Note: there are file systems that return mtime for this call instead of ctime, such as the
     // WindowsFileSystem.
-    return new FileContentsProxy(stat.getLastChangeTime(), stat.getNodeId());
+    return new FileContentsProxy(stat.getLastChangeTime(), stat.getLastModifiedTime(), stat.getNodeId());
   }
 
   @Override
@@ -54,16 +60,31 @@ public final class FileContentsProxy {
     }
 
     FileContentsProxy that = (FileContentsProxy) other;
-    return ctime == that.ctime && nodeId == that.nodeId;
+    return ctime == that.ctime && mtime == that.mtime && nodeId == that.nodeId;
+  }
+
+  /**
+   * Can be used when hardlink reference counter changes
+   * should not be considered a file modification.
+   * Is only comparing mtime and not ctime and is therefore
+   * not detecting changed metadata like permission.
+   */
+  public boolean isModified(FileContentsProxy other) {
+    if (other == this) {
+      return false;
+    }
+    // true if nodeId are different or inode has a new mtime
+    return nodeId != other.nodeId || mtime != other.mtime;
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(ctime, nodeId);
+    return Objects.hash(ctime, mtime, nodeId);
   }
 
   void addToFingerprint(Fingerprint fp) {
     fp.addLong(ctime);
+    fp.addLong(mtime);
     fp.addLong(nodeId);
   }
 
@@ -73,6 +94,6 @@ public final class FileContentsProxy {
   }
 
   public String prettyPrint() {
-    return String.format("ctime of %d and nodeId of %d", ctime, nodeId);
+    return String.format("ctime of %d and mtime of %d and nodeId of %d", ctime, mtime, nodeId);
   }
 }
--- a/src/main/java/com/google/devtools/build/lib/sandbox/AbstractSandboxSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/AbstractSandboxSpawnRunner.java
@@ -131,6 +131,9 @@ abstract class AbstractSandboxSpawnRunner implements SpawnRunner {
       try (SilentCloseable c = Profiler.instance().profile("subprocess.run")) {
         result = run(originalSpawn, sandbox, context.getTimeout(), outErr);
       }
+      try (SilentCloseable c = Profiler.instance().profile("sandbox.verifyPostCondition")) {
+        verifyPostCondition(originalSpawn, sandbox, context);
+      }
 
       context.lockOutputFiles();
       try (SilentCloseable c = Profiler.instance().profile("sandbox.copyOutputs")) {
@@ -148,6 +151,11 @@ abstract class AbstractSandboxSpawnRunner implements SpawnRunner {
       }
     }
   }
+  /**
+   * Override this method if you need to run a post condition after the action has executed
+   */
+  public void verifyPostCondition(Spawn originalSpawn, SandboxedSpawn sandbox,
+      SpawnExecutionContext context) throws IOException, ForbiddenActionInputException {}
 
   private String makeFailureMessage(Spawn originalSpawn, SandboxedSpawn sandbox) {
     if (sandboxOptions.sandboxDebug) {
--- a/src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
@@ -1,0 +1,100 @@
+// Copyright 2016 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.sandbox;
+
+import com.google.common.flogger.GoogleLogger;
+import com.google.devtools.build.lib.exec.TreeDeleter;
+import com.google.devtools.build.lib.sandbox.SandboxHelpers.SandboxInputs;
+import com.google.devtools.build.lib.sandbox.SandboxHelpers.SandboxOutputs;
+import com.google.devtools.build.lib.vfs.FileSystemUtils;
+import com.google.devtools.build.lib.vfs.Path;
+import com.google.devtools.build.lib.vfs.Symlinks;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Creates an execRoot for a Spawn that contains input files as hardlinks to their original
+ * destination.
+ */
+public class HardlinkedSandboxedSpawn extends AbstractContainerizingSandboxedSpawn {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private boolean sandboxDebug = false;
+  public HardlinkedSandboxedSpawn(
+      Path sandboxPath,
+      Path sandboxExecRoot,
+      List<String> arguments,
+      Map<String, String> environment,
+      SandboxInputs inputs,
+      SandboxOutputs outputs,
+      Set<Path> writableDirs,
+      TreeDeleter treeDeleter,
+      @Nullable Path statisticsPath,
+      boolean sandboxDebug) {
+    super(
+        sandboxPath,
+        sandboxExecRoot,
+        arguments,
+        environment,
+        inputs,
+        outputs,
+        writableDirs,
+        treeDeleter,
+        statisticsPath);
+    this.sandboxDebug = sandboxDebug;
+  }
+
+  @Override
+  protected void copyFile(Path source, Path target) throws IOException {
+    hardLinkRecursive(source, target);
+  }
+
+  /**
+   * Recursively creates hardlinks for all files in @param source path, in @param target path.
+   * Symlinks are resolved. If files is located on another disk, hardlink will fail
+   * and a copy will be made instead.
+   * Throws IllegalArgumentException if source path is a subdirectory of target path.
+   */
+  private void hardLinkRecursive(Path source, Path target) throws IOException {
+    if (source.isSymbolicLink()) {
+      source = source.resolveSymbolicLinks();
+    }
+
+    if (source.isFile(Symlinks.NOFOLLOW)) {
+      try {
+        source.createHardLink(target);
+      } catch (IOException e) {
+        if (sandboxDebug) {
+          logger.atInfo().log("File %s could not be hardlinked, file will be copied instead.", source);
+        }
+        FileSystemUtils.copyFile(source, target);
+      }
+    } else if (source.isDirectory()) {
+      if (source.startsWith(target)) {
+        throw new IllegalArgumentException(source + " is a subdirectory of " + target);
+      }
+      target.createDirectory();
+      Collection<Path> entries = source.getDirectoryEntries();
+      for (Path entry : entries) {
+        Path toPath = target.getChild(entry.getBaseName());
+        hardLinkRecursive(entry, toPath);
+      }
+    }
+  }
+}
--- a/src/main/java/com/google/devtools/build/lib/sandbox/LinuxSandboxedSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/LinuxSandboxedSpawnRunner.java
@@ -19,12 +19,16 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
+import com.google.devtools.build.lib.actions.ActionInput;
 import com.google.devtools.build.lib.actions.ExecException;
 import com.google.devtools.build.lib.actions.ExecutionRequirements;
+import com.google.devtools.build.lib.actions.FileArtifactValue;
+import com.google.devtools.build.lib.actions.FileContentsProxy;
 import com.google.devtools.build.lib.actions.ForbiddenActionInputException;
 import com.google.devtools.build.lib.actions.Spawn;
 import com.google.devtools.build.lib.actions.Spawns;
 import com.google.devtools.build.lib.actions.UserExecException;
+import com.google.devtools.build.lib.actions.cache.VirtualActionInput;
 import com.google.devtools.build.lib.analysis.BlazeDirectories;
 import com.google.devtools.build.lib.exec.TreeDeleter;
 import com.google.devtools.build.lib.exec.local.LocalEnvProvider;
@@ -38,17 +42,19 @@ import com.google.devtools.build.lib.server.FailureDetails.Sandbox.Code;
 import com.google.devtools.build.lib.shell.Command;
 import com.google.devtools.build.lib.shell.CommandException;
 import com.google.devtools.build.lib.util.OS;
+import com.google.devtools.build.lib.vfs.FileStatus;
 import com.google.devtools.build.lib.vfs.FileSystem;
 import com.google.devtools.build.lib.vfs.Path;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.build.lib.vfs.Symlinks;
+
+import javax.annotation.Nullable;
 import java.io.File;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
-import javax.annotation.Nullable;
 
 /** Spawn runner that uses linux sandboxing APIs to execute a local subprocess. */
 final class LinuxSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
@@ -229,6 +235,19 @@ final class LinuxSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
           sandboxfsMapSymlinkTargets,
           treeDeleter,
           statisticsPath);
+    } else if (getSandboxOptions().useHermetic) {
+      commandLineBuilder.setHermeticSandboxPath(sandboxPath);
+      return new HardlinkedSandboxedSpawn(
+          sandboxPath,
+          sandboxExecRoot,
+          commandLineBuilder.build(),
+          environment,
+          inputs,
+          outputs,
+          writableDirs,
+          treeDeleter,
+          statisticsPath,
+          getSandboxOptions().sandboxDebug);
     } else {
       return new SymlinkedSandboxedSpawn(
           sandboxPath,
@@ -354,6 +373,42 @@ final class LinuxSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
                 Code.MOUNT_TARGET_DOES_NOT_EXIST));
       }
     }
+  }
+  @Override
+  public void verifyPostCondition(
+      Spawn originalSpawn, SandboxedSpawn sandbox, SpawnExecutionContext context) throws IOException, ForbiddenActionInputException {
+    if(getSandboxOptions().useHermetic){
+      checkForConcurrentModifications(context);
+    }
+  }
+
+  private void checkForConcurrentModifications(SpawnExecutionContext context) throws IOException, ForbiddenActionInputException {
+    for (ActionInput input : (context.getInputMapping(PathFragment.EMPTY_FRAGMENT).values())) {
+      if (input instanceof VirtualActionInput) {
+        continue;
+      }
+
+      FileArtifactValue metadata = context.getMetadataProvider().getMetadata(input);
+      Path path = execRoot.getRelative(input.getExecPath());
+
+      try {
+        if (wasModifiedSinceDigest(metadata.getContentsProxy(), path)) {
+          throw new IOException("input dependency " + path + " was modified during execution.");
+        }
+      } catch (UnsupportedOperationException e) {
+        throw new IOException(
+          "input dependency " + path + " could not be checked for modifications during execution.",
+          e);
+      }
+    }
+  }
+
+  private boolean wasModifiedSinceDigest(FileContentsProxy proxy, Path path) throws IOException {
+      if (proxy == null) {
+        return false;
+      }
+      FileStatus stat = path.statIfFound(Symlinks.FOLLOW);
+      return stat == null || !stat.isFile() || proxy.isModified(FileContentsProxy.create(stat));
   }
 
   @Override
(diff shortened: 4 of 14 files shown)


frazze-jobb:
Related to Issue Make the sandboxed file system more strict: https://github.com/bazelbuild/bazel/issues/7313

aiuto:
It is not clear we had agreement on answer to #7313, so this PR might be premature.

frazze-jobb:
@googlebot I signed it

frazze-jobb:
This is the results from a benchmark on how long one of our builds takes with respective sandbox.

Sandbox type | Average build-time
Process-wrapper | 268.93
Linux-sandbox | 287.26
Hermetic Linux-sandbox | 271.37

justinhorvitz:
I will let philwo handle this review, since I'm not that familiar with sandboxes.

src/main/tools/linux-sandbox-options.cc
```diff
@@ -170,6 +172,19 @@ static void ParseCommandLine(unique_ptr<vector<char *>> args) {
...
+        if (opt.sandbox_root.empty()) {
+          std::string sandbox_root(optarg);
+          // Make sure that the sandbox_root path has no trailing slash.
+          if (sandbox_root.back() == '/') {
+            opt.sandbox_root.assign(sandbox_root, 0, sandbox_root.length() - 1);
+          } else {
+            opt.sandbox_root.assign(sandbox_root);
+          }
+        } else {
+          Usage(args->front(), "Multiple sandbox roots (-S) specified, expected one.");
```
larsrc-google, line 185: It should be a lower-case 's' in the error message.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -310,11 +386,21 @@ static void MakeFilesystemMostlyReadOnly() {
 }
 
 static void MountProc() {
-  // Mount a new proc on top of the old one, because the old one still refers to
-  // our parent PID namespace.
-  if (mount("/proc", "/proc", "proc", MS_NODEV | MS_NOEXEC | MS_NOSUID,
-            nullptr) < 0) {
-    DIE("mount");
+  if(!opt.hermetic){
```
larsrc-google, line 389: Please avoid negated conditions.

larsrc-google (commented):
Not quite done with the C++/Bash parts yet, but here's a couple of notes.

larsrc-google:
Could you fix the tests?

frazze-jobb:
Thanks for your review comments. Regarding the test I thought I knew what to fix, turns out I was wrong, how can I read the logs produced by buildkite?

larsrc-google:
Click the `Details` link at the failing test, it'll show you the log. But that's the log for the entire shard. Click "Artifacts" to see the output from the failing tests. Once you know what tests are failing, run them locally (possibly with ibazel).

frazze-jobb:
I seem to have solved most of the tests, but how do I prevent "buildkite/bazel-bazel-github-presubmit/darwin-openjdk-8-shard-2" from running my hermetic "linux-sandbox" test. I assume linux-sandbox is not supported on that one, and therefore the tests will be failing

larsrc-google:
Looking at [the logs](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/f477f219-eecc-4334-bf35-0f04a984a277/src/test/shell/bazel/bazel_hermetic_sandboxing_test/attempt_1.log), one test is failing because the error message is different, one is passing, and one is failing for actually not being hermetic. The last one is the interesting one. For that, I would suggest doing a platform check [like this](https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/cc_integration_test.sh#L594).

frazze-jobb:
The tests are passing now, I have dealt with @larsrc-google initial review comments. I would appreciate some more review comments @philwo

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
```diff
@@ -21,38 +21,60 @@
 
 /**
  * In case we can't get a fast digest from the filesystem, we store this metadata as a proxy to the
- * file contents. Currently it is a pair of a relevant timestamp and a "node id". On Linux the
- * former is the ctime and the latter is the inode number. We might want to add the device number in
+ * file contents. Currently it is two timestamps and a "node id". On Linux we
+ * use both ctime and mtime and inode number. We might want to add the device number in
```
larsrc-google, line 25: Shouldn't mtime be included in `equals()/hashCode()`, then? And possibly `addToFingerprint()`?

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
```diff
@@ -21,38 +21,60 @@
...
+    this.nodeId = nodeId;
+  }
+
   /**
    * Visible for serialization / deserialization. Do not use this method, but call {@link #create}
    * instead.
    */
   public FileContentsProxy(long ctime, long nodeId) {
     this.ctime = ctime;
+    this.mtime = ctime;
```
larsrc-google, line 58: Don't know if mtime==ctime gives the right result, but not whether it matters either. Preferably I would get rid of this, but I don't know what the serialization requires.

src/main/java/com/google/devtools/build/lib/sandbox/LinuxSandboxedSpawnRunner.java
```diff
@@ -359,6 +372,40 @@ private void validateBindMounts(SortedMap<Path, Path> bindMounts) throws UserExe
...
+
+      FileArtifactValue metadata = context.getMetadataProvider().getMetadata(input);
+      Path path = execRoot.getRelative(input.getExecPath());
+
+      try {
+        if (wasModifiedSinceDigest(metadata.getContentsProxy(), path)) {
+          throw new IOException("input dependency " + path + " was modified during execution.");
+        }
+      } catch (UnsupportedOperationException e) {
+        throw new IOException("input dependency " + path + " could not be checked for modifications during execution.");
```
larsrc-google, line 397: Good idea to add the path information to the exception, but please set the original exception as the cause.

src/main/java/com/google/devtools/build/lib/sandbox/AbstractSandboxSpawnRunner.java
```diff
@@ -140,6 +143,12 @@ private SpawnResult runSpawn(
       }
     }
   }
+  /**
+   * Override this method if you need to run a post condition after the action has executed
+   */
+  public void verifyPostCondition(Spawn originalSpawn, SandboxedSpawn sandbox, SpawnExecutionContext context) throws IOException {
+    return;
```
larsrc-google, line 150: Nit: Return statement is unnecessary.

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
```diff
@@ -21,38 +21,60 @@
...
 /**
  * In case we can't get a fast digest from the filesystem, we store this metadata as a proxy to the
- * file contents. Currently it is a pair of a relevant timestamp and a "node id". On Linux the
- * former is the ctime and the latter is the inode number. We might want to add the device number in
+ * file contents. Currently it is two timestamps and a "node id". On Linux we
+ * use both ctime and mtime and inode number. We might want to add the device number in
  * the future.
  *
- * <p>For a Linux example of why mtime alone is insufficient, note that 'mv' preserves timestamps.
+ * <p>For a Linux example of ctime alone is insufficient, and we add node id,
```
larsrc-google, line 28: This example is no longer correct. `mv` does change `ctime`. Not sure then if `inode` is required, though.

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
```diff
@@ -72,6 +94,20 @@ public boolean equals(Object other) {
...
+   * should not be considered a file modification.
+   * Is only comparing mtime and not ctime and is therefore
+   * not detecting changed metadata like permission.
+   */
+  public boolean isModified(FileContentsProxy other) {
+    if (other == this) {
+      return false;
+    }
+    // true if nodeId are different or inode has a new mtime
+    return nodeId != other.nodeId || mtime != other.mtime;
```
larsrc-google, line 108: Isn't this going to fall prey to the problems in #1525?

```
larsrc@mols:import-bazel-github-13279$ stat /tmp/a /tmp/b
  File: /tmp/a
  Size: 2         	Blocks: 8          IO Block: 4096   regular file
Device: fe01h/65025d	Inode: 2485767     Links: 1
Access: (0640/-rw-r-----)  Uid: (121084/  larsrc)   Gid: (89939/primarygroup)
Access: 2021-04-26 11:32:04.137402754 +0200
Modify: 2021-04-26 11:32:04.137402754 +0200
Change: 2021-04-26 11:32:04.137402754 +0200
 Birth: 2021-04-26 11:31:12.493443234 +0200
  File: /tmp/b
  Size: 2         	Blocks: 8          IO Block: 4096   regular file
Device: fe01h/65025d	Inode: 2485994     Links: 1
Access: (0640/-rw-r-----)  Uid: (121084/  larsrc)   Gid: (89939/primarygroup)
Access: 2021-04-26 11:32:04.137402754 +0200
Modify: 2021-04-26 11:32:04.137402754 +0200
Change: 2021-04-26 11:32:04.137402754 +0200
 Birth: 2021-04-26 11:31:12.493443234 +0200
larsrc@mols:import-bazel-github-13279$ mv /tmp/a /tmp/b
larsrc@mols:import-bazel-github-13279$ stat /tmp/a /tmp/b
stat: cannot statx '/tmp/a': No such file or directory
  File: /tmp/b
  Size: 2         	Blocks: 8          IO Block: 4096   regular file
Device: fe01h/65025d	Inode: 2485767     Links: 1
Access: (0640/-rw-r-----)  Uid: (121084/  larsrc)   Gid: (89939/primarygroup)
Access: 2021-04-26 11:32:04.137402754 +0200
Modify: 2021-04-26 11:32:04.137402754 +0200
Change: 2021-04-26 11:32:36.049377813 +0200
 Birth: 2021-04-26 11:31:12.493443234 +0200
```

src/main/tools/linux-sandbox-options.cc
```diff
@@ -73,6 +73,8 @@ static void Usage(char *program_name, const char *fmt, ...) {
           "  -R  if set, make the uid/gid be root\n"
           "  -U  if set, make the uid/gid be nobody\n"
           "  -D  if set, debug info will be printed\n"
+          "  -h  if set, use whitelisting for improved hermeticity\n"
+          "  -s  The sandbox root directory\n"
```
larsrc-google, line 77: Could you clarify how this is different from the working dir (`-W`)?

larsrc-google (commented):
Here are a few fixes that make it easier for me to test further revisions, and a couple of clarification questions. All tests pass, though.

src/main/tools/linux-sandbox-options.cc
frazze-jobb: I guess working dir is enough, if we can assume that working dir will be sandbox_dir + /execroot/workspace_name>, then it can be derived from that.

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
frazze-jobb: You are right.
I think this example is confusing to me in the way it is written.
Since mtime was not present in the class, but referenced in the documentation,
I assumed they meant ctime here, and that they added nodeId as a complement.
I will revert this change in the documentation. But I still don't know what it is trying to explain, I guess then what it tries to tell me is why ctime is preferred over mtime?

src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Creates an execRoot for a Spawn that contains input files as hardlinks to their original
+ * destination.
+ */
+public class HardlinkedSandboxedSpawn extends AbstractContainerizingSandboxedSpawn {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private boolean sandbox_debug = false;
```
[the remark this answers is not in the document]
frazze-jobb: There is no sandboxDebug in scope as far as I can tell are you suggestng a rename of the variable?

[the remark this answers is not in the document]
larsrc-google: Yes, lowerCamelCase is the standard style for local variables.

src/main/java/com/google/devtools/build/lib/actions/FileContentsProxy.java
frazze-jobb: In your example, mtime is same after the overwrite of (b) but nodeId is different so we will be detecting this modification, if I understand your concern correctly.

larsrc-google:
@janakdr is probably the one who would know about the mtime handling.

janakdr (commented):
@michajlo wrote that test, he may have a better idea. On its face, I don't see that the test is enforcing an important property.

michajlo:
That was a while ago... Skimming history I believe I was showing that mtime was significant to equality (the version I introduced was actually asserting not-equal). At some point it got flipped to assert equal when mtime was removed / deemed unnecessary?

janakdr:
Yes, @michajlo is right, I'm apparently the one who flipped it from not-equal to equal. I'm ok with getting rid of that assertion or changing it back again.

frazze-jobb:
I think I uplifted this branch incorrectly since it added a lot of other contributors.. How do I get out of this mess? Create a new pull request?

lberki:
Cleaning up the reviewer list since it appears that @larsrc-google is the one who's doing the actual review.

frazze-jobb:
@philwo accidentaly removed from reviewer list

frazze-jobb:
Any more comments? @larsrc-google @philwo

thekyz:
any reason why this is not getting merged?

larsrc-google:
My apologies for letting this slip. I haven't fully reviewed the C++ code yet. I'm much less familiar with that code than with the Java code. @philwo, adding your eyes to that would be greatly appreciated.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -310,11 +364,21 @@ static void MakeFilesystemMostlyReadOnly() {
...
-            nullptr) < 0) {
-    DIE("mount");
+  if(opt.hermetic){
+    // Create a bind mount to /proc
+    if (CreateTarget("proc", true) < 0) {
+      DIE("CreateTarget");
+    }
+    if (mount("/proc", "proc", NULL, MS_REC | MS_BIND, NULL) < 0) {
+      DIE("mount");
+    }
```
philwo, line 374: What's the reason that we no longer have to mount a new proc on top of the old one, because the old one still refers to our parent PID namespace, but can just use a bind mount in this case?

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -53,9 +53,83 @@
 #include "src/main/tools/linux-sandbox.h"
 #include "src/main/tools/logging.h"
 #include "src/main/tools/process-tools.h"
+#include "src/main/cpp/util/file_platform.h"
+#include "src/main/cpp/util/file.h"
+#include "src/main/cpp/util/strings.h"
+
+using blaze_util::IsDirectory;
+using blaze_util::ToString;
+using blaze_util::WriteFile;
```
philwo, line 62: Could we replace these imported methods by copying the code over? If I'm missing something, please correct me, but it seems like we're pulling in whole library just to get these three methods. I would prefer to stick to our own versions here and keep it simple:

- `ToString(x)` is just `std::to_string(x)`.
- `IsDirectory(x)` is one `stat` call and a condition.
- A version of `WriteFile` already existed in this file.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -349,8 +413,13 @@ static void SetupNetworking() {
 }
 
 static void EnterSandbox() {
```
philwo, line 415: WDYT, should we rename this to EnterWorkingDirectory?

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -430,6 +499,112 @@ static int WaitForChild() {
...
+    DIE("mount");
+  }
+  if (chdir(opt.sandbox_root.c_str()) < 0) {
+    DIE("chdir(%s)", opt.sandbox_root.c_str());
+  }
+}
+
+static void CreateEmptyFile() {
+  // This is used as the base for bind mounting.
+  CreateTarget("tmp", true);
```
philwo, line 513: Should we check the return value of CreateTarget for errors?

philwo (commented):
In general LGTM, added a few comments.

larsrc-google:
@philwo I was actually the one who suggested using the methods from blaze_util instead of adding yet another reimplementation. Seems like that's what a util class is for. Is this util particularly heavy? Should it maybe be split?

philwo:
@larsrc-google But the one method was already there and worked fine and the other two are single lines of code. I think this is taking "DRY" a bit too far.

The sandbox binary is time-tested, performance and correctness critical code, I can't even tell for sure whether this PR will not break something subtle that we will only discover months down the line. I would really like to keep it as simple as possible so you can read it in one go and reason about it and remove any abstractions or dependencies on "generic" implementations of functions.

larsrc-google:
@philwo My original comment (https://github.com/bazelbuild/bazel/pull/13279/files/0a9c8f7dc80255e925189c5ac94099fca5a172f3#r608589967) was not intended to apply to existing code so much as to the new implementations being added. Adding new implementations where time-tested ones exist makes it more likely that something will break.

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: I could have a look, this code is taken from old linux-sandbox implementation, but maybe its not needed.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -53,9 +53,83 @@
...
 
 static int global_child_pid;
 
+// Helper methods
+static void CreateFile(const char *path) {
+  int handle = open(path, O_CREAT | O_WRONLY | O_EXCL, 0666);
+  if (handle < 0) {
+    DIE("open");
+  }
+  if (close(handle)) {
```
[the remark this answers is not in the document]
frazze-jobb: sure

src/main/tools/linux-sandbox-options.cc
```diff
@@ -170,6 +174,19 @@ static void ParseCommandLine(unique_ptr<vector<char *>> args) {
                 "Cannot write stats to more than one destination.");
         }
         break;
+      case 's':
+        if (opt.sandbox_root.empty()) {
+          std::string sandbox_root(optarg);
+          // Make sure that the sandbox_root path has no trailing slash.
+          if (sandbox_root.back() == '/') {
```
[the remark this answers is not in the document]
frazze-jobb: will fix

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: Have you agreed on this?

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: Yes

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -53,9 +53,83 @@
...
+// This is over two times faster than creating empty files via open() on
+// certain filesystems (e.g. XFS).
+static void LinkFile(const char *path) {
+  if (link("tmp/empty_file", path) < 0) {
+    DIE("link");
+  }
+}
+
+// Recursively creates the file or directory specified in "path" and its parent
+// directories.
```
larsrc-google, line 87: Please add that this sets errno and returns -1 for errors.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -53,9 +53,83 @@
...
+    // If stat failed because of any error other than "the path does not exist",
+    // this is an error.
+    if (errno != ENOENT) {
+      return -1;
+    }
+  }
+
+  // Create the parent directory.
+  if (CreateTarget(dirname(strdupa(path)), true) < 0) {
+    DIE("CreateTarget");
```
larsrc-google, line 119: Adding the name of the directory we failed to create would be helpful here (but easier if CreateTarget dies in all errors).

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -53,9 +53,83 @@
...
+    DIE("close");
+  }
+}
+
+// Creates an empty file at 'path' by hard linking it from a known empty file.
+// This is over two times faster than creating empty files via open() on
+// certain filesystems (e.g. XFS).
+static void LinkFile(const char *path) {
+  if (link("tmp/empty_file", path) < 0) {
+    DIE("link");
```
larsrc-google, line 82: Please add the file we were trying to link to in the message.

larsrc-google (commented):
You seem to have a merge conflict pending.

src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (source.isFile(Symlinks.NOFOLLOW)) {
+      try {
+        source.createHardLink(target);
+      } catch (IOException e) {
+        if (sandboxDebug) {
+          logger.atInfo().log("File %s could not be hardlinked, file will be copied instead.", source);
+        }
+        FileSystemUtils.copyFile(source, target);
```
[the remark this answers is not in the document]
frazze-jobb: We copy since we can't exit the sandbox, and the filepath is probably not bind mounted

src/main/tools/linux-sandbox-options.cc
```diff
@@ -73,6 +73,10 @@ static void Usage(char *program_name, const char *fmt, ...) {
           "  -R  if set, make the uid/gid be root\n"
           "  -U  if set, make the uid/gid be nobody\n"
           "  -D  if set, debug info will be printed\n"
+          "  -h  if set, create a chroot in the sandbox directory (-s) and only "
```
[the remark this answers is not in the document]
frazze-jobb: I agree

src/main/tools/linux-sandbox-pid1.cc
philwo: I agree that using a library instead of reinventing the wheel for complex code is important, but... where is the complex code that we're using from `src/main/tools`? (Honest question, maybe I'm just missing something here!)

larsrc-google: Looking back at the original code, there were fewer new duplicated functions than I thought at the time. So it's fine.

philwo: Thanks for checking! 😊

src/main/tools/linux-sandbox-pid1.cc
philwo: Thanks! I remember this being tricky, but in the end we found a way that worked. 🤔 A good way to check whether /proc is working correctly is that when you manually run `linux-sandbox` e.g. with a shell and you look inside /proc, you shouldn't see any processes from outside the sandbox.

src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
```diff
@@ -0,0 +1,94 @@
...
+      }
+    } else if (source.isDirectory()) {
+      if (source.startsWith(target)) {
+        throw new IllegalArgumentException(source + " is a subdirectory of " + target);
+      }
+      target.createDirectory();
+      Collection<Path> entries = source.getDirectoryEntries();
```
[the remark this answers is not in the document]
frazze-jobb: this seemed not so trivial, so I decided not to do anything about this.
I will take a look if you are really concerned by the current implementation of this.

src/main/tools/linux-sandbox-options.cc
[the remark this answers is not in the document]
frazze-jobb: decided for keeping combining -h and -s <file> flag to just a -h <file> flag

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: done

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: done

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: done

src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
```diff
@@ -0,0 +1,94 @@
...
+        statisticsPath);
+    this.sandboxDebug = sandboxDebug;
+  }
+
+  @Override
+  protected void copyFile(Path source, Path target) throws IOException {
+    hardLinkRecursive(source, target);
+  }
+
+  private void hardLinkRecursive(Path source, Path target) throws IOException {
```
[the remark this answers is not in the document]
frazze-jobb: done

src/main/java/com/google/devtools/build/lib/sandbox/HardlinkedSandboxedSpawn.java
```diff
@@ -0,0 +1,94 @@
...
+      Path sandboxPath,
+      Path sandboxExecRoot,
+      List<String> arguments,
+      Map<String, String> environment,
+      SandboxInputs inputs,
+      SandboxOutputs outputs,
+      Set<Path> writableDirs,
+      TreeDeleter treeDeleter,
+      @Nullable Path statisticsPath,
+      Boolean sandboxDebug) {
```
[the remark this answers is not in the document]
frazze-jobb: done

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: I think it makes sense, done

src/main/tools/linux-sandbox-pid1.cc
frazze-jobb: Yes it works

src/main/tools/linux-sandbox-options.cc
```diff
@@ -73,6 +73,10 @@ static void Usage(char *program_name, const char *fmt, ...) {
...
+          " mount whats been specified with -M/-m for improved hermeticity\n"
+          "  -s  The sandbox root directory where the chroot will be created, -W"
+          " should be a folder inside this sandbox root directory\n"
```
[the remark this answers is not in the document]
frazze-jobb: Added it in ParseCommandLine

frazze-jobb:
Back from vacation now, 
I had to reset the branch, sync it and squash all the changes, because the branch had turned real nasty with my and other peoples commits appearing multiple times, hopefully it doesn't ruin the review to much. I no longer have failing tests at least.

I have acted on all your comments @philwo @larsrc-google. 
Is it good enough for a merge?

larsrc-google:
I agree. I have imported it and all tests pass. There are a few style things to fix, I'll take care of that.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -67,8 +67,100 @@
...
+  } else {
+    // If stat failed because of any error other than "the path does not exist",
+    // this is an error.
+    if (errno != ENOENT) {
+      return -1;
+    }
+  }
+
+  // Create the parent directory.
+  if (CreateTarget(dirname(strdupa(path)), true) < 0) {
```
theimeto, line 149: strdup here and on the next line leaks the memory.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -460,11 +649,22 @@ int Pid1Main(void *sync_pipe_param) {
...
   }
-  MountFilesystems();
-  MakeFilesystemMostlyReadOnly();
-  MountProc();
+
+  if (opt.hermetic) {
+    MountSandboxAndGoThere();
+    CreateEmptyFile();
+    MountDev();
+    MountProc();
```
theimeto, line 657: This mounts /proc on /proc not on sandbox-root/proc; after SYS_pivot_root syscall it will be lost.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -442,6 +518,119 @@ static int WaitForChild() {
...
+      DIE("CreateTarget tmp")
+  }
+  CreateFile("tmp/empty_file");
+}
+
+static void MountDev() {
+  if (CreateTarget("dev", true) < 0) {
+    DIE("CreateTarget /dev");
+  }
+  const char *devs[] = {"/dev/null", "/dev/random", "/dev/urandom", "/dev/zero", NULL};
```
theimeto, line 542: `/dev/shm` might be a good idea (should also be writable) - it is required if python code is to be executed and it happen to use `multiprocessing` module.

ulrfa: The sandbox would be less hermetic if /dev/shm is bind mounted, since then actions could communicate with each other via /dev/shm.

Would it make sense to require an explicit --sandbox_add_mount_pair for use cases where /dev/shm is desired to be bind mounted?

Issue #12559, opened 2020-11-25
NullPointerException when overriding copts and linkopts to None in transition


stefanbucur:
### Description of the problem / feature request:

When overriding the `copts` and `linkopts` build configuration with `None` values in a transition, Bazel crashes with a NullPointerException.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```py
def _my_transition_impl(settings, attr):
    return {
        "//command_line_option:copt": None,
        "//command_line_option:linkopt": None,
    }

my_transition = transition(
    implementation = _my_transition_impl,
    inputs = [],
    outputs = [
        "//command_line_option:copt",
        "//command_line_option:linkopt",
    ],
)
```

### What operating system are you running Bazel on?

Debian-based Linux.

### What's the output of `bazel info release`?

`release 3.7.1`

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```
git@github.com:bazelbuild/rules_fuzzing
f782cd326f15a1182a3c293cf77f3a21f0b24ed0
b0e592b75de5680599ee81ccec6b073e93a21058
```

###  Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

The exception stack trace:

```
Internal error thrown during build. Printing stack trace: java.lang.RuntimeException: Unrecoverable error while evaluating node 'BuildConfigurationValue.Key[f380618cb25b3c02e111c0b5e52b926621d76dbf320be25008319e76d414f029]' (requested by nodes 'ConfiguredTargetKey{label=//examples:re2_fuzz_test_instrum, config=BuildConfigurationValue.Key[8f956e0a5a332769538bb1785bbb03d792f84521f14f8711180a17d7de3e07e8]}')
        at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:513)
        at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
        at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(Unknown Source)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
[...]

PR #13286, opened 2021-04-01
Throw exception if 'None' value given to List-type command line options in Bazel transitions

tobiade:
* Proposed fix for #12559 
* Command line options specified as Java List types should default to empty when given 'None' Starlark value for Bazel transitions

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java
@@ -333,7 +333,7 @@ public class FunctionTransitionUtil {
                           optionValueAsList.stream().map(Object::toString).collect(joining(",")));
             }
           } else if (def.getType() == List.class && optionValue == null) {
-            convertedValue = def.getDefaultValue();
+            throw ValidationException.format("'None' value not allowed for List-type option '%s'. Please use '[]' instead if trying to set option to empty value.", optionName);
           } else if (optionValue == null || def.getType().isInstance(optionValue)) {
             convertedValue = optionValue;
           } else if (def.getType().equals(boolean.class) && optionValue instanceof Boolean) {
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java
@@ -332,6 +332,8 @@ public class FunctionTransitionUtil {
                       .convert(
                           optionValueAsList.stream().map(Object::toString).collect(joining(",")));
             }
+          } else if (def.getType() == List.class && optionValue == null) {
+            convertedValue = def.getDefaultValue();
           } else if (optionValue == null || def.getType().isInstance(optionValue)) {
             convertedValue = optionValue;
           } else if (def.getType().equals(boolean.class) && optionValue instanceof Boolean) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/analysis/starlark/FunctionTransitionUtil.java
```diff
@@ -332,6 +332,8 @@ private static BuildOptions applyTransition(
                       .convert(
                           optionValueAsList.stream().map(Object::toString).collect(joining(",")));
             }
+          } else if (def.getType() == List.class && optionValue == null) {
+            convertedValue = def.getDefaultValue();
```
tobiade, line 336: Please let me know if it's okay to replace with the default value here or if we should be using `StarlarkList.empty()` instead.

tobiade: I'm also curious if it's okay to perform the override here, or if it makes sense to do it within this block of code above:
```
        if (optionValue instanceof NoneType) {
          optionValue = null;
        }
```
I need the `OptionDefinition` to check if it's a `List`, so the current spot seemed more convenient.

gregestren: About the first question, it's a good question. My first intuition was `None` = `empty`, but there really is something to say for it meaning `default`. Especially if we already have ways to express `empty` (`""`, `[]`).

There's a sensible interpretation that says `None` = `unset`, which implies the default. 

@brandjon - does `None` have any incompatible implications in Starlark? Can

```
   "//command_line_option:copt": None,
``` 
(in a transition) reasonably imply unsetting the flag?

gregestren: I just posted an update in the associated issue. I'm still trying to wrap my mind around the three conceptual options:

1. disallow `None`
1. `None` = empty
1. `None` = the flag's default value (as you currently have encoded).

tobiade: Thanks @gregestren. Makes sense. I think if a use-case exists where a user wants to set a flag back to its default (without actually knowing what the default value is), then maybe `None` might be a good way to achieve this.

gregestren: Okay, I also asked some other core devs. including @brandjon who curates the core Starlark API.

The short consensus, which I find compelling, is these are intriguing design issues. But they're not trivial. Not only are all the above interpretations possible, but `None` could potentially be a distinct value in its own right, since it's a legitimate value Starlark variables can take.

So basically, we shouldn't try to answer these questions yet. They deserve a deeper and broader design about what flag parsing looks like and how it mixes with Starlark. So if we simply disallow `None` with a clear error message now, we've solved the crash in https://github.com/bazelbuild/bazel/issues/12559, can give users clear enough guidance with the error message, and aren't narrowing our options yet for however we'd ultimately want this to work.

How does that sound to you?

tobiade: Understood - that makes sense. I'll make that change, thanks for the following up!

tobiade:
@gregestren Updated as discussed.

gregestren (approved):
LGTM, thanks!

gregestren:
Please hold for the PR to auto-merge in a timely manner...

PR #13314, opened 2021-04-08
Fix stripping of macOS loadable bundles

### LINKED ISSUE #11869: macOS frameworks: symbol stripping fails
Hi wonderful Bazel folks,

It seems something has broken in the stripping of macos_bundle in the past few versions. TL;DR: to fix the issue, I think we should pass a slightly less zealous option to strip (-x)--and maybe fix stripopts.

In more detail:
Trying to build a framework with  --compilation_mode=opt leads to 
`error: symbols referenced by indirect symbol table entries that can't be stripped`
Debug builds work fine. iOS frameworks work fine. This worked in Bazel 2.2, but is failing in 3.4.
(You do need to pass --objc_enable_binary_stripping to see the issue, since otherwise Bazel doesn't try to strip objc binaries.)

I'm pretty sure I've seen this issue in Xcode projects a long time ago, when one tries to mistakenly strip "All Symbols" instead of "Non-Global Symbols" (i.e. defaulting to `strip` instead of `strip -x`). You can't pass this in through bazel because --strip_opt doesn't seem to work with macos_bundle...but I've confirmed manually that it fixes the issue.  Apple does note in the man page for strip that "For dynamic shared libraries, the maximum level of stripping is usually -x (to remove all non-global symbols)."

Thanks so much,
Chris

cpsauer:
Adds `-x` flag to stripping of macOS loadable bundles.

Loadable bundles--i.e. truly dynamically loadable libraries on macOS--cannot be stripped without this flag, since you'd be trying to strip away the all symbols, including those used for dynamic loading. Doing so results in `error: symbols referenced by indirect symbol table entries that can't be stripped`.

`-x` instead leads to the removal of the unneeded local symbols. As Apple notes in their man page: "For dynamic shared libraries, the maximum level of stripping is usually -x (to remove all non-global symbols)."

This should fix #11869

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -1116,6 +1116,9 @@ public class CompilationSupport {
     if (Iterables.contains(extraLinkArgs, "-dynamiclib")) {
       return StrippingType.DYNAMIC_LIB;
     }
+    if (Iterables.contains(extraLinkArgs, "-bundle")) {
+      return StrippingType.LOADABLE_BUNDLE;
+    }
     if (Iterables.contains(extraLinkArgs, "-kext")) {
       return StrippingType.KERNEL_EXTENSION;
     }
@@ -1622,6 +1625,7 @@ public class CompilationSupport {
   private enum StrippingType {
     DEFAULT,
     DYNAMIC_LIB,
+    LOADABLE_BUNDLE,
     KERNEL_EXTENSION
   }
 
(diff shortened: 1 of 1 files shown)


cpsauer:
(I understand that --objc_enable_binary_stripping [has other issues](https://github.com/bazelbuild/bazel/pull/13123), but this seems like a clear bugfix that unblocks some useful functionality.)

allevato:
cc @trybka

trybka:
This seems like a perfectly reasonable tweak to improve bundle usage. 👍  Thanks!

allevato (approved):
Hopefully we can get these hardcoded conditions/flags out of Bazel entirely at some point and moved into crosstool features instead, but adding one more check here doesn't really set us back in the meantime, either.

cpsauer:
Sweet! Thanks, all. Where do we go from here on merging?

cpsauer:
[With you, @allevato: Features look like a great, additional layer of indirection to wrap flags that differ across platforms. But in the meantime.]

cpsauer:
Hey all, quick bump. Now That it's approved, what's the path to merging?

lberki:
I sent it to internal code review to @allevato right now

cpsauer:
Sweet! Thanks, all!

PR #13319, opened 2021-04-09
In example code for ctx.actions.run, put args in a list

martinboehme:
Per the documentation for `ctx.actions.run`, the `arguments` parameter should be a list of strings or `actions.args()` objects:

https://docs.bazel.build/versions/master/skylark/lib/actions.html#run

When I try to pass a single `actions.args()` object in my own code, I get an error, so I believe the code sample is wrong.


aiuto (approved):

aiuto:
I'll merge.

PR #13322, opened 2021-04-09
Upgrade Jacoco to 0.8.6 and asm to 9.1

comius:
Fixes https://github.com/bazelbuild/bazel/issues/11674


comius:
This requires a complicated dance of 3 PRs: first add second jacoco to third_party, second use it everywhere, third remove the second jacoco from third_party. @philwo what do you think about merging this first directly and then submitting it over copybara? Or will copybara go crazy?

davido:
This was merged already?

PR #13327, opened 2021-04-12
Don't rely on dirname when invoking wrapped_clang

thii:
Its path can already be resolved from the repo name.


thii:
Doesn't seem to work with `--experimental_disable_external_package` and `--experimental_sibling_repository_layout`.

keith:
was this causing an issue or just hoping to optimize?

thii:
just an optimization, after seeing other repository rules resolve their tool paths like this.

Issue #13330, opened 2021-04-12
Bazel crashes when user sets their own `DEVELOPER_DIR`


thii:
### Description of the problem / feature request:

Bazel crashes when the user sets their own `DEVELOPER_DIR` environment variable.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Set `DEVELOPER_DIR` environment variable to a value in the Apple CROSSTOOL file and build with that.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 4.0.0

### Any other information, logs, or outputs that you want to share?

```
SUBCOMMAND: # //:main [action 'Compiling main.m', configuration: bcb20af40ce0ec204cc0564ea50f3ad16d447e6a99fa61ace3503a019411c0e4, execution platform: @local_config_platform//:host]
(cd /private/var/tmp/_bazel_admin/eac612c5359b6f8a5bb45f21e493eb6a/execroot/__main__ && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=14.4 \
    DEVELOPER_DIR=external/xcode_toolchain/developer_dir \
    PATH=external/xcode_toolchain/developer_dir/Toolchains/XcodeDefault.xctoolchain/usr/bin:/bin:/usr/bin \
    XCODE_VERSION_OVERRIDE=12.4.0.12D4e \
    ZERO_AR_DATE=1 \
[...]

PR #13331, opened 2021-04-12
Use user-defined `DEVELOPER_DIR` environment variable if it's set

thii:
This fixes a Bazel crash if the user defines their own `DEVELOPER_DIR`
environment variable via a custom Xcode toolchain.

Fixes https://github.com/bazelbuild/bazel/issues/13330.


allevato (approved):
I _think_ this is ok and won't interfere with any of our internal build environments. I don't see any places where an externally set `DEVELOPER_DIR` could leak into the action env, but we'll just need to pull it in and test it to be sure I haven't overlooked anything.

Issue #13188, opened 2021-03-09
experimental_sibling_repository_layout seems to create remote execution actions with invalid working_directory/output relationships


illicitonion:
### Description of the problem / feature request:

When testing out an unrelated feature, using remote execution (with buildbarn), I ran into issues where Bazel errored that it could not compile DumpPlatformClassPath.class because:
```
JavaToolchainCompileClasses tools/jdk/platformclasspath_classes/DumpPlatformClassPath.class failed: not all outputs were created or valid
```

I'm running in a workspace with name `example_all`.

As far as I can tell, this error is because on the `Command` we're setting `working_directory: example_all` but also prefixing the `output_files` with `example_all/`, for example the `output_files` for my execution is `example_all/bazel-out/bazel_tools/darwin-fastbuild/bin/tools/jdk/platformclasspath_classes/DumpPlatformClassPath.class`.

The remote execution API specifies that `output_files` should be relative to the `working_directory`, rather than relative to the `input_root`, so I believe this prefixing shouldn't be happening. See this section of the remote API spec:

https://github.com/bazelbuild/remote-apis/blob/0943dc4e70e1414735a85a3167557392c177ff45/build/bazel/remote/execution/v2/remote_execution.proto#L547-L551

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run a buildbarn cluster, and try to build any Java with `--experimental_sibling_repository_layout=true`.

### What operating system are you running Bazel on?

The Bazel client is on macOS, the buildbarn cluster is running Ubuntu Focal.

### What's the output of `bazel info release`?

development version

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

Checked out f1826459f21b5ee8aec4f93686dd7ce8d063466e and ran `bazel build //src:bazel`

If I revert 7149f578006a4ad0d51df69830a6986749b34df5 this error goes away (though others are introduced), so I'll cc @lberki who may have some thoughts :)

lberki:
This is probably because RBE has a bug where the output directory is interpreted relative to `input_root` instead of according to the spec, so this is what Bazel did. 

@eytankidron, do I understand correctly that if Bazel makes it so that output file paths are relative to `working_directory`, it will keep working with RBE since it's been partially fixed there?

@coeuvre , can you take this?

eytankidron:
@lberki that is correct. This bug is now fixed in RBE. Sending output paths relative to the working directory will now work.

[...]

PR #13339, opened 2021-04-13
Remote: Use execRoot as input root and do NOT set working directory by default.

coeuvre:
When --experimental_sibling_external_layout is set, use the parent directory of execRoot as input root and set working directory to the base name of execRoot. Paths of output files are relative to working directory.

When --incompatible_remote_paths_relative_to_input_root is set, paths of output files are relative to input root.

Introduce RemotePathResolver which is used to convert local paths to remote paths and vice versa. We should prefer this class instead of using execRoot directly whenever possible in remote module. execRoot usages in RemoteCache are all replaced.

On Windows, shared action results cache for cl.exe across different workspaces causing header dependency checking to fail. This was initially fixed by #9172, but is broken after cl/356735700. The reason test didn't fail before this change is that two builds from different workspaces do not share the cache since input files are relative to the parent of exec root which contains workspace name. This change fixes that by adding workspace name as platform property for action running on Windows.

Fixes https://github.com/bazelbuild/bazel/issues/13188.


src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -699,6 +697,22 @@ private SpawnResult handleError(
...
   }
 
+  static List<Property> getExtraPlatformProperties(Spawn spawn, Path execRoot) {
+    // Hack on Windows. The included headers dumped by cl.exe in stdout contain absolute paths.
+    // When compiling the file from different workspace, the shared cache will cause header
+    // dependency checking to fail. This was initially fixed by a hack (see
+    // https://github.com/bazelbuild/bazel/issues/9172 for more details), but is broken again due
+    // to cl/356735700. We include workspace name here so action results from different workspaces
+    // on Windows won't be shared.
+    boolean isWindows = System.getProperty("os.name").startsWith("Windows");
```
lberki, line 707: This is the operating system Bazel runs on and that isn't necessarily the same as the remote operating system. In addition, this AFAIU needlessly decreases the cache hit rate for actions that do not invoke MSVC.

hat said, I don't know off the bat how to tell what the remote operating system is. I'm wondering if it's better to fix this by looking at the `PARSE_SHOWINCLUDES` feature in the C++ compile action. It should be possible to add the necessary property to the Spawn constructed in `CppCompileAction.createSpawn()`, which has the additional advantage that the hack stays in C++ code which requires it.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -699,6 +697,22 @@ private SpawnResult handleError(
...
+    if (isWindows) {
+      return ImmutableList
+          .of(Property.newBuilder().setName("workspace").setValue(execRoot.getBaseName()).build());
```
lberki, line 710: Do I understand correctly that `workspace` is not interpreted by RBE and is just to differentiate the cache keys? In that case, call it something more descriptive, e.g. `BAZEL_WORKSPACE_DISAMBIGUATION_FOR_MSVC_HACK`.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * A {@link RemotePathResolver} used when {@code --experimental_sibling_external_layout} is set.
+   * Use parent directory of {@code execRoot} and set {@code workingDirectory} to the base name of
+   * {@code execRoot}.
+   *
+   * The paths of outputs are relative to {@code workingDirectory} if
+   * {@code --incompatible_remote_output_paths_relative_to_input_root} is not set, otherwise,
+   * relative to input root.
+   */
+  class SiblingExternalLayoutResolver implements RemotePathResolver {
```
lberki, line 141: nit: `SiblingRepositoryLayout` for consistency with the name of the flag?

src/main/java/com/google/devtools/build/lib/remote/merkletree/DirectoryTreeBuilder.java
```diff
@@ -61,6 +62,13 @@ static DirectoryTree fromActionInputs(
       throws IOException {
     Map<PathFragment, DirectoryNode> tree = new HashMap<>();
     int numFiles = buildFromActionInputs(inputs, metadataProvider, execRoot, digestUtil, tree);
+    // // Make sure working directory is exists
```
lberki, line 65: debugging code left in?

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -130,16 +131,16 @@ public ActionResult downloadActionResult(
    */
   public ActionResult upload(
```
lberki, line 132: My understanding is that the lifetime of a RemoteCache the a single command. Then why does one need to pass in a `RemotePathResolver` as opposed to making it a member of the class? (same for a number of other places below)

lberki (commented):
Initial round. I'm not in very good shape today so I probably don't grok everything :(

coeuvre:
Note that there is a pending change for REAPI about the semantics of output paths https://github.com/bazelbuild/remote-apis/pull/191. We will update in a separate PR accordingly once it is merge.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
coeuvre: > This is the operating system Bazel runs on and that isn't necessarily the same as the remote operating system.

This is used for remote cache only (which means local execution). For executing cl.exe remotely, RBE has the hack to remove the workspace name on the server side. And this requires other server implementation to do the same hack if they want to support compiling c++.

I had a discussion with @meteorcloudy yesterday about a long term fix for this. It seems we can solve the issue at Bazel side, but requires lots of changes. I will share the details later.

> In addition, this AFAIU needlessly decreases the cache hit rate for actions that do not invoke MSVC.

Note that, if we use parent of exec root as input root, the paths of input files include workspace name (input file paths are used to compute cache key) which implies the same thing: caches across different workspace won't be shared. So I won't concern too much about this.

> I'm wondering if it's better to fix this by looking at the PARSE_SHOWINCLUDES feature in the C++ compile action.

Good point! I will have a look.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
coeuvre: Exactly!

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
coeuvre: `RemotePathResolver` is a drop in replacement for the usages of `execRoot` inside remote module. The question then becomes why did we pass in `execRoot` instead of making it a member of `RemoteCache`. Probably due to the time when we construct `RemoteCache` (inside `beforeCommand`), `execRoot` is not available yet. I can change this in another PR but let's make this PR only focus on changes for output paths semantics.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/merkletree/DirectoryTreeBuilder.java
coeuvre: Oops, removed.

src/main/java/com/google/devtools/build/lib/analysis/platform/PlatformUtils.java
```diff
@@ -101,10 +103,18 @@ public static Platform getPlatformProto(Spawn spawn, @Nullable RemoteOptions rem
...
     }
 
+    platformBuilder.addAllProperties(extraProperties);
+
     sortPlatformProperties(platformBuilder);
     return platformBuilder.build();
   }
 
+  @Nullable
+  public static Platform getPlatformProto(Spawn spawn, @Nullable RemoteOptions remoteOptions)
```
lberki, line 113: nit: `getPlatformProto()` only has four call sites, so this overload is probably unnecessary

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
lberki: ...or else `execRoot` just ended up being there without any rhyme or reason. We'd have to take a look at the change history to tell, but it doesn't really matter, so meh. 

I think the question is more important for `RemotePathResolver` because it's an object with potentially a lot of state, unlike `Path`, which is just a value object. That said, if you want to clean it up later, all the power to you.

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -764,10 +758,11 @@ private ActionResultMetadata parseActionResultMetadata(
 
     ImmutableMap.Builder<Path, FileMetadata> files = ImmutableMap.builder();
     for (OutputFile outputFile : actionResult.getOutputFilesList()) {
+      Path path = remotePathResolver.resolveLocalPath(outputFile.getPath());
```
lberki, line 761: opt: call this `localPath` maybe? (also above). I don't think it's much more descriptive, but given that `RemotePathResolver` seems to make the local/remote path distinction important, it looks like an improvement.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -699,6 +698,16 @@ private SpawnResult handleError(
         .build();
   }
 
+  static List<Property> getExtraPlatformProperties(Spawn spawn, Path execRoot) {
+    if (Spawns.shouldDifferentiateWorkspaceCache(spawn)) {
+      return ImmutableList
+          .of(Property.newBuilder().setName("bazel-workspace").setValue(execRoot.getBaseName())
```
lberki, line 704: nit: make `"bazel-workspace"` also allude to the fact that it's only for splitting cache keys

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+   * Resolves the output path relative to input root for the given {@link Path}.
+   */
+  String resolveOutputPath(Path path);
+
+  /**
+   * Resolves the output path relative to input root for the given {@link PathFragment}.
+   *
+   * @param execPath a path fragment relative to {@code execRoot}.
+   */
+  String resolveOutputPath(PathFragment execPath);
```
lberki, line 56: This overload seems to be unused

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+   * @return a {@link SortedMap} which maps from input paths for remote action to {@link
+   * ActionInput}.
+   */
+  SortedMap<PathFragment, ActionInput> getInputMapping(SpawnExecutionContext context)
+      throws IOException;
+
+  /**
+   * Resolves the output path relative to input root for the given {@link Path}.
+   */
+  String resolveOutputPath(Path path);
```
lberki, line 49: I was about to say that it's probably better to have only one overload for `resolveOutputPath()` and the one taking an execpath seems to be obviously the best, but then I looked at `RemoteCache.UploadManifest()`, which is AFAIU the only place where it's needed and realized that removing `Path` from that call would be quite a bit of work.

It seems always to come from `RemoteSpawnRunner.resolveActionInputs()`, so I think it's still a good idea to do because then you plumb the Path representing the execroot into `UploadManifest` only wrapped in `RemotePathResolver`, but it's probably not for this change. WDYT?

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+  default String resolveOutputPath(ActionInput actionInput) {
+    return resolveOutputPath(actionInput.getExecPath());
+  }
+
+  /**
+   * Resolves the local {@link Path} of an output file.
+   *
+   * @param outputPath the return value of {@link #resolveOutputPath(PathFragment)}.
+   */
+  Path resolveLocalPath(String outputPath);
```
lberki, line 70: opt: maybe it's better to call these methods `localPathToOutputPath` or `outputPathToLocalPath` because the verb "resolve" is quite vague.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * A {@link RemotePathResolver} is used to resolve input/output paths for remote execution from
+ * Bazel's internal path, or vice versa.
+ */
+public interface RemotePathResolver {
+
+  /**
+   * @return the {@code workingDirectory} for a remote action.
+   */
+  @Nullable
```
lberki, line 36: Document when this is null

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
```diff
@@ -1476,21 +1481,25 @@ protected Spawn createSpawn(Map<String, String> clientEnv) throws ActionExecutio
...
+          getDotdFile().getExecPathString());
+    }
+
+    if (shouldParseShowIncludes()) {
+      // Hack on Windows. The included headers dumped by cl.exe in stdout contain absolute paths.
+      // When compiling the file from different workspace, the shared cache will cause header
+      // dependency checking to fail. This was initially fixed by a hack (see
+      // https://github.com/bazelbuild/bazel/issues/9172 for more details), but is broken again due
+      // to cl/356735700. We require execution service to ignore caches from other workspace.
+      executionInfo.put(ExecutionRequirements.DIFFERENTIATE_WORKSPACE_CACHE, "");
```
lberki, line 1494: This implementation is *much* nicer than the previous one, kudos :)

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
```diff
@@ -1476,21 +1481,25 @@ protected Spawn createSpawn(Map<String, String> clientEnv) throws ActionExecutio
...
-              .putAll(executionInfo)
-              .put(
-                  ExecutionRequirements.REMOTE_EXECUTION_INLINE_OUTPUTS,
-                  getDotdFile().getExecPathString())
-              .build();
+      executionInfo.put(ExecutionRequirements.REMOTE_EXECUTION_INLINE_OUTPUTS,
+          getDotdFile().getExecPathString());
+    }
+
+    if (shouldParseShowIncludes()) {
```
lberki, line 1488: nit: why not simply `booleaan shouldParseShowIncludes = featureConfiguration.isEnabled(CppRuleClasses.PARSE_SHOWINCLUDES`?

Every call site of `shouldParseShowIncludes()` is called from this single method, so it's a slightly better because then you have one less method and a teeny tiny bit less computation.

src/test/java/com/google/devtools/build/lib/remote/RemoteSpawnRunnerTest.java
```diff
@@ -145,7 +146,7 @@
 
   // The action key of the Spawn returned by newSimpleSpawn().
   private final String simpleActionId =
-      "b9a727771337fd8ce54821f4805e2d451c4739e92fec6f8ecdb18ff9d1983b27";
+      "eb45b20cc979d504f96b9efc9a08c48103c6f017afa09c0df5c70a5f92a98ea8";
```
lberki, line 149: Mind adding test cases for the new functionality you added?

src/main/java/com/google/devtools/build/lib/analysis/platform/PlatformUtils.java
coeuvre: Done. Changed to use value of `DIFFERENTIATE_WORKSPACE_CACHE` instead of `execRoot.getBaseName()`.

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
coeuvre: Changed to `"bazel-differentiate-workspace-cache"`.

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
coeuvre: You are right. I wanted to only have `resolveOutputPath(PathFragment execPath)` but then realized that `upload` methods take `Path` for output files and it need more work to replace them with `execPath`.

Let's refactor this along with changing `RemotePathResolve` to be a member of `RemoteCache` together in a separate PR!

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
coeuvre: Thanks! :)

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
coeuvre: Since `featureConfiguration` is null when running some unit tests. By using `shouldParseShowIncludes`, we open a chance for tests to mock this method.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
lberki: nit: in that case, consider making the "working directory == input root" case the empty string. Null kinda implies "not available / there isn't any", which is not what you mean here.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
lberki: ...and by doing so, widen the interface of `CppCompileAction`. If that's the motivation to extract this into a separate method, let's not do that and provide a `FeatureConfiguration` to the tests, if need be.

lberki (commented):
(I assume this is not ready for a second round of review, since the CI is red?)

coeuvre:
(I was interrupted when adding more tests) Now it's ready to be reviewed!

src/test/java/com/google/devtools/build/lib/remote/RemoteSpawnRunnerTest.java
coeuvre: Added.

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
coeuvre: Done.

lberki (approved):

alexeagle:
@coeuvre it seems like we're missing the `incompatible-change` tracking bug to flip the default for incompatible_remote_output_paths_relative_to_input_root to true. @meteorcloudy how do those get created?

meteorcloudy:
For incompatible changes, we should still follow https://bazel.build/contribute/breaking-changes

coeuvre:
Created #15131.

PR #13371, opened 2021-04-19
Add config hash to cycle reporting, to aid debugging.

katre:


janakdr (changes requested):
Fix tests?

katre:
Sorry, should have let tests run before I sent this for review.

Unfortunately, the best way to fix the tests involves adding regular expression support to a lot of the existing test primitives. Do you prefer that these changes be in this PR, or split out to a separate PR?

janakdr:
Ok either way

PR #13372, opened 2021-04-19
Fix label_flag and label_setting to not have a dependency on the default

### LINKED ISSUE #11291: Dependency cycle involving label_flag and configuration transition

In rules_go, we have a customizable static analysis system called *nogo*. There's a `nogo` rule that generates a binary that is run alongside the Go compiler when building each `go_library`. Developers use it to define a binary with a customizable set of static analyzers.

I'd like to migrate nogo to the new Starlark Build Configuration system. Ideally, there should be a `label_flag` that points to a `nogo` target to use, and that could be set on the command-line. Currently, each `go_binary` and `go_library` implicitly depends on a default `nogo` target, which is set using some hacks involving repository rules. To break the dependency cycle, `nogo` targets may only depend on targets defined with `go_tool_library`, a bootstrap rule similar to `go_library` except it doesn't depend on `nogo` in order to break the dependency cycle.

Unfortunately, I'm seeing a different kind of dependency cycle. The workspace below is a minimal, standalone example of what I want to do with `nogo`.

* `x_binary` is analogous to `go_binary`.
* `x_checker` is analogous to `nogo`.
* Every `x_binary` implicitly depends on a `label_flag` (with `cfg = "exec"`) that points to an `x_checker` that depends on an `x_binary`.
* The `x_checker` has an incoming transition that sets the flag to a special target with no dependencies to break the cycle.

Any build involving the `label_flag` reports a dependency cycle. It doesn't seem like the current value of the flag is taken into account when the dependency graph is loaded.

This works if `checker_flag` defaults to `//:null_checker` and is set to another value. It just doesn't work when `checker_flag` defaults to a real `x_checker` target.

Run `bazel build //:x` in this workspace:

```
-- WORKSPACE --
# empty file

-- BUILD.bazel --
load(":def.bzl", "null_checker", "x_binary", "x_checker")

x_binary(
    name = "x",
)

label_flag(
    name = "checker_flag",
    build_setting_default = "//:checker",
)

x_checker(
    name = "checker",
    dep = ":checker_bin",
)

x_binary(
    name = "checker_bin",
)

null_checker(
    name = "null_checker",
)

-- def.bzl --
def _x_binary_impl(ctx):
    pass

x_binary = rule(
    implementation = _x_binary_impl,
    attrs = {
        "_checker": attr.label(
            default = "//:checker_flag",
            cfg = "exec",
        ),
    },
)

def _x_checker_transition(settings, attr):
    settings = dict(settings)
    settings["//:checker_flag"] = "//:null_checker"
    return settings

x_checker_transition = transition(
    implementation = _x_checker_transition,
    inputs = ["//:checker_flag"],
    outputs = ["//:checker_flag"],
)

def _x_checker_impl(ctx):
    pass

x_checker = rule(
    implementation = _x_checker_impl,
    attrs = {
        "dep": attr.label(mandatory = True),
        "_whitelist_function_transition": attr.label(
            default = "@bazel_tools//tools/whitelists/function_transition_whitelist",
        ),
    },
    cfg = x_checker_transition,
)

def _null_checker_impl(ctx):
    pass

null_checker = rule(
    implementation = _null_checker_impl,
)
```

Darwin / amd64 (macOS 10.15.4)

release 3.1.0

cc @juliexxia @gregestren @katre

katre:
value.

This prevents an extra analysis, since the dependency should only be on
the value being used.

Fixes #11291.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
@@ -56,6 +56,7 @@ import com.google.devtools.build.lib.packages.RuleClass.Builder.RuleClassType;
 import com.google.devtools.build.lib.packages.RuleClass.Builder.ThirdPartyLicenseExistencePolicy;
 import com.google.devtools.build.lib.packages.RuleFactory.AttributeValues;
 import com.google.devtools.build.lib.packages.Type.ConversionException;
+import com.google.devtools.build.lib.packages.Type.LabelClass;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec.VisibleForSerialization;
 import com.google.devtools.build.lib.util.FileTypeSet;
@@ -941,7 +942,7 @@ public class RuleClass {
             attr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME, type)
                 .nonconfigurable(BUILD_SETTING_DEFAULT_NONCONFIGURABLE)
                 .mandatory();
-        if (BuildType.isLabelType(type)) {
+        if (type.getLabelClass() == LabelClass.DEPENDENCY) {
           defaultAttrBuilder.allowedFileTypes(FileTypeSet.ANY_FILE);
           defaultAttrBuilder.allowedRuleClasses(ANY_RULE);
         }
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptionConverters.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptionConverters.java
@@ -16,6 +16,8 @@ package com.google.devtools.build.lib.analysis.config;
 
 import static com.google.devtools.build.lib.packages.BuildType.LABEL;
 import static com.google.devtools.build.lib.packages.BuildType.LABEL_LIST;
+import static com.google.devtools.build.lib.packages.BuildType.NODEP_LABEL;
+import static com.google.devtools.build.lib.packages.BuildType.NODEP_LABEL_LIST;
 import static com.google.devtools.build.lib.packages.Type.BOOLEAN;
 import static com.google.devtools.build.lib.packages.Type.INTEGER;
 import static com.google.devtools.build.lib.packages.Type.STRING;
@@ -57,6 +59,7 @@ public class CoreOptionConverters {
           .put(STRING_LIST, new CommaSeparatedOptionListConverter())
           .put(LABEL, new LabelConverter())
           .put(LABEL_LIST, new LabelListConverter())
+          .put(NODEP_LABEL, new LabelConverter())
           .build();
 
   /** A converter from strings to Starlark int values. */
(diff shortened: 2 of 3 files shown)


gregestren (approved):

src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
```diff
@@ -941,7 +942,7 @@ public RuleClass build(String name, String key) {
             attr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME, type)
                 .nonconfigurable(BUILD_SETTING_DEFAULT_NONCONFIGURABLE)
                 .mandatory();
-        if (BuildType.isLabelType(type)) {
+        if (type.getLabelClass() == LabelClass.DEPENDENCY) {
```
lberki, line 945: When is this true? It doesn't seem to be true at any call site and AFAIU in order to fix the associated bug, the `build_setting_default` attribute must not create any dependencies, and `allowedFilesTypes()` / `allowedRuleClasses()` only make sense if a dependency is created. Assert that it's not true instead? (either here or in `BuildSetting`)

lberki (approved):
Feel free to submit; I don't want to hold up this change for another day, regardless of the answer to the single comment I had.

Maybe also add a test case?

src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
katre: I changed this from `isLabelType()` to `== LabelClass.DEPENDENCY` specifically because of the change to `NODEP_LABEL`: with this change, the calls to `allowedFileTypes()` and `allowedRuleClasses()` were failing because they only accept real labels, not nodep labels.

Sorry I didn't wait for your review, this is currently internal at cl/369288702.

lberki: Let me rephrase this:  when I look at the code, `RuleClass.getBuildSetting()` never returns a `Type` where `getLabelClass()` would return `DEPENDENCY`. But then why is this branch here?

In addition, if there was any case where that did return `DEPENDENCY`, that would trigger the bug you are fixing here, so why not assert the opposite?

katre: Ah, now I see your point. This entire check should be removed, since it's not possible to trigger it.

Should I add a check in `BuildSetting.create()` to ensure that only `NODEP_LABEL` and `NODEP_LABEL_CLASS` are used? I will make changes in the internal version and test this.

lberki: I think a better approach is to assert `buildSetting.getType().getLabelClass() != LabelClass.DEPENDENCY` beacuse that says exactly what is needed: that the type shouldn't be one that creates a dependeny.

katre: But `allowedFileTypes` can only validly be called when the type is `LabelClass.DEPENDENCY` (see [the definitions](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/Attribute.java;drc=d743e575c29ed954b1a2edbe5a6da66ccb29af8e;l=866)). Since it is now not possible to have an attribute with that type, there's no valid way to call this.

PR #13374, opened 2021-04-19
Update config.md

steve-the-bayesian:
The first line of the file was confusing because it referred to Starlark before defining what Starlark was.  I moved the definition to the first line.


steve-the-bayesian:
@googlebot I signed it!

gregestren (approved):
Looks good to me, thanks!

PR #13375, opened 2021-04-19
Fix regex event matching

katre:
This is the prerequisite to #13371.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/buildtool/CustomRealFilesystemBuildIntegrationTest.java
+++ b/src/test/java/com/google/devtools/build/lib/buildtool/CustomRealFilesystemBuildIntegrationTest.java
@@ -50,6 +50,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/devtools/build/lib/buildtool/CustomRealFilesystemBuildIntegrationTest.java
```diff
@@ -353,11 +354,10 @@ public void ioExceptionReadingBuildFileForDiscoveredInput() throws Exception {
     buildTarget("//hello:hello");
     BuildFailedException e =
         assertThrows(BuildFailedException.class, () -> buildTarget("//hello:hello"));
-    MoreAsserts.assertContainsEventRegex(
-        events.collector(),
+    MoreAsserts.assertContainsEvent(events.collector(), Pattern.compile(
```
janakdr, line 357: pretty sure our internal linter is going to complain about compiling a constant string pattern inline. Not worth refactoring, just FYI.

janakdr (approved):

PR #13377, opened 2021-04-19
Java coverage: fix handling of external files

### LINKED ISSUE #13376: Coverage for java_test doesn't handle external jars

In a Java build using jars from external repositories, running coverage may drop some coverage results due to incorrect path computations.

Bazel writes a file containing the runtime classpath using `LazyWritePathsFileAction`. The `collect_coverage.sh` script then processes that file to generate another file containing absolute paths by prefixing each of the paths with the runfiles directory and the workspace name. It then runs the singlejar tool on this list of jar files to merge them into a single file.

However, for jar files *outside* the main repository, this results in incorrect paths such as `.../MyTest.runfiles/workspace/guava.jar`, where the correct path should be `.../MyTest.runfiles/guava/guava.jar`.

Seems to be OS-independent.

This happens with 4.0.0 and also HEAD.

I have a patch that fixes the paths for coverage w/ `java_test`. However, the `LazyWritePathsFileAction` is used in three more places, and I suspect all of them are incorrect.

ulfjack:
When java_test rules are run with coverage, then Bazel writes a file
containing the runtime classpath using root-relative paths.

The collect_coverage.sh script then processes that file to generate
another file containing absolute paths by prefixing each of the paths
with the runfiles directory and the workspace name. It then runs the
singlejar tool on this list of jar files to merge them into a single
file.

This happens to work for all jar files in the main repository because
the runfiles path is the runfiles directory plus the workspace name
plus the root-relative path.

However, it is broken if some of the jar files come from an external
repository. It appears that singlejar errors out on the first such jar
file, generating a partial output jar file without a central directory.
The error is swallowed by the `collect_coverage.sh` script. Presence of
coverage results may thus depend on the order of classpath entries.

In order to fix this, we change the runtime classpath file to contain
runfiles-relative paths, and the coverage script to prefix them with
the runfiles directory.

Note that we reuse SourceManifestAction here, i.e., the identical code
that is also responsible for generating the runfiles directory in the
first place. This is the only reliable way to get the correct paths
into the test action.

There are more places that use `LazyWritePathsFileAction` and I suspect
that all of them are broken:

- persistent test runner support in BazelJavaSemantics (AFAIK, this
  doesn't work in Bazel anyway)
- coverage w/ metadata jars in BazelJavaSemantics (not used by
  JavaBinary)
- coverage source list file ('-paths-for-coverage.txt'); this probably
  results in non-functioning coverage for *source files* in external
  repositories

Fixes #13376.

Change-Id: Ie9bcc92344f06e190efcb192a3b6ef9905aea352

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/java/JavaBinary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/JavaBinary.java
@@ -33,10 +33,11 @@ import com.google.devtools.build.lib.analysis.RuleContext;
 import com.google.devtools.build.lib.analysis.Runfiles;
 import com.google.devtools.build.lib.analysis.RunfilesProvider;
 import com.google.devtools.build.lib.analysis.RunfilesSupport;
+import com.google.devtools.build.lib.analysis.SourceManifestAction;
+import com.google.devtools.build.lib.analysis.SourceManifestAction.ManifestType;
 import com.google.devtools.build.lib.analysis.TransitiveInfoCollection;
 import com.google.devtools.build.lib.analysis.actions.CustomCommandLine;
 import com.google.devtools.build.lib.analysis.actions.FileWriteAction;
-import com.google.devtools.build.lib.analysis.actions.LazyWritePathsFileAction;
 import com.google.devtools.build.lib.analysis.actions.SpawnAction;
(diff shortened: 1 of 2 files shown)


lberki:
Can I haz test case?

PR #13382, opened 2021-04-20
docs: don't indicate that genrule is for minifying JavaScript

alexeagle:
We have purpose-built rules for that in rules_nodejs, and also, using genrule doesn't work well for nodejs tools since we have
a dedicated rule npm_package_bin that understands how to run these as actions.


lastthyme (approved):

lastthyme (approved):

aiuto:
This looks reasonable to me. I am OOO for much of the next week.
@communikit, can you merge?

PR #13386, opened 2021-04-21
Fix optimized C++ warnings

keith:
These are only used in asserts


keith:
@oquenchil can you review?

oquenchil (approved):
LGTM

PR #13400, opened 2021-04-24
Use boolean expressions to define filters

emusand:
Python boolean expression syntax can now be used to express filters in
--test_tag_filters and --build_tag_filters.

Example:
bazel test ... --test_tag_filters="tag1 and (tag2 or not tag3)"

### DIFF
--- a/site/docs/user-manual.html
+++ b/site/docs/user-manual.html
@@ -1194,13 +1194,37 @@ memory-saving mode</a>.
   setting is high enough.
 </p>
 
-<h4 id='flag--build_tag_filters'><code class='flag'>--build_tag_filters=<var>tag[,tag]*</var></code></h4>
+<h4 id='flag--build_tag_filters'><code class='flag'>--build_tag_filters <var>tag[,tag]*</var> | <var>boolean expression</var></code></h4>
 <p>
-  If specified, Bazel will build only targets that have at least one required tag
-  (if any of them are specified) and does not have any excluded tags. Build tag
-  filter is specified as comma delimited list of tag keywords, optionally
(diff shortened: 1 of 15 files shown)


emusand:
This is a new PR for the same change in https://github.com/bazelbuild/bazel/pull/11900. The author of the original PR, @ehkloaj, was a summer intern at our company, and the Google CLA check failed as his account recently was removed.

See the original PR for more information about this change.

meisterT (commented):
The new file (TagFilter.java) needs the same license header (with updated copyright year) that we have in other files.

emusand:
Thank you for finding the mistake! I have added the license header to TagFilter.java now.

tetromino:
Interesting decision to not quote the tag names! I assumed at first that it would be certain to break lots of things, but in practice I've found only a handful of instances of [Starlark keywords](https://github.com/bazelbuild/starlark/blob/master/spec.md#lexical-elements) being as tag names in Google's monorepo, and I think we can fix those.

So it should be ok.

meisterT:
Hi @emusand, @Kal-Elx,

We are sorry, but after careful consideration we have decided to not merge this PR.

We appreciate the problem you're trying to solve. The challenge we see is keeping Bazel's wider API clear and consistent. The existing --*_tag syntax already has non-trivial uses and documentation references (outside of Bazel). This means extra challenges for migration sequencing, forking the documentation, and sustained migration ownership. If we don't tackle those issues thoroughly, it will lead to new API complexity, even if your specific syntax is clear.

So this isn't about the quality of your technical contribution, but larger project management issues.

As a workaround, you would need to combine bazel query, some postprocessing and bazel test.

Apologies for the late decision which we should have made much sooner.

PR #13403, opened 2021-04-25
stardoc: auto-dedent rule doc and attribute docs

### LINKED ISSUE #13402: Typical formatting of rule docstring is unusable for Markdown output
Commonly a rule with multi-line documentation can have a second paragraph, which developers commonly indent to be readable within the dictionary where it's declared:

```
example_rule = rule(
    implementation = _rule_impl,
    doc = """Small example of rule using a markdown template.

    This paragraph should not have extra intent, as that would make it a code block.
    """,
    attrs = {
        "first": attr.string(doc = "This is the first attribute"),
        "second": attr.string(default = "2"),
    },
)
```

However when producing Markdown-format output, the extra indent on the lines following the first causes that text to be a code block.

> Small example of rule using a markdown template.
> ```
>     This paragraph should not have extra intent, as that would make it a code block.
> ```

Bazel has [code to account for this](https://github.com/bazelbuild/bazel/blob/4.0.0/src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java#L202) by inferring the baseline indentation, however this logic doesn't apply to the rule documentation string.

As a real-world example, see the codeblock that appears here:
https://bazelbuild.github.io/rules_kotlin/kotlin#kt_jvm_binary
and the (ugly) fix required currently:
https://github.com/bazelbuild/rules_kotlin/pull/533/files#diff-4494836edf6b3f0b67056cb72ef24752d21a8c08eb8f2ca4790780179dd76d81R315

alexeagle:
Makes them suitable for markdown output.

Fixes #13402

### DIFF
--- a/src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
+++ b/src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
@@ -58,7 +58,7 @@ public final class DocstringUtils {
    * """
    * }</pre>
    *
-   * @param docstring a docstring of the format described above
+   * @param doc a docstring of the format described above
    * @param parseErrors a list to which parsing error messages are written
    * @return the parsed docstring information
    */
@@ -67,6 +67,63 @@ public final class DocstringUtils {
     DocstringInfo result = parser.parse();
     parseErrors.addAll(parser.errors);
     return result;
+  }
+
+  /**
+   * Removes leading whitespace from docstring lines after the first.
+   *
+   * Allows docstrings to be indented to the context where they appear, without including that indent in the output.
+   * Some output formats have semantic meaning for leading whitespace (such as Markdown).
+   *
+   * @param doc A non-function docstring, without Args or Returns sections.
+   * 
+   */
+  public static String dedentDocstring(String doc) {
+    // Two iterations: first buffer lines and find the minimal indent, then trim to the min
+    List<String> buffer = new ArrayList<>();
+    int indentation = Integer.MAX_VALUE;
+    int lineOffset = 0;
+    int endOfLineOffset = 0;
+
+    do {
+      endOfLineOffset = doc.indexOf("\n", lineOffset);
+      String line = endOfLineOffset < 0 ? doc.substring(lineOffset) : doc.substring(lineOffset, endOfLineOffset + 1);
+      boolean allWhitespace = line.trim().isEmpty();
+      boolean isFirstLine = lineOffset == 0;
+      if (!allWhitespace) {
+        int lineIndent = DocstringParser.getIndentation(line);
+        if (isFirstLine && lineIndent == 0) {
+          // Doesn't count. First line is often not indented, e.g.
+          // doc = """First line
+          //       second line"""
+          // should still result in both lines with no leading whitespace
+        } else {
+          indentation = Math.min(indentation, lineIndent);
+        }
+      }
+      buffer.add(line);
+      // next line
+      lineOffset = endOfLineOffset + 1;
+    } while (endOfLineOffset >= 0);
+
+    if (indentation == Integer.MAX_VALUE) {
+      return doc;
+    }
+
+    boolean firstLine = true;
+    StringBuilder description = new StringBuilder();
+    for (String bufLine : buffer) {
+      if (firstLine) {
+        description.append(bufLine);
+        firstLine = false;
+      } else if (bufLine.trim().isEmpty()) {
+        description.append("\n");
+      } else {
+        String trimmedLine = bufLine.substring(indentation);
+        description.append(trimmedLine);
+      }
+    }
+    return description.toString();
   }
 
   /** Encapsulates information about a Starlark function docstring. */
--- a/src/main/java/com/google/devtools/build/skydoc/fakebuildapi/BUILD
+++ b/src/main/java/com/google/devtools/build/skydoc/fakebuildapi/BUILD
@@ -37,6 +37,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/skydoc/rendering/proto:stardoc_output_java_proto",
         "//src/main/java/net/starlark/java/eval",
         "//src/main/java/net/starlark/java/syntax",
+        "//src/tools/starlark/java/com/google/devtools/starlark/common",
         "//third_party:guava",
         "//third_party:jsr305",
     ],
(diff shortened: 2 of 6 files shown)


alexeagle:
FYI @c-parsons could use someone's help to review/merge :)

tetromino (changes requested):
Thank you for the PR!

Main question is why we need a new `dedentDocstring()` method - can we use the existing `DocstringParser` for dedenting non-function docstrings?

Secondary comment - whatever we do to fix rule docstrings, we ought to also do for providers and aspects; so I think we'd want to dedent in `FakeStarlarkRuleFunctionsApi.asProviderFieldInfo()`, `FakeStarlarkRuleFunctionsApi.forProviderInfo()`, `FakeStarlarkRuleFunctionsApi.aspect()`.

alexeagle:
Yeah I considered changing the existing DocstringParser class, but of course it has a bunch of fields which are specific to docs with structured sections like `Args:` and therefore it's probably a bigger change than this. More work to write and more work to review - and given how PRs often don't find anyone with time to review/merge them I didn't want to do all that work up front.

It's also somewhat risky to make deep changes to existing code, but the test coverage here is pretty good so I think the main issue is just author/reviewer time.

aiuto:
Friendly ping to all.

alexeagle:
My question for @tetromino is whether you have time to review a big refactoring here and if that's what you want, before I spend time digging in to how is could be done.

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
```diff
@@ -69,6 +69,63 @@ public static DocstringInfo parseDocstring(String doc, List<DocstringParseError>
     return result;
   }
 
+  /**
+   * Removes leading whitespace from docstring lines after the first.
+   *
+   * Allows docstrings to be indented to the context where they appear, without including that indent in the output.
+   * Some output formats have semantic meaning for leading whitespace (such as Markdown).
```
aiuto, line 76: This is true, but it points out a design problem. It is completely unscalable to allow different formats for the content of the docstring. Producing a unified doc set and/or pre-rendering the doc in a PR review becomes a guessing game of pick the right render engine. We should insist on markdown, or html, or anything, as long as the rendering rules for it are well defined.

alexeagle: That ship sailed a long time ago I think. There are several old stardoc issues about it.

For this context we can just say "let's not introduce leading white space to accept as many inputs as possible"

aiuto: Actually. The comment is probably fine.  I was just grumbling. I'm totally in favor of this CL.

But to be honest, I am going to propose that for rules to play together well - in terms of mixing them all into a unified doc set - that we pick a single markup syntax and start converging toward that.  But we don't have to make that choice at the stardoc level. It is something that should come from the overall project and push down.

alexeagle:
@tetromino I think the ball is still in your court, I don't want to invest in a refactoring here that doesn't get merged

aiuto:
@tetromino Friendly ping

sgowroji:
Hello @tetromino, Any update on this PR? Please. Thanks!

alexeagle:
🤷🏻‍♂️

aiuto:
@alexeagle   Can you rebase to fix the merge conflict?
@brandjon  Since Alex is out for a few months, can you take a look?

brandjon:
@tetromino will be back in office next month. We can discuss broader questions of docstring formatting then. In the meantime I'll leave this PR open.

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
```diff
@@ -69,6 +69,63 @@ public static DocstringInfo parseDocstring(String doc, List<DocstringParseError>
...
+    }
+
+    boolean firstLine = true;
+    StringBuilder description = new StringBuilder();
+    for (String bufLine : buffer) {
+      if (firstLine) {
+        description.append(bufLine);
+        firstLine = false;
+      } else if (bufLine.trim().isEmpty()) {
+        description.append("\n");
```
tetromino, line 120: Returning `doc` unchanged on line 110 implies that all-whitespace lines become empty lines (even if they contain whitespace beyond `indentation`) if any dedentation is done, but remain unchanged when not dedenting - which is unintuitive behavior.

Would it make more sense to instead dedent all-whitespace lines by up to the same number of spaces as any other line?

src/tools/starlark/java/com/google/devtools/starlark/common/DocstringUtils.java
```diff
@@ -69,6 +69,63 @@ public static DocstringInfo parseDocstring(String doc, List<DocstringParseError>
...
+    // Two iterations: first buffer lines and find the minimal indent, then trim to the min
+    List<String> buffer = new ArrayList<>();
+    int indentation = Integer.MAX_VALUE;
+    int lineOffset = 0;
+    int endOfLineOffset = 0;
+
+    do {
+      endOfLineOffset = doc.indexOf("\n", lineOffset);
+      String line = endOfLineOffset < 0 ? doc.substring(lineOffset) : doc.substring(lineOffset, endOfLineOffset + 1);
+      boolean allWhitespace = line.trim().isEmpty();
```
tetromino, line 91: `line.trim()` removes tabs (as well as all other characters below 0x20, e.g. windows carriage returns), not just spaces. `DocstringParser.getIndentation()` counts only spaces.

tetromino (changes requested):
I apologize for dropping the ball here and not reviewing this sooner. I somehow operated under the impression that this code was only needed for the legacy stardoc extractor - but obviously both extractors need something like this logic.

So overall I agree with the proposed change, but there are a number of corner case problems (see my comments). I  suggest limiting this PR to just adding DocstringUtils.dedentDocstring() and a set of java tests for it to make sure the corner cases are addressed.

As a side effect, limiting the PR in this way would automatically eliminate merge conflicts :)

Then as a followup, we can apply dedentation to everywhere it needs to be applied to (both in the old and new doc extractors, for all documentable entities).

alexeagle:
Thanks for the review! I'm not sure when I'll get time to return to this, but I'll try to get someone to carry it to the finish.

tetromino:
I have an in-progress change (depends on other changes which have not been submitted yet) which fixes trimming/dedenting *everywhere* (rules, aspects, providers, repo rules, module extensions, and attributes of any of above).

alexeagle:
That's awesome, I've been sad this one is languishing. Thanks for the update!

tetromino:
For reference, it's part of https://bazel-review.googlesource.com/c/bazel/+/225435 (unfortunately, Gerrit squashed other in-progress changes into it).

PR #13405, opened 2021-04-26
Allow excluding --config flag sets, instead of just accumulating them.

### LINKED ISSUE #13284: Allow excluding --config flag sets, instead of just accumulating them

`--config` selects flag sets from an rc file. It would be useful to have mutually-exclusive flag sets, e.g. by specifying a `--noconfig` flag (or `--config=-value`, which is probably not so good since `-` is currently allowed, and using this notation might break existing rc files).

We have flag sets for remote build execution defined in our `.bazelrc` file. As we have one main platform, we turned one of these configs on by default by adding `build     --config=...`. However,  some of the flags (e.g. `--extra_execution_platforms`) are accumulative, so this default cannot really be overridden or turned off without commenting out that line when we want to build for another platform. This sort of behaviour is also less safe for non-accumulating flags.

It would be nice to have something like a `--noconfig` flag, and then each RBE configuration could explicitly exclude the others.

Linux

4.0.0

I couldn't find other related issues here or on bazel-discuss. Searching for `--config` on StackOverflow and asking on Slack didn't yield an answer either.

There are some nuances around this feature: `--noconfig` should probably not be a regular accumulative flag, as we want later configs to prevail; it might not do to have a simple comma-separated, non-repeatable flag, though, as there may be several independent groups of flag sets that need to support just one configuration. So `--config` and `--noconfig` may have to be processed for each value separately, in the order they're given. This seems non-trivial to do, but a solution (pre-parsing all the flags, then re-parsing with the right exclusions in place?) might also handle the complication that arises from the two-pass processing `ConfigExpander.expandConfigOptions` does: it first expands all the flags it can and parses them to determine whether `--enable_platform_specific_config` is on, then proceeds to expand the platform-specific config, which may have `--noconfig` options that should affect options from the first pass.

Anyway, I've been toying with implementing this – if the feature sounds good, I'll be happy to send a PR and iterate on the details.

quval:
Some Bazel flags accumulate values and cannot be completely overridden, yet it may be useful to set defaults for them in a `.bazelrc` file. To override these values, use `--noconfig`. `--noconfig=foo` prevents the expansion of the `foo` config, allowing certain config groups to exclude others.

`--noconfig` options are processed before configs are actually applied, so setting `--noconfig=foo` as a default (even through some config group that is applied by default) would prevent `--config=foo` from ever being applied. In particular, whereas alternative configs might exclude the default, the default shouldn't exclude the alternatives. For example,

```
build:rbe-default-platform    # ...
build:rbe-other-platform      # ...
build:rbe-other-platform      --noconfig=rbe-default-platform
build                         --config=rbe-default-platform
```

(Without `--noconfig`, this wouldn't work because of `--extra_execution_platforms`.)

To keep things simpler, we don't support `--noconfig` settings from platform-dependent configuration; it should probably be easy enough to work around any need for `--noconfig` there.

For clarity, it is best to keep any usage of `--noconfig` close to toplevel configs.

Closes #13284.

### DIFF
--- a/site/docs/guide.md
+++ b/site/docs/guide.md
@@ -927,6 +927,22 @@ This syntax does not extend to the use of `startup` to set
 [startup options](#option-defaults), e.g. setting
 `startup:config-name --some_startup_option` in the .bazelrc will be ignored.
 
+#### `--noconfig`
+
+Some Bazel flags accumulate values and cannot be completely overridden, yet it
+may be useful to set defaults for them in a `.bazelrc` file. To override these
+values, use `--noconfig`. `--noconfig=foo` prevents the expansion of the `foo`
+config, allowing certain config groups to exclude others.
+
+`--noconfig` options are processed before configs are actually applied, so
+setting `--noconfig=foo` as a default (even through some config group that is
+applied by default) would prevent `--config=foo` from ever being applied. In
+particular, whereas alternative configs might exclude the default, the default
+shouldn't exclude the alternatives.
+
+For clarity, it is best to keep any usage of `--noconfig` close to toplevel
+configs.
+
 #### Example
 
 Here's an example `~/.bazelrc` file:
--- a/src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
@@ -83,6 +83,19 @@ public class CommonCommandOptions extends OptionsBase {
   public List<String> configs;
 
   @Option(
+      name = "noconfig",
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.UNCATEGORIZED,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      allowMultiple = true,
+      help =
+          "Config sections to avoid selecting from the rc files. Values are collected from config "
+              + "sections before they are actually applied, so noconfig takes precedence. In "
+              + "particular, setting --noconfig=<config> as a default prevents --config=<config> "
+              + "from ever being applied, so use this carefully.")
+  public List<String> noconfigs;
+
+  @Option(
       name = "logging",
       defaultValue = "3", // Level.INFO
       documentationCategory = OptionDocumentationCategory.LOGGING,
--- a/src/main/java/com/google/devtools/build/lib/runtime/ConfigExpander.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/ConfigExpander.java
@@ -29,6 +29,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
+import javax.annotation.Nullable;
 
 /** Encapsulates logic for performing --config option expansion. */
 final class ConfigExpander {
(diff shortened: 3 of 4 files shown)


quval:
As I wrote on the issue itself, I think there are some nuances around this feature. I'm more than happy to iterate over the solution and design. Thanks!

gregestren:
Question (I'm not necessarily the expert here, so this isn't a review question):

Can you get what you need by changing `--bazelrc`? That's not a perfect approach either. But my raw intuition is the more we can do to avoid new API complexity (in this case the `bazelrc` API), the better. 

But yes, this is an interesting dilemma.

quval:
@gregestren: I wholeheartedly agree with the sentiment, and to be perfectly honest, no – I hadn't thought of that. It doesn't solve the problem unless I also specify `--noworkspace_rc`. Either flag can't be given through an rc file, obviously, and I suppose we can live with switching execution (and target) platforms with two rc flags instead of just one `--config` (as well as breaking up our `.bazelrc`)... but something like this would be nicer, if we can get to a nice solution.

I was thinking of maybe limiting `--noconfig` to the command line and the first level of configs from the rc file. Do you think it would be better this way?

quval:
Thanks for the detailed response – I really appreciate it.

I think I get the complexity at stake here. I sent this PR mostly as a way to get the discussion going. I'm ambivalent here myself. Bazel flag parsing is so delicate that I can't think of any clean solution, and it's clearly necessary to be very cautious not to complicate matters further. I was (still am) ready for both the feature request and my suggested implementation to be rejected.

I will say that I felt it was a bit of a universal problem that there's no way around how `--config` interacts with flags whose values cannot be overridden. I imagine this might happen in other cases as well, but the trigger in my case was setting up remote execution as a default while supporting several target platforms with separate build environments, and the solutions I could come up with so far are all pretty awkward (avoid setting a default and have developers add `--config=rbe` all the time; go back and edit an rc file whenever switching target platform; add both `--noworkspacerc` and an appropriate `--bazelrc` to switch target platform; have separate remote executors or try leveraging remote_instance_name somehow to choose the right platform – this last one is less trivial than it seems – simply because the relevant flags can be overridden). The simplest, most general solution seemed to be to just disable a certain config group.

Unfortunately, the implementation didn't turn out to be so straightforward. It might become a bit simpler with the restriction on `--noconfig` depth (to toplevel configs in rc files or just to the actual command line), but the nuances are hard to get right. I can accept that it might not be worth the time and effort to resolve this right now.

lberki:
FWIW, "Bazel's flag parsing is so delicate" is a strong argument *against* making it more complicated...

On a more serious note, I think this should be a utility vs. cost (in maintenance cost, UI complexity, etc.) decision. As you say, neither the semantics nor the implementation is trivial, so the utility must be correspondingly high.

The most obvious alternative is to use a wrapper script that adds `--config=<default>` to the command line if that flag is not on the command line or some variant thereof. The main issue I can think of is that then that script must parse Bazel's command line far enough to at least understand `--config`, but this can be mitigated by using some other syntax (for example, requiring `--config` to be `argv[1]` and erroring out if it's not)

If I wanted to solve this within Bazel, I would probably consider adding a "this is the default configuration if not specified otherwise" stanza to the bazelrc syntax because that looks semantically simpler because then we don't need negation. But I hope the wrapper script approach works well enough.

quval:
My hope was that this feature was independent enough of flag parsing that it wouldn't make it more complicated. But between all of our messages I hope we've found enough workarounds that one of them might work well enough. Shall I close the PR then or will you?

lberki:
I'll close it then. Let me know if you can't achieve what you want to do without this!

Issue #13406, opened 2021-04-26
`@local_jdk` references fail with Ubuntu 20.10 OpenJDK


fmeum:
### Description of the problem / feature request:

As of https://github.com/bazelbuild/bazel/commit/3733065ddd418357e0e74091789ace5884b01fef, references to `@local_jdk` lead to an error when running with the OpenJDK installed by the `openjdk-15-jre-headless` package of Ubuntu 20.10.

The stack trace is as follows:
```
INFO: Repository local_jdk instantiated at:
  /DEFAULT.WORKSPACE.SUFFIX:27:6: in <toplevel>
  /home/fabian/.cache/bazel/_bazel_fabian/018e1cccb754ba18b7604abc85e7cd0d/external/bazel_tools/tools/build_defs/repo/utils.bzl:201:18: in maybe
  /home/fabian/.cache/bazel/_bazel_fabian/018e1cccb754ba18b7604abc85e7cd0d/external/bazel_tools/tools/jdk/local_java_repository.bzl:226:32: in local_java_repository
Repository rule _local_java_repository_rule defined at:
  /home/fabian/.cache/bazel/_bazel_fabian/018e1cccb754ba18b7604abc85e7cd0d/external/bazel_tools/tools/jdk/local_java_repository.bzl:197:46: in <toplevel>
ERROR: An error occurred during the fetch of repository 'local_jdk':
   Traceback (most recent call last):
File "/home/fabian/.cache/bazel/_bazel_fabian/018e1cccb754ba18b7604abc85e7cd0d/external/bazel_tools/tools/jdk/local_java_repository.bzl", line 147, column 105, in _local_java_repository_impl
version = repository_ctx.attr.version if repository_ctx.attr.version != "" else _detect_java_version(repository_ctx, java_bin)
File "/home/fabian/.cache/bazel/_bazel_fabian/018e1cccb754ba18b7604abc85e7cd0d/external/bazel_tools/tools/jdk/local_java_repository.bzl", line 32, column 26, in _detect_java_version
(major, minor, rest) = version_value.split(".", 2)
Error: too few values to unpack (got 1, want 3)
```

With that package installed, the command `java -XshowSettings:properties`, which is invoked in https://github.com/bazelbuild/bazel/commit/3733065ddd418357e0e74091789ace5884b01fef#diff-8d3287c1b6c0d3bbfc4e6a4b8a6f6966a357c29464f58aedb91309dbf7252f24R18, returns a line of the form:
```
[...]

PR #13407, opened 2021-04-27
Handle java versions with a single component

keith:
Fixes https://github.com/bazelbuild/bazel/issues/13406


keith:
@comius can you review?

comius (changes requested):
Can you add a test to src/test/shell/bazel/bazel_with_jdk_test.sh?

keith:
Updated!

comius (approved):

Issue #13379, opened 2021-04-20
Cannot run Bazel without /usr/bin/python, even if /usr/bin/python3 exists


ted-xie:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

Bazel relies on /usr/env/bin python, which no longer exists in the newest update to my operating system.

### Feature requests: what underlying problem are you trying to solve with this feature?

N/A.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Remove /usr/bin/python symlink and run literally any Bazel workload.

### What operating system are you running Bazel on?

Proprietary OS vaguely similar to Ubuntu 20.04.

### What's the output of `bazel info release`?

```
INFO: Invocation ID: f21f4240-c757-4848-ac49-2acef7ed1685
release 4.0.0
```

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A, using pre-built binary.

###  Have you found anything relevant by searching the web?

Looks like https://github.com/bazelbuild/bazel/issues/8665 was never *truly* fixed Bazel-side. Users should not have to create system-level symlinks to run Bazel.

### Any other information, logs, or outputs that you want to share?

N/A.

keith:
This is a dup of https://github.com/bazelbuild/bazel/issues/8685

ted-xie:
I acknowledge that but there has been little/no traction on that bug. I was hoping to revive the discussion/raise the visibility of this issue by filing a new bug.

philwo:
Hey @ted-xie, I'm sorry that this still isn't fixed. We'll get to it hopefully soon.

lberki:
It looks like this is fixed by https://github.com/bazelbuild/bazel/pull/13414 ?

philwo:
@lberki Indeed! This looks like it might be an elegant fix for this :)

sventiffe:
@philwo #13414 however stalled a while ago. Is anything preventing us from moving on? Would that PR also be included in a future 4.x update?

philwo:
We need to merge that PR, I think it's the best fix we have.

PR #13414, opened 2021-04-29
Define version-specific shebangs for the default python runtimes

rohanjoyce:
Prior to this commit, the python stub generated with the autodetected runtimes will use `#!/usr/bin/env python` as the shebang line. For modern python3 systems, this may not exist. It's more reliable to have each specific python runtime version specify a shebang of that particular python version.

### DIFF
--- a/tools/python/toolchain.bzl
+++ b/tools/python/toolchain.bzl
@@ -174,6 +174,7 @@ def define_autodetecting_toolchain(
         name = "_autodetecting_py2_runtime",
         interpreter = ":py2wrapper.sh",
         python_version = "PY2",
+        stub_shebang = "#!/usr/bin/env python2",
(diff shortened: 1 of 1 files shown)


lberki:
@brandjon , WDYT?

Python knowledge is swapped out of my brain, but it looks like this should fix a number of issues e.g. https://github.com/bazelbuild/bazel/issues/13379 .

/cc @comius

craiglabenz:
Any updates on landing this PR?

sventiffe:
@lberki @comius @brandjon

philwo:
Can we just merge this? It's just four lines of change and all tests pass. Or is there any hidden risk that I'm not aware of? 🤔

lberki:
Not that I know of. I'll send an internal change to @comius .

PR #13417, opened 2021-04-30
Add build_workspace_root to repository_ctx

illicitonion:
There are a number of use-cases for this, and I've seen it come up in
`bazel-discuss` and similar. It is possible to abuse this information,
but this information is currently available by looking up something like:

```python
repository_ctx.path(Label("@//:WORKSPACE")).dirname
```

But this is unreliable as `WORKSPACE.bazel` was added as an alternative
to `WORKSPACE`, and trying to look up a label which doesn't exist is a
fatal error.

We're currently using this to work around the fact that labels can't
exist if the file they point at doesn't already exist.
In repository rules we want users to be able to set an attribute to
point at a lockfile, which may not yet exist, and which our rule may
create as a side-effect of running. The nicest way we've worked out how
to do this is to use a `attr.string`, and use `repository_ctx.execute`
to test for the existence of a file, but that relies on being able to
find the path to the root repository.

It may be interesting to try to introduce some kind of fallible
label/path lookup, but that feels like a much bigger change for the
future.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
@@ -102,6 +102,7 @@ public class StarlarkRepositoryContext extends StarlarkBaseExternalContext {
 
   private final Rule rule;
   private final PathPackageLocator packageLocator;
+  private final Path workspaceRoot;
   private final StructImpl attrObject;
   private final ImmutableSet<PathFragment> ignoredPatterns;
   private final SyscallCache syscallCache;
@@ -122,7 +123,8 @@ public class StarlarkRepositoryContext extends StarlarkBaseExternalContext {
       @Nullable ProcessWrapper processWrapper,
       StarlarkSemantics starlarkSemantics,
       @Nullable RepositoryRemoteExecutor remoteExecutor,
-      SyscallCache syscallCache)
+      SyscallCache syscallCache,
+      Path workspaceRoot)
       throws EvalException {
     super(
         outputDirectory,
@@ -137,6 +139,7 @@ public class StarlarkRepositoryContext extends StarlarkBaseExternalContext {
     this.packageLocator = packageLocator;
     this.ignoredPatterns = ignoredPatterns;
     this.syscallCache = syscallCache;
+    this.workspaceRoot = workspaceRoot;
     WorkspaceAttributeMapper attrs = WorkspaceAttributeMapper.of(rule);
     ImmutableMap.Builder<String, Object> attrBuilder = new ImmutableMap.Builder<>();
     for (String name : attrs.getAttributeNames()) {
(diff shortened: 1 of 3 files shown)


illicitonion:
@philwo I've just pushed a rebase, if it's possible to get a review that's be really handy, it's a small but very handy PR!

UebelAndre:
@illicitonion would you be able to resolve the conflicts here?

@philwo @Wyverald @meteorcloudy could someone take a look at the change? It'd be nice to get this in.

illicitonion:
Rebased :)

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -162,6 +165,14 @@ public String getName() {
     return rule.getName();
   }
 
+  @StarlarkMethod(
+      name = "build_workspace_directory",
```
Wyverald, line 169: this concept is usually referred to as the "workspace root". Could you update the name throughout this PR?

UebelAndre: For what it's worth, `build_workspace_directory` makes sense to me and matches the related environment variable that's set on executable targets:
https://docs.bazel.build/versions/main/user-manual.html#run
> The following extra environment variables are also available to the binary:
>
>    `BUILD_WORKSPACE_DIRECTORY`: the root of the workspace where the build was run.
>   ` BUILD_WORKING_DIRECTORY`: the current working directory where Bazel was run from.

illicitonion: Done - thanks!

Wyverald: sorry to be nitpicky -- is there a reason why "build" is in the name? I kind of thought `ctx.workspace_root` would work well.

Also thanks @UebelAndre for the context -- I wasn't aware of the env variables. It makes more sense that those names include "BUILD", since env variables aren't Bazel-specific; but with the `ctx` object, we're obviously in Bazel.

Wyverald: @meteorcloudy Oh boy, `Label.workspace_root` has a completely different meaning, though... I think we should deprecate the `Label.workspace_*` fields and replace them with `Label.repo_*` or something. The terms "repo" and "workspace" used to be used interchangeably but now we mostly think of the "workspace" as the overall space we work in, housing the main repo and the various external repos.

illicitonion: Removed the build :)

meteorcloudy: > Label.workspace_root has a completely different meaning, though...

Indeed, it's actually the workspace root of the repo of the label, `repo_root` is much better.

brentleyjones:
@bazel-io flag

Issue #11128, opened 2020-04-14
Support workspace label prefixes in Starlark config and platform APIs


gregestren:
This is a catch-all to focus the common theme of:

- https://github.com/bazelbuild/bazel/issues/9065 (fixed)
- https://github.com/bazelbuild/bazel/issues/9177 (fixed)
- https://github.com/bazelbuild/bazel/issues/10152 (fixed)
- https://github.com/bazelbuild/bazel/issues/10499
- https://github.com/bazelbuild/bazel/issues/10927 (@katre is looking into this)
- https://github.com/bazelbuild/bazel/issues/10975

At the narrowest this may require unrelated fixes to `build_setting` and toolchain resolution logic. But the general theme of the problem is clear, however the fixes are made.

We're (Bazel configurability) trying to prioritize this for the current quarter.

gregestren:
Also https://github.com/bazelbuild/bazel/issues/10118.

aherrmann:
https://github.com/bazelbuild/bazel/issues/11734 is a similar issue in aspects.

gregestren:
@juliexxia - can you advise on how much of this issue is still relevant?

nacl:
With 3.7, our experience with this is mostly the same as what's described in #9177, except Bazel no longer crashes, instead, nothing happens.

Interestingly, when I run these two commands (within the [example in #9177](https://github.com/ulysses4ever/bazel-user-settings-example/)), it seems like the prefixed and non-prefixed versions are being treated as separate, but still work :

```console
> bazel build //:my_drink --//:favorite_flavor=True
DEBUG: .../tmp/bazel-user-settings-example/rules.bzl:5:14: Get the opposite of default (True)
...
INFO: Build completed successfully, 1 total action
> bazel build //:my_drink --@bsws//:favorite_flavor=True
INFO: Build options --//:favorite_flavor and --@bsws//:favorite_flavor have changed, discarding analysis cache.
DEBUG: .../tmp/bazel-user-settings-example/rules.bzl:7:14: Get the default (False)
...
INFO: Build completed successfully, 1 total action
```

Referencing build settings from external WORKSPACEs (https://github.com/bazelbuild/bazel/issues/9177#issuecomment-621090904) seems to work well.

stefanbucur:
I am experiencing the same behavior when it comes to using the flags in Bazel transitions.

For a WORKSPACE file defined as `workspace(name = "my_workspace")` and a build setting in my BUILD file as:

```
[...]

PR #13418, opened 2021-05-01
Support `--no@//:bool_flag` on the command line.

beasleyr-vmw:
This applies label canonicalization to starlark flags having a `--no`
prefix.

Relates to #11128.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
@@ -259,7 +259,10 @@ public class StarlarkOptionsParser {
           buildSettingTarget.getAssociatedRule().getRuleClassObject().getBuildSetting();
       if (current.getType().equals(BOOLEAN)) {
         // --boolean_flag or --noboolean_flag
-        unparsedOptions.put(name, new Pair<>(String.valueOf(booleanValue), buildSettingTarget));
+        // Ditto w/r/t canonical form.
+        unparsedOptions.put(
+            buildSettingTarget.getLabel().getCanonicalForm(),
+            new Pair<>(String.valueOf(booleanValue), buildSettingTarget));
       } else {
         if (!booleanValue) {
           // --no(non_boolean_flag)
--- a/src/test/java/com/google/devtools/build/lib/starlark/StarlarkOptionsParsingTest.java
+++ b/src/test/java/com/google/devtools/build/lib/starlark/StarlarkOptionsParsingTest.java
@@ -214,6 +214,18 @@ public class StarlarkOptionsParsingTest extends StarlarkOptionsTestCase {
     assertThat(result.getResidue()).isEmpty();
   }
 
+  // test --no@main_workspace//:bool_flag
+  @Test
+  public void testNoPrefixedBooleanFlag_withWorkspace() throws Exception {
+    writeBasicBoolFlag();
+
+    OptionsParsingResult result = parseStarlarkOptions("--no@//test:my_bool_setting");
+
+    assertThat(result.getStarlarkOptions()).hasSize(1);
+    assertThat(result.getStarlarkOptions().get("//test:my_bool_setting")).isEqualTo(false);
+    assertThat(result.getResidue()).isEmpty();
+  }
+
   // test --noint_flag
   @Test
   public void testNoPrefixedNonBooleanFlag() throws Exception {


gregestren:
Oh, nice! While I like this contribution, I vaguely recall some effort to move away from the `--no<flag>` syntax for *all* flags (both Starlark and native) as a best practice concern. I may be wrong on that, so as next step let me dig into discussion history and see if I can confirm/deny my claim.

beasleyr-vmw:
Aha, good to know. Thanks for looking into it!

FYI, without this we can crash Bazel (even at master/HEAD) by mixing `--no//:flag` and `--no@//:flag` on the command line.

```console
$ bazelisk build --no//:bool_flag --no@//:bool_flag
INFO: Invocation ID: 960e93ef-d3ac-416d-8fb8-b53d90d430ff
WARNING: Usage: bazel build <options> <targets>.
Invoke `bazel help build` for full description of usage and options.
Your request is correct, but requested an empty set of targets. Nothing will be built.
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.IllegalStateException: Duplicate key //:bool_flag (attempted merging values false and false)
        at java.base/java.util.stream.Collectors.duplicateKeyException(Unknown Source)
        at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Unknown Source)
        at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(Unknown Source)
        at com.google.common.collect.CollectSpliterators$1WithCharacteristics.lambda$forEachRemaining$1(CollectSpliterators.java:67)
        at java.base/java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Unknown Source)
        at com.google.common.collect.CollectSpliterators$1WithCharacteristics.forEachRemaining(CollectSpliterators.java:67)
        at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
        at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.base/java.util.stream.ReferencePipeline.collect(Unknown Source)
        at com.google.devtools.build.lib.analysis.config.BuildOptions.labelizeStarlarkOptions(BuildOptions.java:85)
        at com.google.devtools.build.lib.analysis.config.BuildOptions.of(BuildOptions.java:147)
        at com.google.devtools.build.lib.analysis.ConfiguredRuleClassProvider.createBuildOptions(ConfiguredRuleClassProvider.java:726)
        at com.google.devtools.build.lib.runtime.BlazeRuntime.createBuildOptions(BlazeRuntime.java:749)
        at com.google.devtools.build.lib.buildtool.BuildTool.buildTargets(BuildTool.java:136)
        at com.google.devtools.build.lib.buildtool.BuildTool.processRequest(BuildTool.java:401)
        at com.google.devtools.build.lib.runtime.commands.BuildCommand.exec(BuildCommand.java:97)
        at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.execExclusively(BlazeCommandDispatcher.java:579)
        at com.google.devtools.build.lib.runtime.BlazeCommandDispatcher.exec(BlazeCommandDispatcher.java:231)
        at com.google.devtools.build.lib.server.GrpcServerImpl.executeCommand(GrpcServerImpl.java:543)
        at com.google.devtools.build.lib.server.GrpcServerImpl.lambda$run$1(GrpcServerImpl.java:606)
        at io.grpc.Context$1.run(Context.java:579)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
```

gregestren:
Okay, so this simply extend the existing `--no` support to support workspace prefixes too. It's reasonable to treat that consistently, and not crash.

My comments about `--no<flag>` above was indeed wrong. There was a drive to remove `--no_<flag>` (i.e. a flag that starts with `no` and an underscore). But that's unrelated.

There was also discussion about extra subtleties in the Starlark vs. native flag parser. Julie Xia, who best knew this and has since moved onto other things, wrote:

> The starlark options parser operates on the residue of the regular options parser. Take the cases of bazel build //foo --//bar value vs bazel build //foo --//bar=value vs bazel build //foo --//bool/typed/flag //some/other/target/to/build. The regular options parser knows to put args that start with "--//" in residue to pass to the starlark options parser, but has no way to know when it should also grab the next arg. It can't do this without loading the starlark option to figure out what type it is. Currently the regular options parser makes no effort to figure out if it needs to grab the next arg, it just doesn't.
> 
> So even though the starlark options parser can support it, the regular options parser can't formulate the right kind of residue to give to the starlark options parser. So right now, remove support for the space syntax from the starlark options parser. We can always reimplement correctly if there's a need. This simplifies the whole thing nicely and I don't see a major argument for supporting the space syntax right now.

That led to https://github.com/bazelbuild/bazel/commit/28442ae70099812f57bd4db052b5cc63ecb31845.

If I'm understanding this correctly, that means `--//some_starlark_flag --some_other_value` is universally considered two flags. That should be consistent with boolean flags. I don't think anything you're doing adds to these complications. But I just wanted to fill out the fuller considerations on desirable flag syntax.

gregestren (approved):

PR #13427, opened 2021-05-05
Fixes incorrect install names on darwin platforms

csmulhern:
#12304 added support to bazel for setting install names for dynamic
libraries on darwin platforms. This would set LC_ID_DYLIB to
@rpath/{library_name}, so that RPATH would be used to locate these
libraries at runtime. However, the code was using a utility method that
assumed the library name was mangled, which is often not the case. Given
that the output path should already have been determined with the
mangled or unmangled name, we should be able to just use the base name
of the artifact. The test that was added in #12304 has been updated to
actually use dynamic libaries, and passes with the changes made in this
commit.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
@@ -872,8 +872,7 @@ public class CppLinkActionBuilder {
               getLinkType().linkerOrArchiver().equals(LinkerOrArchiver.LINKER),
               configuration.getBinDirectory(repositoryName).getExecPath(),
               output.getExecPathString(),
-              SolibSymlinkAction.getDynamicLibrarySoname(
-                  output.getRootRelativePath(), /* preserveName= */ false),
+              output.getRootRelativePath().getBaseName(),
               linkType.equals(LinkTargetType.DYNAMIC_LIBRARY),
               paramFile != null ? paramFile.getExecPathString() : null,
               thinltoParamFile != null ? thinltoParamFile.getExecPathString() : null,
--- a/src/test/shell/bazel/cpp_darwin_integration_test.sh
+++ b/src/test/shell/bazel/cpp_darwin_integration_test.sh
@@ -124,41 +124,59 @@ EOF
 }
 
 function test_cc_test_with_explicit_install_name() {
-  mkdir -p cpp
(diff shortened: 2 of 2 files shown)


csmulhern:
@oquenchil, have you had the chance to look at this?

oquenchil:
It looks like this change affects more than darwin, but all Bazel presubmits and internal tests pass so it can go in.

I may have to rollback though.

oquenchil (approved):
LGTM

det:
I just ran into this issue (runtime_solib_name is not the mangled name) after upgrading to bazel 5.1 from 4.2. Any reason this did not land?

det:
I believe this change is incorrect and it appears to have broken mac toolchains.

The only use of `runtime_solib_name` in the bazel repo is this:
```
tools/osx/crosstool/cc_toolchain_config.bzl:                            "-Wl,-install_name,@rpath/%{runtime_solib_name}",
```
This line is now broken.

The way it worked previously is bazel will inject rpaths into an executable, ie:
```
path @loader_path/../../_solib_arm64/
```
This directory contains symlinks to all the mangled names of shared libraries that have been built.

This can only work if the install_name for a shared library is set to the mangled name. This PR change changes it so that `runtime_solib_name` is no longer mangled. ie, it went from `libexternal_Sboost_Slibasio.dylib` to `libasio.dylib`.

Also, the test as it existed was also correct as cc_test will link its dependencies using dynamic linking by default. The new test is broken as it does not capture the use case this feature was written to support.

Can we please revert this PR? @csmulhern @oquenchil

det:
I believe this change is incorrect and it appears to have broken mac toolchains.

The only use of `runtime_solib_name` in the bazel repo is this:
```
tools/osx/crosstool/cc_toolchain_config.bzl:                            "-Wl,-install_name,@rpath/%{runtime_solib_name}",
```
This line is now broken.

The way it worked previously is bazel will inject rpaths into an executable, ie:
```
path @loader_path/../../_solib_arm64/
```
This directory contains symlinks to all the mangled names of shared libraries that have been built.

This can only work if the install_name for a shared library is set to the mangled name. This PR change changes it so that `runtime_solib_name` is no longer mangled. ie, it went from `libexternal_Sboost_Slibasio.dylib` to `libasio.dylib`.

Also, the test as it existed was also correct as cc_test will link its dependencies using dynamic linking by default. The new test is broken as it does not capture the use case this feature was written to support.

Can we please revert this PR? @csmulhern @oquenchil

PR #13431, opened 2021-05-06
Simplify macOS arch setup

keith:
Quite a few settings only differ based on the architecture passed. This
hoists the architecture decision out of each individual action. There
are more places that can benefit from this but I only changed 1 here to
reduce the churn. This adds the `apply_simulator_compiler_flags` to more
configurations, but that flag is a no-op for non-simulator
configurations, so we don't need to branch on it in both places.

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -141,6 +141,11 @@ def _impl(ctx):
     else:
         fail("Unreachable")
 
+    if ctx.attr.cpu == "armeabi-v7a":
+        arch = "<architecture>"
+    else:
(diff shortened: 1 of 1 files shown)


tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -141,6 +141,30 @@ def _impl(ctx):
     else:
         fail("Unreachable")
 
+    if (ctx.attr.cpu == "ios_arm64" or
```
trybka, line 144: `.endswith("arm64")` and similar below?

keith:
merged into https://github.com/bazelbuild/bazel/pull/13437 for ease of merging internally

PR #13432, opened 2021-05-06
Deduplicate macOS default_compile_flags_feature

keith:
This didn't differ across these different CPUs so we don't need this
conditional


trybka:
LGTM

PR #13433, opened 2021-05-06
Delete macOS link_cocoa feature

keith:
This feature doesn't appear to be necessary at this point, importing
this framework will link it successfully.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/packages/util/mock/osx_cc_toolchain_config.bzl
+++ b/src/test/java/com/google/devtools/build/lib/packages/util/mock/osx_cc_toolchain_config.bzl
@@ -6855,29 +6855,6 @@ def _impl(ctx):
         provides = ["profile"],
     )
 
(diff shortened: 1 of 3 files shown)


keith:
I don't know the history here so maybe there was a larger purpose before

trybka:
This is added in rules_apple https://github.com/bazelbuild/rules_apple/blob/bd873c4ed8de4c0cd75a5fc5fa0ad14c73cdce19/apple/macos.bzl#L63 to implicitly add it for UI applications (vs. command-line).

That being said, this is a no-op in our crosstool. Just not sure if someone out there could be depending on it?

keith:
ah thanks, I didn't think to look there, testing to see if any tests are covering it https://github.com/bazelbuild/rules_apple/pull/1149 based on the git blame it dates all the way back to the initial commit to those rules

keith:
Seems like this is still required for the common case, so I guess we can leave it for now

PR #13434, opened 2021-05-06
Fix NullPointerException in ConfigCommand.diffOptions()

moroten:
`@Nullable FragmentOptionsForOutput options1` should be checked for being `null` before use. `bazel config` can otherwise crash, e.g. as the test configuration nowadays is trimmed.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/ConfigCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/ConfigCommand.java
@@ -61,6 +61,7 @@ import com.google.devtools.common.options.OptionsParsingResult;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


gregestren:
@sdtwigg 

Interesting observation. Does this apply to both printing and diffing a configuration? i.e. both `bazel config foo` and `bazel config foo bar`?

moroten:
I only tried `bazel config foo bar` for diffing and is now running my patched version ;)

moroten:
`bazel config`, `bazel config foo` and `bazel config bar` works, but `bazel config foo bar` crashes.

moroten:
Sorry for being unclear. I've now also run all four variants with both patched and unpatched version.
Unpatched version: `bazel config`, `bazel config foo` and `bazel config bar` works, but `bazel config foo bar` crashes.
Patched version: `bazel config`, `bazel config foo`, `bazel config bar` and `bazel config foo bar` all works.

Instead of crashing with `NullPointerException`, the patched version outputs:
```
Displaying diff between configs a8aaf8d028e4c9ba3ee0518043a87e8f514729d15df0ef0dd0f49bd0918cf512 and 9029ef98aed4ba5e0c34310cde7197acbdcf8a841fc6d5a90ca6303fd6d78a87
FragmentOptions com.google.devtools.build.lib.analysis.test.TestConfiguration$TestOptions {
  cache_test_results: null, AUTO
  coverage_report_generator: null, @bazel_tools//tools/test:coverage_report_generator
  coverage_support: null, @bazel_tools//tools/test:coverage_support
  experimental_cancel_concurrent_tests: null, false
  experimental_fetch_all_coverage_outputs: null, false
...
  trim_test_configuration: null, true
}
```

gregestren (approved):
Thanks so much!

Issue #12671, opened 2020-12-07
Swift host tools compiled with wrong cpu on Apple Silicon


thii:
Host tools are being compiled for `x86_64` cpu when compiling a target that uses them on an arm Mac, with an arm Bazel.

```
# BUILD
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_binary")

genrule(
    name = "hello_swift",
    outs = ["hello.swift"],
    cmd = """echo 'print("Hello")' > $@""",
)

swift_binary(
    name = "hello_swift_bin",
    srcs = [":hello_swift"],
)

genrule(
    name = "a",
    outs = ["a.out"],
    cmd = """$(execpath :hello_swift_bin) > $@""",
    tools = [":hello_swift_bin"],
)
```

```
$ bazel build a
...
SUBCOMMAND: # //:hello [action 'Compiling Swift module hello [for host]', configuration: 130cff6c5881621671f254bf93eadfab04237d1f78aac0f4e2233c4b12d50eee, execution platform: @local_config_platform//:host]
(cd /private/var/tmp/_bazel_user/c65ad10e7591af2b858f41477e1f1a20/execroot/build_bazel_rules_swift && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=11.0 \
    XCODE_VERSION_OVERRIDE=12.2.0.12B45b \
  bazel-out/host/bin/tools/worker/worker swiftc @bazel-out/host/bin/hello.swiftmodule-0.params)
SUBCOMMAND: # //:hello [action 'Linking hello [for host]', configuration: 130cff6c5881621671f254bf93eadfab04237d1f78aac0f4e2233c4b12d50eee, execution platform: @local_config_platform//:host]
(cd /private/var/tmp/_bazel_user/c65ad10e7591af2b858f41477e1f1a20/execroot/build_bazel_rules_swift && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=11.0 \
    PATH=/bin:/usr/bin \
    XCODE_VERSION_OVERRIDE=12.2.0.12B45b \
[...]

PR #13440, opened 2021-05-06
Update DEFAULT_MACOS_CPU to match host

keith:
With Apple Silicon machines host tools that use the apple_binary, or
apple fragment to fetch the CPU would be built for the wrong
architecture, even though the C++ logic would correctly identify the
host CPU. This changes this hardcoded value to correctly determine Apple
Silicon as arm64, and otherwise fallback to x86_64.

Fixes https://github.com/bazelbuild/bazel/issues/12671


keith:
There are 2 things to note about this change:

1. if folks still want to build for x86_64 from a Apple Silicon host they can pass `--macos_cpus`
2. attempting to run bazel with `arch -x86_64` on Apple Silicon isn't actually respect, it seems to be related to the JVM arch instead, which means host tools built on Apple Silicon will _always_ be built for arm64, even if you use a rosetta simulated terminal

thii (approved):
Thanks for fixing this.

susinmotion (approved):

Issue #10628, opened 2020-01-21
Feature Request to have child's uses-permissions exported to top level manifest


nkoroste:
### Description of the problem / feature request:

> Replace this line with your answer.

Related: https://github.com/bazelbuild/bazel/issues/5411

Both Gradle and Buck have an option to merge all uses-permissions  manifest enteries from child libs to the top level final binary manifest. It would be nice if Bazel supported that as well.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Have a android_binary target that depends on an android_library target that defines some permissions in it's manifest. 
2. build the android_binary target and inspect it's final AndroidManifest.xml

Expected the permission from library to be present
Actual it's not present.

### What's the output of `bazel info release`?

1.2.1

### Any other information, logs, or outputs that you want to share?

btw, there is a current bug in bazel where it actually inherits **some** permissions from child aar_imports. So if you have android_binary that deps on android_library that deps on aar_import that has permissions -- those permissions will be exported.

Issue #5411, opened 2018-06-15
uses-permissions in AndroidManifest.xml are not exported to App


steeve:
### Description of the problem / feature request:

As per the documentation, `uses-permissions` attributes are never exported to the main app. What is the rationale behind this decision?
Is there any way to have an `android_library` export its permissions to an app that includes it ?

Thanks!

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

```
release 0.14.1-homebrew
```

###  Have you found anything relevant by searching the web?

They are manually removed here:
https://github.com/bazelbuild/bazel/blob/bb4ffd680cf102d6dc2b32895b7bedde8e0cd695/src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java#L157-L185

jin:
I have no idea - @asteinb @ahumesky?

ahumesky:
It's mostly because changing the permissions of your app prevents auto-updating on the play store.

If you have a code base where there are many apps that are independently owned by different teams, and you have many libraries that are similarly independently owned by other teams, and if you inherit permissions from your dependencies, then it means that other teams can change the permissions of your app without your permission or noticing. Another way to put it is that app owners should own the permissions their app asks for. If a new permission gets by and ends up in a release, then your app can't be auto-updated.

One possible response is that this could instead be taken care of in a test where the test asserts the set of permissions that the app has. But this results in basically the same situation: a flat list somewhere of the permissions the app is allowed, just in a test rather than the top-level manifest.

That said, it should be possible to add an attribute to `android_binary` that will enable permissions inheritance. It should also be possible to write a skylark aspect that traverses the dependencies of an `android_binary` that will at least give you all the permissions from the dependencies.

steeve:
I'm closing this as the rationale makes sense.

PR #13445, opened 2021-05-07
Support uses-permission merging in manifest merger

Bencodes:
Adding support for conditionally merging `uses-permissions`.

https://github.com/bazelbuild/bazel/issues/10628
https://github.com/bazelbuild/bazel/issues/5411

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
+++ b/src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
@@ -62,6 +62,7 @@ import org.xml.sax.SAXException;
  *       --manifestValues key value pairs of manifest overrides
  *       --customPackage package to write for library manifest
  *       --manifestOutput path to write output manifest
+ *       --mergeManifestPermissions merge manifest uses-permissions
  * </pre>
  */
 public class ManifestMergerAction {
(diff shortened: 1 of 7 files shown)


src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
```diff
@@ -176,8 +187,7 @@ private static Path removePermissions(Path manifest, Path outputDir)
       }
     }
     // Write resulting manifest to the output directory, maintaining full path to prevent collisions
-    Path output = outputDir.resolve(manifest.toString().replaceFirst("^/", ""));
```
Bencodes, line 179: This was failing on windows machines causing the original files to be modified in place.

Bencodes:
@ahumesky mind taking a look at this?

Bencodes:

src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
ahumesky: It will be easier to import this PR if we can evaluate this change separately. Is it right that this was an existing problem and not needed for merging permissions?

Bencodes: Correct. I can pull this out into another PR that can be imported independently of this, but this PR will need to be rebased against that for the tests to pass.

Bencodes: Pulled this out into https://github.com/bazelbuild/bazel/pull/13760

Bencodes: @ted-xie The tests for this wont pass on windows until the other PR lands. I'll rebase this once that lands.

PR #13448, opened 2021-05-08
Remove unnecessary macOS crosstool configuration

keith:
This cleans up some unnecessary duplication / branching the macOS crosstool

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -251,68 +251,29 @@ def _impl(ctx):
     xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
     xcode_execution_requirements = xcode_config.execution_info().keys()
 
-    if (ctx.attr.cpu == "tvos_arm64" or
-        ctx.attr.cpu == "tvos_x86_64"):
-        cpp_header_parsing_action = action_config(
-            action_name = ACTION_NAMES.cpp_header_parsing,
-            implies = [
-                "preprocessor_defines",
-                "include_system_dirs",
-                "version_min",
-                "objc_arc",
-                "no_objc_arc",
-                "apple_env",
-                "user_compile_flags",
-                "sysroot",
-                "unfiltered_compile_flags",
-                "compiler_input_flags",
-                "compiler_output_flags",
-                "unfiltered_cxx_flags",
-            ],
-            tools = [
-                tool(
-                    path = "wrapped_clang",
-                    execution_requirements = xcode_execution_requirements,
-                ),
-            ],
-        )
-    elif (ctx.attr.cpu == "armeabi-v7a" or
-          ctx.attr.cpu == "darwin_x86_64" or
-          ctx.attr.cpu == "darwin_arm64" or
-          ctx.attr.cpu == "darwin_arm64e" or
-          ctx.attr.cpu == "ios_arm64" or
-          ctx.attr.cpu == "ios_arm64e" or
-          ctx.attr.cpu == "ios_armv7" or
-          ctx.attr.cpu == "ios_i386" or
-          ctx.attr.cpu == "ios_x86_64" or
-          ctx.attr.cpu == "watchos_arm64_32" or
-          ctx.attr.cpu == "watchos_armv7k" or
-          ctx.attr.cpu == "watchos_i386" or
-          ctx.attr.cpu == "watchos_x86_64"):
-        cpp_header_parsing_action = action_config(
-            action_name = ACTION_NAMES.cpp_header_parsing,
-            implies = [
-                "preprocessor_defines",
-                "include_system_dirs",
-                "version_min",
-                "objc_arc",
-                "no_objc_arc",
-                "apple_env",
-                "user_compile_flags",
-                "sysroot",
-                "unfiltered_compile_flags",
-                "compiler_input_flags",
-                "compiler_output_flags",
-            ],
-            tools = [
-                tool(
-                    path = "wrapped_clang",
-                    execution_requirements = xcode_execution_requirements,
-                ),
-            ],
-        )
-    else:
-        cpp_header_parsing_action = None
+    cpp_header_parsing_action = action_config(
+        action_name = ACTION_NAMES.cpp_header_parsing,
+        implies = [
+            "preprocessor_defines",
+            "include_system_dirs",
+            "version_min",
+            "objc_arc",
+            "no_objc_arc",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
+        ],
+        tools = [
+            tool(
+                path = "wrapped_clang",
+                execution_requirements = xcode_execution_requirements,
+            ),
+        ],
+    )
 
     objc_compile_action = action_config(
         action_name = ACTION_NAMES.objc_compile,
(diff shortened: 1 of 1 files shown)


tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -251,68 +251,29 @@ def _impl(ctx):
...
+            "version_min",
+            "objc_arc",
+            "no_objc_arc",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
```
keith, line 268: This action only differed by having `unfiltered_cxx_flags` in 1 branch and not the other, this action exists for all archs, it's just a no-op for some, so we don't need to `if` on that in both places

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -347,578 +308,93 @@ def _impl(ctx):
...
-        )
-    else:
-        objcpp_executable_action = None
+    objcpp_executable_action = action_config(
+        action_name = "objc++-executable",
+        flag_sets = [
+            flag_set(
+                flag_groups = [
+                    flag_group(flags = ["-stdlib=libc++", "-std=gnu++11"]),
+                    flag_group(flags = ["-arch", arch]),
```
keith, line 317: This action only differed for different architectures by hardcoding them here, now that we have a variable for this we can remove that duplication

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -347,578 +308,93 @@ def _impl(ctx):
...
+            "linkstamps",
+            "output_execpath_flags",
+            "runtime_root_flags",
+            "input_param_flags",
+            "strip_debug_symbols",
+            "linker_param_file",
+            "version_min",
+            "apple_env",
+            "sysroot",
+            "cpp_linker_flags",
```
keith, line 389: This action only differed by having `cpp_linker_flags` in 1 branch and not the other, this action exists for all archs, it's just a no-op for some, so we don't need to if on that in both places

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -937,131 +413,53 @@ def _impl(ctx):
...
+            "version_min",
+            "objc_arc",
+            "no_objc_arc",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
```
keith, line 430: This action only differed by having `unfiltered_cxx_flags` in 1 branch and not the other, this action exists for all archs, it's just a no-op for some, so we don't need to if on that in both places

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -937,131 +413,53 @@ def _impl(ctx):
...
```
keith, line 454: This action only differed by having `unfiltered_cxx_flags` in 1 branch and not the other, this action exists for all archs, it's just a no-op for some, so we don't need to if on that in both places

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -347,60 +308,191 @@ def _impl(ctx):
...
+        ],
+        tools = [
+            tool(
+                path = "cc_wrapper.sh",
+                execution_requirements = xcode_execution_requirements,
+            ),
+        ],
+    )
+
+    cpp_link_static_library_action = action_config(
```
keith, line 399: this just moved

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -3314,187 +2712,9 @@ def _impl(ctx):
         requires = [feature_set(features = ["coverage"])],
     )
 
-    if (ctx.attr.cpu == "ios_arm64"):
-        default_link_flags_feature = feature(
```
keith, line 3318: default_link_flags has 3 cases, the macOS archs which just differed by arch, the normal case which just differed by arch, and the opt outs that don't pass `-target` which weren't changed here

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -1399,64 +797,27 @@ def _impl(ctx):
...
+            "objc_arc",
+            "no_objc_arc",
+            "include_system_dirs",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
```
keith, line 812: This action only differed by `unfiltered_cxx_flags` which exists for all archs but just no-ops for some, so we don't have to duplicate that conditional

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -1399,127 +797,51 @@ def _impl(ctx):
...
+            "version_min",
+            "objc_arc",
+            "no_objc_arc",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
```
keith, line 836: This action only differed by `unfiltered_cxx_flags` which exists for all archs but just no-ops for some, so we don't have to duplicate that conditional

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -347,1179 +308,245 @@ def _impl(ctx):
...
+        ],
+        tools = [
+            tool(
+                path = "wrapped_clang_pp",
+                execution_requirements = xcode_execution_requirements,
+            ),
+        ],
+    )
+
+    objcpp_compile_action = action_config(
```
keith, line 464: This action only differed on arch, the `apply_simulator_compiler_flags` feature, which exists for all archs, just no-ops for some, so that's fine, and some actions used `wrapped_clang` instead of `wrapped_clang_pp` which seems to have been an unintentional difference

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -347,1437 +308,276 @@ def _impl(ctx):
...
+            flag_set(
+                flag_groups = [
+                    flag_group(
+                        flags = _deterministic_libtool_flags(ctx) + [
+                            "-no_warning_for_no_symbols",
+                            "-static",
+                            "-filelist",
+                            "%{obj_list_path}",
+                            "-arch_only",
+                            target_system_name,
```
keith, line 563: This action only differed on arch

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -937,1385 +413,237 @@ def _impl(ctx):
...
+                            "-fobjc-link-runtime",
+                            "-ObjC",
+                        ],
+                    ),
+                ],
+                with_features = [with_feature_set(not_features = ["kernel_extension"])],
+            ),
+            flag_set(
+                flag_groups = [
+                    flag_group(flags = ["-arch", target_system_name]),
```
keith, line 602: this action only differed on arch

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -937,1450 +413,262 @@ def _impl(ctx):
...
+            "output_execpath_flags",
+            "runtime_root_flags",
+            "input_param_flags",
+            "force_pic_flags",
+            "strip_debug_symbols",
+            "linker_param_file",
+            "version_min",
+            "apple_env",
+            "sysroot",
+            "cpp_linker_flags",
```
keith, line 663: this action only differed on `cpp_linker_flags` which exists for all archs but no-ops for some

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -2405,135 +693,55 @@ def _impl(ctx):
...
+            "version_min",
+            "objc_arc",
+            "no_objc_arc",
+            "apple_env",
+            "user_compile_flags",
+            "sysroot",
+            "unfiltered_compile_flags",
+            "compiler_input_flags",
+            "compiler_output_flags",
+            "unfiltered_cxx_flags",
```
keith, line 710: this action only differed on `unfiltered_cxx_flags` which exists for all archs but no-ops for some

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -4842,44 +2147,13 @@ def _impl(ctx):
...
-                    ],
-                    flag_groups = [
-                        flag_group(
-                            flags = [
-                                "-no-canonical-prefixes",
-                                "-Wno-builtin-macro-redefined",
-                                "-D__DATE__=\"redacted\"",
-                                "-D__TIMESTAMP__=\"redacted\"",
-                                "-D__TIME__=\"redacted\"",
+                                target_system_name,
```
keith, line 2150: this feature only differed on arch, except for `armeabi-v7a` which excludes the `-target` arg

keith:
this diff is obviously quite large, happy to break it up however we want, right now reviewing by commit should be easiest, and I commented on every feature I changed. I aimed for 0 behavior changes here

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -937,2054 +413,1262 @@ def _impl(ctx):
...
+        flag_sets = [
+            flag_set(
+                actions = [
+                    ACTION_NAMES.c_compile,
+                    ACTION_NAMES.cpp_compile,
+                    ACTION_NAMES.cpp_module_compile,
+                    ACTION_NAMES.cpp_header_parsing,
+                    ACTION_NAMES.objc_compile,
+                    ACTION_NAMES.objcpp_compile,
+                    "objc-executable",
```
trybka, line 1642: should these use ACTION_NAMES....executable?

trybka (commented):
I reviewed this to the best of my ability, and it seems reasonable.

Some notes for @keith :

Most of the remaining conditionals still seem unnecessarily verbose. In particular, can we remove armeabi-v7a? Also, the other pattern I see that could possibly be removed is like:
if (one condition) ... elif (other condition) ... else (default case which should probably never happen, i.e. CPU is not anything we recognize)

Maybe those could be collapsed a bit in a follow-up? 
Certainly wouldn't block this on such a cleanup, though.

keith:
Thanks for reviewing! If you're ok with me doing those in a followup I'd definitely prefer that at this point, and those should be much less churn than this diff. I've been trying to remember why this armeabi-v7a thing exists again but I'm happy to remove it in a follow up if it's not needed anymore

keith:
Working on removing armeabi-v7a from this here https://github.com/bazelbuild/bazel/pull/13449 but I would still prefer we land this first so I don't have to resolve conflicts between them, and in case I miss something there having to revert the whole thing would be tough

trybka:
Follow-ups and conflict resolution all make sense to me. I'm happy to help land this as-is and then process the follow-up(s)

PR #13449, opened 2021-05-08
Remove armeabi-v7a from Apple crosstool

keith:
This is still needed in general, but instead of baking it into this crosstool we can use the separate setup for it.

### DIFF
--- a/tools/osx/crosstool/BUILD.toolchains
+++ b/tools/osx/crosstool/BUILD.toolchains
@@ -1,6 +1,26 @@
+load(":osx_archs.bzl", "OSX_TOOLS_ARCHS")
+
 package(default_visibility = ["//visibility:public"])
 
-load(":osx_archs.bzl", "OSX_TOOLS_ARCHS", "OSX_TOOLS_CONSTRAINTS")
+# Target constraints for each arch.
+# TODO(apple-rules): Rename osx constraint to macOS.
+OSX_TOOLS_CONSTRAINTS = {
+    "armeabi-v7a": ["@platforms//cpu:arm"],
+    "darwin_arm64": ["@platforms//os:osx", "@platforms//cpu:aarch64"],
+    "darwin_arm64e": ["@platforms//os:osx", "@platforms//cpu:aarch64"],
+    "darwin_x86_64": ["@platforms//os:osx", "@platforms//cpu:x86_64"],
+    "ios_arm64": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
+    "ios_arm64e": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
+    "ios_armv7": ["@platforms//os:ios", "@platforms//cpu:arm"],
+    "ios_i386": ["@platforms//os:ios", "@platforms//cpu:x86_32"],
+    "ios_x86_64": ["@platforms//os:ios", "@platforms//cpu:x86_64"],
+    "tvos_arm64": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
+    "tvos_x86_64": ["@platforms//os:ios", "@platforms//cpu:x86_64"],
+    "watchos_arm64_32": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
+    "watchos_armv7k": ["@platforms//os:ios", "@platforms//cpu:arm"],
+    "watchos_i386": ["@platforms//os:ios", "@platforms//cpu:x86_32"],
+    "watchos_x86_64": ["@platforms//os:ios", "@platforms//cpu:x86_64"],
+}
 
 OSX_DEVELOPER_PLATFORM_CPUS = [
     "aarch64",
@@ -19,5 +39,6 @@ OSX_DEVELOPER_PLATFORM_CPUS = [
         toolchain = "@local_config_cc//:cc-compiler-" + arch,
         toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
     )
-    for arch in OSX_TOOLS_ARCHS for cpu in OSX_DEVELOPER_PLATFORM_CPUS
+    for arch in OSX_TOOLS_ARCHS + ["armeabi-v7a"]
+    for cpu in OSX_DEVELOPER_PLATFORM_CPUS
 ]
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -55,31 +55,6 @@ def _deterministic_libtool_flags(ctx):
     return []
 
 def _impl(ctx):
-    if (ctx.attr.cpu == "armeabi-v7a"):
-        host_system_name = "armeabi-v7a"
-    elif (ctx.attr.cpu == "darwin_x86_64" or
-          ctx.attr.cpu == "darwin_arm64" or
-          ctx.attr.cpu == "darwin_arm64e" or
-          ctx.attr.cpu == "ios_arm64" or
-          ctx.attr.cpu == "ios_arm64e" or
-          ctx.attr.cpu == "ios_armv7" or
-          ctx.attr.cpu == "ios_i386" or
-          ctx.attr.cpu == "ios_x86_64" or
-          ctx.attr.cpu == "tvos_arm64" or
-          ctx.attr.cpu == "tvos_x86_64" or
-          ctx.attr.cpu == "watchos_arm64_32" or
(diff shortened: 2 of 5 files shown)


keith:
Checking to see if this breaks anything

keith:
example failure if you remove this https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/8780#9ad3b275-6f59-48f9-b0d7-38c4a6793694

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -19,5 +20,17 @@ OSX_DEVELOPER_PLATFORM_CPUS = [
         toolchain = "@local_config_cc//:cc-compiler-" + arch,
         toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
     )
-    for arch in OSX_TOOLS_ARCHS for cpu in OSX_DEVELOPER_PLATFORM_CPUS
+    for arch in OSX_TOOLS_ARCHS
+    for cpu in OSX_DEVELOPER_PLATFORM_CPUS
```
keith, line 24: buildifier

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -1,5 +1,6 @@
 package(default_visibility = ["//visibility:public"])
 
+load("@local_config_platform//:constraints.bzl", "HOST_CONSTRAINTS")
```
keith, line 3: This works but I assume this is safe to add since the C++ logic does this too https://github.com/bazelbuild/bazel/blob/8a42645ec500874b0440475763ab680d5efc1e6a/tools/cpp/BUILD.toolchains.tpl#L1

tools/osx/crosstool/osx_archs.bzl
```diff
@@ -26,32 +26,10 @@ OSX_TOOLS_NON_DEVICE_ARCHS = [
...
-    "watchos_x86_64": ["@platforms//os:ios", "@platforms//cpu:x86_64"],
-    "tvos_x86_64": ["@platforms//os:ios", "@platforms//cpu:x86_64"],
-    "armeabi-v7a": ["@platforms//cpu:arm"],
-    "ios_armv7": ["@platforms//os:ios", "@platforms//cpu:arm"],
-    "ios_arm64": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
-    "ios_arm64e": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
-    "watchos_armv7k": ["@platforms//os:ios", "@platforms//cpu:arm"],
-    "watchos_arm64_32": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
-    "tvos_arm64": ["@platforms//os:ios", "@platforms//cpu:aarch64"],
-}
```
keith, line 57: This was only used in 1 file so I moved it there since it also slightly diffs now in keys vs the archs lists

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -19,5 +39,6 @@ OSX_DEVELOPER_PLATFORM_CPUS = [
         toolchain = "@local_config_cc//:cc-compiler-" + arch,
         toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
     )
-    for arch in OSX_TOOLS_ARCHS for cpu in OSX_DEVELOPER_PLATFORM_CPUS
+    for arch in OSX_TOOLS_ARCHS + ["armeabi-v7a"]
```
keith, line 42: we add this manually here so we can exclude it from the `OSX_TOOLS_ARCHS` so it doesn't get an autocreated toolchain in the other loops, but we still need to reference the manually created one here, so we keep it in the `OSX_TOOLS_CONSTRAINTS` dictionary

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,31 +55,6 @@ def _deterministic_libtool_flags(ctx):
     return []
 
 def _impl(ctx):
-    if (ctx.attr.cpu == "armeabi-v7a"):
```
keith, line 58: there's some more to cleanup in this file, but I would prefer we land https://github.com/bazelbuild/bazel/pull/13448 first so I don't have to resolve conflicts in that

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -113,18 +86,11 @@ def _impl(ctx):
...
-        arch = ctx.attr.cpu.split("_", 1)[-1]
-
-    if (ctx.attr.cpu == "armeabi-v7a"):
-        target_libc = "armeabi-v7a"
-    elif (ctx.attr.cpu == "ios_arm64" or
-          ctx.attr.cpu == "ios_arm64e" or
-          ctx.attr.cpu == "ios_armv7" or
-          ctx.attr.cpu == "ios_i386" or
-          ctx.attr.cpu == "ios_x86_64"):
+    if (ctx.attr.cpu == "ios_arm64" or
```
trybka, line 89: `cpu.startswith("ios_")` ?

tools/osx/crosstool/BUILD.toolchains
keith: Yep. Without this you get this error when running android tests https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/8780#9ad3b275-6f59-48f9-b0d7-38c4a6793694

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -183,6 +145,9 @@ def _impl(ctx):
     else:
         fail("Unreachable")
 
+    host_system_name = "x86_64-apple-macosx"
```
trybka, line 148: Apologies if this is addressed elsewhere, but what about the arm variant? Or is that not supported yet?

trybka (commented):
Overall much cleaner, thanks!

tools/osx/crosstool/cc_toolchain_config.bzl
keith: do you know how this value is actually used? it was hardcoded before so i guess it doesn't entirely matter, but yea I was thinking the same thing

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -113,75 +86,20 @@ def _impl(ctx):
...
-        target_libc = "watchos"
     else:
-        fail("Unreachable")
+        target_libc = ctx.attr.cpu.split("_")[0]
 
-    if (ctx.attr.cpu == "armeabi-v7a"):
-        abi_version = "armeabi-v7a"
-    elif (ctx.attr.cpu == "darwin_x86_64"):
+    if ctx.attr.cpu == "darwin_x86_64":
+        abi_libc_version = "darwin_x86_64"
```
keith, line 95: I also wonder if this condition should be updated for Apple Silicon

tools/osx/crosstool/cc_toolchain_config.bzl
trybka: It looks like it is now vestigial. https://cs.opensource.google/search?q=lang:java%20host_system_name%20-f:Test.java&sq=&ss=bazel

keith: Nice, I can remove this entirely (probably separately) but that might be a breaking change?

trybka:
@keith it looks like this may have broken some things in rules_cc: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2047#9b7530e6-d708-4852-938d-321d0af6e90c

keith:
Thanks! https://github.com/bazelbuild/rules_cc/pull/105

meteorcloudy:
TensorFlow is failing with Bazel@HEAD due to this change: https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/1143#a46ef18c-b4dc-43c1-9118-4a7620e72c1a

Any idea how to fix it?

keith:
Welp, looks like they use rules_cc's toolchain https://github.com/tensorflow/tensorflow/blob/cbc1ddc5b52ac13a5d2563f7b5661e72173d7084/tensorflow/workspace0.bzl#L106 but also reference some files from this repo, so there is an incompatibility here.

I've submitted https://github.com/tensorflow/tensorflow/pull/53440 to just drop rules_cc

keith:
@meteorcloudy this is fixed on tensorflow HEAD now, thanks for pinging this!

PR #13452, opened 2021-05-09
Compile Apple tools as fat binaries if possible

keith:
The Apple toolchain has 2 native binaries that are inputs to every
single action. Because of this if you want to share caches between Apple
Silicon machines and Intel machines, you either need to force them to be
x86_64 binaries and suffer the performance loss on Apple Silicon
machiens, or use fat binaries so the sha's match on both architectures,
which is what this change does. These binaries are so small that the
size impact of this doesn't matter. Since Apple Silicon support requires
Xcode 12 this falls back to compiling the single architecture binary if
it fails, under the assumption that means you're on Xcode 11 or lower.
We don't have a better indication at this point of what Xcode version
you're using, so this seems like a fine workaround until Xcode 12 is the
minimum supported version.


keith:
Note this doesn't touch xcode-locator because that's not an input to other actions (or the embedded version is used instead)

tools/cpp/osx_cc_configure.bzl
```diff
@@ -83,6 +85,32 @@ def _compile_cc_file(repository_ctx, src_name, out_name):
...
+        "--sdk",
+        "macosx",
+        "clang",
+        "-mmacosx-version-min=10.9",
+        "-std=c++11",
+        "-lc++",
+        "-arch",
+        "arm64",
+        "-arch",
+        "x86_64",
```
keith, line 104: This is the only difference from the function above, but since you can't recursively call functions in starlark I had to duplicate this

keith:
I tested this with Xcode 11.7 and it works as expected by falling back to the single arch binary

brentleyjones (approved):

brentleyjones:
@dmaclach @allevato Any thoughts on this?

brentleyjones:
AFAIK, this is needed in order to kick off macOS remote execution from an Apple Silicon machine where the executors are Intel.

brentleyjones:
@kaylathar Per this https://github.com/bazelbuild/bazel/pull/13873#issuecomment-901984761, could you please take a look at this? 🙏

thii:
@keith Are you using this internally? We noticed significant cache misses after using this patch.

keith:
We're not yet, we're only testing M1s atm. Did you debug the misses back to this?

thii:
I did a quick debug and it turned out that the fat wrapped_clang is not deterministic. Might be a bug in clang.

thii:
I can confirm that our cache hit rate went back to normal after reverting this change.

The root cause seems because clang adhoc signs binaries for arm64, and since the signatures are different each time, the result fat binaries would be different between builds. This results in a lot of cache misses because wrapped_clang is an input of most actions when building for Apple platforms.

tools/cpp/osx_cc_configure.bzl
```diff
@@ -83,6 +84,32 @@ def _compile_cc_file(repository_ctx, src_name, out_name):
...
+        "-O3",
+        "-o",
+        out_name,
+        src_name,
+    ], 30)
```
thii, line 108: Here's what works for us:

```starlark
    repository_ctx.execute([
        "env",
        "-i",
        "DEVELOPER_DIR={}".format(env.get("DEVELOPER_DIR", default = "")),
        "xcrun",
        "--sdk",
        "macosx",
        "clang",
        "-mmacosx-version-min=10.9",
        "-std=c++11",
        "-lc++",
        "-O3",
        "-arch",
        "arm64",
        "-arch",
        "x86_64",
        "-Wl,-no_adhoc_codesign",
        "-o",
        out_name,
        src_name,
    ], 30)
    repository_ctx.execute([
        "env",
        "-i",
        "codesign",
        "--force",
        "--sign",
        "-",
        out_name,
    ], 30)
```

keith: is the second codesign reproducible here?

thii: Reproducible with the same machine type (Intel vs Intel, M1 vs M1).

keith: updated with what we learned here

Issue #13443, opened 2021-05-07
try-import dies on Windows if path is invalid


erenon:
### Description of the problem / feature request:

From The Users Guide: "The difference between import and try-import is that Bazel fails if the import‘ed file is missing (or can’t be read), but not so for a try-import‘ed file." However, if a  `.bazelrc` contains `try-import /no/such/file`, on windows, bazel dies. (On Linux, the import ignored as expected).

### Feature requests: what underlying problem are you trying to solve with this feature?

I'd like to make this .bazelrc work:

```
try-import /mnt/shared/defaults.bazelrc
try-import E:/shared/defaults.bazelrc
```

This would allow cross-platform projects to pick up platform specific defaults easily.
(Note: --enable_platform_specific_config is not good enough for two reasons: (a) it can't differentiate different windows and linux versions (b) it still needs a single path that can be imported on both platforms (windows and linux))

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

On windows:

```
> cat .bazelrc
try-import /mnt/shared/defaults.bazelrc
> bazel info
[...] network paths are unsupported
```

And it dies.

### What operating system are you running Bazel on?

Windows 10

### What's the output of `bazel info release`?

Bazel 4.0.0

###  Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

Root cause: `RcFile::ParseFile` invokes `blaze_util::ReadFile`, that converts the string filename to a `Path`, and the `Path` constructor invokes `BAZEL_DIE`. The issue can be easily avoided e.g: by making `ParseFile` use iostreams instead of blaze_util. I'm happy to provide a PR, if this is viable.

aiuto:
It seems like we should make the posix and windows ReadFile implementations the same rather than add a special purpose Readfile for parsing RC files.

erenon:
That's OK. However, I couldn't figure out what the windows implementation does in extra (I have only Linux experience). I do no understand why the windows Path implementation rejects certain kinds of paths. Thank for taking a look!

lberki:
Routing to @meteorcloudy .

My non-binding opinion is that the reason why this happens is that we thought calling `BAZEL_DIE()` when Bazel encounters a path it cannot parse was both easy and correct. This is almost always true, except during parsing `try-import` because then errors are supposed to be ignored.

[...]

PR #13453, opened 2021-05-09
Make try-import ignore unsupported paths

erenon:
Previously on windows, `try-import /foo/bar` invoked BAZEL_DIE.
Fixes https://github.com/bazelbuild/bazel/issues/13443

### DIFF
--- a/src/main/cpp/rc_file.cc
+++ b/src/main/cpp/rc_file.cc
@@ -15,6 +15,8 @@
 #include "src/main/cpp/rc_file.h"
 
 #include <algorithm>
+#include <fstream>
+#include <sstream>
 #include <utility>
 
 #include "src/main/cpp/blaze_util.h"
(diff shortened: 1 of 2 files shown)


aiuto:
This feels like the wrong fix.  We should not be masking the divergent behavior of ReadFile on windows vs. posix. We should be bringing Windows into alignment with posix.

sgowroji:
Hello @erenon, Did you get a chance to check the above [comment](https://github.com/bazelbuild/bazel/pull/13453#issuecomment-840748852). Can you please update the PR and share us the latest changes.

erenon:
Hi, the comment above is not specific enough to be actioned by me. The change was good enough for my use-case. If the bazel team plans to straighten the path handling on windows (esp. the handling of reparse points), a clear strategy needs to be articulated first.

meteorcloudy:
Closing since https://github.com/bazelbuild/bazel/issues/13443 is already fixed by https://github.com/bazelbuild/bazel/pull/13698

Issue #3709, opened 2017-09-08
Better Invalid SHA256 checksum message


ianthehat:
When you specify an incorrect checksum, the message you get is
```
Invalid SHA256 checksum.
```
It would be nice if it told you what the correct checksum was. This would be an easy way of updating a repository version. Update the reference, set the sha to an invalid value, try to build, cut'n'paste in the correct checksum, profit.

dungeonmaster18:
Hi, Can someone help me to get started with this issue. I am new here.

keir:
The error message is not clear, especially if you are new to Bazel and are gluing together pieces of other Bazel WORKSPACE files to get a project going. Here's a suggested alternate message that would make it more clear:

```
The SHA256 checksum specified in the WORKSPACE file does not match the SHA256
checksum of the downloaded file. The checksum exists to ensure a secure build,
since remote files can change. If the file is correct, update the SHA256 checksum.
("sha256 = ") in your WORKSPACE file. Otherwise, this could be a security compromise.

File:
/home/keir/foo/bazel-build/..../foo.tar.gz"

Expected SHA256 (from WORKSPACE):
1ee310e5d0a19b9d584a855000434bb724aa744745d5b8ab1855c85bff8a8e21

Actual SHA256 from file:
6e16c8bc91b1310a44f3965e616383dbda48f83e8c1eaa2370a215057b00cabe
```

deepaknalore:
Hi! I would like to pick this up.

jin:
@deepaknalore go for it!

mig35:
This should be closed because it is already implemented and merged in https://github.com/bazelbuild/bazel/pull/5991

jin:
#5991 isn't merged. It's still open here: https://github.com/bazelbuild/bazel/pull/5996

pscheid2:
I'm not sure this issue is still relevant. The error message to be improved ("Invalid SHA256 checksum.") is no longer anywhere in the repo and files touched in the old PRs on this issue don't have any special checksum validation logic.

redone9211:
i would like to contribute to this issue , can you let me know how i can solve it.

PR #5996, opened 2018-08-27
Issue #3709: Better Invalid SHA256 checksum message

deepaknalore:
This is related to issue: #3709
I have updated the error message from "Invalid 256 checksum" to
"The SHA256 checksum specified in the WORKSPACE file does not match the SHA256
checksum of the downloaded file. The checksum exists to ensure a secure build,
since remote files can change. If the file is correct, update the SHA256 checksum.
("sha256 = ") in your WORKSPACE file. Otherwise, this could be a security compromise.

URL:
http://127.0.0.1:$fileserver_port/repo.zip

Actual SHA26 (from WORKSPACE):
a random string is not valid"

Updated the test case as well.

PR #13460, opened 2021-05-10
Attempted fix for issue #3709

anandramakris:
Notation: sha256 refers to the user-provided string, and SHA256 refers to the checksum of the file specified in the URL.
An “invalid” sha256 is a string that does not pass RepositoryCache.isValid().
The function fails when the string is not a 64-character hexadecimal value.

Issue #3709 asked for an error message that, when the user provides an invalid sha256 string as a parameter of http_archive(), http_file(), or http_jar(), displays to the user both the correct SHA256 and the user’s incorrect sha256.
I decided the best way to solve the problem was to create an entirely new error message rather than modifying the message shown when an InvalidChecksumException is thrown.

I created two new messages within DownloadManager.download() using EventHandler.
I made an info message when the user’s string is null or empty, and an error message when the user’s string is not null or empty but is still invalid.
The reason I created an info message rather than an error message in the first case was that an error message would cause test_query_cached() within external_integration_test.sh to fail as it does not pass an sha256 string at all.
I modified test_sha_weird() in external_integration_test.sh so that the expected error message is my new error rather than the InvalidChecksumException() one.

I tested many different cases for both the sha256 and urls fields when using http_file().
For the URL, I tested the cases where it was null, empty, not a working URL, and a working URL.
For sha256, I tested what happens if it was null, empty, invalid but not null or empty, valid but not equal to the working URL’s SHA256, and valid but equal to the working URL’s SHA256.
I tested all combinations of these cases, both if the file linked to in the URL was previously downloaded and if it was not.

This message now appears in the cases when the URL is a working URL and when sha256 is null or empty:
"User provided a null or empty SHA256 checksum."
This message now appears in the case when the URL is a working URL and when sha256 is invalid but not null or empty:
"Invalid SHA256 checksum (sha256). The correct checksum is (SHA256)."
As expected, the other cases have exactly the same result as they had before I changed the code.


anandramakris:
@googlebot I signed it!

src/test/shell/bazel/external_integration_test.sh
```diff
@@ -1109,11 +1109,11 @@ EOF
   # Simulate going offline by removing the external archive
   rm -f "${EXTREPODIR}/ext.zip"
   bazel query 'deps("@ext//:bar")' > "${TEST_log}" 2>&1 \
-    || fail "expected success"
+   || fail "expected success"
```
janakdr, line 1112: nit: Better to leave unrelated whitespace alone (here and below)

src/test/shell/bazel/external_integration_test.sh
```diff
@@ -910,7 +910,7 @@ http_archive(
 )
 EOF
   bazel build @repo//... &> $TEST_log && fail "Expected to fail"
-  expect_log "[Ii]nvalid SHA-256 checksum"
+  expect_log "Invalid SHA256 checksum 'a random string'. The correct checksum is '$(sha256sum repo.zip | head -c 64)'."
```
janakdr, line 913: Is the original "[Ii]nvalid SHA-256 checksum" string also being printed? If so, can we merge those messages so there's no duplication? I think the approach taken in https://github.com/bazelbuild/bazel/pull/5996/files might be the correct one.

janakdr (commented):
Thanks for the contribution!

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
```diff
@@ -31,6 +31,12 @@ private InvalidChecksumException(String msg) {
     }
   }
 
+  public static final class ICExceptionExtended extends Exception {
```
janakdr, line 34: nit: it's generally best practice to avoid abbreviations in class names unless they're standard. The reader may not know what "IC" stands for here. Also "extended" makes it sound like this exception extends another exception, but it doesn't. Maybe "MismatchedChecksumException" would be a better name?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
```diff
@@ -31,6 +31,12 @@ private InvalidChecksumException(String msg) {
...
+    private ICExceptionExtended(KeyType keyType, String hash, String correctHash) {
+      super("Invalid " + keyType + " checksum '" + hash + "'. The correct checksum is '" + correctHash + "'.");
```
janakdr, line 36: nit: line length, and also I think that exception messages generally don't have trailing periods.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
```diff
@@ -47,6 +53,12 @@ public static Checksum fromString(KeyType keyType, String hash) throws InvalidCh
     return new Checksum(keyType, HashCode.fromString(hash));
   }
 
+  public static void stringComparison(String hash, String correctHash) throws ICExceptionExtended {
+    if(hash != correctHash) {
```
janakdr, line 57: You're using reference equality to compare the strings here. With a few exceptions, such a comparison will always fail. You want to use semantic equality, .equals(). You would need to investigate why there are no failing tests with this code: it's likely this would cause crashes in production, so it would be good to make sure there's test coverage for this.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
```diff
@@ -235,6 +236,15 @@ public Path download(
     } else if (repositoryCache.isEnabled()) {
       String newSha256 = repositoryCache.put(destination, KeyType.SHA256, canonicalId);
       eventHandler.handle(Event.info("SHA256 (" + rewrittenUrls.get(0) + ") = " + newSha256));
+      
+      if(!sha256.isEmpty()) {
+        try {
+          Checksum.stringComparison(sha256, newSha256);
+        }
+        catch (Checksum.ICExceptionExtended e) {
```
janakdr, line 244: nit: `} catch ...` should all be on one line.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
```diff
@@ -235,6 +236,15 @@ public Path download(
...
+          throw new IOException(e.getMessage());
```
janakdr, line 245: If you're throwing an IOException with the message of the thrown exception, then why not just make ICExceptionExtended a subclass of IOException, and not catch it here at all?

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/SuperChecksum.java
```diff
@@ -0,0 +1,32 @@
...
+    Checksum checksum = null;
+    
+    /* The if statement makes sure Checksum.fromString() will never return an exception. */
+    try {
+      if(keyType == KeyType.SHA256 && keyType.isValid(hash)) {
+        checksum = Checksum.fromString(keyType, hash);
+      }
+    }
+    catch(Checksum.InvalidChecksumException e) {
+    }
```
janakdr, line 28: if the try {} block should never throw an exception, then this catch block should crash, via something like `throw new IllegalStateException("Impossible exception from " + keyType + " and " + hash, e)`.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/SuperChecksum.java
```diff
@@ -0,0 +1,32 @@
...
+
+    return new SuperChecksum(checksum);
+  }
+}
```
janakdr, line 32: nit: missing newline at end of file.

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/SuperChecksum.java
```diff
@@ -0,0 +1,32 @@
+package com.google.devtools.build.lib.bazel.repository.downloader;
+
+import com.google.common.base.Optional;
+import com.google.common.hash.HashCode;
+import com.google.devtools.build.lib.bazel.repository.cache.RepositoryCache.KeyType;
+import com.google.devtools.build.lib.bazel.repository.downloader.Checksum;
+public class SuperChecksum {
```
janakdr, line 7: Please add documentation for this class. Right now, it appears to be a utility class that either stores null or a checksum inside it. I don't see the benefit of that versus a static method like `@Nullable Checksum maybeGetChecksum(KeyType keyType, String hash)`.

nit: put a newline before the class declaration.

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -634,13 +635,28 @@ public StructImpl download(
...
-    } catch (RepositoryFunctionException e) {
-      checksum = Optional.<Checksum>absent();
-      checksumValidation = e;
+    Checksum checksum;
+    
+    if(!sha256.isEmpty() && !integrity.isEmpty()) {
+      /* Terminates the program if both sha256 and integrity are present */
+      throw Starlark.errorf("Expected either 'sha256' or 'integrity', but not both");
+    }
+    else if(!sha256.isEmpty()) {
```
janakdr, line 644: nit: `} else ...` on one line, here and below. And put a space after `if` and before `(`

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/Checksum.java
```diff
@@ -47,6 +53,12 @@ public static Checksum fromString(KeyType keyType, String hash) throws InvalidCh
...
+      throw new ICExceptionExtended(KeyType.SHA256, hash, correctHash);
+    }
+  }
+
   /** Constructs a new Checksum from a hash in Subresource Integrity format. */
   public static Checksum fromSubresourceIntegrity(String integrity)
```
janakdr, line 63: Should this be annotated @Nullable since the returned checksum can now be null?

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -634,13 +635,28 @@ public StructImpl download(
...
             /* checksumGiven= */ !Strings.isNullOrEmpty(sha256)
                 || !Strings.isNullOrEmpty(integrity));
-    Optional<Checksum> checksum;
-    RepositoryFunctionException checksumValidation = null;
-    try {
-      checksum = validateChecksum(sha256, integrity, urls);
-    } catch (RepositoryFunctionException e) {
-      checksum = Optional.<Checksum>absent();
-      checksumValidation = e;
+    Checksum checksum;
```
janakdr, line 638: should this be annotated @Nullable now?

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -884,13 +898,28 @@ public StructImpl downloadAndExtract(
...
+      }
+      catch(Checksum.InvalidChecksumException e) {
+        throw new RepositoryFunctionException(Starlark.errorf(
+          "Definition of repository %s: %s at %s",
+          rule.getName(), e.getMessage(), rule.getLocation()),
+          Transience.PERSISTENT);
+      }
+    }
+    else {
+      checksum = null;
```
janakdr, line 922: this seems like largely duplicated code that wasn't duplicated in the original. Can you not keep the utility method to avoid duplication?

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -1029,40 +1056,6 @@ private Checksum calculateChecksum(Optional<Checksum> originalChecksum, Path pat
     }
   }
 
-  private Optional<Checksum> validateChecksum(String sha256, String integrity, List<URL> urls)
-      throws RepositoryFunctionException, EvalException {
```
janakdr, line 1033: A suggestion about how you might be able to do this more minimally: define a small utility class:

```
private static class ChecksumOrValidationError {
  Optional<Checksum> checksum;
  Optional<String> messageForValidationError;

  private ChecksumOrValidationError(@Nullable Checksum checksum, @Nullable String messageForValidationError) {
    Preconditions.checkState(
        checksum == null || messageForValidationError == null, "checksum %s and message %s both non-null", checksum, messageForValidationError);
    this.checksum = Optional.of(checksum);
    this.messageForValidationError = Optional.of(messageForValidationError);
  }

  static void forNullableChecksum(@Nullable Checksum checksum) {
    return new ChecksumOrValidationError(checksum, null);
  }

  static void forValidationError(String messageForValidationError) {
    return new ChecksumOrValidationError(Preconditions.checkNotNull(messageForValidationError));
  }

  void maybeThrowError(Rule rule, Checksum correctChecksum) throws RepositoryFunctionException {
    if (messageForValidationError.isPresent()) {
      throw new RepositoryFunctionException(
              Starlark.errorf(
                  "Definition of repository %s: %s at %s (correct checksum is %s)",
                  rule.getName(), messageForValidationError.get(), rule.getLocation(), correctChecksum),
              Transience.PERSISTENT);
    }
  }
}
```

Then I think you can return an instance of this utility class from validateChecksum, and pass in the natural arguments where the old code checks the `checksumValidation` variable. The only issue will be accessing the actual sha256 hash from this function, since it's currently only available in the Download Manager. For that purpose, I would change DownloadManager to return a Pair<Path, Checksum> instead of a Path. Then you can just read out the Checksum from it.

sgowroji:
Hello @anandramakris, Can you please take a look on above comments from @janakdr . Thanks!

PR #13461, opened 2021-05-10
Add `--host_macos_cpus` flag

keith:
When your macOS build has tools that build with the Apple rules, such as
`macos_command_line_application`, you likely want to share caches
between Apple Silicon and Intel machines. Today your tool would build
for the host architecture (ideally, see https://github.com/bazelbuild/bazel/pull/13440)
which means the tools would differ between the different architectures,
leading to split caches. This flag allows you to pass
`--host_macos_cpus=arm64,x86_64` in order to produce a fat binary for
host tools. This has the downside of increasing binary size for your
tools, but likely that trade-off is worth it for almost everyone versus
having split caches.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/apple/AppleCommandLineOptions.java
@@ -322,6 +322,16 @@ public class AppleCommandLineOptions extends FragmentOptions {
   public List<String> macosCpus;
 
   @Option(
+          name = "host_macos_cpus",
+          allowMultiple = true,
+          converter = CommaSeparatedOptionListConverter.class,
+          defaultValue = "null",
+          documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+          effectTags = {OptionEffectTag.LOSES_INCREMENTAL_STATE, OptionEffectTag.LOADING_AND_ANALYSIS, OptionEffectTag.AFFECTS_OUTPUTS},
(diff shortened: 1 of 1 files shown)


keith:
Here's an example alternative for `macos_command_line_application` specifically https://github.com/bazelbuild/rules_apple/pull/1155

allevato:
@dmaclach Any concerns here?

brentleyjones:
I wanted to make sure my comment on the other PR was seen here as well: https://github.com/bazelbuild/rules_apple/pull/1155#issuecomment-837579275

brentleyjones:
@dmaclach @allevato Friendly ping 🤗

brentleyjones:
This is also very useful for remote execution, being able to have the host and remote executor be on differing architectures.

brentleyjones:
Seems this is needed to get https://github.com/bazelbuild/rules_apple/commit/73c041db161f5058f2e4cc124b7591300618b5a6 to work for host tools.

thii:
Can we please have this in? Although it's possible to do this with Starlark, it's a pain to update the rules everywhere; if we have this, it would be just a single flip.

keith:
Last I checked it required more changes to work

sgowroji:
Hello @keith, Could you please confirm whether all the required changes are added in this PR ?  Asking you this query with respect to the comment mentioned [here](https://github.com/bazelbuild/bazel/pull/13461#issuecomment-970403448). So that i can request  for review and try for merge.

keith:
dropping this for now, not sure if we really need anymore or not, we can bring back up if needed

PR #13462, opened 2021-05-11
Use `xcrun` to invoke `install_name_tool`

thii:
`/usr/bin/install_name_tool` is a shim; it invokes `install_name_tool`
in the current active Xcode.

This allows using this template file to configure toolchain on
non-macOS. This doesn't change the behavior as both will end up invoking
the same tool.

### DIFF
--- a/tools/cpp/osx_cc_wrapper.sh.tpl
+++ b/tools/cpp/osx_cc_wrapper.sh.tpl
@@ -27,7 +27,7 @@
 #
 set -eu
 
-INSTALL_NAME_TOOL="/usr/bin/install_name_tool"
+INSTALL_NAME_TOOL=/usr/bin/xcrun install_name_tool
 
 LIBS=
 LIB_DIRS=


tools/cpp/osx_cc_wrapper.sh.tpl
```diff
@@ -110,7 +110,7 @@ for rpath in ${RPATHS}; do
         if [[ -n "${libname-}" ]]; then
             libpath=$(get_library_path ${lib})
             if [ -n "${libpath}" ]; then
-                ${INSTALL_NAME_TOOL} -change $(get_otool_path "${libpath}") \
+                "${INSTALL_NAME_TOOL}" -change $(get_otool_path "${libpath}") \
```
keith, line 113: Maybe just in-line this? With this currently I think your quotes are off?

thii: Thanks

tools/cpp/osx_cc_wrapper.sh.tpl
```diff
@@ -27,7 +27,7 @@
 #
 set -eu
 
-INSTALL_NAME_TOOL="/usr/bin/install_name_tool"
+INSTALL_NAME_TOOL="/usr/bin/xcrun install_name_tool"
```
keith, line 30: i think it would be better to inline this now that it has a space to be more careful around folks accidentally quoting this

sgowroji:
Hello @thii, Could you please resolve the above review comments.Thanks!

keertk:

keith:
https://github.com/bazelbuild/bazel/pull/16951

PR #13465, opened 2021-05-11
Update remote worker to use read offset of a byte stream read request

coeuvre:
This fixes digest mismatch issues when downloading outputs from remote worker.


meisterT:
What's the status of this PR?

coeuvre:
I don't remember the reason behind this PR. Closing it for now.

Issue #11986, opened 2020-08-21
Missing -target arg in cpp compilation for native cpp rules


jerrymarino:
### Description of the problem / feature request:

When I run `bazel build //ios-app --cpu ios_arm64` it's missing the argument for `-target` on objc compilation for `native.objc_library` sources

### Feature requests: what underlying problem are you trying to solve with this feature?

Compile an iOS app correctly. While this doesn't seem to error out, I noticed this was caught by Xcode 12 ( beta )

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Build an iOS app with the native rules

### What operating system are you running Bazel on?

macOS catalina

### What's the output of `bazel info release`?

3.4.1

### Any other information, logs, or outputs that you want to share?

I believe this used to work at one point, and seems like a regression. It is still passing this flag to the linker commands correctly

oquenchil:
Thanks for the early warning.

 @googlewalt Have you bumped into this?

googlewalt:
Internally we use a different crosstool file, so we don't have this issue.  It does appear that the bazel osx crosstool does not set `-target` for objc/objc++ compile actions, though AFAICT it's been like that for at least a couple years.

oquenchil:
Alright, thank you Walt. I will mark this as "help wanted" in case someone from the community wants to contribute a fix.

keith:
I see `-target` being passed here, @jerrymarino can you post the full command you see that doesn't pass `-target`?

keith:
we can close unless this comes up again or we have repro steps, merging https://github.com/bazelbuild/bazel/pull/13466 is probably good to do anyways to be more correct

PR #13466, opened 2021-05-11
Pass `-target` in addition to `-arch` in compiling commands in Apple crosstool

thii:
This allows using this crosstool file for cross-compiling from non-Apple
host platforms. On macOS, `-arch <arch>`  and
`-m<platform>-version-min=<version_min>` will be expanded to
`-triple <arch>-apple-<platform><version_min>` in the frontend
invocation, but on a Linux host, without `-target`, the expanded triple
flags would be `-triple <arch>-pc-linux-gnu`.

Keeping `-arch` here, since `-target arm64-apple-<platform>` alone may
end up being rewritten to `aarch64-apple-<platform>` (see
https://github.com/mstorsjo/xcode-cross/commit/4efe326b18bb1f592b85f6ac0cf019f2873f79b0)


keith:
If we're passing target we can probably drop the arch arg? Also see https://github.com/bazelbuild/bazel/pull/13448 which should make this necessary in fewer places

thii:
I'll rebase this after your PR is merged.

I've actually never seen the target being rewritten to `aarch64-apple-darwin` as mentioned in the linked commit, maybe it's only with older clang? If there's no problem I can drop the `-arch` argument.

thii:
Couldn't reproduce myself. Rebased and removed `-arch`.

Also it seems this will fix https://github.com/bazelbuild/bazel/issues/11986.

keith (approved):

keith:
looks like the failure is related

thii:
There's still some places in this file that can be simplified. I did a cleanup of it a while ago while implementing another toolchain, and the target system name can already be derived from `arch` like:

```starlark
def _target_system_name(cpu):
    platform, _, cpu = cpu.partition("_")
    if platform == "darwin":
        platform = "macosx"
    return "{}-apple-{}".format(cpu, platform)
```

https://github.com/apple-cross-toolchain/rules_applecross/blob/e59e4a02aa41b918b919d3711bb9c167b4474961/toolchain/cc_toolchain_config.bzl.tpl#L31-L35

keith:
I think this is good to go, wdyt?

thii:
It's what returned by `cc_toolchain.target_gnu_system_name`.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
         return ["-D"]
     return []
 
+def _target_os_version(ctx):
+    platform_type = ctx.fragments.apple.single_arch_platform.platform_type
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    return xcode_config.minimum_os_for_platform_type(platform_type)
```
keith, line 61: can this actually be known at the time this rule is configured? I assume the value we want here is actually more like the one you had before, which was replaced at compile time with the one for the current build

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
...
+
 def _impl(ctx):
+    target_os_version = _target_os_version(ctx)
```
keith, line 64: I guess can you split out the original change you had here around `s/arch/target` and the potentially replacing version_min? the latter seems harder

tools/osx/crosstool/cc_toolchain_config.bzl
thii: This is what was used by `xcode_swift_toolchain` before, I think this should return the right one.

Is that possible to expand cc toolchain variable like `%{version_min}` here? I tried but the `%{version_min}` string is passed verbatim if you pass it to a rule attribute.

tools/osx/crosstool/cc_toolchain_config.bzl
thii: OK I'll keep this change to just replacing arch to target.

thii: https://github.com/bazelbuild/bazel/pull/14898

keith (approved):

brentleyjones (approved):

thii:
Superseded by https://github.com/bazelbuild/bazel/pull/14898.

PR #13468, opened 2021-05-12
Use provided stub_shebang when building with --build_python_zip

codeschlosser:
Use `stub_shebang` when building with --build_python_zip. Escapes single quotes within `stub_shebang` before passing it to `echo`

[#8685](https://github.com/bazelbuild/bazel/issues/8685)

Tests (Debian 10):
$ bazelisk test //src/test/java/com/google/devtools/build/lib/rules/python/...:all
$ bazel test //src/test/java/com/google/devtools/build/lib/rules/python/...:all


codeschlosser:
@googlebot I signed it!

codeschlosser:
Will reopen with signed user commits

PR #13481, opened 2021-05-16
Remove reference to archived project in docs.

uri-canva:
The buildbarn project at https://github.com/buildbarn/bb-event-service/ was archived over a year ago, as discussed in https://github.com/buildbarn/bb-browser/issues/21, so it's not a good example reference anymore.


aiuto:
Handing this off to @meisterT because we might want to not just delete the reference. Perhaps we should update to something better.

meisterT:
@coeuvre is the Remote Execution person.

coeuvre (approved):

Issue #12460, opened 2020-11-11
Bazel 3.7.0 can't be compiled on Alpine linux


kigero:
### Description of the problem / feature request:

Bazel 3.7.0 can't be compiled on Alpine linux; I get the following error:
```
$ wget https://github.com/bazelbuild/bazel/releases/download/3.7.0/bazel-3.7.0-dist.zip && unzip bazel-3.7.0-dist.zip
$ env EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk" bash ./compile.sh
Building Bazel from scratch......
Building Bazel with Bazel.
<snip>
    Compiling src/main/tools/linux-sandbox.cc; 0s local
src/main/tools/linux-sandbox-pid1.cc: In function 'int WaitForChild()':
src/main/tools/linux-sandbox-pid1.cc:402:23: error: 'TEMP_FAILURE_RETRY' was not declared in this scope
  402 |     const pid_t pid = TEMP_FAILURE_RETRY(wait(&status));
      |                       ^~~~~~~~~~~~~~~~~~
```

It seems that TEMP_FAILURE_RETRY is only available on GNU libc, where Alpine uses musl libc.  The last Bazel version I tried on Alpine was 3.5.0 and that was compiling/running fine.

### What operating system are you running Bazel on?

Alpine linux, 3.13.0_alpha20200917.  Running within docker using the `alpine:edge` tag.

###  Have you found anything relevant by searching the web?

1. The line that errors was added in commit 1517af14afea988083756bd33a476d1100f46cff.
1. I found a definition for TEMP_FAILURE_RETRY [on StackOverflow](https://stackoverflow.com/a/41446972/1322) that does seem to work (meaning, compilation succeeds) when added to the `linux-sandbox-pid1.cc` file, but this is not my area of expertise so I can't tell if it's actually doing what it's supposed to do, or if it just looks like it is.

philwo:
@kigero Sorry for that. I'll see that we revert this part or fix it in a way that works with non-glibcs.

foopub:
Same problem with bazel 4.0 and musl. This fix seems fairly straightforward, other projects seem to be implementing too:

https://git.sailfishos.org/mer-core/profiled/merge_requests/5/diffs

I've successfully built Bazel by adding the snippet from the above push to linux-sandbox-pid1.cc as suggested by kigero. 

[...]

PR #13483, opened 2021-05-17
Adding TEMP_FAILURE_RETRY definition for musl

foopub:
Needed to compile under musl. Taken directly from glibc's unistd.h: 
https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=posix/unistd.h;hb=HEAD

### DIFF
--- a/src/main/tools/linux-sandbox-pid1.cc
+++ b/src/main/tools/linux-sandbox-pid1.cc
@@ -49,6 +49,16 @@
 #include <linux/fs.h>
 #endif
 
+// Musl doesn't have TEMP_FAILURE_RETRY. Taken from glibc unistd.h
+#ifndef TEMP_FAILURE_RETRY
+#define TEMP_FAILURE_RETRY(expression) \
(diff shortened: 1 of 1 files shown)


foopub:
@philwo fix for https://github.com/bazelbuild/bazel/issues/12460

foopub:
@googlebot I signed it!

aiuto:
Is there an existing issue that this fixes?

philwo:
@foopub Thanks! I need to clarify something internally first, but should get an answer today, I'll follow up here asap.

philwo:
Hi @foopub, sorry for the delay. The problem is that we can't accept a patch that is copying or based on glibc's code due to licensing issues. I know it's just a small snippet, but we really want to be careful here and do the right thing. 😊 

The good news is that I found an internal source for this macro in our Google repo and can just use that one instead. I'm sorry for the hassle and hope for your understanding!

I'll tag #12460 in my commit so you'll see when it's committed!

PR #13485, opened 2021-05-17
Delete host_system_name

keith:
This field is never read


keith:
used too much still, likely a breaking change

PR #13488, opened 2021-05-18
Fix symlink creation on older Windows versions

### LINKED ISSUE #13169: Bazel fails to create symlinks on slightly older versions of Windows

When symlinks support is enabled for Windows platform with the startup option `--windows_enable_symlinks` it only works on Windows version 1703 (Creators Update, build 15063) and above. Symlink creation fails on earlier versions of Windows even though symlinks are supported since Windows Vista and the process is elevated.

While dev machines tend to be rather up to date, the build servers are often stale. In our company we use Windows Server 2016 version 1607 and Bazel build fails to create symlinks there despite the feature is fully supported in that version of Windows.

[This ](https://cs.opensource.google/bazel/bazel/+/master:src/main/native/windows/file.cc;l=448?q=CreateSymbolicLinkW&ss=bazel)is the place in the Bazel code which causes the problem. The code at fault looks like this:
```
  if (!CreateSymbolicLinkW(name.c_str(), target.c_str(),
                           SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE)) {
```
While symlinks are supported in Windows for 15 years (since Vista) that particular flag **SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE** is relatively new and recognized by Windows only starting from  version 1703 (Creators Update, build 15063). On previous versions it causes `CreateSymbolicLinkW` to fail with GetLastError() returning 0x57 ERROR_INVALID_PARAMETER.

It would be great if Bazel checks the version of the OS and only apply **SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE** when it is supported by the underlying platform.

Windows, Linux, Mac.

4.0.0

khogeland:
This patch ensures that the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag is not passed to `CreateSymbolicLinkW` when the developer mode registry key is not present and enabled on a supported version of Windows. This allows symlinks to be created when Bazel is run with elevated privileges, regardless of Windows version or developer mode.

I also removed the dummy file creation check while refactoring that code for reuse. It seemed overly complicated vs. simply checking the registry and failing during runfiles creation if we're not admin. Please let me know if there's some subtle reason it needed to be done that way.

Fixes #13169 - tested on Windows Server 2016.

### DIFF
--- a/src/main/tools/BUILD
+++ b/src/main/tools/BUILD
@@ -60,12 +60,6 @@ cc_binary(
         "//src/conditions:windows": ["build-runfiles-windows.cc"],
         "//conditions:default": ["build-runfiles.cc"],
     }),
-    linkopts = select({
-        "//src/conditions:windows": [
-            "-DEFAULTLIB:advapi32.lib",  # RegGetValueW
-        ],
-        "//conditions:default": [],
-    }),
     deps = ["//src/main/cpp/util:filesystem"] + select({
         "//src/conditions:windows": ["//src/main/native/windows:lib-file"],
         "//conditions:default": [],
--- a/site/docs/windows.md
+++ b/site/docs/windows.md
@@ -32,8 +32,7 @@ fsutil 8dot3name set 0
 
 ### Enable symlink support
 
-Some features require Bazel to create file symlink on Windows, you can allow Bazel to do that by enabling [Developer Mode](https://docs.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development) on Windows (Only works for Windows 10, version 1703 or newer).
-After enabling the Developer Mode, you should be able to use the following features:
+Some features require Bazel to be able to create file symlinks on Windows, either by enabling [Developer Mode](https://docs.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development) (on Windows 10 version 1703 or newer), or by running Bazel as an administrator. This will enable the following features:
 
 * [\-\-windows_enable_symlinks](command-line-reference.html#flag--windows_enable_symlinks)
 * [\-\-enable_runfiles](command-line-reference.html#flag--enable_runfiles)
--- a/src/main/native/windows/BUILD
+++ b/src/main/native/windows/BUILD
@@ -25,6 +25,9 @@ cc_library(
         "file.h",
         "util.h",
     ],
+    linkopts = [
+        "-DEFAULTLIB:advapi32.lib", # RegGetValueW
+    ],
     visibility = [
         "//src/main/cpp:__subpackages__",
         "//src/main/tools:__pkg__",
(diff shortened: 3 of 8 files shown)


src/main/tools/BUILD
```diff
@@ -60,12 +60,6 @@ cc_binary(
         "//src/conditions:windows": ["build-runfiles-windows.cc"],
         "//conditions:default": ["build-runfiles.cc"],
     }),
-    linkopts = select({
```
meteorcloudy, line 63: Can you move this `linkopts` to https://github.com/bazelbuild/bazel/blob/0a23bd3d294eebf26ab735a5202504bbba2d641f/src/main/native/windows/BUILD#L18 instead of deleting it? Then you can remove `#pragma comment(lib, "advapi32.lib") //RegGetValueW` in file.h.
This will make it a bit easier to figure out what system libraries are required for the cc_library.

meteorcloudy (changes requested):
Thanks! This indeed looks better than the previous check overall.

But one corner case is that: what if you are on a Windows version prior to Windows 10 Creators Update (1703), and you also have developer enabled. In this case, the implementation in this PR will try to create the symlink with the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag, which the system doesn't support. However it should be able to create the symlink if the user has admin right.

philwo:
I was going to say that Windows 1703 (even Enterprise) is long EOL since 2019, but indeed Windows Server 2016 is still supported until 2022 😳

khogeland:
Ah, alright, I misinterpreted [the Bazel docs](https://docs.bazel.build/versions/master/windows.html#enable-symlink-support) and thought those features were introduced at the same time, but that is a possible situation. Creating a dummy link (or at least crafting a call that would fail) might actually be the best way to check for support of the parameter.

src/main/tools/BUILD
khogeland: Sure, I can do it like that instead. For context, I changed it to this style because the dependency would need to be listed twice (in Bazel and in compile.sh).

khogeland:
@meteorcloudy I've added an explicit OS version check to make sure Bazel never passes the flag to an incompatible version of Windows.

src/main/tools/BUILD
meteorcloudy: I see, thanks!

meteorcloudy (approved):
Thank you so much! Getting rid of the dummy link approach is really nice!

meteorcloudy:
@khogeland Can you please sign the CLA so that I can merge this PR? 😃

khogeland:
@meteorcloudy I see the `cla: yes` label, is there something else I need to do? I rewrote my initial commit to use the email on our CCLA

PR #13489, opened 2021-05-18
Make `-c opt` dead strip code for Apple builds by default.

dmaclach:
Previously `objc_enable_binary_stripping` and `-c opt` were required to both
symbol strip and dead strip code. Symbol stripping and dead code stripping are very
different things. Dead code stripping is the elimination of dead code and should be done
in an optimized build. Unused code can be marked as "not dead" by using `exported symbols`
in the Apple build rules. Symbol stripping is the removal of symbol tables from an
executable.

`objc_enable_binary_stripping` now only controls symbol stripping.

Eventually I hope we deprecate `objc_enable_binary_stripping` and just use `strip=`.

### DIFF
--- a/src/test/shell/bazel/apple/bazel_objc_test.sh
+++ b/src/test/shell/bazel/apple/bazel_objc_test.sh
@@ -118,7 +118,8 @@ EOF
       fail "Timestamp of contents of archive file should be zero"
 }
 
-function test_strip_symbols() {
+# Verifies that unused code is dead stripped with `-c opt`.
+function test_symbols_dead_stripped_opt() {
   setup_objc_test_support
 
   rm -rf ios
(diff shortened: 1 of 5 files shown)


allevato:
cc @trybka

trybka (approved):

allevato (approved):

keith (approved):

src/test/shell/bazel/apple/bazel_objc_test.sh
```diff
@@ -159,6 +245,94 @@ EOF
...
+
+  bazel build --verbose_failures \
+      --apple_platform_type=ios \
+      --ios_sdk_version=$IOS_SDK_VERSION \
+      --compilation_mode=opt \
+      //ios:app >$TEST_log 2>&1 || fail "should pass"
+  ls bazel-out/apl-ios_x86_64-opt/bin/ios/app_lipobin \
+    || fail "should generate lipobin (stripped binary)"
+  nm bazel-out/apl-ios_x86_64-opt/bin/ios/app_lipobin | grep addOne \
+    || fail "should fail to find symbol addOne"
```
thomasvl, line 288: Is this right?  Test name seems to imply it shouldn't strip, but test names say it should?

lberki:
aye aye, pulling it in then

lberki:
Update: I have to roll this change back since it breaks targets at Google. I don't know much about Apple builds so I don't know how much work it is to fix; @allevato , if doing so is not the right course action, please do tell me!

Issue #13428, opened 2021-05-05
cquery ignores test config in .bazelrc files


illicitonion:
I expect `cquery` when querying tests to reflect flags like `--test_env` and `--test_arg`. It appears to reflect these if they're specified on the command line, but not if they're in a .bazelrc file:

Reproduction (making a clean workspace for each run, so we're definitely not seeing weird leftover state from previous runs):
```console
$ dir=$(mktemp -d); (cd "${dir}" && touch WORKSPACE && touch test.sh && chmod 0755 test.sh && echo >BUILD 'sh_test(name = "test", srcs = ["test.sh"])' && bazel cquery //:test 2>/dev/null)
//:test (86e5752)

$ dir=$(mktemp -d); (cd "${dir}" && touch WORKSPACE && touch test.sh && chmod 0755 test.sh && echo >BUILD 'sh_test(name = "test", srcs = ["test.sh"])' && bazel cquery //:test --test_arg=beep 2>/dev/null)
//:test (6a51fd2)

$ dir=$(mktemp -d); (cd "${dir}" && touch WORKSPACE && touch test.sh && chmod 0755 test.sh && echo >BUILD 'sh_test(name = "test", srcs = ["test.sh"])' && echo 'test --test_arg=beep' > .bazelrc && bazel cquery //:test 2>/dev/null)
//:test (86e5752)
```

`cquery` _does_ appear to correctly pay attention to `.bazelrc` files for things that would affect the _build_ actions that produce the tests, but appears to be ignored for things that would affect the test actions.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

`release 4.0.0`

/cc @gregestren

gregestren:
@twigg has input on this.

illicitonion:
@sdtwigg I'd love to hear your thoughts, if you wouldn't mind sharing!

sdtwigg:
This is roughly working as expected. The cquery command inherits from the build command so it will only see bazelrc entries with `build` or `cquery` (although I am not familiar with anyone using the latter as a matter of course).

In practice, we generally advise being careful with triggering off `test` in bazelrc and instead prefer `build` because, depending on affected options, it could result in a redoing lot of analysis phase work if users swap between commands.

illicitonion:
> In practice, we generally advise being careful with triggering off `test` in bazelrc and instead prefer `build` because, depending on affected options, it could result in a redoing lot of analysis phase work if users swap between commands.

[...]

PR #13491, opened 2021-05-19
cquery inherits from `test` not `build`

illicitonion:
This makes directives like `test --test_arg=foo` present in `.bazelrc` files be
factored into the configuration hash for test targets.

See https://github.com/bazelbuild/bazel/issues/13428 for extensive
context.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/CqueryCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/CqueryCommand.java
@@ -48,7 +48,12 @@ import java.util.Set;
 @Command(
   name = "cquery",
   builds = true,
-  inherits = {BuildCommand.class},
+  // We inherit from TestCommand so that we pick up changes like `test --test_arg=foo` in .bazelrc
+  // files.
+  // Without doing this, there is no easy way to use the output of cquery to determine whether a
+  // test has changed between two invocations, because the testrunner action is not easily
+  // introspectable.
+  inherits = {TestCommand.class},
   options = {CqueryOptions.class},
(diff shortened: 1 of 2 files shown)


gregestren:
Just want to confirm, after https://github.com/bazelbuild/bazel/issues/13428#issuecomment-860629715, that this is the route you want to go.

Also a nitpick on the description: this means cquery picks up .bazelrc directives of the form:

    test --some_flag

right?

For example, I expect cquery already picks up

    build --test_arg=foo

but not

    test --cpu=bar

src/main/java/com/google/devtools/build/lib/runtime/commands/CqueryCommand.java
```diff
@@ -48,7 +48,7 @@
 @Command(
   name = "cquery",
   builds = true,
-  inherits = {BuildCommand.class},
+  inherits = {TestCommand.class},
```
gregestren, line 51: Quick comment rationale for why this value?

gregestren (commented):
Can we add a quick test to block regression? https://github.com/bazelbuild/bazel/blob/080ef6eba08e8e5329c7135d39378d5841ee417f/src/test/shell/integration/configured_query_test.sh looks like the best place to me (I can't find any dedicated `bazelrc` tests).

src/main/java/com/google/devtools/build/lib/runtime/commands/CqueryCommand.java
illicitonion: Done, and added a test which failed before this PR - PTAL!

gregestren (approved):

Issue #13429, opened 2021-05-06
[Coverage] Provide option to generate exec coverage data under bazel-out


ThomasCJY:
### feature request
Currently, if we run `bazel coverage` on a java repository. There is only 
```
bazel-out/_coverage/_coverage_report.dat
bazel-out/_coverage/lcov_files.tmp
```
being generated. It would be nice if bazel can provide the option to put the original exec coverage data file under this output directory as well.

### Feature requests: what underlying problem are you trying to solve with this feature?
Provide more options of using bazel coverage.

ThomasCJY:
looks like --experimental_fetch_all_coverage_outputs works and will generate coverage data under bazel-testlogs

PR #13498, opened 2021-05-20
[Coverage] Add option to output Jacoco coverage binary data

ThomasCJY:
**Issue**
https://github.com/bazelbuild/bazel/issues/13429
This is my proposed implementation of adding option to output Jacoco coverage binary data. Although bazel coverage is already generating the lcov style report for java code, it will be helpful if we can have an option to output the origin binary coverage output from jacoco.

Let me know you feedback on it so I can update and fix the unit test code.


ThomasCJY:
looks like `--experimental_fetch_all_coverage_outputs` works and will generate coverage data under `bazel-testlogs`. closing pr

PR #13506, opened 2021-05-21
Add --no-log-init flag to docker sandbox.

psigen:
This adds the `--no-log-init` flag (`-l`) to the internal `useradd` command used to initial the docker sandbox environment.

Without this flag, AD/LDAP/SSSD users that have large UID/GID values will be added to `lastlog`/`faillog`, but since docker does not support sparse files, this will cause the docker daemon to attempt to create a `/var/lib/docker/overlay2` entry that may consume all available disk space.

https://github.com/moby/moby/issues/5419#issuecomment-332785867

For one example, my SSSD-assigned uid is `1553201121`, which makes the _sparse_ size of my `lastlog` file 423GB.  If this uid is used by bazel's docker-sandbox, the resulting container attempts to create the full 423GB file, which I confirmed the hard way.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/sandbox/DockerSandboxedSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/DockerSandboxedSpawnRunner.java
@@ -344,7 +344,7 @@ final class DockerSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
               if (uid > 0) {
                 dockerfile.append(
                     String.format(
-                        "RUN [\"useradd\", \"-m\", \"-g\", \"%d\", \"-d\", \"%s\", \"-N\", \"-u\", "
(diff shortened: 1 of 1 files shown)


susinmotion:
@larsrc-google

larsrc-google (approved):
I guess `lastlog` has this format from a time when UIDs were never bigger than 65K or so and somebody thought that extremely fast lookup was more important than disk space. Huh.

psigen:
Well, it's just a bit of a semantics mismatch:
- having a large _sparse_ file doesn't cost anything, so on a baremetal system these lastlogs work great
- docker doesn't support sparse files yet, and it's generally been ok because they aren't used _that_ widely.. except for this.

In any case, thanks for taking a look.  I'm really looking forward to not accidentally locking up my machine :grin:

Issue #12200, opened 2020-09-30
rename 'master' branch to 'main'


aiuto:
Because we should

philwo:
Do you have experience with and guidance on how to do this without breaking all the tools and scripts that currently have the old branch name hardcoded? I acknowledge the importance of this, but I'm a bit at a loss on how to do this technically without breaking everything and everyone.

As a rough estimate of the work involved:

grep over our CI repo:

<img width="1179" alt="Screen Shot 2020-09-30 at 23 18 16" src="https://user-images.githubusercontent.com/504652/94740597-3b8ab300-0373-11eb-8dfd-7917cb037120.png">

Plus the Google-internal stuff, the config in the Buildkite pipeline settings, the gitsync container, the Copybara setup, the Gerrit configuration / ACLs, pending pull requests, ...

This has to be very, very well coordinated.

meisterT:
See https://github.com/github/renaming for some guidance, specifically the mention that this will be more seamless later this year.

aiuto:
I had seen that page before, but not the "seamless" renaming coming later.  I did this with rules_pkg a month ago and learned some pain points
- old PRs are against master.  If you don't remember to rebase them off main, they go to the wrong branch.
- copybara pulls from 'master' by default, not from the default branch. Easy to fix, but the copy.bara.sky files all need update
- CI seems to work without change.

UebelAndre:
Hey, would someone be able to elaborate on:

> Because we should

?

meisterT:
> Hey, would someone be able to elaborate on:
> 
> > Because we should

The term `master` is offensive to some people. There's nothing special about the name in how git treats branches it was just historically the name for the first branch.

philwo:
Hey @UebelAndre :)

Terms like "master / slave" or "whitelist / blacklist" are considered as offensive by many people (e.g. reminding them of slavery or racism). Obviously the *intent* of using these terms was not to provoke these negative feelings - they just somehow happened to become standard terms in the IT, but if there's a good neutral term available to replace them, I think it's just kind to consider doing so. Thus, there has been a recent movement to find better words for what we try to say! Here's some background: https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/.

Often the recommended replacements actually convey the intended meaning better than the historically used words - some examples:
- It's the `main` branch, because it's our... wait for it... main branch, where the development happens and release branches are cut from. :)
[...]

PR #13519, opened 2021-05-26
Serve docs@head under /versions/main/

aiuto:
See  #12200

Another good choice would be to use 'head' instead of main.
References to /master/ are redirected to /main/

Next step:
- wait for deploy and stability
- Change /master/ => /main/ in all the checked in docs.
- gradually do the same in other repositories
- remove the temporary redirection or move hosting to a site that handles regex redirects

### DIFF
--- a/site/BUILD
+++ b/site/BUILD
@@ -167,15 +167,15 @@ jekyll_build(
     srcs = [
         "@jekyll_tree_%s//file" % DOC_VERSION["version"].split("-")[0].replace(".", "_")
         for DOC_VERSION in DOC_VERSIONS
-    ] + [":jekyll-tree"],  # jekyll-tree *must* come last to use layouts and includes from master
+    ] + [":jekyll-tree"],  # jekyll-tree *must* come last to use layouts and includes at head
     bucket = "docs.bazel.build",
 )
 
-# A smaller site build target. Contains just the docs for latest released version and master.
+# A smaller site build target. Contains just the docs for latest released version and head.
 jekyll_build(
     name = "site_latest",
     srcs = [
         "@jekyll_tree_%s//file" % DOC_VERSIONS[0]["version"].split("-")[0].replace(".", "_"),
-        ":jekyll-tree",  # jekyll-tree *must* come last to use layouts and includes from master
+        ":jekyll-tree",  # jekyll-tree *must* come last to use layouts and includes at head
     ],
 )
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -20,7 +20,7 @@ version: "4.1.0"
 
 # This must be kept in sync with //scripts/docs:doc_versions.bzl
 doc_versions:
-  - master
+  - main
   - 4.1.0
   - 4.0.0
   - 3.7.0
--- a/site/jekyll-tree.sh
+++ b/site/jekyll-tree.sh
@@ -45,7 +45,7 @@ readonly TMP=$(mktemp -d "${TMPDIR:-/tmp}/tmp.XXXXXXXX")
 readonly OUT_DIR="$TMP/out"
 trap "rm -rf ${TMP}" EXIT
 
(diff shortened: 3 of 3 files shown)


gregestren:
Interesting. How long might the temporary redirection last? I could imagine a lot of `master` link references scattered across the ecosystem.

aiuto:
We can pick an arbitrary date in the future, but I think at 5.0 or 6.0
would be a reasonable time frame. Doing this today allows us to change the
embedded links in docs asynchronously.

aiuto:
Friendly ping.

philwo:
I would recommend to use https://cloud.google.com/load-balancing/docs/https/traffic-management to implement the redirects, that way it's generic instead of requiring the placeholder files and uses HTTP 301 Moved Permanently instead of a client-side JavaScript or `<meta>` redirect.

Would it be possible to temporarily publish both `main` and `master` in parallel, then we implement the redirect on the infra side and as soon as that's working, we can remove the `master` variant from the config?

aiuto:
On Mon, Jun 7, 2021 at 7:43 AM Philipp Wollermann ***@***.***>
wrote:

> I would recommend to use
> https://cloud.google.com/load-balancing/docs/https/traffic-management to
> implement the redirects, that way it's generic instead of requiring the
> placeholder files and uses HTTP 301 Moved Permanently instead of a
> client-side JavaScript or <meta> redirect.
>
That may be the best solution in the long run, but I would rather we do
that once, after we start serving docs from devsite.

> Would it be possible to temporarily publish both main and master in
> parallel, then we implement the redirect on the infra side and as soon as
> that's working, we can remove the master variant from the config?
>
My intent was to stop displaying the url with /master/ in it as soon as
possible. A redirect is more in line with that than serving under both.

philwo:
> That may be the best solution in the long run, but I would rather we do that once, after we start serving docs from devsite.

Setting up an HTTP redirect is *one click* in the cloud console. Here, I made one as a proof of concept: https://docs.bazel.build/versions/proof-of-concept/be/overview.html

It redirects to https://docs.bazel.build/versions/master/be/overview.html.

Creating a redirect from /versions/master/* to /versions/main/* is the same process.

HTTP based redirects will correctly be handled by all search engines, browsers, command-line tools as opposed to JavaScript/HTML-based redirects.

> My intent was to stop displaying the url with /master/ in it as soon as possible. A redirect is more in line with that than serving under both.

What about we prepare a change that just switches over from "master" to "main" and I submit the CL and make the load balancer change in parallel? Can be done tomorrow if you want.

philwo:
> If you know the process, that's great.

I wasn't actually sure whether Google Cloud Load Balancer could do this now (sometimes there are hidden caveats..), but it seems to be working fine with that proof-of-concept URL. Happy to set this up for master -> main.

> Just to make sure I understand. The CL will be against site, but move to
> main, without the redirect, and we will do that roughtly the same time the
> load balancer redirect is added. Pages might be broken for a few minutes
> during the push/redirect overlap, but then it will settle out. Right?

Exactly! 😊

Actually, here's a slightly tweaked even better idea: I misunderstood a part of this PR in the beginning regarding how our docs structure works, but now that I re-read it, I think we can do this:

1. Import and submit this PR here as is, which will move the content over to `main` and place the client-side redirector HTML files into the `master` directory.
2. Then I can setup the HTTP redirect on the Load Balancer, at which point the HTTP redirect takes precedence over the HTML files in the directory.
3. When we verified that everything is working fine, we can just remove the `function gen_master_redirects` part in a follow-up.

That way we would have zero downtime. WDYT?

If you import this and send it to me for review in Critique, I can LGTM it and then do the infrastructure part afterwards!

philwo (approved):

aiuto:
> That way we would have zero downtime. WDYT?

Yes. That was my plan.  I will send the import CL.

aiuto:
Thanks.

PR #13529, opened 2021-05-27
Reference osx_archs.bzl from bazel_tools

keith:
There doesn't seem to be a reason to symlink before referencing it. It
was brought up that referencing a local label in this toolchain
configuration is a bit strange when using remote execution.

https://bazelbuild.slack.com/archives/CA31HN1T3/p1622142347351500

### DIFF
--- a/tools/cpp/cc_configure.bzl
+++ b/tools/cpp/cc_configure.bzl
@@ -39,7 +39,6 @@ def cc_autoconf_toolchains_impl(repository_ctx):
     paths = resolve_labels(repository_ctx, [
         "@bazel_tools//tools/cpp:BUILD.toolchains.tpl",
         "@bazel_tools//tools/osx/crosstool:BUILD.toolchains",
-        "@bazel_tools//tools/osx/crosstool:osx_archs.bzl",
         "@bazel_tools//tools/osx:xcode_locator.m",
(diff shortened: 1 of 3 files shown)


keith:
cc @oquenchil

aiuto:
It does seem to keep bouncing around.

Since you caught me on triage duty week, I'll try to help. 
- I imported the PR to create a change against our internal code
- That seems to have passed internal presubmit
- I'll send out the change to various apple rules people internally to see if they have problems.
- We will have to do some testing against our blaze on mac version, which has some special magic around xcode selection.

If someone who knows what they are talking about, I can merge it.

aiuto:
@hlopko  You wrote the lines being changed. Does this seem reasonable to you?

aiuto (approved):
It passes CI. Several people at Google believe this won't break our rules.
I'll start the merge now.

keith:
Thanks!

keith:
@Wyverald can you include this one in the next 5.x?

jfirebaugh:
Hmm, what about:
https://github.com/bazelbuild/bazel/blob/46e0be4ef67901cab7896348c89de06396f737b4/tools/osx/crosstool/BUILD.toolchains#L5

PR #13532, opened 2021-05-28
Update user-manual.html

Delwin9999:
The double dash is in the wrong place in the example.

### DIFF
--- a/site/docs/user-manual.html
+++ b/site/docs/user-manual.html
@@ -2199,7 +2199,7 @@ Examples:
   it is used to build <em>and run</em> a single target. Here is a typical session:
 </p>
 <pre>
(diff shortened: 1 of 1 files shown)


Delwin9999:
@googlebot I signed it!

gregestren:
Hi @Delwin9999 . Don't both forms work? Do you think the second is clearer?

Delwin9999:
> Hi @Delwin9999 . Don't both forms work? Do you think the second is clearer?

I tried and both work, it's just the internal examples I've been working from have the dashes after the target.

That seems to make more sense to me since it clearly separates the bazel section from the non-bazel section.

gregestren:
That's fair. I agree.

gregestren (approved):

PR #13547, opened 2021-06-02
Remove unneeded features from @tools/build_defs/pkg:pkg_tar.

aiuto:
@tools/build_defs/pkg:pkg_tar is deprecated and replaced by @rules_pkg.
This PR pushes that concept forward by elminating features that are not used by Bazel.
By the next LTS track (5.x) I think we can reduce this pkg_tar into something
that serves as the tiny tar utility.

cc/ @nacl

### DIFF
--- a/tools/build_defs/pkg/BUILD
+++ b/tools/build_defs/pkg/BUILD
@@ -25,7 +25,7 @@ exports_files(
 py_library(
     name = "archive",
     srcs = ["archive.py"],
-    srcs_version = "PY2AND3",
+    srcs_version = "PY3",
     visibility = ["//visibility:public"],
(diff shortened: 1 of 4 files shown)


lberki (approved):

lberki:
...it looks like this is breaking internal presubmits, though?

aiuto:
I'll see. That is most likely a Blaze project peeking into Bazel source.
Latent tech debt.

lberki:
I didn't check very thoroughly, only noticed the red blobs on the UI

aiuto:
It was because of a BUILD vs. BUILD.oss conflict.

Issue #13531, opened 2021-05-28
Cache hit for remote execution shows Scheduling state while downloading


brentleyjones:
### Description of the problem / feature request:

When an action is a cache hit, instead of showing some sort of running state in the UI, it shows as `[Sched]` the entire time. This is confusing, as it seems the action is stalled out waiting for an executor, when in reality it's making progress.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

While using remote execution, build, from clean, a fully cached action graph. All of the actions will look like this:
```console
    [Sched] //Modules/Illustration:IllustrationResources; 113s
    [Sched] Bundling IconographyResources; 42s
    [Sched] Generating derived files for Swift module Kronos; 12s
    [Sched] Generating derived files for Swift module CollectionDiffing; 12s
    [Sched] Generating derived files for Swift module SwiftProtobuf; 12s
    [Sched] Compiling Swift module CollectionDiffing; 12s
    [Sched] Generating derived files for Swift module Yams; 12s
    [Sched] Compiling Swift module Yams; 12s ..
```

When not using `remote`, it looks like this:
```console
    @com_google_protobuf_protoc_macos_x86_64//:protoc; 15s remote-cache
    Compiling src/google/protobuf/struct.pb.cc [for host]; 14s remote-cache
    //Modules/UserProfilesUI:UserProfilesUIResources; 14s remote-cache
    //Modules/DigitCodeUI:DigitCodeUIResources; 14s remote-cache
    //Modules/RidePasses:RidePassesResources; 14s remote-cache
    .../LastMileCodeEntryUI:LastMileCodeEntryUIResources; 14s remote-cache
    .../AutonomousInRideUI:AutonomousInRideUIResources; 14s remote-cache
    //Modules/NetworkErrorUI:NetworkErrorUIResources; 14s remote-cache
```

### What operating system are you running Bazel on?

macOS 12.3.1

### What's the output of `bazel info release`?

4.0.0

###  Have you found anything relevant by searching the web?

Here the scheduling state is set:
https://github.com/bazelbuild/bazel/blob/6b33bdb1e22514304c0e35ce8e067f2175685245/src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java#L251

[...]

PR #13555, opened 2021-06-04
Remote: Report checking cache status before the action is scheduled to run remotely.

coeuvre:
Add a Caching state to ActionState. This state indicates the action is checking the cache and should be happened before Scheduling state.

Change ProgressStatus from enum to interface so that we can pass more data to the event handler (which is required to report upload/download details later).

Fixes https://github.com/bazelbuild/bazel/issues/13531.


src/main/java/com/google/devtools/build/lib/runtime/UiStateTracker.java
```diff
@@ -193,6 +194,13 @@ String formatNames(int bitmap) {
      */
     boolean scanning;
 
+    /**
+     * Bitmap of strategies that are checking the cache of this action.
+     *
+     * <p>If non-zero, implies that {@link #scanning} is false.
+     */
+    int cachingStrategiesBitmap = 0;
```
philwo, line 202: Can we use an https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/EnumSet.html here instead of manually handling a bitmap?

I see that `schedulingStrategiesBitmap` and `runningStrategiesBitmap ` were already there before, but maybe it's an opportunity to make the code more readable 😊

philwo (approved):

src/main/java/com/google/devtools/build/lib/runtime/UiStateTracker.java
coeuvre: As discussed, since the bit is generated dynamically, we will go with how the existing code does.

PR #13557, opened 2021-06-07
Remote: Display download progress when actions are downloading outputs from remote cache.

coeuvre:
Normally, when executing action with remote cache/execution, the UI only display the "remote"/"remote-cache" strategy:
```
[500 / 1000] 500 actions, 3 running
    [Sched] Executing genrule //:test-1;
    Executing genrule //:test-2; 2s remote
    Executing genrule //:test-3; 3s remote ...
```

However, it doesn't tell users what is happening under the hood. #13555 fixed the confusion which the UI display the action is scheduling while it is actually downloading the outputs.

With this change, Bazel will display the downloads if action is downloading outputs. e.g.
```
[500 / 1000] 500 actions, 3 running
    [Sched] Executing genrule //:test-1; 1s remote
    Executing genrule //:test-2; Downloading 2.out, 20.1 KiB / 100 KiB; 2s remote        
    Executing genrule //:test-3; 3s remote ...
``` 

Add a generic `ActionProgressEvent` which can be reported within action execution to display detailed execution progress for that action.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ActionProgressEvent.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ActionProgressEvent.java
@@ -1,0 +1,42 @@
+package com.google.devtools.build.lib.actions;
+
+import com.google.devtools.build.lib.events.ExtendedEventHandler.ProgressLike;
+
+/**
+ * Notifies that an in-flight action is making progress..
+ */
+public class ActionProgressEvent implements ProgressLike {
+  private final ActionExecutionMetadata action;
+  private final String progressId;
+  private final String progress;
+  private final boolean isFinished;
+
+  public ActionProgressEvent(ActionExecutionMetadata action, String progressId, String progress, boolean isFinished) {
+    this.action = action;
+    this.progressId = progressId;
+    this.progress = progress;
+    this.isFinished = isFinished;
+  }
+
+  /** Gets the metadata associated with the action being scheduled. */
+  public ActionExecutionMetadata getActionMetadata() {
+    return action;
+  }
+
+  /**
+   * The id that uniquely determines the progress among all progress events within an action.
+   */
+  public String getProgressId() {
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/devtools/build/lib/actions/ActionProgressEvent.java
```diff
@@ -0,0 +1,42 @@
+package com.google.devtools.build.lib.actions;
+
+import com.google.devtools.build.lib.events.ExtendedEventHandler.ProgressLike;
+
+/**
+ * Notifies that an in-flight action is making progress..
```
philwo, line 6: nit: remove double `.` at the end

src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
```diff
@@ -332,17 +330,7 @@ public void report(ProgressStatus state, String name) {
 
       // TODO(ulfjack): We should report more details to the UI.
```
philwo, line 331: Do you think we can remove this TODO now? :)

philwo (approved):
In general LGTM, very nice UX improvement! Could you split this up into smaller CLs when you send it for internal review please? 😊

coeuvre:
Thanks! Since this PR is based on #13555, can you also give it a look?

src/main/java/com/google/devtools/build/lib/exec/AbstractSpawnStrategy.java
coeuvre: Yes!

src/main/java/com/google/devtools/build/lib/actions/ActionProgressEvent.java
coeuvre: Done.

PR #13568, opened 2021-06-10
Remote: Add --experimental_capture_corrupted_outputs flag.

coeuvre:
Which when set, Bazel will save outputs whose digest does not match the expected value to the target directories.

Also use OutputDigestMismatchException to indicate the error and include output path in the error message. The message for such an error will become e.g.:

```
com.google.devtools.build.lib.remote.common.OutputDigestMismatchException: Output bazel-out/darwin-fastbuild/bin/output.txt download failed: Expected digest '872af2fe77729717832d0a020ae87a93b8b944146a2af6b3490491e1eaf1dc74/29229' does not match received digest '872af2fe77729717832d0a020ae87a93b8b944146a2af6b3490491e1eaf1dc74/29229'.
```

Used to support https://github.com/bazelbuild/continuous-integration/issues/1175.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
@@ -377,9 +377,14 @@ public class RemoteCache implements AutoCloseable {
                 Path localPath = ((OutputDigestMismatchException) suppressed).getLocalPath();
                 Path dst = captureCorruptedOutputsDir.getRelative(outputPath);
                 dst.createDirectoryAndParents();
+
+                // Make sure dst is still under captureCorruptedOutputsDir, otherwise IllegalArgumentException will be
+                // thrown.
+                dst.relativeTo(captureCorruptedOutputsDir);
+
                 FileSystemUtils.copyFile(localPath, dst);
-              } catch (IOException ioEx) {
-                suppressed.addSuppressed(ioEx);
+              } catch (Exception ee) {
+                ee.addSuppressed(ee);
               }
             }
           }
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
```diff
@@ -355,6 +366,25 @@ public void download(
...
     } catch (Exception e) {
+      if (captureCorruptedOutputsDir != null) {
+        if (e instanceof BulkTransferException) {
+          for (Throwable suppressed : e.getSuppressed()) {
+            if (suppressed instanceof OutputDigestMismatchException) {
+              // Capture corrupted outputs
+              try {
+                String outputPath = ((OutputDigestMismatchException) suppressed).getOutputPath();
+                Path localPath = ((OutputDigestMismatchException) suppressed).getLocalPath();
+                Path dst = captureCorruptedOutputsDir.getRelative(outputPath);
```
philwo, line 378: Is it guaranteed (even with broken or malicious backends) that the output path will never contain `..` references or start with a `/`? Should we check that the resulting path `dst` is still below the `captureCorruptedOutputsDir` just to make sure? 🤔

philwo (approved):

src/main/java/com/google/devtools/build/lib/remote/RemoteCache.java
coeuvre: We are safe here since `outputPath` is what we obtained from skyframe and send to remote execution backend (something like `bazel-out/darwin-fastbuild/bin/...`).

It is a path relative to the working directory (or input root depends on the flags) when the remote backend executing the action.

coeuvre: However, I added the check in case the invariance changed in the future. Thanks for the catch!

Issue #13435, opened 2021-05-06
file handle exhaustion when fetching from remote cache


jablin:
### Description of the problem

When extensively using a remote cache (i.e. 100% of the build results of a large project), `bazel build` (4.0.0) hits "too many file handles" if you use an empty `--disk_cache` at the same time.
This is reproducible. On an 8 core CPU (+ hyperthreading) it usually happens after ~4k targets.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

* setup a project with tens of thousands of targets.
* setup a remote cache
```
# upload all build artifacts to the remote cache
bazel clean --expunge
bazel build --remote_cache=http://<host> --remote_upload_local_results=true //... 

# trigger as many downloads as possible
bazel clean --expunge
rm -fr /tmp/disk-cache
mkdir /tmp/disk-cache
bazel build --remote_cache=http://<host> //... --disk_cache=disk-cache
```
This eventually leads to:
```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'ActionLookupData{actionLookupKey=ConfiguredTargetKey{[...]
        at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
        at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
Caused by: io.netty.channel.ChannelException: Unable to create Channel from class class io.netty.channel.socket.nio.NioSocketChannel
        at io.netty.channel.ReflectiveChannelFactory.newChannel(ReflectiveChannelFactory.java:46)
        at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:310)
        at io.netty.bootstrap.Bootstrap.doResolveAndConnect(Bootstrap.java:155)
        at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:116)
        at io.netty.channel.pool.SimpleChannelPool.connectChannel(SimpleChannelPool.java:265)
[...]

PR #13574, opened 2021-06-14
when writing to local disk cache, open files later in order to avoid "too many open files"

jablin:
Re-use the existing "LazyFileOutputStream" in DiskAndRemoteCacheClient.java
in order to avoid "Too many open files".

Resolves https://github.com/bazelbuild/bazel/issues/13435


coeuvre (approved):
Thanks!

PR #13575, opened 2021-06-14
Fix building with GCC 11 on ARM v8

samcom12:
Getting error 

`error: 'numeric_limits' is not a member of 'std'`

Added fix for same.


philwo:
Thank you very much for the report and proposed fix @samcom12. As this is just an added #include directive, I have made the change directly and sent it to a colleague for review, this PR should be auto-closed as soon as it's submitted.

PR #13583, opened 2021-06-17
Use True instead of 1 to be more expressive

limdor:
linkstatic conceptually can be True or False, using 1 can give the impression that any arbitrary value is allowed.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLibrary.java
@@ -254,7 +254,7 @@ public abstract class CcLibrary implements RuleConfiguredTargetFactory {
 
     if (common.getLinkopts().contains("-static")) {
       ruleContext.attributeWarning("linkopts", "Using '-static' here won't work. "
-                                   + "Did you mean to use 'linkstatic=1' instead?");
+                                   + "Did you mean to use 'linkstatic=True' instead?");
     }
 
     linkingHelper.setShouldCreateDynamicLibrary(createDynamicLibrary);
@@ -526,7 +526,7 @@ public abstract class CcLibrary implements RuleConfiguredTargetFactory {
         && ccCompilationOutputs.getObjectFiles(true).isEmpty()) {
       if (!linkstaticAttribute && appearsToHaveObjectFiles(ruleContext.attributes())) {
         ruleContext.attributeWarning("linkstatic",
-            "setting 'linkstatic=1' is recommended if there are no object files");
+            "setting 'linkstatic=True' is recommended if there are no object files");
       }
     } else {
       if (!linkstaticAttribute && !appearsToHaveObjectFiles(ruleContext.attributes())) {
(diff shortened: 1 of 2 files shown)


comius (approved):

comius (changes requested):
Would you mind also fixing testCcLibraryWithDashStatic, which compares against this message?

limdor:
@comius I just fixed the test. When doing the changes I saw several other places that use 1 instead of True, I will create a different PR for the other cases. I think that this was the most relevant because the consequence is that the users use 1 instead of True.

oquenchil:
Did you discuss this change with anybody from the Bazel team? 1=True and 0=False is super clear, making it more verbose does not make it better in my opinion.

limdor:
> Did you discuss this change with anybody from the Bazel team? 1=True and 0=False is super clear, making it more verbose does not make it better in my opinion.

No I did not, I was planning to discuss it in the PR. If there is a better place to discuss it please let me know.
In our company we advice people to use True and False instead of 1 and 0, for the same reason that booleans were introduced in several languages. It makes it more clear that this is a Boolean value and not an integer, and that for example 2 is not a value that should be used.

oquenchil:
I think that's certainly a valid argument for code. For configuration the argument may have merits too. In any case, if the Bazel team decided to encourage that we would disallow the integers in flags altogether to avoid inconsistencies. I don't think this is a reasonable effort right now.

Thanks for bringing this up though.

laurentlb:
@oquenchil Please see https://docs.bazel.build/versions/main/skylark/build-style.html#other-conventions (as limdor pointed out in the new issue)

oquenchil:
I think my point still stands that these PRs don't make sense if there isn't a major effort to completely forbid integers.

laurentlb:
cc @brandjon

We agreed on a preferred style a few years ago (because consistency is important). Some work was already done in this area: https://github.com/bazelbuild/bazel/commit/9b8e9feaec3568ccb76787c3f0df5d80cc4a894f

I think it makes sense to fix style issues. Error messages that contradict the recommendations should be updated. The documentation looks inconsistent to me: https://docs.bazel.build/versions/main/be/c-cpp.html (search for `linkstatic`, see how it mixes `1` and `= True` in the same page).

I see no reason to reject this PR.

oquenchil:
If there is a preferred style and we actually control what is allowed in the code base, there should be no time wasted fixing this everywhere before enforcing that through code. Otherwise it can easily become inconsistent again at any place and wastes more time.

comius:
Those are changes in the error messages we print out, and they should follow the preferred style, while I doubt there is ever going to be an automated check for this in our error messages, that would enforce it (and I have doubts it is even implementable).

I don't believe the error messages will regress, if we don't enforce checks on them. And I don't think this change should be considered together with any consistency checks in the actual BUILD or .bzl defined targets.

The change in error message, signposts users to use True/False instead of 1/0 - a this results in any fixes made by users to have the proper value - so quite the opposite - it works to make the new code consistent.

As for enforcing the use of True in Starlark, I guess this ship has sailed, by making the design decision 1 == True. This is in my opinion impossible to revert now. The decision I guess follows Python (I'd be disappointed if it's only C relic). Perhaps it's possible to enforce it on some literals that are directly set on the attributes - but I don't imagine anything nice would come out of it.

There is another worry from our side, though, A lot of changes like this pile up and take our time to review them. The time to import a pr is negligible, but it adds up.

What would be helpful is making all such similar+trivial changes in a single PR. If not @oquenchil, I'd be happy to review such PR and take care it is merged.

PR #13584, opened 2021-06-17
Use True/False instead of 1/0

limdor:
This is the second part of https://github.com/bazelbuild/bazel/pull/13583. I'm doing it in a second PR to see the tests because I'm not sure that this part is so straight forward.


limdor:
Considering that https://github.com/bazelbuild/bazel/pull/13583 has been rejected, does not make sense to keep this PR.

Issue #8319, opened 2019-05-14
adding header to cc_library causes all source files to be rebuilt


crorvick:
### Description of the problem / feature request:

Adding a header to a `cc_library` causes a full rebuild of that library, even if that header is added to `srcs` and no source files include that header yet.

### Feature requests: what underlying problem are you trying to solve with this feature?

Do not rebuild source files that do not depend on the new header.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
$ git clone https://github.com/crorvick/bazel-hello.git
Cloning into 'bazel-hello'...
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 38 (delta 0), reused 4 (delta 0), pack-reused 32
Unpacking objects: 100% (38/38), done.
$ cd bazel-hello
$ git co 60dfc8f25eee
HEAD is now at 60dfc8f use library for printing
$ bazel build //...
Starting local Bazel server and connecting to it...
INFO: Analysed 5 targets (22 packages loaded, 471 targets configured).
INFO: Found 5 targets...
INFO: Elapsed time: 2.425s, Critical Path: 0.02s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
$ git checkout 494f76d571da
HEAD is now at 494f76d add unused private header
$ bazel build -s //...
INFO: Analysed 5 targets (1 packages loaded, 6 targets configured).
INFO: Found 5 targets...
SUBCOMMAND: # //src/util:print [action 'Compiling src/util/print.cc']
(cd /home/crorvick/.cache/bazel/_bazel_crorvick/69d6d6c4fc40b0136e835715b0adc69f/execroot/__main__ && \
  exec env - \
    PATH=/home/crorvick/.local/bin:/home/crorvick/bin:/usr/local/bin:/bin:/usr/bin \
    PWD=/proc/self/cwd \
[...]

PR #13592, opened 2021-06-21
CppCompileAction: do not make declaredIncludeSrcs() part of the action key

obruns:
There are two types of input discovery in Bazel:

  * include scanning is hard-disabled [1]
  * depfile parsing is active and the action key is updated accordingly

However, due to declaredIncludeSrcs() being part of the action key, a
change of the declared include srcs, either explicitly or via a glob(),
causes a CppCompileAction to be re-executed even though none of its
inputs (as reported by the compiler) had changed.

When an action does not discover inputs, there is no need to distinguish
between inputs and declared inputs since the inputs already go into the
cache key.
When an action does discover inputs, the action key gets updated after
parsing the depfile.

With this change, Bazel behaves much like Ninja in terms of avoiding
unnecessary recompiles of C++ translation units.

There should probably be an entry in the release notes because this
change makes existing cache entries non-candidates.

[1] https://github.com/bazelbuild/bazel/commit/1df4c71a9cf52f2dd24c4a10da75bb88631e7fde

Fixes: #8319
Hint-by: Lukacs T. Berki <lberki@google.com>

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppCompileAction.java
@@ -1368,8 +1368,6 @@ public class CppCompileAction extends AbstractAction implements IncludeScannable
     fp.addBytes(commandLineKey);
     fp.addBoolean(validateTopLevelHeaderInclusions);
 
-    actionKeyContext.addNestedSetToFingerprint(fp, declaredIncludeSrcs);
-    fp.addInt(0); // mark the boundary between input types
     actionKeyContext.addNestedSetToFingerprint(fp, mandatoryInputs);
     fp.addInt(0);
     actionKeyContext.addNestedSetToFingerprint(fp, prunableHeaders);
--- a/src/test/shell/integration/cpp_test.sh
+++ b/src/test/shell/integration/cpp_test.sh
@@ -129,6 +129,291 @@ EOF
   bazel build //$pkg:a || fail "build failled"
 }
 
+function test_recompile_on_changed_used_headers() {
+  local -r pkg=$FUNCNAME
+  mkdir -p $pkg
+  cat > $pkg/BUILD <<EOF
+cc_binary(name="a", srcs=["a.cc"], deps=["b"])
+cc_library(name="b", includes=["."], hdrs=["b.h"])
+EOF
+
+  cat > $pkg/a.cc <<EOF
+#include "b.h"
+
+int main(void) {
+  return B_RETURN_VALUE;
+}
+EOF
+
+  cat > $pkg/b.h <<EOF
+#define B_RETURN_VALUE 31
+EOF
+
+  bazel build -s //$pkg:a >& $TEST_log || fail "build failed"
+  expect_log "Compiling $pkg/a.cc"
+
+  cat > $pkg/b.h <<EOF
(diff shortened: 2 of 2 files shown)


obruns:
@googlebot I signed it!

lberki:
@oquenchil , can you take care of this? I have a little too much to do this week and the next :(

lberki:
I don't _think_ this is going to work because this would probably break our internal tests for the reasons described [here](https://github.com/bazelbuild/bazel/issues/8319#issuecomment-865934356). 

I can imagine two possible resolutions:
* @oquenchil runs our internal presubmits on this change. If they pass, I got all this wrong. They are extensive enough that in that case, I'm comfortable merging this pull request.
*  You change this pull request to add `getDeclaredIncludeSrcs()` to the action key iff include scanning is enabled. Given that that means that internal behavior does not change, I'm fairly certain that it will not break anything at Google..

obruns:
Thanks for your feedback @lberki. I am happy to proceed with either of your proposals.

About the failing tests on Windows/msvc: `/showIncludes` and parsing should be enabled, AFAICT. I would have to look into why exactly those tests are failing on msvc. Unfortunately, I currently don't have a working Windows/msvc/Bazel setup. Therefore, I could:

* Disable those test cases based on `$is_windows` which is already available in `cpp_test.sh` or
* Spent some time setting things up and proceed with properly fixing those failures (not preferred)

oquenchil:
I agree that the original issue #8319 needs to be addressed. The priority right now is not the highest though, it doesn't happen that often that you add a new header. I will see if I have some time to properly look at this during Q3.

obruns:
@oquenchil and @lberki, can you give a status update? Do you still plan to have a closer look during Q3/2021? Thanks.

lberki:
@oquenchil ?

oquenchil:
No sorry, no time this quarter.

Yannic:
Please consider the following test-case that starts failing with this change:

```bash
function test_compile_on_header_existence_check() {
  local -r pkg=$FUNCNAME
  mkdir -p $pkg
  cat > $pkg/BUILD <<EOF
cc_binary(name="a", srcs=["a.cc"], deps=["b"])
cc_library(name="b", includes=["."], hdrs=["b.h"])
EOF

  cat > $pkg/a.cc <<EOF
#include "b.h"

int main(void) {
  return 0;
}
EOF

  cat > $pkg/b.h <<EOF
#if __has_include("unused.h")
#error unused.h shall not exist
#endif  // __has_include("unused.h")
EOF

  bazel build -s //$pkg:a >& $TEST_log || fail "build failed"
  expect_log "Compiling $pkg/a.cc"

  cat > $pkg/BUILD <<EOF
cc_binary(name="a", srcs=["a.cc"], deps=["b"])
cc_library(name="b", includes=["."], hdrs=["b.h", "unused.h"])
EOF

  cat > $pkg/unused.h <<EOF
=== BANANA ===
EOF

  bazel build -s //$pkg:a >& $TEST_log && fail "unexpected success" || true
  expect_log "Compiling $pkg/a.cc"
}
```

See https://github.com/bazelbuild/bazel/issues/8319#issuecomment-976693110 for more context.

oquenchil:
Closing this PR. If still interested, please file an issue instead so that this is tackled in a different way. It should pass Yannic's test and internal tests.

Issue #13349, opened 2021-04-13
/dev/null in action cache, remote caches, etc  ( `/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` )


jerrymarino:
### Description of the problem / feature request:

Hello! I noticed today about 100x more requests for a certian object in a remote http cache. `/cas/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```bash
git clone https://github.com/bazelbuild/rules_apple.git
cd rules_apple 
bazel build examples/ios/PrenotCalculator:PrenotCalculator --disk_cache=/tmp/disk-cache --appe_platform_type=ios
```
```
grep -r -a 2 e3b0c /tmp/disk_cache/
```

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

Happens on 3x and 4x releases, dev builds

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```
https://github.com/bazelbuild/rules_apple.git
```
###  Have you found anything relevant by searching the web?
https://github.com/bazelbuild/bazel/issues/11063
https://github.com/bazelbuild/bazel/issues/6260

### Any other information, logs, or outputs that you want to share?

```
$ grep -r -B 2 e3b0c /tmp/disk_cache/
/tmp/disk_cache//ac/ac/acc0ca28dbfd1fbba3851db7fab5db5e423dbc2e1e15a5d35a5fe8d8052439b7-Hbazel-out/apl-ios_x86_64-dbg/bin/tests/ios/app/_objs/App_objc/arc/main.dF
/tmp/disk_cache//ac/ac/acc0ca28dbfd1fbba3851db7fab5db5e423dbc2e1e15a5d35a5fe8d8052439b7-@2a034b09ad418213cad1f5da6c4903187ec852c7c76a0dd29f370269424495c6
                                                       2B
/tmp/disk_cache//ac/ac/acc0ca28dbfd1fbba3851db7fab5db5e423dbc2e1e15a5d35a5fe8d8052439b7:@e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855BB
[...]

Issue #11063, opened 2020-04-03
Bazel uploads AC entries that reference non-existent (zero-size) CAS blob


scele:
### Description of the problem / feature request:

Bazel remote cache grpc backend seems to optimize away uploading the zero-sized CAS blob (sha256 = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855). This means that all remote cache implementations must add special logic to handle that digest, for example: https://github.com/buchgr/bazel-remote/issues/233

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I'm running https://github.com/buildbarn/bb-storage/commit/fb0be44cf0e61fe98608748b47f16ee5049d142a as a remote cache at grpc://localhost:9092.

```
$ cat BUILD 
cc_binary(
    name = "test",
    srcs = ["test.cc"],
)

$ bazel clean && bazel build //:test --remote_cache=grpc://localhost:9092 --experimental_remote_grpc_log=$PWD/grpc.log
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
INFO: Invocation ID: 8e4eba6d-d513-4afb-8260-63a2b182eb21
INFO: Analyzed target //:test (13 packages loaded, 47 targets configured).
INFO: Found 1 target...
Target //:test up-to-date:
  bazel-bin/test
INFO: Elapsed time: 0.524s, Critical Path: 0.25s
INFO: 2 processes: 2 linux-sandbox.
INFO: Build completed successfully, 6 total actions
```

Here is the decoded remote grpc log:

<details>
  <summary>Click to expand</summary>

```
metadata {
  tool_details {
    tool_name: "bazel"
    tool_version: "2.2.0"
  }
  action_id: "capabilities"
  tool_invocation_id: "549975db-fdf6-4fcf-8dd8-e1232062e7f6"
  correlated_invocations_id: "6814a271-36e6-4fc7-ac25-ac5ecce1c1a7"
}
status {
}
method_name: "build.bazel.remote.execution.v2.Capabilities/GetCapabilities"
details {
  get_capabilities {
    request {
      instance_name: "local"
    }
    response {
      cache_capabilities {
        digest_function: MD5
        digest_function: SHA1
        digest_function: SHA256
        digest_function: UNKNOWN_ENUM_VALUE_DigestFunction_5
        digest_function: UNKNOWN_ENUM_VALUE_DigestFunction_6
[...]

PR #13594, opened 2021-06-22
Remote: Do not upload empty output to remote cache.

coeuvre:
An unintended side-effect of change cc2b3ecb9283ebdb297fc3fb6407f4697c850ac2 is that zero-sized blob won't be uploaded to gRPC cache. However, the behavior is existed for a long time and the REAPI spec is also updated accordingly. This change makes the behavior explicit and brings it to other remote cache backends.

Context https://github.com/bazelbuild/bazel/issues/11063.

Fixes https://github.com/bazelbuild/bazel/issues/13349.


philwo:
Interesting! 🤔 But if it's in the spec, yeah, why not. ^^

philwo (approved):

Issue #8248, opened 2019-05-07
Store remote metadata in the on disk action cache


buchgr:
This is part of #6862.

The persistent action cache stores the state of Bazel's output base. It does not know about remotely stored files. We need to extend it to do so.

```
$ bazel build :foo --experimental_remote_download_outputs=minimal
$ bazel shutdown
# This will be clean build, because the metadata from the previous build are
# not yet properly persisted to disk
$ bazel build :foo --experimental_remote_download_outputs=minimal
```

moroten:
This problem makes up to several minutes for me. Any updates on priority?

buchgr:
No updates.

moroten:
I saw the related #10233 which talks about refactoring `findMissingBlobs`. Is there more difficulties relating to `--experimental_inmemory_jdeps_files` and `--experimental_inmemory_dotd_files` that needs to be considered? Would the right way to solve this issue be to implement `--disk_cache` support in combination with remote execution or should there be another flag involved?

I do see this phenomenon also when the analysis cache is discarded due to changing configuration flags, even if the server is still alive.

sohaibiftikhar:
@buchgr I'd like to work on this. Any pointers to get started?

coeuvre:
This can be enabled by setting `--experimental_action_cache_store_output_metadata`.

Issue #12665, opened 2020-12-09
Separate downloads from action execution to provide more download options


coeuvre:
In the current implementation of remote execution, outputs can only be downloaded when an action executes. We should 
 separate downloads from action execution so that we can provide more download options to users. For example:
1. Provide a download API so that other code can download outputs of previously executed actions when necessary. Use cases:
    - Currently, we download the outputs after executed an action remotely and wait for the downloads to complete before starting execution of other actions. We can change to download the outputs of current action and start executing next action in the same time. This will improve the performance of full-downloads mode a lot.
    - Some files downloaded by the minimal/toplevel mode are symlink pointing to files we didn't download. There are some issues root-caused by this behavior. We can provide options to download the actual files when reading the symlink. This will fix most of bugs related to symlink produced by remote execution.
    - Dynamic execution can download outputs of previously remotely executed action in the local branch.
2. Add a post-build step for downloads so that we can download files which are only known after the build complete. Use cases:
    - We can combine this with aspects and BEP to download files required by IDE after the build.
    - Tests require some of the outputs to run. There are some issues related to running tests with top-level mode which should be fixed by this change.
3. Add a command line flag to only download arbitrary output files. Use cases:
    - Allow to download anything of the output tree without going through the build steps

We need a way to specify which files to download after the build (option 2) or with the download flag (option 3). For example:
- Reuse the flags --remote_download_toplevel to only download output files of top level targets and --remote_download_minimal to only download necessary files.
- Add a new flag which accept a list of targets whose output files should be downloaded.
- I am not sure how to specify download files at file level. Is that useful (and can be implemented) to support regex search over the output tree? (like `rg` over a directory)

Options for downloading the content of a symlink file:
- Provide a flag to control whether the download should follow the symlinks, this applies to all 3 download options above
- When the code is reading the content of a symlink file (e.g. in the RemoteActionFS), we can follow the symlink using option 1
- When other tools are reading the content of a symlink (e.g. IDE), the tool can use option 3 to download the file.
- When users want to read the content of a symlink, they can use option 3 to download the file manually.

**Related Issues**
- [x] #13625
- [x] #12855
- [x] #12635
- [x] #12592
- [x] #12137
- [x] #11942
- [x] #11920
- [x] #11834
- [x] #11558
- [x] #11119
- [x] #10971
- [x] #10618
- [x] #10525
- [x] #10025
- [x] #9386
- [x] #9248
- [x] #8581

coeuvre:
[...]

Issue #13566, opened 2021-06-09
--explain explain should tell what files changed


metti:
### Description of the problem / feature request:

Running `bazel --explain explain.log` will create `explain.log` with an explanation why a target was rebuilt. Unfortunately it lacks detailed information about which files changed. Instead a typical diagnostic is:

```
Build options: --announce_rc --explain explain.log
Executing action 'BazelWorkspaceStatusAction stable-status.txt': unconditional execution is requested.
Executing action 'Building something //some:target': One of the files has changed.
```

### Feature requests: what underlying problem are you trying to solve with this feature?

I would like to specify `--explain explain.log` for my builds unconditionally to adhoc debug "huh-why-did-this-now-build?"-situations.

Ninja's explain log `-d explain` helped me in the past numerous times to track down dependency issues.

### What's the output of `bazel info release`?

`development version`

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

Recent Bazel used for Android

lberki:
The reason why Bazel doesn't have this admittedly basic piece of functionality is that it's, surprisingly, not doable with the data Bazel keeps about an action.

Since some action have a lot of inputs, Bazel doesn't store the checksums of each input or even the set of inputs for each cached action (except for C++). Instead, what it does it stores just the xor of all the checksums of all `<input path, checksum of input file>` pairs (see `MetadataDigestUtils.fromMetadata()`

So when one of these files changes, it's detected, but Bazel cannot tell which one changed (hell, I don't even know how we detect when the set of inputs files changed...)

In order to fix this, we could do two things:
1. Peek into Skyframe invalidation (that wouldn't work on Bazel server restarts, but it would work most of the time and it would be complicated)
2. Replace the action cache (it's looking increasingly like a tempting target for multiple reasons, but it's not a small amount of work and it has to fulfill some pretty strict CPU and RAM constraints)

lberki:
(/cc @janakdr since we discussed this yesterday)

PR #13604, opened 2021-06-24
Save remote output metadata to ActionCache and load them before checking action cache

coeuvre:
This PR extends `ActionCache.Entry` to store output metadata by having a map of <Path, Metadata>. This map is updated after action execution when we update action cache so that metadata of all outputs of the action are saved. Before checking the action cache (when executing actions), we will load the output metadata into output store if it is remote and the correspondingly local one is missing.

With this change, remote output metadata is saved to disk so build without bytes can use them among server restarts. 
 We can also download outputs after action execution since remote output metadata can be accessed outside.

Part of https://github.com/bazelbuild/bazel/issues/12665.

Fixes https://github.com/bazelbuild/bazel/issues/8248.


coeuvre:
@philwo @janakdr @alexjski PTAL

coeuvre:
Another thought: we can probably move "load metadata cache" and "update metadata cache" to `OutputService` so that this change only affect Bazel's remote execution/cache.

janakdr:
General idea seems reasonable, will defer to @alexjski

src/main/java/com/google/devtools/build/lib/actions/cache/MetadataCache.java
```diff
@@ -0,0 +1,12 @@
+package com.google.devtools.build.lib.actions.cache;
+
+import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.actions.FileArtifactValue;
+
+public interface MetadataCache {
```
alexjski, line 8: We very likely do not need that interface -- we could just call Metadata with the metadata instead.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -289,32 +316,162 @@ public void put(String key, ActionCache.Entry entry) {
...
+        "Must use getTreeMetadata to get tree artifacts and their children: %s",
+        artifact
+    );
+    String key = artifact.getExecPathString();
+    byte[] data = get(metadataMap, key);
+    try {
+      return data != null ? decodeMetadata(data) : null;
+    } catch (IOException e) {
+      // return null if can't decode metadata
+      return null;
```
alexjski, line 572: Maybe log some warning? I assume that would be an abnormal situation.

alexjski (changes requested):

alexjski:
Can you explain why we need separate files for the metadata as opposed to storing that with the existing structure? That could technically open possibilities of entries present in the metadata, but not in the cache and the other way round. Is the worry that we could read the metadata in case of a cache miss?

src/main/java/com/google/devtools/build/lib/actions/cache/MetadataCache.java
coeuvre: Not sure whether I understand this correctly. Could you please be more specific?

coeuvre:
They are technically different type of maps: the action cache is a map of `<Path of action's first output, Digest of action data>` while the metadata cache is a map of `<Path of a output file, Metadata>`. Storing them into different files seems to be reasonable.

> That could technically open possibilities of entries present in the metadata, but not in the cache and the other way round.

I think this is a valid case. If entries present in the metadata but not in the cache, Bazel will just ignore them. If the other way round, it is just the case before this change.

> Is the worry that we could read the metadata in case of a cache miss?

Even if we store metadata into action cache files, a cache miss shouldn't remove the metadata.

However, we can change to store the metadata to the existing action cache file (by suffix the path of output file with `:metadata` for example) if you prefer this way.

alexjski (commented):
> They are technically different type of maps: the action cache is a map of `<Path of action's first output, Digest of action data>` while the metadata cache is a map of `<Path of a output file, Metadata>`. Storing them into different files seems to be reasonable.

They are, but I don't think we need 2 maps -- my suggestion was to merge them in which case the storage would be merged too.

> 
> > That could technically open possibilities of entries present in the metadata, but not in the cache and the other way round.
> 
> I think this is a valid case. If entries present in the metadata but not in the cache, Bazel will just ignore them. If the other way round, it is just the case before this change.

I am afraid that will be a possibility, unless we do something like differ the version of action cache based on whether this feature is enabled (that's technically abuse of version). Another thing we could do is if we had some global indicator in the action_cache file on whether it has metadata, we could reject action cache if it disagrees on whether it should have it.
> 
> > Is the worry that we could read the metadata in case of a cache miss?
> 
> Even if we store metadata into action cache files, a cache miss shouldn't remove the metadata.
> 
> However, we can change to store the metadata to the existing action cache file (by suffix the path of output file with `:metadata` for example) if you prefer this way.

Yeah, that's one way to have a global variance on the option. That sounds reasonable with only the caveat that if we have more options like that, we may end up with multitude of prefixes, but maybe that is not a thing to worry about. Also, there is the case that if you had a prior build storing the metadata, follow up one without it could technically still use it, but I think that is a case not worth optimizing for.

That would take out the possibility of the metadata being absent from prior builds due to changing options -- missing metadata in this case would be an indication of a bug.

I think there is a case with a separate path -> metadata cache could be (made up a start-up option `--store_metadata_in_action_cache`, please forgive me that) problematic:

```
$ bazel --store_metadata_in_action_cache build //foo # generates foo.out with "hello1", stores metadata
# Update BUILD file for foo
$ bazel build //foo # Updates action cache, but not metadata
$ bazel shutdown
$ rm -rf outputs
$ baze --store_metadata_in_action build //foo # that gets an action cache hit from last build, however the metadata for foo.out is stale
```
Is that possible?

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/actions/cache/MetadataCache.java
coeuvre: Removed `MetadataCache` interface. 

Loading metadata from action cache is happened before checking action cache (hence before action execution), we can't inject metadata with `ActionMetadataHandler` (`executionMode` is `false` at that time).

coeuvre:
Updated to merge two maps into one. For metadata out an output, the key is its execPath with a suffix `:metadata`.

src/main/java/com/google/devtools/build/lib/actions/cache/MetadataCache.java
alexjski: That makes sense, thanks!

alexjski:
> Updated to merge two maps into one. For metadata of an output, the key is its `execPath` plus a suffix `:metadata`.

That's good. I think I initially had more negative reaction to separate maps than it calls for (sorry) because I didn't realize the output fingerprint check is still present, which negates all of the risks. Anyway with merged map, we know they both are always in sink which is nice thing in itself.

coeuvre:
Regarding https://github.com/bazelbuild/bazel/pull/13604#issuecomment-868172207, I am thinking separating these metadata methods from `ActionCache` and let `OutputService` do the job (that's why it initially has a `MetadataCache` interface).

The goal of this PR is allowing Bazel to download remote outputs after action execution. To do that, we need:
1. Download metadata of remote outputs and save them during action execution.
2. When the remote output is downloaded, the remote metadata should be removed from cache (so we can re-construct metadata from local file system).

1 is implemented - during the step when we update action cache, we also check for metadata injected during action execution, and save those which is "remote".

2 is only implemented for downloads happened within action execution - if the metadata injected is not "remote", we remove the key from cache during updating the action cache.

Considering the case that the local action cache of a remote action is valid (hence action execution is skipped) but we need to download its outputs. With current implementation, it won't work.

We need a post action execution download procedure. Whichever module implements this feature, it needs to access ActionCache to remove metadata for downloaded outputs. Accessing ActionCache somewhere else is risky IMO.

If we move these metadata methods to the `OutputService` i.e. it is responsible for maintaining the metadata, then we will have following steps when executing an action:
1. Load metadata from OutputService (if it is not null) into `OutputStore` (this is essentially constructing an action-local metadata cache from a global one).
2. Check the ActionCache with `OutputStore`.
3. If cache is invalidated, execute the action and update metadata with method provided by `OutputService` (merge the action-local metadata cache into a global one).
4. Call the post action download method provided by the `OutputService` during which the metadata is updated by itself for downloaded outputs.

WDYT?

alexjski:
> Regarding [#13604 (comment)](https://github.com/bazelbuild/bazel/pull/13604#issuecomment-868172207), I am thinking separating these metadata methods from `ActionCache` and let `OutputService` do the job (that's why it initially has a `MetadataCache` interface).
> 
> The goal of this PR is allowing Bazel to download remote outputs after action execution. To do that, we need:
> 
> 1. Download metadata of remote outputs and save them during action execution.
> 2. When the remote output is downloaded, the remote metadata should be removed from cache (so we can re-construct metadata from local file system).
> 
> 1 is implemented - during the step when we update action cache, we also check for metadata injected during action execution, and save those which is "remote".
> 
> 2 is only implemented for downloads happened within action execution - if the metadata injected is not "remote", we remove the key from cache during updating the action cache.

That may be a reason for 2 maps, so you can still have cache hits for local actions.

> 
> Considering the case that the local action cache of a remote action is valid (hence action execution is skipped) but we need to download its outputs. With current implementation, it won't work.
> 
> We need a post action execution download procedure. Whichever module implements this feature, it needs to access ActionCache to remove metadata for downloaded outputs. Accessing ActionCache somewhere else is risky IMO.
> 
> If we move these metadata methods to the `OutputService` i.e. it is responsible for maintaining the metadata, then we will have following steps when executing an action:
> 
> 1. Load metadata from OutputService (if it is not null) into `OutputStore` (this is essentially constructing an action-local metadata cache from a global one).
> 2. Check the ActionCache with `OutputStore`.
> 3. If cache is invalidated, execute the action and update metadata with method provided by `OutputService` (merge the action-local metadata cache into a global one).
> 4. Call the post action download method provided by the `OutputService` during which the metadata is updated by itself for downloaded outputs.
> 
> WDYT?

My very first reaction is that I think we discussed that before and that was not agreed upon. I see 2 reasons to need to download the outputs:

1. Follow-up local actions.
2. Allow users to inspect the outptus.

Seems like that if you need those, your best bet may be a FUSE-based solution which tracks all outputs which would deem all of these changes unnecessary (see https://github.com/bazelbuild/bazel/pull/12823).

Another thing to consider about downloading results, if you wanted to do that eagerly (I think if we can have FUSE, that would be strictly better) is a subscribing to events and looking at outputs. We already issue those for top-level actions (BEP), there also is CachedActionEvent (you might need to add the metadata to it though). I am not pressing too hard, but that is something to think about -- it looks like we may already have the desired interface there. For reference, we already look at the outputs right there: https://github.com/bazelbuild/bazel/blob/4ccb2cced62e4fc070349e1b8ea6c941074bccbb/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java#L657-L664.

This PR as which only tracks the metadata in action cache, to me can serve an interesting case for remote-only builds, which is something that we discussed before and it is indeed a nice thing to have.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -435,6 +441,100 @@ public void updateActionCache(
...
+
+            if (hasRemote) {
+              TreeArtifactValue.Builder builder = TreeArtifactValue.newBuilder(artifact);
+              for (Map.Entry<TreeFileArtifact, FileArtifactValue> childEntry :
+                  treeArtifactValue.getChildValues().entrySet()) {
+                TreeFileArtifact child = childEntry.getKey();
+                FileArtifactValue childMetadata = childEntry.getValue();
+                if (!childMetadata.isRemote()) {
+                  // Only save remote tree file metadata
+                  childMetadata = MISSING_FILE_MARKER;
```
alexjski, line 513: I wonder if we should have a different marker or at least add a comment that this differs from what that represents -- this is metadata for a file which does not exists at all. In here, we are saying it exists, but locally.

alexjski (changes requested):
I will leave that up to you whether you want to have 2 maps or 1. One high-level comment is that if you are keeping that together, enhancing the ActionCache.Entry may give you a better interface where you do a single update with all remote metadata. This can potentially spare some of the OutputStore manipulations (injections are still needed). Something to think about.

coeuvre:
> My very first reaction is that I think we discussed that before and that was not agreed upon. I see 2 reasons to need to download the outputs:
> 
> 1. Follow-up local actions.
> 2. Allow users to inspect the outptus.
> 

Another case is we can just inject metadata for the outputs of an action and start executing dependent actions while keeping output downloads in the background (today, we have to wait for the download).

> Seems like that if you need those, your best bet may be a FUSE-based solution which tracks all outputs which would deem all of these changes unnecessary (see #12823).
> 

IIUC, we are experimenting with non-FUSE-based solution. The solution in my head is we can build artifacts as normal and use "outputs" flags to tell Bazel which outputs we want to download, e.g. no flags means we want all the outputs, `--remote_download_minimal` means we want a minimal set of outputs (we can also introduce additional "outputs" flags to tell Bazel in a more fine grained level).

If we need outputs that weren't downloaded during previous build, we can then "rebuild" with different "outputs" flags which will hit the local action cache but start downloading missing outputs.

The entire progress for executing an action remotely should be:

1. Check for local action cache. If hit, go to step 6.
2. Remove injected metadata and local outputs if any.
3. Check for remote cache. If hit, go to step 5.
4. Upload inputs and execute.
5. Download and inject metadata.
6. Only download outputs which is requested (remove injected metadata after download).

This works without FUSE but provides similar experience.

> Another thing to consider about downloading results, if you wanted to do that eagerly (I think if we can have FUSE, that would be strictly better) is a subscribing to events and looking at outputs. We already issue those for top-level actions (BEP), there also is CachedActionEvent (you might need to add the metadata to it though). I am not pressing too hard, but that is something to think about -- it looks like we may already have the desired interface there. For reference, we already look at the outputs right there:
> 
> https://github.com/bazelbuild/bazel/blob/4ccb2cced62e4fc070349e1b8ea6c941074bccbb/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java#L657-L664
> 
> .

This is also a good place to consider for post execution downloading. One thing I mentioned above is that we need to remove remote metadata from cache after download. So we need to decide whether should we expose ActionCache to other modules or we keep ActionCache as it is and introduce a separate interface for metadata only.

> This PR as which only tracks the metadata in action cache, to me can serve an interesting case for remote-only builds, which is something that we discussed before and it is indeed a nice thing to have.

janakdr:
Perhaps the word "FUSE" is a bit of a red herring. Let's say, any
filesystem-based output service. An NFS volume would also work. Chi, what
is the reason to avoid such a setup and centralize logic inside Bazel?

brentleyjones:
Just chiming in, but macFUSE performance isn't the best, so any solution that requires FUSE might not be performant on macOS.

coeuvre:
To be clear, I wasn't meant to replace FUSE-based (or any filesystem-based output service) solution but just wanted to improve "build withouts bytes".

> Chi, what is the reason to avoid such a setup and centralize logic inside Bazel?

IIUC, we don't offer the setup for Bazel today. For users who care about network bandwidth, the only choice is "build without bytes". The ways moving forward are probably:
1. Improve "build without bytes".
2. Add support for FUSE like #12823.

I don't think these two ways are incompatible with each other.

philwo:
I would like to go ahead with Chi's idea of having this logic in Bazel.

I think filesystem-based layers are probably the most user-friendly interface when they work, but they come with lots of challenges, too: They are platform-specific, they are often slower than non-filesystem-based APIs (e.g. see sandboxfs performance vs symlink-based sandbox performance), they require additional setup (e.g. install FUSE), they require having special privileges on the system (admin rights, ability to load a kernel extension, ability to mount a filesystem, ...) and they might not be available on all systems due to security policies. As we know from Google they can perform very well in a carefully managed IT environment, but I don't believe we will be able to provide a good out of the box experience for these solutions on typical user systems.

The two approaches should nicely complement each other though!

alexjski:
> I would like to go ahead with Chi's idea of having this logic in Bazel.
> 
> I think filesystem-based layers are probably the most user-friendly interface when they work, but they come with lots of challenges, too: They are platform-specific, they are often slower than non-filesystem-based APIs (e.g. see sandboxfs performance vs symlink-based sandbox performance), they require additional setup (e.g. install FUSE), they require having special privileges on the system (admin rights, ability to load a kernel extension, ability to mount a filesystem, ...) and they might not be available on all systems due to security policies. As we know from Google they can perform very well in a carefully managed IT environment, but I don't believe we will be able to provide a good out of the box experience for these solutions on typical user systems.
> 
> The two approaches should nicely complement each other though!

There already was a discussion about that before and I also believe there was some effort put towards testing how NFS-based solutions perform. I think that in order to have most informed discussion, we should merge those and look at the results from the tests and them make design decisions. I believe that this code review may not be the best place for that.

The main thinking so far was that if we have a FUSE/NFS-based solution available, then the incentive of this may be low. What I was missing from the proposal before was the justification for why the added complexity is needed if we have FUSE/NFS solution. If there is none, I would strongly prefer to i

> > My very first reaction is that I think we discussed that before and that was not agreed upon. I see 2 reasons to need to download the outputs:
> > 
> > 1. Follow-up local actions.
> > 2. Allow users to inspect the outptus.
> 
> Another case is we can just inject metadata for the outputs of an action and start executing dependent actions while keeping output downloads in the background (today, we have to wait for the download).
> 
> > Seems like that if you need those, your best bet may be a FUSE-based solution which tracks all outputs which would deem all of these changes unnecessary (see #12823).
> 
> IIUC, we are experimenting with non-FUSE-based solution. The solution in my head is we can build artifacts as normal and use "outputs" flags to tell Bazel which outputs we want to download, e.g. no flags means we want all the outputs, `--remote_download_minimal` means we want a minimal set of outputs (we can also introduce additional "outputs" flags to tell Bazel in a more fine grained level).
> 
> If we need outputs that weren't downloaded during previous build, we can then "rebuild" with different "outputs" flags which will hit the local action cache but start downloading missing outputs.
> 
> The entire progress for executing an action remotely should be:
> 
> 1. Check for local action cache. If hit, go to step 6.
> 2. Remove injected metadata and local outputs if any.
> 3. Check for remote cache. If hit, go to step 5.
> 4. Upload inputs and execute.
> 5. Download and inject metadata.
> 6. Only download outputs which is requested by the "outputs" flags in the background (remove injected metadata after download).
> 7. If there are more actions, go to step 1.
> 8. Wait for all downloads.
> 
> This works without FUSE but provides similar experience.
> 
> > Another thing to consider about downloading results, if you wanted to do that eagerly (I think if we can have FUSE, that would be strictly better) is a subscribing to events and looking at outputs. We already issue those for top-level actions (BEP), there also is CachedActionEvent (you might need to add the metadata to it though). I am not pressing too hard, but that is something to think about -- it looks like we may already have the desired interface there. For reference, we already look at the outputs right there:
> > https://github.com/bazelbuild/bazel/blob/4ccb2cced62e4fc070349e1b8ea6c941074bccbb/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java#L657-L664
> > 
> > .
> 
> This is also a good place to consider for post execution downloading. One thing I mentioned above is that we need to remove remote metadata from cache after download. So we need to decide whether should we expose ActionCache to other modules or we keep ActionCache as it is and introduce a separate interface for metadata only.

Why do we need to remove the metadata? Also, it may be useful what metadata you mean in this context -- the newly added metadata in action cache or the metadata stored in ActionExecutionValue (skyframe).

> 
> > This PR as which only tracks the metadata in action cache, to me can serve an interesting case for remote-only builds, which is something that we discussed before and it is indeed a nice thing to have.

coeuvre:
> Why do we need to remove the metadata? Also, it may be useful what metadata you mean in this context -- the newly added metadata in action cache or the metadata stored in ActionExecutionValue (skyframe).

I was referring to the newly added metadata in action cache. This PR __only__ save remote metadata and load them into OutputStore before checking the action cache. If we don't remove the remote metadata of one output from action cache after downloaded it, we will still use the remote metadata from action cache instead of re-calculating it from local file system next time when we check action cache (even if the local file content is changed).

Now I think it again, maybe we can change to storing __all__ metadata of outputs into action cache (which helps https://github.com/bazelbuild/bazel/issues/13566). Re-calculating all metadata of outputs for the action from local file system (we do it today as well) and compare them with those stored in the action cache. Then we don't need to remove metadata after downloads.

It looks like saving metadata directly in action cache is a preferred way. Then I would like to make it stored within one map by extending `ActionCache.Entry`.

alexjski:
> > Why do we need to remove the metadata? Also, it may be useful what metadata you mean in this context -- the newly added metadata in action cache or the metadata stored in ActionExecutionValue (skyframe).
> 
> I was referring to the newly added metadata in action cache. This PR **only** save remote metadata and load them into OutputStore before checking the action cache. If we don't remove the remote metadata of one output from action cache after downloaded it, we will still use the remote metadata from action cache instead of re-calculating it from local file system next time when we check action cache (even if the local file content is changed).
> 

I agree, we need to take care of that. My thinking was that it may be a better API to just request a desired end state rather than have fine-grain manipulations. My preference would be to have an API like this:

```
Map<PathFragment, FileArtifactMetadata> newOutputs = ...;
cache.updateMetadata(key, newOutputs);
```

In fact, that could be a single call as part of updating the cache, which we already do.

over:
```
Map<PathFragment, FileArtifactMetadata> oldOutputs = ...;
Map<PathFragment, FileArtifactMetadata> newOutputs = ...;
oldOutputs.keySet().stream().filter(k -> !newOutputs.containsKey(k)).forEach(cache::removeMetadata);
newOutputs.forEach((k, m) -> cache.putMetadata(k, m);
```

In fact, then you can choose in the cache implementation to only track the remote ones and you don't have to worry about removing old entries since the operation resembles `map = newMap` vs `manipulate(map)`.

> Now I think it again, maybe we can change to storing **all** metadata of outputs into action cache (which helps #13566). Re-calculating all metadata of outputs for the action from local file system (we do it today as well) and compare them with those stored in the action cache. Then we don't need to remove metadata after downloads.
> 

Not sure if we need that. My point was merely about the API seeming too fine grained. Logically it makes perfect sense.

> It looks like saving metadata directly in action cache is a preferred way. Then I would like to make it stored within one map by extending `ActionCache.Entry`.

I agree, if we are doing that, the `update(Map<...>)` seems like the simplest choice.

coeuvre:
@alexjski  The code is updated. PTAL.

It works fine to encode/decode `FileArtifactValue` with `ObjectCodecs`. However, when serializing `TreeArtifactValue`, since it has reference to `SpecialArtifact` -> `ArtifactRoot` -> `Path` -> `FileSystem` -> `FileSystem`, there is an infinite loop. Not sure whether the `ObjectCodecs` is used correctly or not, I have to use create `SerializableTreeArtifactValue`.

alexjski:
> @alexjski The code is updated. PTAL.
> 
> It works fine to encode/decode `FileArtifactValue` with `ObjectCodecs`. However, when serializing `TreeArtifactValue`, since it has reference to `SpecialArtifact` -> `ArtifactRoot` -> `Path` -> `FileSystem` -> `FileSystem`, there is an infinite loop. Not sure whether the `ObjectCodecs` is used correctly or not, I have to use create `SerializableTreeArtifactValue`.

I don't think you need that -- in fact, we already do serialize `TreeArtifactValue`. The trick for it is to register the `FileSystem` object as a serialization constant, using [addReferenceConstant](https://github.com/bazelbuild/bazel/blob/97c5019b0f9825170177f6a1da794145db12a507/src/main/java/com/google/devtools/build/lib/skyframe/serialization/ObjectCodecRegistry.java#L341-L344).

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -512,10 +539,51 @@ public synchronized void dump(PrintStream out) {
...
+        String path = (id >= 0 ? indexer.getStringForIndex(id) : null);
+        if (path == null) {
+          throw new IOException("Corrupted file index");
+        }
+
+        byte[] bytes = new byte[VarInt.getVarInt(source)];
+        source.get(bytes);
+        FileArtifactValue metadata;
+        try {
+          metadata = (FileArtifactValue) objectCodecs.deserialize(ByteString.copyFrom(bytes));
```
alexjski, line 555: ditto here -- you can create a CodedOutputStream with both [a ByteBuffer](https://github.com/protocolbuffers/protobuf/blob/b90ec9c242b303e90811deebaa2e3e684b63de91/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L175-L177) and a [byte array slice](https://github.com/protocolbuffers/protobuf/blob/b90ec9c242b303e90811deebaa2e3e684b63de91/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L138-L141).

src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java
```diff
@@ -406,7 +406,6 @@ public void injectFile(Artifact output, FileArtifactValue metadata) {
         !output.isTreeArtifact() && !output.isChildOfDeclaredDirectory(),
         "Tree artifacts and their children must be injected via injectTree: %s",
         output);
-    checkState(executionMode.get(), "Tried to inject %s outside of execution", output);
```
alexjski, line 409: As much as removing checks is scary -- I think that's the right call -- it looks like this boolean literally protects against the metadata cache only (am I missing something?). We set it right after the metadata cache check and start with false otherwise.

If the metadata handler was to hand around after, we could consider a tri-state here maybe (checking cache -> cache hit/execution), but I don't see the point.

alexjski (changes requested):
Did I fail to see entirety of the diff or is this PR missing tests for the new feature?

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -229,6 +235,72 @@ protected boolean unconditionalExecution(Action action) {
...
+                      .getChildValues()
+                      .getOrDefault(childArtifact, MISSING_FILE_MARKER);
+
+              FileArtifactValue childMetadata;
+              // Only use cached metadata if it is remote and is missing from local file system.
+              if (cachedChildMetadata.isRemote() && localChildMetadata == MISSING_FILE_MARKER) {
+                childMetadata = cachedChildMetadata;
+                shouldInjectCached = true;
+              } else {
+                childMetadata = localChildMetadata;
```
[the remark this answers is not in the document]
coeuvre: It seems that we could have a tree metadata that is [mixed](https://github.com/bazelbuild/bazel/blob/aad74a3646d30d0b6a440a566f06aa264697676d/src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java#L528) with remote and local file metadata. Also, there could be a case that the injected tree metadata is entirely remote but the downloading is not complete (maybe interrupted).

In those cases, we need to merge two trees in order to hit the cache.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -229,6 +235,72 @@ protected boolean unconditionalExecution(Action action) {
...
+
+        try {
+          FileArtifactValue localMetadata;
+          try {
+          localMetadata = getMetadataOrConstant(metadataHandler, artifact);
+          } catch (FileNotFoundException e) {
+            localMetadata = MISSING_FILE_MARKER;
+          }
+
+          // Only inject metadata if it is remote and is missing from local file system.
```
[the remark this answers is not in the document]
coeuvre: I think, before the third build, the local output will be deleted before execution so we still end with good state (update-to-date remote metadata, no local output)?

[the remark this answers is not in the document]
alexjski: If you have an action cache hit, you don't execute the action and do not delete the outputs.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -409,13 +487,21 @@ public void updateActionCache(
...
-        FileArtifactValue metadata = getMetadataOrConstant(metadataHandler, output);
-        Preconditions.checkState(metadata != null);
-        entry.addFile(output.getExecPath(), metadata, /* saveExecPath= */ false);
+        if (output.isTreeArtifact()) {
+          SpecialArtifact parent = (SpecialArtifact) output;
+          TreeArtifactValue metadata = metadataHandler.getTreeArtifactValue(parent);
+          entry.addOutputTree(parent, metadata);
+        } else {
+          // Output files *must* exist and be accessible after successful action execution. We use
+          // the
```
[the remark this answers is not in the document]
alexjski: Did you forget to update the PR?

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -409,13 +487,21 @@ public void updateActionCache(
...
       }
       if (!metadataHandler.artifactOmitted(output)) {
-        // Output files *must* exist and be accessible after successful action execution. We use the
-        // 'constant' metadata for the volatile workspace status output. The volatile output
-        // contains information such as timestamps, and even when --stamp is enabled, we don't want
-        // to rebuild everything if only that file changes.
-        FileArtifactValue metadata = getMetadataOrConstant(metadataHandler, output);
-        Preconditions.checkState(metadata != null);
-        entry.addFile(output.getExecPath(), metadata, /* saveExecPath= */ false);
+        if (output.isTreeArtifact()) {
```
alexjski, line 490: We used to just remove the entry from cache on a miss -- why not keep it the same way?

alexjski (changes requested):
I think the 2 big asks will be feature switch for it (do we need to vary the action cache name on it?) and tests.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -287,6 +359,12 @@ public Token getTokenIfNeedToExecute(
       }
     }
     ActionCache.Entry entry = getCacheEntry(action);
+
+    // load remote metadata from action cache
+    if (entry != null) {
+      loadRemoteOutputMetadataIfMissing(action, entry, metadataHandler);
```
alexjski, line 365: I wonder if we shouldn't inject the outputs only if we get an action cache hit (after mustExecute). I understand that outputStore is currently used to get the metadata, but we could change `getMetadataMaybe` to look for the metadata in the entry itself or in some intermediary which has the loaded, but not yet injected files and trees.

My worry here would be if a remote executing action for some reason does not produce an output it should, but got it injected from the output store to begin with -- can we hold onto a stale output this way?

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -409,13 +487,21 @@ public void updateActionCache(
...
+        } else {
+          // Output files *must* exist and be accessible after successful action execution. We use
+          // the
+          // 'constant' metadata for the volatile workspace status output. The volatile output
+          // contains information such as timestamps, and even when --stamp is enabled, we don't
+          // want
+          // to rebuild everything if only that file changes.
+          FileArtifactValue metadata = getMetadataOrConstant(metadataHandler, output);
+          Preconditions.checkState(metadata != null);
+          entry.addOutputFile(output, metadata);
```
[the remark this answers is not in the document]
coeuvre: Changed to only save remote metadata.

src/main/java/com/google/devtools/build/lib/actions/cache/ActionCache.java
```diff
@@ -84,31 +91,121 @@
...
+     */
+    @AutoValue
+    abstract static class SerializableTreeArtifactValue {
+      public static SerializableTreeArtifactValue create(
+          ImmutableMap<String, FileArtifactValue> childData) {
+        return new AutoValue_ActionCache_Entry_SerializableTreeArtifactValue(childData);
+      }
+
+      // A map from parentRelativePath to the file metadata
+      abstract ImmutableMap<String, FileArtifactValue> childData();
```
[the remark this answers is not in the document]
coeuvre: Added the code to handle archived tree artifact.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
coeuvre: OutputStore is [cleared](https://github.com/bazelbuild/bazel/blob/aad74a3646d30d0b6a440a566f06aa264697676d/src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java#L467) right [before](https://github.com/bazelbuild/bazel/blob/aad74a3646d30d0b6a440a566f06aa264697676d/src/main/java/com/google/devtools/build/lib/skyframe/ActionExecutionFunction.java#L806) action execution so we should be safe in that case.

That said, I updated the code to follow your suggestion in order to decouple from that assumption of the current `ActionMetadataHandler`.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
coeuvre: I didn't change this logic. Not sure How the line you commented on are related to this.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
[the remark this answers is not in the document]
coeuvre: If we hit the cache ƒor the third build, the metadata of outputs from the second build are same with those from the first build. So the metadata cache is still in good state.

coeuvre:
Addressed most comments and added flag `--experimental_action_cache_store_output_metadata` to switch the feature. May need more time to add (and fix existing) tests.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -423,66 +452,97 @@ public synchronized void dump(PrintStream out) {
...
+    MetadataDigestUtils.write(entry.getFileDigest(), sink);
+
+    VarInt.putVarInt(entry.discoversInputs() ? files.size() : NO_INPUT_DISCOVERY_COUNT, sink);
+    for (String file : files) {
+      VarInt.putVarInt(indexer.getOrCreateIndex(file), sink);
+    }
+
+    MetadataDigestUtils.write(entry.getUsedClientEnvDigest(), sink);
+
+    VarInt.putVarInt(entry.getOutputFiles().size(), sink);
```
alexjski, line 502: I like that we always store the empty maps. You don't need to make a distinction on the cache format thanks to that. Is there a noticeable difference in the size of the cache with the flag disable due to the extra 2 ints?

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -287,6 +407,12 @@ public Token getTokenIfNeedToExecute(
       }
     }
     ActionCache.Entry entry = getCacheEntry(action);
+    CachedOutputMetadata cachedOutputMetadata = null;
+    // load remote metadata from action cache
+    if (entry != null && !entry.isCorrupted() && cacheConfig.storeOutputMetadata()) {
+      cachedOutputMetadata = loadRemoteOutputMetadataIfMissing(action, entry, metadataHandler);
```
alexjski, line 413: Nice! I like that we skip the load if that is disabled. Allows for a good feature switch.

alexjski (changes requested):

src/main/java/com/google/devtools/build/lib/actions/cache/ActionCache.java
```diff
@@ -84,31 +94,168 @@
...
+      }
+
+      // A map from parentRelativePath to the file metadata
+      abstract ImmutableMap<String, FileArtifactValue> childValues();
+
+      abstract Optional<SerializableArchivedRepresentation> archivedRepresentation();
+    }
+
+    @AutoValue
+    abstract static class SerializableArchivedRepresentation {
```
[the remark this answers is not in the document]
alexjski: About the archived tree artifact -- the prefix is already accounted for in the parent path, so that's not a problem anyway.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -229,6 +266,89 @@ protected boolean unconditionalExecution(Action action) {
...
+        try {
+          FileArtifactValue localMetadata;
+          try {
+            localMetadata = getMetadataOrConstant(metadataHandler, artifact);
+          } catch (FileNotFoundException e) {
+            localMetadata = null;
+          }
+
+          // Only inject metadata if it is remote and is missing from local file system.
+          if (cachedMetadata != null && cachedMetadata.isRemote() && localMetadata == null) {
```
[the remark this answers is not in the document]
coeuvre: Nope. The local metadata is loaded and [cached](https://github.com/bazelbuild/bazel/blob/aad74a3646d30d0b6a440a566f06aa264697676d/src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java#L285) by `OutputStore`.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -409,13 +549,41 @@ public void updateActionCache(
...
+          }
+          // Save tree file metadata if the archived representation is remote
+          saveTreeMetadata =
+              saveTreeMetadata
+                  || metadata
+                      .getArchivedRepresentation()
+                      .map(ar -> ar.archivedFileValue().isRemote())
+                      .orElse(false);
+
+          saveTreeMetadata = saveTreeMetadata && cacheConfig.storeOutputMetadata();
```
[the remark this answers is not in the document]
coeuvre: Good point!

src/main/java/com/google/devtools/build/lib/actions/cache/ActionCache.java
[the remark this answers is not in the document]
coeuvre: Added `derivedPathPrefix` as a member variable of `ActionCacheChecker` and use it when constructing `ArchivedTreeArtifact`.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -229,6 +266,89 @@ protected boolean unconditionalExecution(Action action) {
...
+            TreeArtifactValue.Builder merged = TreeArtifactValue.newBuilder(parent);
+            for (Map.Entry<TreeFileArtifact, FileArtifactValue> child :
+                cachedTreeMetadata.getChildValues().entrySet()) {
+              TreeFileArtifact childArtifact = child.getKey();
+              FileArtifactValue cachedChildMetadata = child.getValue();
+              FileArtifactValue localChildMetadata = localTreeMetadata.getChildValues().get(childArtifact);
+
+              FileArtifactValue childMetadata;
+              // Only use cached metadata if it is remote and is missing from local file system.
+              if (cachedChildMetadata.isRemote() && localChildMetadata == null) {
```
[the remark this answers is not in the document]
coeuvre: Yes, the logic can be simplify with the precondition that only remote metadata is saved. I will keep this open until the mentioned test case is added.

src/test/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutorTest.java
```diff
@@ -623,7 +623,8 @@ public Path getPathFromSourceExecPath(Path execRoot, PathFragment execPath) {
           },
           new ActionKeyContext(),
           Predicates.<Action>alwaysTrue(),
-          null);
+          null,
+          PathFragment.create("bin"));
```
alexjski, line 627: nitnit: It doesn't matter, but typically that should be something like 'bazel-out` -- may make tests easier to read if we use a common one or `out` at least.

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -229,6 +273,90 @@ protected boolean unconditionalExecution(Action action) {
...
+
+            cachedTreeMetadata
+                .archivedFileValue()
+                .map(
+                    fileArtifactValue -> {
+                      Artifact.ArchivedTreeArtifact archivedTreeArtifact =
+                          Artifact.ArchivedTreeArtifact.create(parent, derivedPathPrefix);
+                      return ArchivedRepresentation.create(archivedTreeArtifact, fileArtifactValue);
+                    })
+                .or(localTreeMetadata::getArchivedRepresentation)
```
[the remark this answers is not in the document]
alexjski: Looks tests answered that question.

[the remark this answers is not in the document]
coeuvre: Yes, tests tell us we are still targeting Java 8. :(

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -423,66 +452,97 @@ public synchronized void dump(PrintStream out) {
...
+    for (Map.Entry<String, FileArtifactValue> file : entry.getOutputFiles().entrySet()) {
+      VarInt.putVarInt(indexer.getOrCreateIndex(file.getKey()), sink);
+
+      ByteString.Output resultOut = ByteString.newOutput();
+      CodedOutputStream codedOut = CodedOutputStream.newInstance(resultOut);
```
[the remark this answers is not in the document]
coeuvre: Np! Updated to use custom serialization code.

coeuvre:
Added a few tests. More are coming tomorrow.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    Action action = new InjectOutputMetadataAction(output, FileArtifactValue.createForTesting(output));
+    output.getPath().delete();
+
+    // act
+
+    // Not cached.
+    runAction(action);
+
+    // assert
+    assertThat(output.getPath().exists()).isFalse();
```
alexjski, line 483: nit: That seems like a test of the test a little.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    ActionCache.Entry entry = cache.get(output.getExecPathString());
+    assertThat(entry).isNotNull();
+    assertThat(entry.getOutputFile(output)).isEqualTo(createRemoteMetadata(content));
+    assertStatistics(0, new MissDetailsBuilder().set(MissReason.NOT_CACHED, 1).build());
+  }
+
+  @Test
+  public void saveOutputMetadata_localFileMetadataNotSaved() throws Exception {
```
alexjski, line 465: Should we also test an action which injects remote metadata and creates the file locally. Think running remote runner with the option to materialize all outputs.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+            artifactRoot, PathFragment.create("bin/dummy"), NULL_ARTIFACT_OWNER);
+    String content = "content";
+    Action action = new InjectOutputMetadataAction(output, createRemoteMetadata(content));
+
+    // act
+
+    // Not cached.
+    runAction(action);
+    // Cache hit
+    runAction(action);
```
alexjski, line 528: Ideally, we would clean the statistics before this one. Maybe consider making runAction return the token or a boolean whether it was a cache hit -- that would be easy to check.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+
+    // assert
+    assertThat(output.getPath().exists()).isFalse();
+    assertStatistics(1, new MissDetailsBuilder().set(MissReason.NOT_CACHED, 1).build());
+  }
+
+  @Test
+  public void saveOutputMetadata_localMetadataOverwriteRemoteMetadata() throws Exception {
```
alexjski, line 536: nit: overwrites

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+
+    // act
+
+    // Not cached.
+    runAction(action);
+    // Cache hit
+    runAction(action);
+
+    // assert
+    assertThat(output.getPath().exists()).isFalse();
```
alexjski, line 531: Should we assert the output metadata in the entry? Also, how about checking whether the outputs were injected to the metadata handler?

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -78,7 +86,38 @@ public void setupCache() throws Exception {
...
+        RootType.Output,
+        "bin");
+
+  }
+
+  private byte[] digest(byte[] content) {
+    return digestHashFunction.getHashFunction().hashBytes(content).asBytes();
+  }
+
+  private void enableSaveOutputMetadata() {
```
alexjski, line 108: It looks like you should be able to make that into a helper also called from setupCache (with a boolean flag).

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    // assert
+    assertThat(output.getPath().exists()).isFalse();
+    ActionCache.Entry entry = cache.get(output.getExecPathString());
+    assertThat(entry).isNotNull();
+    assertThat(entry.getOutputFile(output)).isNull();
+    assertStatistics(0, new MissDetailsBuilder().set(MissReason.NOT_CACHED, 1).build());
+  }
+
+  @Test
+  public void saveOutputMetadata_notSavedIfNotEnabled() throws Exception {
```
alexjski, line 491: nitnit: NotEnabled -> Disabled

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+
+    // Not cached
+    runAction(action);
+
+    FileSystemUtils.writeContentAsLatin1(output.getPath(), "content2");
+    // Not cached since local file changed
+    runAction(action);
+
+    // assert
+    assertStatistics(
```
alexjski, line 556: assert the entry has no remote files now?

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: I wonder if that would not be simpler if you didn't use `runAction` for this, but just call `getTokenIfNeedToExecute`.

alexjski (changes requested):

src/main/java/com/google/devtools/build/lib/actions/ActionCacheChecker.java
```diff
@@ -234,6 +283,97 @@ protected boolean unconditionalExecution(Action action) {
...
+        // Load remote child file metadata from cache.
+        cachedTreeMetadata
+            .childValues()
+            .forEach(
+                (key, value) -> merged.putChild(TreeFileArtifact.createTreeOutput(parent, key), value));
+        // Or add local one.
+        localTreeMetadata.getChildValues().forEach(merged::putChild);
+
+        Optional<ArchivedRepresentation> archivedRepresentation;
+        if (cachedTreeMetadata.archivedFileValue().isPresent()) {
```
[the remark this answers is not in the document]
coeuvre: Good catch!

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -368,15 +412,12 @@ public void testMiddleman_differentFiles() throws Exception {
...
-                RootType.Output,
-                "bin"),
+            artifactRoot,
             PathFragment.create("bin/dummy"),
-            ActionsTestUtil.NULL_ARTIFACT_OWNER,
+            NULL_ARTIFACT_OWNER,
             SpecialArtifactType.CONSTANT_METADATA);
     output.getPath().getParentDirectory().createDirectoryAndParents();
-    Action action = new NullAction(output);
+    Action action = new WriteEmptyOutputAction(output);
```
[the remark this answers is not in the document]
coeuvre: Previously, outputs are [created](https://github.com/bazelbuild/bazel/blob/2c059126de8e128d9dc1cc8b7eb3d7fe99bde6e6/src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java#L123) with empty content within `runAction`. Since we need to test against the case that outputs are not created but instead metadata are  injected, we move the creation to the action itself.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -457,7 +627,7 @@ public void resetStatistics() {
   }
 
   /** A null middleman action. */
-  private static class NullMiddlemanAction extends NullAction {
+  private static class NullMiddlemanAction extends WriteEmptyOutputAction {
```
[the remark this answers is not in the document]
coeuvre: Changed back to derive from `NullAction`.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+  public void saveOutputMetadata_localFileMetadataNotSaved() throws Exception {
+    // arrange
+    enableSaveOutputMetadata();
+
+    Artifact.DerivedArtifact output =
+        new Artifact.DerivedArtifact(
+            artifactRoot, PathFragment.create("bin/dummy"), NULL_ARTIFACT_OWNER);
+    output.getPath().getParentDirectory().createDirectoryAndParents();
+    FileSystemUtils.writeContentAsLatin1(output.getPath(), "content");
+    Action action = new InjectOutputMetadataAction(output, FileArtifactValue.createForTesting(output));
```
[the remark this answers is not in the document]
coeuvre: Changed to use `WriteEmptyOutputAction`

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: The entry has remote files since we still inject remote metadata during the second execution. That said, I updated the test to assert the metadata in the entry is equals to the metadata injected from the second run.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: Added test for the case where the local metadata is same as injected remote one.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -423,58 +433,119 @@ public synchronized void dump(PrintStream out) {
...
+    Collection<String> files = entry.getPaths();
+
+    // Estimate the size of the buffer:
+    //   5 bytes max for the actionKey length
+    // + the actionKey itself
+    // + 32 bytes for the digest
+    // + 5 bytes max for the file list length
+    // + 5 bytes max for each file id
+    // + 32 bytes for the environment digest
+    int maxSize =
```
alexjski, line 488: Is that still truly max size?

coeuvre: Fixed.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -391,6 +432,135 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+  @Test
+  public void saveOutputMetadata_localFileMetadataNotSaved() throws Exception {
+    // arrange
+    enableSaveOutputMetadata();
+
+    Artifact.DerivedArtifact output =
+        new Artifact.DerivedArtifact(
+            artifactRoot, PathFragment.create("bin/dummy"), NULL_ARTIFACT_OWNER);
+    output.getPath().getParentDirectory().createDirectoryAndParents();
+    FileSystemUtils.writeContentAsLatin1(output.getPath(), "content");
```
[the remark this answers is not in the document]
coeuvre: The problem with `writeIsoLatin1` is it appends newline `\n` (or `\r\n` on windows).

src/main/java/com/google/devtools/build/lib/actions/cache/ActionCache.java
```diff
@@ -84,31 +92,145 @@
...
+      public abstract ImmutableMap<String, RemoteFileArtifactValue> childValues();
+
+      public abstract Optional<RemoteFileArtifactValue> archivedFileValue();
+    }
 
     public Entry(String key, Map<String, String> usedClientEnv, boolean discoversInputs) {
       actionKey = key;
       this.usedClientEnvDigest = MetadataDigestUtils.fromEnv(usedClientEnv);
       files = discoversInputs ? new ArrayList<String>() : null;
       mdMap = new HashMap<>();
```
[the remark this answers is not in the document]
coeuvre: `mdMap` still contains inputs and outputs.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
coeuvre: Nope. For building Bazel itself, the size of action cache increased from 896K to 904K (0.9%). When the store flag is set, the size increased to 1.8M.

coeuvre:
Tests added. I think we can have another review round.

src/test/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCacheTest.java
```diff
@@ -171,6 +188,154 @@ public void testToStringIsntTooBig() {
...
+          Artifact.ArchivedTreeArtifact artifact =
+              Artifact.ArchivedTreeArtifact.create(parent, derivedPathPrefix);
+          builder.setArchivedRepresentation(
+              TreeArtifactValue.ArchivedRepresentation.create(artifact, metadata));
+        });
+    return builder.build();
+  }
+
+  @Test
+  public void putAndGet_saveRemoteFileMetadata() {
```
alexjski, line 230: optionalnitnit: saves, all names like that.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: I was also thinking about the case when an action both injects the remote metadata and stores a file locally. Wouldn't remote execution with download all do that?

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    assertThat(output.getPath().exists()).isFalse();
+    ActionCache.Entry entry = cache.get(output.getExecPathString());
+    assertThat(entry).isNotNull();
+    assertThat(entry.getOutputFile(output)).isNull();
+    assertStatistics(0, new MissDetailsBuilder().set(MissReason.NOT_CACHED, 1).build());
+  }
+
+  @Test
+  public void saveOutputMetadata_remoteFileMetadataLoaded() throws Exception {
+    cacheChecker = createActionCacheChecker(/* storeOutputMetadata= */ true);
```
alexjski, line 527: nit: `/*storeOutputMetadata=*/` (no spaces).

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    Artifact output = createArtifact(artifactRoot, "bin/dummy");
+    String content = "content";
+    Action action = new InjectOutputFileMetadataAction(output, createRemoteFileMetadata(content));
+    MetadataHandler metadataHandler = new FakeMetadataHandler(derivedPathPrefix);
+
+    // Not cached.
+    runAction(action);
+    Token token = cacheChecker.getTokenIfNeedToExecute(
+        action,
+        null,
```
alexjski, line 537: nit: maybe add the `/*bla=*/` for the nulls and empty maps.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    cacheChecker = createActionCacheChecker(/* storeOutputMetadata= */ true);
+    Artifact output = createArtifact(artifactRoot, "bin/dummy");
+    String content = "content";
+    Action action = new InjectOutputFileMetadataAction(output, createRemoteFileMetadata(content));
+
+    // Not cached
+    runAction(action);
+    writeContentAsLatin1(output.getPath(), content);
+    // Cached since local metadata is same as remote metadata
+    runAction(action);
```
alexjski, line 563: nit: would getToken make sense here too?

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    String content2 = "content2";
+    Action action =
+        new InjectOutputFileMetadataAction(
+            output, createRemoteFileMetadata(content1), createRemoteFileMetadata(content2));
+
+    // Not cached
+    runAction(action);
+    writeContentAsLatin1(output.getPath(), content2);
+    // Not cached since local file changed
+    runAction(action);
```
alexjski, line 589: ditto.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    assertThat(entry).isNotNull();
+    assertThat(entry.getOutputTree(output))
+        .isEqualTo(
+            SerializableTreeArtifactValue.create(
+                ImmutableMap.of(), Optional.of(createRemoteFileMetadata("content"))));
+    assertStatistics(0, new MissDetailsBuilder().set(MissReason.NOT_CACHED, 1).build());
+  }
+
+  @Test
+  public void saveOutputMetadata_emptyTreeMetadata_notSaved() throws Exception {
```
alexjski, line 652: This is hardly an important case, but does this mean that empty trees will not get action cache hits? I would think that actions generating those should also be eligible.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: Another option is to check the stats are empty before the second run. To me, the first run is a part of arrange here.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    Action action =
+        new InjectOutputTreeMetadataAction(
+            output,
+            createTreeMetadata(
+                output,
+                ImmutableMap.of(),
+                Optional.of(FileArtifactValue.createForTesting(fileSystem.getPath("/archive")))));
+    fileSystem.getPath("/archive").delete();
+
+    // Not cached.
```
alexjski, line 713: nitnit: I would hope that is obvious for first run.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -466,15 +986,153 @@ public MiddlemanType getActionType() {
...
+        actionExecutionContext.getMetadataHandler().injectFile(output, metadata[index]);
+      }
+      ++times;
+      return super.execute(actionExecutionContext);
+    }
+  }
+
+  private static class InjectOutputTreeMetadataAction extends NullAction {
+    private final SpecialArtifact output;
+    private final TreeArtifactValue[] metadata;
```
alexjski, line 1118: nit: you could consider that as a Queue with and pop on it.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -466,15 +986,153 @@ public MiddlemanType getActionType() {
...
+      super(output);
+
+      this.output = output;
+      this.metadata = metadata;
+    }
+
+    @Override
+    public ActionResult execute(ActionExecutionContext actionExecutionContext) {
+      int index = Math.min(times, metadata.length - 1);
+      if (index >= 0) {
```
alexjski, line 1108: Shouldn't that not happen? If it should, wouldn't it really mean that metadata is empty?

alexjski: Also, should we fail in else?

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    Action action =
+        new InjectOutputTreeMetadataAction(
+            output,
+            createTreeMetadata(output, children1, Optional.empty()),
+            createTreeMetadata(output, children2, Optional.empty()));
+    MetadataHandler metadataHandler = new FakeMetadataHandler(derivedPathPrefix);
+
+    // Not cached
+    runAction(action);
+    writeIsoLatin1(output.getPath().getRelative("file2"), "modified");
```
alexjski, line 774: Given we create that file, does it matter what children2 contains? I think you could make a stronger test case if those didn't happen to be the same as the state on disk.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: Thank you!

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -385,12 +442,475 @@ public void testDeletedConstantMetadataOutputCausesReexecution() throws Exceptio
...
+    Action action =
+        new InjectOutputTreeMetadataAction(
+            output, createTreeMetadata(output, children, Optional.empty()));
+    MetadataHandler metadataHandler = new FakeMetadataHandler(derivedPathPrefix);
+
+    // Not cached
+    runAction(action);
+    writeContentAsLatin1(output.getPath().getRelative("file1"), "content1");
+    // Cache hit
+    runAction(action, metadataHandler);
```
alexjski, line 863: getToken?

src/test/java/com/google/devtools/build/lib/actions/util/ActionsTestUtil.java
```diff
@@ -982,7 +982,7 @@ public void setDigestForVirtualArtifact(Artifact artifact, byte[] digest) {
     }
 
     @Override
-    public TreeArtifactValue getTreeArtifactValue(SpecialArtifact treeArtifact) {
+    public TreeArtifactValue getTreeArtifactValue(SpecialArtifact treeArtifact) throws IOException {
```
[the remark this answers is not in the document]
coeuvre: I added override in the subclass (FakeMetadataHandler) which will throw IOException.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -1157,6 +1157,46 @@ EOF
...
+)
+EOF
+
+  bazel build \
+    --experimental_action_cache_store_output_metadata \
+    --remote_executor=grpc://localhost:${worker_port} \
+    --remote_download_minimal \
+    //a:foobar || fail "Failed to build //a:foobar"
+
+  (! [[ -f bazel-bin/a/foo.txt ]] && ! [[ -f bazel-bin/a/foobar.txt ]]) \
```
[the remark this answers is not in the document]
coeuvre: `bazel-bin/a` will be created by [`SkyframeActionExecutor`](https://github.com/bazelbuild/bazel/blob/a52b1320da06c85d5aaa1ece9eb46acabd09b33f/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java#L1005) since it's the output directory.

Updated the code to check `bazel-bin/a` is empty.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -1157,6 +1157,46 @@ EOF
...
+
+  bazel shutdown
+
+  bazel build \
+    --experimental_action_cache_store_output_metadata \
+    --remote_executor=grpc://localhost:${worker_port} \
+    --remote_download_minimal \
+    //a:foobar >& $TEST_log || fail "Failed to build //a:foobar"
+
+  expect_log "1 process: 1 internal"
```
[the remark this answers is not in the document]
coeuvre: Thanks for the pointer! Updated.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: Remote execution with download all won't inject remote metadata. but I will probably change it to inject remote metadata to support async download. Test case added.

src/test/java/com/google/devtools/build/lib/actions/util/ActionsTestUtil.java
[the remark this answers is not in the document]
alexjski: Then you certainly can't revert that -- sorry!

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: The tree metadata of that empty tree won't be saved but it will get action cache hits. Updated the checks.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: Changed.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
coeuvre: Changed to use a queue

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: Thanks!

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
alexjski: FYI, that is what we do internally.

alexjski (approved):

PR #13606, opened 2021-06-24
Add blurb about json workers resilience

keith:
This was copied from https://blog.bazel.build/2020/11/11/json-workers.html#how-do-i-make-a-worker-that-uses-json


larsrc-google (approved):

meisterT:
Thanks!

Issue #13599, opened 2021-06-23
json persistent workers GA / --experimental_worker_allow_json_protocol removal


keith:
JSON support for persistent workers launched a few months ago https://blog.bazel.build/2020/11/11/json-workers.html I'm curious what the criteria are for `--experimental_worker_allow_json_protocol` to no longer be required so we can use this in rules without having to force that flag on users?

susinmotion:
cc: @larsrc-google 
There are some changes to the worker protocol in progress, and we wanted to wait for those to be in place before flipping, then removing, the flag. Since we're not currently enforcing that JSON workers be tolerant of unknown fields in the work request, we were afraid we'd break JSON workers by changing the protocol.

Other things we considered were adding an arbitrary field to the work request to enforce resilience to unknown fields, then flipping the flag. We can still do this if there's good reason.

meisterT:
I assume the JSON protocol is basically just push a dict over the wire. Unknown keys in the dict should not hurt any worker implementation, unless they specifically check for it.

susinmotion:
Our own test workers weren't resilient to extra fields! :P

I'm comfortable flipping the flag but I defer to Lars's sense of risk.

larsrc-google:
Our own workers aren't actually quite resilient yet, AFAICT.

keith:
Is the thought that at the time of flipping this flag we would never again add fields though? based on the docs I believe that workers _should_ be resilient to this and therefore that doesn't particularly have to affect this flip

larsrc-google:
Oh, there will be more fields. Sandboxing for multiplex workers will require another field. And who knows what the future will bring? I definitely don't want to tie us to a frozen protocol.

aiuto:
I'm going to raise Keith one.   Workers *must* gracefully handle fields
they don't understand.
That should be a requirement for any code in any project that reads
serialized data.

larsrc-google:
Yes, we should be specific about that in the worker docs. We aren't quite.

keith:
https://github.com/bazelbuild/bazel/pull/13606

keith:
I submitted https://github.com/bazelbuild/bazel/pull/13607 in case we decide we're ready for this

wolfd:
[...]

PR #13607, opened 2021-06-24
Remove --experimental_worker_allow_json_protocol

keith:
Closes https://github.com/bazelbuild/bazel/issues/13599

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerOptions.java
@@ -46,16 +46,6 @@ public class WorkerOptions extends OptionsBase {
       })
   public Void experimentalPersistentJavac;
 
-  @Option(
-      name = "experimental_worker_allow_json_protocol",
-      defaultValue = "false",
-      documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
-      effectTags = {OptionEffectTag.BUILD_FILE_SEMANTICS},
-      help =
-          "Allows workers to use the JSON worker protocol until it is determined to be"
-              + " stable.")
-  public boolean experimentalJsonWorkerProtocol;
-
   /**
    * Defines a resource converter for named values in the form [name=]value, where the value is
    * {@link ResourceConverter.FLAG_SYNTAX}. If no name is provided (used when setting a default),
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/devtools/build/lib/worker/WorkerOptions.java
```diff
@@ -46,16 +46,6 @@
       })
   public Void experimentalPersistentJavac;
 
-  @Option(
```
larsrc-google, line 49: You need to move the flag to the graveyard, or any Bazel invocation passing that flag will fail.

keith: my thought here was that since this was an undocumented flag for an alpha feature removing it would be ok, especially if that's in the 5.x release not the 4.x release. Also flipping the default wouldn't change much AFAICT since workers can only support 1 protocol at once, so flipping it back to false wouldn't be super useful for folks, still happy to make that change if you think we should, but that's why I just deleted it

larsrc-google:
It's fair to remove an experimental flag, but I prefer giving users a bit of notice. 

What bothers me about this flag is exactly that there is no way to negotiate with a worker which protocol to use. If a worker implementation changes from one to the other, there's a real risk of protocol mismatch. But then a global flag doesn't help much, since it would also require other workers to support both. 

I'm currently doing a refactoring of the graveyards, I'll flip and graveyard this flag afterwards.

keith:
Yea workers only being able to support 1 is interesting, but I guess the expectation is users shouldn't need to know which protocol a worker uses anyways?

Either way I submitted https://github.com/bazelbuild/bazel/pull/13745 just to do the flip now, I opted not to move it around further given your comment about refactoring the graveyard logic

aiuto:
In retrospect, we should be clearer in the description of this kind of PR.  It does not remove the feature enabled by experimental_worker_allow_json_protocol. It just removes the flag and makes the feature permanent. When looking at revision history, it will be easier to understand if phrased that way.

keith:
Now that this feature is flipped I'm less worried about this flag existing and just doing nothing, there's no reason to turn it off. I assumed at this point we might want to give users a bit just so the flag doesn't start erroring since it was so recently flipped.

aiuto:
I'm beginning a quest to remove as many unneeded flags as possible from Bazel over the next year. We just have too many.  But, that's just like my opinion.

keith:
I create https://github.com/bazelbuild/bazel/pull/14679

PR #13629, opened 2021-06-30
Fix Windows developer mode symlinks

khogeland:
The OS version check in https://github.com/bazelbuild/bazel/pull/13488 breaks the developer mode symlink behavior.

`IsWindowsVersionOrGreater` does not work as advertised, and returns false on Windows 10 if not called from an executable with an associated application manifest declaring its compatibility for Windows 10. (Very cool, Microsoft.) The other methods of checking OS version are far more verbose and complicated, which doesn't seem warranted here.

As an alternative workaround, this PR replaces the ahead-of-time version check with a retry without the flag if the function reports an invalid argument exception.

@meteorcloudy

### DIFF
--- a/src/main/native/windows/file.cc
+++ b/src/main/native/windows/file.cc
@@ -43,12 +43,8 @@ using std::wstring;
 DWORD DetermineSymlinkPrivilegeFlag() {
   DWORD val = 0;
   DWORD valSize = sizeof(val);
-  if (  // The unprivileged create flag was introduced in Windows 10 build
-        // 14972:
-        // https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/
-      !IsWindowsVersionOrGreater(10, 0, 14972)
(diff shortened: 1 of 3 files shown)


src/main/native/windows/file.cc
```diff
@@ -467,6 +463,13 @@ int CreateSymlink(const wstring& symlink_name, const wstring& symlink_target,
 
   if (!CreateSymbolicLinkW(name.c_str(), target.c_str(),
                            symlinkPrivilegeFlag)) {
+    if (GetLastError() == ERROR_INVALID_PARAMETER) {
```
meteorcloudy, line 466: Nit: This means on older Windows versions with developer mode enabled, each symlink operation triggers CreateSymlinkLink twice, which has some performance implication. But I guess it's small issue which won't impact the vast majority of Windows users.

meteorcloudy (approved):

meteorcloudy:
Thanks for the fix!

PR #13630, opened 2021-07-01
Remote: Merge download and downloadMinimal.

coeuvre:
Cleans up the code structure without introducing new features and regressions. Future changes depend on this.

`remoteCache.download` is used to download outputs for remote executed actions and `remoteCache.downloadMinimal` is used to only download metadata of these outputs to save network time. The decision of which method to call is based on remote options.

This change merge these two methods into `remoteExecutionService.downloadOutputs` so that:
1. One step forward to blur the boundary between two download modes: having a single entry point. The ultimate goal is to use one code path for both modes.
2. Open opportunities to implement "remote output streaming": only download metadata within spawn execution and kick off the downloads into background so we don't need to wait for the downloads before starting another action execution.

Tests for these two methods are also updated:
1. Rename tests following the `testedMethodName_input_expectedOutcome` pattern.
2. Remove some duplicated tests in `GrpcCacheClientTest`.


coeuvre:
Rebased.

PR #13633, opened 2021-07-02
Replace rbe_autoconfig with configuration files generated by rbe_configs_gen

coeuvre:
`rbe_autoconfig` from [`bazel_toolchain`](https://github.com/bazelbuild/bazel-toolchains) was used to generate toolchain configurations to execute actions remotely on RBE. With the release [`5.0.0`](https://github.com/bazelbuild/bazel-toolchains/releases/tag/v5.0.0), `rbe_autoconfig` is deprecated in favor of the new CLI tool `rbe_configs_gen`.

This PR removes the calls to `rbe_autoconfig` and adds the pre-generated toolchains using `rbe_configs_gen`. A shell script `configs/rbe/gen.sh` which is used for the generation is also added.


coeuvre:
It works fine with `rbe_ubuntu1804_java11`, but has the following error on `rbe_ubuntu1604_java8`:

```
(07:49:11) ERROR: /workdir/src/main/tools/BUILD:3:10: Linking src/main/tools/daemonize failed: (Exit 1): clang failed: error executing command
  (cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/execroot/io_bazel && \
  exec env - \
    PATH=/bin:/usr/bin:/usr/local/bin \
    PWD=/proc/self/cwd \
  /usr/bin/clang @bazel-out/k8-fastbuild/bin/src/main/tools/daemonize-2.params)
Execution platform: //:rbe_ubuntu1604_java8_platform
clang: error: invalid linker name in argument '-fuse-ld=/usr/bin/ld.gold'
(07:49:11) ERROR: /workdir/src/BUILD:301:9 Executing genrule //src:package-zip_jdk_minimal failed: (Exit 1): clang failed: error executing command
  (cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/execroot/io_bazel && \
  exec env - \
    PATH=/bin:/usr/bin:/usr/local/bin \
    PWD=/proc/self/cwd \
  /usr/bin/clang @bazel-out/k8-fastbuild/bin/src/main/tools/daemonize-2.params)
Execution platform: //:rbe_ubuntu1604_java8_platform
```

coeuvre:
In order to use custom execution properties, I added line 
```
bazel_toolchains_repositories()
```

to `WORKSPACE` which introduce other dependencies e.g. `rules_docker` and the "bootstrap from distfile" test failed with following error:

```
Error in fail: error running 'git fetch origin refs/heads/*:refs/remotes/origin/* refs/tags/*:refs/tags/*' while working with @io_bazel_rules_docker:
fatal: unable to access 'https://github.com/bazelbuild/rules_docker.git/': Could not resolve host: github.com; Unknown error
```

coeuvre:
Another problem needs to be addressed is that: we need `create_rbe_exec_properties_dict` from `bazel_toolchains` to define custom execution properties but it seems that release 5.0.0 doesn't include that. How do we use `create_rbe_exec_properties_dict` with Bazel 5.0.0?

smukherj1:
A few notes:
1. Use bazel-toolchains release [5.1.0](https://github.com/bazelbuild/bazel-toolchains/releases/tag/v5.1.0) instead of 5.0.0. Also, given you own the tool now, you could also pin to the HEAD commit on master to pull in any patches/fixes needed to get this integration working.
2. `create_rbe_exec_properties_dict` is available in 5.1.0 [here](https://github.com/bazelbuild/bazel-toolchains/blob/27f2db256e54e5748ee1cd9485ccd0d5444bf1c6/rules/exec_properties/exec_properties.bzl#L205). You should be able to import this bzl file directly after importing bazel-toolchains. You shouldn't need to invoke `bazel_toolchains_repositories()` because the dependencies like rules_docker imported by this macro aren't needed for `create_rbe_exec_properties_dict`.
3. Regarding using gcc, clang is currently the hardcoded compiler for Linux specified [here](https://github.com/bazelbuild/bazel-toolchains/blob/v5.1.0/pkg/rbeconfigsgen/options.go#L129) (toolchain constraint for the generated execution platform) and a configurable set of options [here](https://github.com/bazelbuild/bazel-toolchains/blob/v5.1.0/pkg/rbeconfigsgen/options.go#L141) (environment variable that control Bazel's internal C++ toolchain autodetection machinery). 

For (3) above, do the following in the short term:

1. Override the `execution_constraints` in the custom platform inherited from the auto-generated platform and specify gcc instant of clang.
2. Custom C++ config generation env variables can be specified via a JSON file and accepted by the rbe_configs_gen tool via the cmd line flag [here](https://github.com/bazelbuild/bazel-toolchains/blob/v5.1.0/cmd/rbe_configs_gen/rbe_configs_gen.go#L47). I'm not sure what the equivalent values for gcc are though.

coeuvre:
Thanks! It works now.

coeuvre:
@philwo This PR is ready to be reviewed. The generated configuration files are under `//configs/rbe/$TOOLCHAIN_NAME`. Is that an appropriate place or will it causes problems when importing?

configs/rbe/gen.sh
```diff
@@ -0,0 +1,78 @@
...
+#
+# The script is used to generate rbe configurations.
+#
+
+set -eu
+
+cd $(dirname $BASH_SOURCE)/../..
+BASEDIR=configs/rbe
+
+BAZEL_VERSION=4.1.0
```
philwo, line 25: What is this version number used for? The configs checked into the `master` branch should probably be built using the latest Bazel rolling release instead of the latest LTS version.

philwo:
> The generated configuration files are under //configs/rbe/$TOOLCHAIN_NAME. Is that an appropriate place or will it causes problems when importing?

I think that's fine!

Could you add some short documentation maybe to a `configs/README.md` file about the structure of the files (e.g. which files are generated by which tool, which are manually maintained) and how and when to regenerate the configs (e.g. do we have to do it everytime we publish Docker images? How do we handle Bazel version upgrades?).

configs/rbe/gen.sh
coeuvre: The tool will pass this number to `bazelisk` when auto-detecting toolchains in the container. I think the version should match which we use on Bazel CI.

Using rolling release will not re-generate the configuration unless we manually re-run the tool again.

coeuvre:
Added `configs/rbe/README.md`

philwo (approved):

coeuvre:
Please check #13981 instead. Closing.

PR #13641, opened 2021-07-02
Ignore try import failure on Windows

andrewkatson:
Fixes:

https://github.com/bazelbuild/bazel/issues/13443#event-4951673487

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -387,8 +387,13 @@ static vector<string> GetServerExeArgs(const blaze_util::Path &jvm_path,
   set<string> java_library_paths;
   std::stringstream java_library_path;
   java_library_path << "-Djava.library.path=";
+  std::string errorText;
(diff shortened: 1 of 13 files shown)


meteorcloudy:
Thanks!  Can you reference the issue report in your PR description?

meteorcloudy:
I wonder if there is a more clean solution, this looks like a lot of changes.

andrewkatson:
Im open to suggestions of how to make it cleaner. Not sure how though.

meteorcloudy:
I guess you can try to deal with the error in ConvertPath instead of calling BAZEL_DIE unconditionally: https://cs.opensource.google/bazel/bazel/+/master:src/main/cpp/util/path_windows.cc;l=73

See Lukacs's suggestion here: https://github.com/bazelbuild/bazel/issues/13443#issuecomment-855765933

andrewkatson:
This is invalid and I will open a better one

PR #13643, opened 2021-07-03
Support .def file on cc_common.link

jlaxson:
This allows rules to use the same api and crosstool support as the cc_binary windows_def_file


sgowroji:
Helloi @jlaxson, Could you please resolve the above branch conflicts. Thanks!

Issue #12510, opened 2020-11-18
Crash when using execution log and flaky_test_attempts if test fails


mariusgrigoriu:
### Description of the problem / feature request:

Bazel crashes when executing a failing test with both `flaky_test_attempts` > 1 and either execution log flags.

```
$ bazel test :always_fail  --execution_log_json_file=log.json --flaky_test_attempts=2
INFO: Analyzed target //repro:always_fail (28 packages loaded, 325 targets configured).
INFO: Found 1 test target...
FAIL: //repro:always_fail (see /private/var/tmp/_bazel_marius/c39204db2cb7e0a79ba2096a64d88d76/execroot/platform_bootstrap/bazel-out/darwin-fastbuild/testlogs/repro/always_fail/test_attempts/attempt_1.log)
FAIL: //repro:always_fail (see /private/var/tmp/_bazel_marius/c39204db2cb7e0a79ba2096a64d88d76/execroot/platform_bootstrap/bazel-out/darwin-fastbuild/testlogs/repro/always_fail/test.log)

FAILED: //repro:always_fail (Summary)
      /private/var/tmp/_bazel_marius/c39204db2cb7e0a79ba2096a64d88d76/execroot/platform_bootstrap/bazel-out/darwin-fastbuild/testlogs/repro/always_fail/test.log
      /private/var/tmp/_bazel_marius/c39204db2cb7e0a79ba2096a64d88d76/execroot/platform_bootstrap/bazel-out/darwin-fastbuild/testlogs/repro/always_fail/test_attempts/attempt_1.log
Target //repro:always_fail up-to-date:
  bazel-bin/repro/always_fail
INFO: Elapsed time: 2.406s, Critical Path: 0.39s
INFO: 3 processes: 4 darwin-sandbox.
INFO: Build completed, 1 test FAILED, 3 total actions
//repro:always_fail                                                      FAILED in 2 out of 2 in 0.2s
  Stats over 2 runs: max = 0.2s, min = 0.1s, avg = 0.1s, dev = 0.1s
  /private/var/tmp/_bazel_marius/c39204db2cb7e0a79ba2096a64d88d76/execroot/platform_bootstrap/bazel-out/darwin-fastbuild/testlogs/repro/always_fail/test.log
[...]

PR #13650, opened 2021-07-06
Allow multiple outputs in SpawnLog

styurin:
As reported in https://github.com/bazelbuild/bazel/issues/12510 there might be multiple spawns with the same output when flaky tests are executed multiple times.

The fix is to remove the check for presence of duplicate outputs and to extend the sorting algorithm to accept multiple outputs with the same name.


styurin:
@meisterT are there any issues with this PR?

meisterT (approved):

Issue #13050, opened 2021-02-16
The incremental apk generated by the `android_binary` rule isn't compliant with Android 11 behavior changes


veganafro:
### Description of the problem / feature request:

Running `bazel mobile-install //app:app_binary --start_app` fails when installing an incremental apk on a device running Android 11. This only happens when `targetSdkVersion="30"` in my `AndroidManifest.xml`.

The exact error output is
```
Error:
adb command: ['external/androidsdk/platform-tools/adb', 'install', '-r', './bazel-out/darwin-fastbuild/bin/app/src/app_binary_incremental.apk']
return code: 1
stdout: Performing Streamed Install
stderr: adb: failed to install ./bazel-out/darwin-fastbuild/bin/app/src/app_binary_incremental.apk: Failure [-124: Failed parse during installPackageLI: Targeting R+ (version 30 and above) requires the resources.arsc of installed APKs to be stored uncompressed and aligned on a 4-byte boundary]
```

I believe the error could be related to [this](https://developer.android.com/about/versions/11/behavior-changes-11#compressed-resource-file) behavior change for apps targeting Android 11. But, there might also be a case for the error being related to how incremental apks are generated.

I'm able to manually run `adb install bazel-out/darwin-fastbuild/bin/app/src/app_binary.apk` without issue. However, `bazel mobile-install` runs `adb install -r bazel-out/darwin-fastbuild/bin/app/src/app_binary_incremental.apk` which results in a failure.

Also, running `zipalign -c -v 4 bazel-out/darwin-fastbuild/bin/app/src/app_binary_incremental.apk` outputs a `Verification FAILED` message.

### Feature requests: what underlying problem are you trying to solve with this feature?

I'd like to be able to install, run, and test apps targeting sdk 30

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create a sandbox android app with the following configurations:

## WORKSPACE

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

## BEGIN build_bazel_rules_android ##

RULES_ANDROID_VERSION = "0.1.1"
RULES_ANDROID_SHA = "cd06d15dd8bb59926e4d65f9003bfc20f9da4b2519985c27e190cddc8b7a7806"

http_archive(
    name = "build_bazel_rules_android",
    sha256 = RULES_ANDROID_SHA,
    strip_prefix = "rules_android-%s" % RULES_ANDROID_VERSION,
[...]

PR #13653, opened 2021-07-07
Zipalign the incremental APK before install

Bencodes:
This addresses https://github.com/bazelbuild/bazel/issues/13050 which prevents the incremental APK from being installed on Android 11 devices due to the fact that the contents of the APK are not aligned on a 4-byte boundary.


Bencodes:
@ahumesky mind taking a look at this so we can get it cherry picked into the 4.2.0 release?

Bencodes:
Split APK installs are also broken https://github.com/bazelbuild/bazel/pull/13718

ahumesky:
Thanks, I'll start the import on this fix

ahumesky:
Unfortunately this change is failing an internal test, and in somewhat surprising way, so it might take a little while to debug.
And same for https://github.com/bazelbuild/bazel/pull/13718

Bencodes:
@ahumesky have you been able to root cause that internal test failure?

Issue #13632, opened 2021-07-01
Cache uploads shouldn't block dependent local actions


brentleyjones:
### Description of the problem / feature request:

Currently cache uploads take place as the last part of a local action spawn. This ties up the `local_cpu_resources`, and also prevents dependent local action spawns from starting (even though they have the inputs they need now).

Instead, the upload of local actions outputs to the cache should be independent spawns, that don't count towards `local_cpu_resources`, and don't block dependent local actions. Remote actions that need the output as input (found via `FindMissingBlobs`) should still wait for the upload to finish, including the already started upload with the other inputs in `RemoteExecutionCache.ensureInputsPresent()`.

### Feature requests: what underlying problem are you trying to solve with this feature?

By doing this, a non-remote build can complete a lot faster, as it no longer has to wait for uploads to finish to move onto the next action. These uploads will no longer take up `local_cpu_resources`, allowing for more parallel local actions (as long as `--jobs` is set higher than `--local_cpu_resources`). The end of the build can report that it's waiting for uploads to finish, similar to what it does today for BES and Profile uploading.

### What operating system are you running Bazel on?

macOS 11.4

### What's the output of `bazel info release`?

4.1.0 and HEAD

###  Have you found anything relevant by searching the web?

- https://github.com/bazelbuild/bazel/issues/7664
- https://github.com/buildbuddy-io/buildbuddy-cli/blob/171585fc6a5af723f418c935990e97909cd62f8f/cache_proxy/cache_proxy.go#L37-L44
- https://github.com/Asana/bazels3cache#main-features

coeuvre:
Thanks for bringing this up! I will look into it!

BTW, I had similar idea recently but for downloads i.e. downloading the outputs of a remote action shouldn't block dependent remote actions. https://github.com/bazelbuild/bazel/pull/13630 is working towards this feature.

coeuvre:
Fixed by #13655.

PR #13655, opened 2021-07-08
Remote: async upload

coeuvre:
When remote cache is enabled, the outputs of locally executed actions are uploaded to the remote cache. However, the uploads take place as the part of a local action spawn which prevents dependent actions from starting .

This change introduces flag `--experimental_remote_cache_async` flag, which when enabled, makes the uploads happen in the background. Remote actions that need the output as input are still wait for the upload to finish, including the already started upload with the other inputs.

Uploads that are still active after the build are waited to finish with message like:

```
INFO: Build completed successfully, 5001 total actions
Waiting for remote cache: 4911 uploads; 16s
```

This change also de-duplicates calls to FindMissingBlobs as well as file uploads to the remote cache. This is effectively the same as https://github.com/bazelbuild/bazel/pull/13166 but with different implementation.

Fixes https://github.com/bazelbuild/bazel/issues/13632.


brentleyjones:
Thanks for this @coeuvre!

brentleyjones:
How does this interact with BES, and in particular `--bes_upload_mode`?

coeuvre:
BES uses `ByteStreamUploader` to upload local files while RE uses `RemoteCache` which uses `ByteStreamUploader` under the hood.

`ByteStreamUploader` can deduplicate file uploads. This PR updates `RemoteCache` to deduplicate file uploads as well. Technically, they share the same upload instance. But there are some overheads here:

```
upload 1 (RE) ----
                 |---- RemoteCache ---- upload a ----
upload 2 (RE) ----                                  |
                                                    |---- ByteStreamUploader ---- upload ---- Server
                                                    |
upload b (BES) --------------------------------------
```

I will make followup PRs to update BES to use RemoteCache directly and remove the deduplication logic in `ByteStreamUploader` (as well as improvements for #11473).

RE doesn't read `--bes_upload_mode` and behaves like `wait_for_upload_complete`. Adding support for that is not complex but I'm not sure whether it's needed.

brentleyjones:
> I will make followup PRs to update BES to use RemoteCache directly and remove the deduplication logic in ByteStreamUploader (as well as improvements for #11473).

I _really_ look forward to fixes for #11473, in particular to having it respect `no-remote`/`no-cache` tags. Currently I have to set  `--experimental_build_event_upload_strategy=local` when I really want it to be `--experimental_build_event_upload_strategy=remote`.

philwo (approved):
If possible, could you split this into smaller parts during import? 😊

coeuvre:
Imported.

@brentleyjones You can give it a try by setting `--experimental_remote_cache_async`. Feedback is welcomed :)

brentleyjones:
@coeuvre This seems to work as advertised!

Something that confused me though: at the end of a build, I saw the number of items waiting to be uploaded increase as other ones finished? For example, I saw this:

```
Waiting for remote cache: 87 uploads, 11s
```

then after some finished, and the count go down, I saw it increase

```
Waiting for remote cache: 111 uploads
```

that doesn't make sense to me?

---

Also, a small UX feedback: in the profile.json, `upload outputs` now shows for a very small bit, and doesn't actually represent the upload anymore. I think it would be nice if the uploads appeared in their own swimlanes, though maybe the profile isn't the right place for that sort of thing, since it can end before uploads are finished now as well (same with BES uploads, which aren't shown at all).

brentleyjones:
Ohh, another thought: this behaves like `--bes_upload_mode= wait_for_upload_complete`. I believe there is value in a version that behaves like `nowait_for_upload_to_complete` and/or `fully_async`, to prevent the end of the build from being blocked by the uploads (for the same reasons one might adjust the `--bes_upload_mode`).

coeuvre:
Thanks for the feedback!

> then after some finished, and the count went down, I saw it increase

The steps of an async upload task are:
  1. Call `findMissingDigests` to find which files to upload.
  2. For each missing file/blob:
      1. Start uploading the content and increase the counter.
      2. Decrease the counter after finished.
  3. After all files/blobs are uploaded:
      1. Start uploading action result and increase the counter.
      2. Decrease the counter after finished.

After the build finished, there could be multiple upload tasks running in the background and some of them may still waiting for the result of `findMissingDigests`. Once those tasks get the results, the counter will be increased depends on how many missing files/blobs.

That being said, I understand that this may create confusion to users. One way to improve is before calling `findMissingDigests`, we can increase the counter for all the outputs. Decrease counter for files that are already existed in the remote cache immediately after `findMissingDigests` returned. And decrease counter for files that are uploaded after upload finished.

> Also, a small UX feedback: in the profile.json, upload outputs now shows for a very small bit, and doesn't actually represent the upload anymore. I think it would be nice if the uploads appeared in their own swimlanes, though maybe the profile isn't the right place for that sort of thing, since it can end before uploads are finished now as well (same with BES uploads, which aren't shown at all). If not, maybe it should have a different name to avoid confusion with the existing blocking upload behavior.

For `profile.json`, we should probably only keep `upload outputs` for blocking upload. It would nice if we can show uploads (and other background tasks) in their own swimlanes (the format actually supports it with [Async Events](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.jh64i9l3vwa1)) but requires some fundamental changes to the profile code. I can look into it later. 

> Ohh, another thought: this behaves like --bes_upload_mode=wait_for_upload_complete. I believe there is value in a version that behaves like nowait_for_upload_to_complete and/or fully_async, to prevent the end of the build from being blocked by the uploads (for the same reasons one might adjust the --bes_upload_mode).

Noted.

brentleyjones:
@coeuvre Do we need a new issue to track the request for background (post-build) uploads?

coeuvre:
Yes, please help create a new issue to better track that (and other features you think are missing)!

PR #13656, opened 2021-07-08
Check for explicit dead states in VerifyServerProcess

bduffany:
`VerifyServerProcess` should return false if the server is dead. The current logic checks whether the start time can be read from `/proc/[pid]/stat`. However, that file can exist while the server is a zombie or dead. This PR adds a check for explicit "dead" states and treats the server as dead if so.

### DIFF
--- a/src/main/cpp/blaze_util_linux.cc
+++ b/src/main/cpp/blaze_util_linux.cc
@@ -161,8 +161,15 @@ string GetSystemJavabase() {
   return blaze_util::Dirname(blaze_util::Dirname(javac_dir));
 }
 
+// Relevant entries from /proc/[pid]/stat.
+// Run `man proc` and search "/proc/[pid]/stat"
+struct ProcessStatus {
+  string state;
+  string start_time;
+};
+
 // Called from a signal handler!
-static bool GetStartTime(const string& pid, string* start_time) {
+static bool GetProcessStatus(const string& pid, ProcessStatus* status) {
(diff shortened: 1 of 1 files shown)


src/main/cpp/blaze_util_linux.cc
```diff
@@ -161,8 +161,17 @@ string GetSystemJavabase() {
...
+struct ProcessStatus {
+  // Process state. See "(3) state" in `man proc`.
+  string state;
+  // Start time since startup. See "(22) starttime" in `man proc`.
+  string start_time;
+};
+
 // Called from a signal handler!
-static bool GetStartTime(const string& pid, string* start_time) {
+static bool GetProcessStatus(const string& pid, ProcessStatus* status) {
```
bduffany, line 174: I wasn't sure whether it made more sense to have this just return `false` if the process if the zombie, instead of modifying it to return a struct. Perhaps someone more familiar with this code could comment.

michajlo:
Looks like there was another attempted mitigation (https://github.com/buildbuddy-io/buildbuddy/pull/682). Did that work out? I like the idea of detecting that the process is in a zombie state, but I'm wary of treating it the same as the process not existing since it could lead to surprising leaks.

sgowroji:
Hello @bduffany, Could you please respond if you are still looking to submit the above PR. And can you reply on the above [comment](https://github.com/bazelbuild/bazel/pull/13656#issuecomment-876761871). Thanks!

PR #13660, opened 2021-07-09
Use correct exit code on invalid aquery --output

illicitonion:
Currently this exits 0, which is dangerous - scripts may assume a query
succeeded with no results, where actually it never ran.

Instead, properly exit with exit code 2.


joeleba (approved):

joeleba:
Thanks for the PR!

PR #13666, opened 2021-07-12
Add support for clang's libc++ to Unix toolchain

### LINKED ISSUE #13071: System headers not found when using clang with libc++

Building C++ using clang and libc++ results in standard system headers not being found. For example:
```
In file included from external/com_google_protobuf/src/google/protobuf/compiler/
main.cc:31:                                                                     
external/com_google_protobuf/src/google/protobuf/compiler/cpp/cpp_generator.h:40:10: fatal error: 'string' file not found
#include <string>
```

1. Install clang-11 and libc++ (e.g. via Debian packages `clang-11`, `libc++-11-dev`, `libc++abi-dev`)
1. Create a cc_binary target with a C++ source file which includes a standard library system header.
1. Create a `.bazelrc` file in the workspace root with options to enable C++ compilation with clang and libc++ (see below).
1. Build the target using `bazel build`.

`.bazelrc`:
  ```
  build --repo_env='CC=clang'
  build --repo_env='BAZEL_CXXOPTS=-std=c++17:-stdlib=libc++'
  build --repo_env='BAZEL_LINKLIBS=-lc++:-lm'
  ```

Linux (Debian Bullseye)

release 4.0.0

Yes. Relevant information on how to enable clang + libc++ as well as how auto-configure works for cc_toolchain_config.

- https://github.com/bazelbuild/bazel/issues/6926
- https://github.com/envoyproxy/envoy/pull/5563

I suspect the issue is with the combination of where clang lists its include path for libc++ headers and the the `-no-canonical-prefixes` compiler flag added by `unix_cc_configure`. For example, compare the include paths specified by clang using the following commands:

`clang -E -xc++ - -v -stdlib=libc++ -std=c++17`:
```
ignoring nonexistent directory "/include"
#include "..." search starts here:
 /usr/lib/llvm-11/bin/../include/c++/v1
 /usr/local/include
 /usr/lib/llvm-11/lib/clang/11.0.1/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
```

`clang -E -xc++ - -v -stdlib=libc++ -std=c++17 -no-canonical-prefixes`:
```
ignoring nonexistent directory "/include"
#include "..." search starts here:
 /usr/local/include
 /usr/lib/clang/11.0.1/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list
```

The path ` /usr/lib/llvm-11/bin/../include/c++/v1` which resolves to `/usr/lib/llvm-11/include/c++/v1` is missing from the latter. Indeed, dropping `-no-canonical-prefixes` from the `clang` command that's shown when using the `--sandbox_debug` flag results in successful compilation.

fmeum:
Due to two separate issues, it is currently not possible to compile and link against clang's libc++ with Bazel's auto-configured Unix toolchain, e.g. via adding `-stdlib=libc++` as a `cxxopt` and `linkopt` on a `cc_binary`:

1. Since Bazel uses `-no-canonical-prefixes`, clang is not able to find the libc++ headers if it is invoked through a symlink in a different directory. This is fixed by fully resolving the path to the clang binary in the toolchain. 

2. The list of built-in include paths does not contain the base directory of the libc++ headers, which makes Bazel's implicit dependency checker fail on targets with the `cxxopt` `-stdlib=lib++`. This is fixed by adding the search paths obtained from `clang -v` with that option to the list of built-in include paths in the toolchain.

Both 1. and 2. can be partially worked around by passing in `-stdlib=libc++` and a fully resolved `CC` via `--repo_env`. However, the former means that the choice of standard library has to be fixed for the entire workspace and the latter cannot be accomplished with a system-independent entry in `.bazelrc`. Proper fixes in Bazel thus seem vastly superior.

Fixes #13071

### DIFF
--- a/tools/cpp/unix_cc_configure.bzl
+++ b/tools/cpp/unix_cc_configure.bzl
@@ -307,7 +307,15 @@ def _find_generic(repository_ctx, name, env_name, overriden_tools, warn = False,
     return result
 
 def find_cc(repository_ctx, overriden_tools):
-    return _find_generic(repository_ctx, "gcc", "CC", overriden_tools)
+    cc = _find_generic(repository_ctx, "gcc", "CC", overriden_tools)
+    if _is_clang(repository_ctx, cc):
+        # If clang is run through a symlink with -no-canonical-prefixes, it does
+        # not find its own include directory, which includes the headers for
+        # libc++. Resolving the potential symlink here prevents this.
+        result = repository_ctx.execute(["readlink", "-f", cc])
+        if result.return_code == 0:
+            return result.stdout.strip()
+    return cc
 
 def configure_unix_toolchain(repository_ctx, cpu_value, overriden_tools):
     """Configure C++ toolchain on Unix platforms."""
@@ -333,7 +341,7 @@ def configure_unix_toolchain(repository_ctx, cpu_value, overriden_tools):
     repository_ctx.file("tools/cpp/empty.cc", "int main() {}")
     darwin = cpu_value.startswith("darwin")
 
-    cc = _find_generic(repository_ctx, "gcc", "CC", overriden_tools)
+    cc = find_cc(repository_ctx, overriden_tools)
     is_clang = _is_clang(repository_ctx, cc)
     overriden_tools = dict(overriden_tools)
     overriden_tools["gcc"] = cc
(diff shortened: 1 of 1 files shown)


fmeum:
@oquenchil Is this something you would be willing and have the time to look into?

oquenchil:
Thanks!

This looks good to me. I don't see any drawbacks or any ways this change could break someone. Do you?

fmeum:
It may surprise some users that compile commands no longer start with something memorable such as `/usr/bin/clang` but the resolved target of that symlink, but I don't see how this could break anything.

oquenchil (approved):
LGTM

Issue #13558, opened 2021-06-07
Release 4.2 - June 2021


philwo:
# Status of Bazel 4.2

This release will use Bazel 4.1.0 as its baseline and we will apply selected cherry-picks and backports on top of it. Please request cherry-picks that you'd like to get into Bazel 4.2.0 here via a comment.

- Expected release date: June 2021
- [List of release blockers](https://github.com/bazelbuild/bazel/labels/Release%20blocker)

To report a release-blocking bug, please file a bug using the `Release blocker` label, and cc me.

Task list:

- [x] Pick release baseline: https://github.com/bazelbuild/bazel/commit/c4975efdd7ede7b46637bf353209d9ac371181a5
- [x] Create release candidate: https://releases.bazel.build/4.2.0/rc1/
- [x] Check downstream projects: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2105
- [x] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [x] Send for review the release announcement PR: https://github.com/bazelbuild/bazel-blog/pull/258
- [x] Push the release, notify package maintainers: ...
- [x] Update the documentation
- [x] Push the blog post
- [x] Update the [release page](https://github.com/bazelbuild/bazel/releases/tag/4.2.0)

philwo:
### Requested cherry picks

**Unreviewed:**
- [ ] https://github.com/bazelbuild/bazel/commit/8ba48ad869006d87abc84e2f699fbc59eb8470f8 in https://github.com/bazelbuild/bazel/issues/13099#issuecomment-827091633
- [ ] https://github.com/bazelbuild/bazel/commit/48eee8b4b447a2ad11df28dd81a2ccb65562b5f5 in https://github.com/bazelbuild/bazel/issues/13099#issuecomment-827164697
- [ ] https://github.com/bazelbuild/bazel/commit/763dd0ce6e1644bf895231432f616427a11d385a in https://github.com/bazelbuild/bazel/issues/13099#issuecomment-830266538
- [ ] https://github.com/bazelbuild/bazel/commit/c22855857d3fb6c13ef70c48710793bbc4f195ea in https://github.com/bazelbuild/bazel/issues/13099#issuecomment-830509694
[...]

PR #13678, opened 2021-07-13
[4.2.0 Cherry-pick] Fixes incorrect install names on darwin platforms

csmulhern:
JUSTIFICATION:
--incompatible_macos_set_install_name is broken without this change. This cherry-pick applied cleanly without modifications.

ORIGINAL COMMIT:
b06f495

RELEASE NOTES:
* Fixes --incompatible_macos_set_install_name not correctly setting install names.

OLD COMMIT MESSAGE:

#12304 added support to bazel for setting install names for dynamic
libraries on darwin platforms. This would set LC_ID_DYLIB to
@rpath/{library_name}, so that RPATH would be used to locate these
libraries at runtime. However, the code was using a utility method that
assumed the library name was mangled, which is often not the case. Given
that the output path should already have been determined with the
mangled or unmangled name, we should be able to just use the base name
of the artifact. The test that was added in #12304 has been updated to
actually use dynamic libaries, and passes with the changes made in this
commit.

Closes #13427.

PiperOrigin-RevId: 377504015

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
@@ -872,8 +872,7 @@ public class CppLinkActionBuilder {
               getLinkType().linkerOrArchiver().equals(LinkerOrArchiver.LINKER),
               configuration.getBinDirectory(repositoryName).getExecPath(),
               output.getExecPathString(),
-              SolibSymlinkAction.getDynamicLibrarySoname(
-                  output.getRootRelativePath(), /* preserveName= */ false),
+              output.getRootRelativePath().getBaseName(),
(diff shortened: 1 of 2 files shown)


csmulhern:
Linking to the 4.2.0 Release Issue (#13558) for tracking purposes.

CC: @katre.

katre:
Looks good, thanks. I'll merge this when the tests pass.

katre (approved):

katre:
I spoke too soon: looks like the tests are failing. Please investigate and let me know when this is ready.

csmulhern:
The issue with the failing test is that a cc_library without linkshared=1 (libfoo in the test) is being dynamically linked, and then the dylib is not found at runtime due to an incorrect install name. It seems that:

1) There was a change made after the cut that changed the linking behavior of libfoo in the test case.
2) Install names are still not being set correctly in some cases.

Given the above, let's skip the cherry-pick.

PR #13679, opened 2021-07-14
avoid "too many open files" when copying from remote cache to local cache

jablin:
If you fetch a build result from the remote cache and use a local disk cache at the same time, Bazel copies the remote cache entry to the local disk cache. Some strange work loads (probably triggered by build steps with > 1k output files) trigger
the error "too many open files".

This is a cherry-pick of the original commit 5d7df56b88.

##### Fix
Re-use the existing class `LazyFileOutputStream` in `DiskAndRemoteCacheClient.java`

##### Work-Around on Linux
 increase `DefaultLimitNOFILE` in `/etc/systemd/system.conf` and reboot

Resolves https://github.com/bazelbuild/bazel/issues/13435


coeuvre:
Can we close this since the commit is included in #13686?

jablin:
> Can we close this since the commit is included in #13686 <https://github.com/bazelbuild/bazel/pull/13686> ?

Seems OK. Caveat: I don't understand the integration process. Newbie.

Issue #7668, opened 2019-03-07
Documentation for @bazel_tools//src/conditions


laurentlb:
@laszlocsomor used this technique in a bzl file to test if we're on Windows:

```python
        is_windows = select({
            "@bazel_tools//src/conditions:host_windows": True,
            "//conditions:default": False,
        }),
```

I didn't find anything about `@bazel_tools//src/conditions` in the documentation.

laszlocsomor:
I used this pattern in https://github.com/bazelbuild/bazel-skylib/pull/122 and https://github.com/bazelbuild/bazel-skylib/pull/123.

yogurtearl:
Related: 

https://bazel.build/docs/configurable-attributes

https://github.com/bazelbuild/bazel/blame/364b77ed34afcd07d48313472e8fe3b260a8312f/src/conditions/BUILD

aiuto:
Obsolete. We should be removing this in favor of the `@platforms` repo.

PR #13699, opened 2021-07-16
Add config settings for x86_32, x86_64, and any_x86

GMNGeoffrey:
AFAICT individual constraint_values can't be used in a config setting
group, which makes it difficult to specify disjunction with them. I'm
adding this to our build files, but it just seems generally useful.

### DIFF
--- a/src/conditions/BUILD
+++ b/src/conditions/BUILD
@@ -174,6 +174,27 @@ config_setting(
 )
 
 config_setting(
+    name = "x86_32",
+    constraint_values = ["@platforms//cpu:x86_32"],
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "x86_64",
+    constraint_values = ["@platforms//cpu:x86_64"],
+    visibility = ["//visibility:public"],
+)
+
+selects.config_setting_group(
+    name = "any_x86",
+    match_any = [
+        ":x86_32",
+        ":x86_64",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
     name = "remote",
     values = {"define": "EXECUTOR=remote"},
     visibility = ["//visibility:public"],


aiuto:
Is there a particular need for this? AFAIK, we are not planning to ever try to build Bazel for x32.

katre:
I agree with @aiuto: I don't see a place where we'll use these in Bazel itself, and so we shouldn't add them.

Possibly putting this into a gist or the examples repo might make sense, however.

GMNGeoffrey:
Maybe I'm misunderstanding, but wouldn't these be for cases in which Bazel is being used to build something else for x86-32 (so speaking of the target platform)? The documentation for what src/conditions is supposed to cover is pretty much non-existent (https://github.com/bazelbuild/bazel/issues/7668), so maybe there's some understood goal for it that I'm not aware of, but it's used quite a bit in the wild as a place to get pre-defined config settings.

katre:
The `//src/conditions` package is internal to Bazel's own repository: we don't provide it to other projects that use Bazel. Things in the `tools` directory do get built into Bazel's `@bazel_tools` repo, which _is_ available to other projects.

The fact that most of `//src` is exported into `@bazel_tools` is, I think, a mistake, and one that we shouldn't contribute to. So since this isn't needed for Bazel's own build, I think it should be somewhere else.

aiuto:
configuration settings to define ubiquitous platforms or
architectures belong in github.com/bazelbuild/platforms.

And even there, we don't want to get overly specific. For the most part,
individual toolchain creators should
define fine-grained constraints with their toolchains.  "arm" for example,
is pretty much only "not x86 or ppc".
In reality there are hundreds of different arm implementations, so an
individual project may define the
constraints that match their needs.

GMNGeoffrey:
Ok, I can tell you that as a Bazel user the confusing story around common architecture constraints (coupled with the poor ergonomics of select) make this a major weak point for Bazel.

PR #13700, opened 2021-07-16
Clarify that entries in the `packages` attribute of `package_group` can be directories if used with wildcards.

jginsburgn:


aiuto:
That is interesting. It feels like unexpected behavior. 
Before documenting that, I want to verify that this is intended, or see if we should start a bug to tighten that up.

sgowroji:
Hello @jginsburgn, Could you please resolve the above conflicts if you are still looking to submit this PR. Thanks!

jginsburgn:
> Hello @jginsburgn, Could you please resolve the above conflicts if you are still looking to submit this PR. Thanks!

and what about https://github.com/bazelbuild/bazel/pull/13700#issuecomment-889229528?

sgowroji:
Okay let me check and get back to you on this PR. Thank you for reply!!

PR #13704, opened 2021-07-17
Cherrypick for 4.2.0: Fix stripping of macOS loadable bundles

### LINKED ISSUE #11869: macOS frameworks: symbol stripping fails

cpsauer:
Cherrypick of https://github.com/bazelbuild/bazel/commit/1a0285c3b64b121268ced3eb9ad9d5ba396b4905 for 4.2.0 (#13558).

Without this bug fix, building stripped macOS loadable bundles fails--simply fixing a case that had been neglected.

Original commit message below:
----

Adds `-x` flag to stripping of macOS loadable bundles.

Loadable bundles--i.e. truly dynamically loadable libraries on macOS--cannot be stripped without this flag, since you'd be trying to strip away the all symbols, including those used for dynamic loading. Doing so results in `error: symbols referenced by indirect symbol table entries that can't be stripped`.

`-x` instead leads to the removal of the unneeded local symbols. As Apple notes in their man page: "For dynamic shared libraries, the maximum level of stripping is usually -x (to remove all non-global symbols)."

This should fix #11869

Closes #13314.

PiperOrigin-RevId: 368841977

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java
@@ -1074,6 +1074,9 @@ public class CompilationSupport {
     if (Iterables.contains(extraLinkArgs, "-dynamiclib")) {
       return StrippingType.DYNAMIC_LIB;
     }
+    if (Iterables.contains(extraLinkArgs, "-bundle")) {
+      return StrippingType.LOADABLE_BUNDLE;
+    }
     if (Iterables.contains(extraLinkArgs, "-kext")) {
       return StrippingType.KERNEL_EXTENSION;
     }
@@ -1544,6 +1547,7 @@ public class CompilationSupport {
   private enum StrippingType {
     DEFAULT,
     DYNAMIC_LIB,
+    LOADABLE_BUNDLE,
     KERNEL_EXTENSION
   }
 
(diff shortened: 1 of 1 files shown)


cpsauer:
I don't have the permissions to assign, but this cherrypick should go to @katre, per https://github.com/bazelbuild/bazel/issues/13558#issuecomment-879188938

katre (approved):

katre:
This was merged to release-4.2.0-patches (after a quick rebase).

PR #13705, opened 2021-07-17
Cherrypick for 4.2.0: Export proguard specs from aar_import

cpsauer:
Cherrypick of https://github.com/bazelbuild/bazel/commit/c8c0d94a49e1b865d95c6d245c2d152c7c7c9722 for 4.2.0 (#13558).

Without this bug fix, using the proguard feature subtly breaks a wide range of android dependencies, including official ones from Google. 

Original commit message below:
----

**Background**
https://github.com/bazelbuild/bazel/issues/3778

proguard specs from the `aar_import` rule do not get bubbled up to `android_binary`. In this PR, I wire up a `ProguardSpecProvider` from this rule that exports the `proguard.txt` within an AAR if it exists and any transitive proguard specs from the `exports` attribute.

**Changes**
* Add an `aar_embedded_proguard_extractor` script to extract `proguard.txt` from an AAR if it exists otherwise generate an empty proguard specs file
* In AarImport, wire up the proguard extractor action and export results through a `ProguardSpecProvider`.

Once this lands, the android rules would need to be bumped.

**Test Plan**
* Added tests for the extraction python script
* Added tests for the `aar_import` rule changes

Closes #12749.

PiperOrigin-RevId: 359667674

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AarImport.java
@@ -14,6 +14,7 @@
 package com.google.devtools.build.lib.rules.android;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.devtools.build.lib.actions.Action;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.actions.Artifact.SpecialArtifact;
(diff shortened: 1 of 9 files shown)


cpsauer:
I don't have the permissions to assign, but this cherrypick should go to @katre, per https://github.com/bazelbuild/bazel/issues/13558#issuecomment-879188938

katre (approved):

PR #13706, opened 2021-07-18
Prepare for toolchainifying proto rules

Yannic:
Progress on https://docs.google.com/document/d/1go3UMwm2nM8JHhI3MZrtyoFEy3BYg5omGqQmOwcjmNE/edit

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchain.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchain.java
@@ -1,0 +1,58 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.rules.proto;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.analysis.FilesToRunProvider;
+import com.google.devtools.build.lib.analysis.RuleContext;
+import javax.annotation.Nullable;
+
+/** Toolchain for {@code proto_*} rules. */
+public class ProtoToolchain {
+  /** Creates a {@link ProtoToolchain} from a {@link RuleContext}. */
+  @Nullable
+  public static ProtoToolchain fromRuleContext(RuleContext ruleContext) {
+    Preconditions.checkNotNull(ruleContext);
+
+    FilesToRunProvider compiler = ruleContext.getExecutablePrerequisite(":proto_compiler");
+    if (compiler == null) {
+      return null;
+    }
+
+    ProtoConfiguration protoConfiguration = ruleContext.getFragment(ProtoConfiguration.class);
+    if (protoConfiguration == null) {
+      return null;
+    }
+
+    return new ProtoToolchain(compiler, protoConfiguration.protocOpts());
+  }
+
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchain.java
```diff
@@ -0,0 +1,58 @@
...
+package com.google.devtools.build.lib.rules.proto;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.analysis.FilesToRunProvider;
+import com.google.devtools.build.lib.analysis.RuleContext;
+import javax.annotation.Nullable;
+
+/** Toolchain for {@code proto_*} rules. */
+public class ProtoToolchain {
```
katre, line 24: Use @AutoValue with this, so that the class gets reasonable `equals()` and `hashCode()` behavior.

Is this eventually going to be a provider returned from a `proto_toolchain` rule? If so, I'd go ahead and rename it to `ProtoToolchainInfo` now (and also so that there is less confusion for the future `ProtoToolchainRule` and `ProtoToolchain` classes for that).

katre (changes requested):

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoToolchain.java
Yannic: The plan is to implement `proto_toolchain` directly in Starlark and have a factory method `fromStarlark(ToolchainInfo)` that checks that there are `compiler` and `compiler_options` fields of the expected type for usage from native code.

I was advised against creating `ProtoToolchainInfo` that extends from `ToolchainInfo` because Starlark doesn't have inheritance and C++/Java recently migrated away from doing so.

Yannic (commented):
Renamed and `@AutoValue`'d, PTAL

katre:
I was possibly not clear: I agree that this shouldn't inherit from `ToolchainInfo`, but I feel that it should be a provider, which means a) naming it `ProtoToolchainInfo` and b) inheriting from [`NativeInfo`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/NativeInfo.java;drc=94da3a6e00897ca914ae9a678323c32fb3a97f3b;l=32).

There are a number of other things that would need to happen to make the provider available to a Starlark version of a `proto_toolchain` rule, but those can happen in a separate CL.

The current way that I am advising toolchains to work with `ToolchainInfo` is the following:
1. There should be a rule-specific provider (in this case, `ProtoToolchainInfo`)
2. The toolchain rule should provider **both**:
   a. The rule-specific provider directly, and
   b. The rule-specific provider via `ToolchainInfo` (so, something like `return [proto_toolchain_info, ToolchainInfo(proto_toolchain = proto_toolchain_info)]`)
3. Rules that use toolchain resolution get the `ToolchainInfo` provider from `ctx.toolchains`, and then access the rule-specific provider from that.
4. Other rules that want to depend directly on the rule-specific toolchain (ie, have a direct dependency on `proto_toolchain`) can directly access the rule-specific provider.

Most of this can wait for the remaining CLs in this series, but since you are creating the provider-like class now, let's go ahead and make it a full provider.

Yannic:
I was thinking of `ProtoToolchainInfo` as a helper for migration while proto rules are still in Java. Eventually, I was planning to have `ToolchainInfo(compiler=whatever, compiler_options=whatever)` and get rid of `ProtoToolchainInfo` completely because there isn't really a need for it.

If it's recommended to have this indirection, then sure, I can change that. https://docs.bazel.build/versions/main/toolchains.html should probably be updated to reflect that :).

I have one question /re 4: In which scenario would something depend on a `*_toolchain` target directly? Is it for testing?

katre:
Good point, the docs are slightly out of date. I do recommend still having the indirection, because `ToolchainInfo` fields is not type checked, whereas provider fields can be (and can give better error messages). So I do actually suggest using a real provider for proto toolchain data, and having the helper `fromRuleContext` that creates one magically if needed.

Yes, testing is the scenario where rules authors have had direct dependencies on toolchain-providing rules in the past. But (these comments aside) I don't want to dictate to rule authors about what to do.

Yannic:
That makes sense, thank you! PTAL

katre (approved):
I'll merge this today. Thanks!

Issue #10077, opened 2019-10-21
Avdmanager and emulator paths in export_files of androidsdk


geaden:
### Description of the problem / feature request:

Add `'tools/bin/avdmanager'` and `'tools/emulator'` to `export_files` in `external/androidsdk/BUILD.bazel`

### Feature requests: what underlying problem are you trying to solve with this feature?

When building custom rules for android (i.e. like [this one](https://github.com/jin/rules_experimental_android/blob/master/composer/defs.bzl)) you may need to specify paths to `tools/bin/avdmanager` and `tools/emulator`. The only way to do it currently is to update `external/androidsdk/BUILD.bazel`. If you try to use `composer_instrumentation_test` from the link above target analysis will fail without

```python
export_files([
  ...
  'tools/bin/avdmanager',
  'tools/emulator',
])
```

in `external/androidsdk/BUILD.bazel`.

### What operating system are you running Bazel on?

macOS 10.15

### What's the output of `bazel info release`?

`release 1.0.0-homebrew`

###  Have you found anything relevant by searching the web?

Original GitHub issue: https://github.com/jin/rules_experimental_android/issues/5

jinchizhou:
@geaden @jin is there an update on this?

ahumesky:
please feel free to send us a pull request for this, should be straight forward for us to merge it in

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

PR #13716, opened 2021-07-19
Export Android tools in android sdk rule

ThomasCJY:
We saw an issue similar to https://github.com/bazelbuild/bazel/issues/10077 and we need to export some android tools from sdk rules to be able to access them in our bazel rule

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_sdk_repository_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_sdk_repository_template.txt
@@ -80,4 +80,7 @@ create_system_images_filegroups(
 )
 
 exports_files([
+    'tools/bin/apkanalyzer',
+    'tools/bin/avdmanager',
+    'tools/emulator',
 %exported_files%] + glob(["system-images/**"], allow_empty = True))


ThomasCJY:

src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_sdk_repository_template.txt
```diff
@@ -80,4 +80,7 @@ create_system_images_filegroups(
 )
 
 exports_files([
+    'tools/bin/apkanalyzer',
+    'tools/bin/avdmanager',
```
ahumesky, line 84: apkanalyzer and avdmanager are shell scripts that invoke their corresponding java application. In particular, they set up the classpath for particular jars in the tools/lib directory. So I'm not sure it would be enough to export just the shell scripts. Have you tried using these exported scripts as an action or in a genrule?

What you may need to do is write java_import rules for the jars that are set in the classpath in the scripts, and write a java_binary rule that specifies the main class and depends on the java_imports

ThomasCJY: i tired and i will need to add for apk analyzer I will need to pass in JAVA_HOME env variable in bazel command to make it work

sgowroji:
Hello @ThomasCJY, Any update on this PR? Please. Thanks!

PR #13718, opened 2021-07-20
Zipalign the split APK before installing

Bencodes:
Similar to #13050 but prevents the split APK (`--split_apks`) from being installed on Android 11 devices due to the fact that the contents of the APK are not aligned on a 4-byte boundary.


Bencodes:

ahumesky:
Thanks for the fix, I'll start the import on this.

ahumesky:
Unfortunately this change is failing an internal test, and in somewhat surprising way, so it might take a little while to debug.
https://github.com/bazelbuild/bazel/pull/13653

Issue #13713, opened 2021-07-19
[4.2.0rc1] rules_apple fails with NoSuchElementException


katre:
Downstream rules_apple tests for the 4.2.0rc1 release (#13558) are failing with NoSuchElementException in error handling code.

Error stack trace:
```
(16:44:12) FATAL: bazel crashed due to an internal error. Printing stack trace:
[...]

PR #13719, opened 2021-07-20
Remote: Fix a race when reporting action progresses.

coeuvre:
Fixes https://github.com/bazelbuild/bazel/issues/13713.

I will create checkpick request for 4.2.0 once merged into main branch.


katre (approved):
Thanks!

Issue #7448, opened 2019-02-16
Account for /bin/ps in Sandbox configuration


vmax:
### Description of the problem / feature request:

If a binary in system path has SUID bit set, test will fail in runtime with `Operation not permitted`. Sandbox on Linux, however, does not have the same behaviour.

### Feature requests: what underlying problem are you trying to solve with this feature?

I need to read process list via executing `ps` in `sh_test`. It works on Linux whereas on macOS I'm getting
```
/private/var/tmp/_bazel_vmax/8e8cab568c69716a0c9992fe4f39e019/sandbox/darwin-sandbox/1/execroot/__main__/bazel-out/darwin-fastbuild/bin/my.runfiles/__main__/my: line 12: /bin/ps: Operation not permitted
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

`test.sh`:
```
#!/usr/bin/env bash
ps
```

`BUILD`:
```
sh_test(
    name = 'my',
    srcs = ["test.sh"]
)
```

### What operating system are you running Bazel on?

macOS Mojave 10.14.3 where bug is reproducible
Debian 9.5 in Docker where there's no such bug

### What's the output of `bazel info release`?

`release 0.22.0`

###  Have you found anything relevant by searching the web?

—

### Any other information, logs, or outputs that you want to share?

I suspect the reason is somehow that `/bin/ps` has SUID bit set on:
```
ghost:is-ps-forbidden-in-tests vmax$ ls -la $(which ps)
-rwsr-xr-x  1 root  wheel  51280 30 ноя 10:39 /bin/ps
```

I've created my own dummy script with SUID bit in a PATH visible in `bazel`'s runtime and the behaviour is consistent with what is described in the issue
`bazel run //:my`, however, works perfectly

jmmv:
I don't see a way to detect SUID binaries in the `sandbox-exec` configuration. However, it is possible to poke holes for specific binaries. I have found that this works:

```
(version 1)
(allow default)
(allow process-exec (with no-sandbox) (literal "/bin/ps"))
```

so we should probably extend the sandbox configuration to account for `/bin/ps` given that `ps` failing is surprising behavior and it doesn't match Linux.

I do not see other worthwhile binaries to include though:

```
$ ls -l /bin /sbin /usr/bin /usr/sbin | grep rws
-rwsr-xr-x  1 root  wheel    64416 Apr  6 16:39 ps
[...]

PR #13727, opened 2021-07-22
Allow sandboxed macOS actions to run /bin/ps

keith:
Fixes https://github.com/bazelbuild/bazel/issues/7448

On Apple Silicon (maybe not exclusively) using rules_foreign_cc without
this you end up seeing many `/bin/ps` failures when configure scripts
are running.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/sandbox/DarwinSandboxedSpawnRunner.java
+++ b/src/main/java/com/google/devtools/build/lib/sandbox/DarwinSandboxedSpawnRunner.java
@@ -340,6 +340,7 @@ final class DarwinSandboxedSpawnRunner extends AbstractSandboxSpawnRunner {
       out.println("(version 1)");
       out.println("(debug deny)");
       out.println("(allow default)");
+      out.println("(allow process-exec (with no-sandbox) (literal \"/bin/ps\"))");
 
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

tetromino:
On MacOS, non-root users are not allowed to see other users' process list; this means the executables for any general purpose process-listing tool (`/bin/ps`, `/usr/bin/top`, the new-fangled [procs](https://crates.io/crates/procs) tool, etc.) must be suid root.

I don't think there's a good reason for a Bazel action to see other users' process list; but on the other hand, carrying around our own limited, sandbox-friendly variant of ps seems a lot of work for no real security gain.

As far as I can tell from reading the `ps` man page and the [ps.c source](https://opensource.apple.com/source/adv_cmds/adv_cmds-176.100.1/ps/ps.c), the MacOS `/bin/ps` can't do anything nefarious. We ought to allow it.

I have also verified that `(allow process-exec (with no-sandbox) (literal "/bin/ps"))` only allows executing `/bin/ps` - not e.g. `/tmp/bin/ps` (so it wouldn't allow sneaking in of unrelated unsandboxed binaries under the `ps` name).

tetromino (approved):

PR #13739, opened 2021-07-25
local_java_runtime should create a public native#java_runtime

Bencodes:
`local_java_repository` should expose the underlying `java_runtime` as public so that it can be used when creating custom java toolchains.

I'm able to reproduce this using `5.0.0-pre.20210708.4`

WORKSPACE
```
load("@bazel_tools//tools/jdk:local_java_repository.bzl", "local_java_repository")

local_java_repository(
    name = "jdk11",
    java_home = "/tmp/jdk11/11.0.11_9/jdk-unarchived/",
)
```

BUILD.bazel
```
load("@bazel_tools//tools/jdk:default_java_toolchain.bzl", "default_java_toolchain")

default_java_toolchain(
    name = "java11",
    java_runtime = "@jdk11//:jdk11",
    # ...
)
```

Error message
```
(00:44:56) ERROR: .../BUILD.bazel:64:23: in java_toolchain rule //:jdk11_java_toolchain: target '@jdk11//:jdk11' is not visible from target '//:jdk11_java_toolchain'. Check the visibility declaration of the former target if you think the dependency is legitimate
```

### DIFF
--- a/tools/jdk/local_java_repository.bzl
+++ b/tools/jdk/local_java_repository.bzl
@@ -16,7 +16,7 @@
 
 load(":default_java_toolchain.bzl", "JVM8_TOOLCHAIN_CONFIGURATION", "default_java_toolchain")
 
-def _detect_java_version(repository_ctx, java_bin, visibility = ["//visibility:public"]):
+def _detect_java_version(repository_ctx, java_bin):
     properties_out = repository_ctx.execute([java_bin, "-XshowSettings:properties"]).stderr
     # This returns an indented list of properties separated with newlines:
     # "  java.vendor.url.bug = ... \n"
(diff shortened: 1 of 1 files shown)


tools/jdk/local_java_repository.bzl
```diff
@@ -60,6 +60,7 @@ def local_java_runtime(name, java_home, version, runtime_name = None):
         native.java_runtime(
             name = runtime_name,
             java_home = java_home,
+            visibility = ["//visibility:public"],
```
comius, line 63: Perhaps it would be better to add a visibility parameter with default value public. So that it can be changed if needed.

comius (changes requested):

tools/jdk/local_java_repository.bzl
Bencodes: @comius updated

comius (approved):

PR #13745, opened 2021-07-26
Flip --experimental_worker_allow_json_protocol

keith:
Based on the discussion here
https://github.com/bazelbuild/bazel/pull/13607 we want to flip this and
leave it around temporarily, but there are also some refactorings coming
around how to do that. This flips the value now so it can be used
without the flag sooner in rolling releases.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkerOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkerOptions.java
@@ -48,7 +48,7 @@ public class WorkerOptions extends OptionsBase {
 
   @Option(
       name = "experimental_worker_allow_json_protocol",
-      defaultValue = "false",
+      defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {OptionEffectTag.BUILD_FILE_SEMANTICS},
       help =


larsrc-google:
FYI: This change is correct, but I first have to make some internal adjustments. We have some cases where Blaze and workers are released out of sync, so we can't just flip this flag internally, because we can't tell the workers which one to use.

aiuto:
friendly ping.  Can we move this along.  It seems mergable.

keith:
Now that the ndjson change merged I assume the blocking changes are internal? Is that correct @larsrc-google ?

keith:
@Wyverald can we cherry pick this for the next 5.x release?

Wyverald:
We shouldn't flip incompatible flags in the middle of an LTS release. This has to wait until 6.x.

keith:
@Wyverald this isn't a breaking change though, since workers can only be one or the other, so I think this is purely additive?

Wyverald:
Ah, I see -- so flipping this flag only enables an extra feature that was already checked in? Then it indeed seems fine.

keith:
Yep, it was just a warning so that the API could be safely changed and users shouldn't be surprised because they had to pass the flag

aiuto:
Just affirming that this just enables a feature that people have to
willingly invoke.
It should have defaulted to true months ago.

Wyverald:
@bazel-io fork 5.1

PR #13749, opened 2021-07-27
Add new data class for rule transition factories.

katre:
Part of the fix for #13751.


gregestren:
What does this change facilitate?

katre:
Sorry, created an issue in github. This is part of addressing the crash when a user tries to use an attribute-only transition (like the exec transition) directly on a rule. See #13751.

gregestren:
That makes more sense, thanks. Does something in this PR fix the crash, or is there more to come?

I'm being more pedantic than normal since we've noticed we have a lot of abstractions for how transitions are created and applied. Since this adds another abstraction, I want to ensure I understand clearly its role.

katre:
The next commit adds a marker interface so that TransitionFactory can only be created with specified data classes (AttributeTransitionData and the new RuleTransitionData). The final commit adds a type enum and an error message when the wrong type is used.

gregestren (approved):

PR #13750, opened 2021-07-27
[cherrypick 4.2] Migrate ExampleWorker to use WorkRequestHandler.

meisterT:
This means that we're using a real worker implementation in our example and integration tests. It also includes a JSON WorkRequestHandler, but it isn't used except in the tests.

RELNOTES: None.
PiperOrigin-RevId: 345765354


katre:
Why do we want to cherrypick example/test fixes to 4.2.0? What impact will this have on Bazel users who are getting the released version?

meisterT:
Another pile of cherrypick requests gets easier if we cherrypick this, but let's check test failures first.

katre:
Can we have all the cherrypicks as separate commits in the same PR? That will make it easier for me to tell what's going on.

meisterT:
Yep, we want to do that, gives us some time.

PR #13755, opened 2021-07-27
Add TransitionType enum and add checks that transitions aren't used

### LINKED ISSUE #13751: Crash when using exec transition as incoming rule transition
Reproduce:
```
def _impl(ctx):
    pass

demo = rule(
    implementation = _impl,
    cfg = config.exec(),
)
```

Error:
```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.ClassCastException: class com.google.devtools.build.lib.packages.AutoValue_RuleTransitionData cannot be cast to class com.google.devtools.build.lib.packages.AttributeTransitionData (com.google.devtools.build.lib.packages.AutoValue_RuleTransitionData and com.google.devtools.build.lib.packages.AttributeTransitionData are in unnamed module of loader 'app')
	at com.google.devtools.build.lib.analysis.config.ExecutionTransitionFactory.create(ExecutionTransitionFactory.java:39)
	at com.google.devtools.build.lib.analysis.config.TransitionResolver.applyTransitionFromFactory(TransitionResolver.java:114)
	at com.google.devtools.build.lib.analysis.config.TransitionResolver.applyRuleTransition(TransitionResolver.java:99)
	at com.google.devtools.build.lib.analysis.config.TransitionResolver.evaluateTransition(TransitionResolver.java:83)
	at com.google.devtools.build.lib.analysis.AnalysisUtils.targetsToDeps(AnalysisUtils.java:203)
	at com.google.devtools.build.lib.analysis.AnalysisUtils.getTargetsWithConfigs(AnalysisUtils.java:191)
	at com.google.devtools.build.lib.analysis.BuildView.update(BuildView.java:244)
	at com.google.devtools.build.lib.buildtool.AnalysisPhaseRunner.runAnalysisPhase(AnalysisPhaseRunner.java:217)
	at com.google.devtools.build.lib.buildtool.AnalysisPhaseRunner.execute(AnalysisPhaseRunner.java:129)
```

(This is copied from an internal bug report)

katre:
inappropriately.

Fixes #13751.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/ExecutionTransitionFactory.java
@@ -53,6 +53,11 @@ public class ExecutionTransitionFactory
     return new ExecutionTransitionFactory(DEFAULT_EXEC_GROUP_NAME);
   }
 
+  @Override
+  public TransitionType transitionType() {
+    return TransitionType.ATTRIBUTE;
+  }
+
   /**
    * Returns a new {@link ExecutionTransitionFactory} for the given {@link
    * com.google.devtools.build.lib.packages.ExecGroup}.
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/transitions/ComposingTransitionFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/transitions/ComposingTransitionFactory.java
@@ -46,6 +46,9 @@ public abstract class ComposingTransitionFactory<T extends TransitionFactory.Dat
     Preconditions.checkNotNull(transitionFactory1);
     Preconditions.checkNotNull(transitionFactory2);
     Preconditions.checkArgument(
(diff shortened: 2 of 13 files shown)


gregestren:
How do Starlark transitions fit into this? If you define a 1-1 Starlark transition and both apply it to a rule and an attribute, how does that still work? Is that because those don't show up as `TransitionFactory` instances in the `StarlarkRuleClassFunctions` object?

gregestren:
Oh, those are handled at https://github.com/bazelbuild/bazel/blob/525227e08483cda399b7aee4db997b6552e15891/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java#L409?

katre:
Right, plain transitions don't run into this workflow. There's no way to make an actual `TransitionFactory` instance from Starlark, so this only applies to native code.

Native factories that don't care about the data type (like the transition factory for `HostTransition`) can use `TransitionFactory.TransitionType.ANY` to indicate that.

gregestren (approved):

PR #13757, opened 2021-07-27
Allow relative log paths

keith:
This reverts commit bec1ea84caa75c8c3850b96d105627d1a36e322a and ed7f81a5bd0af39c1f94951aaa272e296baa9962

These changes required that users pass absolute paths to `--profile` and `--memory_profile` with the goal of disallowing users to alter the source tree through bazel invocations. Practically speaking it's very useful for users to quickly dump a log like `--profile=foo.log` for testing, where that won't be picked up by any bazel rules, and therefore shouldn't affect anything, or to write their logs to a `.gitignore`d / `.bazelignore`d directory such as `--profile=./tmp/logs/foo.log`. This has the advantage of making sure logs don't conflict between multiple source trees using bazel. Regardless This wouldn't restrict users from adding an absolute path in their bazel wrappers to the source root such as `--profile=$PWD/tmp/logs/foo.log`, so I think we should allow users to self regulate these issues.

More discussion: https://github.com/bazelbuild/bazel/commit/bec1ea84caa75c8c3850b96d105627d1a36e322a#commitcomment-53595211


meisterT (approved):

keith:
Thanks! ❤️

PR #13760, opened 2021-07-28
Fix ManifestMergerAction test case on windows

Bencodes:


src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
```diff
@@ -174,8 +174,7 @@ private static Path removePermissions(Path manifest, Path outputDir)
       }
     }
     // Write resulting manifest to the output directory, maintaining full path to prevent collisions
```
ted-xie, line 176: I think this comment is no longer valid with your change. The full path is no longer maintained; rather, a temporary file is created at `$outputDir/.xml/${gibberish}AndroidManifest`.

ted-xie (changes requested):

PR #13765, opened 2021-07-28
Update RE third party code.

rubensf:
This does come with a slight breaking change in
digest_function->digest_functions on CacheCapabilities.


atetubou:
fix build?

aiuto:
Thanks for the PR, but it obviously does not build. It is also in third_party, which changes the merge process.
I'll keep this open as a place for discussion.

rubensf:
Google instructions for contributing to bazel explicitly say to send third_party changes separately from main code changes... Which seems impossible here :)

I was hoping someone would tell me what to do otherwise.

aiuto:
Right. We can figure out a plan in this PR (or maybe open a distinct issue).
My hunch is that we have to update Bazel so the update will happen,
then we can do the third_party update.

coeuvre:
Discussed with @philwo today. We need some hacks for this case.

PR #13769, opened 2021-07-29
Remote: Cleanup the code that determines whether the spawn should accept/upload results from/to remote cache.

### LINKED ISSUE #13621: `no-remote` and `no-remote-cache` are affecting the disk cache

When an action is tagged with `no-remote` or `no-remote-cache`, it's not properly using the disk cache. Only with `no-cache` should it exclude the disk cache as well.

```shell
bazel build --disk_cache=/tmp/disk_cache -- //some/cacheable:target
bazel clean
bazel build --disk_cache=/tmp/disk_cache --modify_execution_info='.*=+no-remote-cache' -- //some/cacheable:target
```

The second build won't check the disk cache.

macOS 11.4

4.1.0, but is still broken at HEAD.

coeuvre:
Remote cache (and execution) for an action can be controlled with following tags:
  - `no-remote-cache`: prevent remote caching, but allow local caching
  - `no-remote-exec`: prevent remote execution, but allow remote caching
  - `no-remote`: combine no-remote-cache and no-remote-exec
  - `no-cache`: extend no-remote-cache to also prevent local caching

A combined cache is treated as remote cache, hence is disabled for actions that are tagged with `no-remote-cache` unless `--incompatible_remote_results_ignore_disk` is set in which case local component is enabled.

Fixes #13621.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
@@ -288,7 +288,7 @@ public class RemoteExecutionService {
         if (remoteOptions.incompatibleRemoteResultsIgnoreDisk) {
           return Spawns.mayBeCachedRemotely(spawn);
         } else {
-          return remoteOptions.remoteAcceptCached && Spawns.mayBeExecutedRemotely(spawn);
+          return remoteOptions.remoteAcceptCached && Spawns.mayBeCachedRemotely(spawn);
         }
       } else {
         // Remote cache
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -220,19 +221,76 @@ public NetworkTime getNetworkTime() {
...
-    return remoteCache != null && Spawns.mayBeCached(spawn) && Spawns.mayBeCachedRemotely(spawn);
+  private static boolean useRemoteCache(RemoteOptions options) {
+    return !isNullOrEmpty(options.remoteCache) || !isNullOrEmpty(options.remoteExecutor);
   }
 
-  /** Returns {@code true} if the result of spawn may be cached. */
-  public boolean mayBeCached(Spawn spawn) {
-    return remoteCache != null && Spawns.mayBeCached(spawn);
+  private static boolean useDiskCache(RemoteOptions options) {
+    return options.diskCache != null && !options.diskCache.isEmpty();
```
philwo, line 229: Can we use `!isNullOrEmpty(options.diskCache)` here?

philwo (approved):

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
coeuvre: `options.diskCache` is a `PathFragment`, before converting to `String` we have to check for `null`, which makes `isNullOrEmpty` less appealing.

PR #13778, opened 2021-07-30
Update xz to 1.9.

benjaminp:
This makes decompression more than 10% faster.


benjaminp:
@philwo perhaps you can push this? Thank you.

philwo:
Will do! Thanks :)

✅ I verified that the SHA256 of the `xz-1.9.jar` in this PR matches the one from Maven Central.

PR #13788, opened 2021-08-02
Report message from remote on status failure

illicitonion:
If the remote has useful debugging information, it's more useful to show
it than discard it.


coeuvre (changes requested):
Thanks for the PR!

Can you please add a test in `RemoteSpawnRunnerTest`?

illicitonion:
Added the test - thanks!

coeuvre (approved):
Thanks! Importing.

PR #13790, opened 2021-08-02
Fix up @build_tools//tools/java package

illicitonion:
https://github.com/bazelbuild/bazel/pull/12994 added the file, but it's
currently exposed as:
`@build_tools//tools:java/java_stub_template.txt`

This fixes it up to be properly exposed as:
`@build_tools//tools/java:java_stub_template.txt`

Which makes more sense :)


illicitonion:
Ran into a weird issue here - debugging - apologies again for the noise!

tools/java/BUILD
```diff
@@ -19,8 +19,21 @@ genrule(
...
 )
 
+# Generate a BUILD.bazel file to put in the repo, as this one references targets which don't exist in a Bazel distribution.
+# Note: We need to call it BUILD.bazel not BUILD because otherwise it conflicts with this file.
+genrule(
+    name = "generated_BUILD",
+    srcs = ["BUILD.for_buildtools"],
+    outs = ["BUILD.bazel"],
+    cmd = "cp $< $@",
+)
```
comius, line 29: I'd prefer if BUILD.tools file is added to the repo and used in the filegroup, like it's done in other directories (tools/jdk, tools/java/runfiles).

java_stub_template.txt needed genrule with cp, but in general I'd prefer it's avoided.

comius (changes requested):

tools/java/BUILD
illicitonion: Done!

comius (approved):

tools/java/BUILD
comius: You also need to add BUILD.tools to srcs filegroup in tools/java/BUILD

Issue #2575, opened 2017-02-23
Android APK was rejected by Google Play because it's debuggable


fishy:
Please provide the following information. The more we know about your system and use case, the more easily and likely we can help.

### Description of the problem / feature request / question:

I have an android_binary rule, and generated `app_unsigned.apk` file from the rule.

Then I signed it with my release key per instructions on https://github.com/bazelbuild/bazel/issues/1613, as:

```
bazel build java:app --strategy=AndroidAapt=standalone
zipalign -v -p 4 bazel-bin/java/app_unsigned.apk app-unsigned-aligned.apk
apksigner sign --ks release.jks --out app.apk app-unsigned-aligned.apk
rm app-unsigned-aligned.apk
```

When I try to publish the generated `app.apk` file on Google Play, it was rejected as "You uploaded a debuggable APK".

Bazel should provide a way to generate undebuggable APKs.

### Environment info
* Operating System:
macOS 10.12.3

* Bazel version (output of `bazel info release`):
release 0.4.4-homebrew

* If `bazel info release` returns "development version" or "(@non-git)", please tell us what source tree you compiled Bazel from; git commit hash is appreciated (`git rev-parse HEAD`):

### Anything else, information or logs or outputs that would be helpful?
(If they are large, please upload as attachment or provide link).

fishy:
Adding `android:debuggable="false"` into the `application` section in `AndroidManifest.xml` can workaround this issue, but that means I have to maintain two `AndroidManifest.xml` files, one for debug and one for release, which is error-prone. Bazel should provide a mechanism to generate release APK.

iirina:
@ahumesky can you take a look please?

ahumesky:
--compilation_mode=opt (or -copt for short) should set the debuggable flag to false, among other optimizations. We're revamping our android documentation and we'll be sure to mention this.

fishy:
`--compilation_mode=opt` did the trick, but `-copt` doesn't work:

```
$ bazel build java:app --strategy=AndroidAapt=standalone -copt
Invalid options syntax: -copt
```

```
$ bazel build java:app --strategy=AndroidAapt=standalone --copt
Expected value after --copt
```

ahumesky:
whoops, that's a typo, there should be a space between -c and opt ("-c opt") (i edited my comment above)

fishy:
`-c opt` worked. Thanks!

PR #13801, opened 2021-08-04
Support overriding debuggable in AndroidManifest

benjaminRomano:
**Background**
I would like to add `debuggable="false"` into the `AndroidManifest.xml`, but specifying this directly in the `AndroidManifest.xml` does not take affect as `--debug-mode` is passed into `aapt2` which overrides what developers set. [ref](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/tools/aapt2/cmd/Link.h;l=292-294?q=debuggable&ss=android%2Fplatform%2Fsuperproject:frameworks%2Fbase%2Ftools%2Faapt2%2F)

The only way to get a debuggable apk is to require `--compilation-mode opt` which will invalidate the entire dep graph which is not feasible for large apps.
https://github.com/bazelbuild/bazel/issues/2575

**Changes**
* Add if `debuggable: "false"` is specified in `manifestValues` of the `android_binary` target do not set `--debug-mode true` during resource processor step

**Usage**
```
android_binary(
    name = "app",
    dex_shards = 10,
    manifest = "src/bazel/AndroidManifest.xml",
    manifest_values = {
        "minSdkVersion": "19",
        "appName": "...",
        "applicationId": "...",
        "debuggable": "false",
    },
    multidex = "native",
    deps = ["..."],
)
```

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/ProcessedAndroidData.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/ProcessedAndroidData.java
@@ -241,7 +241,7 @@ public class ProcessedAndroidData {
       Map<String, String> manifestValues) {
     return new AndroidResourcesProcessorBuilder()
         // Settings
-        .setDebug(dataContext.useDebug())
+        .setDebug(dataContext.useDebug() && !manifestValues.getOrDefault("debuggable", "true").equals("false"))
         .setJavaPackage(manifest.getPackage())
         .setApplicationId(manifestValues.get("applicationId"))
         .setVersionCode(manifestValues.get("versionCode"))


ahumesky:
The way that the final "debuggable" value gets determined currently and with this change was fiddly enough that I made some tables. (I mostly reasoned about this in my head looking through the code, so it might not be perfect.)

Current:
--compilation_mode | manifest_values debuggable | AndroidManifest.xml debuggable | final debuggable value
-- | -- | -- | --
dbg/fastbuild (default) | X | X | True
opt | X | False | False
opt | X | True | True
opt | False (with substitution) | X | False
opt | True  (with substitution) | X | True

With this change:
--compilation_mode | manifest_values debuggable | AndroidManifest.xml debuggable | final debuggable value
-- | -- | -- | --
dbg/fastbuild | False | True | True
dbg/fastbuild | False | False | False
dbg/fastbuild | True | X | True
dbg/fastbuild | not set | X | True
opt | X | False | False
opt | X | True | True
opt | False (with substitution) | X | False
opt | True  (with substitution) | X | True

(X = "don't care")

So the first thing is that it seems more difficult to reason about what's going on with this change, because there are more cases to think about.

The second is that it might not be trivial to release this change internally at Google, because I'm finding apps that might be affected by this (e.g., they have `manifest_values = { "debuggable": "false" }`, and that might not be doing anything right now, and with this change it might start doing something).

I think what would be preferable is to add a flag that just takes `--compilation_mode` (and aapt2's `--debug-mode`) out of the equation. Something like `--android_opt_compilation_mode_sets_debuggable` that can be put in a bazelrc. That way, the final value is simply either the value that's already in the manifest itself, or from the build file via value substitution:

```
BUILD:
manifest_values = {
   "debuggable": "true",
}

AndroidManifest.xml:
android:debuggable="${debuggable}"
```

It would also be possible to make this configurable on the command line:

```
load("@bazel_skylib//lib:dicts.bzl", "dicts")

config_setting(
  name = "android_debuggable",
  values = {
    "define": "android_debuggable=true",
  }
)

manifest_values = {
  "applicationId": "com.foo.app",
}

android_binary(
    name = "app",
    manifest = "AndroidManifest.xml",
    deps = [
        ":applib",
    ],
    manifest_values = select({
        ":android_debuggable": dicts.add(manifest_values, {
            "debuggable": "true",
        }),
        "//conditions:default": dicts.add(manifest_values, {
            "debuggable": "false",
        }),
    }),
)
```

ahumesky:
If there's still any interest in pursing this, happy to reopen.

EdbertChan:
@ahumesky Had a chat w Corbin about this actually.

https://github.com/bazelbuild/rules_android/pull/238#issuecomment-2156192368

Copy pasting the original for convenience. 

> Per @restingbull (Corbin) offline conversation, the dependency on Compilation.OPT was a matter of convenience from inside of Google. Supposedly, few/nobody who was working on building with the debuggable flag build cared about release so full invalidation was not a problem.
> 
> If you weren't building an optimized binary, you should not be building a release anyway since release should be optimized.
> 
> Outside of Google, that may not be the case. Some developers very specifically want to work on performance.

Issue #13764, opened 2021-07-28
UrlRewriter logging is very noisy


mattem:
### Description of the problem / feature request:

When using `--experimental_downloader_config` to rewrite URLs, the rewritten URLs are [logged at INFO level](https://cs.opensource.google/bazel/bazel/+/ceb498fd6b9323a738f0ba628d12a23a161a0671:src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/UrlRewriter.java;l=106;drc=7649f610c45190735fd7de433b15679b21b2d91b;bpv=0;bpt=1). For a large repository with a lot of rewrites this causes a large amount of log noise.

FR to allow these messages to be suppressed without suppressing all other messages at INFO level, either via `--ui_event_filters` or via another flag. Or, the inverse, only print these messages when a specific flag is passed (eg `--verbose_url_rewrite_info` or something)

### Feature requests: what underlying problem are you trying to solve with this feature?

Less noisy logging output.

### What operating system are you running Bazel on?

MacOS / Linux

### What's the output of `bazel info release`?

```
release 4.0.0
```

alexeagle:
Also the downloader prints SHA256 for each file at INFO level:

 https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java;l=237?q=DownloadManager.java

philwo:
@meteorcloudy @Wyverald Could you please triage this?

meteorcloudy:
I believe this is fixed at https://github.com/bazelbuild/bazel/commit/d21345dea13a8d46475cf29563a64a7fbd2d6b35 ?

Please re-open if the issue still persists.

alexeagle:
The SHA256 issue is also very noisy, I should file a separate issue.

PR #13806, opened 2021-08-05
Remove noisy logs from UrlRewriter.

glukasiknuro:
UrlRewriter prints log for each URL that is rewritten, in form of
`Rewritten %s as %s`. This causes tens of long logs to be printed,
making logs very noisy.

See also https://github.com/bazelbuild/bazel/issues/13764.


src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/UrlRewriter.java
```diff
@@ -102,10 +102,6 @@ public static UrlRewriter getDownloaderUrlRewriter(String configPath, Reporter r
     ImmutableList<URL> rewritten =
         urls.stream().map(rewriter).flatMap(Collection::stream).collect(toImmutableList());
 
-    if (!urls.equals(rewritten)) {
```
denyska, line 105: Is there way to make it print log once for given url ?

denyska (approved):

coeuvre (approved):

PR #13808, opened 2021-08-05
remote: Do not retry bytestream read requests if the file is complete.

benjaminp:
Previously, an error at the end of the Read stream would result in a pointless retry request for 0 bytes of data.


coeuvre (approved):
Thanks!

PR #13812, opened 2021-08-06
bes: Allow additional headers to be passed to the server.

benjaminp:
There is already a --remote_header option that passes extra headers to the remote cache and executor. This revision allows extra headers to be passed along to a BES server with the --bes_header option.

RELNOTES[NEW]: Add `--bes_header` flag to pass extra headers to the BES server.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
@@ -28,13 +28,18 @@ import com.google.devtools.build.v1.PublishBuildToolEventStreamRequest;
 import com.google.devtools.build.v1.PublishBuildToolEventStreamResponse;
 import com.google.devtools.build.v1.PublishLifecycleEventRequest;
 import io.grpc.CallCredentials;
+import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannel;
+import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.StatusException;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.AbstractStub;
+import io.grpc.stub.MetadataUtils;
 import io.grpc.stub.StreamObserver;
 import java.time.Duration;
+import java.util.List;
+import java.util.Map;
 import javax.annotation.Nullable;
 
 /** Implementation of BuildEventServiceClient that uploads data using gRPC. */
@@ -48,11 +53,24 @@ public class BuildEventServiceGrpcClient implements BuildEventServiceClient {
   private final PublishBuildEventBlockingStub besBlocking;
 
   public BuildEventServiceGrpcClient(
-      ManagedChannel channel, @Nullable CallCredentials callCredentials) {
-    this(
-        withCallCredentials(PublishBuildEventGrpc.newStub(channel), callCredentials),
-        withCallCredentials(PublishBuildEventGrpc.newBlockingStub(channel), callCredentials),
-        channel);
+      ManagedChannel channel,
+      @Nullable CallCredentials callCredentials,
+      List<Map.Entry<String, String>> headers) {
+    ClientInterceptor interceptor = null;
+    if (!headers.isEmpty()) {
+      Metadata extraHeaders = new Metadata();
+      headers.forEach(
+          header ->
+              extraHeaders.put(
+                  Metadata.Key.of(header.getKey(), Metadata.ASCII_STRING_MARSHALLER),
+                  header.getValue()));
+      interceptor = MetadataUtils.newAttachHeadersInterceptor(extraHeaders);
+    }
+    this.besAsync =
+        configureStub(PublishBuildEventGrpc.newStub(channel), callCredentials, interceptor);
+    this.besBlocking =
+        configureStub(PublishBuildEventGrpc.newBlockingStub(channel), callCredentials, interceptor);
+    this.channel = channel;
   }
 
   @VisibleForTesting
@@ -65,9 +83,10 @@ public class BuildEventServiceGrpcClient implements BuildEventServiceClient {
     this.channel = channel;
   }
 
-  private static <T extends AbstractStub<T>> T withCallCredentials(
-      T stub, @Nullable CallCredentials callCredentials) {
+  private static <T extends AbstractStub<T>> T configureStub(
+      T stub, @Nullable CallCredentials callCredentials, @Nullable ClientInterceptor interceptor) {
     stub = callCredentials != null ? stub.withCallCredentials(callCredentials) : stub;
+    stub = interceptor != null ? stub.withInterceptors(interceptor) : stub;
     return stub;
   }
 
--- a/src/test/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceGrpcClientTest.java
+++ b/src/test/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceGrpcClientTest.java
@@ -15,37 +15,45 @@
 package com.google.devtools.build.lib.buildeventservice;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.when;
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
```diff
@@ -65,9 +83,10 @@ protected BuildEventServiceGrpcClient(
     this.channel = channel;
   }
 
-  private static <T extends AbstractStub<T>> T withCallCredentials(
-      T stub, @Nullable CallCredentials callCredentials) {
+  private static <T extends AbstractStub<T>> T configureStub(
+      T stub, @Nullable CallCredentials callCredentials, @Nullable ClientInterceptor interceptor) {
     stub = callCredentials != null ? stub.withCallCredentials(callCredentials) : stub;
+    stub = interceptor != null ? stub.withInterceptors(interceptor) : stub;
     return stub;
```
lfpino, line 90: What happens if both callCredentials and interceptor are present? Is that possible?

benjaminp: The stub should be wrapped by with call-credentials and then the header interceptor.

lfpino (approved):

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
```diff
@@ -48,11 +53,24 @@
...
   public BuildEventServiceGrpcClient(
-      ManagedChannel channel, @Nullable CallCredentials callCredentials) {
-    this(
-        withCallCredentials(PublishBuildEventGrpc.newStub(channel), callCredentials),
-        withCallCredentials(PublishBuildEventGrpc.newBlockingStub(channel), callCredentials),
-        channel);
+      ManagedChannel channel,
+      @Nullable CallCredentials callCredentials,
+      List<Map.Entry<String, String>> headers) {
+    ClientInterceptor interceptor = null;
```
michaeledgar, line 59: If the `BuildEventServiceGrpcClient` took the `ClientInterceptor` as a `@Nullable` constructor argument, then this class can remain focused on the business logic needed for a fully-constructed gRPC client.

The caller (`BazelBuildEventServiceModule`) can do the work of munging the configuration's key-value-pair headers into Metadata and a ClientInterceptor. In the future, if we need to change the Metadata or the ClientInterceptor (eg. adding tracing functionality), that change remains isolated to the Module.

This may also improve unit-testability of the `BuildEventServiceGrpcClient` by allowing tests to interpose the Interceptor on the client side.

src/test/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceGrpcClientTest.java
```diff
@@ -59,54 +67,92 @@ public void onError(Throwable t) {}
...
+   */
+  private static void withServer(ServerServiceDefinition service, TestWithServer callback)
+      throws Exception {
+    String uniqueName = UUID.randomUUID().toString();
+    Server server =
+        InProcessServerBuilder.forName(uniqueName).directExecutor().addService(service).build();
+    server.start();
     ManagedChannel channel = InProcessChannelBuilder.forName(uniqueName).directExecutor().build();
-    grpcClient = new BuildEventServiceGrpcClient(channel, null);
+    try {
```
michaeledgar, line 86: Test helpers which encapsulate the control flow of the entire test case tend to obscure the actual test behavior, which can frustrate investigation when tests break. In this case, the value of the helper method is abstracting the construction and teardown of the server. We can define a helper for that, without requiring the caller provide a callback.

The static helper can return an instance of an `AutoCloseable` class which includes the `#shutdown()` calls in its `#close()` method. Then the test cases can be written as:

```
@Test
public void besHeaders() throws Exception {
    ArrayList<Metadata> seenHeaders = new ArrayList<>();
    try (TestChannelAndServer channelAndServer = startInprocessServer(ServerInterceptors.intercept(...))) {
         BuildEventServiceGrpcClient grpcClient =
              new BuildEventServiceGrpcClient(
                  channel, null, ImmutableList.of(Maps.immutableEntry("metadata-foo", "bar")));
          assertThat(grpcClient.openStream(ack -> {}).getStatus().get()).isEqualTo(Status.OK);
          assertThat(seenHeaders).hasSize(1);
          Metadata headers = seenHeaders.get(0);
          assertThat(headers.get(Metadata.Key.of("metadata-foo", Metadata.ASCII_STRING_MARSHALLER)))
              .isEqualTo("bar");
    }
}
```

michaeledgar (commented):
Thanks for sending this pull request - the functionality makes sense and looks good!

I just have a couple of cleanups to suggest before merging.

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
benjaminp: Ah, yes, I went back and forth myself on this issue when writing this CL. I agree that the munging would be more comfortable in `BazelBuildEventServiceModule`, but I will need some guidance to get there. Ultimately, I took uneasily took the current approach because

1. `BuildEventServiceGrpcClient` is already "polluted" by taking in `CallCredentials` and having to mess with that, so what more is some header configuration?
2. More pragmatically, I found it significantly easier to unittest `BuildEventServiceGrpcClient` than `BazelBuildEventServiceModule`. ``BazelBuildEventServiceModule` is basically a "plugin" for its superclass, and it looks tricky to unittest what I want without disturbing too much of that ecosystem. (There used to be `BazelBuildEventServiceModuleTest.java` until it slunk away in 4ca2b6d5c4d201009ff21a3abbeddd9d58d4c95b. I'm not sure what infrastructure exists for testing `BazelBuildEventServiceModule` within Google and don't want to reinvent it. Additionally, I see that `BazelBuildEventServiceModule#newGrpcChannel` has a comment explaining `newGrpcChannel is only defined so it can be overridden in tests to not use a real network link`, but no such overrides exist in the public test suite. This makes me nervous about doing surgery on this class without a full view of the codebase.)

src/test/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceGrpcClientTest.java
benjaminp: Thanks—I like this suggestion a lot.

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
michaeledgar: Thanks for your quick followup!

Re: #1: My thought here is that `CallCredentials` is a gRPC domain object, from the `io.grpc` package, and the `ClientInterceptor` would be also. So we'd be handing a few gRPC objects to construct a gRPC client, which gets to be focused entirely on gRPC logic.

Re: #2: It looks like an oversight that `BazelBuildEventServiceModuleTest.java` is not open-sourced. Thanks for pointing that out, I'll remedy that ASAP! If you want, you can pull down [the patchset](https://bazel-review.googlesource.com/c/bazel/+/175231) and work off of that. You could honestly just make the Flag-and-String-munging be a static method on `BazelBuildEventServiceModule` and drop some unit tests into `BazelBuildEventServiceModuleTest` that call the static method. We already have a `#testKeywords()` test in there for what amounts to testing Flag manipulation.

michaeledgar: Actually, it looks like open-sourcing `BazelBuildEventServiceModuleTest.java` will be a bit more work than I expected, so I'm going to approve this and merge it in.

benjaminp: Alright,; thank you for your reviews and assistance on this.

PR #13822, opened 2021-08-09
Add authentication support to the downloader rewriter

ma-oli:
The external repository downloader rewriter currently allows configuring
bazel to pull external dependencies from alternative locations. However,
it doesn't take credentials in the url into account.

This change fixes that by reading the user information from the URL
object and updating the authHeaders accordingly.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
@@ -18,6 +18,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.devtools.build.lib.bazel.repository.cache.RepositoryCache;
 import com.google.devtools.build.lib.bazel.repository.cache.RepositoryCache.KeyType;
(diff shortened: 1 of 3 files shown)


ma-oli:
Just a friendly reminder :)

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
```diff
@@ -101,8 +102,11 @@ public Path download(
     }
 
     List<URL> rewrittenUrls = originalUrls;
+    Map<URI, Map<String, String>> rewrittenAuthHeaders = ImmutableMap.copyOf(authHeaders);
```
coeuvre, line 105: nit: `rewrittenAuthHeaders = authHeaders` to save the copy?

coeuvre (approved):

src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/DownloadManager.java
coeuvre: Please address this comment or I can do it during import if you don't mind.

keith: @ma-oli I submitted https://github.com/ma-oli/bazel/pull/1 to do this in case that's helpful!

coeuvre: I will import the PR and address this during that. Thanks for the PR!

keith: Thanks!

ma-oli: Hey; sorry about that. I'm not super familiar with Github and I didn't receive these on the right email. Thanks a lot @keith for following up.

PR #13824, opened 2021-08-10
Test that cc_binary can specify -stdlib=libc++

fmeum:
cc_binary can successfully specify -stdlib=libc++ as copt and linkopt as of f1531dc2 when the compiler is clang. This commit adds a test for this.

@oquenchil This is the test that I should have already included in #13666.

### DIFF
--- a/src/test/shell/bazel/cc_integration_test.sh
+++ b/src/test/shell/bazel/cc_integration_test.sh
@@ -1382,4 +1382,40 @@ EOF
       fail "bazel build should've succeeded with --features=compiler_param_file"
 }
 
+function test_clang_libcxx() {
+  local -r clang_tool=$(which clang)
+  if [[ ! -x ${clang_tool:-/usr/bin/clang_tool} ]]; then
+    echo "clang not installed. Skipping test."
+    return
+  fi
+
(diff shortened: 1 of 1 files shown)


fmeum:
Is it possible that the tests fail because they haven't picked up https://github.com/bazelbuild/bazel/commit/f1531dc2ebfb2db721a94f6f19081fa89af536da yet? The failure message seems to indicate that is the case.

oquenchil (changes requested):
Thank you!

Could you please add this test as part of cc_integration_test.sh instead of adding it in a separate file?

fmeum:
> Thank you!
> 
> Could you please add this test as part of cc_integration_test.sh instead of adding it in a separate file?

Done. However, this PR is blocked on https://github.com/bazelbuild/continuous-integration/pull/1212 as the Bazel CI images do not have libc++ installed.

oquenchil:
Ah understood. Can you please ping this again when the blocker is resolved?

philwo:
There's a different error now: `clang: error: invalid linker name in argument '-fuse-ld=/usr/bin/ld.gold'` - but maybe that's fixed by rebasing the PR?

PR #13827, opened 2021-08-11
Don't call `FindMissingBlobs` when not uploading

brentleyjones:
If `--disk_cache` is used, along with `--incompatible_remote_results_ignore_disk` and `--noremote_upload_local_results`, results aren't uploaded to the remote cache, even if they are uploaded to the disk cache, so there is no reason to call `FindMissingBlobs`. This change removes that call when it's not needed.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
@@ -103,18 +103,22 @@ public final class DiskAndRemoteCacheClient implements RemoteCacheClient {
   @Override
   public ListenableFuture<ImmutableSet<Digest>> findMissingDigests(
       RemoteActionExecutionContext context, Iterable<Digest> digests) {
-    ListenableFuture<ImmutableSet<Digest>> remoteQuery =
-        remoteCache.findMissingDigests(context, digests);
     ListenableFuture<ImmutableSet<Digest>> diskQuery =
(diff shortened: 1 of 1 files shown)


brentleyjones:
@coeuvre

src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
```diff
@@ -103,18 +103,22 @@ public void close() {
...
         diskCache.findMissingDigests(context, digests);
-    return Futures.whenAllSucceed(remoteQuery, diskQuery)
-        .call(
-            () ->
-                ImmutableSet.<Digest>builder()
-                    .addAll(remoteQuery.get())
-                    .addAll(diskQuery.get())
-                    .build(),
-            MoreExecutors.directExecutor());
+    if (!options.incompatibleRemoteResultsIgnoreDisk || options.remoteUploadLocalResults) {
```
coeuvre, line 108: Can we use the helper method `shouldUploadLocalResultsToRemoteCache` here?

coeuvre (changes requested):

src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
brentleyjones: Sure. I was copying the same check that was used in the method above.

coeuvre (approved):
Thanks!

Issue #13834, opened 2021-08-12
Crash with 4.2.0rc3: Unrecoverable error while evaluating node


thii:
Doesn't happen with `last_green` Bazel, so it must have been fixed somewhere.

https://buildkite.com/bazel/rules-swift-swift/builds/2741#19c25d8a-5530-4aff-8cca-d0998059ac84 (the log was for a build with 4.1.0 but I have reproduced this locally with 4.2.0rc3).

cc @katre

```
(08:35:41) FATAL: bazel crashed due to an internal error. Printing stack trace:
  | java.lang.RuntimeException: Unrecoverable error while evaluating node 'ConfiguredTargetKey{label=//examples/xplatform/hello_world:hello_world, config=BuildConfigurationValue.Key[58765066c537a2addebe865dea2674b9eb2786d3980bb8070f08dabcf685a2a8]}' (requested by nodes )
  | at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
  | at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:398)
  | at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
  | at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
  | Caused by: net.starlark.java.eval.Starlark$UncheckedEvalException: java.lang.IllegalArgumentException: hello_world.modulewrap.o (Starlark stack: [_swift_binary_impl@/workdir/swift/internal/swift_binary_test.bzl:343:61, _swift_linking_rule_impl@/workdir/swift/internal/swift_binary_test.bzl:249:50, register_link_binary_action@/workdir/swift/internal/linking.bzl:255:26, link@<builtin>])
  | at net.starlark.java.eval.Starlark.fastcall(Starlark.java:610)
  | at net.starlark.java.eval.Eval.evalCall(Eval.java:642)
[...]

PR #13838, opened 2021-08-12
[Cherry-pick request 4.2.0] Refactors CompilationSupport for objc to use existing API

thii:
Fixes https://github.com/bazelbuild/bazel/issues/13834

Original commit: 31b689ba7238abe5629a0de87fb6e245910526bb

---

This CL makes CompilationSupport use CcLinkingHelper instead of CppLinkActionBuilder. The former is the Java class used by the existing Starlark linking API.

This is in preparation for a future (unknown when) re-write of objc rules to
Starlark.

This is a rollforward after fixing an issue with implicit outputs of proto_libraries which had the j2objc_aspect applied on it. These artifacts were created with the genfiles dir while the C++ sandwich was always creating them in the bin dir.

RELNOTES:none
PiperOrigin-RevId: 353879792
(cherry picked from commit 31b689ba7238abe5629a0de87fb6e245910526bb)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingContext.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingContext.java
@@ -526,7 +526,7 @@ public class CcLinkingContext implements CcLinkingContextApi<Artifact> {
       return this;
     }
 
-    Builder addLinkstamps(List<Linkstamp> linkstamps) {
+    public Builder addLinkstamps(List<Linkstamp> linkstamps) {
       hasDirectLinkerInput = true;
       linkerInputBuilder.addLinkstamps(linkstamps);
       return this;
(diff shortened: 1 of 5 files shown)


katre:
Did you test this with the reproduction for #13834?

thii:
I'm testing it now.

thii:
I can confirm that this fixes the crash on Linux.

katre:
I don't understand all the Windows failures. @oquenchil, do you have any ideas?

katre:
Looks like the fix is 83dd93b8a11a734877a4f171c5b24cca43e9b17e. @meteorcloudy will send a PR for that, and then we can rebase this PR and get it merged.

Thanks!

meteorcloudy:
Please cherry pick https://github.com/bazelbuild/bazel/pull/13846

katre:
Just pushed the rebased version. If the tests pass, I'll merge this today.

Issue #12841, opened 2021-01-17
--build_python_zip fails when runfiles include file containing '=' character (affects pyspark==2.4.6)


thundergolfer:
### Description of the problem / feature request:

A Bazel Python 'zipapp' cannot be built using `--build_python_zip` when the underlying `py_binary` target depends on `pyspark==2.4.6`. I think this is because `pyspark` contains files that include the "=" character in their filename, which breaks some logic in the `--build_python_zip` action. 

**Example Error:**

```
INFO: Analyzed 2 targets (22 packages loaded, 708 targets configured).
INFO: Found 2 targets...
ERROR: /Users/jonathon/work/reproduce_zipapp_bug/spark_hello_world/BUILD:4:10: PythonZipper spark_hello_world/main.zip failed (Exit 255): zipper failed: error executing command external/bazel_tools/tools/zip/zipper/zipper cC bazel-out/darwin-fastbuild/bin/spark_hello_world/main.zip @bazel-out/darwin-fastbuild/bin/spark_hello_world/main.zip-0.params

Use --sandbox_debug to see verbose messages from the sandbox zipper failed: error executing command external/bazel_tools/tools/zip/zipper/zipper cC bazel-out/darwin-fastbuild/bin/spark_hello_world/main.zip @bazel-out/darwin-fastbuild/bin/spark_hello_world/main.zip-0.params

Use --sandbox_debug to see verbose messages from the sandbox
File kittens/date=2018-01/not-image.txt=external/pypi/pypi__pyspark/pyspark/data/mllib/images/partitioned/cls=kittens/date=2018-01/not-image.txt does not seem to exist.
INFO: Elapsed time: 2.814s, Critical Path: 0.35s
INFO: 2 processes: 2 internal.
FAILED: Build did NOT complete successfully
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I have a full public reproduction over in this repo: https://github.com/thundergolfer/bazel-build_python_zip-bug-reproduction (instructions in the README)

### What operating system are you running Bazel on?

MacOS Catalina 10.15.7

### What's the output of `bazel info release`?

`release 3.7.2`

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

> Replace this line with your answer.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```bash
[...]

PR #13850, opened 2021-08-15
Support `|` delimiter in ijar zipper to support `py_binary` zipping files with `=` in file name

benjaminRomano:
**Background**
https://github.com/bazelbuild/bazel/issues/12841

`py_binary` build zip fails when a file contains `=` since that character is used as a delimiter. I added support for another mode in the ijar zipper to support `|` character as a delimiter which should be less likely to occur as it is a restricted character on windows systems. However, it does not fully mitigate the issue. A more stable solution is multi-char delimiter but requires larger structural changes.

**Changes**
* Add support for `|` delimiter to ijar zipper

**Test Plan**
* Tests pass. E2E tested on pybinary repo


sgowroji:
Hello @benjaminRomano, Can you please fix the above presubmit buildkite failures ? Thanks!

PR #13851, opened 2021-08-15
Update default stub shebang to Python 3

olekw:
Last part of #11201.
@philwo


philwo:
Hi @olekw, thanks! I have the same change pending internally and already LGTM'd from my colleague, but I'm still trying to fix that one test on Windows that's failing...

I just gave it another try, let's see how it goes. I'll mark this PR as fixed once my change is submitted!

philwo:
The Windows version of Python doesn't ship a `python3.exe`, only a `python.exe`. 🤦🏻 

I'm going to fix this with `mklink "C:\path\to\symlink\python3.exe" "C:\path\to\Python3\python.exe`... (https://github.com/bazelbuild/continuous-integration/pull/1216)

PR #13856, opened 2021-08-16
Remove unused build target

katre:
This was in error because AndroidTest_lib was removed in
0f2c599582c73d93f8605bdd21cf16c945108571.

It's clearly unused since nothing else was bothered by this.


Wyverald (approved):
thanks!

PR #13867, opened 2021-08-18
Properly re-throw errors in aspect loading.

katre:
Apparently a description is required.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/BuildTopLevelAspectsDetailsFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/BuildTopLevelAspectsDetailsFunction.java
@@ -89,19 +89,13 @@ public class BuildTopLevelAspectsDetailsFunction implements SkyFunction {
       // duplicate aspects by keeping the first occurrence of each aspect.
       topLevelAspectsClasses = ImmutableSet.copyOf(topLevelAspectsClasses).asList();
 
-      try {
-        // Then return a list of indenpendent aspects to be applied on the top-level targets
-        ImmutableList<AspectDetails> aspectsDetailsList =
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/BuildTopLevelAspectsDetailsFunction.java
```diff
@@ -88,13 +89,19 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
-        return null; // some aspects are not loaded
+      try {
+        // Then return a list of indenpendent aspects to be applied on the top-level targets
+        ImmutableList<AspectDetails> aspectsDetailsList =
+            getIndependentTopLevelAspects(env, topLevelAspectsClasses);
+        if (aspectsDetailsList == null) {
+          return null; // some aspects are not loaded
+        }
+        return new BuildTopLevelAspectsDetailsValue(aspectsDetailsList);
+      } catch (AspectCreationException e) {
```
mai93, line 100: Can we add this exception catch to `getIndependentTopLevelAspects` to be similar to `getTopLevelAspects`?

mai93:
I am curious to know what cases the absence of this re-throw can affect? I can add a test for that to take care of it in the future as well.

src/main/java/com/google/devtools/build/lib/skyframe/BuildTopLevelAspectsDetailsFunction.java
katre: Done, now they are more similar.

katre:
I added a test that shows the case I observed: I was using an aspect from the command line and hadn't removed the `attrs`, so it couldn't be created, but the error message was swallowed.

mai93 (approved):

Issue #13859, opened 2021-08-17
dwp-less toolchains gracelessly handled


benjaminp:
```
$ bazel build --fission=yes --features=per_object_debug_info //src/main/cpp:*
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'ConfiguredTargetKey{label=//src/main/cpp:client, config=BuildConfigurationValue.Key[4df5be2a0e5e5dfafc200f2ec879cf999e610b09925c1e5f6ed5424aabd4d850]}' (requested by nodes 'ConfiguredTargetKey{label=//src/main/cpp:client.stripped, config=BuildConfigurationValue.Key[4df5be2a0e5e5dfafc200f2ec879cf999e610b09925c1e5f6ed5424aabd4d850]}', 'ConfiguredTargetKey{label=//src/main/cpp:client.dwp, config=BuildConfigurationValue.Key[4df5be2a0e5e5dfafc200f2ec879cf999e610b09925c1e5f6ed5424aabd4d850]}')
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:644)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:382)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
Caused by: java.lang.IllegalStateException
	at com.google.common.base.Preconditions.checkState(Preconditions.java:494)
	at com.google.devtools.build.lib.rules.cpp.CcBinary.createDebugPackagerActions(CcBinary.java:978)
	at com.google.devtools.build.lib.rules.cpp.CcBinary.init(CcBinary.java:627)
[...]

PR #13868, opened 2021-08-18
Remove precondition check related to toolchain dwp_files.

benjaminp:
This check always vacuously passed until recently. Even if it was desirable to check for non-empty dwp_files, erroring out should happen in a user-friendly way rather than crashing with a Preconditions failure.

Fixes https://github.com/bazelbuild/bazel/issues/13859.


oquenchil:
I agree, I guess this comes from a time when every toolchain was controlled by the Bazel team. In any case, to keep it as before would you mind doing as you say and give a user friendly error?

benjaminp:
I don't think an error would be correct because the toolchain might have an empty `dwp_files` but say the `dwp` executable is on the system, say `/usr/bin/dwp`.

oquenchil (approved):
LGTM

PR #13871, opened 2021-08-19
Wire up include scanning.

benjaminp:
Enable Bazel to use C/C++ include scanning if the --experimental_include_scanning flag is given. Include scanning can improve performance and incrementality by decreasing the size of compilation input trees. This change gives the community an opportunity to try it out. Include scanning must not be enabled by default because limitations of the builtin include parser can break builds.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/Bazel.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/Bazel.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.devtools.build.lib.analysis.BlazeVersionInfo;
 import com.google.devtools.build.lib.bazel.repository.starlark.StarlarkRepositoryDebugModule;
+import com.google.devtools.build.lib.includescanning.IncludeScanningModule;
 import com.google.devtools.build.lib.runtime.BlazeModule;
 import com.google.devtools.build.lib.runtime.BlazeRuntime;
 import java.io.IOException;
@@ -77,7 +78,8 @@ public final class Bazel {
           com.google.devtools.build.lib.metrics.PostGCMemoryUseRecorder.GcAfterBuildModule.class,
           com.google.devtools.build.lib.packages.metrics.PackageMetricsModule.class,
           com.google.devtools.build.lib.metrics.MetricsModule.class,
-          BazelBuiltinCommandModule.class);
+          BazelBuiltinCommandModule.class,
+          IncludeScanningModule.class);
 
   public static void main(String[] args) {
     BlazeVersionInfo.setBuildInfo(tryGetBuildInfo());
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
@@ -15,6 +15,7 @@
 package com.google.devtools.build.lib.bazel.rules.cpp;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.analysis.RuleContext;
 import com.google.devtools.build.lib.analysis.RuleErrorConsumer;
 import com.google.devtools.build.lib.analysis.TransitiveInfoCollection;
@@ -31,6 +32,7 @@ import com.google.devtools.build.lib.rules.cpp.CppActionNames;
 import com.google.devtools.build.lib.rules.cpp.CppCompileActionBuilder;
 import com.google.devtools.build.lib.rules.cpp.CppConfiguration;
 import com.google.devtools.build.lib.rules.cpp.CppConfiguration.HeadersCheckingMode;
+import com.google.devtools.build.lib.rules.cpp.CppFileTypes;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.SerializationConstant;
 
 /** C++ compilation semantics. */
@@ -78,15 +80,20 @@ public class BazelCppSemantics implements AspectLegalCppSemantics {
       RuleErrorConsumer ruleErrorConsumer) {
     CcToolchainProvider toolchain = actionBuilder.getToolchain();
     if (language == Language.CPP) {
+      CppConfiguration cppConfig = configuration.getFragment(CppConfiguration.class);
+      Artifact sourceFile = actionBuilder.getSourceFile();
       actionBuilder
           .addTransitiveMandatoryInputs(
-              configuration.getFragment(CppConfiguration.class).useSpecificToolFiles()
-                      && !actionBuilder.getSourceFile().isTreeArtifact()
+              cppConfig.useSpecificToolFiles() && !actionBuilder.getSourceFile().isTreeArtifact()
                   ? (actionBuilder.getActionName().equals(CppActionNames.ASSEMBLE)
                       ? toolchain.getAsFiles()
                       : toolchain.getCompilerFiles())
                   : toolchain.getAllFiles())
-          .setShouldScanIncludes(false);
+          .setShouldScanIncludes(
+              cppConfig.experimentalIncludeScanning()
+                  && featureConfiguration.getRequestedFeatures().contains("cc_include_scanning")
+                  && !sourceFile.isFileType(CppFileTypes.ASSEMBLER)
+                  && !sourceFile.isFileType(CppFileTypes.CPP_MODULE));
     } else {
       actionBuilder
           .addTransitiveMandatoryInputs(toolchain.getAllFilesIncludingLibc())
@@ -110,7 +117,7 @@ public class BazelCppSemantics implements AspectLegalCppSemantics {
 
   @Override
   public boolean allowIncludeScanning() {
-    return false;
+    return true;
   }
 
   @Override
--- a/src/main/java/com/google/devtools/build/lib/includescanning/LegacyIncludeScanner.java
+++ b/src/main/java/com/google/devtools/build/lib/includescanning/LegacyIncludeScanner.java
@@ -537,19 +537,13 @@ public class LegacyIncludeScanner implements IncludeScanner {
     }
     ArtifactRoot root = includer.getRoot();
     Artifact sourceArtifact =
-        artifactFactory.resolveSourceArtifactWithAncestor(
-            name, parentDirectory, root, RepositoryName.MAIN);
+        artifactFactory.resolveSourceArtifactWithAncestor(name, parent, root, RepositoryName.MAIN);
     if (sourceArtifact == null) {
       // If the name had up-level references, this path may not be under any package. Otherwise,
       // we must have gotten an artifact, since it should be under the same package as the
       // including artifact.
       Preconditions.checkState(
-          name.containsUplevelReferences(),
-          "%s %s %s %s",
-          name,
-          parentDirectory,
-          rootRelativePath,
-          root);
+          name.containsUplevelReferences(), "%s %s %s %s", name, parent, rootRelativePath, root);
     }
     return sourceArtifact;
   }
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
@@ -1006,6 +1006,25 @@ public class CppOptions extends FragmentOptions {
   public boolean useArgsParamsFile;
 
   @Option(
+      name = "experimental_unsupported_and_brittle_include_scanning",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.BUILD_TIME_OPTIMIZATION,
+      effectTags = {
+        OptionEffectTag.LOADING_AND_ANALYSIS,
+        OptionEffectTag.EXECUTION,
+        OptionEffectTag.CHANGES_INPUTS
+      },
+      help =
+          "Whether to narrow inputs to C/C++ compilation by parsing #include lines from input"
+              + " files. This can improve performance and incrementality by decreasing the size of"
+              + " compilation input trees. However, it can also break builds because the include"
+              + " scanner does not fully implement C preprocessor semantics. In particular, it does"
+              + " not understand dynamic #include directives and ignores preprocessor conditional"
+              + " logic. Use at your own risk. Any issues relating to this flag that are filed will"
+              + " be closed.")
+  public boolean experimentalIncludeScanning;
+
+  @Option(
       name = "experimental_objc_include_scanning",
       defaultValue = "false",
       documentationCategory = OptionDocumentationCategory.BUILD_TIME_OPTIMIZATION,
@@ -1175,6 +1194,7 @@ public class CppOptions extends FragmentOptions {
     host.parseHeadersSkippedIfCorrespondingSrcsFound = parseHeadersSkippedIfCorrespondingSrcsFound;
     host.strictSystemIncludes = strictSystemIncludes;
     host.useArgsParamsFile = useArgsParamsFile;
+    host.experimentalIncludeScanning = experimentalIncludeScanning;
 
     // Save host options for further use.
     host.hostCoptList = hostCoptList;
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ActionExecutionFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ActionExecutionFunction.java
@@ -93,6 +93,7 @@ import com.google.devtools.build.lib.util.Pair;
 import com.google.devtools.build.lib.util.io.FileOutErr;
 import com.google.devtools.build.lib.util.io.TimestampGranularityMonitor;
 import com.google.devtools.build.lib.vfs.FileSystem;
+import com.google.devtools.build.lib.vfs.OutputService.ActionFileSystemType;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import com.google.devtools.build.lib.vfs.Root;
 import com.google.devtools.build.skyframe.SkyFunction;
@@ -274,7 +275,11 @@ public class ActionExecutionFunction implements SkyFunction {
     }
 
     CheckInputResults checkedInputs = null;
-    NestedSet<Artifact> allInputs = state.allInputs.getAllInputs();
+    NestedSet<Artifact> allInputs =
+        state.allInputs.getAllInputs(
+            !skyframeActionExecutor
+                .actionFileSystemType()
+                .equals(ActionFileSystemType.STAGE_REMOTE_FILES));
 
     Iterable<SkyKey> depKeys = getInputDepKeys(allInputs, state);
     // We do this unconditionally to maintain our invariant of asking for the same deps each build.
@@ -614,37 +619,57 @@ public class ActionExecutionFunction implements SkyFunction {
       Preconditions.checkState(env.valuesMissing(), action);
       return null;
     }
-    return new AllInputs(allKnownInputs, actionCacheInputs, resolver.packageLookupsRequested);
+    return new AllInputs(
+        allKnownInputs,
+        actionCacheInputs,
+        action.getAllowedDerivedInputs(),
+        resolver.packageLookupsRequested);
   }
 
   static class AllInputs {
     final NestedSet<Artifact> defaultInputs;
+    @Nullable final NestedSet<Artifact> allowedDerivedInputs;
     @Nullable final List<Artifact> actionCacheInputs;
     @Nullable final List<ContainingPackageLookupValue.Key> packageLookupsRequested;
 
     AllInputs(NestedSet<Artifact> defaultInputs) {
       this.defaultInputs = checkNotNull(defaultInputs);
       this.actionCacheInputs = null;
+      this.allowedDerivedInputs = null;
       this.packageLookupsRequested = null;
     }
 
     AllInputs(
         NestedSet<Artifact> defaultInputs,
         List<Artifact> actionCacheInputs,
+        NestedSet<Artifact> allowedDerivedInputs,
         List<ContainingPackageLookupValue.Key> packageLookupsRequested) {
       this.defaultInputs = checkNotNull(defaultInputs);
+      this.allowedDerivedInputs = checkNotNull(allowedDerivedInputs);
       this.actionCacheInputs = checkNotNull(actionCacheInputs);
       this.packageLookupsRequested = packageLookupsRequested;
     }
 
-    NestedSet<Artifact> getAllInputs() {
+    /**
+     * Compute the inputs to request from Skyframe.
+     *
+     * @param prune If true, only return default inputs and any inputs from action cache checker.
+     *     Otherwise, return default inputs and all possible derived inputs of the action. Bazel's
+     *     {@link com.google.devtools.build.lib.remote.RemoteActionFileSystem} requires the metadata
+     *     from all derived inputs to know if they are remote or not during input discovery.
+     */
+    NestedSet<Artifact> getAllInputs(boolean prune) {
       if (actionCacheInputs == null) {
         return defaultInputs;
       }
 
       NestedSetBuilder<Artifact> builder = new NestedSetBuilder<>(Order.STABLE_ORDER);
-      // actionCacheInputs is never a NestedSet.
-      builder.addAll(actionCacheInputs);
+      if (prune) {
+        // actionCacheInputs is never a NestedSet.
+        builder.addAll(actionCacheInputs);
+      } else {
+        builder.addTransitive(allowedDerivedInputs);
+      }
       builder.addTransitive(defaultInputs);
       return builder.build();
     }
--- a/src/test/shell/bazel/cc_integration_test.sh
+++ b/src/test/shell/bazel/cc_integration_test.sh
@@ -1382,4 +1382,35 @@ EOF
       fail "bazel build should've succeeded with --features=compiler_param_file"
 }
 
+function test_include_scanning_smoketest() {
+  # Make sure there are no packages containing tools/cpp/INCLUDE_HINTS to exercise that case in
+  # IncludeHintsFunction.
+  rm -rf BUILD tools
+  mkdir pkg
+  cat > pkg/BUILD <<EOF
+cc_binary(
+  name = 'bin',
+  srcs = ['bin.cc'],
+  deps = [':spurious_dep'],
+)
+
+cc_library(
+  name = 'spurious_dep',
+  hdrs = ['dep.h'],
+)
+EOF
+
+  cat > pkg/bin.cc <<EOF
+#define NASTY "dep.h"
+#include NASTY
+int main() { return 0; }
+EOF
+
+  touch pkg/dep.h
+
+  bazel build --experimental_unsupported_and_brittle_include_scanning --features=cc_include_scanning //pkg:bin &>"$TEST_log" && fail 'include scanning did not (wrongly) remove dependency' || true
+  expect_log "Include scanning enabled. This feature is unsupported."
+  expect_log "fatal error: '\?dep.h'\?"
+}
+
 run_suite "cc_integration_test"
--- a/src/main/java/com/google/devtools/build/lib/bazel/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/BUILD
@@ -139,6 +139,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/lib/bazel/rules/ninja/actions",
         "//src/main/java/com/google/devtools/build/lib/buildeventservice",
         "//src/main/java/com/google/devtools/build/lib/dynamic",
+        "//src/main/java/com/google/devtools/build/lib/includescanning",
         "//src/main/java/com/google/devtools/build/lib/metrics:memory-use-recorder",
         "//src/main/java/com/google/devtools/build/lib/metrics:metrics_module",
         "//src/main/java/com/google/devtools/build/lib/network:noop_connectivity",
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/CcRules.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/CcRules.java
@@ -80,7 +80,7 @@ public class CcRules implements RuleSet {
     builder.addRuleDefinition(new BazelCppRuleClasses.CcLibraryBaseRule());
     builder.addRuleDefinition(new BazelCcLibraryRule());
     builder.addRuleDefinition(new BazelCcImportRule());
-    builder.addRuleDefinition(new CcIncludeScanningRule());
+    builder.addRuleDefinition(new CcIncludeScanningRule(/* addGrepIncludes= */ false));
     builder.addRuleDefinition(new FdoProfileRule());
     builder.addRuleDefinition(new FdoPrefetchHintsRule());
     builder.addRuleDefinition(new CcLinkingRule());
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BUILD
@@ -42,6 +42,7 @@ java_library(
     name = "bazel_cpp_semantics",
     srcs = ["BazelCppSemantics.java"],
     deps = [
+        "//src/main/java/com/google/devtools/build/lib/actions:artifacts",
         "//src/main/java/com/google/devtools/build/lib/analysis:analysis_cluster",
         "//src/main/java/com/google/devtools/build/lib/analysis:config/build_configuration",
         "//src/main/java/com/google/devtools/build/lib/analysis:transitive_info_collection",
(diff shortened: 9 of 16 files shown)


lberki:
/cc @comius 

At the very least, let's call this `--experimental_include_scanning` instead?

oquenchil:
We are not yet sure about adding support to include scanning in Bazel. We are discussing ways internally in which we can replace it with something else that will make the Bazel codebase healthier. For the time being what you can do is to write your build with single *.cc files in cc_libraries and use the implementation_deps attribute. That will have the same effect as include scanning and no unnecessary headers would be propagated.

ulfjack:
Hi @oquenchil!

If you would be so kind as to share with us what you're discussing, then we may be able to adjust our plans to help with that. We're happy to help replace include scanning with something better (if such a thing exists) and provide feedback based on prototypes (which we can also build).

Rewriting all C/C++ rules in all transitive dependencies to be fine-grained is clearly not a short term option and it's not clear whether that would be feasible (try convincing all C++ projects that use Bazel to rewrite all their rules!?!). It's not even clear if that would improve end-to-end build times. E.g., the increased analysis time and memory usage in Bazel could outweigh the savings in the remote execution overhead, not to mention the increased maintenance costs.

Note that the rest of the world doesn't necessarily run on Google's timeline - we want faster remote C/C++ builds and include scanning is the best option we are aware of right now. Last I tried, include scanning made about a 2x performance difference. We looked into `clang-scan-deps` as an option to avoid some of the shortcomings with the existing include scanning implementation, but that would still be conceptually very similar.

Let us know!

Thanks,

-- Ulf

oquenchil:
Hi Ulf,

Nice to hear from you. All the ideas that I have revolve around migrating our internal repo to implementation_deps; of course if this were ever to be done, it would have to be done automatically. The idea would be to use the grep_includes binary outside the build to apply fixes to the BUILD files and a presubmit check could be run for new changes to make sure deps/implementation_deps were used correctly. 

It could be that that cuts down headers by so much that include scanning is not needed anymore. I don't have numbers yet. If that were not enough then I would hope that the cases where include scanning has to be run can be replaced by a separate action (marked to be run locally), which puts all the pruned headers in a tree artifact and then passes them to the remotely executed compile action. I have yet to figure out how you could avoid being forced to run that local action (which will probably be slower than the current include scanning implementation) for every target downstream once we detect a single target upstream that needs it.

Those are ideas which might or might not work but I think are worth exploring. I would be very happy to use your expertise on this. If you can think of anything else that would allow ripping include scanning special casing out from the Bazel core codebase and push the complexity to a Starlark implementation of the C++ rules I would like to hear.

ulfjack:
Hi @oquenchil, we have tried some builds with implementation deps and include scanning and the results aren't promising so far, i.e., include scanning is about 4 times better than implementation deps. Any progress on your side?

ulfjack:
Hi @bjacklyn, we will share more data shortly. Also, apologies for the misunderstanding - we're seeing ~4x fewer header files sent to remote, not 4x performance difference.

oquenchil:
Hey Ulf,

Would you mind providing more data about how the graph for those builds you tried looks like? The idea of implementation_deps won't work too well if most of your libraries have more than one public header or have more than one source. With the former you can see how you will start accumulating unneeded headers, with the latter you can't distinguish which of the deps should actually be passed to each of the compilation actions of the source files in srcs.

Regarding tooling, we are planning to add support to build_cleaner to take care automatically about placing the dependencies in the correct place. If the headers of a dependency are used from hdrs, they go in deps, if they are used from srcs, they go in implementation_deps. For this to work. it's also important that if the cc_library has any private headers only used from srcs, that they are listed in srcs rather than hdrs.

For third party base libraries that cannot be modified which contain lots of headers, the idea would be to make sure those files stay cached in the remote execution machine.

I can tell you that although nothing has been settled completely the idea is gaining traction with C++ leads outside the Bazel team because of all the benefits it will have. [Quoting] It will likely be a huge speed boost for Bazel:
- Include scanning is the tall pole in C++-heavy builds
- We would start to benefit from memory gains as more inputs could be represented as nested sets
- We wouldn't need to store all the filenames in the persistent action cache
- We could actually properly represent architecture-specific includes (include scanner doesn't understand #ifs), although we'd might have to do those manually and not via build_cleaner
- (probably more)

Apart from those benefits I can tell you that making Starlarkification easier seems now the least significant incentive for everyone else. The current implementation of include-scanning is causing problems for other projects that are happening now due to the complexity it introduces.

As part of that work, this quarter I will be hard-coding implementation_deps in cc_proto_library. The main reason for tackling this first doesn't affect Bazel directly since it's related to module pruning but if you have a lot of proto_libraries in your build, it should definitely help reduce the number of headers propagated.

Let's find out where unneeded headers are getting propagated in the builds you have tried and maybe we can fix that.

ulfjack:
We have published our tools and results here: https://github.com/EngFlow/bazel-include-scanning-test

We ran our tests on the LLVM project (we had to fix a few things to make it work with remote execution). From the data, it looks like moving to `implementation_deps` requires splitting up every `cc_library` into separate libraries for each source file. This seems extraordinarily expensive in terms of migration cost as well as post-migration maintenance unless there is extensive automatic tooling in place to handle that.

I am doubtful that all of these things can be put into place *for projects outside of Google* in a reasonable amount of time, and I'd argue in favor of merging this patch as a temporary solution, even if there is confidence that they will land in the medium term.

AFAIUI, IWYU isn't widely used outside of Google, and if `build_cleaner` isn't oss right now, it obviously can't be used by anyone right now.

oquenchil:
I will need some time to go over your experimental results but at first glance I'm very surprised. What the results mean is that almost every header in your build is used by at least one public header and that the chain is almost never broken by a dependency only being used by a source file and none of its public headers.

ulfjack:
@oquenchil, we're happy to meet over VC if that helps.

oquenchil:
We can do that but first I'd like to understand more about the experiment. From the description on the link I am not able to see what was the top level target you built which has 588 header inputs with include-scanning and 2069 without.

Could you please give me the target name and the command needed to build it?

I want to understand the graph for that particular project using bazel query and perhaps an aspect that I write for debugging.

Thanks!

phildawson-ct:
The target we used to generate the data was [this one][aquery-target], added to [llvm bazel][llvm]. We applied [these patches][patches] in order to get the build working, though most of the changes to the .bazelrc and WORKSPACE are just remote execution configuration and can be ignnored.

cc: @ulfjack

[patches]: https://github.com/EngFlow/bazel-include-scanning-test/tree/main/benchmark-config/llvm-project
[llvm]: https://github.com/llvm/llvm-project/tree/main/utils/bazel
[aquery-target]: https://github.com/EngFlow/bazel-include-scanning-test/blob/main/benchmark-config/llvm-project/aquery-build-target.patch

oquenchil:
I think this project fits under:

> For third party base libraries that cannot be modified which contain lots of headers, the idea would be to make sure those files stay cached in the remote execution machine.

That's the idea that was floated around and there will definitely need to be a mechanism like that.

Bazel will have to provide a way to solve this. What I cannot tell you right now is whether implementation_deps will be judged to be a better solution long-term than include-scanning. I will try to get an answer to that by the end of this quarter. Meanwhile without further data proving that implementation_deps (plus the change used for base libraries that just contain headers like your example) does not suffice I'm hesitant to accept this PR.

If the answer is that implementation_deps cannot possibly replace include-scanning then we can revisit this.

ulfjack:
I think that the problem isn't that the files need to stay cached; commonly used files should always stay cached and any manual configuration on that is likely to go out of sync with reality. The problem is that *the client*, i.e., Bazel, has to hash and put those into the merkle tree. This is a serious bottleneck for mostly cached builds.

My concern is that a) you won't have enough data by EOQ, and b) even if there is a decision to go with `implementation_deps`, building and rolling out all the automation necessary to do that w/ Bazel is going to take years, even if Google were to put serious manpower behind it. As an outsider, I'm concerned that Google might just do whatever seems best for Google and leave the community in the cold. I would really appreciate if we could work on this together.

bjacklyn:
@oquenchil hopefully you don't mind me adding my own 2 cents here, but I generally agree with @ulfjack and have the same concerns about Google solving this such that it works for Google and not for the wider community. It seems like Google is quite disconnected from the pain the rest of us have with the cc_rules because Google has a bunch of things hooked up in bazel that the rest of us don't. For example, a few months ago a heuristic was added to estimate the ram required to run compilations based on the number of hdrs in the dependency graph. At Google it looks like internally it worked great, but anyone who using a few libraries from rules_boost could easily be transitively be pulling in 10,000+ hdrs. This caused the heuristic to estimate 7+ GB was needed per compilation severely reducing the parallelism of our builds. My guess for why it wasn't caught by Google is because include scanning would have pruned the majority of hdrs leading to a reasonable estimate of ram.

Regarding `impl_deps`, in my mind it won't be adopted by the community without significant tooling that can:

1. Automatically generate single hdr/src cc_libraries
2. Automatically manage the deps/impl_deps of these libraries

Speaking of rules_boost this means the maintainer will need this tooling to convert the 156 existing libraries into 16,231 libraries, which might cause other issues with bazel memory usage, and more time required to parse all these libraries/BUILD files.
```
external/boost $ find . -name *.hpp | wc -l
16231

external/boost $ grep -r boost_library . | wc -l
156
```

And the tooling will need to deal with everything from combining circular dependencies into same libraries (i.e. when foo.cpp --> bar.hpp and bar.cpp --> foo.hpp you have to stick foo and bar in the same library), handling deps/impl_deps when `#include`s are behind custom/platform `#ifdef` flags, possibly handling `#include <macro>` statements, etc.

So I'm wondering what your thoughts are on what the success criteria for impl_deps will be such that this PR will not be necessary. I'm using boost as an example because it is one of the most popular c++ projects out there, and being able easily to port it (and other popular OSS projects) to impl_deps is the minimal success criteria in my mind, but I'm curious WDYT?

Also hopefully I'm not coming off as too negative here because I really appreciate your efforts on this topic. This & the lack of c++ (clang? cpp20?) modules/precompiled hdrs support are probably the top 2 pain points my org has with the cc_rules in terms of build incrementality & build speed. I would also love it if the existing clang module support in bazel could be wired up to an experimental flag as well, as well as any other cc_rules goodies Google has enabled in blaze that are disabled for the community in bazel.

lberki:
I don't think "leaving the community in the cold" is a fair assessment of the situation. We simply don't want to commit to the indefinite maintenance of some complicated machinery that, to the best of our knowledge, has only marginal benefit.

How about this:

1. Y'all (@ulfjack @oquenchil ) agree upon whether include scanning is useful in the "ideal" future. To the best of my knowledge, your investigations yielded diametrically opposite results and I would like to get to the bottom of it.
   
   From first principles, the answer should be the same for Google's monorepo and for every other use case; at least as far as C++ is concerned, said monorepo is not _that_ special (@oquenchil : feel free to talk freely about the general shape of the C++ dependency graph there, I don't think it's a big secret e.g. how many headers per target we have or how deep is our dependency graph) 

   In case the answer is different for Google and everyone else, I would like to know why. In case changes to e.g. Bazel's RBE interface are needed to get Google and everyone else to agree, that's fine; the include scanning machinery is complicated enough and has enough potential for bugs that its eventual removal is worth a lot to us.
2. We draw up a roadmap as to how to get to that ideal state; doesn't need to be very detailed, what's important is rough timelines from which we can discern whether "temporary" solutions are needed.
3. Once we have a roadmap, we agree upon what to do in the near future. If it turns out that include scanning is needed indefinitely, well, we'll need to swallow that bitter pill. If not, I can imagine any number of temporary solutions. For example:
   * Hiding include scanning behind a `--experimental_dont_even_bother_filing_bugs_this_can_go_away_in_any_point_release` flag
   * Offering some sort of public interface (eventually) behind which include scanning can be implemented if we eventually delete it at Google
   * Something else that I haven't thought of while writing this comment?

ulfjack:
That seems like a surprising statement. The last time I compared include scanning vs. no include scanning, I saw *fully cached* builds of a medium sized Google C++ project run ~2x faster end-to-end with include scanning, *including* the overhead of the include scanning implementation.

I note that AOSP also uses include scanning for remote execution with Soong / Ninja (last I checked), as well as a 'dedicated' include scanning project in upstream clang.

The reason for the benefit is that it reduces the per-file costs for Skyframe + action cache + remote execution cache key, which are all on the critical path for fully cached builds.

I find it doubtful that `implementation_deps` can match those benefits, given that:

1. It requires an increase in the number of `cc_library` rules by a significant factor (additional rules increase loading+analysis cost as well as skyframe nodes + pre-execution dependencies; both CPU+Memory)
2. We don't currently have a solution for toolchain headers

In addition:

3. It requires substantial automation of `cc_library` dependencies to avoid increasing manual maintenance costs
4. It requires a substantial migration for all major C++ projects using Bazel

There are (at least) two special properties of the Google monorepo that are applicable here:

1. Google has significant bespoke automation for changing and maintaining the code; if my point 3 is correct, then Bazel users will have a hard time adopting `implementation_deps` at scale without such tooling
2. Google has an empty WORKSPACE file - that means it largely doesn't depend on BUILD files maintained by third parties; for a Bazel user to see a benefit from `implementation_deps`, a potentially large number of upstream projects have to migrate

Wrt. working together, I find this statement particularly disappointing:
> *I* will try to get an answer to that by the end of this quarter.

Why not work together on this instead?

janakdr:
> I think that the problem isn't that the files need to stay cached; commonly used files should always stay cached and any manual configuration on that is likely to go out of sync with reality. The problem is that _the client_, i.e., Bazel, has to hash and put those into the merkle tree. This is a serious bottleneck for mostly cached builds.

Internally at Google, we are trying to support directories as inputs more natively. This effort is first focused on output directories (so we can finally achieve parity with RBE!), but we expect it to also encompass input directories. This would hopefully drive @ulfjack's prior work on "directory artifacts" to completion. Toolchain headers could be such an input directory. With FUSE, Blaze can completely avoid digesting such a directory, relying on a precomputed xattr provided by the filesystem. Without FUSE, Bazel would need to digest the directory once per server lifetime, but could cache the (essentially Merkle) hash and reference it on incremental builds. I don't know if that is enough for the "mostly cached" case you're referencing here, @ulfjack, but I assume you're not focused on cold incremental builds, so perhaps it would be sufficient?

oquenchil:
Thank you everyone.

I'm happy to go with Lukacs's idea so that we don't incur the maintenance/support costs until we figure this out:
> Hiding include scanning behind a --experimental_dont_even_bother_filing_bugs_this_can_go_away_in_any_point_release flag

If we get to a state where the alternative to include-scanning is a reasonable match and the available tools make migration/maintenance of open source projects realistic, we can follow by setting a generous deadline after which include-scanning will be deleted (if we ever get there, if on the other hand it becomes clear we won't we can remove the experimental flag).

src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
```diff
@@ -78,15 +78,15 @@ public void finalizeCompileActionBuilder(
       RuleErrorConsumer ruleErrorConsumer) {
     CcToolchainProvider toolchain = actionBuilder.getToolchain();
     if (language == Language.CPP) {
+      CppConfiguration cppConfig = configuration.getFragment(CppConfiguration.class);
```
oquenchil, line 81: Let's also add a warning at the top of this method:

`ruleErrorConsumer.ruleWarning("Include scanning is turned on. This feature is unsupported.");`

oquenchil (changes requested):

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
```diff
@@ -63,20 +65,14 @@ public IncludeHintsFunction(PathFragment hintsFile) {
...
-        String reasonForNoContainingPackage = hintsLookupValue.getReasonForNoContainingPackage();
-        String message =
-            String.format(
-                "INCLUDE_HINTS file %s was not in a package%s",
-                hintsFile,
-                reasonForNoContainingPackage != null ? ": " + reasonForNoContainingPackage : "");
-        throw new IncludeHintsFunctionException(
-            new EnvironmentalExecException(
-                createFailureDetail(message, Code.INCLUDE_HINTS_FILE_NOT_IN_PACKAGE)));
+        return HintsRules.EMPTY;
```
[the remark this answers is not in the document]
benjaminp: It's possible to have a valid workspace with no package containing `tools/cpp/` if there are no `BUILD` files along that path.

src/main/java/com/google/devtools/build/lib/skyframe/ActionExecutionFunction.java
```diff
@@ -274,7 +275,11 @@ private SkyValue computeInternal(SkyKey skyKey, Environment env)
     }
 
     CheckInputResults checkedInputs = null;
-    NestedSet<Artifact> allInputs = state.allInputs.getAllInputs();
+    NestedSet<Artifact> allInputs =
+        state.allInputs.getAllInputs(
+            !skyframeActionExecutor
```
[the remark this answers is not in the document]
benjaminp: My understanding is that blaze can always assume derived input headers exist (at least virtually) and can be statted. Ulf and I discussed this, but he couldn't recall exactly how it all fits together internally. (E.g., when remote `grep-includes` is used, how is metadata for the header found? The forge client must be able to find metadata by path somehow.) I assume janakr knows the full story.

src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
```diff
@@ -78,15 +78,15 @@ public void finalizeCompileActionBuilder(
...
           .addTransitiveMandatoryInputs(
-              configuration.getFragment(CppConfiguration.class).useSpecificToolFiles()
-                      && !actionBuilder.getSourceFile().isTreeArtifact()
+              cppConfig.useSpecificToolFiles() && !actionBuilder.getSourceFile().isTreeArtifact()
                   ? (actionBuilder.getActionName().equals(CppActionNames.ASSEMBLE)
                       ? toolchain.getAsFiles()
                       : toolchain.getCompilerFiles())
                   : toolchain.getAllFiles())
-          .setShouldScanIncludes(false);
+          .setShouldScanIncludes(cppConfig.shouldScanIncludes());
```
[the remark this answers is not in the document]
benjaminp: This brings up an idea. Do you think it would be better to remove the `--experimental_include_scanning` flag and instead have a `experimental_unsupported_include_scanning` feature that users can enable with `--features`? That would be more configurable and more consistent with the internal world. (Or is the condition before the first `&&` you pasted also checking an option?)

src/main/java/com/google/devtools/build/lib/bazel/Bazel.java
```diff
@@ -76,7 +76,8 @@
           com.google.devtools.build.lib.metrics.PostGCMemoryUseRecorder.GcAfterBuildModule.class,
           com.google.devtools.build.lib.packages.metrics.PackageMetricsModule.class,
           com.google.devtools.build.lib.metrics.MetricsModule.class,
-          BazelBuiltinCommandModule.class);
+          BazelBuiltinCommandModule.class,
+          com.google.devtools.build.lib.includescanning.IncludeScanningModule.class);
```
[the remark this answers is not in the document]
benjaminp: Sure; it's not clear what the style of this `BLAZE_MODULES` list is supposed to be. Most but not all of the entries are fully-qualified.

src/main/java/com/google/devtools/build/lib/includescanning/LegacyIncludeScanner.java
```diff
@@ -523,8 +523,7 @@ private Artifact locateRelative(
     if (!isFile(execPath, name, includer.isSourceArtifact(), headerData)) {
       return null;
     }
-    PathFragment parentDirectory = includer.getRootRelativePath().getParentDirectory();
-    PathFragment rootRelativePath = parentDirectory.getRelative(name);
+    PathFragment rootRelativePath = includer.getRootRelativePath().getRelative(name);
```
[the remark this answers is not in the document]
benjaminp: Hmm, yes I should revert these two lines, which should hopefully fix those tests.

I do have to modify below, though. `ArtifactFactory.resolveSourceArtifactWithAncestor` was being called incorrectly. The second parameter to that method is supposed to be a execpath, but previously a root-relative path, `parentDirectory` was being passed. That matters when locating relative includes within external repositories.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
```diff
@@ -1004,6 +1004,25 @@ public Label getPropellerOptimizeLabel() {
...
+      name = "experimental_unsupported_and_brittle_include_scanning",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.BUILD_TIME_OPTIMIZATION,
+      effectTags = {
+        OptionEffectTag.LOADING_AND_ANALYSIS,
+        OptionEffectTag.EXECUTION,
+        OptionEffectTag.CHANGES_INPUTS
+      },
+      help =
+          "Whether to narrow inputs to C/C++ compilation by scrapping dependencies from input"
```
janakdr, line 1017: nit: is this supposed to be "scraping dependencies"? Might be worth rephrasing either way.

src/test/shell/bazel/cc_integration_test.sh
```diff
@@ -1382,4 +1382,29 @@ EOF
...
+
+  cat > bin.cc <<EOF
+#define NASTY "dep.h"
+#include NASTY
+int main() { return 0; }
+EOF
+
+  touch dep.h
+
+  bazel build --experimental_unsupported_and_brittle_include_scanning --features=cc_include_scanning //:bin && fail 'include scanning did not (wrongly) remove dependency' || true
```
janakdr, line 1407: please add an assertion on the error message so the test doesn't break for some other reason

src/main/java/com/google/devtools/build/lib/skyframe/ActionExecutionFunction.java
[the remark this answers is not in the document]
janakdr: There's nothing special happening internally, and it shouldn't be different from Bazel. Blaze will explicitly build any derived input headers, and their digests are available in memory. The only difference between Bazel and Blaze here should be BwtB, where a generated (but not downloaded) header needs to be locally include scanned. Blaze would handle this via the FUSE mount for output files. How is Bazel going to handle it?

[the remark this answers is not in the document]
benjaminp: Bazel BwoB requires having the metadata available to know if a file needs to be downloaded. As you say, derived input headers are necessarily built—or in the case of BwoB, in a remote cache—by the time dependent include scanning runs. But, without the changes in this file, derived input header metadata is not available to the include scanning action to know if the derived headers need to be downloaded.

When I was looking into this issue, I thought there would be an analogous issue with remote include scanning. Because in order to run `grep-includes` on a header, you need to have the metadata for the header to tell the remote execution system. How does the blaze execution client figure out the metadata of the inputs to this spawn? https://github.com/bazelbuild/bazel/blob/5149408a6c2500cb1701880575c41640d4fd1c2f/src/main/java/com/google/devtools/build/lib/includescanning/SpawnIncludeScanner.java#L354-L355 Is it just statting them in the FUSE output tree?

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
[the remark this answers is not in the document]
benjaminp: I will adapt the existing `test_include_scanning_smoketest` test to exercise this case.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
benjaminp: Good catch. I will replace it with "parsing".

src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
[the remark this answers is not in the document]
oquenchil: Yes, the condition before && is also a flag. So I think checking the cc_include_scanning feature and the experimental flag is closer to what we have internally. Even though there aren't CPP_MODULES in Bazel, if you add that to the condition as well then it's a perfect match except for the flag names.

src/main/java/com/google/devtools/build/lib/bazel/Bazel.java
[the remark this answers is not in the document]
oquenchil: I see, I didn't notice the others. importing is the right way according to our presubmit checks. You can leave the others above as they are since you didn't touch those.

src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
oquenchil: Indeed. Let's have the warning once per invocation by adding it in IncludeScanningModule.beforeCommand when the experimental flag is passed.

src/main/java/com/google/devtools/build/lib/includescanning/LegacyIncludeScanner.java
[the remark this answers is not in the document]
oquenchil: Yes, both pass now. Thanks!

src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCppSemantics.java
benjaminp: added

oquenchil: Thanks Benjamin. Once the other threads where Janak is helping are resolved I can merge.

benjaminp:
@janakdr How do you want to move forward?

janakdr (approved):
> @janakdr How do you want to move forward?

Sorry, github apparently forked my comment so I didn't see your reply since I was looking at the second copy: 
image

benjaminp:
@oquenchil I believe we have all the necessary sign-offs now.

oquenchil:
Sorry, for some reason I thought we were waiting for @coeuvre and @alexjski but I have just realized they were only cc'd. I will merge now.

oquenchil (approved):
LGTM

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
```diff
@@ -990,9 +990,24 @@ public static CcDebugInfoContext mergeCcDebugInfoContexts(
         .collect(ImmutableList.toImmutableList());
   }
 
+  private static final Label NOOP_GREP_INCLUDES =
+      Label.parseAbsoluteUnchecked("@bazel_tools//tools/cpp:grep-includes");
+
   public static Artifact getGrepIncludes(RuleContext ruleContext) {
```
lberki, line 996: Why are these changes needed? As far as I understand, this method is called even at HEAD from many places and this pull request doesn't alter the fact that this method returns null (my immediate instinct would be to move the function to get the dependency to CppSemantics, but I don't know the background)

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
```diff
@@ -63,20 +65,14 @@ public IncludeHintsFunction(PathFragment hintsFile) {
...
       }
       hintsPackageRoot = hintsLookupValue.getContainingPackageRoot();
-      env.getValueOrThrow(FileValue.key(RootedPath.toRootedPath(hintsPackageRoot, hintsFile)),
-          IOException.class);
-    } catch (IOException | BuildFileNotFoundException e) {
+      env.getValueOrThrow(
```
lberki, line 71: What is the plan for tools/cpp/INCLUDE_HINTS in Bazel? The file is at kind of a strange place for Bazel because it's the only place where Bazel reads a file from a specific location in the main repository.

Can we get away with never reading it so that the Bazel code path is a bit simpler?

lberki (commented):
These questions are kinda simple, but it looks like no one has asked them on this review thread yet. Or maybe I skimmed over that part of the conversation?

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
benjaminp: This change is to make `getGrepIncludes` always return `null` with Bazel. Bazel has a `grep-includes.sh` in `tools/cpp/grep-includes.sh` that unconditionally fails. It's harmless to pass that around as an `Artifact` until include scanning decides it wants to make a remote include scanning spawn, which is a possibility now for Bazel under this CL. If `getGrepIncludes` always returns `null`, include scanning will never attempt remote include scanning.

Moving to the `*Semantics` classes sounds fine to me if that's what you prefer. There are two process hiccups: 1) the CL size would go up due to modifying the "many" callsites you mention. 2) `GoogleCppSemantics.java` is outside my ability to edit, making the import process more complex.

The change here is admittedly not the nicest. Before modifying this method, I tried a simpler approach: removing `CcIncludeScanningRule`, which provides `grep-includes` as an implicit attribute, as an ancestor from all the Bazel cc rules. That worked elegantly for C++. Unfortunately the Objective-C rules use `CcIncludeScanningRule`, too, and do not appear to be forked between Blaze and Bazel.

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
benjaminp: I intended that `INCLUDE_HINTS` be functional with Bazel because it may be useful for third party code with computed includes. That said, even though you can provide a `INCLUDE_HINTS` file under this CL, it's fairly useless, since, for example, it doesn't accept hints outside of a hardcoded `third_party/` directory. I thought that could be fixed in the future if desired.

I could bypass the include hints code in Bazel only. However, working on workspaces that don't have an `INCLUDE_HINTS` file isn't a wild extension in functionality or code complexity for `IncludeHintsFunction`; `git` says `IncludeHintsFunction.java` has 12 deletions and 10 insertions in this CL.

On a more technical note, I don't know what the preferred method to make a product-based branch for this code would be. A new semantics method? `BlazeRuntime.getProductName().equals("Bazel")`?

benjaminp:
> These questions are kinda simple, but it looks like no one has asked them on this review thread yet. Or maybe I skimmed over that part of the conversation?

Thank you for the comments; I don't recall addressing these particular issues before here.

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
lberki: I dont mind too much; since this is experimental functionality, ugliness like reading a random file with a random name from the main repository is not too bad, but I'd still prefer to code to be clean if we can help.

Looking at the code, I think the best alternative would be to set `IncludeScanningModule.INCLUDE_HINTS_FILENAME` to null (we already override it within Google so it would be a no-op for our monorepo) and change `IncludeScanningModule.getSkyFunctions()` and `IncludeHintsFunction` so that if the value is null, they simply return the empty include hints.  Looks like this is doable with a minimal amount of effort.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
lberki: Yeah, this hack stuck out like a sore thumb when reviewing this change :)

I can imagine two possible approaches:
1. The more hacky (but still much nice than this one) approach is to give the constructor of `CcIncludeScanningRule` a Boolean flag that tells whether to add the extra attribute or not. This requires a single line of change in our internal-only code.
2. The more principled approach is moving the "use remote include scanning" bit to `CppSemantics`. The internal-only changes caused by this would be pretty straightforward but it's indeed a complication and I'd much rather to this refactoring atomically. 

So I guess (1) it is, unless @oquenchil thinks otherwise?

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
benjaminp: I've now implemented your suggestion to `IncludeScanningModule` and reverted the changes to `IncludeHintsFunction`.

src/main/java/com/google/devtools/build/lib/rules/cpp/CppHelper.java
benjaminp: I've now implemented approach 1, which is definitely less lines of hacks than messing with `getGrepIncludes`.

lberki: Thanks. It looks much nicer :)

src/main/java/com/google/devtools/build/lib/includescanning/IncludeHintsFunction.java
lberki: Thanks! One fewer hackj.

lberki:
Looks good to me. @oquenchil , if you are find with this change as it is, can you import it?

benjaminp:
Thank you lberki, janakr, and oquenchil for your reviews and assistance in getting this submitted.

PR #13872, opened 2021-08-19
Remove --ios_cpu flag

keith:
This flag appears to be unused, it was set once but never read. Instead
users should prefer `--ios_multi_cpus` or `--cpu=ios_XXX`


allevato:
This one would need some internal repo cleanup before we could accept it; we still have a few teams using `--ios_cpu` in some scripts and CI config files.

aiuto:
@allevato   Should we close this PR and assume you will do it after the internal cleanup?

allevato:
Yes, this is cleanup that we'll most likely do after we've finished the transition over to platforms. (I don't see a need to rush it at this point; it would be better to tidy up all of that legacy stuff holistically.)

aiuto:
@allevato, cc: @gregestren, @katre 

Tony: I see fewer ios_cpu instances each month.  Do you have a status update on internal movement.

keith:
seems like this was mostly handled in https://github.com/bazelbuild/bazel/commit/ecf9ac90f34c72562fc4a3fd4525c660082ef371

PR #13874, opened 2021-08-19
Remove unused apple API

keith:
These starlark methods have been deprecated for a long time, with viable
replacements, and are not used in the core iOS rules (or anywhere else I
could find). This also inlines a few helpers that were only used once
after this cleanup.


allevato:
I _think_ this one should be safe. A dumb regex search didn't turn up any `.bzl` files still using `\.ios_cpu` or `\.ios_cpu_platform` anywhere in our monorepo.

allevato (approved):

comius (changes requested):
I checked internally and on github and it looks like this can be safely removed.

Please also remove the failing tests that cover those two functions.

comius (approved):

comius:
cc @kaylathar

PR #13879, opened 2021-08-19
Cache merkle trees

### LINKED ISSUE #10875: Reuse input map and merkle tree for tools with lots of inputs

moroten:
MerkleTree calculations are now cached for each node in the input NestedSets (depsets). This drastically improves the speed when checking for remote cache hits. One example reduced the Merkle tree calculation time from 78 ms to 3 ms for 3000 inputs.

This caching can be disabled using --remote_cache_merkle_trees=false which will reduce the memory footprint. The caching is discarded after each build to free up memory, the cache setup time is negligible.

Fixes #10875.


src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -133,6 +136,7 @@
   @Nullable private final RemoteExecutionClient remoteExecutor;
   private final ImmutableSet<PathFragment> filesToDownload;
   @Nullable private final Path captureCorruptedOutputsDir;
+  private final ConcurrentHashMap<Object, MerkleTree> merkleTreeCache = new ConcurrentHashMap<>();
```
moroten, line 139: Shall I use soft references to the `MerkleTree`s or some other kind of map to reduce the memory footprint?

coeuvre:
Thanks for the PR! I like the performance boost.

It touches some areas that I'm not familiar with yet so it may take me a little while to give the feedback.

Can we change the switch flag to `--experimental_remote_cache_merkle_trees` and make it `false` by default?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
moroten: I got out of memory using `ConcurrentHashMap`. Switching to `Cache<Object, MerkleTree> merkleTreeCache = CacheBuilder.newBuilder().softValues().build()` works for me. I'll fix that and all the other comments (and test errors) next week.

moroten:
Fixed a recursion problem with `inputBytes` and `inputFiles` which made the previous attempt too slow. Also, introduced an explicit parameter for the cache size due to out of memory problems. Try values in the range 1000-10000.

A test is now available, but it looks a bit clumsy. Is that kind of test the right way to go to check if the caching is actually taking effect?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -136,7 +137,8 @@
   @Nullable private final RemoteExecutionClient remoteExecutor;
   private final ImmutableSet<PathFragment> filesToDownload;
   @Nullable private final Path captureCorruptedOutputsDir;
-  private final ConcurrentHashMap<Object, MerkleTree> merkleTreeCache = new ConcurrentHashMap<>();
+  private final Cache<Object, MerkleTree> merkleTreeCache =
+      CacheBuilder.newBuilder().softValues().build();
```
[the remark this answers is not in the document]
moroten: Introduced `--experimental_remote_merkle_tree_cache_size` for now.

moroten:
I don't have any updated measurements and is not working actively on this. One old example of 3000 inputs cut the calculation time from 78 ms to 0.7 ms (or even less). With the fast alias hash I've tried before, I got down to 0.1 ms. Bazel has more overhead, in the range of 10 ms for my example action, so 1 ms or 0.1 ms doesn't really matter.

There is a bug in the new `getInputMapping()` that I've heard about. More details and/or a fix will come as soon as I know more.

torgil:
> There is a bug in the new `getInputMapping()` that I've heard about. More details and/or a fix will come as soon as I know more.

It's a change of behavior for the case where the same file appears in different nested sets but with different casing in Windows.

An example leading to this is the winsock2 interface library in Microsoft CLibs package which is named "um/x64/WS2_32.Lib", a file ending which cc_import currently doesn't allow. It's however fine with "um/x64/WS2_32.lib" and if the same file is also a dependency from a glob in the toolchain we end up with two objects for the same file in different nested sets with different casing.

Assume we have nested set with a ".lib" version and a nested set with a ".Lib" version of the same file. 

`old getInputMapping():`
"expandArtifacts" is called which loops over inputs.toList() which ignores the file ending with .Lib
resulting inputMap entry:  key= ".lib" PathFragment,  value = ".lib" ActionInput

`new getInputMappipng():`
inputMap.putAll is called for each set which overrides the value but not the key
resulting inputMap entry:  key= ".lib" PathFragment,  value = ".Lib" ActionInput

This mismatch later on confuses ActionMetadataHandler which fails with java.lang.IllegalStateException on this line:
`checkState(forInputDiscovery, "%s is not present in declared outputs: %s", artifact, outputs);`

The immediate solution for us here is to allow cc_import to accept ".Lib" file-endings and maybe fail if rules specify the wrong casing instead (to avoid similar issues). Fine-grain dependencies such that the file only occurs once also avoids this issue.

Given solution above, I'm okay with this patch. If you want to retain previous behavior you could modify or replace "putAll" but I'm unsure what the proper solution is here.

Wyverald:
@coeuvre is out of office until next week. I'll ping him again by then. Sorry for the delay!

moroten:
I've rebased and resolved the merge conflicts.

> Sorry for the delay. LGTM generally but would like to have some changes before merging:
> 
> 1. Can we move the newly added `walkInputs` method and its related interfaces/implementations to a place that is private to remote module? Reasons:
>    
>    * `SpawnExecutionContext` is also used internally. Adding methods there makes it hard to import.
>    * This PR tries to improve merkle tree performance for remote execution. Keeping the change within remote module reduce the risk to introduce regressions for other spawn strategies.

`SpawnInputExpander` needs to implement `walkInputs()` and the only reference is from the private member `AbstractSpawnStrategy.spawnInputExpander`. Is it okay to make `spawnInputExpander` and `SpawnExecutionContextImpl` "protected" so they can be overridden and extended by `RemoteSpawnStrategy`? Then `RemoteExecutionService.buildInputMerkleTree()` can check if `context` is of a special `RemoteSpawnExecutionContext` (or a better name) which then should support `walkInputs()`. Does anyone know another better way of routing?

> 2. `SpawnInputExpander#getInputMapping` is also used by other spawn strategies and getting it right is crucial. Can we keep its original implementation untouched (rather than reuse `walkInputs`) as starter and make sure we use the original one when the switch is off?

Done, removed the commit _Reimplement getInputMapping() using walkInputs()_. `SpawnInputExpander` does not have access to `RemoteOptions`, so it would be a bit of plumbing to switch `getInputMapping()` depending on the flag. When point 1 gets sorted, it might be possible to switch on the flag again.

> 3. Can we use two flags, one boolean flag for whether enabling merkle tree cache and another one for controlling the cache size?

Done. Default size is set to unlimited, but maybe 1000 or 10000 would be more usable.

moroten:
> > In some way `spawnInputExpander` needs to be accessed. What about adding `SpawnExecutionContext.getSpawnInputExpander()`? Then `RemoteExecutionService` would call `context.getSpawnInputExpander().walkInputs()`. Instead of polluting the interfaces with `walkInputs()`, the `spawnInputExpander` will be exposed. Is that a good enough solution?
> 
> SGTM.

Done the implementation but some tests fail. Will check that in the afternoon.

sluongng:
A warning for whoever wanting to enable this flag in their build: do NOT use the default value of `experimental_remote_merkle_tree_cache_size` at zero. 

It essentially an **in-memory** cache, constructed using https://github.com/ben-manes/caffeine/, to help speed up input resolver before using remote cache or remote exec. The flag being in remote execution section of Bazel 5.0 blog was giving me a false impression that this cache things on the remote_cache instead of Bazel JVM memory.

And because this is in-memory and not remotely cached, you would definitely want to cap it at a certain size using `experimental_remote_merkle_tree_cache_size` to avoid OOM killing Bazel.

I would suggest setting the default cache size to 1000-3000 by default just to give folks a bit of a sane value to start with. And perhaps updating the doc that optimal cache size value needs to be calibrated per-workspace.

@moroten wdyt?

sluongng:
I created https://github.com/bazelbuild/bazel/pull/14959

EricBurnett:
@coeuvre is there a plan, or evaluation you're looking for, to turn this on by default? 

If anyone on this bug has data on the memory / latency tradeoff of turning on this cache for their own builds, especially at the default size (1000 from #14959), that might be useful.

(I ask only as an interested observer - if the cache is as useful as it sounds like, i.e. the experiment is considered a success, I'd like everyone to get the benefit of it by default and not have to know they need to manually opt in.)

sluongng:
I think a more mature way to approach this is:

1. Provide a way to measure these adhoc cache out of the box. I have been looking into this during my off hours and I think we could include metrics of all the caffeine cache stores into buildMetrics event in bes.proto if `--memory_profile` is set. However that does require some work to wire up everything which I don't have the capacity today

2. Provide a saner default caching strategy. Perhaps we could provide some sort of "adaptive" strategy where Bazel set the size of this cache conditionally based on some criterias? It's way better than `1000`, which is a completely abitrary non-zero value to ensure caffeine cache performance.

If we have either one of the above, I think it would be a lot easier to transition this into non-experimental.

coeuvre:
I am not confident enough to turn this on by default:
  - Discussion in #16054. I am not sure enabling this could improve it or make it worse.
  - I remember @tjgq did the benchmark of building Bazel with this flag turned on and the result was not promising. /cc @tjgq
  - I would like to see more data/benchmarks for this flag, not only for the wall time but also from the memory perspective.
  - and yes, `1000` is probably not a good default value.

I agree with what @sluongng said but I don't have the workload to work on the improvement.

chris-codaio:
I just tried this experimental setting out on a typescript heavy codebase and it resulted in increasing our fully-remote-cached test run time from 1.5 minutes to near 10 minutes.

moroten:
That sounds really bad. I've mostly been looking at the output from `--profile --noslim_profile` to find out where the bottlenecks have been on a high level. Is `--experimental_remote_merkle_tree_cache_size=1000` too small? Also, I don't have experience of the typescript repositories and rules.

chris-codaio:
FWIW, I tried bumping the cache size up to 4000 but saw no improvement.

moroten:
The [Remote Execution API meeting notes](https://docs.google.com/document/d/1EtQMTn-7sKFMTxIMlb0oDGpvGCMAuzphVcfx58GWuEM/edit#heading=h.a2wi9zy5ah2) 2022-10-11 reads:

> Using remote Merkle tree caching (--experimental_remote_merkle_tree_cache) outcome
Ed: The flag itself didn’t have a big impact, but cherry-picking Bazel [PR#16118](https://github.com/bazelbuild/bazel/pull/16118) did.

It seems like it is not working as intended. I haven't looked into it for a long time so I don't know if something in Bazel have changed in an unfortunate way for this patch.

PR #13884, opened 2021-08-21
Enable llvm_coverage_map_format for darwin by default

keith:
Coverage with clang and the default toolchain on darwin requires this
for coverage, previous you had to pass `--experimental_use_llvm_covmap`
always, but this removes that requirement. If folks have a reason to
this could still be disabled with `--features=-llvm_coverage_map_format`


keith:
I'm not 100% sure about the statement about gcc vs llvm coverage, but I can't seem to get the gcc stuff working right now, not sure if it should be possible? but linking fails because of missing llvm symbols in that case

trybka:
I have no issue with it, but that's up to the bazel folks. I'm not sure what their flag removal policy is these days.

keith:
dropping for now

PR #13888, opened 2021-08-21
LLVM BC support

omarzl:
Using [rules_swift](https://github.com/bazelbuild/rules_swift) it is possible to generate LLVM IR BC files using `swift.emit_bc` feature, some days ago there was a [refactor](https://github.com/bazelbuild/rules_swift/commit/8cf36354655e6baa276fd7a5bae29b00ee5757c3#diff-535d0dbea9614c6905a30dfdb0e708df3b16fc68f8930d6aba519342509dc657R1948) to use `create_compilation_outputs` from `cc_common`, it is giving us the following error since the bc extension is missing, this pr fixes it

```
Error in create_compilation_outputs: 'Somefile.swift.bc' has wrong extension. The list of possible extensions for 'objects' is: .o,.obj,.pic.o,.opb
```


omarzl:
@googlebot I fixed it.

Issue #13221, opened 2021-03-15
incompatible_disallow_resource_jars


comius:
**Flag:** `--incompatible_disallow_resource_jars `
**Available since:** 2018-11-12
**Flipped in:** Bazel 5.0.0
**Will be removed:** after 5.0.0

# Migration

Disables the `resource_jars` attribute on `java_library`; use `java_import` and `deps` or `runtime_deps` instead.

For example:
```starlark
java_import(
  name = "resources",
  jars = ["resource.jar"],
)

java_library(
  name = "lib",
  runtime_deps = [":resources"],
)
```

# Unflipping

Use:
```
--noincompatible_disallow_resource_jars --experimental_builtins_injection_override=-java_library,-java_plugin 
```

The flag is correlated with Starlark `java_library`. In order to flip it back, also implementation of `java_library` and `java_plugin` need to be flipped back to native implementation.

cheister:
Using a java_import for the resource jar and adding the java_import to the deps or runtimes_deps does not include the contents of the resource jar in the jar created by the java_library.  

Is there still a way to include a jar of resources in the java_libary jar without resource_jars?

comius:
> Using a java_import for the resource jar and adding the java_import to the deps or runtimes_deps does not include the contents of the resource jar in the jar created by the java_library.

This is true. However the `jar` file is on either compile or runtime classpath. So it is accessible from the java_library's implementation.

> Is there still a way to include a jar of resources in the java_libary jar without resource_jars?

Is there a use case where the resources need to be in that particular java_library jar?

jdai8:
Removing this from `java_plugin` rules is problematic since they don't have a `runtime_deps` attribute (not sure why).

We can use `deps` instead, but I think it's worth calling out if this is the intended workaround.

eustas:
> Is there a use case where the resources need to be in that particular java_library jar?

Yes. LICENSE file.

comius:
> Removing this from `java_plugin` rules is problematic since they don't have a `runtime_deps` attribute (not sure why).
> 
> We can use `deps` instead. Is this the intended workaround?

Using `deps` is ok - resources will be in this case on both compilation and runtime classpaths.

>> Is there a use case where the resources need to be in that particular java_library jar?

> Yes. LICENSE file.

If you use `java_library`'s `resources` attribute, the LICENSE file will be part of the library's `jar`.

eustas:
Alas, `resources` have quite a weird path resolution rules.

[...]

PR #13891, opened 2021-08-23
Flip incompatible_disallow_resource_jars flag.

comius:
The resource_jars is not supported by java_common.compile function. In
order to rewrite java_library to Starlark the call needs to be extended or the
feature disabled. Because the migration is relatively simple I'm opting
to flip the flag and remove it.

RELNOTES[INC]: Flipped --incompatible_disallow_resource_jars (see https://github.com/bazelbuild/bazel/issues/13221).


comius:
Downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2150

lberki (approved):

PR #13893, opened 2021-08-23
[Cherrypick 4.2.1] Revert 80c59de in 4.2 branch

### LINKED ISSUE #13890: Transitions cannot read Starlark flags as `@repository_name//:flag_name` in 4.2.0
Bazel 4.2.0 introduced a regression in the handling of Starlark flags: In a transition, the value of such a flag cannot be read if the flag is referred to as `@repository_name//:flag_name` instead of `//:flag_name`, which as far as I know is required for that transition to be usable by other workspaces. Reading the flag does not work even if it is set via both `--@repository_name//:flag_name` and `--//:flag-name` on the commandline.

I noticed this while working on [rules_fuzzing](https://github.com/bazelbuild/rules_fuzzing), whose tests no longer pass with Bazel 4.2.0. Usage of `rules_fuzzing` by other workspaces does not seem to be affected.

I [created a reproducer](https://gist.github.com/fmeum/4e2b5d08aa9a7ed8df39e4362f4bfa3b) that prints 

- `regressed`with
  - Bazel 4.2.0
- `not_regressed` with
  - Bazel 4.0.0
  - Bazel 4.1.0
  - Bazel 5.0.0-pre.20210810.4
  - Bazel 4.2.0 with 80c59dea59 reverted

Unfortunately 98d376faeb206f14838156ce4cb305ddbfce08fa cannot be cleanly cherry-picked on top of Bazel 4.2.0 by itself, so reverting 80c59dea59 appears to be the safest way to mitigate this regression.

_Originally posted by @fmeum in https://github.com/bazelbuild/bazel/issues/13558#issuecomment-903488624_

katre:
This fixes #13890 by reverting 80c59dea59d4dce39d4b5d21665c3d7313197358 in the 4.2 series.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/StarlarkOptionsParser.java
@@ -109,8 +109,6 @@ public class StarlarkOptionsParser {
     ImmutableMap.Builder<String, Object> parsedOptions = new ImmutableMap.Builder<>();
     for (Map.Entry<String, Pair<String, Target>> option : unparsedOptions.entrySet()) {
       String loadedFlag = option.getKey();
-      // String loadedFlag =
-      // Label.parseAbsoluteUnchecked(option.getKey()).getDefaultCanonicalForm();
       String unparsedValue = option.getValue().first;
       Target buildSettingTarget = option.getValue().second;
       BuildSetting buildSetting =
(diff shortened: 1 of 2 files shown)


gregestren:
Is the bug saying that if we also had https://github.com/bazelbuild/bazel/commit/98d376faeb206f14838156ce4cb305ddbfce08fa then all would be well? So it's a patch sequencing error?

katre:
Yes, a bisect indicates that 98d376faeb206f14838156ce4cb305ddbfce08fa should fix this, but I couldn't figure out a set of commits to cherrypick that gets it to apply cleanly. Since the original request to include 80c59de doesn't have any justification, I feel like it's reasonable to remove.

gregestren (approved):

Issue #12696, opened 2020-12-14
Bazel branch coverage ignores Jacoco filtering


gergelyfabian:
### Description of the problem / feature request:

Bazel counts branch coverage also for code that would be filtered out by Jacoco.
Function and line coverage (when looking at reports generated by genhtml) seem to be respecting Jacoco filtering.

Jacoco is constantly improving filtering (https://www.jacoco.org/jacoco/trunk/doc/changes.html), but these changes (e.g. filtering out generated bytecode for Java and Kotlin) aren't then taken by Bazel branch coverage.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Check out https://github.com/gergelyfabian/bazel-scala-example/tree/jacoco_coverage_fix.
2. In bazel repo: bazel build src/java_tools/junitrunner/java/com/google/testing/coverage:JacocoCoverage_jarjar_deploy.jar
3. Copy bazel's bazel-bin/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverage_jarjar_deploy.jar to bazel-scala-example's tools/
4. Run tools/coverage.sh
5. You'll see in the coverage report branch coverage for lines that are not covered by line coverage (example-lib/src/main/scala/mypackage/Foo.scala):
```
      23         [ +  + ]:            :       d <- if (foo.a < 10) {
      24                 :            :         List(4)
      25                 :            :       } else {
      26                 :            :         List(14)
      27                 :            :       }
```
6. Branch coverage is 5/30 for example-lib/src/main/scala/mypackage/Foo.scala.

Note: These lines are not covered because of a bug in Jacoco 0.8.3 (that Bazel currently uses), it was fixed in Jacoco 0.8.5.

When in contrast I follow the instructions to upgrade Jacoco in https://github.com/bazelbuild/bazel/issues/11674#issuecomment-739725257 (to Jacoco 0.8.3 with the backported Scala 2.12 lambda fix) and then run the above steps the line coverage changes (the above lines 23-27 are now covered) and the branch coverage stays the same for example-lib/src/main/scala/mypackage/Foo.scala (5/30).

Compare with plain Jacoco:

1. Download http://search.maven.org/remotecontent?filepath=org/jacoco/jacoco/0.8.5/jacoco-0.8.5.zip and http://search.maven.org/remotecontent?filepath=org/jacoco/jacoco/0.8.3/jacoco-0.8.3.zip.
[...]

PR #13896, opened 2021-08-24
Apply Jacoco's instruction filtering to Java branch coverage

c-mita:
Jacoco applies a set of filters during its coverage analysis to remove
various compiler constructs that could lead to surprising outputs in
coverage results, leaving behind coverage data that closer fits the
code given to the compiler.

A simple java example is the function:

```
public int foo(String input) {
  switch (input) {
    case "A":
      return 1;
    case "B":
      return 2;
    case "C":
      return 3;
    default:
      return -1
  }
}
```

This would generate at least double the number of expected branches as
the compiler generates a second set of branches to operate on hashCode.
Jacoco would normally ignore the first set of branches, reducing the
number to the expected 4.

Because Bazel applies its own branch analysis step, Bazel's branch
coverage output does not benefit from this filtering.

This change adapts the custom analyzer to record the necessary
information during the ASM Tree walk in the same manner as Jacoco's
regular analyzer in order to apply the filters.

The filters have three output operations:

 * ignore - ignore a range of instructions
 * merge - merges the coverage information of two instructions
 * replaceBranches - retarget the output branches of an instruction

The first of these, ignore, is by far the simplest to implement and
covers the vast majority of use cases in Java.

By mapping the AbstractInsnNode objects recorded during the ASM Tree
walk to the Instruction objects used during branch analysis, we can
simply skip over Instruction objects that have had their associated
AbstractInsnNode object ignored by a filter.

The remaining operations, merge and replaceBranches, are left
unimplemeted for now; these require more care to handle, are harder
to test, and affect only a minority of cases, specifically:

 * merge is only used to handle duplication of finally blocks
 * replaceBranches is used by Kotlin and Eclipse compiler filters

Part of #12696

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/BranchDetailAnalyzer.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/BranchDetailAnalyzer.java
@@ -114,7 +114,7 @@ public class BranchDetailAnalyzer extends Analyzer {
 
   // Generate the line to probeExp map so that we can evaluate the coverage.
   private Map<Integer, BranchExp> mapProbes(final ClassReader reader) {
-    final ClassProbesMapper mapper = new ClassProbesMapper();
+    final ClassProbesMapper mapper = new ClassProbesMapper(reader.getClassName());
     final ClassProbesAdapter adapter = new ClassProbesAdapter(mapper, false);
     reader.accept(adapter, 0);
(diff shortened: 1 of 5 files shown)


c-mita:
To be clear, I don't think supporting the other filter operations is impossible; the main issue is testing, especially for the `replaceBranches` call.

PR #13914, opened 2021-08-28
Update to Error Prone 2.9.0

cushon:


katre:
This commit needs to be split, we can't apply changes to third_party and other files at once.
Probably we'll want to merge the third_party changes first and the src change right after? If the postsubmit breaks for a few minutes it's survivable.

cushon:
@katre thanks, done. The `//third_party/checker_framework_javacutil` dep is now unused, but I can clean up the references to it and then delete it as separate follow-up changes to avoid needing to make an atomic change that spans `third_party`.

katre (approved):
Thanks, I'll import this shortly.

Issue #12010, opened 2020-08-26
Downloads can be truncated and report an incorrect sha256sum


b0ri5:
### Description of the problem / feature request:

A CircleCI (continuous integration service) build seems to have downloaded a truncated version of the file https://dl.google.com/go/go1.15.linux-amd64.tar.gz and thought the download was complete, computed the sha256sum and reported that it was different than what was expected, which is the published version currently at https://golang.org/dl/.

Expected sha256sum: 2d75848ac606061efe52a8068d0e647b35ce487a15bb52272c427df485193602
Unexpected sha256sum: 22c0e73b372fba0186c83b1b8d6572946b3d50415a5d1eb2699695dff96dd0ec

I asked internally and someone figured out that
$ head -c 3407856 go1.15.linux-amd64.tar.gz | sha256sum
22c0e73b372fba0186c83b1b8d6572946b3d50415a5d1eb2699695dff96dd0ec

produces the unexpected hash. To find the full internal conversation search internally for the unxpected hash.

Their recommendation was to audit the error reporting code that does the downloading since the code seems to have considered the request complete before the Content-Size header was fulfilled.

My expectation is that rather than reporting the wrong hash, that some other error be reported if the file could not be downloaded.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Audit the download code and try to write a unit test for anything that could cause a truncated file to be hashed.

### What operating system are you running Bazel on?

Linux? This occurred on a CircleCI machine.

### What's the output of `bazel info release`?

I'm not sure, probably the latest release as I was using bazelisk.

###  Have you found anything relevant by searching the web?

No, I couldn't find anything else matching the error.

### Any other information, logs, or outputs that you want to share?

Attached the logs.
[bazel-wrong-sha.txt](https://github.com/bazelbuild/bazel/files/5133109/bazel-wrong-sha.txt)

philwo:
Seems like we just hit this on Bazel CI: https://buildkite.com/bazel/bazel-bazel/builds/14172#d7d32a9c-47f3-4d71-87cb-d2ccd872b831

```
[...]

PR #13918, opened 2021-08-30
Correctly report error when HTTP connection was closed before Content-Length was fulfilled.

coeuvre:
Instead of reporting checksum mismatch, the downloader will now report 

```
Connection was closed before Content-Length was fulfilled. Bytes read %s but wanted %s
```

when HTTP connection is closed before Content-Length is fulfilled.

Fixes https://github.com/bazelbuild/bazel/issues/12010.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpStreamTest.java
+++ b/src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpStreamTest.java
@@ -39,7 +39,6 @@ import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.SocketTimeoutException;
 import java.net.URL;
-import java.util.Arrays;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.zip.GZIPOutputStream;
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpStreamTest.java
```diff
@@ -205,6 +206,26 @@ public void bigDataWithInvalidChecksum_throwsIOExceptionAfterCreateOnEof() throw
     }
   }
 
+  @Test
+  public void bigDataTruncated_throwsExpectedError() throws Exception {
+    byte[] bigData = new byte[HttpStream.PRECHECK_BYTES + 70001];
+    randoCalrissian.nextBytes(bigData);
+    byte[] truncated = Arrays.copyOfRange(bigData, 0, bigData.length - 1);
```
whsiao, line 213: A nice to have test coverage can be added.  A test case or test data for `bigData.length + 1` to cover the other half of the not equal check at https://github.com/bazelbuild/bazel/pull/13918/files#diff-e49ed82007084b0b163fc77f9f8a46b5336587b9ea739ad5bfb01a4f7eed8befR57.

philwo (approved):

src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpStreamTest.java
coeuvre: Added. Thanks!

whsiao (approved):
Approval for the new test cases.  Get philwo's approval for the other parts of your changes outside of the test cases, since I'm not familiar with those parts.

PR #13927, opened 2021-08-31
Add warnings as errors cc feature

### LINKED ISSUE #11550: Default C++ toolchains feature for setting Werror

It's quite commonly desirable to be able to set `Werror` for treating warnings as errors. Currently one way of doing is by passing this in `copts`. However, it would be nicer if it was part of the default toolchain and therefore could be enabled with an `error` feature. This would also allow easy disabling it for certain targets, an  example being external libraries which don't build with `Werror`.

An easier way of configuring treating warning as errors.

jheaff1:
This PR adds a `feature` to cc toolchains to treat warnings as errors. The `feature` is disabled by default.

Closes #11550

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -1181,6 +1181,20 @@ def _impl(ctx):
         ],
     )
 
+    treat_warnings_as_errors_feature = feature(
+        name = "treat_warnings_as_errors",
+        flag_sets = [
+            flag_set(
+                actions = [ACTION_NAMES.c_compile, ACTION_NAMES.cpp_compile],
+                flag_groups = [flag_group(flags = ["-Werror"])],
+            ),
+            flag_set(
+                actions = all_link_actions,
+                flag_groups = [flag_group(flags = ["-Wl,-fatal-warnings"])],
+            ),
+        ],
+    )
+
     is_linux = ctx.attr.target_libc != "macosx"
 
     # TODO(#8303): Mac crosstool should also declare every feature.
@@ -1231,6 +1245,7 @@ def _impl(ctx):
             user_compile_flags_feature,
             sysroot_feature,
             unfiltered_compile_flags_feature,
+            treat_warnings_as_errors_feature,
         ] + layering_check_features(ctx.attr.compiler)
     else:
         features = [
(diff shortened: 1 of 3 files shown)


tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -1181,6 +1181,16 @@ def _impl(ctx):
         ],
     )
 
+    treat_warnings_as_errors_feature = feature(
+        name = "treat_warnings_as_errors",
+        flag_sets = [
+            flag_set(
+                actions = [ACTION_NAMES.c_compile, ACTION_NAMES.cpp_compile],
+                flag_groups = [flag_group(flags = ["-Werror"])],
```
keith, line 1189: I wonder if we should add `-Wl,-fatal-warnings` to link invocations here as well?

jheaff1: According to [this](http://web.mit.edu/rhel-doc/3/rhel-as-en-3/w.html), `-fatal-warnings` is an argument for the GNU assembler (`as`). So I should add `-fatal-warnings` for assemble actions?

keith: checkout https://linux.die.net/man/1/ld

jheaff1: Thanks, I have added `-Wl,-fatal-warnings` for link invocations

keith (approved):
we probably want this in the darwin toolchain as well, but that could be done as a follow up if this lands

jheaff1:
> we probably want this in the darwin toolchain as well, but that could be done as a follow up if this lands

Sadly I don’t own a Mac so would not be able to test changes to the Darwin toolchain

jheaff1:
Can this PR be merged? :)

keith:
Someone from Google will have to review

jheaff1:
@philwo Thoughts on this PR?

gregestren:
I'm not sure if @philwo or @comius would have the best input on this. Both are CC'ed.

jheaff1:
@oquenchil Hi, any update on this PR :)

aiuto:
@oquenchil  Friendly ping.

oquenchil (approved):
LGTM

aiuto:
I started a merge. 
@oquenchil  look for a CL in the morning.

PR #13929, opened 2021-08-31
Fix an instance of -Wunused-but-set-variable.

pkasting:
Bug: chromium:1203071


gregestren:
Assigned to @comius for triage. This is related to ijars?

comius (approved):

Issue #2231, opened 2016-12-12
OS X Sandbox should mount ~/Library/Logs and ~/Library/Cache read-write by default


damienmg:
Most of the tutorial failures come from tools trying to write there and getting permission denied. See #2189 for more context.

damienmg:
Assigning to Philip since @hermione521 is busy with coverage.

philwo:
I think we can make these paths read-write without decreasing hermeticity of builds (much). I'll send a patch.

Of course it would be even better if we could mount a private tmpfs there, so that tools can write there, but the written files won't persist, but I don't know of any way how to do this on macOS.

damienmg:
+1 if we can do with tmpfs but meh :/

philwo:
Fix is in code review.

PR #13945, opened 2021-09-05
Fix typo in the cache directory name on macOS

tetsuok:
This fixes the cache directory name on macOS.
The issue #2231 mentioned ~/Library/Cache, but such directory
doesn't exist. The correct cache directory should be ~/Library/Caches.


philwo (approved):
Oops! Thanks for catching that!

PR #13946, opened 2021-09-05
Delete remaining code on src/objc_tools

tetsuok:
The targets, `//src/objc_tools` and the directory were removed in 32246b2b60, but there are files to mention them.
This PR is a follow-up fix of 32246b2b60 to remove the remaining.


philwo (approved):
Excellent, thanks!

PR #13952, opened 2021-09-08
Improve libtool.sh script

keith:
This is just some basic cleanup to pass shellcheck, and error if
toolchain setup didn't happen correctly.

### DIFF
--- a/tools/objc/libtool.sh
+++ b/tools/objc/libtool.sh
@@ -35,7 +35,8 @@ fi
 WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
 
 if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
-    echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  exit 1
 elif "${MY_LOCATION}"/libtool_check_unique "$@"; then
   # If there are no duplicate .o basenames,
   # libtool can be invoked with the original arguments.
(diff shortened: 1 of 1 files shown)


tools/objc/libtool.sh
```diff
@@ -96,7 +97,6 @@ while [[ $# -gt 0 ]]; do
      ARG="$1"
      shift
      ARGS+=("${ARG}")
-     OUTPUTFILE="${ARG}"
```
keith, line 99: unused

tools/objc/libtool.sh
```diff
@@ -120,7 +120,7 @@ while [[ $# -gt 0 ]]; do
      ;;
    # Remaining args are input objects
    *)
-     ARGS+=("$(echo "$(hash_objfile "${ARG}")")")
```
keith, line 123: no need to echo the output from another call

tools/objc/libtool.sh
```diff
@@ -44,7 +45,7 @@ elif "${MY_LOCATION}"/libtool_check_unique "$@"; then
 fi
 
 TEMPDIR="$(mktemp -d "${TMPDIR:-/tmp}/libtool.XXXXXXXX")"
-trap "rm -rf \"${TEMPDIR}\"" EXIT
```
keith, line 47: this way it's evaluated later in case the value changes

tools/objc/libtool.sh
```diff
@@ -35,7 +35,8 @@ fi
 WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
 
 if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
-    echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  exit 1
```
keith, line 39: now that this script has exited for a while I think erroring if it doesn't exist is reasonable

gregestren:
<puts on PR triaging hat> @keith do you know the best people to review PRs like this?

keith:
I'll close this one, I ended up folding it into https://github.com/bazelbuild/bazel/pull/13953

PR #13953, opened 2021-09-08
Fix darwin libtool with params files

keith:
This shim only exists to avoid the warning about linking object files
with the same basenames into a single library. Since we moved to params
files this hasn't worked because that file wasn't opened by
libtool_check_unique.cc, and therefore duplicate basenames were never
discovered.

### DIFF
--- a/tools/objc/libtool.sh
+++ b/tools/objc/libtool.sh
@@ -35,7 +35,8 @@ fi
 WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
 
 if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
-    echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
+  exit 1
 elif "${MY_LOCATION}"/libtool_check_unique "$@"; then
   # If there are no duplicate .o basenames,
   # libtool can be invoked with the original arguments.
@@ -44,7 +45,7 @@ elif "${MY_LOCATION}"/libtool_check_unique "$@"; then
 fi
 
 TEMPDIR="$(mktemp -d "${TMPDIR:-/tmp}/libtool.XXXXXXXX")"
-trap "rm -rf \"${TEMPDIR}\"" EXIT
+trap 'rm -rf "$TEMPDIR"' EXIT
 
 # Creates a symbolic link to the input argument file and returns the symlink
 # file path.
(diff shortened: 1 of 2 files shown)


tools/objc/libtool.sh
```diff
@@ -67,62 +68,77 @@ if [[ ! -x "$python_executable" ]]; then
...
+        ;;
+    esac
+  fi
+}
+
+for arg in "$@"; do
+  parse_option "$arg"
 done
 
+if [[ "$use_params_file" == "1" ]]; then
```
keith, line 139: maybe we should just always do this at this point?

tools/objc/libtool_check_unique.cc
```diff
@@ -32,43 +37,71 @@ string getBasename(const string &path) {
...
+unordered_set<string> parseArgs(std::vector<string> args) {
   unordered_set<string> basenames;
-  const regex libRegex = regex(".*\\.a$");
-  const regex noArgFlags =
-      regex("-static|-s|-a|-c|-L|-T|-D|-no_warning_for_no_symbols");
-  const regex singleArgFlags = regex("-arch_only|-syslibroot|-o");
-  // Set i to 1 to skip executable path
-  for (int i = 1; argv[i] != nullptr; i++) {
-    const string arg = argv[i];
+  for (auto it = args.begin(); it != args.end(); ++it) {
```
michaeleisel, line 55: nit: use `<` here in case a badly set of args would overrun?

tools/objc/libtool_check_unique.cc
```diff
@@ -32,43 +37,71 @@ string getBasename(const string &path) {
...
+        }
+      }
     } else if (regex_match(arg, noArgFlags)) {
     } else if (regex_match(arg, singleArgFlags)) {
-      i++;
+      ++it;
     } else if (arg[0] == '-') {
-      return EXIT_FAILURE;
+      exit(EXIT_FAILURE);
       // Unrecognized flag, let the wrapper deal with it
```
michaeleisel, line 82: unrelated, but thinking more about this case, i wonder what the odds are of flags being added (or have they been already?) in which case this would silently become a performance regression. i wonder if it makes sense to output a line somewhere saying libtool has to fall back to the slow path

michaeleisel: ah but looking at this again, it looks like we're confined to a specific set of args anyway or it errors out

tools/objc/libtool.sh
michaeleisel: that seems reasonable

tools/objc/libtool_check_unique.cc
keith: sounds like this isn't really best practice https://stackoverflow.com/a/6673775/902968

tools/objc/libtool_check_unique.cc
keith: yea so whenever libtool.sh is updated to handle new args, we just have to remember to do it here too

tools/objc/libtool_check_unique.cc
michaeleisel: i guess you have technically have UB either way, but the UB of over-incrementing an iterator seems trivial. anyways i'm fine either way

michaeleisel: that is, UB in the case of a buffer overrun

michaeleisel:
lgtm

tools/objc/libtool_check_unique.cc
```diff
@@ -32,43 +37,71 @@ string getBasename(const string &path) {
...
+
+// Returns 0 if there are no duplicate basenames in the object files (via
+// -filelist, params files, and shell args), 1 otherwise
+int main(int argc, const char *argv[]) {
+  std::vector<string> args;
+  // Set i to 1 to skip executable path
+  for (int i = 1; argv[i] != nullptr; i++) {
+    args.push_back(argv[i]);
+  }
+  parseArgs(args);
```
trybka, line 105: I imagine this would compile with an unused-result warning or something. Dunno if it's worth handling that or not.

tools/objc/libtool_check_unique.cc
trybka: May be worth adding a comment to that effect?

tools/objc/libtool_check_unique.cc
```diff
@@ -32,43 +37,71 @@ string getBasename(const string &path) {
   return path.substr(idx + 1);
 }
 
-// Returns 0 if there are no duplicate basenames in the object files (both via
-// -filelist as well as shell args), 1 otherwise
-int main(int argc, const char *argv[]) {
+std::vector<string> readFile(const string path) {
+  std::vector<string> lines;
```
trybka, line 41: nit: there are as many uses of vector as of unordered_set. for consistency, maybe add a using statement?

tools/objc/libtool_check_unique.cc
keith: AFAICT that only happens if you have ` __attribute__((warn_unused_result))` on the decl

keith:
thanks @trybka, ready for another pass

trybka:
LGTM

PR #13961, opened 2021-09-09
Fixes #13887: Add downloaded_file_path param to http_jar

### LINKED ISSUE #13887: http_jar rule doesn't allow specifying file name when downloading

> http_jar rule doesn't have a parameter to specify downloaded file name. This is ok in most cases but if a http_jar is used in conjunction with docker rules this becomes a problem because multiple files are called `downloaded.jar`. In this case all but one file gets replaced. 
A simple fix would be to include a parameter to specify the name of the file. I believe keeping the name as downloaded.jar would ensure backward compatibility as well.

rabisg:
If multiple http_jar rules are used the downloaded jar is always named `downloaded.jar`. This is a problem when the jar
file is being copied, for example in the case of docker rules, since one jar overrides the other which is not the desired
behaviour in most cases.

### DIFF
--- a/src/test/shell/bazel/external_integration_test.sh
+++ b/src/test/shell/bazel/external_integration_test.sh
@@ -311,7 +311,7 @@ function test_jar_download() {
   cat >> $(create_workspace_with_default_repos WORKSPACE) <<EOF
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_jar")
 http_jar(name = 'endangered', url = 'http://127.0.0.1:$nc_port/lib.jar',
-         sha256='$sha256', file_name="foo.jar")
+         sha256='$sha256', downloaded_file_name="foo.jar")
 EOF
 
   mkdir -p zoo
--- a/src/test/shell/bazel/external_integration_test.sh
+++ b/src/test/shell/bazel/external_integration_test.sh
@@ -338,7 +338,7 @@ EOF
   kill_nc
   expect_log "Tra-la!"
   output_base=$(bazel info output_base)
-  jar_dir=$output_base/external/endangered/file
+  jar_dir=$output_base/external/endangered/jar
   [[ -f ${jar_dir}/foo.jar ]] || fail "${jar_dir}/foo.jar not found"
 }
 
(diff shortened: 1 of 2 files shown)


rabisg:
Not sure if there are any corresponding tests to write. Please point me towards it if there are.

meteorcloudy:
Please add a test here:
https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/external_integration_test.sh;l=379?q=http_file&ss=bazel&start=11

tools/build_defs/repo/http.bzl
```diff
@@ -502,6 +503,10 @@ unless it was added to the cache by a request with the same canonical id.
     "auth_patterns": attr.string_dict(
         doc = _AUTH_PATTERN_DOC,
     ),
+    "downloaded_file_path": attr.string(
```
meteorcloudy, line 506: Can it just be a file name instead of a path? Or at least, we should make sure this can only be a relative path.

philwo: Just a file name is probably best.

meteorcloudy (approved):
Thanks you!

src/test/shell/bazel/external_integration_test.sh
```diff
@@ -311,7 +311,7 @@ function test_jar_download() {
   cat >> $(create_workspace_with_default_repos WORKSPACE) <<EOF
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_jar")
 http_jar(name = 'endangered', url = 'http://127.0.0.1:$nc_port/lib.jar',
-         sha256='$sha256')
+         sha256='$sha256', file_name="foo.jar")
```
meteorcloudy, line 314: `downloaded_file_name` ?

meteorcloudy (changes requested):

meteorcloudy:
Please fix the test failures

src/test/shell/bazel/external_integration_test.sh
rabisg: Ya sorry.. unintended. Fixed

rabisg:
> Please fix the test failures

Done

meteorcloudy (approved):

rabisg:
@meteorcloudy Please feel free to merge this PR. I dont have permissions to do it

meteorcloudy:
Yes, I've imported this, waiting for final LGTM and submit.

PR #13971, opened 2021-09-10
Trim input build settings after transition

### LINKED ISSUE #12888: Build failure caused by no-op starlark transition (friction between Starlark None and java null)

Our [fuzzing rules library](https://github.com/bazelbuild/rules_fuzzing) implements a Starlark [transition](https://github.com/bazelbuild/rules_fuzzing/blob/744cfe3ea683f5fa9f0e327663acfcac437e4ee8/fuzzing/private/binary.bzl#L79) to provide compiler instrumentation options when building fuzz test executables. In order to reduce build size output, we would like to provide a "no-op" mode in which the transition is disabled through a bool [config flag](https://github.com/bazelbuild/rules_fuzzing/blob/744cfe3ea683f5fa9f0e327663acfcac437e4ee8/fuzzing/BUILD#L30) and the parent configuration is passed through unchanged.

However, when we attempt a build in this "no-op" transition configuration, we're hitting build failures of the form:

```
ERROR: file 'external/re2/_objs/re2/rune.pic.o' is generated by these conflicting actions:
Label: @re2//:re2
RuleClass: cc_library rule
Configuration: 3c05312acc4ad4afcdf2f011f312660ad214f21abaa06671cad80d482163789b, 2e241eb8731382b18fcb12c9a4b7eb174abce65a2e229b9754f25836277ec281
Mnemonic: CppCompile
Action key: e064b11b5c25436d8c57c05d4e04baa44c8de4fdc572c688519590053b2a3df1
Progress message: Compiling util/rune.cc
PrimaryInput: File:[/usr/local/google/home/sbucur/.cache/bazel/_bazel_sbucur/2527f57e21cfbf2e414b894f7610e705/external/re2[source]]util/rune.cc
PrimaryOutput: File:[[<execution_root>]bazel-out/k8-fastbuild/bin]external/re2/_objs/re2/rune.pic.o
Owner information: ConfiguredTargetKey{label=@re2//:re2, config=BuildConfigurationValue.Key[3c05312acc4ad4afcdf2f011f312660ad214f21abaa06671cad80d482163789b]}, ConfiguredTargetKey{label=@re2//:re2, config=BuildConfigurationValue.Key[2e241eb8731382b18fcb12c9a4b7eb174abce65a2e229b9754f25836277ec281]}
MandatoryInputs: are equal
Outputs: are equal
ERROR: com.google.devtools.build.lib.skyframe.ArtifactConflictFinder$ConflictException: com.google.devtools.build.lib.actions.MutableActionGraph$ActionConflictException: for external/re2/_objs/re2/rune.pic.o, previous action: action 'Compiling util/rune.cc', attempted action: action 'Compiling util/rune.cc'
```

Running `bazel config` on the two configuration yields the following:

```
$ bazel config 3c05312acc4ad4afcdf2f011f312660ad214f21abaa06671cad80d482163789b 2e241eb8731382b18fcb12c9a4b7eb174abce65a2e229b9754f25836277ec281
INFO: Displaying diff between configs 3c05312acc4ad4afcdf2f011f312660ad214f21abaa06671cad80d482163789b and 2e241eb8731382b18fcb12c9a4b7eb174abce65a2e229b9754f25836277ec281
Displaying diff between configs 3c05312acc4ad4afcdf2f011f312660ad214f21abaa06671cad80d482163789b and 2e241eb8731382b18fcb12c9a4b7eb174abce65a2e229b9754f25836277ec281
FragmentOptions user-defined {
  @rules_fuzzing//fuzzing:cc_engine_instrumentation: none, null
  @rules_fuzzing//fuzzing:cc_engine_sanitizer: none, null
  @rules_fuzzing//fuzzing:cc_fuzzing_build_mode: false, null
}
```

The only differences in config seem to be related to input configs in the transition that we only read and not list in the output.

The transition is defined as follows:

```
fuzzing_binary_transition = transition(
    implementation = _fuzzing_binary_transition_impl,
    inputs = [
        "@rules_fuzzing//fuzzing:cc_engine_instrumentation",
        "@rules_fuzzing//fuzzing:cc_engine_sanitizer",
        "@rules_fuzzing//fuzzing:cc_fuzzing_build_mode",
        "//command_line_option:copt",
        "//command_line_option:conlyopt",
        "//command_line_option:cxxopt",
        "//command_line_option:linkopt",
    ],
    outputs = [
        "//command_line_option:copt",
        "//command_line_option:conlyopt",
        "//command_line_option:cxxopt",
        "//command_line_option:linkopt",
    ],
)
```

```
$ git clone git@github.com:bazelbuild/rules_fuzzing.git -b no-op-transitions
$ cd rules_fuzzing
$ bazel test //examples/...
```

Debian-based Linux.

`release 4.0.0`

This issue seems to be loosely related to conversations in #11196.

fmeum:
By convention, BuildOptions should not contain Starlark build settings
that are set to their default values. However, if a Starlark transition
has a build setting as an input, the value of that setting has to be
added to the input BuildOptions temporarily so that it can be read by
the transition.

Before this commit, if a build setting was an input but not an output of
a transition, its default value remained in the BuildOptions and could
cause action conflicts. With this commit, all build settings rather than
only output build settings are trimmed post-transition if they are set
to their default value.

Fixes #12888.


src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -445,6 +446,48 @@ public static BuildOptions addDefaultStarlarkOptions(
     return optionsWithDefaults == null ? fromOptions : optionsWithDefaults.build();
   }
 
+  /*
+   * Removes all build settings explicitly set to their default values from the output build options
```
gregestren, line 450: Nit: what precisely does "explicitly" mean here?

i.e. is this equivalent to "removes all build settings set to their default values"? Or is it possible for "implicit" defaults to get in there somehow?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -445,6 +446,48 @@ public static BuildOptions addDefaultStarlarkOptions(
...
+   * of a transition. This cancels the effect of addDefaultStarlarkOptions post-transition if called
```
gregestren, line 451: Can you elaborate for me this condition? Is there any risk of not calling it on the same buildSettingPackages? Or is this just asserting expected code invariants?

gregestren (commented):
Just for posterity, do you see any conceptual risks or blemishes to this change? Or does it look fully "elegant"?

I suspect the latter but want to sanity check with you.

gregestren:
@sdtwigg relevant to your transition naming refactoring plans.

sdtwigg:
This looks good and is a reasonable idea conceptually.

@gregestren  can you verify that this 'sits in the right place' wrt where you added Starlark transition caching?

fmeum:
> Just for posterity, do you see any conceptual risks or blemishes to this change? Or does it look fully "elegant"?
> 
> I suspect the latter but want to sanity check with you.

It didn't feel that fully elegant and that's because it used the wrong approach - I missed that there already was logic that was trimming default values, just not all of them. In hindsight that is obvious: If it weren't the case, all no-op transitions would have caused action conflicts (rather than only those with input-only build settings).

I updated the PR and would now call it fully "elegant".

gregestren:
> @gregestren can you verify that this 'sits in the right place' wrt where you added Starlark transition caching?

Yes. This code only triggers when the transition is actually being executed, and its output is prepared. Caching should purely wrap around that.

gregestren:
> It didn't feel that fully elegant and that's because it used the wrong approach - I missed that there already was logic that was trimming default values, just not all of them. In hindsight that is obvious: If it weren't the case, all no-op transitions would have caused action conflicts (rather than only those with input-only build settings).
> 
> I updated the PR and would now call it fully "elegant".

Very nice catch! And indeed much simpler this round. Thanks for that update. And solid test coverage.

gregestren (approved):

gregestren (commented):
Re: https://github.com/bazelbuild/bazel/issues/12888#issuecomment-927166979, can you confirm the `return {}` code path also get this optimization? Maybe a test case for that too?

fmeum:
> Re: [#12888 (comment)](https://github.com/bazelbuild/bazel/issues/12888#issuecomment-927166979), can you confirm the `return {}` code path also get this optimization? Maybe a test case for that too?

I added a test case verifying that input build settings are trimmed even for `return {}` transitions. Is that what you meant?

sdtwigg:
LGTM on my end.

PS: Really like how elegant it is now :D

gregestren:
> > Re: [#12888 (comment)](https://github.com/bazelbuild/bazel/issues/12888#issuecomment-927166979), can you confirm the `return {}` code path also get this optimization? Maybe a test case for that too?
> 
> I added a test case verifying that input build settings are trimmed even for `return {}` transitions. Is that what you meant?

Yes. Thank you.

PR #13979, opened 2021-09-11
Updated a deprecated link

Sayanta66:
Fixed the link that was showing error

### DIFF
--- a/site/docs/android-build-performance.md
+++ b/site/docs/android-build-performance.md
@@ -12,7 +12,7 @@ apps specifically. For general build performance optimization with Bazel, see
 ## Recommended flags
 
 The flags are in the
-[`bazelrc` configuration syntax](guide.html#bazelrc-syntax-and-semantics), so
+[`bazelrc` configuration syntax](/guide.md#bazelrc-syntax-and-semantics), so
 they can be pasted directly into a `bazelrc` file and invoked with
 `--config=<configuration_name>` on the command line.
 


Sayanta66:
@googlebot I signed it!

site/docs/android-build-performance.md
```diff
@@ -12,7 +12,7 @@ apps specifically. For general build performance optimization with Bazel, see
 ## Recommended flags
 
 The flags are in the
-[`bazelrc` configuration syntax](guide.html#bazelrc-syntax-and-semantics), so
+[`bazelrc` configuration syntax](/guide.md#bazelrc-syntax-and-semantics), so
```
aiuto, line 15: The link should be .html That is what they appear as.

Sayanta66: But in the github when clicking on that link it is showing error 404

Sayanta66: Do you think it's better to put this link directly https://docs.bazel.build/versions/main/guide.html#bazelrc-syntax-and-semantics

gregestren: This is about a discrepancy between what links look like on `docs.bazel.build` (rendered through Jekyll)  vs. `GitHub`, right? 

I don't think putting in the full link will work since that'll legitimately break the versioning that should be user-controllable. I actually don't know the answer here, but I can run this idea by some tech writers who might have some ideas.

floriographygoth: Responding here: unfortunately, the rendered content needs to have the file type set to `.html` to render correctly, so we can't accept this pull request. However, we're looking at possible fixes for the future. :) Thanks for bringing this up & submitting a pull request, though.

aiuto:
Closing, as per discussion. The link with .html is the correct one.

PR #13982, opened 2021-09-13
Skylark create compilation outputs based on another object

torgil:
This patch adds a "compilation_outputs" argument to cc_common.create_compilation_outputs. All files in this object will be merged into the new object. This enables starlark code to modify object files from compilation phase while keeping all other files (coverage, debug, ...) intact.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCcModule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCcModule.java
@@ -227,10 +227,15 @@ public class BazelCcModule extends CcModule
       Object objectsObject,
       Object picObjectsObject,
       Object ltoCopmilationContextObject,
+      Object compilationOutputs,
       StarlarkThread thread)
       throws EvalException {
     return super.createCompilationOutputsFromStarlark(
-        objectsObject, picObjectsObject, ltoCopmilationContextObject, thread);
+        objectsObject,
+        picObjectsObject,
+        ltoCopmilationContextObject,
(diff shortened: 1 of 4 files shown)


oquenchil:
Can you use merge_compilation_outputs instead?

torgil:
> Can you use merge_compilation_outputs instead?

merge_compilation_outputs will merge the compilation outputs, keeping all objects. The patch is intended to replace the object files, which it didn't. Fixed.

My use-case is to mock calls to system functions which the test also uses for other purposes (say malloc). The test in this case needs to define the mock under a different name and the code under test needs to call into this function instead. Here I use objcopy to redefine symbol names on the tested code.

oquenchil:
Not sure I understand, but if you wanted to write a function in Starlark to create a compilation_outputs object the way you want to create it, is there anything you would be missing from the API?

I'd prefer to make getters available for what's currently there rather than add this to the constructor for a single use case.

torgil:
> I'd prefer to make getters available for what's currently there rather than add this to the constructor for a single use case.

Agree, this would be more flexible. It'd need a test that assures that all members (java-side) are copied (currently lto and dwo files are missing at starlark side), i'm not sure how to that. Is it something like this had in mind?

```
def clone_compilation_info(compilation_outputs, **kwargs):
    args = {key: getattr(compilation_outputs, key) for key in dir(compilation_outputs)}
    args.update(kwargs)
    return cc_common.create_compilation_context(**args)
```

oquenchil:
We should solve this via methods in Starlark, not with more API implemented natively in Java.

Issue #10642, opened 2020-01-23
_lcov_merger late-bound test attribute


keith:
I went to add `_lcov_merger` as an attribute for iOS tests here https://github.com/bazelbuild/rules_apple/pull/691 and there was some discussion about making this tool a late-bound attribute so we wouldn't incur overhead for this tool always. It doesn't seem like there's an API that currently allows us to do this.

There was also some discussion of if the attribute makes sense to be on every test rule or if it should be more global, maybe with a flag like `--lcov_merger` with this tool as the default.

I wanted to file this here so we could discuss this and unblock iOS tests supporting coverage.

ulfjack:
This looks like a dupe of #8736.

Issue #8736, opened 2019-06-27
Expose CoverageOutputGenerator's label in a configuration


iirina:
In the native rules we have the late-bound attribute `:lcov_merger` which looks at the build configuration to see if coverage is enabled and returns the label of `CoverageOutputGenerator` or null (see [BaseRuleClasses#115](https://source.bazel.build/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java;l=115?q=BaseRuleClasses&ss=bazel)).

We need the Starlark rules to be able to define `_lcov_merger` as a late bound attribute in the same way. For this to happen, the `CoverageOutputGenerator`s label has to be exposed on a class that extends BuildConfiguration.Fragment.

This is helpful for example for `rules_go` (see #8670), in order for `go_test` to be compatible with bazel's native test rules (e.g. `java_test`, `cc_test`).

fmeum:
@c-mita I think that I generally know how to resolve this issue, but would need to know which `Fragment` I should add expose the label on. Could you provide guidance here as per https://github.com/bazelbuild/bazel/pull/13983#issuecomment-981646173 or point me to someone who could?

PR #13983, opened 2021-09-13
Provide the default lcov_merger to all Starlark tests

### LINKED ISSUE #13978: Custom test rule without `_lcov_merger` attribute always fails under `bazel coverage`

When a custom Starlark test rule without explicit coverage support in the form of the `_lcov_merger` attribute is invoked via `bazel coverage` with Bazel after 5b216b27435aeb9eb9c3bd3c552d6498e1050cc7, it fails with
```
FAIL: //:custom_test (see /home/fmeum/.cache/bazel/_bazel_fhenneke/767fd922858d516eb34a4dddb64ca1a0/execroot/__main__/bazel-out/k8-fastbuild/testlogs/custom_test/test.log)
ERROR: /home/fmeum/git/bazel-custom-test-coverage/BUILD:3:12: output 'custom_test/coverage.dat' was not created
ERROR: /home/fmeum/git/bazel-custom-test-coverage/BUILD:3:12: Testing //:custom_test failed: not all outputs were created or valid
Target //:custom_test up-to-date:
  bazel-bin/custom_test
INFO: Elapsed time: 5.635s, Critical Path: 0.17s
INFO: 7 processes: 1 disk cache hit, 5 internal, 1 linux-sandbox.
//:custom_test                                                           FAILED in 0.0s
  /home/fmeum/.cache/bazel/_bazel_fhenneke/767fd922858d516eb34a4dddb64ca1a0/execroot/__main__/bazel-out/k8-fastbuild/testlogs/custom_test/test.log
```

The log says:
```
Executing tests from //:custom_test
--
Coverage runner: cannot locate file
```

With Bazel 4.2.1, the build succeeds:
```
INFO: Build completed successfully, 5 total actions
//:custom_test                                                           PASSED in 0.0s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command
INFO: Build completed successfully, 5 total actions
```

```sh
#!/usr/bin/env sh
touch WORKSPACE

cat << EOF > BUILD
load("//:rules.bzl", "custom_test")

custom_test(
    name = "custom_test",
)
EOF

cat << EOF > rules.bzl
def _custom_test_impl(ctx):
    executable = ctx.actions.declare_file(ctx.attr.name)
    ctx.actions.write(executable, "", is_executable = True)
    return [
        DefaultInfo(
            executable = executable,
        )
    ]

custom_test = rule(
    implementation = _custom_test_impl,
    test = True,
)
EOF

bazel coverage //:custom_test
```

Linux

release 5.0.0-pre.20210831.2 

I bisected the origin of this issue back to 5b216b27435aeb9eb9c3bd3c552d6498e1050cc7.

With `5b216b27435aeb9eb9c3bd3c552d6498e1050cc7`, the `custom_test` rule in the reproducer now provides an `InstrumentedFilesInfo`, which [causes TestActionBuilder to collect coverage](https://github.com/bazelbuild/bazel/blob/ded65d28832ea419e0b8056aa88f9b135a25dabb/src/main/java/com/google/devtools/build/lib/analysis/test/TestActionBuilder.java#L201). However, since the rule does not have the [`:lcov_merger`](https://github.com/bazelbuild/bazel/blob/ded65d28832ea419e0b8056aa88f9b135a25dabb/src/main/java/com/google/devtools/build/lib/analysis/test/TestActionBuilder.java#L268) attribute set, the `LCOV_MERGER` env variable isn't set when [`collect_coverage.sh`](https://github.com/bazelbuild/bazel/blob/ded65d28832ea419e0b8056aa88f9b135a25dabb/tools/test/collect_coverage.sh#L93) is invoked, which causes the error message reproduced above.

fmeum:
Since 5b216b27, all Starlark-defined test rules either need to set the `_lcov_merger` attribute or fail if run with `bazel coverage`. This is fixed by attaching the default lcov merger as a late-bound default to all Starlark rules via a new attribute and falling back to that attribute in the TestActionBuilder.

This commit ensures backwards compatibility is maintained with existing Starlark test rules that set the `_lcov_merger` attribute. This is also verified in the included integration test.

Fixes #13978.


fmeum:
@oquenchil Would you be able to take a look? The issue this resolves is labeled with team-Rules-CPP.

meteorcloudy:
@c-mita Can you help review this change?

c-mita:
Unfortunately, I can't accept this PR as is.
Internally, things are setup a little differently for coverage collection. We use a different `collect_coverage.sh` script and generally do not set the `:lcov_attribute`; this change does not play well with the internal setup.

I think the best (only?) way to handle this at the moment is by changing Bazel's `collect_coverage.sh` to simply do nothing if  the `LCOV_MERGER` variable has not been set.

fmeum:
@c-mita I see the problem. While your proposed solution does resolve the linked issue, it would mean that there would still be no way for Starlark tests to access the merge as a late-bound attribute.

How about the following path forward:
1. You make the required change to `collect_coverage.sh` to resolve the release blocking #13978 and I close the current PR.
2. Someone who is familiar with Google's internal coverage setup provides some guidance on either #8736 or #10642 on how a solution to these issues could look like that would have a chance to be accepted.
3. I will look into providing such a solution as a separate PR.

c-mita:
I'm ok with this approach; I've got a change ready to unblock the build.
It isn't ideal, but it's a way forward.

Issue #13766, opened 2021-07-28
Sandbox doesn't work on macOS Big Sur when the sandbox directory is in a firm linked directory


ob:
macOS Catalina (10.*) introduced a security feature where the [root filesystem is mounted read-only](https://support.apple.com/en-ca/HT210650), and a mechanism for creating directories or symlinks at boot time called [firmlinks](http://www.swiftforensics.com/2019/10/macos-1015-volumes-firmlink-magic.html). Since the root filesystem is read-only, a new volume mounted at `/System/Volumes/Data` is overlaid on top of the root filesystem and contains the mutable portions of the filesystem (more info [here](https://eclecticlight.co/2021/01/13/big-sur-boot-volume-layout/)). 

When building with Bazel on a location that is firm linked from the root, and moving the sandbox to that same location by using the `--experimental_sandbox_base` flag:

In macOS Catalina (10.15.7 to be precise), it works as expected.

```
$ cat /etc/synthetic.conf
export	/System/Volumes/Data/export
$ ls -l /export
lrwxr-xr-x  1 root  wheel  27 May 30 22:42 /export -> /System/Volumes/Data/export
$ cd /export/home/tester
$ git clone https://github.com/bazelbuild/bazel.git
Cloning into 'bazel'...
remote: Enumerating objects: 567875, done.
remote: Counting objects: 100% (1761/1761), done.
remote: Compressing objects: 100% (924/924), done.
remote: Total 567875 (delta 783), reused 1374 (delta 586), pack-reused 566114
Receiving objects: 100% (567875/567875), 757.61 MiB | 27.05 MiB/s, done.
Resolving deltas: 100% (354168/354168), done.
Updating files: 100% (9123/9123), done.
$ cd bazel
$ bazel build --experimental_sandbox_base=$(pwd)/sandbox-dir //src:bazel
INFO: Elapsed time: 92.474s, Critical Path: 29.31s
INFO: 2915 processes: 2034 remote cache hit, 167 internal, 533 darwin-sandbox, 1 local, 180 worker.
INFO: Build completed successfully, 2915 total actions
```

However, that exact same sequence of operations on macOS Big Sur (11.4), produces the following error:

```
$ bazel build --experimental_sandbox_base=$(pwd)/sandbox-dir //src:bazel
<output elided>
[...]

PR #13984, opened 2021-09-13
Don't resolve symlinks for --sandbox_base

ob:
On macOS BigSur, the sandbox-exec command behaves slightly different than on
Catalina when firm links are present.

Resolving symlinks can prevent the sandbox for allowing write operations to the
sandbox base.

This effectively reverts a piece of 656a0ba, namely:

>  When using --experimental_sandbox_base, ensure that symlinks in the path are
>  resolved. Before this, you had to check whether on your system /dev/shm is a
>  symlink to /run/shm and then use that instead. Now it no longer matters, as
>  symlinks are resolved.

See https://github.com/bazelbuild/bazel/issues/13766 for full details.


ob:
CC @philwo since 656a0ba was your commit.

meisterT:
cc @larsrc-google

brentleyjones:
@Wyverald I think this should go in a 5.x release.

Wyverald:
@bazel-io fork 5.1

PR #13999, opened 2021-09-16
Fix remote-exec with disk cache.

ron-stripe:
Fixed uploadBlob/File in DiskAndRemoteCacheClient to always upload to remote when doing remote exec.

Added a new method to check for should uploadContent vs uploadActionResult.

I think I didn't catch this before because I had already cached my inputs in my remote build system.
Now that I tried again with the release, I was seeing problems with our remote build system not having access to the inputs.

This fixes it.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
@@ -15,6 +15,7 @@ package com.google.devtools.build.lib.remote.disk;
 
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static com.google.devtools.build.lib.remote.util.Utils.shouldAcceptCachedResultFromRemoteCache;
+import static com.google.devtools.build.lib.remote.util.Utils.shouldUploadContentToRemoteCache;
 import static com.google.devtools.build.lib.remote.util.Utils.shouldUploadLocalResultsToRemoteCache;
 
 import build.bazel.remote.execution.v2.ActionResult;
@@ -75,7 +76,7 @@ public final class DiskAndRemoteCacheClient implements RemoteCacheClient {
   public ListenableFuture<Void> uploadFile(
       RemoteActionExecutionContext context, Digest digest, Path file) {
     ListenableFuture<Void> future = diskCache.uploadFile(context, digest, file);
-    if (shouldUploadLocalResultsToRemoteCache(options, context.getSpawn())) {
+    if (shouldUploadContentToRemoteCache(options, context.getSpawn())) {
       future =
           Futures.transformAsync(
               future, v -> remoteCache.uploadFile(context, digest, file), directExecutor());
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
```diff
@@ -75,7 +76,7 @@ public void close() {
   public ListenableFuture<Void> uploadFile(
       RemoteActionExecutionContext context, Digest digest, Path file) {
     ListenableFuture<Void> future = diskCache.uploadFile(context, digest, file);
-    if (shouldUploadLocalResultsToRemoteCache(options, context.getSpawn())) {
+    if (shouldUploadContentToRemoteCache(options, context.getSpawn())) {
```
coeuvre, line 79: Can we changed to `options.isRemoteExecutionEnabled() || shouldUploadLocalResultsToRemoteCache()`?
  1. The "content" in `shouldUploadContent...` is not clear.
  2. To reflect that we check `options.isRemoteExecutionEnabled()` inside `findMissingDigests()` as well.

src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
```diff
@@ -87,7 +88,7 @@ public void close() {
   public ListenableFuture<Void> uploadBlob(
       RemoteActionExecutionContext context, Digest digest, ByteString data) {
     ListenableFuture<Void> future = diskCache.uploadBlob(context, digest, data);
-    if (shouldUploadLocalResultsToRemoteCache(options, context.getSpawn())) {
+    if (shouldUploadContentToRemoteCache(options, context.getSpawn())) {
```
coeuvre, line 91: Ditto.

coeuvre (changes requested):
Thanks for the fix!

Do we have similar issues for downloads e.g. outputs of actions that are tagged with `no-cache` should be downloaded after executed remotely?

Can you please add tests to cover uploads/downloads with disk cache and remote-exec?

src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
ron-stripe: I did as you suggested, but I do think shouldUploadLocalResultsToRemoteCache isn't named very well as it is being used to check inputs to remote actions not "Results".

ron-stripe:
@coeuvre rfal.  If you have suggestions on other tests, let me know. The matrix is getting complicated.
I tried adding with a no-remote-exec tag, but it just complained that it couldn't run with the remote genrule spawn strategy.
I'm not sure if getting it run with a different strategy would exercise the right code path.

coeuvre (approved):
LGTM. Thanks!

src/main/java/com/google/devtools/build/lib/remote/disk/DiskAndRemoteCacheClient.java
coeuvre: `shouldUploadLocalResultsToRemoteCache` was intended for checking local results as the code only executed when upload local results before this PR.

Now there are two cases to upload files:
    1. Upload as inputs for remote actions.
    2. Upload outputs of local actions.

IMO, the updated code is more clear for the two purposes. Thanks for the update!

ron-stripe:
Thanks, I still think it would be good to update the code to populate the disk cache on the first run of the remote exec action.
Right now, the disk cache will only get populated when we pull it on the next run from the remote-cache.
However, this code is functional and we discussed this in the previous PR.  So please merge when you get a chance.

ron-stripe:
Hi, I think this still waiting to be merged.
Thanks.

coeuvre:
Oops, sorry that I imported it immediately after LGTM but forgot to send it out for reviewing. Should be merged  in a few days.

PR #14002, opened 2021-09-16
docs: document that digest can be empty in some cases

thesayyn:
WorkRequest proto does not specify whether the digest is present all the time. Since digest can be null in some cases, it should be documented here to clear any confusion.

See: https://github.com/bazelbuild/bazel/blob/97643ad92d86179d21b2526e9f9804045d025f21/src/main/java/com/google/devtools/build/lib/worker/WorkerSpawnRunner.java#L238

### DIFF
--- a/src/main/protobuf/worker_protocol.proto
+++ b/src/main/protobuf/worker_protocol.proto
@@ -27,7 +27,8 @@ message Input {
   string path = 1;
 
   // A hash-value of the contents. The format of the contents is unspecified and
-  // the digest should be treated as an opaque token.
+  // the digest should be treated as an opaque token. This can be empty in some cases
+  // should not assumed to be present all the time.
   bytes digest = 2;
 }
 


thesayyn:
Also seen on https://github.com/bazelbuild/rules_nodejs/issues/2953

thesayyn:
Even though we know that it might be empty in some cases, am not certain about which cases are there. Would like to be pointed in the right direction on this.

larsrc-google:
The `getMetaData` call we get the digest from says it may be null due to "artifact non-existence, lookup errors, or any other reason" but also that "If #mayGetGeneratingActionsFromSkyframe is true and the generating action is not immediately available, this method returns null to signify that a skyframe restart is necessary to obtain the metadata for the requested Artifact.DerivedArtifact." I don't know if the latter can happen at this point, though.

thesayyn:
> The `getMetaData` call we get the digest from says it may be null due to "artifact non-existence, lookup errors, or any other reason" but also that "If #mayGetGeneratingActionsFromSkyframe is true and the generating action is not immediately available, this method returns null to signify that a skyframe restart is necessary to obtain the metadata for the requested Artifact.DerivedArtifact." I don't know if the latter can happen at this point, though.

Should I clarify the description a bit more to give more details from above?

larsrc-google:
Yes, please, that would be useful.

PR #14003, opened 2021-09-16
Document --watchfs limitation when NFS is used

erenon:
If the source is on NFS, and edited from host A, then built on a different host B, that runs bazel with --watchfs, bazel might not notice the change made on the host A.


meisterT:
Let's rather update the non-startup option docs here: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/skyframe/LocalDiffAwareness.java;l=48;drc=2e7030ae75c64e607d28e516f141b3b9edfc2a83

I want to remove the startup option but don't know when I will get to it.

erenon:
Thanks! If you can, please also keep the extra doc on the deprecated startup option - otherwise it'll look like that option doesn't have this limitation.

PR #14010, opened 2021-09-17
Add {objc,objcpp}_compile actions to action sets

jparise:
The Objective-C compile actions weren't included in the
'all_compile_actions', 'preprocessor_compile_actions', and
'codegen_compile_actions' sets.

Adding them makes these action sets a bit more useful as
demonstrated by the fact that we can now use them to set
some default compiler flags below.


jparise:
Alternatively, we could remove these default sets entirely because none of them are used here other than `all_link_actions`.

oquenchil:
@googlewalt can you take a look?

googlewalt (approved):

brentleyjones:
Can/should this be cherry-picked into 6.1?

googlewalt:
It's an NFC, so probably not worth it?

Issue #13819, opened 2021-08-09
Shared libraries with transitions aren't always available where expected in runtime


quval:
### Description of the problem:

When an executable has dynamically-linked dependencies that have transitions, shared libraries aren't always placed where the binary expects them in runtime. This means, for example, that such tests cannot be run in remote execution.

The rpath where shared libraries with transitions are looked up is `$ORIGIN/_solib_k8/../../../k8-fastbuild-ST-<...>/bin/_solib_k8`, but this path is not always available:

1. Locally, `bazel-out/k8-fastbuild/bin/_solib_k8/` may not always exist (e.g. if all shared libraries have transitions). ld doesn't seem to follow `../` from nonexistent paths.
2. In remote execution (or when deploying a target by copying its runfiles directory), the configuration directory `k8-fastbuild-ST-<...>` isn't available at all. The right files exist under `_solib_k8`, but because of the rpath, ld doesn't look for them there.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run `bazel test :b` on this: https://gist.github.com/quval/5077b2edc902e9dc0bcad2311a245ffa
(Or change `cc_test` to `cc_binary` with `linkstatic = 0` and use `bazel run`).

### What operating system are you running Bazel on?

Linux

### What's the output of `bazel info release`?

4.1.0 (also happens at master)

###  Have you found anything relevant by searching the web?

Issue #9357 seems to be related (though Bazel doesn't crash here).

### Any other information, logs, or outputs that you want to share?

After a90c4084315a490c4fecc864d0b3435cc7b2bd64 the same library may not be linked more than once, so a solution may be to always add the default `solib` directory to the rpath. (When running locally, the wrong version may then be loaded for libraries that have been built for more than one configuration – this is already a potential issue; the default `solib` may have to go last if added as a fallback – or the configuration could be encoded in the mangled name of the library to prevent this altogether.) I'll be happy to send a PR if any of this makes sense.

quval:
Here's a draft: https://github.com/quval/bazel/commit/2e63f5af05559e2d967d816beb049fc6ea81fd70. Doesn't create the default `solib` directory, but does the rest. Should I send this as a PR?

quval:
@oquenchil: sorry about the ping – sort of a general question, though: is it be better to just send my patch as a PR and iterate on it, or to discuss the general approach on the ticket before sending a PR?

oquenchil:
[...]

PR #14011, opened 2021-09-19
Add the default solib dir to the rpath for shared libs with transitions.

quval:
When an executable has dynamically-linked dependencies that have transitions, shared libraries are looked up under rpaths like this:

```
$ORIGIN/(../)*_solib_k8/../../../k8-fastbuild-ST-[0-9a-f]+/bin/_solib_k8
```

Unless running under the execroot, the transitioned solib directory may not be available at all; the right files will be present under the default solib directory, so it should be on the rpath.

Adding the default solib directory to the rpath may cause the wrong version of a shared library to be loaded, e.g. if it has been compiled in the default configuration. To prevent this, we also add the transition mnemonic to the mangled name of the library (adding the default solib last to the rpath won't really help, because it can legitimately be added first).

#13819

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CcLinkingHelper.java
@@ -693,7 +693,9 @@ public final class CcLinkingHelper {
             ImmutableList.of(
                 "-Wl,-soname="
                     + SolibSymlinkAction.getDynamicLibrarySoname(
-                        linkerOutput.getRootRelativePath(), /* preserveName= */ false));
+                        linkerOutput.getRootRelativePath(),
+                        /* preserveName= */ false,
+                        actionConstructionContext.getConfiguration().getMnemonic()));
       }
     }
 
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
@@ -267,7 +267,10 @@ public class LibrariesToLinkCollector {
         // When COPY_DYNAMIC_LIBRARIES_TO_BINARY is enabled, dynamic libraries are not symlinked
         // under solibDir, so don't check it and don't include solibDir.
         if (!featureConfiguration.isEnabled(CppRuleClasses.COPY_DYNAMIC_LIBRARIES_TO_BINARY)) {
-          if (libDir.equals(solibDir)) {
+          // The first fragment is bazel-out, and the second may contain a configuration mnemonic.
+          // We should always add the default solib dir because that's where libraries will be found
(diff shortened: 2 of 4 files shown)


quval:
Hey @oquenchil, do you think this might make it on time to Bazel 5? This is mostly the same patch as before, with the change you asked me to make. Thanks!

oquenchil:
Sorry, I need to carefully review this change and see that it doesn't affect anything negatively internally. I haven't had time to do so yet.

oquenchil:
Thanks for being so patient.

This looks good and internally the tests pass with minor modifications that I can take care of. But I'm still trying to wrap my head around all this, so please bear with me.

> Unless running under the **execroot**, the transitioned solib directory may not be available at all; the right files will be present under the default solib directory, so it should be on the rpath.

When you say the execroot, you mean the repository root where you run `bazel build`, do you mean where actions themselves run or something else?

As I understand it, if you move the location of the executable, ORIGIN changes, but the code still works regardless of where you move the binary, doesn't it? So where is the solibdir when you have `$ORIGIN/../_solib_k8/`? You go one directory up from where you are running the binary and then there should be a directory called `solib_k8`? Do you have to put it there manually? Is it part of runfiles and you just have to copy the runfiles?

That's for the default solibdir. Why doesn't the same thing apply for the transitioned solibdir, i.e.: ../_solib_k8/../../../k8-fastbuild-ST-? If you placed the default solibdir there manually, why not this one as well? If the default solibdir was in runfiles and you copied runfiles, how come that the transitioned solibdir isn't in runfiles too?

I realize that it should be the other way around where I'm answering the questions but since you have a better understanding of what the actual problem is I thought you could help.

Thanks!

quval:
I should be the one to thank you – I'm sorry I've been nagging you about it, and I really appreciate your taking the time to work on this. I'm happy to help any way I can; if you'd like to have a chat over this, I can try and demo what I'm seeing – please just let me know.

There seems to be a difference between the layout in local execution (i.e. where `bazel build` is run: this is what I meant by 'under the execroot') and the directory structure where actions are run remotely. `$ORIGIN` is the directory where the executable resides, and the structure of the runfiles dir is such that there's one solib directory there, where all the correctly transitioned SOs are available. When depending on transitioned libraries, rpath entries are added that go all the way up to `k8-fastbuild-ST-...` directories for other configurations. Locally, they do exist and everything works well. In remote execution (and deployment by copying the executable and its runfiles), we only have the runfiles directory, so the only rpath entry that can be actually used is the one that points to the default solib directory. But we don't get an rpath entry pointing there unless we're linking a non-transitioned library. If we don't, all the libraries that Bazel put there can't be found.

Indeed, the trigger for filing the original issue was a cc_test that always failed in remote execution, but always passed locally. I found that it depended only on transitioned libraries – so the default solib dir wasn't on the rpath, whereas that's where all the SOs were in RE. I put together a gist that I hope demonstrates this: https://gist.github.com/quval/5077b2edc902e9dc0bcad2311a245ffa.

PR #14015, opened 2021-09-20
feat: Add missing flag annotations to flags documentation protocol

mattem:
Adds three fields to the `bazel_flags` proto to allow fetching the option effects and metadata tags, and associated documentation category when calling `bazel help flags-as-proto`.  These extra fields can be useful if programmatically parsing and displaying flag documentation.

This does not add any new functionality to flags themselves. It just exposes already existing flag metadata to documentation pipelines.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
@@ -280,7 +280,7 @@ public final class HelpCommand implements BlazeCommand {
     flagBuilder.addAllEffectTags(optionEffectTags);
 
     List<String> optionMetadataTags = Arrays.stream(option.getOptionMetadataTags())
-        .filter(tag -> OptionMetadataTag.INTERNAL.equals(tag))
+        .filter(tag -> !OptionMetadataTag.INTERNAL.equals(tag))
         .map(Enum::toString)
         .collect(Collectors.toList());
     flagBuilder.addAllMetadataTags(optionMetadataTags);
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
@@ -53,6 +53,7 @@ import com.google.devtools.common.options.OptionsParser;
 import com.google.devtools.common.options.OptionsParser.HelpVerbosity;
 import com.google.devtools.common.options.OptionsParsingResult;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
@@ -64,6 +65,7 @@ import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 /** The 'blaze help' command, which prints all available commands as well as specific help pages. */
 @Command(
@@ -271,6 +273,24 @@ public final class HelpCommand implements BlazeCommand {
     flagBuilder.setHasNegativeFlag(option.hasNegativeOption());
     flagBuilder.setDocumentation(option.getHelpText());
     flagBuilder.setAllowsMultiple(option.allowsMultiple());
+
+    List<String> optionEffectTags = Arrays.stream(option.getOptionEffectTags())
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllEffectTags(optionEffectTags);
+
+    List<String> optionMetadataTags = Arrays.stream(option.getOptionMetadataTags())
+        .filter(tag -> OptionMetadataTag.INTERNAL.equals(tag))
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllMetadataTags(optionMetadataTags);
+
+    if (option.getDocumentationCategory() != null &&
+        !OptionDocumentationCategory.UNDOCUMENTED.equals(option.getDocumentationCategory()) &&
+        !OptionDocumentationCategory.UNCATEGORIZED.equals(option.getDocumentationCategory())) {
+      flagBuilder.setDocumentationCategory(option.getDocumentationCategory().toString());
+    }
+
     if (option.getAbbreviation() != '\0') {
       flagBuilder.setAbbreviation(String.valueOf(option.getAbbreviation()));
     }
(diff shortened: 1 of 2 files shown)


mattem:
I couldn't see existing test coverage for `flags-as-proto`, but happy to add some as needed

alexeagle:
This looks really useful, @aiuto any idea who's the best reviewer for this area?

aiuto:
@alexeagle We have a rotation for github issue triage so they can be assigned to the right team and/or reviewer. @oquenchil gave it to configurability, which seems about right.

src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
```diff
@@ -274,6 +276,22 @@ private void emitFlagsAsProtoHelp(BlazeRuntime runtime, OutErr outErr) {
...
+        .collect(Collectors.toList());
+    flagBuilder.addAllEffectTags(optionEffectTags);
+
+    List<String> optionMetadataTags = Arrays.stream(option.getOptionMetadataTags())
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllMetadataTags(optionMetadataTags);
+
+    if (option.getDocumentationCategory() != null &&
+        !OptionDocumentationCategory.UNCATEGORIZED.equals(option.getDocumentationCategory())) {
```
aiuto, line 291: Please also drop UNDOCUMENTED

src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
```diff
@@ -274,6 +276,22 @@ private void emitFlagsAsProtoHelp(BlazeRuntime runtime, OutErr outErr) {
...
+
+    List<String> optionEffectTags = Arrays.stream(option.getOptionEffectTags())
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllEffectTags(optionEffectTags);
+
+    List<String> optionMetadataTags = Arrays.stream(option.getOptionMetadataTags())
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllMetadataTags(optionMetadataTags);
```
aiuto, line 288: Please drop OptionMetadataTag.INTERNAL

aiuto:
@mattem:  I rewrote the PR description to be a little more precise. PTAL.

@brandjon: Please take a quick look to see if you have any concerns w.r.t. StarDoc.
@comius: FYI

aiuto (changes requested):

src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
```diff
@@ -271,6 +273,24 @@ private static void emitFlagsAsProtoHelp(BlazeRuntime runtime, OutErr outErr) {
...
     flagBuilder.setDocumentation(option.getHelpText());
     flagBuilder.setAllowsMultiple(option.allowsMultiple());
+
+    List<String> optionEffectTags = Arrays.stream(option.getOptionEffectTags())
+        .map(Enum::toString)
+        .collect(Collectors.toList());
+    flagBuilder.addAllEffectTags(optionEffectTags);
+
+    List<String> optionMetadataTags = Arrays.stream(option.getOptionMetadataTags())
+        .filter(tag -> OptionMetadataTag.INTERNAL.equals(tag))
```
aiuto, line 283: Is this backwards?  Won't that drop everything except INTERNAL?

aiuto (changes requested):
Please try not to force push. github seems to lose the ability to diff since the last commit, making review harder.

aiuto (commented):
I've been thinking about this, and perhaps my requests don't make sense.  If that is true, I apologize for the needless work.   It's not that we don't want to add the UNDOCUMENTED and INTERNAL tags to the proto. Instead we should not generate any output for the internal tags.

Do you have a before and after sample that can show the impact of the change? My hunch is that you won't see anything that is undocumented in the first place, in which case, not setting the tag doesn't do anything useful.  I don't necessarily need a test, just a statement in the PR description that we aren't emitted docs for the undocumented flags.  If we *are* doing that, then we have a different problem.

aiuto:
Any thoughts?

mattem:
Sorry dropped the ball here. I see what you are saying. So taking a look at the raw proto out, we don't get any `INTERNAL` flags through if we remove the filtering, but we do get `UNCATEGORIZED` and `UNDOCUMENTED` flags in the proto.

With the filtering removed:
```
1 {
  1: "experimental_incremental_dexing_after_proguard"
  2: 0
  3: "Whether to use incremental dexing tools when building proguarded Android binaries.  Values > 0 turn the feature on, values > 1 run that many dexbuilder shards."
  4: "aquery"
  4: "build"
  4: "canonicalize-flags"
  4: "clean"
  4: "config"
  4: "coverage"
  4: "cquery"
  4: "info"
  4: "mobile-install"
  4: "print_action"
  4: "run"
  4: "test"
  6: 0
  7: "LOADING_AND_ANALYSIS"
  7: "LOSES_INCREMENTAL_STATE"
  9: "UNDOCUMENTED"
}
```

Vs still getting the flag though but without the tags:

```
1 {
  1: "experimental_incremental_dexing_after_proguard"
  2: 0
  3: "Whether to use incremental dexing tools when building proguarded Android binaries.  Values > 0 turn the feature on, values > 1 run that many dexbuilder shards."
  4: "aquery"
  4: "build"
  4: "canonicalize-flags"
  4: "clean"
  4: "config"
  4: "coverage"
  4: "cquery"
  4: "info"
  4: "mobile-install"
  4: "print_action"
  4: "run"
  4: "test"
  6: 0
  7: "LOADING_AND_ANALYSIS"
  7: "LOSES_INCREMENTAL_STATE"
}
```

Can remove them from the output by filtering above:

```
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/HelpCommand.java
@@ -244,7 +244,9 @@ public final class HelpCommand implements BlazeCommand {
     BiConsumer<String, OptionDefinition> visitor =
         (commandName, option) -> {
           if (ImmutableSet.copyOf(option.getOptionMetadataTags())
-              .contains(OptionMetadataTag.INTERNAL)) {
+              .contains(OptionMetadataTag.INTERNAL) ||
+              OptionDocumentationCategory.UNDOCUMENTED.equals(option.getDocumentationCategory()) ||
+              OptionDocumentationCategory.UNCATEGORIZED.equals(option.getDocumentationCategory())) {
             return;
           }
           BazelFlagsProto.FlagInfo.Builder info =
```

Let me know if that's the desired state.

aiuto:
UNCATEGORIZED should obviously be output.
I think it is fine to emit UNDOCUMENTED flags, as long as you are emitting the tag too.
That way the renderer can choose what to do.  Either mark them *undocumented* or drop them entirely.

mattem:
Okay, so I think the latest commit has the behaviour we are going for. It will completely drop `INTERNAL` flags from the proto (existing behaviour), but will keep both `UNDOCUMENTED` and `UNCATEGORIZED` flags but will always tag them as such, as therefore left up to the consumer to decide how to present them.

aiuto (approved):

aiuto:
I'll start a merge.

PR #14018, opened 2021-09-21
Fix error in worker declaration example sniplet

bozaro:
With line:
```
  executable=ctx.attr.worker,
```
I got error like:
```
Error in run: in call to run(), parameter 'executable' got value of type 'Target', want 'File, string, or FilesToRunProvider'
```


gregestren:
@susinmotion - could you do a quick domain (expert) review?

susinmotion (approved):
Thanks for fixing!

PR #14019, opened 2021-09-21
Add a test execgroup to Starlark test rules, if not manually defined.

### LINKED ISSUE #13986: "test" exec_group is not present on starlark defined rules

The "test" exec_group added in https://github.com/bazelbuild/bazel/commit/0b3a5aad5818dd5bc1fd3c6b5c3ada77de85f328 does not work for starlark defined rules. It appears to only work for native rules.

All TestRunner actions should have the "test" exec_group

An example repo where I have been playing with remote exec & exec_groups is https://github.com/djmarcin/bazel-remote-exec, but the easiest way to reproduce the bug is to attempt to set the "test" exec group on a `rust_test` rule directly and observe the following error:
```
ERROR: Analysis of target '//:hello_rs_test' failed; build aborted: Tried to set properties for non-existent exec groups: test.
```

Linux, macOS

release 4.2.1
or
release 5.0.0-pre.20210831.2

n/a

n/a

The bazel documentation seems to indicate that this ought to work https://docs.bazel.build/versions/main/exec-groups.html#execution-groups-for-native-rules

Issue filed as a result of discussion at https://github.com/bazelbuild/bazel/pull/13110#issuecomment-917345432
@katre

quval:
Fixes #13986.

### DIFF
--- a/site/docs/exec-groups.md
+++ b/site/docs/exec-groups.md
@@ -39,11 +39,11 @@ namely any constraints via `exec_compatible_with` and toolchain types via
 my_rule = rule(
     _impl,
     exec_groups = {
-        “link”: exec_group(
-            exec_compatible_with = [ "@platforms//os:linux" ]
+        "link": exec_group(
+            exec_compatible_with = ["@platforms//os:linux"],
             toolchains = ["//foo:toolchain_type"],
         ),
-        “test”: exec_group(
+        "test": exec_group(
             toolchains = ["//foo_tools:toolchain_type"],
         ),
     },
@@ -61,6 +61,9 @@ attribute param and the
 module. The module exposes an `exec` function which takes a single string
 parameter which is the name of the exec group for which the dependency should be
 built.
+
+As on native rules, the `test` execution group is present by default on Stalark
+test rules.
 
 ### Execution group inheritance
 
@@ -80,7 +83,7 @@ those specified by the rule itself. In other words, given the following:
 my_rule = rule(
     _impl,
     exec_groups = {
-        “copied”: exec_group(
+        "copied": exec_group(
             copy_from_rule = True,
             # This will inherit exec_compatible_with and toolchains.
             # Setting them here directly would be an error, however.
@@ -194,7 +197,7 @@ my_rule = rule(
     toolchains = ["//foo:toolchain_type"],
     exec_groups = {
         # The following two groups have the same toolchains and constraints:
-        “foo”: exec_group(copy_from_rule = True),
+        "foo": exec_group(copy_from_rule = True),
         "bar": exec_group(
             exec_compatible_with = [ "@platforms//os:linux" ],
             toolchains = ["//foo:toolchain_type"],
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
@@ -15,6 +15,7 @@
 package com.google.devtools.build.lib.analysis.starlark;
 
 import static com.google.devtools.build.lib.analysis.BaseRuleClasses.RUN_UNDER;
+import static com.google.devtools.build.lib.analysis.BaseRuleClasses.TEST_RUNNER_EXEC_GROUP;
 import static com.google.devtools.build.lib.analysis.BaseRuleClasses.TIMEOUT_DEFAULT;
 import static com.google.devtools.build.lib.analysis.BaseRuleClasses.getTestRuntimeLabelList;
 import static com.google.devtools.build.lib.packages.Attribute.attr;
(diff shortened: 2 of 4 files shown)


quval:
Unfortunately, the one specific shard that had a legitimate failure before, failed again on some technical issue – just when I thought I fixed the problem... is there a way to rerun that specific shard?

katre:
I hit the "retry" button for the failing shard, we'll see what happens.

site/docs/exec-groups.md
```diff
@@ -39,11 +39,11 @@ namely any constraints via `exec_compatible_with` and toolchain types via
 my_rule = rule(
     _impl,
     exec_groups = {
-        “link”: exec_group(
-            exec_compatible_with = [ "@platforms//os:linux" ]
+        "link": exec_group(
```
katre, line 42: Thanks for fixing these.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
```diff
@@ -402,6 +403,9 @@ public StarlarkCallable rule(
       }
       builder.addExecGroups(execGroupDict);
     }
+    if (test && (execGroupDict == null || !execGroupDict.containsKey(TEST_RUNNER_EXEC_GROUP))) {
```
katre, line 406: The actual `execGroups` data in `RuleClass.Builder` is a map: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/RuleClass.java;drc=8965d25af90ba975cd207c84fa461b499d8d51c9;l=820

Would it be easier to expose a `Set<String> getExecGroupNames()` or `boolean hasExecGroup(String name)` method on `RuleClass.Builder`?

This works but it's not elegant and requires keeping track of the existing groups. It's fine for Starlark, but native rules can have inheritance which would make this trickier.

src/test/shell/integration/exec_group_test.sh
```diff
@@ -297,6 +297,71 @@ EOF
...
+    },
+    constraint_values = [":local"],
+)
+EOF
+
+  bazel test --extra_execution_platforms="${pkg}:my_platform" ${pkg}:a --execution_log_json_file out.txt || fail "Test failed"
+  grep "platform_key" out.txt || fail "Did not find the platform key"
+  grep "test_value" out.txt || fail "Did not find the test-action value"
+
+  bazel test --extra_execution_platforms="${pkg}:my_platform" ${pkg}:b --execution_log_json_file out.txt || fail "Test failed"
```
katre, line 360: Can you split this into two separate tests? One for the default test exec group, one for the manual? I think that will make any future errors easier to diagnose.

katre (changes requested):
This looks good, I have a few questions and requests.

Thanks very much for taking this on!

src/test/shell/integration/exec_group_test.sh
quval: Done.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
quval: Done. I was wondering about the same thing – my reasoning against this was that I didn't see getters on any other builder classes, and I wasn't sure what you'd think of it.

site/docs/exec-groups.md
quval: Sure!

src/test/shell/integration/exec_group_test.sh
```diff
@@ -297,6 +297,106 @@ EOF
...
+
+  cat > ${pkg}/defs.bzl <<EOF
+def _impl(ctx):
+    out_file = ctx.actions.declare_file("$script_name")
+    ctx.actions.write(out_file, "$script_content", is_executable=True)
+    return [DefaultInfo(executable = out_file)]
+
+starlark_test = rule(
+    implementation = _impl,
+    test = True,
```
katre, line 320: Add a comment here that the "test" exec group exists by default in "test" rules.

src/test/shell/integration/exec_group_test.sh
```diff
@@ -297,6 +297,106 @@ EOF
...
+def _impl(ctx):
+    out_file = ctx.actions.declare_file("$script_name")
+    ctx.actions.write(out_file, "$script_content", is_executable=True)
+    return [DefaultInfo(executable = out_file)]
+
+starlark_with_manual_test_exec_group_test = rule(
+    implementation = _impl,
+    test = True,
+    exec_groups = {
+        "test": exec_group(),
```
katre, line 370: Add a comment here that this overrides the default "test" exec group.

Can we add a property or something here and verify it's set when the target is built?

katre (changes requested):
LGTM, some small test tweaks and this is ready to import.

src/test/shell/integration/exec_group_test.sh
quval: Done (I thought it was enough that the test name said it).

src/test/shell/integration/exec_group_test.sh
quval: Done. Initially I used `run_shell` instead of `write` and tested that we aren't seeing the value for the test execgroup on `bazel build`, but unfortunately this seemed not to build on Windows, and `write` actions don't show up in the execution log. So I'm not sure how to do this.

katre: Hmm, you're right, I don't see an easy way to do that. This is better than what we have now, so that's good.

katre (approved):
Thanks! I'll merge this shortly.

PR #14021, opened 2021-09-22
Fix _is_shared_library_extension_valid

erenon:
Consider libfoo-1.0.so.1.0 valid.
Do not fail if the name does not contain a '.', return False instead.


erenon:
Build is green now, ptal.

PR #14039, opened 2021-09-26
Fix invalid javadoc

cyberdelia:
Fix invalid Javadoc that leads to a ` NullPointerException` when running the `javadoc` tool against it.


tetromino (approved):
Thanks!

Issue #13344, opened 2021-04-13
Compression Support


bansalvinayak:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:
Remote Execution API has released support for Compressed Transfer (https://github.com/bazelbuild/remote-apis/pull/168), and it would be good for bazel to take advantage of it

### Feature requests: what underlying problem are you trying to solve with this feature?
Save network bytes and speed up upload and download transfers for network constrained clients

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
NA

### What operating system are you running Bazel on?
NA

### What's the output of `bazel info release`?
All versions are in scope

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.
NA

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?
NA

###  Have you found anything relevant by searching the web?

> Replace these lines with your answer.
NA

### Any other information, logs, or outputs that you want to share?

> Replace these lines with your answer.
>
> If the files are large, upload as attachment or provide link.

PR #12437, opened 2020-11-09
Add zstd-jni to //third_party.

philwo:
Extracted from #11968, as I have to merge this separately from the rest
of that PR.

PR #11968, opened 2020-08-18
Support zstd decompression for external dependencies

glukasiknuro:
Bug: [Add support for .tar.zst to http_archive](https://github.com/bazelbuild/bazel/issues/10342)

Handling of jni headers and loading system library is very similar to one already in bazel for native os library and possibly can be unified at some point.

Checked repo with big external deps, and noticed following improvements when checking bazel tracing profile:

```
size (tar.gz) |  extracting tar.gz  | extracting tar.zst
--------------------------------------------------------
    773M      |       13 s          |       3.2 s
    421M      |       13 s          |         5 s
    346M      |      9.2 s          |       3.3 s
    261M      |      7.5 s          |         3 s
```

One thing to check is whether bazel distributes binaries compiled with `-c opt` - personally tested with version compiled with this option.

PR #14041, opened 2021-09-26
Add support for compression on gRPC cache

AlessandroPatti:
Add support for compressed transfers from/to gRPC remote caches. Tested with [bazel-remote](https://github.com/buchgr/bazel-remote) 2.13.

Relates to https://github.com/bazelbuild/bazel/issues/13344


AlessandroPatti:
@googlebot I signed it!

src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
```diff
@@ -345,12 +352,41 @@ public static String getResourceName(String instanceName, Digest digest) {
...
+                }
+              }
+
               @Override
               public void onNext(ReadResponse readResponse) {
                 ByteString data = readResponse.getData();
                 try {
-                  data.writeTo(out);
-                  offset.addAndGet(data.size());
+                  if (options.cacheByteStreamCompression) {
```
coeuvre, line 381: Instead of handling two cases here, can we wrap the output stream ahead of time if compression is enabled e.g.
```
if (options.cacheByteStreamCompression) {
    out = decompress(out);
}
```

so whenever we write to the new `out`, the content is decompressed and redirected to the original `out`.

With that, we can keep the code here unchanged.

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -146,7 +191,35 @@ public void seek(long toOffset) throws IOException {
       reset();
     }
     maybeInitialize();
-    ByteStreams.skipFully(data, toOffset - offset);
+    if (compressed && toOffset > 0) {
```
coeuvre, line 194: Things are getting complicated here. Can we create a subclass `ZstdChunker` instead to handle zstd compression?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -364,6 +364,14 @@ public String getTypeDescription() {
               + "symlinks and represent them as files. See #6631 for details.")
   public boolean incompatibleRemoteSymlinks;
 
+  @Option(
+      name = "experimental_cache_bytestream_compression",
```
coeuvre, line 368: Can we change to `experimental_remote_cache_compression`?

coeuvre (changes requested):
Thanks for the PR!

For reasons, we can't merge PR that contains both `non third_party` and `third_party` changes. Can you split `third_party` changes into another PR?

benjaminp:
Note that zstd-jni has a sad history here: #12437, #11968

AlessandroPatti:
> Note that zstd-jni has a sad history here: #12437, #11968

Thank you @benjaminp for pointing that out. @philwo seems like #12437 got reverted, what was the reason of the revert? From #11968, it seems there was an issue with the version of apache commons compress, which is not used here. Can zstd-jni be merged back?

philwo:
> seems like #12437 got reverted, what was the reason of the revert?

From the rollback commit: "We're rolling back this feature before cutting Bazel 4.0 LTS, as we can't merge the remaining part in time and are worried that the newly added patched JNI dependency will make it hard for distributions like Debian to accept. Let's revisit this when we find a suitable pure Java version of Zstd."

The only pure Java version I could find was https://github.com/airlift/aircompressor, which still doesn't support Big Endian (and it doesn't look like they care about that). So, we'd be OK with adding zstdlib and zstd-jni now, probably.

> Can zstd-jni be merged back?

- @meteorcloudy @olekw how difficult would this additional dependency be for Debian (it seems like it's currently not packaged, but maybe it could easily be)?
- We can't accept prebuilt binary JARs (I know we still have some, but we really need to get rid of them) and instead would like to build all dependencies from source. We'd have to replace the checked in JAR files in this PR with instead downloading zstdlib and zstd-jni's sources in our WORKSPACE file, creating BUILD files for them and then using that to build the whole thing from source. I'd love to open-source the BUILD files that we use internally for these two libraries for your reference, let me see whether I can easily do this...
- @meisterT should have another look at it from a performance / binary size increase perspective then

coeuvre:
The import is almost done except that we use an older zstd-jni version internally which doesn't contains the fix https://github.com/luben/zstd-jni/commit/3d51bdcb82c0003f943416cae7aba3f3001b1163. So the test `streamCanBeDecompressedOneByteAtATime` is stuck in an infinite loop.

@AlessandroPatti, while we are waiting for upgrading internal zstd-jni (it's not an easy task and I am afraid that it won't be done soon), is there any workaround we could apply?

AlessandroPatti:
> The import is almost done except that we use an older zstd-jni version internally which doesn't contain the fix [luben/zstd-jni@3d51bdc](https://github.com/luben/zstd-jni/commit/3d51bdcb82c0003f943416cae7aba3f3001b1163). So the test `streamCanBeDecompressedOneByteAtATime` is stuck in an infinite loop.
> 
> @AlessandroPatti, while we are waiting for upgrading internal zstd-jni (it's not an easy task and I am afraid that it won't be done soon), is there any workaround we could apply?

What version are you using internally? I can try to reproduce on my end and see if there is any workaround.
Alternatively, can we disable the test for lower zstd versions (e.g. `if (ZstdVersion.VERSION != "1.5.0-4") return;`) to make the Pr merge cleanly? As long as the compression does not get enabled, this code path wouldn't get hit internally and it can wait for `zstd-jni` to be upgraded.

coeuvre:
We are at https://github.com/luben/zstd-jni/commit/190e4f2e690d1d6316f2c41947dbbe2434ead399 with some custom patches.  

> Alternatively, can we disable the test for lower zstd versions (e.g. if (ZstdVersion.VERSION != "1.5.0-4") return;) to make the Pr merge cleanly? 

SGTM.

coeuvre:
Imported and waiting for the final review.

Some zstd tests are not tested internally but should be fine since they are always tested on Bazel CI.

brentleyjones:
@coeuvre What's your feeling on the risk level of getting this into 5.0?

coeuvre:
It should be fine to get this into 5.0 since it is guarded by an experimental flag.

werkt:
So I got up close and personal with this implementation when I added zstd support to buildfarm, per the supplied streams in the zstd package, and I feel this needs to be reworked.

The overall stream is not being compressed for transport in a tidily shorter sequence of upload/download chunks. Instead, each chunk is being compressed and sent as a shorter WriteRequest/ReadResponse. This results in drastically poorer overall byte ratio of identity:compressed compared to a simple `zstd` invocation on blob content, with a higher possibility of blob *inflation* due to the reduced context and overhead of the smaller blocks.

This was discovered when the upload went through buildfarm's rechunking mechanism for shards, where request chunks would be combined or split on 16k boundaries by default, and the stream processor only worked on the current chunk, discarding appended bytes, and eliciting decompression corruption.

Overall, this implementation is in conflict with the zstd compressor description in the remote apis, which makes no mention of sequential compressor streams in compressed packed blob content, but instead expects the full content to be deliverable to a compressor engine in a single context.

I'm going to try to adjust this to get a full blob's worth of compression going, with even block sizes and no per-request/response semantics for the stream.

PR #14049, opened 2021-09-27
Enable the toolchain transition for all toolchains.

### LINKED ISSUE #11584: incompatible_override_toolchain_transition: Migrate rules to use the toolchain transition
This flag unconditionally makes all rule uses the new [toolchain transition](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-toolchain-transitions.md). This means that any rule that uses [toolchain resolution](https://docs.bazel.build/versions/3.2.0/toolchains.html) to fetch tools will need to migrate so that tools are in the correct configuration.

**Flag:** `--incompatible_override_toolchain_transition`

**Migration:**

First, wait for the new feature to be released (presumably in the release after 3.3.0).

See [toolchain transition migration](https://github.com/bazelbuild/proposals/blob/master/designs/2020-02-07-toolchain-transition-migration.md) for details, but the process is fairly simple:
1. For rules that **create** toolchains (i.e., `cc_toolchain`, `go_toolchain`, `java_toolchain`), add the proper `cfg` attribute.
   1. If the attribute is a tool used in builds (a compiler, a linker, anything that is executed), add `cfg = 'exec'` so that it will be built for the correct execution platform.
   1. If the attribute is intended to be part of the final target (a library, included source, anything else), add `cfg = 'target'`, so that it will be built in the correct target configuration.
   1. At this point, there will be no change, because toolchains still use the host transition, and that overrides anything set on the toolchain rule itself.
2. For each rule that **uses** a toolchain (i.e., `cc_library`, `go_binary`, `java_test`), add the `incompatible_use_toolchain_transition = True` parameter to the rule definition in Starlark.
   1. This enables the toolchain transition, instead of the host transition, when this rule type depends on a toolchain.
   1. Every toolchain type this rule uses should have been migrated in step 1 above.
3. Once every rule has been migrated, wait for Bazel to flip the value of the `--incompatible_override_toolchain_transition` flag. This will take several releases for all rules to finish steps 1. and 2.
4. Once the flag has flipped, remove the `incompatible_use_toolchain_transition` parameter from the rule declaration.
5. Bazel will eventually deprecate the flag and rule parameter and remove them.

katre:
RELNOTES: The toolchain transition is now enabled for all toolchains.

Closes #11584.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/PlatformOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/PlatformOptions.java
@@ -216,7 +216,7 @@ public class PlatformOptions extends FragmentOptions {
 
   @Option(
       name = "incompatible_override_toolchain_transition",
-      defaultValue = "false",
+      defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = OptionEffectTag.LOADING_AND_ANALYSIS,
       metadataTags = {OptionMetadataTag.INCOMPATIBLE_CHANGE},


katre:
Downstream tests: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2190

Filed https://github.com/bazelbuild/rules_go/issues/2967 about the one error.

gregestren (approved):

PR #14063, opened 2021-09-30
Add cc_toolchain.target_transition_for_inputs attribute.

jin:
If this attribute is enabled, all file inputs to the cc_toolchain
actions will be transitioned to the target platform instead of the
execution platform. This allows bootstrap cc_toolchains to build
toolchain libraries from source.


jin:
@lberki was this something that you were envisioning? I had to add the else case in DependencyResolver to add the executionPlatformLabel, since the attributes are now using a non-ExecutionTransitionFactory, which resulted in the label not being set.

jin:
@oquenchil this attribute is used in AOSP's cc_toolchain (https://android-review.googlesource.com/c/platform/prebuilts/clang/host/linux-x86/+/1841853) to build source libraries to be used in toolchain features for link actions.

oquenchil:
Hi Jin,

This looks like it meets perfectly the requirements of the crosstool-dependencies-as-source doc. Are there any differences I'm missing?

jin:
@oquenchil Not that I know of, based on my understanding of the doc. If it looks right, I can go ahead and write some tests and send it internally for your review.

oquenchil:
Yes, thank you very much Jin.

Issue #5373, opened 2018-06-12
Pseudoterminals don't work in the Linux sandbox


lee-bin:
### Description of the problem / feature request:

When I run a python script, which uses the 'sh' module,  from bazel genrule, it failed with this:
```
INFO: Analysed target //src:foo_gen (8 packages loaded).
INFO: Found 1 target...
ERROR: /home/libin11/workspace/test/test/src/BUILD:1:1: Executing genrule //src:foo_gen failed (Exit 1)
Traceback (most recent call last):
  File "src/test.py", line 2, in <module>
    sh.touch("foo.bar")
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 1427, in __call__
    return RunningCommand(cmd, call_args, stdin, stdout, stderr)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 767, in __init__
    self.call_args, pipe, process_assign_lock)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 1784, in __init__
    self._stdout_read_fd, self._stdout_write_fd = pty.openpty()
  File "/usr/lib/python2.7/pty.py", line 29, in openpty
    master_fd, slave_name = _open_terminal()
  File "/usr/lib/python2.7/pty.py", line 70, in _open_terminal
    raise os.error, 'out of pty devices'
OSError: out of pty devices
Target //src:foo_gen failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 2.143s, Critical Path: 0.12s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```
### Feature requests: what underlying problem are you trying to solve with this feature?

I want to integrate a thirdparty project to my own. The third party project is built with a python script, so I would like to build the project with bazel genrule. 

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Example file list:
```
.
├── src
│   ├── BUILD
│   └── test.py
└── WORKSPACE
```
WORKSPACE is empty, BUILD is:
```
genrule(
    name = "foo_gen",
    srcs = glob(["**/*"]),
    outs = ["foo.bar"],
    cmd = "python $(location test.py)",
)
```
test.py is:
```
import sh
sh.touch("foo.bar")
```
And run:
```
bazel build //src:foo_gen
```
### What operating system are you running Bazel on?

Ubuntu 16.04

### What's the output of `bazel info release`?

release 0.14.1

[...]

PR #14072, opened 2021-10-04
Functionality for pseudoterminals in linux sandbox

crydell-ericsson:
As brought up in issue #5373 , the Linux sandbox does not allow processes that run inside it to open pseudoterminals. These changes enable this by addressing the two main underlying issues:

- `/dev/pts` can not be read-only if a new pseudoterminal is to be created. These changes make `dev/pts` writable when remounting file systems during sandbox initialization.
- The group associated with pseudoterminals is "tty". After creating a new pseudoterminal, its gid has to be changed. If there is no gid mapping in the user namespace that corresponds to "tty", this group will not be known inside the sandbox. This causes issues in some Linux distributions, since they do not allow changing the group of a file to one that is not known inside the current user namespace. These changes map the gid of the user to the one corresponding to "tty" inside the sandbox in order to avoid this issue.

These changes introduce the `-P` flag to `linux-sandbox` in order to control whether or not the changes are applied, and the `--sandbox-explicit-pseudoterminal` to `bazel` in order to set this when calling bazel.

### DIFF
--- a/src/main/tools/linux-sandbox-pid1.cc
+++ b/src/main/tools/linux-sandbox-pid1.cc
@@ -231,6 +231,11 @@ static void SetupUserNamespace() {
     inner_uid = global_outer_uid;
     inner_gid = global_outer_gid;
   }
+  if (opt.enable_pty) {
+    // Change the group to 'tty' regardless of what was previously set
+    inner_gid = 5;
+  }
+
   WriteFile("/proc/self/uid_map", "%d %d 1\n", inner_uid, global_outer_uid);
   WriteFile("/proc/self/gid_map", "%d %d 1\n", inner_gid, global_outer_gid);
 }
@@ -302,6 +307,10 @@ static void MountFilesystems() {
 // returns true.
 static bool ShouldBeWritable(const std::string &mnt_dir) {
   if (mnt_dir == opt.working_dir) {
+    return true;
+  }
+
+  if (opt.enable_pty && mnt_dir == "/dev/pts") {
     return true;
   }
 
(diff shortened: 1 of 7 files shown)


philwo:
Thank you @crydell-ericsson! Ideally I'd like to avoid adding yet another flag... is there a way to automatically detect what the right thing to do is?

Could we just always mount `/dev/pts` and somehow map or make the `tty` group usable inside the sandbox without explicitly changing to it?

crydell-ericsson:
Just thought I'd check in on the progress of this pull request; is there any additional information needed for this?

philwo:
Thank you for the information, @crydell-ericsson. It's too bad that the best solution requires CAP_SETGID. 😞 I agree with your thoughts.

@larsrc-google Let's merge this as is? Looks clean to me and all changed behavior is behind a flag.

src/main/tools/linux-sandbox-pid1.cc
```diff
@@ -232,8 +233,9 @@ static void SetupUserNamespace() {
     inner_gid = global_outer_gid;
   }
   if (opt.enable_pty) {
-    // Change the group to 'tty' regardless of what was previously set
-    inner_gid = 5;
+    // Change the group to "tty" regardless of what was previously set
+    struct group *g = getgrnam("tty");
```
larsrc-google, line 237: A friendly internal linter suggests using getgrnam_r for extra thread safety.

larsrc-google:
@crydell-ericsson This PR is out of date, could you update it?

PR #14075, opened 2021-10-04
Upgrade JDK11 version

comius:


comius:
Downstream test at: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2200

comius:
cc @hvadehra @cushon

PR #14080, opened 2021-10-05
fix resource leak due to Files.newDirectoryStream

lujiefsi:
JDK has point out that we shoud close the stream opened by Files.newDirectoryStream , as below

`When not using the try-with-resources construct, then directory
     * stream's {@code close} method should be invoked after iteration is
     * completed so as to free any resources held for the open directory. `

but the code at https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/tools/jdk/DumpPlatformClassPath.java#L185 

do not the close the stream. This pull request use try-with-resources to fix this problem.

### DIFF
--- a/tools/jdk/DumpPlatformClassPath.java
+++ b/tools/jdk/DumpPlatformClassPath.java
@@ -182,8 +182,10 @@ public class DumpPlatformClassPath {
     List<Path> jars = new ArrayList<>();
     Path extDir = javaHome.resolve("jre/lib/ext");
     if (Files.exists(extDir)) {
-      for (Path extJar : Files.newDirectoryStream(extDir, "*.jar")) {
-        jars.add(extJar);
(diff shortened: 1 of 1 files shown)


aiuto (changes requested):
Please update the PR description to point to a bug that describes the problem.
If there is no existing bug, please provide enough information so someone can reason about the change. Then we'll find the right reviewer.

sgowroji:
Hello @lujiefsi,  Did you get a chance to check the above [comment](https://github.com/bazelbuild/bazel/pull/14080#pullrequestreview-772143763) and can you kindly reply. Thanks!

ckolli5:
@lujiefsi - I'm going to go ahead and close this PR, because it seems to have stalled. If you're still interested in pursing this (and responding to my comments), please feel free to reopen!

PR #14085, opened 2021-10-05
Fix overly aggressive regex when filtering resource classess

mauriciogg:
The regex in place will filter any class that ends in R.class or R$.*.class (e.g. leakcanary/internal/ObjectsKt$NO_OP_HANDLER$1.class)


ahumesky:
Thanks, I'll start the import on this

PR #14086, opened 2021-10-06
[Cherry pick] LLVM BC support

omarzl:
Cherry pick request from #13889

----

Using rules_swift it is possible to generate LLVM IR BC files using swift.emit_bc feature, some days ago a refactor to use create_compilation_outputs from cc_common was merged, it is giving us the following error since the bc extension is missing, this pr fixes it

```
Error in create_compilation_outputs: 'Somefile.swift.bc' has wrong extension. The list of possible extensions for 'objects' is: .o,.obj,.pic.o,.opb
```


thii:
Since 4.2.1 has been released, I think you have to request a new patch release 4.2.2 to the release manager https://github.com/bazelbuild/bazel/issues/13558

PR #14096, opened 2021-10-12
Refactor Incompatible Target Skipping

### LINKED ISSUE #12897: Cannot use target_compatible_with to ignore MacOS only targets

I'm trying to use the `target_compatible_with` attribute to automatically skip MacOS only targets on other platforms, but I just get an error every time. I have a simple `objc_library` rule with `target_compatible_with = [ "@platforms//os:macos" ]` but whenever I run `bazel build //...:all` on non MacOS platforms I get this:

```log
ERROR: /root/.cache/bazel/_bazel_root/eab0d61a99b6696edb3d2aff87b585e8/external/local_config_cc/BUILD:47:19: in cc_toolchain_suite rule @local_config_cc//:toolchain: cc_toolchain_suite '@local_config_cc//:toolchain' does not contain a toolchain for cpu 'darwin_x86_64'
ERROR: Analysis of target '//ble/platform/macos:ble_macos' failed; build aborted: Analysis of target '@local_config_cc//:toolchain' failed
```

Any `objc_library` rule fails instead of being ignored.
```bazel
objc_library(
    name = 'ble',
    target_compatible_with = [
        "@platforms//os:macos",
    ],
)
```

I added a simple example here: [jagobagascon/bazel-target-compatible](https://github.com/jagobagascon/bazel-target-compatible)
It's just an empty `objc_library` rule but you can se how the Ubuntu build fails while the MacOS just works: [jagobagascon/bazel-target-compatible/actions/runs/512144112](https://github.com/jagobagascon/bazel-target-compatible/actions/runs/512144112)

I'm running it on a Docker image based on `ubuntu:focal-20200423`, but it also happens on a GitHub-hosted `ubuntu-18.04`.

release 4.0.0

philsc:
This patch reworks how Incompatible Target Skipping is implemented.

The biggest change is that incompatibility is now checked earlier. This
allows us to avoid evaluating dependencies (such as toolchains) in
situations where a target is "directly" incompatible. "Direct incompatibility"
refers to incompatibility due to a target's `target_compatible_with` value.

Moving the incompatibility checking earlier had the undesired effect of
visibility restrictions being ignored on incompatible targets. This is
tracked in #16044. As per
https://github.com/bazelbuild/bazel/pull/14096#issuecomment-1189589626,
we will fix it in a separate patch.

Fixes #12897.


src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -440,6 +542,62 @@ public SkyValue compute(SkyKey key, Environment env)
     }
   }
 
+  // Taken from ConfiguredTargetFactory.convertVisibility().
+  // TODO(phil): Figure out how to de-duplicate somehow.
```
philsc, line 546: @gregestren , can you help me figure out how to de-duplicate this function?

I can _almost_ do it. The part I'm struggling with is figuring out how to deal with the `PackageGroupsRuleVisibility` section. The thing that stumps me is this line here: https://github.com/bazelbuild/bazel/blob/7f74878b9da57666f04cb607a3d7a36122143863/src/main/java/com/google/devtools/build/lib/analysis/ConfiguredTargetFactory.java#L158

What's the significance of the configuration being `null`? I wanted to query the `visibility` labels similar to how I am currently querying the `target_compatible_with` labels above. That, however, I'm doing with this: https://github.com/bazelbuild/bazel/pull/14096/files#diff-6a800890b53e39d722c5457065ec7694a044c12b6f8ac93b2147c15ab7ac95f8R362-R366
That has a non-null configuration.

I'll dig into it more. I'm probably overlooking something simple.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -314,6 +338,63 @@ public SkyValue compute(SkyKey key, Environment env)
                 getPrioritizedDetailedExitCode(causes)));
       }
 
+      Rule rule = null;
+      if (target instanceof Rule) {
+        rule = (Rule) target;
```
gregestren, line 343: Note Rule.getAssociatedRule() also works, so you can combine the conditions.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
```diff
@@ -882,159 +882,4 @@ public static IncompatibleCheckResult checkForIncompatibility(ConfiguredTarget t
     return IncompatibleCheckResult.create(
         target.get(IncompatiblePlatformProvider.PROVIDER) != null, target);
   }
-
```
gregestren, line 885: Sanity check: all this removed logic remains present in the other files?

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -314,6 +338,63 @@ public SkyValue compute(SkyKey key, Environment env)
...
+      } else if (target instanceof OutputFile) {
+        rule = ((OutputFile) target).getAssociatedRule();
+      }
+
+      PlatformInfo platformInfo = unloadedToolchainContexts != null ? unloadedToolchainContexts.getTargetPlatform() : null;
```
gregestren, line 348: Is there any reason we can't keep most of the code in RuleContextConstraintSemantics, even with a few API tweaks to make it compatible from calling here?

Broadly speaking your logic looks reasonable to me. My main thoughts on following up are:

- ConfiguredTargetFunction is a notoriously complex class. I'd like to minimize it's ongoing complexity. Since incompatible target skipping is itself a bunch of logic I think that's a good argument for isolating it in its own dedicated helper file, with a narrow API, and calling it from here with as little change in this file as possible.
- Aside from which file the code is in, I'll want to do some iterations with you to see how we can absolutely minimize the code itself. Most importantly not replicating logic that we may be able to tap into elsewhere, as you note in your comment about visibility checking. That said, it is possible we'll just have a bunch of code, which isn't the worst thing if it's isolated in a file
- Largely related, I want to make sure we stay careful about reimplementing *behavior*: in this case the logic that determines if a platform satisfies a set of constraints. I think this PR doesn't add new custom logic for that, and I think the existing logic is already pretty streamlined. But I want to keep that in the back of my mind.

So basically, this looks like a great effort. But given the complex code and behavior being addressed I'd like to focus on code structure to keep it all lean, efficient, readable, resilient. And get a final PR out of that.

So my next step suggestion is to address the "complex logic in a dedicated" file question, both here and my comment about all the code removed from RuleContextConstraintSemantics. I'd like to understand better why you removed the code from there, and help understand if we can land code there or in some other `IncompatibleDeterminingHelper.java` file or something.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: Noted. Let's hold off on this pending my file comments.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -314,6 +338,63 @@ public SkyValue compute(SkyKey key, Environment env)
...
+      Label platformLabel = platformInfo != null ? platformInfo.label() : null;
+
```
gregestren, line 350: Do you think this will work on aliases? https://github.com/bazelbuild/bazel/commit/1c3a2456c95fd19974a5b2bd33c5ebdb2b2277e4 fixed a related problem, but that only applies when the alias has a `select()`.

The key point is that aliases are some rare rule that don't opt into toolchain resolution: https://github.com/bazelbuild/bazel/blob/1c3a2456c95fd19974a5b2bd33c5ebdb2b2277e4/src/main/java/com/google/devtools/build/lib/rules/Alias.java#L86

That means, as I understand, that if the alias has no `select()`, then none of the toolchain logic kicks in, which means `platformInfo == null`, which would skip all this checking.

Can we verify that on a simple example alias? We historically had similar problems with filegroup but I think that was fixed (checks notes... yup: https://github.com/bazelbuild/bazel/commit/dfb70ea4cae2ffffb76e9741d86c96505a6d05ad)

Here's the full list: https://github.com/bazelbuild/bazel/search?q=%22useToolchainResolution%22. It looks like there are no other rule types to worry about. I'm bringing this point up since people use aliases and I want to protect us against "feature X works except for target Y" bugs as much as we can foresee them.

gregestren (commented):
Left a whole bunch of comments! Thanks for all this effort. Looking forward to working through this with you.

philsc:
Thanks for the feedback @gregestren , I really appreciate it!
I'll try to get back to incorporating it next week.

gregestren:
No worries. Thanks for the update!

Just FYI there's been some unrelated commits on `ConfiguredTargetFunction` to make it more efficient recently. If you run into any issues merging with head please let me know.

keith:
I'd love to see this moving again! We hit this in the swift rules today

gregestren:
FYI @keith : https://github.com/bazelbuild/bazel/issues/15271#issuecomment-1100484502

philsc:
@keith , yeah. Sorry about the delay. Now that robotics season is over, I'll get back to it.

kkpattern:
We also hit this issue. Thanks for the good work. Waiting for it to land.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
philsc: Very nice. Thanks!

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleContextConstraintSemantics.java
philsc: Yes. Now in a file called `RuleConstraintSemantics.java`. I apologize for the name. I will rename it.

The reason it's removed here is that we no longer have access to a `RuleContext`.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
```diff
@@ -335,6 +336,19 @@ public SkyValue compute(SkyKey key, Environment env)
...
       }
 
+      RuleConstraintSemantics.IncompatibleTargetCreationResult incompatibleTargetCreationResult =
+          RuleConstraintSemantics.createDirectlyIncompatibleTarget(
+              targetAndConfiguration,
+              configConditions,
+              env,
+              unloadedToolchainContexts,
+              state.transitivePackagesForPackageRootResolution);
+      if (incompatibleTargetCreationResult.needsMoreConfiguredTargets()) {
```
katre, line 346: (This leads from an earlier comment) Instead of having needsMoreConfiguredTargets, createDirectlyIncompatibleTarget should just return null to signal the need for a skyframe restart.

katre (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+import com.google.devtools.build.lib.packages.Target;
+import com.google.devtools.build.lib.skyframe.ConfiguredTargetAndData;
+import com.google.devtools.build.lib.skyframe.RuleConfiguredTargetValue;
+import com.google.devtools.build.lib.skyframe.UnloadedToolchainContext;
+import com.google.devtools.build.lib.util.OrderedSetMultimap;
+import com.google.devtools.build.skyframe.SkyFunction.Environment;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class RuleConstraintSemantics {
```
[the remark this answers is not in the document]
philsc: I will add this in the next iteration. Apologies. Not having this makes reviewing harder I realize.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+    Target target = targetAndConfiguration.getTarget();
+    Rule rule = target.getAssociatedRule();
+    PlatformInfo platformInfo =
+        unloadedToolchainContexts != null ? unloadedToolchainContexts.getTargetPlatform() : null;
+    Label platformLabel = platformInfo != null ? platformInfo.label() : null;
+    BuildConfigurationValue configuration = targetAndConfiguration.getConfiguration();
+
+    // Check if this target is directly incompatible. In other words, check if it's incompatible
+    // because of its "target_compatible_with" value.
+    if (rule != null && !rule.getRuleClass().equals("toolchain")) {
```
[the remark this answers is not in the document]
philsc: Great callout.

Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+      if (platformInfo != null) {
+        ConfiguredAttributeMapper attrs =
+            ConfiguredAttributeMapper.of(rule, configConditions.asProviders(), configuration);
+        if (attrs.has("target_compatible_with", BuildType.LABEL_LIST)) {
```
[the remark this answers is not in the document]
philsc: It is indeed a configurable attribute. That is how we let users implement OR logic: https://bazel.build/docs/platforms#expressive-constraints

Can you elaborate on the scary part? I have thus far not thought of this as a bad thing. I hope I'm not wrong :P

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+          ImmutableList.Builder<TransitiveInfoCollection> constraintProvidersBuilder =
+              ImmutableList.builder();
+          for (Label constraintLabel : labels) {
+            Dependency constraintDep =
+                Dependency.builder()
+                    .setLabel(constraintLabel)
+                    .setConfiguration(configuration)
+                    .build();
+            ConfiguredTargetValue ctv =
+                (ConfiguredTargetValue) env.getValue(constraintDep.getConfiguredTargetKey());
```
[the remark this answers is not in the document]
philsc: I didn't know that. Thank you!

Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+            if (ctv == null) {
+              return IncompatibleTargetCreationResult.create(true, null);
+            }
+            constraintProvidersBuilder.add(ctv.getConfiguredTarget());
```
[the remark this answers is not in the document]
philsc: Good idea.
Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+    // Check if any dependencies are incompatible. If any are, then this target is also
+    // incompatible.
+    if (rule != null && !rule.getRuleClass().equals("toolchain")) {
+      ImmutableList<ConfiguredTarget> incompatibleDeps =
+          depValueMap.values().stream()
+              .map(ConfiguredTargetAndData::getConfiguredTarget)
+              .filter(
+                  dep ->
+                      RuleContextConstraintSemantics.checkForIncompatibility(dep).isIncompatible())
+              .collect(ImmutableList.toImmutableList());
```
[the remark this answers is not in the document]
philsc: I had no idea this was a thing. Thanks!

Done

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
philsc: I would like to simplify it down to that if possible. Let's discuss further in the earlier comment.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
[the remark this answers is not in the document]
philsc: That's fair. Marking this as resolved.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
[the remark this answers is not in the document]
philsc: Added a javadoc comment. Let me know if you would like more detail and such.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -0,0 +1,240 @@
...
+          for (Label constraintLabel : labels) {
+            Dependency constraintDep =
+                Dependency.builder()
+                    .setLabel(constraintLabel)
+                    .setConfiguration(configuration)
+                    .build();
+            ConfiguredTargetValue ctv =
+                (ConfiguredTargetValue) env.getValue(constraintDep.getConfiguredTargetKey());
+            if (ctv == null) {
+              return IncompatibleTargetCreationResult.create(true, null);
```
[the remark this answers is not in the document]
philsc: I thought about this more and realized that `ConfiguredTargetFunction` already has a pattern for this. When `ConfiguredTargetFunction` calls various functions, the first thing it does is check `env.valuesMissing()`. If that's `true`, then it requests a restart. If it's `false`, then it evaluates the return value.  I decided to follow that pattern here in order to simplify this logic and reduce the return value to a nullable `RuleConfiguredTargetValue.

Let me know what you think of this. It _seemed_ to fit the pattern, but it's possible I'm misunderstanding.

src/main/java/com/google/devtools/build/lib/analysis/constraints/RuleConstraintSemantics.java
```diff
@@ -108,22 +108,29 @@ public static IncompatibleTargetCreationResult createDirectlyIncompatibleTarget(
 
     // Resolve the constraint labels.
     List<Label> labels = attrs.get("target_compatible_with", BuildType.LABEL_LIST);
-    ImmutableList<SkyKey> constraintKeys = labels.stream()
-        .map(label -> Dependency.builder().setLabel(label)
-          .setConfiguration(configuration).build().getConfiguredTargetKey())
-        .collect(toImmutableList());
+    ImmutableList<SkyKey> constraintKeys =
```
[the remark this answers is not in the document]
philsc: Done

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
philsc: Decided to rename the file to `IncompatibleDeterminingHelper.java` because I couldn't think of anything better.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -0,0 +1,259 @@
...
+    return new RuleConfiguredTargetValue(
+        configuredTarget,
+        transitivePackagesForPackageRootResolution == null
+            ? null
+            : transitivePackagesForPackageRootResolution.build());
+  }
+
+  // Taken from ConfiguredTargetFactory.convertVisibility().
+  // TODO(phil): Figure out how to de-duplicate somehow.
+  private static NestedSet<PackageGroupContents> convertVisibility(Target target) {
```
philsc, line 245: Next task on my list is to figure out how to deduplicate this.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
+ */
+public class IncompatibleDeterminingHelper {
+  /**
+   * Creates an incompatible configured target if it is "directly incompatible".
+   *
+   * <p>In other words, this function checks if a target is incompatible because of its
+   * "target_compatible_with" attribute.
+   *
+   * <p>After this function returns, the caller must immediatly check if "env.valuesMissing()"
+   * returns true. If that check succeeds, the return value of this function is guaranteed to be
```
katre, line 90: Nit: "if valuesMIssing() returns {@code true}, the return value of this..."

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
+   * null. The caller must request a skyframe restart. If the check fails, then the return value is
+   * either null or an incompatible configured target. If null is returned, the target is not
+   * directly incompatible and analysis of this target should proceed.
+   */
   @Nullable
-  public static IncompatibleTargetCreationResult createDirectlyIncompatibleTarget(
+  public static RuleConfiguredTargetValue createDirectlyIncompatibleTarget(
```
katre, line 96: I would still prefer that this return an Optional<RuleConfiguredTargetValue>. This helps to clarify the semantics that a missing RCTV is allowed and is not the same as a skyframe restart.

This technically gives you three states of return values:
- null: skyframe restart
- Optional.empty() - target is not directly incompatible
- !Optional.empty() - target is directly incompatible

But it's good that the docs clarify that users should check env.valuesMissing directly rather than relying on that.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
       TargetAndConfiguration targetAndConfiguration,
       ConfigConditions configConditions,
       Environment env,
-      ToolchainCollection<UnloadedToolchainContext> unloadedToolchainContexts,
+      PlatformInfo platformInfo,
```
katre, line 100: This can be `@Nullable`, since callers do not need to verify that platformInfo is non-null.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
+   * Creates an incompatible target if it is "indirectly incompatible".
+   *
+   * <p>In other words, this function checks if a target is incompatible because of one of its
+   * dependencies. If a dependency is incompatible, then this target is also incompatible.
+   *
+   * <p>This function returns either null or an incompatible configured target. If null is returned,
+   * the target is not indirectly incompatible and analysis of this target should proceed.
+   */
   @Nullable
   public static RuleConfiguredTargetValue createIndirectlyIncompatibleTarget(
```
katre, line 166: For simplicity, this should also return an Optional if the previous method does.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
       TargetAndConfiguration targetAndConfiguration,
       OrderedSetMultimap<DependencyKind, ConfiguredTargetAndData> depValueMap,
       ConfigConditions configConditions,
-      ToolchainCollection<UnloadedToolchainContext> unloadedToolchainContexts,
+      PlatformInfo platformInfo,
```
katre, line 170: `@Nullable`

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
+ *
+ * <p>A target is considered incompatible if any of the following applies:
+ *
+ * <ol>
+ *   <li>The target's <code>target_compatible_with</code> attribute specifies a constraint that is
+ *       not present in the target platform.
+ *   <li>One or more of the target's dependencies is incompatible.
+ * </ol>
+ *
+ * The intent of these helpers is that they get called as early in the analysis phase as possible.
```
katre, line 75: I assume that this is why there are two methods, so that the direct check can be done ASAP. Call that out explicitly, I was initially looking to see if it was possible to unify the methods.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
```diff
@@ -54,141 +53,161 @@
...
+ * Doing this as early as possible allows us to skip analysing unused dependencies and ignore unused
+ * toolchains.
+ *
+ * <p>See https://bazel.build/docs/platforms#skipping-incompatible-targets for more information on
+ * incompatible target skipping.
+ */
+public class IncompatibleDeterminingHelper {
```
katre, line 82: IncompatibleTargetChecker?

katre (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: That sounds reasonable. I switched to an `Optional` and updated the comments.

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: Neat. I didn't realize you can decorate function arguments. Thanks!

Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: I like it.

Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleDeterminingHelper.java
philsc: Superseded by the move to `Optional`.

katre (approved):
Thanks for the changes. This looks good to me, I'll check with Greg if he wants to give a look over before we import.

philsc:
Thanks @katre , I appreciate it!

There are 2 things that I still need to address:
1. Deduplicate `convertVisibility()` in `IncompatibleTargetChecker.java`. E.g. it doesn't currently handle package groups.
2. Revert my changes in `RuleConfiguredTargetValue.java`. The `IncompatibleTargetChecker.java` file doesn't have access to the constructor as-is. Need to figure out what to do there. If you have any opinions, I'd appreciate hearing those.

katre:
It looks like the visibility conversion in ITC is much simpler than the version in ConfiguredTargetFactory. Do you think the extra details are needed? Otherwise I don't think this is terrible duplication.

Similarly, I don't see a problem with making RuleConfiguredTargetValue have a public constructor.

philsc:
>It looks like the visibility conversion in ITC is much simpler than the version in ConfiguredTargetFactory. Do you think the extra details are needed? Otherwise I don't think this is terrible duplication.

It's much simpler in `IncompatibleTargetChecker` because it doesn't support `package_group()`.  I had dropped it because I couldn't find a clean way way to add it. But now I gotta deal with it. E.g. in the "directly incompatible" case, I have to fully resolve the `visibility` attribute. Definitely doable. I will give it a try.

> Similarly, I don't see a problem with making RuleConfiguredTargetValue have a public constructor.

Sounds good. I removed my TODO and kept it public.

philsc:
Okay. It turns out that visibility doesn't get checked until a good chunk after the calls to `createDirectlyIncompatibleTarget` and `createIndirectlyIncompatibleTarget` (i.e. normally during `RuleContext` construction). I had hoped it happens during `ConfiguredTargetFunction.computeDependencies`, but that's not the case.
```
at com.google.devtools.build.lib.analysis.CommonPrerequisiteValidator.handleVisibilityConflict(CommonPrerequisiteValidator.java:151)
at com.google.devtools.build.lib.analysis.CommonPrerequisiteValidator.validateDirectPrerequisiteVisibility(CommonPrerequisiteValidator.java:95)
at com.google.devtools.build.lib.analysis.CommonPrerequisiteValidator.validate(CommonPrerequisiteValidator.java:45)
at com.google.devtools.build.lib.analysis.RuleContext$Builder.validateDirectPrerequisite(RuleContext.java:2185)
at com.google.devtools.build.lib.analysis.RuleContext$Builder.createTargetMap(RuleContext.java:1873)
at com.google.devtools.build.lib.analysis.RuleContext$Builder.build(RuleContext.java:1688)
at com.google.devtools.build.lib.analysis.ConfiguredTargetFactory.createRule(ConfiguredTargetFactory.java:316)
at com.google.devtools.build.lib.analysis.ConfiguredTargetFactory.createConfiguredTarget(ConfiguredTargetFactory.java:188)
at com.google.devtools.build.lib.skyframe.SkyframeBuildView.createConfiguredTarget(SkyframeBuildView.java:1179)
at com.google.devtools.build.lib.skyframe.ConfiguredTargetFunction.createConfiguredTarget(ConfiguredTargetFunction.java:1201)
at com.google.devtools.build.lib.skyframe.ConfiguredTargetFunction.compute(ConfiguredTargetFunction.java:416)
at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:573)
at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:382)
at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(Unknown Source)
at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
```

I think the question becomes: If target A depends on an incompatible target B whose visibility restriction excludes A, what do you want to happen?
The current code makes it so that A becomes incompatible without any visibility errors.
If we want visibility errors to be thrown, then I'll have to do some serious thinking. It would be nice if we could call into `CommonPrerequisiteValidator`, but as-is that's not an option.

philsc:
To clarify: I think the behaviour of the code in this PR here is incorrect. Visibility errors _should_ take priority over incompatibility propagation. What do you think about this?

I'm just not sure yet how to reconcile priority with how late in the analysis phase visibility is checked.

katre:
Can we move visibility checking earlier? I haven't reviewed that code recently, @gregestren may have some ideas.

philsc:
I will look into what it would take to move visibility checking earlier.

src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
```diff
@@ -404,7 +404,7 @@ public RuleClass build(RuleClass.Builder builder, RuleDefinitionEnvironment env)
               attr("distribs", DISTRIBUTIONS)
                   .nonconfigurable("Used in core loading phase logic with no access to configs"))
           // Any rule that has provides its own meaning for the "target_compatible_with" attribute
```
gregestren, line 406: Super minor comment typo nit: "has provides"?

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleTargetChecker.java
```diff
@@ -0,0 +1,276 @@
...
+ * <p>A target is considered incompatible if any of the following applies:
+ *
+ * <ol>
+ *   <li>The target's <code>target_compatible_with</code> attribute specifies a constraint that is
+ *       not present in the target platform. The target is said to be "directly incompatible".
+ *   <li>One or more of the target's dependencies is incompatible. The target is said to be
+ *       "indirectly incompatible."
+ * </ol>
+ *
+ * The intent of these helpers is that they get called as early in the analysis phase as possible.
```
gregestren, line 77: Minor bureaucratic style nit: Start this paragraph with a \<p>?

gregestren: As an aside, I think this Javadoc is a clear and concise guide to the meaning behind this code. Very nice maintainability contribution!

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
gregestren: That sounds fine to me. Honestly, I don't remember the full context for my previous comment (i.e. what I meant by saying "which would skip all this checking"? Do you still see anything relevant in that comment? If not I'm happy to resolve, especially since you're showing alias test coverage.

gregestren:
> I will look into what it would take to move visibility checking earlier.

I agree with you that principally visibility checking should happen first - that's the right user experience.

That said, I think it's going to be hard to refactor visibility. Visibility's implementation has a lot of subtle dependencies and I don't think it's easy to, like, just move some `checkVisibility()` function from class A to class B. I say this from experience. :p 

If you see a reasonable way forward on that, let's push on it. If it looks like it's a serious blocker to this feature, I could be convinced for us to replace that with a `TODO: get visibility / incompatible error precedence right`, in the name of getting a nice user experience improvement today at the cost of a somewhat obscure error scenario.

gregestren:
> > I will look into what it would take to move visibility checking earlier.
> 
> I agree with you that principally visibility checking should happen first - that's the right user experience.
> 
> That said, I think it's going to be hard to refactor visibility. Visibility's implementation has a lot of subtle dependencies and I don't think it's easy to, like, just move some `checkVisibility()` function from class A to class B. I say this from experience. :p
> 
> If you see a reasonable way forward on that, let's push on it. If it looks like it's a serious blocker to this feature, I could be convinced for us to replace that with a `TODO: get visibility / incompatible error precedence right`, in the name of getting a nice user experience improvement today at the cost of a somewhat obscure error scenario.

And most of the point of this PR was to invoke the incompatible checking logic before all "normal" rule logic kicks in. So this problem sort of comes up by definition. :)

src/main/java/com/google/devtools/build/lib/analysis/BaseRuleClasses.java
philsc: Good call.
Done

src/main/java/com/google/devtools/build/lib/analysis/constraints/IncompatibleTargetChecker.java
philsc: Forgot to run the formatter on this. Fixed now. Thanks!

philsc:
Fair enough. I spent a couple of hours trying a couple of things, but didn't really get anywhere. The visibility code is more complicated than I expected. I would like to spend another couple hours on it to see if I can figure it out. But otherwise I am okay with creating a ticket for fixing the visibility/incompatibility problem and merging this as-is.

src/main/java/com/google/devtools/build/lib/skyframe/ConfiguredTargetFunction.java
philsc: Resolving sounds good to me. I am pretty sure if there _is_ an issue with this, it should be easy to resolve. That being said, I should make sure that this patch works as expected with  #15426.

philsc: This patch together with #15426 still passes all the tests. Resolving this comment.

gregestren:
> Fair enough. I spent a couple of hours trying a couple of things, but didn't really get anywhere. The visibility code is more complicated than I expected. I would like to spend another couple hours on it to see if I can figure it out. But otherwise I am okay with creating a ticket for fixing the visibility/incompatibility problem and merging this as-is.

Sounds good to me.

philsc:
I won't have time to experiment further with moving visibility checking this week. So I'll do that next week.

That being said, there is a question you could answer for me in case I still don't get anywhere.
I will file a Github issue to keep track of the visibility/incompatibility strangeness. Where should I document that behaviour? Is https://docs.bazel.build/versions/main/platforms.html#skipping-incompatible-targets a good place for this?

gregestren:
I think that's good for a start. We could always expend as needed, but no need to overdo documentation at first.

philsc:
I will keep looking into fixing the visibility issue. I filed #16044 and referenced it in the code and in the docs. Let me know if that looks reasonable.

katre (approved):
This looks good to me, I'll wait for Greg's review as well.

gregestren (approved):
LGTM to meet too.

PR #14099, opened 2021-10-12
Update remote-execution-caching-debug.md

ecngtng:
1. When '--execution_log_json_file' is used, we don't need to download code and use execution log parser to convert the logs to 
    text anymore. We could directly use our favourite text differ to diff.
   This option is more convenient than '--execution_log_binary_file'.

2. Add information about 'internal' and update example with that.

### DIFF
--- a/site/docs/remote-execution-caching-debug.md
+++ b/site/docs/remote-execution-caching-debug.md
@@ -23,12 +23,14 @@ from a remote cache is displayed as `remote cache hit`.
 
 For example:
 
-        INFO: 7 processes: 3 remote cache hit, 4 remote.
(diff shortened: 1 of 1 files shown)


ecngtng:
@googlebot I signed it!

ecngtng:
@googlebot I signed it!

ulrfa:
> By using `src/tools/execlog:parser`, the order of the actions are synced. Is the json output ordered in a way that diffing is easy, or does it need to be post-processed?

The json file produced by --execution_log_json_file is sorted, so no post-processing is needed before textual comparison.

PR #14119, opened 2021-10-15
Remove unnecessary cc_binary

yasfatft:


aiuto:
Closing.
- We should not delete the rule, because it illustrates the point that you can depend on generated source code.
- the PR is very stale, and the changed files have moved.

PR #14123, opened 2021-10-17
Retain runfiles when forwarding DelegatingDefaultInfo

### LINKED ISSUE #9442: Propagating `DefaultInfo` loses runfiles
Having a rule return the `DefaultInfo` for a dependency appears to lose runfiles from that dep compared to explicitly creating a new `DefaultInfo` that just copies all of the dependencies data. This seems like a bug.

Context: see discussion in https://github.com/bazelbuild/bazel/issues/9391#issuecomment-532139150, which was worked around in #9398. The following change allowed runfiles for the propagated rule to be handled correctly:

Before:

https://github.com/bazelbuild/bazel/blob/b8e1bb18fae89898b0d28535c6355e67b7d4239e/tools/jdk/java_toolchain_alias.bzl#L52

After:

https://github.com/bazelbuild/bazel/blob/66d099e9760aa49b8d93844bd81e9b8c147225dc/tools/jdk/java_toolchain_alias.bzl#L52-L58

fmeum:
Before this change, returning a DefaultInfo obtained from a native rule target from a Starlark rule would lose the data_runfiles and
default_runfiles. This could only be worked around by creating a DefaultInfo provider instance in Starlark explicitly setting these
fields.

This is fixed by having the DelegatingDefaultInfo return null instead of an empty Runfiles object for the stateless runfiles, which indicates to addSimpleProviders in StarlarkRuleConfiguredTargetUtils that default and data runfiles should be used instead. Previously, this logic would only use the always empty stateless runfiles.

Fixes #9442.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -110,3 +110,4 @@ Andy Scott <andy.g.scott@gmail.com>
 Jamie Snape <jamie.snape@kitware.com>
 Irina Chernushina <ichern@google.com>
 C. Sean Young <csyoung@google.com>
+Fabian Meumertzheim <fabian@meumertzhe.im>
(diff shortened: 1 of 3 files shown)


fmeum:
@comius Could you maybe take a look since you recently touched this code in 811eca3a046a3a3e1936923f98e58f6555753080?

This PR is a one-line fix with test coverage for a long-standing Starlark headscratcher. It is not a regression as your commit just refactored around the bug, but it would be great if this could be resolved in Bazel 5.

Issue #14135, opened 2021-10-19
Unable to build bazel on FreeBSD due to 'error: use of undeclared identifier 'SIGKILL' '


yesudeep:
### Description of the problem / feature request:

```
❯ bazel build //src:bazel
INFO: Analyzed target //src:bazel (5 packages loaded, 310 targets configured).
INFO: Found 1 target...
INFO: From Generating Java (Immutable) proto_library @googleapis//:google_bytestream_bytestream_proto:
google/bytestream/bytestream.proto:19:1: warning: Import google/api/annotations.proto is unused.
INFO: From Generating Java (Immutable) proto_library @googleapis//:google_devtools_build_v1_build_status_proto:
google/devtools/build/v1/build_status.proto:20:1: warning: Import google/api/annotations.proto is unused.
INFO: From Generating Java (Immutable) proto_library @googleapis//:google_api_auth_proto:
google/api/auth.proto:19:1: warning: Import google/api/annotations.proto is unused.
INFO: From Generating Java (Immutable) proto_library @googleapis//:google_devtools_build_v1_build_events_proto:
google/devtools/build/v1/build_events.proto:20:1: warning: Import google/api/annotations.proto is unused.
INFO: From Generating Java (Immutable) proto_library //src/main/protobuf:execution_statistics_proto:
[libprotobuf WARNING external/com_google_protobuf/src/google/protobuf/compiler/java/java_file.cc:247] The optimize_for = LITE_RUNTIME option is no longer supported by protobuf Java code generator and is ignored--protoc will always generate full runtime code for Java. To use Java Lite runtime, users should use the Java Lite plugin instead. See:
  https://github.com/protocolbuffers/protobuf/blob/master/java/lite.md
ERROR: /usr/home/yesudeep/code/yesudeep/bazel/src/main/tools/BUILD:14:11: Compiling src/main/tools/process-tools-linux.cc failed: (Exit 1): clang failed: error executing command /usr/bin/clang -U_FORTIFY_SOURCE '-D_FORTIFY_SOURCE=1' -fstack-protector -Wall -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 28 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
src/main/tools/process-tools-linux.cc:24:14: error: use of undeclared identifier 'SIGKILL'
  kill(-pid, SIGKILL);
             ^
[...]

PR #14136, opened 2021-10-19
fix(#14135): error: use of undeclared identifier SIGKILL on FreeBSD 13.0

yesudeep:
Fixes https://github.com/bazelbuild/bazel/issues/14135 

@philwo


yesudeep:
Hi @philwo Could you please take a look at this?

Issue #14137, opened 2021-10-20
[FreeBSD/OpenBSD] Unable to build bazel on systems where `bash` does not exist at `/bin/bash` 


yesudeep:
### Description of the problem / feature request:

Tools used by Bazel to bootstrap itself and others written in Bash assume a non-portable location for the `bash` executable binary.
The binary is located at different locations on different operating systems. For example:

- [ ] GNU/Linux distributions (generally): `/bin/bash`
- [ ] MacOS: `/bin/bash`
- [ ] FreeBSD: `/usr/local/bin/bash`
- [ ] OpenBSD: `/usr/local/bin/bash`

The use of `#!/bin/bash` in the shebang line for shell scripts is non-portable and may require a lot of selective patching for Bazel to build successfully on multiple operating systems if done on a case by case basis. Fortunately, there appears to be a portable workaround if one is acceptable. Please consider using:

`#!/usr/bin/env bash` instead of `#!/bin/bash` in the shebang line.

### Feature requests: what underlying problem are you trying to solve with this feature?

- [ ] Allow Bazel to build portably on multiple operating systems.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Attempt to build Bazel from source on FreeBSD 13.0 using

```
❯ bazel build //src:bazel
INFO: Analyzed target //src:bazel (5 packages loaded, 310 targets configured).
INFO: Found 1 target...
ERROR: /usr/home/yesudeep/code/yesudeep/bazel/src/main/starlark/builtins_bzl/BUILD:16:8: Executing genrule //src/main/starlark/builtins_bzl:builtins_bzl_zip failed: (Exit 126): bash failed: error executing command /usr/local/bin/bash -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
/usr/local/bin/bash: bazel-out/host/bin/src/zip_builtins: /bin/bash: bad interpreter: No such file or directory
Target //src:bazel failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.776s, Critical Path: 0.01s
INFO: 4 processes: 3 internal, 1 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

### What operating system are you running Bazel on?

```
❯ uname -a
FreeBSD ghostname 13.0-STABLE FreeBSD 13.0-STABLE GENERIC  amd64
```

[...]

PR #14138, opened 2021-10-20
fix(#14137): s/ '#!/bin/bash' / '#!/usr/bin/env bash'

yesudeep:
Fixes: https://github.com/bazelbuild/bazel/issues/14137


ted-xie (changes requested):
This PR is too huge; there are 300+ files here and even loading the diff actually causes my browser to stutter. Please break this up into sub-PRs (maybe alphabetically, with ~10 files changed per PR) if this is still affecting you. Breaking it up into sub-PRs will also help the debug/rollback process if changing the shebangs cause any issues (which should not be the case, but you never know).

Also worth noting that public Google style guides prefer `#!/bin/bash` as the GNU bash shebang. See here: https://chromium.googlesource.com/chromiumos/docs/+/HEAD/styleguide/shell.md#standard-selection

I'm not saying we have to treat the Chromium shell style guide as the golden truth, but there is precedent to use `#!/bin/bash`. I can imagine that keeping shell versions in sync for developer environments is the main reason for this, at least internally. For external-facing projects such as Bazel, I can see some benefit to allowing for the more flexible `#!/usr/bin env bash` style.

ted-xie:
Please also provide a less esoteric commit message. I would imagine most Bazel users are familiar with the sed-style regex syntax, but something straightforward like `Replace all shebangs with #!/usr/bin/env bash` would be a lot easier to read and understand at a glance.

ted-xie:
Some more thoughts on this:

Is there a reason to change _every_ shebang in every shell script? I think only a subset of these scripts are directly executed in a way that the shebang even matters (such as `./a/b/c/d.sh`). Maybe there's something we can change in the job orchestrator that transcends the problem altogether.

aiuto:
I suggest breaking the work along clear boundaries rather than by size or name.

- get bazel to dist-build
- get tests to run
- the long tail of other stuff.

sgowroji:
Hello @yesudeep, Could you please respond to the above [comments](https://github.com/bazelbuild/bazel/pull/14138#issuecomment-1132192936) and resolve the conflicts. Thanks!

src/main/native/windows/build_windows_jni.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
```
aiuto, line 1: I'm not sure this is right for scripts that only run on windows.

aiuto:
From the conversation thread, it is pretty clear we should just close this PR and try again in a more disciplined way.

PR #14147, opened 2021-10-20
desugar tests: Use toolchains rather than the legacy jdk targets.

benjaminp:


katre:
The PR title really worried me. :)

When you send this for review please clarify the scope of the removal.

ahumesky:
Thanks! I can import this now

PR #14158, opened 2021-10-23
Use .dylib extension for macOS dynamic libraries in Unix toolchain

### LINKED ISSUE #11082: C++ shared libraries generated on OSX get file extension .so instead of .dylib

MacOS 10.16

2.2.0

fmeum:
The autoconfigured Unix toolchain previously did not set any artifact_name_patterns, which meant that all artifacts used the default
(Linux) patterns. This was incorrect in the case of dynamic libraries on macOS, which should have the extension `.dylib` rather than `.so`.

Fixes #11082.

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -17,6 +17,7 @@
 load(
     "@bazel_tools//tools/cpp:cc_toolchain_config_lib.bzl",
     "action_config",
+    "artifact_name_pattern",
     "feature",
     "feature_set",
     "flag_group",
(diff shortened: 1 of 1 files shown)


fmeum:
@oquenchil Would you have time to take a look? Since the change is very small, it could potentially be cherry-picked into 5.0.

fmeum:
@oquenchil Friendly ping.

oquenchil (approved):
LGTM

keith:
We should probably add this to the Xcode crosstool as well? https://github.com/bazelbuild/bazel/blob/59e16e944200555da377799aa0d9e8d0674d2e27/tools/osx/crosstool/cc_toolchain_config.bzl#L3010

fmeum:
> We should probably add this to the Xcode crosstool as well?
> 
> https://github.com/bazelbuild/bazel/blob/59e16e944200555da377799aa0d9e8d0674d2e27/tools/osx/crosstool/cc_toolchain_config.bzl#L3010

Submitted https://github.com/bazelbuild/bazel/pull/14354.

PR #14171, opened 2021-10-26
Enable Bzlmod for Bazel

meteorcloudy:
Bazel can now be built with Bzlmod with dependencies fetched from [the Bazel Central Registry](https://github.com/bazelbuild/bazel-central-registry) (https://github.com/bazelbuild/bazel-central-registry/issues/7).

Changes in this PR:
- Added MODULE.bazel file for Bazel
- Added MODULE.bazel file for third_party/remoteapis and third_party/googleapis to make them work as Bazel modules
- Updated gson to 2.8.6, because we build with protobuf 3.19.0 with Bzlmod which needs a newer version of gson.
- Added WORKSPACE.bzlmod, this eventually should be empty, but for now we still need some bind rules to make some jar dependencies available for protobuf. 
- package-bazel.sh: use wildcard to locate the directory of the platforms repo, whose canonical repo name is `platform.<version>` when building with Bzlmod.

To build Bazel with Bzlmod:
- Copy WORKSPACE.bzlmod to WORKSPACE
- Run `USE_BAZEL_VERSION=last_green bazelisk build --experimental_enable_bzlmod //src:bazel_nojdk`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,7 +4,7 @@
 
 module(
   name = "bazel",
-  version = "5.0.0-pre",
+  version = "6.0.0-pre",
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.0.3")
--- a/WORKSPACE.bzlmod
+++ b/WORKSPACE.bzlmod
@@ -1,4 +1,4 @@
-# Copy this file to replace the original WORKSPACE file.
+# Replace the original WORKSPACE file with this file to build Bazel with bzlmod.
 # We still need the bind rules to make protobuf work.
 workspace(name = "io_bazel")
 


aiuto:
> > Then about the *_deps.bzl, would it be fine to add some comment describing a bit what is the meaning?
> 
> @aiuto Since you created the distdir_deps.bzl file, can you help with this?

I think there is a PR floating around to put a warning on distdir_deps.bzl to point out that it is internal to the process of building Bazel and that end users should not rely on it.

> > Indeed I still did not understand why for example Bazel LTS releases have dependencies pointing to specific commit hashes instead of releases.
> 
> I agree we shouldn't depend on random commits, but unfortunately that's the current case, I think we should address this in a separate issue.

Yes. That is a separate issue. Some of the things we build on do not have strong release processes.
This is another reason why distdir_deps should never be seen by end users. Let's continue in #14217.

Issue #13546, opened 2021-05-31
ijar should produce valid class files with code attributes


szeiger:
### Description of the problem / feature request:

ijar strips out all code attributes, thus producing invalid classfiles that cannot be loaded into the JVM. Instead of removing code attributes it should replace them with a dummy implementation that throws an Exception.

### Feature requests: what underlying problem are you trying to solve with this feature?

Our Scala tooling uses Zinc which may load dependent classes from the compiler classpath into the JVM when analyzing classes produced by javac in mixed Scala/Java compilation. This fails because the classfiles produced by ijar do not pass classfile validation:

```
java.lang.ClassFormatError: Absent Code attribute in method that is not native or abstract in class file org/apache/spark/ml/linalg/VectorUDT
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:757)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:419)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:352)
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.privateGetPublicMethods(Class.java:2902)
	at java.lang.Class.getMethods(Class.java:1615)
	at sbt.internal.inc.ClassToAPI$.toDefinitions0(ClassToAPI.scala:164)
	at sbt.internal.inc.ClassToAPI$.$anonfun$toDefinitions$1(ClassToAPI.scala:121)
	at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:82)
	at sbt.internal.inc.ClassToAPI$.toDefinitions(ClassToAPI.scala:121)
	at sbt.internal.inc.ClassToAPI$.$anonfun$process$1(ClassToAPI.scala:28)
[...]

PR #14177, opened 2021-10-27
[ijar] Mark stripped methods as native

gabrielrussoc:
ijar strips out the "Code" attribute from methods which generates invalid class files.

[JVMS 4.7.3](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.3): "If the method is either native or abstract, its method_info structure must not have a Code attribute. Otherwise, its method_info structure must have exactly one Code attribute."

As a hack around this, we can mark all the stripped methods as `native` and generate a valid class file. I couldn't find a case where this causes issues to downstream compilation.

The main motivation behind this is #13546

### DIFF
--- a/third_party/ijar/classfile.cc
+++ b/third_party/ijar/classfile.cc
@@ -82,9 +82,10 @@ enum ACCESS {
   ACC_BRIDGE = 0x0040,
   ACC_VOLATILE = 0x0040,
   ACC_TRANSIENT = 0x0080,
+  ACC_NATIVE = 0x0100,
   ACC_INTERFACE = 0x0200,
   ACC_ABSTRACT = 0x0400,
-  ACC_SYNTHETIC = 0x1000
+  ACC_SYNTHETIC = 0x1000,
 };
 
 // See Table 4.7.20-A in Java 8 JVM Spec.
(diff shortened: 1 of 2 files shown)


comius:
Unlike previous proposals, this change probably wouldn't cause any regressions.
I would prefer though, that the methods are set to abstract instead of native - because this has a closer semantics to "method is not implemented". Than native, that's saying the method is implemented in a library.

I'll defer the review to @cushon, since I'm not the primary expert in ijar.

cushon:
I don't think this change is safe. For example, an annotation processors can inspect the modifiers of an element, and this would incorrectly cause it to see all methods as being `native`.

comius:
Closing because this change is considered unsafe and because there was no response in 20 days.

szeiger:
Annotation processors can see this difference but I think this is purely a theorectical problem. Annotation processors that rely on the official API cannot see a method's implementation so it doesn't make sense to treat native and non-native methods differently. And if an annotation processor does look at the implementation (by casting down to internal compiler APIs) it's already broken because ijar currently produces invalid classfiles without any code attribute.

Marking methods as `abstract` would be wrong. This flag is used by compilers and it changes semantics significantly (unlike `native` which is usually ignored).

cushon:
I agree annotation processors can't use supported APIs to inspect method bodies, but they can use support APIs to check if the `native` modifier is present, and we can't guarantee that there are no annotation processors that look for the `native` modifier and would be affected by ijar adding it to every method it processes.

If making ijar outputs work with classloading is prioritized, I think having ijar produce stub code attributes is a better solution than having it unconditionally mark all methods as `native`.

gabrielrussoc:
for completion, this is indeed unsafe and I found a case where compilation fails with `java.lang.ClassFormatError: Method <init> in class. [...] has illegal modifiers: 0x101`

PR #14187, opened 2021-10-28
Inclusion of mnemonics in profiling

crydell-ericsson:
We have a use case wherein we wish to use the profiling files produced by Bazel to analyze the cumulative use of specific tools. In order to simplify this, this pull request adds information about each Action's mnemonic to the JSON profiles. The additions to the profiling files follow the [Chrome Trace Event Format Specification.](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit) In order to make the processing of profiling files by external programs cleaner and more consistent, mnemonic information is included in two parts of each Action event:

### Mnemonics as Categories

According to the specification, the `"cat"` field of a trace event can specify multiple categories as a comma-separated string. For events that describe Actions, the mnemonic of the Action is included as one of the categories here. 

### Mnemonics as Arguments

The specification also describes the `"args"` field, in which arbitrary data can be placed, and *is* already placed by Bazel with the inclusion of the `"target"` field. For events that describe Actions, a `"mnemonic"` field is added to `"args"`, in which the mnemonic of the action is placed.

### Motivation of Double Inclusion

While it may be seen as superfluous to include the mnemonic twice in the same event object, we believe that doing so is beneficial. The inclusion of the mnemonic as a category is mainly intended for filtering purposes. The trace event format specification describes how "the categories can be used to hide events in the Trace Viewer UI," and it would thereby be reasonable for any other external tools that use the format to implement the same functionality for categories. By including the mnemonic as a category, filtering can be done without taking non-category fields into special regard.

Aside from filtering, it is also of interest to collect cumulative information about Actions grouped by their mnemonics. In this case, the names of the mnemonics may not be initially known during analysis. It may at first seem sufficient to only include the mnemonic as a category for this. With how profiling files are currently constructed in Bazel, the mnemonic of an event can be consistently derived by first making sure it has two categories, then ensuring that one of the categories is `"action processing"`, and then picking the other category. However, this is not future-proof, as the trace event specification would permit the addition of more categories in the future, should Bazel ever be changed to do so. By including an explicit specification of the mnemonic in the `"args"` object, this issue is removed. 

In summary, the double inclusion of the mnemonic is done to ensure that it can consistently be used for filtering, as well as for accessing initially unknown mnemonics from arbitrary Actions.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/profiler/MnemonicData.java
+++ b/src/main/java/com/google/devtools/build/lib/profiler/MnemonicData.java
@@ -1,3 +1,16 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 package com.google.devtools.build.lib.profiler;
 
 /** Representation of mnemonic data used by the {@link Profiler}.
@@ -7,6 +20,7 @@ package com.google.devtools.build.lib.profiler;
  * actions into the types of actions they correspond to more clearly.
  */
 public class MnemonicData {
+  private static MnemonicData emptyMnemonic = null;
   private final String mnemonic;
 
   private MnemonicData() {
@@ -23,10 +37,13 @@ public class MnemonicData {
    * To keep the internal representation encapsulated, {@link MnemonicData#hasBeenSet}
    * can be used to check whether or not a mnemonic has been set for the datum.
    *
-   * @return an empty mnemonic datum
+   * @return a singleton of an empty mnemonic datum
    */
   public static MnemonicData getEmptyMnemonic() {
-    return new MnemonicData();
+    if (emptyMnemonic == null) {
+      emptyMnemonic = new MnemonicData();
+    }
+    return emptyMnemonic;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


crydell-ericsson:
I think that's reasonable, I changed it to only be included in "args".

src/main/java/com/google/devtools/build/lib/profiler/MnemonicData.java
```diff
@@ -0,0 +1,49 @@
+package com.google.devtools.build.lib.profiler;
```
meisterT, line 1: Please include the copyright header, see the other files.

src/main/java/com/google/devtools/build/lib/profiler/MnemonicData.java
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Get a mnemonic datum without any set mnemonic.
+   *
+   * To keep the internal representation encapsulated, {@link MnemonicData#hasBeenSet}
+   * can be used to check whether or not a mnemonic has been set for the datum.
+   *
+   * @return an empty mnemonic datum
+   */
+  public static MnemonicData getEmptyMnemonic() {
+    return new MnemonicData();
```
meisterT, line 29: please make this a singleton

src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
```diff
@@ -138,27 +138,47 @@ public ProfilerTask getType() {
...
       this.duration = duration;
       this.description = description;
     }
 
+    /**
+     * Get the value of the datum's "cat" field for JSON output.
+     *
+     * @return the category value if one can be created, otherwise null
+     */
+    public String getJsonCategory() {
```
meisterT, line 174: I don't think that function carries its own weight.

crydell-ericsson: While it doesn't do much, I left this in after reverting the use of the "cat" field since I think it's a bit clearer to use this than to access `data.type.description` directly. I do admit that it's a bit out of place since more clarity could be brought with a larger refactoring, which would be a separate issue from this adding of mnemonics, so I can revert this as well if you still think it's apt to do so.

meisterT (approved):

Issue #10252, opened 2019-11-16
Per-target exec_properties overwrites (not merges with) --remote_default_exec_properties


jmillikin-stripe:
### Description of the problem / feature request:

`exec_properties` is a universal target attribute documented as:
> A dictionary of strings that will be added to the `exec_properties` of a platform selected for this target. 

This doesn't seem to be working as expected when combined with `--remote_default_exec_properties`. Instead of merging, the default exec properties get completely ignored.

The build is not using a `platform()` rule (the `container-image` name is too dynamic to easily put in a source file), so I would have expected `--remote_default_exec_properties` to be used.

### What operating system are you running Bazel on?

Linux

### What's the output of `bazel info release`?

release 1.0.0

brentleyjones:
This recently bit us as well.

PR #14193, opened 2021-10-29
Remote: Merge target-level exec_properties with --remote_default_exec_properties

coeuvre:
Per-target `exec_properties` was introduced by 0dc53a2217f32da737410883158d42c41b6d1d61 but it didn't take into account of `--remote_default_exec_properties` which provides default exec properties for the execution platform if it doesn't already set with `exec_properties`. So the per-target `exec_properties` overrides `--remote_default_exec_properties` instead of merging with them which is wrong.

Fixes https://github.com/bazelbuild/bazel/issues/10252.


brentleyjones:
Can this get cherry-picked into the 5.0 RC? 🙏 Thanks!

coeuvre:
Sure!

PR #14197, opened 2021-10-29
Fix the analysis tests

andrewkatson:
I changed from comparing objects to comparing the paths the objects correspond to. The tests are looking for ordering based on the path ordering not the object ordering. It is unclear why objects compare differently on linux and windows.

### DIFF
--- a/.bazelci/postsubmit.yml
+++ b/.bazelci/postsubmit.yml
@@ -240,7 +240,6 @@ tasks:
       - "-//src/test/java/com/google/devtools/build/android/desugar/scan/..."
       - "-//src/test/java/com/google/devtools/build/android/desugar/typeannotation/..."
       - "-//src/test/java/com/google/devtools/build/android/r8/..."
-      - "-//src/test/java/com/google/devtools/build/lib/analysis/..."
       - "-//src/test/java/com/google/devtools/build/lib/query2/cquery/..."
       - "-//src/test/java/com/google/devtools/build/lib/query2/engine/..."
       - "-//src/test/java/com/google/devtools/build/lib/versioning/..."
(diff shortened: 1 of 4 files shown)


meteorcloudy:
Thanks! Can you also enable the corresponding test in the BUILD file by removing the "no_windows" tag?

andrewkatson:
There doesnt seem to be a no_windows tag anywhere in the BUILD file.

meteorcloudy:
Oh, right, it's in the presubmit.yml and postsubmit.yml file.

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -54,7 +54,7 @@
   private static final String GUID = "07459553-a3d0-4d37-9d78-18ed942470f4";
 
   private static final Comparator<Map.Entry<PathFragment, Artifact>> ENTRY_COMPARATOR =
-      (path1, path2) -> path1.getKey().compareTo(path2.getKey());
+      (path1, path2) -> path1.getKey().getPathString().compareTo(path2.getKey().getPathString());
```
meteorcloudy, line 57: Interesting, can you elaborate in the PR description why this fixes the test on Windows?

.bazelci/postsubmit.yml
```diff
@@ -240,7 +240,6 @@ tasks:
       - "-//src/test/java/com/google/devtools/build/android/desugar/scan/..."
       - "-//src/test/java/com/google/devtools/build/android/desugar/typeannotation/..."
       - "-//src/test/java/com/google/devtools/build/android/r8/..."
-      - "-//src/test/java/com/google/devtools/build/lib/analysis/..."
```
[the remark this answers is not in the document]
andrewkatson: Done

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
andrewkatson: Basically we are comparing objects but the tests are comparing path ordering.  Ill add that to the description.

meteorcloudy: OK, this case we actually need it to be case-sensitive to get the correct order, then this change makes sense.

meteorcloudy (approved):

Issue #13649, opened 2021-07-06
Databinding: Build failure when enabling `--persistent_android_resource_processor`


arunkumar9t2:
### Description of the problem / feature request:

When databinding is used, specifically if layout xml has databinding expressions, persistent workers for Android resource work cannot be enabled. Enabling it causes the build to fail with the following message.

```bash
Merging compiled Android resources for //:A failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log, file at /private/var/tmp/_bazel_arun.sampathkumar/b612b0579661b9b8cf8dbfbec098862c/bazel-workers/worker-18-AndroidCompiledResourceMerger.log ---8<---8<---
Exception in thread "main" com.google.common.base.VerifyException: layout/activity_main.xml-0 is not a valid resource name. Expected ^((?<package>[^:]+):)?(?<type>\w+)/(?<name>[A-Za-z0-9_.$]+)$
        at com.google.common.base.Verify.verify(Verify.java:424)
        at com.google.devtools.build.android.ResourceName.parse(ResourceName.java:54)
        at com.google.devtools.build.android.AndroidCompiledDataDeserializer.consumeCompiledFile(AndroidCompiledDataDeserializer.java:530)
        at com.google.devtools.build.android.AndroidCompiledDataDeserializer.read(AndroidCompiledDataDeserializer.java:675)
        at com.google.devtools.build.android.SerializedAndroidData.deserialize(SerializedAndroidData.java:108)
        at com.google.devtools.build.android.AndroidResourceMerger.mergeCompiledData(AndroidResourceMerger.java:258)
        at com.google.devtools.build.android.AndroidCompiledResourceMergingAction.main(AndroidCompiledResourceMergingAction.java:231)
        at com.google.devtools.build.android.ResourceProcessorBusyBox$Tool$3.call(ResourceProcessorBusyBox.java:81)
        at com.google.devtools.build.android.ResourceProcessorBusyBox.processRequest(ResourceProcessorBusyBox.java:233)
        at com.google.devtools.build.android.ResourceProcessorBusyBox.runPersistentWorker(ResourceProcessorBusyBox.java:198)
        at com.google.devtools.build.android.ResourceProcessorBusyBox.main(ResourceProcessorBusyBox.java:175)
---8<---8<--- End of log ---8<---8<---
Target //:app failed to build
```

### Feature requests: what underlying problem are you trying to solve with this feature?

Ability to enable workers for android resource work for performance.

[...]

PR #14198, opened 2021-10-29
Prevent temporary intermediate files from being picked up by Android ResourceCompiler when building with workers.

arunkumar9t2:
When `--persistent_android_resource_processor` is used alongside databinding, `ResourceCompiler` reads `databinding-processed-resources` folder during resource merging and that folder can contain `*.params` file used for the `PROCESS_DATABINDING` action. This causes resource compiler's file name validation to fail. This CL simply ignores staging the intermediate files for resource compilation.

Additionally this CL also excludes `*.tmp` files which were observed when using dynamic remote execution.

Fixes https://github.com/bazelbuild/bazel/issues/13649


ahumesky:
Thanks, I'll start importing this now

PR #14201, opened 2021-11-01
Add support for wrapping system streams in WorkRequestHandler

Bencodes:
There are often [places](https://github.com/bazelbuild/bazel/blob/ea19c17075478092eb77580e6d3825d480126d3a/src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java#L188) where persistent workers need to swap out the standard system streams to avoid tools poisoning the worker communication streams by writing logs/exceptions to it.

This pull request extracts that pattern into an optional WorkerIO wrapper can be used to swap in and out the standard streams without the added boilerplate.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
@@ -18,11 +18,15 @@ import com.google.devtools.build.lib.worker.WorkerProtocol.WorkRequest;
 import com.google.devtools.build.lib.worker.WorkerProtocol.WorkResponse;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.sun.management.OperatingSystemMXBean;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.management.ManagementFactory;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
@@ -317,8 +321,16 @@ public class WorkRequestHandler implements AutoCloseable {
    * then writing the corresponding {@link WorkResponse} to {@code out}. If there is an error
    * reading or writing the requests or responses, it writes an error message on {@code err} and
    * returns. If {@code in} reaches EOF, it also returns.
+   *
+   * <p>This function also wraps the system streams in a {@link WorkerIO} instance that prevents
+   * calls to {@link System#out} and {@link System#err} from corruption the worker streams. When the
+   * while loop exits, the original system streams will be swapped back into {@link System}.
    */
   public void processRequests() throws IOException {
+    // Wrap the system streams into a WorkerIO instance that can be used
+    // to capture output and error streams that aren't being written to the worker output directly.
+    WorkerIO workerIO = WorkerIO.capture();
+
     try {
       while (!shutdownWorker.get()) {
         WorkRequest request = messageProcessor.readWorkRequest();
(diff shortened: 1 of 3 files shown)


Bencodes:
@larsrc-google mind taking a look at this?

src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
```diff
@@ -496,4 +500,100 @@ private void maybePerformGc() {
...
+   * {@link System#out} and {@link System#err} which corrupt the persistent worker.
+   *
+   * <p>WorkerIO implements {@link AutoCloseable} and will swap the original streams back into {@link
+   * System} once close has been called.
+   */
+  public static class WorkerIO implements AutoCloseable {
+
+    private final InputStream originalInputStream;
+    private final PrintStream originalOutputStream;
+    private final PrintStream originalErrorStream;
```
Bencodes, line 519: Wrapping `System.err` isn't actually be necessary for a wrapper worker io.

larsrc-google:
Sorry for not reacting earlier, need to update mail filters.

This is a good thing to do, something I have considered but not gotten around to yet. One question: Why an optional wrapper? Why not just make it part of WorkRequestHandler? Can you imagine a case where a user of WorkRequestHandler would not want to redirect stdout?

Bencodes:
> Why an optional wrapper? Why not just make it part of WorkRequestHandler?

No reason other than it felt easier to reason about merging this into Bazel as an isolated tool first, instead of making changes to `WorkRequestHandler` directly.

> Can you imagine a case where a user of WorkRequestHandler would not want to redirect stdout?

I can't, although the java workers seem to be managing fine without wrapped system streams.

src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
larsrc-google: Correct. Also `System.in` is already handled properly. Only `System.out` needs to be wrapped.

Bencodes: > Also System.in is already handled properly. Only System.out needs to be wrapped.

The motivation behind wrapping all of the system streams in a `WorkerIO` class is that there is a clear and consistent place to access the root system streams while setting up the work request handler. For `System.in` and `System.err` we can just return the original streams when calling `getOriginalInputStream`.

Bencodes:
@larsrc-google is there a path forward for merging some form of this?

src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
Bencodes: > wrapping the inside of the WorkRequestCallback would be clearer

By this do you mean calling `capture` from within the implementation of `WorkRequestHandler`, or in the lambda that gets passed into the `WorkRequestHandler`? I like idea of this being hidden inside the `WorkRequestCallback` because we hide `WorkerIO#readCapturedAsUtf8String` entirely which is kind of an awkward API to need to expose.

> That change would also mean that the WorkerIO doesn't need to reset the 

Would each work request call `WorkerIO#capture` swapping in it's own buffer?

> A definite benefit of either approach is that stdout and stderr get merged into a single stream, that will be easier to understand.

Wrapping the `stderr` could be a bit error-prone because if there's ever an unrecoverable error that causes the worker to not send a work response, you just won't see an exception come out of the worker.

larsrc-google: > By this do you mean calling capture from within the implementation of WorkRequestHandler, or in the lambda that gets passed into the WorkRequestHandler?

From within the implementation, in `respondToRequest`. Putting it in there means the stdio redirection does not touch on the use of stdio for the server-worker protocol. 

> Would each work request call WorkerIO#capture swapping in it's own buffer?

Yes. Throw-away object is simplest object.

> Wrapping the stderr could be a bit error-prone because if there's ever an unrecoverable error that causes the worker to not send a work response, you just won't see an exception come out of the worker.

I believe that's already the case today, but I'll have to consider the various possible failures some more before I can opine on it.

How does your current solution deal with multiplex workers?

Bencodes: > How does your current solution deal with multiplex workers?

@larsrc-google I took another pass at building this into the `WorkRequestHandler` so that it's a bit more transparent, and with the multiplex worker cases in mind. The new approach uses the existing `WorkerIO` implementation, but wraps the system streams when you call `WorkRequestHandler#processRequests`. 

The `WorkerIO` instance gets passed down into `respondToRequest` so that there is a single source of truth for the wrapped streams, and so that we don't have to deal with multiple `WorkerIO` instances being created inside of `respondToRequest` which doesn't really make sense in a multiplex-world where `respondToRequest` may get called concurrently.

larsrc-google: Oh, I've been so focused on the implementation details that I forgot the larger problem: Redirecting the stderr messages into the `output` field of the response changes semantics. The `output` field is for shorter messages to be displayed to the user, `stderr` goes into the logs. See https://docs.bazel.build/versions/main/creating-workers.html#work-responses. It's known that multiplex workers only have one log between them, but the messages shown to the user should not interleave.

And of course, since the multiplex threads still share one `System.err` instance regardless of what we set, we can't possibly disentangle that. But we can still fix the main issue of stdout getting corrupted by setting it to the same stream as stderr.

Bencodes: Just to make sure i'm understanding next steps here - You'd prefer that we avoid capturing the tool output streams and instead just swap the `System.err` in for `System.out` to not pollute the work response output?

larsrc-google: Yes. Having System.err and System.out go to the same place (the logs) is good. Changing the work response output would cause a lot of noise.

Bencodes: Pushed the changes for swapping `System.err` in for `System.out` and updated the tests.

src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
```diff
@@ -496,4 +508,70 @@ private void maybePerformGc() {
...
+     * <p>{@link WorkerIO#close} must be called to swap the original streams back in.
+     */
+    public static WorkerIO redirectSystemStreams() throws IOException {
+      // Save the original streams
+      PrintStream originalOutputStream = System.out;
+      PrintStream originalErrorStream = System.err;
+
+      // Redirect System.out to System.err
+      System.setOut(originalErrorStream);
+      System.setErr(originalErrorStream);
```
larsrc-google, line 549: This looks a lot simpler. I don't see any reason to do anything to System.err, though.

Bencodes: Updated

src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java
```diff
@@ -311,6 +315,8 @@ public WorkRequestHandler build() {
    * returns. If {@code in} reaches EOF, it also returns.
    */
   public void processRequests() throws IOException {
+    WorkerIO workerIO = WorkerIO.redirectSystemStreams();
```
larsrc-google, line 318: Looking at https://github.com/bazelbuild/bazel/pull/14623, you're not the only one with this idea. You might be right in making it optional, as some workers do their own handling. Fortunately that could easily be done by adding a method to the builder.

nkoroste:
Any updates on this?

larsrc-google:
Sorry for the silence, I was out sick for quite a while. I believe being optional is not as big a deal as I indicated, so if you can fix the conflicts, we should be able to merge this.

Bencodes:
@larsrc-google I have this a rebase. Please give it another once over to make sure that the conflicts got resolved correctly against the new changes that were in master.

brentleyjones:
@bazel-io flag

brentleyjones:
(Above was from a discussion with @larsrc-google.)

kshyanashree:
@bazel-io fork 6.1.0

Issue #14178, opened 2021-10-27
Bazel 4.2.1 doesn't respect --remote_max_connections


Vertexwahn:
### Description of the problem:

There is a problem with Bazel 4.2.1. For some reason, it does not respect the [--remote_max_connections flag](https://docs.bazel.build/versions/4.2.1/command-line-reference.html#flag--remote_max_connections). Builds that run with this version of Bazel open up thousands of sockets when they should in fact be restricted to 100.

We do not see this behavior with Bazel 4.0.0.

When using socketstat (we do here always a build with a remote cache - the system has around 240 sockets open before Bazel starts) we see a large increase of max sockets:

| RUN_ID |  Bazel version |  Max Sockets |
| -------- | ------------| ------------- |
| 1           | 4.0.0         |     341           |
| 2           | 4.2.1         |    1241          |
| 3           | 4.0.0         |     332          |
| 4           | 4.2.1         |    1804          |
| 5           | 4.0.0         |     360           |
| 6           | 4.2.1         |    1813          |
| 7           | 4.2.1         |    2643          |

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

TODO (we work on it)

### What operating system are you running Bazel on?

Ubuntu 20.04 LTS

### What's the output of `bazel info release`?

`release 4.0.0` or `release 4.2.1`

###  Have you found anything relevant by searching the web?

Nope.

### Further notes

We are using Buildbarn

brentleyjones:
@Wyverald Can this get the release blocker milestone instead of https://github.com/bazelbuild/bazel/pull/14202, and can it be reopened until the cherry-pick is in?

Wyverald:
Cherrypicked \o/

PR #14202, opened 2021-11-01
Remote: Limit max number of gRPC connections by --remote_max_connections.

coeuvre:
`--remote_max_connections` is only applied to HTTP remote cache. This PR makes it apply to gRPC cache/executor as well.

Note that `--remote_max_connections` limits the number of concurrent connections. For HTTP remote cache, one connection could handle one request at one time. For gRPC remote cache/executor, one connection could handle 100+ concurrent requests. So the default value `100` means we could make up to `100` concurrent requests for HTTP remote cache or `10000+` concurrent requests for gRPC remote cache/executor.

Fixes: #14178.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -64,8 +64,7 @@ public final class RemoteOptions extends OptionsBase {
       effectTags = {OptionEffectTag.HOST_MACHINE_RESOURCE_OPTIMIZATIONS},
       help =
           "Limit the max number of concurrent requests to remote cache/executor. By default the"
-              + " value is 100. Setting this to 0 will make Bazel choose the number of connections"
-              + " automatically.\n"
+              + " value is 100. Setting this to 0 means no limitation.\n"
               + "For HTTP remote cache, one TCP connection could handle one request at one time, so"
               + " it could open up to --remote_max_connections TCP connections.\n"
               + "If --incompatible_remote_max_connections_grpc is set, this also applies to the"
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -68,7 +68,7 @@ public final class RemoteOptions extends OptionsBase {
               + "For HTTP remote cache, one TCP connection could handle one request at one time, so"
               + " it could open up to --remote_max_connections TCP connections.\n"
               + "If --incompatible_remote_max_connections_grpc is set, this also applies to the"
-              + " gRPC remote cache/executor, for which, one TCP connection could usually handle"
+              + " gRPC remote cache/executor, for which, one gRPC channel could usually handle"
               + " 100+ concurrent requests, so it could open up to `--remote_max_connections / 100`"
               + " gRPC channels. Each gRPC channel could open 1 or more TCP connections depending"
               + " on the number of resolved server IPs.")
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -63,11 +63,16 @@ public final class RemoteOptions extends OptionsBase {
       documentationCategory = OptionDocumentationCategory.REMOTE,
       effectTags = {OptionEffectTag.HOST_MACHINE_RESOURCE_OPTIMIZATIONS},
       help =
-          "The max. number of concurrent network requests to the HTTP remote cache. By default the"
+          "Limit the max number of concurrent requests to remote cache/executor. By default the"
               + " value is 100. Setting this to 0 will make Bazel choose the number of connections"
               + " automatically.\n"
-              + "If --incompatible_remote_max_connections_grpc is set, this also applies"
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -59,13 +59,13 @@
 
   @Option(
       name = "remote_max_connections",
-      defaultValue = "100",
+      defaultValue = "0",
```
brentleyjones, line 62: Isn't 0 for HTTP the same as no limit?

brentleyjones:
Btw, this would be nice to get into the 5.0 RC if possible.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
coeuvre: Yes. I am actually concerned about changing to 0.

The number of HTTP connections has been limited to 100 by default. By changing to 0, it could grow rapidly for a large build.

On the other hand, the number of gRPC channels was 1 (<`3.7.0`), then changed to based on `--jobs` (< `4.1.0`) and finally changed to dynamic connection pool (>= `4.1.0`) which doesn't has an upper-bound. Note that one gRPC channel could usually handle 100+ concurrent requests. So leaving `--remote_max_connections` to 100 by default means we can handle 10000+ max concurrent requests by default which looks find to me. By building Bazel itself with this change, I didn't see the number of gRPC channels reaching the max.

WDTY if we keep the 100 as default?

coeuvre:
> Btw, this would be nice to get into the 5.0 RC if possible.

Sure, I can get this into 5.0.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
brentleyjones: Since projects ran into issues before this flag was added, which they would possibly encounter again if this became `0`, I say we need to leave it at `100`.

I would adjust the help text to indicate how it affects HTTP versus gRPC, especially since people might not know it now affects the gRPC side. I would be surprised that if this is `100` it would allow "10,000 concurrent requests", as I'm telling it "100 max connections". It would be nice if it spelled out that "connection" == "gRPC channel" as well.

coeuvre: Thanks for the input! SGTM.

If we keep the default of `--remote_max_connections` to `100` and make it to the number of max concurrent requests on the fly for gRPC, the number of gRPC channels is limited to `1` by default which could cause problems for some projects. I need to guard this feature with an incompatible flag which could be flipped and removed in `6.x`.

coeuvre: Just want to emphasize here: for each gRPC channel, it could create one or more HTTP/2 connections depending on resolved IP. For each connection, it could handle `MAX_CONCURRENT_STREAMS` (which is a per-server settings and is usually set to `100` or `128`) concurrent requests. (See more details in https://github.com/bazelbuild/bazel/issues/11801)

AFAIK, there isn't an API to check the max concurrent requests a channel could handle, so we assume the number is `100`. Then the max number of gRPC channels we can create is calculated by `--remote_max_connections / 100`.

brentleyjones: I guess I wonder if `--remote_max_connections` _should_ be "max concurrent requests on the fly for gRPC". https://github.com/bazelbuild/bazel/issues/14178's concern is about the number of sockets. I assume each concurrent stream of a gRPC channel opens a socket? If so, then yes, this change (making it `/ 100` for gRPC) makes sense, otherwise I think it should equal the maximum number of channels.

coeuvre: > I assume each concurrent stream of a gRPC channel opens a socket? 

Nope.

To make it easier to understand, assuming the server name is resolved to one IP address for the following discussion.

One gRPC channel opens one TCP connection. It can use this one TCP connection to fire off 100+ requests at one time by leveraging HTTP/2 multiplexing. So with the latest commit, if `--remote_max_connections` is set to `100`, Bazel will create only one gRPC channel which opens only one TCP connection but can make 100 concurrent requests.

The semantic of `--remote_max_connections` was to limit max connections for HTTP remote cache. Since we only send one request at one time per HTTP connection, this implies to limiting the max concurrent requests. 

However, since gRPC can more requests at one time per connection, this becomes ambiguous.

1. If we want `--remote_max_connections` limit the max number of TCP connections (or sockets), Bazel could make up to `--remote_max_connections * 100` concurrent requests to the server.
2. If we want `--remote_max_connections` limit the max number of concurrent requests, Bazel could open up to  `--remote_max_connections / 100` TCP connections (or sockets) to the server.

Both seem reasonable to me, but in terms of implementation and backward compatibility, I prefer 1 since we don't need to introduce an incompatible flag (the default value is high enough to consider unlimited for gRPC).

WDTY in your use case?

brentleyjones: I agree, it should be 1, as "connections" == "sockets" in my mind, not requests.

Wyverald:
What's the status of this?

coeuvre:
The PR is ready. I will import and send out for review.

PR #14211, opened 2021-11-02
Typo fix in build-ref.html

amaslak0v:


amaslak0v:
@googlebot I signed it!

brentleyjones:
The isn't actually a typo. It's correct that the `WORKSPACE` forms another workspace.

oquenchil:
I agree with Brentley. This isn't a typo.

Issue #14217, opened 2021-11-02
Encourage users to use release versions for dependencies used by Bazel


limdor:
### Description of the problem / feature request:

Ticket to continue the discussion started in https://github.com/bazelbuild/bazel/pull/14171#issuecomment-953813615

In our company we had several times the situation that we want to update Bazel and we need to update also the different Bazel rules. In order to decide which ones to use there two approaches:
* You can take the same version of the dependencies that Bazel is using for the specific release that you want to use. For example in Bazel 4.2.1 you could take the same version of the dependencies like specified in https://github.com/bazelbuild/bazel/blob/02ad3e3bc6970db11fe80f966da5707a6c389fdd/WORKSPACE#L147
* You can take the latest release version of the rules that you want to update
Regarding the first option I do not think it is a good practice, Bazel generally depends on specific commit hashes and not in releases. I would assume that releases are more tested than specific hashes (correct me if I'm wrong).

Note that in addition there are two additional approaches that currently are not possible but were or will be possible:
* In the past, when Bazel federation was alive, the recommended one by Bazel federation was taken
* In the future, once BazelMod comes into play, you could take the latest version specified in BazelMod

### Feature requests: what underlying problem are you trying to solve with this feature?

My request does not involve any code change, just a comment in https://github.com/bazelbuild/bazel/blob/master/distdir_deps.bzl#L14
The idea would be to add a comment in this file saying that people do not have to use this specific versions and that they are encourage to use the latest releases for the dependencies. The exact text can be discussed but the idea would be to discourage the use of specific commits and encourage the use of releases for Bazel rules.

### What's the output of `bazel info release`?

4.2.1

###  Have you found anything relevant by searching the web?

This is the ticket created after the discussion in https://github.com/bazelbuild/bazel/pull/14171#issuecomment-953813615

limdor:
@meteorcloudy @aiuto here the ticket. If you agree I could start drafting the PR.

limdor:
@meteorcloudy I created a PR. If you could review it or point to the person that could it that would be great

aiuto:
I think the comment should be simply that end users should never look at distdir_deps.bzl. It is internal source code and independent of the user experience.  The advice about specific versions is only going to get us in to trouble because
- we don't follow that advice
- sometimes it is impossible for users to follow the advice

limdor:
[...]

PR #14223, opened 2021-11-03
Add clarification note in bazel dependency file

limdor:
As discussed in #14217 the list of dependencies here if it is seen
by the users could be seen like a blessed list of versions.
Adding this message makes it clear that it is not the case.

### DIFF
--- a/distdir_deps.bzl
+++ b/distdir_deps.bzl
@@ -11,7 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""List the distribution dependencies we need to build Bazel."""
+"""
+List the distribution dependencies we need to build Bazel.
+
+Note for Bazel users: This list is not intended to tell you what version
+you should use for each dependency. Indeed in this list you will find that
+some dependencies do not use official releases but specific commits.
+This is something that in general we discourage but in Bazel itself sometimes
+it is necessary.
+Our recomendation is always to use the latest version of the dependency that
+has been released and it works for the Bazel version that you are using.
+"""
 
 DIST_DEPS = {
     ########################################


distdir_deps.bzl
```diff
@@ -11,7 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""List the distribution dependencies we need to build Bazel."""
+"""
+List the distribution dependencies we need to build Bazel.
+
+Note for Bazel users: This list is not intended to tell you what version
+you should use for each dependency. Indeed in this list you will find that
```
meteorcloudy, line 18: I don't quite understand this. This is only for Bazel itself, not a recommendation for users, right?

limdor: yes, the idea would be that if a user comes here to this code to try to figure out what version to use. This should kick the user away and say that this is not the place to look. The question is how to phrase it.
I already found several users pointing me to this file to justify the usage of some specific version of a dependency

meteorcloudy: Oh, sorry. I missed the "not" in "This list is is not intended"..

meteorcloudy (approved):

distdir_deps.bzl
limdor: Maybe I should have put a **NOT** ;-)

aiuto (commented):
Honestly. I don't see the point.
The bazel sources are not bazel.  This PR is for a few people who build bazel from source and look at it. They should know better.
For normal users, this is irrelevant because they never look at the source.
This kind of guidance should be in a general document describing how we build and test bazel, rather than in each file.

limdor:
> Honestly. I don't see the point. The bazel sources are not bazel. This PR is for a few people who build bazel from source and look at it. They should know better. For normal users, this is irrelevant because they never look at the source. This kind of guidance should be in a general document describing how we build and test bazel, rather than in each file.

I understand that this can be confusing but just to make it clear the list of dependencies here is not affecting only the people that compile Bazel from source but also the Bazel users. Some of these dependencies are leaking and in the end they are exposed to the user. In the direction of reproducibility if you want to freeze these dependencies and not depend on the version that your direct dependency is using, you might want to know what is currently used and here is the only place that can be checked.
One case for us for example is rules_cc, even we are not using rules_cc anywhere, we still need to download it because Bazel (even compiled) is using it and trying to fetch it. If we want to have a good overview of the version what we do is to first of all choose the rules_cc version that we want and then Bazel will not pull the version specified in this file.

Anyway I understand that this comment can be then confusing for the Bazel developers that go to this file. For me just the discussion stating that users should not look here is already very valuable. Thanks a lot for that.

I will close the PR

PR #14225, opened 2021-11-04
Bzlmod: add workspace_name attribute to module directive

meteorcloudy:
Modules can specify the workspace_name attribute to allow accessing its
own targets under the repo name.

This helps make migration easier when the preferred workspace name is
different from the module name of the project.

Context: https://github.com/bazelbuild/bazel-central-registry/pull/28#issuecomment-954741081


meteorcloudy:
This PR may not be needed in favour of https://github.com/protocolbuffers/protobuf/pull/9187

PR #14230, opened 2021-11-05
Add support for rendering HTML in markdown table cells in skydoc

cvarier:
Fixes https://github.com/bazelbuild/stardoc/issues/78.

There are multiple use cases in which we would want to support the ability to render HTML in table cells.

As described in the issue above, rendering proper lists is not possible without providing a way to avoid escaping the HTML tags in the docstring. This solution provides the ability to toggle the escaping of HTML in `markdownCellFormat`, without affecting existing usages of this method (particularly those in custom Velocity templates living in separate repos).

### DIFF
--- a/src/main/java/com/google/devtools/build/skydoc/rendering/MarkdownUtil.java
+++ b/src/main/java/com/google/devtools/build/skydoc/rendering/MarkdownUtil.java
@@ -35,21 +35,47 @@ public final class MarkdownUtil {
   private static final int MAX_LINE_LENGTH = 100;
 
   /**
-   * Return a string that formats the input string so it is displayable in a markdown table cell.
-   * This performs the following operations:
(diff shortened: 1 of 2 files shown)


Cvarier:
@googlebot I signed it!

brentleyjones:
@sgowroji Can you ping someone to get this reviewed?

brandjon:
tetromino@ will be back in office next month. When he returns we can discuss the broader design question of how/whether we should handle more advanced markup in docstrings. I'll leave this PR open until then.

tetromino:
Thanks for your contribution! Unfortunately, the decision of which content to escape in markdown table cells is a fair bit more complex than this PR suggests, because not all markdown markup is permitted in a table cell. A more complete fix was implemented in https://github.com/bazelbuild/stardoc/pull/167.

PR #14231, opened 2021-11-05
Remove DigestUtils.getDigestInExclusiveMode() now that SsdModule has …

### LINKED ISSUE #14034: Drop experimental_multi_threaded_digest
This options defaults to true since a while and should just be removed, likely with the whole `SsdModule`.

bjacklyn:
…been removed

There used to be a cli flag --experimental_multi_threaded_digest which defaulted to true
and therefore would not compute the file digest in exclusive mode unless overridden to false.
When that flag was removed this code was accidentally left behind which instead causes bazel to always
compute file digests in exclusive mode for all files larger than MULTI_THREADED_DIGEST_MAX_FILE_SIZE.

This causes bazel to take 5-6x longer to 'check cached actions' in my org's builds,
specifically increasing the time from 30 secs to 2 mins 30 secs for fully cached no-op builds.

Fixes #14034

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/vfs/DigestUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/vfs/DigestUtils.java
@@ -41,16 +41,8 @@ import java.io.IOException;
  * fail.
  */
 public class DigestUtils {
-  // Object to synchronize on when serializing large file reads.
-  private static final Object DIGEST_LOCK = new Object();
-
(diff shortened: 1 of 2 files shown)


bjacklyn:
cc @meisterT @Wyverald

Wyverald (approved):
Thanks for the fix! I'll merge this shortly.

PR #14233, opened 2021-11-05
Remote: Update FilesystemValueChecker to handle the case that a file was remote but is now staged.

### LINKED ISSUE #13912: When using `--remote_download_toplevel`, some remote inputs are staged (downloaded) for local execution, even when they already exist on disk

When using `--remote_download_toplevel`, and only some of the outputs of another action are consumed as inputs to an action that is run locally, those inputs are staged (downloaded) each time the local action is executed, instead of reusing the already downloaded outputs, if not all of the outputs of the producing action have been downloaded.

`example_rules.bzl`:
```starlark
ProducerInfo = provider(
    fields = {
        "a": "a",
        "b": "b",
    },
)

def _producer_impl(ctx):
    a = ctx.actions.declare_file("{}/a.txt".format(ctx.attr.name))
    b = ctx.actions.declare_file("{}/b.txt".format(ctx.attr.name))

    ctx.actions.run_shell(
        command = """
head -c 100 </dev/urandom > $1
head -c 100 </dev/urandom > $2
""",
        arguments = [
            a.path,
            b.path,
        ],
        outputs = [a, b],
        mnemonic = "Producer",
    )

    return [
        ProducerInfo(
            a = depset([a]),
            b = depset([b]),
        ),
        OutputGroupInfo(
            a = depset([a]),
            b = depset([b]),
        ),
    ]

producer = rule(
    implementation = _producer_impl,
)

def _consumer_impl(ctx):
    b = ctx.attr.producer[ProducerInfo].b
    output = ctx.actions.declare_file("{}/out.txt".format(ctx.attr.name))

    args = ctx.actions.args()
    args.add_all(b)

    ctx.actions.run_shell(
        command = """
readonly output="$1"
shift
cat $@ > "$output"
""",
        arguments = [
            output.path,
            args,
        ],
        inputs = depset([ctx.info_file], transitive = [b]),
        mnemonic = "Consumer",
        outputs = [output],
        execution_requirements = {
            "no-cache": "1",
            "no-remote": "1",
        },
    )

    return [
        DefaultInfo(
            files = depset([output]),
        )
    ]

consumer = rule(
    implementation = _consumer_impl,
    attrs = {
        "producer": attr.label(providers = [[ProducerInfo]]),
    },
)
```

`BUILD`:
```starlark
load("@//:example_rules.bzl", "producer", "consumer")

producer(
    name = "producer",
)

consumer(
    name = "consumer",
    producer = ":producer",
)
```

```shell
# Run it once to prime the cache
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel -- //:consumer
INFO: Invocation ID: 63bbdc7b-b5f8-4895-96dd-3ff99f1b287e
INFO: Analyzed target //:consumer (4 packages loaded, 8 targets configured).
INFO: Found 1 target...
Target //:consumer up-to-date:
  bazel-bin/consumer/out.txt
INFO: Elapsed time: 0.183s, Critical Path: 0.05s
INFO: 3 processes: 1 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 3 total actions

# Clean to force cache to be used
❯❯❯ bazel clean
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel -- //:consumer
INFO: Invocation ID: 0c1ed250-40f2-4d55-ad76-5c59d8ebb7be
INFO: Analyzed target //:consumer (4 packages loaded, 8 targets configured).
INFO: Found 1 target...
Target //:consumer up-to-date:
  bazel-bin/consumer/out.txt
INFO: Elapsed time: 0.143s, Critical Path: 0.02s
INFO: 3 processes: 1 remote cache hit, 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 3 total actions

# Subsequent runs re-download (see the "1 remote cache hit")
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel -- //:consumer
INFO: Invocation ID: 43e4f249-3c96-4dc1-b725-4b161a5054c2
INFO: Analyzed target //:consumer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:consumer up-to-date:
  bazel-bin/consumer/out.txt
INFO: Elapsed time: 0.094s, Critical Path: 0.02s
INFO: 3 processes: 1 remote cache hit, 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 3 total actions

# If we download the other part of the output...
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel --output_groups=+a -- //:producer
INFO: Invocation ID: d8ba3512-ff22-44da-b19a-95f04c7ba594
INFO: Analyzed target //:producer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:producer up-to-date:
  bazel-bin/producer/a.txt
INFO: Elapsed time: 0.088s, Critical Path: 0.00s
INFO: 2 processes: 1 remote cache hit, 1 internal.

# ... then we no longer get extra downloads
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel -- //:consumer
INFO: Invocation ID: 7dee6046-a432-4264-a23b-ee76cccaec35
INFO: Analyzed target //:consumer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:consumer up-to-date:
  bazel-bin/consumer/out.txt
INFO: Elapsed time: 0.105s, Critical Path: 0.03s
INFO: 2 processes: 1 internal, 1 darwin-sandbox.

# If we delete that output, then we get extra downloads again
❯❯❯ rm -f bazel-bin/producer/a.txt
❯❯❯ bazel build --disk_cache=/tmp/cache --embed_label=$(date) --remote_download_toplevel -- //:consumer
INFO: Invocation ID: 190c6993-564c-48a8-8b83-98f4fa8e9e3e
INFO: Analyzed target //:consumer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:consumer up-to-date:
  bazel-bin/consumer/out.txt
INFO: Elapsed time: 0.147s, Critical Path: 0.03s
INFO: 3 processes: 1 remote cache hit, 1 internal, 1 darwin-sandbox.
INFO: Build completed successfully, 3 total actions
```

macOS 11.5.2

release 5.0.0-pre.20210817.2

https://github.com/bazelbuild/bazel/issues/12855#issuecomment-906885622

coeuvre:
When using BwtB, intermediate outputs are not downloaded. When an action is going to execute locally (either is decided by dynamic execution or is forced to run locally), Bazel will download all the inputs which are outputs of dependent actions and the downloaded file are kept after build.

For a following build, Bazel doesn't realize the downloaded files are the same as remote ones. It marks those generating actions dirty and try to re-execute. Normally, this re-execution will be skipped due to the action cache. However, if only part of the outputs are downloaded, the action cache can't help.

This PR update FilesystemValueChecker to compare between previous remote metadata and current local file metadata so the generating action isn't invalidated if the output files are staged and are same with remote ones.

This PR only fixes output files. A following PR is needed to handle output tree artifacts.

Fixes #13912.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
@@ -468,9 +468,8 @@ public class FilesystemValueChecker {
       FileArtifactValue fileMetadata =
           ActionMetadataHandler.fileArtifactValueFromArtifact(file, null, tsgm);
 
-      if (lastKnownData.isRemote()
-          && !fileMetadata.isRemote()
-          && fileMetadata.getType() == FileStateType.REGULAR_FILE) {
+      if (lastKnownData.isRemote() && lastKnownData.getType() == FileStateType.REGULAR_FILE
+          && !fileMetadata.isRemote() && fileMetadata.getType() == FileStateType.REGULAR_FILE) {
         // If the file was remote but is now staged, we compare the digests.
         byte[] digest = DigestUtils.manuallyComputeDigest(file.getPath(), fileMetadata.getSize());
         if (Arrays.equals(digest, lastKnownData.getDigest())
--- a/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
@@ -42,6 +42,7 @@ import com.google.devtools.build.lib.skyframe.TreeArtifactValue.ArchivedRepresen
 import com.google.devtools.build.lib.util.Pair;
 import com.google.devtools.build.lib.util.io.TimestampGranularityMonitor;
 import com.google.devtools.build.lib.vfs.BatchStat;
+import com.google.devtools.build.lib.vfs.DigestUtils;
 import com.google.devtools.build.lib.vfs.Dirent;
 import com.google.devtools.build.lib.vfs.FileStatusWithDigest;
 import com.google.devtools.build.lib.vfs.ModifiedFileSet;
@@ -56,6 +57,7 @@ import com.google.devtools.build.skyframe.SkyKey;
 import com.google.devtools.build.skyframe.SkyValue;
 import com.google.devtools.build.skyframe.WalkableGraph;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -465,6 +467,20 @@ public class FilesystemValueChecker {
     try {
       FileArtifactValue fileMetadata =
           ActionMetadataHandler.fileArtifactValueFromArtifact(file, null, tsgm);
+
+      if (lastKnownData.isRemote()
+          && !fileMetadata.isRemote()
+          && fileMetadata.getType() == FileStateType.REGULAR_FILE) {
+        // If the file was remote but is now staged, we compare the digests.
+        byte[] digest = DigestUtils.manuallyComputeDigest(file.getPath(), fileMetadata.getSize());
+        if (Arrays.equals(digest, lastKnownData.getDigest())
+            && fileMetadata.getSize() == lastKnownData.getSize()) {
+          // TODO(chiwang): Find a way to update lastKnownData to fileMetadata so we don't need to
+          // calculate digest again.
+          return false;
+        }
+      }
+
       boolean trustRemoteValue =
           fileMetadata.getType() == FileStateType.NONEXISTENT
               && lastKnownData.isRemote()
@@ -498,6 +514,7 @@ public class FilesystemValueChecker {
       }
     }
 
+    // TODO(chiwang): Check the case that tree artifacts were previously remote but are now staged.
     for (Map.Entry<Artifact, TreeArtifactValue> entry :
         actionValue.getAllTreeArtifactValues().entrySet()) {
       TreeArtifactValue tree = entry.getValue();
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -3171,4 +3171,130 @@ EOF
   expect_log "-r-xr-xr-x"
 }
 
+function test_downloads_toplevel_partial_intermediate_outputs() {
+  # Test that when only partial outputs of an intermediate action are
+  # downloaded, rebuild doesn't re-execute the action. See
+  # https://github.com/bazelbuild/bazel/issues/13912.
+
+  mkdir -p a
+
+  # producer generates a.txt and b.txt
+  # consumer only depends on b.txt and generates out.txt
+  cat > a/rules.bzl <<EOF
+ProducerInfo = provider(
+    fields = {
+        "a": "a",
+        "b": "b",
+    },
+)
+
+def _producer_impl(ctx):
+    a = ctx.actions.declare_file("{}/a.txt".format(ctx.attr.name))
+    b = ctx.actions.declare_file("{}/b.txt".format(ctx.attr.name))
+
+    ctx.actions.run_shell(
+        command = """
+head -c 100 </dev/urandom > \$1
+head -c 100 </dev/urandom > \$2
+""",
+        arguments = [
+            a.path,
+            b.path,
+        ],
+        outputs = [a, b],
+        mnemonic = "Producer",
+    )
+
+    return [
+        ProducerInfo(
+            a = depset([a]),
+            b = depset([b]),
+        ),
+    ]
+
+producer = rule(
+    implementation = _producer_impl,
+)
+
+def _consumer_impl(ctx):
+    b = ctx.attr.producer[ProducerInfo].b
+    output = ctx.actions.declare_file("{}/out.txt".format(ctx.attr.name))
+
+    args = ctx.actions.args()
+    args.add_all(b)
+
+    ctx.actions.run_shell(
+        command = """
+readonly output="\$1"
+shift
+cat \$@ > "\$output"
+""",
+        arguments = [
+            output.path,
+            args,
+        ],
+        inputs = depset([ctx.info_file], transitive = [b]),
+        mnemonic = "Consumer",
+        outputs = [output],
+        execution_requirements = {
+            "no-cache": "1",
+            "no-remote": "1",
+        },
+    )
+
+    return [
+        DefaultInfo(
+            files = depset([output]),
+        )
+    ]
+
+consumer = rule(
+    implementation = _consumer_impl,
+    attrs = {
+        "producer": attr.label(providers = [[ProducerInfo]]),
+    },
+)
+EOF
+  cat > a/BUILD <<EOF
+load(":rules.bzl", "producer", "consumer")
+
+producer(
+    name = "producer",
+)
+
+consumer(
+    name = "consumer",
+    producer = ":producer",
+)
+EOF
+
+  # Populate the cache
+  bazel build \
+      --remote_cache=grpc://localhost:${worker_port} \
+      --remote_download_toplevel \
+      //a:consumer >& $TEST_log \
+      || fail "Failed to build"
+
+  bazel clean >& $TEST_log || fail "Failed to clean"
+
+  # Only downloads out.txt and b.txt
+  bazel build \
+      --remote_cache=grpc://localhost:${worker_port} \
+      --remote_download_toplevel \
+      //a:consumer >& $TEST_log \
+      || fail "Failed to build"
+
+  ([[ -f bazel-bin/a/producer/b.txt ]] && [[ -f bazel-bin/a/consumer/out.txt ]] ) \
+    || fail "Expected out.txt and b.txt are downloaded"
+  (! [[ -f bazel-bin/a/producer/a.txt ]] ) \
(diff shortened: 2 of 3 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
```diff
@@ -465,6 +467,19 @@ private boolean artifactIsDirtyWithDirectSystemCalls(
...
       FileArtifactValue fileMetadata =
           ActionMetadataHandler.fileArtifactValueFromArtifact(file, null, tsgm);
+
+      if (lastKnownData.isRemote() && lastKnownData.getType() == FileStateType.REGULAR_FILE
+          && !fileMetadata.isRemote() && fileMetadata.getType() == FileStateType.REGULAR_FILE) {
+        // If the file was remote but is now staged, we compare the digests.
+        byte[] digest = DigestUtils.manuallyComputeDigest(file.getPath(), fileMetadata.getSize());
+        if (Arrays.equals(digest, lastKnownData.getDigest())
+            && fileMetadata.getSize() == lastKnownData.getSize()) {
+          // TODO(chiwang): Find a way to update lastKnownData to fileMetadata so we don't need to
```
alexjski, line 477: One way to do this could be to attach a FileContentsProxy to a RemoteFileArtifactValue when we download the input. I think whether that idea makes sense would be informed by how we download the inputs -- how do we avoid re-downloading inputs today? By re-downloading, I mean a case of Action B depends on Action A and say we change something with Action B, but not A -- we would want not to materialize remote outputs of Action A again.

If we did that, we should see if the extra pointer takes more memory (for non-RBE case, we already have the actionId we don't use). Then, this code may not need any changes -- all you would need is to allow that content proxy comparison to work between RegularFileArtifactValue and RemoteFileArtifactValue.

alexjski (changes requested):
Would we need this change if we enabled action cache to store the remote metadata? My main worry is about the extra IO in case when we are not using an output service, namely in the case when a digest does not match -- will we end up computing a digest of of a stale output twice (once when checking the outputs, once when checking action cache for it)?

In particular, due to lack of sharing of these digest, we can imagine that 1 change to an output can trigger 2x digesting of N outputs for a rule which has N of them.

I can see that potentially helping a bit in the case of no changes to outputs, namely we would not dirty some ActionExecutionValues which later get change-pruned.

coeuvre:
> Would we need this change if we enabled action cache to store the remote metadata?

If we store remote metadata in action cache, we don't need this change. But I am wondering do we want these extra steps in this case i.e. invalidating generating actions, checking action cache and change-pruning.

> My main worry is about the extra IO in case when we are not using an output service, namely in the case when a digest does not match -- will we end up computing a digest of of a stale output twice (once when checking the outputs, once when checking action cache for it)?

True.

I think, the best we can do is, after downloading the inputs, replacing RemoteFileArtifactValue with RegularFileArtifactValue in the ActionExecutionValues. Is that possible?

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3171,4 +3171,130 @@ EOF
...
+            files = depset([output]),
+        )
+    ]
+
+consumer = rule(
+    implementation = _consumer_impl,
+    attrs = {
+        "producer": attr.label(providers = [[ProducerInfo]]),
+    },
+)
```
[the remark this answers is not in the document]
coeuvre: The critical part of this test case is that consumer only need b.txt from producer (i.e. only part of outputs from its dependency). I don't know how to achieve that with genrules (sorry for my poor starlark skill).

src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
coeuvre: > how do we avoid re-downloading inputs today?

We guard the download with `FileArtifactValue.isRemote`.

After changing Action B, the following build will check Action A via action cache during which the RegularFileArtifactValues of staged outputs are generated. If Action A hit the cache, we will use these RegularFileArtifactValues for Action B. So there is no extra downloads for Action B in this run.

coeuvre:
> I think, the best we can do is, after downloading the inputs, replacing RemoteFileArtifactValue with RegularFileArtifactValue in the ActionExecutionValues. Is that possible?

Let skyframe invalidate the generating actions and redo the action cache checking is actually an indirect way to update ActionExeciontValues with RegularFileArtifactValues. Do we have more direct way to achieve this immediately after downloads?

src/test/shell/bazel/remote/remote_execution_test.sh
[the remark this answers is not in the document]
alexjski: No worries -- that is really fancy Starlark. I would think about something like this:

```
genrule(name="dep", outs=["a.txt", "b.txt"], cmd="touch $(SRCS)")
genrule(name="top", srcs=[":b.txt"], outs=["top.out"], cmd="cp $< $@")
```

src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
alexjski: > how do we avoid re-downloading inputs today?

Imagine we have action A executed remotely and action B locally.
 
1. Build 1 -- we run A remotely, download outputs of A to run action B.
2. We change a source of Action B so that it needs to be rerun.
2. Build 2 -- we run `bazel build :b` -- for A, we have a Skyframe cache hit, B we execute.

For A, we get an `ActionExecutionValue` from Skyframe, no action cache checks needed, hence the metadata for all its outputs should stay as remote. Will we download the inputs again in this case even though they are already available locally from prior build?

What happens if I change the downloaded file on disk between bazel builds?

alexjski:
> > Would we need this change if we enabled action cache to store the remote metadata?
> 
> If we store remote metadata in action cache, we don't need this change. But I am wondering do we want these extra steps in this case i.e. invalidating generating actions, checking action cache and change-pruning.
> 
> > My main worry is about the extra IO in case when we are not using an output service, namely in the case when a digest does not match -- will we end up computing a digest of of a stale output twice (once when checking the outputs, once when checking action cache for it)?
> 
> True.
> 
> I think, the best we can do is, after downloading the inputs, replacing RemoteFileArtifactValue with RegularFileArtifactValue in the ActionExecutionValues. Is that possible?

The unfortunate part is that this would make ActionExecutionValue not-immutable anymore + you would need to synchronize that. It may be easier to do so at the level of the RemoteFileArtifactValue (imagine you add a transitive field there with the content proxy).

src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
coeuvre: > Will we download the inputs again in this case even though they are already available locally from prior build?


In theory, we would download the inputs again because the ActionExecutionValue for Action A contains RemoteFileArtifactValue. However, in practice, and I have done some local tests, for some reason, Action A is invalidated and we check the action cache against it so the metadata becomes RegularFileArtifactValue when we run Action B. I don't totally understand the reason why Action A is invalidated but maybe related to the original issue.


> What happens if I change the downloaded file on disk between bazel builds?


The generating action (in this case, Action A) is invalidated and re-executed. Before execution, the downloaded (and modified) file will be deleted. Action A is executed remotely and then we get the action result without downloading the outputs.
1. If the result of Action A is same as before, Action B won't be re-executed because of action cache.
2. Otherwise, execute Action B, download the updated outputs of Action A.

coeuvre:
I think replacing RemoteFileArtifactValue with RegularFileArtifactValue after we downloaded the files (either with some hacks in prefetchInputs or via action rewinding) is a proper fix. This can fix the original issue and can prevent the same files be re-downloaded as well.

I can see a lot of issues fixed by implementing action rewinding but it's not that easy to do. I will try with some hacks in prefetchInputs first.

coeuvre:
Closing since this change adds extra digest computations and I would like to explore other ways to fix the issue. In the meantime, a workaround is to use `--experimental_action_cache_store_output_metadata`.

alexjski:
> I think replacing RemoteFileArtifactValue with RegularFileArtifactValue after we downloaded the files (either with some hacks in prefetchInputs or via action rewinding) is a proper fix. This can fix the original issue and can prevent the same files be re-downloaded as well.
> 
> I can see a lot of issues will be fixed by implementing action rewinding but it's not that easy to do. I will try with some hacks in prefetchInputs first.

The problem there is that in case of rewinding, you would want a special case of that -- one which does know about the prior result to avoid re-digesting the downloaded files. I think the best bet may be to update the actionExcecutionValue in prefetchInputs. Actually, my favorite would be to add nullable contentproxy to `RemoteActionFileArtifactValue`, update that after a download and use that to avoid the re-download as well.

src/main/java/com/google/devtools/build/lib/skyframe/FilesystemValueChecker.java
alexjski: That's surprising -- I would think one should be able to invalidate a top-action without needing to change the base one. I would be more than happy to try to help you offline with constructing an example doing that.

PR #14234, opened 2021-11-05
bzlmod: Add support for WORKSPACE.bzlmod

meteorcloudy:
When enabling Bzlmod, if a WORKSPACE.bzlmod file is presented at the
source root, it will replace the content of the WORKSPACE file.
In contrast to the WORKSPACE file, no prefix or suffix are added to the WORKSPACE.bzlmod file.

This file is for migration purpose in case some of the dependencies
still need to be introduced via the WORKSPACE mechanism, eventually
it should be removed.

RELNOTES: None
GOOGLE:

### DIFF
--- a/WORKSPACE.bzlmod
+++ b/WORKSPACE.bzlmod
@@ -1,31 +1,425 @@
-# Replace the original WORKSPACE file with this file to build Bazel with bzlmod.
-# We still need the bind rules to make protobuf work.
+# Replace the original WORKSPACE file with this file to build Bazel with Bzlmod.
+# When using the WORKSPACE.bzlmod, no WORKSPACE prefix or suffix are added.
+# Therefore, we need to keep some legacy content to make the build work, eventually this
+# file should be empty after fully migrating to Bzlmod.
 workspace(name = "io_bazel")
 
 # Required by @com_google_protobuf//java/util:util in protobuf 3.19.0
 # TODO(pcloudy): Remove those bind rules, when com_google_protobuf can
 # fetch jar dependencies via rules_jvm_external with module extension.
 bind(
-  name = "error_prone_annotations",
-  actual = "//third_party:error_prone_annotations",
+    name = "error_prone_annotations",
+    actual = "//third_party:error_prone_annotations",
 )
 
 bind(
-  name = "j2objc_annotations",
-  actual = "//third_party/java/j2objc-annotations:j2objc-annotations",
+    name = "j2objc_annotations",
+    actual = "//third_party/java/j2objc-annotations:j2objc-annotations",
 )
 
 bind(
-  name = "gson",
-  actual = "//third_party:gson",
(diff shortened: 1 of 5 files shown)


WORKSPACE.bzlmod
```diff
@@ -1,31 +1,425 @@
-# Replace the original WORKSPACE file with this file to build Bazel with bzlmod.
-# We still need the bind rules to make protobuf work.
+# Replace the original WORKSPACE file with this file to build Bazel with Bzlmod.
```
Wyverald, line 1: this line is no longer necessary, right?

src/main/java/com/google/devtools/build/lib/skyframe/WorkspaceFileFunction.java
```diff
@@ -100,27 +101,49 @@ public SkyValue compute(SkyKey skyKey, Environment env)
...
+
+    Optional<RootedPath> resolvedFile = Preconditions.checkNotNull(
+        RepositoryDelegatorFunction.RESOLVED_FILE_INSTEAD_OF_WORKSPACE.get(env));
+    boolean useWorkspaceResolvedFile = resolvedFile.isPresent();
+
+    boolean useWorkspaceBzlmodFile = false;
+    RootedPath workspaceBzlmodFile = RootedPath.toRootedPath(workspaceFile.getRoot(),
+        workspaceFile.getRootRelativePath().replaceName("WORKSPACE.bzlmod"));
+    // We only need to check WORKSPACE.bzlmod when the resolved file isn't used.
+    if (!useWorkspaceResolvedFile) {
```
Wyverald, line 121: this condition can be simply rewritten as:

```java
if (!useWorkspaceResolvedFile && RepositoryDelegatorFunction.ENABLE_BZLMOD.get(env)) {
  FileValue ... = ...;
  if (...) { ... }
  useWorkspaceBzlmodFile = ...;
}
```

IMO the logic is much clearer that way.

WORKSPACE.bzlmod
meteorcloudy: Oh, right...

src/main/java/com/google/devtools/build/lib/skyframe/WorkspaceFileFunction.java
meteorcloudy: Oh, absolutely,, thanks!

WORKSPACE.bzlmod
meteorcloudy: Indeed..

meteorcloudy:
Merged in https://github.com/bazelbuild/bazel/commit/09fe7cce9a2b8716476b001be221c42d3d03b180#diff-98160be0a821b608863421665e00bc0af7276035713a2f309a19fdd426117faf

PR #14238, opened 2021-11-05
Fix Cquery

andrewkatson:
It needed to expect .exe when there are binaries produced

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/query2/testutil/AbstractQueryTest.java
+++ b/src/test/java/com/google/devtools/build/lib/query2/testutil/AbstractQueryTest.java
@@ -76,9 +76,9 @@ public abstract class AbstractQueryTest<T> {
   private static final String DEFAULT_UNIVERSE = "//...:*";
 
   protected static final String BAD_PACKAGE_NAME =
-      "package names may contain "
-          + "A-Z, a-z, 0-9, or any of ' !\"#$%&'()*+,-./;<=>?[]^_`{|}~' "
-          + "(most 7-bit ascii characters except 0-31, 127, ':', or '\\')";
+          "package names may contain "
+                  + "A-Z, a-z, 0-9, or any of ' !\"#$%&'()*+,-./;<=>?[]^_`{|}~' "
+                  + "(most 7-bit ascii characters except 0-31, 127, ':', or '\\')";
 
   protected MockToolsConfig mockToolsConfig;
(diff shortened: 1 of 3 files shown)


meteorcloudy (approved):
Thanks!

gregestren:
What was the failure this fixed?

src/test/java/com/google/devtools/build/lib/query2/testutil/AbstractQueryTest.java
```diff
@@ -404,24 +437,42 @@ public void testAttrOperator() throws Exception {
...
+      // Using "empty" pattern will just check existence of the attribute.
+      assertThat(evalToString("attr(cmd,'', c:*)")).isEqualTo("//c:c");
+      assertThat(evalToString("attr(linkshared, 0, c:*)")).isEqualTo("//c:d");
+      assertThat(evalToString("attr('data', 'r', c:*)")).isEqualTo("//c:d");
+      // Empty list attribute value always resolves to '[]'. If list attribute has
+      // more than one value, the will be delimited with ','.
+      assertThat(evalToString("attr('deps', '\\[\\]', c:*)")).isEqualTo("//c:d");
+      assertThat(evalToString("attr('deps', '^..$', c:*)")).isEqualTo("//c:d");
+      assertThat(evalToString("attr('srcs', '\\[[^,]+\\]', c:*)")).isEqualTo("//c:d");
+
```
meteorcloudy, line 454: Can we not duplicated those lines? Instead we can optional add ".exe" based on the OS using Strings.format

meteorcloudy (commented):
Thanks, but can you figure out a way to fix the indentation? Or you can just send this change as an internal CL, internal tool should also be able to fix it for you. ;)

meteorcloudy:
@gregestren 

I think it's just those tests are not enabled on Windows due to the missing ".exe" extension. Andrew is helping us enabling those tests.

src/test/java/com/google/devtools/build/lib/query2/testutil/AbstractQueryTest.java
andrewkatson: Okay ill fix that

sgowroji:
Hello @andrewkatson, Could you please reply on the above [comment](https://github.com/bazelbuild/bazel/pull/14238#issuecomment-1116232745). And let us know the status of the PR?

PR #14244, opened 2021-11-08
correcting vs2019 VC buildtools path.

perezd:


perezd:
Don't actually change this.

PR #14245, opened 2021-11-09
Fix test_cc_starlark_api_additional_inputs when used with lld.

benjaminp:
Unlike the GNU linkers, lld does not insert an absolute symbol for every version in the version script into the symbol table. Check the version script is working by instead assigning a particular symbol a version.

### DIFF
--- a/src/test/shell/bazel/cc_integration_test.sh
+++ b/src/test/shell/bazel/cc_integration_test.sh
@@ -490,8 +490,8 @@ EOF
 
   cat > $pkg/script.lds << EOF
 VERS_42.0 {
-  local:
-    *;
+  global:
+    f;
 };
 EOF
 
@@ -560,9 +560,8 @@ function test_cc_starlark_api_additional_inputs() {
   cat >> "$pkg"/BUILD << EOF
 cc_bin(
     name = "g",
-    srcs = ["e.cc"],
-    data = [":f"],
-    linkstatic = 1,
+    srcs = ["f.cc"],
+    linkshared = 1,
     additional_linker_inputs = ["script.lds"],
     user_link_flags = [
         "-ldl",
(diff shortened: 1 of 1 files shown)


src/test/shell/bazel/cc_integration_test.sh
```diff
@@ -603,9 +603,8 @@ function test_cc_starlark_api_additional_inputs() {
   cat >> "$pkg"/BUILD << EOF
 cc_bin(
     name = "g",
-    srcs = ["e.cc"],
```
buildbreaker2021, line 606: Cannot the test pass with an already existing target? Why is this modification necessary?

src/test/shell/bazel/cc_integration_test.sh
```diff
@@ -603,9 +603,8 @@ function test_cc_starlark_api_additional_inputs() {
...
-    data = [":f"],
-    linkstatic = 1,
```
buildbreaker2021, line 608: Cannot the test pass with an already existing target? Why is this change necessary?

src/test/shell/bazel/cc_integration_test.sh
```diff
@@ -622,9 +621,7 @@ EOF
   bazel build --experimental_cc_skylark_api_enabled_packages=, --verbose_failures \
     //"$pkg":g  &>"$TEST_log" || fail "Build failed"
 
-  nm bazel-bin/"$pkg"/g  | grep VERS_42.0 || fail "VERS_42.0 not in binary"
-
-  bazel-bin/"$pkg"/g | grep a1a2bcddcb2a1a || fail "output is incorrect"
```
buildbreaker2021, line 627: What's the reason behind deleting this line? Can the be preserved?

benjaminp:
Thank you for the review.

To hopefully answer both of your questions: The intention of this test appears to be to check that a linker script can be successfully passed using the `additional_inputs` flag to `cc_common.link`. The test was previous relying on the linker inserting a symbol from the version script into the binary regardless of whether any symbols were given that version. However, lld does not have this behavior. Therefore, I modified the test binary and linker script to assign a version to a specific symbol, `f`.

buildbreaker2021 (approved):

PR #14251, opened 2021-11-09
Fixes for the Starlark transition hash computation

fmeum:
@Wyverald These two commits fix a whole cluster of issues around wrongly calculated transition hashes, leading to action conflicts on otherwise correct builds. I think they would be very useful to have in Bazel 5.

@sdtwigg FYI


Wyverald:
Thanks!

Issue #13882, opened 2021-08-20
Build failure from missing files when using `--experimental_action_cache_store_output_metadata`


brentleyjones:
### Description of the problem / feature request:

When using `--experimental_action_cache_store_output_metadata`, which was introduced in 4e29042fd77eae565a711655df6150500cb6e915, you can get build failures when going from using a remote cache to not using one.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```shell
bazel build --remote_cache=grpc://... --remote_download_toplevel --experimental_action_cache_store_output_metadata -- //...
bazel build --remote_download_toplevel --experimental_action_cache_store_output_metadata -- //...
```

The second command will fail with errors like this:

```shell
(06:29:47) INFO: Writing explanation of rebuilds to 'tmp/logs/explanation.log'
(06:29:49) ERROR: /private/var/tmp/_bazel_iosci/4e089aeca56f6c6b5c9850816d5a6522/external/build_bazel_rules_swift/tools/worker/BUILD:85:10: Linking external/build_bazel_rules_swift/tools/worker/worker [for host] failed: (Aborted): cc_wrapper.sh failed: error executing command
  (cd /private/var/tmp/_bazel_iosci/4e089aeca56f6c6b5c9850816d5a6522/sandbox/darwin-sandbox/9/execroot/repo && \
  exec env - \
    APPLE_SDK_PLATFORM=MacOSX \
    APPLE_SDK_VERSION_OVERRIDE=11.3 \
    PATH=/usr/bin:/bin \
    RELATIVE_AST_PATH=true \
    XCODE_VERSION_OVERRIDE=12.5.1.12E507 \
    ZERO_AR_DATE=1 \
  external/local_config_cc/cc_wrapper.sh @bazel-out/host/bin/external/build_bazel_rules_swift/tools/worker/worker-2.params)
Execution platform: @local_config_platform//:host
 
Use --sandbox_debug to see verbose messages from the sandbox
clang: error: no such file or directory: 'bazel-out/host/bin/external/build_bazel_rules_swift/tools/worker/_objs/worker/worker_main.o'
clang: error: no such file or directory: 'bazel-out/host/bin/external/build_bazel_rules_swift/tools/worker/libcompile_without_worker.a'
clang: error: no such file or directory: 'bazel-out/host/bin/external/build_bazel_rules_swift/tools/worker/libcompile_with_worker.a'
[...]

PR #14252, opened 2021-11-10
Remote: Fix "file not found" error when remote cache is changed from enabled to disabled.

coeuvre:
When use BwtB, intermediate outputs are not downloaded. If a following build disables remote cache, a "file not found" error will be thrown if an action doesn't have its inputs downloaded in previous build because those files cannot be downloaded in this build.

This change fix this issue by:
  1. Do not load remote metadata from action cache if `--experimental_action_cache_store_output_metadata` is set and remote cache is disabled.
  2. Invalidate action nodes if previous build use BwtB and remote cache is changed from enabled to disabled.

Fixes https://github.com/bazelbuild/bazel/issues/13882.

### DIFF
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -3216,8 +3216,6 @@ function download_toplevel_when_turn_remote_cache_off() {
   # Test that BwtB doesn't cause build failure if remote cache is disabled in a following build.
   # See https://github.com/bazelbuild/bazel/issues/13882.
 
-  local extra_build_flags="$@"
-
   cat > .bazelrc <<EOF
 build --verbose_failures
 EOF
@@ -3250,7 +3248,7 @@ EOF
   bazel build \
     --remote_cache=grpc://localhost:${worker_port} \
     --remote_download_toplevel \
-    $extra_build_flags \
+    "$@" \
     //a:consumer >& $TEST_log || fail "Failed to download outputs"
   [[ -f bazel-bin/a/a.txt ]] || [[ -f bazel-bin/a/b.txt ]] \
     && fail "Expected outputs of producer are not downloaded"
@@ -3259,7 +3257,7 @@ EOF
   echo 'bar' > a/in.txt
   bazel build \
     --remote_download_toplevel \
-    $extra_build_flags \
+    "$@" \
     //a:consumer >& $TEST_log || fail "Failed to build without remote cache"
 }
 
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -3204,4 +3204,72 @@ EOF
   expect_not_log "WARNING: Writing to Remote Cache:"
 }
 
+function test_download_toplevel_when_turn_remote_cache_off() {
+  # Test that BwtB does cause build failure if remote cache is disabled in a following build.
+  # See https://github.com/bazelbuild/bazel/issues/13882.
+
+  mkdir -p a
+  cat > a/BUILD <<EOF
+genrule(
+    name = "producer",
+    outs = ["a.txt", "b.txt"],
+    cmd = "touch \$(OUTS)",
+)
+
+genrule(
+    name = "consumer",
+    outs = ["out.txt"],
+    srcs = [":b.txt", "in.txt"],
+    cmd = "cat \$(SRCS) > \$@",
+)
+EOF
+  echo 'foo' > a/in.txt
+
+  # populate the cache
+  bazel build \
+    --remote_cache=grpc://localhost:${worker_port} \
+    --remote_download_toplevel \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to populate the cache"
+
+  bazel clean || fail "Failed to clean"
+
+  # download top level outputs without remote metadata
+  bazel build \
+    --remote_cache=grpc://localhost:${worker_port} \
+    --remote_download_toplevel \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to download outputs without remote metadata"
+  (! [[ -f bazel-bin/a/a.txt ]] && ! [[ -f bazel-bin/a/b.txt ]]) \
+  || fail "Expected outputs of producer are not downloaded without remote metadata"
+
+  # build without remote cache without remote metadata
+  echo 'bar' > a/in.txt
+  bazel build \
+    --remote_download_toplevel \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to build without remote cache without remote metadata"
+
+  bazel clean || fail "Failed to clean"
+  echo 'foo' > a/in.txt
+
+  # download top level outputs with remote metadata
+  bazel build \
+    --remote_cache=grpc://localhost:${worker_port} \
+    --remote_download_toplevel \
+    --experimental_action_cache_store_output_metadata \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to download outputs with remote metadata"
+  (! [[ -f bazel-bin/a/a.txt ]] && ! [[ -f bazel-bin/a/b.txt ]]) \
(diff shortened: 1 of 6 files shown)


src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3204,4 +3204,72 @@ EOF
   expect_not_log "WARNING: Writing to Remote Cache:"
 }
 
+function test_download_toplevel_when_turn_remote_cache_off() {
+  # Test that BwtB does cause build failure if remote cache is disabled in a following build.
+  # See https://github.com/bazelbuild/bazel/issues/13882.
+
+  mkdir -p a
+  cat > a/BUILD <<EOF
```
alexjski, line 3212: nit: you can use `EOF` and not need to escape the `$` in the content.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3204,4 +3204,72 @@ EOF
   expect_not_log "WARNING: Writing to Remote Cache:"
 }
 
+function test_download_toplevel_when_turn_remote_cache_off() {
+  # Test that BwtB does cause build failure if remote cache is disabled in a following build.
+  # See https://github.com/bazelbuild/bazel/issues/13882.
+
+  mkdir -p a
```
alexjski, line 3211: supernit: `-p` not needed in this case

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3204,4 +3204,72 @@ EOF
...
+  echo 'foo' > a/in.txt
+
+  # populate the cache
+  bazel build \
+    --remote_cache=grpc://localhost:${worker_port} \
+    --remote_download_toplevel \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to populate the cache"
+
+  bazel clean || fail "Failed to clean"
```
alexjski, line 3235: super nit: `bazel clean >& "${TEST_log}" || fail ...`

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3204,4 +3204,72 @@ EOF
...
+
+  # download top level outputs without remote metadata
+  bazel build \
+    --remote_cache=grpc://localhost:${worker_port} \
+    --remote_download_toplevel \
+    --verbose_failures \
+    //a:consumer >& $TEST_log || fail "Failed to download outputs without remote metadata"
+  (! [[ -f bazel-bin/a/a.txt ]] && ! [[ -f bazel-bin/a/b.txt ]]) \
```
alexjski, line 3243: nit: can we reverse that to `[[ -f bazel-bin/a/a.txt ]] || [[ -f bazel-bin/a/b.txt ]] && fail ...`?

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3204,4 +3204,72 @@ EOF
...
+  || fail "Expected outputs of producer are not downloaded without remote metadata"
+
+  # build without remote cache without remote metadata
```
alexjski, line 3246: Is this the build which fails without your change? The reason I am asking is because this test may be merging 2 test cases in here. In general, it is a little hard to follow what is arrange-act-assert in here.

Meta-comment -- is there a way to run the remote tests in a Java test using BuildIntegrationTestCase? In those, it is much easier to parameterize etc.

alexjski (changes requested):
I would recommend splitting that into 2 logically separate changes as in the description.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Split into 2 test cases.

> Meta-comment -- is there a way to run the remote tests in a Java test using BuildIntegrationTestCase? In those, it is much easier to parameterize etc.

Not sure. These remote tests run a remote worker process which is a real remote server. Do we prefer BuildIntegrationTestCase over shell in general? If so, I can try to make it work with remote tests and write new tests there in future PRs.

coeuvre (commented):
Thanks for the comments. I will split this PR when importing.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3232,44 +3232,73 @@ EOF
...
   echo 'bar' > a/in.txt
   bazel build \
     --remote_download_toplevel \
     --verbose_failures \
-    //a:consumer >& $TEST_log || fail "Failed to build without remote cache without remote metadata"
+    //a:consumer >& $TEST_log || fail "Failed to build without remote cache"
+}
 
-  bazel clean || fail "Failed to clean"
+function test_download_toplevel_when_turn_remote_cache_off_with_metadata() {
```
alexjski, line 3254: These 2 test cases differ by 1 option only -- why not put the test case as a separate helper function and call that twice with different parameters (Bash tests parameterization). Example [here](https://github.com/bazelbuild/bazel/blob/bb2e0ae6021da27cb4d6044bcb5c26e00d2b3db3/src/test/shell/integration/client_sigint_test.sh#L80-L86).

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3232,44 +3232,73 @@ EOF
...
 
-  bazel clean || fail "Failed to clean"
+  bazel clean >& $TEST_log || fail "Failed to clean"
 
-  # download top level outputs without remote metadata
+  # download top level outputs
   bazel build \
     --remote_cache=grpc://localhost:${worker_port} \
     --remote_download_toplevel \
     --verbose_failures \
```
alexjski, line 3241: supernit: `--verbose_failures` is unlikely needed/we could add that to the blazerc for the whole test if that aids debugging failures.

src/test/shell/bazel/remote/remote_execution_test.sh
alexjski: We generally do. I wouldn't tell you to rewrite these tests, but you may want to prefer Java-based ones when applicable in the future. Sent you a document talking about that offline.

alexjski (approved):

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
coeuvre: Done.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3286,8 +3250,7 @@ EOF
   bazel build \
     --remote_cache=grpc://localhost:${worker_port} \
     --remote_download_toplevel \
-    --experimental_action_cache_store_output_metadata \
-    --verbose_failures \
+    $extra_build_flags \
```
alexjski, line 3253: I suspect bash linter may not like that -- your best bet may be to use the "$@" array directly or declare and use extra_build_flags as an array and use like that (personally, I recommend `"$@"`):

```
local -r extra_build_flags=("$@")
...
"${extra_build_flags[@]}"
```

alexjski (approved):

alexjski (approved):

PR #14253, opened 2021-11-10
(WIP) Bzlmod: export DEFAULT_SYSTEM_JAVABASE to MODULE.bazel

meteorcloudy:
`DEFAULT_SYSTEM_JAVABASE` is an default variable in WORKSPACE file, its value is controlled by `--default_system_javabase` flag, we need this information to locate the local JDK path in rules_java.

Required by https://github.com/bazelbuild/rules_java/pull/45#issuecomment-965197860

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
@@ -229,7 +229,7 @@ public class BazelRepositoryModule extends BlazeModule {
         .addSkyFunction(SkyFunctions.REPOSITORY_DIRECTORY, repositoryDelegatorFunction)
         .addSkyFunction(
             SkyFunctions.MODULE_FILE,
-            new ModuleFileFunction(registryFactory, directories.getWorkspace()))
+            new ModuleFileFunction(registryFactory, directories.getWorkspace(), directories.getLocalJavabase()))
         .addSkyFunction(SkyFunctions.BAZEL_MODULE_RESOLUTION, new BazelModuleResolutionFunction())
         .addSkyFunction(SkyFunctions.SINGLE_EXTENSION_EVAL, singleExtensionEvalFunction)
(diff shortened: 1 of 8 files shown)


Wyverald:
cc @comius 

I think DEFAULT_SYSTEM_JAVABASE as a symbol in WORKSPACE was an ugly hack, and I would really like to avoid having it in MODULE.bazel too. [Off-topic: In fact it feels weird that we can specify `--default_system_javabase` on the command line, but cannot specify, for example, `--default_system_pythonbase` or something (i.e. why is Java special?).]

Anyhow, I think my preferred solution would be to make it available under `ctx.os.default_system_javabase`, or something to that effect. That way, module extensions and repo rules get access to it, but not MODULE.bazel files. WDYT?

comius:
Hey, I wish you asked about this earlier. I think there are no obstacles to completely remove DEFAULT_SYSTEM_JAVABASE logic and its passing down to the workspace. 

Proper implementation should leverage repository api in the local_java_repository repo rule. I believe the logic that happens in C++ can be completely reproduced in local_java_repository's implementation. It would perhaps be even better, to simplify it and only resolve to JAVA_HOME environment variable (this is what's advertised in the documentation, but we actually also look in PATH).

This was one thing I wanted to do, but there was no proper motivation to do it, that is "tech debt", but not much impact when removing it. Now with bzlmod, it has much bigger impact to remove it, because you can keep the codebase simple.

meteorcloudy:
OK, looks like the java base is detected in the Bazel client, and each platform has its own implementation:
https://cs.opensource.google/search?q=GetSystemJavabase&sq=&ss=bazel

meteorcloudy:
> to simplify it and only resolve to JAVA_HOME environment variable

Can we require user to always set JAVA_HOME manually? That's a breaking change, right?

comius:
> Can we require user to always set JAVA_HOME manually? That's a breaking change, right?

You're right, Java compilations on systems without JAVA_HOME set would start failing. JAVA_HOME or PATH it is.

meteorcloudy:
Closing this one as we prefer to get rid of `DEFAULT_SYSTEM_JAVABASE`

PR #14258, opened 2021-11-11
Retry on HTTP remote cache fetch failure

aherrmann:
Bazel's previous behavior was to rebuild an artifact locally if fetching
it from an HTTP remote cache failed. This behavior is different from
GRPC remote cache case where Bazel will retry the fetch.

The lack of retry is an issue for multiple reasons: On one hand
rebuilding locally can be slower than fetching from the remote cache, on
the other hand if a build action is not bit reproducible, as is the case
with some compilers, then the local rebuild will trigger cache misses on
further build actions that depend on the current artifact.

This change aims to avoid theses issues by retrying the fetch in the
HTTP cache case similarly to how the GRPC cache client does it.

Some care needs to be taken due to the design of Bazel's internal remote
cache client API. For a fetch the client is given an `OutputStream`
object that it is expected to write the fetched data to. This may be a
temporary file on disk that will be moved to the final location after
the fetch completed. On retry, we need to be careful to not duplicate
previously written data when writing into this `OutputStream`. Due to
the generality of the `OutputStream` interface we cannot reset the file
handle or write pointer to start fresh. Instead, this change follows the
same pattern used in the GRPC cache client. Namely, keep track of the
data previously written and continue from that offset on retry.

With this change the HTTP cache client will attempt to fetch the data
from the remote cache via an HTTP range request. So that the server only
needs to send the data that is still missing. If the server replies with
a 206 Partial Content response, then we write the received data directly
into the output stream, if the server does not support range requests
and instead replies with the full data, then we drop the duplicate
prefix and only write into the output stream from the required offset.

This patch has been running successfully in production [here](https://github.com/digital-asset/daml/pull/11238).

cc @cocreature

**TODO**
- [x] Testing - this PR does not include any testing, yet. If anyone has any pointers on how to best test this in the Bazel code base please let me know. One option I briefly explored was to add a test-case to `src/test/java/com/google/devtools/build/lib/remote/http/HttpCacheClientTest.java`. But, that seemed to require implementing support for HTTP range requests in `src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/http/HttpCacheServerHandler.java`. Perhaps there is a better way to test this change.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -246,6 +246,8 @@ public final class RemoteModule extends BlazeModule {
                      String msg = e.getMessage().toLowerCase();
                      if (msg.contains("connection reset by peer")) {
                        retry = true;
+                     } else if (msg.contains("operation timed out")) {
+                       retry = true;
                      }
                    }
                    return retry;
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -103,6 +103,7 @@ import io.grpc.CallCredentials;
 import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannel;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 import java.io.IOException;
 import java.net.URI;
@@ -226,7 +227,11 @@ public final class RemoteModule extends BlazeModule {
             if (e instanceof ClosedChannelException) {
               retry = true;
             } else if (e instanceof HttpException) {
-              retry = true;
+              HttpResponseStatus status = ((HttpException) e).response().status();
+              retry = status == HttpResponseStatus.INTERNAL_SERVER_ERROR
+                  || status == HttpResponseStatus.BAD_GATEWAY
+                  || status == HttpResponseStatus.SERVICE_UNAVAILABLE
+                  || status == HttpResponseStatus.GATEWAY_TIMEOUT;
             } else if (e instanceof IOException) {
               String msg = e.getMessage().toLowerCase();
               if (msg.contains("connection reset by peer")) {
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -59,9 +59,11 @@ import com.google.devtools.build.lib.exec.ExecutorBuilder;
 import com.google.devtools.build.lib.exec.ModuleActionContextRegistry;
 import com.google.devtools.build.lib.exec.SpawnStrategyRegistry;
 import com.google.devtools.build.lib.remote.RemoteServerCapabilities.ServerCapabilitiesRequirement;
+import com.google.devtools.build.lib.remote.common.CacheNotFoundException;
 import com.google.devtools.build.lib.remote.common.RemoteCacheClient;
 import com.google.devtools.build.lib.remote.common.RemoteExecutionClient;
 import com.google.devtools.build.lib.remote.downloader.GrpcRemoteDownloader;
+import com.google.devtools.build.lib.remote.http.HttpException;
 import com.google.devtools.build.lib.remote.logging.LoggingInterceptor;
 import com.google.devtools.build.lib.remote.options.RemoteBuildEventUploadMode;
 import com.google.devtools.build.lib.remote.options.RemoteOptions;
@@ -105,6 +107,7 @@ import io.reactivex.rxjava3.plugins.RxJavaPlugins;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.channels.ClosedChannelException;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
@@ -230,7 +233,26 @@ public final class RemoteModule extends BlazeModule {
               credentials,
               authAndTlsOptions,
               Preconditions.checkNotNull(env.getWorkingDirectory(), "workingDirectory"),
-              digestUtil);
+              digestUtil,
+              new RemoteRetrier(
+                 remoteOptions,
+                 (e) -> {
+                   boolean retry = false;
+                   if (e instanceof ClosedChannelException) {
+                     retry = true;
+                   } else if (e instanceof HttpException) {
+                     retry = true;
+                   } else if (e instanceof IOException) {
+                     String msg = e.getMessage().toLowerCase();
+                     if (msg.contains("connection reset by peer")) {
+                       retry = true;
+                     }
+                   }
+                   return retry;
+                 },
+                 retryScheduler,
+                 Retrier.ALLOW_ALL_CALLS)
+              );
     } catch (IOException e) {
       handleInitFailure(env, e, Code.CACHE_INIT_FAILURE);
       return;
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -115,6 +115,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
@@ -219,6 +220,24 @@ public final class RemoteModule extends BlazeModule {
     return capabilities;
   }
 
+  public static final Predicate<? super Exception> RETRIABLE_HTTP_ERRORS =
+          e -> {
+            boolean retry = false;
+            if (e instanceof ClosedChannelException) {
+              retry = true;
+            } else if (e instanceof HttpException) {
+              retry = true;
+            } else if (e instanceof IOException) {
+              String msg = e.getMessage().toLowerCase();
+              if (msg.contains("connection reset by peer")) {
+                retry = true;
+              } else if (msg.contains("operation timed out")) {
+                retry = true;
+              }
+            }
+            return retry;
+          };
+
   private void initHttpAndDiskCache(
       CommandEnvironment env,
       Credentials credentials,
@@ -236,22 +255,7 @@ public final class RemoteModule extends BlazeModule {
               digestUtil,
               new RemoteRetrier(
                  remoteOptions,
-                 (e) -> {
-                   boolean retry = false;
-                   if (e instanceof ClosedChannelException) {
-                     retry = true;
-                   } else if (e instanceof HttpException) {
-                     retry = true;
-                   } else if (e instanceof IOException) {
-                     String msg = e.getMessage().toLowerCase();
-                     if (msg.contains("connection reset by peer")) {
-                       retry = true;
-                     } else if (msg.contains("operation timed out")) {
-                       retry = true;
-                     }
-                   }
-                   return retry;
-                 },
+                 RETRIABLE_HTTP_ERRORS,
                  retryScheduler,
                  Retrier.ALLOW_ALL_CALLS)
               );
--- a/src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
@@ -20,24 +20,18 @@ import com.google.common.collect.ImmutableList;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
-import io.netty.handler.codec.http.DefaultFullHttpRequest;
-import io.netty.handler.codec.http.HttpContent;
-import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpHeaderValues;
-import io.netty.handler.codec.http.HttpMethod;
-import io.netty.handler.codec.http.HttpObject;
-import io.netty.handler.codec.http.HttpRequest;
-import io.netty.handler.codec.http.HttpResponse;
-import io.netty.handler.codec.http.HttpResponseStatus;
-import io.netty.handler.codec.http.HttpUtil;
-import io.netty.handler.codec.http.HttpVersion;
-import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http.*;
 import io.netty.handler.timeout.ReadTimeoutException;
 import io.netty.util.internal.StringUtil;
+
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.OptionalInt;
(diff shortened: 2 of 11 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -229,7 +232,26 @@ private void initHttpAndDiskCache(
...
+                   if (e instanceof ClosedChannelException) {
+                     retry = true;
+                   } else if (e instanceof HttpException) {
+                     retry = true;
+                   } else if (e instanceof IOException) {
+                     String msg = e.getMessage().toLowerCase();
+                     if (msg.contains("connection reset by peer")) {
+                       retry = true;
+                     }
+                   }
```
aherrmann, line 249: This list of exceptions is based on what we observed on CI, see https://github.com/digital-asset/daml/pull/11238 and https://github.com/digital-asset/daml/pull/11445. It may be that further rules should be added.

As far as I understand, `ClosedChannelException` can also occur in case of build failure when Bazel aborts the build. Is there a way to avoid retry in this case and only retry on legitimate network errors?

ulrfa:
Do you know the reason to the fetch failures / closed connections? Is there high package loss due to package congestion? Or overloaded server? Something else?

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
ulrfa: What kind of HttpException do you expect to catch and retry?

E.g. if server is overloaded and answer with 503 Service Unavailable, then it could make sense to do the opposite, back off and not retry, at least until what is specified by Retry-After header.

aherrmann: Sorry for the late reply.

An example that we [saw in our logs](https://github.com/digital-asset/daml/pull/11238#issue-1025314540) was
```
RETRYING: com.google.devtools.build.lib.remote.http.HttpException: 502 Bad Gateway
```

> E.g. if server is overloaded and answer with 503 Service Unavailable, then it could make sense to do the opposite, back off and not retry, at least until what is specified by Retry-After header.

AFAIK, `RemoteRetrier` currently uses an exponential backoff. So, it does already back-off. It's true that it wouldn't respect Retry-After. I'm not too familiar with the internals of `RemoteRetrier`. Does it support that kind of use-case?

aherrmann: Just yesterday I encountered a Bazel `HttpException` on a local build with this patch enabled:
```
RETRYING: com.google.devtools.build.lib.remote.http.HttpException: 503 Service Unavailable
Service Unavailable
```
The retry succeeded and the build continued from there on.

avdv: Maybe the retries should be limited to a specific number of status codes in case of an `HttpException`?

Does it even make sense to retry for *any* client error as resending the request will probably not change the response?! (e.g. for 400, 401, 403 or 405)

Otherwise, retrying on 503 Service Unavailable with the back-off mechanism of the `Retrier` seems reasonable to me, what do you think @ulrfa?

avdv:
Hi @sgowroji,

is there anything we should do / change with this PR to get it merged?

fmeum:
@brentleyjones @coeuvre Would one of you be available to review or suggest a reviewer? It would be really unfortunate if this effort stalled.

aherrmann:
@coeuvre We spoke at BazelCon, I mentioned this PR to you, and you asked me to ping you on it.
What would you say is missing and what's the best way to get this merged?

coeuvre (changes requested):
Thanks for the PR and sorry for the delay.

It seems that this PR tries to do two things:

1. Retry the download/upload if certain errors happened.
2. Upon retry of download, continue the last download at certain offset.

IMO, it's risky to combine these two within one PR. Can you split 2) into another PR?

Also, can you please add some tests for 1)?

aherrmann:
> It seems that this PR tries to do two things:
> 
>     1. Retry the download/upload if certain errors happened.
> 
>     2. Upon retry of download, continue the last download at certain offset.
> 
> 
> IMO, it's risky to combine these two within one PR. Can you split 2) into another PR?

Correct, it does these two things. Unfortunately, it doesn't seem to be possible to separate the two without much larger changes to the Bazel code-base as described in the PR description above:

> Some care needs to be taken due to the design of Bazel's internal remote
> cache client API. For a fetch the client is given an `OutputStream`
> object that it is expected to write the fetched data to. This may be a
> temporary file on disk that will be moved to the final location after
> the fetch completed. On retry, we need to be careful to not duplicate
> previously written data when writing into this `OutputStream`. **Due to
> the generality of the `OutputStream` interface we cannot reset the file
> handle or write pointer to start fresh.** Instead, this change follows the
> same pattern used in the GRPC cache client. Namely, keep track of the
> data previously written and continue from that offset on retry.

Please let me know if you see a good way to separate the two concerns.

---

> Also, can you please add some tests for 1)?

Without being able to separate 1) and 2) (as described above), could you give some pointers toward this question from the original PR message:

> If anyone has any pointers on how to best test this in the Bazel code base please let me know. One option I briefly explored was to add a test-case to `src/test/java/com/google/devtools/build/lib/remote/http/HttpCacheClientTest.java`. But, that seemed to require implementing support for HTTP range requests in `src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/http/HttpCacheServerHandler.java`. Perhaps there is a better way to test this change.

coeuvre:
> Unfortunately, it doesn't seem to be possible to separate the two without much larger changes to the Bazel code-base as described in the PR description above:

Sorry,  I missed that part. Yes, you are right. Then I am ok to merge them at once if we have enough test coverage.

> Without being able to separate 1) and 2) (as described above), could you give some pointers toward this question from the original PR message:

I think the best place for the tests are in `src/test/java/com/google/devtools/build/lib/remote/http/HttpCacheClientTest.java`. It's not necessary to implement range requests for `HttpCacheServerHandler`. Using mocks is fine as long as we are able to cover the edge cases added in the client code.

ulrfa:
> > Instead, this change follows the
> > same pattern used in the GRPC cache client. Namely, keep track of the
> > data previously written and continue from that offset on retry.

The GRPC cache client seems to do that for CAS downloads, but not when retrieving AC, right? And this PR seems to do it also for AC, right?

I think it is safe to do for CAS where the key represent a specific content, but a retried download of AC key could result in different ActionResult. And concatenating chunks from different ActionResult is not reliable. Do you have ideas about how to address that?

src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
```diff
@@ -159,16 +186,36 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable t) {
...
+  private HttpRequest buildRequest(String path, String host, long offset) {
     HttpRequest httpRequest =
         new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, path);
     httpRequest.headers().set(HttpHeaderNames.HOST, host);
     httpRequest.headers().set(HttpHeaderNames.CONNECTION, HttpHeaderValues.KEEP_ALIVE);
     httpRequest.headers().set(HttpHeaderNames.ACCEPT, "*/*");
     httpRequest.headers().set(HttpHeaderNames.ACCEPT_ENCODING, HttpHeaderValues.GZIP);
+    if (offset != 0) {
+      httpRequest.headers().set(HttpHeaderNames.RANGE, String.format("%s=%d-", HttpHeaderValues.BYTES, offset));
+    }
```
ulrfa, line 198: How is the RANGE implementation in this PR working in combination with the also supplied `ACCEPT-ENCODING: gzip` header?

It seems this PR expects ranges to be applied to the uncompressed data, right? but RFC 9110 states: _“If the representation data has a content coding applied, each byte range is calculated with respect to the encoded sequence of bytes, not the sequence of underlying bytes that would be obtained after decoding.“_

For the gPRC cache protocol, there were lots of discussions and considerations in https://github.com/bazelbuild/remote-apis/pull/168 about combining compression with read offsets. E.g. regarding if offsets should be applied to the compressed or non-compressed stream, regarding when offsets don’t align with stored compressed chunk sizes on the server side, etc.

How often do you expect retries of already partially downloaded files? **Would it be sufficient to always re-request the whole file and simply skip already received CAS chunks after receiving them again?** Would that avoid much of the complexities in both code and test cases?

[Bazel-remote](https://github.com/buchgr/bazel-remote) supports zstd compression for both the gRPC and HTTP protocols. Bazel is still only supporting it for gRPC. I’m afraid the RANGE implementation part of this PR would make it harder for future work of adding support also for `ACCEPT-ENCODING: zstd` to the HTTP cache client.

aherrmann: Correct, the offset is calculated on uncompressed bytes.

> Bazel is still only supporting it for gRPC. I’m afraid the RANGE implementation part of this PR would make it harder for future work of adding support also for ACCEPT-ENCODING: zstd to the HTTP cache client.

I see, fair enough. Perhaps a solution that takes encoding into account could introduce a new step [into the download pipeline before the inflater](https://github.com/bazelbuild/bazel/blob/5e763b4bd25a7a9991953b79beeaef4852f77c5e/src/main/java/com/google/devtools/build/lib/remote/http/HttpCacheClient.java#L399) to determine the correct download offset in light of encoding.

> How often do you expect retries of already partially downloaded files? **Would it be sufficient to always re-request the whole file and simply skip already received CAS chunks after receiving them again?** Would that avoid much of the complexities in both code and test cases?

The failures were sporadic, usually a handful in one CI pipeline. So, this may well be good enough. That would help make the tests simpler as the mock server wouldn't have to worry about range requests. I'll try this route.

aherrmann:
> I think it is safe to do for CAS where the key represent a specific content, but a retried download of AC key could result in different ActionResult. And concatenating chunks from different ActionResult is not reliable. Do you have ideas about how to address that?

@ulrfa I think that's possible. The blob download takes an external output stream, so that's where we don't have control and need to continue from the offset. But, the AC download controls the output stream, so, if we perform the retry at a higher level, then we'd be able to do a fresh download.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -233,7 +236,28 @@ private void initHttpAndDiskCache(
...
+                     retry = true;
+                   } else if (e instanceof IOException) {
+                     String msg = e.getMessage().toLowerCase();
+                     if (msg.contains("connection reset by peer")) {
+                       retry = true;
+                     } else if (msg.contains("operation timed out")) {
+                       retry = true;
+                     }
+                   }
+                   return retry;
```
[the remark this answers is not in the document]
aherrmann: I can factor it out. Unfortunately, I cannot place it next to the grpc error predicate as that causes a dependency cycle:
* `//.../remote/http:http` depends on `//.../remote:Retrier` (due to the retry logic in `HttpCacheClient`).
* `//.../remote:Retrier` depends on `//.../remote/http:http` (due to the dependency on `HttpException`).

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
aherrmann: Good point, I've added logic to only retry on certain HTTP errors.

aherrmann:
@coeuvre @ulrfa @avdv Thank you for the review!

I’ve changed the retry logic such that retries of `ActionResult` fetches start a new download from scratch and write into a fresh buffer, whereas CAS downloads skip the previously downloaded prefix and write to the existing output stream from the offset.

I’ve removed RANGE requests. As pointed out above the implementation would not have handled compressed streams correctly. RANGE requests could still be introduced later on, but they could be seen as an optimization over this PR to avoid sending redundant bytes over the wire, and as an improvement in that fewer bytes to send should imply a lower likelihood of repeated intermittent failure.

I’ve factored out the logic deciding whether to retry into `RETRIABLE_HTTP_ERRORS`. And changed it to only retry on certain HTTP error codes instead of all `HttpException`s.

I’ve added tests for the retry mechanism on `HttpDownloadHandler` and `HttpCacheClient`.

I’ve also tested the changes against [a test-setup](https://github.com/aherrmann/bazel-remote-retry-testing) and observed retries occurring as expected.

coeuvre (approved):
Thanks for the update! I agree that we should leave RANGE for another PR.

I will import after @ulrfa has reviewed.

src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
```diff
@@ -51,6 +45,10 @@ final class HttpDownloadHandler extends AbstractHttpHandler<HttpObject> {
   private long contentLength = -1;
   /** the path header in the http request */
   private String path;
+  /** the offset at which to download */
+  private long offset;
+  /** the bytes to skip in a full or chunked response */
+  private OptionalInt skipBytes;
```
ulrfa, line 51: What is the reason for having both `offset` and `skipBytes` as fields? Could the code be simplified by removing the `offset` field and instead initialize `skipBytes` where `offset` is currently initialized?

ulrfa:
Thanks for the update @aherrmann! I think the logic is fine and good test cases! Now I only have the question about the `offset` field, but that is only a minor implementation detail.

src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
aherrmann: True, one field is enough. I've dropped `offset` in favor of `skipBytes`.

coeuvre (approved):

src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
```diff
@@ -137,6 +142,7 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
     DownloadCommand cmd = (DownloadCommand) msg;
     out = cmd.out();
     path = constructPath(cmd.uri(), cmd.digest().getHash(), cmd.casDownload());
+    skipBytes = (int)cmd.offset();
```
ulrfa, line 145: What happens if the long offset is larger than Integer.MAX_VALUE?

Would it make sense to store skipBytes as a long field and instead around line 105 set skipNow=Integer.MAX_VALUE if skipBytes > Integer.MAX_VALUE?

ulrfa (changes requested):
Thanks @aherrmann!

I did not notice the long to int cast in the previous commit, but noticed it now when it moved to other place.  :smiley:

src/main/java/com/google/devtools/build/lib/remote/http/HttpDownloadHandler.java
aherrmann: Good catch! IIRC I chose `long` for `offset` because `contentLength` is also `long`, but then found that `content.readableBytes` is `int`, so had to convert. But, you're right that this is not the right way to do that conversion.
I've changed `skipBytes` to be `long` and moved the conversion into the branch where `skipBytes < readableBytes`. Since `readableBytes` is `int` it's safe to convert in that case.

ulrfa (approved):
Good work @aherrmann!

aherrmann:
@coeuvre I see this is labeled as `awaiting-PR-merge` but not yet merged. Is it stuck on anything internally? Anything I can help with to get this merged?

sgowroji:
Hi @aherrmann,  Above PR merging is in progress. @coeuvre is merging the above PR.

PR #14262, opened 2021-11-11
Refactor system suspend event handling.

dmaclach:
We are looking to add more build anomaly reporting, so refactor the current system suspend handling
to match the new model where we have a module and an event that we can record.

Note this deprecates the AnomalyRecord out of BES for the time being.


dmaclach:
Applying via standard google process.

PR #14268, opened 2021-11-12
Add --experimental_repository_cache_urls_as_default_canonical_id to help detect broken repository URLs

### LINKED ISSUE #14128: Flag to include URLs in repository cache key when canonical_id not specified

It would be nice to have a flag, e.g. `--experimental_use_urls_as_default_canonical_id`, which includes the URLs in the repository cache key if `canonical_id` isn't set manually in the call to `download`/`download_and_extract`. This can be used in a CI system to verify that invalid repository updates don't get checked in.

It seems that users frequently run into issues with repository rules using cached downloads based on SHA256 despite some other detail (especially URLs) being changed and making it invalid. This isn't noticed until building without that cache entry. See #10630, #10612, #10353, #5144.

The way to prevent this is to use `canonical_id` to force redownloading, for example by setting it to the URL. The problem is that requiring developers to set `canonical_id` is prone to error. You could make custom wrappers for e.g. `http_file` to automatically use the URL as `canonical_id`, but in a large codebase it's possible this will be accidentally circumvented with the upstream rule definition.

macOS 11.6

```
release 4.2.1
```

kmicklas:
This new flag can be used to force redownloading when repository URLs are changed. Otherwise, it's possible broken URLs can be masked by the presence of a repository cache entry with the same hash. Specifying a `canonical_id` works as before, overriding this behavior.

Closes #14128.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
@@ -275,6 +275,7 @@ public class BazelRepositoryModule extends BlazeModule {
       if (repoOptions.repositoryDownloaderRetries >= 0) {
         downloadManager.setRetries(repoOptions.repositoryDownloaderRetries);
       }
+      downloadManager.setUrlsAsDefaultCanonicalId(repoOptions.urlsAsDefaultCanonicalId);
 
       repositoryCache.setHardlink(repoOptions.useHardlinks);
       if (repoOptions.experimentalScaleTimeouts > 0.0) {
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/RepositoryOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/RepositoryOptions.java
@@ -215,6 +215,19 @@ public class RepositoryOptions extends OptionsBase {
               + " to escalate it to a resolution failure.")
   public CheckDirectDepsMode checkDirectDependencies;
 
+  @Option(
+      name = "experimental_repository_cache_urls_as_default_canonical_id",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.BAZEL_CLIENT_OPTIONS,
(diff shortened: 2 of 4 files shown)


linzhp (approved):

davido:
I tested this PR and it works as expected, see my comment in the corresponding issue.

Code looks sane. Would it be feasible to add a test?
Looking into the code base:

```bash
  $ cd src/test/java && git grep -i CanonicalId
com/google/devtools/build/lib/bazel/repository/cache/RepositoryCacheTest.java:  public void testCanonicalId() throws Exception {
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:              "testCanonicalId",
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:              "testCanonicalId",
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:              "testCanonicalId",
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:            "testCanonicalId",
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:                "testCanonicalId",
com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java:            "testCanonicalId",
com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java:    final String canonicalId = "";
com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloaderTest.java:        canonicalId,
```

I see a number of occurrences of `CanonicalId` in test code.

kmicklas:
I'll take a look at adding a test on Monday, thanks!

meteorcloudy:
Thanks!
An integration test can be added here: https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/bazel_repository_cache_test.sh

kmicklas:
I added an integration test for the new flag and the default behavior.

meteorcloudy (approved):
Thanks!

PR #14274, opened 2021-11-14
Fix typo in example on testing.md.

cfredric:
The example says that `foo` should correspond to `provider_contents` earlier in the text.

### DIFF
--- a/site/docs/skylark/testing.md
+++ b/site/docs/skylark/testing.md
@@ -170,7 +170,7 @@ Then:
 *   its test rule type should be named `foo_test` (`provider_contents_test`)
 
 *   the label of the target of this rule type should be `foo`
-    (`provider_contents_test`)
+    (`provider_contents`)
 
 *   the implementation function for the testing rule should be named
     `_foo_test_impl` (`_provider_contents_test_impl`)
--- a/site/docs/skylark/testing.md
+++ b/site/docs/skylark/testing.md
@@ -169,8 +169,8 @@ Then:
 
 *   its test rule type should be named `foo_test` (`provider_contents_test`)
 
-*   the label of the target of this rule type should be `foo`
-    (`provider_contents`)
+*   the label of the target of this rule type should be `foo_test`
+    (`provider_contents_test`)
 
 *   the implementation function for the testing rule should be named
     `_foo_test_impl` (`_provider_contents_test_impl`)


gregestren:
I don't have domain expertise on this, but note https://github.com/bazelbuild/bazel/commit/5a74742d83f1ff68e4d019eca6e4a0643a8eb2cb switched this in the other direction. I'll see if I can CC the original contributor for input.

gregestren:
[@sfreilich](https://github.com/sfreilich) - what do you think?

cfredric:
That commit looks like a big one focused on formatting; I'd be surprised if the switch of `provider_contents` to `provider_contents_test` in that commit was intentional.

gregestren:
So reviewing some more, it looks like the `foo` part is wrong, not the actual name.

In other words, the current text is:

```the label of the target of this rule type should be foo (provider_contents_test)```

If you look at the code it's referencing:

```
# Testing rule.
    provider_contents_test(name = "provider_contents_test",
  ...
```

That's an example of a `foo_test` name (`provider_contents_test`).

What do you think?

gregestren:
When you said "earlier in the text", which text specifically did you mean?

cfredric:
I'm referring to the following text:

    For consistency, follow the recommended naming convention: Let `foo` stand for
    the part of the test name that describes what the test is checking
    (`provider_contents` in the above example).

starting on [line 160](https://github.com/bazelbuild/bazel/blob/master/site/docs/skylark/testing.md?plain=1#L160-L162).

gregestren:
Thanks. So my reading of the whole text is:

- `foo` -> `provider_contents`
- macro that generates the test: `def _test_foo` -> `def _test_provider_contents():`
- test rule type: `foo_test` -> `provider_contents_test = analysistest.make(_provider_contents_test_impl)`
- label of the target: `foo_test` -> `provider_contents_test(name = "provider_contents_test",`
- implementation function: `_foo_test_impl` -> `def _provider_contents_test_impl(ctx):`
- label of targets under test: `foo_*` -> `target_under_test = ":provider_contents_subject")`

So the part of the text reading `the label of the target of this rule type should be foo` is incorrect: there's no label in the example code matching `foo`.

cfredric:
Hmm, you're probably right; it wouldn't make much sense to have a target named `foo` whose rule type is `foo_test`. In that case, I think the bullet should be changed to:

    * the label of the target of this rule type should be `foo_test` (`provider_contents_test`)'

WDYT?

gregestren:
Yes, I think that'd work!

gregestren (approved):

Issue #14273, opened 2021-11-14
C++ rules broken on macOS 12 Monterey and command line developer tools


nresare:
### Description of the problem / feature request:

There are mainly two ways to install the basic apple provided clang compiler toolchain on macOS. Either you install the Command Line developer tools by invoking `xcode-select --install`, or you install the full XCode development environment, which includes a graphical IDE and tooling for application development on all of Apple's operating systems. The latter is a significantly larger download at >10GiB.

Bazel 4.2.1 will detect and use the C++ compiler and linker that is available through the command line tools, but as of macOS 12 Monterey, attempting to build C++ binaries will succeed but the resulting binaries will fail on startup with a Segmentation Fault error. This is because the linker invocation generating the binary is not invoked with the -lc++ flag, indicating that the libc++ library should be loaded by the dynamic linker on startup.

The [suggested solution](https://github.com/bazelbuild/bazel/issues/14206#issuecomment-957356014) to this problem is to install the full XCode IDE, which apparently works around this problem, but it would be neat to have this work out-of-the-box

### Feature requests: what underlying problem are you trying to solve with this feature?

Make bazel more user friendly by having it build functional c++ binaries without non-obvious workarounds such as installing XCode on a pretty standard setup on one of the more common platforms.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. On a vanilla macOS 12 Monterey machine, install the command line devloper tools with `xcode-select --install`
2. download the latest stable bazel (I ran bazelisk that pulled bazel 4.2.1)
3. Clone the [bazelbuild/examples](https://github.com/bazelbuild/examples) repository 
4. and build the [hello world example](https://github.com/bazelbuild/examples/tree/main/cpp-tutorial/stage1):
4.1 `cd cpp-tutorial/stage1`
4.2 `bazelisk build //main:hello-world`
5. Attempt to execute the resulting binary: `./bazel-bin/main/hello-world`

```shell
% ./bazel-bin/main/hello-world
zsh: segmentation fault  ./bazel-bin/main/hello-world
```

### What operating system are you running Bazel on?

I have reproduce this with macOS 12.0.1 Monterey on an Apple Silicon M1 MacBook as well as an Intel MacBook Pro.
```shell
% gcc --version
[...]

Issue #14216, opened 2021-11-02
bazel-built protoc segfaults on mac os monterey


pcj:
Users that have upgraded their operating system to `12.0.1` and have a dependency on @com_google_protobuf//:protoc are experiencing segfaults when running the `protoc` tool.

See https://github.com/protocolbuffers/protobuf/issues/8884 for context.

keith:
Probably related https://github.com/bazelbuild/bazel/issues/14206 sounds like user's tools setup might be the issue

Wyverald:
FWIW, I can no longer build Bazel itself on macOS Monterey -- exact same issue here: `protoc` segfaults.

@susinmotion any ideas?

Wyverald:
A snippet of the error message:

```
ERROR: /private/var/tmp/_bazel_wyv/01cd51505ca764e566d7f93a5254e0db/external/googleapis/BUILD.bazel:180:14: Generating Java (Immutable) proto_library @googleapis//:google_bytestream_bytestream_proto failed: (Segmentation fault): protoc failed: error executing command bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/com_google_protobuf/protoc '--proto_path=external/googleapis' ... (remaining 10 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Target //src:bazel_nojdk failed to build
```

demorgan104:
@Wyverald have you tried cleaning the cache ... ? I could build protoc in the end using the steps I mentioned in my issue...

susinmotion:
Sadly I don't have any wisdom to add here :(
It sounds like this is an issue with using Monterrey without Xcode 13.0? It would be nice to have a less bad user experience, but I do hope that it'll be come common knowledge that Xcode 13 is required.

dtbartle:
I don't understand why, but it looks like the difference (between full xcode and the CLI tools) is how the binary is linked (`otool -L`:
```
xcode:
bazel-bin/protoc:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1200.3.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1853.0.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)

tools:
bazel-bin/protoc:
[...]

Issue #14013, opened 2021-09-20
Release 5.0 - January 2022


Wyverald:
# Status of Bazel 5.0

- Target baseline: 2021-10-25
- Expected release date: 2022-01-19
- [5.0 release blockers](https://github.com/bazelbuild/bazel/milestone/31)

To report a release-blocking bug, please add it to the 5.0 release blocker milestone, and cc me.

Task list:

- [x] Pick release baseline: 8d66a4171baddcbe1569972f019e54130111202c
- [x] Create release candidate: 5.0.0rc1
- [x] Check downstream projects:
- [x] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [x] Send for review the release announcement PR: https://github.com/bazelbuild/bazel-blog/pull/263
- [x] Push the release, notify package maintainers: https://github.com/bazelbuild/bazel/issues/14013#issuecomment-1016560372
- [x] Update the documentation
- [x] Push the blog post
- [x] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

aaliddell:
Would you consider looking at https://github.com/bazelbuild/bazel/issues/7386 for Bazel 5? This was missed as a change on Bazel 4 and has been tripping people up for a long time. The proposed steps to change the default were begun with the addition of an incompatible flag but have got stuck there. Perhaps we can now move to flipping the default to disabling the creation of `__init__.py` files?

But perhaps this is too short notice now?

danieljanes:
...and (related to @aaliddell's comment above): #12238

Wyverald:
Thanks for flagging this; I'll check with @brandjon.

Although yeah, the purpose of this thread is more on the side of making sure any in-flight changes can make it into 5.0, rather than start prioritizing stuff now. Nevertheless, I see that this has been a long-standing issue, so let's see if we have the cycles to make it happen. (Alternatively, feel free to send PRs if you have the time!)

Wyverald:
re: `__init__.py`, please see @brandjon's update at https://github.com/bazelbuild/bazel/issues/7386#issuecomment-924320303

matthewjh:
What about #13879?

Wyverald:
I'll check with @coeuvre.

philwo:
Please add (potential) release blockers to this milestone: https://github.com/bazelbuild/bazel/milestone/31

[...]

PR #14275, opened 2021-11-14
Enable user_link_flags_feature for macosx cc_toolchain_config

nresare:
This will cause the values passed in the link_libs parameters
to not get silently dropped on macOS 12 Monterey, and fix
the problem where the command line dev tools toolchain
will output C++ binaries fails to start with a segmentation
fault error.

Fixes: #14273

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -1243,6 +1243,7 @@ def _impl(ctx):
             coverage_feature,
             default_compile_flags_feature,
             default_link_flags_feature,
+            user_link_flags_feature,
             fdo_optimize_feature,
             supports_dynamic_linker_feature,
             dbg_feature,


keith:
@nresare looks like you need to sign the CLA here

nresare:
@googlebot I signed it!

keith (approved):
I can't reproduce this, I assume because I installed Xcode which altered the system some way. But This not being here seems like an oversight anyways so it seems like a good fix

nresare:
@google-cla I signed it!

keith:
@gregestren since you were triaging related issues maybe you can help get this merged? It should fix https://github.com/bazelbuild/bazel/issues/14216

// @oquenchil

EdSchouten (approved):

EdSchouten:
Maybe worth cherry-picking this onto the 4.2 branch and releasing a 4.2.2? Seems pretty serious if you ask me.

gregestren:
@keith happy to but since I'm not a domain expert here my main role is going to be routing.

I assigned to the Rules / CPP team and @comius team lead according to triage policy.

Also CC'ing @meteorcloudy and @Wyverald who are listed at https://github.com/bazelbuild/bazel/issues/14013 re: qualifying for a 4.2 branch.

Issue #14287, opened 2021-11-17
Bazel HEAD OOMs during analysis phase when JavaPluginInfo provides Java outputs information


Bencodes:
### Description of the problem / feature request:

Bazel HEAD runs out of memory during the analysis phase of a build.

I did a git bisect tracking the issue back to this commit https://github.com/bazelbuild/bazel/commit/f73e28cd44d439e6bcecd73f45813ca13c352a93 as the culprit.

The stacktrace that eventually comes out:
```
(19:11:00) FATAL: bazel ran out of memory and crashed. Printing stack trace:
net.starlark.java.eval.Starlark$UncheckedEvalError: OutOfMemoryError thrown during Starlark evaluation (//instant-features/rider/last-mile/fieldwork/stationdetail/plugins:lib_kt)
        at <starlark>.compile(<builtin>:0)
        at <starlark>._run_kt_java_builder_actions(/private/var/tmp/_bazel_blee/499a001013731d09bffd82f8601a3161/external/io_bazel_rules_kotlin/kotlin/internal/jvm/compile.bzl:780)
        at <starlark>.kt_jvm_produce_jar_actions(/private/var/tmp/_bazel_blee/499a001013731d09bffd82f8601a3161/external/io_bazel_rules_kotlin/kotlin/internal/jvm/compile.bzl:517)
        at <starlark>.kt_jvm_library_impl(/private/var/tmp/_bazel_blee/499a001013731d09bffd82f8601a3161/external/io_bazel_rules_kotlin/kotlin/internal/jvm/impl.bzl:235)
Caused by: java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.Arrays.copyOf(Unknown Source)
        at com.google.common.collect.ImmutableList.asImmutableList(ImmutableList.java:371)
        at com.google.common.collect.ImmutableList$Builder.build(ImmutableList.java:871)
        at com.google.devtools.build.lib.rules.java.JavaPluginInfo.merge(JavaPluginInfo.java:179)
        at com.google.devtools.build.lib.rules.java.JavaInfoBuildHelper.mergeExportedJavaPluginInfo(JavaInfoBuildHelper.java:238)
        at com.google.devtools.build.lib.rules.java.JavaInfoBuildHelper.createJavaCompileAction(JavaInfoBuildHelper.java:298)
        at com.google.devtools.build.lib.rules.java.JavaStarlarkCommon.createJavaCompileAction(JavaStarlarkCommon.java:118)
[...]

PR #14288, opened 2021-11-17
Remove merging of java_outputs in JavaPluginInfo.

comius:
JavaPluginInfo merging of java_outputs caused a significant regression when a large number of plugins is present.

This value is not used by native code and it does not get exposed to Starlark. The only use case for java_outputs is IDE integrations, where the value is used from a single java_plugin target.

Fixes https://github.com/bazelbuild/bazel/issues/14287


Bencodes (approved):

Bencodes:
Confirmed that this does reduce the heap significantly. https://github.com/bazelbuild/bazel/issues/14287#issuecomment-971316153

PR #14300, opened 2021-11-19
Fix remote spawn tests for remote_merkle_tree_cache=true

moroten:
When setting `--experimental_remote_merkle_tree_cache=true` as default and running the test suite, some errors became visible. This change fixes those errors.


coeuvre:
This is rolled back in https://github.com/bazelbuild/bazel/commit/4256d46327bad8638df91be1a5d4ef83b12b74c7 because it breaks some internal tests.

meteorcloudy:
@coeuvre Hmm, should we also rollback this in the 5.0.0rc2? Or are we going to rollfoward with a fix?

moroten:
This PR was not meant for 5.0, only for master. It is a preparation for potential default activation of the Merkle tree caching feature, but I'll leave it as is because I can't fix the Google internal tests.

PR #14302, opened 2021-11-19
Delete marker file before fetching an external repository

### LINKED ISSUE #8993: Investigate behavior of interrupted execution of long-to-execute repository rules
Copying the link and text of original [description:]

(https://github.com/bazelbuild/rules_nodejs/issues/802#issuecomment-515061871)

Canceling a build at its loading phase will also trigger this.

The first build:

```
Loading:
Loading: 0 packages loaded
Loading: 0 packages loaded
Loading: 0 packages loaded
# Received cancellation signal, interrupting
```

The second build:

```
Loading:
Loading: 0 packages loaded
ERROR: error loading package '': in /var/lib/buildkite-agent/builds/buildkite-agents-preemptible-nzjp-2/<redacted>/javascript/npm.bzl: Every .bzl file must have a corresponding package, but '@npm//:install_bazel_dependencies.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
ERROR: error loading package '': in /var/lib/buildkite-agent/builds/buildkite-agents-preemptible-nzjp-2/<redacted>/javascript/npm.bzl: Every .bzl file must have a corresponding package, but '@npm//:install_bazel_dependencies.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
INFO: Elapsed time: 0.200s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
INFO: Build Event Protocol files produced successfully.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

My speculation is that repository_ctx doesn't have a way to tell if a repository has been fully loaded or not, so it will use a corrupted external repository without loading it again.

fmeum:
Fetching a repository is a long-running operation that can easily be
interrupted. If it is and the marker file exists on disk, a new
evaluation of the RepositoryDelegatorFunction may treat this repository
as valid even though it is in an inconsistent state.

Clearing the marker file before initiating the fetch and only recreating
it after the fetch is complete prevents this scenario.

Fixes #8993.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/repository/RepositoryDelegatorFunction.java
@@ -339,6 +339,11 @@ public final class RepositoryDelegatorFunction implements SkyFunction {
     }
 
     if (isFetch.get()) {
+      // Fetching a repository is a long-running operation that can easily be interrupted. If it is
+      // and the marker file exists on disk, a new call of this method may treat this repository as
+      // valid even though it is in an inconsistent state. Clear the marker file and only recreate
+      // it after fetching is done to prevent this scenario.
+      DigestWriter.clearMarkerFile(directories, repositoryName);
       // Fetching enabled, go ahead.
       RepositoryDirectoryValue.Builder builder =
           fetchRepository(skyKey, repoRoot, env, digestWriter.getMarkerData(), handler, rule);


philwo:
LGTM, but let's ask Yun, too! :D

meteorcloudy:
Hmm, I don't think this is necessary. Because:
 - If `fetchRepository` is executed, then that means the existing marker file isn't consistent anymore.
 - When `fetchRepository` is interrupted, the marker file shouldn't be rewritten because we do that after `fetchRepository` is done.
 - Therefore, the next time you execute this SkyFunction, the marker data should still be inconsistent and `fetchRepository` should be re-executed.

Do you have a minimal reproducible case that we can look into?

fmeum:
> Hmm, I don't think this is necessary. Because:
> 
> * If `fetchRepository` is executed, then that means the existing marker file isn't consistent anymore.
> * When `fetchRepository` is interrupted, the marker file shouldn't be rewritten because we do that after `fetchRepository` is done.

This is correct.

> * Therefore, the next time you execute this SkyFunction, the marker data should still be inconsistent and `fetchRepository` should be re-executed.

As far as I can tell, this is not true when dealing with a managed directory: The marker file only tracks the *existence* of the managed directory, not its contents. If the `SkyFunction` is interrupted at a point where the managed directory has been created but its contents aren't consistent with what the ruleset expects, it will not rerun `fetchRepository`.

Every project using `rules_nodejs`'s `yarn_install` rule should serve as a reproducer. For example, try the following:
1. `git clone https://github.com/bazelbuild/rules_nodejs`
2. `cd cd examples/react_webpack`
3. `bazel build //...`
4. `rm -rf node_modules`
5. `bazel build //...`, interrupting the build soon after it shows `Fetching @npm; Running yarn install on //:package.json`. If your internet connection or local yarn cache is too fast, this might require adding new and large dependencies to `package.json` so that you have a window to interrupt the execution of the repository rule.
6. `bazel build //...` should fail with an error such as
```
ERROR: error loading package '': Every .bzl file must have a corresponding package, but '@npm//webpack-cli:index.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
```

At that point, the build failure is persistent and can only be resolved by deleting `node_modules` (or running `bazel clean --expunge`).

meteorcloudy:
I see, that makes sense now, basically unfinished `fetchRepository` could accidentally cause the previous marker data to be considered consistent due to the use of managed directories.

meteorcloudy (approved):
Thanks for the fix and explaination!

brentleyjones:
This is a nice little bug fix. Think we can cherry-pick it into 5.0?

fmeum:
> This is a nice little bug fix. Think we can cherry-pick it into 5.0?

I submitted it for both 5.0.0 and 4.2.2.

PR #14310, opened 2021-11-23
Support select() on constraint_value for aliases.

gregestren:
This implements approach #4 of
https://github.com/bazelbuild/bazel/issues/13047#issuecomment-805309450.

The basic change adds a new toolchain resolution mode: "resolve iff the target has a select()". It then sets alias() to that mode.

We could remove this special casing if we ever ubiquitously provide platform info to *all* rules (https://github.com/bazelbuild/bazel/issues/12899#issuecomment-767759147).

RELNOTES: alias() can now select() directly on constraint_value()

Fixes https://github.com/bazelbuild/bazel/issues/13047.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
@@ -221,6 +221,19 @@ public class RuleClass {
     ENABLED,
     /** The rule should not use toolchain resolution. */
     DISABLED,
+    /**
+     * The rule instance uses toolchain resolution if it has a select().
+     *
+     * <p>This is for rules that don't intrinsically use toolchains but have select()s on {@link
+     * com.google.devtools.build.lib.rules.platform.ConstraintValue}, which are part of the build's
+     * platform. Such instances need to know what platform the build is targeting, which Bazel won't
+     * provide unless toolchain resolution is enabled.
+     *
+     * <p>Ultimately, we should remove this when <a
+     * href="https://github.com/bazelbuild/bazel/issues/12899#issuecomment-767759147}#12899</a>is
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
```diff
@@ -221,6 +221,19 @@ public int getPathSegment() {
...
+     * <p>This is for rules that don't intrinsically use toolchains but have select()s on {@link
+     * com.google.devtools.build.lib.rules.platform.ConstraintValue}, which are part of the build's
+     * platform. Such instances need to know what platform the build is targeting, which Bazel won't
+     * provide unless toolchain resolution is enabled.
+     *
+     * <p>Ultimately, we should remove this when <a
+     * href="https://github.com/bazelbuild/bazel/issues/12899#issuecomment-767759147}#12899</a>is
+     * addressed, so platforms are unconditionally provided for all rules.
+     */
+    HAS_SELECT,
```
katre, line 236: I had initially thought this was checking the Target (in ConfiguredTargetFunction) to see if any selects were available, not that we were hard-coding this for `alias` only. I'm fine with this choice, but add a comment here clarifying that this doesn't do any inspection, it's still set statically on a rule.

katre (changes requested):
LGTM, once issue.

src/main/java/com/google/devtools/build/lib/packages/RuleClass.java
gregestren: Done.

I'm open to even following it the way you suggest.I just wanted to be conservative in how much we expand this functionality given the law of unpredictable consequences for large diffs.

katre (approved):

Issue #13861, opened 2021-08-17
Bazel cquery fails if paths are incompatible for platform


kwc:
### Description of the problem / feature request:

Bazel `cquery` is a useful command for finding build dependencies you wish to avoid/remove in a specific configuration.

If you use Bazel `cquery` to find `somepath` or `allpaths` to a incompatible target, it will error with:

```
bazel --nohome_rc cquery  'allpaths(//foo:bar, //foo:baz)'
```
```
ERROR: Target //foo:baz is incompatible and cannot be built, but was explicitly requested.
```

As this is failing in a `bazel build` step, it actually does produce a relevant result, but accidentally, and if there are no query results, it simply returns an error instead of 0 results.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Using the following trivial `BUILD.bazel`, with one incompatible target:
```
sh_test(
    name = "foo",
    srcs = ["BUILD.bazel"],
)

sh_test(
    name = "bar",
    srcs = ["BUILD.bazel"],
)

sh_test(
    name = "baz",
    srcs = ["BUILD.bazel"],
    target_compatible_with = ["@platforms//:incompatible"],
)
```

A `cquery` between the targets shows how `cquery` will `ERROR` if the target is an incompatible target.
```
$ bazel cquery 'somepath(:foo, :bar)'
INFO: Analyzed 2 targets (1 packages loaded, 430 targets configured).
INFO: Found 2 targets...
INFO: Empty query results
INFO: Elapsed time: 0.770s
INFO: 0 processes.
INFO: Build completed successfully, 0 total actions
$ bazel cquery 'somepath(:foo, :baz)'
ERROR: Target //:baz is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //:baz   <-- target platform didn't satisfy constraint @platforms//:incompatible
INFO: Elapsed time: 0.243s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 3 targets configured)
```

Repeating the same command with `bazel query`:
```
$ bazel query 'somepath(:foo, :baz)'
INFO: Empty results
$ bazel query 'somepath(:foo, :bar)'
INFO: Empty results
Loading: 0 packages loaded
```

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

`release 4.1.0`

###  Have you found anything relevant by searching the web?

No

[...]

PR #14311, opened 2021-11-23
cquery: support queries over incompatible targets.

gregestren:
The implementation is a simple matter of adding an extra "should we report incompatible errors?" bit to the build request. Keep it true on builds, false on cquery.

Fixes https://github.com/bazelbuild/bazel/issues/13861.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
@@ -434,8 +434,7 @@ public class BuildView {
               topLevelTargetsWithConfigsResult,
               /*includeExecutionPhase=*/ true);
     } else {
-      Set<ConfiguredTarget> targetsToSkip;
-
+      Set<ConfiguredTarget> targetsToSkip = ImmutableSet.of();
       if (reportIncompatibleTargets) {
         TopLevelConstraintSemantics topLevelConstraintSemantics =
             new TopLevelConstraintSemantics(
@@ -462,8 +461,6 @@ public class BuildView {
                         skyframeAnalysisResult.getConfiguredTargets()),
                     platformRestrictions.targetsToSkip())
                 .immutableCopy();
-      } else {
-        targetsToSkip = ImmutableSet.of();
       }
 
       result =
--- a/src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
@@ -202,6 +202,7 @@ public class BuildView {
       boolean checkForActionConflicts,
       int loadingPhaseThreads,
       TopLevelArtifactContext topLevelOptions,
+      boolean reportIncompatibleTargets,
       ExtendedEventHandler eventHandler,
       EventBus eventBus,
       boolean includeExecutionPhase,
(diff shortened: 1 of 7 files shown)


gregestren:
@philsc I'd love your review too (but I can't assign to you).

src/main/java/com/google/devtools/build/lib/buildtool/AnalysisPhaseRunner.java
```diff
@@ -225,6 +225,7 @@ private static AnalysisResult runAnalysisPhase(
             request.getCheckForActionConflicts(),
             request.getLoadingPhaseThreadCount(),
             request.getTopLevelArtifactContext(),
+            request.reportIncompatibleTargets(),
```
philsc, line 228: I admittedly don't know anything about the Java style guide, but this seems more readable as `request.incompatibleTargetsShouldBeReported()`. Though obviously more wordy. Otherwise, it makes it seem like this function actually reports targets and returns success/failure of the report maybe? Not sure. Basically phrase it like an assertion.

Anyway, just a stream of consciousness that I had. I guess I should read up on the Java style guide.

EDIT: looking at all the other booleans in this code, looks like this is being consistent. So nevermind!

philsc (approved):
This is super cool!  I say ship it :D

konste:
So what `cquery` is going to do with the incompatible targets - remove them from the output or keep them in the output? It looks like this change prevents `cquery` from failing when it hits incompatible target, but still outputs that target. Is that right? We really need a way to skip incompatible targets from the output. May be we can have a flag for that?

philsc:
@konste , you can filter out incompatible targets like so: https://docs.bazel.build/versions/main/platforms.html#detecting-incompatible-targets-using-bazel-cquery

gregestren:
I agree with @philsc .

There are contradictory needs here. One need is as you describe: identifying and skipping incompatible targets. The other need is to be able to identify them in the first place, including where they appear in the dependency graph and how the build gets to them. 

I don't think cquery should be in the business of worrying about the distinction between incompatible vs. compatible targets. What it should do is not fail, which is what this PR handles. For those who do want to identify incompatible targets I'd hope the right specific query would be the way to do that, as @philsc links.

Is that workable for you?

konste:
@philsc yes, this is what we do right now. But it looks... bulky? Also every time incompatible target is skipped from `cquery` output it is replaced with the empty line, which makes processing the output more cumbersome. But all inconveniences aside - yes, it does the job.

@gregestren if it is not too much work I would like to have a flag which skips incompatible targets from `cquery` output OR alternatively flag them somehow in the output so they can be skipped by the following code. But I admit it is just the convenience thing and not a blocker.

gregestren:
I don't think anyone wants cquery to actually error on incompatible targets, so I think his PR is a pure improvement regardless of how we address your concerns.

In that spirit, can we consider the issues separately? i.e. submit this PR as-is and file a separate issue for your concerns that we'll consider independently?

For reasons described we're really reluctant to bake in too much specific functionality, especially when there are reasonable alternatives through Starlark, where we try to route as much user customization power as we can. I'd remain really resistant to a cquery flag. One of the goals of cquery's Starlark output was to save us from creating new output formatters (`--output=...`) for the myriad ways users want to consume the output.

But I am more open to the idea of somehow flagging the output. i.e. cquery results already show configuration hashes and whether they're in the HOST configuration. Adding a `(incompatible)` suffix could be consistent. I don't want to guarantee that without further discussion, but that's where I'd focus the feature request.

gregestren:
@katre I think I'm ready for your code owner review.

konste:
Yes @gregestren I agree that this PR is a pure improvement and should go forward regardless. I will open a separate feature request for annotating incompatible `cquery` outputs.

src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
```diff
@@ -433,31 +434,37 @@ public AnalysisResult update(
...
-                  platformRestrictions.targetsToSkip())
-              .immutableCopy();
+        targetsToSkip =
+            Sets.union(
+                    topLevelConstraintSemantics.checkTargetEnvironmentRestrictions(
+                        skyframeAnalysisResult.getConfiguredTargets()),
+                    platformRestrictions.targetsToSkip())
+                .immutableCopy();
+      } else {
+        targetsToSkip = ImmutableSet.of();
```
katre, line 466: Nit: Set this back on line 437 (where targetsToSkip is declared) and remove the `else` clause.

src/main/java/com/google/devtools/build/lib/buildtool/BuildRequest.java
```diff
@@ -131,6 +132,20 @@ public Builder setCheckforActionConflicts(boolean checkForActionConflicts) {
       return this;
     }
 
+    /**
+     * If true, build status depends on whether or not requested targets are platform-compatible
+     * ({@link com.google.devtools.build.lib.analysis.IncompatiblePlatformProvider}. If false, this
```
katre, line 137: Nit: You have an open paren and no close paren.

katre (changes requested):
LGTM, with a few small nits to clear up.

src/main/java/com/google/devtools/build/lib/analysis/BuildView.java
gregestren: Done.

src/main/java/com/google/devtools/build/lib/buildtool/BuildRequest.java
gregestren: Done.

katre (approved):

PR #14318, opened 2021-11-24
Remote: Limit max number of gRPC connections by --remote_max_connecti… (cherry pick for 4.2.2)

limdor:
…ons.

`--remote_max_connections` is only applied to HTTP remote cache. This PR makes it apply to gRPC cache/executor as well.

Note that `--remote_max_connections` limits the number of concurrent connections. For HTTP remote cache, one connection could handle one request at one time. For gRPC remote cache/executor, one connection could handle 100+ concurrent requests. So the default value `100` means we could make up to `100` concurrent requests for HTTP remote cache or `10000+` concurrent requests for gRPC remote cache/executor.

Fixes: #14178.

Closes #14202.

PiperOrigin-RevId: 410249542
(cherry picked from commit 8d5973d29d60c0c615838c534ee27f93377cf5af)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannel.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ReferenceCountedChannel.java
@@ -56,8 +56,13 @@ public class ReferenceCountedChannel extends Channel implements ReferenceCounted
       };
 
   public ReferenceCountedChannel(ChannelConnectionFactory connectionFactory) {
+    this(connectionFactory, /*maxConnections=*/ 0);
+  }
+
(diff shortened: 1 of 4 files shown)


limdor:
I took a look to the failing tests in the CI but I do not see how this can be related. It seems some issue with a test regarding floating point. Any way to trigger the same tests for the branch release-4.2.2-patches to see if the issue also appears there?

meteorcloudy:
I'm retrigging the test again 🤔

limdor:
It looks like the issue is also present in the branch https://buildkite.com/bazel/bazel-bazel/builds?branch=release-4.2.2-patches, but indeed there are more failures there :thinking:

meteorcloudy:
Succeeded at 4th try 😅

PR #14326, opened 2021-11-25
Revert "Set --experimental_enable_aspect_hints default to false"

thii:
This reverts commit 60ebb105dbf34f0b267ea7573157246b4c9bfcaf.

Some rulesets like the Swift rules have started depending on the
`aspect_hints` attribute since the time it was introduced, and some time
before the `--experimental_enable_aspect_hints` flag was added. Many
changes have stacked up since then, so setting this default to false
makes it difficult to support the rules users now.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -244,7 +244,7 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
 
   @Option(
       name = "experimental_enable_aspect_hints",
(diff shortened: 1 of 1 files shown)


thii:
cc @scentini since you're the author of the original change.

scentini:
cc @comius

comius:
The flag is experimental - meaning the contract is, that it may be removed at anytime.

The `aspect_hints` feature was put behind a flag, because we know there are issues with its design and we would like to keep our options open. Potentially this feature is going to be designed differently.

Keeping the flag to `false`, will help containing it.

I'll comment on the issue and close this PR for now.

PR #14328, opened 2021-11-25
Fail on Xcode configure errors, add timeout

jerrymarino:
Bazel puts this error in the build file which is a problem on the CI because it
swallows the error and it requires a user to clean --expunge. Bazel requires
the user to expunge because it thinks the repository was successfully written.

Addtionally I am constatly hititng this problem with github actions. The
default timeout of 120s is not long enough. This is because the performance of
github actions is unstable, probably from over-subscribing the hosts. The fix
for that is to to provide custom workers and use Buildkite.

I'm planning to use this attribute to increase the timeout in my repository. A
retry mechanism isn't appropriate because the operations aren't flaky in any
capacity, the host CPU is just over subscribed.

Here is an example where I've forked and patched the `fail()` change on but
prior to increasing the timeout
https://github.com/bazel-ios/rules_ios/runs/4319115106?check_suite_focus=true

```
Error in fail: Generating xcode-locator-bin failed.  return code 256, stderr: Timed out, stdout:
...
62
    File "/Users/runner/work/rules_ios/rules_ios/tools/toolchains/xcode_configure/xcode_configure.bzl", line 148, column 13, in run_xcode_locator
63
	    fail(error_msg.replace("\n", " "))
64
Error in fail: Generating xcode-locator-bin failed.  return code 256, stderr: Timed out, stdout:`
```

### DIFF
--- a/tools/osx/xcode_configure.bzl
+++ b/tools/osx/xcode_configure.bzl
@@ -17,8 +17,6 @@
    installed on the local host.
 """
 
-_EXECUTE_TIMEOUT = 120
-
 def _search_string(fullstring, prefix, suffix):
     """Returns the substring between two given substrings of a larger string.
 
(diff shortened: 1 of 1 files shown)


keith (approved):
I think this is a good compromise but technically there is a behavior change here that got a change of mine reverted in the past. Previously if you had _invalid_ xcode directories, which IIRC google has internally in some SVN repo that checks in all of Xcode, this rule would not fail, but you just wouldn't get an Xcode toolchain. And with this change that case will fail instead.

jerrymarino:
Thanks for the update - this one is probably possible for us to just tackle outside of Bazel

PR #14329, opened 2021-11-25
Correct help message for experimental_strict_action_env

tsawada:
I might be simply wrong but it seems to me that `experimental_strict_action_env` does not affect TMPDIR actually. I wasn't able to find relevant code, and also in my experiments `echo $TMPDIR` in `sh_test` behave the same with or without the flag.


aiuto:
Assigning to lberki as original author of the comment.

meisterT:
seems reasonable looking at the code: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java;l=198;drc=982bbce2cfe12942eb47de214a2393e2cfd56c04

meisterT (approved):

PR #14333, opened 2021-11-25
Undocument --bes_best_effort

Yannic:
We've heared from several people that they tried passing the flag and it did
not have any efect (as expected from a no-op), so let's undocument the flag
so it's no longer discoverable.

On a related note, we should document `--bes_upload_mode` which seems to be
what people want to do.


Yannic:
cc @lfpino

lfpino (approved):
We should just delete this flag at this point. It's been a no-op for years.

Yannic:
Removing sounds tricky to me, looks like lots of builds would breaks because people are setting it. If we can cherry-pick to 5.x, we should be able to remove in 6.0

meisterT:
Move it to the graveyard then? https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRulesModule.java;l=54;drc=2c6fa349822312ee953d5961fc42dafe775848b7

Yannic:
@meisterT I can do that, but that graveyard seems Bazel specific. Can you verify that moving it there won't break Blaze (or take care of also adding it to the Blaze graveyard when merging)?

meisterT:
Yep, that works for both Blaze and Bazel.

Yannic:
Done, PTAL

meisterT (approved):

Yannic:
What's the status of this? I'd like to get this into 5.0

/cc @wyverald

Wyverald:
@meisterT started the import process last year, but he's still on vacation until next week. In the meantime, @michaeledgar left the following comment:

> I think this needs to go in AllCommandGraveyardOptions, because BES options are applied to many commands: see `BazelBuildEventServiceModule#allowedCommands()` and `BlazeBuildEventServiceModule#allowedCommands()`.

Yannic:
Sorry for the delay, done. PTAL

Yannic:
Any updates here? @meisterT @michaeledgar

Yannic:
ping? I'd like to get this into 5.1

Wyverald:
I'll check with @meisterT

meisterT:
Sorry for the delay, giving it a kick internally.

meisterT:
@Yannic will you create the cherrypick against the release branch?

Wyverald:
I can do it.

PR #14335, opened 2021-11-26
Find runfiles in directories that are themselves runfiles

### LINKED ISSUE #14336: Runfiles libraries don't find runfiles in directories that are themselves runfiles when using manifest

In the way they are currently implemented, the Bazel runfiles libraries at `@bazel_tools//tools/*/runfiles` do not always find the identical set of runfiles when using a manifest compared to when using a runfiles directory.

Specificially, if a target has runfiles `foo` (a directory) and `foo/a.txt` (a file in that directory), then [`Runfiles#filterListForObscuringSymlinks`](https://github.com/bazelbuild/bazel/blob/aaf87f4ef3e19cbf1e5ab28f50cd8fb90dfd115e/src/main/java/com/google/devtools/build/lib/analysis/Runfiles.java#L331) will filter out `foo/a.txt`. As a result:
1. when using a directory-based runfiles lookup, `foo/a.txt` will be found since the symlink `foo` resolves to the correct directory containing `a.txt`;
2. when using a manifest-based runfiles lookup, `foo/a.txt` is not found since it is not listed in the manifest, only `foo` itself is.

Note that it is pretty easy to pick up both a directory and a file contained in that directory as runfiles through transitive dependencies and everything will work well on Linux/macOS, only to non-obviously break on Windows. I encountered this issue while working on a ruleset that would allow for compile-time checked `rlocation` lookups for C++ and Java, but found the current issue to block this effort.

```
#!/usr/bin/env bash
touch WORKSPACE

mkdir -p foo
touch foo/a.txt

cat <<EOF >main.cc
int main() {}
EOF

cat <<EOF >BUILD.bazel
cc_binary(
    name = "main",
    data = [
        "foo",
        "foo/a.txt",
    ],
    srcs = ["main.cc"],
)
EOF

bazel build //:main
cat bazel-bin/main.runfiles_manifest
```

correctly prints
```
__main__/foo /tmp/repro/foo
__main__/main /home/user/.cache/bazel/_bazel_user/.../execroot/__main__/bazel-out/k8-fastbuild/bin/main
```

When using `@bazel_tools//tools/cpp/runfiles` with the runfiles manifest, e.g. on Windows, then `Rlocation("foo/a.txt")` will return an empty string instead of the correct path for the declared runfile.

Any, but this mostly applies to Windows.

Any version is affected, including current `master`.

###  Have you found anything relevant by searching the web?

No.

I prepared a PR (#14335) that extends the C++, Java, and Python runfiles libraries so that they also find runfiles within runfiles directories when using the runfiles manifest. I could also fix this for the bash implementation, but would like to wait for general feedback before I start working on this.

fmeum:
When a target has a runfile that is contained in a directory that is itself one of its runfiles, the runfile will be shadowed by the `SymlinkEntry` for the directory. While this still allows to resolve the file in the runfiles symlink tree, a manifest-based lookup will fail. Since the manifest is used to create the symlink tree for which it is important that no path is a prefix of another, this can only be fixed in the runfiles libraries.

This PR extends the lookup logic of the Bash, C++, Java, and Python runfiles libraries to also find runfiles contained within directories that are themselves runfiles. It does so by searching the manifest not only for the exact provided rlocation path, but also for all path prefixes. If a prefix is looked up successfully, the corresponding suffix is resolved relative to the looked up path.

Fixes #14336.


fmeum:
@mai93 Would you be able to take a look? You reviewed and worked on the last fixes to the runfiles libraries.

fmeum:
@meteorcloudy Could you review this PR in case @mai93 is busy?

tools/java/runfiles/testing/RunfilesTest.java
```diff
@@ -249,10 +250,16 @@ public void testManifestBasedRlocation() throws Exception {
         new MockFile(
             ImmutableList.of(
                 "Foo/runfile1 C:/Actual Path\\runfile1",
-                "Foo/Bar/runfile2 D:\\the path\\run file 2.txt"))) {
+                "Foo/Bar/runfile2 D:\\the path\\run file 2.txt",
+                "Foo/Bar/Dir E:\\Actual Path\\Directory"))) {
       Runfiles r = Runfiles.createManifestBasedForTesting(mf.path.toString());
       assertThat(r.rlocation("Foo/runfile1")).isEqualTo("C:/Actual Path\\runfile1");
       assertThat(r.rlocation("Foo/Bar/runfile2")).isEqualTo("D:\\the path\\run file 2.txt");
```
fmeum, line 257: This existing test case seems to go against the specification of `rlocation` in https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub: It demands the return value to be an absolute path that uses `\` as the directory separator, whereas the design doc demands all paths to use `/` as path separators. 

Depending on what is correct, I would replace the uses of OS-specific path joining logic with simple concenations with `/` as the separator.

meteorcloudy: I think the documentation is correct, in reality, only `/` is used in the runfiles manifest file, feel free to fix this test case! Let's also use `/` for this case to be consistent and less error prone.

meteorcloudy (changes requested):
Thanks for fixing this use case!

tools/cpp/runfiles/runfiles_src.cc
```diff
@@ -48,6 +49,12 @@ using std::vector;
 
 namespace {
 
+#ifdef _WIN32
```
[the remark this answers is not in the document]
fmeum: Yes, I got rid of it.

tools/java/runfiles/testing/RunfilesTest.java
fmeum: I'm using `/` as a separator in all cases. I will leave the test case I mentioned as is: I assume it exists more to show that manifest entries are left as is and not modified.

fmeum:
I gave the Bash implementation a try and it wasn't that bad. This PR now includes fixes for all runfiles libraries shipped with Bazel.

Let me know if you would prefer me to squash the four commits in case you want to import them together.

tools/bash/runfiles/runfiles.bash
```diff
@@ -122,16 +122,42 @@ function rlocation() {
...
+          [[ -z "$prefix_result" ]] && continue
+          local -r candidate="${prefix_result}${1#"${prefix}"}"
+          if [[ -e "$candidate" ]]; then
+            if [[ "${RUNFILES_LIB_DEBUG:-}" == 1 ]]; then
+              echo >&2 "INFO[runfiles.bash]: rlocation($1): found in manifest as ($candidate) via prefix ($prefix)"
+            fi
+            echo "$candidate"
+            return 0
+          fi
+          # There is only ever one matching prefix in the runfiles manifest.
```
meteorcloudy, line 147: Shouldn't this `break` statement be inside the if statement?

meteorcloudy (changes requested):

tools/bash/runfiles/runfiles.bash
fmeum: I actually think it's correct as is, but agree this is confusing and thus added an extensive comment. Let me know what you think.

fmeum:
I squashed the commits and added a descriptive message.

tools/bash/runfiles/runfiles.bash
meteorcloudy: Thanks! I missed the continue statement in the middle and thought the while loop would never run a second time. Thanks for the explanation!

tools/bash/runfiles/runfiles.bash
```diff
@@ -122,16 +122,51 @@ function rlocation() {
...
+            echo "$candidate"
+            return 0
+          fi
+          # At this point, the manifest lookup of prefix has been successful,
+          # but the file at the relative path given by the suffix does not
+          # exist. We do not continue the lookup with a shorter prefix for two
+          # reasons:
+          # 1. Manifests generated by Bazel never contain a path that is a
+          #    prefix of another path.
+          # 2. Runfiles libraries for other libraries do not check for file
```
meteorcloudy, line 153: nit: for other languages?

meteorcloudy (approved):

tools/bash/runfiles/runfiles.bash
fmeum: Good catch, fixed.

Issue #7260, opened 2019-01-25
incompatible_enable_cc_toolchain_resolution: Turn on toolchain resolution for cc rules


katre:
Flag: `--incompatible_enable_cc_toolchain_resolution`
Available since: 0.23
Will be flipped in: 7.0.0

### Description

C++ toolchain resolution is an improved mechanism that selects a proper C++ toolchain (compiler) based on a pair of target and execution platforms (think cross compilation from linux to mac). It’s an improvement over the old mechanism that sets the `--crosstool_top` flag. More information about platforms and toolchain resolution on https://bazel.build/reference/be/platforms-and-toolchains

### Current status

- C++ toolchain resolution has been in use at large repository at Google for a year without significant problems.
- rules_apple and rules_swift lack toolchainization, however they work using `platform_mapping`
- Bazel's C++ toolchain autoconfiguration already supports toolchain resolution.

### Migration

If you're not using C++, or custom `--crosstool_top`, `--cpu`, or `--compiler`,
you don't use `select` on these options, you can stop reading now, there is
nothing to migrate for you.

Migration before the flag is flipped (only in case you're building for custom platforms, using custom C++ toolchains or using custom rules invoking `cc_common.compile` and `cc_common.link`):

1. Add C++ platform definitions, register C++ toolchains
2. Add missing C++ toolchain requirements to rules (Already added to OSS rules)
3. Update blazerc configuration to use custom platforms
4. Update Starlark configuration transitions to use platforms (ones affecting --cpu and --crostool_top, in cases `platform_mappings` isn’t used)
5. Update or fix configurations (`bazelrc`) to use `--platforms` (where `--cpu` or `--crosstool_top` is used)

Migration after the flag is flipped:

1. Replace cpu and os based selects with platform based selects (for example selects `cc_target_os`, `target_cpu` and other)
2. Remove `--crosstool_top`, `--cpu` flags from `bazelrcs` and from Starlark configuration transitions
3. Remove part of `platform_mappings` configuration needed for C++
4. Remove `_cc_toolchain` implicit dependency

##### 1. Add C++ platform definitions, register C++ toolchains

Please use `constraint_settings` and `constraint_values` from
the canonical [Bazel Platforms
Repository](https://github.com/bazelbuild/platforms) (don't hesitate to propose
missing targets!). If there is a need for C++ specific constraints, feel free to
upload a PR to the [`rules_cc`](https://github.com/bazelbuild/rules_cc)
[...]

PR #14340, opened 2021-11-27
Enable native support for Windows on arm64

nsait-linaro:
This PR will enable cross-compilation of Bazel binaries for win/arm64 from win/x64

`bazel build --subcommands --copt=-w --host_copt=-w -c opt --cpu=x64_arm64_windows //src:bazel.exe`

Generated bazel executable can be used for native compilation in win/arm64

`bazel.exe build --subcommands //main:hello-world`

Following changes are included

- Add grpc patch to workaround win/arm64 specific build issues
- Add win/arm64 JDK 17
- Fix AutoCpuConverter.java to identify win/arm64 CPU
- Extend build_bazel_binaries.yml to cross-compile for win/arm64
- Fix msvc toolchain to look for tools in HostX86 directory as well

The patch is only a draft at this moment and there are a couple of things required before merging the PR

- Upload missing JDK targets for win/arm64 to Bazel mirror and add rules to use them
-  Rename `x64_arm64_windows` to `arm64_windows` or add an alias for `arm64_windows` as it might be a better name with native compilation
- Update test/documentation


nsait-linaro:
I am not sure how the CI failures are related to this patch. I've tried to run the test locally on top of master and the tests are failing there too.

philwo:
Sorry for the delay @nsait-linaro. I just saw your e-mail on bazel-dev@!

@meteorcloudy @mai93 Could you have a look at this PR when you have time? It will be very hard for us to test as we don't have any Windows arm64 machines either personally nor on CI, but it looks pretty exciting and not too complex!

@nsait-linaro Could you try to upstream your change to gRPC and link the PR here? We can apply a patch in our local copy, but we generally like to avoid keeping these around for too long to reduce complexity on our side.

philwo:
I manually flipped the CLA label to green, because signcla says we have a valid CLA.

meteorcloudy:
`Auto-Configuration Error: 'PROCESSOR_ARCHITECTURE' environment variable is not set`
Can you please fix this error in the tests?

nsait-linaro:
> `Auto-Configuration Error: 'PROCESSOR_ARCHITECTURE' environment variable is not set` Can you please fix this error in the tests?

Sure.  I am on holiday now.  I will have a look when I am back.

meteorcloudy:
@nsait-linaro Thanks for working on this, this is awesome!

> def_parser binaries cannot handle arm64 objects yet

We probably need to sync with the upstream code in CMake: 
https://github.com/Kitware/CMake/commits/master/Source/bindexplib.cxx
ARM64 support: https://github.com/Kitware/CMake/commit/8950183b3834dc2179dc4965138b1091e291ae9f

> clang-cl toolchain support needs to be added (TensorFlow cannot be compiled with MSVC due to XNNPACK dependency having a lot of assembly code that cannot be compiled with MSVC)

Do you mean XNNPACK cannot be compiled with MSVC on ARM64, because at least the TF x64 build is working with MSVC, right? Also, we already support clang-cl in Bazel, see https://docs.bazel.build/versions/main/windows.html#build-c-with-clang. But I think last time I tried to build TF with `clang-cl`, I got some compilation errors.


> I would also like to ask your opinion on renaming the CPU name from x64_arm64_windows to arm64_windows

Agree, we should rename it. FYI @mai93

nsait-linaro:
@meteorcloudy and @philwo - Apologies for the delay. Finally, I managed to put together a more complete patch to enable windows on arm. Waiting for review!

aiuto:
This is not really my wheelhouse but I would feel more comfortable if this were two distinct PRs.

The first would add enough toolchain support so that bazel for x86_64 can cross compile for arm64. 
If we have our core features right, that should require almost no Bazel changes. We might even be able to build for windows arm64 by cross compiling from linux.

The followup PR would add support to build bazel itself for windows arm.

philwo (approved):

philwo:
Seems like a lot of tests are failing with this error now:

> ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/5933/execroot/io_bazel/_tmp/0f0d5afbc0f3af3b7e882cc5dbea196c/root/dbac4e5567f590e08c93c2d7d21a8944/external/bazel_tools/src/conditions/BUILD:145:27: configurable attribute "actual" in @bazel_tools//src/conditions:host_windows doesn't match this configuration. Would a default condition help?
> 
> Conditions checked:
>  @bazel_tools//src/conditions:host_windows_x64_constraint
>  @bazel_tools//src/conditions:host_windows_arm64_constraint
> 
> To see a condition's definition, run: bazel query --output=build <condition label>.
> 
> This instance of @bazel_tools//src/conditions:host_windows has configuration identifier dafa3bb. To inspect its configuration, run: bazel config dafa3bb.
> 
> For more help, see https://docs.bazel.build/configurable-attributes.html#why-doesnt-my-select-choose-what-i-expect.

Could you please have a look? I think this is fine to merge once tests pass. It's basically just adding one more case to our existing multi-arch/platform support. FYI @meteorcloudy is out of office for two days, I think he'll review after he is back.

nsait-linaro:
> Seems like a lot of tests are failing with this error now:
> 
> > ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/5933/execroot/io_bazel/_tmp/0f0d5afbc0f3af3b7e882cc5dbea196c/root/dbac4e5567f590e08c93c2d7d21a8944/external/bazel_tools/src/conditions/BUILD:145:27: configurable attribute "actual" in @bazel_tools//src/conditions:host_windows doesn't match this configuration. Would a default condition help?
> > Conditions checked:
> > @bazel_tools//src/conditions:host_windows_x64_constraint
> > @bazel_tools//src/conditions:host_windows_arm64_constraint
> > To see a condition's definition, run: bazel query --output=build .
> > This instance of @bazel_tools//src/conditions:host_windows has configuration identifier dafa3bb. To inspect its configuration, run: bazel config dafa3bb.
> > For more help, see https://docs.bazel.build/configurable-attributes.html#why-doesnt-my-select-choose-what-i-expect.
> 
> Could you please have a look? I think this is fine to merge once tests pass. It's basically just adding one more case to our existing multi-arch/platform support. FYI @meteorcloudy is out of office for two days, I think he'll review after he is back.

Thanks @philwo. I've pushed a new commit to fix the above issue. How do I see the error log for the remaining failures?

philwo:
Hi @nsait-linaro,

Thanks! To see the error logs, first click the Details link next to one of the failing Buildkite checks in GitHub:
Screenshot 2022-02-01 at 14 11 14

Then you can see the individual CI jobs and whether they passed or failed. If a test failed, there will be a test log in the Artifacts tab of the job:
Screenshot 2022-02-01 at 14 12 33

If it's a build failure, the error message will instead be in the main Bazel log shown for the job.

In this case, it seems like there's only a small tweak left to do in order to make everything pass:

```
//:srcs filegroup do not contains all the sources, missing:
tools/windows/BUILD
tools/windows/windows_config.bzl
```

You'll have to add your `//tools/windows:srcs` filegroup that you already created in your new BUILD file to this list: https://github.com/bazelbuild/bazel/blob/master/tools/BUILD#L4 :)

third_party/grpc/BUILD
```diff
@@ -21,6 +21,7 @@ licenses(["notice"])  # Apache v2
 exports_files([
     "grpc_1.31.1.patch",
     "grpc_1.41.0.patch",
+    "grpc_1.41.0.win_arm64.patch"
```
meteorcloudy, line 24: Is this change upstreamed? Can we remove this patch after upgrading grpc to a version that contains that change?

tools/cpp/windows_cc_configure.bzl
```diff
@@ -323,6 +323,9 @@ def _get_vc_env_vars(repository_ctx, vc_path, msvc_vars_x64, target_arch):
         lib = msvc_vars_x64["%{msvc_env_lib_x64}"]
         full_version = _get_vc_full_version(repository_ctx, vc_path)
         tools_path = "%s\\Tools\\MSVC\\%s\\bin\\HostX64\\%s" % (vc_path, full_version, target_arch)
+        # For native windows on arm64 builds host toolchain runs in an emulated x86 environment
```
meteorcloudy, line 326: Wow, do you mean there is no C++ build tools that runs natively on Windows arm64 platform? 

Also, can you make it a parameter instead of using "replace", I think that's safer.
```
tools_path = "%s\\Tools\\MSVC\\%s\\bin\\%s\\%s" % (vc_path, full_version, host_arch, target_arch)
```

tools/cpp/windows_cc_configure.bzl
```diff
@@ -786,11 +792,13 @@ def configure_windows_toolchain(repository_ctx):
     template_vars = dict()
     msvc_vars_x64 = _get_msvc_vars(repository_ctx, paths, "x64")
     template_vars.update(msvc_vars_x64)
-    template_vars.update(_get_clang_cl_vars(repository_ctx, paths, msvc_vars_x64))
+    template_vars.update(_get_clang_cl_vars(repository_ctx, paths, msvc_vars_x64, "x64"))
     template_vars.update(_get_msys_mingw_vars(repository_ctx))
     template_vars.update(_get_msvc_vars(repository_ctx, paths, "x86", msvc_vars_x64))
     template_vars.update(_get_msvc_vars(repository_ctx, paths, "arm", msvc_vars_x64))
-    template_vars.update(_get_msvc_vars(repository_ctx, paths, "arm64", msvc_vars_x64))
+    msvc_vars_arm64 = _get_msvc_vars(repository_ctx, paths, "arm64", msvc_vars_x64)
```
meteorcloudy, line 799: nit, just:
```
    template_vars.update(_get_msvc_vars(repository_ctx, paths, "arm64", msvc_vars_x64))
```
?

src/conditions/BUILD
```diff
@@ -167,6 +167,43 @@ config_setting(
...
+
+config_setting(
+    name = "windows_arm64_constraint",
+    constraint_values = [
+        "@platforms//os:windows",
+        "@platforms//cpu:arm64",
+    ],
+)
+
+config_setting(
```
meteorcloudy, line 187: I think we won't need them after https://github.com/bazelbuild/bazel/issues/7260 is fixed, can you put a TODO here?

@oquenchil Can you confirm this?

meteorcloudy (changes requested):

src/test/py/bazel/bazel_windows_cpp_test.py
```diff
@@ -1033,6 +1033,24 @@ def testBuildCppBinaryWithMsysGCC(self):
...
+    ])
+    self.ScratchFile('main.cc', [
+        'int main() {',
+        '  return 0;',
+        '}',
+    ])
+    exit_code, _, stderr = self.RunBazel(
+        ['build', '-s', '--cpu=arm64_windows', '//:main'])
+    self.AssertExitCode(exit_code, 0, stderr)
+    self.assertIn('arm64/cl.exe', ''.join(stderr))
```
meteorcloudy, line 1052: Has to be `'arm\\cl.exe'`

tools/cpp/BUILD.windows.tpl
```diff
@@ -489,6 +492,72 @@ toolchain(
...
+    name = "clang_cl_arm64",
+    cpu = "arm64_windows",
+    compiler = "clang-cl",
+    host_system_name = "local",
+    target_system_name = "aarch64-pc-windows-msvc",
+    target_libc = "msvcrt",
+    abi_version = "local",
+    abi_libc_version = "local",
+    toolchain_identifier = "clang_cl_arm64",
+    msvc_env_tmp = "%{clang_cl_env_tmp}",
```
meteorcloudy, line 520: Those env var values are also different for x86 and arm64, this is probably why the [tests are failing on Windows](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/df9ece11-19b6-40d3-947d-4c5d10403786/src%5Ctest%5Cpy%5Cbazel%5Cbazel_windows_cpp_test%5Ctest.log). The second call of `_get_clang_cl_vars` will just override the result from the first call.

meteorcloudy (changes requested):

tools/cpp/BUILD.windows.tpl
nsait-linaro: Yes of course! Thanks a lot for pointing it out.

tools/cpp/windows_cc_configure.bzl
nsait-linaro: `msvc_vars_arm64` is required later (line 801) so have to store it in a variable

tools/cpp/windows_cc_configure.bzl
nsait-linaro: Yes, they all run in emulated mode now.

For Windows 11, x64 toolchain is available so I want to keep the logic of checking for x64 toolchain first and then fallback to x86 if x64 toolchain is not available.

third_party/grpc/BUILD
nsait-linaro: Yes, part of it (renaming move64 function) is upstreamed and is available from grpc v1.43.2. But the other part (changes cares build config) is not yet upstreamed.

nsait-linaro:
> This is not really my wheelhouse but I would feel more comfortable if this were two distinct PRs.
> 
> The first would add enough toolchain support so that bazel for x86_64 can cross compile for arm64. If we have our core features right, that should require almost no Bazel changes. We might even be able to build for windows arm64 by cross compiling from linux.
> 
> The followup PR would add support to build bazel itself for windows arm.

The cross-compilation already works and this patch adds support for windows/arm64 host

src/conditions/BUILD
oquenchil: I don't know much about #7260, about what's needed or not. Can you clarify why these config settings won't be needed?

meteorcloudy: If toolchain resolution for C++ is enabled, we no longer need to use `--cpu=x64_windows` to select toolchain, then this config setting won't be needed.

third_party/grpc/BUILD
meteorcloudy: I see, it will be great if the fix for cares.BUILD can also be upstreamed.

tools/cpp/windows_cc_configure.bzl
meteorcloudy: Got it, thanks!

meteorcloudy (approved):
Thank you, this is great!

meteorcloudy:
@nsait-linaro Can you please split this PR into two? The first one contains changes under `third_party/`, the second one contains the rest changes. We have to merge them separately due to our setup.

nsait-linaro:
> @nsait-linaro Can you please split this PR into two? The first one contains changes under `third_party/`, the second one contains the rest changes. We have to merge them separately due to our setup.

Sure, I will create a separate PR for `third_party/` and keep this PR for the rest of the changes. Is that okay ?

nsait-linaro:
> I think that's OK, but you'll have to rebase after we merge the first one.

Ok, First PR with `third_party` contents: https://github.com/bazelbuild/bazel/pull/14689

I will rebase this patch once that is merged

meteorcloudy:
OK, https://github.com/bazelbuild/bazel/pull/14689 is merged.

WORKSPACE
```diff
@@ -254,6 +254,13 @@ http_file(
     urls = ["https://mirror.bazel.build/openjdk/azul-zulu11.37.17-ca-jdk11.0.6/zulu11.37.17-ca-jdk11.0.6-win_x64.zip"],
 )
 
+http_file(
+    name = "openjdk_win_arm64_vanilla",
+    downloaded_file_path = "zulu-win-arm64.zip",
+    sha256 = "811d7e7591bac4f081dfb00ba6bd15b6fc5969e1f89f0f327ef75147027c3877",
+    urls = ["https://cdn.azul.com/zulu/bin/zulu17.30.15-ca-jdk17.0.1-win_aarch64.zip"],
```
meteorcloudy, line 261: Please change to `https://mirror.bazel.build/cdn.azul.com/zulu/bin/zulu17.30.15-ca-jdk17.0.1-win_aarch64.zip` to make sure this is more realiable.

.bazelci/build_bazel_binaries.yml
```diff
@@ -49,3 +49,13 @@ platforms:
     build_targets:
       - "//src:bazel.exe"
       - "//src:bazel_nojdk.exe"
+  windows_arm64:
```
meteorcloudy, line 52: We don't support `windows_arm64` on our CI yet because we don't have the arm64 machine, but we can cross-compile a Windows arm64 binary just like we did for macos arm64. Have you tried to cross-compile a arm64 Bazel binary, does it work?

.bazelci/build_bazel_binaries.yml
```diff
@@ -49,3 +49,13 @@ platforms:
...
+    build_flags:
+      - "--copt=-w"
+      - "--host_copt=-w"
+      - "-c"
+      - "opt"
+      - "--cpu=x64_arm64_windows"
```
meteorcloudy, line 58: Shouldn't it just be `--cpu=arm64_windows`?

.bazelci/build_bazel_binaries.yml
nsait-linaro: Yes, cross-compilation should work.

.bazelci/build_bazel_binaries.yml
nsait-linaro: Yes, I had it as `arm64_windows` but changed it thinking there might be bootstrapping issue. At least for the first ARM64 `Bazel` build, we will be using `Bazel` which doesn't recognize `arm64_windows` as CPU.

meteorcloudy: Oh, I see, you are right.

meteorcloudy: Thanks, I removed some unnecessary build flags.

meteorcloudy: Adding the support for `windows_arm64` here: https://github.com/bazelbuild/continuous-integration/pull/1320

meteorcloudy (approved):

meteorcloudy:
BTW, can normal x64 binaries run on Windows arm64? Bazel pulls [java_tools](https://github.com/bazelbuild/java_tools/releases) as an external dependency, which contains some x64 binaries. Can those binaries run seamlessly just like on Apple silicon? Eventually, we should publish java_tools for arm64 platforms.

nsait-linaro:
> BTW, can normal x64 binaries run on Windows arm64? Bazel pulls [java_tools](https://github.com/bazelbuild/java_tools/releases) as an external dependency, which contains some x64 binaries. Can those binaries run seamlessly just like on Apple silicon? Eventually, we should publish java_tools for arm64 platforms.

Yes, Windows 11 supports x64 emulation so that should work.

nsait-linaro:
Thanks, @meteorcloudy and @philwo for your help! It is merged now 🥳 

I would like to check one of the bazel(arm64) builds generated by CI system. Does it produce any nightly builds I could try ? Or should I wait for the next release.

meteorcloudy:
Thanks! It'll be great if you could help check if everything is working correctly on Windows ARM64 before we adding it the official release.

We actually build Bazel at every commit, the binary should be available at https://storage.googleapis.com/bazel-builds/artifacts/windows_arm64/c2ddbd1954af5baab63b93f2b055a410a27832c8/bazel. I just [enabled](https://github.com/bazelbuild/continuous-integration/commit/0ae9eb7150046d9afeebd9198f5bdc894a680698) publishing binary for windows_arm64, you'll have to use latest commits to fetch the binary. 

If you are interested, you could also try to make [Bazelisk](https://github.com/bazelbuild/bazelisk) work on Windows arm64, then it'll be even easier to fetch Bazel binary at your desired version/commit.

nsait-linaro:
> Bazelisk

Thanks, @meteorcloudy. I will try the binary.

Yes, `Bazelisk` would be nice as well. I will have a look

nsait-linaro:
> Thanks! It'll be great if you could help check if everything is working correctly on Windows ARM64 before we adding it to the official release.
> 
> We actually build Bazel at every commit, the binary should be available at https://storage.googleapis.com/bazel-builds/artifacts/windows_arm64/c2ddbd1954af5baab63b93f2b055a410a27832c8/bazel. I just [enabled](https://github.com/bazelbuild/continuous-integration/commit/0ae9eb7150046d9afeebd9198f5bdc894a680698) publishing binary for windows_arm64, you'll have to use latest commits to fetch the binary.
> 
> If you are interested, you could also try to make [Bazelisk](https://github.com/bazelbuild/bazelisk) work on Windows arm64, then it'll be even easier to fetch 


Thanks, @meteorcloudy. I can confirm that Bazel binary is working as expected. Tried native and cross-compilation builds for arm64 and x64 windows targets with MSVC and clang and seems to work. [Log](https://github.com/bazelbuild/bazel/files/7993525/bazel.log) attached for reference.

I think we need to also add `exe` suffix to the published binary. Now it is published as `bazel` and windows doesn't run it until I add an `exe` extension to it.

meteorcloudy:
That's awesome! Have you also tried to build Java binary? 

> I think we need to also add exe suffix to the published binary. Now it is published as bazel and windows doesn't run it until I add an exe extension to it.

Yes, you'll have to rename it, but it won't be a problem with Bazelisk, it will do it for you!

nsait-linaro:
> That's awesome! Have you also tried to build Java binary?
> 
> > I think we need to also add exe suffix to the published binary. Now it is published as bazel and windows doesn't run it until I add an exe extension to it.
> 
> Yes, you'll have to rename it, but it won't be a problem with Bazelisk, it will do it for you!

@meteorcloudy , Unfortunately, the java build didn't work due to missing JDK configurations. I've created https://github.com/bazelbuild/bazel/pull/14700 which should fix the java builds now. Can you please review ?

PR #14352, opened 2021-11-30
Exit collect_coverage.sh early if LCOV_MERGER is not set.

### LINKED ISSUE #13978: Custom test rule without `_lcov_merger` attribute always fails under `bazel coverage`

c-mita:
Rules that provide InstrumentedFilesInfo but do not define the
_lcov_merger attribute will result in this script being called without
the required LCOV_MERGER environment variable being set.

Since _all_ rules now provide an InstrumentedFilesInfo provider by
default, but are not required to provide the attribute, this scenario
will be encountered far more frequently and should not be an error.

Because of the way coverage collection within Google differs to Bazel,
it isn't safe to change TestActionBuilder to simply not prepare the
collect_coverage.sh script the way it would if no coverage was to be
collected, so we have to change the script itself.

Fixes #13978


src/test/shell/bazel/bazel_coverage_starlark_test.sh
```diff
@@ -0,0 +1,110 @@
...
+        || fail "Coverage run failed but should have succeeded."
+}
+
+function test_starlark_rule_with_custom_lcov_merger() {
+
+    cat <<EOF > lcov_merger.sh
+for var in "\$@"
+do
+    if [[ "\$var" == "--output_file="* ]]; then
+        path="\${var:14}"
```
lberki, line 67: nit: if you want to do string manipulation like this instead, how about `${var##--output_file=}` instead so that one doesn't need to know that this is the number of characters in that string?

lberki (commented):
Remind me again, why isn't it feasible to add the `:lcov_merger` attribute to all Starlark test rules?

src/test/shell/bazel/bazel_coverage_starlark_test.sh
c-mita: Done.

c-mita:
> Remind me again, why isn't it feasible to add the `:lcov_merger` attribute to all Starlark test rules?
1. It's a breaking change, since rules that already define _lcov_merger will break.
2. It breaks things internally; rules within Google work a little differently w.r.t. setting up the lcov_merger tool (which is why the original PR, #13983, wasn't viable). It's typical for Google's test rules to leave it undefined and let the different `collect_coverage.sh` handle things.

Other little minor issues arise, but the second point is the big one.

lberki:
On second thought, there isn't any downside for this change, so LGTM. It would be nice to have an overarching design for adding coverage collection for any language, but that shouldn't be a blocker for this change.

lberki (approved):

Issue #12050, opened 2020-09-04
gRPC resource exhaustion errors during BEP upload


SrodriguezO:
### Description of the problem:

gRPC resource exhaustion errors during BEP upload when a very large build completes very quickly due to cache hits.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

We often run into it when a very large build completes very quickly due to cache hits (and `--bes_backend` is specified).

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

release 3.3.0

###  Have you found anything relevant by searching the web?

Similar issue encountered:

- https://github.com/buildbarn/bb-deployments/issues/2#issuecomment-489724682 

### Any other information, logs, or outputs that you want to share?

Error logs:

> ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=RESOURCE_EXHAUSTED, description=grpc: received message larger than max (10913322 vs. 4194304), cause=null}' RESOURCE_EXHAUSTED: RESOURCE_EXHAUSTED: grpc: received message larger than max (10913322 vs. 4194304) RESOURCE_EXHAUSTED: RESOURCE_EXHAUSTED: grpc: received message larger than max (10913322 vs. 4194304)

Bazel Exit Code:

> 38

Other Info:
- We are using [BuildBuddy](https://github.com/buildbuddy-io/buildbuddy) as our BES backend.
- Our current workaround is to disable BEP for the build in question.

SrodriguezO:
Hey @zachgrayio, thanks for your quick response. Is this something that gets configured on the server or on the client? I found this [grpc-gateway comment](https://github.com/grpc-ecosystem/grpc-gateway/issues/943#issuecomment-500345264) and this [Stack Overflow response](https://stackoverflow.com/a/54139696/10448182) suggesting it was a client-side config.

Maybe it needs to be configured in both places? I can't find references to the grpc MaxXMsgSize configs on the Bazel project though. 

Also, someone ran into this issue [with Buildbarn](https://github.com/buildbarn/bb-deployments/issues/2#issuecomment-489724682) as well (granted, they could also have the bug if it's a server-side thing).

zachgrayio:
In this case it's a missing server option I think (grpc.MaxRecvMsgSize()).

ulfjack:
[...]

PR #14353, opened 2021-11-30
Set --legacy_important_outputs to false by default.

benjaminp:
This flag reduces the largest proto size, which helps avoid sharp edges with remote execution systems (e.g., https://github.com/bazelbuild/bazel/issues/12050).

RELNOTES[INC]: --legacy_important_outputs now has a default of false.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventstream/BuildEventProtocolOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventstream/BuildEventProtocolOptions.java
@@ -24,7 +24,7 @@ public class BuildEventProtocolOptions extends OptionsBase {
 
   @Option(
     name = "legacy_important_outputs",
-    defaultValue = "true",
+    defaultValue = "false",
     documentationCategory = OptionDocumentationCategory.LOGGING,
     effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
     help = "Use this to suppress generation of the legacy important_outputs field in the "


michaeledgar:
This is a good change for a future Bazel release. I'll start importing this change.

haxorz:
@michaeledgar What's the status of this? Can it be merged into Bazel? In internal issue 212691656 I see you already prepped for a change in the default value.

soloffame:
I just now actually sent cl/418966673 out - it slipped through the cracks.

ashleyrichter:
ResultStore customers, including TensorFlow, still depend on this. We can't make this change until we have a migration plan in place for them.

keith:
shouldn't result store users just flip this flag to true then?

ashleyrichter:
We're working on migrating ResultStore off `important_outputs`. That should be done mid-2023.

In the meantime, I've found more users of this functionality that need to be migrated off it.

brentleyjones:
Regardless, those users can set `--legacy_important_outputs` in their `.bazelrc`. Most users would benefit from the flag being flipped. We aren't asking for the flag to be flipped _and removed_, just flipped.

brentleyjones:
cc: @bazelbuild/triage

iancha1992:
cc: @meteorcloudy @Wyverald

ashleyrichter:
I think my team could own the flipping of this flag. I'm not sure how this interacts with Bazel breaking changes policy, since it is a breaking change albeit one that is fixable by reenabling the flag. I will consult BEP owners.

brentleyjones:
It can go into `master` just fine, but we might not be able to cherry-pick it into `7.x` depending on the breaking change policy.

ashleyrichter:
I also got confirmation from wyv@ that this should go into `8.0`, which might be EoY.

Maybe We'll be able to migrate users before then.

PR #14354, opened 2021-11-30
 Use .dylib extension for macOS dynamic libraries in XCode toolchain

### LINKED ISSUE #11082: C++ shared libraries generated on OSX get file extension .so instead of .dylib

fmeum:
This is a follow-up to #14158.

Fixes #11082.

### DIFF
--- a/src/test/shell/bazel/cc_integration_test.sh
+++ b/src/test/shell/bazel/cc_integration_test.sh
@@ -505,8 +505,14 @@ EOF
 
 using namespace std;
 
+#ifdef __APPLE__
+#define DYNAMIC_LIB_EXT "dylib"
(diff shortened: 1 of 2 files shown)


keith (approved):

fmeum:
@oquenchil I forgot about the XCode toolchain in #14158. It's the same change as there.

fmeum:
The tests now pass with only a small change to the `cc_integration_test`.

oquenchil (approved):
LGTM

fmeum:
@oquenchil I don't see the PR at https://bazel-review.googlesource.com yet, has it been imported with the approval?

PR #14366, opened 2021-12-02
Remove JavaScript getting started docs

alexeagle:
They are two years out-of-date and now very inaccurate. They suggest using an approach which rules_nodejs documents as not recommended for new code.


alexeagle:
Note, I'm suggesting removing these docs from docs.bazel.build because I think that community-maintained rulesets should be documented in the community repo instead. It's sufficient that I don't think the Bazel tech writing staff has the resources to maintain documentation like this. I think that https://docs.bazel.build/versions/main/rules.html is sufficient to direct users to the right place to learn about language-specifics of community rules.

aiuto (approved):
This looks fine to me.
@philwo  You have a force push in the PR log.  Are you looking at this?

philwo:
@aiuto I had to force-push the `master` branch, because during the Bazel 4.2.2 release we accidentally ran the release scripts twice, resulting in a duplicated broken commit and a 4.2.2 tag pointing to the wrong commit. First time I see this message on GitHub, I don't think this should impact anything though? (There were no commits yet after the problematic ones and we were very quick fixing this.)

alexeagle:
The force push to master appeared on all PRs which target that branch.

aiuto:
SGTM. I'll merge.

aiuto (approved):

PR #14367, opened 2021-12-02
[apple] fix issues compiling C in objc_library for watchos/armv7k

dflems:
Reproducible example here: https://github.com/dflems/bazel-repro-armv7k. This was tested against 5.0.0 rc2 and latest master and fails on both. The build succeeds after this fix is applied.

Basically, if you have any non-objc files (like C sources) in `objc_library` rules, they'll get compiled for the wrong architecture when targeting watchOS/armv7k.

The target triplet for the `watchos_armv7k` CPU was set to `armv7-apple-watchos` instead of `armv7k-apple-watchos` (which is what Xcode sets when compiling/linking for this architecture). It's been like this for a few years, but I think this has always been wrong.

Before the recent starlark rewrite of the objc rules, it appears that all of the sources (objc or not) were compiled with the `-arch armv7k` arguments applied. After the rewrite, objc still gets compiled this way but C files in the same library get the target triplet from the crosstool, which is incorrect. Then when `libtool` creates the archive, the object files are filtered out because they don't match `-arch_only armv7k`.

If approved, I think this should be cherry-picked into the 5.0 release.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/packages/util/MockObjcSupport.java
+++ b/src/test/java/com/google/devtools/build/lib/packages/util/MockObjcSupport.java
@@ -459,7 +459,7 @@ public final class MockObjcSupport {
         .withCompiler("compiler")
         .withToolchainIdentifier("watchos_armv7k")
         .withHostSystemName("x86_64-apple-ios")
-        .withTargetSystemName("armv7-apple-watchos")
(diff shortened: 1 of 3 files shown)


keith (approved):

keith:
@oquenchil can you review

Wyverald:
Thanks for the fix! I'll import and cherrypick this shortly.

PR #14369, opened 2021-12-03
 Use .dylib extension for macOS dynamic libraries in XCode toolchain

### LINKED ISSUE #11082: C++ shared libraries generated on OSX get file extension .so instead of .dylib

fmeum:
This is a follow-up to #14158.

Fixes #11082.

Closes #14354.

PiperOrigin-RevId: 413885597
(cherrypicked from 6e1b440)

### DIFF
--- a/src/test/shell/bazel/cc_integration_test.sh
+++ b/src/test/shell/bazel/cc_integration_test.sh
@@ -505,8 +505,14 @@ EOF
 
 using namespace std;
 
+#ifdef __APPLE__
+#define DYNAMIC_LIB_EXT "dylib"
+#else
+#define DYNAMIC_LIB_EXT "so"
+#endif
+
 int main() {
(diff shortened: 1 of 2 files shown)


fmeum:
@Wyverald Could you merge #14360 and this PR for Bazel 5?

meteorcloudy:
@oquenchil Is this change safe to get into 5.0?

meteorcloudy:
We decided not to merge this for 5.0, see https://github.com/bazelbuild/bazel/pull/14360#issuecomment-986892875

jsharpe:
@meteorcloudy just so you're aware on macOS python C extensions require a `.so` extension so if this PR means that this can no longer be created then this is a regression and breaking change for anyone compiling python extensions on macOS.

PR #14371, opened 2021-12-03
Add documentation for using code coverage

TLATER:
Hi, sorry for the out-of-the-blue PR, but we've been working on getting code coverage to run in remote execution downstream, and written up these docs.

The basic local workflow is fairly easy (though I think it's still useful to document how `genhtml` is intended to be used), but there are a handful of issues with remote execution. The doc as-is just lists and explains them, as well as workarounds, but we intend to fix at least some those problems over the coming weeks.

The python workflow is also quite rough at the moment, we're looking at making this simpler (see nedbat/coveragepy#587).

Nonetheless, it would be nice to have some review on this in the mean time, so that we can get it merged a bit more quickly, and at least have code coverage documented a bit better even if it is broken in some ways.

I think I may not have been diligent enough with checking whether it builds, so apologies if this just ends up causing a build failure. I hope CI will help a little with this.

### DIFF
--- a/site/docs/coverage.md
+++ b/site/docs/coverage.md
@@ -120,8 +120,84 @@ Java should work out-of-the-box with the default configuration. The
 [bazel toolchains][bazel_toolchains] contain everything necessary for
 remote execution, as well, including JUnit.
 
+### Python
+
+#### Prerequisites
+
+Running coverage with python has some prerequisites:
+
+- A bazel binary that includes [b01c859][python_coverage_commit],
+  which should be any Bazel >3.0.
+- A [modified version of coverage.py][modified_coveragepy].
+<!-- TODO: Upstream an lcov implementation so that this becomes usable  -->
+
+#### Consuming the modified coverage.py
+
+A way to do this is via [rules_python][rules_python], this provides
+the ability to use a `requirements.txt` file, the requirements listed
+in the file are then created as bazel targets using the
+[pip_install][pip_install_rule] repository rule.
+
+The `requirements.txt` should have the following entry:
+
+```text
+git+https://github.com/ulfjack/coveragepy.git@lcov-support
+```
+
+The `rules_python`, `pip_install`, and the `requirements.txt` file should then be used in the WORKSPACE file as:
+
+```python
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_python",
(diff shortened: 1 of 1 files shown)


TLATER:
@lfpino has had a glance over this before, can't seem to request a review from him, so I'll just add him here.

site/docs/coverage.md
```diff
@@ -0,0 +1,203 @@
...
+### Python
+
+#### Prerequisites
+
+Running coverage with python has some prerequisites:
+
+- A bazel binary that includes [b01c859][python_coverage_commit],
+  which should be any Bazel >3.0.
+- A [modified version of coverage.py][modified_coveragepy].
+<!-- TODO: Upstream an lcov implementation so that this becomes usable  -->
```
lfpino, line 132: nit: could you file issues for these TODOs once you get the approval from the Bazel team?

lfpino (approved):

site/docs/coverage.md
TLATER: Yes, of course :)

site/docs/coverage.md
```diff
@@ -0,0 +1,203 @@
...
+  although this of course incurs a heavy cost in terms of test times.
+- `--experimental_split_coverage_postprocessing` and
+  `--experimental_fetch_all_coverage_outputs`
+  - Usually coverage is run as part of the test action, and so by
+    default, we don't get all coverage back as outputs of the remote
+    execution by default. These flags override the default and obtain
+    the coverage data. See [this issue][split_coverage_issue] for more
+    details.
+
+## Language-specific configuration
```
c-mita, line 115: Might be worth leaving a note regarding C++ as well (which should also "just work", assuming the required tools are available).

c-mita (commented):
This looks fine; I just have one comment re. C++, but I don't think it's pressing if you don't feel you can answer it.

site/docs/coverage.md
TLATER: We planned on testing/documenting that too, originally. @lfpino has used it for another project, but it's in a similar state as Python as far as we can tell, with environment variables being thrown around for the tooling. We wanted to make sure we actually have a minimal working example before doing so, though, and it sort of fell under the radar.

It's probably not a lot of work to try and replicate some coverage with C++, but I think I'd rather do that as a follow up - we're kind of at capacity, given holidays and such.

c-mita: SGTM

c-mita (approved):

PR #14376, opened 2021-12-03
Return clang as macOS compiler

keith:
When using the default macOS toolchain you always get a clang compiler
since this is based on Apple's default Xcode installation. If you use a
config setting like this to determine the compiler, you want this to
correctly reflect that, just as if you used BAZEL_COMPILER on other
platforms.

```
config_setting(
    name = "clang_build",
    flag_values = {
        "@bazel_tools//tools/cpp:compiler": "clang",
    },
)
```

### DIFF
--- a/tools/osx/crosstool/BUILD.tpl
+++ b/tools/osx/crosstool/BUILD.tpl
@@ -83,7 +83,7 @@ cc_toolchain_suite(
 [
     cc_toolchain_config(
         name = arch,
-        compiler = "compiler",
+        compiler = "clang",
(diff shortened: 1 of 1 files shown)


keith:
@oquenchil can you review?

trybka:
This is reasonable to me. The only caveat here is that folks who have been checking for the old default of "compiler" may break. I don't really know how to mitigate that.

keith:
@googlewalt bump

googlewalt (approved):

PR #14377, opened 2021-12-03
Sort resources to produce consistent zip and help cache hit

nkoroste:
Start sorting resources prior to zipping them in order to guarantee consistent shasum on the `Symbols.zip` file which in turn will improve cache hit for action that use these files as inputs.

Requires a new version of Android Tools.


nkoroste:
ping

PR #14379, opened 2021-12-04
Add clarification note in bazel dependency file

limdor:
As discussed in #14217 the list of dependencies here if it is seen
by the users could be seen like a blessed list of versions.
Adding this message makes it clear that it is not the case.


aiuto (approved):

aiuto:
I'll merge.

PR #14383, opened 2021-12-06
Add incompatible flag to disable cfg="host" from Starlark

Yannic:
Host transitions are problematic where the host and exec environment are
different (e.g., with Remote Execution). This change adds an incompatible
flag that fails analysis if Starlark rules use `cfg = "host"`.

Migrating is pretty straigt forward (replace `host` with `exec`) and
should be automatable by buildifier (although I haven't tried to do so).


Yannic:
@katre and/or @gregestren PTAL

katre:
@comius is actually working on something similar, so I'm assigning this to him so the two of you can discuss.

comius:
cc @susinmotion

I would attempt the more aggressive path first, where we just change all "host" transitions to "exec". This would reduce the churn substantially.

We're already fixing some issues that happen with that plan (within Bazel).

If this plan fails, this PR becomes viable alternative.

Yannic:
In my experiments, it seemed like tools build in exec config do get `--host_*` flags, so that plan doesn't look obviously broken (at least for local exec, didn't try remote exec yet but I should probably do that). SGTM

I think we should probably still make `cfg="host"` an error eventually even after `cfg="host" == cfg="exec"` (similar to what was done to `cfg="data"` which first became a no-op and was the removed), but we can handle that at a later point.

Should I send the documentation change in a separate PR? That seems useful regardless of whether we add the flag or not.

gregestren:
@susinmotion 's been focusing more on this recently. @susinmotion - do you have recommendations for @Yannic 's latest comment?

susinmotion:
Cool, thanks for thinking about this. A separate PR for documentation sounds great--we're really close to making this change.

I'm torn about making cfg="host" an error after the migration. On the one hand, it will make people's rules less confusing in the long term, since they'll say "exec" when they mean "exec". On the other hand, it causes breakages with things that aren't actually a problem, and that's a bummer for users. Can we do some kind of migration script here instead?

Yannic:
@susinmotion thanks for the update!

> Can we do some kind of migration script here instead?

Yes! Added a `buildifier` warning/fix for this: https://github.com/bazelbuild/buildtools/pull/1053

comius:
Hey, can you resolve the conflicts? Host transition has been removed now, so a flag to completely disable it is useful.

comius (changes requested):
Please resolve the conflicts.

comius:
> Please resolve the conflicts.

Friendly ping.

Yannic:
@comius Updated, PTAL

comius (approved):

src/main/java/com/google/devtools/build/lib/packages/semantics/BuildLanguageOptions.java
```diff
@@ -532,6 +532,17 @@
...
 
+  @Option(
+      name = "incompatible_disable_starlark_host_transitions",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.STARLARK_SEMANTICS,
+      metadataTags = {OptionMetadataTag.INCOMPATIBLE_CHANGE},
+      effectTags = {OptionEffectTag.LOADING_AND_ANALYSIS},
+      help =
+          "If set to true, rule attributes cannot set 'cfg = \"host\"'. Rules should set "
+              + "'cfg = \"host\"' instead.")
```
comius, line 543: Second host should be exec.

Yannic:
```suggestion
              + "'cfg = \"exec\"' instead.")
```

brentleyjones:
@bazel-io flag

PR #14384, opened 2021-12-06
Add mirror link for bazelci_rules

coeuvre:


meteorcloudy (approved):

larsrc-google (approved):
While you're at it, wanna also fix the URLs for `remote_coverage_tools`? Or is that file only available in the mirror somehow?

coeuvre:
That file was generated and [pushed](https://github.com/bazelbuild/bazel/blob/master/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator/build_and_upload.sh) directly to the mirror.

PR #14399, opened 2021-12-09
Remove some aborts from wrapped_clang

keith:
These propagated a misleading error code when failing. Now we always
exit 1 in this case instead. Ideally we would exit with the specific
failure code when possible, but it seems that caused issues in the past.

https://github.com/bazelbuild/bazel/commit/f341e0d376cdb517a2fb6c68d835657c89e46b9a
https://github.com/bazelbuild/bazel/commit/f24958a11e3a54f563429aadc12bce033be047fa

### DIFF
--- a/tools/osx/crosstool/wrapped_clang.cc
+++ b/tools/osx/crosstool/wrapped_clang.cc
@@ -110,7 +110,7 @@ std::vector<const char *> ConvertToCArgs(const std::vector<std::string> &args) {
 
 // Spawns a subprocess for given arguments args. The first argument is used
 // for the executable path.
-void RunSubProcess(const std::vector<std::string> &args) {
+bool RunSubProcess(const std::vector<std::string> &args) {
   std::vector<const char *> exec_argv = ConvertToCArgs(args);
   pid_t pid;
   int status = posix_spawn(&pid, args[0].c_str(), nullptr, nullptr,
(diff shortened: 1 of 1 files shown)


keith:
cc @bbarenblat since you wrote some of the previous versions of this, it would be helpful to know if this would cause similar issues to why this was reverted in the past

bbarenblat:
Alas, I remember very little about this work. There’s probably a record inside Google, but I’m not working there right now and don’t have access to it. @trybka might have more information or a better memory – Tom, do you know why CL 202658759 got rolled back?

trybka:
It looks like it was rolled back due to not propagating error codes. :-)

keith:
Thanks. My change is pretty different from that one (explicitly ignoring errno and just propagating failure), so I think we're probably good

keith:
@sgowroji bump

tools/osx/crosstool/wrapped_clang.cc
```diff
@@ -400,7 +402,10 @@ int main(int argc, char *argv[]) {
     }
```
googlewalt, line 406: Any reason not to get rid of all aborts: lines 171, 353, 399?  Can use exit(EXIT_FAILURE) to short circuit any failure.

keith: Done

googlewalt (approved):

PR #14404, opened 2021-12-10
Add `--experimental_check_external_repository_files`

### LINKED ISSUE #14400: I need a way to not scan all external files on every command

I'd like an equivalent feature to `--noexperimental_check_output_files` but for all external files. These files aren't touched by developers in our CI.

We have a large monorepo at Robinhood with many many external files (tons of `pypi` and `npm` packages). In our CI testing environment we run `bazel test //...` for every commit. We also run other small `bazel run` commands and `bazel query` commands in the test runner script to do various other tasks but each of them currently takes 24 seconds irrespective of their complexity because of the `fsvc.getDirtyKeys` task. In the logs we see things like:
```
211209 17:31:21.590:I 1004 [com.google.devtools.build.lib.skyframe.FilesystemValueChecker.lambda$getDirtyValues$2] Spent 24383190001 nanoseconds checking 2987282 filesystem nodes (10254570 scanned)
```

Make a project with 10M external files. See that no matter what, your `bazel run` and even `bazel info` times will be 24 seconds.

Linux and OSX

```
release 4.2.0
```

(not a public repo)

> Replace these lines with your answer.
>
> Places to look:
> - StackOverflow: http://stackoverflow.com/questions/tagged/bazel
> - GitHub issues: https://github.com/bazelbuild/bazel/issues
> - email threads on https://groups.google.com/forum/#!forum/bazel-discuss

> Replace these lines with your answer.
>
> If the files are large, upload as attachment or provide link.

ptarjan:
This flag mirrors `--experimental_check_output_files` to allow installations with a large number of external files to not scan them for changes on every bazel invocation.

If used in combination with `--experimental_check_output_files` and `--watchfs` you can now avoid scanning the cache for invalidations entirely allowing large projects to keep many things cached while still executing small `bazel run` commands quickly.

Fixes #14400

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java
@@ -189,7 +189,9 @@ public class PackageOptions extends OptionsBase {
     effectTags = {OptionEffectTag.UNKNOWN},
     help =
         "Check for modifications made to the output files of a build. Consider setting "
-            + "this flag to false to see the effect on incremental build times."
+            + "this flag to false if you don't expect these files to change outside of bazel "
+            + "since it will speed up subsequent runs as they won't have to check a "
+            + "previous run's cache."
   )
   public boolean checkOutputFiles;
 
--- a/src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
@@ -38,6 +38,7 @@ import com.google.devtools.build.lib.analysis.ConfiguredTarget;
 import com.google.devtools.build.lib.analysis.ConfiguredTargetValue;
 import com.google.devtools.build.lib.analysis.WorkspaceStatusAction.Factory;
 import com.google.devtools.build.lib.analysis.configuredtargets.RuleConfiguredTarget;
+import com.google.devtools.build.lib.bazel.repository.RepositoryOptions;
 import com.google.devtools.build.lib.bugreport.BugReporter;
 import com.google.devtools.build.lib.buildtool.BuildRequestOptions;
 import com.google.devtools.build.lib.cmdline.LabelConstants;
@@ -268,8 +269,10 @@ public final class SequencedSkyframeExecutor extends SkyframeExecutor {
         tsgm,
         options);
     long startTime = System.nanoTime();
+    RepositoryOptions repoOptions = options.getOptions(RepositoryOptions.class);
+    boolean checkExternalRepositoryFiles = repoOptions == null || repoOptions.checkExternalRepositoryFiles;
     WorkspaceInfoFromDiff workspaceInfo =
-        handleDiffs(eventHandler, packageOptions.checkOutputFiles, options);
+        handleDiffs(eventHandler, packageOptions.checkOutputFiles, checkExternalRepositoryFiles, options);
     long stopTime = System.nanoTime();
     Profiler.instance().logSimpleTask(startTime, stopTime, ProfilerTask.INFO, "handleDiffs");
     long duration = stopTime - startTime;
@@ -328,12 +331,12 @@ public final class SequencedSkyframeExecutor extends SkyframeExecutor {
       dropConfiguredTargetsNow(eventHandler);
       super.lastAnalysisDiscarded = false;
     }
-    handleDiffs(eventHandler, /*checkOutputFiles=*/ false, OptionsProvider.EMPTY);
+    handleDiffs(eventHandler, /*checkOutputFiles=*/ false, /*checkExternalRepositoryFiles=*/ true, OptionsProvider.EMPTY);
   }
 
   @Nullable
   private WorkspaceInfoFromDiff handleDiffs(
-      ExtendedEventHandler eventHandler, boolean checkOutputFiles, OptionsProvider options)
+      ExtendedEventHandler eventHandler, boolean checkOutputFiles, boolean checkExternalRepositoryFiles, OptionsProvider options)
       throws InterruptedException, AbruptExitException {
     TimestampGranularityMonitor tsgm = this.tsgm.get();
     modifiedFiles = 0;
@@ -376,6 +379,7 @@ public final class SequencedSkyframeExecutor extends SkyframeExecutor {
         tsgm,
         pathEntriesWithoutDiffInformation,
         checkOutputFiles,
+        checkExternalRepositoryFiles,
         managedDirectoriesChanged,
         fsvcThreads);
     handleClientEnvironmentChanges();
@@ -467,15 +471,16 @@ public final class SequencedSkyframeExecutor extends SkyframeExecutor {
       TimestampGranularityMonitor tsgm,
       Set<Pair<Root, ProcessableModifiedFileSet>> pathEntriesWithoutDiffInformation,
       boolean checkOutputFiles,
+      boolean checkExternalRepositoryFiles,
       boolean managedDirectoriesChanged,
       int fsvcThreads)
       throws InterruptedException {
 
     ExternalFilesKnowledge externalFilesKnowledge = externalFilesHelper.getExternalFilesKnowledge();
     if (!pathEntriesWithoutDiffInformation.isEmpty()
-        || (externalFilesKnowledge.anyOutputFilesSeen && checkOutputFiles)
-        || repositoryHelpersHolder != null
-        || externalFilesKnowledge.anyFilesInExternalReposSeen
+        || (checkOutputFiles && externalFilesKnowledge.anyOutputFilesSeen)
+        || (checkExternalRepositoryFiles && repositoryHelpersHolder != null)
+        || (checkExternalRepositoryFiles && externalFilesKnowledge.anyFilesInExternalReposSeen)
         || externalFilesKnowledge.tooManyNonOutputExternalFilesSeen) {
       // We freshly compute knowledge of the presence of external files in the skyframe graph. We
       // use a fresh ExternalFilesHelper instance and only set the real instance's knowledge *after*
@@ -507,33 +512,46 @@ public final class SequencedSkyframeExecutor extends SkyframeExecutor {
         diffPackageRootsUnderWhichToCheck.add(pair.getFirst());
       }
 
-      EnumSet<FileType> fileTypesToCheck =
-          EnumSet.of(FileType.EXTERNAL_REPO, FileType.EXTERNAL_IN_MANAGED_DIRECTORY);
+      EnumSet<FileType> fileTypesToCheck = EnumSet.noneOf(FileType.class);
+      Iterable<SkyValueDirtinessChecker> dirtinessCheckers = ImmutableList.of();
+
+      if (!diffPackageRootsUnderWhichToCheck.isEmpty()) {
+        dirtinessCheckers = Iterables.concat(
+            dirtinessCheckers,
+            ImmutableList.of(new MissingDiffDirtinessChecker(diffPackageRootsUnderWhichToCheck)));
+      }
+      if (checkExternalRepositoryFiles && repositoryHelpersHolder != null) {
+        dirtinessCheckers = Iterables.concat(
+            dirtinessCheckers,
+            ImmutableList.of(repositoryHelpersHolder.repositoryDirectoryDirtinessChecker()));
+      }
+      if (checkExternalRepositoryFiles) {
+        fileTypesToCheck =
+            EnumSet.of(FileType.EXTERNAL_REPO, FileType.EXTERNAL_IN_MANAGED_DIRECTORY);
+      }
+      if (externalFilesKnowledge.tooManyNonOutputExternalFilesSeen
+        || !externalFilesKnowledge.nonOutputExternalFilesSeen.isEmpty()) {
+        fileTypesToCheck.add(FileType.EXTERNAL);
+      }
       // See the comment for FileType.OUTPUT for why we need to consider output files here.
       if (checkOutputFiles) {
         fileTypesToCheck.add(FileType.OUTPUT);
       }
-      if (externalFilesKnowledge.tooManyNonOutputExternalFilesSeen
-          || !externalFilesKnowledge.nonOutputExternalFilesSeen.isEmpty()) {
-        fileTypesToCheck.add(FileType.EXTERNAL);
+      if (!fileTypesToCheck.isEmpty()) {
+        dirtinessCheckers = Iterables.concat(
+            dirtinessCheckers,
+            ImmutableList.of(new ExternalDirtinessChecker(tmpExternalFilesHelper, fileTypesToCheck)));
       }
+      Preconditions.checkArgument(!Iterables.isEmpty(dirtinessCheckers));
+
       logger.atInfo().log(
           "About to scan skyframe graph checking for filesystem nodes of types %s",
           Iterables.toString(fileTypesToCheck));
       ImmutableBatchDirtyResult batchDirtyResult;
       try (SilentCloseable c = Profiler.instance().profile("fsvc.getDirtyKeys")) {
-        batchDirtyResult =
-            fsvc.getDirtyKeys(
-                memoizingEvaluator.getValues(),
-                new UnionDirtinessChecker(
-                    Stream.of(
-                            new ExternalDirtinessChecker(tmpExternalFilesHelper, fileTypesToCheck),
-                            new MissingDiffDirtinessChecker(diffPackageRootsUnderWhichToCheck),
-                            repositoryHelpersHolder == null
-                                ? null
-                                : repositoryHelpersHolder.repositoryDirectoryDirtinessChecker())
-                        .filter(Objects::nonNull)
-                        .collect(toImmutableList())));
+        batchDirtyResult = fsvc.getDirtyKeys(
+            memoizingEvaluator.getValues(),
+            new UnionDirtinessChecker(ImmutableList.copyOf(dirtinessCheckers)));
       }
       handleChangedFiles(
           diffPackageRootsUnderWhichToCheck,
--- a/src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
+++ b/src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
@@ -22,16 +22,22 @@ import com.google.devtools.build.lib.blackbox.framework.PathUtils;
 import com.google.devtools.build.lib.blackbox.framework.ProcessResult;
 import com.google.devtools.build.lib.blackbox.junit.AbstractBlackBoxTest;
 import com.google.devtools.build.lib.util.ResourceFileLoader;
+import com.google.devtools.build.lib.util.OS;
+import com.google.testing.junit.testparameterinjector.TestParameter;
+import com.google.testing.junit.testparameterinjector.TestParameterInjector;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 
 /** Tests for managed directories. */
+@RunWith(TestParameterInjector.class)
 public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
   private Random random;
   private Integer currentDebugId;
@@ -51,7 +57,7 @@ public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
   }
 
   @Test
-  public void testNodeModulesDeleted() throws Exception {
+  public void testNodeModulesDeleted(@TestParameter boolean watchFs) throws Exception {
     generateProject();
     buildExpectRepositoryRuleCalled();
     checkProjectFiles();
@@ -60,7 +66,7 @@ public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
     assertThat(nodeModules.toFile().isDirectory()).isTrue();
     PathUtils.deleteTree(nodeModules);
 
-    buildExpectRepositoryRuleCalled();
+    buildExpectRepositoryRuleCalled(false, watchFs);
     checkProjectFiles();
   }
 
@@ -162,16 +168,7 @@ public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
   }
 
   @Test
-  public void testFilesUnderChangedManagedDirectoriesRefreshed() throws Exception {
-    doTestFilesUnderManagedDirectoriesRefreshed(false);
-  }
-
-  @Test
-  public void testFilesUnderChangedManagedDirectoriesRefreshedWatchFs() throws Exception {
-    doTestFilesUnderManagedDirectoriesRefreshed(true);
-  }
-
-  private void doTestFilesUnderManagedDirectoriesRefreshed(boolean watchFs) throws Exception {
+  public void testFilesUnderManagedDirectoriesRefreshed(@TestParameter boolean watchFs) throws Exception {
     generateProject();
     buildExpectRepositoryRuleCalled(false, watchFs);
     checkProjectFiles();
@@ -238,19 +235,8 @@ public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
   }
 
   @Test
-  public void testManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneously()
-      throws Exception {
-    doTestManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneously(false);
-  }
-
-  @Test
-  public void testManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneouslyWatchFs()
-      throws Exception {
-    doTestManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneously(true);
-  }
-
-  private void doTestManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneously(
-      boolean watchFs) throws Exception {
+  public void testManagedDirectoriesSettingsAndManagedDirectoriesFilesChangeSimultaneously(
+      @TestParameter boolean watchFs) throws Exception {
     generateProject();
     buildExpectRepositoryRuleCalled(false, watchFs);
     checkProjectFiles();
@@ -390,6 +376,36 @@ public class ManagedDirectoriesBlackBoxTest extends AbstractBlackBoxTest {
                 "WORKSPACE file can not be a symlink if incrementally updated directories are"
                     + " used."))
         .isTrue();
+  }
+
+  @Test
+  public void testNoCheckFiles(@TestParameter boolean allSkips, @TestParameter boolean watchFs) throws Exception {
+    // On Darwin CI, --watchfs is nondeterministic if this passes or fails
+    Assume.assumeFalse(OS.DARWIN.equals(OS.getCurrent()));
+
+    generateProject();
+    buildExpectRepositoryRuleCalled(false, watchFs);
+    checkProjectFiles();
+
+    Path nodeModules = context().getWorkDir().resolve("node_modules");
+    assertThat(nodeModules.toFile().isDirectory()).isTrue();
+    PathUtils.deleteTree(nodeModules);
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // As compared to testNodeModulesDeleted, we don't check that the external file disappeared with this flag so the build is broken
+    BuilderRunner bazel = bazel(watchFs).withFlags("--noexperimental_check_external_repository_files");
+    if (allSkips) {
+      bazel = bazel.withFlags("--noexperimental_check_output_files");
+    }
+
+    ProcessResult result = bazel.shouldFail().build("//...");
+    assertThat(findPattern(result, "Not found package.json")).isTrue();
+
+    // it doesn't make the file on disk
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // In a perfect world we would be able to fix the build by rebuilding here without the flags, but we don't
+    // invalidate the cache correctly so the server would have to shut down
   }
 
   private void generateProject() throws IOException {
--- a/src/test/shell/bazel/BUILD
+++ b/src/test/shell/bazel/BUILD
@@ -1330,4 +1330,5 @@ sh_test(
         ":test-deps",
         "@bazel_tools//tools/bash/runfiles",
     ],
+    tags = ["no_windows"],
 )
--- a/src/test/shell/bazel/BUILD
+++ b/src/test/shell/bazel/BUILD
@@ -1322,3 +1322,12 @@ sh_test(
         "requires-network",
     ],
 )
+
+sh_test(
+    name = "check_external_files_test",
+    srcs = ["check_external_files_test.sh"],
+    data = [
+        ":test-deps",
+        "@bazel_tools//tools/bash/runfiles",
+    ],
+)
--- a/src/test/shell/bazel/check_external_files_test.sh
+++ b/src/test/shell/bazel/check_external_files_test.sh
@@ -191,7 +191,7 @@ run_override_repository_isnt_affected() {
 test_override_repository_isnt_affected() {
   run_override_repository_isnt_affected "--nowatchfs"
 }
-test_override_repository_isnt_affected() {
+test_override_repository_isnt_affected_with_skips() {
   run_override_repository_isnt_affected "--noexperimental_check_output_files --watchfs"
 }
 
--- a/src/test/shell/bazel/check_external_files_test.sh
+++ b/src/test/shell/bazel/check_external_files_test.sh
@@ -1,0 +1,266 @@
+#!/bin/bash
+#
+# Copyright 2019 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Verify that archives can be unpacked, even if they contain strangely named
+# files.
+
+# --- begin runfiles.bash initialization ---
+set -euo pipefail
+if [[ ! -d "${RUNFILES_DIR:-/dev/null}" && ! -f "${RUNFILES_MANIFEST_FILE:-/dev/null}" ]]; then
+  if [[ -f "$0.runfiles_manifest" ]]; then
+    export RUNFILES_MANIFEST_FILE="$0.runfiles_manifest"
+  elif [[ -f "$0.runfiles/MANIFEST" ]]; then
+    export RUNFILES_MANIFEST_FILE="$0.runfiles/MANIFEST"
+  elif [[ -f "$0.runfiles/bazel_tools/tools/bash/runfiles/runfiles.bash" ]]; then
+    export RUNFILES_DIR="$0.runfiles"
+  fi
+fi
+if [[ -f "${RUNFILES_DIR:-/dev/null}/bazel_tools/tools/bash/runfiles/runfiles.bash" ]]; then
+  source "${RUNFILES_DIR}/bazel_tools/tools/bash/runfiles/runfiles.bash"
+elif [[ -f "${RUNFILES_MANIFEST_FILE:-/dev/null}" ]]; then
+  source "$(grep -m1 "^bazel_tools/tools/bash/runfiles/runfiles.bash " \
+            "$RUNFILES_MANIFEST_FILE" | cut -d ' ' -f 2-)"
+else
+  echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
+  exit 1
+fi
+# --- end runfiles.bash initialization ---
+
+source "$(rlocation "io_bazel/src/test/shell/integration_test_setup.sh")" \
+  || { echo "integration_test_setup.sh not found!" >&2; exit 1; }
+
+case "$(uname -s | tr [:upper:] [:lower:])" in
+msys*|mingw*|cygwin*)
+  declare -r is_windows=true
+  ;;
+*)
+  declare -r is_windows=false
+  ;;
+esac
+
+if "$is_windows"; then
+  export MSYS_NO_PATHCONV=1
+  export MSYS2_ARG_CONV_EXCL="*"
+fi
+
+function get_extrepourl() {
+  if $is_windows; then
+    echo "file:///$(cygpath -m $1)"
+  else
+    echo "file://$1"
+  fi
+}
+
+setup_remote() {
+  WRKDIR=$(mktemp -d "${TEST_TMPDIR}/testXXXXXX")
+  cd "${WRKDIR}"
+
+  mkdir remote
+  (
+    cd remote
+    echo 'genrule(name="g", outs=["go"], cmd="echo GO > $@")' > BUILD
+  )
+  tar cvf remote.tar remote
+  rm -rf remote
+
+  mkdir main
+  cd main
+  cat >> "$(create_workspace_with_default_repos WORKSPACE)" <<EOF
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+  name="remote",
+  strip_prefix="remote",
+  urls=["$(get_extrepourl $WRKDIR)/remote.tar"],
+)
+EOF
+}
+
+setup_local() {
+  WRKDIR=$(mktemp -d "${TEST_TMPDIR}/testXXXXXX")
+  cd "${WRKDIR}"
+
+  mkdir local_rep
+  (
+    cd local_rep
+    create_workspace_with_default_repos WORKSPACE
+    echo 'genrule(name="g", outs=["go"], cmd="echo GO > $@")' > BUILD
+  )
+
+  mkdir main
+  cd main
+  cat >> "$(create_workspace_with_default_repos WORKSPACE)" <<EOF
+local_repository(
+  name="local_rep",
+  path="../local_rep",
+)
+EOF
+}
+
+test_check_external_files() {
+  setup_remote
+  bazel build @remote//:g >& "$TEST_log" || fail "Expected build to succeed"
+
+  echo "broken file" > bazel-main/external/remote/BUILD
+  # The --noexperimental_check_external_repository_files flag doesn't notice the file is broken
+  bazel build --noexperimental_check_external_repository_files @remote//:g >& "$TEST_log" || fail "Expected build to succeed"
+
+  bazel build @remote//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_log "no such target '@remote//:g'"
+}
+
+test_check_all_flags_fast() {
+  setup_remote
+  msg="About to scan skyframe graph checking for filesystem nodes"
+
+  bazel build --watchfs @remote//:g >& "$TEST_log" || fail "Expected build to succeed"
+  instances=$(grep -c "$msg" "$(bazel info server_log)")
+  [[ $instances -eq 1 ]] || fail "Should have only been 1 instance, got $instances"
+
+  echo "broken file" > bazel-main/external/remote/BUILD
+
+  bazel build \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& "$TEST_log" || fail "Expected build to succeed"
+
+  instances=$(grep -c "$msg" "$(bazel info server_log)")
+  [[ $instances -eq 1 ]] || fail "Should have only been 1 instance (from the first build), got $instances"
+}
+
+run_local_repository_isnt_affected() {
+  local -r extra_args="$1"
+  shift
+
+  setup_local
+  bazel build @local_rep//:g >& "$TEST_log" || fail "Expected build to succeed"
+
+  echo "broken file" > ../local_rep/BUILD
+  # The --noexperimental_check_external_repository_files flag still notices the file is broken
+  bazel build \
+    --noexperimental_check_external_repository_files \
+    $extra_args \
+    @local_rep//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  bazel build --noexperimental_check_external_repository_files @local_rep//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_log "no such target '@local_rep//:g'"
+}
+
+test_local_repository_isnt_affected() {
+  run_local_repository_isnt_affected "--nowatchfs"
+}
+
+test_local_repository_isnt_affected_with_skips() {
+  run_local_repository_isnt_affected "--noexperimental_check_output_files --watchfs"
+}
+
+run_override_repository_isnt_affected() {
+  local -r extra_args="$1"
+  shift
+
+  setup_local
+  create_workspace_with_default_repos WORKSPACE
+  bazel build @local_rep//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_log "no such package '@local_rep//'"
+
+  argv="--override_repository=local_rep=$(pwd)/../local_rep"
+  bazel build "$argv" $extra_args @local_rep//:g >& "$TEST_log" || fail "Expected build to succeed"
+
+  echo "broken file" > ../local_rep/BUILD
+  # The --noexperimental_check_external_repository_files flag still notices the file is broken
+  bazel build \
+    --noexperimental_check_external_repository_files \
+    "$argv" \
+    $extra_args \
+    @local_rep//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_log "no such target '@local_rep//:g'"
+}
+
+test_override_repository_isnt_affected() {
+  run_override_repository_isnt_affected "--nowatchfs"
+}
+test_override_repository_isnt_affected() {
+  run_override_repository_isnt_affected "--noexperimental_check_output_files --watchfs"
+}
+
+test_no_fetch_then_fetch() {
+  setup_remote
+  bazel build \
+    --nofetch \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_log "no such package '@bazel_tools//tools/build_defs/repo'"
+  bazel build \
+    --fetch \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& "$TEST_log" || fail "Expected build to pass"
+}
+
+test_no_build_doesnt_break_the_cache() {
+  setup_remote
+  bazel build \
+    --nobuild \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& "$TEST_log" || fail "Expected build to pass"
+  [[ ! -f bazel-main/external/remote/BUILD ]] || fail "external files shouldn't have been made"
+  bazel build \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& "$TEST_log" || fail "Expected build to pass"
+}
+
+test_symlink_outside_still_checked() {
+  mkdir main
+  cd main
+  create_workspace_with_default_repos WORKSPACE
+  echo 'sh_test(name = "symlink", srcs = ["symlink.sh"])' > BUILD
+
+  mkdir ../foo
+  echo 'exit 0' > ../foo/foo.sh
+  chmod u+x ../foo/foo.sh
+  ln -s ../foo/foo.sh symlink.sh
+
+  bazel test \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    :symlink >& "$TEST_log" || fail "Expected build to succeed"
+
+  bazel test \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    :symlink >& "$TEST_log" || fail "Expected build to succeed"
+  expect_log '//:symlink.*cached'
+
+  echo 'exit 1' > ../foo/foo.sh
+
+  bazel test \
+    --noexperimental_check_external_repository_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    :symlink >& "$TEST_log" && fail "Expected build to fail" || true
+  expect_not_log '//:symlink.*cached'
+  expect_log '1 test FAILED'
+}
+
+run_suite "check_external_files tests"
--- a/src/main/java/com/google/devtools/build/lib/bazel/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/BUILD
@@ -30,6 +30,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/lib/bazel/bzlmod:resolution_impl",
         "//src/main/java/com/google/devtools/build/lib/bazel/commands",
         "//src/main/java/com/google/devtools/build/lib/bazel/repository",
+        "//src/main/java/com/google/devtools/build/lib/bazel/repository:repository_options",
         "//src/main/java/com/google/devtools/build/lib/bazel/repository/cache",
         "//src/main/java/com/google/devtools/build/lib/bazel/repository/downloader",
         "//src/main/java/com/google/devtools/build/lib/bazel/repository/starlark",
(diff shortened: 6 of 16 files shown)


ptarjan:
cc @haxorz and you worked with `--noexperimental_check_output_files` in 60108839363 (6 years ago!)

ptarjan:
Thinking a bit more about this, I can't see a reason why we would want to scan the filesystem for changes for anything from an `http_archive` or `git_repository`. Yes, for `local_repository` we absolutely should check the filesystem, but should I not make this a flag and instead just skip it all the time for those `bazel`-owned files?

ptarjan:
@alexeagle pointed out that maybe scanning the file system by default is good for build reproducibility and this flag is only valuable in CI. Starting with a flag is the least impact so I guess starting here makes sense.

keith:
What about if you're using `--override_repository`? In that case does this disabling scanning for changes in those? In which case developing rule sets would be affected if you passed this

keith:
it might be nice to carve out that case if possible since then you could avoid that distinction. We use `experimental_check_output_files` for developers as well to reduce incremental build time, and would similarly like to use this if it work'd well enough that someone would have to be doing something quite weird to break it.

ptarjan:
Interesting point. I'm scared to enable it in dev as it won't guarantee
hermetic builds but I'm happy to exclude those since we obviously won't be
using them in CI.

Having said that, the main speedup here is not walking the whole cache
structure at all rather than skipping the filesystem stats so you'll lose
the speed if you have any of those overrides at all. So enabling it in dev
won't give you much.

Would the maintainers want me to set a bit if you have those and still do
the scan if they are present? I'm inclined to do that as a follow-up if
needed since that experience will be much slower.

Wyverald:
Hi Paul, I'm following along the discussion. Unfortunately I'm not quite familiar with this scanning code, so I was hoping you could reach consensus with Keith before devoting time to understand this part of the codebase. So I don't really have much advice to offer (yet).

Let's see what Keith thinks, and I'll try to carve out some time tomorrow to look at this.

keith:
Having this is clearly an improvement for the stated use case, so I think it's safe to merge now, but for folks who use `local_repository` or `--override_repository` I assume they won't be able to enable it locally, even if normally this would feel safe to enable locally if you can "trust" developers not to mess with files in the bazel root. I think if this is feasible it could be done later.

ptarjan:
Thanks for being flexible! I’ll work on a second follow up PR which at
least doesn’t allow you to use those two features together and if I can
figure out a way to distinguish files that are overridden vs owned by bazel
I’ll still allow skipping the stating since that saves some time.

ptarjan:
@Wyverald is there anything else on your end that I need to do to get this PR reviewed?

ptarjan:
@keith I added a warning for `--override_repository` here. I don't see an easy way to do something similar for `--local_repository` but I'll hunt around a bit more.

Wyverald:
Sorry about the delay -- I'll import this shortly.

src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
```diff
@@ -360,6 +360,11 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
       }
 
       if (repoOptions.repositoryOverrides != null) {
+        if (pkgOptions != null && !pkgOptions.checkExternalFiles) {
+          env.getReporter()
+              .handle(Event.warn("--noexperimental_check_external_files is making it so we don't check the disk for changes to your repositories overriden with --override_repository"));
+        }
```
janakdr, line 366: nit: this is phrased a little strangely to my eye. I might say instead:

"Changes to external repositories overridden by --override_repository will not be detected because --noexperimental_check_external_files is set".

src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
```diff
@@ -360,6 +360,11 @@ public void beforeCommand(CommandEnvironment env) throws AbruptExitException {
       }
 
       if (repoOptions.repositoryOverrides != null) {
+        if (pkgOptions != null && !pkgOptions.checkExternalFiles) {
+          env.getReporter()
+              .handle(Event.warn("--noexperimental_check_external_files is making it so we don't check the disk for changes to your repositories overriden with --override_repository"));
```
janakdr, line 365: Does this actually only apply to repositories overridden with override_repository, or any external repositories on disk? Sorry for not knowing the difference.

src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java
```diff
@@ -189,10 +189,23 @@ public ParallelismConverter() throws OptionsParsingException {
...
     help =
         "Check for modifications made to the output files of a build. Consider setting "
-            + "this flag to false to see the effect on incremental build times."
+            + "this flag to false if you don't expect these files to change outside of bazel "
+            + "since it will speed up incremental build times."
   )
   public boolean checkOutputFiles;
 
+  @Option(
+    name = "experimental_check_external_files",
```
janakdr, line 198: Can we add a test (ideally a Java integration test) for this functionality?

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,14 +519,20 @@ private void handleDiffsWithMissingDiffInformation(
...
+        }
+      }
       // See the comment for FileType.OUTPUT for why we need to consider output files here.
       if (checkOutputFiles) {
         fileTypesToCheck.add(FileType.OUTPUT);
       }
-      if (externalFilesKnowledge.tooManyNonOutputExternalFilesSeen) {
-        fileTypesToCheck.add(FileType.EXTERNAL);
+      if (!checkExternalFiles && !checkOutputFiles && diffPackageRootsUnderWhichToCheck.isEmpty()) {
+        throw new AssertionError("should have already early exited");
```
janakdr, line 535: nit: It's a little more idiomatic IMO to do a Preconditions.checkState call here

janakdr (commented):
Sorry to jump in here with a last-minute review: was assigned to review the import of the PR, and had these comments.

src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java
ptarjan: Sure! Can you point me at a test which is a good one to copy? I see lots of different types of ones (but none for the param which I'm copying here)

src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java
ptarjan: This branch of the code (if I understand it correctly) is only ran when `--override-repository` is passed in. At least that was my intention (and I don't see the warning when using this bazel on my repository which doesn't use overrides).

src/main/java/com/google/devtools/build/lib/pkgcache/PackageOptions.java
janakdr: One is here: https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/skyframe/OutputsInvalidationIntegrationTest.java. You should be able to do something pretty straightforward in test/.../lib/bazel/repository where you inherit from BuildIntegrationTestCase, do a trivial build with a repository, modify a file, and observe no change when you do a second build.

ptarjan: Thanks for the pointers. I've been bashing my head against it for a while. I have something semi-working but it isn't registering the repo. I have this file:
```
package com.google.devtools.build.lib.bazel.repository;

import static com.google.common.truth.Truth.assertThat;

import com.google.devtools.build.lib.buildtool.BuildResult;
import com.google.devtools.build.lib.buildtool.util.BuildIntegrationTestCase;
import com.google.devtools.build.lib.testutil.MoreAsserts;
import com.google.devtools.build.lib.vfs.FileSystemUtils;
import com.google.testing.junit.testparameterinjector.TestParameterInjector;
import com.google.testing.junit.testparameterinjector.TestParameters;
import org.junit.Test;
import org.junit.runner.RunWith;

@RunWith(TestParameterInjector.class)
public class ExternalsInvalidationIntegrationTest extends BuildIntegrationTestCase {

  @TestParameters({
      "{checkExternalFiles: true}",
      "{checkExternalFiles: false}",
  })
  @Test
  public void noCheckExternalFiles_ignoresModifiedFile(boolean checkExternalFiles) throws Exception {
    addOptions("--experimental_check_external_files="+checkExternalFiles);
    write("WORKSPACE", FileSystemUtils.readLinesAsLatin1(getWorkspace().getRelative("WORKSPACE"))
        + "\n"
        + "load(\"@bazel_tools//tools/build_defs/repo:http.bzl\", \"http_archive\")\n"
            + "http_archive(\n"
            + "    name = \"rules_python\",\n"
            + "    url = \"https://github.com/bazelbuild/rules_python/releases/download/0.5.0/rules_python-0.5.0.tar.gz\",\n"
            + "    sha256 = \"cd6730ed53a002c56ce4e2f396ba3b3be262fd7cb68339f0377a45e8227fe332\",\n"
            + ")");
    write("foo/BUILD", "load(\"@rules_python//python:defs.bzl\", \"py_binary\")\n"
        + "py_binary(name = 'foo', srcs = ['main.py'])");
    write("foo/main.py", "print('hi')");
    buildTarget("//foo");
    String scanString = "About to scan skyframe graph checking for filesystem nodes of types";
    MoreAsserts.assertContainsEvent(events.collector(), scanString);

    events.collector().clear();
    writeAbsolute(getOutputBase().getRelative("external/rules_python/python/defs.bzl"), "a broken file");
    BuildResult result = buildTarget("//foo");
    assertThat(result.getSuccess()).isEqualTo(checkExternalFiles);

    MoreAsserts.assertDoesNotContainEvent(events.collector(),  scanString);
  }

}
```

but when running it I get:
```
com.google.devtools.build.lib.skyframe.BzlLoadFunction$BzlLoadFunctionException: com.google.devtools.build.lib.skyframe.BzlLoadFunction$BzlLoadFailedException: Unable to find package for @rules_python//python:defs.bzl: The repository '@rules_python' could not be resolved: Repository '@rules_python' is not defined.
```

ptarjan: Next up I'll editing a `local_repository` which is pre-registered but @keith wanted that to not be skipped by this flag so I was hoping to avoid that since the semantics for that aren't locked in yet.

ptarjan: Oh I see, it is all mocked out in `BazelAnalysisMock.java`

ptarjan: I banged my head against this for a few more hours and got something to execute, but no matter what changing that file is observed by the cache. Maybe `local_repository` isn't cached? Anyways, this test is 100x harder to write than the original code :( 

```
package com.google.devtools.build.lib.bazel.repository;

import static com.google.common.truth.Truth.assertThat;

import com.google.devtools.build.lib.buildtool.BuildResult;
import com.google.devtools.build.lib.buildtool.util.BuildIntegrationTestCase;
import com.google.devtools.build.lib.testutil.MoreAsserts;
import com.google.devtools.build.lib.testutil.Scratch;
import com.google.devtools.build.lib.vfs.FileSystemUtils;
import com.google.devtools.build.lib.vfs.ModifiedFileSet;
import com.google.devtools.build.lib.vfs.Path;
import com.google.devtools.build.lib.vfs.Root;
import com.google.testing.junit.testparameterinjector.TestParameterInjector;
import com.google.testing.junit.testparameterinjector.TestParameters;
import java.util.logging.Logger;
import org.junit.Test;
import org.junit.runner.RunWith;

@RunWith(TestParameterInjector.class)
public class ExternalsInvalidationIntegrationTest extends BuildIntegrationTestCase {

  @TestParameters({
      "{checkExternalFiles: true}",
      "{checkExternalFiles: false}",
  })
  @Test
  public void noCheckExternalFiles_ignoresModifiedFile(boolean checkExternalFiles) throws Exception {
    addOptions("--noexperimental_check_output_files");
    addOptions("--experimental_check_external_files="+checkExternalFiles);
    write("fake_external/pkg/my_alias.bzl",
        "def my_alias(name, other):",
        "    print('my_alias ran')");
    write("fake_external/WORKSPACE", "workspace(name = 'fake_external')");
    write("fake_external/pkg/BUILD", "exports_files(['my_alias.bzl'])");
    write("WORKSPACE", FileSystemUtils.readLinesAsLatin1(getWorkspace().getRelative("WORKSPACE"))
        + "\n"
        + "local_repository(name = 'fake_external', path = 'fake_external')\n"
    );
    write("foo/BUILD", "load(\"@fake_external//pkg:my_alias.bzl\", \"my_alias\")\n"
        + "genrule(name='foo', outs=['foo.out'], cmd='echo genrule ran; touch $@')\n"
        + "my_alias(name='bar', other='foo')");
    buildTarget("//foo");

    writeAbsolute(getOutputBase().getRelative("external/fake_external/pkg/my_alias.bzl"), "a broken file");
    try {
      buildTarget("//foo");
      if (!checkExternalFiles) {
        throw new Exception("Should have raised");
      }
    } catch (Exception e) {
      if (checkExternalFiles) {
        throw new Exception("Should not have raised");
      }
    }
  }

}
```

janakdr:
Sorry for the issues, @ptarjan . It does look like our Java integration framework does not properly support repositories. cc @michajlo who might know where to route an issue to track fixing that. Could you try adding a shell test instead, in file underneath src/test/shell/bazel? There are a number of tests with repositories there, so it shouldn't be too onerous to add one with your flag.

Also, just confirming, when you run with a custom-built Bazel and your flag, you do observe the correct behavior (that is, the Java test is not picking up an actual bug)?

ptarjan:
Thanks for the pointers! I'll try a shell test. 

> Also, just confirming, when you run with a custom-built Bazel and your flag, you do observe the correct behavior (that is, the Java test is not picking up an actual bug)?
Yes. I'm using my patch in production for our build pipelines and it is very happily running builds for the last few days. And the speeds for builds where the cache is full populated is down to 100ms from 24 seconds, so it has the desired effect.

ptarjan:
Thanks for the pointer to that test. That was much easier to write!

ptarjan:
@keith you'll be happy to know that this change doesn't affect `local_repository` nor `--override_repository`. I added a test showing that. So you should be good to use this flag locally.

keith:
Thanks!

src/test/shell/bazel/check_external_files.sh
```diff
@@ -0,0 +1,124 @@
...
+    --noexperimental_check_external_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g || fail "Expected build to succeed"
+
+  cat $(bazel info server_log) | \
+    grep 'Skipping scanning the filesystem for cache invalidations, yay' || fail "Should have found text"
+}
+
+test_local_repository_isnt_affected() {
```
janakdr, line 101: what happens for this and the following test if you also pass `--nocheck_experimental_output_files` and run with `--watchfs`?

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -376,13 +376,18 @@ private WorkspaceInfoFromDiff handleDiffs(
...
     handleDiffsWithCompleteDiffInformation(
         tsgm, modifiedFilesByPathEntry, managedDirectoriesChanged, fsvcThreads);
-    handleDiffsWithMissingDiffInformation(
-        eventHandler,
-        tsgm,
-        pathEntriesWithoutDiffInformation,
-        checkOutputFiles,
-        managedDirectoriesChanged,
-        fsvcThreads);
+    if (checkOutputFiles || checkExternalFiles || !pathEntriesWithoutDiffInformation.isEmpty()) {
```
janakdr, line 379: Sorry I didn't notice this before:

Unfortunately the github PR interface doesn't let me comment on unchanged lines, but I'm a bit concerned about lines 493-498 in this PR: the guard of when to check inside handleDiffs seems to be both stricter and looser than this check. In substance, this flag seems to be inconsistently assuming that "external repository" files should count as "external" files. Do you want them to? I would naively think that you would want this flag only for external non-repository files. I don't think it should apply to both non-repository and repository files, we'd probably want separate flags. Here are the inconsistencies I see:

(1) This check doesn't respect the value of customDirtinessCheckers. I think the only place that's set is in BazelRepositoryModule, with a RepositoryDirectoryDirtinessChecker. Based on that class's javadoc, it looks like this check would lead to a bug where a `--nofetch` build followed by a `--fetch` build didn't properly update repositories when running with `--watchfs`, `--noexperimental_check_external_files` and `--noexperimental_check_output_files`.

(2) Similarly, it looks like *either* checkExternalFiles should make us ignore externalFilesKnowledge.anyFilesInExternalRepoSeen or it should be overridden. Right now, there's a redundant check, which I think you can exercise by doing a `--nowatchfs build --nobuild` on a project with a file in an external repository, then on an incremental build do `--nowatchfs build --nocheck_external_files` , and having a file in an external repository. That will do a full filesystem scan. However, running with `--watchfs` (and `--noexperimental_check_output_files`) will skip the full filesystem scan. That seems wrong. Also, anyFIlesInExternalRepoSeen can refer either to "normal" external repository files (which maybe you want to suppress checking for) and to "managed directories" (see ExternalFilesHelper) which is a pretty tricky concept. Is it actually the case that you want to disable checks in managed directories with this flag, even if you do want to suppress repository checks? I think probably not, although I'm not positive when managed directories are expected to be updated by the corresponding repository rules. If you don't want to disable such checks, you'll probably need to split the ExternalFilesKnowledge boolean into two, and track them separately in ExternalFilesHelper.

src/test/shell/bazel/BUILD
```diff
@@ -1360,3 +1360,10 @@ sh_test(
         "requires-network",
     ],
 )
+
+sh_test(
+    name = "check_external_files",
+    srcs = ["check_external_files.sh"],
+    data = [":test-deps"],
+    tags = ["no_windows"],
```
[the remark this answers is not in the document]
ptarjan: I was just cargo-culting most of the other tests in this package. I'll remove and try it!

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -333,12 +333,12 @@ public void handleDiffsForTesting(ExtendedEventHandler eventHandler)
       dropConfiguredTargetsNow(eventHandler);
       super.lastAnalysisDiscarded = false;
     }
-    handleDiffs(eventHandler, /*checkOutputFiles=*/ false, OptionsProvider.EMPTY);
+    handleDiffs(eventHandler, /*checkOutputFiles=*/ false, /*checkExternalFiles=*/ true, OptionsProvider.EMPTY);
```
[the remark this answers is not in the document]
ptarjan: I was just maintaining the status quo. Keeping it to true means that the builds are more hermetic but will be slower with large number of externals.

src/test/shell/bazel/check_external_files.sh
ptarjan: The `local_repository()` one passes but the `--override_repository` one won't notice the file changed with the current implementation.

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
ptarjan: Thank you for the catch!

> In substance, this flag seems to be inconsistently assuming that "external repository" files should count as "external" files. Do you want them to? 

I don't know the difference. In my case I want to not scan any files which a normal users would think are owned by bazel. In my case, that is a bunch of `http_archive`s and packages registered by `rules_python` and `rules_nodejs`.  Basically things owned by repository rules which I don't manage the files for. I'm assuming that is everything in the `externals/` folder? It was surprising to me that these needed to be scanned for changes on every invocation (especially in CI where my users don't have direct filesystem writing and where doing anything outside of the bazel sandbox is very unorthodox).

Thanks for those extra cases! I'll add tests for those and look at updating this logic to match the same as are on 493-498. I'm assuming I'd want to check my flag in that section too just like `checkOutputFiles` is used there

ptarjan: I added a test for (1) and it seems to pass. Maybe I didn't understand the case you were trying to get at:
```
  test_no_fetch_then_fetch() {
    setup_remote
    bazel build \
      --nofetch \
      --noexperimental_check_external_files \
      --noexperimental_check_output_files \
      --watchfs \
      @remote//:g >& $TEST_log && fail "Expected build to fail" || true
    expect_log "no such package '@bazel_tools//tools/build_defs/repo'"
    bazel build \
      --fetch \
      --noexperimental_check_external_files \
      --noexperimental_check_output_files \
      --watchfs \
      @remote//:g >& $TEST_log || fail "Expected build to pass"
  }
```

ptarjan: And for (2) I didn't really follow. I tried adding a test which I think did as you said but the second build still works. Did you want me to edit the filesystem in the meantime and show that the scan didn't happen? Or did you think it should have done a scan for that?

ptarjan: Just doing a normal build from my tests there is a `com.google.devtools.build.lib.rules.repository.RepositoryDirectoryDirtinessChecker@6c76d4a0` in that variable. Can I short circuit that if my flag is set?

ptarjan:
The windows build is failing to find the support file that is used in this dir https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/9954#db7d5072-8a72-4ad1-9832-141eb57c11d7

src/test/shell/bazel/check_external_files.sh
ptarjan: With your suggestion of just using that same guard they both ignore the flag now

ptarjan:
@janakdr I think I'm going to need some direction as I've been banging my head against this for many hours now. I think I incorporated your feedback by inlining the check for `checkExternalFiles` into the corresponding `if` statements instead of skipping the whole thing. In the process I tried to break apart the checkers so only the applicable checkers will run (as that seemed the least surprising to me). That wasn't mandatory for my goals so I'm happy to remove that if you'd like.

Having done this, I added a test to the `ManagedDirectoriesBlackBoxTest` since getting a `managed_directory` setup working in my test was hard. But what I found was that if I ran with my flag, then the next run without my flag still wouldn't check the removed managed directory. So I'm assuming that with my flag I'm marking it that I already scanned that directory when in reality I didn't. Do you know where this could be? Or is there something else in that function that I should be skipping doing with my flag so subsequent builds have to keep scanning?

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +392,50 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+      bazel = bazel.withFlags("--noexperimental_check_output_files");
+    }
+    ProcessResult result = bazel.shouldFail().build("//...");
+    assertThat(findPattern(result, "Not found package.json")).isTrue();
+
+    // and it doesn't make the file
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // but a normal build fixes it
+    // TODO: These don't check the directories and they should
```
ptarjan, line 434: @janakdr here is the subsequent build that should re-scan the `node_modules` and fix it, but instead it just leaves it broken

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +392,50 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+    // buildExpectRepositoryRuleCalled(false, watchFs);
```
ptarjan, line 435: @janakdr here is the subsequent build that should re-scan the `node_modules` and fix it, but instead it just leaves it broken

janakdr (commented):
It sounds like this flag is only intended for external *repositories*, not external files in general. In that case, I have a few suggestions. First, change the name to experimental_check_external_repository_files. Second, move the flag to RepositoryOptions. You can still access it in SequencedSkyframeExecutor via the `options` variable.

About the test cases, for (1) the --fetch flag seems to be in a pretty sorry state, so I'm not sure what the semantics actually are right now, and I don't know that it's worth digging too much deeper. My suspicion is that something else is causing invalidation (maybe a transient error). You might be able to trigger it by having a repository that is fetched, but then is updated to a new version and the build is run with --nofetch, so the updated version isn't actually used. But I'm not positive. Any Google folks who actually maintain this functionality are welcome to weigh in. Maybe @Wyverald knows?

For (2), I think that you *don't* want to affect the checks for external files that are not in repositories. That specifically means that ExternalFilesKnowledge#tooManyNonOutputExternalFilesSeen and ExternalFilesKnowledge#tooManyNonOutputExternalFilesSeen#nonOutputExternalFilesSeen should not be connected to your flag at all. You can enforce that in a test by just creating a symlink to a random other file on disk, modifying the file, and verifying that we rebuild even with all checks disabled.

For (3), I suspect that the issue is that on line 598 (in your most recent commit), we set external files knowledge for the graph based on the "scan" that just happened. That setting should probably be conditional, or be a union, if we didn't actually check the whole graph above.

ptarjan:
Thanks for the reply. I’ll go through it more thoroughly soon. I will add
that my main goal with the flag is to avoid walking the cache at all since
that is the slow part of our executions.

janakdr:
Apologies for (again) thinking of this quite late, but my initial comment was asking for too much: I don't think that with `--noexperimental_check_output_files` we detect the case of: build, modify output file, build without checking outputs, *build checking outputs and discover modified file. So I don't think your new flag needs to do that either. You can leave your existing test cases if you like, but I think we basically don't support switching such flags back and forth without shutting down the Bazel server or taking some other action that invalidates the in-memory cache.

src/test/shell/bazel/check_external_files_test.sh
```diff
@@ -0,0 +1,194 @@
...
+  expect_log "no such package '@bazel_tools//tools/build_defs/repo'"
+  bazel build \
+    --fetch \
+    --noexperimental_check_external_files \
+    --noexperimental_check_output_files \
+    --watchfs \
+    @remote//:g >& $TEST_log || fail "Expected build to pass"
+}
+
+test_no_build_then_nocheck() {
```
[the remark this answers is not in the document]
ptarjan: I added a check to show what I was testing. I was trying to show (as per your original comment) that using the `--no-build` doesn't negatively cache something which breaks the build, even with my new flag).

ptarjan:
@janakdr Thank you for the thorough review! Yeah, it does seem the state of cache invalidation for the other flag that I'm copying isn't great either, so maybe we can file a ticket somewhere to get them both shored up at some point? Would that be an issue on github here?

I do like the test cases I have out there as they make it explicit what the behavior is now. I tried to leave comments in the areas where the current behavior is unexpected so a future person can use them to validate the cache invalidation got better. I also like my refactor of the dirtyness checker to only run the needed checkers as it gives me a good spot to put my invariant that you are walking the whole cache for a reason.

Given all that, do you think this is ready for merging or is there something else I should tackle? Thanks again.

src/test/shell/bazel/BUILD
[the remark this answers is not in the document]
ptarjan: Sadly it still fails with the same error. E.g. https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/9978#2528ac52-3dac-4bae-a8d3-0c26a33783dc
```
Executing tests from //src/test/shell/bazel:check_external_files_test
C:\\b\\sk77rbjx\\execroot\\io_bazel\\bazel-out\\x64_windows-fastbuild\\bin\\src\\test\\shell\\bazel\\check_external_files_test: line 21: /c/b/sk77rbjx/execroot/io_bazel/bazel-out/x64_windows-fastbuild/bin/src/test/shell/bazel/../integration_test_setup.sh: No such file or directory
integration_test_setup.sh not found!
```

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,29 +515,40 @@ private void handleDiffsWithMissingDiffInformation(
...
+        }
+      }
       // See the comment for FileType.OUTPUT for why we need to consider output files here.
       if (checkOutputFiles) {
         fileTypesToCheck.add(FileType.OUTPUT);
       }
-      if (externalFilesKnowledge.tooManyNonOutputExternalFilesSeen) {
-        fileTypesToCheck.add(FileType.EXTERNAL);
+      if (checkExternalFiles || checkOutputFiles) {
+        Preconditions.checkArgument(!fileTypesToCheck.isEmpty());
```
ptarjan, line 539: I used `checkArgument` since this variable is derived from the argument and that's what the docs seemed to want. Let me know if I should use `checkState` instead.

ptarjan:
@janakdr gentle ping now that the holidays are nearing a close

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,29 +515,40 @@ private void handleDiffsWithMissingDiffInformation(
...
+      EnumSet<FileType> fileTypesToCheck = EnumSet.noneOf(FileType.class);
+      Iterable<SkyValueDirtinessChecker> dirtynessCheckers = ImmutableList.of();
+
+      if (!diffPackageRootsUnderWhichToCheck.isEmpty()) {
+        dirtynessCheckers = Iterables.concat(
+            dirtynessCheckers,
+            ImmutableList.of(new MissingDiffDirtinessChecker(diffPackageRootsUnderWhichToCheck)));
+      }
+      if (checkExternalFiles) {
+        dirtynessCheckers = Iterables.concat(dirtynessCheckers, customDirtinessCheckers);
```
janakdr, line 527: It is currently true that this custom checker is only used for external files, but it isn't known locally. Let me see if I can clean this up.

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,29 +515,40 @@ private void handleDiffsWithMissingDiffInformation(
...
+        fileTypesToCheck =
+            EnumSet.of(FileType.EXTERNAL_REPO, FileType.EXTERNAL_IN_MANAGED_DIRECTORY);
+        if (externalFilesKnowledge.tooManyNonOutputExternalFilesSeen) {
+          fileTypesToCheck.add(FileType.EXTERNAL);
+        }
```
janakdr, line 532: As I've noted elsewhere, general "external" files should not be mixed up with "external repository" files, they are unrelated.

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
janakdr: I don't generally care to distinguish them. not a big deal either way

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,29 +515,40 @@ private void handleDiffsWithMissingDiffInformation(
         diffPackageRootsUnderWhichToCheck.add(pair.getFirst());
       }
 
-      EnumSet<FileType> fileTypesToCheck =
-          EnumSet.of(FileType.EXTERNAL_REPO, FileType.EXTERNAL_IN_MANAGED_DIRECTORY);
+      EnumSet<FileType> fileTypesToCheck = EnumSet.noneOf(FileType.class);
+      Iterable<SkyValueDirtinessChecker> dirtynessCheckers = ImmutableList.of();
```
janakdr, line 519: nit: I think dirt*i*nessCheckers reads better and is more consistent with the other types and variables here.

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
```diff
@@ -513,29 +515,40 @@ private void handleDiffsWithMissingDiffInformation(
...
-        batchDirtyResult =
-            fsvc.getDirtyKeys(
-                memoizingEvaluator.getValues(),
-                new UnionDirtinessChecker(
-                    Iterables.concat(
-                        customDirtinessCheckers,
-                        ImmutableList.<SkyValueDirtinessChecker>of(
-                            new ExternalDirtinessChecker(tmpExternalFilesHelper, fileTypesToCheck),
-                            new MissingDiffDirtinessChecker(diffPackageRootsUnderWhichToCheck)))));
+        batchDirtyResult = fsvc.getDirtyKeys(memoizingEvaluator.getValues(), new UnionDirtinessChecker(dirtynessCheckers));
```
janakdr, line 551: I think it's probably worth taking an ImmutableList#copyOf this dirtynessCheckers list, since I think the iterable gets traversed for every node in the graph.

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +377,40 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+    generateProject();
+    buildExpectRepositoryRuleCalled(false, watchFs);
+    checkProjectFiles();
+
+    Path nodeModules = context().getWorkDir().resolve("node_modules");
+    assertThat(nodeModules.toFile().isDirectory()).isTrue();
+    PathUtils.deleteTree(nodeModules);
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // We don't check the external file disappeared so the build is broken
```
janakdr, line 391: nit: s/check the/check that the/

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +377,40 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+      buildSucceeded(result);
+    } else {
+      ProcessResult result = bazel.shouldFail().build("//...");
+      assertThat(findPattern(result, "Not found package.json")).isTrue();
+    }
+
+    // it doesn't make the file on disk
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // But in a perfect world a normal build would fix it. TODO: uncomment this
```
janakdr, line 409: I don't think we need to keep a TODO here, or commented out code. Just a comment explaining that we don't check for the missing file until something else changes.

src/test/shell/bazel/BUILD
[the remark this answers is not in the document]
janakdr: Can you use the same preamble as archive_contents_test.sh uses? I think everything up to line 58 should probably be copied from it.

janakdr (commented):
> It sounds like this flag is only intended for external _repositories_, not external files in general. In that case, I have a few suggestions. First, change the name to experimental_check_external_repository_files. Second, move the flag to RepositoryOptions. You can still access it in SequencedSkyframeExecutor via the `options` variable.

Please address this.

> For (2), I think that you _don't_ want to affect the checks for external files that are not in repositories. That specifically means that ExternalFilesKnowledge#tooManyNonOutputExternalFilesSeen and ExternalFilesKnowledge#tooManyNonOutputExternalFilesSeen#nonOutputExternalFilesSeen should not be connected to your flag at all. You can enforce that in a test by just creating a symlink to a random other file on disk, modifying the file, and verifying that we rebuild even with all checks disabled.

Please address this too.

> For (3), I suspect that the issue is that on line 598 (in your most recent commit), we set external files knowledge for the graph based on the "scan" that just happened. That setting should probably be conditional, or be a union, if we didn't actually check the whole graph above.

I'm going to try to fix this, actually, which should make your PR work. It's too ugly a bug to leave lying around.

janakdr:
As of 8ef29dd57a41efe18b858f725d2143338623953c we now have better test coverage here. You can fix the TODO at https://github.com/bazelbuild/bazel/blob/8ef29dd57a41efe18b858f725d2143338623953c/src/test/java/com/google/devtools/build/lib/skyframe/LocalDiffAwarenessIntegrationTest.java#L185 by adding your flag there, and also add your flag to [ignoreOutputFilesThenCheckAgainDoesCheck](https://github.com/bazelbuild/bazel/blob/8ef29dd57a41efe18b858f725d2143338623953c/src/test/java/com/google/devtools/build/lib/skyframe/LocalDiffAwarenessIntegrationTest.java#L131) so that it actually tests what it's supposed to. As a result of that commit, you should now get proper detection of modified external repository files on the subsequent build.

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
ptarjan: Oh! I thought I saw `externalFilesKnowledge.tooManyNonOutputExternalFilesSeen` set when I had too many files in external repositories. But if that isn't the case, then I'm happy to decouple.

ptarjan:
Thanks for doing your fix! That greatly helps my end. I rebased and am working through some of your requests. One that is giving me trouble is:

> You can enforce that in a test by just creating a symlink to a random other file on disk, modifying the file, and verifying that we rebuild even with all checks disabled.

If this is the requirement with my flag on, then I don't think I've succeeded. I need to not walk the in-memory cache at all since our monorepo loads up a ton of things and walking it for invalidation currently takes 4 minutes just to do a `bazel run :something_small`. I think the symlink check is done by the `repositoryHelpersHolder.repositoryDirectoryDirtinessChecker` right? If so, I need to not do that either. Is that skippable by using `--watchfs` if the symlink is to a file which was also in the repo root?

src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java
janakdr: No, this should be only for EXTERNAL-type files, which are different from EXTERNAL_REPO-type files.

janakdr (commented):
> > You can enforce that in a test by just creating a symlink to a random other file on disk, modifying the file, and verifying that we rebuild even with all checks disabled.
> 
> If this is the requirement with my flag on, then I don't think I've succeeded. I need to not walk the in-memory cache at all since our monorepo loads up a ton of things and walking it for invalidation currently takes 4 minutes just to do a `bazel run :something_small`. I think the symlink check is done by the `repositoryHelpersHolder.repositoryDirectoryDirtinessChecker` right? If so, I need to not do that either. Is that skippable by using `--watchfs` if the symlink is to a file which was also in the repo root?

Let's make sure we have our terminology clear. My understanding is that your project has a lot of external *repositories*. These are defined in your WORKSPACE file and live under bazel-out/external if they're downloaded, or in a specific local directory in some cases. The "random other file" I'm talking about above is an external file not in a repository, just the target of a symlink, where the symlink itself is in the project. My understanding is that this second type of external file is *not* a concern for your project. Is that correct?

Even if none of the conditions for a full graph scan are met, we will check individual targets of symlinks when the targets are not in the repo: https://github.com/bazelbuild/bazel/blob/43bcbb623e241e6381149c08e2be653433cc9407/src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java#L548-L551
That shouldn't appreciably slow the build down, since you should have 0 of these files in your repo.

RepositoryDirectoryDirtinessChecker is unrelated to this, and I think it's fine for your flag to disable it.

ptarjan:
> Let's make sure we have our terminology clear.
Thank you! I'm pretty new to working in `bazel`'s guts.

> My understanding is that your project has a lot of external repositories. These are defined in your WORKSPACE file and live under bazel-out/external if they're downloaded, or in a specific local directory in some cases. 
Correct

> The "random other file" I'm talking about above is an external file not in a repository, just the target of a symlink, where the symlink itself is in the project. My understanding is that this second type of external file is not a concern for your project. Is that correct?
Sort of. The main problem is just the act of calling `fsvc.getDirtyKeys` costs 4 minutes, regardless of what checkers are present. Even if I pass in an empty array, just walking the cache itself walks the in-memory cache and since we have so many external repositories (basically every `npm` and `pip` package out there) that takes forever.

So in CI I want to avoid entering that function call at all. I thought I wanted to set some flags that says to `bazel` that no files will change on the filesystem without its knowledge since users don't have write access there. Should I also add another flag then that says there aren't any symlinks outside of the repo? Or are there other ways to manifest an `EXTERNAL` file?

I pushed a copy of something that would work for me. It doesn't call the `repositoryHelpersHolder.repositoryDirectoryDirtinessChecker()` when my flag is on, but the test still seems to find the symlink change correctly (as shown in `test_symlink_still_checked` https://github.com/ptarjan/bazel/commit/d63e66b994525a5250005e1a2be75c46477b6581#diff-24dab69c0b04fba6de1375566b6bb73f6727def5d76b4aac4aafcd8e09134060R195)

janakdr:
> > The "random other file" I'm talking about above is an external file not in a repository, just the target of a symlink, where the symlink itself is in the project. My understanding is that this second type of external file is not a concern for your project. Is that correct?
> 
> Sort of. The main problem is just the act of calling `fsvc.getDirtyKeys` costs 4 minutes, regardless of what checkers are present. Even if I pass in an empty array, just walking the cache itself walks the in-memory cache and since we have so many external repositories (basically every `npm` and `pip` package out there) that takes forever.
> 
> So in CI I want to avoid entering that function call at all. I thought I wanted to set some flags that says to `bazel` that no files will change on the filesystem without its knowledge since users don't have write access there. Should I also add another flag then that says there aren't any symlinks outside of the repo? Or are there other ways to manifest an `EXTERNAL` file?

The graph scan that causes you issues is guarded by https://github.com/bazelbuild/bazel/blob/416ae3754f64cd86744c913e026eeb501555d6b0/src/main/java/com/google/devtools/build/lib/skyframe/SequencedSkyframeExecutor.java#L475-L479

Note that *any* external repo file encountered will pass the guard (which is why your flag is needed), but only `externalFilesKnowledge.tooManyNonOutputExternalFilesSeen` will pass the guard, not just *any* external non-repo file. The "too many" there is [2500](https://github.com/bazelbuild/bazel/blob/04dff44f5307aa4b0c74e09924f0a659d7debd02/src/main/java/com/google/devtools/build/lib/skyframe/ExternalFilesHelper.java#L46).

So, as long as your filesystem is --watchfs-compatible, and you aren't linking to thousands of other files that are neither in your workspace nor in external repositories, you can guarantee that none of the conditions in the guard are satisfied.

> I pushed a copy of something that would work for me. It doesn't call the `repositoryHelpersHolder.repositoryDirectoryDirtinessChecker()` when my flag is on, but the test still seems to find the symlink change correctly (as shown in `test_symlink_still_checked` [ptarjan@d63e66b#diff-24dab69c0b04fba6de1375566b6bb73f6727def5d76b4aac4aafcd8e09134060R195](https://github.com/ptarjan/bazel/commit/d63e66b994525a5250005e1a2be75c46477b6581#diff-24dab69c0b04fba6de1375566b6bb73f6727def5d76b4aac4aafcd8e09134060R195))

Is the PR ready for me to review?

ptarjan:
> First, change the name to experimental_check_external_repository_files. Second, move the flag to RepositoryOptions. You can still access it in SequencedSkyframeExecutor via the options variable.

If I try to reference a `RepositoryOptions` from `SequencedSkyframeExecutor` then I need to add `"//src/main/java/com/google/devtools/build/lib/bazel/repository` to the `src/main/java/com/google/devtools/build/lib/skyframe/BUILD` but then I get:
```
ERROR: /Users/paul.tarjan/robinhood/opensource/bazel/src/main/java/com/google/devtools/build/lib/BUILD:219:13: in java_library rule //src/main/java/com/google/devtools/build/lib:runtime: cycle in dependency graph:
    //src/test/java/com/google/devtools/build/lib/blackbox/tests:ManagedDirectoriesBlackBoxTest
    //src/test/java/com/google/devtools/build/lib/blackbox/tests:common_deps
    //src/test/java/com/google/devtools/build/lib/blackbox/junit:test_base
    //src/test/java/com/google/devtools/build/lib/blackbox/bazel:common_tools
    //src/test/java/com/google/devtools/build/lib/blackbox/bazel:default
    //src:bazel
    //src:bazel-bin
    //src:package-zip_jdk_minimal
    //src:install_base_key_jdk_minimal
    //src:install_base_key-file_jdk_minimal
    //src/main/java/com/google/devtools/build/lib/bazel:BazelServer_deploy.jar
    //src/main/java/com/google/devtools/build/lib/bazel:BazelServer
    //src/main/java/com/google/devtools/build/lib/bazel:main
.-> //src/main/java/com/google/devtools/build/lib:runtime
|   //src/main/java/com/google/devtools/build/lib/analysis:build_view
|   //src/main/java/com/google/devtools/build/lib/skyframe:skyframe_cluster
|   //src/main/java/com/google/devtools/build/lib/bazel/repository:repository
`-- //src/main/java/com/google/devtools/build/lib:runtime
```

Any suggestions how to resolve that?

ptarjan:
> you can guarantee that none of the conditions in the guard are satisfied.

I guess I'm having trouble figuring out how to make the `repositoryHelpersHolder != null` part of the guard not be true. It seems to always be present so if I don't add my flag to that one guard, I still do the full cache scan.

ptarjan:
> Is the PR ready for me to review?

Not quite. I still have to work on the rename you asked for (questioned above) and fixing the test you linked to.

janakdr:
> > you can guarantee that none of the conditions in the guard are satisfied.
> 
> I guess I'm having trouble figuring out how to make the `repositoryHelpersHolder != null` part of the guard not be true. It seems to always be present so if I don't add my flag to that one guard, I still do the full cache scan.

Your flag should definitely apply to repositoryHelpersHolder: that's why I created that object, so it was clear that it was repository-related, and so your flag should neutralize it.

> .-> //src/main/java/com/google/devtools/build/lib:runtime
> |   //src/main/java/com/google/devtools/build/lib/analysis:build_view
> |   //src/main/java/com/google/devtools/build/lib/skyframe:skyframe_cluster
> |   //src/main/java/com/google/devtools/build/lib/bazel/repository:repository
> `-- //src/main/java/com/google/devtools/build/lib:runtime
> ```
> 
> Any suggestions how to resolve that?

Sorry about that. I would just break RepositoryOptions out of the full `repository:repository` target and put it in its own `repository:repository_options` target: it doesn't depend on anything else in the repository bundle.

ptarjan:
> Your flag should definitely apply to repositoryHelpersHolder: that's why I created that object, so it was clear that it was repository-related, and so your flag should neutralize it.

Ok! Perfect. I guess that didn't jump out at me at first since I think I'm anchoring on the word external more than repository. Should I delete the word external from my flag?

> I would just break RepositoryOptions out

worked great!


It should be ready for review now (assuming the tests all pass).

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +377,39 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+    generateProject();
+    buildExpectRepositoryRuleCalled(false, watchFs);
+    checkProjectFiles();
+
+    Path nodeModules = context().getWorkDir().resolve("node_modules");
+    assertThat(nodeModules.toFile().isDirectory()).isTrue();
+    PathUtils.deleteTree(nodeModules);
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // We don't check that the external file disappeared so the build is broken
```
janakdr, line 391: I don't understand this statement. Why is the build broken, as opposed to spuriously succeeding because we don't actually execute any actions that depend on the missing files? What causes us to try to rebuild anything at all? I would expect that the missing files would go unnoticed until an in-repo file was modified that caused a downstream action to re-run, with that downstream action consuming a file from the deleted managed directory.

That being said, I'm also fine with your just deleting this test. Managed directories is a super gnarly feature, and I'm satisfied that your flag isn't making anything worse.

src/test/shell/bazel/BUILD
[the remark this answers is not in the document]
janakdr: I don't see the common preamble?

janakdr (commented):
> > Your flag should definitely apply to repositoryHelpersHolder: that's why I created that object, so it was clear that it was repository-related, and so your flag should neutralize it.
> 
> Ok! Perfect. I guess that didn't jump out at me at first since I think I'm anchoring on the word external more than repository. Should I delete the word external from my flag?

No, that's fine, they're called "external repositories" as opposed to the "mono repo(sitory)" that the main code lives in. Terminology isn't great, but that's where we are.

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
ptarjan: I was writing this comment comparing to the above test `testNodeModulesDeleted` which notices that we deleted this file and the rebuild fixes it. I updated the comment to reference it. Do you think it is still useful to keep this around like this?

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +377,39 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+      buildSucceeded(result);
+    } else {
+      ProcessResult result = bazel.shouldFail().build("//...");
+      assertThat(findPattern(result, "Not found package.json")).isTrue();
+    }
+
+    // it doesn't make the file on disk
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // In a perfect world we would be able to fix the build by rebuildling here without the flags, but we don't
```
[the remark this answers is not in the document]
ptarjan: Sadly yes, it still fails if I put 
```
      buildExpectRepositoryRuleCalled(false, watchFs);
      checkProjectFiles();
```

ptarjan:
Windows failure here: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/5c2d7165-f4e8-4fa4-86d7-92a37725897c/src%5Ctest%5Cshell%5Cbazel%5Ccheck_external_files_test%5Ctest.log

ptarjan:
OSX failure that I worry might be non-deterministic. I'll try a few more times and skip the test on OSX if that's the case.: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/eb5b7280-d923-4f91-9b84-3858e7e1c859/src/test/java/com/google/devtools/build/lib/blackbox/tests/ManagedDirectoriesBlackBoxTest/test.log

ptarjan:
And a failure the other way. I'll just remove it for OSX. https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/4dc49353-950f-4e23-b8d2-c53dccca487e/src/test/java/com/google/devtools/build/lib/blackbox/tests/ManagedDirectoriesBlackBoxTest/test.log

ptarjan:
@janakdr this is ready for review

src/test/java/com/google/devtools/build/lib/blackbox/tests/manageddirs/ManagedDirectoriesBlackBoxTest.java
```diff
@@ -392,6 +377,38 @@ public void testWorkspaceSymlinkThrowsWithManagedDirectories() throws Exception
...
+    }
+
+    ProcessResult result = bazel.shouldFail().build("//...");
+    assertThat(findPattern(result, "Not found package.json")).isTrue();
+
+    // it doesn't make the file on disk
+    assertThat(nodeModules.toFile().isDirectory()).isFalse();
+
+    // In a perfect world we would be able to fix the build by rebuilding here without the flags, but we don't
+    // invalidate the cache correctly so the server would have to shutdown
```
janakdr, line 409: nit-nit: s/shutdown/shut down

src/test/shell/bazel/check_external_files_test.sh
```diff
@@ -0,0 +1,270 @@
...
+  rm -rf remote
+
+  mkdir main
+  cd main
+  cat >> "$(create_workspace_with_default_repos WORKSPACE)" <<EOF
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+  name="remote",
+  strip_prefix="remote",
+  urls=["file://${WRKDIR}/remote.tar"],
```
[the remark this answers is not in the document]
meteorcloudy: Yes, I think this is what you need:
https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/external_patching_test.sh;l=86-92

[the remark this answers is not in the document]
ptarjan: Thanks! I'll add that in. Is there a similar fix for `ln -s`?  i.e. fixing this:
```
ERROR: Could not create symlink to repository "local_rep" (absolute path: "/c/b/ef64ujiv/execroot/io_bazel/_tmp/98581429905a6f3c8d2b19e2a074195b/testP9EDZf/local_rep"): Cannot create junction (name=C:\b\ef64ujiv\execroot\io_bazel\_tmp\98581429905a6f3c8d2b19e2a074195b\root\fsdvwpty\external\local_rep, target=\c\b\ef64ujiv\execroot\io_bazel\_tmp\98581429905a6f3c8d2b19e2a074195b\testP9EDZf\local_rep): ERROR: src/main/native/windows/file-jni.cc(99): nativeCreateJunction(\\?\C:\b\ef64ujiv\execroot\io_bazel\_tmp\98581429905a6f3c8d2b19e2a074195b\root\fsdvwpty\external\local_rep, \\?\\c\b\ef64ujiv\execroot\io_bazel\_tmp\98581429905a6f3c8d2b19e2a074195b\testP9EDZf\local_rep): ERROR: src/main/native/windows/file.cc(182): CreateJunction(\\?\\c\b\ef64ujiv\execroot\io_bazel\_tmp\98581429905a6f3c8d2b19e2a074195b\testP9EDZf\local_rep): expected an absolute Windows path for junction_target
```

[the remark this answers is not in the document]
meteorcloudy: It seems an unix path (`/c/b/ef64ujiv/execroot/io_bazel/_tmp/98581429905a6f3c8d2b19e2a074195b/testP9EDZf/local_rep`) was somehow passed to the Windows API for creating a junction (directory symlink on Windows). I'm not sure how that could happen with the current code, can you please try to enable the test on Windows so we could see the failure on CI?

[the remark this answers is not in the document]
ptarjan: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/10074#d6eba862-0166-4d29-a83f-6c19acfb7fcd is a failed windows build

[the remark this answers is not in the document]
ptarjan: 🤦 . https://github.com/bazelbuild/bazel/pull/14404/commits/a192661c44d702cf96f7736ac36b9911660ca5cc. I guess windows was trying to simlink something but the other OSes are fine with that?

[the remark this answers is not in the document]
ptarjan: Nevermind, that was a red herring. Now it fails twice with the symlink problem.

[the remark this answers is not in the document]
ptarjan: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/d6eba862-0166-4d29-a83f-6c19acfb7fcd/src%5Ctest%5Cshell%5Cbazel%5Ccheck_external_files_test%5Ctest.log is the failure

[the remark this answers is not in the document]
ptarjan: I pushed it with disabling windows to unblock this diff

janakdr (approved):
Thanks for sticking through this extremely long review! Apologies for how much time it took me to get up to speed on the issues.

ptarjan:
> Thanks for sticking through this extremely long review! Apologies for how much time it took me to get up to speed on the issues.

Amazing! Thanks so much for the thorough review and for sticking with it too. Let me know if there is any follow ups here and hopefully future contributions from me won't be as difficult.

ptarjan:
Do you know what release this code will be targeting? I'd like to use it internally without being on a fork for too long if possible.

brentleyjones:
There will be a rolling 6.0 release that includes it in 1-2 weeks.

ptarjan:
Ok, thank you. Is there a targeted release date for that?

brentleyjones:
One or two are cut per week internally at Google, and then they go public after internal testing.

ptarjan:
Thank you. I do see a release pushed but when I try to put it in my `.bazelversion` I get this on linux:
```
$ bazel version
2022/01/20 23:28:19 could not resolve the version '6.0.0-pre.20220112.2' to an actual version number: Invalid version '6.0.0-pre.20220112.2'
```

(it does seem to work on mac). Am I doing it as you would expect? And is this a `bazel` thing or a `bazelisk` thing?

brentleyjones:
It's a bazelisk thing, and you are doing it the right way. I would update bazelisk on linux, since rolling releases are a new(er) feature for it.

jheaff1:
Looking forward to this being available in Bazel 6.0 as I have a fairly small repo which builds Qt and LLVM (both massive repos) from source. As Qt and LLVM are so large, the RAM usage of bazel is quite significant (>1GB). Will this `experimental_check_external_repository_files` flag reduce RAM usage? If not, is there a flag that could help with this?

PR #14407, opened 2021-12-11
Remove dataproc

aiuto:
Remove an unneeded google API.
It was the only mention of log4j in the source tree. Since the tests pass without it, that proves it was just a comment and not a use.


aiuto:
I'm not sure we actually want to do this, because it becomes an update headache, but it does prove that we don't use log4j.

PR #14416, opened 2021-12-13
Remote: Don't blocking-get when acquiring gRPC connections.

### LINKED ISSUE #14363: bazel 5.0.0rc2 hangs when both --bes_backend and --build_event_binary_file specified

`--bes_backend specified` in our .bazelrc
`--build_event_binary_file` added by intellij plugin

when specifying both bazel sometimes hang.

```
bazel build //... --bes_backend=grpc://localhost:6000 --build_event_binary_file=build.log '--override_repository=intellij_aspect=/Users/darl/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/212.5457.46/IntelliJ IDEA.app.plugins/ijwb/aspect' --output_groups=intellij-resolve-java-direct-deps,intellij-info-generic,intellij-info-java-direct-deps --aspects=@intellij_aspect//:intellij_info_bundled.bzl%intellij_info_aspect 
```

MacOS, Linux

release 5.0.0rc2

Nothing similar

Thread dump of bazel when it hangs:
https://gist.github.com/darl/94141514150e09b2460030f283cc9f21

coeuvre:
With recent change to limit the max number of gRPC connections by default, acquiring a connection could suspend a thread if there is no available connection.

gRPC calls are scheduled to a dedicated background thread pool. Workers in the thread pool are responsible to acquire the connection before starting the RPC call.

There could be a race condition that a worker thread handles some gRPC calls and then switches to a new call which will acquire new connections. If the number of connections reaches the max, the worker thread is suspended and doesn't have a chance to switch to previous calls. The connections held by previous calls are, hence, never released.

This PR changes to not use blocking get when acquiring gRPC connections.

Fixes #14363.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/remote/BUILD
@@ -138,9 +138,11 @@ java_library(
     ],
     deps = [
         "//src/main/java/com/google/devtools/build/lib/remote/grpc",
+        "//src/main/java/com/google/devtools/build/lib/remote/util",
         "//third_party:guava",
         "//third_party:jsr305",
         "//third_party:netty",
+        "//third_party:rxjava3",
         "//third_party/grpc:grpc-jar",
     ],
 )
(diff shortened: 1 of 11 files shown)


brentleyjones:
Since 8d5973d29d60c0c615838c534ee27f93377cf5af is in the 5.0 RC, can we get this cherry-picked into the release as well? @Wyverald

brentleyjones:
FYI, 5.0 cherry-pick is here: https://github.com/bazelbuild/bazel/pull/14421

Issue #14249, opened 2021-11-09
Bazel 5.0rc1 issue with linking Java 17 records with explicit constructors from a different package


eikemeier:
### Description of the problem / feature request:

Bazel 5.0rc1 has an issue with linking Java 17 records with explicit constructors from a different package.

Demonstration in [eikemeier/bazel_jdk17](https://github.com/eikemeier/bazel_jdk17).

### Feature requests: what underlying problem are you trying to solve with this feature?

Compile a Java 17 program with bazel.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Check out https://github.com/eikemeier/bazel_jdk17, run `bazelisk run //src/main/java/com/example`.

### What operating system are you running Bazel on?

macOS 10.15

### What's the output of `bazel info release`?

release 5.0.0rc1

### What's the output of `git remote get-url origin ; git rev-parse main ; git rev-parse HEAD` ?

ssh://git@github.com/eikemeier/bazel_jdk17.git
2c85df80a77291600764d3f193437dec07a41c51
2c85df80a77291600764d3f193437dec07a41c51

###  Have you found anything relevant by searching the web?

Not really, if you don't count the general Java 17-discussion.

### Any other information, logs, or outputs that you want to share?

See the the [README](https://github.com/eikemeier/bazel_jdk17#readme).

cushon:
```
ERROR: [...]/bazel_jdk17/src/main/java/com/example/hello/BUILD:3:13: Compiling Java headers src/main/java/com/example/hello/libhello-hjar.jar (1 source file) failed: (Exit 1): java failed: error executing command external/remotejdk17_macos/bin/java [...]

src/main/java/com/example/hello/Hello.java:6: error: unexpected token: {
  public Hello {
               ^
```

```
$ cat src/main/java/com/example/hello/Hello.java
package com.example.hello;

import java.util.Objects;

public record Hello(String one) {
  public Hello {
    Objects.requireNonNull(one);
  }
}
```

eikemeier:
@cushon: When you move the `Hello` record in the same package as the `Main` class the error vanishes. So _just_ the record isn't enough.

cushon:
That's because header compilation actions only run on Java targets that are dependencies of other targets. The header jar is used as a compile-time optimization when compiling downstream targets, and the action isn't run if record is being built directly.

eikemeier:
[...]

PR #14423, opened 2021-12-14
Update java_tools v11.6

eikemeier:
Closes #14296.

PiperOrigin-RevId: 410870809

### DIFF
--- a/distdir_deps.bzl
+++ b/distdir_deps.bzl
@@ -281,11 +281,11 @@ DIST_DEPS = {
             "remote_java_tools_test",
             "remote_java_tools_for_testing",
         ],
-        "archive": "java_tools-v11.5.zip",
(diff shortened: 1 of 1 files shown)


eikemeier:
Pinging @Wyverald

eikemeier:
Also, to explain why this is good: If fixes #14249, so you can use [compact constructors](https://docs.oracle.com/en/java/javase/14/language/records.html).

Wyverald:
I'm reluctant to cherrypick this commit at this stage, since it's unclear how much it changes and we risk delaying the release further. Plus you can get the benefits by using the latest java_tools in your WORKSPACE anyhow.

Wyverald:
Given that there are other release blockers for 5.0 now (and rc3 is definitely not the last RC), I'm happy the cherrypick this.

PR #14424, opened 2021-12-14
Make AndroidCompiledResourceMergingAction worker compatible

Bencodes:
Calling `System#exit` from within a worker compatible action will shut down unexpectedly.


meisterT:
cc @larsrc-google

larsrc-google (approved):

PR #14425, opened 2021-12-14
Avoid leaking Aapt2ResourcePackagingAction args between requests

Bencodes:
Keeping the `Aapt2ResourcePackagingAction` options in a static field allows these options to leak between requests in a multiplex world.


meisterT:
cc @larsrc-google

Issue #6904, opened 2018-12-12
Android manifest merger error messages are unnecessarily dropped


jin:
Repro: compile an Android project with the following invalid manifest:

```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.example.bazel"
    android:versionCode="1"
    android:versionName="1.0" />

  <uses-sdk
      android:minSdkVersion="21"
      android:targetSdkVersion="26" />

</manifest>
```

Note the incorrect `/>` in the opening `<manifest>` element.

Bazel crashes with: 

```
INFO: Found 1 target...
ERROR: /usr/local/google/home/jingwen/code/examples/android/tutorial-test/src/main/BUILD:1:1: Merging manifest for //src/main:greeter_app failed (Exit 1) ResourceProcessorBusyBox failed: error executing command bazel-out/host/bin/external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox --tool MERGE_MANIFEST -- --manifest src/main/AndroidManifest.xml ... (remaining 8 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Target //src/main:greeter_app failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 9.480s, Critical Path: 3.73s
INFO: 9 processes: 8 linux-sandbox, 1 worker.
FAILED: Build did NOT complete successfully
```

It is expected to crash with:

```
ERROR: /usr/local/google/home/jingwen/code/examples/android/tutorial-test/src/main/BUILD:1:1: Merging manifest for //src/main:greeter_app failed (Exit 1) ResourceProcessorBusyBox failed: error executing command bazel-out/host/bin/external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox --tool MERGE_MANIFEST -- --manifest src/main/AndroidManifest.xml ... (remaining 8 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Dec 12, 2018 1:23:25 PM com.google.devtools.build.android.ManifestMergerAction main
SEVERE: Encountered error while attempting to merge manifests
[...]

PR #14427, opened 2021-12-14
Make ManifestMergerAction worker compatible

Bencodes:
Calling `System#exit` kills the worker during the build. Passing the exception up to the worker should be enough for it to end up in the worker or local execution output.

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
+++ b/src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
@@ -245,10 +245,10 @@ public class ManifestMergerAction {
         Files.copy(manifest, options.manifestOutput, StandardCopyOption.REPLACE_EXISTING);
       }
     } catch (AndroidManifestProcessor.ManifestProcessingException e) {
-      // We special case ManifestProcessingExceptions here to indicate that this is
-      // caused by a build error, not an Bazel-internal error.
-      logger.log(SEVERE, "Error during merging manifests", e);
-      System.exit(1); // Don't duplicate the error to the user or bubble up the exception.
+      // ManifestProcessingExceptions represent build errors that should be delivered directly to
+      // ResourceProcessorBusyBox where the exception can be delivered with a non-zero status code
+      // to the worker/process
+      throw e;
     } catch (Exception e) {
       logger.log(SEVERE, "Error during merging manifests", e);
       throw e; // This is a proper internal exception, so we bubble it up.
--- a/src/test/java/com/google/devtools/build/android/ManifestMergerActionTest.java
+++ b/src/test/java/com/google/devtools/build/android/ManifestMergerActionTest.java
@@ -15,6 +15,7 @@ package com.google.devtools.build.android;
 
 import static com.google.common.truth.Truth.assertThat;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertThrows;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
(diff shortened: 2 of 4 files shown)


src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
```diff
@@ -231,11 +231,6 @@ public static void main(String[] args) throws Exception {
         // to the expected location of the output.
         Files.copy(manifest, options.manifestOutput, StandardCopyOption.REPLACE_EXISTING);
       }
-    } catch (AndroidManifestProcessor.ManifestProcessingException e) {
```
Bencodes, line 234: The root `Exception` can catch everything, minus the `System.exit`.

meisterT:
cc @larsrc-google

src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
larsrc-google: This was explicitly added to fix https://github.com/bazelbuild/bazel/issues/6904. We should not unfix that.

Bencodes: Thoughts on raising this specific logic up into `ResourceProcessorBusyBox` where we can log the exception and return the proper exit code inside `processRequest`?

Looking at some of the other options, they don't really

larsrc-google: @ahumesky

ahumesky: Well, this is a bit strange. There are actually two exception classes with this name:
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/tools/android/java/com/google/devtools/build/android/ManifestProcessingException.java
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/tools/android/java/com/google/devtools/build/android/AndroidManifestProcessor.java#L60

And they're even both used in the same class:
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/tools/android/java/com/google/devtools/build/android/AndroidManifest.java#L85
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/tools/android/java/com/google/devtools/build/android/AndroidManifest.java#L145

We should converge onto one to make this less confusing....

Looks like the top-level class is also being used to capture two different kinds of exceptions:
https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/tools/android/java/com/google/devtools/build/android/AndroidManifest.java#L84-L86

IOException is one thing, but XMLStreamException might be from badly formed xml...

Either way, if we just throw here, this exception will get caught by the `catch (Exception e)` in ResourceProcessorBusyBox, and do the same basic logging. So probably don't need to make ResourceProcessorBusyBox aware of lower level exceptions.

Bencodes: > Either way, if we just throw here, this exception will get caught by the catch (Exception e) in ResourceProcessorBusyBox, and do the same basic logging. So probably don't need to make ResourceProcessorBusyBox aware of lower level exceptions.

The generic `catch (Exception e) {` in `ResourceProcessorBusyBox` will re-throw the exception after logging the error and bubble that up as an exception coming from `main`, while `catch (AndroidManifestProcessor.ManifestProcessingException e) {`  gracefully handles the exception with an exit code of `1` being returned directly via the work response (Prior to this PR is was being returned as a `System#exit` with 1). Is this slight behavioral change fine to merge?

Bencodes: Probably isn't a big deal given that this PR wraps all unhandled exceptions with a proper exit code of 1 in order to confirm to `WorkRequestHandler`.

https://github.com/bazelbuild/bazel/pull/14428/files#diff-6bbb0aaf4cf2d425734021815617e2b6203223abe7dc76ca8030a90e9f043873R213

Bencodes: @larsrc-google is this in a mergable state so that https://github.com/bazelbuild/bazel/pull/14428/ is unblocked?

nkoroste:
Any updates on this?

src/tools/android/java/com/google/devtools/build/android/ManifestMergerAction.java
ted-xie: Sorry, I'm catching up on this thread right now -- I'll probably add in a comment in the ManifestProcessingException case handler to explain why its handler is almost the same as the generic case, minus the logging statement. I think it should be fine -- either we have essentially two debug log lines of the same error message, or we have a slightly funky-looking try/catch block. Cutting down on noise in the console seems preferable in this case.

ted-xie:
@nkoroste We've merged this upstream!

nkoroste:
Can we add it to 5.2 while we're at it? https://github.com/bazelbuild/bazel/issues/15210

ckolli5:
@bazel-io fork 5.3.0

ckolli5:
Hello @ted-xie, I cherry-picked these changes and raised a [PR](https://github.com/bazelbuild/bazel/pull/15699) against release-5.3.0 branch. Presubmit checks are failing in that PR. Could you please cherry picked with appropriate commits and raise a PR against [release-5.3.0](https://github.com/bazelbuild/bazel/tree/release-5.3.0)? Thanks!

PR #14428, opened 2021-12-14
Support multiplex workers in ResourceProcessorBusyBox

Bencodes:


meisterT:
cc @larsrc-google

src/main/java/com/google/devtools/build/lib/rules/android/BusyBoxActionBuilder.java
```diff
@@ -344,10 +344,12 @@ public void buildAndRegister(String message, String mnemonic) {
...
     if (dataContext.isPersistentBusyboxToolsEnabled()) {
       commandLine.add("--logWarnings=false");
-      spawnActionBuilder
-          .addCommandLine(commandLine.build(), WORKERS_FORCED_PARAM_FILE_INFO);
-
+      spawnActionBuilder.addCommandLine(commandLine.build(), WORKERS_FORCED_PARAM_FILE_INFO);
       executionInfo.putAll(ExecutionRequirements.WORKER_MODE_ENABLED);
+
+      if (dataContext.isPersistentMultiplexBusyboxTools()) {
+        executionInfo.putAll(ExecutionRequirements.WORKER_MULTIPLEX_MODE_ENABLED);
```
larsrc-google, line 351: So this flag would override `--noexperimental_worker_multiplex`?

Bencodes: Passing `--noexperimental_worker_multiplex` would disable multiplex workers entirely while `isPersistentMultiplexBusyboxTools` just instructs `BusyBoxActionBuilder` to register it's actions with with `supports-multiplex-worker` specified as part of it's execution info.

larsrc-google: D'uh. You're right.

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,36 +184,43 @@ private static int runPersistentWorker() throws Exception {
...
+                        } catch (Exception e) {
+                          e.printStackTrace(pw);
+                          return 1;
+                        } finally {
+                          try {
+                            buf.flush();
+                            String s = buf.toString().trim();
+                            buf.reset();
+                            if (!s.isEmpty()) {
+                              pw.print(s);
```
larsrc-google, line 208: This sends everything sent to stdout/err into the `output` field of the response. However, since System.out/err are global, output from all threads gets interleaved here. These should be sent to `realStdErr`, any actual error messages from processing this request should go to `pw`. See discussion in https://github.com/bazelbuild/bazel/pull/14201

Bencodes: Merging https://github.com/bazelbuild/bazel/pull/14201 first would be the preferred solution here so that we aren't having to duplicate any of the system stream remapping logic between workers.

Bencodes: Pushed the changes to swap `System.err` in for `System.out`.

Bencodes: Also worth noting that swapping `System.err` in for `System.out` technically changes where the outputs are delivered for this rule, where the prior implementation would write the captured outputs directly to the work response output vs. the new implementation dumps them to `System.err` and only writes the contents of the `PrintWriter` provided by `WorkRequestHandler` to the work response output.

larsrc-google: Then you need to be sure that that's the desired behaviour. Things the user should see should go into the work response output, things that should go into the log should go to `System.out`.

Bencodes: I'm going to roll back to the previous behavior to preserve the existing logs/error delivery that exists today.

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,43 +182,51 @@ private static int runPersistentWorker() throws Exception {
...
+      exitCode = processRequest(args);
+    } catch (Exception e) {
+      e.printStackTrace(pw);
+      exitCode = 1;
+    } finally {
+      // Write the captured buffer to the work response
+      String captured = buf.toString().trim();
+      buf.reset();
+      if (!captured.isEmpty()) {
+        pw.print(captured);
```
Bencodes, line 223: This preserves the previous logging behavior seen here https://github.com/bazelbuild/bazel/pull/14428/files#diff-6bbb0aaf4cf2d425734021815617e2b6203223abe7dc76ca8030a90e9f043873L204

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,43 +182,51 @@ private static int runPersistentWorker() throws Exception {
...
       System.setErr(realStdErr);
     }
     return 0;
   }
 
+  private static int processRequest(List<String> args, PrintWriter pw, ByteArrayOutputStream buf) {
+    int exitCode;
+    try {
+      // Process the actual request and grab the exit code
+      exitCode = processRequest(args);
```
Bencodes, line 214: In the prior implementation, errors thrown by `processRequest` would crash the worker with an unexpected exception. This should mirror that behavior by delivering the exception as part of the work response with a non-zero exit code.

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,43 +182,55 @@ private static int runPersistentWorker() throws Exception {
...
+    } catch (Exception e) {
+      e.printStackTrace(pw);
+      exitCode = 1;
+    } finally {
+      // Write the captured buffer to the work response
+      String captured = buf.toString().trim();
```
larsrc-google, line 224: You have a race condition here: `buf` is shared among all threads, so if something gets written between this line and the next, it gets lost. Synchronizing on `buf` should be OK, though.

Bencodes: Updated but synchronizing on a static object to avoid having to synchronize on function arguments.

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,43 +188,58 @@ private static int runPersistentWorker() throws Exception {
...
+    } catch (Exception e) {
+      e.printStackTrace(pw);
+      exitCode = 1;
+    } finally {
+      // Write the captured buffer to the work response. We synchronize to avoid race conditions
+      // while reading from and calling reset on the shared ByteArrayOutputStream.
+      synchronized (LOCK) {
+        String captured = buf.toString().trim();
+        buf.reset();
+        if (!captured.isEmpty()) {
```
larsrc-google, line 234: The `if` is not really necessary. Printing nothing should not be an issue.

larsrc-google (approved):

src/tools/android/java/com/google/devtools/build/android/ResourceProcessorBusyBox.java
```diff
@@ -183,43 +188,58 @@ private static int runPersistentWorker() throws Exception {
...
+    try {
+      // Process the actual request and grab the exit code
+      exitCode = processRequest(args);
+    } catch (Exception e) {
+      e.printStackTrace(pw);
+      exitCode = 1;
+    } finally {
+      // Write the captured buffer to the work response. We synchronize to avoid race conditions
+      // while reading from and calling reset on the shared ByteArrayOutputStream.
+      synchronized (LOCK) {
```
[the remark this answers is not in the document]
Bencodes: Updated

ted-xie:
Hi @Bencodes @nkoroste!

We've merged this PR upstream; thanks so much for your contribution! In order for this feature to actually work, we'll have to update some remote repository pointers. I'll tag this PR in the description for that change, and you'll be notified when it lands.

Bencodes:
Awesome, thanks for taking this one across the finish line @ted-xie!

Issue #14285, opened 2021-11-16
Android manifest merger unable to merge manifests containing queries


Bencodes:
### Description of the problem / feature request:

The Android manifest merger being used by the Bazel `android_tools` is unable to merge manifests that contain queries that have providers listed. This can be reproduced using the rolling releases and HEAD.

The merge fails with the below exception:

```
Error: [@maven//:androidx_car_app_app] /private/var/tmp/_bazel_ekamp/490385dd022af32a8a24bff076cf882c/bazel-workers/multiplex-worker-1-ManifestMerger/instantandroid/bazel-out/android-arm64-v8a-dbg/bin/external/maven/androidx_car_app_app_processed_manifest/AndroidManifest.xml:25:9-71 Error:
	Missing 'name' key attribute on element provider at [@maven//:androidx_car_app_app] AndroidManifest.xml:25:9-71
Error: [@maven//:androidx_car_app_app] /private/var/tmp/_bazel_ekamp/490385dd022af32a8a24bff076cf882c/bazel-workers/multiplex-worker-1-ManifestMerger/instantandroid/bazel-out/android-arm64-v8a-dbg/bin/external/maven/androidx_car_app_app_processed_manifest/AndroidManifest.xml Error:
	Validation failed, exiting
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This can be reproduced by including the latest version of https://developer.android.com/jetpack/androidx/releases/car-app, which has the following listed in it's manifest:

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="androidx.car.app" >
    
    ...

    <queries>
        <provider
          android:authorities="androidx.car.app.connection"
      />
    ...

```

### What operating system are you running Bazel on?

MacOS Big Sur

###  Have you found anything relevant by searching the web?

Support for queries got [added in API 30](https://developer.android.com/guide/topics/manifest/queries-element) while the `android_tools` are [still using v25.0.0](https://github.com/bazelbuild/bazel/tree/master/third_party/android_common).

### Any other information, logs, or outputs that you want to share?

```
[...]

PR #14429, opened 2021-12-14
Patch manifest merger 25.0.0 with queries support

Bencodes:
Built using the instructions seen [here](https://github.com/bazelbuild/bazel/issues/14285).

https://github.com/bazelbuild/bazel/issues/14285


Bencodes:
I'm hesitant to actually have this one merged because updating the Android tools dependencies seems like a might better long term solution.

Bencodes:
cc: @ahumesky

ahumesky:
We're actually looking to update from 25 to the latest version. It might take a few weeks though given how much has changed since 25

nkoroste:
Any updates on this? curious to know what's the best way to go about working around this issue

PR #14439, opened 2021-12-15
[apple] support tvos_sim_arm64 in toolchain

dflems:


keith (approved):

thii:
This needs to be updated too https://github.com/bazelbuild/bazel/blob/519cb4cc9aa93c38ebb94543f96472769ab30dd9/src/main/java/com/google/devtools/build/lib/rules/apple/AppleConfiguration.java#L233-L245

keith:
ah yea be sure to grep for all `IOS_FORCED_SIMULATOR_CPU_PREFIX` and `ios_sim` in general

brentleyjones:
@Wyverald This would be really nice to get into 5.0 if we are creating another RC (which is looks like we might be?).

dflems:
Updated! Now it just checks for the prefix on all CPUs and chops it off if `removeSimPrefix` is true

brentleyjones (approved):

brentleyjones:
@kaylathar Can we get this reviewed and imported? I think this should be included in 5.0.

cc: @Wyverald.

Wyverald:
Unless there's a strong argument that this is crucial, I'm reluctant to cherrypick any non-trivial changes at this point, for fear of delaying the release even further. We can look at cherrypicking this into a 5.1 release after 5.0 gets out.

dflems:
@Wyverald This would be really useful for us as we ship AppleTV and Watch apps. iOS and macOS aren't the only M1-compatible platforms. For the watch especially, we've been having problems running it under rosetta.

brentleyjones:
This would fall under OS support in my mind, which is a key feature of the LTS releases.

kaylathar:
I have concerns on some portions of this change, in particular tvOS related change. The watchOS related change I will review more closely shortly but I have less concern on. The tvOS change adds more technical debt by adding an additional "_sim_" CPU as we are moving to platforms and will require internal changes in some cases as well. I will evaluate more fully later today though.

dflems:
@kaylathar I can break these up into two PRs if need be. The watch integration is more pressing for us in the short term as the TV app can be built for intel and run under rosetta in the simulator. We're having problems doing that for the watch app.

keith:
It seems to me that this tech debt is pretty localized, and users don't have another option if they need to build for this simulator in the meantime. So can we land that piece as well knowing it's something we'll have to cleanup when the ideal way to implement this lands? I think the worst case scenario here is the ideal solution continues to lag and folks are blocked in the meantime

gregestren:
@kaylathar is there any reference material coming up on how the architectures map differently for platforms? For example, what about `sim` adds more complication than adding another non-`sim` CPU? Or for that matter how is the concept of `sim` is expressed in the new API?

I'm interested both to understand your intentions but also to support the possibility of collaborative improvements. i.e. if you're working with a fixed assumption of what the platforms API can support, possibly at the expense of a clean Apple API, my team might be able to find optimizations to support cleaner Apple APIs. I understand Apple rules particularly struggle with their complex configuration needs.

dflems:
@kaylathar Updated this old PR to fix the conflicts. Still think `tvos_sim_arm64` is a necessary evil until we have proper platform support. Otherwise it's impossible to build for the native tvOS simulator on M1 macs.

kaylathar:
LGTM - I'm fine landing this change given the time it's going to take to get platforms functional.

brentleyjones:
@bazel-io flag

Wyverald:
@bazel-io fork 5.1

PR #14441, opened 2021-12-16
Remove old http cache put deduplication

ulrfa:
Removing HTTP cache specific deduplication logic that is not needed any
more, thanks to the new generic deduplication in RemoteCache.java,
which was introduced by the "Remote: Async upload" set of commits,
completed by commit 581c81a854.


ulrfa:
@coeuvre, what do you think about this? Do you agree that the removed logic is not needed any more, thanks to your "Remote: Async upload" commits?

coeuvre (approved):
LGTM. Thanks!

Note that, there is one difference: `storedBlobs` was used for upload both outputs and action results, but the deduplication in remote cache is only used for upload outputs.

Since action result is uploaded once per action, and actions are deduplicated by skyframe, the difference shouldn't introduce any behaviour changes.

ulrfa:
Thanks @coeuvre!

PR #14442, opened 2021-12-16
Avoid too verbose warnings in terminal when cache issues

ulrfa:
Deduplicate warnings in terminal. This was working in earlier bazel
versions both for read and write, but become broken when write was
moved to RemoteExecutionService.java by the "Remote: Async upload"
set of commits, completed by commit 581c81a854.

Use same phrase "Remote Cache" for both read and write, for
deduplication to work better.

Avoid printing short warnings on multiple lines for reads, as it
already was for writes.


ulrfa:
@coeuvre, what do you think about this? I'm asking you since related to your changes in "Remote: Async upload". Thanks!

PR #14447, opened 2021-12-19
Allow Label instances as keys in select

### LINKED ISSUE #14259: Bzlmod: a project cannot provide a macro that uses select statement
For example in rules_pkg, [//private:private_stamp_detect](https://github.com/bazelbuild/rules_pkg/blob/b83be9ce26ff9708c9b9ebe658a175527cc74ff1/pkg/pkg.bzl#L50) is a config setting in rules_pkg for filling the value of [private_stamp_detect](https://github.com/bazelbuild/rules_pkg/blob/b83be9ce26ff9708c9b9ebe658a175527cc74ff1/pkg/pkg.bzl#L494), which is defined in a macro `pkg_tar`.

In rules_pkg, the final condition label string is calculated as:
```
_stamp_condition = str(Label("//private:private_stamp_detect"))
```
This is because:

- we cannot just use `//private:private_stamp_detect` as the select key, because it will be relative to the main repo that uses the `pkg_tar` macro.
- then rules_pkg has to convert it to a label with a specific repo name via `Label(//private:private_stamp_detect)`, this made sure the package path is relative to rules_pkg.
- But **`select` doesn't accept label as key**, then rules_pkg has to convert it back to a string.

In Bzlmod, we enforces repo level [strict deps](https://github.com/bazelbuild/bazel/issues/13793) via repo mapping, each repo has a repo mapping that contains all the repos it can see. In this case, if A depends on rules_pkg at version 0.5.1, then there is a mapping `@rules_pkg`(local name) -> `@rules_pkg.0.5.1`(canonical name). We resolve the repo mapping when constructing a Label from a label string.

This will break the above use case, because `_stamp_condition` will have to go through repo mapping twice. The first time, `//private:private_stamp_detect` is resolved to `@rules_pkg.0.5.1//private:private_stamp_detect` with rules_pkg's repo mapping, and the second time the `select` statement tries to resolve `@rules_pkg.0.5.1//private:private_stamp_detect` with A's repo mapping but only finds out `@rules_pkg.0.5.1` (local name) doesn't exist because the local repo name should be `@rules_pkg`.

The solution is to let `select` accept Label as the key type, therefore we can skip the second repo mapping.

fmeum:
When a macro specifies a label string as a key in a select, this label
is resolved relative to the site of use rather than the .bzl file the
macro is defined in. The resolution will lead to incorrect results if
the repository that uses the macro has a different repo mapping, e.g.
because it is created by another Bazel module.

This can be solved by allowing macros to specify label instances created
with the `Label` constructor instead of label strings everywhere, which
previously was not possible in select.

This commit also updates the docs for Label, select and macros.

Fixes #14259.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
@@ -330,8 +330,10 @@ public final class StarlarkLibrary {
               name = "x",
               positional = true,
               doc =
-                  "A dict that maps configuration conditions to values. Each key is a label string"
-                      + " that identifies a config_setting instance."),
+                  "A dict that maps configuration conditions to values. Each key is a "
+                      + "<a href=\"$BE_ROOT/../skylark/lib/Label.html\">Label</a> or a label string"
+                      + " that identifies a config_setting, constraint_setting, or constraint_value"
+                      + " instance."),
           @Param(
               name = "no_match_error",
               defaultValue = "''",
--- a/site/docs/skylark/macros.md
+++ b/site/docs/skylark/macros.md
@@ -143,6 +143,38 @@ macro), use the function [native.package_name()](lib/native.html#package_name).
 Note that `native` can only be used in `.bzl` files, and not in `WORKSPACE` or
 `BUILD` files.
 
+## Label resolution in macros
+
+Since macros are evaluated in the [loading phase](concepts.md#evaluation-model),
+label strings such as `"//foo:bar"` that occur in a macro are interpreted
+relative to the `BUILD` file in which the macro is used rather than relative to
+the `.bzl` file in which it is defined. This behavior is generally undesirable
+for macros that are meant to be used in other repositories, e.g. because they
+are part of a published Starlark ruleset.
+
+To get the same behavior as for Starlark rules, wrap the label strings with the
+[`Label`](lib/Label.html#Label) constructor:
+
+```python
+# @my_ruleset//rules:defs.bzl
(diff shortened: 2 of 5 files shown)


fmeum:
@Wyverald @meteorcloudy

Wyverald:
What a nice PR! Thank you, Fabian! I'll import this shortly.

src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
```diff
@@ -330,8 +330,10 @@ public Depset depset(
               name = "x",
               positional = true,
               doc =
-                  "A dict that maps configuration conditions to values. Each key is a label string"
-                      + " that identifies a config_setting instance."),
+                  "A dict that maps configuration conditions to values. Each key is a "
```
gregestren, line 333: I agree this is a nice PR!

My one quibble is documentation that says "A may be X or Y" without differentiating details is fundamentally ambiguous.

You explain the differentiation particularly well in `macros.md`. I wonder if a cross-reference there would help this.

fmeum: Thanks for the suggestion. I created #14458 as a follow-up, could you take a look?

meteorcloudy:
Nice, thanks for your contribution! @fmeum

fmeum:
@Wyverald Could this and #14458 be included in 5.1?

brentleyjones:
@bazel-io flag

meteorcloudy:
@bazel-io fork 5.1

PR #14455, opened 2021-12-20
Rename --project_id to --bes_instance_name

Yannic:
This makes it clearer that this flag affects the BES and not, e.g.,
remote cache or execution. `--bes_instance_name` was chosen to match
`--remote_instance_name` as closely as possible.

`--project_id` always seemed very Google/GCP centric, and now that
there are multiple other BES implementations like BuildBuddy or EngFlow
Build and Test UI, it makes sense to change the terminology to follow
the conventions from the REAPI. We will keep `--project_id` as an alias
for the foreseable future.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
@@ -88,13 +88,14 @@ public class BuildEventServiceOptions extends OptionsBase {
   public boolean besLifecycleEvents;
 
   @Option(
-    name = "project_id",
+    name = "bes_instance_name",
+    oldName = "project_id",
     defaultValue = "null",
     documentationCategory = OptionDocumentationCategory.LOGGING,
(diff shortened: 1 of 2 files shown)


Yannic:
@meisterT PTAL

src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceOptions.java
```diff
@@ -88,13 +88,14 @@
   public boolean besLifecycleEvents;
 
   @Option(
-    name = "project_id",
+    name = "bes_instance_name",
+    oldName = "project_id",
     defaultValue = "null",
     documentationCategory = OptionDocumentationCategory.LOGGING,
     effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
     help = "Specifies the BES project identifier. Defaults to null."
```
michaeledgar, line 96: Can we update the help string, eg. "Specifies the instance name under which BES will store uploaded BEP. Defaults to null."

michaeledgar:
This looks good - I'll pull this in and update the help string while I'm there.

PR #14458, opened 2021-12-21
Improve documentation for select()

fmeum:
* Reference documentation that explains when to use Label rather than a label string.
* Remove mention of constraint_settings as they can't be used as select keys.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
@@ -332,8 +332,9 @@ public final class StarlarkLibrary {
               doc =
                   "A dict that maps configuration conditions to values. Each key is a "
                       + "<a href=\"$BE_ROOT/../skylark/lib/Label.html\">Label</a> or a label string"
-                      + " that identifies a config_setting, constraint_setting, or constraint_value"
-                      + " instance."),
+                      + " that identifies a config_setting or constraint_value instance. See the"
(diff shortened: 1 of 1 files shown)


fmeum:
@gregestren

gregestren (approved):
Thanks for this followup.

 I didn't even notice the bug about `constraint_setting`, nice catch.

FYI because of the holidays this PR may take more time than usual to get merged. Apologies for that, but we'll get it in.

src/main/java/com/google/devtools/build/lib/packages/StarlarkLibrary.java
```diff
@@ -332,8 +332,9 @@ public Depset depset(
               doc =
                   "A dict that maps configuration conditions to values. Each key is a "
                       + "<a href=\"$BE_ROOT/../skylark/lib/Label.html\">Label</a> or a label string"
-                      + " that identifies a config_setting, constraint_setting, or constraint_value"
-                      + " instance."),
+                      + " that identifies a config_setting or constraint_value instance. See the"
+                      + " <a href=\"$BE_ROOT/../skylark/macros.html#label-resolution-in-macros>"
```
gregestren, line 336: Oops, I found a small bug! No closing \" on the href. See https://docs-staging.bazel.build/5813/versions/main/skylark/lib/globals.html#select.

fmeum: Thanks for catching this, fixed.

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 5.1

PR #14470, opened 2021-12-23
Adding CPU Measurements and Unit Tests

ecngtng:
This commit is about reporting cpu user time and cpu system time, 
that is already available, but not printed in any log.

### **Part one**
The solution adds CPU metrics after a build as seen below:

INFO: 2 processes: 1 internal, 1 linux-sandbox.
INFO: Total action wall time 70.05s
INFO: Critical path 70.08s (setup 0.00s, action wall time 70.05s)
INFO: Elapsed time 71.58s (preparation 1.45s, execution 70.13s)
_INFO: CPU time 5.31s (user 0.00s, system 0.01s, bazel 5.29s)_

If any of the values is not  collected, "???" is printed instead like:

INFO: 415 processes: 1 internal, 414 processwrapper-sandbox.
INFO: Total action wall time 127.84s
INFO: Critical path 2.58s (setup 0.00s, action wall time 2.51s)
INFO: Elapsed time 38.37s (preparation 3.98s, execution 34.39s)
_INFO: CPU time ???s (user 79.45s, system 22.39s, bazel ???s)_
INFO: Build completed successfully, 415 total actions

**Note:**
The values is only presented if the  following flag is set:
--experimental_stats_summary

The following flags must  be set to  get the  measurements to be
ollected:
--experimental_collect_local_sandbox_action_metrics
--experimental_profile_cpu_usage (set but default)

### **Part two**
Unit Test is added in runtime-tests.
Test command:
./misc/bazel test --experimental_stats_summary //src/test/java/com/google/devtools/build/lib:runtime-tests

**Note:**
Due to Reporter class has keywords 'final', it'll block mock in UT.
We didn't find a better way to solve this issue.
So in this commit, this keyword 'final' in Reporter was removed.


ecngtng:
Hello, who knows how to see the fail log as below:
(10:58:31) FAIL: //src/test/java/com/google/devtools/build/lib:runtime-tests (see /private/var/tmp/_bazel_buildkite/5e1880647af908bfbf43bdcc50066287/execroot/io_bazel/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/google/devtools/build/lib/runtime-tests/test_attempts/attempt_1.log)

ecngtng:
Hello, due to I couldn't reproduce the issue locally. Who knows where to get the the log as below:
FAILED: //src/test/java/com/google/devtools/build/lib:runtime-tests (Summary)
/private/var/tmp/_bazel_buildkite/5e1880647af908bfbf43bdcc50066287/execroot/io_bazel/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/google/devtools/build/lib/runtime-tests/test.log
/private/var/tmp/_bazel_buildkite/5e1880647af908bfbf43bdcc50066287/execroot/io_bazel/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/google/devtools/build/lib/runtime-tests/test_attempts/attempt_1.log
/private/var/tmp/_bazel_buildkite/5e1880647af908bfbf43bdcc50066287/execroot/io_bazel/bazel-out/darwin-fastbuild/testlogs/src/test/java/com/google/devtools/build/lib/runtime-tests/test_attempts/attempt_2.log

ecngtng:
Here is the answer for my question above.:
1.	Choose the failed test suite, by pressing ”details” link in github. That will take you to https://buildkite.com
2.	Choose the failed test suite again in buildkite.com.
3.	Switch tab from ”Log” to ”Artifacts” for the suite in buildkite.com, to access the log files.

ecngtng:
New commit fixed the issue in Mac OS. It's due to Mac OS uses "," to  divide integer and decimal fraction instead of '.' in other OS.

E.g.: In Mac OS, it's '9,00'. In other OS, it's '9.00'.

larsrc-google:
@ecngtng It's entirely possible that that difference in punctuation is not due to being on Mac per se, but due to Mac doing different i18n.

ecngtng:
@larsrc-google  So in this case, is it reasonable to distinguish the punctuation with related key words in OS name? Or?

ulrfa:
Distinguishing on OS might cause problems for users in other countries with different default Locale.

One alternative could be to get the JVMs decimal separator char for current `Locale.getDefault()`, via `DecimalFormatSymbols.getInstance().getDecimalSeparator()` and use that as condition.

Another alternative could be to use `String.format` also when constructing the expected result in the test case.

ecngtng:
@ulrfa @larsrc-google Thanks for your good comments. Codes have been updated with your comment.

meisterT:
Is this the right place to expose it? Most users likely don't care about CPU time.

The idea behind `--experimental_stats_summary` is to provide the most useful information in a more concise way.

If a user cares, they can always look it up in the BEP and the trace profile.

ulrfa:
Thank you @meisterT  for reviewing!

AFAIK, BEP provides CPU time only for:
- The JVM process.

and the `--profile` json file contains CPU usage only for:
-	The JVM Process.
-	All processes running on local host (including other builds and other concurrently executing processes).

I assume the metrics **above** are relevant mostly for the developers of bazel codebase itself. But **this PR** provides CPU time for:
-	The JVM process. (“bazel”)
-	All **subprocesses spawned** in sandboxes for the particular build. (“user” and “system”)

I think the CPU time, for the JVM and **subprocesses** combined, is one of the most relevant metrics for a user that would like to grasp the overall resource consumption of a build.

In the future, I imagine storing subprocess's CPU time **per action** in the profile json file, and then calculate summary grouped by mnemonic, to see how much CPU each tool in the build chain consumes. But I think this PR is a good start and makes information conveniently available.

Perhaps `--experimental_stats_summary` is not suitable for this. @meisterT, do you have ideas about other ways to control verbosity?

ecngtng:
@meisterT  What are your comments to @ulrfa's response?  :   )

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -109,7 +116,20 @@ public void executionPhaseFinish(@SuppressWarnings("unused") ExecutionFinishedEv
   @AllowConcurrentEvents
   public void actionResultReceived(ActionResultReceivedEvent event) {
     spawnStats.countActionResult(event.getActionResult());
-  }
+    Optional<Duration> cpuUserTimeForActionsDuration = event.getActionResult().cumulativeCommandExecutionUserTime();
+
+    if(cpuUserTimeForActionsDuration.isPresent() &&  (cpuUserTimeForActions != UNKNOWN)) {
+      cpuUserTimeForActions = cpuUserTimeForActions + cpuUserTimeForActionsDuration.get().toMillis();
```
larsrc-google, line 122: You can use += here and below.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -60,6 +63,10 @@
   private long executionEndMillis;
   private SpawnStats spawnStats;
   private Path profilePath;
+  private static final long UNKNOWN = -1;
+  private long cpuUserTimeForActions = 0;
```
larsrc-google, line 67: Please document the behaviour of these fields WRT the "UNKNOWN" value - as I read the code, if any action does not have CPU time, the field becomes UNKNOWN and no other CPU time values are recorded.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -109,7 +116,20 @@ public void executionPhaseFinish(@SuppressWarnings("unused") ExecutionFinishedEv
   @AllowConcurrentEvents
   public void actionResultReceived(ActionResultReceivedEvent event) {
     spawnStats.countActionResult(event.getActionResult());
-  }
+    Optional<Duration> cpuUserTimeForActionsDuration = event.getActionResult().cumulativeCommandExecutionUserTime();
+
+    if(cpuUserTimeForActionsDuration.isPresent() &&  (cpuUserTimeForActions != UNKNOWN)) {
```
larsrc-google, line 121: Nit: Unnecessary extra parentheses.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -206,6 +237,25 @@ public void buildComplete(BuildCompleteEvent event) {
...
+
+  private static String formatCpuTime(long milliseconds) {
+    if (milliseconds == UNKNOWN) {
+      return "???s";
+    } else {
+      return String.format("%.2fs", milliseconds / 1000.0);
+    }
+  }
+
+  private static long sumCpuTimes(long a, long b, long c) {
```
larsrc-google, line 254: Since you're hardcoding this to three parameters, you might as well name them with what kinds of CPU times they are.

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private ActionResultReceivedEvent createActionEvent(Duration userTime, Duration systemTime) {
+    ActionResult result = mock(ActionResult.class);
+
+    if(userTime == null){
+      when(result.cumulativeCommandExecutionUserTime()).thenReturn(Optional.empty());
+    }
+    else{
+      when(result.cumulativeCommandExecutionUserTime()).thenReturn(Optional.of(userTime));
```
larsrc-google, line 81: You can simplify these with `Optional.ofNullable()`.

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
```diff
@@ -0,0 +1,168 @@
...
+    ActionResultReceivedEvent action1 = createActionEvent(Duration.ofSeconds(50), Duration.ofSeconds(20));
+    ActionResultReceivedEvent action2 = createActionEvent(Duration.ofSeconds(5), Duration.ofSeconds(2));
+    buildSummaryStatsModule.actionResultReceived(action1);
+    buildSummaryStatsModule.actionResultReceived(action2);
+    Field field1 = tBSSM.getDeclaredField("cpuTimeForBazelJvm");
+    field1.setAccessible(true);
+    field1.setLong(buildSummaryStatsModule, 11000);
+    buildSummaryStatsModule.buildComplete(createBuildEvent());
+    if(Comma) {
+      verify(reporterMock).handle(Event.info("CPU time 88,00s (user 55,00s, system 22,00s, bazel 11,00s)"));
```
larsrc-google, line 112: I think the comma problem would be easier to understand if you use `String.format` here instead of having the `Comma` variable. A note on why it's done that way would be good, though.

larsrc-google (commented):
I started a review, then realized a general problem: Action CPU time is readily available for standalone/sandboxed actions. For singleplex worker actions, it could be found by taking the processes CPU time before and after making a request (I don't think we do that by default, but we could). But multiplex worker actions simply don't have a well-defined CPU time concept (actually the singleplex action would paper over some things like GC between requests). Since even a single action without CPU time makes the whole thing unknown, it will probably end up being unknown more often than not.

The way around this would be to measure non-worker actions the way you currently do, but measure worker actions by taking the CPU time at start and end (of build or of worker).

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
ecngtng: For this comment, I think you want it to be changed like:
verify(reporterMock).handle(Event.info(**String.format("CPU time %.2fs (user %.2fs, system %.2fs, bazel %.2fs)",88.00, 55.00, 22.00, 11.00))**);
Is it right?

larsrc-google: Yes, that should do it. Plus a comment on why it's not just a fixed string.

ecngtng: I hope, in this way, it could be determined by the system to use '.' or ','  to expresss the floating number.

ckolli5:
Hello @ecngtng, could you please confirm that all the review comment are taken care? Thanks.

ecngtng:
@ckolli5 yes, all the comments are taken care.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -60,6 +63,11 @@ public class BuildSummaryStatsModule extends BlazeModule {
   private long executionEndMillis;
   private SpawnStats spawnStats;
   private Path profilePath;
+  private static final long UNKNOWN_CPU_TIME = -1;
+  //If anyone of the CPU time below is UNKNOWN, the total CPU time will be UNKNOWN. 
+  private long cpuUserTimeForActions = 0;
+  private long cpuSystemTimeForActions = 0;
```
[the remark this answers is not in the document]
ecngtng: The usage of these variables is pretty simple. Are there any benefits to change them to duration?  :   )

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -191,6 +215,14 @@ public void buildComplete(BuildCompleteEvent event) {
                     (now - commandStartMillis) / 1000.0,
                     overheadTime / 1000.0,
                     executionTime / 1000.0)));
+        reporter.handle(
+            Event.info(
+                String.format(
+                    "CPU time %s (user %s, system %s, bazel %s)",
```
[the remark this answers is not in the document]
ecngtng: Do you mean "CPU time %s (user %s, system %s, bazel **jvm** %s)"?

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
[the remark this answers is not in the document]
larsrc-google: Using a Duration would make it much harder to make unit errors. Right now, these are in... seconds? Nanoseconds? Fortnights? Microcenturies? I can't know! If this was a proto that we created millions of instances of, I could see using a primitive type would save some space, but here clarity is more important.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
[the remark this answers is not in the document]
wilwell: Yes

ecngtng:
Code has been changed according to comments and all checks have passed.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -206,6 +229,34 @@ public void buildComplete(BuildCompleteEvent event) {
         criticalPathComputer = null;
       }
       profilePath = null;
+      cpuUserTimeForActions = Duration.ofMillis(0);
```
wilwell, line 232: It's better to initialise this variables on other method, e.g. on `executorInit`.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -60,6 +63,11 @@ public class BuildSummaryStatsModule extends BlazeModule {
   private long executionEndMillis;
   private SpawnStats spawnStats;
   private Path profilePath;
+  private static final long UNKNOWN_CPU_TIME = -1;
+  //If anyone of the CPU time below is UNKNOWN, then the total CPU time become UNKNOWN.
```
wilwell, line 67: I think, that this comment make more sense as documentation of method `sumCpuTimes `. Please move it there.

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.runners.JUnit4;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
+
+@RunWith(JUnit4.class)
+public class BuildSummaryStatsModuleTest {
+  private BuildSummaryStatsModule buildSummaryStatsModule;
+  private Reporter reporterMock;
+  private ActionKeyContext actionKeyContextMock;
```
wilwell, line 42: You don't need this filed in tests. Create it only on `setUp()`

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
```diff
@@ -0,0 +1,126 @@
...
+    buildSummaryStatsModule.buildComplete(createBuildEvent());
+    verify(reporterMock).handle(Event.info(String.format("CPU time ???s (user %.2fs, system ???s, bazel jvm ???s)", 50.00)));
+  }
+
+  @Test
+  public void reusedBuildSummaryStatsModuleIsClearedBetweenBuilds() throws Exception {
+    ActionResultReceivedEvent action1 = createActionEvent(Duration.ofSeconds(50),
+                                                          Duration.ofSeconds(20));
+    buildSummaryStatsModule.actionResultReceived(action1);
+    Field field1 = tBSSM.getDeclaredField("cpuTimeForBazelJvm");
```
wilwell, line 117: You could mock `Profiler` and get rid of this reflection too.

wilwell (changes requested):

ecngtng:
I just re-requested review from [wilwell](https://github.com/wilwell). 
But I didn't remove request for [meisterT](https://github.com/meisterT) and [larsrc-google](https://github.com/larsrc-google).
I don't know why they two were removed by system. Or any wrong action was done by me?
Anyway, any comments are welcome!

src/test/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModuleTest.java
```diff
@@ -0,0 +1,125 @@
...
```
wilwell, line 116: Please create `@VisibleOnlyForTesting` method with setter of this field, and use this method instead of using this mechanism

wilwell: Same for other places, where you are using this mechanism.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
wilwell: Could you move please this inititalization on `executorInit`?

wilwell (changes requested):

ecngtng:
@wilwell  thanks for your comments. 
And for your commnets about 'Could you move please this inititalization on executorInit?',  I tried as you mentioned to move the initialization on executorInit. But some test case failed after it. It seems that it's more safe to keep them here.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -228,6 +251,47 @@ public void buildComplete(BuildCompleteEvent event) {
...
+    if(sumDuration.isPresent() && (termDuration.toMillis() !=  UNKNOWN_CPU_TIME)) {
+      termDuration = termDuration.plus(sumDuration.get());
+    } else {
+      termDuration = Duration.ofMillis(UNKNOWN_CPU_TIME);
     }
+    return termDuration;
+  }
+
+  @VisibleForTesting
+  void setcpuTimeForBazelJvm(Duration time)
```
wilwell, line 287: Please rename on `setCpuTimeForBazelJvm`

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -228,6 +251,47 @@ public void buildComplete(BuildCompleteEvent event) {
...
+  {
+    this.cpuTimeForBazelJvm = time;
+  }
+
+  @VisibleForTesting
+  void setstatsSummary(boolean summary)
```
wilwell, line 293: You could call `executorInit` with options, instead of creating this method. Try to use` @VisibleForTesting` methods, only if there is no other alternatives.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
wilwell: I assume that you could move this to `executorInit` and call this method on `setUp()` on tests.

wilwell (changes requested):

ecngtng:
@wilwell As your comments 'Call executorInit with options, instead of creating this method', I made some code change. The test passed in my local environment. But there were 2 failing fired when they were pushed to github. After checking related logs, the root cause was not dug out yet. Could you please help to check and give any clue? Thanks!

ecngtng:
From log, we found the failure was because of 'one test failed in Unit tests for tools/test/collect_cc_code_coverage.sh' (i.e.: test_cc_test_coverage_gcov).  And 'collect_cc_code_coverage.sh' is to collect code coverage for C++ sources.  But our code change doesn't relate to C++. I'm curious how it could be impacted?

wilwell:
The problem with tests looks strange. Have you tried to merge/pull master branch into yours?

ecngtng:
> The problem with tests looks strange. Have you tried to merge/pull master branch into yours?

Do you mean to rebase my commit?

wilwell:
Yes, rebase on master your commit and rerun tests please.

ecngtng:
@wilwell all checks have passed. If you have any further comments, please feel free to let's know.

wilwell (approved):

wilwell:
@larsrc-google could you please also take a look ob this change. If everything is ok, I am gonna merge it.

ecngtng:
@larsrc-google could you please help to take a look for this change as it has passed so long time?  :  )

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
larsrc-google: I would also prefer having these on `executorInit` or some other part of startup, to be sure that the values are zeroed out regardless of how the previous build ended.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -131,6 +139,10 @@ public void executionPhaseFinish(@SuppressWarnings("unused") ExecutionFinishedEv
   @AllowConcurrentEvents
   public void actionResultReceived(ActionResultReceivedEvent event) {
     spawnStats.countActionResult(event.getActionResult());
+    Optional<Duration> cpuUserTimeForActionsDuration = event.getActionResult().cumulativeCommandExecutionUserTime();
```
larsrc-google, line 142: Nit: You don't need to have `Duration` at the end of this variable name, that's obvious from the type. In fact, since the variable is so short-lived, use a shorter variable name, or just inline it.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -228,6 +251,41 @@ public void buildComplete(BuildCompleteEvent event) {
...
+  //If anyone of the CPU time(cpuUserTimeActions, cpuSystemTimeActions and cpuTimeBazelJvm) is UNKNOWN, then the total CPU time become UNKNOWN.
+  private static long sumCpuTimes(long cpuUserTimeActions, long cpuSystemTimeActions, long cpuTimeBazelJvm) {
+    if ((cpuUserTimeActions == UNKNOWN_CPU_TIME) || (cpuSystemTimeActions == UNKNOWN_CPU_TIME) || (cpuTimeBazelJvm == UNKNOWN_CPU_TIME)) {
+      return UNKNOWN_CPU_TIME;
+    } else {
+      return cpuUserTimeActions + cpuSystemTimeActions + cpuTimeBazelJvm;
+    }
+  }
+  
+  private static Duration addCPUTime(Optional<Duration> sumDuration, Duration termDuration) {
```
larsrc-google, line 277: This method needs to be threadsafe, since it's called from a method marked `@AllowConcurrentEvents`.

ecngtng:
@larsrc-google Codes have been changed according to your comments. Could you please help to check? Thanks!

src/main/java/com/google/devtools/build/lib/events/Reporter.java
```diff
@@ -39,7 +39,7 @@
  * <p>Thread-safe: calls to {@code #report} may be made on any thread. Handlers may be run in an
  * arbitrary thread (but right now, they will not be run concurrently).
  */
-public final class Reporter implements ExtendedEventHandler {
```
larsrc-google, line 42: Rather than making Reporter non-final for mocking purposes, can't you use the [EventCollector](src/main/java/com/google/devtools/build/lib/events/EventCollector.java) as in for instance [FoundationTestCase.java](src/test/java/com/google/devtools/build/lib/testutil/FoundationTestCase.java)?

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
```diff
@@ -182,6 +194,9 @@ public void buildComplete(BuildCompleteEvent event) {
         // Since the BEP currently shuts down at the BuildCompleteEvent, we cannot just move posting
         // the BuildToolLogs to afterCommand of this module.
         try {
+          if(Profiler.getProcessCpuTimeMaybe() != null) {
```
larsrc-google, line 197: Nit: I'd prefer to not call `getProcessCpuTimeMaybe()` twice.

src/main/java/com/google/devtools/build/lib/runtime/BuildSummaryStatsModule.java
larsrc-google: Zeroing them on startup would also make it clearer that they are properly initialized.

larsrc-google:
Thank you. You missed one comment from before, and I added a couple small things.

larsrc-google:
I don't see the zeroing happening in `executorInit()`, only in `buildComplete()`. If you zeroed them in `executorInit()`, that would be sufficient.

ecngtng:
@larsrc-google  You're right.  After checking, I found they were really not in last commits. I added them back now.  (Every interesting, I ever checked codes and confirmed that they merged several days ago.)

ecngtng:
@larsrc-google Codes have been changed and all checks passed. Could you please help to check? Thanks!

larsrc-google:
Looks nice. I'll do some extra checks internally, and then I think we can merge it.

larsrc-google:
Getting bogged down in all the technical bits, I forgot about @meisterT's comments above. Sorry. We want to reduce the terminal output, not increase it. Putting it in the log (as indicated in the original comment) would be fine, though.

sgowroji:
Hi @wilwell, Since I can see that this PR has been approved, please let me know whether I should proceed with importing it.

larsrc-google:
No, not yet. As per my last comment adding this info to the output from every build is not desirable.

meisterT:
Closing as per above's comment. Feel free to re-open and making the info available in the BEP or the info log.

PR #14498, opened 2021-12-29
Changes http_archive's netrc attr to a Label.

jiawen:
This makes it possible to use a netrc file in another repository (e.g.,
the current repository) rather than referring to a netrc *file* inside
the external repository (which usually does not make sense).

Closes: 13662

RELNOTES[INC]: this incompatible change breaks old instances of http_archive that specified netrc as an absolute path. It is unlikely there are many instances in the wild since the path would refer to a netrc file inside the external repository by absolute path. Migration should be straightforward.

### DIFF
--- a/src/test/shell/bazel/starlark_repository_test.sh
+++ b/src/test/shell/bazel/starlark_repository_test.sh
@@ -2008,16 +2008,12 @@ function test_http_archive_netrc() {
   tar cvf x.tar x
   sha256=$(sha256sum x.tar | head -c 64)
   serve_file_auth x.tar
-  netrc_dir="$(pwd)"
-  if "$is_windows"; then
-    netrc_dir="$(cygpath -m ${netrc_dir})"
-  fi
   cat > WORKSPACE <<EOF
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
   name="ext",
   url = "http://127.0.0.1:$nc_port/x.tar",
-  netrc = "${netrc_dir}/.netrc",
+  netrc = "@//:.netrc",
   sha256="$sha256",
 )
 EOF
@@ -2045,16 +2041,12 @@ function test_http_archive_auth_patterns() {
   tar cvf x.tar x
   sha256=$(sha256sum x.tar | head -c 64)
   serve_file_auth x.tar
-  netrc_dir="$(pwd)"
-  if "$is_windows"; then
-    netrc_dir="$(cygpath -m ${netrc_dir})"
-  fi
   cat > WORKSPACE <<EOF
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
   name="ext",
   url = "http://127.0.0.1:$nc_port/x.tar",
-  netrc = "${netrc_dir}/.netrc",
+  netrc = "@//:.netrc",
   sha256="$sha256",
   auth_patterns = {
     "127.0.0.1": "Bearer <password>"
--- a/tools/build_defs/repo/http.bzl
+++ b/tools/build_defs/repo/http.bzl
@@ -444,8 +444,9 @@ unless it was added to the cache by a request with the same canonical id.
 Each entry must be a file, http or https URL. Redirections are followed.
 Authentication is not supported.""",
     ),
-    "netrc": attr.string(
-        doc = "Location of the .netrc file to use for authentication",
+    "netrc": attr.label(
+        doc = "A label that specifies the location of the .netrc file to use for authentication.",
+        allow_single_file = True,
     ),
     "auth_patterns": attr.string_dict(
         doc = _AUTH_PATTERN_DOC,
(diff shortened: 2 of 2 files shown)


src/test/shell/bazel/starlark_repository_test.sh
```diff
@@ -2045,16 +2041,12 @@ function test_http_archive_auth_patterns() {
...
-  if "$is_windows"; then
-    netrc_dir="$(cygpath -m ${netrc_dir})"
-  fi
   cat > WORKSPACE <<EOF
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
   name="ext",
   url = "http://127.0.0.1:$nc_port/x.tar",
-  netrc = "${netrc_dir}/.netrc",
+  netrc = "@//:.netrc",
```
jiawen, line 2049: This seems fine as existing tests still work.

Let me know if you need more tests - e.g., create yet another external repository (`local_repository`?) with just the netrc in it and read it. Seems like overkill though.

tools/build_defs/repo/http.bzl
```diff
@@ -241,8 +241,9 @@ to omit the checksum as remote files can change._ At best omitting this
 field will make your build non-hermetic. It is optional to make development
 easier but either this attribute or `sha256` should be set before shipping.""",
     ),
-    "netrc": attr.string(
+    "netrc": attr.label(
```
jiawen, line 244: Let me know if this needs additional comments - or if there's additional documentation that needs updating.

meteorcloudy:
Thanks! Technically this is a breaking change, because the old way of specifying an absolute path won't work anymore. But I agree making "netrc" a Label is much better than before, and it won't be hard to migrate.

Can you put:
"RELNOTES[INC]: \<description of the incompatible change\>" at the bottom of your PR description?

src/test/shell/bazel/starlark_repository_test.sh
meteorcloudy: We are also introducing the new external dependency system for Bazel: https://docs.bazel.build/versions/main/bzlmod.html#strict-deps, which will make "@//" no longer valid in some cases, but I think it's fine for now in this test.

tools/build_defs/repo/http.bzl
meteorcloudy: Maybe "A label that specifies the location of the .netrc file to use for authentication" is slightly better?

meteorcloudy (changes requested):

jiawen:
> Thanks! Technically this is a breaking change, because the old way of specifying an absolute path won't work anymore. But I agree making "netrc" a Label is much better than before, and it won't be hard to migrate.
> 
> Can you put: "RELNOTES[INC]: <description of the incompatible change>" at the bottom of your PR description?

Done.

src/test/shell/bazel/starlark_repository_test.sh
jiawen: Got it. I've been eagerly following bzlmod waiting for it to be "production ready". In the meantime, and probably going forward, my projects need `http_archive` for data dependencies (e.g., test data).

tools/build_defs/repo/http.bzl
jiawen: Done.

jiawen:
I just noticed `http_file` and `http_jar`. Shall I update those (and their tests) as well, for consistency?

meteorcloudy (approved):
Thanks for making this better!

meteorcloudy:
Unfortunately this is breaking many projects in downstream, I'll have to rollback it first.

jiawen:
Shucks. No problem. I'm happy to continue working in this but would like some advice as to how.

AFAICT, downstream projects aren't in a monorepo with Bazel. How does one make this kind of breaking API change atomically with known dependencies?

An incompatible flag is too heavyweight. Perhaps a new `netrc_label` attr? It sadly has a terrible name and will need tests...

meteorcloudy:
This is indeed tricky, the best thing I can think of is:

1. Keep `netrc` as a attr.string()
2. Before invoking [read_netrc()](https://cs.opensource.google/bazel/bazel/+/master:tools/build_defs/repo/http.bzl;l=79), we check if `ctx.attr.netrc` is a label-like string (starting with `//`), if so convert it to a path via `ctx.path(Label(ctx.attr.netrc))`

The upside is that we can pass both path and "label", but the downside is that this won't be compatible with Bzlmod, because the string was converted to Label in `http.bzl`, which will use the repository mapping of @bazel_tools instead of the repo who defines the `http_archive`. See more details here: https://docs.bazel.build/versions/main/bzlmod.html#repository-names

Nevertheless, we can then get the chance to migrate [rules_jvm_external](https://github.com/bazelbuild/rules_jvm_external/pull/306), which is blocking the downstream. Print a warning message when path is used for netrc attribute, and eventually change the type to Label.

@comius Have you ever encountered a case where you have to change a type of an attribute? How did that proceed?

FYI @Wyverald in case you have better ideas ;)

Wyverald:
The "string-but-treat-as-label-sometimes" setup doesn't work very well even without Bzlmod -- if a user passes in "//some:label" it should refer to "@//some:label", but if we just treat it as a label it would point to "@bazel_tools//some:label". Not to mention that the user could pass in a relative label (":label"), which doesn't work for a Label() call.

IMO the `netrc_label` attribute idea is the only feasible one.

meteorcloudy:
> The "string-but-treat-as-label-sometimes" setup doesn't work very well even without Bzlmod ...

Agree, there is too much risk with this approach. 

> IMO the netrc_label attribute idea is the only feasible one.

@jiawen Sounds like this is the only way forward

PR #14499, opened 2021-12-29
update version of build rules to latest available

skofgar:
The version numbers in the doc are quite far behind. This PR is updating them to the latest versions available.

### DIFF
--- a/site/docs/tutorial/ios-app.md
+++ b/site/docs/tutorial/ios-app.md
@@ -116,25 +116,25 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 git_repository(
     name = "build_bazel_rules_apple",
     remote = "https://github.com/bazelbuild/rules_apple.git",
-    tag = "0.19.0",
+    tag = "0.32.0",
 )
 
 git_repository(
     name = "build_bazel_rules_swift",
     remote = "https://github.com/bazelbuild/rules_swift.git",
-    tag = "0.13.0",
+    tag = "0.24.0",
 )
 
 git_repository(
     name = "build_bazel_apple_support",
     remote = "https://github.com/bazelbuild/apple_support.git",
-    tag = "0.7.2",
+    tag = "0.11.0",
 )
 
 git_repository(
     name = "bazel_skylib",
     remote = "https://github.com/bazelbuild/bazel-skylib.git",
-    tag = "0.9.0",
+    tag = "1.1.1",
 )
 ```
 


skofgar:
I updated my CLA. Please do a CLA rescan

aiuto (commented):
@keith  Does it make sense to continue to have this tutorial in bazelbuild/bazel?  Should we, instead, move it to rules_apple?

keith:
It's a good question. Some portion of the docs today are about native providers. But the "how to build your first iOS app" should live in the rules

skofgar:
@keith thanks for approving the PR. Can you also re-run the `cla/google` verification?

aiuto:
@skofgar Would you be willing to try to create an initial tutorial in rules_apple?

FWIW, I am trying to move us towards fixing problems as we see them, rather than kicking technical debt down the roads. There are a bunch of documents alongside the Bazel source that simply don't belong there. I am hoping to get interested parties to do the deeper fixes.

cc: @kaylathar My suspicion is that this bazel+ios tutorial is divergent enough from ios usage inside Google that it would probably be less confusing to users internally if it were moved to rules_apple.

skofgar:
Hey @aiuto,
I'm fairly new to Bazel (and Tulsi). My question would be, if the tutorial is moved to [`rules_apple`](https://github.com/bazelbuild/rules_apple), it looks like there would no longer be a tutorial on the Bazel website - for example: https://docs.bazel.build/versions/main/tutorial/ios-app.html ? It looks like it's currently getting content from `bazelbuild/bazel/site/docs/...`

It looks like only the Bazel repository has a `site` folder that will be published as a website. 

(Or would it make sense to create a "documentation website" repository that would import/combine the tutorials from each of the projects and then publish them as a single documentation website?)

aiuto:
Ideally, we would get the tutorials out of the versioned docs (that is, anything under https://docs.bazel.build/versions/main/...).  The main site would point to the tutorials that site with the rules they use. That let's rule authors update their tutorials without being blocked by the Bazel team.

aiuto:
I opted for something else.
As a first step, let's move this page to bazelbuild/examples, since the page points to that. That will make it possible to iterate on the tutorial text and the sample code in the same PR.

https://github.com/bazelbuild/examples/pull/237

aiuto:
On Tue, Jan 4, 2022 at 12:38 PM Roland ***@***.***> wrote:

> Hey @aiuto <https://github.com/aiuto>,
> I'm fairly new to Bazel (and Tulsi). My question would be, if the tutorial
> is moved to rules_apple <https://github.com/bazelbuild/rules_apple>, it
> looks like there would no longer be a tutorial on the Bazel website - for
> example: https://docs.bazel.build/versions/main/tutorial/ios-app.html ?
> It looks like it's currently getting content from
> bazelbuild/bazel/site/docs/...
>
We should move the tutorials out of the main source and into places where
they are easier to contribute to.

> It looks like only the Bazel repository has a site folder that will be
> published as a website.
>
(Or would it make sense to create a "documentation website" repository that
> would import/combine the tutorials from each of the projects and then
> publish them as a single documentation website?)
>
We have the examples directory.   I did first move here:
https://github.com/bazelbuild/examples/pull/237
Waiting for the approve from @gregce

meisterT:
Closing as obsolete.

PR #14507, opened 2022-01-03
[5.x] Rename --project_id to --bes_instance_name

Yannic:
This makes it clearer that this flag affects the BES and not, e.g.,
remote cache or execution. `--bes_instance_name` was chosen to match
`--remote_instance_name` as closely as possible.

`--project_id` always seemed very Google/GCP centric, and now that
there are multiple other BES implementations like BuildBuddy or EngFlow
Build and Test UI, it makes sense to change the terminology to follow
the conventions from the REAPI. We will keep `--project_id` as an alias
for the foreseable future.

Closes #14455.

PiperOrigin-RevId: 418972581


Yannic:
@Wyverald PTAL

/cc @michaeledgar

Issue #8305, opened 2019-05-13
native.bazel_version is not available for macros or rules


haberman:
### Description of the problem / feature request:

I would like access to `native.bazel_version` from BUILD macros and rules. It appears that `native.bazel_version` only works from WORKSPACE macros (as tested [here](https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/skylark_repository_test.sh#L1159-L1199)).

This also means that [versions.bzl from bazel-skylib](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/versions.bzl) only works from `WORKSPACE` files.  This is very surprising and appears to be an undocumented limitation. If you try to invoke it from a `BUILD` file you get:

```
ERROR: /usr/local/google/home/haberman/.cache/bazel/_bazel_haberman/17449526a2508fe9f4def2619c761b7e/external/com_google_protobuf/BUILD:284:2: in //:build_defs.bzl%_upb_proto_libra
ry_aspect aspect on proto_library rule @com_google_protobuf//:field_mask_proto:
Traceback (most recent call last):                                                                                                                                                 
        File "/usr/local/google/home/haberman/.cache/bazel/_bazel_haberman/17449526a2508fe9f4def2619c761b7e/external/com_google_protobuf/BUILD", line 284
                //:build_defs.bzl%_upb_proto_library_aspect(...)                                                                                                                   
        File "/usr/local/google/home/haberman/code/upb/build_defs.bzl", line 383, in _upb_proto_aspect_impl
                cc_library_func(ctx = ctx, name = ctx.rule.attr.na..., <3 more arguments>)
        File "/usr/local/google/home/haberman/code/upb/build_defs.bzl", line 287, in cc_library_func
                versions.get()                                                                                                                                                      
        File "/usr/local/google/home/haberman/.cache/bazel/_bazel_haberman/17449526a2508fe9f4def2619c761b7e/external/bazel_skylib/lib/versions.bzl", line 20, in versions.get       
                native.bazel_version                                                                                                                                                
no native function or rule 'bazel_version'  
```

It took a lot of searching to figure out the reason for this error: `versions.bzl` only works from `WORKSPACE` files.

Since the `native` module is only available during loading (not analysis), it might be better to have a separate module for `bazel_version` that is available during both loading and analysis. Then bazel-skylib's `versions.bzl` could use that. That way it could work from any context.

### Feature requests: what underlying problem are you trying to solve with this feature?

[...]

PR #14508, opened 2022-01-04
Added `native.bazel_version` for use in BUILD and bzl files

UebelAndre:
This change is intended to improve rules maintainer's ability to avoid breaking changes by gating certain functionality behind a version check. Users can now write rules which check the version of Bazel as an indication that certain functionality will be available.

[@rules_python//python/pip_install:requirements.bzl](https://github.com/bazelbuild/rules_python/blob/0.5.0/python/pip_install/requirements.bzl#L77-L82) is one such example where this rule could be used.
```diff
diff --git a/python/pip_install/requirements.bzl b/python/pip_install/requirements.bzl
index b040ebd..219466f 100644
--- a/python/pip_install/requirements.bzl
+++ b/python/pip_install/requirements.bzl
@@ -75,7 +75,7 @@ def compile_pip_requirements(
     }

     # cheap way to detect the bazel version
-    _bazel_version_4_or_greater = "propeller_optimize" in dir(native)
+    _bazel_version_4_or_greater = native.bazel_version[0:2] not in ("0.", "1.", "2.", "3.")

     # Bazel 4.0 added the "env" attribute to py_test/py_binary
     if _bazel_version_4_or_greater:
```

Or, for users of [bazel_skylib](https://github.com/bazelbuild/bazel-skylib):
```diff
diff --git a/python/pip_install/requirements.bzl b/python/pip_install/requirements.bzl
index b040ebd..f5dd545 100644
--- a/python/pip_install/requirements.bzl
+++ b/python/pip_install/requirements.bzl
@@ -1,5 +1,6 @@
 """Rules to verify and update pip-compile locked requirements.txt"""

+load("@bazel_skylib//lib:versions.bzl", "versions")
 load("//python:defs.bzl", "py_binary", "py_test")
 load("//python/pip_install:repositories.bzl", "requirement")

@@ -75,7 +76,7 @@ def compile_pip_requirements(
     }

     # cheap way to detect the bazel version
-    _bazel_version_4_or_greater = "propeller_optimize" in dir(native)
+    _bazel_version_4_or_greater = versions.is_at_least("4.0.0", native.bazel_version)

     # Bazel 4.0 added the "env" attribute to py_test/py_binary
     if _bazel_version_4_or_greater:
```
    

closes https://github.com/bazelbuild/bazel/issues/8305

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/BazelStarlarkEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/BazelStarlarkEnvironment.java
@@ -70,7 +70,7 @@ public final class BazelStarlarkEnvironment {
     this.ruleFunctions = ruleFunctions;
     this.uninjectedBuildBzlNativeBindings =
         createUninjectedBuildBzlNativeBindings(
-            ruleFunctions, packageFunction, environmentExtensions);
+            ruleFunctions, packageFunction, environmentExtensions, version);
     this.workspaceBzlNativeBindings = createWorkspaceBzlNativeBindings(ruleClassProvider, version);
     this.uninjectedBuildBzlEnv =
         createUninjectedBuildBzlEnv(ruleClassProvider, uninjectedBuildBzlNativeBindings);
@@ -143,7 +143,8 @@ public final class BazelStarlarkEnvironment {
   private static ImmutableMap<String, Object> createUninjectedBuildBzlNativeBindings(
       Map<String, ?> ruleFunctions,
       Object packageFunction,
-      List<PackageFactory.EnvironmentExtension> environmentExtensions) {
+      List<PackageFactory.EnvironmentExtension> environmentExtensions,
+      String version) {
     ImmutableMap.Builder<String, Object> env = new ImmutableMap.Builder<>();
     env.putAll(StarlarkNativeModule.BINDINGS_FOR_BUILD_FILES);
     env.putAll(ruleFunctions);
@@ -151,6 +152,7 @@ public final class BazelStarlarkEnvironment {
     for (PackageFactory.EnvironmentExtension ext : environmentExtensions) {
       ext.updateNative(env);
     }
+    env.put("bazel_version", version);
     return env.build();
   }
 
--- a/src/test/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryIntegrationTest.java
+++ b/src/test/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryIntegrationTest.java
@@ -537,4 +537,46 @@ public class StarlarkRepositoryIntegrationTest extends BuildViewTestCase {
     getConfiguredTarget("@foo//p:BUILD"); // (loadPackage(@foo//p) would suffice)
     assertContainsEvent("repo=@foo pkg=p");
   }
+
+  @Test
+  public void testNativeBazelVersion() throws Exception {
+    reporter.removeHandler(failFastHandler);
+    scratch.file(
+        "repo_bzl_ver.bzl",
+        "def _repo_impl(repository_ctx):",
+        "  repository_ctx.file('BUILD', '')",
+        "  repository_ctx.file('defs.bzl', 'BAZEL_VERSION = \"' + native.bazel_version + '\"')",
+        "  repository_ctx.file('WORKSPACE', '')",
+        "",
+        "repo = repository_rule(",
+        "    implementation=_repo_impl,",
+        "    local=True)");
+    scratch.file(
+        "bzl_ver.bzl",
+        "load('@repo2//:defs.bzl', 'BAZEL_VERSION')",
+        "def _impl(ctx):",
+        "  if BAZEL_VERSION == native.bazel_version:",
+        "    print('native.bazel_version matches')",
+        "  else:",
+        "    fail(BAZEL_VERSION + ' does not equal ' + native.bazel_version)",
+        "  output = ctx.actions.declare_file(ctx.label.name + '.txt')",
+        "  ctx.actions.write(output = output, content = native.bazel_version)",
+        "  return DefaultInfo(files = depset([output]))",
+        "",
+        "check = rule(",
+        "    implementation=_impl)");
+    scratch.file("BUILD", "load('@//:bzl_ver.bzl', 'check')", "check(name = 'check')");
+    scratch.overwriteFile(
+        rootDirectory.getRelative("WORKSPACE").getPathString(), 
+        new ImmutableList.Builder<String>()
+        .addAll(analysisMock.getWorkspaceContents(mockToolsConfig))
+        .add("load('@//:repo_bzl_ver.bzl', 'repo')")
+        .add("repo(name = 'repo2')")
+        .build());
+
+    invalidatePackages();
+
+    getConfiguredTarget("//:check");
+    assertContainsEvent("native.bazel_version matches");
+  }
 }


UebelAndre:
Hey, @meteorcloudy, philwo suggested that you might be a good reviewer for this change. Do you have bandwidth to take a look?

meteorcloudy:
See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/WorkspaceFactory.java;l=382;bpv=1;bpt=0

`native.bazel_version` is already available if your .bzl file is used in the WORKSPACE file (repository rules). The change you made is making it also available to the bzl files used in the BUILD files (build rules). I think it's fine, but the test you added is testing the repository rule, so can you please fix that?

FYI @Wyverald We should also make `native.bazel_version` available in Bzlmod?

Wyverald:
I feel a bit uneasy about introducing this. It's essentially another source of non-hermetic input, so it somewhat makes sense for repo rules, but this change would make it available to build rules (and even build macros). IMO the use case is rare enough to warrant the need to jump through hoops (e.g. by using a repo rules to store that information) in favor of the API essentially endorsing its use. cc @brandjon @comius as this is really about the build API.

Nevertheless we should indeed look at adding this for bzlmod.

UebelAndre:
@meteorcloudy The rest I added tests both. I just wasn't sure where to add a test. Happy to move it if you can let me know where

UebelAndre:
> I feel a bit uneasy about introducing this. It's essentially another source of non-hermetic input, so it somewhat makes sense for repo rules, but this change would make it available to build rules (and even build macros).

Can you elaborate here? Why would this be more or less hermetic than getting the value from a generated repository? At the end of the day, both cases produce a string in starlark and not some generated target or file. Having to write a repository rule and load it from there just feels like unnecessary boiler-plate. It now requires consumers or rules to load a new macro to provide the repository instead of just having access to the Bazel version.

UebelAndre:
> Nevertheless we should indeed look at adding this for bzlmod.

@Wyverald @meteorcloudy is that something that should be done in this PR? Where would that code live? I'm pretty unfamiliar with Bazel's code structure.

comius:
I think that we should not add the ability to do version checks in BUILD or bzl files.

Version checks are hard to understand for the readers of the code. This could be amended with extended documentation about, what feature was added and when, in which version of Bazel. Considering the given Python use case, the "feature check" (`hasattr(native, "xxx")`) is on other hand self-documenting (except that instead of storing the result in `_bazel_version_4_or_greater`, the result should be used where a feature is needed or worked around)

Version checks need to be tested against all supported Bazel versions and removed when they are no longer tested. I'm also afraid this will never happen. Maintaining version checking code can potentially become a nightmare. Usually a developer has a single version of Bazel and modifying the version checking code, people will never check if it works with different Bazel versions.

As we are moving language specific rules and modules out of Bazel, version checks should in the future be mostly against modules the current module depends on, and not Bazel itself. bzlmod provides functionality to resolve correct dependent modules. And furthermore also to limit Bazel version with which a version of the module is suppose to work.

There is also a proposal by @philwo, how to version the rules repositories. https://docs.google.com/document/d/1s_8AihGXbYujNWU_VjKNYKQb8_QGGGq7iKwAVQgjbn0/edit#heading=h.3d0j52sbj6iq

So instead of this PR, I propose cleaning up rules_python, to rather do feature checks where needed and in the future to depend on bzlmod functionality to select compatible versions of other modules or enforce Bazel version constraints.

UebelAndre:
@comius how would I handle cases where a bug fix went into a newer version of bazel that's required for enabling certain functionality? I'd run into this case and wanted to inform users that they needed to update Bazel if they wanted to use a particular rule or attribute but was unable to do so as there was no existing repository rule which exposed `native.bazel_version` for me to do the check :(

comius:
> @comius how would I handle cases where a bug fix went into a newer version of bazel that's required for enabling certain functionality? I'd run into this case and wanted to inform users that they needed to update Bazel if they wanted to use a particular rule or attribute but was unable to do so as there was no existing repository rule which exposed native.bazel_version for me to do the check 😞

A principled answer is to let this be handled by Bazel, fixing at head and cherry-picking the fixes to LTS versions we are currently maintaining. And not additionally warning off user.

I see some problems if you try to inform/warn the user. Before there is a fix you don't know at what version it is going to be fixed. You could make a guess guess, but you could also make a wrong guess (this happened to me with Java toolchainisation in bazel_toolchains repo and caused a lot of problems on each Bazel release).

After there is a fix in Bazel, I think the warning for the user has only a really limited value. It would only happen to users who upgraded just the rules repo without upgrading Bazel.

I agree feature checks are hard or impossible in case of bugs. Hopefully bzlmod comes to the rescue.

To keep rules' implementation and macros simple, I would prefer to close this PR now without merging it.
If it happens that bzlmod is not sufficient solution we can revisit this.

UebelAndre:
> A principled answer is to let this be handled by Bazel, fixing at head and cherry-picking the fixes to LTS versions we are currently maintaining. And not additionally warning off user.
> 
> I see some problems if you try to inform/warn the user. Before there is a fix you don't know at what version it is going to be fixed. You could make a guess, but you could also make a wrong guess (this happened to me with Java toolchainisation in bazel_toolchains repo and caused a lot of problems on each Bazel release).

I feel like there is a disconnect here. When there's a bug in Bazel itself, failures may not be obvious as to what the issue is and will require users to investigate something that may already be a known issue. This can be very time consuming. I don't think it makes any sense to perform a version check as a means for warning users that functionality is broken and to guess an arbitrary release where it might be fixed (doing this sounds kinda crazy to me). However, there may be cases (and I've experienced this with `rust_toolchains` https://github.com/bazelbuild/rules_rust/issues/635) where a bug was found that required a fix in upstream Bazel. The fix went in but users continued to run into this issue. To avoid unnecessary headache for my team, I added a version check using a repository rule to detect the version of Bazel and inform them that they needed to update. The repository rule required an additional `load` and would not have worked in the upstream `rules_rust` without forcing users to also load another macro in their workspace file which (IMO) made the change feel unacceptable to upstream to `rules_rust`. The introduction of `native.bazel_version` for rules and macros is not a new concept since it *can* be done using the repository rules but is generally an unusable pattern because of the loads needed.

> I agree feature checks are hard or impossible in case of bugs. Hopefully bzlmod comes to the rescue.
> 
> To keep rules' implementation and macros simple, I would prefer to close this PR now without merging it. If it happens that bzlmod is not sufficient solution we can revisit this.

My understanding of `bzlmod` is that it may address issues for problems in rules, not Bazel itself. I feel these changes are allowing rule authors to solve for the issues in Bazel itself in an effective way. I agree with the desire to keep rules clean just as much as I agree with the desire for there to be no bugs. But because the latter is impossible, I think it requires that rules and macros have a few extra lines of code to sufficiently handle or warn about issues. 

If you feel strongly that these changes should still not be merged (which would be disappointing 😞), I would say the linked issue (https://github.com/bazelbuild/bazel/issues/8305) should also be closed.

comius:
What about adding a call `native.check_bazel_version(compatible_versions: str, message: str)` that fails when the Bazel version doesn't match what's in `compatible_versions`? 

`compatible_versions` might be define the same way as in https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit#

Such a call would support the use case you mention and it would potentially prevent a lot of problems that could be introduced with `native.bazel_version` kind of checks.

UebelAndre:
> What about adding a call `native.check_bazel_version(compatible_versions: str, message: str)` that fails when the Bazel version doesn't match what's in `compatible_versions`?
> 
> `compatible_versions` might be define the same way as in https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit#
> 
> Such a call would support the use case you mention and it would potentially prevent a lot of problems that could be introduced with `native.bazel_version` kind of checks.

I think I like that. How would I go about implementing that? Again, the Bazel repo is pretty unfamiliar to me. Where do you think the functionality would live and where would the tests go?

comius:
I think the location for such implementation is the same as you have at the moment. For the testing StarlarkIntegrationTest might be a better location.

comius:
I believe we reached an agreement for an alternative implementation. As such I will close this PR.

PR #14512, opened 2022-01-04
[apple] support watchos_arm64 in toolchain

dflems:
The watch-only changes from #14439

### DIFF
--- a/tools/osx/crosstool/BUILD.toolchains
+++ b/tools/osx/crosstool/BUILD.toolchains
@@ -52,6 +52,10 @@ OSX_TOOLS_CONSTRAINTS = {
         "@platforms//os:ios",
         "@platforms//cpu:x86_64",
     ],
+    "watchos_arm64": [
+        "@platforms//os:ios",
+        "@platforms//cpu:aarch64",
+    ],
     "watchos_arm64_32": [
         "@platforms//os:ios",
         "@platforms//cpu:aarch64",
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -76,6 +76,8 @@ def _impl(ctx):
         target_system_name = "x86_64-apple-ios"
     elif (ctx.attr.cpu == "ios_sim_arm64"):
         target_system_name = "arm64-apple-ios-simulator"
+    elif (ctx.attr.cpu == "watchos_arm64"):
+        target_system_name = "arm64-apple-watchos-simulator"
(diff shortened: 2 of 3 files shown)


tools/osx/crosstool/BUILD.toolchains
```diff
@@ -52,6 +52,10 @@ OSX_TOOLS_CONSTRAINTS = {
         "@platforms//os:ios",
         "@platforms//cpu:x86_64",
     ],
+    "watchos_arm64": [
+        "@platforms//os:ios",
+        "@platforms//cpu:aarch64",
+    ],
```
dflems, line 58: not sure if this will conflict with the one below or if there's a different CPU that should be used for one or the other?

dflems:
cc @Wyverald @kaylathar I broke this out of that other PR (#14439). The `watchos_arm64` CPU already existed. This PR doesn't add any new ones, just integration with the default osx toolchain.

kaylathar:
This LGTM - no concerns

brentleyjones:
@oquenchil or @Wyverald can you import?

tools/osx/crosstool/BUILD.toolchains
aiuto: It looks like it may in the future. If the constraints are the same, toolchain resolution would pick this one first.
But, what is up with the name  watchos_arm64_*32*? 
That entry is strange with a name that implies 32 bits, while using aarch64.

aiuto (commented):
Not really a review. I don't have enough knowledge of the Apple naming conventions.

tools/osx/crosstool/BUILD.toolchains
brentleyjones: > arm64_32 is a variant of arm64 with 32-bit pointer sizes, used on Apple Watch Series 4 and later.

Wyverald:
I'll do the import.

allevato (approved):

Wyverald:
@katre asks:

> Not @platforms//os:watchos?

> I don't know why they are all set like this, but this feels like a misconfiguration to me.

brentleyjones:
I don't think this change should change those, as that's a much larger and riskier change for this current addition.

katre:
Someone should investigate these, but I don't know enough about the difference between the ios variants to know what's correct.

tools/osx/crosstool/BUILD.toolchains
dflems: I wasn't sure about these. But `ios_sim_arm64` and `ios_arm64` are identical here too, so if it's wrong what's in here already is wrong.

Issue #14500, opened 2021-12-30
Manifest-based runfiles lookup is non-hermetic


fmeum:
### Description of the problem / feature request:

The Bazel runfiles libraries behave non-hermetically with `--enable_runfiles=false`.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Clone https://github.com/fmeum/bazel-non-hermetic-runfiles
2. Run `bazel run //parent:Parent` and observe that the runfiles lookup succeeds.
3. Comment out the `data` dependency of `//child`.
4. Run `bazel build //child`.
5. Revert the change made in step 3.
6. *Do not run* `bazel build //child`.
7. Run `bazel run //parent:Parent` and observe that the runfiles lookup fails.
### What operating system are you running Bazel on?

Linux, but applies to all OSes. This is particularly important on Windows where `--enable_runfiles` defaults to `false`.

### What's the output of `bazel info release`?

5.0.0rc3

###  Have you found anything relevant by searching the web?

The change that introduced the regression is https://github.com/bazelbuild/bazel/commit/bfdfa6ebfd21b388f1c91f512291c848e1a92a96. The same behavior can't be reproduced with Bazel 4.2.2, since the runfiles env variables are checked first there before the runfiles manifest is looked up next to the binary.

The root cause of the issue is that the first manifest-based lookup returns a path in the Bazel cache. But the contents of the cache are not hermetic and may contain outdated runfiles manifests for non-top-level binaries. If any of these binaries is invoked and first searches for the manifest next to itself instead of using the manifest set in `RUNFILES_MANIFEST_FILE`, it will behave non-hermetically.

The same issue can be reproduced with `--enable_runfiles=true` if for some reason the parent process resolves the runfiles tree symlink to the child process before launching it.

The process described in the [original design doc](https://docs.google.com/document/d/e/2PACX-1vSDIrFnFvEYhKsCMdGdD40wZRBX3m3aZ5HhVj4CtHPmiXKDCxioTUbYsDydjKtFDAzER5eg7OjJWs3V/pub) is also affected. I do not know whether it has been updated after the breaking commit or has always been buggy in this way.

### Any other information, logs, or outputs that you want to share?

The output of the reproducer with Bazel 4.2.2:
```
Runfiles variables in parent:
[...]

PR #14515, opened 2022-01-05
[5.x] Automated rollback of commit bfdfa6e.

meteorcloudy:
*** Reason for rollback ***

Due to https://github.com/bazelbuild/bazel/issues/14500

*** Original change description ***

Update find runfiles manifest file logic

This pull request updates the `FindManifestFileImpl` function to look for the runfiles manifest file of the binary first. If it is not found, it can use the manifest file specified in the environment variables `RUNFILES_MANIFEST_FILE` if it exists. The same logic is applied in `runfiles_src.cc` to find runfiles manifest file / directory of cc binaries.

This is done to avoid using the manifest file...

***

PiperOrigin-RevId: 419548921


meteorcloudy:
ping @Wyverald

PR #14540, opened 2022-01-10
Starlark: Add dict union operators | and |=

fmeum:
Implements the Starlark spec addition of https://github.com/bazelbuild/starlark/pull/215.

Work towards #12457

### DIFF
--- a/src/main/java/net/starlark/java/eval/Dict.java
+++ b/src/main/java/net/starlark/java/eval/Dict.java
@@ -72,7 +72,7 @@ import net.starlark.java.annot.StarlarkMethod;
             + "0 in d, \"a\" in d  # (True, False)\n"
             + "[(k, v) for k, v in d.items()]  # [(0, \"a\"), (1, 1), (2, \"b\")]\n"
             + "</pre>\n"
-            + "<p>There are three ways to construct a dictionary:\n"
+            + "<p>There are four ways to construct a dictionary:\n"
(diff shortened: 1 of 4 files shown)


aiuto:
Is there an issue for this? 
It seems like we shouljd have discussed it first, so the other Starlark implementations can more easily stay in sync.

fmeum:
@aiuto Sorry for the confusion, I created this as a draft PR specifically because I first wanted to ask about the proper procedure at https://github.com/bazelbuild/starlark/issues/14. I found that issue to convey positive signal towards whether this should be implemented, so I thought that having one concrete implementation including test cases would aid the final discussion. If you prefer to keep the open PR count low, I can close the PR until I have gotten a reply over at https://github.com/bazelbuild/starlark/issues/14.

aiuto:
No need to close the issue. But let's finish the discussion in the original FR bug.  So, maybe just convert to draft, so it stays around.

fmeum:
> No need to close the issue. But let's finish the discussion in the original FR bug. So, maybe just convert to draft, so it stays around.

It's been a draft all along, but I agree that we should continue the discussion on the other repo.

fmeum:
The spec has been updated, so this is now ready for review.

laurentlb (approved):
lgtm

(import is pending @brandjon or @tetromino 's approval)

Issue #14395, opened 2021-12-08
Bazel 4.2.2: C fails to link on macOS x86_64


uhthomas:
### Description of the problem / feature request:

C fails to link on macOS x86_64. This works fine on Bazel 4.2.1 and so I would consider this a regression.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I have a repository which contains a reproduction for the issue.

https://github.com/uhthomas/bazel-issue-14395

```
git clone https://github.com/uhthomas/bazel-issue-14395
cd bazel-issue-14395
USE_BAZEL_VERSION=4.2.2 bazel build //:bin
```

The project builds okay with Bazel 4.2.1.

```
USE_BAZEL_VERSION=4.2.1 bazel build //:bin
```

### What operating system are you running Bazel on?

```sh
❯ uname -mrs
Darwin 20.6.0 x86_64
```

### What's the output of `bazel info release`?

```sh
❯ USE_BAZEL_VERSION=4.2.2 bazel info release
release 4.2.2
```

###  Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

```sh
❯ USE_BAZEL_VERSION=4.2.2 bazel build //:bin
INFO: Analyzed target //:bin (46 packages loaded, 8076 targets configured).
INFO: Found 1 target...
ERROR: /private/var/tmp/_bazel_thomas/9c861fc6d98a2c5a9cf5eae5dc642b60/external/raze__memchr__2_4_1/BUILD.bazel:40:19: Compiling Rust bin memchr_build_script_ v2.4.1 (35 files) failed: (Exit 1): process_wrapper failed: error executing command bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/rules_rust/util/process_wrapper/process_wrapper --subst 'pwd=${pwd}' -- external/rust_darwin_x86_64/bin/rustc external/raze__memchr__2_4_1/build.rs ... (remaining 23 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
error: linking with `external/local_config_cc/cc_wrapper.sh` failed: exit status: 1
  |
[...]

PR #14542, opened 2022-01-11
Remove uses of -lstdc++ on darwin

keith:
This flag has been invalid seemingly since macOS 10.9 from 2013. By
default clang remaps this back to -lc++, but if your linkopts pass
-nodefaultlibs this is not the case, in which case this invalid flag
fails the link. This happened for rust builds where the rust driver
passes this flag intentionally.

For more info see https://github.com/bazelbuild/bazel/issues/14395#issuecomment-1009490593

Fixes https://github.com/google/cargo-raze/issues/247 https://github.com/bazelbuild/bazel/issues/14395 https://github.com/bazelbuild/rules_rust/issues/226

### DIFF
--- a/tools/cpp/cc_toolchain_config.bzl
+++ b/tools/cpp/cc_toolchain_config.bzl
@@ -395,7 +395,7 @@ def _impl(ctx):
                     flag_groups = [
                         flag_group(
                             flags = [
-                                "-lstdc++",
+                                "-lc++",
                                 "-undefined",
                                 "dynamic_lookup",
                                 "-headerpad_max_install_names",
--- a/tools/cpp/unix_cc_configure.bzl
+++ b/tools/cpp/unix_cc_configure.bzl
@@ -400,7 +400,7 @@ def configure_unix_toolchain(repository_ctx, cpu_value, overriden_tools):
         False,
     ), ":")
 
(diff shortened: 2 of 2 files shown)


uhthomas:
Does this also fix https://github.com/bazelbuild/rules_rust/issues/978?

keith:
I don't think so, I think we'd also have to change https://github.com/bazelbuild/bazel/blob/26d4aadaf4006efd6ff20004aa1430e97263b30c/tools/cpp/cc_toolchain_config.bzl#L374 and expand my new conditional to match that. Can you trace back the history of libstdc++ vs libc++ to verify that is a safe change? Either way I guess we can do that in a separate PR just in case.

oquenchil (approved):
LGTM

uhthomas:
> I guess we can do that in a separate PR just in case.

I agree, that sounds like the best thing to do.

For now, is it possible to get this change merged? We're essentially stuck on Bazel 4.2.1 until then.

brentleyjones:
@Wyverald This would be nice to get into 5.0.1.

keith:
I submitted https://github.com/bazelbuild/bazel/pull/14668 for the bsds

Wyverald:
@bazel-io fork 5.1

PR #14558, opened 2022-01-12
run buildifier over the repo

ptarjan:
As a first time contributor I was surprised there wasn't a linter which just automatically enforced my build files were formatted. Is there any reason why we don't enforce this?

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -485,7 +485,6 @@ dist_http_archive(
     build_file = "@local_jdk//:BUILD.bazel",
     patch_cmds = EXPORT_WORKSPACE_IN_BUILD_BAZEL_FILE,
     patch_cmds_win = EXPORT_WORKSPACE_IN_BUILD_BAZEL_FILE_WIN,
-
 )
 
 dist_http_archive(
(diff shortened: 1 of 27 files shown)


sventiffe (approved):

sventiffe:
Actually I've learned the hard way that the approval has been pre-mature. We require PRs to be either only affecting src/ or third_party/ so the PR needs eventually to be split. Also, I've had some actual trouble importing some changes.

But I agree that it makes sense to run buildifier on our own BUILD files. Let's see how we get the changes in with minimal effort.

ptarjan:
@sventiffe Thanks for taking a look! I can run this PR as the core one and submit another one for third_party/. Would that help?

ptarjan:
https://github.com/bazelbuild/bazel/pull/14596 is the third-party one

ptarjan:
Great! Thank you for taking a look at this. Do let me know if I can help somewhere, but I'm sure you can run `buildifier -r .` yourself :)

sventiffe:
Importing these changes will take a bit as some formatting issues actually result from the export (and automatic modification) of some files from Google to the GitHub repo. This is looks like a good opportunity to look for improvements on that end as well.  

Mentioning this to manage expectations why merging these trivial changes takes more time than you would expect if this was only the repo here on GitHub.

ptarjan:
No worries. Thank you. I managed a similar process at Facebook for many
years so understand the complexities. Good luck with the improvements. I’m
totally fine if you instead run buildifier internally and then see if that
gets closer to a clean repo externally.

cushon:
> This first export that picked up the change is https://github.com/bazelbuild/bazel/commit/59f8e9bfc4fe2d4dfc32e79c0fb22b2b486dd01c. There are still a couple of formatting issues in `src/main/starlark/builtins_bzl/` that I'm working on a follow-up for.

The follow-up config change was picked up here: https://github.com/bazelbuild/bazel/commit/43c7b97e3d1889a57fadbc3fca2e633c12f57d75. Everything outside of `third_party/` should be fixed now.

cushon:
I forgot to follow up here, but the `third_party/` changes were applied as 801bb4b13ccad94af63f1cc4632329ad4674fb64

PR #14581, opened 2022-01-14
Fix build after rc4 cherrypicks

Wyverald:


Wyverald:
cc @meteorcloudy

meteorcloudy:
Nice!

PR #14595, opened 2022-01-18
Don't resolve symlinks for --sandbox_base

ob:
On macOS BigSur, the sandbox-exec command behaves slightly different than on
Catalina when firm links are present.

Resolving symlinks can prevent the sandbox for allowing write operations to the
sandbox base.

This effectively reverts a piece of 656a0ba, namely:

>  When using --experimental_sandbox_base, ensure that symlinks in the path are
>  resolved. Before this, you had to check whether on your system /dev/shm is a
>  symlink to /run/shm and then use that instead. Now it no longer matters, as
>  symlinks are resolved.

See https://github.com/bazelbuild/bazel/issues/13766 for full details.

Closes #13984.

PiperOrigin-RevId: 422319807
(cherry picked from commit 0de7bb95022057e8b89334f44759cf6f950e131f)


ob:
This is a small localized fix that would be worth including in 5.0 as it affects macOS builds when firm links are set up in a particular way. See #13766 for all the details. CC @Wyverald

PR #14600, opened 2022-01-19
Add the runfiles directory to the runtime library search paths

EdSchouten:
When executing a C++ binary through REv2 directly (e.g., directly as a
genrule, not as a dependency of another script/executable), the input
root will only contain a \_solib\_arch directory inside the executable's
runfiles directory. This means that library lookups also need to
consider rpath $ORIGIN/$executable.runfiles/$workspace/\_solib\_$arch/.
Only looking at ../../../\_solib\_$arch/ is not sufficient.

This issue isn't noticeable when doing local execution, for the reason
that rtld/dyld will first do a realpath() on the executable, meaning the
lookup of shared objects takes place outside of the sandbox, inside of
bazel-out/ directly.

This change extends LibrariesToLinkCollector to duplicate all runtime
library search paths. One for bare execution, and one for execution
inside a runfiles directory.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
@@ -17,6 +17,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.analysis.RuleErrorConsumer;
 import com.google.devtools.build.lib.cmdline.LabelConstants;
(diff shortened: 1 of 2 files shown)


EdSchouten:
@oquenchil PTAL!

oquenchil (approved):
LGTM. Could you please verify there aren't merge conflicts? Then I will merge.

EdSchouten:
Thanks, @oquenchil! I've just addressed the merged conflicts. Please take another look.

oquenchil (approved):
LGTM

Wyverald:
No action items for you, @EdSchouten. Our team will merge this. Thanks!

oquenchil:
This causes targets to fail internally when there is a `,` in the target name. 

```
ld: error: cannot open SOME_NAME.runfiles/workspace_name/_solib_k8/: No such file or directory
ld: error: cannot open  SOME_NAME.runfiles/workspace_name/some_name-compiler-k8-llvm/: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Could you please take a look and make sure it takes that into account?

oquenchil:
On the crosstool? Would that break anyone?

fmeum:
Yes, that would be set on the toolchain. GCC and clang support both flags. I know that there are compilers that support `-Xlinker` but not `-Wl`. I can't rule out that there is a compiler that only supports `-Wl` though.

oquenchil:
I'm not aware either, so quite possibly out of ignorance no compatibility concerns.

Issue #14503, opened 2022-01-02
Documentation issue: Label - What happens when label sections are blank?


psigen:
Documentation URL: https://docs.bazel.build/versions/main/skylark/lib/Label.html

Would it be possible to add documentation around what happens with this API when portions of a label are empty?

Here are some examples I was having trouble with:
- if no package is specified, i.e. `//:abc`, does `package` return `"."` or `""` or `None`?
- if the name is not specified, i.e. `//pkg`, does it automatically set `name` to `"pkg"`?
- if the workspace is not specified, i.e. `//pkg:abc`, does `workspace_name` automatically resolve to `""`  or `"__main__"` or dynamically resolve to the workspace name specified in `WORKSPACE`?
- if the workspace is not specified, i.e. `//pkg:abc`, does `workspace_root` get set to `"."` or `None` or some other value?

Wyverald:
A number of documentation improvements in this area have been made to address some of the concerns raised here. Other questions like `if the name is not specified, i.e. //pkg, does it automatically set name to "pkg"?` can be answered by simply reading the docs (in this case, `//pkg` behaves exactly the same as `//pkg:pkg`). So I'll go ahead and close this.

PR #14609, opened 2022-01-20
Do not mention redundant use case for Label constructor

fmeum:
Label strings passed to the `default` argument of `attr.label()` are resolved to a `Label` immediately relative to the call site of
`attr.label()`, so wrapping the argument with `Label(...)` is entirely redundant.

More context at:
https://github.com/bazelbuild/bazel-central-registry/pull/63#issuecomment-1017669585

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/StarlarkRuleFunctionsApi.java
+++ b/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/StarlarkRuleFunctionsApi.java
@@ -579,12 +579,11 @@ public interface StarlarkRuleFunctionsApi<FileApiT extends FileApi> {
   @StarlarkMethod(
       name = "Label",
       doc =
-          "Creates a Label referring to a BUILD target. Use this function when you want to give a"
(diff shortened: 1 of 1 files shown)


fmeum:
@Wyverald This part of the docs has always confused me. Maybe you could give this particular page a read and verify that everything is up-to-date now that `Label` will experience its rise to fame with bzlmod.

Wyverald:
Thanks, this looks good -- but yeah I should really rewrite this maybe. It took myself a long time to understand all the intricacies of label semantics.

fmeum:
> Thanks, this looks good -- but yeah I should really rewrite this maybe. It took myself a long time to understand all the intricacies of label semantics.

This might also solve https://github.com/bazelbuild/bazel/issues/14503 in one go.

fmeum:
@Wyverald Do you want to open an issue about updating the docs for `Label`? I would then close this PR, which I guess only covers a very small part of the required overhaul.

Wyverald:
Filed https://github.com/bazelbuild/bazel/issues/15821. Thanks!

PR #14623, opened 2022-01-22
Support workers in CompatDexBuilder

Bencodes:
This also fixes the `--use_workers_with_dexbuilder` flag which would pass the worker execution requirements to `CompatDexBuilder` without it knowing how to handle the work requests.

Using a worker here should help with some OOMs issues, and drastically improves build performance. We saw a 29% improvement in clean build times and a 14% improvement in incremental build times where the ABI was invalidated.

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/r8/BUILD
+++ b/src/tools/android/java/com/google/devtools/build/android/r8/BUILD
@@ -41,6 +41,7 @@ java_library(
     }),
     visibility = ["//src/test/java/com/google/devtools/build/android/r8:__pkg__"],
     deps = [
+        "//src/main/java/com/google/devtools/build/lib/worker:work_request_handlers",
         "//src/main/java/com/google/devtools/common/options",
         "//src/tools/android/java/com/google/devtools/build/android:android_builder_lib",
         "//third_party:asm",
--- a/src/tools/android/java/com/google/devtools/build/android/r8/CompatDexBuilder.java
+++ b/src/tools/android/java/com/google/devtools/build/android/r8/CompatDexBuilder.java
@@ -26,14 +26,23 @@ import com.android.tools.r8.DexIndexedConsumer;
 import com.android.tools.r8.DiagnosticsHandler;
 import com.android.tools.r8.origin.ArchiveEntryOrigin;
 import com.android.tools.r8.origin.PathOrigin;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.io.ByteStreams;
+import com.google.devtools.build.lib.worker.ProtoWorkerMessageProcessor;
+import com.google.devtools.build.lib.worker.WorkRequestHandler;
 import com.google.devtools.common.options.OptionsParsingException;
+
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PrintStream;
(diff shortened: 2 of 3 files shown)


larsrc-google:
This looks good from a worker perspective. At a quick glance, CompatDexBuilder is thread-safe, which is good. It looks like it never truly supported workers, though, so maybe release the worker support first to weed out bugs in that before adding the complexity of multiplex workers?

Bencodes:
> so maybe release the worker support first to weed out bugs in that before adding the complexity of multiplex workers?

I can make that change if you think that's the safest path forward. We've been using a version of this internally for quite some time and have seen no issues so far (both worker and multiplex worker impls).

Bencodes:
Reverted the multiplex changes out of this PR and will follow up in a second PR with those additional changes.

larsrc-google:
How does the CompatDexBuilder normally output warnings and errors for the user to see?

Bencodes:
`D8#run` (the tool that `CompatDexBuilder` calls into) dump warnings and errors directly into the standard system streams by way of `DiagnosticsHandler` https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/DiagnosticsHandler.java.

Bencodes:
We can provide out own implementation of `DiagnosticsHandler` that gets passed into `D8Command#builder` and redirect that info directly into the `PrintWriter` that the work request provides.

Bencodes:
Pulled some benchmarks for this worker changes (includes multiplex). We saw a 29% improvement in clean build times and a 14% improvement in incremental build times where the ABI was invalidated.

Bencodes:
@larsrc-google are there next steps here for getting in?

src/tools/android/java/com/google/devtools/build/android/r8/CompatDexBuilder.java
```diff
@@ -77,11 +85,60 @@ public void finished(DiagnosticsHandler handler) {
...
+    } catch (OptionsParsingException e) {
+      pw.println("CompatDexBuilder raised OptionsParsingException: " + e.getMessage());
+      return 1;
+    } catch (IOException | InterruptedException | ExecutionException e) {
+      e.printStackTrace();
+      return 1;
+    } finally {
+      // Write the captured buffer to the work response
+      synchronized (buf) {
+        String captured = buf.toString().trim();
```
larsrc-google, line 133: You should apply a charset (UTF_8) to the toString() method.

src/tools/android/java/com/google/devtools/build/android/r8/BUILD
```diff
@@ -41,6 +41,8 @@ java_library(
     }),
     visibility = ["//src/test/java/com/google/devtools/build/android/r8:__pkg__"],
     deps = [
+        "//src/main/java/com/google/devtools/build/lib/worker:work_request_handlers",
+        "//src/main/java/com/google/devtools/build/lib/worker:worker",
```
larsrc-google, line 45: Do you actually need to depend on `:worker`? That should only be for internal (server-side) use.

larsrc-google (approved):

ted-xie:
Hi Ben,

Thanks for this PR. We're in the process of importing it upstream now. We've discovered that this PR actually requires a manual update to the remote Android tools repo, and then a Bazel release cycle, so we'll need a little longer before we can close this out.

PR #14634, opened 2022-01-24
Remote: Only waits for background tasks from remote execution.

### LINKED ISSUE #14620: `--bes_upload_mode=fully_async` is ignored in Bazel 5

Running a build with the `--bes_upload_mode=fully_async` doesn't have the expected effect. The build events are still uploaded synchronously after the command is run. This seems to happen when remote cache is used.

### Bugs: what's the simplest, easiest way to reproduce this bug? 

```
git clone git@github.com:bazelbuild/rules_java.git && cd rules_java
```
And then run at least twice so the Bazel server is running and the cache is populated:
```
env USE_BAZEL_VERSION=5.0.0 bazelisk build --remote_cache=grpcs://remote.buildbuddy.io --bes_backend=grpcs://remote.buildbuddy.io //java:srcs --bes_upload_mode=fully_async
```

After the build is complete it still takes a bit of time to do something, which I believe is uploading the BEP events. Changing `USE_BAZEL_VERSION` to 4.2.0 makes it snappier. This upload time is more noticeable in short builds that run more targets.

macOS 12.0.1

`release 5.0.0`

This seems similar to https://github.com/bazelbuild/bazel/issues/14576.

coeuvre:
We added the block waiting behaviour after each command in remote module to wait for background uploads when introducing async upload. However, not all background uploads should be waited, e.g. uploads from BES module but with flag `--bes_upload_mode=fully_async`.

This PR updates remote module so that only uploads initiated by remote module are waited after the command. This also enable us to implement something like `--remote_upload_mode=fully_async` in the future.

Fixes #14620.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
@@ -142,6 +142,7 @@ import java.util.SortedMap;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Phaser;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


brentleyjones:
> This also enable us to implement something like --remote_upload_mode=fully_async in the future.

❤️

meisterT (approved):

brentleyjones:
@Wyverald This fixes a 5.0 regression. Can we add it to the 5.0.1 milestone?

Wyverald:
(for the record, the linked issue is tracked on the 5.1 milestone)

Issue #14393, opened 2021-12-08
Stack Overflow during build due to inability to remove _virtual_includes dir


Warchant:
### Description of the problem / feature request:

Stack overflow: 
![image](https://user-images.githubusercontent.com/1867551/145259219-a8f66bef-d7c8-4dff-bd4e-b49b4b73e49e.png)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I don't have reproducer, and I don't know exact reason why this happens, but it has to be related to OS being unable to remove one of `_virtual_includes` folders under `execution root`.

### What operating system are you running Bazel on?

Win11.

### What's the output of `bazel info release`?

release 4.1.0

###  Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

```
$ bazel clean --expunge
INFO: Starting clean.
ERROR: C:/users/$user/_bazel_$user/hrr5qgf6/execroot/prj/bazel-out/cpu-opt/bin/path/to/dir/_virtual_includes/lib_h (Permission denied)
```

meteorcloudy:
On Windows, you cannot delete files which are opened by other process. Can you check is other running process are holding the file handle of the directory?

Warchant:
@meteorcloudy I am sure that other processes opened those files. Bazel should not die with Stack Overflow because of that.

meteorcloudy:
Yes, normally it should exit with an error message like
`ERROR: C:/users/$user/_bazel_$user/hrr5qgf6/execroot/prj/bazel-out/cpu-opt/bin/path/to/dir/_virtual_includes/lib_h (Permission denied)`

Do you have a minimal reproduce case that we could use to debug the crash?

Warchant:
@meteorcloudy sorry, no :( but you have stacktrace, that should be enough to find the root.

meteorcloudy:
Sorry, without a reproducible case, we don't know what caused the stackoverflow error. The code is here: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java;l=466-470

Warchant:
@meteorcloudy `deleteOutput` calls itself recursively. If `path.delete()` fails over and over (when process holds a file), then it will call itself recursively until StackOverflow is thrown.

```diff
public static void deleteOutput(Path path, @Nullable Root root) throws IOException {
    try {
[...]

PR #14637, opened 2022-01-25
Fix stack overflow exception in deleteOutput

Warchant:
Fixes https://github.com/bazelbuild/bazel/issues/14393

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java
@@ -494,13 +494,14 @@ public abstract class AbstractAction extends ActionKeyCacher implements Action, 
 
       Path parentDir = path.getParentDirectory();
       if (!parentDir.isWritable() && root.contains(parentDir)) {
-        // Retry deleting after making the parent writable.
         parentDir.setWritable(true);
-        deleteOutput(path, root);
-      } else if (path.isDirectory(Symlinks.NOFOLLOW)) {
+      }
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java
```diff
@@ -497,6 +497,12 @@ public static void deleteOutput(Path path, @Nullable Root root) throws IOExcepti
         // Retry deleting after making the parent writable.
         parentDir.setWritable(true);
         deleteOutput(path, root);
```
meteorcloudy, line 499: This recursion call here is really confusing and unneccessary. I think we can just replace it with:
```
if (path.isDirectory(Symlinks.NOFOLLOW)) {
  path.deleteTree();
} else {
  path.delete();
}
```

Then we don't need the special check here.

meteorcloudy (changes requested):
Thanks!

src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java
Warchant: Done

Warchant:
Is PR merged or just closed?

meteorcloudy:
Merged at https://github.com/bazelbuild/bazel/commit/1746d6341d22278a615a3707f2ab9e5a9d83df0b!

meteorcloudy:
We have to import this change to internal code base, then export to github.

Issue #14316, opened 2021-11-24
`extractArgumentsForDynamicLinkParamFile` causes arguments that start with `@rpath` to be interpreted as a param file


brentleyjones:
### Description of the problem / feature request:

While updating rules_apple we couldn't use `@rpath` arguments as it would result in a build failure. Instead we had to pass the arguments in a different way:

https://github.com/bazelbuild/rules_apple/pull/1207/files#diff-d84498974951b9a579baeeaf19d52ffb0a42a15b520c4359dac92448e2b806ceR670

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This code needs to be updated to support arguments that start with `@rpath`:

https://github.com/bazelbuild/bazel/blob/1c3a2456c95fd19974a5b2bd33c5ebdb2b2277e4/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkCommandLine.java#L295

### What operating system are you running Bazel on?

macOS 11.6.1

### What's the output of `bazel info release`?

5.0.0rc2

keith:
FWIW this was a bit of an intentional tradeoff when we added params files support. In the shell scripts we check if the file exists, and assume that if it doesn't it's an intentional arg otherwise, so we could do the same here to fix it probably.

sventiffe:
So is this working as intended?

keith:
We should fix anywhere we can easily, but most of the time the recommendation is to change whoever is passing the flag to use the `-Wl,@rpath` format to avoid this.

keith:
https://github.com/bazelbuild/bazel/pull/14650

brentleyjones:
@bazel-io fork 5.1

PR #14650, opened 2022-01-26
Fix aggressive params file assumption

keith:
Most programs that accept params files use the `@file` syntax. For Apple
platform builds `@` can be the start of non-params file arguments as
well, such as `-rpath @executable_path/Frameworks`. There is a small
list of options where this is the case, so this new behavior no longer
assumes params files if args start with `@`, they also have to not start
with one of the 3 keywords used with this (from `man dyld` on macOS).
This should always hold since params files generated by bazel should
always start with `bazel-out`, if someone renames the symlinks to one of
the keywords, they're on their own.

Previously the workaround was to always make sure to pass the
`-Wl,-rpath,@executable_path` form of these arguments, but this makes
users not have to worry about this.

In a few other places we check this by checking if the file exists,
which is likely more accurate, but feels excessive and potentially
dangerous in this context.

Related: https://github.com/bazelbuild/bazel/pull/13148
Fixes: https://github.com/bazelbuild/bazel/issues/14316

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkCommandLine.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkCommandLine.java
@@ -270,13 +270,13 @@ public final class LinkCommandLine extends CommandLine {
         paramFile, linkTargetType, forcedToolPath, featureConfiguration, actionName, variables);
   }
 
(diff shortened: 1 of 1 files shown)


oquenchil (approved):
LGTM

PR #14651, opened 2022-01-26
Fix missing warning on macOS

keith:
If you're building on macOS you're likely using clang, and
`-Wno-maybe-uninitialized` isn't a valid warning with Apple clang


gregestren:
Assigned to @coeuvre because of the package's commit history.

coeuvre (approved):
@meteorcloudy Please help import.

PR #14660, opened 2022-01-28
Normalize rpath entries to guard against missing default solib dir

fmeum:
When all dynamic deps in a build are built in transitioned configurations, the default solib dir is not created. However, while
resolving paths, the dynamic linker stops at the first directory that does not exist, even when followed by `../`.

Before this commit, all rpath entries would consist of the relative path to the default solib dir followed by the relative path to the
particular library's solib dir. Thus, if the default solib dir was missing, the dynamic linker wouldn't resolve any of these paths.

This commit ensures that the relative path entries are normalized and thus contain no references to non-existing directories assuming the normalized path itself exists.

Work towards #13819.


oquenchil (approved):
LGTM

fmeum:
@oquenchil Friendly ping. Is this simply waiting to be imported or does it perhaps fail some internal tests?

PR #14665, opened 2022-01-28
Fix `ctx.fragments.apple.single_arch_cpu` returning incorrect cpu for tools when host cpu and exec cpu are different

thii:


gregestren:
@keith @brentleyjones are you comfortable reviewing PRs like this (especially given pending OWNERS updates)? Or do you recommend elsewhere?

keith (approved):

keith:
Yes, also in this case @katre since you investigated this issue as well maybe you have some thoughts?

katre (approved):
Looks good, I'll import this and get it in front of the internal experts today.

PR #14666, opened 2022-01-29
fix: Add missing commas

mrshu:
* Add missing commas to `ctexplain` text


gregestren (approved):
Out of curiosity, what's your interest in this code?

mrshu:
None at all to be honest -- it was just a potential bug (albeit in this case it almost certainly is not) :slightly_smiling_face:

gregestren:
Ah, thanks. I was curious because I wrote it and it's only partially checked in. So I'd need to support getting the other parts checked in to make it useful to anyone who cared.

Thanks for the fix, still.

PR #14668, opened 2022-01-29
Fix uses of std++ on bsd

keith:
This fix is the same as https://github.com/bazelbuild/bazel/pull/14542
but for freebsd and openbsd

Hopefully fixes https://github.com/bazelbuild/rules_rust/issues/978


keith:
@uhthomas can you test this change to see if it fixes https://github.com/bazelbuild/rules_rust/issues/978 for you?

oquenchil (approved):
LGTM

brentleyjones:
@bazel-io fork 5.1

PR #14676, opened 2022-01-31
Add the ability to emit *.runfiles_manifest as JSON

EdSchouten:
The *.runfiles_manifest files currently use a pretty simple format: two
pathnames separated by a space character. This can be problematic at
times. On Windows, usernames may contain spaces, meaning home
directories with spaces are not uncommon. If you look at the macOS
ecosystem, it's also not uncommon to have applications or packages
containing files with spaces in their names. The goal of this change is
to start addressing this issue, by moving *.runfiles_manifest to newline
delimited JSON.

Runfiles libraries for each programming language will need to be
adjusted to support this new format. Because not all runfiles libraries
are bundled/versioned together with Bazel, those libraries will need to
seamlessly support both formats. Tuples are encoded as lists, JSON
entries will always start with an opening square bracket. Because paths
are generally prefixed with a workspace name, it's extremely unlikely
that this causes any conflict. Runfiles libraries can thus distinguish
both formats as follows:

    if (line[0] == '[') {
      // Parse as JSON.
    } else {
      // Parse as legacy format.
    }

To prevent any breakage, we place this feature behind a flag named
--incompatible_json_source_manifests. Even if this flag is disabled, we
emit JSON entries for paths that are not representable in the legacy
format. This should at least make it possible to build targets and
instantiate their runfiles directories with build-runfiles, regardless
of whether the manifest can be parsed at runtime.

Before considering JSON, I experimented with CSV instead. Though easier
to parse, RFC 4180 requires the use of carriage return newlines. Because
both Bazel and runfiles libraries tend to open runfiles manifests in
text mode, this turned out to be hard to get right.

Given the fact that Bazel does not permit non-printable characters and
backslashes in pathnames, parsing the resulting JSON turns out to be
remarkably easy. If it's not realistic to let a given runfiles library
depend on a third-party library to do the JSON parsing, it's most likely
acceptable to handroll a simplified parser.

This PR already contains updates to build-runfiles, the Bash runfiles
library and the Python runfiles library to support the new format.
Updates to the C++ and Java runfiles libraries remain to be done.

Issue: #4327

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BUILD
@@ -446,6 +446,7 @@ java_library(
         "//third_party:auto_value",
         "//third_party:caffeine",
         "//third_party:flogger",
+        "//third_party:gson",
         "//third_party:guava",
         "//third_party:jsr305",
         "//third_party/protobuf:protobuf_java",
--- a/src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
@@ -348,7 +348,8 @@ public final class RunfilesSupport {
                 context.getActionOwner(),
                 inputManifest,
                 runfiles,
-                context.getConfiguration().remotableSourceManifestActions()));
+                context.getConfiguration().remotableSourceManifestActions(),
+                context.getConfiguration().jsonSourceManifests()));
 
     if (!createSymlinks) {
       // Just return the manifest if that's all the build calls for.
--- a/src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
@@ -16,6 +16,7 @@ package com.google.devtools.build.lib.analysis;
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.CharMatcher;
 import com.google.devtools.build.lib.actions.ActionExecutionContext;
 import com.google.devtools.build.lib.actions.ActionKeyContext;
 import com.google.devtools.build.lib.actions.ActionOwner;
@@ -28,6 +29,7 @@ import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
 import com.google.devtools.build.lib.events.EventHandler;
 import com.google.devtools.build.lib.util.Fingerprint;
 import com.google.devtools.build.lib.vfs.PathFragment;
+import com.google.gson.stream.JsonWriter;
 import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.OutputStream;
@@ -71,7 +73,7 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
      * @param symlink (optional) symlink that resolves the above path
      */
     void writeEntry(Writer manifestWriter, PathFragment rootRelativePath,
-        @Nullable Artifact symlink) throws IOException;
+        @Nullable Artifact symlink, boolean useJson) throws IOException;
 
     /**
      * Fulfills {@link com.google.devtools.build.lib.actions.AbstractAction#getMnemonic()}
@@ -101,6 +103,7 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
   private final Runfiles runfiles;
 
   private final boolean remotableSourceManifestActions;
+  private final boolean jsonSourceManifests;
 
   /**
    * Creates a new AbstractSourceManifestAction instance using latin1 encoding to write the manifest
@@ -114,7 +117,13 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
   @VisibleForTesting
   SourceManifestAction(
       ManifestWriter manifestWriter, ActionOwner owner, Artifact primaryOutput, Runfiles runfiles) {
-    this(manifestWriter, owner, primaryOutput, runfiles, /*remotableSourceManifestActions=*/ false);
+    this(
+        manifestWriter,
+        owner,
+        primaryOutput,
+        runfiles,
+        /*remotableSourceManifestActions=*/ false,
+        /*jsonSourceManifests=*/ false);
   }
 
   /**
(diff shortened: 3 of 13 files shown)


gregestren:
@EdSchouten do you have a recommended reviewer for this PR? Or at least team (https://bazel.build/maintaining/maintainers-guide.html#team-labels)?

I can route as make sense to me, but if we can route to anyone we already know is paying attention that offers a better chance of timely response.

EdSchouten:
Hey Greg! Thanks for responding!

> @EdSchouten do you have a recommended reviewer for this PR? Or at least team (https://bazel.build/maintaining/maintainers-guide.html#team-labels)?
> 
> I can route as make sense to me, but if we can route to anyone we already know is paying attention that offers a better chance of timely response.

I suspect team-Core would be the best fit for this PR, considering that runfiles handling is pretty core to Bazel. Happy to bring other teams in the loop when it comes to reviewing some of the runfiles libraries changes, but we can always do that later.

gregestren:
I believe `Core` is more focused on the core execution engine, i.e. Skyframe.

I assigned `team-Rules-Server` for lack of a better label. @lberki should actually be a good contact but he's on a temporary tech leave at the moment.

Generic runfiles I think suffers from weak ownership.

gregestren:
So basically, I'm trying to get this proper attention in accordance with https://bazel.build/maintaining/maintainers-guide.html#lifecycle-of-a-pull-request.

Ping me specifically if you worry about responsiveness, but as a heads up I'd expect this one to take some time.

EdSchouten:
Hey Greg,

Sorry for the late response; I missed the notification.

> Please note I'm not super-familiar with the runfiles implementation logic, so I have some pretty basic questions.
> 
> Is build-runfiles the logic that Bazel uses to construct a target's `foo.runfiles` directory, which it reads the `foo.runfiles_manifest` file as input?

Yes. But there are also cases in which Bazel constructs 'foo.runfiles' directories without going through 'build-runfiles'. It may construct the runfiles directory directly, based on its in-memory bookkeeping.
 
> How much of #4327 does this PR address? The issue is titled as runfiles-related, but I think it's actually a broader issue with spaced path names in general, of which runfiles is one major but not the only example? i.e. the failure of `genrule` `$(SRCS)` at [#4327 (comment)](https://github.com/bazelbuild/bazel/issues/4327#issuecomment-468274037), and maybe the `--output_user_root` issue?

It by no means addresses all issues related to whitespace in filenames. For my specific use case I want to drag in a large directory hierarchy with pre-generated contents into the runfiles directory of a tool of ours. My intent is not to call `genrule()` on any of those files. That said, those issues are also interesting and we should also work towards fixing those in the long run.

> Do you know what's going on at [#4327 (comment)](https://github.com/bazelbuild/bazel/issues/4327#issuecomment-960722697) - which is setting `--nobuild_runfile_links` but still failing with a runfiles creation error?

So if I understand it correctly, `bazel run` is broken, while `bazel test` does work.
- `bazel test ` works, because it needs to instantiate a runfiles directory in the sandbox context. Bazel is capable of doing that without using runfiles_manifest at all. It constructs the runfiles directory from its own in-memory bookkeeping.
- `bazel run` doesn't work, because it does call into build-runfiles to construct a runfiles directory in bazel-out/. The `--nobuild_runfile_links` flag prevents this from happening during `bazel build`, but if you do a `bazel run` it _must_ be done to satisfy the execution requirements of the tool.

> Aside from these specifics I'm curious how many logic holes exist throughout the code base w.r.t. spaces in paths, and how much harder of a problem that is vs. what I think is a subset of the problem being addressed here.

Yeah, there are likely many other things that are broken. That said, we don't necessarily need to support all these use cases. My hope is that we can at least support being able to `filegroup()` a directory structure containing spaces in names and adding those to `data =` of tools.

I hope that answers all your questions!

gregestren:
Hi @EdSchouten . @brandjon and I discussed this yesterday.

Would you be willing to output both a `runfiles_manifest` and `runfiles_manifest.json` file with the corresponding formats? That would offer format consistency: a file's format is fixed, period. It doesn't depend on conditions that the user may or may not understand.

The ultimate end game could be to remove the `runfiles_manifest` version, but that's of course a few more feature PRs down the road.

EdSchouten:
Hey Greg,

> Would you be willing to output both a `runfiles_manifest` and `runfiles_manifest.json` file with the corresponding formats? That would offer format consistency: a file's format is fixed, period. It doesn't depend on conditions that the user may or may not understand.

Just so I get this right: Does that mean that we should also add a foo.runfiles/MANIFEST.json?

I suppose we should also introduce a `RUNFILES_MANIFEST_JSON_FILE` environment variable for this, right? Or are you suggesting that code should always try to open `${RUNFILES_MANIFEST_FILE}.json` as well?

gregestren:
> Hey Greg,

> 
> Just so I get this right: Does that mean that we should also add a foo.runfiles/MANIFEST.ndjson?

Does your current patch adjust both files?

Are you preferring ndjson vs. full json because it's easier to implement, or you think it's cleaner, or both?

It's probably most consistent to update both `foo.runfiles_manifest` and `foo.runfiles/MANIFEST` atomically in the same PR.

> 
> I suppose we should also introduce a `RUNFILES_MANIFEST_NDJSON_FILE` environment variable for this, right? Or are you suggesting that code should always try to open `${RUNFILES_MANIFEST_FILE}.ndjson` as well?

I'd suggest the latter. I'm wary of reliance of environment variables for reproducibility reasons, but maybe I'm not appreciating this use case for them.

EdSchouten:
> > Hey Greg,
> 
> > Just so I get this right: Does that mean that we should also add a foo.runfiles/MANIFEST.ndjson?
> 
> Does your current patch adjust both files?

Yes!

> Are you preferring ndjson vs. full json because it's easier to implement, or you think it's cleaner, or both?

NDJSON has the advantage that it can be parsed line by line. Especially for stuff like runfiles.bash it's easier if we allow that. Allowing arbitrary JSON in there (which may contain newlines at arbitrary spots) can be tricky to deal with in those languages.

> It's probably most consistent to update both `foo.runfiles_manifest` and `foo.runfiles/MANIFEST` atomically in the same PR.

Yeah, I think they also have to be in sync. I seem to remember there are some places in the tree where runfiles_manifest and runfiles/MANIFEST are compared for equality. If they don't match, the runfiles directory is rebuilt. The runfiles/MANIFEST file is thus used as a marker to indicate whether the runfiles directory is up to date.

> > I suppose we should also introduce a `RUNFILES_MANIFEST_NDJSON_FILE` environment variable for this, right? Or are you suggesting that code should always try to open `${RUNFILES_MANIFEST_FILE}.ndjson` as well?
> 
> I'd suggest the latter. I'm wary of reliance of environment variables for reproducibility reasons, but maybe I'm not appreciating this use case for them.

Gotcha. Will try to get this PR updated next week. Thanks!

matthewjh:
Any update, @EdSchouten? Really appreciate the work you have put in here, it's going to help a lot of people!

keertk:

matthewjh:
@keertk  @gregestren how can we reopen this?

gregestren:
@matthewjh we just need ownership. Would you be willing to check around to see who might want to take this on, or if @EdSchouten wants to re-focus? Ask around on Slack,  https://github.com/bazelbuild/bazel/discussions, etc?

Issue #14436, opened 2021-12-15
To run coverage with python, setting PYTHON_COVERAGE in every test env is required


bradb423:
## Description of the problem / feature request:
Currently, to obtain coverage for projects in python, the location macro is used to point to the entry point of a python coverage tool. This has to be repeated for each test in a given project. Perhaps an environment variable could be created e.g: `PYTHON_COVERAGE_TARGET`, to make the process simpler.

## Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

For an example project with the following python files:

```python
import sys


def main() -> int:
    return 0


if __name__ == "__main__":
    sys.exit(main())

```

```python
import unittest

from main import main


class TestMain(unittest.TestCase):
    def test_main(self):
        self.assertEqual(main(), 0)


if __name__ == "__main__":
    unittest.main()

```

To achieve coverage for the files above, one can use the `location` macro to point to a specific coverage tool. This is shown below in the BUILD file:

```python
alias(
    name = "python_coverage_tools",
    actual = entry_point("coverage"),
    visibility = ["//src:__subpackages__"],
)

py_test(
    name = "test",
    srcs = ["test.py"],
    env = {
        "PYTHON_COVERAGE": "$(location :python_coverage_tools)",
    },
    deps = [
        ":main",
        ":python_coverage_tools",
    ],
)
```

The problem for this is that for additional python test files, the `PYTHON_COVERAGE` must be set each time. While this process could be automated, it may be easier to use an environment variable that would allow us to use --test_env flag instead of the `location` macro each time.

## What operating system are you running Bazel on?
Debian 10(Buster)

## Bazel Version
bazel version 4.2.1 and above

@c-mita

TLATER:
@lfpino will be interested too :)

Issue #10660, opened 2020-01-27
Python Coverage does not work


ulfjack:
I thought that there was an existing issue for this, but I can't find it.

I was able to successfully collect coverage for python code with a modified version of coverage.py:

1. Use a Bazel binary that include b01c85962d88661ec9f6c6704c47d8ce67ca4d2a
2. Check out the modified coverage.py from https://github.com/ulfjack/coveragepy (this is based on the patch at https://github.com/nedbat/coveragepy/pull/863 with a small change to make it work with Bazel)
3. Run bazel:
```
bazel coverage --test_env=PYTHON_COVERAGE=/path/to/coveragepy/coverage/__main__.py //python:lib_test
```

Test coverage should be written to ```bazel-testlogs/python/lib_test/coverage.dat``` in lcov format.

ulfjack:
In order for cross-language coverage to work in Bazel, all coverage data needs to be converted into a common format. The above workaround uses an upstream patch to coverage.py. As an alternative, we could also change LcovMerger to support coverage.py's output format.

Reference to upstream's issue: https://github.com/nedbat/coveragepy/issues/587

dmadisetti:
The issue you were looking for may have been under python_rules https://github.com/bazelbuild/rules_python/issues/43

Thanks for the suggested workaround!

zuerst:
@ulfjack  thank you for the suggestion. Unfortunately I am seeing empty coverage.dat file with some warnings in in the `test.log` file. Any idea what I might be missing?

**test.log**
```
Jul 08, 2020 6:35:19 PM com.google.devtools.coverageoutputgenerator.Main getTracefiles
INFO: No lcov file found.
Jul 08, 2020 6:35:19 PM com.google.devtools.coverageoutputgenerator.Main getGcovInfoFiles
INFO: No gcov info file found.
Jul 08, 2020 6:35:19 PM com.google.devtools.coverageoutputgenerator.Main getProfdataFileOrNull
INFO: No .profdata file found.
Jul 08, 2020 6:35:19 PM com.google.devtools.coverageoutputgenerator.Main main
WARNING: There was no coverage found.
```

ulfjack:
The most likely reason for coverage not to work is an incorrect `--instrumentation_filter`, and the second most likely is having sources in a `*_test` rule - test rules do not collect any coverage. Beyond that, there are a number of things that could go wrong, but I don't know which one it is. You'd have to post a repro if you need more help.

[...]

PR #14677, opened 2022-01-31
Use upstream coveragepy's new lcov support

TLATER:
As explained in 8c3ea99c2e65edbf400f739cc5cb3115900ac4eb, [coveragepy recently added lcov support](https://github.com/nedbat/coveragepy/pull/1289). This patch set intends to tackle #10660 (and bazelbuild/rules_python#43), and make python coverage fully usable out-of-the box. It builds on work by @ulfjack in #10433, but now with actual support from upstream.

This also incidentally fixes a bug where `bazel test` would no longer work if the `PYTHON_COVERAGE` environment variable was set (can report/fix separately if desired), as well as #14436 (this should perhaps be raised in a separate PR, but made much more sense after refactoring here).

See the commit messages for more detailed comments.

This is a draft, but I would like some review of the overall approach, and whether this is workable in this state. I'm raising this as a draft for these reasons:

- [ ] A commit to allow coveragepy resolution by label is included. It's not implemented correctly at the moment (see commit message), and I'd like feedback on how this could be achieved better.
- [ ] The comments regarding `os.execv` usage seem to have a strange history (see commit message), and I would like to figure out what happened here before continuing the legacy.
- [ ] More consideration for other coverage tools might be more important than I think.
- [ ] Testing for this is slim because it depends on coveragepy and I'm unsure how/if I should integrate that into CI.
- [ ] The python coverage documentation should be updated.

@c-mita and @lfpino may also be interested.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -281,6 +281,17 @@ def Deduplicate(items):
           seen.add(it)
           yield it
 
+def target_to_filepath(build_target, module_space):
+  # TODO (tlater): Find a way to properly resolve this at execution
+  # time, rather than relying on label syntax matching up to a
+  # specific location in runfiles.
+  build_target = build_target.replace("@", "/")
+  build_target = build_target.replace("//", "/")
+  build_target = build_target.replace(":", "/")
+
+  coverage_entry_point = os.path.join(module_space, build_target.lstrip("/") + ".py")
+  return coverage_entry_point
+
 def ExecuteFile(python_program, main_filename, args, env, module_space,
                 coverage_tool=None, workspace=None):
   """Executes the given python file using the various environment settings.
@@ -400,12 +411,14 @@ def Main():
   # COVERAGE_DIR is set iff the instrumentation is configured for the
   # file and coverage is enabled.
   if os.environ.get('COVERAGE_DIR'):
-    if 'PYTHON_COVERAGE' in os.environ:
+    if 'PYTHON_COVERAGE_TARGET' in os.environ:
+      cov_tool = target_to_filepath(os.environ.get('PYTHON_COVERAGE_TARGET'), module_space)
+    elif 'PYTHON_COVERAGE' in os.environ:
       cov_tool = os.environ.get('PYTHON_COVERAGE')
     else:
       raise EnvironmentError(
         'No python coverage tool set, '
-        'set PYTHON_COVERAGE '
+        'set PYTHON_COVERAGE or PYTHON_COVERAGE_TARGET '
         'to configure the coverage tool'
       )
 
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -281,6 +281,71 @@ def Deduplicate(items):
           seen.add(it)
           yield it
 
+def ExecuteFile(python_program, main_filename, args, env, module_space,
+                coverage_tool=None, workspace=None):
+  """Executes the given python file using the various environment settings.
+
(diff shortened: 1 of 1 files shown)


TLATER:
Hm, interesting, I'll have to look into those test failures.

TLATER:
The failing test is `//src/test/shell/bazel:python_version_test`, the culprit is an accidental f-string. New habits die hard, it seems.

comius:
cc @rickeylev for Python

adam-azarchs:
Not to distract from what this PR is trying to accomplish, but it's only part of the puzzle when it comes to hermetic python toolchains.  In order to support those, we'd need to also add an attribute to `py_runtime` to supply a target for coveragepy, and I guess plumb that through to `PYTHON_COVERAGE`.  That sort of thing is particularly important in a remote build scenario where the developer may have no control over whether there exists a system-wide installation of `coverage` on the remote builder.

TLATER:
That sounds like a good idea. We were using `rules_python` to provide it instead, but it also currently lacks toolchain support (and did things rather more impurely than I'm comfortable with).

rickeylev:
> we'd need to also add an attribute to py_runtime to supply a target for coveragepy, and I guess plumb that through to PYTHON_COVERAGE.

I'm not too familiar with how coverage works, so forgive me: why does the `py_runtime` rule need a coverage attribute? That would imply that a particular runtime and coverage are intrinsically linked (e.g. a particular coverage implementation only works with a particular runtime, and vice versa). IIUC, coverage isn't so strictly tied to a runtime, so it seems more like something that would go on the toolchain or rules (which wire together the runtime with other parts).

adam-azarchs:
[`py_runtime`](https://bazel.build/reference/be/python#py_runtime) is how you set up the python toolchain.  A python toolchain setup looks something like
```starlark
py_runtime(
    name = "python3_runtime",
    files = [":python"],
    interpreter = ":python_interpreter_exe",
    python_version = "PY3",
)

py_runtime_pair(
    name = "python_runtime_pair",
    py3_runtime = ":python3_runtime",
)

toolchain(
    name = "python_toolchain",
    exec_compatible_with = ["@platforms//os:linux"],
    target_compatible_with = ["@platforms//os:linux"],
    toolchain = ":python_runtime_pair",
    toolchain_type = "@bazel_tools//tools/python:toolchain_type",
)
```

While one could just put `coverage` into the `files` attribute of `py_runtime`, making it a separate attribute would have a few advantages:
1. No need for the user to muck around with environment variables (which always seems a little but "un-bazely" to me).
2. Only depending on the coverage library when running with code coverage.
3. Being able to supply different coverage libraries to different runtimes, e.g. probably an older version for python 2, if anyone is still using that.

rickeylev:
Ahhhh, huh. Thanks. I didn't realize py_runtime was, in effect, the toolchain (we don't have toolchains enabled for Python at Google, so this isn't an area I'm very familiar with). Well, that's kind of unfortunate naming, but yeah, putting the coverage settings on the toolchain could make sense. TBH, I don't fully understand the pros/cons of a toolchain over an implicit attribute of the rule, but both are pretty transparent to the user, so I'm not sure if it matters much.

adam-azarchs:
The advantage of a toolchain is that it can be configured.  An implicit attribute of the rule has to be a fixed target.  In our use case we very specifically cannot use the automatically-configured python toolchain based on the system-wide installation, because that is usually not the right version of python.

adam-azarchs:
Adding an attribute to `py_runtime` doesn't look like it would be that hard to do; I'd probably have already made a PR to do it except it's [still in the java side of things](https://github.com/bazelbuild/bazel/blob/bea8fb1aaf923ad4d372280868707e2de06a92f5/src/main/java/com/google/devtools/build/lib/rules/python/PyRuntimeRule.java#L29-L30), with which I have a lot less experience.  But also, as I said earlier, it's kind of orthogonal to this PR which, as far as I understand it, would still be required in order to make things actually work.

TLATER:
Superseded by @adam-azarchs ' #15590, thanks :)

PR #14679, opened 2022-01-31
Remove --experimental_worker_allow_json_protocol

keith:
This flag has been flipped, and unlike other experimental flags there is no reason to disable it because workers can only support a single format at once, so if they rely on this, you just wouldn't be able to build.

Closes https://github.com/bazelbuild/bazel/issues/13599

### DIFF
--- a/site/en/docs/creating-workers.md
+++ b/site/en/docs/creating-workers.md
@@ -9,9 +9,7 @@ would benefit from cross-action caching, you may want to implement your own
 persistent worker to perform these actions.
 
 The Bazel server communicates with the worker using `stdin`/`stdout`. It
-supports the use of protocol buffers or JSON strings. Support for JSON is
-experimental and thus subject to change. It is guarded behind the
-`--experimental_worker_allow_json_protocol` flag.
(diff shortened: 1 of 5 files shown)


keith:
cc @aiuto

keith:
Enabling this feature by default should be cherry picked into 5.x, so if we merge this now and leave until 6.x for folks to remove this flag from their configs, that seems safe

fweikert:
I'm not that familiar with workers, so I'd like to get an expert review from @larsrc-google

aiuto:
My 2 cents:
The feature works and is permanent. We don't need the experimental flag to turn it on, since users select the protocol through other means. So, this CL make sense, AND should be backported to 5.x.
The only think I am not sure of is the one line change of of the protocol format selection.

larsrc-google (approved):

larsrc-google:
@susinmotion The flag is already flipped internally and doesn't seem to be causing any harm. So we can indeed remove this. I don't see any reason to backport it, though.

keith:
updated the docs

PR #14695, opened 2022-02-02
docs: add breaking change note to 5.0.0

davidxia:


davidxia:
cc @laurentlb

Wyverald:
The changelog file is unfortunately a bit messed up for the 5.0 release (it does not contain any relnotes); we're trying to fix that for the next release. Unfortunately this PR doesn't really fix the problem either (the commit in question is not a cherry-pick; moreover, it only calls out this specific breaking change, implying there aren't any others). In the meantime, please refer to the [blog post for 5.0](https://blog.bazel.build/2022/01/19/bazel-5.0.html).

davidxia:
Thanks for the quick response. The blog post doesn’t mention this breaking change either. Any chance it can be updated?

davidxia:
@Wyverald i'm happy to make a PR for the blog post to update it if that's possible. Lmk!

Wyverald:
Sorry I missed your previous comment. Yes the blog post should be updatable ([sample PR](https://github.com/bazelbuild/bazel-blog/pull/264)). I just realized this change didn't show up in any changelogs because it didn't correctly have the RELNOTES section in its commit message.

PR #14697, opened 2022-02-03
Fix fail message construction in cc_shared_library

fmeum:
Previously, if there were many unaccounted for libraries, this would simply fail with:
```
Error in join: 'string' is not iterable
```


fmeum:
@oquenchil This is a one-line Starlark fix.

oquenchil (approved):
LGTM

ankit-agarwal1999:
Can we cherrypick this onto 5.1.1. It kind of blocks a core migration workflow (i.e analyzing unaccounted for deps)

fmeum:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

PR #14724, opened 2022-02-04
Expose CoverageOutputGenerator on a Fragment

### LINKED ISSUE #8736: Expose CoverageOutputGenerator's label in a configuration

### LINKED ISSUE #10642: _lcov_merger late-bound test attribute

fmeum:
This allows Starlark rules to define _lcov_merger as a late-bound Starlark configuration_field in order not to pay the cost of building it when coverage isn't enabled.

Fixes #8736 
Fixes #10642


fmeum:
@c-mita

keith:
@c-mita can you review?

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/test/CoverageConfigurationApi.java
```diff
@@ -0,0 +1,34 @@
...
+    category = DocCategory.CONFIGURATION_FRAGMENT,
+    doc = "A configuration fragment representing the coverage configuration.")
+public interface CoverageConfigurationApi extends StarlarkValue {
+
+  @StarlarkMethod(
+      name = "output_generator",
+      allowReturnNones = true,
+      structField = true,
+      doc =
+          "Returns label pointed to by "
```
c-mita, line 21: Can you add a note saying that this will be None if coverage collection isn't enabled?

c-mita (commented):
LGTM, maybe add a comment on the output_generator doc indicating it may be None, since I don't think that's obvious behaviour.

c-mita:
_Ideally_, I'd prefer this was guarded by an experimental flag, but I don't think that's feasible without making the feature clunky to use, so I'll let that pass.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/test/CoverageConfigurationApi.java
fmeum: Done, I also slightly reworded the comment.

c-mita (approved):
LGTM.
Thanks.

Issue #2954, opened 2017-05-05
support for pure C programs?


rtsai:
Please provide the following information. The more we know about your system and use case, the more easily and likely we can help.

### Description of the problem / feature request / question:

We would like to have better support for building pure C programs (not C++). Specifically, we are cross-compiling to a target platform where the boot-time file system does not have libstdc++, but the cc_binary output for that C program has a dynamic link dependency on -lstdc++. The CROSSTOOL linker_flag points to -lstdc++, which is what we want for all the other C++ programs that run after the root file system has been mounted. We can work around this by using "-static" for that particular program, but then we also lose dynamic linking for other things we *do* have (e.g., libc.so).

### If possible, provide a minimal example to reproduce the problem:

Write and build a pure C program. "ldd" on the resulting output will show a dependency on libstdc++

### Environment info

* Operating System: Ubuntu-16.04.2 LTS
* Bazel version (output of `bazel info release`): 0.4.5

###  Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/issues/2548

### Anything else, information or logs or outputs that would be helpful?

https://groups.google.com/d/msg/bazel-discuss/iYzl5XfMDWE/HGV9MUx3AAAJ

One thing we think might work is to create a new CROSSTOOL that is basically a copy of the regular one, but with "linker_flag -lstdc++" removed. This seems to work, but is undesirable, since we do in fact predominantly use C++. Using a different crosstool_top for different packages is undesirable because package-type rules ("create a tarball") don't seem to support depending on outputs of a different build configuration ("create a package of mostly C++ programs, except for this C program").

So the next step might be to introduce a new cc_{binary,library} attribute "lang"

 cc_binary(..., lang="c | c++", ...) which defaults to "c++".

The implementation for "c" would then forbid depending on "c++" providers

The CROSSTOOL configuration would have to add a "cxx_linker_flag" field to be used only when linking C++ programs. The existing "linker_flag" can continue to be used for both C and C++ programs.

kchodorow:
@katre, could you comment on this from a configurations perspective?  Right now, Bazel can only use one toolchain for a build, although we are working on fixing that. 

One thing you might be able to work around this is to go ahead and create a custom C toolchain without the `-lstdc++` flag and then create a macro like:

def cpp_library(name, srcs, hdrs, deps):
  native.cc_library(
    name = name,
    srcs = srcs,
    hdrs = hdrs,
    deps = deps,
[...]

PR #14736, opened 2022-02-06
Add pure C support to unix cc toolchain (#2954)

jheaff1:
This commit splits the user_link_flags_feature so that the default link
libs (i.e libstdc++ and libm) can be removed from the linker command
line by specifying "-default_link_libs" as a cc "feature".

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -891,7 +891,18 @@ def _impl(ctx):
                         iterate_over = "user_link_flags",
                         expand_if_available = "user_link_flags",
                     ),
-                ] + ([flag_group(flags = ctx.attr.link_libs)] if ctx.attr.link_libs else []),
+                ],
+            ),
+        ],
+    )
+
+    default_link_libs_feature = feature(
+        name = "default_link_libs",
+        enabled = True,
+        flag_sets = [
+            flag_set(
+                actions = all_link_actions + lto_index_actions,
+                flag_groups = [flag_group(flags = ctx.attr.link_libs)] if ctx.attr.link_libs else [],
             ),
         ],
     )
@@ -1240,6 +1251,7 @@ def _impl(ctx):
             default_link_flags_feature,
             libraries_to_link_feature,
             user_link_flags_feature,
+            default_link_libs_feature,
             static_libgcc_feature,
             fdo_optimize_feature,
             supports_dynamic_linker_feature,
@@ -1271,6 +1283,7 @@ def _impl(ctx):
             default_compile_flags_feature,
             default_link_flags_feature,
             user_link_flags_feature,
+            default_link_libs_feature,
             fdo_optimize_feature,
             supports_dynamic_linker_feature,
             dbg_feature,


jheaff1:
Any chance of a review of this PR?

jheaff1:
Friendly ping

oquenchil:
Is this a breaking change?

jheaff1:
I don’t believe so, as the default behaviour is unchanged

oquenchil (approved):
LGTM

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -891,7 +891,18 @@ def _impl(ctx):
...
                     ),
-                ] + ([flag_group(flags = ctx.attr.link_libs)] if ctx.attr.link_libs else []),
+                ],
+            ),
+        ],
+    )
+
+    default_link_libs_feature = feature(
+        name = "default_link_libs",
+        enabled = True,
```
oquenchil, line 901: Why is this enabled by default? The original one from which this was split isn't enabled by default.

jheaff1: The starlark implementation of `user_link_flags` doesn't appear to be enabled by default here, yet if i try to disable it using `--features=-user_link_flags`, I get the error message:

```
The C++ toolchain '@local_config_cc//:toolchain' unconditionally implies feature 'user_link_flags', which is unsupported by this rule. This is most likely a misconfiguration in the C++ toolchain.
```

AFAICT, the feature is added unconditionally here: https://github.com/bazelbuild/bazel/blob/b21132597e90c42fc8dac10459b9326f2437ce5c/src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java#L842

oquenchil: That's adding it to the CROSSTOOL but not necessarily enabling it, I can't see where it is enabled by default.

jheaff1: After inspecting the bazel codebase, it seems that the `user_link_flags` is marked as implied by the `c++-compile` action in the `getLegacyActionConfigs` function here:

https://github.com/bazelbuild/bazel/blob/cc55400e614e4e301244cc7c338ee3ea89424ce0/src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java#L1113-L1129

These legacy actions are then added to the `actionConfigList` here:
https://github.com/bazelbuild/bazel/blob/cc55400e614e4e301244cc7c338ee3ea89424ce0/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java#L1224-L1235.

The `actionConfigList` is then provided to the CcToolchainConfigInfo constructor here:
https://github.com/bazelbuild/bazel/blob/cc55400e614e4e301244cc7c338ee3ea89424ce0/src/main/java/com/google/devtools/build/lib/rules/cpp/CcModule.java#L1245-L1262

A CcToolchainFeatures is created which uses this CcToolchainConfigInfo to assemble its `implies` member variable here:
https://github.com/bazelbuild/bazel/blob/cc55400e614e4e301244cc7c338ee3ea89424ce0/src/main/java/com/google/devtools/build/lib/rules/cpp/CcToolchainFeatures.java#L1612

The CcToolchainFeatures formsm part of the `FeatureConfiguration`, which is used in the following lines to detect unconditionally implied features:

https://github.com/bazelbuild/bazel/blob/cc55400e614e4e301244cc7c338ee3ea89424ce0/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCommon.java#L1163-L1172



According to [this](https://docs.bazel.build/versions/main/cc-toolchain-config-reference.html#well-known-features) documentation, it should be possible to disable legacy features using the `no_legacy_features` feature, however this is not implemented in [unix_cc_configure.bzl ](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/unix_cc_configure.bzl)

hlopko:
Drive by comment, is it an option to use `BAZEL_LINKLIBS` and `BAZEL_LINKOPTS` (relevant code is around here: https://cs.opensource.google/bazel/bazel/+/master:tools/cpp/unix_cc_configure.bzl;l=419)?

PR #14738, opened 2022-02-07
Add "arch" struct field to repository_os

### LINKED ISSUE #14685: Provide the exec CPU to repository rules

Repository rules already have access to `repository_os` which allows rules to change their behaviour based on the exec OS. Repository rules should also know the exec architecture. Repository rules currently work around this limitation by  `uname -m`, which isn't ideal. See [rules_go](https://github.com/bazelbuild/rules_go/blob/fd11dd2768669dc2cc1f3a11f2b0b81d84e81c32/go/private/sdk.bzl#L253).

Repository rules create potentially brittle and strange workarounds to get the exec CPU. There is no standardisation and no official guidance on how best to create repository rules which target multiple platforms.

Linux x86_64

```sh
❯ bazel info release
release 5.0.0
```

- https://github.com/bazelbuild/bazel/issues/11655#issuecomment-734212603
- https://github.com/bazelbuild/bazel/issues/4636 (request was target, not exec)

N/A

fmeum:
This new field provides access to the Java "os.arch" property to
repository rules, which previously had to rely on uname (Unix) or
additional env variables (Windows) to detect the architecture.

This also fixes a small issue in the existing implementation of
repository_ctx.os.name, which should use the root locale when converting
the value of the "os.name" property to lowercase.

Existing sites of manual architecture detection in shipped repository
rules as well as redundant calls to lower() on the value of
repository_ctx.os.name are cleaned up.

Fixes #14685

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkOS.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkOS.java
@@ -17,12 +17,15 @@ package com.google.devtools.build.lib.bazel.repository.starlark;
 import com.google.common.collect.ImmutableMap;
 import com.google.devtools.build.docgen.annot.DocCategory;
 import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
+import java.util.Locale;
 import java.util.Map;
 import net.starlark.java.annot.StarlarkBuiltin;
 import net.starlark.java.annot.StarlarkMethod;
 import net.starlark.java.eval.StarlarkValue;
 
-/** A Starlark structure to deliver information about the system we are running on. */
+/**
+ * A Starlark structure to deliver information about the system we are running on.
+ */
 @StarlarkBuiltin(
     name = "repository_os",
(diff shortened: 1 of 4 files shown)


Wyverald:
Looks great, thank you! Any more changes planned before you un-draft this?

Wyverald:
Ah, I see. It seems fine to me to use `os.arch` since we are running in a JVM after all, but let's wait for consensus on the other thread. cc @meteorcloudy

meteorcloudy:
Nice, thanks for adding this!

@katre With this change, we probably can starlarkify https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/repository/LocalConfigPlatformFunction.java ?

meteorcloudy (approved):

fmeum:
@meteorcloudy Marked as ready, thanks for the review.

katre:
@meteorcloudy Yes, we could possibly use this to rewrite LocalConfigPlatformFunction, although it's at such a low level in the hierarchy of dependencies I'd be very worried about correctness and performance.

Issue #11997, opened 2020-08-22
py_binary located via manifest-based runfiles lib fails when its own runfiles tree isn't built


pcjanzen:
### Description of the problem / feature request:

Consider a Starlark-generated executable target, which depends at runtime on a py_binary by placing it in its data attr, and which locates the path to that py_binary using a runfiles library that uses runfiles_manifest files. 

In this scenario, the target will execute the py_binary using the "real" path of the binary inside `bazel-bin`, not a path relative to the original target's runfiles.

Bazel doesn't necessarily build the py_binary's own runfiles tree, so the python stub, which only looks at its own `sys.argv[0]`, can't find its module space.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Expand [Python-binary-failure-example.zip](https://github.com/bazelbuild/bazel/files/5113103/Python-binary-failure-example.zip) then run `bazel run //:exe`

### What operating system are you running Bazel on?

macosx

### What's the output of `bazel info release`?

release 3.4.1

pauldraper:
> so the python stub, which only looks at its own sys.argv[0], can't find its module space.

The usual fix is to require calling executables to set `RUNFILES_DIR` environment variable, and then use that if it is set.

rickeylev:
A few observations.

`$(rlocation)` is returning a path that is outside of runfiles. `$(rlocation workspace/inner_bin)` is returning

`/home/<user>/.cache/bazel/_bazel_<user>/<hash>/execroot/<workspace>/bazel-out/<config>/bin/inner_bin` (I'm calling this the "canonical" path, since its the real file everything else symlinks to)

not e.g. `/<snip>/outer_bin.runfiles/<workspace>/inner_bin`. (when invoked like this, it works)

If `outer_bin` has a data dependency on `inner_bin`, then the official location (wrt outer bin) of `inner_bin` is `$outer_bin_runfiles_dir/<workspace>/inner_bin`. Sure, the one under outer bin's runfiles is a symlink, but this still seems problematic (also, this sort of thing _could_ be viewed as a potential path-escape vulnerability, as you're forced to accept arbitrary paths that may not actually be in runfiles and shouldn't be accessible). FWIW, internally at Google, I think I've only ever seen the runfiles-relative paths used in e.g. shell programs, e.g. `$RUNFILES/inner_bin --whatever", and, similarly, internally, we always use a runfiles-relative path for accessing runfiles data.

[...]

PR #14740, opened 2022-02-07
Let Python stubs respect RUNFILES_* while finding the module space

EdSchouten:
When invoking a py_binary() through an sh_binary() using the Bash
runfiles library, the location of the py_binary() will be resolved from
the runfiles manifest file. This means that the argv[0] of the Python
stub script will not point to a location under the runfiles directory of
the shell script, but to a location in Bazel's execroot.

Normally this does not lead to any issues, as argv[0] + ".runfiles"
happens to point to be a valid runfiles directory as well. It does
become problematic when --nobuild_runfile_links is provided, as in that
case only the outer shell script is guaranteed to have a runfiles
directory; not the inner Python script.

This change extends the Python stub template to also consider
RUNFILES_DIR when no runfiles directory can be found. Even though it's
not technically correct, we also attempt to derive the runfiles
directory from RUNFILES_MANIFEST_FILE. I suspect that this is a
necessity as long as py_binary()s cannot operate purely using a manifest
file. They currently depend on a concrete instantiation of the runfiles
directory.

Fixes: #11997

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -129,8 +129,24 @@ def CreatePythonPathEntries(python_imports, module_space):
   parts = python_imports.split(':')
   return [module_space] + ['%s/%s' % (module_space, path) for path in parts]
 
-def FindModuleSpace():
+def FindModuleSpace(main_rel_path):
   """Finds the runfiles tree."""
+  # When the calling process used the runfiles manifest to resolve the
+  # location of this stub script, the path may be expanded. This means
+  # argv[0] may no longer point to a location inside the runfiles
+  # directory. We should therefore respect RUNFILES_DIR and
+  # RUNFILES_MANIFEST_FILE set by the caller.
+  runfiles_dir = os.environ.get('RUNFILES_DIR', None)
+  if not runfiles_dir:
+    runfiles_manifest_file = os.environ.get('RUNFILES_MANIFEST_FILE', '')
+    if (runfiles_manifest_file.endswith('.runfiles_manifest') or
+        runfiles_manifest_file.endswith('.runfiles/MANIFEST')):
+      runfiles_dir = runfiles_manifest_file[:-9]
+  # Be defensive: the runfiles dir should contain our main entry point. If
+  # it doesn't, then it must not be our runfiles directory.
+  if runfiles_dir and os.path.exists(os.path.join(runfiles_dir, main_rel_path)):
+    return runfiles_dir
+
   stub_filename = sys.argv[0]
   if not os.path.isabs(stub_filename):
     stub_filename = os.path.join(os.getcwd(), stub_filename)
@@ -413,10 +429,17 @@ def Main():
 
   new_env = {}
 
+  # The main Python source file.
+  # The magic string percent-main-percent is replaced with the runfiles-relative
+  # filename of the main file of the Python binary in BazelPythonSemantics.java.
+  main_rel_path = '%main%'
+  if IsWindows():
+    main_rel_path = main_rel_path.replace('/', os.sep)
+
   if IsRunningFromZip():
     module_space = CreateModuleSpace()
   else:
-    module_space = FindModuleSpace()
+    module_space = FindModuleSpace(main_rel_path)
 
   python_imports = '%imports%'
   python_path_entries = CreatePythonPathEntries(python_imports, module_space)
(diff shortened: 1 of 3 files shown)


src/test/shell/bazel/bazel_example_test.sh
```diff
@@ -151,15 +151,17 @@ function test_shell() {
 function test_python() {
   assert_build "//examples/py:bin"
 
-  ./bazel-bin/examples/py/bin >& $TEST_log \
+  env -u RUNFILES_DIR -u RUNFILES_MANIFEST_FILE \
```
EdSchouten, line 154: This was necessary, because the test itself also happens to drag in `//examples/py` from the Bazel source tree. This means that with the new logic added, the Python stub script would end up picking the wrong runfiles directory now.

I suspect that this part of the change is acceptable, because lookups of runfiles were already guaranteed to fail within the called process.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -95,8 +95,22 @@ def CreatePythonPathEntries(python_imports, module_space):
...
+  # argv[0] may no longer point to a location inside the runfiles
+  # directory. We should therefore respect RUNFILES_DIR and
+  # RUNFILES_MANIFEST_FILE set by the caller.
+  runfiles_dir = os.environ.get('RUNFILES_DIR', None)
+  if not runfiles_dir:
+    runfiles_manifest_file = os.environ.get('RUNFILES_MANIFEST_FILE', '')
+    if (runfiles_manifest_file.endswith('.runfiles_manifest') or
+        runfiles_manifest_file.endswith('.runfiles/MANIFEST')):
+      runfiles_dir = runfiles_manifest_file[:-9]
+  if runfiles_dir and os.path.exists(os.path.join(runfiles_dir, rel_path)):
```
EdSchouten, line 111: This is a safety belt to ensure that we only respect RUNFILES_DIR if we know it points to a valid runfiles directory that (most likely) belongs to the Python script. If it happens to point to a disjoint runfiles directory owned by the parent, it'll end up ignoring it.

EdSchouten:
Friendly ping!

EdSchouten:
Hi @brandjon; happen to have a couple of spare cycles to look into this? Thanks!

EdSchouten:
Hey @comius, would you be interested in reviewing this change for me? Thanks!

EdSchouten:
Hmmm... The CI failures for "buildkite/bazel-bazel-github-presubmit/darwin-openjdk-11-xcode-shard-*" seem to be caused by infrastructure failure. The underlying system ran out of disk space?

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
rickeylev: Did you see this case occur? If so, how did it come up?

I think this logic is fine, just add a comment:
```suggestion
  # Be defensive: the runfiles dir should contain our main entry point. If
  # it doesn't, then it must not be our runfiles directory.
  if runfiles_dir and os.path.exists(os.path.join(runfiles_dir, rel_path)):
```

EdSchouten: Yes! I saw a bunch of shell tests in the Bazel repository itself fail if this logic wasn't present. I've added the comment you've suggested.

src/test/shell/bazel/bazel_example_test.sh
EdSchouten: Done!

EdSchouten:
Hi @rickeylev, I think I've addressed all of your comments. PTAL!

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
rickeylev: Do you recall which, by chance?
I just want to look if the tests you saw fail have the same property for some tests I saw failing while I was working on this same change for our Google-version of the stub script. The thing I saw were some shell tests directly embedding a blaze-bin path directly into the tests rather than going through $0.runfiles.

rickeylev (approved):

EdSchouten:
Hey @rickeylev, just tagging you because two days have passed. :D

sgowroji:
Hello @EdSchouten, Thank you for the patience. I will update once it gets merged.

rickeylev:
Just to give an update: I've been fighting with our internal windows test environment to make the test added to python_stub_test.sh pass for the last day or so. For whatever reason, it seems to be missing a lot of the supporting code/config that the github environment has. If I don't have it figured out by Monday, I'll just disable this test in our internal builds for now.

rickeylev:
Woo, got everything happy! Code is out for review internally; should land early next week.

PR #14741, opened 2022-02-07
Enable r8 support in bazel

mauriciogg:
This cl introduces R8 support in Bazel for proguarding and dexing.
R8 is enabled by passing --experimental_enable_r8=true to the
blaze build invocation and --proguard_top="//bzl/r8:proguard_compat"

When enabled the build graph was changed from
 - jar -> proguard -> dex -> apk
To
 - jar -> r8 -> apk

 I.e R8 takes care of proguarding and dexing.

There are a few known limitations:

 - Only native multidex (Android version L+) is supported.
   All android_binary targets declare native_multidex so it does not
   make sense to implement legacy multidex with R8
 - Mobile-Install is not supported: mobile-install relies on how dexes
   are shard. R8 dexes the deploy jar in a single monolithic action
   which is cant be done incrementally (I don't think mobile-install is
   supported anymore)
 - Other features which are not implemented in bazel at the moment:
   e.g. dynamic delivery
 - proguard_compat needs to be available (not in this commit)

see https://github.com/bazelbuild/rules_android/issues/31

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidBinary.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidBinary.java
@@ -25,6 +25,7 @@ import com.google.common.base.Functions;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.devtools.build.lib.actions.Artifact;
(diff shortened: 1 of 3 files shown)


nkoroste:
@ahumesky @ted-xie

gabrielborglund:
@mauriciogg Why is it better to pass R8 via the proguard_top flag rather than changing how AndroidBinary works? That is why not implement something like applyR8 instead of applyProguard?

mauriciogg:
> @mauriciogg Why is it better to pass R8 via the proguard_top flag rather than changing how AndroidBinary works? That is why not implement something like applyR8 instead of applyProguard?

Technically this code is being deprecated so I wouldnt want to add modifications to android binary, besides this just works out of the box with the rest of the toolchain. The only extra piece needed is a shim binary that translates between proguard args and r8 args. This is very unlikely to get merge though.

Issue #14727, opened 2022-02-04
Update CC rules to use optional toolchains


katre:
Once #14726 is done, the CC rules need to use optional toolchains, then we can remove the dummy cc toolchain.

PR #14742, opened 2022-02-07
Add a helper method for rules to depend on the cpp toolchain type.

### LINKED ISSUE #14728: Add a helper function for rules that need the CC toolchain
Currently, rules that require the CC toolchain via toolchain resolution use this syntax:
```
foo = rule(
    ...
    toolchains = ["@bazel_tools//tools/cpp:toolchain_type"],
)
```

(Possibly with other toolchain types present)

With [optional toolchains](https://github.com/bazelbuild/proposals/blob/main/designs/2022-01-21-optional-toolchains.md), the syntax is changing, and will wait on a Bazel release.

We should provide a simple helper function that rules can use to be forward compatible when the new optional toolchains are release:
```
load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "use_cpp_toolchain")

foo = rule(
    ...
    toolchains = [other toolchains] + use_cpp_toolchain(mandatory = True),
)
```

The implemenation of `use_cpp_toolchain` will update with Bazel and will either return the bare-string toolchain type (as used currently) or the more complicated `config.toolchain_type` call. Rules won't need to make changes (beyond changing the value of `mandatory` if desired).

The only parameter will be `mandatory`, which will default to `True` for backwards compatibility.

katre:
Fixes #14728, part of #14727.

Rules can begin using this as soon as it is released (hopefully in Bazel
5.1), and then future versions of Bazel will update the functionality as
optional toolchains are added and C++ rules are updated to use them.

### DIFF
--- a/tools/cpp/cc_flags_supplier.bzl
+++ b/tools/cpp/cc_flags_supplier.bzl
@@ -15,7 +15,7 @@
 
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "CC_FLAGS_MAKE_VARIABLE_ACTION_NAME")
 load("@bazel_tools//tools/cpp:cc_flags_supplier_lib.bzl", "build_cc_flags")
-load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain", "use_cpp_toolchain")
 
 def _cc_flags_supplier_impl(ctx):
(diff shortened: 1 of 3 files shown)


brentleyjones:
> Will this make migrations to use optional toolchains easier, or will it just add an extra precondition because users want to support pre-5.1 versions of Bazel?

It will make it possible to be forward compatible, but only once a ruleset only supports Bazel 5.1+. That's better than nothing imo.

fmeum:
I created https://github.com/bazelbuild/bazel/pull/14795 to get this cherry-picked into 5.1.0.

katre:
Thank you!

PR #14755, opened 2022-02-08
[5.1] Allow Label instances as keys in select

### LINKED ISSUE #14259: Bzlmod: a project cannot provide a macro that uses select statement

### LINKED ISSUE #14721: [5.1] Allow Label instances as keys in select
Forked from #14447

brentleyjones:
When a macro specifies a label string as a key in a select, this label
is resolved relative to the site of use rather than the .bzl file the
macro is defined in. The resolution will lead to incorrect results if
the repository that uses the macro has a different repo mapping, e.g.
because it is created by another Bazel module.

This can be solved by allowing macros to specify label instances created
with the `Label` constructor instead of label strings everywhere, which
previously was not possible in select.

This commit also updates the docs for Label, select and macros.

Fixes #14259.

(cherry picked from commit 69f8b172ae8495587751564284fc126300cae6fe)

Closes #14721.

### DIFF
--- a/site/docs/skylark/macros.md
+++ b/site/docs/skylark/macros.md
@@ -143,6 +143,38 @@ macro), use the function [native.package_name()](lib/native.html#package_name).
 Note that `native` can only be used in `.bzl` files, and not in `WORKSPACE` or
 `BUILD` files.
 
+## Label resolution in macros
+
+Since macros are evaluated in the [loading phase](concepts.md#evaluation-model),
+label strings such as `"//foo:bar"` that occur in a macro are interpreted
+relative to the `BUILD` file in which the macro is used rather than relative to
+the `.bzl` file in which it is defined. This behavior is generally undesirable
+for macros that are meant to be used in other repositories, e.g. because they
+are part of a published Starlark ruleset.
+
+To get the same behavior as for Starlark rules, wrap the label strings with the
+[`Label`](lib/Label.html#Label) constructor:
+
+```python
+# @my_ruleset//rules:defs.bzl
+def my_cc_wrapper(name, deps = [], **kwargs):
+  native.cc_library(
+    name = name,
+    deps = deps + select({
+      # Due to the use of Label, this label is resolved within @my_ruleset,
+      # regardless of its site of use.
+      Label("//config:needs_foo"): [
+        # Due to the use of Label, this label will resolve to the correct target
+        # even if the canonical name of @dep_of_my_ruleset should be different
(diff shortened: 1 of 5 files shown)


aiuto:
Who is the release manager for the 5.x track?. They should be the reviewer of all feature backports.

aiuto (approved):

brentleyjones:
@Wyverald

Wyverald:
> Who is the release manager for the 5.x track?. They should be the reviewer of all feature backports.

I'm managing 5.1 still. I'll look into better labeling PRs that are not going into the main branch.

PR #14760, opened 2022-02-08
config doesn't error on duplicate `--define` values

illicitonion:
When `--define` flags are interpreted by other commands, the last one
wins. Currently `bazel config` causes a server crash when interpreting
duplicate `--define` values.

This change makes `config` consistent with the other commands, and
re-uses the same deduplication code across all call-sites.


katre (approved):
Looks good, I'll get this merged once the tests pass.

illicitonion:
@bazel-io fork 5.2

keith:
@bazel-io fork 5.2

ckolli5:
@bazel-io fork 5.2.0

PR #14762, opened 2022-02-08
config doesn't error on duplicate `--define` values

illicitonion:
When `--define` flags are interpreted by other commands, the last one
wins. Currently `bazel config` causes a server crash when interpreting
duplicate `--define` values.

This change makes `config` consistent with the other commands, and
re-uses the same deduplication code across all call-sites.

Closes #14760.

PiperOrigin-RevId: 427266039


gregestren:
Hi @oquenchil. What do I need to do if this is going into the 5.1 branch? 

i.e. we already have this in `bazelbuild:master`: https://github.com/bazelbuild/bazel/pull/14760

illicitonion:
@Wyverald it looks like we missed this getting into the 5.1 branch?

PR #14765, opened 2022-02-09
Add support for fission on android

keith:
This change adds the per_object_debug_info feature to the android cc
toolchain to support fission.

This also removes this same feature from the macOS crosstool because
clang targeting darwin does nothing when passed this flag, which leads
to failures if you enable this for android, but are on a macOS host
because host tools fail to produce the dwo files.


Bencodes:
cc: @ahumesky

keith:
This one is also supported by the new NDK rules https://github.com/bazelbuild/rules_android_ndk/blob/40e17b3af75406ab10219313f16a404937f09874/ndk_cc_toolchain_config.bzl#L886-L894 so I will close this once it's more clear those are the future

keith:
dropping for the new ndk rules

Issue #14726, opened 2022-02-04
Implement Optional Toolchains


katre:
Now that https://github.com/bazelbuild/proposals/blob/main/designs/2022-01-21-optional-toolchains.md is approved, this is the tracking issue for the core implemenation.

Updating rules to use the new feature will be tracked by separate issues.

PR #14770, opened 2022-02-09
Add non-Starlark optional toolchain API

katre:

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/ToolchainTypeRequirement.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/ToolchainTypeRequirement.java
@@ -1,0 +1,47 @@
+// Copyright 2022 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/analysis/config/ToolchainTypeRequirement.java
```diff
@@ -0,0 +1,47 @@
...
+  /** Returns a builder for a new {@Link ToolchainTypeRequirement}. */
+  public static Builder builder() {
+    return new AutoValue_ToolchainTypeRequirement.Builder()
+        .mandatory(true);
+  }
+
+  /** The toolchain type this requirement is for. */
+  public abstract Label toolchainType();
+
+  /** Whether the toolchain type is mandatory or optional. */
```
gregestren, line 33: Can we contextualize this comment a bit more? Just another sentence suggesting the implication of these terms?

katre: Done.

gregestren (approved):

Issue #11561, opened 2020-06-07
Documentation incorrect --fission has no effect


andrewstevens-infineon:
### Description of the problem / feature request:

The documentation suggestes that the --fission build option isimplement.   

> --fission (yes|no|[dbg][,opt][,fastbuild])
> 
> Enables Fission, which writes C++ debug information to dedicated .dwo files instead of .o files, where it would otherwise go. This substantially reduces the input size to links and can reduce link times.
> 
> When set to [dbg][,opt][,fastbuild] (example: --fission=dbg,fastbuild), Fission is enabled only for the specified set of compilation modes. This is useful for bazelrc settings. When set to yes, Fission is enabled universally. When set to no, Fission is disabled universally. Default is dbg. 

At least for linux/gcc this is not the case.  The option has no effect whatsoever.
To function correct with gcc the -gsplit-dwarf option should be used. This does not occur.
Additionally --gdb-index should be used with the gold linker.  Neither occurs

### Feature requests: what underlying problem are you trying to solve with this feature?

Documentation matching actual implementation status.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Use bazel C++ [examples]( https://github.com/bazelbuild/examples).

`cd examples/cpp-tutorial/stage1`

`bazel buiild  --fission=yes -c dbg --subcommands=yes //main:hello-world`

The gcc command-line still selects -g not -gsplit-dwarf.

`cat bazel-out/k8-dbg/bin/main/hello-world-2.params`

The linker command line has no -Wl,--gdb-index.

### What operating system are you running Bazel on?

Ubuntu 18.04.

### What's the output of `bazel info release`?

release 3.2.0

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

https://github.com/bazelbuild/examples

###  Have you found anything relevant by searching the web?

Yes. Very old article discussion formum article [here](https://groups.google.com/forum/#!searchin/bazel-discuss/fission%7Csort:date/bazel-discuss/GhP-gODplX0/aQBl7XUwCQAJ)

Inidicating fission option was non-functional for default toolchains.

Riatre:
First, the [user manual](https://docs.bazel.build/versions/master/user-manual.html#flag--fission) say:

[...]

PR #14777, opened 2022-02-09
Enable per_object_debug_info feature by default

keith:
This feature is used as a flag to determine that the current toolchain
supports per object file debug info, it is only consulted when
`--fission` is enabled for the current compilation mode. Previously you
had to enable fission and this flag, with this change enabling fission
by itself should be enough.

Related: https://github.com/bazelbuild/bazel/issues/11561


tomaspipes:
Any reason for this not to be integrated?

fmeum:
It was merged in https://github.com/bazelbuild/bazel/commit/5f51d21e918a4c7286c5024c3c50f2d931662af9

PR #14805, opened 2022-02-12
Update rules.md

Vertexwahn:
xcode rules do not exist anymore

### DIFF
--- a/site/docs/rules.md
+++ b/site/docs/rules.md
@@ -72,8 +72,6 @@ BUILD files without a `load` statement.
   - [`bind`](be/workspace.html#bind)
   - [`local_repository`](be/workspace.html#local_repository)
   - [`new_local_repository`](be/workspace.html#new_local_repository)
-  - [`xcode_config`](be/workspace.html#xcode_config)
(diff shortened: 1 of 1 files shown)


keith:
Hrm these rules definitely still exist and are in use. Maybe their docs disappearing (or moving?) is a bug?

Here's xcode_config: https://github.com/bazelbuild/bazel/blob/d533996b2c8eabaa8e796f3aad57bca13878f460/src/main/java/com/google/devtools/build/lib/rules/apple/XcodeConfigRule.java#L30

Here's xcode_version https://github.com/bazelbuild/bazel/blob/d533996b2c8eabaa8e796f3aad57bca13878f460/src/main/java/com/google/devtools/build/lib/rules/apple/XcodeVersionRule.java#L27

gregestren:
I assume this is related to the recent website refresh?

gregestren:
Looks like the links resolve now. Can I get a second opinion?

keith:
yep looks like the links work now from https://bazel.build/rules?hl=en

Issue #14803, opened 2022-02-12
On Apple M1, objc_library compiled with darwin_x86_64 but cc_library compiled with darwin_arm64, and failed at link time


merckhung:
### Description of the problem / feature request:

I have a project uses both cc_library and objc_library to compile for MacOS X platform.
The project could compile and run properly until I switched to Apple M1 based MacOS X.

I discovered that on M1 MacOS X, Bazel compiles objc_library with darwin_x86_64 platform and compiles
cc_library/cc_binary with darwin_arm64 platform. Therefore, the project no longer links these two libraries successfully.
The compilation fails at link-time with this error message:

ld: warning: ignoring file bazel-out/applebin_macos-darwin_x86_64-dbg-ST-cfb676c6944e/bin/test/libmacfile.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
Undefined symbols for architecture arm64:

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I have a minimal example project to reproduce this problem, please find the attached file.

[objc_apple_m1_linker.zip](https://github.com/bazelbuild/bazel/files/8052579/objc_apple_m1_linker.zip)

unzip objc_apple_m1_linker.zip
cd objc_apple_m1_linker
bazel build -c dbg //test -s
ls -al bazel-out/

drwxr-xr-x  3 merck  wheel   96 Feb 11 18:38 applebin_macos-darwin_x86_64-dbg-ST-cfb676c6944e
drwxr-xr-x  4 merck  wheel  128 Feb 11 18:38 darwin_arm64-dbg

* objc_library was compiled into applebin_macos-darwin_x86_64-dbg-ST-cfb676c6944e
* cc_binary was compiled into darwin_arm64-dbg, they failed to link with each other.

### What operating system are you running Bazel on?

Apple M1 (ARM64) MacOS X 12.2.1
bazel version is 6.0.0-pre.20220112.2 (Darwin ARM64 variant bazel-6.0.0-pre.20220112.2-darwin-arm64)

### What's the output of `bazel info release`?

release 6.0.0-pre.20220112.2

###  Have you found anything relevant by searching the web?

I tried, but I didn't find anything similar.

keith:
Thanks for the repro case, fix here! https://github.com/bazelbuild/bazel/pull/14816

brentleyjones:
@bazel-io fork 5.1

keith:
[...]

PR #14816, opened 2022-02-14
Fix default CPU for macOS and iOS

keith:
Since M1 support was added we started defaulting the CPU for macOS and
iOS builds to the host CPU in the case another CPU wasn't passed. This
change mirrors https://github.com/bazelbuild/bazel/pull/13873, https://github.com/bazelbuild/bazel/pull/13440 and https://github.com/bazelbuild/rules_apple//commit/99e5b631bf060358241a8eaabd285be5c120490f
in the newer starlark transition.

Without this you end up with architecture mismatches for builds. Fixes https://github.com/bazelbuild/bazel/issues/14803


keith:
cc @oquenchil

keith:
@oquenchil can you help land so we can get this in 5.1

PR #14823, opened 2022-02-15
Update AspectDefinition to use ToolchainTypeRequirement.

katre:

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/DexArchiveAspect.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/DexArchiveAspect.java
@@ -45,6 +45,7 @@ import com.google.devtools.build.lib.analysis.actions.CustomCommandLine.VectorAr
 import com.google.devtools.build.lib.analysis.actions.SpawnAction;
 import com.google.devtools.build.lib.analysis.config.BuildConfigurationValue;
 import com.google.devtools.build.lib.analysis.config.ExecutionTransitionFactory;
+import com.google.devtools.build.lib.analysis.config.ToolchainTypeRequirement;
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/devtools/build/lib/rules/android/DexArchiveAspect.java
```diff
@@ -182,8 +183,12 @@ public AspectDefinition getDefinition(AspectParameters params) {
                     // For proto_lang_toolchain rules, where we just want to get at their runtime
                     // deps.
                     ImmutableSet.of(ProtoLangToolchainProvider.class)))
-            .addRequiredToolchains(
-                Label.parseAbsoluteUnchecked(toolsRepository + sdkToolchainLabel))
+            .addToolchainType(
+                ToolchainTypeRequirement.builder()
+                    .toolchainType(
+                        Label.parseAbsoluteUnchecked(toolsRepository + sdkToolchainLabel))
+                    .mandatory(true)
```
katre, line 190: I'm a little unhappy with the verbosity here. Should I add a shorter version?

One thought:
```
.addToolchainType(ToolchainTypeRequirement.builder(typeLabel))
```

Thoughts?

gregestren: Sure. As currently written I had some trouble identifying when the builder resolves and how it generally fits with the pattern.

katre: Okay, added them as #14825, once that's in I'll rebase this.

katre: Rebased.

gregestren (approved):

Issue #14826, opened 2022-02-16
ActionConflictException for SolibSymlink action created for dynamic_runtime_lib 


glukasiknuro:
### Description of the problem

`ActionConflictException` shows up when using combination of CC toolchains with runtime libs, and rules that require specific execution platforms - `rules_docker` in the reproduction below, but it can be reproduced for e.g. tensorflow as well.

Looks like the problematic action is `SolibSymlink` action. Changing the action to ignore execution platforms and exec properties fixes the issue (WIP: https://github.com/bazelbuild/bazel/pull/14827). Similar change was made to Middleman in https://github.com/bazelbuild/bazel/commit/49269551269cafeec733442477569d96883ef58f, possibly it can be applied for SolibSymlink as well?

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Code to reproduce as well and instructions: https://github.com/glukasiknuro/bazel_solibsymlink_conflict_issue

Error message- `executionPlatformLabel` seems to be the problem:
```
./bazel-5.0.0-linux-x86_64 build //:test_py_image

Starting local Bazel server and connecting to it...
ERROR: file '_solib__toolchain_Ck8_Ucc_Utoolchain/foo.so' is generated by these conflicting actions:
Label: //toolchain:k8_cc_toolchain
RuleClass: cc_toolchain rule
Configuration: 30ef7fa7561d68e72a06426dbdb1ff29fe154ce431958a511b45458ec146af56
Mnemonic: SolibSymlink
Action key: fb9103b4f577c25c225b84492070aef8fb18733871a75006b27c69ee7311572c, 1fb5751a4d74877adaf30c2a75ff8ae43324906fb1e7f94f6bf94b426e3566d6
Progress message: (null)
PrimaryInput: File:[/(...)/bazel_solibsymlink_conflict_issue[source]]toolchain/libs/foo.so
PrimaryOutput: File:[[<execution_root>]bazel-out/k8-py2-opt-ST-4a519fd6d3e4/bin]_solib__toolchain_Ck8_Ucc_Utoolchain/foo.so
Owner information: 

ToolchainDependencyConfiguredTargetKey{
label=//toolchain:k8_cc_toolchain,
config=BuildConfigurationValue.Key[30ef7fa7561d68e72a06426dbdb1ff29fe154ce431958a511b45458ec146af56], 
executionPlatformLabel=@io_bazel_rules_docker//platforms:local_container_platform}, 

[...]

PR #14827, opened 2022-02-16
SolibSymlinkAction does not need exec platform or properties

glukasiknuro:
https://github.com/bazelbuild/bazel/issues/14826


jlaxson:
@bazel-io flag 

I'd like to suggest this should be a blocker for the 5.x series.  It is a regression from 4.2, appears to have a viable fix ready, and the underlying issue doesn't seem to have a workaround.

jlaxson:
@bazel-io flag

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

oquenchil (approved):
LGTM

PR #14836, opened 2022-02-16
[5.1] Fix bazel coverage false negative

### LINKED ISSUE #14820: [5.1] Fix bazel coverage false negative
Forked from #14807

brentleyjones:
Previously this short circuit meant the tests weren't actually run, and
they would always exit 0, in the case the test rule didn't set the lcov
related attributes.

More context: https://github.com/bazelbuild/bazel/issues/13978

RELNOTES: Fixed an issue where Bazel could erroneously report a test passes in coverage mode without actually running the test.
(cherry picked from commit 16de03595e21f7bf31818e717505b23c953b3b7d)

Closes #14820.


brentleyjones:
@Wyverald

PR #14614, opened 2022-01-21
add review ownership for apple and configurability

aiuto:
cc: @keith

PR #14844, opened 2022-02-17
Correct cpu and os values of `local_config_cc_toolchains` targets

thii:
These aren't being used now, but some of them are clearly defined
incorrectly. Fix them now to avoid surprises in the future. Also change
`aarch64` to `arm64` since `aarch64` is an alias of `arm64`.

### DIFF
--- a/tools/osx/crosstool/BUILD.toolchains
+++ b/tools/osx/crosstool/BUILD.toolchains
@@ -10,11 +10,11 @@ OSX_TOOLS_CONSTRAINTS = {
     "armeabi-v7a": ["@platforms//cpu:arm"],
     "darwin_arm64": [
         "@platforms//os:osx",
-        "@platforms//cpu:aarch64",
+        "@platforms//cpu:arm64",
     ],
     "darwin_arm64e": [
         "@platforms//os:osx",
-        "@platforms//cpu:aarch64",
+        "@platforms//cpu:arm64",
     ],
     "darwin_x86_64": [
         "@platforms//os:osx",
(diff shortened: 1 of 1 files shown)


gregestren:
@keith

allevato (approved):
Being pedantic, `arm64` should probably be an alias for `aarch64` and not the other way around, but I don't know whether it's worth drilling that deep at this point.

These changes seem fine, otherwise.

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -10,76 +10,76 @@ OSX_TOOLS_CONSTRAINTS = {
...
-        "@platforms//os:ios",
+        "@platforms//os:tvos",
         "@platforms//cpu:x86_64",
     ],
     "tvos_sim_arm64": [
-        "@platforms//os:ios",
-        "@platforms//cpu:aarch64",
+        "@platforms//os:tvos",
+        "@platforms//cpu:arm64",
     ],
```
keith, line 58: I wondered this in another context but is this one correct? since it's not actually different from `tvos_arm64`? maybe the name difference is enough?

keith:
@gregestren it seems like the github owners aren't working correctly since we didn't get cc'd here?

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -10,76 +10,76 @@ OSX_TOOLS_CONSTRAINTS = {
...
     "ios_x86_64": [
         "@platforms//os:ios",
         "@platforms//cpu:x86_64",
     ],
     "ios_sim_arm64": [
         "@platforms//os:ios",
-        "@platforms//cpu:aarch64",
+        "@platforms//cpu:arm64",
     ],
```
keith, line 46: some question as tvos

keith (approved):

tools/osx/crosstool/BUILD.toolchains
allevato: A while back we checked in environment constraints for simulator, device, and catalyst into https://github.com/bazelbuild/apple_support/blob/master/constraints/BUILD, but I don't know enough about Bazel's toolchains in a multi-repo situation to know if those would be accessible from here.

keith: ah yea, those wouldn't from here

gregestren:
> @gregestren it seems like the github owners aren't working correctly since we didn't get cc'd here?

Github is reporting errors on https://github.com/bazelbuild/bazel/pull/14614/files. @aiuto can you diagnose?

gregestren:
I'm not sure if you want a merge now or if the comments above are still pending. I'm happy to merge on the signal.

keith:
I think we can merge now, this is an improvement as is, we might need to make more improvements in the future, but without this change it's far off in a lot of places

brentleyjones:
@bazel-io fork 5.1

PR #14857, opened 2022-02-17
Set keywords on appropriate lifecycle events.

benjaminp:
The docs for the PublishLifecycleEventRequest say that the notification_keywords field should be set if the build event is InvocationAttemptStarted or BuildEnqueued. However, Bazel did not conform to this spec.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceProtoUtil.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceProtoUtil.java
@@ -161,6 +161,14 @@ public final class BuildEventServiceProtoUtil {
     if (projectId != null) {
       builder.setProjectId(projectId);
     }
+    switch (lifecycleEvent.getEventCase()) {
+      case BUILD_ENQUEUED:
+      case INVOCATION_ATTEMPT_STARTED:
+        builder.addAllNotificationKeywords(getKeywords());
+        break;
+      default:
+        break;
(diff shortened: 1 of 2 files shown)


Yannic:
cc @lfpino @ulfjack

lfpino:
@michaeledgar could you please take a look?

lfpino (approved):

src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceProtoUtil.java
```diff
@@ -161,6 +161,12 @@ public PublishBuildToolEventStreamRequest publishBuildToolEventStreamRequest(
     if (projectId != null) {
       builder.setProjectId(projectId);
     }
+    switch (lifecycleEvent.getEventCase()) {
+      case BUILD_ENQUEUED:
+      case INVOCATION_ATTEMPT_STARTED:
+        builder.addAllNotificationKeywords(getKeywords());
+        break;
```
meisterT, line 168: please add an explicit `default: // fall through` here

benjaminp: Done, though I assumed a `break;` would be preferred over a fall-through comment.

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #14872, opened 2022-02-20
Test remove android stub toolchain

keith:
I'm curious if the hard NDK requirement changes makes this no longer
necessary

Previous context: https://github.com/bazelbuild/bazel/pull/13449


keith:
https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/10364#03367f1d-7924-43ed-b568-c0891ecd185b

PR #14873, opened 2022-02-20
Add archiver flags to cpp crosstool config

AlessandroPatti:
There are three changes proposed in this pr that do not alter the current behaviour but that allow more flexibility when defining a toolchain:
  - Move default macos archiver flags from [CppActionConfigs.java#L622](https://github.com/bazelbuild/bazel/blob/feea781b30788997c0b97ad9103a13fdc3f627c8/src/main/java/com/google/devtools/build/lib/rules/cpp/CppActionConfigs.java#L622) to the crosstools definition.
  - Add `archive_flag` attribute to allow passing additional flags to the archive action.
  - Add libtool feature to switch between macos `libtool`'s flags and `ar`'s.

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -977,6 +977,26 @@ def _impl(ctx):
                         expand_if_available = "output_execpath",
                     ),
                 ],
+                with_features = [
+                    with_feature_set(
+                        not_features = ["libtool"],
+                    ),
+                ],
+            ),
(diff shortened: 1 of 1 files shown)


AlessandroPatti:
@oquenchil @buildbreaker2021 any thoughts on this?

buildbreaker2021:
I will have a good look at it today or tomorrow.

keith:
what's the benefit in using ar instead of libtool here?

buildbreaker2021 (approved):

PR #14878, opened 2022-02-21
Let RunfilesTreeUpdater use OutputService.createSymlinkTree()

EdSchouten:
The canonical way of creating runfiles directories during a build is
through SymlinkTreeStrategy. That code already has logic for calling
into OutputService to do bulk creating of symbolic links, if one is
available. If not, it calls into the build-runfiles helper program.

When --nobuild_runfile_links is provided and no-sandbox actions are
executed, an alternative mechanism for creating runfiles directories is
used, namely the one in RunfilesTreeUpdater. This one is seemingly only
capable of using build-runfiles; it cannot use OutputService.

This change makes RunfilesTreeUpdater more complete by also considering
the use of OutputService. This tends to be faster, as implementations of
OutputService can generally do batch creation.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/exec/BUILD
@@ -182,6 +182,7 @@ java_library(
     srcs = ["RunfilesTreeUpdater.java"],
     deps = [
         ":bin_tools",
+        ":spawn_input_expander",
         ":symlink_tree_helper",
         "//src/main/java/com/google/devtools/build/lib/actions",
         "//src/main/java/com/google/devtools/build/lib/actions:artifacts",
@@ -189,6 +190,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/lib/profiler",
         "//src/main/java/com/google/devtools/build/lib/util/io:out-err",
         "//src/main/java/com/google/devtools/build/lib/vfs",
+        "//src/main/java/com/google/devtools/build/lib/vfs:output_service",
         "//src/main/java/com/google/devtools/build/lib/vfs:pathfragment",
         "//third_party:guava",
         "//third_party:jsr305",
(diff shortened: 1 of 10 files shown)


EdSchouten:
Friendly ping!

EdSchouten:
@janakdr, I see that you've made changes to `src/main/java/com/google/devtools/build/lib/exec/RunfilesTreeUpdater.java` in the past. Would you be the right person to review this?

janakdr:
Sorry, I'm no longer working on Bazel.

alexjski (changes requested):
I think this is a great idea! I have couple of high-level questions here:

1. Did you look into trying to share more of the code with `SymlinkTreeStrategy`. Admittedly, I just looked briefly at that, it looks like there is some overlap there (please feel free to tell me to dig more into that to give more concrete).
2. Did you consider adding some test coverage for that? I think an integration test with a mocked outputService could be a way to go.

EdSchouten:
Hi Alex,

Sorry for the late reply.

> 1. Did you look into trying to share more of the code with `SymlinkTreeStrategy`. Admittedly, I just looked briefly at that, it looks like there is some overlap there (please feel free to tell me to dig more into that to give more concrete).

That's a good question. I just spent an hour or two juggling code around, but in the end I came to the conclusion it's not as straight-forward/elegant as I would like:
- When `RunfilesTreeUpdater` and `SymlinkTreeStrategy` need to create the symlink tree using a command, the logic is identical, which is why it can call into `SymlinkTreeHelper`.
- When `RunfilesTreeUpdater` and `SymlinkTreeStrategy` need to create the symlink tree directly or using the output service, the logic to compute the set of symlinks is quite different. The former operates on `ActionInput`s, while the latter uses `Artifact`s.

Considering that the logic for computing the set of symlinks is far more complex than calling into the output service, there isn't a lot that can be factored out.

> 2. Did you consider adding some test coverage for that? I think an integration test with a mocked outputService could be a way to go.

The issue right now is that Bazel Open Source does not ship with any implementations of OutputService that can talk to an external process. My plan is to polish up #12823 after this PR lands, which should make that easier, I guess.

meisterT:
I assume this is obsolete by now that we properly support a Bazel output service

PR #14882, opened 2022-02-22
Fix broken Tables and Images on android tutorial

rifqimfahmi:
- Tables not formatted properly https://bazel.build/tutorials/android-app. for example:
image

- ~~Image path seems to be not pointing to the correct asset path but renders fine on the website but is broken when you preview it on GitHub https://bazel.build/tutorials/android-app. anyway, I adjust the image path.~~

### DIFF
--- a/site/en/tutorials/android-app.md
+++ b/site/en/tutorials/android-app.md
@@ -79,7 +79,7 @@ Take a look at the source files for the app.
                         ├── MainActivity.java
                         └── res
                             ├── layout
-                            │   └── activity_main.xml
+                            │   └── activity_main.xml
                             └── values
                                 ├── colors.xml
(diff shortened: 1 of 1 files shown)


gregestren:
Sorry for confusion, but which preview do you mean? I'm more confused than normal because of the recent changing of the docs backend.

fweikert:
The image URLs have to be relative to en/, otherwise the website will break. The incompatibility with GitHub's preview feature is annoying, but I'm afraid that we have to live with it.

rifqimfahmi:
I agree it is a little bit annoying. Anyway, I adjusted the changes to just fix the table if it's still relevant

fweikert (approved):

Issue #14890, opened 2022-02-23
Python zip action is memory hungry even if zipping not used - 80% of heap in our case


glukasiknuro:
### Description of the problem / feature request:

While investigating memory usage during analysis phase noticed that py_binary and py_test rules were responsible for majority of memory usage in analysis phase. Tracked it down to zip action that is always created - irrespective whether `--build_python_zip` is used, or what platform it runs on.

Looks like the below code iterates over all runfiles and creates strings for those, if using a lot of dependencies this adds on pretty quickly:

https://github.com/bazelbuild/bazel/blob/841c04a8b42a5e25bb3b6d5e4d432a3d29c3e6fd/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L370

Commenting out the above code caused 80% reduction of heap usage in our case as reported by `bazel info used-heap-size-after-gc` after building codebase with `--nobuild` option. Even simply calling `.intern()` on the created String reduced heap usage by 70%, but it maybe very code-base dependent.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I can create a reproduction if necessary using rules_python and reasonable number of dependencies, let me know whether is useful.

### What operating system are you running Bazel on?

Ubuntu 20.04

### What's the output of `bazel info release`?

bazel 5

glukasiknuro:
I was able to reproduce the problem on part of tensorflow codebase, but there are additional factors to it.

Setup:
```
git clone git@github.com:tensorflow/tensorflow.git
```

Using https://bazel.build/rules/performance#enabling-memory-tracking

```
bazel ${STARTUP_FLAGS} build --nobuild //tensorflow/python/...
```

**Released bazel 5**
```
$ bazel ${STARTUP_FLAGS} dump --rules | grep py_
py_test                              1,654      14,967    380,780,088      230,217  
py_library                           1,085           0      1,048,712          966  
py_binary                               32         192      6,030,736      188,460  

$ bazel ${STARTUP_FLAGS} info used-heap-size-after-gc
626MB
```

**Bazel 5 with change to intern strings**
```
$ bazel ${STARTUP_FLAGS} dump --rules | grep py_
py_test                              1,654      14,967     65,358,744       39,515  
py_library                           1,085           0      1,572,952        1,449  
py_binary                               32         192        786,480       24,577  

$ bazel ${STARTUP_FLAGS} info used-heap-size-after-gc
289MB
```

[...]

PR #14892, opened 2022-02-23
Intern strings for python zip runfiles

glukasiknuro:
Provides substantial memory savings for repos using heavily python dependencies.

Possibly even better change could be created that avoids creating those strings, but this small change by itself provides most of the benefits without to much risk. 

See  https://github.com/bazelbuild/bazel/issues/14890


src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
```diff
@@ -370,9 +370,10 @@ private static void createPythonZipAction(
     for (Artifact artifact : runfilesSupport.getRunfilesArtifacts().toList()) {
       if (!artifact.equals(executable) && !artifact.equals(zipFile)) {
         argv.addDynamicString(
-            getZipRunfilesPath(artifact.getRunfilesPath(), workspaceName, legacyExternalRunfiles)
-                + "="
-                + artifact.getExecPathString());
+            (getZipRunfilesPath(artifact.getRunfilesPath(), workspaceName, legacyExternalRunfiles)
+                    + "="
+                    + artifact.getExecPathString())
+                .intern());
```
comius, line 376: You should use BlazeInterners.weakInterner for this, like the rest of the codebase. https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/concurrent/BlazeInterners.java;l=38;drc=ec8fb7969614c6b3752392089e2310708d103225

Interning is not available in Starlark, so this will have to be optimised in a different way.

Even better would be if you replaced call `argv.addDynamicString` with one of the calls that accepts artifact.getExecPath() as an argument and set format to `getZipRunfilesPath(artifact.getRunfilesPath(), workspaceName, legacyExternalRunfiles) + "="`.  

Can you also provide some benchmarking results? baseline vs the change?
I'm mostly interested in used-heap-size-after-gc displayed by `bazel info` after building the project. (You should do 1 warm-up run before collecting the results).

Unfortunately I don't have appropriate project to run benchmarks for you.

comius (changes requested):

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
comius: Nice results!

Those can be better + Starlarkifiable. Using lazy argument construction, Bazel wouldn't retain any of those strings.

Change the whole for loop to something like:
`argv.addAll(VectorArg.of(runfilesSupport.getRunfilesArtifacts()).mapped(MAP_FN))`

where
```
MAPFN = (artifact, args) ->
{ 
if (!artifact.equals(executable) && !artifact.equals(zipFile)) {
          args.accept(getZipRunfilesPath(artifact.getRunfilesPath(), workspaceName, legacyExternalRunfiles)
                    + "="
                    + artifact.getExecPathString());
     }
}
```

Example:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/java/OneVersionCheckActionBuilder.java;l=121-128;drc=103ea9d3f3a0d85a0d7c1ad98fd8dbb422505d95

comius:
Friendly ping.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
glukasiknuro: Implemented second solution by adding `CommandLineItem.DefinedParamsCacheMapFn` , PTAL

glukasiknuro:
@comius friendly ping. Do you think we can proceed with this PR?

src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java
comius: I removed `CapturingMapFn`, because it was unused. I think the solution using it would be nice. Let me rollback this removal.

comius: A sec, before I rollback removal of `CapturingMapFn`.

> Using ParametrizedMapFn requires specifying INT_MAX in maxInstancesAllowed

When implementing this in Starlark it uses ParametrizedMapFn with maxInstancesAllowed set to Integer.MAX_VALUE. See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkCustomCommandLine.java;l=893-976;drc=8f0034ce1e7854521e259a5eaf71859b1e6f95bc

The requirement in ParametrizedMapFn is:
```
  * <p>The user promises that the number of distinct instances constructed is closer to O(rule
   * class count) than O(rule count).
```

In any case, I'd prefer if we don't add new things to `CommandLineItem` and `NestedSetFingerprintCache`

comius: Rolled back removal of `CapturingMapFn` in https://github.com/bazelbuild/bazel/commit/c70e0721f0ed506fad62ab89561804a60ab9b976

comius: > @comius friendly ping. Do you think we can proceed with this PR?

There's a second PR that handles this with `CapturingMapFn`: https://github.com/bazelbuild/bazel/pull/15037

comius:
Since the second PR (https://github.com/bazelbuild/bazel/pull/15037) has been merged I'll close this one.

PR #14898, opened 2022-02-24
Configure Apple crosstool to return a complete target triple that includes minimum OS version and target environment

thii:
This is required for https://github.com/bazelbuild/rules_swift/pull/768

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
         return ["-D"]
     return []
 
+def _target_os_version(ctx):
+    platform_type = ctx.fragments.apple.single_arch_platform.platform_type
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    return xcode_config.minimum_os_for_platform_type(platform_type)
+
 def _impl(ctx):
+    target_os_version = _target_os_version(ctx)
+
     if (ctx.attr.cpu == "ios_arm64"):
-        target_system_name = "arm64-apple-ios"
+        target_system_name = "arm64-apple-ios{}".format(target_os_version)
     elif (ctx.attr.cpu == "tvos_arm64"):
-        target_system_name = "arm64-apple-tvos"
+        target_system_name = "arm64-apple-tvos{}".format(target_os_version)
     elif (ctx.attr.cpu == "watchos_arm64_32"):
-        target_system_name = "arm64_32-apple-watchos"
+        target_system_name = "arm64_32-apple-watchos{}".format(target_os_version)
     elif (ctx.attr.cpu == "ios_arm64e"):
-        target_system_name = "arm64e-apple-ios"
+        target_system_name = "arm64e-apple-ios{}".format(target_os_version)
     elif (ctx.attr.cpu == "ios_armv7"):
-        target_system_name = "armv7-apple-ios"
+        target_system_name = "armv7-apple-ios{}".format(target_os_version)
     elif (ctx.attr.cpu == "watchos_armv7k"):
-        target_system_name = "armv7k-apple-watchos"
+        target_system_name = "armv7k-apple-watchos{}".format(target_os_version)
     elif (ctx.attr.cpu == "ios_i386"):
-        target_system_name = "i386-apple-ios"
+        target_system_name = "i386-apple-ios{}-simulator".format(target_os_version)
     elif (ctx.attr.cpu == "watchos_i386"):
-        target_system_name = "i386-apple-watchos"
+        target_system_name = "i386-apple-watchos{}-simulator".format(target_os_version)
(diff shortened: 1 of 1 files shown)


tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
         return ["-D"]
     return []
 
+def _target_os_version(ctx):
+    platform_type = ctx.fragments.apple.single_arch_platform.platform_type
+    xcode_config = ctx.attr._xcode_config[apple_common.XcodeVersionConfig]
+    return xcode_config.minimum_os_for_platform_type(platform_type)
+
 def _impl(ctx):
+    target_os_version = _target_os_version(ctx)
```
keith, line 64: continuing the convo from the other PR, I guess I would assume this was only fetched at toolchain setup time, and thus wouldn't differ between targets?

thii: This is a normal rule, not a repo rule, so this value is evaluated at build time.

thii: I think the toolchain setup just symlinks this file to the `local_config_cc` repo.

thii: I have tested a Bazel binary built from this patch with all unit tests in rules_swift with the above mentioned rules_swift change.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
...
+        target_system_name = "arm64-apple-ios{}-simulator".format(target_os_version)
     elif (ctx.attr.cpu == "tvos_sim_arm64"):
-        target_system_name = "arm64-apple-tvos-simulator"
+        target_system_name = "arm64-apple-tvos{}-simulator".format(target_os_version)
     elif (ctx.attr.cpu == "watchos_arm64"):
-        target_system_name = "arm64-apple-watchos-simulator"
+        target_system_name = "arm64-apple-watchos{}-simulator".format(target_os_version)
     elif (ctx.attr.cpu == "darwin_x86_64"):
-        target_system_name = "x86_64-apple-macosx"
+        target_system_name = "x86_64-apple-macosx{}".format(target_os_version)
```
thii, line 91: Both `macos` and `macosx` work here. But the expanded `-triple` flag use `macosx` so I just keep it as is.

tools/osx/crosstool/cc_toolchain_config.bzl
keith: ah my bad

keith (approved):
this matches what xcode does today

keith:
cc @oquenchil can you review?

thii:
@googlewalt Can you review?

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -55,41 +55,48 @@ def _deterministic_libtool_flags(ctx):
...
+        target_system_name = "arm64-apple-macosx{}".format(target_os_version)
     elif (ctx.attr.cpu == "darwin_arm64e"):
-        target_system_name = "arm64e-apple-macosx"
+        target_system_name = "arm64e-apple-macosx{}".format(target_os_version)
     elif (ctx.attr.cpu == "tvos_x86_64"):
-        target_system_name = "x86_64-apple-tvos"
+        target_system_name = "x86_64-apple-tvos{}-simulator".format(target_os_version)
     elif (ctx.attr.cpu == "watchos_x86_64"):
-        target_system_name = "x86_64-apple-watchos"
+        target_system_name = "x86_64-apple-watchos{}-simulator".format(target_os_version)
```
googlewalt, line 99: I think the ability to specify os version in the triple was added in Xcode 9.3, so this change will in theory break for Xcodes preceding that.  Are the OSS rules_apple/swift maintainers ok with that (@keith, etc.)?

Separately, https://bazel.build/install/os-x references Xcode 6.1 and ios_* rules so seems very dated.  Can we replaced with a link to the rules_apple repository that says "go here to see the current platform requirements", and have the OSS maintainers of that repo to keep a section of their README updated with whatever the rules support?

keith: I'm fine with dropping Xcode 9 support, according to https://developer.apple.com/news/?id=ib31uj1j you have to be using 12 for the app store, and I can't imagine there are a ton of non-app store cases out there

keith: +1 for deferring docs over to the rules, since most folks using this are likely using the rules as well, can probably be done separately

googlewalt: Ok I'll start the merge.

googlewalt (approved):

brentleyjones:
@bazel-io fork 5.2

PR #14824, opened 2022-02-15
Add removeprefix/removesuffix to Starlark strings

fmeum:
Implements https://github.com/bazelbuild/starlark/issues/185 in Java Starlark.

PR #14899, opened 2022-02-24
Add removeprefix/removesuffix to Starlark strings

fmeum:
Implements https://github.com/bazelbuild/starlark/issues/185 in Java Starlark.

Closes #14824.

PiperOrigin-RevId: 430556229

### DIFF
--- a/src/main/java/net/starlark/java/eval/StringModule.java
+++ b/src/main/java/net/starlark/java/eval/StringModule.java
@@ -1022,4 +1022,36 @@ final class StringModule implements StarlarkValue {
   private static boolean substringStartsWith(String str, int start, int end, String prefix) {
     return start + prefix.length() <= end && str.startsWith(prefix, start);
   }
+
+  @StarlarkMethod(
+      name = "removeprefix",
+      doc =
+          "If the string starts with <code>prefix</code>, returns a new string with the prefix "
+              + "removed. Otherwise, returns the string.",
+      parameters = {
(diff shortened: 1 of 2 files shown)


fmeum:
@Wyverald In general, do you want cherry-pick commits to add `RELNOTES`? Also, if possible, you might want to disable code owners for non-master branches, as otherwise reviewers will get pinged for all these cherry picks.

Wyverald:
> In general, do you want cherry-pick commits to add RELNOTES?

It's probably fine since most minor releases get release notes from entire commit messages, instead of relnotes. The number of cherry-pick commits in a minor release is very manageable. So let's so "no" for now.

> Also, if possible, you might want to disable code owners for non-master branches, as otherwise reviewers will get pinged for all these cherry picks.

I can't find a setting that does that. Please let me know if I'm missing something!

fmeum:
@Wyverald You would have to delete the `CODEOWNERS` on the release branch or, even better, replace its content with a wildcard that always assigns the current release manager. See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location for the details.

Wyverald:
Hmm, this idea has been brought up before. I guess we should just replace `CODEOWNERS` with the release manager wildcard as the first commit in each `release-X` branch. There shouldn't be any downsides. Thanks for the suggestion!

Issue #13817, opened 2021-08-08
config.string with allow_multiple=True should use list for build_setting_default


Yannic:
I'm trying to migrate a native flag (`--protocopt`) to a Starlark flag. For this, we need a flag that a) can be supplied multiple times (like with `allow_multiple=True` - `config.string_list` doesn't allow this) and b) has an empty default.

Today, it's impossible to get an empty list or a list with multiple elements as `ctx.build_setting_value` when using `config.string` with `allow_multiple=True`.

```starlark
# foo.bzl

def _impl(ctx):
    pass

foo = rule(
    implementation = _impl,
    build_setting = config.string(flag = True, allow_multiple = True),
)

# BUILD
load(":foo.bzl", "foo")

foo(name="foo", build_setting_default="")  # <-- `ctx.build_setting_value` becomes `[""]`.
```

There are ways to work around this (i.e. setting `build_setting_default` to a nonsensical value and checking for `ctx.build_setting_value` with one element that is the specified default value), but that seems unnecessarily complex.

From a technical perspective, changing the allowed type for `build_setting_default` when `allow_multiple=True` looks pretty straight forward to me.

`allow_multiple` was added in https://cs.opensource.google/bazel/bazel/+/a13f590b69bcbcaa10b1a49bfd9a4607dfbd8f47 after cutting Bazel 4.0.0, so its usage in the wild should be low.

fmeum:
@gregestren What would you think of allowing both "list of <type>" and "<type>" as default values, with the latter continuing to be converted to a singleton list for backwards compatibility? Since there is already a `string_list` type, having the `BuildSetting` return a type that dynamically depends on the value of its `allowMultiple` member may actually not be that much work.

fmeum:
I ultimately came to the conclusion that the source of issues is that repeatable string flags should be represented as `string_list` flags with a different CLI argument passing style rather than `string` flags that conditionally behave more like `string_list` flags. I created #14911 to create a replacement for the current `allow_multiple` flag that requires significantly less special casing internally and should also be more transparent to users.

gregestren:
Thanks. I'll take a look. Also cc @aranguyen FYI who's been looking recently at flag parsing logic in general.

aranguyen:
@fmeum  thanks for sending out the proposal. I don't know how to make comments on the proposal you linked so I'm dropping them here instead.
[...]

PR #14908, opened 2022-02-24
Allow literal default for allow_multiple build setting in transitions

### LINKED ISSUE #14894: Build settings defined with `allow_multiple = True` cannot pass their default values to transition inputs without also being outputs

`foo.bzl`
```Starlark

def _s_impl(ctx):

def _t_impl(settings, attr):
    return {"//:b": ["bar"]}
  else:
    return {"//:b": ["baz"]}

def _r_impl(ctx):

s = rule(
  implementation = _s_impl,
)

t = transition(
  implementation = _t_impl,
  inputs = ["//:a"],
  outputs = ["//:b"],
)

r = rule(
  implementation = _r_impl,
  attrs = {
  },
)
```

`BUILD`
```Starlark
load(
  "foo.bzl",
  "s",
  "r",
)

s(
  name = "a",
)

s(
  name = "b",
)

r(
  name = "c",
  setting = ":b"
)
```

release 5.0.0

N/A

fmeum:
Instead of requiring the new value for a build setting with
allow_multiple = True specified by a transition to always be a list, we
make an exception for the literal default value for two reasons:

1. It ensures that specifying a setting with allows multiple as an input
   but not as an output of a transition does not run into the exception
   below.
2. It provides a clear way to reset the setting to its (literal) default.

Fixes #14894 

RELNOTES: Fixed a bug that prevented build settings with
`allow_multiple = True` from appearing only as inputs but not as
outputs of a transition.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
@@ -329,11 +329,20 @@ public abstract class StarlarkTransition implements ConfigurationTransition {
         if (allowsMultiple) {
           // if this setting allows multiple settings
           if (!(newValue instanceof List)) {
-            throw new TransitionException(
-                String.format(
-                    "'%s' allows multiple values and must be set"
-                        + " in transition using a starlark list instead of single value '%s'",
-                    actualSetting, newValue));
+            if (newValue.equals(rule.getAttr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME))) {
+              // We make an exception for the default value here for two reasons:
+              // 1. It ensures that specifying a setting with allow_multiple as an input but not as
+              //    an output of a transition does not run into the exception below.
+              // 2. It provides a clear way to reset the setting to its (literal) default from a
+              //    transition.
+              newValue = ImmutableList.of(newValue);
+            } else {
+              throw new TransitionException(
+                  String.format(
+                      "'%s' allows multiple values and must be set"
+                          + " in transition using a starlark list instead of single value '%s'",
+                      actualSetting, newValue));
+            }
           }
           List<?> rawNewValueAsList = (List<?>) newValue;
           List<Object> convertedValue = new ArrayList<>();
--- a/src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java
+++ b/src/test/java/com/google/devtools/build/lib/analysis/StarlarkAttrTransitionProviderTest.java
@@ -2549,4 +2549,61 @@ public final class StarlarkAttrTransitionProviderTest extends BuildViewTestCase 
         new EventBus());
     assertNoEvents();
   }
+
+  @Test
+  public void testTransitionPreservesAllowMultipleDefault() throws Exception {
+    writeAllowlistFile();
+    scratch.file(
+        "test/starlark/rules.bzl",
+        "P = provider(fields = ['value'])",
+        "def _s_impl(ctx):",
+        "    return [P(value = ctx.build_setting_value)]",
+        "def _t_impl(settings, attr):",
+        "    if 'foo' in settings['//test/starlark:a']:",
+        "        return {'//test/starlark:b': ['bar']}",
+        "    else:",
+        "        return {'//test/starlark:b': ['baz']}",
+        "def _r_impl(ctx):",
+        "    pass",
+        "s = rule(",
+        "    implementation = _s_impl,",
+        "    build_setting = config.string(allow_multiple = True, flag = True),",
+        ")",
+        "t = transition(",
+        "    implementation = _t_impl,",
+        "    inputs = ['//test/starlark:a'],",
+        "    outputs = ['//test/starlark:b'],",
+        ")",
+        "r = rule(",
+        "    implementation = _r_impl,",
+        "    attrs = {",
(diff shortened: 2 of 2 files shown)


Wyverald:
@gregestren @lberki could we get this looked at? This potentially fixes a 5.0 regression and it would great to have it in 5.1.

gregestren:
@Wyverald there are a couple of PRs and issues (at least 4) related to `allow_multiple`.  I'm trying to wrap my head around them. It's on my agenda but that's why I haven't responded yet.

I'll comment on them by today but I'm not ready as of this message to advocate any specific step forward.

gregestren:
To be clear:

- https://github.com/bazelbuild/bazel/issues/13817
- https://github.com/bazelbuild/bazel/issues/14894
- https://github.com/bazelbuild/bazel/pull/14908 (this)
-  https://github.com/bazelbuild/bazel/pull/14911

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -329,11 +329,20 @@ private static Target getActual(
         if (allowsMultiple) {
           // if this setting allows multiple settings
           if (!(newValue instanceof List)) {
-            throw new TransitionException(
-                String.format(
-                    "'%s' allows multiple values and must be set"
-                        + " in transition using a starlark list instead of single value '%s'",
-                    actualSetting, newValue));
+            if (newValue.equals(rule.getAttr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME))) {
```
gregestren, line 332: It really took me some time to realize `changedSettingToRule` from line 310 is now a misleading name.  As I understand, the reality is:

1. `changedSettingToRule` was originally written to encapsulate settings the transition changed
2. `build_setting` inputs are implicitly changed by transition code as an implementation detail, from non-existent to their explicit default values, to guarantee the transition can read them
3. https://github.com/bazelbuild/bazel/pull/13971 causes the original logic, which historically only checked user-changed settings, to also include these implicit changes. For the purpose of guaranteeing the default that the transition added are then removed.

I'm worried about these overloading of concepts, even if it keeps the code concise.

What if we broke out the logic some: revert https://github.com/bazelbuild/bazel/pull/13971/files, then do a separate distinct call to `getRelevantStarlarkSettingsFromTransition(Settings.INPUTS)` which *only* replicates the part of `validate` that removes an option if it's at its default value?

I realize that takes more refactoring, likely breaking out a helper method for `validate()`. But I'd really like to keep the code as clear as possible in transition logic, since it's super-complicated as it is.

At the very leas I'd like the naming and commenting (including method Javadoc) to remain accurate. I worry it could be more work to clearly make the commenting accurate than do the refactoring I'm suggesting.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -329,11 +329,20 @@ private static Target getActual(
...
+              // We make an exception for the default value here for two reasons:
+              // 1. It ensures that specifying a setting with allow_multiple as an input but not as
+              //    an output of a transition does not run into the exception below.
+              // 2. It provides a clear way to reset the setting to its (literal) default from a
```
gregestren, line 336: 2. here feels a bit API-shaky to me. Basically, a transition has to set a setting to a certain type unless it's set to a certain special value in which case it's exempt. I get the intended use case described here. But I can also see transitions that, say, read their values from attributes, which happen to set values to defaults - or not - by coincidence. i.e. whoever's setting the values has no insight into this and may be confused why some arbitrary string works while another fails.

To the point, do you think https://github.com/bazelbuild/bazel/pull/14911 obsoletes all this complexity?

gregestren (commented):
Generally speaking, how much of this PR remains relevant after https://github.com/bazelbuild/bazel/pull/14911?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
fmeum: That's a very good point - input build settings should be cleared unconditionally rather than undergo validation. I created https://github.com/bazelbuild/bazel/pull/14972 for this refactoring and a better fix for #14894.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
fmeum: I agree it looks very hacky and is another reason to consider #14972 the better kind of fix. That said, I don't think this accidental resetting would be very realistic: The "magic" literal default value is the only valid one of its type (string rather than list of string) due to type validation and any attribute would only ever yield one type.

> To the point, do you think https://github.com/bazelbuild/bazel/pull/14911 obsoletes all this complexity?

The new representation of repeated string flags would no longer require any special handling for transitions, so we could eventually delete the entire `allowsMultiple` branch. But given that finalizing the design of #14911 will take some more time, either this PR or #14972 are probably needed to deal with the regression for 5.1.0.

gregestren: Agreed on your last point - I'm happy to fast track a temporary fix for the regression. I just wanted to verify the long-term expected state.

fmeum:
Superseded by #14972.

PR #14912, opened 2022-02-25
Fixes typo in toolchains documentation page

alloveras:
## Intent
To fix what I believe it's a typo in `site/en/docs/toolchains.md`.

### DIFF
--- a/site/en/docs/toolchains.md
+++ b/site/en/docs/toolchains.md
@@ -450,7 +450,7 @@ with preference given to earlier items in the list.
 The resolution steps are as follows.
 
 1. A `target_compatible_with` or `exec_compatible_with` clause *matches* a
-   platform iff, for each `constraint_value` in its list, the platform also has
+   platform if, for each `constraint_value` in its list, the platform also has
    that `constraint_value` (either explicitly or as a default).
 
    If the platform has `constraint_value`s from `constraint_setting`s not


site/en/docs/toolchains.md
```diff
@@ -450,7 +450,7 @@ with preference given to earlier items in the list.
 The resolution steps are as follows.
 
 1. A `target_compatible_with` or `exec_compatible_with` clause *matches* a
-   platform iff, for each `constraint_value` in its list, the platform also has
+   platform if, for each `constraint_value` in its list, the platform also has
```
gregestren, line 453: The original was a shorthand for "if and only if" (https://math.stackexchange.com/questions/566565/are-if-and-iff-interchangeable-in-definitions), which is logically more precise.

I think for the purposes of user documentation `if` is reasonable here though: less conceptual overhead and I don't think we need such logical precision to get the point across.

gregestren (approved):

site/en/docs/toolchains.md
alloveras: Oh, I wasn't aware that _iff_ was shorthand for _if and only if_. 

I saw that you approved the PR already as it is right now but, if you feel strongly about it, I am happy to either:
- Revert it to back _iff_
- Change it to _if and only if_

gregestren: My approval was an endorsement. :) 

Basically, I don't think we need to be that logically pedantic here. *if* makes just as much sense, without the confusion and extra mental processing *iff* requires. A win for clear documentation in my view.

Issue #14866, opened 2022-02-18
http_archive reporting 404 Not Found instead of 302 Redirect


tali:
### Description of the problem:

Bazel `http_archive()` does not follow a HTTPS redirect.
It complains about a _404 File not found_ while `wget` sees a _302 Found_ and then follows to the redirected URL.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Try to build Gerrit when not all resources are in the cache.

### What operating system are you running Bazel on?

Debian Linux (testing), but seeing the same problem on MacOS.

### What's the output of `bazel info release`?

```
INFO: Invocation ID: 0439b56f-c214-4841-8aa1-34bd32721cc8
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=169
INFO: Reading rc options for 'info' from /home/martin/src/gerrit/.bazelrc:
  Inherited 'build' options: --workspace_status_command=python3 ./tools/workspace_status.py --repository_cache=~/.gerritcodereview/bazel-cache/repository --action_env=PATH --disk_cache=~/.gerritcodereview/bazel-cache/cas --java_language_version=11 --java_runtime_version=remotejdk_11 --tool_java_language_version=11 --tool_java_runtime_version=remotejdk_11 --incompatible_strict_action_env --announce_rc
release 5.0.0
```
### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```
https://gerrit.googlesource.com/gerrit
547c9e46b9c81fb0192e82dc73d085cfa182cd88
547c9e46b9c81fb0192e82dc73d085cfa182cd88
```

###  Have you found anything relevant by searching the web?

Similar bug, but http->https redirect: 
* #527: https://groups.google.com/g/bazel-discuss/c/8BZvhwGQsjg/m/b7OreBOcAAAJ

### Any other information, logs, or outputs that you want to share?

```sh
$ bazel build //:gerrit
INFO: Invocation ID: c07bac0d-a149-4de0-968c-65b179dc2364
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=169
INFO: Reading rc options for 'build' from /home/martin/src/gerrit/.bazelrc:
[...]

PR #14922, opened 2022-03-01
Avoid merging URLs in HttpUtils

Bencodes:
`mergeUrls` does not need to rebuild the URL from scratch if user information exists on the original URL. This behavior can actually break the 302 redirect due to subtle changes in the URL/encoding and should be avoided when possible.

This fixes https://github.com/bazelbuild/bazel/issues/14866 by correcting the implementation of `mergeUrls` to match the documentation that was added instead of rebuilding the URL from scratch which breaks the encoding of signed URLs.


Wyverald (approved):

brentleyjones:
@bazel-io fork 5.1

PR #14924, opened 2022-03-01
Normalize rpath entries to guard against missing default solib dir

fmeum:
When all dynamic deps in a build are built in transitioned configurations, the default solib dir is not created. However, while
resolving paths, the dynamic linker stops at the first directory that does not exist, even when followed by `../`.

Before this commit, all rpath entries would consist of the relative path to the default solib dir followed by the relative path to the
particular library's solib dir. Thus, if the default solib dir was missing, the dynamic linker wouldn't resolve any of these paths.

This commit ensures that the relative path entries are normalized and thus contain no references to non-existing directories assuming the normalized path itself exists.

Work towards #13819.

Closes #14660.

PiperOrigin-RevId: 431671888

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
@@ -331,8 +331,13 @@ public class LibrariesToLinkCollector {
         commonParent = commonParent.getParentDirectory();
       }
 
-      rpathRootsForExplicitSoDeps.add(
-          rpathRoot + dotdots + libDir.relativeTo(commonParent).getPathString());
+      // When all dynamic deps are built in transitioned configurations, the default solib dir is
+      // not created. While resolving paths, the dynamic linker stops at the first directory that
+      // does not exist, even when followed by "../". We thus have to normalize the relative path.
+      String relativePathToRoot =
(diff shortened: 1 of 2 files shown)


fmeum:
@Wyverald Not sure whether this is too late to get into 5.1.0. What do you think?

oquenchil:
This was merged 7 minutes ago: https://github.com/bazelbuild/bazel/pull/14660

Why this PR?

fmeum:
@oquenchil Sorry, you should not have gotten pinged about this: This is not a PR against `master` but rather against the branch intended for cherry picks into 5.1.0rc1. Unfortunately, `CODEOWNERS` is the same as on `master` for this branch, which means that the same reviewers will get pinged again even though only the release manager, in this case @Wyverald, should have to be notified. I discussed this with him over at https://github.com/bazelbuild/bazel/pull/14899#issuecomment-1049803462 and we found a workaround, so there shouldn't be any redundant pings starting with the next release.

Sorry again for the redundant ping and thank you very much for reviewing the original PR!

Wyverald:
Yeah please send a PR against `release-5.1.0`. The `rcX` branches will be used as tags (easier to create on the GitHub UI). Sorry for the confusion!

fmeum:
Closed in favor of https://github.com/bazelbuild/bazel/pull/14929

PR #14926, opened 2022-03-01
Fix bug in BUILD.tools for jdk8

ianoc:
The DEFAULT_TOOLCHAIN_CONFIGURATION doesn't work for JDK8, uses compact string options


ianoc:
Not sure how most of the test failures listed could have been caused by this change?

ianoc:
So my understanding of whats i happening here:

- Previously for source/target version 8, we were actually getting the the options that are incompatible and only good for 9+
- Once the jdk8 toolchain gets fixed then its evaluated, `JVM8_TOOLCHAIN_CONFIGURATION` was previously for toolchain resolution purposes dead code.
- `JVM8_TOOLCHAIN_CONFIGURATION` points at @bazel_tools//tools/jdk:jdk_8, which looks for a runtime for 8. But in bazel these aren't configured (anymore?). So the toolchain resolution fails.

It may not be the prettiest, but by default i've set the java_runtime to be jdk11. Which is what it was previously, but with jdk8 options. I think this should be ~enough for anyone who has setup remote_repositories for jdk8, and sets the runtime version when they've done that.

sgowroji:
Hello @ianoc, Can you please check the  buildkite presubmit failures. Thanks!

sgowroji:
I'm going to go ahead and close this PR, because it seems to have stalled. If you're still interested in pursing this (and responding to my comments), please feel free to reopen!

Issue #11183, opened 2020-04-22
Provide a tiny tar utility for use in tests


aiuto:
Proposal:  Create a rule(s) that can create a zip/tar archives from a set of sources. This would not be available at Bazel runtime as part of @tools (even though the source might go in //tools for lack of a better place).  The capabilities would be to bundle up a list of sources and rename some files along the way. It would be stripped down version of the rules from pkg_tar.

Need:
1. Some workspace tests often require building a tar archive as part of the test. These currently depend on pkg_tar from @tools/build_defs/pkg, which is going away (#8857 ).  https://github.com/bazelbuild/bazel/search?q=bazel_tools/tools/build_defs/pkg&unscoped_q=pkg_tar 
Rewriting them to depend on the rules_pkg distribution is possible, but introduces an external dependency buried in code. The package checksum ends up appearing in Java test code that is writing a WORKSPACE file. This is extremely ugly.
2. Building the Bazel distribution makes frequent use of pkg_tar to bundle up some things. We currently depend on bazelbuild/rules_pkg for this.  It looks like we could use the mini tool to eliminate an upstream dependency.

As an alternative, we could minimize chance of drift between rules_pkg in the build and tests by reusing the imported repo.
- re-export the rules_pkg.tar.gz that we import into the Bazel WORKSPACE. (maybe an http_file duplicating the http_archive)
- tests which need to import \@rules_pkg into the WORKSPACE files they write will add a dep on this tarball.

aiuto:
My thinking as of late is to remove all the extra features from the internal pkg_tar and let it become the mini-tar.
Users of the esoteric ones will have to do the easy migration to [rules_pkg](https://github.com/bazelbuild/rules_pkg)

aiuto:
I picked this up again for now.  Rough plan:
- PR 1: create mini_tar using the remains of pkg_tar.  Same build_tar, but few attributes.
- migrate google off of bazel's pkg_tar to rules_pkg.pkg_tar
- PR 2: drop more attributes from pkg_tar
- PR 3: delete pkg.bzl
- PR 4: consolidate archive.py into build_tar.py
- PR 5: get rid of the shell tests.

It is very back burner, so it may take a few months.

PR #14935, opened 2022-03-02
Create mini_tar and switch to it for our blackbox tests.

aiuto:
Add `@bazel_tools/build_defs/pkg/tar.bzl%mini_tar`.
This is limited functionality tar file builder. Its purpose is mainly for building tar files inside tests, where we do not want to take a dependency on the external repository @rules_pkg.

This means bazel itself does not depend on tools/build_defs/pkg/pkg.bzl at all any more, so we are free to delete that as soon as we have migrated the last of the users away.

Advances https://github.com/bazelbuild/bazel/issues/11183


aiuto:
I can do that. But it's going to take some time to detangle the test.
Also, will want to refactor the tar builder so it is a single file first, so there is less to bundle.
I'll come back to it after perf time and close this out for now to avoid clutter.

PR #14941, opened 2022-03-02
Fix a number of doc s/blaze/bazel/g

hanneskaeufler:


sgowroji:
"Blaze" text changes belongs to serverside options class, Which breaks the internal pipelines. We are closing this PR now.

PR #14357, opened 2021-11-30
Update Guava to `31.0.1`.

alexjski:
Update Guava so that we can use the new APIs like `ImmutableMap.Builder::buildOrThrow`.

The jars are pulled from Maven central:
- [guava-31.0.1-jre.jar](https://repo1.maven.org/maven2/com/google/guava/guava/31.0.1-jre/guava-31.0.1-jre.jar)
- [guava-testlib-31.0.1-jre.jar](https://repo1.maven.org/maven2/com/google/guava/guava-testlib/31.0.1-jre/guava-testlib-31.0.1-jre.jar)

PR #14942, opened 2022-03-02
Update Guava to `31.1`.

haxorz:
Main motivation is to be able to use the new
`ImmutableMap.Builder#buildKeepingLast`.

Jars were downloded from Maven Central:
* https://repo1.maven.org/maven2/com/google/guava/guava/31.1-jre/
* https://repo1.maven.org/maven2/com/google/guava/guava-testlib/31.1-jre/

This is the same as commit 5d55643, but I'm doing it as a GitHub PR
rather than a Gerrit change, so that way we can run it through the CI
pipeline and figure out why it didn't work (5d55643 was reverted in
b9bfde8).


third_party/BUILD
```diff
@@ -499,7 +499,10 @@ distrib_java_import(
 # For bootstrapping JavaBuilder
 distrib_jar_filegroup(
     name = "guava-jars",
-    srcs = ["guava/guava-31.0.1-jre.jar"],
+    srcs = [
+      "guava/failureaccess-1.0.1.jar",
```
meteorcloudy, line 503: You probably need to split this into two filegroups
```
filegroup(
    name = "guava-jar",
    srcs = ["guava/guava-31.1-jre.jar"],
)

filegroup(
    name = "guava-failureaccess-jar",
    srcs = ["guava/guava-31.1-jre.jar"],
)

distrib_jar_filegroup(
    name = "guava-jars",
    srcs = [
        ":guava-jar",
        ":guava-failureaccess-jar",
    ],
    enable_distributions = ["debian"],
)
```
Then you can add them separately above:
```
"--classpath_entry $(location //third_party:guava-jar) " +
"--classpath_entry $(location //third_party:guava-failureaccess-jar) " +
```

haxorz:
Okay, `6fbe89e` passed CI.

@meteorcloudy Is the restriction mentioned in https://github.com/bazelbuild/bazel/pull/14357#issuecomment-983026651 still a thing? If so, I will need to do 3 PRs: One to add a new `//third_party:guava-failureaccess-jar` target to `third_party/BUILD`, one to preemptively use it in `//src/test/java/com/google/devtools/build/android/desugar:desugar_guava_at_head`, and a final one to update the jars in `//third_party:guava-jars`. Then I will do 2 more PRs after that (one to update `compile.sh` and one to delete the old Guava jars). Please confirm this is what I need to do. Then I will add a comment explaining the issue `5d55643` tickled.

third_party/BUILD
haxorz: Experimentally, it seems like it's not necessary to do literally that. `6fbe89e` passes CI. Is that commit fine? And do I really need to have used `distrib_jar_filegroup` for the new `//third_party:guava-failureaccess-jar`?

haxorz: Also, if `srcs` in `:guava-jars` has more than one element then we run into the issue mentioned in `b98be959`.

meteorcloudy: I see, it looks good. Using `distrib_jar_filegroup` is for the Bazel Debian build to switch to system installed libraries easily (eg. `/usr/share/java/guava.jar`), it's better to keep this pattern in case we want to package a new Bazel for Debian.

meteorcloudy:
Yes, https://github.com/bazelbuild/bazel/pull/14357#issuecomment-983026651 is still true, what you described sounds good to me!

meteorcloudy (approved):

haxorz:
> [me] Then I will add a comment explaining the issue 5d55643 tickled.

@cushon and @michajlo figured this out. Guava 31.1 newly contains https://github.com/google/guava/commit/46093e2517b4f49c019ecff6dbe90cc22e8003b5. They verified that code causing the Desugaring step at https://github.com/bazelbuild/bazel/blob/ae41448cea071f45b4844bf64fc74bd45e1e5dc7/src/test/java/com/google/devtools/build/android/desugar/BUILD#L2094 to think about class `InternalFutureFailureAccess`, whereas it previously didn't need that class. So we now need to make sure `Desugar` has the jar containing that class on its classpath.

haxorz:
My sequence of changes needs to involve 5 commits. @katre advised me to do 5 separate PRs, so that he can vet and merge each one separately. So I'll close this one and then mail those out.

PR #14948, opened 2022-03-03
Update starlark apis that consume toolchain types to use

katre:
ToolchainTypeRequirement.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
@@ -101,10 +101,9 @@ import com.google.devtools.build.lib.util.FileType;
 import com.google.devtools.build.lib.util.FileTypeSet;
 import com.google.devtools.build.lib.util.Pair;
 import com.google.errorprone.annotations.FormatMethod;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import javax.annotation.Nullable;
 import net.starlark.java.eval.Debug;
 import net.starlark.java.eval.Dict;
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleClassFunctions.java
```diff
@@ -1049,13 +1027,62 @@ public ExecGroup execGroup(
...
+      Sequence<?> rawToolchains, StarlarkThread thread) throws EvalException {
+    ImmutableSet.Builder<ToolchainTypeRequirement> toolchainTypes = new ImmutableSet.Builder<>();
+    Set<Label> toolchainTypesSeen = new HashSet<>();
+    LabelConverter converter = LabelConverter.forThread(thread);
+
+    for (Object rawToolchain : rawToolchains) {
+      ToolchainTypeRequirement toolchainType = parseToolchainType(converter, rawToolchain);
+      if (toolchainTypesSeen.contains(toolchainType.toolchainType())) {
+        // throw an error for duplicates.
+        throw Starlark.errorf("Duplicate toolchain type %s", toolchainType.toolchainType());
```
katre, line 1050: I'm actually not sure about this, we may need to instead merge TTRs, keeping the most strict. Thoughts?

katre: Mandatory is more strict than optional. So if a rule requests `(@bazel_tools//tools/cpp:toolchain_type, Optional)` and `(@bazel_tools//tools/cpp:toolchain_type, Mandatory)`, we'd collapse that to just the mandatory version.

gregestren: I see. Toolchain types can't be inherited, can they? Is there a Starlark mechanism for rule definitions to inherit from other rule definitions?

With inheritance I could see an argument for allowing further locking down --> turning an optional into a required but not vice versa. 

Without inheritance, is there a good reason to actually have that repetition? I'd be inclined to be strict on that. We can always of course loosen if people run into problem and actually need the more complicated logic.

katre: There's no way to do inheritance in Starlark like there is in native. There may be cases where rule authors provide helpers to create the rule configurations and they combine oddly, however.

The more I think, the more I like deduping instead of failing.

gregestren: Why?

katre: In code, the toolchain type requirements act like a set (and are actually implemented with ImmutableSet). I think this is cleaner, and easier for the user (rule author) to understand.

gregestren: I'm missing the scenario in which the user would set the same toolchain for both types. Why is that more than just obvious repetition? Bazel, for example, reports an error on

```
cc_binary(
    name = "appl",
    srcs = ["main.c", "main.c"],
)
```

But I'm hearing this isn't the right comparison?

katre: A lot of rules are written by composing helpers, and those helpers are not always completely independent. We currently have many places where rules are creating the `attrs` argument by creating a dict and using that to handle merging name conflicts, so this is something that happens.

I've updated the code to merge toolchain type requirements. In this case, mandatory requirements are considered stricter than optional, since they have tighter constraints, and so that if both are present the requirement is considered to be mandatory.

gregestren: Thanks for all the elaboration. I don't want to block progress on this and your explanation makes sense. Resolving this thread and I'll do another review over the current state of the code.

gregestren (approved):

PR #14952, opened 2022-03-04
Make protocOpts() public.

comius:
Cherry pick public protoc opts for 5.1.0


Wyverald:
This is causing a lot of presubmit failures: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/10453

comius:
> This is causing a lot of presubmit failures: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/10453

Thanks for a ping. Last commit should fix that.

src/main/java/com/google/devtools/build/lib/rules/proto/ProtoConfiguration.java
```diff
@@ -201,6 +201,7 @@ public ProtoConfiguration(BuildOptions buildOptions) {
     this.options = options;
   }
 
+  @StarlarkMethod(name = "experimental_protoc_opts", structField = true, documented = false)
```
Wyverald, line 204: needs import for `StarlarkMethod`, it looks like

PR #14959, opened 2022-03-04
merkle_tree_cache: change default size to 1000

sluongng:
The implementation of Merkle Tree Cache maintains an in-process map
which cache the Input merkle tree's nodes to save traversal time when
Bazel have to interact with Remote Cache and/or RBE services.

This cache implementation used Caffeine's Soft Values cache which should
be used with predictable maximum cache size.

Set the default cache size to 1000 instead of unlimited(0).
Adjust doc to note that default might not be the most optimal value.


coeuvre (approved):

brentleyjones:
@bazel-io fork 5.1

PR #14972, opened 2022-03-06
Do not validate input-only settings in transitions

### LINKED ISSUE #14894: Build settings defined with `allow_multiple = True` cannot pass their default values to transition inputs without also being outputs

fmeum:
Starlark transition logic temporarily explicitly sets all input build
settings of a transition to their defaults. Since #13971, these values
are cleared after the transition. However, since then they have also
been subject to type validation, which is not only unnecessary, but
also breaks in the special case of a string build setting with
allow_multiple.

With this commit, input-only build settings are unconditionally
stripped from the post-transition BuildOptions and do not undergo
validation. This is made possible by a refactoring of
`StarlarkTransition#validate` that extracts the validation logic into a
separate function and also improves some variable names.

Fixes #14894

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
@@ -267,31 +267,39 @@ public abstract class StarlarkTransition implements ConfigurationTransition {
       Map<PackageValue.Key, PackageValue> buildSettingPackages,
       Map<String, BuildOptions> toOptions)
       throws TransitionException {
-    // Collect settings changed during this transition and their types. This includes settings that
-    // were only used as inputs as to the transition and thus had their default values added to the
-    // fromOptions, which in case of a no-op transition directly end up in toOptions.
-    Map<Label, Rule> changedSettingToRule = Maps.newHashMap();
+    // Collect settings that are inputs or outputs of the transition together with their types.
+    // Output setting values will be validated and removed if set to their default.
+    Map<Label, Rule> inputAndOutputSettingsToRule = Maps.newHashMap();
+    // Collect settings that were only used as inputs to the transition and thus possibly had their
+    // default values added to the fromOptions. They will be removed if set to ther default, but
+    // should not be validated.
+    Set<Label> inputOnlySettings = Sets.newHashSet();
     root.visit(
         (StarlarkTransitionVisitor)
             transition -> {
-              ImmutableSet<Label> changedSettings =
+              ImmutableSet<Label> inputAndOutputSettings =
                   getRelevantStarlarkSettingsFromTransition(
                       transition, Settings.INPUTS_AND_OUTPUTS);
-              for (Label setting : changedSettings) {
-                changedSettingToRule.put(
+              ImmutableSet<Label> outputSettings =
+                  getRelevantStarlarkSettingsFromTransition(transition, Settings.OUTPUTS);
+              for (Label setting : inputAndOutputSettings) {
+                inputAndOutputSettingsToRule.put(
                     setting, getActual(buildSettingPackages, setting).getAssociatedRule());
+                if (!outputSettings.contains(setting)) {
+                  inputOnlySettings.add(setting);
+                }
               }
             });
 
-    // Return early if no starlark settings were affected.
-    if (changedSettingToRule.isEmpty()) {
+    // Return early if the transition has neither inputs nor outputs (rare).
+    if (inputAndOutputSettingsToRule.isEmpty()) {
       return toOptions;
     }
 
     ImmutableMap.Builder<Label, Label> aliasToActualBuilder = new ImmutableMap.Builder<>();
-    for (Map.Entry<Label, Rule> changedSettingWithRule : changedSettingToRule.entrySet()) {
-      Label setting = changedSettingWithRule.getKey();
-      Rule rule = changedSettingWithRule.getValue();
+    for (Map.Entry<Label, Rule> settingWithRule : inputAndOutputSettingsToRule.entrySet()) {
+      Label setting = settingWithRule.getKey();
+      Rule rule = settingWithRule.getValue();
       if (!rule.getLabel().equals(setting)) {
         aliasToActualBuilder.put(setting, rule.getLabel());
       }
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -381,6 +338,74 @@ private static Target getActual(
...
+  private static boolean validateAndCheckIfAtDefault(Rule buildSettingRule, BuildOptions options,
+      Label maybeAliasSetting, Set<Label> inputAndOutputSettings) throws TransitionException {
+    // If the build setting was *not* referenced in the transition by an alias, this is the same
+    // value as {@code maybeAliasSetting}.
+    Label actualSetting = buildSettingRule.getLabel();
+    Object newValue = options.getStarlarkOptions().get(actualSetting);
+    // TODO(b/154132845): fix NPE occasionally observed here.
+    Preconditions.checkState(
+        newValue != null,
+        "Error while attempting to validate new values from starlark"
```
fmeum, line 363: @gregestren Do you know whether this is still needed? It is the only use of the parameter `inputAndOutputSettings`.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -267,31 +267,38 @@ private static Target getActual(
       Map<PackageValue.Key, PackageValue> buildSettingPackages,
       Map<String, BuildOptions> toOptions)
       throws TransitionException {
-    // Collect settings changed during this transition and their types. This includes settings that
-    // were only used as inputs as to the transition and thus had their default values added to the
-    // fromOptions, which in case of a no-op transition directly end up in toOptions.
-    Map<Label, Rule> changedSettingToRule = Maps.newHashMap();
+    // Collect settings that are input or outputs of the transition together with their types.
+    // Output setting values will be validated and cleaned if set to the default.
```
gregestren, line 271: Can we quickly comment-clarify what "cleaned" means? i.e. it means "removed" vs. "normalized", "linted", etc.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -267,31 +267,38 @@ private static Target getActual(
       Map<PackageValue.Key, PackageValue> buildSettingPackages,
       Map<String, BuildOptions> toOptions)
       throws TransitionException {
-    // Collect settings changed during this transition and their types. This includes settings that
-    // were only used as inputs as to the transition and thus had their default values added to the
-    // fromOptions, which in case of a no-op transition directly end up in toOptions.
-    Map<Label, Rule> changedSettingToRule = Maps.newHashMap();
+    // Collect settings that are input or outputs of the transition together with their types.
```
gregestren, line 270: typo nit: "inputs or outputs"?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -267,31 +267,38 @@ private static Target getActual(
...
+    // Output setting values will be validated and cleaned if set to the default.
+    Map<Label, Rule> inputAndOutputSettingsToRule = Maps.newHashMap();
+    // Collect settings that were only used as inputs to the transition and thus had their default
+    // values added to the fromOptions. They always have to be cleaned.
+    Set<Label> inputOnlySettings = Sets.newHashSet();
```
gregestren, line 275: Does `getRelevantStarlarkSettingsFromTransition(Settings.INPUTS)` compute this more directly?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -267,31 +267,38 @@ private static Target getActual(
...
                     setting, getActual(buildSettingPackages, setting).getAssociatedRule());
+                if (!outputSettings.contains(setting)) {
+                  inputOnlySettings.add(setting);
+                }
               }
             });
 
     // Return early if no starlark settings were affected.
-    if (changedSettingToRule.isEmpty()) {
+    if (inputAndOutputSettingsToRule.isEmpty()) {
```
gregestren, line 294: Which transitions return early here? Transitions with no inputs at all that don't change their outputs? Doesn't `getRelevantStarlarkSettingsFromTransition` itemize all inputs/outputs regardless of their values? Is this conditional actually possible aside from a trivial no-op transition?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -307,69 +314,19 @@ private static Target getActual(
...
-            if (cleanedOptions == null) {
-              cleanedOptions = options.toBuilder();
-            }
-            cleanedOptions.removeStarlarkOption(rule.getLabel());
+        Label maybeAliasSetting = settingWithRule.getKey();
+        Rule rule = settingWithRule.getValue();
+        // Input-only settings had their default value added to the BuildOptions and thus always
+        // have to be cleaned, whereas output settings have their value validated, converted and
+        // stripped if it is equivalent to the setting's default value.
+        if (inputOnlySettings.contains(maybeAliasSetting) || validateAndCheckIfAtDefault(
```
gregestren, line 324: I'm still hazy on exactly what `getRelevantStarlarkSettingsFromTransition` returns. If an input is at a non-default value, would it not be part of `inputOnlySettings`? Which logic excludes it?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -381,6 +338,74 @@ private static Target getActual(
     return cleanedOptionMap.buildOrThrow();
   }
 
+  /**
+   * Validate the value of a particular build setting after a transition has been applied.
```
gregestren, line 342: Javadoc nit: can we get a blank line after this?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -381,6 +338,74 @@ private static Target getActual(
...
+   * @param buildSettingRule the build setting to validate.
+   * @param options the {@link BuildOptions} reflecting the post-transition configuration.
+   * @param maybeAliasSetting the label used to refer to the build setting in the transition,
+   *                          possibly an alias. This is only used for error messages.
```
gregestren, line 346: nit: normalize the indenting here and below?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
```diff
@@ -381,6 +338,74 @@ private static Target getActual(
...
+   * @param inputAndOutputSettings the transition input and output settings. This is only used for
+   *                               error messages.
+   * @return {@code true} if and only if the setting is set to its default value after the
+   *         transition.
+   * @throws TransitionException if the value returned by the transition for this setting has an
+   *                             invalid type.
+   */
+  private static boolean validateAndCheckIfAtDefault(Rule buildSettingRule, BuildOptions options,
```
gregestren, line 354: Is this method pretty much a boiler plate copy of the original logic, or are there specific changes I should pay attention to?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
gregestren: Good question. The bug says there were instances of NPes but we haven't figured out how to reproduce them. So this at least provides a clear error message vs. an arbitrary crash.

Given that I'd keep this, even with the extra complication it adds.

gregestren (commented):
Lots of nitpicky and sanity checking comments on my side, but overall I love the direction of this!

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkTransition.java
gregestren: Ah, good point.

gregestren (approved):
Looks solid!

PR #14981, opened 2022-03-07
Do not hide BulkTransferException messages when there were more than one exception

thii:
Previously, when there were more than one BulkTransferException, it
would be reported like this:

```
Executing genrule //:foo failed: Exec failed due to IOException: 221 errors during bulk transfer
```

which didn't include the underlying exception messages. The only case
that underlying exceptions were included was when there was only one
exception.

This change patches the error message to include all the exception
messages, which helps diagnose BulkTransferException.


coeuvre (approved):

brentleyjones:
@bazel-io fork 5.1

Issue #14985, opened 2022-03-07
[5.1] Do not hide BulkTransferException messages when there were more than one exception


bazel-io:
Forked from #14981

PR #14986, opened 2022-03-07
[5.1] Do not hide BulkTransferException messages when there were more than one exception

brentleyjones:
Previously, when there were more than one BulkTransferException, it
would be reported like this:

```
Executing genrule //:foo failed: Exec failed due to IOException: 221 errors during bulk transfer
```

which didn't include the underlying exception messages. The only case
that underlying exceptions were included was when there was only one
exception.

This change patches the error message to include all the exception
messages, which helps diagnose BulkTransferException.

(cherry picked from commit 113eaca5862c48797654ae2a3acbb6e15d761485)

Closes https://github.com/bazelbuild/bazel/issues/14985.


brentleyjones:
@Wyverald

Issue #14714, opened 2022-02-03
Release 5.1 - March 2022


Wyverald:
# Status of Bazel 5.1

- Expected release date: 2022-03-24
- [Release blockers milestone](https://github.com/bazelbuild/bazel/milestone/35)

To report a release-blocking bug, please add a comment with the text `@bazel-io flag` to the issue. A release manager will triage it and add it to the milestone.

To cherry-pick a mainline commit into 5.1, simply send a PR against the `release-5.1.0` branch.

Task list:

- [x] [Create draft release announcement](https://docs.google.com/document/d/1wDvulLlj4NAlPZamdlEVFORks3YXJonCjyuQMUQEmB0/edit)
- [x] Send for review the release announcement PR: https://github.com/bazelbuild/bazel-blog/pull/271
- [x] Push the release, notify package maintainers:
- [x] Update the documentation
- [x] Push the blog post
- [x] Update the [release page](https://github.com/bazelbuild/bazel/releases/)

martis42:
Please cherry-pick this fix https://github.com/bazelbuild/bazel/issues/14646#issuecomment-1028969558

comius:
Please cherry-pick 0e73219fbe4f8053ca196753fc7fac152f02a7d6

sgammon:
We humbly request a #14724 be added to unblock coverage in `rules_apple`

comius:
> @comius could you send a cherry-pick PR against `release-5.1.0`? This commit doesn't look trivial to cherry-pick.

Done:
https://github.com/bazelbuild/bazel/pull/14952

sluongng:
As discussed in https://github.com/bazelbuild/bazel/pull/13879#issuecomment-1058851178 I just opened https://github.com/bazelbuild/bazel/pull/14959

Hoping that it can be picked to 5.1 if possible 🙇‍♂️

Wyverald:
Status update: there are still a few rather key release blockers (a couple internal ones too), so we'll have to postpone the release again.

kmicklas:
Any chance we can get f9882e41382fff23f4862841631da593c9fe2344 cherry-picked?

Wyverald:
Thank you so much, @brentleyjones!

@linzhp @kmicklas For future reference, please send a cherry-pick PR against the branch `release-5.1.0`. It's much easier to keep track of PRs than comments in this issue.

Wyverald:
5.1 is now out!

[...]

PR #14989, opened 2022-03-07
[5.1] Add --experimental_repository_cache_urls_as_default_canonical_id to help detect broken repository URLs

brentleyjones:
This new flag can be used to force redownloading when repository URLs are changed. Otherwise, it's possible broken URLs can be masked by the presence of a repository cache entry with the same hash. Specifying a `canonical_id` works as before, overriding this behavior.

(cherry picked from commit f9882e41382fff23f4862841631da593c9fe2344)


brentleyjones:
@Wyverald

brentleyjones:
Requested here: https://github.com/bazelbuild/bazel/issues/14714#issuecomment-1060940771

Issue #14994, opened 2022-03-08
[5.1] Correct cpu and os values of `local_config_cc_toolchains` targets


bazel-io:
Forked from #14844

PR #14995, opened 2022-03-08
[5.1] Correct cpu and os values of `local_config_cc_toolchains` targets

brentleyjones:
These aren't being used now, but some of them are clearly defined
incorrectly. Fix them now to avoid surprises in the future. Also change
`aarch64` to `arm64` since `aarch64` is an alias of `arm64`.

(cherry picked from commit b858ec39aebd7e586af5438aa2035db2adebf9a4)

Closes https://github.com/bazelbuild/bazel/issues/14994.


brentleyjones:
@Wyverald

thii:
Seems like the failures were because of timeout. Can you force push to rerun the tests?

PR #15000, opened 2022-03-08
Clear build configurations on --discard_analysis_cache

### LINKED ISSUE #14725: Ability to clean stale configuration data from config/cquery without shutting all the way down

Currently, running `bazel config` or `bazel cquery` may return stale data. In particular:
* https://docs.bazel.build/versions/main/cquery.html has a warning about `cquery` picking up results from past queries
* Running `bazel cquery ... && bazel config && bazel cquery --test_arg=foo && bazel config` - the second `bazel config` will return configurations from the first invocation which would not be returned if a `bazel shutdown` was performed between the runs.

As a workaround to this, I have scripts which run `bazel shutdown` between these invocations, but this is very expensive - repositories need re-fetching, and build files need re-loading from scratch.

It would be useful to either give these commands a `--clear_previous_state` flag, which drops a targeted subset of state to avoid this, or to have a top-level `bazel drop-stale-configs` or similar command.

I'd be happy to try out implementing either of these solutions, if someone can point me at roughly what the solution would look like.

/cc @gregestren

illicitonion:
This means that `bazel config` no longer returns stale configurations
after a build with `--discard_analysis_cache`.

Unfortunately, because `--discard_analysis_cache` may still preserve
part of the graph, we don't unconditionally do this (as doing so would
cause an inconsistent graph).

Rather than walking the graph to find out which configurations are
actually needed, we either drop all or none. This could be refined in
the future.

Alternatively, perhaps there is a need for a standalone Command:
`bazel discard-analysis-cache` which can avoid the
preserve-top-level-target edge-cases of
`bazel build --discard_analysis_cache`.

Fixes #14725

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeExecutor.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeExecutor.java
@@ -1084,6 +1084,14 @@ public abstract class SkyframeExecutor implements WalkableGraphFactory, Configur
           }
         }
       }
+      if (topLevelTargets.isEmpty() && topLevelAspects.isEmpty()) {
+        // Top-level ConfiguredTargets may still reference configurations, so we leak them
+        // if any are being retained.
+        // Ideally this would only preserve the configurations reachable from those top-level
+        // targets, but this at least lets people run `bazel build --discard_analysis_cache`
+        // to blow away stale output from `bazel config`.
+        memoizingEvaluator.delete(k -> k.functionName().equals(SkyFunctions.BUILD_CONFIGURATION));
+      }
     }
   }
 


gregestren:
Note to self: check up on this next week.

gregestren:
What do you think about https://github.com/bazelbuild/bazel/commit/2f4ed67b2539762a36c4e0a28018d62ef8811477?

There were some internal comments. Particularly:

> I will flip the value of this flag after the next Bazel release to decouple it from the release.

It didn't get flipped. But the intention looked pretty similar to yours.

illicitonion:
Thanks for following up! Functionality-wise, it looks like it does the job great for me, the only weird corner is that it appears to need two builds to take effect.

If I perform the following sequence of commands:
```
bazel cquery 'deps(...)' >/dev/null
bazel build --discard_analysis_cache --noexperimental_keep_config_nodes_on_analysis_discard
bazel config
```

I still get configurations printed. But if I run:

```
bazel cquery 'deps(...)' >/dev/null
bazel build --discard_analysis_cache
bazel build --noexperimental_keep_config_nodes_on_analysis_discard
bazel config
```

or

```
bazel cquery 'deps(...)' >/dev/null
bazel build --discard_analysis_cache --noexperimental_keep_config_nodes_on_analysis_discard
bazel build --noexperimental_keep_config_nodes_on_analysis_discard
bazel config
```

I get a clear list as expected.

I guess this is because of the funky "mark that stuff needs cleaning up at the start of the next build" behaviour of `--discard_analysis_cache`, but it'd be nice to not need two invocations if possible.

gregestren:
I'm actually surprised those last two scenarios clear as expected.

I'm only somewhat informed on these flags (I'm basically familiar with `--discard_analysis_cache` but not an expert, and as you know, the most pedantic details matter in code). So this intuition could be wrong. But my understanding is:

- `--discard_analysis_cache` clears the analysis cache immediately when the first build finishes. See https://github.com/bazelbuild/bazel/blob/4e6153eeb6adb3489bf2686f5c84135df37fdae4/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java#L454
- `--noexperimental_keep_config_nodes_on_analysis_discard` clears configs at the start of the next build. See https://github.com/bazelbuild/bazel/blob/e060edfdd7ad08ee381b18cc3a5f65719639cde0/src/main/java/com/google/devtools/build/lib/skyframe/SkyframeExecutor.java#L2669

But why do you need `bazel config` to be clear? Wasn't the main cause to have it only retain the info from the last `cquery` or `build` call?

illicitonion:
> But why do you need `bazel config` to be clear? Wasn't the main cause to have it only retain the info from the last `cquery` or `build` call?

I think I can actually get away without it being clear :) I had two reasons for that check:
1. Verifying that the analysis cache has actually been discarded before performing more queries (as far as I can tell, all other ways of inspecting the contents of the analysis cache _also_ have the side effect of populating it, whereas `bazel config` does not).
2. Parsing out which configuration hash represents the host config. Before this PR, running `bazel config` may return multiple configurations labelled as `host`, and being able to know which configuration is host can be useful. But switching to use the proto output provides more information about these configurations anyway, so I should probably just do that in my consuming code.

Thanks for the digging and thoughts!

gregestren:
FYI https://github.com/bazelbuild/bazel/commit/2f9f8429be9651d4bb94b425b6ae5f11e95bea16

illicitonion:
Thanks for the heads up!

Issue #8546, opened 2019-06-03
incompatible_disable_static_cc_toolchains: Remove @bazel_tools//tools/cpp:default-toolchain and friends


scentini:
**Flag:** --incompatible_disable_static_cc_toolchains
**Available since:** 0.28
**Will be flipped:** 1.0

Currently we have a [`@bazel_tools//tools/cpp:default-toolchain`](https://github.com/bazelbuild/bazel/blob/2b3c879a6507f17030fdf9fba2a1704cb09eb5f3/tools/cpp/BUILD#L87) `cc_toolchain_suite`, along with multiple `cc_toolchain` targets that Bazel doesn't use.

Users use it by specifying `--host_crosstool_top=@bazel_tools//tools/cpp:default-toolchain` and/or `--crosstool_top=@bazel_tools//tools/cpp:default-toolchain`.

The `incompatible_disable_static_cc_toolchains` flag will disable using the [`@bazel_tools//tools/cpp:default-toolchain`](https://github.com/bazelbuild/bazel/blob/2b3c879a6507f17030fdf9fba2a1704cb09eb5f3/tools/cpp/BUILD#L87) toolchain suite, as well as all the individual `cc_toolchain`s under [`tools/cpp/BUILD`](https://github.com/bazelbuild/bazel/blob/2b3c879a6507f17030fdf9fba2a1704cb09eb5f3/tools/cpp/BUILD)

### Migration
Users should use `--host_crosstool_top=@bazel_tools//tools/cpp:toolchain` and/or `--crosstool_top=@bazel_tools//tools/cpp:toolchain` instead, which uses the auto-configuration mechanism. Users can also omit the `--crosstool_top` and `--host_crosstool_top` options altogether, as they point to `@bazel_tools//tools/cpp:toolchain` by default.

laurentlb:
If it's going to be flipped for 1.0, please add the label `breaking-change-1.0`.

PR #15003, opened 2022-03-08
Remove cpp:default-toolchain

keith:
This was disabled in bazel 1.0 but there were still a lot of remnants
around.

https://github.com/bazelbuild/bazel/issues/8546


keith:
@oquenchil given no failures here do you have more context on what should break here?

oquenchil (approved):
LGTM

PR #15006, opened 2022-03-09
Fix lld detection on macOS

keith:
The MachO version of LLD is being used in production at this point, but
the detection incorrectly detected the linker name on macOS. Since the
detection was using `string.find("lld")` it matched this flag instead
`-debugger-tuning=lldb`. This now verifies that the match is against the
expected format which starts with double quotes.

Fixes https://github.com/bazelbuild/bazel/issues/15126

### DIFF
--- a/tools/cpp/unix_cc_configure.bzl
+++ b/tools/cpp/unix_cc_configure.bzl
@@ -203,20 +203,11 @@ def _find_linker_path(repository_ctx, cc, linker, is_clang):
     if not is_clang:
         return linker
 
-    for line in result.stderr.splitlines():
-        if line.find(linker) == -1:
-            continue
-        for flag in line.split(" "):
-            if flag.find(linker) == -1:
-                continue
-
-            # flag looks like "/usr/lib/ld.gold".
-            return flag.strip(" \"'")
-    auto_configure_warning(
-        "CC with -fuse-ld=" + linker + " returned 0, but its -v output " +
-        "didn't contain '" + linker + "', falling back to the default linker.",
-    )
-    return None
+    # Extract linker path from:
+    # /usr/bin/clang ...
+    # "/usr/bin/ld.lld" -pie -z ...
+    linker_command = result.stderr.splitlines()[-1]
+    return linker_command.strip().split(" ")[0].strip("\"'")
 
 def _add_compiler_option_if_supported(repository_ctx, cc, option):
     """Returns `[option]` if supported, `[]` otherwise. Doesn't %-escape the option."""


ckolli5:
Hello @keith, Can you please check the buildkite presubmit failures. Thanks!

keith:
@ckolli5 updated

keith:
@googlewalt ping

keith:
@googlewalt @sgowroji ping

tools/cpp/unix_cc_configure.bzl
```diff
@@ -208,6 +208,8 @@ def _find_linker_path(repository_ctx, cc, linker, is_clang):
         for flag in line.split(" "):
```
googlewalt, line 208: t seems this code has a lot of legacy cruft that we can clean up.  IIUC this only needs to support clang now, and we are just looking for the linker path, right?  So instead of looping through the entire commandline can we just check the first arg to see if it matches the linker (with possibly a space before it)?

keith: unix_cc_configure does support GCC as well, it's just on darwin that it will realistically only ever be clang. Related: https://github.com/bazelbuild/bazel/commit/ef3f05852b3bf542903f091184da96b96354a13d

googlewalt: But I think this code fragment needs to only support clang, due to line 202, from https://github.com/bazelbuild/bazel/pull/12729.

googlewalt: More to the point, this code evolved from back when the code needed to look for "-fuse-ld=<linker>" arg for gcc, but the linker binary for clang.  That was why it iterated through the whole commandline.  Now that it is only looking for the linker binary for clang, it should only need to look at the first arg.

dom96:
It seems that this is only used on macOS when Xcode isn't installed. I'm curious why that is, does Bazel side-step this entirely if Xcode is found? I wonder if it wouldn't be better to just assume Xcode exists and have bazel fail fast if it doesn't with a descriptive error message asking for it to be installed.

keith:
Having the Xcode command line tools installed without a full Xcode installation is a supported (albeit not well tested) configuration.

tools/cpp/unix_cc_configure.bzl
keith: good point, I changed the parsing to extract the first argument, from the last line, of the output. For reference here's the output I see on macOS:

```
% clang "tools/cpp/empty.cc" -o /dev/null -fuse-ld=lld -v
Apple clang version 14.0.0 (clang-1400.0.29.102)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
 "/Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" -cc1 ...
clang -cc1 version 14.0.0 (clang-1400.0.29.102) default target arm64-apple-darwin21.6.0
ignoring nonexistent directory "/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks"
#include "..." search starts here:
 /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1
 /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/14.0.0/include
 /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
 "/Users/ksmiley/bin/ld64.lld" -demangle -no_deduplicate -dynamic -arch arm64 -platform_version macos 12.0.0 12.3 -syslibroot /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -o /dev/null -L/usr/local/lib /var/folders/gj/wf3swl0x215b2sq1qy84kzkm0000gn/T/empty-ff40e4.o -lSystem /Applications/Xcode-14.0.0-RC1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/14.0.0/lib/darwin/libclang_rt.osx.a
```

tools/cpp/unix_cc_configure.bzl
```diff
@@ -203,20 +203,11 @@ def _find_linker_path(repository_ctx, cc, linker, is_clang):
...
-    auto_configure_warning(
-        "CC with -fuse-ld=" + linker + " returned 0, but its -v output " +
-        "didn't contain '" + linker + "', falling back to the default linker.",
-    )
-    return None
+    # Extract linker path from:
+    # /usr/bin/clang ...
+    # "/usr/bin/ld.lld" -pie -z ...
+    linker_command = result.stderr.splitlines()[-1]
+    return linker_command.strip().split(" ")[0].strip("\"'")
```
googlewalt, line 210: I think we still need to handle the case where the requested linker can't be found?

keith: in that case the `clang` command fails and we hit the codepath above

googlewalt: Never mind.  Checking the return_code above should suffice.

googlewalt (approved):

fmeum:
@keith Do you know of a workaround that applies to earlier versions of Bazel? A recent regression in Bazel on Windows prevents me from using https://github.com/bazelbuild/bazel/releases/tag/6.0.0-pre.20220922.1.

keith:
I don't see an easy way. I guess if you called `cc_autoconf_impl` yourself and passed `overridden_tools` you could probably override the compiler to return something special in this case, but that's probably too much overhead 😕

PR #15007, opened 2022-03-09
Remove -U_FORTIFY_SOURCE when thin_lto is enabled

keith:
Clang does not accept `-U` flags when doing lto backend actions:

```
clang: error: argument unused during compilation: '-U _FORTIFY_SOURCE' [-Werror,-Wunused-command-line-argument]
```

This isn't a problem for `-D` flags because of a special case:

https://github.com/llvm/llvm-project/blob/10e99eb7e4adbb8e2407e2f6ae3d2e6420d060c9/clang/lib/Driver/ToolChains/Clang.cpp#L5579-L5582


brentleyjones:
@bazel-io flag

Issue #14474, opened 2021-12-25
Overriding default java toolchain with source version 17 is failing in com.google.turbine.options.LanguageVersion: invalid -source version: 17


davido:
On Bazel@HEAD (46ed5d55def27d124afcdeaccd17f3cd16070816) I'm trying to override `default_java_toolchain` to promote Error Prone bug patterns to ERROR severity:

```python
default_java_toolchain(
    name = "error_prone_warnings_toolchain_java17",
    package_configuration = [
        ":error_prone",
    ],
    java_runtime = "@bazel_tools//tools/jdk:remotejdk_17",
    source_version = "17",
    target_version = "17",
    visibility = ["//visibility:public"],
)

# Error Prone errors enabled by default; see ../.bazelrc for how this is
# enabled. This warnings list is originally based on:
# https://github.com/bazelbuild/BUILD_file_generator/blob/master/tools/bazel_defs/java.bzl
# Additionally, items used internally in google is added. Such items have
# the same or higher verbosity level than in google.
# However, feel free to add any additional errors. Thus far they have all been pretty useful.
# All warnings are commented to avoid noise in the output.
# Newer versions of error-prone have XepDisableAllWarnings flag which could
# be used instead of commenting. Bazel should be updated to use a new version
# of error-prone.
java_package_configuration(
    name = "error_prone",
    javacopts = [
        "-XepDisableWarningsInGeneratedCode",
[...]
```

and register this custom toolchain in `WORKSPACE`:

```python
register_toolchains("//tools:error_prone_warnings_toolchain_java17_definition")
```

and activate it like this in `.bazelrc`:

```
# Builds using remote_jdk17, executes using remote_jdk17 or local_jdk
build --java_language_version=17
build --java_runtime_version=17
build --tool_java_language_version=17
build --tool_java_runtime_version=17
```

Now, the build is failing:

```
  $ bazeldev build --verbose_failures java/com/google/gerrit/common:server

NFO: Invocation ID: 49cd923d-3a0b-429c-a910-7f038e35dbae
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=219
INFO: Reading rc options for 'build' from /home/davido/projects/gerrit/.bazelrc:
[...]

PR #15009, opened 2022-03-09
Update use of `default_java_toolchain` in "Java and Bazel" to work correctly

kjcartledge:
There are really two changes here:

1. Fix the reference to `//tools/jdk:remote_jdk11` to have `@bazel_tools`. Without this, a direct copy/paste will fail to even build.
2. Change from `JDK9_JVM_OPTS` to `BASE_JDK9_JVM_OPTS` as the former includes `--patch-module=java.compiler=` which overrides `source_version` and can also cause build failures. See [this issue](https://github.com/bazelbuild/bazel/issues/14474#issuecomment-1001071398) for an example.

### DIFF
--- a/site/en/docs/bazel-and-java.md
+++ b/site/en/docs/bazel-and-java.md
@@ -242,16 +242,16 @@ Example toolchain configuration:
 ```python
 load(
   "@bazel_tools//tools/jdk:default_java_toolchain.bzl",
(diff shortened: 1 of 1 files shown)


comius (approved):

comius:
Thanks @kjcartledge!

PR #15023, opened 2022-03-10
If a test failure isn't found in the test case map, try to assign it to the suite matching a class that is

gmishkin:
Mockito unnecessary stubbing detection ([`org.mockito.exceptions.misusing.UnnecessaryStubbingException`](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/exceptions/misusing/UnnecessaryStubbingException.html)) fires off failures outside the context of a test case, but since the Description isn't found in `testsMap`, the details aren't written to the test.xml (however it still goes to stdout and the overall bazel test fails).

This change attempts to look up a matching suite in the test map by its Class. If found, it creates a new test case node and marks it as failed. I tried using the `dynamicTestFailure` thingy directly on the suite, and that would increase the failure count, but the failure details wouldn't be added to the XML.

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
@@ -30,6 +30,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nullable;
@@ -91,7 +92,27 @@ public class TestSuiteModel {
 
   private TestNode getTest(Description description) {
     // The description shouldn't be null, but in the test runner code we avoid throwing exceptions.
-    return description == null ? null : testsMap.get(description);
+    if (description == null) {
+      return null;
+    }
+
+    TestNode test = testsMap.get(description);
+    if (test != null) {
+      return test;
+    }
+
+    final Optional<Description> matchingSuite = testsMap.keySet().stream()
+            .filter(d -> d != null && d.getTestClass() != null && d.isSuite() && d.getTestClass().equals(description.getTestClass()))
+            .findAny();
+    if (matchingSuite.isPresent()) {
+      final TestSuiteNode suite = (TestSuiteNode)testsMap.get(matchingSuite.get());
+      test = new TestCaseNode(description, suite);
+      suite.addTestCase((TestCaseNode)test);
+      testsMap.put(description, test);
+      return test;
+    }
+
+    return null;
   }
 
   // VisibleForTesting


src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
```diff
@@ -183,6 +184,18 @@ public void testFailure(Description description, Throwable throwable) {
         test.testFailure(throwable, now());
       }
     }
+    else {
+      final Optional<Description> matchingSuite = testsMap.keySet().stream()
```
c-mita, line 188: I think it's worth adding a comment indicating how this scenario occurs and what the motivation for the code is.

src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
```diff
@@ -183,6 +184,18 @@ public void testFailure(Description description, Throwable throwable) {
...
+          .filter(d -> d.isSuite() && d.getTestClass().equals(description.getTestClass()))
+          .findAny();
+      if (matchingSuite.isPresent()) {
+        final TestSuiteNode suite = (TestSuiteNode)testsMap.get(matchingSuite.get());
+        final TestCaseNode newTest = new TestCaseNode(description, suite);
+        suite.addTestCase(newTest);
+        testsMap.put(description, newTest);
+        newTest.testFailure(throwable, now());
```
c-mita, line 196: TestNode differentiates between `testFailure` and `dynamicTestFailure`. I'm not familiar with this code, but would it not be better to do this as well?

Perhaps the code should be written so the logic to get a `TestNode` object (and update the map) comes at the top, and then the original code just operates on that?

gmishkin: @c-mita for this particular kind of exception, at least, when I try replacing the call to `testFailure` with `dynamicTestFailure`, I don't get a full stack trace in the final XML, despite passing through the Throwable.

So I'm a little hesitant to reorder the logic, because I don't understand the circumstances under which a DynamicTestException will appear in the top part of the if statement.

sgowroji:
Hello @gmishkin, Can you please resolve  the above comments and send back for review. Thanks!

src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
c-mita: After a little more investigation, I _believe_ that `DynamicTestException` is unused within Bazel.

It is used within Google however, so some care is needed. I'll need to investigate a bit more as changes here could have impacts beyond Bazel's test exection.

gmishkin:
@sgowroji I am just waiting for the outcome of the internal investigation in https://github.com/bazelbuild/bazel/pull/15023#discussion_r856316769 before proceeding further. I could start typing up a code comment for the first review comment above, but it'd simply be restarting the pull request description, and I'd rather wait til the other comment is resolved in case it affects the patch or wording.

src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
c-mita: That using `dynamicTestFailure` all the time surpresses full stack traces doesn't surprise me, because that appears to be part of its intent. Its origin lies in handling Javascript test suites (which is an area I have no expertise in).

But the code I had in mind was this:
```
if (throwable instanceof DynamicTestException) {
  DynamicTestException dynamicFailure = (DynamicTestException) throwable;
  newTest.dynamicTestFailure(dynamicFailure.getTest(), dynamicFailure.getCause(), now());
} else {
  newTest.testFailure(throwable, now());
}
```
Essentially replicating what's done in the method above.

gmishkin: Moved the logic into getTest() think that's what you meant originally.

gmishkin:
@c-mita the tests that are failing don't look like they're related to these changes, please advise if you think there's something I need to take a closer look at.

src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
c-mita: This means this new logic will run for more than just test failures; I don't know if this is a good thing or not.

src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/TestSuiteModel.java
```diff
@@ -91,7 +92,27 @@ private TestCaseNode getTestCase(Description description) {
...
+    if (description == null) {
+      return null;
+    }
+
+    TestNode test = testsMap.get(description);
+    if (test != null) {
+      return test;
+    }
+
+    final Optional<Description> matchingSuite = testsMap.keySet().stream()
```
c-mita, line 104: This still needs a comment explaining why this is being done.

Especially considering that now a "getTest" method can result in internal state being changed.

c-mita:
> @c-mita the tests that are failing don't look like they're related to these changes, please advise if you think there's something I need to take a closer look at.

After cursory glance, I don't think your change is causing those failures.

c-mita (changes requested):

PR #15025, opened 2022-03-11
Fix unused features in cc toolchains

keith:
These came up with the new buildifier unused variable warnings

### DIFF
--- a/tools/cpp/unix_cc_toolchain_config.bzl
+++ b/tools/cpp/unix_cc_toolchain_config.bzl
@@ -729,41 +729,6 @@ def _impl(ctx):
         ],
     )
 
-    llvm_coverage_map_format_feature = feature(
-        name = "llvm_coverage_map_format",
-        flag_sets = [
-            flag_set(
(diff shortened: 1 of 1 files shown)


keith:
These features all appear to have been accidentally omitted from the features lists

oquenchil:
All failing tests seem related to coverage.

ckolli5:
Hi @keith, Can you please check the buildkite presubmit failures. Thanks!

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -761,7 +761,6 @@ def _impl(ctx):
             ),
         ],
         requires = [feature_set(features = ["coverage"])],
-        provides = ["profile"],
```
keith, line 764: This matches the macOS toolchain which uses similar features, since this requires coverage I think that's safe, since coverage provides profile

oquenchil (approved):
LGTM

PR #15030, opened 2022-03-12
Run tests in runfiles directory without `LCOV_MERGER`

fmeum:
`collect_coverage.sh` has the execution root as it's working directory, but tests should still be run from their runfiles directory  if `LCOV_MERGER` is missing.


fmeum:
@keith @c-mita

keith:
Thanks for submitting this! Seeing another issue with this setup code I submitted this alternative that I think would be more future proof, but it's up to @c-mita to decide on which approach makes sense. Can you add a test similar to the runfiles tests I added in my PR to catch regressions in the case @c-mita wants to go with this change?

keith:ks/fix-coverage-runfiles-directory-issue

fmeum:
I also think that the approach of #15031 is more future-proof. I can't edit the PR at the moment, but will do so in case the other PR can't be used for some reason.

PR #15031, opened 2022-03-12
Fix coverage runfiles directory issue

keith:
As discovered in https://github.com/bazelbuild/bazel/pull/15030 there
was another bug with the short circuiting coverage logic where tests
were not run from the correct directory. In that PR this issue is fixed
directly, with this change we instead make missing LCOV_MERGER be
deferred until after all test setup and run, which I think is more
future proof to other changes here, and also allows users to know their
tests are being run with coverage, and output the correct files, even in
the case they don't setup the coverage merger infrastructure.


keith:
@c-mita checkout this comment https://github.com/bazelbuild/bazel/pull/15030#issuecomment-1065950390 for more context, I think this is the better long term option so we don't have more issues like this, and so users can rely on `$COVERAGE == 1` even if they don't use coverage merging, but up to you!

PR #15032, opened 2022-03-14
Add more notes about intermediary macro targets

bpcreech:
Adding some best practice my team has discovered and are practicing for "internal" (intermediary) targets generated by widely-used macros.


bpcreech:
Created https://github.com/bazelbuild/bazel/issues/15033 to go with this

gregestren:
This sounds reasonable to me. CC'ing @brandjon @comius who have shared interest in the build API.

I'll default to approve in a week if we don't get blocking comments beforehand (please ping the PR then to guarantee it doesn't get lost my inbox!)

site/en/rules/bzl-style.md
```diff
@@ -141,11 +141,14 @@ not instantiated by that macro), follow these best practices:
...
-    an underscore (`_`), followed by the name attribute. For instance, if the
-    macro is supplied with the name *resources*, internal targets should have
-    names beginning with *_resources*. They should also have restricted
-    visibility.
+*   All other targets defined by a macro should:
+    *   Have their names preceded with an underscore (`_`), followed by the
+        value of the `name` attribute. For instance, if the macro is supplied
+        with the name *resources*, internal targets should have names
+        beginning with *_resources*.
+    *   Have restricted visibility (`//visibility:private`), and
```
brandjon, line 149: It's possible that a macro could generate an internal target for use by the same macro called in other packages. For instance, the macro could take `name="foo"` and generate "_foo_helper". Then in another package, it could take in `deps=["//pkg:foo"]` and munge that string into "//pkg:_foo_helper" to get the private target. In that case, the helper can't have private visibility.

But I don't think that's necessarily a style we should encourage, and anyone clever enough to do that is also clever enough to realize this advice doesn't apply to them anyway. So the current wording is fine.

brandjon (approved):

PR #15038, opened 2022-03-14
Cquery can report transitions

### LINKED ISSUE #14610: cquery doesn't filter ruleInputs according to used configurations

Cquery follows selects across the build graph, but does not appear to perform this filtering to the `Rule.rule_input` field.

Accordingly, when trying to ascertain what files may affect a target in the _current_ configuration, irrelevant files may be returned.

For example, when trying to query the transitive affecting files of this kind of target:
```starlark
java_test(
    name = "ExampleTest",
    srcs = ["ExampleTest.java"],
)
```

Following the `rule_input` fields when evaluating `bazel cquery 'deps(//java/example:ExampleTest) --output=jsonproto`, there's the chain:

```
//java/example:ExampleTest -> @bazel_tools//tools/test:xml_writer -> @bazel_tools//tools/test:xml.exe
```

However, when the currently configured platform is not Windows, there should be no dependency on `xml.exe`, as that's Window-specific.

I think it would be reasonable to have to perform some logic to do this filtering (though it just being presented completely up-front would be great), but right now I don't think it's possible to follow any algorithm to decide that `xml.exe` isn't actually a dependency in the current configuration...

Coming up with a hash for each target so that we can diff graphs across different commits of source code. See also https://github.com/bazelbuild/bazel/issues/7962.

With the above BUILD file, run a `bazel cquery` for that target, and try determine whether `@bazel_tools//tools/test:xml.exe` is depended on by the target.

release 4.2.1

/cc @gregestren

### LINKED ISSUE #14617: cquery returns rule_input targets which are missing from deps list

Chasing the chain of `rule_input`s from running `cquery` on a `java_test` target, I see the following chain:

```
//java/example:ExampleTest
->@bazel_tools//tools/cpp:toolchain
->@local_config_cc//:toolchain
->@local_config_cc//:cc-compiler-armeabi-v7a
->@bazel_tools//tools/objc:host_xcodes
```

In particular, here's where I see `@bazel_tools//tools/objc:host_xcodes`:

```
% bazel cquery '@local_config_cc//:cc-compiler-armeabi-v7a' --output=jsonproto 2>/dev/null | jq '.results[].target.rule.ruleInput'
[
  "@bazel_tools//tools/build_defs/cc/whitelists/parse_headers_and_layering_check:disabling_parse_headers_and_layering_check_allowed",
  "@bazel_tools//tools/build_defs/cc/whitelists/starlark_hdrs_check:loose_header_check_allowed_in_toolchain",
  "@bazel_tools//tools/cpp:interface_library_builder",
  "@bazel_tools//tools/cpp:link_dynamic_library",
  "@bazel_tools//tools/objc:host_xcodes",
  "@local_config_cc//:empty",
  "@local_config_cc//:osx_tools_armeabi-v7a",
  "@local_config_cc//:stub_armeabi-v7a"
]
[
  "@bazel_tools//tools/build_defs/cc/whitelists/parse_headers_and_layering_check:disabling_parse_headers_and_layering_check_allowed",
  "@bazel_tools//tools/build_defs/cc/whitelists/starlark_hdrs_check:loose_header_check_allowed_in_toolchain",
  "@bazel_tools//tools/cpp:interface_library_builder",
  "@bazel_tools//tools/cpp:link_dynamic_library",
  "@bazel_tools//tools/objc:host_xcodes",
  "@local_config_cc//:empty",
  "@local_config_cc//:osx_tools_armeabi-v7a",
  "@local_config_cc//:stub_armeabi-v7a"
]
[
  "@bazel_tools//tools/build_defs/cc/whitelists/parse_headers_and_layering_check:disabling_parse_headers_and_layering_check_allowed",
  "@bazel_tools//tools/build_defs/cc/whitelists/starlark_hdrs_check:loose_header_check_allowed_in_toolchain",
  "@bazel_tools//tools/cpp:interface_library_builder",
  "@bazel_tools//tools/cpp:link_dynamic_library",
  "@bazel_tools//tools/objc:host_xcodes",
  "@local_config_cc//:empty",
  "@local_config_cc//:osx_tools_armeabi-v7a",
  "@local_config_cc//:stub_armeabi-v7a"
]
```

And this target seems to exist:

```
% bazel cquery '@bazel_tools//tools/objc:host_xcodes' 2>/dev/null                                                           
@bazel_tools//tools/objc:host_xcodes (51eca89)
```

But it's not returned in `deps(//java/example:ExampleTest)` or `somepath(//java/example:ExampleTest, @bazel_tools//tools/objc:host_xcodes)`:

```
% bazel cquery 'somepath(//java/example:ExampleTest, @bazel_tools//tools/objc:host_xcodes)' 2>/dev/null

% bazel cquery 'deps(//java/example:ExampleTest)' 2>/dev/null | grep -c '@bazel_tools//tools/objc:host_xcodes'
0
```

The target _does_ show up using `query` not `cquery`, so maybe this is a duplicate of https://github.com/bazelbuild/bazel/issues/14610 where the root issue is that `rule_input`s don't get configuration filtered, but target dep edges do?

Either way, it doesn't seem like there's a way to trace "Here are the files which impact this target in the current configuration", except possibly by performing a per-target `deps(target)` query - trying to batch this into a single `deps(...)` query seems to take away the ability to filter by configuration?

macOS

release 4.2.1

/cc @gregestren

illicitonion:
When the existing `--transitions` flag is set to a non-None value,
cquery will report the configuration its dependencies are configured in.
This allows for pruning away values from ruleInputs which are not
required in the current configuration.

Example output from running:
`bazel cquery --transitions=lite --output=jsonproto 'deps(...)'`:
```
[...]
        "ruleInput": ["@bazel_tools//src/conditions:host_windows", "@bazel_tools//src/tools/launcher:launcher", "@bazel_tools//tools/launcher:launcher_windows"],
        "configuredRuleInput": [{
          "label": "@bazel_tools//src/tools/launcher:launcher",
          "configurationChecksum": "01ec5513a9fc41b2a15570123817f3c2200ad9aeb21b1181d588a4b4f91d5693",
          "configurationId": 3
        }]
[...]
```

Previously on a non-Windows platform it was not possible to determine
that the two windows-specific ruleInputs are not actually required - now
we can see from the configuredRuleInput section that the
windows-specific dependencies have been pruned by selection, and we can
see that a transition has re-configured the ruleInput into the exec
configuration.

For dependencies which were not subject to transition (e.g. because
they're in a non-transition attribute), or which had no configuration
(e.g. because they're a source file), we add the label as a
ConfiguredRuleInput _without_ any configuration information. This
indicates that the dependency has not been pruned, but that the caller
should determine the correct configuration from context (probably be
determining whether it's a source file, and if so, considering it
un-configured, otherwise propagating the contextual ConfiguredTarget's
configuration).

Fixes #14610
Fixes #14617

Implementation-wise, this took roughly the following shape:
1. Extract TransitionResolver from being inline in TransitionsOutputFormatterCallback
2. Extract KnownTargetsDependencyResolver from TransitionsOutputFormatterCallback.FormatterDependencyResolver
3. Conditionally call these from ProtoOutputFormatterCallback depending
   on the --transitions flag.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
@@ -1,3 +1,16 @@
+// Copyright 2022 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 package com.google.devtools.build.lib.query2.cquery;
 
 import com.google.devtools.build.lib.analysis.DependencyKind;
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
@@ -1,0 +1,35 @@
+package com.google.devtools.build.lib.query2.cquery;
+
+import com.google.devtools.build.lib.analysis.DependencyKind;
+import com.google.devtools.build.lib.analysis.DependencyResolver;
+import com.google.devtools.build.lib.analysis.TargetAndConfiguration;
+import com.google.devtools.build.lib.causes.Cause;
+import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder;
+import com.google.devtools.build.lib.packages.Target;
+import com.google.devtools.build.lib.util.OrderedSetMultimap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class KnownTargetsDependencyResolver extends DependencyResolver {
+
+  private final Map<Label, Target> knownTargets;
+
+  public KnownTargetsDependencyResolver(Map<Label, Target> knownTargets) {
+    this.knownTargets = knownTargets;
+  }
+
+  @Override
+  protected Map<Label, Target> getTargets(
+      OrderedSetMultimap<DependencyKind, Label> labelMap,
+      TargetAndConfiguration fromNode,
+      NestedSetBuilder<Cause> rootCauses) {
+    return labelMap.values().stream()
+        .distinct()
+        .filter(Objects::nonNull)
+        .filter(knownTargets::containsKey)
+        .collect(Collectors.toMap(Function.identity(), knownTargets::get));
+  }
+}
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
@@ -13,6 +13,11 @@ import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+/**
+ * KnownTargetsDependencyResolver is a DependencyResolver which resolves statically over a known
+ * set of targets. It can be useful when performing queries over a known pre-resolved universe of
+ * targets.
+ */
 public class KnownTargetsDependencyResolver extends DependencyResolver {
 
   private final Map<Label, Target> knownTargets;
--- a/src/main/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallback.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallback.java
@@ -18,11 +18,16 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Ordering;
 import com.google.devtools.build.lib.actions.BuildConfigurationEvent;
 import com.google.devtools.build.lib.analysis.AnalysisProtosV2;
 import com.google.devtools.build.lib.analysis.AnalysisProtosV2.Configuration;
+import com.google.devtools.build.lib.analysis.DependencyResolver;
+import com.google.devtools.build.lib.analysis.InconsistentAspectOrderException;
+import com.google.devtools.build.lib.analysis.config.BuildOptions;
 import com.google.devtools.build.lib.analysis.config.ConfigMatchingProvider;
+import com.google.devtools.build.lib.analysis.config.transitions.TransitionFactory;
 import com.google.devtools.build.lib.buildeventstream.BuildEventStreamProtos;
 import com.google.devtools.build.lib.cmdline.Label;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
@@ -30,9 +35,14 @@ import com.google.devtools.build.lib.packages.Attribute;
 import com.google.devtools.build.lib.packages.AttributeFormatter;
 import com.google.devtools.build.lib.packages.ConfiguredAttributeMapper;
 import com.google.devtools.build.lib.packages.Rule;
+import com.google.devtools.build.lib.packages.RuleTransitionData;
+import com.google.devtools.build.lib.query2.cquery.CqueryOptions.Transitions;
 import com.google.devtools.build.lib.query2.engine.QueryEnvironment.TargetAccessor;
 import com.google.devtools.build.lib.query2.proto.proto2api.Build;
+import com.google.devtools.build.lib.query2.proto.proto2api.Build.ConfiguredRuleInput;
 import com.google.devtools.build.lib.query2.proto.proto2api.Build.QueryResult;
+import com.google.devtools.build.lib.query2.proto.proto2api.Build.Target;
+import com.google.devtools.build.lib.query2.proto.proto2api.Build.Target.Builder;
 import com.google.devtools.build.lib.query2.query.aspectresolvers.AspectResolver;
 import com.google.devtools.build.lib.query2.query.output.ProtoOutputFormatter;
 import com.google.devtools.build.lib.skyframe.BuildConfigurationKey;
@@ -47,6 +57,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /** Proto output formatter for cquery results. */
 class ProtoOutputFormatterCallback extends CqueryThreadsafeCallback {
@@ -101,9 +112,11 @@ class ProtoOutputFormatterCallback extends CqueryThreadsafeCallback {
   private final SkyframeExecutor skyframeExecutor;
   private final ConfigurationCache configurationCache = new ConfigurationCache();
   private final JsonFormat.Printer jsonPrinter = JsonFormat.printer();
+  @Nullable private final TransitionFactory<RuleTransitionData> trimmingTransitionFactory;
 
   private AnalysisProtosV2.CqueryResult.Builder protoResult;
 
+  private final HashMap<Label, com.google.devtools.build.lib.packages.Target> partialResultMap;
   private KeyedConfiguredTarget currentTarget;
 
   ProtoOutputFormatterCallback(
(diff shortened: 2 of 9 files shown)


gregestren:
FYI we have different people assigned to the associated issues. I think the right followup from our side is to sync our understandings, then provide feedback here. Please stay tuned.

src/main/java/com/google/devtools/build/lib/query2/cquery/KnownTargetsDependencyResolver.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder;
+import com.google.devtools.build.lib.packages.Target;
+import com.google.devtools.build.lib.util.OrderedSetMultimap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class KnownTargetsDependencyResolver extends DependencyResolver {
```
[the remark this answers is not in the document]
illicitonion: This is exactly a `FormatterDependencyResolver` - no new filter, it was just extracted to be a top-level class. Because `FormatterDependencyResolver` uses state from its outer class, extracting it out to be explicit about which state it depends on felt useful. The problem is that we need to use it from `ProtoOutputFormatterCallback` as well as `TransitionsOutputFormatterCallback` - these are separately invoked by different options to `cquery`, so in order to use this logic from both it either needs extracting or duplicating.

src/main/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallback.java
```diff
@@ -184,7 +206,35 @@ public void processOutput(Iterable<KeyedConfiguredTarget> partialResult)
...
+
+                targetBuilder.getRuleBuilder().addConfiguredRuleInput(
+                    ConfiguredRuleInput.newBuilder().setLabel(resolvedTransition.label().toString())
+                        .setConfigurationChecksum(options.checksum())
+                        .setConfigurationId(configurationId));
+              }
+            }
+          }
+        } catch (DependencyResolver.Failure | InconsistentAspectOrderException e) {
+          // This is an abuse of InterruptedException.
```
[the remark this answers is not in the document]
illicitonion: This was copied and pasted from `TransitionsOutputFormatterCallback`. I think what it's saying is: `processOutput` only has two checked exceptions it's allowed to throw, `IOException` and `InterruptedException`, so rather than changing the `processOutput` API (a larger change), or throwing a `RuntimeException` (generally frowned upon in Java, but 🤷 seems better than this), hide the error in an `InterruptedException`.

I tried adding the exception to the interface, but ended up with a cyclic dependency because it ended up needing the `com.google.devtools.build.lib.cmdline.BatchCallback` interface (very high-level) to depend on the analysis package (very low level).

I could use a `RuntimeException`, or introduce a new Exception type, but I assume whoever originally wrote this code decided it was not worthwhile.

[This code](https://github.com/bazelbuild/bazel/blob/6594e0ca76305464f7f43ee236b21699ac4d3892/src/main/java/com/google/devtools/build/lib/query2/engine/OutputFormatterCallback.java#L45-L59) has an even sketchier hack, we could do something similar if we needed?

illicitonion:
Thanks for the review - I added a bunch of javadoc to `KnownTargetsDependencyResolver` and `TransitionResolver` which hopefully clears some of this up - PTAL!

illicitonion:
@aranguyen would you be able to give this another pass? Thanks!

aranguyen:
@illicitonion I completely missed this, very sorry about that. I will have this reviewed again by eod today.

src/main/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallback.java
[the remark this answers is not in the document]
aranguyen: ok thanks for the explanation. Looking at the previous commit, it seems to be part of a major change to reporting of Starlark errors and the call stack, so I would leave it.

src/test/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallbackTest.java
```diff
@@ -213,6 +217,25 @@ public void testConfigurations() throws Exception {
             .findAny()
             .orElseThrow();
     assertThat(fileTargetProto.getConfigurationId()).isEqualTo(0);
+
+    // Targets whose deps have no transitions should appear without configuration information.
+    assertThat(Iterables.getOnlyElement(parentRuleProto.getTarget().getRule().getConfiguredRuleInputList()))
+        .isEqualTo(ConfiguredRuleInput.newBuilder().setLabel("//test:transition_rule").build());
```
aranguyen, line 223: Maybe this could be simplified to ?
```
assertThat(parentRuleProto.getTarget().getRule().getConfiguredRuleInputList())
        .containsExactly(
            ConfiguredRuleInput.newBuilder().setLabel("//test:transition_rule").build());
```

src/test/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallbackTest.java
```diff
@@ -29,11 +29,13 @@
 import com.google.devtools.build.lib.events.NullEventHandler;
```
aranguyen, line 29: Could you please make sure this test target build after the change? I did an import changelist and it did not build successfully. Even though the checks in this PR seem to have passed, the test for `ProtoOutputFormatterCallbackTest` was not included looking at the artifacts (only TransitionsOutputFormatterTest was tested). I'll do a bit more debugging tomorrow morning.

src/main/java/com/google/devtools/build/lib/query2/cquery/TransitionResolver.java
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * The configuration(s) this edge results in. This is a collection because a split transition
+     * may lead to a single attribute requesting a dependency in multiple configurations.
+     *
+     * If a target is depended on via two attributes, separate ResolvedTransitions should be used,
+     * rather than combining the two into a single ResolvedTransition with multiple options.
+     *
+     * If no transition was applied to an attribute, this collection will be empty.
+     */
+    abstract Collection<BuildOptions> options();
```
aranguyen, line 66: Is it possible to use `ImmutableCollection` here instead because AutoValue instances should be deeply immutable?

src/main/java/com/google/devtools/build/lib/query2/cquery/TransitionResolver.java
```diff
@@ -0,0 +1,168 @@
...
+    ImmutableMap<Label, ConfigMatchingProvider> configConditions =
+        keyedConfiguredTarget.getConfigConditions();
+
+    // Get a ToolchainContext to use for dependency resolution.
+    ToolchainCollection<ToolchainContext> toolchainContexts =
+        accessor.getToolchainContexts(target, config);
+    // We don't actually use fromOptions in our implementation of
+    // DependencyResolver but passing to avoid passing a null and since we have the information
+    // anyway.
+    deps =
```
aranguyen, line 127: would it be clearer to do `OrderedSetMultimap<DependencyKind, DependencyKey> deps = ...` and delete line 117?

aranguyen (commented):
thanks for adding the documentation, that was really helpful. I added a few more comments and sorry again for overlooking your responses last time.

sgowroji:
Hello @illicitonion, Could you please reply on the review comments. Thanks!

src/test/java/com/google/devtools/build/lib/query2/cquery/ProtoOutputFormatterCallbackTest.java
illicitonion: It worked for me in open source:
```
//src/test/java/com/google/devtools/build/lib/query2/cquery:ProtoOutputFormatterCallbackTest PASSED in 32.9s
```

I introduced a compile break to it, and it stopped compiling, so in a Bazel world this looks fine... I'm not sure what would be different internally, sorry!

src/main/java/com/google/devtools/build/lib/query2/cquery/TransitionResolver.java
illicitonion: Absolutely it would! I think originally this code before I moved it was in a try-catch, hence the separation :) Fixed

illicitonion:
@aranguyen All addressed - thanks for the review! Let me know if I can be helpful with the test compiling issue at all

aranguyen:
@illicitonion FYI, I had to do a small modification to your code for the internal tests to pass (instead of having `ProtoOutputFormatter.toTargetProtoBuffer` to return  `Build.Target.Builder`, I'm having it returning `Build.Target`. The import now awaiting internal review from @sdtwigg

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

gregestren:
> For dependencies which were not subject to transition (e.g. because
they're in a non-transition attribute), or which had no configuration
(e.g. because they're a source file), we add the label as a
ConfiguredRuleInput without any configuration information.

Hey @illicitonion . We're exploring some `cquery --output=proto` cleanups and rediscovered this point.

Was this choice important? What would you think if we universally include the configuration info? While redundant, it does make it easier to directly identify all metadata for any input.

illicitonion:
I would love to universally include configuration info. Yes please!

gregestren:
Do you know why that wasn't done in the first place?

illicitonion:
I believe the data wasn't readily available - the existing transition following mechanism only applied to transitioned targets, and there wasn't convenient access to the configuration of non-transitioned targets. If it's convenient (or useful) to get that information, it'd be great to do so - the ambiguity ("is this a source file with no configuration, or just untransitioned and in the current configuration") is annoying, just not annoying enough for me to get around to fixing myself :)

gregestren:
I see. I *think* we can handle the source file / non-source difference. I'll share a PR with whatever I can do.

PR #15039, opened 2022-03-14
Add thin_lto to NDK CC toolchain features

keith:
This is a direct copy of the unix toolchain thin_lto setup

Related: https://github.com/bazelbuild/bazel/issues/13287

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_ndk_cc_toolchain_config_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_ndk_cc_toolchain_config_template.txt
@@ -38,11 +38,17 @@ def _impl(ctx):
         ACTION_NAMES.lto_backend,
     ]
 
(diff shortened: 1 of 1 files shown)


jgavris:
Thanks @keith as always!

keith:
Looks like CI has an older version of the NDK than I was testing, I guess we might also have to force on lld in this case, which I found to be a bit of an issue with rules_foreign_cc in my testing, but might just need to be fixed there

jgavris:
Is there any test coverage for NDK 19 in CI (especially on macOS, since most of our mobile devs have Apple laptops)? Looking at the Buildkite pipeline I don't have a lot of confidence.

cpsauer:
Hey all! Was setting up the new NDK toolchain (https://github.com/bazelbuild/rules_android_ndk) and noticed it has a thin_lto feature built in. Maybe that obsoletes this?

keith:
dropping for new ndk rules

PR #15045, opened 2022-03-15
Use python3 on macOS

ob:
macOS 12.3 removed `/usr/bin/python2.7` and `/usr/bin/python`. We need to use `python3` now.

### DIFF
--- a/tools/objc/libtool.sh
+++ b/tools/objc/libtool.sh
@@ -62,9 +62,9 @@ function hash_objfile() {
   echo "$SYMLINK_NAME"
 }
 
-python_executable=/usr/bin/python2.7
+python_executable=/usr/bin/python3
 if [[ ! -x "$python_executable" ]]; then
-  python_executable=python
+  python_executable=python3
 fi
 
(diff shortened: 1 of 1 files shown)


keith (approved):
Tested a case where the python script is run and this works, the script does not need updates. I didn't catch this in previous testing because we don't have any libraries that hit the slow path of libtool.sh

kersson:
Our teams needed additional changes to unblock us: https://github.com/ButterflyNetwork/bazel/commit/938c427f351e1df50abdd857c69c134b60260e62

We're still on 4.2.2, so that's the base of that commit. We only tested these changes on macOS.

Wyverald:
@googlewalt asks:

> Do we need to set LANG="en_US.UTF-8" [_on line 81_]?  When I did the analogous change internally in [Google], not setting LANG gets me these errrors: [redacted]

keith:
I tested and did not have to set that, without knowing more about the error I'm not sure what conditions that shows up in to repro

PR #15047, opened 2022-03-15
Fix coverage runfiles directory issue

fmeum:
As discovered in https://github.com/bazelbuild/bazel/pull/15030 there
was another bug with the short circuiting coverage logic where tests
were not run from the correct directory. In that PR this issue is fixed
directly, with this change we instead make missing LCOV_MERGER be
deferred until after all test setup and run, which I think is more
future proof to other changes here, and also allows users to know their
tests are being run with coverage, and output the correct files, even in
the case they don't setup the coverage merger infrastructure.

Closes #15031.

PiperOrigin-RevId: 434715347


fmeum:
@Wyverald @keith

PR #15052, opened 2022-03-15
Always collect FileProvider's filesToBuild as data runfiles

### LINKED ISSUE #15043: sh_binary's data dependencies are not triggered during the build

When a `sh_binary` has a target in the `data` attribute, the target is not built automatically when the `sh_binary` target is built.

* Check out [rules_docker](https://github.com/bazelbuild/rules_docker/tree/0b1a033eff28eeff474573c7ed2d5b9d9bec4af1)
* `bazel clean`
* `bazel build //docs:update`
* `ls bazel-bin/docs/container.md_`
Although `bazel-bin/docs/container.md_` is the output of `//docs:container_doc`, which is depended by `//docs:update`, it is not built.

macOS

release 5.0.0-homebrew

fmeum:
Guidelines for Starlark rules suggest that `data`-like attributes on
rules should always merge the default outputs of rule targets into the
transitive runfiles. See:
https://docs.bazel.build/versions/main/skylark/rules.html#runfiles-features-to-avoid

As a result, Starlark rules generally don't (and shouldn't) explicitly
include their default outputs into their runfiles. Before this commit,
native rules did not merge these outputs in the same way as idiomatic
Starlark rules, which led to unexpectedly absent runfiles.

Fixes #15043

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/Runfiles.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/Runfiles.java
@@ -935,7 +935,8 @@ public final class Runfiles implements RunfilesApi {
     /** Collects runfiles from data dependencies of a target. */
     @CanIgnoreReturnValue
     public Builder addDataDeps(RuleContext ruleContext) {
-      addTargets(getPrerequisites(ruleContext, "data"), RunfilesProvider.DATA_RUNFILES);
+      addTargets(getPrerequisites(ruleContext, "data"), RunfilesProvider.DATA_RUNFILES,
+          ruleContext.getConfiguration().alwaysIncludeFilesToBuildInData());
       return this;
     }
 
@@ -952,16 +953,18 @@ public final class Runfiles implements RunfilesApi {
     @CanIgnoreReturnValue
     public Builder addTargets(
         Iterable<? extends TransitiveInfoCollection> targets,
-        Function<TransitiveInfoCollection, Runfiles> mapping) {
+        Function<TransitiveInfoCollection, Runfiles> mapping,
+        boolean alwaysIncludeFilesToBuildInData) {
       for (TransitiveInfoCollection target : targets) {
-        addTarget(target, mapping);
+        addTarget(target, mapping, alwaysIncludeFilesToBuildInData);
       }
       return this;
     }
 
     public Builder addTarget(TransitiveInfoCollection target,
-        Function<TransitiveInfoCollection, Runfiles> mapping) {
-      return addTargetIncludingFileTargets(target, mapping);
+        Function<TransitiveInfoCollection, Runfiles> mapping,
+        boolean alwaysIncludeFilesToBuildInData) {
+      return addTargetIncludingFileTargets(target, mapping, alwaysIncludeFilesToBuildInData);
     }
 
     @CanIgnoreReturnValue
@@ -976,7 +979,8 @@ public final class Runfiles implements RunfilesApi {
     }
 
     private Builder addTargetIncludingFileTargets(TransitiveInfoCollection target,
-        Function<TransitiveInfoCollection, Runfiles> mapping) {
+        Function<TransitiveInfoCollection, Runfiles> mapping,
+        boolean alwaysIncludeFilesToBuildInData) {
       if (target.getProvider(RunfilesProvider.class) == null
           && mapping == RunfilesProvider.DATA_RUNFILES) {
         // RuleConfiguredTarget implements RunfilesProvider, so this will only be called on
@@ -986,6 +990,15 @@ public final class Runfiles implements RunfilesApi {
         // of the memory use, though, since we have a whole lot of FileConfiguredTarget instances.
         addTransitiveArtifacts(target.getProvider(FileProvider.class).getFilesToBuild());
         return this;
+      }
+
+      if (alwaysIncludeFilesToBuildInData && mapping == RunfilesProvider.DATA_RUNFILES) {
+        // Ensure that `DefaultInfo.files` of Starlark rules is merged in so that native rules
+        // interoperate well with idiomatic Starlark rules..
+        // https://bazel.build/extending/rules#runfiles_features_to_avoid
+        // Internal tests fail if the order of filesToBuild is preserved.
+        addTransitiveArtifacts(NestedSetBuilder.<Artifact>stableOrder()
+            .addTransitive(target.getProvider(FileProvider.class).getFilesToBuild()).build());
       }
 
       return addTargetExceptFileTargets(target, mapping);
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -435,7 +435,7 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
 
   @Option(
       name = "incompatible_always_include_files_in_data",
-      defaultValue = "true",
+      defaultValue = "false",
       documentationCategory = OptionDocumentationCategory.OUTPUT_SELECTION,
       effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
       metadataTags = {OptionMetadataTag.INCOMPATIBLE_CHANGE},
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -434,6 +434,17 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
   public boolean legacyExternalRunfiles;
 
   @Option(
+      name = "incompatible_always_include_files_in_data",
+      defaultValue = "true",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_SELECTION,
+      effectTags = {OptionEffectTag.AFFECTS_OUTPUTS},
+      metadataTags = {OptionMetadataTag.INCOMPATIBLE_CHANGE},
+      help = "If true, native rules add <code>DefaultInfo.files</code> of data dependencies to "
+          + "their runfiles, which matches the recommended behavior for Starlark rules ("
+          + "https://bazel.build/extending/rules#runfiles_features_to_avoid).")
+  public boolean alwaysIncludeFilesToBuildInData;
+
+  @Option(
       name = "check_fileset_dependencies_recursively",
       defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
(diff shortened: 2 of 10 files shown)


src/test/java/com/google/devtools/build/lib/starlark/StarlarkIntegrationTest.java
```diff
@@ -723,6 +723,66 @@ public void testDefaultInfoWithRunfilesConstructor() throws Exception {
...
+        .contains("cr.out");
+    assertThat(
+        ActionsTestUtil.baseArtifactNames(
+            target.getProvider(RunfilesProvider.class).getDataRunfiles().getAllArtifacts()))
+        .contains("cr.out");
+  }
+
+  @Test
+  public void testDefaultInfoFilesAddedToCcBinaryTargetRunfiles() throws Exception {
+    helperDefaultInfoFilesAddedToNativeBinaryTargetRunfiles("cc_binary");
```
comius, line 761: Readability and testing best practices: 
Please inline `helperDefaultInfoFilesAddedToNativeBinaryTargetRunfiles` in all 3 tests. 
This helps debugging when the test breaks.

comius (changes requested):

comius:
This looks reasonable. Mind that we're rewriting all the rules where this applies to Starlark.

I'll run internal unit tests, to see if this breaks anything, but I don't expect it to.

comius:
> This looks reasonable. Mind that we're rewriting all the rules where this applies to Starlark.
> 
> I'll run internal unit tests, to see if this breaks anything, but I don't expect it to.

There are some breakages on internal unit tests, but it seems just the tests have to be updated.

src/test/java/com/google/devtools/build/lib/starlark/StarlarkIntegrationTest.java
fmeum: Done. I will keep this in mind for future contributions.

fmeum:
> This looks reasonable. Mind that we're rewriting all the rules where this applies to Starlark.
> 
> I'll run internal unit tests, to see if this breaks anything, but I don't expect it to.

Thanks!

fmeum:
@comius Did you get a chance to look into the internal unit tests blocking this?

comius:
> @comius Did you get a chance to look into the internal unit tests blocking this?

I've put this on my backlog, I will need to do a bit more research and that will take more time.

What happens is that `.ifso` files from data dependency of a dynamic `cc_binary` (`linkstatic=0`) get added to runfiles, and there's tests, that verifies this doesn't happen.

And `.so` and `.a` gets added from data dep in the case of static cc_binary (`linkstatic=1`)

fmeum:
@comius Is there anything I can do to help get this into 5.3.0?

The issue fixed by the PR is, based on my personal experience, one of the top gotchas for users writing their first custom rules (including myself and many of my coworkers at some point). It also frequently comes up on the Bazel Slack (most recently at https://bazelbuild.slack.com/archives/CA31HN1T3/p1657044541910759?thread_ts=1657044541.910759&cid=CA31HN1T3). Getting it fixed in Bazel 5 would also help rule authors ditch `data_runfiles` possibly 9 months earlier.

Would it help if I guarded the change behind a flag that you could flip back inside Google? Or, alternatively, can the failing tests be open-sourced? I could then try to fix the runfile pollution issues in the cc rules.

fmeum:
@comius Friendly ping. This is still coming up on Bazel Slack over and over again as a major inconsistency. 

Do you think we can get this into Bazel 6? It looks like Starlark `cc_binary` already has the fixed behavior, otherwise the test wouldn't pass. Presumably the internal tests have been fixed?

src/main/java/com/google/devtools/build/lib/analysis/Runfiles.java
```diff
@@ -980,6 +980,13 @@ private Builder addTargetIncludingFileTargets(TransitiveInfoCollection target,
         return this;
       }
 
+      if (mapping == RunfilesProvider.DATA_RUNFILES) {
+        FileProvider fileProvider = target.getProvider(FileProvider.class);
```
comius, line 984: Can you remove the branch above L#972 to L#981, but keep the TODO comment? I believe the new branch completely covers it.

The case there's no RunfilesProvider should not raise any exceptions. Also calling `addTargetExceptFileTargets` where there's no RunfilesProvider shouldn't make a difference.

comius (changes requested):
I did the changes needed internally - and I'm waiting for feedback from reviewers.

comius (changes requested):
I have a lot of failures running this change through the depot, I still need to analyse them thought. 

It makes sense to put this behind an incompatible flag - because others might have similar problems (and if I don't fix all the failures internally we can still release this in Bazel).

One bug that I already located in the run and you can probably fix from stack trace:
```
Caused by: java.lang.IllegalArgumentException: Order mismatch: preorder != postorder
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:441)
	at com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder.addTransitive(NestedSetBuilder.java:138)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addTransitiveArtifacts(Runfiles.java:777)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addTargetIncludingFileTargets(Runfiles.java:988)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addTarget(Runfiles.java:964)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addTargets(Runfiles.java:957)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addDataDeps(Runfiles.java:938)
	at com.google.devtools.build.lib.analysis.Runfiles$Builder.addRunfiles(Runfiles.java:910)
	at com.google.devtools.build.lib.view.sh.ShBinary.create(ShBinary.java:71)
	at com.google.devtools.build.lib.view.sh.ShBinary.create(ShBinary.java:30)
	at com.google.devtools.build.lib.analysis.ConfiguredTargetFactory.createRule(ConfiguredTargetFactory.java:371)
```

fmeum:
I'm actually quite confused by the stack trace: I expected both `filesToBuild` and runfiles to use stable order nested set, but the exception clearly shows that *neither* are using the stable order in this case. Is that possible and something that the code should be able to handle?

comius:
> I'm actually quite confused by the stack trace: I expected both `filesToBuild` and runfiles to use stable order nested sets, but the exception clearly shows that _neither_ are using the stable order in this case. Is that possible and something that the code should be able to handle?

You can rewrap NestedSet to a stable order which will prevent the exception. Example is Runfiles.addTransitiveArtifactsWrappedInStableOrder

fmeum:
Since `Runfiles.Builder` is instantiated in ~100 locations, I chose a different strategy to pass in the flag.

@comius Should the flag be flipped right away (as is done in this PR to get realistic test coverage) or not?

jheaff1:
Sorry to butt in on this PR but I seem to experiencing the issue that this PR aims to resolve but I tested with this PR and sadly the issue still occurs.

My scenario is essentially a `sh_binary` being fed into a `native_binary` from skylib, the output of which is used as a “tool” in the dependent rule.

If I do `Bazel build <dependent rule>`, the runfiles of the `sh_binary` don’t get built. But if I do `bazel build <native_binary>` then the runfiles of the `sh_binary` do get built.

I can provide a repro if that’s helpful.

fmeum:
@jheaff1 A repo would be helpful, yes.

At a glance, it *may be the case* that `native_binary` is affected by the same issue this PR is trying to solve. Looking at https://github.com/bazelbuild/bazel-skylib/blob/8a6ab72c6ece96cab91e17aaf4fb5339615d1d3d/rules/native_binary.bzl#L41
it seems that it is merging in the `default_runfiles` from data dependencies, which is neither what native rules do (they would merge in `data_runfiles`) nor what Starlark rules should do (merge in `default_runfiles` and `files`). Of course this may not be the reason for what you are seeing, it was just something I noticed.

jheaff1:
@fmeum Please see repro here:

https://github.com/jheaff1/issue-repros/tree/main/bazel/runfiles-not-built

fmeum:
@jheaff1 I don't have access to a Windows machine at the moment, but I think I now what's causing this behavior. I don't think it is related to this PR or the issue it fixes, but originates from the interplay between skylib's `copy_file` and `native_binary`.

In your repro, since the `copy_file` rule doesn't have the `is_executable` attribute set, its `DefaultInfo` will have the `native_bin` script in `files` and `data_runfiles` (see https://github.com/bazelbuild/bazel-skylib/blob/6a17363a3c27dde70ab5002ad9f2e29aff1e1f4b/rules/private/copy_file_private.bzl#L85). But `native_binary` only merges the `default_runfiles` of its `data` dependencies (see https://github.com/bazelbuild/bazel-skylib/blob/6a17363a3c27dde70ab5002ad9f2e29aff1e1f4b/rules/native_binary.bzl#L35-L38), which don't contain `native_bin`.

A workaround would be to set `is_executable` on `copy_file` so that `native_bin` is included in the `default_runfiles`, but the proper fix is to have `native_binary` merge in the `files` of its `data` dependencies.

jheaff1:
@fmeum Thank you very much for investigating the issue, it’s greatly appreciated!

Apologies for hijacking your PR.

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -433,6 +433,17 @@ public ExecConfigurationDistinguisherSchemeConverter() {
               + ".runfiles/wsname/external/repo (in addition to .runfiles/repo).")
   public boolean legacyExternalRunfiles;
 
+  @Option(
+      name = "incompatible_always_include_files_in_data",
+      defaultValue = "true",
```
comius, line 438: I'd need "false" here, if I want to import. I can flip it to true as soon as it's released internally.

comius (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
fmeum: Changed

fmeum:
@bazel-io flag

fmeum:
@meteorcloudy If it isn't too late for that, could you run a downstream pipeline with this flag flipped? If still possible, what would I (rather than a googler) need to do to get this flipped for Bazel 6?

comius (approved):

fmeum:
@comius @meteorcloudy Sorry, forgot to update the tests. Should be fixed now.

comius (approved):

meteorcloudy:
@fmeum Can you follow https://bazel.build/contribute/breaking-changes to file a tracking issue for this incompatible change? 

Once this is merged, I'll add the correct labels and it'll be tested in the [incompatible flag pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags)

fmeum:
@meteorcloudy I created https://github.com/bazelbuild/bazel/issues/16654.

fmeum:
@bazel-io flag

fmeum:
@meteorcloudy This has been merged, could you add it to the pipeline?

meteorcloudy:
@bazel-io fork 6.0.0

meteorcloudy:
@fmeum This flag is already tested in https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1325, currently failing with "Unrecognized option" error, we'll get some real result tomorrow ;)

PR #15060, opened 2022-03-17
Set C++ archive action mnemonic to CppArchive

keith:
Previously the archiver action shared the CppLink mnemonic which made it
impossible to differentiate with this like --modify_execution_info


keith:
Considering this is public API I'm not sure what considerations there are about merging something like this?

ckolli5:
Hello @keith, can you please check presubmit failures. Thanks!

keith:
@ckolli5 updated

brentleyjones:
@bazel-io flag

brentleyjones:
Actually, this is probably a breaking change, and can't get included in 5.x. You can just remove the label @ckolli5.

meisterT:
Agreed, this should be in 6.0 but not 5.x.

PR #15061, opened 2022-03-17
remote: Proactively close the ZstdInputStream in ZstdDecompressingOutputStream.

benjaminp:
ZstdInputStream hangs onto some native memory, which should be released as soon as ZstdDecompressingOutputStream is done being used rather than when the finalizer runs.


coeuvre (approved):
Thanks!

brentleyjones:
@bazel-io fork 5.2

PR #15066, opened 2022-03-17
[5.1] Undocument --bes_best_effort

Wyverald:
We've heared from several people that they tried passing the flag and it did
not have any efect (as expected from a no-op), so let's undocument the flag
so it's no longer discoverable.

On a related note, we should document `--bes_upload_mode` which seems to be
what people want to do.

Closes #14333.

PiperOrigin-RevId: 435317406


Yannic:
Thanks!

PR #15068, opened 2022-03-17
[5.1] Fix conflicting actions error when specifying --host_macos_minimum_os

brentleyjones:
The flag was introduced in #13001 but not usable internally, because
the host flag was not saved when doing a transition.  Host flags
should be saved, like in:

https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java#L1212

(cherry picked from commit b8a2ee2ca8ae1f66b40baea8fc413b2732e7ebaa)

Fixes issues like https://github.com/bazelbuild/rules_swift/issues/777.


brentleyjones:
@Wyverald

lberki:
Given that Tony reviewed #13001, I assume he is the right person for this one, too?

/cc @gregestren

thii:
This is only a cherry-pick of a change that is already landed.

PR #15075, opened 2022-03-18
Support dict select union

AlessandroPatti:
Starlark has recentely added support for union operations over dictionaries (https://github.com/bazelbuild/starlark/pull/215). The syntax is already supported in bazel as of https://github.com/bazelbuild/bazel/pull/14540, but the same operation with selects of dictionaries is still usupported.

Related issue: #12457

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
@@ -142,7 +142,12 @@ public final class SelectorList implements StarlarkValue, HasBinary {
   @Override
   @Nullable
   public SelectorList binaryOp(TokenKind op, Object that, boolean thisLeft) throws EvalException {
-    if (op == TokenKind.PLUS) {
+    if (getNativeType(that).equals(Dict.class)) {
+      if (op == TokenKind.PIPE) {
+        return thisLeft ? concat(this, that) : concat(that, this);
+      }
+    }
+    else if (op == TokenKind.PLUS) {
       return thisLeft ? concat(this, that) : concat(that, this);
     }
     return null;
--- a/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
@@ -121,7 +121,7 @@ public final class SelectorList implements StarlarkValue, HasBinary {
       }
       if (!canConcatenate(getNativeType(firstValue), getNativeType(value))) {
         throw Starlark.errorf(
-            "'+' operator applied to incompatible types (%s, %s)",
+            "Cannot combine incompatible types (%s, %s)",
             getTypeName(firstValue), getTypeName(value));
       }
     }
@@ -142,11 +142,6 @@ public final class SelectorList implements StarlarkValue, HasBinary {
   @Override
   @Nullable
   public SelectorList binaryOp(TokenKind op, Object that, boolean thisLeft) throws EvalException {
-    if (!canConcatenate(getNativeType(this), getNativeType(that))) {
-      throw Starlark.errorf(
-          "'%s' operator applied to incompatible types (%s, %s)",
-          op.toString(), getTypeName(thisLeft ? this : that), getTypeName(thisLeft ? that : this));
-    }
     if (getNativeType(that).equals(Dict.class)) {
       if (op == TokenKind.PIPE) {
         return thisLeft ? concat(this, that) : concat(that, this);
--- a/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
+++ b/src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
@@ -103,9 +103,9 @@ public final class SelectorList implements StarlarkValue, HasBinary {
    * Creates a list from the given sequence of values, which must be non-empty. Each value may be a
    * native type, a select over that type, or a selector list over that type.
    *
-   * @throws EvalException if all values don't have the same underlying type
+   * @throws IllegalStateException if all values don't have the same underlying type
    */
-  static SelectorList of(Iterable<?> values) throws EvalException {
+  static SelectorList of(Iterable<?> values) {
     Preconditions.checkArgument(!Iterables.isEmpty(values));
     ImmutableList.Builder<Object> elements = ImmutableList.builder();
     Object firstValue = null;
@@ -120,9 +120,10 @@ public final class SelectorList implements StarlarkValue, HasBinary {
         firstValue = value;
       }
       if (!canConcatenate(getNativeType(firstValue), getNativeType(value))) {
-        throw Starlark.errorf(
-            "'+' operator applied to incompatible types (%s, %s)",
-            getTypeName(firstValue), getTypeName(value));
+        throw new IllegalStateException(
+            String.format(
+                "This list should have been validated on creation but found incompatible types (%s, %s)",
+                getTypeName(firstValue), getTypeName(value)));
       }
     }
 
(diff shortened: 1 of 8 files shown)


fmeum:
@tetromino Would you be able to review this? It's the reason why I worked on introducing the dict union operator in the first place.

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
```diff
@@ -110,7 +110,12 @@ static SelectorList concat(Object x, Object y) throws EvalException {
 
   @Override
   public SelectorList binaryOp(TokenKind op, Object that, boolean thisLeft) throws EvalException {
-    if (op == TokenKind.PLUS) {
+    if (getNativeType(that).equals(Dict.class)) {
```
tetromino, line 113: The error message thrown when concatenating the wrong types using the wrong operator (like 146 of SelectorList.java) is now misleading (currently it's hard-coded as "'+' operator applied ..."). To fix, you'd need to pass the operator name from binaryOp to canConcatenate.

tetromino, line 113: You'd need to also make the same change to SelectorValue.binaryOp (and ensure it's tested)

src/main/java/com/google/devtools/build/lib/packages/Type.java
```diff
@@ -559,6 +559,15 @@ public Map<KeyT, ValueT> convert(Object x, Object what, Object context)
       return ImmutableMap.copyOf(result);
     }
 
+    @Override
+    public Map<KeyT, ValueT> concat(Iterable<Map<KeyT, ValueT>> iterable) {
```
tetromino, line 563: Docstring of `Type.concat()` is now misleading - it should be something like `Implementation of concatenation for this type, as if by {@code elements[0] + ... + elements[n-1]} for scalars or lists, or {@code elements[0] | ... | elements[n-1]} for dicts`

tetromino (changes requested):
Please add tests for type/operator mismatches: dict | list, list | dict, dict + list, list + dict.

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
AlessandroPatti: Updated it but I'm usure how to test it. As far as a can tell, `SelectorValue` objects never live on their own, but only as elements of `SelectorList`, so there would never be a binary operation involving SelectorValue evaluated from Starlark. I could test the native code, but since `SelectorValue.binaryOp` defers to `SelectorList.binaryOp` after this change, I'm not sure how much value that would bring. @tetromino WDYT?
cc @alandonovan since I see a few comment about merging SelectorValue and SelectorList and this seems relevant

AlessandroPatti: Ended up moving the validation up to `binaryOp`, since `SelectorList.of` is called in other places where there is not a binary operator available (e.g.  [BuildType.java#L493](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/packages/BuildType.java#L493) and [StarlarkNativeModule.java#L704](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/packages/StarlarkNativeModule.java#L704))

fmeum:
@comius AFAIK tetromino won't be available until July. Is there anybody else who could take over the review?

src/main/java/com/google/devtools/build/lib/packages/Type.java
```diff
@@ -559,6 +560,15 @@ public Map<KeyT, ValueT> convert(Object x, Object what, Object context)
...
+      LinkedHashMap<KeyT, ValueT> output = new LinkedHashMap<>();
```
adonovan, line 565: Presumably LinkedHashMap is used instead of ImmutableMap.Builder (at the cost of a copy) to avoid an exception when the keys intersect. A comment to this effect would be good.

Better still: use Dict.Builder, which avoids the exception, and will take advantage of the exception-free builder semantics (buildKeepingLast) when they become available.

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
adonovan: I'm no longer on the Blaze team and never got as far as merging these two classes ([three](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/BuildType.java;l=416;bpv=1;bpt=1?q=BuildType.SelectorList&ss=bazel%2Fbazel) actually). I think your analysis is basically right, though I vaguely recall having seen edge cases in which a SelectorValue escapes from a SelectorList and is seen as a first class Starlark value. In other words, if we merged all three classes, some programs might observe an incompatible change, but I don't think that's a strong reason not to make the change, which would be a significant simplification.

@brandjon may remember more.

src/main/java/com/google/devtools/build/lib/packages/Type.java
fmeum: `buildKeepingLast` is in fact already available: https://cs.opensource.google/bazel/bazel/+/4eb62cd07393e495ab9ead49b771871d1050ee51:src/main/java/com/google/devtools/build/lib/actions/ActionEnvironment.java;l=111

AlessandroPatti: Updated to use `buildKeepingLast`, although from the docs:
> In the current implementation, all values associated with a given key are stored in the Builder object, even though only one of them will be used in the built map. If there can be many repeated keys, it may be more space-efficient to use a [LinkedHashMap](https://docs.oracle.com/javase/9/docs/api/java/util/LinkedHashMap.html?is-external=true) and [ImmutableMap.copyOf(Map)](https://guava.dev/releases/31.1-jre/api/docs/com/google/common/collect/ImmutableMap.html#copyOf(java.util.Map)) rather than ImmutableMap.Builder.

AlessandroPatti:
@fmeum @tetromino @comius Any update on this? I'd be awesome to have it in the next release

fmeum:
@AlessandroPatti Not a Googler, so I can't do more than stating that this LGTM.

@brandjon Do you happen to have cycles to get this into 5.3.0?

fmeum:
@comius @tetromino Friendly ping. This has been in review for almost half a year now, it would be great to finally get it merged.

AlessandroPatti:
@comius @tetromino @gregestren Friendly ping on this, is there any blocker for merging this?

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
```diff
@@ -132,15 +133,25 @@ static SelectorList of(Iterable<?> values) throws EvalException {
...
+    if (!canConcatenate(getNativeType(this), getNativeType(that))) {
+      throw Starlark.errorf(
+          "'%s' operator applied to incompatible types (%s, %s)",
+          op.toString(), getTypeName(thisLeft ? this : that), getTypeName(thisLeft ? that : this));
+    }
+    if (getNativeType(that).equals(Dict.class)) {
+      if (op == TokenKind.PIPE) {
+        return thisLeft ? concat(this, that) : concat(that, this);
+      }
+    }
```
tetromino, line 153: This means we are pointlessly checking canConcatenate twice: in `binaryOp` and in `of()`.

A solution, IMHO, would be to make `concat` a private method which simply does `return new SelectorList(getNativeType(x), ImmutableList.of(x, y))` without performing any `canConcatenate` check. It should be easy to do since we're no longer calling `concat` from SelectorValue.

tetromino:
My apologies for the very late review!

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
```diff
@@ -102,9 +102,9 @@ static SelectorList of(SelectorValue selector) {
    * Creates a list from the given sequence of values, which must be non-empty. Each value may be a
    * native type, a select over that type, or a selector list over that type.
    *
-   * @throws EvalException if all values don't have the same underlying type
+   * @throws IllegalStateException if all values don't have the same underlying type
    */
-  static SelectorList of(Iterable<?> values) throws EvalException {
+  static SelectorList of(Iterable<?> values) {
```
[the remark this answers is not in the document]
AlessandroPatti: From my understanding, the other two places where this function is called are not performing any evaluation, but rather exposing this data structure in a human-friendly way. This is why I assumed that this should never happen (assuming the validation during the creation of the `SelectorList` is correct) and it is thus an "illegal state" if it happens. Regardless, reverted to throw an EvenException if this happens to be on the safe side.

src/main/java/com/google/devtools/build/lib/packages/SelectorList.java
AlessandroPatti: Tried making `concat` private and return a new SelectorList directly but
  - A few tests use `concat`, so the visibility cannot be private, it would require using `@VisibileForTesting`
  - `of` flattens the SelectorLists elements, which we would not do here instantiating the SelectorList directly, causing errors like `expected type <type> for attribute <attr>, but got select`

Instead I removed the check entirely so we rely on `of` to fail if the types cannot be combined.

tetromino (approved):
Thank you, looks good now!

PR #15081, opened 2022-03-19
Let JacocoCoverageRunner identify the classpath on JDK 16+

fmeum:
The Unsafe-backed extraction of the classpath in
`JacocoCoverageRunner#getClassLoaderUrls` has to be adapted for JDK 16+
as the private `ucp` field on `AppClassLoader` has been moved to its
superclass.

With this commit, the existing coverage integration tests pass for the
JDK 17 toolchain.

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
@@ -398,8 +398,17 @@ public class JacocoCoverageRunner {
         field.setAccessible(true);
         Unsafe unsafe = (Unsafe) field.get(null);
 
-        // jdk.internal.loader.ClassLoaders.AppClassLoader.ucp
-        Field ucpField = classLoader.getClass().getDeclaredField("ucp");
+        Field ucpField;
+        try {
+          // Java 9-15:
+          // jdk.internal.loader.ClassLoaders.AppClassLoader.ucp
+          ucpField = classLoader.getClass().getDeclaredField("ucp");
+        } catch (NoSuchFieldException e) {
+          // Java 16+:
+          // jdk.internal.loader.BuiltinClassLoader.ucp
+          // https://github.com/openjdk/jdk/commit/03a4df0acd103702e52dcd01c3f03fda4d7b04f5#diff-32cc12c0e3172fe5f2da1f65a75fa1cb920c39040d06323c83ad2c4d84e095aaL147
+          ucpField = classLoader.getClass().getSuperclass().getDeclaredField("ucp");
+        }
         long ucpFieldOffset = unsafe.objectFieldOffset(ucpField);
         Object ucpObject = unsafe.getObject(classLoader, ucpFieldOffset);
 
--- a/src/test/shell/bazel/BUILD
+++ b/src/test/shell/bazel/BUILD
@@ -514,7 +514,8 @@ sh_test(
         ],
         tags = ["no_windows"],
     )
-    for java_version in JAVA_VERSIONS_COVERAGE
+    # FIXME: Update JAVA_VERSIONS_COVERAGE on the next java_tools release.
+    for java_version in JAVA_VERSIONS_COVERAGE + ("17",)
 ]
 
 sh_test(


comius:
cc @c-mita

src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
```diff
@@ -398,8 +398,17 @@ private static String getUniquePath(String pathTemplate, String suffix) throws I
         field.setAccessible(true);
         Unsafe unsafe = (Unsafe) field.get(null);
 
-        // jdk.internal.loader.ClassLoaders.AppClassLoader.ucp
-        Field ucpField = classLoader.getClass().getDeclaredField("ucp");
+        Field ucpField;
+        try {
+          // Java 9-15:
+          // jdk.internal.loader.ClassLoaders.AppClassLoader.ucp
+          ucpField = classLoader.getClass().getDeclaredField("ucp");
```
comius, line 405: I think`ucpField = classLoader.getClass().getField("ucp");` looks also into super up. Using it would be nicer than catching an exception an retrying.

comius (changes requested):

comius (approved):

fmeum:
@comius Friendly ping. Is the import the only remaining step?

src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverageRunner.java
cushon: I wonder if `getClassLoaderUrls` has requirements that aren't met by [Guava's ClassPath](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/reflect/ClassPath.html) or [ClassGraph](https://github.com/classgraph/classgraph)?

It would be nice to avoid layering on reflection here, especially since it's getting harder to reflect on internal APIs with [JEP-403](https://openjdk.java.net/jeps/403) etc.

fmeum: > I wonder if `getClassLoaderUrls` has requirements that aren't met by [Guava's ClassPath](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/reflect/ClassPath.html) or [ClassGraph](https://github.com/classgraph/classgraph)?

I think that either of those two could be used here. Most of the complexity ClassGraph deals with, in particular special class loaders offeres by frameworks such as Spring, isn't relevant here, but of course that's not a reason not to use it.

The function in ClassGraph that implements the equivalent of `getClassLoaderUrls` is [here]. Note that it does use a native library as a helper to work around restrictions on reflection usage.

> It would be nice to avoid layering on reflection here, especially since it's getting harder to reflect on internal APIs with [JEP-403](https://openjdk.java.net/jeps/403) etc.

I think that at least JEP-403 is not something we need to worry about here: AFAIU, strong encapsulation means that `setAccessible(true)` calls on private or protected fields will fail, but `getClassLoaderUrls` already circumvents this by using `Unsafe` to directly read the field contents from memory. This strategy should be future-proof for as long as `sun.misc.Unsafe` itself remains accessible.

cushon: Thanks. I realized we had a separate internal fix for a related issue that's doing something similar to what `ClassPath` is doing here: https://github.com/google/guava/blob/2a268346329c3ce17ad1269c5c0206ddd2747a17/guava/src/com/google/common/reflect/ClassPath.java#L642

Porting that fix to `JacocoCoverageRunner` would look like this: https://bazel-review.googlesource.com/c/bazel/+/194693

What do you think about that as an alternative to the reflection here?

I agree that this approach is going to work for existing releases, but I'd still prefer to avoid `Unsafe` if possible.

fmeum: > What do you think about that as an alternative to the reflection here?

The alternative should be correct in the most common cases (although I just noticed that the Bazel CL you linked also removes the special handling for wrapped jars without replacement, is it no longer needed?). 

It could cause issues with Java agents since those can append to the system or bootstrap class loader path via [`Instrumentation.appendToSystemClassLoaderSearch`](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#appendToSystemClassLoaderSearch-java.util.jar.JarFile-). The `premain` of a Java agent would run before `JacocoCoverageRunner`, but since these functions modify `ucp` directly, the new classpath entries wouldn't be reflected in `java.class.path` (see https://github.com/classgraph/classgraph/blob/e0ca965dae5ac74299be7c3d7b26f003f7163575/src/main/java/nonapi/io/github/classgraph/classloaderhandler/JPMSClassLoaderHandler.java#L94 for more details on this).

This is not even that theoretical: I'm currently working on improvements to `bazel coverage` because I want to get full end-to-end coverage support for [`Jazzer`](https://github.com/CodeIntelligenceTesting/jazzer), which has most of its code in a Java agent. I would thus like to keep the ability to get coverage information for agents.

cushon: I see, thanks. Would you be willing to investigate the `ClassPath` or `ClassGraph`-based approaches?

fmeum: Yes, I will go over the possible approaches and let you know about their pros and cons.

fmeum: I looked into the different approaches in detail and found the following.

**Guava's `ClassPath`**: Parses `java.class.path` and returns a list of `URL`s via [`ClassPath.parseJavaClassPath`](https://github.com/google/guava/blob/6ec853ce614e1184876be90d195e1180ca3ab6c8/guava/src/com/google/common/reflect/ClassPath.java#L640), but doesn't expose this functionality as public API. Efforts to change that are tracked in b/65488446, but given that the class doc recommends using `ClassGraph` instead, I don't know how likely this is to be fixed.
A general downside of this approach is that `java.class.path` doesn't reflect the full class path of the system class loader: Jars specified via the `-javaagent` JVM option and those added via [`Instrumentation#appendToSystemClassLoaderSearch`](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#appendToSystemClassLoaderSearch-java.util.jar.JarFile-) are not added to the properties value. Going with this approach would thus mean that coverage wouldn't be collected correctly for Java agents without workarounds.

**ClassGraph**: Reads the `ucp` field of the system class loader. To do this on JDK 17, it delegates to [Narcissus](https://github.com/toolfactory/narcissus) or [JVM-Driver](https://github.com/toolfactory/jvm-driver), both of which ultimately rely on a JNI library to circumvent strong encapsulation. Narcissus [doesn't seem to support M1 Macs or Windows on ARM yet](https://github.com/toolfactory/narcissus/tree/main/lib), but that could be worked around by building the dependency from source with Bazel instead of importing a jar. New versions of the JDK that change the relevant internals would still require an update of this dependency.
An advantage of ClassGraph is that it could handle many custom class loaders if those are used as the system class loader via the `java.system.class.loader` property. I don't know how common that is though.

**https://bazel-review.googlesource.com/c/bazel/+/194693**: Essentially does what Guava does, sharing the pros (only uses public APIs) and cons (doesn't handle Java agents).

**This PR**: Uses the same approach as ClassGraph. It doesn't require a JNI library or third-party dependency.

Based on my analysis, I would say that neither ClassGraph nor Guava's `ClassPath` are viable alternatives at this point (although this may change as part of b/65488446). Given that I have a concrete use case for Java agent coverage collection, I would prefer a solution that reads `ucp` of the system class loader. Given that this part of the code has only required changes twice since Java 8 and would only break fundamentally if `Unsafe` were to be removed without replacement, I would personally find the maintenance burden acceptable. But I can understand if you arrive at a different conclusion and would, in that case, propose to go with your Bazel CL, possibly with the wrapper jar logic added back.

@cushon What do you think?

cushon: Thanks for the thorough investigation here, I agree there don't seem to be great alternatives that would provide support for instrumenting agents.

For my understanding, is the Jazzer use-case to get Java coverage support when running Jazzer's own unit tests using the `java_test` rule?

fmeum: > For my understanding, is the Jazzer use-case to get Java coverage support when running Jazzer's own unit tests using the `java_test` rule?

Yes. Jazzer itself is a `cc_binary` that starts a JVM with a `java_binary` deploy jar attached as an agent. We have a `java_test`-based macro that runs Jazzer on a number of example projects as an end-to-end test and would use these tests to collect coverage reports for Jazzer itself. [rules_jni](https://github.com/fmeum/rules_jni/commits/main) recently gained the ability to do this, but it requires `JacocoCoverageRunner` to see java agent jars.

cushon (approved):

fmeum:
@cushon How does the process for a new java_tools release look like? Do I have to do anything else to get the change in now that it has been merged to master?

comius:
> @cushon How does the process for a new java_tools release look like? Do I have to do anything else to get the change in now that it has been merged to master?

cc @hvadehra

Issue #7069, opened 2019-01-09
Support `android_instrumentation_test` on macOS


jin:
Tracking issue for `android_instrumentation_test` support on macOS.

jin:
This is blocked on the availability of Project Nitrogen publicly.

jin:
While integration with Nitrogen is WIP, I've made a experimental rule that uses the Composer test runner, and it works on a Mac: https://github.com/jin/rules_experimental_android#composer-integration

jinchizhou:
Is `android_instrumentation_test` still not supported on macOS? I'm still getting https://github.com/bazelbuild/bazel/issues/9452.

nkoroste:
@jin why is this not higher priority? I think being able to run tests on a Mac is a pretty big deal

jin:
It's been a while since this was being worked on (I no longer work on Android rules). AFAIK, we were waiting for the multiplatform Nitrogen test runner (as I commented above), instead of making the legacy test infrastructure work on more platforms.

arturdryomov:
Ah, Nitrogen. Announced two years ago, not a single byte of it published. But there is always hope!

PR #15092, opened 2022-03-21
Fix android emulator darwin_arm64 select

keith:
This fixes this condition printing a strange error if you try to run
these on macOS. In general this still doesn't work because of
https://github.com/bazelbuild/bazel/issues/7069 but if we ever want to
support it this is a pre-requisite to supporting arm macs

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_sdk_repository_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_sdk_repository_template.txt
@@ -69,7 +69,8 @@ filegroup(
 filegroup(
     name = "qemu2_x86",
     srcs = ["emulator/emulator"] + select({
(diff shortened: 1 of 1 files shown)


ted-xie:
Hi Keith,

Thank you for this PR! At a glance it looks pretty low-risk. It looks like this is related to compiling Android projects on M1 Macs. Do you know how we could validate this fix on our own M1 machines, and have you done the same test on your own machine? What bug did this PR solve? In general, some more documentation around the select would be helpful context for maintainers.

ted-xie:
Thanks for the background. Overall looks good to me; I'll start importing this and let you know if there are any show stoppers.

ted-xie (approved):

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15093, opened 2022-03-21
Have README.md in java_tools release reference the right target

nresare:
I was trying to figure out what produced the files included in https://github.com/bazelbuild/java_tools/releases/download/java_v11.6/java_tools-v11.6.zip that happens to have some instructions on how to re-produce the contents of the file. 

However, it turns out that the instructions in README.md references the wrong build target, which made it a bit confusing. This change updates the target to one that produces a zip that seem to contain the right files.

### DIFF
--- a/src/upload_java_tools.sh
+++ b/src/upload_java_tools.sh
@@ -75,7 +75,6 @@ while [[ -n "$@" ]]; do
 done
 
 java_tools_zip=$(rlocation io_bazel/${java_tools_zip_name})
-platform=${platform:+"_"}${platform:-}
 
 # Create a temp directory and a writable temp zip file to add a README.md file to
 # the initial zip.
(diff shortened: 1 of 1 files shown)


src/upload_java_tools.sh
```diff
@@ -97,7 +97,7 @@ To build from source the same zip run the commands:
 
 $ git clone https://github.com/bazelbuild/bazel.git
 $ git checkout ${commit_hash}
-$ bazel build //src:java_tools_prebuilt.zip
+$ bazel build //src:java_tools.zip
```
comius, line 100: I think the same script is used for both prebuilt and non-prebuilt zips. This change would fix one but break 3 other platform. Please use `${platform}` variable here.

comius (changes requested):

comius:
Friendly ping.

comius:
Second ping.

nresare:
Hi. I did spend some time on this, but unfortunately it seems with the latest macOS update bazel no longer compiles so I couldn't test it easily. I have now pushed an (untested) update to the script

nresare:
bumping the grpc dependency to it's latest released version fixed the compilation problem and I could test drive this. Now adding target platform as well when it is applicable.

In all cases where `java_upload_tools.sh` is used currently, it holds true that the output of `basename $java_tools_zip` is the same name as the target that is being packaged. It is a little magical, and might break in the future, but at least it is a lot better now than it was.

comius (approved):

PR #15096, opened 2022-03-22
Collect coverage from cc_binary data deps of java_test

fmeum:
Before this commit, if a java_test executed a cc_binary from its data
deps, coverage for this cc_binary would not be collected. This is fixed
by adding the implicit `$collect_cc_coverage` attribute to
BazelJavaTestRule, similar to how this is already done for
BazelShTestRule.

Fixes the Java part of #15098.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BazelJavaTestRule.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/BazelJavaTestRule.java
@@ -62,6 +62,13 @@ public final class BazelJavaTestRule implements RuleDefinition {
             attr(":lcov_merger", LABEL)
                 .cfg(ExecutionTransitionFactory.create())
                 .value(BaseRuleClasses.getCoverageOutputGeneratorLabel()))
+        // Add the script as an attribute in order for java_test to output code coverage results for
+        // code covered by CC binaries invocations.
+        .add(
+            attr("$collect_cc_coverage", LABEL)
+                .cfg(ExecutionTransitionFactory.create())
+                .singleArtifact()
+                .value(env.getToolsLabel("//tools/test:collect_cc_coverage")))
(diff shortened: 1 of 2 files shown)


fmeum:
@c-mita Did you intentionally disable the integration test in https://github.com/bazelbuild/bazel/commit/065e2e8e76c5bf21bd797f3c8baebe909e85a6cf#diff-6dd28028af8d621974915318831c0addec4e9fd91386698481c63e65e09d9d93? It does seem to pass both with and without this PR and the commit message does not mention the new tags.

c-mita:
Yes, they were intentionally disabled, I guess I forgot to mention it in the commit.

Those tests would have failed until java_tools was updated (the Java coverage stuff is part of it). Notice that the "head" tests (which build a fresh version of the java tooling) were not disabled.

Looks like we simply forgot to re-enable the tests afterwards.

Created https://github.com/bazelbuild/bazel/pull/15100 to fix.

c-mita:
Reviewing the rest of your PR will take me a little longer; I need to convince myself that nothing strange will happen.

fmeum:
@c-mita Thanks! I'm currently looking into Java/C++ coverage in detail in order to get multi-language coverage reports working for [rules_jni](https://github.com/fmeum/rules_jni). This PR is part of fixing the issues I encountered, with https://github.com/bazelbuild/bazel/pull/15081 being another one.

c-mita (approved):
Part of me is surprised that this just "works". But I don't think there's an issue here; `java_test` already depends on the C++ toolchain.

I would like a "principled" way of handling coverage for multiple languages in a test rule, but this doesn't add anything in the way of complexity that would make me oppose it at this time.

fmeum:
@c-mita Let me know if you want to discuss possible principled ways to collect multi-language coverage in tests. I am aware of a design doc on this topic, but don't know what became of it.

fmeum:
@bazel-io flag

Issue #14437, opened 2021-12-15
Bazel crashes when using bzlmod and a registry with an incorrect `bazel_registry.json` file 


danielmachlab:
### Description of the problem / feature request:

Bazel crashes due to an internal error when using a bzlmod registry that doesn't correctly specify a `bazel_registry.json` file. The stack-trace printed by Bazel when it crashes is included in the last section below.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

To demonstrate this bug, I created [a fork of the bazel-central-repository] and made a [single commit] that deletes the contents of the `bazel_registry.json` file. I also created an [bazelmod_example repository] that has a dependency on `zlib`.  

To reproduce the bug:
1.  Install `bazelisk`
2. `git clone https://github.com/danielmachlab/bzlmod_example`
3. `bazelisk build src:hello-world --experimental_enable_bzlmod` to demonstrate the example works when using the bazel-central-repository
4. `bazelisk build src:hello-world --experimental_enable_bzlmod --registry https://danielmachlab.github.io/bazel-central-registry/` to reproduce the bug by using the forked version of the bazel-central-repository, which contained an empty `bazel_registry.json` file

[a fork of the bazel-central-repository]: https://github.com/danielmachlab/bazel-central-registry
[bazelmod_example repository]:https://github.com/danielmachlab/bzlmod_example
[single commit]: https://github.com/danielmachlab/bazel-central-registry/commit/5561e9718d1c89aa6f33a45a28e9f54ad28fbea7

### What operating system are you running Bazel on?

MacOS Big Sur 11.5.2 

### What's the output of `bazel info release`?
```
2021/12/15 15:52:43 Using unreleased version at commit 4c6e324dfab9444fc51b9fb3f5074a01889725e1
Starting local Bazel server and connecting to it...
development version
```

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

I am using `bazelisk` with a `.bazeliskrc` file containing:
```
USE_BAZEL_VERSION=last_green
```

###  Have you found anything relevant by searching the web?

I haven't found anything specific to this issue, but I have been referencing the following resources:
[bzlmod user guide](https://docs.bazel.build/versions/main/bzlmod.html)
[Bazel External Dependencies Overhaul Google Doc](https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit#)

[...]

PR #15109, opened 2022-03-24
fix(bzlmod): throw on json parse exception

Tomsons:
[Fixes](https://github.com/bazelbuild/bazel/issues/14437)

Get rid of new lines in `bazel_registry.json` and return an `Optional.empty()` aka registry descriptor without mirrors and throw an IOException for malformed JSON to stick to official json.org specs

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/IndexRegistry.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/IndexRegistry.java
@@ -117,7 +117,7 @@ public class IndexRegistry implements Registry {
       return Optional.empty();
     }
     String jsonString = new String(bytes.get(), UTF_8);
-    if (jsonString.strip().equals("")) {
+    if (jsonString.strip().isEmpty()) {
       return Optional.empty();
     }
     try {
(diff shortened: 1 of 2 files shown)


meteorcloudy (approved):
Thank you!

src/main/java/com/google/devtools/build/lib/bazel/bzlmod/IndexRegistry.java
```diff
@@ -115,7 +117,14 @@ private String constructUrl(String base, String... segments) {
       return Optional.empty();
     }
     String jsonString = new String(bytes.get(), UTF_8);
-    return Optional.of(gson.fromJson(jsonString, klass));
+    if (jsonString.strip().equals("")) {
```
Wyverald, line 120: why is the empty string treated specially?

also prefer using `String#isEmpty` over `equals("")`

Tomsons: Sure, will come with another commit.

The empty string is treated specially because  the `bazel_registry.json` can be empty without being explicitly declared as an empty json object `{}`, this should not throw an exception as it is simply ignored at `getRepoSpec`

```
    if (bazelRegistryJson.isPresent() && bazelRegistryJson.get().mirrors != null) {
      for (String mirror : bazelRegistryJson.get().mirrors) {
        try {
          new URL(mirror);
        } catch (MalformedURLException e) {
          throw new IOException("Malformed mirror URL", e);
        }

        urls.add(constructUrl(mirror, sourceUrl.getAuthority(), sourceUrl.getFile()));
      }
    }
```

Wyverald: I'd still rather the file always be valid JSON, but this is such a minor point that I won't push it. Thanks for the contribution!

Tomsons:
You're welcome @Wyverald @meteorcloudy

brentleyjones:
@bazel-io fork 5.2

Wyverald:
@bazel-io fork 5.2

Issue #14043, opened 2021-09-27
Bazel query error with square braces


rodentskie:
### Description of the problem / feature request:

Currently I'm working with a monorepo and has a NextJS project inside it, and it uses square braces 
for its dynamic url parameters. Like `web/pages/profile/[id]/index.tsx`. The problem is that bazel query returns
an error because of the square braces.

Here is the result:
```
❯ bazel query web/pages/profile/[id]/index.tsx
ERROR: Error while parsing 'web/pages/profile/[id]/index.tsx': unexpected token '[id' after query expression 'web/pages/profile/'
Loading: 0 packages loaded
```
### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

A dummy NextJS project will do, with a directory containing a `square braces` name on it.

### What operating system are you running Bazel on?

> Ubuntu 18.04 and in Github Action

### What's the output of `bazel info release`?
```
❯ bazel info release
INFO: Invocation ID: b87be1ab-a297-4aae-acf7-d61dba5b13f8
release 4.2.1
```
### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

> Replace this line with your answer.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

```
It is a private repo btw, I'm just a newbie devops here.
```

oquenchil:
So that character should be allowed in a package name but it might require you to quote it, see [here](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/cmdline/LabelValidator.java#L33).

Try quoting the file path and perhaps you have to add a backslash before each square bracket.

rodentskie:
> So that character should be allowed in a package name but it might require you to quote it, see [here](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/cmdline/LabelValidator.java#L33).
> 
> Try quoting the file path and perhaps you have to add a backslash before each square bracket.

You mean like this? 
`bazel query "web/pages/profile/\[id\]/index.tsx"`

I also tried this one:
`bazel query "web/pages/profile/[id]/index.tsx"`

Here are the results:

```
❯ bazel query "web/pages/profile/\[id\]/index.tsx"
[...]

PR #15115, opened 2022-03-24
Fix square braces in lexer

Tomsons:
[Fixes 14043](https://github.com/bazelbuild/bazel/issues/14043)


meisterT:
Please also update https://github.com/bazelbuild/bazel/blob/master/site/en/reference/query.md

meisterT (approved):

PR #15118, opened 2022-03-24
Collect C++ lcov coverage if runtime object not in runfiles

fmeum:
Before this commit, collecting C++ coverage in lcov format would fail
at the llvm-cov export step if a shared library listed in the
runtime_objects_list.txt was not contained in the runfiles of the top-
level target. This can happen e.g. if a cc_library depends on a
java_binary that has a cc_binary shared library in its resources.

This is fixed by not including objects that don't exist at runtime
in the llvm-cov invocation.

### DIFF
--- a/src/test/shell/bazel/bazel_coverage_cc_test_llvm.sh
+++ b/src/test/shell/bazel/bazel_coverage_cc_test_llvm.sh
@@ -146,7 +146,77 @@ LH:5
 LF:7
 end_of_record"
 
-  assert_equals "$(cat $(get_coverage_file_path_from_test_log))" "$expected_result"
+  assert_equals "$expected_result" "$(cat $(get_coverage_file_path_from_test_log))"
 }
+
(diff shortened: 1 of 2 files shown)


fmeum:
@c-mita I suspect this is also something for you to take a look at. The fix is the most direct one I could think of, but I am completely open to other suggestions (e.g. filtering the runtime objects list in the first place).

fmeum:
@c-mita Do you think that it will be possible to get this PR reviewed in time for the 5.2.0 release? It's the final blocker for cross-language coverage reports for us and now that `java_test` has the `$collect_cc_coverage` attribute, our current workaround of manually setting `CC_CODE_COVERAGE_SCRIPT` to point to a patched version of this script doesn't work anymore.

c-mita (approved):
LGTM.

Sorry for the delay.

fmeum:
@bazel-io flag

Issue #15022, opened 2022-03-10
Support `key` callback parameter for `min()`, `max()` starlark builtin functions


tetromino:
See https://github.com/bazelbuild/starlark/blob/master/spec.md#max

tetromino:
I light of #15022 - let me clarify how I see the feature needs to be implemented. Naturally, we do not want to call `sorted` (finding a min or max is an O(n) time and O(1) space problem; creating a temporary a sorted copy of the list is O(n log n) time and O(n) space). But we also do not want trivial comparator-based implementations that might evaluate `key` twice on each element of the list - the `key` function might be heavy, it might hypothetically have side effects, we want to call it only once on each element.

Thus ideally, I'd like to see an efficient, lightweight sort-key helper that would be shared between the implementations of `sorted`, `min` and `max`.

fmeum:
@bazel-io fork 7.2.0

iancha1992:
A fix for this issue has been included in [Bazel 7.2.0 RC1](https://github.com/bazelbuild/bazel/releases/tag/7.2.0rc1). Please test out the release candidate and report any issues as soon as possible.
If you're using Bazelisk, you can point to the latest RC by setting USE_BAZEL_VERSION=7.2.0rc1. Thanks!

PR #15122, opened 2022-03-25
Starlark mix max key callback param

Tomsons:
Handles request [#15022](https://github.com/bazelbuild/bazel/issues/15022)

Used `sorted` builtin mechanism to provide to key callback parameter

### DIFF
--- a/src/main/java/net/starlark/java/eval/BUILD
+++ b/src/main/java/net/starlark/java/eval/BUILD
@@ -35,6 +35,7 @@ java_library(
         "Printer.java",
         "RangeList.java",
         "Sequence.java",
+        "SortKeyHelper.java",
         "Starlark.java",
         "StarlarkCallable.java",
         "StarlarkFloat.java",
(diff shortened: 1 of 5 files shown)


tetromino:
You are proposing an O(n log n) time and O(n) memory solution for an O(n) time and O(1) memory problem.

Tomsons:
Ok, great thank you for the feedback. But why do you close instead of requesting changes and iterate over it until the actual implementation fulfills your request?

tetromino:
I apologize for misjudging the situation; I will reopen the PR.

tetromino (changes requested):
Please don't call `sorted` in min/max: finding a maximum or minimum requires O(n) time, O(1) temp memory; `sorted` is O(n log n) time, O(n) temp memory.

Note that `key` may be heavy and could have side effects, so it should be called at most once for each element - please add a test verifying that fact.

Tomsons:
No worries, I'll push something new in the coming days

Tomsons:
@tetromino, I've just pushed something new with a `SortKeyHelper.java` class, can you check if this kind of implementation could suit. I'll do some extensive testing later on

PR #15125, opened 2022-03-25
Shade com.google.thirdparty in JacocoCoverage.jarjar

### LINKED ISSUE #15124: JacocoCoverage_jarjar_deploy.jar's shading of Guava is incomplete, causing some user tests to fail with bazel coverage

While upgrading our repo to Bazel 5.1 from 4.2.2, we found some Java tests that depend on Guava were now failing but only with `bazel coverage`, not with `bazel test`. Test failures gave this stacktrace:

```
java.lang.NoSuchFieldError: EXACT
	at com.google.common.net.InternetDomainName.findSuffixOfType(InternetDomainName.java:170)
	at com.google.common.net.InternetDomainName.<init>(InternetDomainName.java:151)
	at com.google.common.net.InternetDomainName.from(InternetDomainName.java:209)
        at <ourcode>
```

When stepping through in the debugger, I noticed that `com.google.thirdparty.publicsuffix.PublicSuffixPatterns.EXACT` was being loaded from JacocoCoverage_jarjar_deploy.jar, not the Guava jar being pulled in by the code under test. It seems JacocoCoverage has [shaded com.google.common using JarJar](https://github.com/bazelbuild/bazel/blob/6a219b5aa4a44682f25fdd7bdf7108aed7b0b816/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverage.jarjar), but not com.google.thirdparty, which is also part of Guava.

Run `bazel coverage` on a JUnit test that invokes `com.google.common.net.InternetDomainName.from("example.com")`

macOS 12.3

release 5.1.0

This looks easy to fix, JacocoCoverage.jarjar just needs a rule to shade com.google.thirdparty.

thirtyseven:
Fixes #15124

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverage.jarjar
+++ b/src/java_tools/junitrunner/java/com/google/testing/coverage/JacocoCoverage.jarjar
@@ -1,3 +1,4 @@
 rule com.google.common.** com.google.testing.coverage.jarjar.@0
+rule com.google.thirdparty.** com.google.testing.coverage.jarjar.@0
 rule org.apache.commons.** com.google.testing.coverage.jarjar.@0
 rule org.objectweb.asm.** com.google.testing.coverage.jarjar.@0


thirtyseven:
I have a failing test case for this too that I haven't pushed yet, but I realize there will have to be a new release of java_tools before it can pass. LMK if I should include it in this PR.

comius (approved):

comius:
cc @c-mita, @cushon

PR #15129, opened 2022-03-28
Ensure command line reference genrule respects TMPDIR

jfgrimm:
Avoids build failures when `/tmp` is not writeable


sgowroji (approved):
Thank you!

Issue #13905, opened 2021-08-26
tools/http: Allow http_archive to extract deb archives


uhthomas:
### Description of the feature request:

Allow the `http_archive` rule to extract deb archives.

### Feature requests: what underlying problem are you trying to solve with this feature?

Debian archives are standard and widely used. There are many scenarios which require the extraction of deb archives during the loading phase. It would be great to have this natively in `http_archive`.

### What operating system are you running Bazel on?

`linux/amd64`

### What's the output of `bazel info release`?

`release 4.0.0`

###  Have you found anything relevant by searching the web?

There isn't a whole lot of great information. I wrote a simple `http_deb` repository rule which essentially mirrors the behaviour of `http_archive` but is able to extract deb archives.

```py3
ctx.execute(["dpkg-deb", "-x", path, "."])
```

Issue #15130, opened 2022-03-28
Add support for .ar and .deb files to repository_ctx


kjlubick:
### Description of the problem / feature request:

I intend to add support for .ar archive files and .deb files (which are the same file format).

### Feature requests: what underlying problem are you trying to solve with this feature?

[download_and_extract](https://docs.bazel.build/versions/main/skylark/lib/repository_ctx.html#download_and_extract) supports many compressed or packaged file formats already. Adding .ar and .deb files will make it easier for toolchains to download and extract .deb files into their sysroot without requiring an external tool.

[Example hermetic toolchain](https://github.com/google/skia/blob/7527f901d1c551ec36e9b09af12707595870f163/toolchain/build_toolchain.bzl#L111) that downloads .deb files and extracts them.

###  Have you found anything relevant by searching the web?

Apache Commons Compress [already has support for .ar files](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/ar/package-summary.html) and this package is already used by Bazel for [Compressed Tar files](https://github.com/bazelbuild/bazel/blob/45b225d5bdc5a1e2f1414d4a5a27ddf267dc4e15/src/main/java/com/google/devtools/build/lib/bazel/repository/CompressedTarFunction.java#L33) so we shouldn't need any additional third party dependencies.

aiuto:
I would like to find a way to do this by making it an external set of repository rules that can have it's own development team. The issue is that most of the core Bazel team never see .deb files. They won't be able to adequately maintain or enhance it.  I;m tied up this week and meteorcloudy@ is out on vacation. We can talk next week.

aiuto:
The more I think about it, the more I believe this belongs to a new rule set. That would follow the pattern of jvm_external, where we built new repository rules outside of bazel itself, so they can advance at different rates. Looking at your patch, you will decompress the .deb file, but a fully fleshed out rule could have so much more.  For example, installing a package and then making the libraries and headers available as a .so.

Since I have some  .deb file experience (on the building them side from rules_pkg), I can offer to help on parts of this, but what you need is a place to host it.  Could you create a repository under Skia's control? Then we can point to that as the recommended place for support of .deb install from the workspace.

meteorcloudy:
[...]

PR #15132, opened 2022-03-28
Add support for .ar archives (and .deb files)

kjlubick:
This implements #[15130](https://github.com/bazelbuild/bazel/issues/15130).

As I was updating the docs for .ar and .deb formats, I also addressed some previous formats that had been added but not propagated through to all the documentation places.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/ArFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/ArFunction.java
@@ -18,7 +18,11 @@ import com.google.common.io.ByteStreams;
 import com.google.devtools.build.lib.bazel.repository.DecompressorValue.Decompressor;
 import com.google.devtools.build.lib.vfs.Path;
 
-import java.io.*;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.Date;
 import org.apache.commons.compress.archivers.ar.ArArchiveEntry;
 import org.apache.commons.compress.archivers.ar.ArArchiveInputStream;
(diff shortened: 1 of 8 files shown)


aiuto:
I would like to find a way do to this without adding more code in Bazel. Perhaps we can do this mostly in starlark defined workspace rules.  I'm very familiar with the .deb and .ar formats on the writing side, so I'll take a look at this, but it might not be until next week.

meteorcloudy:
>  Perhaps we can do this mostly in starlark defined workspace rules.

How do you plan to do that? I think this PR extends the functionality of `repository_ctx.download_and_extract`, which is implemented in native code. Not sure how it could be implemented in pure starlark.

kjlubick:
I agree with @meteorcloudy, to implement this in pure starlark would involve decoding .ar files in Starlark code. This is not impossible, but certainly difficult given that we wouldn't really be able to use libraries. Another option is to shell out to a native executable, but that requires users to install or download such a binary.

src/main/java/com/google/devtools/build/lib/bazel/repository/ArFunction.java
```diff
@@ -0,0 +1,79 @@
...
+                    continue;
+                } else {
+                    // We do not have to worry about symlinks in .ar files - it's not supported
+                    // by the .ar file format.
+                    try (OutputStream out = filePath.getOutputStream()) {
+                        ByteStreams.copy(arStream, out);
+                    }
+                    filePath.chmod(entry.getMode());
+
+                    Date lastModified = entry.getLastModifiedDate();
```
meteorcloudy, line 68: Getting this error message from internal presubmit check:
"Date has a bad API that leads to bugs; prefer java.time.Instant or LocalDate."

src/main/java/com/google/devtools/build/lib/bazel/repository/ArFunction.java
```diff
@@ -0,0 +1,79 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.bazel.repository;
+
+import com.google.common.io.ByteStreams;
+import com.google.devtools.build.lib.bazel.repository.DecompressorValue.Decompressor;
+import com.google.devtools.build.lib.vfs.Path;
+
+import java.io.*;
```
meteorcloudy, line 21: Wildcard imports, static or otherwise, should not be used

meteorcloudy (changes requested):

ckolli5:
@bazel-io fork 5.2.0

ckolli5:
@bazel-io fork 5.2.0

uhthomas:
Fixes https://github.com/bazelbuild/bazel/issues/13905?

PR #15151, opened 2022-03-31
Remote: Don't check declared outputs for failed action

coeuvre:
Fix a bug that Bazel print message

```
Invalid action cache entry ...: expected output ... does not exist.
```

instead of the underlying error message when remote action failed.


brentleyjones:
@bazel-io fork 5.2

Wyverald:
@bazel-io fork 5.1.1

kjlubick:
I'm still seeing this behavior in 5.2.0

coeuvre:
I see. This check was implemented to intentionally avoid that issue, see https://github.com/bazelbuild/bazel/issues/14543.

Issue #6354, opened 2018-10-10
Use @argument files in the Java binary wrapper script


cushon:
The Java binary wrapper script works around command line length limits by creating a jar file with a `Class-Path` manifest entry containing the long classpath (related: #6289).

JDK 9 adds support for flag files using the `@argument-file` syntax, see "java Command-Line Argument Files" in https://docs.oracle.com/javase/9/tools/java.htm.

We should use that instead of the manifest jar hack when targeting JDK 9.

cushon:
(previous bug about `@argfile`s: #1392)

cushon:
@petroseskinder I filed this as a reminder to look at `@argfile`s if/when it's worth adding features that only work for JDK > 9. I think we need a different short-term fix for #6289.

petroseskinder:
I see. Thank you. Does the suggestion by @brown suffice for the short term (pre-JDK >9) fix?

cushon:
I actually didn't realize there was a patch there because github mangled the formatting :/
I left a comment: https://github.com/bazelbuild/bazel/issues/6289#issuecomment-430730371

ittaiz:
Should this be fixed? Like @cushon I have a hunch (no data unfortunately) that using an argument file will perform better than a classpath jar

cushon:
I think the main challenge is that we can only rely on this feature for JDK 9 and newer.

There was some related discussion prompted by https://github.com/bazelbuild/bazel/pull/15487/files

One option is to add some information to `java_runtime` that tells Bazel if it can rely on argument files, e.g.

* `java_runtime(..., supports_params_files = True)`
* `java_runtime(..., feature_version = 9)`
* `java_runtime(..., features = ["params_file"], ...)`

ittaiz:
I see. Thanks.
For better and for worse rules_scala has its own fork of java_stub_template so it can do this for itself.

ittaiz:
One last thing (and really sorry for spamming) but from a few runs this is shaving for us 1s-1.5s internally and we have thousands of not more of test targets with large class paths (partially because we have fine grain build targets)

ittaiz:
Thanks!
We'll roll out the change today or on Sunday probably but this is a divergence I obviously would like to remove.
Btw, the measurement is on my M1. We still have many people on intel chips and my colleagues tell me that there this preparation takes ~4s on many targets.
Really interested in the full rollout 😁🤞🏽

ittaiz:
[...]

PR #15159, opened 2022-04-01
Default to a UTF-8 locale in Java stub template

### LINKED ISSUE #15106: Non-ascii filenames  cause `InvalidPathException` in `java_test`
It seems that the environment Bazel executes the JVM in on Linux does not allow for non-ascii filenames.

Example:

Test.java:
```
import java.nio.file.Files;
import java.io.IOException;

class Test {
    public static void main(String[] args) throws IOException {
        Files.createTempFile("æøå", null);
    }
}
```

BUILD:
```
java_test(
    name = "test",
    srcs = ["Test.java"],
    main_class = "Test",
    test_class = "Test",
)
```

Executing this (on Linux):

```
bazel run --java_runtime_version=remotejdk_11 //:test
```

Fails with this:
```
Exception in thread "main" java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: ???5404569911566851009.tmp
	at java.base/sun.nio.fs.UnixPath.encode(UnixPath.java:145)
	at java.base/sun.nio.fs.UnixPath.<init>(UnixPath.java:69)
	at java.base/sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:279)
	at java.base/java.nio.file.TempFileHelper.generatePath(TempFileHelper.java:59)
	at java.base/java.nio.file.TempFileHelper.create(TempFileHelper.java:126)
	at java.base/java.nio.file.TempFileHelper.createTempFile(TempFileHelper.java:160)
	at java.base/java.nio.file.Files.createTempFile(Files.java:913)
	at Test.main(Test.java:6)
```

This issue seems to only happen on Linux (but I've only tried Ubuntu 20.04.3 LTS). On Mac, it works without issues. I have not tried on Windows.

Bazel version:
```
$ bazel info release
release 5.0.0
```

I have found some issues of people getting `InvalidPathException`, but they seem to be exceptions in Bazel itself. This is an exception from the application being tested instead.

I have also tried setting `LANG` via `--test_env` and `--action_env`, and setting the `sun.jnu.encoding` and `file.encoding` JVM parameters, but to no avail.

fmeum:
On non-macOS Unix, without any locale variable set, the OpenJDK defaults
to using ASCII rather than UTF-8 as the encoding for file system paths
(i.e., the value of the `sun.jnu.encoding` property).

Fixes #15106

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
@@ -94,6 +94,16 @@ function is_macos() {
   [[ "${OSTYPE}" =~ darwin* ]]
 }
 
+function available_utf8_locale() {
+  # Both C.UTF-8 and en_US.UTF-8 do not cause any language-specific effects
+  # when set as LC_CTYPE, but neither is certain to exist on all systems.
+  if [[ $(LC_CTYPE=C.UTF-8 locale charmap 2>/dev/null) == "UTF-8" ]]; then
+    echo "C.UTF-8"
+  elif [[ $(LC_CTYPE=en_US.UTF-8 locale charmap 2>/dev/null) == "UTF-8" ]]; then
+    echo "en_US.UTF-8"
+  fi
+}
+
 # Parse arguments sequentially until the first unrecognized arg is encountered.
 # Scan the remaining args for --wrapper_script_flag=X options and process them.
 ARGS=()
@@ -360,6 +370,17 @@ if [ -z "$CLASSPATH_LIMIT" ]; then
   # Windows per-arg limit MAX_ARG_STRLEN == 8k
   # Linux per-arg limit MAX_ARG_STRLEN == 128k
   is_windows && CLASSPATH_LIMIT=7000 || CLASSPATH_LIMIT=120000
+fi
+
+# On non-macOS Unix, without any locale variable set, the JVM would use
+# using ASCII rather than UTF-8 as the encoding for file system paths.
+if ! is_macos; then
+  if [ -z ${LC_CTYPE+x} ] && [ -z ${LC_ALL+x} ] && [ -z ${LANG+x} ]; then
+    UTF8_LOCALE=$(available_utf8_locale)
+    if [[ -n "$UTF8_LOCALE" ]]; then
+      export LC_CTYPE="$UTF8_LOCALE"
+    fi
+  fi
 fi
 
 if (("${#CLASSPATH}" > ${CLASSPATH_LIMIT})); then
--- a/src/test/shell/bazel/unicode_filenames_test.sh
+++ b/src/test/shell/bazel/unicode_filenames_test.sh
@@ -206,4 +206,35 @@ function test_utf8_source_artifact_in_bep() {
   expect_log '"name":"pkg/srcs/ünïcödë fïlë.txt"'
 }
 
+function test_utf8_filename_in_java_test() {
+  # Intentionally do not check for available locales: Either C.UTF_8 or
+  # en_US.UTF-8 should exist on all CI machines - if not, we want to learn about
+  # this so that the Java stub template can be adapted accordingly.
+
+  touch WORKSPACE
+  mkdir pkg
+
+  cat >pkg/BUILD <<'EOF'
+java_test(
+    name = "Test",
+    srcs = ["Test.java"],
+    main_class = "Test",
+    use_testrunner = False,
+)
+EOF
+
+  cat >pkg/Test.java <<'EOF'
+import java.nio.file.Files;
+import java.io.IOException;
+
+class Test {
+    public static void main(String[] args) throws IOException {
+        Files.createTempFile("æøå", null);
+    }
+}
+EOF
+
+  bazel test //pkg:Test --test_output=errors 2>$TEST_log || fail "Test should pass"
+}
+
 run_suite "Tests for handling of Unicode filenames"


fmeum:
The test failures indicate that there are all sorts of compatibility issues with this approach:
1. Not all systems (CentOS) have a UTF-8 locale installed by default.
2. Setting a UTF-8 locale messes with tools output as reported by Bazel (see https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/48d00f3b-d64f-4940-b45d-bfc8d8df4928/src/test/shell/integration/run_test/shard_1_of_3/test.log).

fmeum:
Given the large number of test failures and the fact that the test encyclopedia explicitly documents that all locale variables are unset, I gave up on the original approach of setting `LC_ALL` for all actions and am now only setting it in `java_stub_template.txt` and only if none of the locale variables is set explicitly.

comius:
There is some prior art in JavaCompileAction and JavaHeaderCompileAction: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/rules/java/JavaCompileActionBuilder.java;l=208;drc=cb01bde9369f8197cb54c34b14aa41f0dcc903cf

If something along this way was possible for java_test rule, I would prefer this instead of making the stub template more complex. Testing such change internally would probably be more involved, but I don't mind doing it if it's the right thing.

It seems to me the right thing to do would be to have it on all test rules, because I don't see Java anything special in the bug report (test in other languages could use UTF-8 filenames - and this is also another reason why not to put it into Java stub), however this would be more involved and would probably require migration by parts.

> 1. Not all systems seem have a UTF-8 locale installed by default in a way picked up by all tools
Does setting LANG instead of LC_TYPE help here?

> 2. Setting a UTF-8 locale messes with tools output as reported by Bazel
Is there a way to fix this? How does the output look like? Should just the test be fixed?

fmeum:
It seems that `C.UTF-8` is not universally supported, see https://unix.stackexchange.com/questions/597962/how-widespread-is-the-c-utf-8-locale. I will try again with `en-US.UTF-8`, which should be.

@comius Assuming this would require a migration anyway, would you prefer to set this on all actions or just on all test actions?

comius:
> @comius Assuming this would require a migration anyway, would you prefer to set this on all actions or just on all test actions?

All test actions is where such filenames may occur.

All actions is too wide. This might conflict some other requirements of specific actions, so let's avoid it.

fmeum:
@comius This seems pretty subtle based on the CI results: The Ubuntu 18.04 runner doesn't have the `en_US.UTF-8` locale, the CentOS runner doesn't have the `C.UTF-8` locale. That makes it difficult to determine a working locale statically as setting a non-existent locale can cause warnings to be printed. I'm not sure I see options other than determining a suitable locale at runtime in the Java stub, but of course that doesn't handle non-Java tests.

sgowroji:
Hello @fmeum, Thank you for sharing your contributions in PR. Can you please fix the above buildkite presubmit build failures ?

fmeum:
@comius Given the problem that neither `C.UTF-8` nor `en_US.UTF-8` can be assumed to exist, what kind of approach do you suggest? Would including conditional logic in the Java stub be fine knowing that a static choice is difficult?

cushon:
> The Ubuntu 18.04 runner doesn't have the en_US.UTF-8 locale, the CentOS runner doesn't have the C.UTF-8 locale. That makes it difficult to determine a working locale statically

Could we use platforms for this, to differentiate between those target environments and then statically pick the locale to use here?

Also, for my understanding, is this going to make the default behaviour the JDK will have after [JEP 400: UTF-8 by Default](http://openjdk.java.net/jeps/400)?

fmeum:
> Could we use platforms for this, to differentiate between those target environments and then statically pick the locale to use here?

That may be possible. It isn't going to be easy though if the available locales already differ between major distributions - should those be reflected in platform constraints? 

What I'm thinking: Maybe a prominent hint in the docs instructing users to set either `--action_env=LC_CTYPE=C.UTF-8` or `--action_env=LC_CTYPE=C.en_US.UTF-8` in their `.bazelrc` would serve the same purpose with less additional complexity if we can't find something that "just works".

> Also, for my understanding, is this going to make the default behaviour the JDK will have after [JEP 400: UTF-8 by Default](http://openjdk.java.net/jeps/400)?

I wasn't aware of this effort, but based on a quick glance I would say yes. It even mentions `sun.jnu.encoding`.

cushon:
> should those be reflected in platform constraints?

My sense is that it would be nice if platforms could easily express this, but that it probably isn't a straightforward approach. I don't have as crisp an answer to 'what belongs in platform constraints' as I'd like. Maybe someone else reading this does?

> Would including conditional logic in the Java stub be fine knowing that a static choice is difficult?

What does the change to do this in the stub template look like? All else being equal it'd be nice to keep more logic out of the stub template, but there's precedent for putting work-arounds in the stub template (e.g. `create_and_run_classpath_jar` / https://github.com/bazelbuild/bazel/issues/6354), and it doesn't seem like any of the options here are perfect, so that seems work considering to me.

fmeum:
> > Would including conditional logic in the Java stub be fine knowing that a static choice is difficult?
> 
> What does the change to do this in the stub template look like? All else being equal it'd be nice to keep more logic out of the stub template, but there's precedent for putting work-arounds in the stub template (e.g. `create_and_run_classpath_jar` / #6354), and it doesn't seem like any of the options here are perfect, so that seems work considering to me.

It would essentially amount to querying `locale` for `C.UTF-8`, `en_US.UTF-8`, and `POSIX` in order, setting the first one that is available as `LC_CTYPE` (setting `POSIX` explicitly is probably not necessary though, as it seems to be the default).

cushon:
> It would essentially amount to querying `locale` for `C.UTF-8`, `en_US.UTF-8`, and `POSIX` in order, setting the first one that is available as `LC_CTYPE` (setting `POSIX` explicitly is probably not necessary though, as it seems to be the default).

Thanks, that sounds reasonable to me.

fmeum:
@cushon I implemented this approach.

cushon:
This makes sense to me, although I also want @comius's input.

What do you think about also putting this change behind an `--incompatible_` flag? Internally we'd need to do more testing and deal with tests relying on the current behaviour before making this change, and I imagine other users will have the same issue, and a flag is the usual way to manage that.

fmeum:
I will wait for @comius input on the general approach, but agree that this should be moved behind a flag.

cushon (approved):

comius (approved):

comius:
This implementation looks like the most reasonable default behaviour.

> What do you think about also putting this change behind an `--incompatible_` flag? Internally we'd need to do more testing and deal with tests relying on the current behaviour before making this change, and I imagine other users will have the same issue, and a flag is the usual way to manage that.

I'd avoid additional flag if possible, because we already have many and I wouldn't expect a significant breakages to happen. In this case Bazel LTS policy doesn't require us to do the flag dance - introducing it for one release, flipping it and removing it.

I'd also assume the stub template is something, that is probably already overriden/replaced by large users (like Google does)

fmeum:
@bazel-io flag

PR #15166, opened 2022-04-03
Optionally enable LLVM profile continuous mode

fmeum:
LLVM's profile instrumentation offers a continuous mode in which
counters are continuously synced to a file rather than being dumped
once in an exit hook. This is useful for coverage runs that include
binaries that exit abnormally (e.g. in failure tests), but may require
additional compiler flags and can negatively impact runtime performance
and memory usage.

Enabling continuous mode requires adding the "%c" modifier to the value
of the LLVM_PROFILE_FILE environment variable. With this commit, the
collect_coverage.sh script adds the modifier if the test environment has
the variable LLVM_PROFILE_CONTINUOUS_MODE set. This allows both all and
individual tests to use continuous mode by setting the variable, either
via --test_env or the env attribute.

### DIFF
--- a/tools/test/collect_coverage.sh
+++ b/tools/test/collect_coverage.sh
@@ -93,6 +93,11 @@ if [[ -z "$GCOV_PREFIX_STRIP" ]]; then
 fi
 export GCOV_PREFIX="${COVERAGE_DIR}"
 export LLVM_PROFILE_FILE="${COVERAGE_DIR}/%h-%p-%m.profraw"
+if [[ -n "$LLVM_PROFILE_CONTINUOUS_MODE" ]]; then
+  # %c enables continuous mode but expands out to nothing, so the position
+  # within LLVM_PROFILE_FILE does not matter.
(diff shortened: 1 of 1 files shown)


fmeum:
@c-mita This is a minimal change required to make LLVM profile's continuous mode usable with Bazel at all (everything else can be configured via existing knobs). Since I need this mode to properly generate complete coverage reports for [Jazzer](https://github.com/CodeIntelligenceTesting/jazzer), I am very interested in getting at least this change into Bazel.

If you would appreciate a better integration, I could submit a follow-up PR that adds the required compiler flags as toolchain features together with an experimental flag, similar to the existing `--experimental_use_llvm_covmap`.

fmeum:
@c-mita Sorry for the high PR load, but I assume this one is also yours to review

fmeum:
@c-mita Friendly ping

fmeum:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

PR #15170, opened 2022-04-04
Specify fixedEnv/inheritedEnv interaction in ActionEnvironment

fmeum:
Previously, ActionEnvironment did not publicly document how fixed and
inherited environment variables interact, but still cautioned users to
keep the two sets disjoint without enforcing this. As a result, neither
could users rely on the interaction nor could ActionEnvironment benefit
from the additional freedom of not specifying the behavior.

With this commit, ActionEnvironment explicitly specifies that the values
of environment variable inherited from the client environment take
precedence over fixed values and codifies this behavior in a test.
This has been the effective behavior all along and has the advantage
that users can provide overrideable defaults for environment variables.

Also renames addVariables and addFixedVariables to better reflect that
they are operating on an immutable type.


fmeum:
@comius This is the first follow-up to https://github.com/bazelbuild/bazel/pull/14849.

comius (approved):

PR #15171, opened 2022-04-04
Intern trivial ActionEnvironment and EnvironmentVariables instances

fmeum:
When an ActionEnvironment is constructed out of an existing one, the
ActionEnvironment and EnvironmentVariables instances, which are
immutable, can be reused if no variables are added.

Also renames addVariables and addFixedVariables to better reflect that
they are operating on an immutable type.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ActionEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ActionEnvironment.java
@@ -81,10 +81,10 @@ public final class ActionEnvironment {
    */
   static class CompoundEnvironmentVariables implements EnvironmentVariables {
 
-    static EnvironmentVariables create(
-        Map<String, String> fixedVars, Set<String> inheritedVars, EnvironmentVariables base) {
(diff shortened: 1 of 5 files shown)


fmeum:
@comius This is the second follow-up to https://github.com/bazelbuild/bazel/pull/14849 and based on https://github.com/bazelbuild/bazel/pull/15170, which is why I labeled it as a draft.

fmeum:
@comius I rebased onto master and marked this as ready for review.

comius (approved):

fmeum:
@comius Is there a public way to quickly benchmark the memory impact of a Bazel change?

comius:
> @comius Is there a public way to quickly benchmark the rough memory impact of a Bazel change?

If you have a sizable project you can check `used-heap-size-after-gc` displayed by `bazel info`. See https://github.com/bazelbuild/bazel/issues/14890

I don't know about any public scripts that would do it for you.

cc @meisterT

meisterT:
You can also use https://github.com/bazelbuild/bazel-bench

PR #15173, opened 2022-04-04
Allow customizing bazel wrapper location checked into repository

kageiit:
Allow customizing location of Bazel wrapper checked into the repository using an env var $BAZEL_WRAPPER. There is no strong reason this needs to be hardcoded to `tools/bazel`. For eg: we keep most of our tools in the top-level `tooling` directory and would like to organize the wrapper location better and avoid confusion caused by two tool-related directories at the top `tools` and `tooling`.

### DIFF
--- a/scripts/packages/bazel.sh
+++ b/scripts/packages/bazel.sh
@@ -166,11 +166,12 @@ if [[ ! -x ${BAZEL_REAL} && -x ${bazel_real_path} ]]; then
   fi
 fi
 
-# If the repository contains a checked-in executable called tools/bazel, we
-# assume that they know what they're doing and have their own way of versioning
-# Bazel. Thus, we don't have to print our helpful messages or error out in case
-# we couldn't find a binary.
-readonly wrapper="${workspace_dir}/tools/bazel"
+# If the repository contains a checked-in executable set by $BAZEL_WRAPPER
+# (defaults to tools/bazel), we assume that they know what they're doing and
+# have their own way of versioning Bazel. Thus, we don't have to print our
+# helpful messages or error out in case we couldn't find a binary.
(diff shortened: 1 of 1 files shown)


brettwooldridge:
Hey @kageiit! Glad to see you over here on the Bazel project!

kageiit:
> 

Glad to see you here as well @brettwooldridge . Hoping to contribute more moving forward as we are moving towards bazel from buck

meteorcloudy:
Hi @kageiit since this would require ensuring that 'BAZEL_WRAPPER=tooling/bazel' is set somewhere, presumably in your developer's bashrc, I wonder if you considered an `alias bazel=tooling/bazel` (or equivalently a `function bazel() { ... }`), since that would do the same thing.

kageiit:
> Hi @kageiit since this would require ensuring that 'BAZEL_WRAPPER=tooling/bazel' is set somewhere, presumably in your developer's bashrc, I wonder if you considered an `alias bazel=tooling/bazel` (or equivalently a `function bazel() { ... }`), since that would do the same thing.

Hey @meteorcloudy we plan to set this in a top level bazelwrapper script which would be called by developers like `./bazelw build foo`. We also have other ways to ensure the var is set using tools like direnv. The main change here is to ensure we have a way to programmatically control the wrapper and have be provisioned automatically as part of repo toolng than have developers set it manually.

Regardless of that, I do believe there is no strong reason to hardcode this value (there are other values that are overridable via env vars like USE_BAZEL_VERSION in the current bazel wrapper script, this one is an exception so far) as we would still like to use tools like bazelisk through our wrappers and it makes things less maintainable there more we deviate from what is available in upstream.

meteorcloudy:
Thanks for the explanation!

meteorcloudy (approved):

Issue #14278, opened 2021-11-15
Bazel Remote Cache: Cannot use WIF-generated credentials


kylekurz:
### Description of the problem / feature request:

Bazel cannot use credentials from Workload Identity Federation, exits with error:

```
ERROR: Failed to init auth credentials: Error reading credentials from stream, 
'type' value 'external_account' not recognized. 
Expecting 'authorized_user' or 'service_account'.
```

### Feature requests: what underlying problem are you trying to solve with this feature?

Using remote cache without providing a permanent Service Account JSON credential in a secret or directly in my repository.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Set up a Google Cloud Storage bucket and configure WIF via these instructions: https://github.com/google-github-actions/auth. Once you've set it up, use the auth module to create a job that utilizes a credential file for remote cache access:

```
name: GitHub Actions CI - Build Artifacts
on: [pull_request]
jobs:
  pr-build:
    permissions:
      contents: 'read'
      id-token: 'write'
    runs-on: ubuntu-20.04
    steps:
      - name: Check out repository
        uses: actions/checkout@v2
      - id: 'auth'
        name: 'Authenticate to Google Cloud'
        uses: 'google-github-actions/auth@v0.4.0'
        with:
          token_format: 'access_token'
          workload_identity_provider:<identity_provider>
          service_account: <service_account_email>
          create_credentials_file: true
      - name: Verify that all dependencies are set up correctly
        run: |
          bazelisk run \
            --google_credentials=${{ steps.auth.outputs.credentials_file_path }} \
            --remote_cache=<cache_url> \
            //:gazelle
```

### What operating system are you running Bazel on?

Ubuntu 20.04 on a GitHub Hosted Runner

### What's the output of `bazel info release`?

`release 4.2.1`

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

No

### Any other information, logs, or outputs that you want to share?

N/A

bazaglia:
Facing the same issue. @kylekurz, could you find a workaround while the third-party libraries aren't updated?

kylekurz:
@bazaglia I've resigned myself to using a SA credential in a GitHub secret for now, with notes to remove it once this is fixed.

[...]

PR #15176, opened 2022-04-04
Upgrade Google Auth Version

ThomasCJY:
upgrade google auth to use the latest v1.6.0 instead of the outdated 0.17 in bazel


caseyduquettesc:
For some more context, in 1.5.3 built-in retries were added, which improves the reliability of fetching credentials from the metadata server - https://github.com/googleapis/google-auth-library-java/commit/f9a9b8ace0199e6b75ed42c7bacfa3be30c34111

bazaglia:
I believe this is also gonna fix https://github.com/bazelbuild/bazel/issues/14278

I hope this gets merged soon, the current library is outdated by 3 years and currently prevents me using remote cache with Workload Identity Federation.

third_party/BUILD
```diff
@@ -163,8 +163,10 @@ distrib_java_import(
     jars = [
         "api_client/google-api-client-1.22.0.jar",
         "api_client/google-api-client-jackson2-1.22.0.jar",
+        "api_client/google-api-client-gson-1.22.0.jar",
```
coeuvre, line 166: Does that mean the newer version uses `gson` instead of `jackson2`? It would be nice if we can remove `jackson2`.

third_party/BUILD
```diff
@@ -163,8 +163,10 @@ distrib_java_import(
...
         "api_client/google-http-client-1.22.0-SNAPSHOT.jar",
         "api_client/google-http-client-jackson2-1.22.0.jar",
+        "api_client/google-http-client-gson-1.22.0.jar",
     ],
     runtime_deps = [
         ":jackson2",
```
coeuvre, line 172: We probably need add `:gson`  to `runtime_deps` on here.

coeuvre, line 172: We probably need add runtime_deps on `:gson` here.

third_party/BUILD
ThomasCJY: yes looks like they switched to gson in 0.23.0 https://github.com/googleapis/google-auth-library-java/pull/531/files. Removed jackson dependency in pr

third_party/BUILD
ThomasCJY: added

coeuvre (commented):
Thanks!

I didn't realize removing `jackson2` will introduce non third_party changes. Unfortunately, we can't import PR that include both third_party and non third_party changes. In this case, we have to split the PR into 3 small PRs:
1. Add Gson and upgrade Auth (third_party only changes)
2. Replace `jackson2` usages in non third_party packages
3. Remove `jackson2`

Sorry the import process is not easy enough.

ThomasCJY:
@coeuvre https://github.com/bazelbuild/bazel/pull/15215 part 1 PR is here

coeuvre:
This PR has been submitted as 3 sub PRs. Closing.

brentleyjones:
@bazel-io flag

Issue #11961, opened 2020-08-17
[Android] Mobile-Install runtime crash caused by rewriting try-with-resources statements


ThomasCJY:
**Background**
* Test repository and branch to repro the issue ([link](https://github.com/ThomasCJY/bazel_example/pull/2))

While using `mobile-intall` to install our android app target, we noticed an instant crash on app open with a stack trace:
```
 E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.bazel, PID: 3197
    java.lang.IllegalAccessError: Illegal class access: 'com.google.devtools.build.android.desugar.runtime.ThrowableExtension' attempting to access 'com.google.devtools.build.android.desugar.runtime.ThrowableExtension$NullDesugaringStrategy' (declaration of 'com.google.devtools.build.android.desugar.runtime.ThrowableExtension' appears in /data/app/~~y9VAI4_h6duHtmrR-6cmMw==/com.example.bazel-F-lfB2AsjpB-6l3LfcfTGQ==/base.apk)
        at com.google.devtools.build.android.desugar.runtime.ThrowableExtension.<clinit>(ThrowableExtension.java:67)
        at com.google.devtools.build.android.desugar.runtime.ThrowableExtension.printStackTrace(ThrowableExtension.java:96)
        at com.google.common.base.Throwables.getStackTraceAsString(Throwables.java:352)
        at com.example.bazel.MainActivity.onCreate(MainActivity.java:27)
```
And the problematic line of code is:
```
import com.google.common.base.Throwables;
....
String s = Throwables.getStackTraceAsString(e);
```
After digging into the issue, I noticed that bazel is using [TryWithResourcesRewriter ](https://github.com/bazelbuild/bazel/blob/master/src/tools/android/java/com/google/devtools/build/android/desugar/TryWithResourcesRewriter.java#L97-L101) to replace `Throwables` with `ThrowableExtension`. Apparently, this change causes issue during runtime where dependency and access permissions has not been handled gracefully. 

**Repro Steps**
* Test repository and branch to repro the issue ([link](https://github.com/ThomasCJY/bazel_example/pull/2))
* Run `bazelisk mobile-install //src/main:app` and open the app, app will instantly crash

nkoroste:
Any updates on this?

arunkumar9t2:
[...]

PR #15184, opened 2022-04-05
Make Desugaring Classes Public

jlaxson:
When an android application is installed via mobile-install, some uses of Throwable (such as getting the stack trace) will cause java.lang.IllegalAccessError as described in #11961.

This applies a fix as described in that thread, making the classes public.


jlaxson:
@ahumesky @ted-xie @kevin1e100 polite bump on this

ted-xie:
No major concerns from my end based on this diff. I see there were a couple presubmit failures; I'll take a look at those.

ted-xie (approved):

PR #15199, opened 2022-04-08
Add a flag to expose undeclared test outputs in unzipped form.

### LINKED ISSUE #14568: option to not zip test results

Bazel should allow test results to be left as is as opposed to putting them into a zip file.

I would like to make it easier to browse test output which is an interlinked set of HTML files directly from Buildbarn Browser without having to download and extract the zip file and then open the index page in the browser.

Initially raised to buildbarn browser: https://github.com/buildbarn/bb-browser/issues/35 where they suggested this should be fixed in Bazel instead.

tjgq:
Closes #14568.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/test/TestConfiguration.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/test/TestConfiguration.java
@@ -286,6 +286,14 @@ public class TestConfiguration extends Fragment {
         help = "If true, then Bazel will run coverage postprocessing for test in a new spawn.")
     public boolean splitCoveragePostProcessing;
 
+    @Option(
+        name = "zip_undeclared_test_outputs",
+        defaultValue = "true",
+        documentationCategory = OptionDocumentationCategory.TESTING,
+        effectTags = {OptionEffectTag.TEST_RUNNER},
+        help = "If true, undeclared test outputs will be archived in a zip file.")
+    public boolean zipUndeclaredTestOutputs;
+
     @Override
     public FragmentOptions getHost() {
       TestOptions hostOptions = (TestOptions) getDefault();
@@ -347,11 +355,11 @@ public class TestConfiguration extends Fragment {
     return options.testShardingStrategy;
   }
 
-  public Label getCoverageSupport(){
+  public Label getCoverageSupport() {
     return options.coverageSupport;
   }
 
-  public Label getCoverageReportGenerator(){
+  public Label getCoverageReportGenerator() {
     return options.coverageReportGenerator;
   }
 
@@ -386,6 +394,10 @@ public class TestConfiguration extends Fragment {
 
   public boolean splitCoveragePostProcessing() {
     return options.splitCoveragePostProcessing;
+  }
+
+  public boolean getZipUndeclaredTestOutputs() {
+    return options.zipUndeclaredTestOutputs;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/devtools/build/lib/analysis/test/TestRunnerAction.java
```diff
@@ -391,11 +391,10 @@ public boolean showsOutputUnconditionally() {
         builder.add(
             Pair.of(TestFileNameConstants.TEST_WARNINGS, resolvedPaths.getTestWarningsPath()));
       }
-      if (resolvedPaths.getUndeclaredOutputsZipPath().exists()) {
+      if (resolvedPaths.getUndeclaredOutputsDir().exists()) {
         builder.add(
             Pair.of(
-                TestFileNameConstants.UNDECLARED_OUTPUTS_ZIP,
```
coeuvre, line 397: Not sure whether internal code relies on this name. Please keep an eye on the internal usages and tests when importing.

src/test/shell/bazel/bazel_test_test.sh
```diff
@@ -695,7 +695,7 @@ EOF
   assert_equals "fail" "$(awk "NR == $(wc -l < $TEST_log)" $TEST_log)"
 }
 
-function test_undeclared_outputs_are_zipped_and_manifest_exists() {
```
coeuvre, line 698: Since we had this issue when developing, I think we can also add a test for the case that changing the value of `--zip_undeclared_test_outputs` will rerun the test.

src/main/java/com/google/devtools/build/lib/analysis/test/TestConfiguration.java
```diff
@@ -286,6 +286,14 @@
         help = "If true, then Bazel will run coverage postprocessing for test in a new spawn.")
     public boolean splitCoveragePostProcessing;
 
+    @Option(
+        name = "zip_undeclared_test_outputs",
```
coeuvre, line 290: FYI, we usually start with `experimental_` for a new feature flag and set its default value to `false` if it is a boolean value. But in this case, calling `zip_undeclared_test_outputs` and defaulting to `true` make more sense.

coeuvre (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/test/TestConfiguration.java
tjgq: Acknowledged.

src/test/shell/bazel/bazel_test_test.sh
tjgq: Done.

src/test/shell/bazel/bazel_test_test.sh
```diff
@@ -714,36 +714,107 @@ sh_test(
...
+
+  bazel test -s //dir:test &> $TEST_log || fail "expected success"
+
+  [ -s $outputs_zip ] || fail "$outputs_zip was not present after test"
+
+  bazel test -s --nozip_undeclared_test_outputs //dir:test &> $TEST_log || fail "expected success"
+
+  [ -s $output_text ] || fail "$output_text was not present after test"
+  [ -s $output_html ] || fail "$output_html was not present after test"
+}
```
coeuvre, line 799: Will the zip file be deleted after the build? Also considering the "unzipped and zipped" case, will the text and html files be deleted after second build?

src/main/java/com/google/devtools/build/lib/analysis/test/TestRunnerAction.java
tjgq: Thanks for the heads up. The preexisting constant must stay. I've also made changes to handle the zipped/unzipped cases differently, as some internal code relies on the zip file (not the directory) being a spawn output.

src/test/shell/bazel/bazel_test_test.sh
tjgq: Let's test all combinations, then.

coeuvre (approved):

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15203, opened 2022-04-08
Default experimental_inmemory_dotd_files to true

keith:
I asked around a bit about this flag and it sounds like there isn't a
downside to flipping this on by default. The upside is that if you
switch between using `--remote_download_minimal` and not, or switch
between a remote execution config that set this flag, but weren't
setting it globally, you would thrash your outputs and rebuild the
world. This was just a subtle gotcha if you were scoping this flag to
one of these specific configurations.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppOptions.java
@@ -690,7 +690,7 @@ public class CppOptions extends FragmentOptions {
 
   @Option(
       name = "experimental_inmemory_dotd_files",
-      defaultValue = "false",
+      defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.BUILD_TIME_OPTIMIZATION,
       effectTags = {
         OptionEffectTag.LOADING_AND_ANALYSIS,


comius:
@keith are there any "stronger" upsides?

cc @ericfelly Are you aware of any problems to flip this by default? Or why it is not the case at the moment? (Similar PR is https://github.com/bazelbuild/bazel/pull/15204)

One problem I see is that functionality like this is not implementable in Starlark. But this is a bit far fetched, because this in not the only thing not-implementable.

keith:
I think having fewer flags that developers have to know about is an upside, and this would ideally remove these from the list. I know @aiuto is particularly interested in removing flags, which these could be a candidate for. In general if we don't know of a downside having things work correctly for users out of the box is a strong upside.

ericfelly:
It would probably be OK to make this the default in most organizations. At Google, with ObjFS, we are able to amortize the cost of downloading the same .d file across several builds (on the same machine). But that is not available with in-memory mode.

If an amortizing file system is not in use in the output tree, then in-memory mode is probably the right default. Even with something like ObjFS, in-memory output can be the right choice as disk bandwidth decreases and / or network bandwidth increases.

You can see why making a global decision here is hard.

aiuto:
I  am generally in favor of turning experimental flags true as soon as possible and then deleting them. But @comius raised the most important issue - that this is not implementable in Starlark. If we keep enabling even more special accelerations in the C++ rules, do we run the risk of blocking starlark migration entirely?

keith:
If using a custom FS is the only case where it makes sense to default this to false maybe we should flip it globally and google can flip it off internally? I agree it's never going to be obvious 100% what the default should be for some cases, but if what you say about this being the right default for "most organizations", then it sounds like we should flip it?

@aiuto that's a good question, right now the starlark implementation just depends on this native logic, but I'm not sure, is there a specific plan documented somewhere about how these 2 features would would work or if they would ever be migrated?

lberki:
My $.02: I would be happy if this was the only thing standing in the way of migrating to Starlark. Not enabling this by default also doesn't make the migration to Starlark simpler, because this flag is heavily used at Google so we'll need to find a solution to it no matter what.

Thus, the decision whether to flip the flag should not hinge upon whether it's implementable in pure Starlark or not.

lberki:
I'll follow up in private (I can't publish Google performance data here)

ericfelly:
I wouldn't oppose a global flip. We should probably make this non-experimental while we're at it. Waiting on private metric results.

keith:
Yea happy to change the name separately from the default flip if folks are ok with that

comius (approved):

PR #15204, opened 2022-04-08
Default experimental_inmemory_jdeps_files to true

keith:
I asked around a bit about this flag and it sounds like there isn't a
downside to flipping this on by default. The upside is that if you
switch between using `--remote_download_minimal` and not, or switch
between a remote execution config that set this flag, but weren't
setting it globally, you would thrash your outputs and rebuild the
world. This was just a subtle gotcha if you were scoping this flag to
one of these specific configurations.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/java/JavaOptions.java
@@ -195,7 +195,7 @@ public class JavaOptions extends FragmentOptions {
 
   @Option(
       name = "experimental_inmemory_jdeps_files",
-      defaultValue = "false",
+      defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.BUILD_TIME_OPTIMIZATION,
       effectTags = {
         OptionEffectTag.LOADING_AND_ANALYSIS,


comius:
I'll also ask this here: are there any measurements, how much the flip of this flag improves/regresses the builds? (Without the part where flip "thrash-es your outputs and rebuild-s the world")

keith:
Would you expect this to have a significant difference on things? Considering this is required for build without the bytes I guess we've just been accepting any cost

comius (approved):

comius:
Update: Bazel's team is working on merging this cl. It revealed a minor problem internally. We will merge it after that is resolved.

gregestren:
FYI the internal problem is a test failure on config-stripped actions, i.e. remote executor actions that turn `bazel-out/x86-fastbuild/bin/...` to `bazel-out/bin/..` for better caching.

I've been prototyping that for a while (too long) and been running experiments on Google's remote executor, hence that being a Google test.

I'm also collaborating with @fmeum to expose this functionality to Bazel, at which point stuff like this should be more transparent.

brentleyjones:
@comius @coeuvre Think we can now rename `experimental_inmemory_jdeps_files` and `experimental_inmemory_dotd_files` to remove the `experimental_` prefix?

coeuvre:
SGTM, but I will leave the final decision to @comius in case he has other thoughts.

brentleyjones:
They don't expand to them, but they are enabled by default. Was the requirement that they be enabled at all removed at head?

coeuvre:
> They don't expand to them, but they are enabled by default. 

That's why Tiago suggested to flip and remove it.

> Was the requirement that they be enabled at all removed at head?

Yes.

We don't do this internally so we would like to remove it if possible to make the code simpler.

brentleyjones:
I believe some people use the in memory dotd files to work around a cache poisoning issue with those files (they contain absolute paths, and the stuff that influences their creation isn't part of the action key). So I think that would need to be fixed first.

Am I recalling correctly @keith?

keith:
Yea, I don't remember the exact repro but something with changing bazel's allowed header include directories after you had a failure about an undeclared header exclusion leaves you in a broken state since the `.d` files are in the disk cache at that point.

PR #15205, opened 2022-04-08
Remove noisy include scanning warning

keith:
Given that this flag is called `--experimental_unsupported_and_brittle_include_scanning` I don't think it's necessary to have another warning about this being unsupported.


oquenchil:
Adding this came up during the review. We do want it to be this noisy. In case this has to be removed in the future, we don't want any misunderstandings.

PR #15208, opened 2022-04-11
Fix broken link in glossary.md

nathyong:
The link was both malformed and did not point anywhere in particular.


sgowroji (approved):
LGTM. Thanks

Issue #14900, opened 2022-02-24
Bazel uploads outputs of no-remote actions when used in combination with a disk-cache


BalestraPatrick:
### Description of the problem / feature request:

I've seem to have a discovered a bug that is slowing down our remote execution builds. Our build has some pretty large blobs that we don't want to upload to the remote cache since it would take way too long, so this is an example of how we disable those specific actions from running remotely and uploading to the remote cache:
```
build --modify_execution_info=^(CppLink|ObjcLink)$=+no-remote
```

With Bazel 5, we also enabled the use of a disk cache with our remote exec build. This is now causing long uploads for actions that are marked as `no-remote`.

What seems to be happening is the following:
1. When using a combined cache, Bazel wrongly asks the remote cache if it contains a certain blob without respecting the `no-remote` tag. 
2. If the remote cache doesn't contain it, Bazel will then upload it even though it shouldn't.

I've verified that disabling the disk cache or setting the action mnemonics to `no-cache` works around the issue.

### Feature requests: what underlying problem are you trying to solve with this feature?

Building with remote execution and disk cache should be efficient. This means being able to prevent certain action mnemonics inputs/outputs to be uploaded without disabling the disk cache completely.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This needs a working remote exec cluster, but I believe it should be reproducible by building https://github.com/bazelbuild/rules_apple/blob/master/examples/ios/HelloWorld/BUILD with the following flags:
```
build --disk_cache=~/.cache/bazel_disk
build --modify_execution_info=^(CppLink|ObjcLink)$=+no-remote
build --remote_executor=your.remote.exec
build --remote_cache=your.remote.cache
```

### What operating system are you running Bazel on?

macOS 12.1

### What's the output of `bazel info release`?

release 5.0.0

### Any other information, logs, or outputs that you want to share?

This is a snippet of our gRPC log with some of the interesting fields:
```
---------------------------------------------------------
metadata {
  tool_details {
    tool_name: "bazel"
    tool_version: "5.0.0"
  }
  action_mnemonic: "ObjcLink"
  target_id: "//:target"
}
status {
}
method_name: "build.bazel.remote.execution.v2.ContentAddressableStorage/FindMissingBlobs"
details {
  find_missing_blobs {
    request {
      blob_digests {
        hash: "1a0fe7ea9f46605fa721fd83d8498ccbae3b2bfa25c98420f429980076022c88"
[...]

PR #15212, opened 2022-04-11
Remote: Fix a bug that outputs of actions tagged with no-remote are u…

coeuvre:
…ploaded to remote cache when remote execution is enabled.

Fixes https://github.com/bazelbuild/bazel/issues/14900.

Also fixes an issue that action result from just remotely executed action is not saved to disk cache. The root cause is the action result is inlined in the execution response hence not downloaded through remote cache, hence not saved to disk cache. This results in the second build misses the disk cache, but it can still hit the remote cache and fill the disk cache. The third build can hit disk cache.


src/main/java/com/google/devtools/build/lib/remote/common/RemoteActionExecutionContext.java
```diff
@@ -13,37 +13,71 @@
...
+    UPLOAD_INPUTS,
+    EXECUTE_REMOTELY,
+    UPLOAD_OUTPUTS,
+    DOWNLOAD_OUTPUTS,
+    UPLOAD_BES_FILES,
   }
 
-  /** Returns the {@link Type} of the context. */
-  Type getType();
+  private final Spawn spawn;
```
tjgq, line 35: Mark as `@Nullable`?

src/main/java/com/google/devtools/build/lib/remote/common/RemoteActionExecutionContext.java
```diff
@@ -13,37 +13,71 @@
...
+  private final RequestMetadata requestMetadata;
+  private final NetworkTime networkTime;
+
+  @Nullable private ExecuteResponse executeResponse;
+  private Step step;
+
+  public RemoteActionExecutionContext(
+      Spawn spawn, RequestMetadata requestMetadata, NetworkTime networkTime) {
```
tjgq, line 43: Same as above.

tjgq: Also - should we make the ctor private to enforce construction through one of the `create()` methods?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -258,7 +259,7 @@ static Command buildCommand(
   public static class RemoteAction {
     private final Spawn spawn;
     private final SpawnExecutionContext spawnExecutionContext;
-    private final RemoteActionExecutionContext remoteActionExecutionContext;
+    private final RemoteActionExecutionContext context;
```
tjgq, line 262: Is it a good idea to rename this, given that there are two different context objects?

src/main/java/com/google/devtools/build/lib/remote/common/RemoteActionExecutionContext.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/common/RemoteActionExecutionContext.java
coeuvre: Done.

tjgq (approved):

Issue #13989, opened 2021-09-14
Failing to run d8_dexbuilder with buildtools 31


Kernald:
### Description of the problem / feature request:

Using buildtools 31, Bazel can't build dex files anymore. With d8 disabled (default):
```
ERROR: /private/var/tmp/_bazel_kernald/dc8fa02be43b6315d36f6bf2bf056429/external/androidsdk/BUILD.bazel:13:25: Extracting interface @androidsdk//:dx_jar_import failed: missing input file 'external/androidsdk/build-tools/31.0.0/lib/dx.jar', owner: '@androidsdk//:build-tools/31.0.0/lib/dx.jar'
ERROR: /private/var/tmp/_bazel_kernald/dc8fa02be43b6315d36f6bf2bf056429/external/androidsdk/BUILD.bazel:13:25 Extracting interface @androidsdk//:dx_jar_import failed: 1 input file(s) do not exist
```

With d8 enabled with the following flags:
```
# Enable d8 merger
build --define=android_dexmerger_tool=d8_dexmerger

# Flags for the d8 dexer
build --define=android_incremental_dexing_tool=d8_dexbuilder
build --define=android_standalone_dexing_tool=d8_compat_dx
build --nouse_workers_with_dexbuilder
```

```
ERROR: /private/var/tmp/_bazel_kernald/dc8fa02be43b6315d36f6bf2bf056429/external/maven/BUILD:4748:11: Dexing external/maven/_dx/androidx_transition_transition/classes_and_libs_merged.jar_desugared.jar with applicable dexopts [] failed: (Exit 1): d8_dexbuilder failed: error executing command bazel-out/darwin-py2-opt-exec-2B5CBBC6/bin/external/bazel_tools/tools/android/d8_dexbuilder ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Error: Could not find or load main class com.android.tools.r8.compatdexbuilder.CompatDexBuilder
Caused by: java.lang.ClassNotFoundException: com.android.tools.r8.compatdexbuilder.CompatDexBuilder
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

In a `WORKSPACE`:
```
android_sdk_repository(
    name = "androidsdk",
    api_level = 31,
    build_tools_version = "31.0.0",
)
```

(I don't think the `api_level` actually matters here.)

Try to build anything that requires dexing.

### What operating system are you running Bazel on?

[...]

PR #15213, opened 2022-04-11
Replacing Compat Dx With Reference To Internal Bazel Build's Dx

EdbertChan:
**Initial Issue:**
Bazel fails to build on build_tools_version 31.

Error: Could not find or load main class com.android.tools.r8.compatdx.CompatDx

https://github.com/bazelbuild/bazel/commit/ceb498fd6b9323a738f0ba628d12a23a161a0671 should have fixed the issue. However, I am not able to compile.

The error I get from --verbose_failures boils down to this line
bazel-out/k8-opt-exec-2B5CBBC6/bin/external/androidsdk/d8_compat_dx --dex '--num-threads=5' '--output=bazel-out/k8-fastbuild/bin/apps/presidio/helix/app-apk/_mobile_install/bin_debug/stub_application/classes.dex' bazel-out/k8-fastbuild/bin/apps/presidio/helix/app-apk/_mobile_install/bin_debug/stub_deploy.jar

When I look into the d8_compat_dx, I only see that a manifest file exists in it where it says it was created by "Target-Label: @androidsdk//:d8_compat_dx" and no other classes. Looking at this line in [android_sdk_repository.bzl,](https://github.com/bazelbuild/bazel/blob/66d07f096cb697bdadb6f9d9fbc1c4fe33be6d59/tools/android/android_sdk_repository_template.bzl#L306) it looks like Bazel is still trying to use the d8 packaged in build-tools in the Android SDK rather than trying to create its own.

See https://github.com/bazelbuild/bazel/issues/13989 for the original discussion.

**Fix**:

Make the d8_compat_dx reference the internal Bazel built dx tools.

We tested this by running bazel mobile-install on all our apps.

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/r8/BUILD.tools
+++ b/src/tools/android/java/com/google/devtools/build/android/r8/BUILD.tools
@@ -22,3 +22,10 @@ java_binary(
     visibility = ["//tools/android:__subpackages__"],
     runtime_deps = [":r8"],
 )
+
+java_binary(
+    name = "r8_binary",
+    main_class = "com.google.devtools.build.android.r8.CompatDx",
+    visibility = ["@androidsdk//:__subpackages__", "//tools/android:__subpackages__"],
+    runtime_deps = [":r8"],
+)
--- a/tools/android/android_sdk_repository_template.bzl
+++ b/tools/android/android_sdk_repository_template.bzl
@@ -296,10 +296,9 @@ def create_android_sdk_rules(
         name = "dx_jar_import",
         jars = ["build-tools/%s/lib/dx.jar" % build_tools_directory],
     )
-    java_binary(
+    native.alias(
         name = "d8_compat_dx",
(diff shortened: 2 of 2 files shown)


ahumesky:
Ah thanks for the change here, but coincidentally as part of other improvements, we've made almost identical changes that should address this that we're pushing out likely today

EdbertChan:
@ahumesky Could you link the commit where this change took place in the main branch just for future reference in case other people encounter this?

EdbertChan:
Nevermind I see it here:
[b93f828133e74cb3589ba7ffcfe74d2fe72430cd](https://github.com/bazelbuild/bazel/commit/b93f828133e74cb3589ba7ffcfe74d2fe72430cd)

Closing out

ahumesky:
@EdbertChan Here's the commit: https://github.com/bazelbuild/bazel/commit/b93f828133e74cb3589ba7ffcfe74d2fe72430cd

ahumesky:
Ah I see you already found it :)

PR #15215, opened 2022-04-11
[Step 1] Upgrade Google Auth Version

ThomasCJY:
upgrade google auth to use the latest v1.6.0 instead of the outdated 0.17 in bazel

pt 1: Add Gson and upgrade Auth (third_party only changes)

part of https://github.com/bazelbuild/bazel/pull/15176


third_party/BUILD
```diff
@@ -163,8 +163,10 @@ distrib_java_import(
     jars = [
         "api_client/google-api-client-1.22.0.jar",
         "api_client/google-api-client-jackson2-1.22.0.jar",
```
coeuvre, line 165: I think we can remove two jackson2 jars in this PR.

third_party/BUILD
```diff
@@ -163,8 +163,10 @@ distrib_java_import(
...
+        "api_client/google-api-client-gson-1.22.0.jar",
         "api_client/google-http-client-1.22.0-SNAPSHOT.jar",
         "api_client/google-http-client-jackson2-1.22.0.jar",
+        "api_client/google-http-client-gson-1.22.0.jar",
     ],
     runtime_deps = [
```
coeuvre, line 171: need `:gson` here.

coeuvre (changes requested):

third_party/BUILD
ThomasCJY: wouldn't that require us to update test code `import com.google.api.client.json.jackson2.JacksonFactory;` as well?

third_party/BUILD
ThomasCJY: thanks. updated

third_party/BUILD
coeuvre: Right. I thought only `:jackson2` is required for the test code. Please add them back and we are ready to merge!

ThomasCJY: pr updated

coeuvre (approved):
Thanks!

@meteorcloudy Can you help import this PR?

meteorcloudy:
@ThomasCJY Can you squash all commits into a single one? That's easier for importing.

ThomasCJY:
@meteorcloudy @coeuvre ^PTAL

PR #15232, opened 2022-04-13
Let Starlark executable rules specify their environment

### LINKED ISSUE #7364: Add common env attribute for *_binary and *_test

In the same way as we can define in code that bazel passes on cli args on `bazel run` and `bazel test` it would be great to be able to define env vars in the same way. 

Some applications/libraries/tests/test frameworks prefer environment variables over cli args and being able to define that on a per target level in code would save the need to use external tools or `test_env` in a lot of cases. We also have cases where we just need env variables in ci for certain tests (that use docker) and this could also accommodate for that. Also with make var substitution or stamping we could even pass in non static env vars this way. 

This could also help solving part of the issue described in #6111

Also I noticed on our team, developers where quite surprised when they saw one can set `args` in code but not `env`. So it seems some developers even have some expectations around this behaviour. 

One could of course work around this by wrapping the binary into a `sh_binary` or similar and that is probably a reasonable solution, but the same would be true for `args` so parity on this would be ideal. 

Not really, maybe #955 is somewhat related.

### LINKED ISSUE #15224: Allow testing.TestEnvironment / TestEnvironmentInfo to be read from Starlark
### Description of the bug:

Given a target that returns the testing.TestEnvironment / TestEnvironmentInfo, that provider cannot be read from the target.

This makes it impossible to easily test that the rule constructs an appropriate TestEnvironmentInfo object.

```
def _provides_info(ctx):
  return [testing.TestEnvironment({})]

provides_info = rule(implementation = _provides_info)

def _reads_info(ctx):
  print(ctx.attr.target[testing.TestEnvironment])

reads_info = rule(implementation=_reads_info, attrs={"target": attr.label()})

provides_info(name="provides")
reads_info(name="reads", target=":provides")
```

Actual behavior:

An error like below occurs:
```
Error: Type Target only supports indexing by object constructors, got builtin_function_or_method instead
```

Expected behavior:

The TestEnvironmentInfo provider instance is returned.

Linux

google build

Google build

```text
n/a
```

None found

_No response_

### LINKED ISSUE #15225: testing.TestEnvironment not equal to itself
### Description of the bug:

The `testing.TestEnvironment` object is a bit strange: it isn't equal to itself. In comparison, most provider keys are equal to themselves.

This makes it difficult to write code that can properly detect when the testing.TestEnvironment provider key was requested, e.g.

```
def get_provider(key):
  if key == testing.TestEnvironment: ...
```
(in my particular case, I'm writing helper code for tests to assert TestEnvironment)

Also, I'm not sure if this should be TestEnvironment or TestEnvironmentInfo -- the public symbol is `testing.TestEnvironment`, but the docs say it returns `TestEnvironmentInfo` (note no leading "testing" namespace and the extra Info suffix).

```
print(testing.TestEnvironment == testing.TestEnvironment)
```

actual: prints false

expected: prints true

linux

google build

n/a

```text
n/a
```

[15224 ](https://github.com/bazelbuild/bazel/issues/15224) is sort of related -- same category of "TestEnvironemt provider acts weird"

_No response_

fmeum:
The new RunEnvironmentInfo provider allows any executable or test
Starlark rule to specify the environment for when it is executed, either
as part of a test action or via the run command.

Refactors testing.TestEnvironment to construct a RunEnvironmentInfo and
adds a warning (but not an error) if the provider constructed in this
way is returned from a non-executable non-test rule. If a
RunEnvironmentInfo is constructed directly via the Starlark constructor,
this warning becomes an error.

Fixes #7364
Fixes #15224 
Fixes #15225


fmeum:
@comius Alternative names I considered are `EnvironmentInfo` (which I like better than `RunEnvironmentInfo`, but may be too generic) or `ExecutionEnvironmentInfo`. What do you think?

lberki:
Ivo, are you the right person for this or is it @brandjon ?

comius:
> Ivo, are you the right person for this or is it @brandjon ?

I've reviewed the `TestEnvInfo` already, so it seems I'm the right person.

src/main/java/com/google/devtools/build/lib/analysis/RunEnvironmentInfo.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public RunEnvironmentInfoApi constructor(Dict<?, ?> environment,
+        Sequence<?> inheritedEnvironment) throws EvalException {
+      return new RunEnvironmentInfo(
+          Dict.cast(environment, String.class, String.class, "environment"),
+          StarlarkList.immutableCopyOf(
+              Sequence.cast(inheritedEnvironment, String.class, "inherited_environment")),
+          true);
```
rickeylev, line 91: Not sure if Bazel follows this style, but it's fairly popular within Google: Use an inline comment for bool literals:
`/* shouldErrorOnNonExecutableRule= */ true`

This helps the reader infer what it does without having to jump over the class, find the corresponding ctor, and match the args to the call site.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/RunEnvironmentInfoApi.java
```diff
@@ -0,0 +1,96 @@
...
+@StarlarkBuiltin(
+    name = "RunEnvironmentInfo",
+    category = DocCategory.PROVIDER,
+    doc = "A provider that can be returned from executable rules to control the environment in"
+        + " which their executable is executed.")
+public interface RunEnvironmentInfoApi extends StructApi {
+
+  @StarlarkMethod(
+      name = "environment",
+      doc = "A dict containing environment variables which should be set when executing the rule"
```
rickeylev, line 44: Does this doc show up somewhere? It'd be nice if it repeated the key/value data types, or referred the reader to the ctor-doc. Similar for "what happens if the same keys are in inherited_environment"

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/test/TestEnvironmentInfoApi.java
```diff
@@ -14,25 +14,9 @@
...
-      structField = true)
-  Map<String, String> getEnvironment();
-
-  @StarlarkMethod(
-      name = "inherited_environment",
-      doc = "A list of variables that the test action should inherit from the shell environment.",
-      structField = true)
-  List<String> getInheritedEnvironment();
-}
+public interface TestEnvironmentInfoApi extends RunEnvironmentInfoApi {}
```
rickeylev, line 22: Maybe just delete TestEnvironmentInfoAPI, too? It looks like the only usage of it is in TestingModuleApi. The TestEnvironmentInfo symbol was never exposed to Starlark, and Starlark code was only able to see instances of it as "<provider>". So changing its type from TestEnvironmentInfo to RunEnvironmentInfo is probably fine.

rickeylev:
I'm interested in using this functionality, so I reviewed what I could. I'm not too familiar with apis, so my review wasn't very deep.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/test/TestEnvironmentInfoApi.java
fmeum: I agree that keeping the type around doesn't achieve much, the chance of breakage here seems entirely theoretical. I removed it.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/RunEnvironmentInfoApi.java
fmeum: I'm not so sure, but just in case, I copied the docs as I don't know whether it's possible to include a linked cross ref.

fmeum:
> I'm interested in using this functionality, so I reviewed what I could. I'm not too familiar with apis, so my review wasn't very deep.

Thanks for the comments @rickeylev, I pushed a new commit to address them (and resolved the merge conflicts).

rickeylev:
LGTM.

@comius can you review?

comius (approved):

fmeum:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

fmeum:
@ckolli5 I created https://github.com/bazelbuild/bazel/pull/15766 with the additional required cherry-picks.

Issue #14982, opened 2022-03-07
Android NDK 32 bit Arm should be marked "armv7" not "arm"


ttiurani:
### Description of the problem / feature request:

When building Android NDK with `--fat_apk_cpu=armeabi-v7a` the platform it hits is cpu `arm` and not cpu `armv7`, the value here:

https://github.com/bazelbuild/rules_android/blob/master/android/platforms/BUILD#L6 

### Feature requests: what underlying problem are you trying to solve with this feature?

Android platforms should work right, this came up with trying to get rust to work with Android, and [this kind of special platform](https://github.com/ttiurani/examples/blob/main/android/ndk-rust/BUILD.bazel#L4) was needed. This problem makes it hard to target the right rust target `thumbv7neon-linux-androideabi` in rules_rust, which could be mapped if the right `armv7` was set.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This repository shows the needed special platform workaround needed:
https://github.com/ttiurani/examples/tree/main/android/ndk-rust

### What operating system are you running Bazel on?

Same issue on OSX and linux.

### What's the output of `bazel info release`?

release 5.0.0 (but also latest 6.0.0 preview has the problem)
> Replace this line with your answer.

###  Have you found anything relevant by searching the web?

This was talked about on slack in this thread:

https://bazelbuild.slack.com/archives/CSV56UT0F/p1645029913244809

gregestren:
Are you setting `--platforms` or `--android_platforms` on your build? Or only `--fat_apk_cpu`?

I'm asking because we're moving toward the former via https://bazel.build/concepts/platforms-intro but it's Work In Progress. So users have to actively opt into making `platform` matter, and that extra context can help us evaluate.

ttiurani:
I'm setting platforms via `platform_mappings`:

https://github.com/ttiurani/examples/blob/main/android/ndk-rust/platform_mappings

This is because rules_rust uses platforms and the stable android doesn't. The README here:

https://github.com/ttiurani/examples/blob/main/android/ndk-rust

describes also the build command I use for the project, i.e. just `--fat_apk_cpu`.

I was aware of the platforms WIP branch, but didn't use it here.

keith:
[...]

PR #15248, opened 2022-04-13
Fix android armeabi-v7a constraint

keith:
Previously this was arm instead of armv7. It seems that armv7 is the
correct constraint for this case.

Fixes https://github.com/bazelbuild/bazel/issues/14982


tools/osx/crosstool/BUILD.toolchains
```diff
@@ -7,7 +7,10 @@ exports_files(["osx_archs.bzl"])
 # Target constraints for each arch.
 # TODO(apple-rules): Rename osx constraint to macOS.
 OSX_TOOLS_CONSTRAINTS = {
-    "armeabi-v7a": ["@platforms//cpu:armv7"],
+    "armeabi-v7a": [
+        "@platforms//cpu:armv7",
+        "@platforms//os:android",
```
katre, line 12: This can't be right, shouldn't this be osx?

keith: IIUC this CPU is just a shim because the android SDK requires it to exist even if it isn't valid. With the macOS toolchain it will actually just generate a dummy toolchain that does nothing. https://github.com/bazelbuild/bazel/pull/13449

I tried to remove it entirely again recently, but it's still hit in some cases https://github.com/bazelbuild/bazel/pull/14872

katre: I see, thanks for explaining.

katre:
Ran the internal tests and, as I suspected, there's no impact. I'm going to work on getting this submitted in the next few days.

cpsauer:
Yay, @keith! Yay, @katre!

PR #15249, opened 2022-04-14
Do not run GrpcCacheClientTest twice.

benjaminp:
Factor common code from GrpcCacheClientTest into GrpcCacheClientTestExtraBase, so GrpcCacheClientTestExtraExtra can use it without getting every test case, too.

(It is not clear to me why GrpcCacheClientTestExtraExtra even exists. Google cannot run this test?)


coeuvre (approved):
Thanks!

Yes, we don't run `GrpcCacheClientTestExtra` internally. https://github.com/bazelbuild/bazel/pull/14041#issuecomment-960496515

PR #15251, opened 2022-04-14
Fix broken code block in "Java and Bazel"

kjcartledge:
Without this blank line, the code block fails to render correctly.


gregestren (approved):
Indeed: https://bazel.build/docs/bazel-and-java

Thanks for catching.

PR #15256, opened 2022-04-14
Store start/end timestamps of locally executed actions in the disk cache.

### LINKED ISSUE #14426: Test report fails to get a cache hit the second time it runs

When running a test twice using remote-cache or disk-cache, bazel fails to cache the action that generates the xml test report (`external/bazel_tools/tools/test/generate-xml.sh`) on the second invocation. This happens because the action use as arguments the walltime duration of the previous action (the action running the test, `external/bazel_tools/tools/test/test-setup.sh`), and that time is zero when the action gets a cache hit on the second invocation.

Using the following (repository)[https://github.com/segonpin/bazel_test_cache_miss] (or any simple test that takes longer than 1 second to run), just run the test three times.
```
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec1.json
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec2.json
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec3.json
```

The first time there are 0 cache-hits, the third time everything gets a cache-hit, but the second time one action does not get a cache-hit, and that seems incorrect to me.

Ubuntu 20.04

release 6.0.0-pre.20211117.1

When we run the test for the first time. It will run 7 actions, of which 4 are linux-sandbox. Notice that the test takes 5 seconds to run.
```
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec1.json
...
INFO: 7 processes: 3 internal, 4 linux-sandbox.
INFO: Build completed successfully, 7 total actions
//:sleep_test                                                            PASSED in 5.0s
...
```

Then we run the test for the second time. It will run 7 actions, and now only 1 is linux-sandbox. None should be. The reason for the cache-miss is that the test duration is used as input for the missed action, and because the test was cached now, the duration has changed from 5 seconds to 0 seconds, so the commandArgs was different for the action.
```
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec2.json
...
INFO: 7 processes: 3 disk cache hit, 3 internal, 1 linux-sandbox.
INFO: Build completed successfully, 7 total actions
//:sleep_test                                                   (cached) PASSED in 0.0s
...
```

Now we run the test again, and now the test run in 0 seconds, as the previous time, so that the action printing the xml report get a cache hit from before.
```
$ bazelisk clean; bazelisk test //... --disk_cache=/tmp/bazel_disk_cache --execution_log_json_file=exec3.json
...
INFO: 7 processes: 4 disk cache hit, 3 internal.
INFO: Build completed successfully, 7 total actions
//:sleep_test                                                   (cached) PASSED in 0.0s
...
```

The action getting the cache miss runs
```
# in the first invocation
"commandArgs": ["external/bazel_tools/tools/test/generate-xml.sh", "bazel-out/k8-fastbuild/testlogs/sleep_test/test.log", "bazel-out/k8-fastbuild/testlogs/sleep_test/test.xml", "5", "0"],
# second and third invocation
"commandArgs": ["external/bazel_tools/tools/test/generate-xml.sh", "bazel-out/k8-fastbuild/testlogs/sleep_test/test.log", "bazel-out/k8-fastbuild/testlogs/sleep_test/test.xml", "0", "0"],
```
And the generated test report looks like follows (using 5 the first invocation, and 0 the second and third invocations)
```.xml
$ cat bazel-out/k8-fastbuild/testlogs/sleep_test/test.xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="sleep_test" tests="1" failures="0" errors="0">
    <testcase name="sleep_test" status="run" duration="5" time="5"></testcase>
      <system-out>
Generated test.log (if the file is not UTF-8, then this may be unreadable):
<![CDATA[exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //:sleep_test
-----------------------------------------------------------------------------]]>
      </system-out>
    </testsuite>
</testsuites>
```

In my opinion the test report should be always the first one (with the duration="5"), or otherwise the last action should not be remoteCacheable.

tjgq:
This is necessary to ensure that the generate-xml.sh action associated with
every test, which embeds the test runner wall time, doesn't execute again when
run on a warm disk cache.

This PR makes the following changes:

1. Add SpawnResult#getStartTime alongside existing #getWallTime and populate
   it in various SpawnRunner implementations.
2. Populate the ExecutedActionMetadata with the start/wall times for a
   locally executed spawn when storing it in the disk cache.
3. Populate the SpawnResult with the start/wall times for a spawn
   retrieved from the disk cache.
4. Add a regression test for the scenario described above.

In order to avoid changes to the remote execution protocol, the start/end
time are stored in the existing ActionResult.execution_metadata fields.
Since there is some mismatch between this field and the metrics available
in a SpawnResult, this might be revisited later.

Fixes #14426.

### DIFF
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -3706,4 +3706,40 @@ EOF
   expect_log "Executing genrule .* failed: (Exit 1):"
 }
 
+
+function test_local_test_execution_with_disk_cache() {
+  # Tests that the wall time for a locally executed test is correctly cached.
+  # If not, the generate-xml.sh action, which embeds the wall time, will be
+  # considered stale on a cache hit.
+  # Regression test for https://github.com/bazelbuild/bazel/issues/14426.
+
+  mkdir -p a
+  cat > a/BUILD <<EOF
+sh_test(
+  name = 'test',
+  srcs = ['test.sh'],
+)
+EOF
+  cat > a/test.sh <<EOF
+sleep 1
+EOF
+  chmod +x a/test.sh
+
+  CACHEDIR=$(mktemp -d)
+
+  bazel test \
+    --disk_cache=$CACHEDIR \
+    //a:test >& $TEST_log || "Failed to build //a:test"
+
+  expect_log "5 processes: 3 internal, 2 .*-sandbox"
+
+  bazel clean
+
+  bazel test \
+    --disk_cache=$CACHEDIR \
+    //a:test >& $TEST_log || "Failed to build //a:test"
+
+  expect_log "5 processes: 2 disk cache hit, 3 internal"
+}
+
 run_suite "Remote execution and remote cache tests"
--- a/src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/SpawnResult.java
@@ -26,6 +26,7 @@ import com.google.devtools.build.lib.vfs.Path;
 import com.google.protobuf.ByteString;
 import java.io.InputStream;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.Locale;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -170,6 +171,14 @@ public interface SpawnResult {
   String getRunnerSubtype();
 
   /**
+   * Returns the start time for the {@link Spawn}'s execution.
+   *
+   * @return the measurement, or empty in case of execution errors or when the measurement is not
+   *     implemented for the current platform
+   */
+  Optional<Instant> getStartTime();
+
+  /**
    * Returns the wall time taken by the {@link Spawn}'s execution.
    *
    * @return the measurement, or empty in case of execution errors or when the measurement is not
@@ -270,6 +279,7 @@ public interface SpawnResult {
     private final String runnerName;
     private final String runnerSubtype;
     private final SpawnMetrics spawnMetrics;
+    private final Optional<Instant> startTime;
     private final Optional<Duration> wallTime;
     private final Optional<Duration> userTime;
     private final Optional<Duration> systemTime;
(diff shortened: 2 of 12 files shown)


sgowroji:
Hello @tjgq, Could you please take care of the failure buildkite presubmit checks. Thanks!

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -3706,4 +3706,38 @@ EOF
...
+
+  mkdir -p a
+  cat > a/BUILD <<EOF
+sh_test(
+  name = 'test',
+  srcs = ['test.sh'],
+)
+EOF
+  cat > a/test.sh <<EOF
+sleep 5
```
coeuvre, line 3722: Can we change the test to not use `sleep` as it will slow down the whole test?

coeuvre (approved):

src/test/shell/bazel/remote/remote_execution_test.sh
tjgq: The generate-xml.sh action embeds the wall time at second granularity [1], so we must sleep for at least a second to reproduce the bug this PR fixes. I added a comment above to make this clearer.

[1] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java;l=522;drc=1f33504af3ebe3a43aa91e607e94edc07b1807f8

coeuvre: Make sense.

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15259, opened 2022-04-14
Flip strict_test_suite default

keith:
It seems like this flag is more of an incompatible flag, and the default
makes more sense to be true so new users don't accidentally start
relying on this.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/query2/query/output/QueryOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/query2/query/output/QueryOptions.java
@@ -158,7 +158,7 @@ public class QueryOptions extends CommonQueryOptions {
 
   @Option(
     name = "strict_test_suite",
-    defaultValue = "false",
+    defaultValue = "true",
     documentationCategory = OptionDocumentationCategory.QUERY,
     effectTags = {OptionEffectTag.BUILD_FILE_SEMANTICS, OptionEffectTag.EAGERNESS_TO_EXIT},
     help =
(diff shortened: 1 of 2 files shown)


keith:
@haxorz bump

Wyverald:
Could you add a RELNOTES section in the commit message so that this is called out in the release notes?

IMO we should be able to flip this.

re @haxorz's concerns:

> (1) Is flipping the default value of this flag desired from a Bazel product perspective?

If, as you said, the ideal situation is one where this flag is flipped, then I think it's desired. People who want the old behavior can still set the flag to false.

> (2) What's the process to follow? Does https://bazel.build/release/backward-compatibility apply? If so, should we introduce a new flag named --incompatible_strict_test_suite (whose semantics are to override the existing --strict_test_suite) and flip that? Or is it fine to use the existing --strict_test_suite?

No renaming is needed due to the new incompatible change policy, so we can just use the existing --strict_test_suite. I also tried to run the incompatible flag pipeline but it fails due to the flag only working for `bazel query`, so we can't really do further checks.

Wyverald:
Could you rebase the PR to HEAD and try the checks again? The cc_coverage thing might have been fixed already.

keith:
yea i didn't actually rebase before, so should be good after this run

Wyverald (approved):

keith:
@Wyverald can you land this?

sgowroji:
@keith, It is already in process through internally. I will update once it is finished.

Wyverald:
FYI this turned out to be rather hard to merge, since it interacts negatively with some internal systems. With all the scrambling going on before the 6.0 cut, I haven't had time to properly make this work, so it'll probably have to wait until 7.0.

keith:
Can you flip externally and disable it internally in the google3 bazelrc?

Wyverald:
Unfortunately it's not that easy. There's a long-running service that Google uses to answer large scale Blaze queries, and its protocol definition sticks to the default value of the flag, instead of reading the .blazerc file. I tried making it work back in September, but couldn't make everything happy.

PR #15261, opened 2022-04-14
Revert "Fixes incorrect install names on darwin platforms"

### LINKED ISSUE #15214: #13427 Broke mac C++ toolchains
### Description of the bug:

The only use of `runtime_solib_name` in the bazel repo is this:
```
tools/osx/crosstool/cc_toolchain_config.bzl:                            "-Wl,-install_name,@rpath/%{runtime_solib_name}",
```
This line is now broken.

The way it worked previously is bazel will inject rpaths into an executable, ie:
```
path @loader_path/../../_solib_arm64/
```
This directory contains symlinks for all the mangled names of shared libraries that have been built.

This can only work if the install_name for a shared library is set to the mangled name. #13427 changes it so that `runtime_solib_name` is no longer mangled. ie, it went from `libexternal_Sboost_Slibasio.dylib` to `libasio.dylib`.

Also, the test as it existed was also correct as cc_test will link its dependencies using dynamic linking by default. The new test is broken as it does not capture the use case this feature was written to support (and I expect the change caused the test to start failing before it was rewritten).

Also see the description of the variable:
```
  /** The name of the runtime solib symlink of the shared library. */
  RUNTIME_SOLIB_NAME("runtime_solib_name");
```
which states that it is the mangled name that exists in the `_solib_*` dir.

Can we please revert #13427? @csmulhern @oquenchil 

_No response_

Macos

5.1.1

_No response_

_No response_

_No response_

_No response_

det:
This reverts commit b06f49507e02384cb4ff1ff862da2a673e9a081f.

Also had to add a missing mnemonic function call parameter to `getDynamicLibrarySoname`. Copy-pasted the parameter from the other use of that function in the codebase.

Tested manually with a custom toolchain and there is also a test from the reverted code.

Fixes #15214

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
@@ -820,7 +820,8 @@ public class CppLinkActionBuilder {
               getLinkType().linkerOrArchiver().equals(LinkerOrArchiver.LINKER),
               configuration.getBinDirectory(repositoryName).getExecPath(),
               output.getExecPathString(),
-              output.getRootRelativePath().getBaseName(),
+              SolibSymlinkAction.getDynamicLibrarySoname(
+                  output.getRootRelativePath(), /* preserveName= */ false),
               linkType.equals(LinkTargetType.DYNAMIC_LIBRARY),
               paramFile != null ? paramFile.getExecPathString() : null,
               thinltoParamFile != null ? thinltoParamFile.getExecPathString() : null,
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkActionBuilder.java
@@ -821,7 +821,9 @@ public class CppLinkActionBuilder {
               configuration.getBinDirectory(repositoryName).getExecPath(),
               output.getExecPathString(),
               SolibSymlinkAction.getDynamicLibrarySoname(
-                  output.getRootRelativePath(), /* preserveName= */ false),
+                  output.getRootRelativePath(),
+                  /* preserveName= */ false,
+                  actionConstructionContext.getConfiguration().getMnemonic()),
               linkType.equals(LinkTargetType.DYNAMIC_LIBRARY),
               paramFile != null ? paramFile.getExecPathString() : null,
               thinltoParamFile != null ? thinltoParamFile.getExecPathString() : null,
--- a/src/test/shell/bazel/cpp_darwin_integration_test.sh
+++ b/src/test/shell/bazel/cpp_darwin_integration_test.sh
@@ -124,59 +124,41 @@ EOF
 }
 
 function test_cc_test_with_explicit_install_name() {
-  mkdir -p cpp/install_name
-  cat > cpp/install_name/BUILD <<EOF
+  mkdir -p cpp
+  cat > cpp/BUILD <<EOF
 cc_library(
   name = "foo",
   srcs = ["foo.cc"],
-)
-cc_binary(
-  name = "libbar.so",
-  srcs = ["bar.cc"],
-  linkshared = 1,
-)
-cc_binary(
-  name = "libbaz.dylib",
-  srcs = ["baz.cc"],
-  linkshared = 1,
+  hdrs = ["foo.h"],
 )
 cc_test(
   name = "test",
-  srcs = ["test.cc", ":libbar.so", ":libbaz.dylib"],
+  srcs = ["test.cc"],
   deps = [":foo"],
 )
 EOF
-  cat > cpp/install_name/foo.cc <<EOF
-  int foo() { return 2; }
-EOF
-  cat > cpp/install_name/bar.cc <<EOF
-  int bar() { return 12; }
-EOF
-  cat > cpp/install_name/baz.cc <<EOF
-  int baz() { return 42; }
-EOF
-  cat > cpp/install_name/test.cc <<EOF
+  cat > cpp/foo.h <<EOF
   int foo();
-  int bar();
-  int baz();
+EOF
+  cat > cpp/foo.cc <<EOF
+  int foo() { return 0; }
+EOF
+  cat > cpp/test.cc <<EOF
+  #include "cpp/foo.h"
   int main() {
-    int result = foo() + bar() + baz();
-    if (result == 56) {
-      return 0;
-    } else {
-      return result;
-    }
+    return foo();
   }
 EOF
 
-  bazel test --incompatible_macos_set_install_name //cpp/install_name:test || \
-      fail "bazel test //cpp/install_name:test failed"
+  bazel test --incompatible_macos_set_install_name //cpp:test || \
+      fail "bazel test //cpp:test failed"
   # Ensure @rpath is correctly set in the binary.
-  ./bazel-bin/cpp/install_name/test || \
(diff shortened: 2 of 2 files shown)


keith:
should we get input from @csmulhern here who originally implemented this in case there's an easy way to fix this foward?

det:
> should we get input from @csmulhern here who originally implemented this in case there's an easy way to fix this foward?

FYI, I pinged him on #15214 9 days ago. Maybe someone has an internal Google way of contacting him?

lberki:
Friendly ping sent to @csmulhern .

csmulhern:
Sorry for the delayed response.

I'm not actually a bazel engineer, so I'm sure there are people with better context here than me. My motivation for the original change was to fix issues I was experiencing with --incompatible_macos_set_install_name, as detailed in b06f495. TL;DR: When using `linkshared = 1` or wrapping a dylib in cc_library, the library name was not mangled, and so the RPATH was being set incorrectly. The former seems like it might have been a behavioral change in bazel in the recent past (i.e. between the initial commit and my "fix"). Regardless, in my view the updated test case seems like it should stay, as it actually exercises the behavior of using a shared library. It appears there are still cases in which mangled library names are used, and the `output` object in `CppLinkAction.build` is not yet properly configured to have its path set as mangled vs unmangled.

In that case, it seems the correct fix would be to basically use the original code:

```
SolibSymlinkAction.getDynamicLibrarySoname(
                   output.getRootRelativePath(),
                   /* preserveName= */ false,
                   actionConstructionContext.getConfiguration().getMnemonic())
```

but to set preserveName based on whether the filename will be mangled or not (which doesn't seem straightforward to figure out). It would be good to get input from a bazel engineer who knows when this would or would not be the case.

csmulhern:
For reference, here is what I observe when running the existing test at HEAD today.

```
> otool -L <test_binary>
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	@rpath/libbar.so (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libbaz.dylib (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1858.112.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

```
> otool -l <test_binary>| rg RPATH -A 2
          cmd LC_RPATH
      cmdsize 80
         path @loader_path/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest (offset 12)
```

```
> ls <test_binary>/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest
libbar.so    libbaz.dylib
```

As you can see, the test binary has its RPATH set to the solib symlink directory. The dynamic libraries are located by searching the RPATH, and the library names in the binary (and the symlink names) are unmangled.

csmulhern:
And with this change:

```
> otool -L <test_binary>
   /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
   @rpath/libtest_Slibbar.so (compatibility version 0.0.0, current version 0.0.0)
   @rpath/libtest_Slibbaz.dylib (compatibility version 0.0.0, current version 0.0.0)
   /System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1858.112.0)
   /usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
   /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

```
> otool -l <test_binary>| rg RPATH -A 2
          cmd LC_RPATH
      cmdsize 80
         path @loader_path/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest (offset 12)
```

```
> ls <test_binary>/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest
libbar.so    libbaz.dylib
```

The test fails as the library names are wrong (e.g. libtest_Slibbar.so in the binary, but libbar.so in the solib symlink directory).

csmulhern:
Note: the above output is when using `--incompatible_macos_set_install_name`, as that is the only time the argument being modified in `CppLinkAction.build` is used.

When not using this flag, the binary locates the shared libraries using @loader_path, and this behavior is the same with or without this change.

```
> otool -L <test_binary>
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	@loader_path/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest/libbar.so (compatibility version 0.0.0, current version 0.0.0)
	@loader_path/../_solib_darwin_arm64/_U_S_Stest_Ctest___Utest/libbaz.dylib (compatibility version 0.0.0, current version 0.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1858.112.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

pcjanzen:
The test that was introduced in b06f4950 is IMO a little borderline. The docs for [cc_binary.linkshared](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_binary.linkshared) explicitly state

> However, for build purposes it will never be linked into the dependent binary, as it is assumed that shared libraries built with a [cc_binary](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_binary) rule are only loaded manually by other programs, so it should not be considered a substitute for the [cc_library](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_library)

so since you are technically not supposed to link to such a binary, only dlopen it, it does not matter what its install_name is.

The whole point of `runtime_solib_name` is to allow cc_toolchains that have the `supports_dynamic_linker` feature to create dynamic `cc_library`s with the correct soname, and #13427 breaks that.

Unfortunately, between the time that #12304 was committed and the time that #13427 was submitted, the default darwin toolchain stopped using `supports_dynamic_linker` (ec5553352), so in the existing test, `//cpp:foo` is now linked statically, never dynamically, and there is no install_name to test.

csmulhern:
Thanks for bringing that to my attention. What would be the appropriate test case then? Is there existing infrastructure for testing with a custom toolchain that has `supports_dynamic_linker` set to true?

If I have a prebuilt dylib that I bring in to bazel using cc_import, I believe that will also result in an unmangled name, and that this should be a supported use case. As such, we would still want to support mangled and unmangled names within `runtime_solib_name`.

csmulhern:
> There are no link actions for a prebuilt dylib, so I'm not sure how you would use `runtime_solib_name` in that case.

You're right. I was under the impression that `--incompatible_macos_set_install_name` was affecting the install name in this case, but it appears that's not the case. The test I had in mind that was failing is in the rules_rust repository, but it appears that that is using `cc_import` on a `cc_binary` rule: https://github.com/bazelbuild/rules_rust/blob/main/examples/ffi/rust_calling_c/c/BUILD.bazel#L30. It sounds like that is just not a supported use case for cc_binary.

I think going ahead and reverting sounds like the right course of action here then, which should fix cc toolchains using `supports_dynamic_linker`. Ideally the test case should be updated to actually test `--incompatible_macos_set_install_name`.

oquenchil (approved):
LGTM

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15274, opened 2022-04-16
Restore old execution log diffing docs

keith:
While the new json execution log file is handy in some cases, it doesn't
provide the same diffing behavior as the binary file and the execlog
parser tool. The json file contains `walltime` attributes for each
action that (likely) differ per run, and it's not in any guaranteed
order.

This is a partial revert of b35944e4fedbfa65a7b081021959584ab9a07a84.


coeuvre:
`--execution_log_json_file` is easier to get started and it is worth mentioning it in the doc. Instead of replacing it, I would suggest add another section for `--execution_log` when diffing matters.

keith:
Added a sentence on that flag

coeuvre (approved):
Thanks!

Issue #6393, opened 2018-10-15
TreeArtifacts output directories aren't created with remote execution enabled


cushon:
Previously: #6262

Repro:

```
cd $(mktemp -d)
git clone git@github.com:cushon/bazel-i6393.git
cd bazel-i6393/
bazel --bazelrc=/dev/null build --config=remote ...
```

```
ERROR: /tmp/tmp.pti7hTUec5/bazel-i6393/BUILD:3:1: error executing shell command: '/bin/bash -c cd bazel-out/k8-fastbuild/bin/a_dir && pwd' failed (Exit 1)
/bin/bash: line 0: cd: bazel-out/k8-fastbuild/bin/a_dir: No such file or directory
Target //:a failed to build
```

ola-rozenfeld:
> The expected behaviour is for the TreeArtifact output directory to be created before the action is executed.

Wait, why? We've always had the opposite expectation, the action itself should create the output directory, just like it should create the output files. This is what the current tests test for, too.

cushon:
> We've always had the opposite expectation

That isn't how the local, forge, or sandbox (post-9b8b79002dc9e62bc3fa56525dc4b01c3c2fe805) strategies handle this case.

My model is that the outputs are the contents of the directory. What is the advantage of requiring the action to create the directory?

The issue that prompted the original bug was that earlier versions of javac write output to a directory, but expect the directory to be pre-existing. We could work around that by creating a wrapper script for javac that creates the directory, but it seems strictly better to just guarantee the directory exists.

cushon:
@philwo for 9b8b79002dc9e62bc3fa56525dc4b01c3c2fe805

cushon:
also @lberki who said the behaviour [of not creating the directories] was 'surprising' in cl/214229700

cushon:
> If we want remote execution to create the output directories before running the action, this must be explicitly specified in the remote execution API

The alternative is to "fix" all of the other spawn strategies to stop creating the output directories, is that the proposal? It feels like a regression to me, but it's probably better than the inconsistency.

> This means no distinction between an empty output directory and a non-existing output directory. Why would we want to add this limitation? 

In what sense is it a limitation? If an action declares an output directory and doesn't create it, currently it succeeds under remote execution and just doesn't produce output corresponding to that directory. If instead an empty directory was created and the action didn't write any output to that directory, wouldn't that have effectively the same result?

> The wrapper idea sounds like the proper fix to me.

[...]

PR #15276, opened 2022-04-18
Unify sandbox/remote handling of empty TreeArtifact inputs

fmeum:
Actions that take a TreeArtifact as input should see a corresponding
directory even if the TreeArtifact is empty.

Previously, SandboxHelpers contained special handling for the case of
empty TreeArtifact action inputs to ensure that they are added to the
sandbox as empty directories. As pointed out in a comment, this logic
should live in SpawnInputExpander, where it would also apply to remote
execution.

This commit adds a integration tests for this previously untested case,
extracts the logic into SpawnInputExpander and adapts
DirectoryTreeBuilder to handle empty TreeArtifacts when creating the
Merkle trees.

Note: The Merkle tree builder now reports an error when it encounters a
non-empty TreeArtifact. Such an artifact should have been expanded by
SpawnInputExpander and if it wasn't, e.g. because it wasn't properly
registered with Skyframe, it can't be expanded at this point anyway.
The tests uncovered that the spawn for split coverage postprocessing declared the
coverage dir artifact as such an input. In this case, it can simply be
removed as the coverage script creates the coverage dir if it doesn't
exist.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ActionInputHelper.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ActionInputHelper.java
@@ -103,12 +103,17 @@ public final class ActionInputHelper {
   }
 
   /**
-   * Expands middleman artifacts in a sequence of {@link ActionInput}s.
+   * Expands middleman and tree artifacts in a sequence of {@link ActionInput}s.
    *
-   * <p>Non-middleman artifacts are returned untouched.
+   * The returned list never contains middleman artifacts. It contains tree artifacts only if
+   * {@code keepEmptyTreeArtifacts} is true. In this case, all returned tree artifacts have empty
+   * expansion under the provided {@link ArtifactExpander}.
+   *
+   * <p>Non-middleman, non-tree artifacts are returned untouched.
    */
   public static List<ActionInput> expandArtifacts(
-      NestedSet<? extends ActionInput> inputs, ArtifactExpander artifactExpander) {
+      NestedSet<? extends ActionInput> inputs, ArtifactExpander artifactExpander,
+      boolean keepEmptyTreeArtifacts) {
     List<ActionInput> result = new ArrayList<>();
     List<Artifact> containedArtifacts = new ArrayList<>();
     for (ActionInput input : inputs.toList()) {
@@ -118,7 +123,8 @@ public final class ActionInputHelper {
       }
       containedArtifacts.add((Artifact) input);
     }
-    Artifact.addExpandedArtifacts(containedArtifacts, result, artifactExpander);
+    Artifact.addExpandedArtifacts(containedArtifacts, result, artifactExpander,
+        keepEmptyTreeArtifacts);
     return result;
   }
 
--- a/src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
@@ -130,7 +130,8 @@ public class SpawnInputExpander {
           if (localArtifact.isTreeArtifact()) {
             List<ActionInput> expandedInputs =
                 ActionInputHelper.expandArtifacts(
-                    NestedSetBuilder.create(Order.STABLE_ORDER, localArtifact), artifactExpander);
+                    NestedSetBuilder.create(Order.STABLE_ORDER, localArtifact), artifactExpander,
+                    /* keepEmptyTreeArtifacts= */ false);
             for (ActionInput input : expandedInputs) {
               addMapping(
                   inputMap,
@@ -222,7 +223,11 @@ public class SpawnInputExpander {
       NestedSet<? extends ActionInput> inputFiles,
       ArtifactExpander artifactExpander,
       PathFragment baseDirectory) {
-    List<ActionInput> inputs = ActionInputHelper.expandArtifacts(inputFiles, artifactExpander);
+    // Actions that accept TreeArtifacts as inputs generally expect the directory corresponding
+    // to the artifact to be created, even if it is empty. We explicitly keep empty TreeArtifacts
+    // here to signal consumers that they should create the directory.
+    List<ActionInput> inputs = ActionInputHelper.expandArtifacts(inputFiles, artifactExpander,
+        /* keepEmptyTreeArtifacts= */ true);
     for (ActionInput input : inputs) {
       addMapping(inputMap, input.getExecPath(), input, baseDirectory);
     }
@@ -230,8 +235,10 @@ public class SpawnInputExpander {
 
   /**
    * Convert the inputs and runfiles of the given spawn to a map from exec-root relative paths to
-   * {@link ActionInput}s. The returned map does not contain tree artifacts as they are expanded to
-   * file artifacts.
+   * {@link ActionInput}s. The returned map does not contain non-empty tree artifacts as they are
+   * expanded to file artifacts. Tree artifacts that would expand to the empty set under the
+   * provided {@link ArtifactExpander} are left untouched so that their corresponding empty
+   * directories can be created.
    *
    * <p>The returned map never contains {@code null} values.
    *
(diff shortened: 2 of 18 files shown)


brentleyjones:
@bazelbuild/remote-execution

fmeum:
Still working on this one, there are test failures I don't understand yet.

fmeum:
@brentleyjones The tests are passing now.

fmeum:
@cushon I just learned about https://github.com/bazelbuild/bazel/issues/6393, which seems quite related to this PR. Would you appreciate a PR that fixes #6393 by having Bazel (rather than the remote executor) create the output directories? Fixing these issues would help ongoing work of @gregestren and me on #6526.

cushon:
Thanks for looping me in @fmeum! I am still interested in a fix for #6393, but I haven't paged in all of the discussion since it was filed in 2018, and it's not a change I can approve by myself. But if you're interested in proposing a solution I can try to at least help find reviewers for it.

fmeum:
@coeuvre Would you be available to review this PR or should I ask someone else? It's my first PR in the area, so I don't know the usual process yet.

coeuvre:
Yes, I can take this one. Thanks for your PR!

src/main/java/com/google/devtools/build/lib/remote/merkletree/DirectoryTreeBuilder.java
```diff
@@ -140,6 +142,18 @@ private static int buildFromActionInputs(
                   metadataProvider.getMetadata(input),
                   "missing metadata for '%s'",
                   input.getExecPathString());
+          // A TreeArtifact's metadata has type REGULAR_FILE, not DIRECTORY, and thus requires
```
coeuvre, line 145: Consider moving this to `build()` where we create (parent) directories. So this function can keep the precondition that all inputs are not tree artifact nor middleman.

src/main/java/com/google/devtools/build/lib/sandbox/SandboxHelpers.java
```diff
@@ -487,27 +487,8 @@ private static void writeVirtualInputTo(VirtualActionInput input, Path target)
    */
   public SandboxInputs processInputFiles(
       Map<PathFragment, ActionInput> inputMap,
-      Spawn spawn,
-      ArtifactExpander artifactExpander,
       Path execRoot)
       throws IOException {
-    // SpawnInputExpander#getInputMapping uses ArtifactExpander#expandArtifacts to expand
-    // middlemen and tree artifacts, which expands empty tree artifacts to no entry. However,
-    // actions that accept TreeArtifacts as inputs generally expect that the empty directory is
```
coeuvre, line 496: > actions that accept TreeArtifacts as inputs generally expect that the empty directory is created. So we add those explicitly here.

I think this comment is important. Maybe move to an appropriate place?

src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
```diff
@@ -222,7 +223,8 @@ private static void addInputs(
       NestedSet<? extends ActionInput> inputFiles,
       ArtifactExpander artifactExpander,
       PathFragment baseDirectory) {
-    List<ActionInput> inputs = ActionInputHelper.expandArtifacts(inputFiles, artifactExpander);
+    List<ActionInput> inputs = ActionInputHelper.expandArtifacts(inputFiles, artifactExpander,
```
coeuvre, line 226: The returned collection of `SpawnInputExpander#getInputMapping` was expected to not contain tree artifacts. but now it is changed. Can you add a comment to the function's javadoc?

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
```diff
@@ -586,7 +586,6 @@ private static Spawn createCoveragePostProcessingSpawn(
             .addTransitive(action.getInputs())
             .addAll(expandedCoverageDir)
             .add(action.getCollectCoverageScript())
-            .add(action.getCoverageDirectoryTreeArtifact())
```
coeuvre, line 589: I don't get why this is removed. Can you explain a bit more?

src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
```diff
@@ -222,7 +223,8 @@ private static void addInputs(
...
+        true);
```
coeuvre, line 227: Please add an inline argument comment `/* keepEmptyTreeArtifacts= */` for boolean value. Same below.

coeuvre (changes requested):
I think this PR only covers the case of empty dir as input.

src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
fmeum: Good point, I augmented the comment. I tried to distinguish between "empty TreeArtifact" and "TreeArtifact expanding to the empty set with the given ArtifactExpander", as the coverage dir issue shows these are different concepts. Let me know if this is overly verbose.

src/main/java/com/google/devtools/build/lib/exec/SpawnInputExpander.java
fmeum: I went over the usages and added inline comments.

fmeum (commented):
Thanks for the review. I addressed the comments and also updated the commit/PR title to reflect the fact that it only deals with inputs.

src/main/java/com/google/devtools/build/lib/actions/ActionInputHelper.java
```diff
@@ -103,12 +103,13 @@ public PathFragment getExecPath() {
   }
 
   /**
-   * Expands middleman artifacts in a sequence of {@link ActionInput}s.
+   * Expands middleman and tree artifacts in a sequence of {@link ActionInput}s.
```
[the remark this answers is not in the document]
fmeum: Done.

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
coeuvre: I see. `SpawnInputExpander` can only expand inputs which are existing prior to action execution. This post-processing spawn is collecting inputs generated by previous spawn which are not visible to `SpawnInputExpander` so it is always expand to nothing -- that's the reason why we manually expand tree artifact here by reading directly from metadata provider.

I think it's safe to remove this line here. but @oquenchil can you confirm since you originally wrote this?

coeuvre (approved):
Overall looks good to me, but I will wait for @oquenchil to confirm the change of coverage before importing. Thanks!

fmeum:
@coeuvre I am investigating an alternative approach that may be overall more elegant, especially when it comes to also fixing #6393. Could you wait for a bit before importing? I will ping you on an alternative PR or post here if the other approach doesn't work.

fmeum:
@coeuvre I pushed a small commit to fix an issue with `--experimental_sibling_repository_layout` and added test coverage for this flag.

The alternative approach I pursued in https://github.com/bazelbuild/bazel/pull/15371 - expanding empty `TreeArtifact`s to a synthetic `TreeFileArtifact` to leverage the existing logic to create parent directories of input & output files - ended up not being as promising as expected. It requires too many changes to other users of `SpawnInputExpander` and would probably only be worth it if Bazel had many more affected strategies than just sandboxed and remote. The current PR is just my best attempt, feel free to import once @oquenchil has given the okay.

fmeum:
I resolved merge conflicts and used the opportunity to squash the commits.

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
oquenchil: Yes, it looks safe to me.

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15280, opened 2022-04-19
Fix checking remote cache for omitted files in buildevent file

exoson:
Enabling both `--noremote_upload_local_results` and
`--incompatible_remote_build_event_upload_respect_no_cache` caused
ByteStreamBuildEventArtifactuploader not to check if some files already
existed remotely because local files were not to be uploaded. When using
remote execution some of these files might exist remotely, so we want to
check remote cache for those files even if we would not upload them if
missing. The test case included depicts this failure case.


coeuvre (approved):
Thanks!

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15295, opened 2022-04-19
[Step 2] Update non-third party change for jackson2

ThomasCJY:
Follow up for https://github.com/bazelbuild/bazel/pull/15215 and Step two for https://github.com/bazelbuild/bazel/pull/15176


ThomasCJY:
@coeuvre @meteorcloudy ^ ptal

PR #15297, opened 2022-04-19
Remove unused windows cc toolchain feature

keith:
This isn't referenced, and it sounds like this isn't generally supported
on Windows.


oquenchil (approved):
LGTM

PR #15300, opened 2022-04-20
main: do not declare `environ` on Windows

compnerd:
When building with `/MD` on Windows, `environ` is a macro expanding to
`__p_environ` which has dll-export storage.  However, the redeclaration
does not annotate the DLL storage causing a mismatch warning
(`-Winconsistent-dllstorage`) which will be treated as an error.  Let
the declaration from the system headers be the declaration required.

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -80,7 +80,9 @@
 
 using blaze_util::GetLastErrorString;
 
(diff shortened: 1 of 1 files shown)


meteorcloudy (approved):

meteorcloudy:
@sgowroji Ready to be merged ;)

Issue #14533, opened 2022-01-09
`toolchain_vanilla` works with locally defined jdk 11, but `toolchain_java11` breaks the jdk compiler


james-deee:
### Description of the problem / feature request:

Building in Java 11 works using `toolchain_vanilla`, but changing to `toolchain_java11` does not compile due to a JDK bug (that should be fixed). Does the toolchain_java11 not use a locally defined `host_javabase` and `javabase`?

I assume that the vanilla one is using the local, but does that change to remote if i use toolchain_java11? Is there any way to change this so that I can define the jdk to use in that toolchain?

We define things like this in our .bazelrc:
```
build --define=ABSOLUTE_JAVABASE=//tools:jdk11
build --javabase=//tools:jdk11
build --host_javabase=//tools:jdk11
build --java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
```

The `jdk11` we use is this (via an alias)
```
http_archive(
    name = "jdk11_linux",
    build_file_content = "java_runtime(name = 'runtime', srcs =  glob(['**']), visibility = ['//visibility:public'])",
    sha256 = "3b1c0c34be4c894e64135a454f2d5aaa4bd10aea04ec2fa0c0efe6bb26528e30",
    strip_prefix = "jdk-11.0.13+8",
    urls = ["https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jdk_x64_linux_hotspot_11.0.13_8.tar.gz"],
)
```

### Feature requests: what underlying problem are you trying to solve with this feature?

Get our build to work.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I think that it might have to do with our own giant code base, so it's not easily reproducible that I've found. 

### What operating system are you running Bazel on?

Linux - Ubuntu 21.04

### What's the output of `bazel info release`?

```
release 1.2.1
```

###  Have you found anything relevant by searching the web?

Yes, it looks like the bug we are hitting is a bug in the JDK: https://bugs.openjdk.java.net/browse/JDK-8210483

### Any other information, logs, or outputs that you want to share?
```
[...]

PR #15302, opened 2022-04-20
Update to latest available JDK 11 update releases

cushon:
The new binaries are:

```
e064b61d93304012351242bf0823c6a2e41d9e28add7ea7f05378b7243d34247 https://cdn.azul.com/zulu/bin/zulu11.56.19-ca-jdk11.0.15-linux_x64.tar.gz
fc7c41a0005180d4ca471c90d01e049469e0614cf774566d4cf383caa29d1a97 https://cdn.azul.com/zulu-embedded/bin/zulu11.56.19-ca-jdk11.0.15-linux_aarch64.tar.gz
9750e11721282a9afd18a07743f19c699b2b71ce20d02f3f0a906088b9ae6d9a https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.14.1+1/OpenJDK11U-jdk_ppc64le_linux_hotspot_11.0.14.1_1.tar.gz
79a27a4dc23dff38a5c21e5ba9b7efcf0aa5e14ace1a3b19bec53e255c487521 https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.14.1+1/OpenJDK11U-jdk_s390x_linux_hotspot_11.0.14.1_1.tar.gz
2614e5c5de8e989d4d81759de4c333aa5b867b17ab9ee78754309ba65c7f6f55 https://cdn.azul.com/zulu/bin/zulu11.56.19-ca-jdk11.0.15-macosx_x64.tar.gz
6bb0d2c6e8a29dcd9c577bbb2986352ba12481a9549ac2c0bcfd00ed60e538d2 https://cdn.azul.com/zulu/bin/zulu11.56.19-ca-jdk11.0.15-macosx_aarch64.tar.gz
a106c77389a63b6bd963a087d5f01171bd32aa3ee7377ecef87531390dcb9050 https://cdn.azul.com/zulu/bin/zulu11.56.19-ca-jdk11.0.15-win_x64.zip
```

This is a step towards https://github.com/bazelbuild/bazel/issues/14533

PR #15303, opened 2022-04-20
Remove custom jdk_compiler and java_compiler jars

cushon:
The javac changes that Bazel depends on are now available in the latest JDK 11 update release.

https://github.com/bazelbuild/bazel/issues/14533


cushon:
This depends on https://github.com/bazelbuild/bazel/pull/15302, which should be merged first

cushon:
Also I'm not sure if removing the `:jdk_compiler_jar` etc. targets from the Java tools repo is acceptable because users could be referring to those labels directly. We could also remove the uses of them in the default toolchain config, but leave the artifacts around for now, and delete them later.

comius (approved):

Issue #14602, opened 2022-01-19
`--use_top_level_targets_for_symlinks` doesn't apply through `alias()` targets


keith:
With this BUILD file:

```bzl
load("@build_bazel_rules_apple//apple:macos.bzl", "macos_command_line_application")

alias(
    name = "alias",
    actual = "cli",
)

cc_library(
    name = "lib",
    srcs = ["main.c"],
)

macos_command_line_application(
    name = "cli",
    minimum_os_version = "10.10",
    deps = ["lib"],
)
```

Building the CI directly with `--use_top_level_targets_for_symlinks` correctly results in a top level symlink:

```
% bazel build cli  --use_top_level_targets_for_symlinks
INFO: Analyzed target //:cli (23 packages loaded, 387 targets configured).
INFO: Found 1 target...
Target //:cli up-to-date:
  bazel-bin/cli
INFO: Elapsed time: 2.727s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

But building the alias target still uses the transition path:

```
% bazel build alias --use_top_level_targets_for_symlinks
INFO: Analyzed target //:alias (0 packages loaded, 1 target configured).
INFO: Found 1 target...
Target //:cli up-to-date:
  bazel-out/applebin_macos-darwin_x86_64-fastbuild-ST-45701c223259/bin/cli
INFO: Elapsed time: 0.218s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

repro project: [aliasrepro.zip](https://github.com/bazelbuild/bazel/files/7901022/aliasrepro.zip)

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 5.0.0 (not a regression, also repros on 4.2.2)

joeleba:
Reassigning to Janak who added `--use_top_level_targets_for_symlinks` for triage. Thanks!

janakdr:
Looks like this was actually added by the configurability team (I just moved the flag around). Passing over.

sdtwigg:
This seems to be that the logic in //src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java is not sophisticated enough to consider 'look-through' aliases when doing the filtering.

[...]

PR #15316, opened 2022-04-22
Fix --use_top_level_targets_for_symlinks with aliases

keith:
Fixes https://github.com/bazelbuild/bazel/issues/14602

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
+++ b/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
@@ -732,6 +732,7 @@ public class ExecutionTool {
     Set<BuildConfigurationValue> targetConfigurations =
         buildRequestOptions.useTopLevelTargetsForSymlinks()
             ? analysisResult.getTargetsToBuild().stream()
+                .map(ConfiguredTarget::getActual)
                 .map(ConfiguredTarget::getConfigurationKey)
                 .filter(Objects::nonNull)
                 .distinct()


third_party/googleapis/google/devtools/build/v1/publish_build_event.proto
```diff
@@ -118,6 +118,14 @@ message PublishLifecycleEventRequest {
   // This should match the project used for the initial call to
   // PublishLifecycleEvent (containing a BuildEnqueued message).
   string project_id = 6;
+
+  // Whether to require a previously received matching parent lifecycle event
```
gregestren, line 122: What are the proto changes doing?

src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
```diff
@@ -725,6 +725,7 @@ private ImmutableList<ConvenienceSymlink> createConvenienceSymlinks(
     Set<BuildConfigurationValue> targetConfigurations =
         buildRequestOptions.useTopLevelTargetsForSymlinks()
             ? analysisResult.getTargetsToBuild().stream()
+                .map(ConfiguredTarget::getActual)
```
gregestren, line 728: What does `bazel build //:all` report in this context? Does it automatically fold the alias and actual into the same output report or does it keep them as separate entries?

gregestren (commented):
Approach looks good to me: just a few nitty questions.

third_party/googleapis/google/devtools/build/v1/publish_build_event.proto
keith: Sorry reverted!

src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
keith: It looks like with and without my change the behavior when building both the alias and the target (from my example in the issue) is the same:

```
% bazel-dev build :all --config=ora --use_top_level_targets_for_symlinks
INFO: Analyzed 2 targets (1 packages loaded, 4 targets configured).
INFO: Found 2 targets...
INFO: Elapsed time: 0.111s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
% bazelisk build :all --config=ora --use_top_level_targets_for_symlinks
INFO: Analyzed 2 targets (59 packages loaded, 976 targets configured).
INFO: Found 2 targets...
WARNING: cleared convenience symlink(s) bazel-bin, bazel-testlogs because their destinations would be ambiguous
INFO: Elapsed time: 7.108s, Critical Path: 0.38s
INFO: 8 processes: 4 internal, 3 darwin-sandbox, 1 local.
INFO: Build completed successfully, 8 total actions
```

keith: before:

```
% bazelisk build :cli :alias  --use_top_level_targets_for_symlinks --show_result=2
INFO: Analyzed 2 targets (59 packages loaded, 976 targets configured).
INFO: Found 2 targets...
WARNING: cleared convenience symlink(s) bazel-bin, bazel-testlogs because their destinations would be ambiguous
Target //:cli up-to-date:
  bazel-out/applebin_macos-darwin_arm64-fastbuild-ST-9e7e65650639/bin/cli
Target //:cli up-to-date:
  bazel-out/applebin_macos-darwin_arm64-fastbuild-ST-9e7e65650639/bin/cli
INFO: Elapsed time: 7.179s, Critical Path: 0.39s
INFO: 8 processes: 4 internal, 3 darwin-sandbox, 1 local.
INFO: Build completed successfully, 8 total actions
```

after:

```
% bazel-dev build :cli :alias  --use_top_level_targets_for_symlinks --show_result=2
INFO: Analyzed 2 targets (0 packages loaded, 0 targets configured).
INFO: Found 2 targets...
Target //:cli up-to-date:
  bazel-bin/cli
Target //:cli up-to-date:
  bazel-bin/cli
INFO: Elapsed time: 0.083s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

keith: (all vs specifying multiple targets is the same)

keith: So this is an improvement for this, as previously the alias vs target were different configurations it seems

gregestren: Interesting. The reporting is still a big confusing since it doesn't acknowledge why `//:cli` is reported twice. But that's orthogonal to what you're doing - just an observation.

Thanks for the explanation. I don't see any scenario where the alias's direct configuration is relevant for this.

gregestren (approved):

brentleyjones:
@bazel-io flag

Issue #10241, opened 2019-11-14
Make d8 dex builder compatible with persistent worker mode


gyias:
### Description of the problem / feature request:

Make d8 dex builder compatible with the persistent worker mode.

### Feature requests: what underlying problem are you trying to solve with this feature?

D8 dex builder doesn't support the local persistent worker mode, so users have to specify --nouse_workers_with_dexbuilder flag additionally. Even worse, failing to do so can lead to an obscure behavior that can be challenging to debug. Fix it so that the flag is not required.

tokudu:
@gyias i was chatting with @nkoroste, and it seems that https://github.com/bazelbuild/bazel/issues/10844 is no longer an issue. The only issue we are seeing is this one. 

I would be happy to contribute if you point me the right direction. Thanks!

jongerrish:
@sgjesse - I noticed you've made a number of commits to add `//src/tools/android/java/com/google/devtools/build/android/r8:DexBuilder`. We are very interested in a persistent D8 Builder and D8 Desugaring - what's the status of this work?

jongerrish:
I tried changing `//tools/android:dexbuilder` to use `//src/tools/android/java/com/google/devtools/build/android/r8:DexBuilder` rather than `:dexbuilder` but get build failures when building the tool from our workspace due to missing dependencies. (com.android.tools.r8.CompatDxSupport etc). I don't see this class in any recent versions of d8 (https://r8.googlesource.com/r8/+/refs/heads/master/src/main/java/com/android/tools/r8)

Any pointers as we'd love to try this out.

sgjesse:
The reason you do not find `CompatDxSupport` in the R8 repository any more is that it was moved into the bazel repository (https://github.com/bazelbuild/bazel/blob/a3a854dfc084e8e5115c524cc0030653f9df0778/src/tools/android/java/com/android/tools/r8/CompatDxSupport.java) but still in the `com.android.tools.r8` package to allow access to R8 package private APIs.

The work I have been doing is moving blaze desugaring to use the D8 desugaring engine, and that is still WIP. Supporting persistent workers can be a side effect of this work, as this will also move the bazel specific code out of the R8 repository so that bazel support only uses the public D8/R8 API.

jongerrish:
@sgjesse - Great to hear, persistent worker support for D8 is something we're expecting to help improve our build times.

[...]

PR #15321, opened 2022-04-22
Add multiplex worker support to dexbuilder

Bencodes:
Adding full support for multiplexed dexbuilder workers. This PR is a continuation of https://github.com/bazelbuild/bazel/pull/14623#issuecomment-1019852321 where we added basic worker support.

Related issue: https://github.com/bazelbuild/bazel/issues/10241

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/ExecutionRequirements.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/ExecutionRequirements.java
@@ -176,6 +176,9 @@ public class ExecutionRequirements {
   public static final ImmutableMap<String, String> WORKER_MODE_ENABLED =
       ImmutableMap.of(SUPPORTS_WORKERS, "1");
 
+  public static final ImmutableMap<String, String> WORKER_MULTIPLEX_MODE_ENABLED =
+      ImmutableMap.of(SUPPORTS_MULTIPLEX_WORKERS, "1");
+
   /**
    * Requires local execution without sandboxing for a spawn.
    *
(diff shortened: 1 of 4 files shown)


nkoroste:
Any updates on this?

ted-xie:
Oops, sorry for messing with the labels - I got some of my browser tabs mixed up. From a glance I don't see any major red flags; I'll discuss this internally.

ted-xie (changes requested):
Overall looks good - would you mind adding a test for this flag in src/test/shell/bazel/android/android_integration_test.sh?

ted-xie (approved):
Actually, disregard my previous comment. I'll take care of the integration test. Overall LGTM, I'll push this through the internal processes.

ahumesky:
Rather than adding a separate flag, there's `--modify_execution_info`:

https://bazel.build/reference/command-line-reference#flag--modify_execution_info

So I think we could do the same thing with:

`--modify_execution_info=DexBuilder=+supports-multiplex-workers`

I believe that `--modify_execution_info` was added after `--use_workers_with_dexbuilder` was added, so that's why we have that hardcoded flag. We could actually remove `--use_workers_with_dexbuilder` as well and replace it with `--modify_execution_info`.

`--modify_execution_info` is admittedly a little clunkier than a specially purposed flag, but on the other hand, it might be better than proliferating several more flags.

larsrc-google:
Very much agree that we shouldn't add another tool-specific flag.

I'm starting to regret the `--experimental_worker_multiplex` flag being a boolean instead of a include/exclude list.

Bencodes:
What about instead of flag proliferation we introduce an `expansion` CLI flag that sets all of the necessary `--modify_execution_info` requirements for Android multiplex workers?

Bencodes:
The CLI expansion solution would look something like this. https://github.com/bazelbuild/bazel/pull/15950

ahumesky:
Thanks everyone, let's go with https://github.com/bazelbuild/bazel/pull/15950

nkoroste:
should we close this in favor of https://github.com/bazelbuild/bazel/pull/15950 to avoid confusion

ahumesky:
Yes that sounds good

Issue #4057, opened 2017-11-08
cc_library warning: file has no symbols


jayconrod:
### Description of the problem / feature request / question:

On macOS, if a source file in a static `cc_library` doesn't define any global symbols, we see a warning like the one below:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: bazel-out/darwin_x86_64-fastbuild/bin/libempty.a(empty.o) has no symbols
```

This is a problem in rules_go. We use a tool in the Go toolchain (cgo) that generates some .c files that get compiled with `cc_library`. One of the generated files in each library frequently has no symbols, but it's hard to know ahead of time whether that's the case (the file is not empty).

It would be nice to suppress these warnings altogether. There may be a lot of libraries with cgo in a Go project, so our users may see a lot of warnings.

It looks like ranlib on macOS supports a flag `-no_warnings_for_no_symbols`. ar does not support this flag. The actual command that's emitting this warning looks like this: 

```
external/local_config_cc/wrapped_ar rcs bazel-out/darwin_x86_64-fastbuild/bin/libempty.a bazel-out/darwin_x86_64-fastbuild/bin/_objs/empty/empty.o
```

Maybe this flag can be added in [wrapped_ar.tpl](https://github.com/bazelbuild/bazel/blob/837e1b3d4859140d29aaa6bbab8fbb008e6d701e/tools/osx/crosstool/wrapped_ar.tpl)?

### If possible, provide a minimal example to reproduce the problem:

```
cc_library(
    name = "empty",
    srcs = ["empty.c"],
    linkstatic = 1,
)
```

empty.c is an empty file. Note that there are additional warnings if the entire library has no symbols. Those warnings don't affect rules_go, but they should probably be suppressed, too.

### Environment info
* Operating System: macOS Sierra 10.12.6

* Bazel version (output of `bazel info release`):  release 0.7.0

robbertvanginkel:
Was this fixed with https://github.com/bazelbuild/bazel/issues/5981 / https://github.com/bazelbuild/bazel/commit/81b59f05cb0853d27055edf77e62b619ff0aa4ba ?

cpsauer:
It looks to me like there's still an issue outstanding here somewhere. Not sure if it's a regression since #5981/https://github.com/bazelbuild/bazel/commit/81b59f05cb0853d27055edf77e62b619ff0aa4ba, or an edge case that's slipping through.

[...]

PR #15325, opened 2022-04-23
Filter libtool warning about table of contents

cpsauer:
The warning is unlikely to indicate a real issue, and Bazel already silences parallel warnings with `-no_warning_for_no_symbols`.

For more context, see https://github.com/bazelbuild/bazel/issues/4057

### DIFF
--- a/tools/objc/libtool.sh
+++ b/tools/objc/libtool.sh
@@ -32,7 +32,14 @@ if [ -z ${MY_LOCATION+x} ]; then
   fi
 fi
 
-WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
+function libtool() {
+  # Just invoke libtool via xcrunwrapper
+  "${MY_LOCATION}/xcrunwrapper.sh" libtool "$@" \
+  2> >(grep -v "the table of contents is empty (no object file members in the"`
+              `" library define global symbols)$" >&2)
+  # ^ Filtering a warning that's unlikely to indicate a real issue
+  # ...and not silencable via a flag.
+}
 
 if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
   echo "libtool_check_unique not found. Please file an issue at github.com/bazelbuild/bazel"
@@ -40,7 +47,7 @@ if [ ! -f "${MY_LOCATION}"/libtool_check_unique ] ; then
 elif "${MY_LOCATION}"/libtool_check_unique "$@"; then
   # If there are no duplicate .o basenames,
   # libtool can be invoked with the original arguments.
-  "${WRAPPER}" libtool "$@"
+  libtool "$@"
   exit
 fi
 
@@ -131,4 +138,4 @@ for arg in "$@"; do
 done
 
 printf '%s\n' "${ARGS[@]}" > "$TEMPDIR/processed.params"
-"${WRAPPER}" libtool "@$TEMPDIR/processed.params"
+libtool "@$TEMPDIR/processed.params"


cpsauer:
(Tagging @keith, since this came out of our discussion.)

tools/objc/libtool.sh
```diff
@@ -32,15 +32,22 @@ if [ -z ${MY_LOCATION+x} ]; then
   fi
 fi
 
-WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
+function libtool() {
+  # Just invoke libtool via xcrunwrapper
+  "${MY_LOCATION}/xcrunwrapper.sh" libtool "$@" \
+  2> >(grep -v "the table of contents is empty (no object file members in the"`
+              `" library define global symbols)$" >&2)
```
keith, line 39: does this grep not matching cause an unexpected failure?

tools/objc/libtool.sh
```diff
@@ -32,15 +32,22 @@ if [ -z ${MY_LOCATION+x} ]; then
   fi
 fi
 
-WRAPPER="${MY_LOCATION}/xcrunwrapper.sh"
+function libtool() {
```
keith, line 35: probably worth naming this something slightly different to avoid confusion that we're calling `libtool` directly, maybe `libtool_func` (or ideally something better)

tools/objc/libtool.sh
keith: I think you could also be less strict here if you wanted

tools/objc/libtool.sh
cpsauer: Would `invoke_libtool` be okay?

tools/objc/libtool.sh
cpsauer: When running, I'm seeing things succeed fine when the grep doesn't match. Does that resolve the concern?

cpsauer: (I kinda like the full thing to self-document the error, but can remove if you want.)

tools/objc/libtool.sh
keith: sounds good

tools/objc/libtool.sh
keith: Yes that resolves the concern, I was worried grep would still error. it looks like it's because of the `-v` that it doesn't?

cpsauer: Yep, I think so. I was concerned about that, too, when writing

cpsauer:
All checks are green and pass on the first round. I'll make  that change `libtool` -> `invoke_libtool`.

Then, all good to go, @keith? Thanks for being so lightning fast, and always a delight to work with.

keith (approved):
Yep I think this is safe to filter out! Thanks for the positivity! Someone from google will still have to review and land

oquenchil (approved):
LGTM

cpsauer:
Wahoo! Thanks @oquenchil x2

brentleyjones:
@bazel-io flag

cpsauer:
@brentleyjones, could I ask what the tags, like yours above, mean? I keep seeing them and am curious :)

brentleyjones:
@cpsauer I'm flagging them as desired to get into 5.2. If the release manager agrees they do the fork command to create an issue to track the inclusion. After that issue is made someone (preferably the author) can make a cherry-pick PR. I think now that the branch exists the author can just make the PR though.

Since I might not be able to make the PR immediately, I like to flag ones I think would be great candidates for the next release.

cpsauer:
Got it! Thanks for a great explanation, @brentleyjones. Honored to have made your cut :)

ckolli5:
@bazel-io fork 5.2.0

Issue #8195, opened 2019-04-29
incompatible_disallow_empty_glob: fail if a glob doesn't match anything


laurentlb:
The [`glob()`](https://docs.bazel.build/versions/master/be/functions.html#glob) function tends to be error-prone, because any typo in a path will silently return an empty list. This has been a source of confusion and bugs for many users.

`glob` has an optional Boolean argument `allow_empty`. For example, let's say your code looks like:
```python
glob([pattern_a, pattern_b], exclude = [pattern_c], allow_empty = False)
```

This code will fail if `pattern_a` or `pattern_b` doesn't match anything. It also fails if the whole function (after excluding `pattern_c`) doesn't match anything. We believe this behavior is safer in general.

In rare cases, this is intentional that a pattern doesn't match anything, e.g. when a source file is optional. When it happens, users should tell their intent explicitly by setting `allow_empty = True`.

* What's going to change? The default value of `allow_empty` used to be True. It will change to False.

* How to update your code? If a glob fails, use `allow_empty = True` to keep the old behavior.

In Bazel 8.0, the flag `--incompatible_disallow_empty_glob` will default to `true`. This means that the `allow_empty` argument defaults to `False` on all `glob` calls.

Iulia-Soimaru:
It seems like `glob` function has a bug when using `allow_empty = True`, the [documentation](https://docs.bazel.build/versions/2.2.0/be/functions.html#glob) for bazel v 2.2.0 has this api for the function:
```bazel
glob(include, exclude=[], exclude_directories=1, allow_empty=True)
```

when I try to apply it in the `filegroup` as
```bazel
filegroup(
    name = "icons_svg",
    srcs = glob(
        [
            "src/icons/*.svg",
        ],
        allow_empty = True,
    ),
)
```

and then use this filegroup in another bazel rule
```bazel
ts_node(
    name = "icons_gen",
    args = [
        "src/scripts/myScript.ts",
        # more ts-node cli arguments ...
        "$(rootpaths :icons_svg)",
    ],
    data = [
        ":icons_svg",
    ],
)
```
I am getting an error
```bash
in args attribute of nodejs_binary rule //:icons_gen: label '//:src/icons' in $(location) expression is not a declared prerequisite of this rule. Since this rule was created by the macro 'ts_node', the error might have been caused by the macro implementation
```

alandonovan:
Status: still awaiting cleanup and flag flip.

martis42:
[...]

PR #15330, opened 2022-04-24
Be explicit about alowing empty globs

limdor:

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -50,6 +50,7 @@ filegroup(
     srcs = glob(
         [".git/**"],
         exclude = [".git/**/*[*"],  # gitk creates temp files with []
+        allow_empty = True,
     ),
 )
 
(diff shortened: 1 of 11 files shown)


BUILD
```diff
@@ -50,6 +50,7 @@ filegroup(
     srcs = glob(
         [".git/**"],
         exclude = [".git/**/*[*"],  # gitk creates temp files with []
+        allow_empty = True,
```
aiuto, line 53: ???? Why do we even have this filegroup.
AFAICT, it is not used in Bazel.
Have you tried the PR without this change?

OTOH, I don't think your change is incorrect, presuming we do need it.
Perhaps it is fine to keep the change in place and to a targeted PR to just remove that target. Let's not mingle the different issues.

aiuto (approved):

aiuto:
I'll start a merge.

BUILD
limdor: I tried without the change, and it was failing if I remember right. In general I followed the approach of flipping the flag, and go test by test that was failing and only add the allow_empty where it was needed.
I did not try to remove the filegroup completely.

PR #15327, opened 2022-04-24
Flip disallow empty glob

limdor:
This PR is to flip the default value of `incompatible_disallow_empty_glob`. This flag is tracked in https://github.com/bazelbuild/bazel/issues/8195 and it is available since Bazel 0.29 from August 2019.
Once this is merged, users still can allow empty globs without being explicit setting `--incompatible_disallow_empty_glob=false`.

The motivation of the flip is https://github.com/bazel-contrib/SIG-rules-authors/issues/37 and the fact of not keeping an incompatible flag for 4 years without flipping it.

What needs to be done before flipping this:

- [x] Merge https://github.com/protocolbuffers/upb/pull/584
- [x] Update upb to a version that has the change in https://github.com/protocolbuffers/upb/pull/584
- [x] Merge https://github.com/protocolbuffers/upb/pull/745
- [x] Bring the changes from https://github.com/protocolbuffers/upb/pull/745 to the 21.x branch
- [x] Bring the changes from https://github.com/protocolbuffers/upb/commit/e5f26018368b11aab672e8e8bb76513f3620c579 to the 21.x branch (https://github.com/protocolbuffers/upb/pull/781)
- [x] Update upb to the latest commit of the 21.x branch (https://github.com/bazelbuild/bazel/pull/16343)
- [x] Merge https://github.com/bazelbuild/bazel/pull/15374
- [x] Merge https://github.com/bazelbuild/bazel/pull/15339
- [x] Merge https://github.com/bazelbuild/bazel/pull/15330
- [x] Merge https://github.com/bazelbuild/bazel/pull/16431
- [x] Merge https://github.com/bazelbuild/bazel/pull/16468
- [x] Get green checks for this PR
- [ ] Fix failures in downstream projects: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags
  - [ ] Fix empty globs from IntellIJ Plugin (https://github.com/bazelbuild/intellij)
    - [ ] Issue: https://github.com/bazelbuild/intellij/issues/4040
    - [ ] Android Studio Plugin
    - [ ] Android Studio Plugin Google
    - [ ] CLion Plugin
    - [ ] CLion Plugin Google
    - [ ] IntelliJ Plugin
    - [ ] IntelliJ Plugin Aspect
    - [ ] IntelliJ Plugin Aspect Google
    - [ ] IntelliJ Plugin Google
    - [ ] IntelliJ UE Plugin
    - [ ] IntelliJ UE Plugin Google
    - Opened PRs
[...]

PR #15339, opened 2022-04-25
Be explicit about alowing empty globs (second part)

limdor:

### DIFF
--- a/third_party/googleapis/BUILD.bazel
+++ b/third_party/googleapis/BUILD.bazel
@@ -24,7 +24,7 @@ exports_files(["LICENSE"])
 
 filegroup(
     name = "srcs",
-    srcs = glob(["**"]),
+    srcs = glob(["**"], allow_empty = True),
     visibility = ["@io_bazel//third_party:__pkg__"],
 )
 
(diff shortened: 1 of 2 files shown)


aiuto (approved):
I'll start the merge.

third_party/googleapis/BUILD.bazel
```diff
@@ -24,7 +24,7 @@ exports_files(["LICENSE"])
 
 filegroup(
     name = "srcs",
-    srcs = glob(["**"]),
+    srcs = glob(["**"], allow_empty = True),
```
limdor, line 27: @aiuto it does not look like this change got applied

aiuto: third_party has to be done in a separate PR from anything else.
Do things really fail without this?  That glob should never have anything empty.
If we can not do the change in third_party it would be nicer, because then we don't have to do a patch whenever we update in the future.

limdor: I'll double check.

limdor: @aiuto it seems that it was not needed in the end. I managed to get the CI green without this change https://github.com/bazelbuild/bazel/pull/15327
Because you worked on packaging, you might be interested in https://github.com/bazelbuild/bazel/pull/15327/files#diff-29ed1a1c7d731f96c7fe253b871566952b697712ea2d8d835f703b41dbbd13faR35

aiuto: Thanks.  That bit in tools/build_defs/pkg is irrelevant to anything. By Bazel 7.x it won't exist any more.

PR #15344, opened 2022-04-26
Fix StarlarkIndexable.getIndex implementations that return null.

katre:
They should return Starlark.NONE instead.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
@@ -233,7 +233,11 @@ public abstract class ConstraintCollection
 
   @Override
   public Object getIndex(StarlarkSemantics semantics, Object key) throws EvalException {
-    return get(convertKey(key));
+    Object result = get(convertKey(key));
+    if (result == null) {
+      result = Starlark.NONE;
+    }
+    return result;
   }
 
   @Override
--- a/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
@@ -236,7 +236,7 @@ public abstract class ConstraintCollection
     ConstraintSettingInfo constraintSettingInfo = convertKey(key);
     Object result = get(constraintSettingInfo);
     if (result == null) {
-      throw Starlark.errorf("Unknown constraint_setting '%s'", constraintSettingInfo);
+      return Starlark.NONE;
     }
     return result;
   }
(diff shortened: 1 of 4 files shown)


katre:
I was originally going to fix this in [`EvalUtils.index`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/net/starlark/java/eval/EvalUtils.java;bpv=1;bpt=1;l=447?ss=bazel&q=EvalUtils&gsn=index&gs=kythe%3A%2F%2Fgithub.com%2Fbazelbuild%2Fbazel%3Flang%3Djava%3Fpath%3Dnet.starlark.java.eval.EvalUtils%23f94ddbb499c163996efb03f37dff5e0dd2622f00b563f7e6457f45e5892604a6), but the comments there make me think that isn't a preferred change. Let me know if that would be easier.

tetromino (changes requested):
I would argue that for consistency with `dict` (and python), getIndex() should throw an EvalException if the key is missing.

sgowroji:
Hello @katre, Can you please share the update on the review comments. Thanks!

katre:
@sgowroji: Please note the timestamps: @tetromino commented at about 10pm ET last night, it is currently not even 8am ET now. I appreciate that you are keeping on top of PRs, but please give all contributors at least 12 hours to read and respond.

@tetromino: I will respond once I actually start my work day and can think through your response.

katre:
There are three cases here:
- ConstraintCollection and the test can just throw the exception.
- StarlarkToolchainContext is more complicated
  - If the toolchain type (the key) is known and a toolchain exists, return it
  - If the toolchain type is known and there is no toolchain, return `None`
  - If the toolchain type is unknown, throw an exception

The entire point of the work I am doing is to allow for [optional toolchains](https://github.com/bazelbuild/proposals/blob/main/designs/2022-01-21-optional-toolchains.md), and the design we settled on there was that `ctx.toolchains[/missing/optional/toolchain/type]` would return `None`.

Does this make sense, @tetromino ?

src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
```diff
@@ -233,9 +233,10 @@ public Sequence<ConstraintSettingInfo> constraintSettings() {
 
   @Override
   public Object getIndex(StarlarkSemantics semantics, Object key) throws EvalException {
-    Object result = get(convertKey(key));
+    ConstraintSettingInfo constraintSettingInfo = convertKey(key);
+    Object result = get(constraintSettingInfo);
     if (result == null) {
-      result = Starlark.NONE;
```
tetromino: I am really sorry for being unclear (and for being mistaken in my first comment) - the previous version of the code you hard (returning Starlark.NONE) was correct. For consistency with `dict`'s behavior, we don't want getIndex(x) to throw if containsKey(x) returns true.

tetromino (changes requested):

src/main/java/com/google/devtools/build/lib/analysis/platform/ConstraintCollection.java
katre: Oh, I see. Changed it back.

tetromino (approved):

PR #15345, opened 2022-04-26
Rewind on --remote_download_minimal upload FileNotFounds

illicitonion:
Currently, when building with `--remote_download_minimal`, if the remote
storage has evicted an output of a previous action, Bazel will throw a
FileNotFoundException for it.

This change detects that scenario, and re-throws it as a
LostInputsExecException, tracking enough metadata on the way to do so.

If rewinding is enabled for a build, this will cause rewinding to
automatically kick in.


brentleyjones:
Friendly ping on this @ericfelly.

coeuvre:
Closing as we are moving forward with lease service. https://github.com/bazelbuild/bazel/pull/16660

PR #15348, opened 2022-04-26
Record additional profiling information for remotely executed actions.

tjgq:


meisterT (commented):
Can you post an example screenshot of a profile with this new info?

tjgq:
> Can you post an example screenshot of a profile with this new info?

It's the bits under "execute remotely".

image

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
```diff
@@ -303,7 +313,34 @@ public SpawnResult exec(Spawn spawn, SpawnExecutionContext context)
     }
   }
 
-  /** conversion utility for protobuf Timestamp difference to java.time.Duration */
+  private static void profileAccounting(ExecutedActionMetadata executedActionMetadata) {
+    MillisSinceEpochToNanosConverter converter = BlazeClock.createMillisSinceEpochToNanosConverter();
+
+    logProfileTask(converter, executedActionMetadata.getQueuedTimestamp(),
+        executedActionMetadata.getWorkerStartTimestamp(),
+        REMOTE_QUEUE, "Remote execution queuing time");
```
meisterT, line 321: Consider using just `REMOTE_QUEUE.toString()` here (and similar below) instead of the full description. I think that the enum names are still easy to read for humans and much shorter, so you don't see abbreviations like `Remote execu...` like you do now.

coeuvre: Also consider removing `Remote ...` prefix as they are all under `execute remotely` span.

coeuvre (approved):

src/main/java/com/google/devtools/build/lib/remote/RemoteSpawnRunner.java
tjgq: Done.

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15349, opened 2022-04-26
Remove 'experimental' from inmemory jdeps/dotd files flags

keith:


coeuvre:
Thank for the PR. But I am not confident enough to remove `experimental` prefix:
1. These flags in bazel are mainly for build without bytes. As you might know, there are still many existing issues around it. I want to make sure most build without bytes issues are addressed before we start to thinking remove `experimental` prefix for related flags.
2. There is effort to reimplement c++/java rules in starlark. I am not sure whether these flags are relevant anymore in that case. cc @comius
3. These flags are also used internally where I don't have much experience with (yet). cc @alexjski

So I can't approve this PR at this moment but I will keep it open for future discussion.

PR #15329, opened 2022-04-24
Fix downloading remote execution output files inside output dirs.

jmillikin:
Adds a check to prevent creating multiple download futures for output files that are children of output directories.

Fixes #15328

PR #15356, opened 2022-04-27
Remote: Support the case where output files are under an output direc…

coeuvre:
…tory and are omitted from ActionResult.

- Handle another case when outputs are under output directory.
- Fix unit test and one more.
- Add integration test
- Update RemoteWorker to handle this case.

Supplement for #15329.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
@@ -910,6 +910,56 @@ public class RemoteExecutionService {
         files.buildOrThrow(), symlinks.buildOrThrow(), directories.buildOrThrow());
   }
 
+  /** Check that all mandatory outputs are created. */
+  private void checkOutputs(RemoteAction action, ActionResultMetadata metadata) throws IOException {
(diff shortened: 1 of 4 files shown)


jmillikin:
I might be misunderstanding, but according to the Remote Execution API protobuf at https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto, the `ActionResult` you describe is an error. If a command has an output file under an output directory, the executor should return an `ActionResult` containing entries for both.

```
  // The output files of the action. For each output file requested in the
  // `output_files` or `output_paths` field of the Action, if the corresponding
  // file existed after the action completed, a single entry will be present
  // either in this field, or the `output_file_symlinks` field if the file was
  // a symbolic link to another file (`output_symlinks` field after v2.1).
  // [...]
  repeated OutputFile output_files = 2;
```

coeuvre:
Technically, yes, executor should include declared output files inside the `ActionResult`. But there are implementations omit them when I was testing your PR. I am not sure whether they omit them on purpose or is just a bug. I think it wouldn't hurt if Bazel can detect this and continue the build.

coeuvre:
Closing as this PR is out of date and I don't have the workload to move it forward.

PR #15357, opened 2022-04-27
[WIP] Update ToolchainResolutionFunction to handle optional toolchain types

katre:

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ToolchainResolutionFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ToolchainResolutionFunction.java
@@ -161,6 +161,10 @@ public class ToolchainResolutionFunction implements SkyFunction {
       return new AutoValue_ToolchainResolutionFunction_ToolchainType(
           toolchainTypeRequirement, toolchainTypeInfo);
     }
+
+    public boolean mandatory() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/skyframe/ToolchainResolutionFunction.java
```diff
@@ -428,44 +432,40 @@ private static void determineToolchainImplementations(
...
 
-    if (!missingToolchains.isEmpty()) {
-      throw new UnresolvedToolchainsException(missingToolchains);
+    // Verify that all mandatory toolchain types have a toolchain.
+    if (!missingMandatoryToolchains.isEmpty()) {
+      throw new UnresolvedToolchainsException(missingMandatoryToolchains);
     }
 
     if (valuesMissing) {
       throw new ValueMissingException();
```
gregestren, line 462: Don't remember: does this just signify Skyframe restart?

katre: Yes, MissingValueException is caught at the top level, which then returns a null to signify a skyframe restart. It saves having to add in handling for returning null through all of the levels of this code.

gregestren (approved):

Issue #13661, opened 2021-07-09
Mobile install fails to load native libraries provided by AARs


Bencodes:
### Description of the problem / feature request:

Mobile install fails to load native libraries embedded in the incremental APK while using mobile-install.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Use `mobile-install` to build and install an Android app that includes an AARs containing native libraries. The application will crash at runtime when you load the library using `System#loadLibrary`.

```java
// DemoApplication.java

package cm.ben.android.bazel.bug;

import android.app.Application;

public class DemoApplication extends Application {

    @Override
    public void onCreate() {
        super.onCreate();
        System.loadLibrary("bugsnag-ndk");
    }
}
```
```starlark
#### BUILD

load("@build_bazel_rules_android//android:rules.bzl", "android_binary")

android_binary(
    name = "app",
    custom_package = "cm.ben.android.bazel.bug",
    manifest = "AndroidManifest.xml",
    srcs = ["DemoApplication.java"],
    deps = [
        "@maven//:com_bugsnag_bugsnag_android",
        "@maven//:com_bugsnag_bugsnag_android_core",
        "@maven//:com_bugsnag_bugsnag_plugin_android_anr",
        "@maven//:com_bugsnag_bugsnag_plugin_android_ndk",
    ],
)


#### WORKSPACE

workspace(name = "bazel_mobile_install_bug")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

## JVM External

_RULES_JVM_EXTERNAL_VERSION = "4.1"
_RULES_JVM_EXTERNAL_SHA = "f36441aa876c4f6427bfb2d1f2d723b48e9d930b62662bf723ddfb8fc80f0140"

http_archive(
    name = "rules_jvm_external",
    sha256 = _RULES_JVM_EXTERNAL_SHA,
    strip_prefix = "rules_jvm_external-{}".format(_RULES_JVM_EXTERNAL_VERSION),
    urls = [
        "https://github.com/bazelbuild/rules_jvm_external/archive/{}.zip".format(_RULES_JVM_EXTERNAL_VERSION),
    ],
)

load("@rules_jvm_external//:defs.bzl", "maven_install")

maven_install(
    artifacts = [
        "com.bugsnag:bugsnag-android:4.22.2",
        "com.bugsnag:bugsnag-android-core:4.22.2",
[...]

PR #14965, opened 2022-03-04
Traverse parent DexClassLoaders for native libraries in DelegateClassLoader

Bencodes:
Fixes https://github.com/bazelbuild/bazel/issues/13661

PR #15361, opened 2022-04-27
Fix Android native libs loading on apps deployed with 'mobile-install'

oliviernotteghem:
Fixes https://github.com/bazelbuild/bazel/issues/13661 by specifying the full list of (parent) native libs location in Bazel's IncrementalClassLoader.

StubApplication (used by Bazel 'mobile-install') fails to load native library defined in incremental APK. Issue is fixed by specifying the full set of native libs pathes (obtained via reflection from parent classloader) directly into IncrementalClassLoader, so that classloader doesn't have to iterate through its parent chain to find the native library. Iterating through parent (which is what https://github.com/bazelbuild/bazel/pull/14965 does - thank you to @Bencodes for this and the inspiration for the fix  works in most cases, however it doesn't handled cases where **native** libs needs to load other native lids. In this case, dlopen() will fail since it can only look into IncrementalClassLoader() native libs path, and not parents.

This will crashes for instances with Android applications using flipper when they're deployed via 'mobile-install', in addition to any app that relies on native libs deployed that way.

### DIFF
--- a/src/tools/android/java/com/google/devtools/build/android/incrementaldeployment/IncrementalClassLoader.java
+++ b/src/tools/android/java/com/google/devtools/build/android/incrementaldeployment/IncrementalClassLoader.java
@@ -18,6 +18,7 @@ import android.util.Log;
 import dalvik.system.BaseDexClassLoader;
 import java.io.File;
 import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


Bencodes:
cc: @ahumesky @ted-xie

Bencodes:
cc: @ahumesky @ted-xie

ted-xie:
@oliviernotteghem I'm going to close this PR, since I think the new Starlark mobile-install in rules_android should address this (really, the Studio Deployer should handle this more elegantly than native mobile-install did).

oliviernotteghem:
I can revert the fix from our repo and see if devs see regression

PR #15374, opened 2022-04-29
Make AndroidTests pass with incompatible_disallow_empty_glob

limdor:

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_ndk_misc_libraries_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_ndk_misc_libraries_template.txt
@@ -1,6 +1,6 @@
 cc_library(
     name = "cpufeatures",
-    srcs = glob(["ndk/sources/android/cpufeatures/*.c"]),
-    hdrs = glob(["ndk/sources/android/cpufeatures/*.h"]),
+    srcs = glob(["ndk/sources/android/cpufeatures/*.c"], allow_empty = True),
+    hdrs = glob(["ndk/sources/android/cpufeatures/*.h"], allow_empty = True),
     linkopts = ["-ldl"],
 )
--- a/src/test/java/com/google/devtools/build/lib/analysis/mock/BazelAnalysisMock.java
+++ b/src/test/java/com/google/devtools/build/lib/analysis/mock/BazelAnalysisMock.java
@@ -186,7 +186,7 @@ public final class BazelAnalysisMock extends AnalysisMock {
         "java_host_runtime_alias(name = 'current_host_java_runtime')",
         "filegroup(name='langtools', srcs=['jdk/lib/tools.jar'])",
         "filegroup(name='bootclasspath', srcs=['jdk/jre/lib/rt.jar'])",
-        "filegroup(name='extdir', srcs=glob(['jdk/jre/lib/ext/*']))",
+        "filegroup(name='extdir', srcs=glob(['jdk/jre/lib/ext/*'], allow_empty = True))",
         "filegroup(name='java', srcs = ['jdk/jre/bin/java'])",
         "filegroup(name='JacocoCoverage', srcs = ['JacocoCoverage_deploy.jar'])",
         "exports_files([",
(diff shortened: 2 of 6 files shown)


ahumesky:
So were each of these globs actually failing with `--incompatible_disallow_empty_glob`? Checking a few of these for example, the globs in [android_ndk_misc_libraries_template.txt](https://github.com/bazelbuild/bazel/pull/15374/files#diff-55bc0c199c17906c675fe132f08e72e163c9fa20aa94b378e9aa154856b22ae9), the cpufeatures files are still in the NDK, so those globs should match something

limdor:
How I came up with these globs and not others?
These allow_empty True were needed to flip the flag https://github.com/bazelbuild/bazel/pull/15327
The same for https://github.com/bazelbuild/bazel/pull/15339 and https://github.com/bazelbuild/bazel/pull/15330
I splitted in 3 to make it easier but I can also put them together or split them per folder.
Just let me know what would be the preference.

Once these 3 are merged, the flag still cannot be flipped because of https://github.com/protocolbuffers/upb/pull/584

ahumesky:
So adding `allow_empty = True` would move us closer to enabling the flag, but it would seem to be better to investigate why each of these globs is empty. And again in the example of cpufeatures, the NDK still has for example `sources/android/cpufeatures/cpu-features.c` so that glob shouldn't be returning nothing -- there might actually be something wrong with the glob or something else.

Is there someone on the bazel team who you're working with to drive this effort?

At minimum it would be good to create a separate issue to attach as a TODO to each glob that has `allow_empty = True` added to it so that we know why the code is the way that it is.

limdor:
> Is there someone on the bazel team who you're working with to drive this effort?

No, who would be the right person?

> And again in the example of cpufeatures, the NDK still has for example sources/android/cpufeatures/cpu-features.c so that glob shouldn't be returning nothing -- there might actually be something wrong with the glob or something else.

The problem is not with the example but when running the test //src/test/java/com/google/devtools/build/lib/bazel/rules/android:AndroidTests

> At minimum it would be good to create a separate issue to attach as a TODO to each glob that has allow_empty = True added to it so that we know why the code is the way that it is.

If you prefer I can do that

limdor:
@ahumesky I've updated the changes to include only the changes strictly needed for AndroidTests. If you think that this is not ok then probably the tests should be adapted instead.

src/test/java/com/google/devtools/build/lib/analysis/mock/BazelAnalysisMock.java
```diff
@@ -186,7 +186,7 @@ public void setupMockClient(MockToolsConfig config, List<String> workspaceConten
         "java_host_runtime_alias(name = 'current_host_java_runtime')",
         "filegroup(name='langtools', srcs=['jdk/lib/tools.jar'])",
         "filegroup(name='bootclasspath', srcs=['jdk/jre/lib/rt.jar'])",
-        "filegroup(name='extdir', srcs=glob(['jdk/jre/lib/ext/*']))",
```
limdor, line 189: Just one comment about this. It does not look like android specific but it is needed also to make the android tests work. Maybe someone else apart from the android team should look at it.

limdor:
@ahumesky @ted-xie friendly reminder.
Could you give me an update on what would be needed in your opinion in order to move forward? Thanks

limdor:
Gentle reminder

src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_ndk_misc_libraries_template.txt
```diff
@@ -1,6 +1,6 @@
 cc_library(
     name = "cpufeatures",
-    srcs = glob(["ndk/sources/android/cpufeatures/*.c"]),
-    hdrs = glob(["ndk/sources/android/cpufeatures/*.h"]),
+    srcs = glob(["ndk/sources/android/cpufeatures/*.c"], allow_empty = True),
```
aiuto, line 3: It doesn't make sense to me that these could ever result in failed globs in real life, but maybe they do in our tests.
OTOH, I don't think it is a problem that we hide the glob failure. We will simply push the error from glob time until link time, but the build will still fail, so all is good.

The same question applies everywhere.

aiuto (commented):
LGTM, but I'm no nothing about these rules.

ted-xie:
It also seems strange to me that any of these globs would return empty. I would expect downstream failures if various NDK features are missing source files. I'm OK with pushing this in upstream, but we should all be aware that something else entirely may be broken that's causing the source files not to show up.

PR #15378, opened 2022-05-01
Log one file per command if --experimental_workspace_rules_log_file is a directory

gonzojive:
This is helpful for keeping track of all workspace events across multiple runs of bazel for the https://github.com/gonzojive/bazelgopackagesdriver project. Without this behavior, every bazel invocation needs to be customized to specify the path of the output.

I would probably prefer to have a different flag rather than detecting whether the output location is a directory. Please let me know if you agree.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/debug/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/debug/BUILD
@@ -18,6 +18,7 @@ filegroup(
 proto_library(
     name = "workspace_log_proto",
     srcs = ["workspace_log.proto"],
+    visibility = ["//visibility:public"],
 )
 
 java_proto_library(
@@ -56,6 +57,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/lib/util:abrupt_exit_exception",
         "//src/main/java/com/google/devtools/build/lib/util:detailed_exit_code",
         "//src/main/java/com/google/devtools/build/lib/util/io",
+        "//src/main/java/com/google/devtools/build/lib/vfs",
         "//src/main/java/com/google/devtools/build/lib/vfs:pathfragment",
         "//src/main/java/com/google/devtools/common/options",
         "//src/main/protobuf:failure_details_java_proto",
--- a/src/main/java/com/google/devtools/build/lib/bazel/debug/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/debug/BUILD
@@ -54,6 +54,7 @@ java_library(
         ":workspace-rule-event",
         "//src/main/java/com/google/devtools/build/lib:runtime",
         "//src/main/java/com/google/devtools/build/lib/events",
+        "//src/main/java/com/google/devtools/build/lib/util",
         "//src/main/java/com/google/devtools/build/lib/util:abrupt_exit_exception",
         "//src/main/java/com/google/devtools/build/lib/util:detailed_exit_code",
         "//src/main/java/com/google/devtools/build/lib/util/io",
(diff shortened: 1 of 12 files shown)


meteorcloudy:
My feeling is adding a `--experimental_workspace_rules_log_file` is better, but on the other hand, we do want as few flags as possible in Bazel. Can you figure out a way to work this around instead of changing Bazel?

gonzojive:
> My feeling is adding a `--experimental_workspace_rules_log_file` is better, but on the other hand, we do want as few flags as possible in Bazel. Can you figure out a way to work this around instead of changing Bazel?

1) I assume you mean `--experimental_workspace_rules_log_dir` (distinct from the existing `experimental_workspace_rules_log_file` flag).

2) What about using variable substitution to allow a value like "/tmp/workspace_log_${command_id}.log" for  the `--experimental_workspace_rules_log_dir` flag? Is there a standard string templating syntax used in other parts of the project that could be used here for consistency? That would alleviate the need for the flag and provide the functionality I desire.

I can work around this... but the proposal is a 5 line change vs the workaround, which will require writing a wrapper for the bazel binary.

meteorcloudy:
> I assume you mean --experimental_workspace_rules_log_dir (distinct from the existing experimental_workspace_rules_log_file flag).

Ah, yes, sorry for the mistake, fixed.

> What about using variable substitution to allow a value like "/tmp/workspace_log_${command_id}.log" for the --experimental_workspace_rules_log_dir flag? Is there a standard string templating syntax used in other parts of the project that could be used here for consistency? That would alleviate the need for the flag and provide the functionality I desire.

I don't think we currently support such syntax.

> I can work around this... but the proposal is a 5 line change vs the workaround, which will require writing a wrapper for the bazel binary.

This isn't just a 5 line change. It's the Bazel team who will need to maintain or clean up this in the future. Also if we allow using a dir for the workspace debug log, what about flags for other log files, it's hard to justify why we need to special case this one but not others.

gonzojive:
> > I assume you mean --experimental_workspace_rules_log_dir (distinct from the existing experimental_workspace_rules_log_file flag).
> 
> Ah, yes, sorry for the mistake, fixed.
> 
> > What about using variable substitution to allow a value like "/tmp/workspace_log_${command_id}.log" for the --experimental_workspace_rules_log_dir flag? Is there a standard string templating syntax used in other parts of the project that could be used here for consistency? That would alleviate the need for the flag and provide the functionality I desire.
> 
> I don't think we currently support such syntax.

I updated the pull request to support `{{command_id}}` substitution in all of these flags. The idea behind using `{{x}}` instead of `${x}`  is to make it sufficiently different from bash substitution that it's apparent something else is going on to an unfamiliar user.

> 
> > I can work around this... but the proposal is a 5 line change vs the workaround, which will require writing a wrapper for the bazel binary.
> 
> This isn't just a 5 line change. It's the Bazel team who will need to maintain or clean up this in the future. Also if we allow using a dir for the workspace debug log, what about flags for other log files, it's hard to justify why we need to special case this one but not others.

I understand. I was hoping to find a solution with a minimal maintenance burden.

meteorcloudy:
@gonzojive Thanks for this PR, I understand it's a nice thing to have for you. But I checked this flag is an experimental feature that's currently unmaintained. I strongly prefer to not add more complexity to Bazel while there are simple work arounds.

gonzojive:
That's fine. I understand wanting to keep the codebase easy to understand. I'll work around this by wrapping the bazel binary a la bazelisk.

The PR affects non-experimental flags as well. The main complexity is the introduction of a templating syntax. The ability to put templated strings in configuration files and flags is pretty valuable in some cases. I am thinking of VS Code's configuration files, for instance - often the configuration values can include strings like ${workspaceDir}.

Issue #10091, opened 2019-10-23
Support running persistent workers remotely


tsiq-charliem:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

Looking for support for running persistent workers on remote hosts.

### Feature requests: what underlying problem are you trying to solve with this feature?

Currently, actions can be run with the `remote` strategy or the `worker` strategy, but we'd like a way to get the benefits of a persistent worker on a remote build. Without this, our local builds with persistent workers outperform remote builds.

### What operating system are you running Bazel on?

Ubuntu 16.04

### What's the output of `bazel info release`?

1.0.0

buchgr:
@tsiq-charliem there's the https://blog.bazel.build/2019/02/01/dynamic-spawn-scheduler.html that will get you the best of both.

We currently have no plans for to work on worker support for remote execution.

meisterT:
Closing as we don't plan to add persistent worker support for remote execution.

ulfjack:
I have a patch for this. It's pretty small, so it may be acceptable to check it in?

meisterT:
Interesting, how does that look like?

ulfjack:
Patch is here: https://github.com/ulfjack/bazel/tree/remote-persistent-worker

meisterT:
cc @coeuvre @larsrc-google

ulfjack:
Support for remote persistent workers is one of our most requested features and we've seen significant performance improvements in some real-world scenarios with proprietary codebases. I've rebased my change to HEAD, but I still need to add some tests.

bergsieker:
Your patch seems reasonable as far as adding the tool signature to the Platform. As far as I read it, it's just adding the hash of the tool paths, not trying to get a stronger signature like the digest of the binaries, right? It's possible that splitting into multiple cache keys, one for each referenced tool, might be desirable because it gives the server more scheduling flexibility in terms of prioritizing one tool over another, but I'm not sure how frequent multi-tool actions are so it may not make much difference in practice.

[...]

PR #15381, opened 2022-05-01
[worker] Move WorkerProtocolImpl into its own package

Yannic:
Splitting it out of the main worker package makes re-using the code
for other implementations for dispatching requests to workers (e.g.,
for remote persistent workers, #10091) easier.


Yannic:
@larsrc-google PTAL

Yannic:
Ping?

Yannic:
ping?

Yannic:
@wilwell can you help getting this merged please?

wilwell (approved):
LGTM. @larsrc-google please review these changes too.

larsrc-google:
Why do you need to split it into an entirely separate package, rather than just into a separate target?

Yannic:
> Why do you need to split it into an entirely separate package, rather than just into a separate target?

Having it in a dedicated package makes it clearer that this is its own unit. I prefer the package solution, but I'm not strictly against just making it its own target. Please let me know what you think

larsrc-google:
Now a separate target.

PR #15382, opened 2022-05-01
ijar: read Code attribute

Yannic:
JVMS 4.7.3: "If the method is either native or abstract, its method_info
structure must not have a Code attribute. Otherwise, its method_info structure
must have exactly one Code attribute."

As hack around this, we instead emit a (dummy) Code attribute that does not
depend on the actual implementation of the method. This should fix compatibility
with tools like https://spotbugs.github.io which seem to care about the Code
attribute of dependencies and currently fail with ijar enabled.

Since there are some performance concerns about emitting a code attribute at
all, we start with only reading the Code attribute in this PR and add the code
to emit the dummy attributes in a follow-up to make benchmarking easier.

Progress on https://github.com/bazelbuild/bazel/issues/13546

### DIFF
--- a/third_party/ijar/classfile.cc
+++ b/third_party/ijar/classfile.cc
@@ -454,6 +454,82 @@ struct HasAttrs {
   }
 };
 
+// See sec.4.7.3 of JVM spec.
+//
+// Code_attribute {
+//   u2 max_stack;
+//   u2 max_locals;
+//   u4 code_length;
+//   u1 code[code_length];
+//   u2 exception_table_length;
+//   {
+//     u2 start_pc;
+//     u2 end_pc;
+//     u2 handler_pc;
+//     u2 catch_type;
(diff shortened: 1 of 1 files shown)


Yannic:
@cushon PTAL

cushon:
> Since there are some performance concerns about emitting a code attribute at
all, we start with only reading the Code attribute in this PR and add the code
to emit the dummy attributes in a follow-up to make benchmarking easier.

I don't understand why would need to read code attributes.

If it ever supports writing them, I'd expect it to write a code attribute that did something like unconditionally throw for non-void methods, which doesn't require inspecting the original real code attribute.

cushon:
> This should fix compatibility with tools like https://spotbugs.github.io/ which seem to care about the Code attribute of dependencies and currently fail with ijar enabled.

Can you add context in #13546?

keertk:

PR #15389, opened 2022-05-02
Add tableOfContents to stardoc

keith:
This adds a new replacement in stardoc headers for a simple table of
contents that prints each different type of documented symbol, and a
link to each one that will be on the page. This is much easier than
having consumers enumerate all the symbols in a file to build this up
themselves.


keith:
@c-parsons bump

keith:
https://github.com/bazelbuild/stardoc/pull/165

PR #15393, opened 2022-05-03
Export `BAZEL_TEST=1` for `bazel test` executables

mattyclarkson:
[Catch2][catch2] has [learnt][pr2399] how to output JUnit to the `XML_OUTPUT_FILE`. However, the upstream developers
[felt][comment] that `XML_OUTPUT_FILE` is too generic to blanket enable JUnit output to the file as there are various
XML reporters for Catch2.

This patch enables exporting `BAZEL_TEST` environment variable so that Catch2 can enable JUnit output support
unconditionally in their build when both `BAZEL_TEST` _and_ `XML_OUTPUT_FILE` are available.

[catch2]: https://github.com/catchorg/Catch2
[pr2399]: https://github.com/catchorg/Catch2/pull/2399
[comment]: https://github.com/catchorg/Catch2/pull/2399#issuecomment-1092965791

### DIFF
--- a/site/en/reference/test-encyclopedia.md
+++ b/site/en/reference/test-encyclopedia.md
@@ -452,6 +452,11 @@ The initial environment block shall be composed as follows:
              class="external">JUnit test result schema</a>.</td>
     <td>optional</td>
   </tr>
+  <tr>
+    <td><code>BAZEL_TEST</code></td>
+    <td>Signifies test executable is being driven by <code>bazel test</code></td>
+    <td>required</td>
+  </tr>
 </table>
 
 The environment may contain additional entries. Tests should not depend on the
--- a/tools/test/test-setup.sh
+++ b/tools/test/test-setup.sh
@@ -37,6 +37,11 @@ function is_absolute {
 # root.
 EXEC_ROOT="$PWD"
 
+# Declare that the executable is running in a `bazel test` environment
+# This allows test frameworks to enable output to the unprefixed environment variable
+# For example, if `BAZEL_TEST` and `XML_OUTPUT_FILE` are defined, write JUnit output
+export BAZEL_TEST=1
+
 # Bazel sets some environment vars to relative paths to improve caching and
 # support remote execution, where the absolute path may not be known to Bazel.
 # Convert them to absolute paths here before running the actual test.


meteorcloudy:
It is possible to pass extra env var with  [--test_env](https://bazel.build/docs/user-manual#test-env), for example, `--test_env=BAZEL=1`. Can you please check if this works for you?

mattyclarkson:
> Can you please check if this works for you?

Well, yes, that would work.

However, do we expect anyone that uses {{catch2}} (or other test frameworks) to add {{--test_env}} to their {{.bazelrc}}. I would say that would be a UX/DX regression for usage of test frameworks. Unless we can propagate a test environment variable upon the usage of `@catch2//:catch2_main`, for example.

Bazel does [test framework specific stuff in the test runner](https://github.com/bazelbuild/bazel/blob/def221916a1c724bcfd41f62ff4ff4c8d3672792/tools/test/test-setup.sh#L107-L112):

```sh
# Tell googletest about Bazel sharding.
if [[ -n "${TEST_TOTAL_SHARDS+x}" ]] && ((TEST_TOTAL_SHARDS != 0)); then
  export GTEST_SHARD_INDEX="${TEST_SHARD_INDEX}"
  export GTEST_TOTAL_SHARDS="${TEST_TOTAL_SHARDS}"
fi
export GTEST_TMP_DIR="${TEST_TMPDIR}"

# TODO(ulfjack): Update Gunit to accept XML_OUTPUT_FILE and drop this env
# variable.
GUNIT_OUTPUT="xml:${XML_OUTPUT_FILE}"
```

Removing `GUNIT_OUTPUT` would mean that GTest has to accept a very generic `XML_OUTPUT_FILE`, so seems like introducing a `BAZEL=1` into the test runner would also allow GTest to check for `BAZEL`/`XML_OUTPUT_FILE` combo.

meteorcloudy:
> so seems like introducing a BAZEL=1 into the test runner would also allow GTest to check for BAZEL/XML_OUTPUT_FILE combo.

I see, seems like we should go with this option. But maybe choosing a more explicit name? Like `BAZEL_TEST_SETUP`?

mattyclarkson:
Sure. Maybe `BAZEL_TEST=1` as it's running under `bazel test`?

I'm not here for bikeshedding though and `BAZEL_TEST_SETUP` would be fine 😄 

Depending on the naming I can respin the patch and PR description to match.

meteorcloudy (approved):
Thank you!

meteorcloudy:
@sgowroji Please help import this PR once the tests pass ;)

PR #15402, opened 2022-05-04
Remove --action_env from RBE configuration.

tjgq:
Bazel builds fine without it.


coeuvre (approved):
I believe the `--action_env=PATH` was added because Bazel was using the process PATH for  remote action's env. But no evidence shows it is still doing that today. Let's remove it.

tjgq:
Update: it turns out that Bazel indeed propagates the host PATH to remote actions, therefore `--action_env` is necessary to avoid remote cache pollution. I believe flipping `--incompatible_strict_action_env` to true needs to happen before we can do this.

PR #15403, opened 2022-05-04
Add feature to produce serialized diagnostics files

### LINKED ISSUE #15191: FR: Add ability to generate clang serialized diagnostics files for CC builds

Allow (probably via a `feature`) the ability for C/C++/Obj-C files to generate clang serialized diagnostics files (`--serialized-diagnostics`) for each source file. The file should be handled similar to the `.d` file, and be a declared dependency.

I have an IDE integration (Xcode) use case where I want `--serialized-diagnostics source_file.dia` to be applied to each C/C++/Obj-C compilation, as the IDE consumes this file to report warnings, errors, notes, remarks and fix-its. These files need to be fetched from cache as well, so they need to be a declared output.

macOS

release 5.1.0

_No response_

_No response_

_No response_

I'm willing to contribute this change, if it will be accepted.

brentleyjones:
Using the `serialized_diagnostics_file` feature will add the `--serialized-diagnostics` flag to C/C++/Objective-C/Objective-C++ compiles, causing a declared`.dia` file output to be produced.

Closes #15191.


meteorcloudy:
@googlewalt Do you think you are the right person to review this?

@oquenchil Since this feature seems not only apply to objc rules, maybe you should also take a look?

brentleyjones:
Sure, I'll take a look today to see if I can use that.

brentleyjones:
From the looks of it, that will shift a small amount of code from `CppCompileActionBuilder.setOutputs()` to `CppCompileActionBuilder.setAdditionalOutputs()`.

Because of it being a `feature_flag`, we still need to know about the extension and whatnot for `addStringVariable()`. I'm not really sure there is much of a win in refactoring this, and honestly there is a higher chance of me messing things up when I'm pretty sure the change as is works. I think it can get better if we eventually move to the inject like design (https://github.com/bazelbuild/bazel/issues/15191#issuecomment-1103953386) but that is out of scope for this change, as I would like to get this into the 5.2 release as well.

oquenchil (approved):
LGTM

PR #15408, opened 2022-05-06
Unify URL/URLs parameter code across http_archive, http_file, http_jar

cpsauer:
Hi wonderful Bazel folks,

I--like probably everyone else--make heavy use of the (super handy) http repository rules.

While using them, I'd noticed that the docs for the URL attributes they have in common had started drifting. For example, http_file's URLs array doc was mostly same as http_archive's, except that it forgot to specify that the URLs were tried in the order they were specified. jar_file's were much less complete. Similarly, only http_file was missing the URL parameter, which seems a shame, since probably most people only want to supply one authoritative source most of the time. When I looked at the code, it became clear that this was because the implementations were copy-pasted, and being updated separately.

So I spent a few minutes cleaning things up and making these commonly used features a little more consistent, fixing the issues above. Hopefully that's welcome. 

I'd love your thoughtful consideration and review!

Thanks, wonderful Bazel folks, for all you do!

Cheers,
Chris
(ex-Googler)

P.S. I've given edit access with the "allow edits by maintainers" box. If there are things you'd like changed, feel free also to just make them directly!

### DIFF
--- a/tools/build_defs/repo/http.bzl
+++ b/tools/build_defs/repo/http.bzl
@@ -41,6 +41,38 @@ load(
 )
 
 # Shared between http_jar, http_file and http_archive.
+
+_URL_DOC = """A URL to a file that will be made available to Bazel.
+
+This must be a file, http or https URL. Redirections are followed.
+Authentication is not supported.
+
+More flexibility can be achieved by the urls parameter that allows
+to specify alternative URLs to fetch from."""
+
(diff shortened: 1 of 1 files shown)


Wyverald (approved):
Thanks, this looks great!

cpsauer:
Sweet! Thanks @Wyverald for the review and approval!

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

PR #15413, opened 2022-05-06
Include spawn metrics in the execution log.

tjgq:
The SpawnMetrics proto is intended to match the existing Java class of the
same name. Note that the SpawnExec.walltime field becomes redundant, but it
is kept to ensure backwards compatibility.

This change makes it easier to correlate the metrics for corresponding actions
in two different builds.


tjgq:
Put the inclusion of spawn metrics in the log behind a flag, per offline discussion.

coeuvre (approved):

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

PR #15416, opened 2022-05-06
Ensure Chunkers are always reset (closed).

benjaminp:
Chunkers hold open files and memory. Normally, they are closed when the input is fully read. However, we want to make sure resources are freed in error cases, too.


coeuvre (changes requested):
Can you please also add a test for the case where the chunker is not closed?

coeuvre (approved):
Thanks!

src/test/java/com/google/devtools/build/lib/remote/GrpcCacheClientTest.java
```diff
@@ -151,6 +154,51 @@ public void onError(Throwable t) {
...
+              public void onCompleted() {}
+
+              @Override
+              public void onError(Throwable t) {}
+            };
+          }
+        });
+    byte[] data = new byte[20];
+    Digest digest = DIGEST_UTIL.compute(data);
+    CountDownLatch latch = new CountDownLatch(1);
```
benjaminp, line 181: @coeuvre I see that when this PR landed in 5df79b6513c6dbb0fadf2e65fa95f37e2b7ab540, this part of the test was changed to have an `AtomicBoolean` `closed` flag instead of the latch I propose in this PR. The reason for the latch is that the `.close()` is called in a listener to the future returned by `uploadChunker`. Therefore, in theory, `.close()` could be called slightly after `.get()` on the future returns.

coeuvre: I see. Thanks for the explanation. Will change it back.

PR #15426, opened 2022-05-09
Prevent aspects from executing on incompatible targets

### LINKED ISSUE #15271: Aspects get falsely evaluated against incompatible targets
### Description of the bug:

It looks like aspects run on incompatible targets when they shouldn't.

I think the answer here is that the aspects should not run and instead just provide the `IncompatiblePlatformProvider` to signify that it's incompatible as well.

The absolute minimal example:
`example.bzl`:
```python
def _custom_rule_impl(ctx):
    pass

custom_rule = rule(
    implementation = _custom_rule_impl,
)

def _example_aspect_impl(target, ctx):
    print("Running aspect on " + str(target))
    return []

_example_aspect = aspect(
    implementation = _example_aspect_impl,
    attr_aspects = ["dep"],
)

def _basic_rule_with_aspect_impl(ctx):
    pass

basic_rule_with_aspect = rule(
    implementation = _basic_rule_with_aspect_impl,
    attrs = {
        "dep": attr.label(
            aspects = [_example_aspect],
        ),
    },
)
```
`BUILD`:
```python
load(":example.bzl", "basic_rule_with_aspect", "custom_rule")

custom_rule(
    name = "target1",
)

basic_rule_with_aspect(
    name = "target2",
    dep = ":target1",
)
```
When you build this target, you'll see that the `print()` statement in the aspect gets executed:
```console
$ bazel build //:target2
DEBUG: /home/pschrader/work/test_repo/example.bzl:9:10: Running aspect on <target //:target1>
ERROR: Target //:target2 is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //:target2
    //:target1   <-- target platform (@local_config_platform//:host) didn't satisfy constraint @platforms//:incompatible
INFO: Elapsed time: 0.070s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 2 targets configured)
```
This can lead to undesirable effects. Largely because the aspect evaluation can generate errors before bazel gets a chance to tell the user about the incompatible target being requested.

Ubuntu 18.04

release 5.1.0

N/A

```text
N/A
```

_No response_

_No response_

philsc:
This patch prevents aspect functions from being evaluated when their
associated targets are incompatible. Otherwise aspects can generate
errors that should never be printed at all. For example, an aspect
evaluating an incompatible target may generate errors about missing
providers. This is not the desired behaviour.

The implementation in this patch short-circuits the `AspectValue`
creation if the associated target is incompatible.

I had tried to add an `IncompatiblePlatformProvider` to the
corresponding `ConfiguredAspect` instance, but then Bazel
complained about having duplicate `IncompatiblePlatformProvider`
instances.

Fixes  #15271


philsc:
@gregestren , are you a good person to review this?

Happy to add more test cases too.

gregestren:
Sure, happy to review!

src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
```diff
@@ -186,6 +187,15 @@ public SkyValue compute(SkyKey skyKey, Environment env)
     ConfiguredTarget associatedTarget = state.initialValues.associatedTarget;
     Target target = state.initialValues.target;
 
+    if (associatedTarget.get(IncompatiblePlatformProvider.PROVIDER) != null) {
```
gregestren, line 190: Sanity check: this just means that we return an aspect that does nothing?

gregestren (commented):
It took me too long to get to this. I thought it was going to be a lot more code than it is. What a nice, concise surprise! 

It basically looks good to me, including the test coverage. I just have some sanity check questions to make sure I get all the implications.

Basic question: as I understand aspects can automatically propagate down a target's dependencies. If an aspect hits an `IncompatiblePlatformProvider` target, it gets short-circuited, which is great.

Will it continue to propagate down that target's dependencies? If they're not also `IncompatiblePlatformProvider` targets, would that do something strange?

I think your test coverage shows the opposite? How when the bottom-most dependency is incompatible, everything depending on it also gets skipped? 

What happens when that's reversed?

philsc:
> Will it continue to propagate down that target's dependencies? If they're not also IncompatiblePlatformProvider targets, would that do something strange?

That is a great question. I will add a test case for this and fix it if it behaves unexpectedly.

src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
philsc: That's my understanding, yes.

And more importantly, as far as I can tell, it doesn't trigger any dependent aspects from getting evaluated. Which should address the question you had I think. The one about what happens if an incompatible target depends on a compatible one. The aspect does _not_ get evaluated for the compatible dependencies.

src/test/shell/integration/target_compatible_with_test.sh
```diff
@@ -1082,4 +1082,164 @@ function test_aquery_incompatible_target() {
...
 }
 
+# Use aspects to interact with incompatible targets and validate the behaviour.
+function test_aspect_skipping() {
+  cat >> target_skipping/BUILD <<EOF
+load(":defs.bzl", "basic_rule", "rule_with_aspect")
+
+basic_rule(
+    name = "basic_universal_target",
+)
```
philsc, line 1092: I added this target here that is depended on by `basic_foo3_target`. The idea being that if `basic_foo3_target` is incompatible, then aspects should never reach `basic_universal_target`.

As far as I can tell, this works as expected:
* The `print()` statement in the aspect never gets executed for `basic_universal_target` if `basic_foo3_target` is incompatible.
* The `print()` statement for `basic_universal_target` _does_ get executed when `basic_foo3_target` is compatible.

The test now checks for the debug message on this target.

gregestren: That sounds great, thanks.

Would you mind comment-documenting basic_universal_target to clarify the point you made? That this is to cover the incompatible -> compatible dependency order?

And maybe a similar comment on the target that shows the opposite?

src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
gregestren: Oh, I honestly can't remember where aspect propagation is implemented (AspectFunction vs. ConfiguredTargetFunction vs. DependencyResolver vs ...).

But thanks to your test coverage we have good formalized behavior expectations, so my question doesn't matter!

Thanks for verifying.

gregestren (commented):
No other comments besides asking for a bit more line documentation in the test code.

src/test/shell/integration/target_compatible_with_test.sh
philsc: That's a good idea!

Done

gregestren (approved):

gregestren:
I'm running this by @sdtwigg for the internal push and he asked if this works as expected with aliases (which always seem to work differently than everything else). Was that something you clarified?

philsc:
> I'm running this by @sdtwigg for the internal push and he asked if this works as expected with aliases (which always seem to work differently than everything else). Was that something you clarified?

Haha, totally fair. I haven't validated that it works with aliases. I will add test cases and validate the behaviour.

sdtwigg:
I mostly just saw that you did the change right before the special casing for alias handling and was wondering what the proper order between those two special cases is.

philsc:
> I mostly just saw that you did the change right before the special casing for alias handling and was wondering what the proper order between those two special cases is.

It's a good question. As far as I can tell, the order of those two if statements doesn't matter. If you prefer, I can move it below the alias special casing.
I added some `alias()` targets to the test to make sure it all still works. I couldn't think of any corner cases to cover.

gregestren:
This looks good to me, still. Thank you.

Let me confer with @sdtwigg and then we'll get this imported. Only caveat: @sdtwigg will be out tomorrow through the weekend and I'll be out this Friday and next week.

If you're a bit patient we'll follow up. :)

philsc:
Sounds great @gregestren, thank you!

philsc:
Looks like this patch would help avoid work arounds like grailbio/bazel-compilation-database#100.

gregestren:
Thanks for the ping now (and I'm back online).  I'll get back on this...

PR #15428, opened 2022-05-09
Document @ as allowed in a package name

illicitonion:
Empirically, this works, and some rulesets (e.g. `rules_nodejs`)
explicitly create packages whose names contain `@` characters.

We should document this reality (or forbid it, but that would be a
massively breaking change to the ecosystem).

Additional context: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1651850227929389?thread_ts=1651832164.144609&cid=CEZUUKQ6P

### DIFF
--- a/site/en/concepts/labels.md
+++ b/site/en/concepts/labels.md
@@ -143,8 +143,8 @@ relative to the top-level directory of the containing repository.
 For example: `my/app`.
 
 Package names must be composed entirely of characters drawn from the set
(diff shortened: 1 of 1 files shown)


Wyverald (approved):
Thanks! You can clearly see that `@` is an allowed character in the code https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/cmdline/LabelValidator.java#L55, so this is a no-brainer.

PR #15434, opened 2022-05-09
Catch NumberFormatException while trying to parse thread id.

### LINKED ISSUE #14949: Uncaught NumberFormatException caused by --experimental_execution_log_file path

Bazel sometimes crashes when the path passed to `--experimental_execution_log_file` contains a number greater than `Long.MAX_VALUE` (`9223372036854775807`). This can also happen if you set `--execution_log_json_file` or `--execution_log_binary_file` because a temp file is created for the execution log file and it will have a large number in it.
Stack Trace:
```
java.lang.NumberFormatException: For input string: "99999999999999999999999999999999"
        at java.base/java.lang.NumberFormatException.forInputString(Unknown Source)
        at java.base/java.lang.Long.parseLong(Unknown Source)
        at java.base/java.lang.Long.parseLong(Unknown Source)
        at com.google.devtools.build.lib.profiler.Profiler$JsonTraceFileWriter.getSortIndex(Profiler.java:983)
        at com.google.devtools.build.lib.profiler.Profiler$JsonTraceFileWriter.enqueue(Profiler.java:851)
        at com.google.devtools.build.lib.profiler.Profiler.logTask(Profiler.java:557)
        at com.google.devtools.build.lib.profiler.Profiler.logSimpleTask(Profiler.java:580)
        at com.google.devtools.build.lib.unix.UnixFileSystem$ProfiledNativeFileOutputStream.write(UnixFileSystem.java:555)
        at java.base/java.io.BufferedOutputStream.write(Unknown Source)
        at java.base/java.io.FilterOutputStream.write(Unknown Source)
        at com.google.devtools.build.lib.util.io.AsynchronousFileOutputStream.lambda$new$0(AsynchronousFileOutputStream.java:67)
```
This happens because
1. The [Profiler](https://github.com/bazelbuild/bazel/blob/46e365facff476d785b4cd433124aa4096eac344/src/main/java/com/google/devtools/build/lib/profiler/Profiler.java#L1024) tries to parse a number from every thread name
2. The [`SpawnLogModule` creates an `AsynchronousFileOutputStream`](https://github.com/bazelbuild/bazel/blob/46e365facff476d785b4cd433124aa4096eac344/src/main/java/com/google/devtools/build/lib/bazel/SpawnLogModule.java#L103-L104), which spawns a thread with the [name of the file in the thread's name](https://github.com/bazelbuild/bazel/blob/46e365facff476d785b4cd433124aa4096eac344/src/main/java/com/google/devtools/build/lib/util/io/AsynchronousFileOutputStream.java#L83).

It's not deterministic because it depends on some async thread stuff, but I can get it to happend pretty regularly by building stuff with `--experimental_execution_log_file=/tmp/exec99999999999999999999999999999999.log`

### Possible Fixes

1. Check the length of the number before you try to parse it
2. Catch the `NumberFormatException` exception and `return MAX_SORT_INDEX;`
3. Verify the thread name is one we care about before parsing the number

If you have any preference as to which solution to implement, I'd be happy to submit a PR.

Ubuntu 20.04

`release 5.0.0`

ckolli5:
Fixes #14949.

PiperOrigin-RevId: 439775241

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
+++ b/src/main/java/com/google/devtools/build/lib/profiler/Profiler.java
@@ -980,7 +980,14 @@ public final class Profiler {
         return MAX_SORT_INDEX;
       }
 
-      long extractedNumber = Long.parseLong(numberMatcher.group());
+      long extractedNumber;
+      try {
+        extractedNumber = Long.parseLong(numberMatcher.group());
+      } catch (NumberFormatException e) {
+        // If the number cannot be parsed, e.g. is larger than a long, the actual position is not
+        // really relevant.
+        return MAX_SORT_INDEX;
+      }
 
       if (threadName.startsWith("skyframe-evaluator")) {
         return SKYFRAME_EVALUATOR_SHIFT + extractedNumber;


ckolli5:
https://github.com/bazelbuild/bazel/issues/14949

PR #15462, opened 2022-05-10
Allow any CcInfo providing rule in objc_library deps

keith:
Previously it was invalid to have any custom rules in the deps of
objc_library unless they provided `apple_common.Objc` which very few
rules do, but they do provide `CcInfo` for this type of integration. I
hit this issue with a `rust_library` being depended on by an
`objc_library`.

### DIFF
--- a/src/main/starlark/builtins_bzl/common/objc/attrs.bzl
+++ b/src/main/starlark/builtins_bzl/common/objc/attrs.bzl
@@ -16,6 +16,7 @@
 
 load("@_builtins//:common/objc/semantics.bzl", "semantics")
 
+CcInfo = _builtins.toplevel.CcInfo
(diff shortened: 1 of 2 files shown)


keith:
interested to see if there's something this breaks on CI

keith:
there are more apple related rules with similar restrictions that I imagine would need similar changes

keith:
partially done in https://github.com/bazelbuild/bazel/commit/540892d788727f414cc0cd3ea2ae4d4445914366, I need to check to see if the other part is still needed

keith:
In my repro case of an objc_library depending on a rust_library, the commit linked above was enough, so I'm not sure about the other part of this change, but will reopen if we find a new case.

PR #15469, opened 2022-05-11
Delete CompletableFuture.

benjaminp:
I noticed that CompletableFuture became more prevalant in https://github.com/bazelbuild/bazel/commit/a226eed7a48aa3637c8846b475cf429b0b41caf7. However, I find it's generally simpler to use plain-old ListenableFuture.addListener to propagate cancellations.


src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
```diff
@@ -366,84 +364,86 @@ private ListenableFuture<Long> requestRead(
...
-              try {
-                ((ZstdDecompressingOutputStream) out).closeShallow();
-              } catch (IOException e) {
-                logger.atWarning().withCause(e).log("failed to cleanly close output stream");
-              }
-            }
-          }
-        });
+    Context.CancellableContext grpcContext = Context.current().withCancellation();
+    future.addListener(() -> grpcContext.cancel(null), MoreExecutors.directExecutor());
```
coeuvre, line 375: do we need to check `future.isCancelled()` here?

coeuvre (commented):
I agree, thanks!

src/main/java/com/google/devtools/build/lib/remote/GrpcCacheClient.java
coeuvre: I see. Thanks!

PR #15482, opened 2022-05-12
Safer interaction with framework paths

maxwellE:
Prevents a possible crash when operating on paths without
the proper `.framework` prefix


maxwellE:
@comius I think this is ready for review

keith:
should it reject this bad input instead? still better than crashing at least

Issue #14886, opened 2022-02-22
Robolectric / android_local_test runtime failures with NoClassDefFoundError caused by long classpath argument


oliviernotteghem:
### Description of the problem / feature request:

Running android_local_test with a very long classpath in Bazel currently uses [classpath jar](https://github.com/bazelbuild/bazel/blob/102ce6ddc04063d26165c6a2167059f2348026bf/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L268) in order to workaround command line limitation of the OS (~7k on windows, ~120k on linux). 

However, this approach has two problems :
1. It's not performant, and it requires processing a huge string (in our case, more than 247k long)
2. If the number of the entry in the classpath in more than ~1480 (number determined heuristically), tests fail consistently. It's not clear if it's a generic issue with JVM and/or with classpath jars, as classpath jar do not have apparently any size limits.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

You will only hit this issue with x-large codebase, as in most case, the number of jar dependencies needed to execute android robolectric test will be less than ~1480. Once you have such a large robolectric test target, you will hit the issue consistently

### What operating system are you running Bazel on?

Linux / OpenJDK 11 / Bazel 4.2

###  Suggested fix

For java 9+, leverage java argument file instead of classpath jar.

### Any other information, logs, or outputs that you want to share?

```

There were 3 failures:
--
1) initializationError(com.ubercab.emobility.on_trip.StressFreeBookingInteractorTest)
java.lang.TypeNotPresentException: Type [unknown] not present
at java.base/sun.reflect.annotation.TypeNotPresentExceptionProxy.generateException(TypeNotPresentExceptionProxy.java:46)
at java.base/sun.reflect.annotation.AnnotationInvocationHandler.invoke(AnnotationInvocationHandler.java:86)
at com.sun.proxy.$Proxy2.value(Unknown Source)
at org.junit.internal.builders.AnnotatedBuilder.runnerForClass(AnnotatedBuilder.java:86)
at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:70)
at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:37)
at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:70)
[...]

PR #14887, opened 2022-02-22
Fix robolectric test runtime failures due to large classpath w/ 1.5k+ entries

oliviernotteghem:
fixes #14886

PR #15487, opened 2022-05-12
Fix robolectric test runtime failures due to x-large classpath 

oliviernotteghem:
fixes https://github.com/bazelbuild/bazel/issues/14886
replaces https://github.com/bazelbuild/bazel/pull/14887


src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
```diff
@@ -363,8 +363,17 @@ if [ -z "$CLASSPATH_LIMIT" ]; then
 fi
 
 if (("${#CLASSPATH}" > ${CLASSPATH_LIMIT})); then
-  export JACOCO_IS_JAR_WRAPPED=1
-  create_and_run_classpath_jar
+  java_ver=$(JAVABIN --version 2>&1) | head -1
+  if [[ "$java_ver" =~ [^.0-9]1[.][1-8] ]]; then
```
ted-xie, line 367: This regex will accept versions 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8. Just want to double-check that this is the expected behavior. Do you care about version 1.0?

src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt
```diff
@@ -363,8 +363,17 @@ if [ -z "$CLASSPATH_LIMIT" ]; then
...
+    # Arguments file is not supported by Java SDK version 1.8 and earlier, use classpath jar as fallback
+    export JACOCO_IS_JAR_WRAPPED=1
+    create_and_run_classpath_jar
+  else
+    # Using argument file is more robust and efficient than Classpath jar
+    ARGS_FILE="$(mktemp -t XXXXXXXX.args)"
```
ted-xie, line 373: Should this be cleaned up after the $JAVABIN call returns?

ted-xie (changes requested):
Hi @oliviernotteghem!

I discussed this PR internally and we cannot accept these changes in their current state. Calling $JAVABIN just to get the version number takes around 100-200 ms; while this may not seem like a lot, if there are actions where this pattern occurs many times sequentially, that will quickly add up to be a major performance regression. Furthermore, we don't want to set the precedent of significantly bifurcating rule behavior based on the Java version, since it's a maintenance burden and in general not good for code health.

Instead of checking if the target JDK supports params based on the version number, it would be a better strategy to instead enforce this requirement in the platform runtime itself. This can be achieved by adding `features = ["params_file"],` to the various runtime declarations. There are a few places to make this modification:

* All of the java_repository-s in src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE.tmpl
* Modify remote_java_repository definition in rules_java/toolchains/remote_java_repository.bzl  to accept features.
* Modify rules_java/toolchains/jdk.BUILD so that the java_runtime in there gets the features attribute.

Please let me know if you have any questions!

-Ted

oliviernotteghem:
@sgowroji : I will take a look shortly and will get back to you before end of week.

oliviernotteghem:
@ted-xie : Thanks for the input. However, I am not sure I will time to make your suggested changes without additional guidance. Can you explain the syntax and purpose of file remote_java_repository.bzl? Also, not sure what you mean by modifying jdk.BUILD to get the feature attribute. thanks.

PR #15489, opened 2022-05-12
Switch to the NoFinalizer variant of ZstdOutputStream.

benjaminp:
Finalizers are on the long path to deprecation in Java. Now that we are careful to always close ZstdCompressingInputStream (https://github.com/bazelbuild/bazel/commit/f923a5bb1617045aecbc6babcc43b696bc192a6d), we can stop relying on the finalizer.


coeuvre (approved):
Thanks!

PR #15490, opened 2022-05-12
Add a flag to force Bazel to download certain artifacts

lfpino:
When using `--remote_download_minimal` Bazel only downloads to the client the minimum number of artifacts necessary for the build to succeed. This can be sometimes problematic when local development (i.e. IDE) requires *some* artifacts to do local work (e.g. syntax completion). We add a flag --experimental_force_downloads_regex that allow specifying a regular expression that will force certain artifacts to be forced to downloaded.

TODO: Add integration test

Tested locally that syntax errors disappear in IntelliJ when compiling Bazel itself with `--remote_download_minimal` and `--experimental_force_downloads_regex=".*jar$"`

Signed-off-by: Luis Fernando Pino Duque <luis@engflow.com>


lfpino:
Please ignore this PR, this is WIP

PR #15491, opened 2022-05-12
Parser stable sort based on shard value

Jason-TianrenTan:
# Parser stable sort based on shard value
## About
Bazel now supports sorting by shards when converting two exec logs together. A new parser option `--shard_sorted` is added, and is false by default.
## Instructions
First, build the parser
`bazel build src/tools/execlog:all`
Then run the converting command, and add '--shard_sorted=true'.
```
bazel-bin/src/tools/execlog/parser --log_path=/tmp/exec1.log \
                                       --log_path=/tmp/exec2.log \
                                       --output_path=/tmp/exec1.log.txt \
                                       --output_path=/tmp/exec2.log.txt --shard_sorted=true
```
This enables bazel to sort the output shards in `exec2.log.txt` ascendingly based on the index number.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@
 # User-specific .bazelrc
 user.bazelrc
 bin/bazel
+tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@
 .bazelversion
 # User-specific .bazelrc
 user.bazelrc
+bin/bazel
--- a/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
+++ b/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
@@ -145,15 +145,14 @@ final class ExecLogParser {
 
     private void processInputFile(Parser input) throws IOException {
       sameActions = new PriorityQueue<>((e1, e2) -> (e1.position - e2.position));
+      //determine whether to use arraydeque or priority queue with shard_sorted parameter
       if (golden.isShardSorted) {
         uniqueActions = new PriorityQueue<>((e1, e2) -> {
           int shard_1 = getShard(e1), shard_2 = getShard(e2);
           return shard_1 - shard_2;
         });
-        System.out.println("Sort with shard.");
       } else {
         uniqueActions = new ArrayDeque<>();
-        System.out.println("Sort with unshard.");
       }
 
       SpawnExec ex;
--- a/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
+++ b/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
@@ -125,7 +125,7 @@ final class ExecLogParser {
 
     ReorderingParser(Golden golden, Parser input, String shardSorted) throws IOException {
       this.golden = golden;
-      this.golden.isShardSorted = (shardSorted == "true") ? true : false;
+      this.golden.isShardSorted = (shardSorted.equals("true")) ? true : false;
       processInputFile(input);
     }
 
--- a/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
+++ b/src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
@@ -32,6 +32,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.PriorityQueue;
 import java.util.Queue;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * A tool to inspect and parse the Bazel execution log.
@@ -86,6 +88,7 @@ final class ExecLogParser {
       // Value: the position of the action in the file (e.g., 0th, 1st etc).
       private final Map<String, Integer> positions;
       private int index;
+      private boolean isShardSorted;
 
       public Golden() {
         positions = new HashMap<>();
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -26,3 +26,5 @@
 .bazelversion
 # User-specific .bazelrc
 user.bazelrc
+bin/bazel
+tmp/
```
meteorcloudy, line 30: I don't think it's necessary to add those two to the .gitignore.

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ParserOptions.java
```diff
@@ -60,4 +60,13 @@ public class ParserOptions extends OptionsBase {
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "If set, only output the executions that used the given runner.")
   public String restrictToRunner;
+
+  @Option(
+          name = "shard_sorted",
```
meteorcloudy, line 65: What was the original order? Can we just sort them by default and not have this option at all?

meteorcloudy (changes requested):

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ParserOptions.java
Jason-TianrenTan: Hi meteorcloudy, thanks for the feedback! The original order was to reorder `/tmp/exec2.log` to match actions detected in `/tmp/exec1.log`. This causes the shard indices to be rearranged in an unorganized order in the output log. Therefore, we added a new option so that the user could choose whether to use the original rearranging approach or to sort by shards.

If this new sorting method is satisfactory and you wish to make it as default, we are happy to remove the newly added option while keeping the sorting code.

meteorcloudy: I think having only the sorted result should be fine.

Jason-TianrenTan: Ok, we deleted the option and now it sorts by shards by default.

.gitignore
```diff
@@ -25,4 +25,4 @@
 # Bazelisk version file
 .bazelversion
 # User-specific .bazelrc
-user.bazelrc
+user.bazelrc
```
meteorcloudy, line 28: Can you revert the change in this file?

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ParserOptions.java
```diff
@@ -60,4 +60,5 @@ public class ParserOptions extends OptionsBase {
       effectTags = {OptionEffectTag.UNKNOWN},
       help = "If set, only output the executions that used the given runner.")
   public String restrictToRunner;
+
```
meteorcloudy, line 63: Remove empty line

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
```diff
@@ -130,9 +132,22 @@ public Element(int position, SpawnExec element) {
     // actions in input that are not in the golden, in order received.
     Queue<SpawnExec> uniqueActions;
 
+    // use regex expression to find shard value in file
+    private int getShard(SpawnExec ex) {
+      Pattern pattern = Pattern.compile("shard_[1-9][0-9]?");
```
meteorcloudy, line 137: Not sure if the shard number can go over 100, but can we make this pattern more general and robust in case it does?

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
```diff
@@ -86,6 +87,7 @@ public static class Golden {
       // Value: the position of the action in the file (e.g., 0th, 1st etc).
       private final Map<String, Integer> positions;
       private int index;
+      private boolean isShardSorted;
```
meteorcloudy, line 90: No longer used, right?

meteorcloudy (changes requested):

.gitignore
Jason-TianrenTan: Will do.

Jason-TianrenTan (commented):
- Deleted isShardSorted option. 
- Removed extra lines
- Reverted uneccesary changes

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
Jason-TianrenTan: Deleted.

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ExecLogParser.java
Jason-TianrenTan: Made it into 1 - 1000.

src/tools/execlog/src/main/java/com/google/devtools/build/execlog/ParserOptions.java
Jason-TianrenTan: Removed

Jason-TianrenTan:
Hey @sgowroji @meteorcloudy, please take a look at this new revision when you can. Thanks!

meteorcloudy (approved):
Thanks!

sgowroji:
Hello @Jason-TianrenTan, Could you please check the above comment as the PR is awaiting to merge. Thanks!

Issue #15386, opened 2022-05-02
Upgrade zlib to version 1.2.12


jsharpe:
### Description of the bug:

Bazel embeds zlib version 1.2.11 which is subject to know vulenerabilities. 1.2.12 has been released which addresses these vulnerabilites and so the version used in bazel should be updated.

### Any other information, logs, or outputs that you want to share?

I believe that this should be addressed before the release of 5.2. zlib 1.2.12 should be a drop in replacement.
@bazel-io flag

ckolli5:
@bazel-io fork 5.2.0

daohu527:
hi, how do I temporarily skip this question first, the fixed version is a terrible disaster

meteorcloudy:
Closed due to https://github.com/bazelbuild/bazel/pull/15497

PR #15497, opened 2022-05-13
Upgrade zlib to 1.2.12

meteorcloudy:
Fixes https://github.com/bazelbuild/bazel/issues/15386


meteorcloudy:
I was planning to replace zlib with an external repo in WORKSPACE, but turns out `@bazel_tools` embeds zlib sources currently, so it's harder to set it up. We can revisit after clean up `@bazel_tools` (should be easier when it's a Bazel module).

Wyverald (approved):

meteorcloudy:
@sgowroji Please help merge this one, thanks!

Issue #15502, opened 2022-05-14
python_stub_template.txt doesn't find runfiles associated with pyz file.


jonathan-enf:
### Description of the bug:

In this scenario:
* a py_binary file ("foo.py) has a data dependency on another py_binary ("bar.py")
* foo.py calls bar as a subprocess.

This works if invoked as "blaze run :foo" because "foo" executes in a context where a "foo.runfiles" directory exists, and all of "bar.py"'s dependencies are easily found in "foo.runfiles" directories.

However, if "foo.zip" is built using the "--build_python_zip" option, "python3 foo.zip" is invoked, execution of the "bar" resource will fail because the python stub that is compiled into "bar" is calling FindModuleSpace.  FindModuleSpace searches for any runfiles directory that is named dot-runfiles (".runfiles").  Unfortunately, the runfiles directory created when CreateModuleSpace unzips the python zip archive is only named "runfiles" (no dot).  Thus, execution fails.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
git clone git@github.com:jonathan-enf/bazel.git
cd bazel/src/test/py/bazel
bazel run :foo  # this succeeds
bazel build :foo-zip
python3 ../../../../bazel-bin/src/test/py/bazel/foo.zip  # this fails
```

### Which operating system are you running Bazel on?

linux

### What is the output of `bazel info release`?

release 5.1.1

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
$ git remote get-url origin; git rev-parse master; git rev-parse HEAD
git@github.com:jonathan-enf/bazel.git
96fa291cb1515706ece1be7f9510ff4ed2a40625
96fa291cb1515706ece1be7f9510ff4ed2a40625
```

### Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

My reproduction case:

```
$ blaze run :foo
INFO: Analyzed target //src/test/py/bazel:foo (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //src/test/py/bazel:foo up-to-date:
  bazel-bin/src/test/py/bazel/foo
INFO: Elapsed time: 0.165s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
[...]

PR #15503, opened 2022-05-14
fix python_stub_template.txt for work with python zip archives

jonathan-enf:
This is a patch that fixes the bug I just opened: https://github.com/bazelbuild/bazel/issues/15502

This addresses the issue by modifying LoadModuleSpace to search first for a ".runfiles" directory, and failing that, to search
for a "/runfiles/" directory such as the one created by CreateModuleSpace when the stub is invoked in the context of a zipfile.

This PR also includes a reproduction case: src/test/py/bazel/{foo,bar,bum}.py

I will fill out the contributor agreement momentarily...

Tested:

1. Verified that my repro case reproduces the issue: `bazel build :foo.zip; python3 ../../../../bazel-bin/src/test/py/bazel/foo.zip` fails.
2. Built a patched bazel-dev binary.
3. Verified that my repro case now passes: `bazel-dev build :foo.zip; python3 ../../../../bazel-bin/src/test/py/bazel/foo.zip` passes.

I tested this patch by building bazel-dev from source, and using

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -90,14 +90,14 @@ dist_http_archive(
     patch_cmds_win = EXPORT_WORKSPACE_IN_BUILD_FILE_WIN,
 )
 
-# This is a mock version of bazelbuild/rules_python that contains only
-# @rules_python//python:defs.bzl. It is used by protobuf.
-# TODO(#9029): We could potentially replace this with the real @rules_python.
-new_local_repository(
+http_archive(
     name = "rules_python",
-    build_file = "//third_party/rules_python:BUILD",
-    path = "./third_party/rules_python",
-    workspace_file = "//third_party/rules_python:rules_python.WORKSPACE",
+    sha256 = "cdf6b84084aad8f10bf20b46b77cb48d83c319ebe6458a18e9d2cebf57807cdd",
+    strip_prefix = "rules_python-0.8.1",
+    urls = [
+        "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
+    ],
 )
 
 local_repository(
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -106,7 +106,14 @@ def FindModuleSpace():
     if os.path.isdir(module_space):
       return module_space
 
-    runfiles_pattern = r'(.*\.runfiles)' + (r'\\' if IsWindows() else '/') + '.*'
+    slash_pattern = r'\\' if IsWindows() else '/'
+    runfiles_pattern = r'(.*\.runfiles)' + slash_pattern + '.*'
+    matchobj = re.match(runfiles_pattern, stub_filename)
+    if matchobj:
+      return matchobj.group(1)
+
+    # bazel pyz archives create a directory named "runfiles".  Try that next:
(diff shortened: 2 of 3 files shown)


jonathan-enf:
I've submitted my CLA.  I suspect the build breakage is my attempt to integrate external @rules_python.  I'll revert that momentarily.

WORKSPACE
```diff
@@ -100,6 +100,16 @@ new_local_repository(
...
+# This is the real rules_python.  It's needed for a specific bug reproduction case.
+http_archive(
+    name = "real_rules_python",
+    sha256 = "cdf6b84084aad8f10bf20b46b77cb48d83c319ebe6458a18e9d2cebf57807cdd",
+    strip_prefix = "rules_python-0.8.1",
+    urls = [
+        "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
+    ],
+)
+
```
comius, line 112: If this is a Bazel issue, shouldn't it be possible to test it without bringing in a new dependency?

We are trying to keep the list of dependencies minimal. Additional one just for a single test is not an option.

src/test/py/bazel/foo.py
```diff
@@ -0,0 +1,18 @@
+# standard libraries
```
comius, line 1: If this is a test is should be name this way. It should be structured similar to other tests - arrange, act, assert. There are not asserts here.

src/test/py/bazel/bar.py
```diff
@@ -0,0 +1,4 @@
+from src.test.py.bazel import bum
```
comius, line 1: Files that are used in the tests should be under testdata directory or generated/arranged from within the test. See other tests as an example.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -106,7 +106,14 @@ def FindModuleSpace():
     if os.path.isdir(module_space):
       return module_space
 
-    runfiles_pattern = r'(.*\.runfiles)' + (r'\\' if IsWindows() else '/') + '.*'
+    slash_pattern = r'\\' if IsWindows() else '/'
+    runfiles_pattern = r'(.*\.runfiles)' + slash_pattern + '.*'
```
comius, line 110: This change comparing to the old implementation reduces readability in my opinion.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -106,7 +106,14 @@ def FindModuleSpace():
...
+    matchobj = re.match(runfiles_pattern, stub_filename)
+    if matchobj:
+      return matchobj.group(1)
+
+    # bazel pyz archives create a directory named "runfiles".  Try that next:
+    runfiles_pattern = r'(.*' + slash_pattern + 'runfiles)' + slash_pattern + '.*'
```
comius, line 116: This should be addressed in pyz archive - but doing it there will break users. So providing a second archive that's correctly packaged is probably better.

Runfiles are used by other rules as well. Fixing it in Python stub doesn't work for them.

WORKSPACE
jonathan-enf: This is a bazel issue -- the bug is in python_stub_template.txt.

This is a bazel issue.  It can be exposed without the rules_python package, but that would require recreating about 100 lines of python code from rules_python.  It's probably best to test this functionality using the version of rules_python that our customers are using.

src/test/py/bazel/foo.py
jonathan-enf: naming fixed.

keertk:

Issue #15507, opened 2022-05-16
There should be a target which represents the consolidated LICENSE for everything included in the bazel binary


aiuto:
### Description of the feature request:

Bazel's binary includes third_party code. We need to provide the capability for someone to gather the transitive licenses for all of that.

### What underlying problem are you trying to solve with this feature?

Audit compliance

### Which operating system are you running Bazel on?

n/a

### What is the output of `bazel info release`?

n/a

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

n/a

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
n/a
```

meteorcloudy:
@aiuto Are you actively working on this, should this be P1?

aiuto:
I am actively working on this. It is absolutely a P1.

Issue #11753, opened 2020-07-13
Apply rules_license to third_party tree


aiuto:
We should use rules_license in Bazel. This woudl serve as
- a proof of concept of the licensing rules
- a more precise way of delineating what licenses are used in bazel itself and/or the build process.

aiuto:
This will also be useful for the debian packaging effort.

aiuto:
Most of third_party now has `license` rules. The need to have `package_info` rules too.

PR #15509, opened 2022-05-17
Initial import of rules_license

aiuto:
Initial import of rules_license.

This is a pull from https://github.com/bazelbuild/rules_license/releases/tag/0.0.2, with some tweaking to adopt the 'srcs' construct for Bazel. No other code changes were made.

Since this repository must become an external repository name rules_license, it follows the model of googleapis and remoteapis. Thus, the visibility rules and use of self referential names is different from most other imports.
NOTE: This can not be a pure third_party only PR.  You can't build bazel-srcs without having the local repository from the workspace. An alternative would be to remove WORKSPACE and third_party/BUILD from this PR, do the code only, then have a small PR which just modifies those two files. 

See #15507, #11753


aiuto:
Hi reviewers. Since this is third_party I need the special import.

WORKSPACE
```diff
@@ -100,6 +100,14 @@ new_local_repository(
     workspace_file = "//third_party/rules_python:rules_python.WORKSPACE",
 )
 
+# Rules license must always be referred to as @rules_license even if imported
+# in to third_party. That allows the full classname of the providers to be
+# aligned across all rule sets.
+local_repository(
```
meteorcloudy, line 106: Why not just use http_archive? We really shouldn't add more things in the third_party directory.

meteorcloudy (changes requested):

PR #15514, opened 2022-05-18
Restructure blob upload code.

benjaminp:
The previous blob upload code had a lot of nesting and state shared between Future callbacks that made it hard to follow and modify. There is some fundamental complexity in the problem this code is solving; namely, retrying uploads can resume from the last comitted offset rather than always the beginning of the blob. I hope my rewrite makes the code easier to understand and paves the way for some future behavior modifications I want to make in this area.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java
@@ -23,6 +23,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import build.bazel.remote.execution.v2.Digest;
 import com.google.bytestream.ByteStreamGrpc;
 import com.google.bytestream.ByteStreamGrpc.ByteStreamFutureStub;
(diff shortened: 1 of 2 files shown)


coeuvre:
What future modifications do you have in mind? We plan to refactor this (and eventually the whole remote module) with RxJava and I made a [prototype](https://github.com/coeuvre/bazel/blob/refactor-bytestream-uploader/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java) a while back.

benjaminp:
A few things I have in mind:
- Fix a bug related to compressed uploads
- Make opening of the input files more lazy
- Add an upload progress indicator to the UI

I wouldn't be offended if the RxJava version lands instead of or quickly after this change. :) Just as long as the code ends up in a clearer state.

ulfjack:
Ping? There are pending *bug fixes* that are affecting users of compressed uploads that are pending after this PR.

jgavris:
+1 Also holding off on compressed uploads and we'd really like them.

coeuvre (approved):
LGTM. Sorry for the delay, was on vacation.

PR #7519, opened 2019-02-22
Make Xcode setup more resilient

keith:
Previously if you tried to build a project from a entirely clean bazel
state, and ctrl-c'd while bazel was fetching the local_config_xcode
information, you would end up in a state where the
local_config_xcode/BUILD file contained error information, which made
the Apple crosstool unusable until you did a `clean --expunge`. Now the
Xcode setup calls `fail()` in those cases, which makes bazel re-run it
again the next time you try to build.

It is still possible to reproduce if you ctrl-c while the `xcodebuild
-version` command that determines if you have Xcode installed at all
runs. Because we want to support users who don't have Xcode installed on
macOS, we cannot fail in that case.

Fixes https://github.com/bazelbuild/bazel/issues/6056

PR #15517, opened 2022-05-18
Allow skipping Xcode locator runs

keith:
For builds that require Xcode, running xcode_locator during toolchain
setup is potentially duplicative. The only things it's used for is to
check that they exist if it's expected, and to add to the allowed
include directories. The include directories already includes
`/Applications` which is where _most_ people likely have Xcode
installed. This saves a few seconds when the toolchain is reconfigured.

https://bazelbuild.slack.com/archives/CD3QY5C2X/p1652819625121889

### DIFF
--- a/tools/cpp/osx_cc_configure.bzl
+++ b/tools/cpp/osx_cc_configure.bzl
@@ -157,17 +157,19 @@ def configure_osx_toolchain(repository_ctx, cpu_value, overriden_tools):
 
     env = repository_ctx.os.environ
     should_use_xcode = "BAZEL_USE_XCODE_TOOLCHAIN" in env and env["BAZEL_USE_XCODE_TOOLCHAIN"] == "1"
+    skip_xcode_fetch = "BAZEL_SKIP_XCODE_FETCH" in env and env["BAZEL_SKIP_XCODE_FETCH"] == "1"
     xcode_toolchains = []
 
     # Make the following logic in sync with //tools/cpp:cc_configure.bzl#cc_autoconf_toolchains_impl
-    (xcode_toolchains, xcodeloc_err) = run_xcode_locator(
-        repository_ctx,
-        paths["@bazel_tools//tools/osx:xcode_locator.m"],
-    )
-    if should_use_xcode and not xcode_toolchains:
-        fail("BAZEL_USE_XCODE_TOOLCHAIN is set to 1 but Bazel couldn't find Xcode installed on the " +
-             "system. Verify that 'xcode-select -p' is correct.")
-    if xcode_toolchains:
+    if not skip_xcode_fetch:
+        (xcode_toolchains, xcodeloc_err) = run_xcode_locator(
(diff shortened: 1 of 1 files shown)


oquenchil:
@googlewalt can you please take a look?

tools/cpp/osx_cc_configure.bzl
```diff
@@ -157,17 +157,19 @@ def configure_osx_toolchain(repository_ctx, cpu_value, overriden_tools):
 
     env = repository_ctx.os.environ
     should_use_xcode = "BAZEL_USE_XCODE_TOOLCHAIN" in env and env["BAZEL_USE_XCODE_TOOLCHAIN"] == "1"
+    skip_xcode_fetch = "BAZEL_SKIP_XCODE_FETCH" in env and env["BAZEL_SKIP_XCODE_FETCH"] == "1"
```
googlewalt, line 160: How is this expected to be use?  Does it require setting BAZEL_USE_XCODE_TOOLCHAIN?  If so, perhaps add a comment indicating such?

tools/cpp/osx_cc_configure.bzl
```diff
@@ -157,17 +157,19 @@ def configure_osx_toolchain(repository_ctx, cpu_value, overriden_tools):
...
-    (xcode_toolchains, xcodeloc_err) = run_xcode_locator(
-        repository_ctx,
-        paths["@bazel_tools//tools/osx:xcode_locator.m"],
-    )
-    if should_use_xcode and not xcode_toolchains:
-        fail("BAZEL_USE_XCODE_TOOLCHAIN is set to 1 but Bazel couldn't find Xcode installed on the " +
-             "system. Verify that 'xcode-select -p' is correct.")
-    if xcode_toolchains:
+    if not skip_xcode_fetch:
+        (xcode_toolchains, xcodeloc_err) = run_xcode_locator(
```
googlewalt, line 165: There is a reference to xcodeloc_err on line 231 that caused an error when I tested it.  Is it reasonable to delete that and just fail hard here if there is an error?

googlewalt (changes requested):

tools/cpp/osx_cc_configure.bzl
googlewalt: Trace the error handling back to https://github.com/bazelbuild/bazel/commit/cc21998c299b4d1f97df37b961552ff8168da17f#diff-c731a543f25a190052c3f045e894da8088b94ca4b0e067381fe281a46f26f3e0R671.  It's unfortunate the error handling grew farther and farther away over time.  I still don't understand if there are reasons why we don't just fail fast.

This may also be relevant: https://github.com/bazelbuild/bazel/pull/7519.

keith:
Since this logic is moving out of this repo in https://github.com/bazelbuild/bazel/pull/16619, I'm going to close this and replace it with https://github.com/bazelbuild/apple_support/pull/191

Issue #12522, opened 2020-11-19
Cannot bootstrap bazel on Haiku


extrowerk:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

Bootstrap fails on 64 bit Haiku with openjdk8.

### Feature requests: what underlying problem are you trying to solve with this feature?

To get bazel working on Haiku.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Try to bootstrap

### What operating system are you running Bazel on?

None. I try to get it work on Haiku.

### What's the output of `bazel info release`?

NA.

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

```
/RAMDisk> env EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk" bash ./compile.sh
�  Building Bazel from scratch
ERROR: JAVA_HOME (/packages/openjdk14-14.0.2.12-1/.self/lib/openjdk14) is not a path to a working JDK.
/RAMDisk> export JAVA_HOME=/boot/system/lib/openjdk8
/RAMDisk> env EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk" bash ./compile.sh
�  Building Bazel from scratch......
�  Building Bazel with Bazel.
DEBUG: /tmp/bazel_1XjWedpI/out/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:59:14: 
Current running Bazel is not a release version and one was not defined explicitly in rbe_autoconfig target. Falling back to '3.1.0'
DEBUG: /tmp/bazel_1XjWedpI/out/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:103:14: rbe_ubuntu1804_java11 not using checked in configs as detect_java_home was set to True 
DEBUG: /tmp/bazel_1XjWedpI/out/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:59:14: 
Current running Bazel is not a release version and one was not defined explicitly in rbe_autoconfig target. Falling back to '3.1.0'
[...]

PR #15522, opened 2022-05-18
Allow building the client on Haiku.

Tevo45:
This contains the client portion of the changes for my Haiku port, as mentioned on https://github.com/bazelbuild/bazel/issues/12522#issuecomment-1121321856.

### DIFF
--- a/src/main/cpp/blaze.cc
+++ b/src/main/cpp/blaze.cc
@@ -1433,7 +1433,7 @@ static map<string, EnvVarValue> PrepareEnvironmentForJvm() {
 
   const char *want_locale = "en_US.ISO-8859-1";
   bool override_locale = true;
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(__HAIKU__)
   locale_t iso_locale = newlocale(LC_CTYPE_MASK, want_locale, (locale_t)0);
   if (iso_locale == 0) {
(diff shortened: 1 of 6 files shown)


Tevo45:
Seems like CI will be failling for all of my PRs since they depend on bazelbuild/platforms#41, which is as of yet unmerged. I imagine I'd have to wait for that to go in, submit a change updating the version on the Bazel repository and then rebase all my changes on that?

src/main/cpp/blaze.cc
```diff
@@ -1433,7 +1433,7 @@ static map<string, EnvVarValue> PrepareEnvironmentForJvm() {
 
   const char *want_locale = "en_US.ISO-8859-1";
   bool override_locale = true;
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(__HAIKU__)
```
aiuto, line 1436: Is there a way to express this in a positive way? Something like
#if defined(LC_CTYPE_MASK) or defined(__USE_XOPEN2K8)

aiuto: Or maybe _XOPEN_VERSION >= 700

Tevo45: `LC_CTYPE_MASK` sounds better to me, as it tests for the locale API directly rather than advertised conformace to the corresponding POSIX/XSI version (as to avoid false negatives).

PR #15528, opened 2022-05-19
Delete unused osx_wrapper

keith:
I missed this in 17273615633abbbdcd0b0f7b0945a7d479b709e4


keith:
@sgowroji bump

Wyverald (approved):

brentleyjones:
Does this need its labels adjusted now?

Wyverald:
Thanks. @sgowroji we discussed this before and I remember the conclusion being that we can just merge if the PR has approval, with or without the "awaiting-pr-merge" label. Is that still true?

Wyverald:
We chatted offline and decided that we'll keep applying the "awaiting-pr-merge" label.

PR #15532, opened 2022-05-19
Ensure cc rules "cpp_link" exec group works for cc rules

chancila:
The docs in https://docs.bazel.build/versions/main/exec-groups.html#execution-groups-for-native-rules state that there is a `cpp_link` execution group that allows you to customize the execution of the link actions of cc rules.

This only seems to work for the objc rule classes however since the normal cc linking do not inherit from CcLinkingRule rule class.

This PR ensures consistency in the execution groups across cc rules that link.


chancila:
I didn't find any tests verifying the `cpp_link` exec group behavior

chancila:
ping

oquenchil:
Closing as stale, cc_binary is in Starlark now.

PR #7424, opened 2019-02-13
Git repo cache

Globegitter:
This picks up the work mentioned here https://github.com/bazelbuild/bazel/pull/5928#issuecomment-429125045 and should fix #5116

I refactored things a bit and made sure that the git cache actually works offline as well as without a cache and added a simple test. I have tested it in our monorepo and it seems to be working well. 

Not sure what the `TODO needs a proper implementation.` in the `_hash` method refers to maybe you now @aehlig or @siedentop ?

Issue #12227, opened 2020-10-07
FR: Share extracted external libraries across workspaces


dan-cohn:
> ATTENTION! Please read and follow:
> - if this is a _question_ about how to build / test / query / deploy using Bazel, or a _discussion starter_, send it to bazel-discuss@googlegroups.com
> - if this is a _bug_ or _feature request_, fill the form below as best as you can.

### Description of the problem / feature request:

I would like Bazel to be able to share _extracted_ repos across workspaces. In other words, the "repository cache" should go deeper than simply making downloaded archives available in a centralized location. It would help save time and space if the unpacked contents were also shared.

### Feature requests: what underlying problem are you trying to solve with this feature?

Bazel's repository cache is shared across workspaces, which avoids downloading the same external libraries multiple times. This saves a lot of time and network utilization, especially in CI environments where builds occur across many different repos and versions. The problem is that only the original downloaded artifacts are shared and not the extracted files. This means that Bazel has to repeatedly unzip/unpack archives. This takes a lot of time, consumes CPU, and fills the filesystem with multiple copies of the exact same extracted libraries.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

N/A

### What operating system are you running Bazel on?

RHEL 7/8 and CentOS 7/8

### What's the output of `bazel info release`?

release 3.4.1

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

N/A

###  Have you found anything relevant by searching the web?

No.

### Any other information, logs, or outputs that you want to share?

N/A

Wyverald:
Hi dan-cohn, we're working on a new design for Bazel's external dependencies, and this idea is being incorporated into the new design. Please consider joining the external-deps@bazel.build mailing list or following along on the #external-deps channel on Slack.

tim-chaplin-dd:
> Hi dan-cohn, we're working on a new design for Bazel's external dependencies, and this idea is being incorporated into the new design. Please consider joining the [external-deps@bazel.build](mailto:external-deps@bazel.build) mailing list or following along on the #external-deps channel on Slack.

[...]

PR #15536, opened 2022-05-20
Add support for repository cache to `git_repository`

thii:
If a `BAZEL_GIT_REPOSITORY_CACHE` environment variable is set to a path,
use that as the cache directory for `git_repository`.

Based on https://github.com/bazelbuild/bazel/pull/7424

### DIFF
--- a/src/test/shell/bazel/starlark_git_repository_test.sh
+++ b/src/test/shell/bazel/starlark_git_repository_test.sh
@@ -686,4 +686,38 @@ EOF
   expect_log "shallow_since not allowed if a tag is specified; --depth=1 will be used for tags"
 }
 
+# Verifies that:
+# 1. Rule creates a git_repository rule with a commit hash and caches it.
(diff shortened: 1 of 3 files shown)


aaronmondal:
GitHub stated in https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/ that there won't be support for hash stability for commit-level tar archives. This might require users living at head to switch from the `http` rules to `git` rules sooner or later. I believe that those users tend to be the ones most dependent on good caching properties.

In light of this it might be a good idea to bump the priority of this PR.

Wyverald:
Thank you for your contribution, @thii! And sorry for the long delay. I kept pushing this off since it adds quite a bit of complexity to the repo rule, which is extra maintenance burden, etc etc. But the good news is that this will likely be subsumed by the [true repo cache](https://github.com/bazelbuild/bazel/issues/12227#issuecomment-1710887802) feature, which will likely land in 7.1 or 7.2. So I'll go ahead and close this one now.

PR #15539, opened 2022-05-20
Add test cases for HttpDownloader

KaneBetter:
Add three test cases for uncovered methods in `HttpDownloader` class to test whether the `download` method can successfully download the URL. The tests are under `com.google.devtools.build.lib.bazel.repository.downloader`. Please feel free to provide any feedback and request changes.


Wyverald (approved):
Looks like these are essentially copies of some test cases above, but using `HttpDownloader#download` instead of `DownloadManager#download`. So it's all good.

Issue #15430, opened 2022-05-09
Crash with empty value of --repo_env


layus:
### Description of the bug:

When a --repo_env var lacks the value part, Bazel crashes instead of using the value of that variable from the environment.

```
Starting local Bazel server and connecting to it...
Loading: 0 packages loaded
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'REPOSITORY_DIRECTORY:@debug' (requested by nodes 'IGNORED_PACKAGE_PREFIXES:@debug')
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:674)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:382)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NullPointerException: null value in entry: FOO=null
	at com.google.common.collect.CollectPreconditions.checkEntryNotNull(CollectPreconditions.java:32)
	at com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:99)
	at com.google.common.collect.RegularImmutableMap.fromEntries(RegularImmutableMap.java:73)
	at com.google.common.collect.ImmutableMap.copyOf(ImmutableMap.java:460)
	at com.google.common.collect.ImmutableMap.copyOf(ImmutableMap.java:433)
	at com.google.devtools.build.lib.bazel.repository.starlark.StarlarkRepositoryFunction.fetch(StarlarkRepositoryFunction.java:174)
	at com.google.devtools.build.lib.rules.repository.RepositoryDelegatorFunction.fetchRepository(RepositoryDelegatorFunction.java:415)
	at com.google.devtools.build.lib.rules.repository.RepositoryDelegatorFunction.compute(RepositoryDelegatorFunction.java:349)
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:590)
	... 4 more
```

[...]

PR #15540, opened 2022-05-20
Fix repo_env crash with empty value

rustberry:
Hi,

As reported by #15430, --repo_env option is now unable to read from environment variables.

This problem used to be fixed in [#12886](https://github.com/bazelbuild/bazel/pull/12886), but seems to be introduced again in commit [c2bdd034014f66ce14529cc353cda18a32320f6c](https://github.com/bazelbuild/bazel/commit/c2bdd034014f66ce14529cc353cda18a32320f6c), line 256 and 257 [link here](https://github.com/bazelbuild/bazel/blob/c2bdd034014f66ce14529cc353cda18a32320f6c/src/main/java/com/google/devtools/build/lib/runtime/CommandEnvironment.java#L256) where the value put into `repoEnv` is still the original one that might be `null`. This commit would revert those changes and put the modified value into `repoEnv`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/CommandEnvironment.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/CommandEnvironment.java
@@ -266,8 +266,8 @@ public class CommandEnvironment {
         value = System.getenv(name);
       }
       if (value != null) {
-        repoEnv.put(entry.getKey(), entry.getValue());
-        repoEnvFromOptions.put(entry.getKey(), entry.getValue());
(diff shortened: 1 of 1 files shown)


SalmaSamy (approved):

rustberry:
Hi there, friendly ping @meteorcloudy :)

meteorcloudy:
This is already fixed in https://github.com/bazelbuild/bazel/commit/ed886ee4727d2f882013f2d75ecc17f5596c6c86

PR #15542, opened 2022-05-20
Content-based path mapping for Starlark rules

fmeum:
Starlark actions with mnemonics listed in the flag have their input and output paths mapped according to the following scheme when staged for sandboxed or remote execution:

* source file paths are preserved as is;
* paths of generated input files of the form `bazel-out/k8-fastbuild/bin/pkg/file` are replaced with `bazel-out/<hash>/bin/pkg/file`, where `<hash>` is a digest of the contents of the file;
* paths of output files of the form `bazel-out/k8-fastbuild/bin/pkg/file` are replaced with `bazel-out/out/bin/pkg/file`.

This path mapping is applied both to the location at which files are staged in the sandbox or during remote execution and to paths emitted into structured command lines. In order to build successfully with this flag, action command lines

* must not contain `dirname`, `path`, or `root.path` of any `File` object evaluated at analysis time;
* must not reference `File` objects that aren't declared inputs or outputs of the action.

Spawns derived from actions opted into content-based path mapping can only be executed in symlink-based sandboxes or remotely.

Work towards #6526

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/Artifact.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/Artifact.java
@@ -1675,7 +1675,9 @@ public abstract class Artifact
 
     /**
      * Any two instances are considered equal as they should not cause the Starlark method lookup
-     * cache to be invalidated.
+     * cache to be invalidated. This is purely a performance/memory optimization to prevent the
+     * cache in {@link net.starlark.java.eval.CallUtils#getCacheValue(Class, StarlarkSemantics)}
+     * from growing too large and becoming ineffective.
      */
     @Override
     public boolean equals(Object obj) {
--- a/src/main/java/com/google/devtools/build/lib/actions/Artifact.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/Artifact.java
@@ -42,6 +42,7 @@ import com.google.devtools.build.lib.skyframe.serialization.SerializationExcepti
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec.VisibleForSerialization;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.SerializationConstant;
 import com.google.devtools.build.lib.starlarkbuildapi.FileApi;
+import com.google.devtools.build.lib.starlarkbuildapi.FileRootApi;
 import com.google.devtools.build.lib.util.FileType;
 import com.google.devtools.build.lib.util.FileTypeSet;
 import com.google.devtools.build.lib.vfs.Path;
@@ -61,6 +62,8 @@ import javax.annotation.Nullable;
 import net.starlark.java.eval.EvalException;
 import net.starlark.java.eval.Printer;
 import net.starlark.java.eval.Starlark;
+import net.starlark.java.eval.StarlarkSemantics;
+import net.starlark.java.eval.StarlarkSemantics.Key;
 
 /**
  * An Artifact represents a file used by the build system, whether it's a source file or a derived
@@ -564,10 +567,15 @@ public abstract class Artifact
    *
    * <p> The directory name is always a relative path to the execution directory.
    */
-  @Override
   public final String getDirname() {
     PathFragment parent = execPath.getParentDirectory();
     return (parent == null) ? "/" : parent.getSafePathString();
+  }
+
+  @Override
+  public final String getDirnameForStarlark(StarlarkSemantics starlarkSemantics)
+      throws EvalException {
+    return starlarkSemantics.get(ArtifactPathMapper.STARLARK_SEMANTICS_KEY).getMappedDirname(this);
   }
 
   /**
@@ -629,9 +637,14 @@ public abstract class Artifact
    * package-path entries (for source Artifacts), or one of the bin, genfiles or includes dirs (for
    * derived Artifacts). It will always be an ancestor of getPath().
    */
-  @Override
   public final ArtifactRoot getRoot() {
     return root;
+  }
+
+  @Override
+  public final FileRootApi getRootForStarlark(StarlarkSemantics starlarkSemantics)
+      throws EvalException {
+    return starlarkSemantics.get(ArtifactPathMapper.STARLARK_SEMANTICS_KEY).getMappedRoot(this);
   }
 
   @Override
@@ -690,9 +703,15 @@ public abstract class Artifact
   }
 
   /** Returns this.getExecPath().getPathString(). */
-  @Override
   public final String getExecPathString() {
     return execPath.getPathString();
+  }
+
+  @Override
+  public final String getExecPathStringForStarlark(StarlarkSemantics starlarkSemantics)
+      throws EvalException {
+    return starlarkSemantics.get(ArtifactPathMapper.STARLARK_SEMANTICS_KEY)
+        .getMappedExecPathString(this);
   }
 
   public final String getRootRelativePathString() {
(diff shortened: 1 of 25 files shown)


fmeum:
@sgowroji Since I'm working on a bit of a larger change, I'm already using the CI to catch potential issues. There is no need to assign a reviewer yet as this would have to go through a higher-level design review first anyway.

src/main/java/com/google/devtools/build/lib/actions/PrecomputedPathMapper.java
```diff
@@ -0,0 +1,358 @@
...
+        throw new CommandLineExpansionException(
+            String.format("Failed to get metadata for %s", input), e);
+      }
+      Preconditions.checkNotNull(md);
+      byte[] digest = md.getDigest();
+      if (digest == null) {
+        return NOOP;
+      }
+      fp.addBytes(digest);
+      virtualConfigDigest = DigestUtils.xor(virtualConfigDigest, fp.digestAndReset());
```
fmeum, line 176: Note: This suffers from the issue explained in https://github.com/bazelbuild/bazel/issues/15660. It can easily be fixed by collecting all pairs in a set first and then hashing it.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/PathMapperApi.java
```diff
@@ -0,0 +1,64 @@
...
+import net.starlark.java.eval.StarlarkValue;
+
+@StarlarkBuiltin(
+    name = "PathMapper",
+    category = DocCategory.BUILTIN
+)
+public interface PathMapperApi extends StarlarkValue {
+
+  @StarlarkMethod(
+      name = "dirname",
```
gregestren, line 30: Where are these symbols used?

gregestren (commented):
Summarizing: you're adding Starlark hooks that Starlark rules can use to

a) declare path mapping
b) set up their command lines to work properly with it?

And you're adding the concept of "path mapping" over the existing concept of "path stripping", with two advantages that 

a) it works better on actions with mixed-config inputs
b) it materializes results on the local output tree to avoid path skew vs. the executor?

src/main/java/com/google/devtools/build/lib/analysis/actions/SpawnAction.java
```diff
@@ -403,32 +419,70 @@ public Spawn getSpawn(ActionExecutionContext actionExecutionContext)
...
+          pathMapper = PrecomputedPathMapper.createPerAction(artifactExpander, metadataHandler,
+              getInputs(), getPrimaryOutput());
+        } else {
+          pathMapper = PrecomputedPathMapper.createContentBased(artifactExpander, metadataHandler,
+              getInputs(), getPrimaryOutput());
+        }
+      }
+      if (executionInfo.containsKey(ExecutionRequirements.REQUIRES_MATERIALIZED_INPUT_PATHS)) {
+        try {
+          // TODO(fmeum): This is most likely not the right place to materialize the mapped paths:
```
gregestren, line 463: This is probably getting advanced, but does this work with virtual inputs, like .params files? See, for example, https://github.com/bazelbuild/bazel/blob/361ce673ad2b959b73e859a121ff2e996feb561b/src/main/java/com/google/devtools/build/lib/exec/ExecutionOptions.java#L123 which defaults to false.

More broadly, actions don't necessarily write outputs to the file system. Would this still try to symlink them?

src/main/java/com/google/devtools/build/lib/remote/common/RemotePathResolver.java
```diff
@@ -114,30 +125,36 @@ public void walkInputs(
...
     @Override
     public String localPathToOutputPath(PathFragment execPath) {
-      return execPath.getPathString();
+      return actionStager.strip(execPath).getPathString();
     }
 
     @Override
     public Path outputPathToLocalPath(String outputPath) {
-      return execRoot.getRelative(outputPath);
+      return execRoot.getRelative(actionStager.unstrip(PathFragment.create(outputPath)));
```
gregestren, line 146: This is for OutputPathsMode.strip?

fmeum:
I just had an idea that may obviate the need for new Starlark API. I will give it a try and update the PR if it works.

fmeum:
@gregestren @aranguyen As we discussed at BazelCon, I pushed a new version of this PR that is stripped down to the essentials. Most of the commits are the same, with the following changes compared to the previous version:
1. I removed the per-action path mapping scheme in favor of the simpler per-input, content-based scheme. This scheme should be good enough for all non-C++-like actions and has the added benefit that it also applies to actions that write file manifests (since input paths remain the same across actions).
2. It no longer requires new Starlark API. Instead, paths in `map_each` functions are mapped automatically by injecting a `PathMapper` into `StarlarkSemantics`. I found this to work very well in practice (rules_go). It also means that supporting this feature requires no code changes.
3. Staying in line with the "no code changes" required theme, I replaced the `execution_requirements` opt-in with a simple flag that lists mnemonics for easier experimentation.

You can try out the scheme on [this rules_go branch](https://github.com/bazelbuild/rules_go/tree/support-path-mapping) with this command line:
```
bazel build --experimental_content_based_path_mapping=GoCompilePkg,GoLink,GoToolchainBinary,JavaToolchainCompileBootClasspath,JavaToolchainCompileClasses,GoTestGenTest,GoInfo,GoStdlibList -- //... -//tests/core/cgo:opts -//tests/core/cgo:opts_test
```

As you can see, essentially all Go actions as well as some Bazel-internal Java actions can be opted in. The only test case that doesn't compile is one that depends on generated C headers: Since the native providers track include paths as hardcoded strings rather than `dirname`s of `File`s, they can't be remapped.

src/main/java/com/google/devtools/build/lib/actions/Artifact.java
```diff
@@ -563,12 +566,17 @@ public SpecialArtifact getParent() {
...
-  @Override
   public final String getDirname() {
     PathFragment parent = execPath.getParentDirectory();
     return (parent == null) ? "/" : parent.getSafePathString();
   }
 
+  @Override
+  public final String getDirnameForStarlark(StarlarkSemantics starlarkSemantics)
+      throws EvalException {
+    return starlarkSemantics.get(ArtifactPathMapper.STARLARK_SEMANTICS_KEY).getMappedDirname(this);
```
gregestren, line 577: Not sure I'm reading this right: is this providing the ability for `dirname` to resolve to mapped paths? How does that fit with

> must not contain `dirname`, ...

?

src/main/java/com/google/devtools/build/lib/actions/PathStripper.java
```diff
@@ -163,6 +215,60 @@ public interface CommandAdjuster {
      */
     List<String> stripCustomStarlarkArgs(List<String> args);
 
+    /**
+     * Creates a an {@link ArtifactPathMapper} that can be set on a Starlark thread's semantics
+     * object to automatically rewrite paths returned by the File API in map_each callbacks.
+     */
+    default ArtifactPathMapper toArtifactPathMapper() {
```
gregestren, line 222: I'm reading three meaningful abstractions now: `ArtifactPathMapper`, `PathMapper`, and `PathStripper`. I'm getting confused especially between the first two (sometimes I didn't understand a piece of code because I thought it was using a `PathMapper` vs. an `ArtifactPathMapper` or vice versa).

I'm not sure what the right answer is but I'd like to keep our abstractions as clear as we possibly can.

I'm also wondering if there's some way to better merge `PathMapper` and `PathStripper`. i.e. the latter is an instance of the former? And maybe avoid the current connections where `PathStripper` is broken down into distinct subcomponents `ActionStager` and `CommandAdjuster` and `PathMapper` re-combines them.

The idea behind the distinct subcomponents was distinct pieces of code were referencing each. Code that constructs command lines is concerned with `CommandAdjuster`. Code that stages actions is concerned with `ActionStager`. The thought was keeping each abstraction clear & focused for its intended consumer would be helpful for the API.

gregestren (commented):
Somewhat reviewed the abstractions (not so much their usages yet). I have enough comments already so I'll stop there for this iteration.

Also to sanity check:

1) Is the current wording of the PR description accurate?
2) How many of the old comments are still relevant? Can we close any as obsolete?

I'm mostly asking because it'll be really easy to get confused given the breadth of concepts here and I want to make sure we're all precisely on the same page for effective review.

src/main/java/com/google/devtools/build/lib/analysis/actions/SpawnAction.java
```diff
@@ -219,7 +227,8 @@ public SpawnAction(
       boolean executeUnconditionally,
       ExtraActionInfoSupplier extraActionInfoSupplier,
       Consumer<Pair<ActionExecutionContext, List<SpawnResult>>> resultConsumer,
-      boolean stripOutputPaths) {
+      boolean stripOutputPaths,
+      boolean applyContentBasedPathMapping) {
```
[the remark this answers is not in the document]
fmeum: Let's discuss the options we have tomorrow. I could see merging my flag into yours, but we need to figure out the right flag API for that.

katre:
I have no context on this. Is there anything I should review specifically that @gregestren and @aranguyen can't cover?

fmeum:
@katre No need to do anything, I simply can't override the default reviewers.

fmeum:
Superseded by other PRs.

PR #15552, opened 2022-05-23
Add `--output=files` mode to cquery

### LINKED ISSUE #8739: Obtaining output paths of the build artifacts
There seem to be no easy way of getting location of the artifacts built by bazel, which results in a lot of unnecessary complexity and potentially faulty assumptions in tooling that needs to access build artifacts.

Most difficulties arise from the fact that `bazel` adds various sub-folders in the output such as "external", "$platform_$arch_[stripped|opt]", "$target_name.runfiles", etc. which requires any tooling that needs to be able to find the location know information about the build target, environment and all `bazel` layout conventions.

Buck solved this problem by introducing `buck targets --show-output` [command](https://buck.build/command/targets.html) that prints location of each build target in the build output folder.

For example:
```
buck targets --show-output //java/com/myproject:binary
> //java/com/myproject:binary buck-out/gen/java/com/myproject/binary.apk
```

The only potential work-around that I was able to find was using `bazel aquery` as described [here](https://stackoverflow.com/questions/47859615/bazel-how-do-you-get-the-path-to-a-generated-file) but it's not easy to use.

It would be nice to have equivalent in bazel, is there anything that prevents us from exposing this information as part of `bazel query` command?

Can we have `bazel query mytarget --show-output` or other simple equivalent?

fmeum:
With the new output mode `--output=files`, cquery lists all files advertised by the matched targets in the currently requested output groups.

This new mode has the following advantages over `--output=starlark` combined with an appropriate handcrafted `--starlark:expr`:
* provides a canonical answer to the very common "Where are my build outputs?" question
* is more friendly to new users as it doesn't require knowing about providers and non-BUILD dialect Starlark
* takes the value of `--output_groups` into account
* stays as close to the logic for build summaries printed by `bazel build` as possible

Fixes #8739

RELNOTES: `cquery`'s new output mode [`--output=files`](https://bazel.build/docs/cquery#files-output) lists the output files of the targets matching the query. It takes the current value of `--output_groups` into account.


aiuto:
I am very hesitant about adding this since it seems to be just a convenience for what can be done with starlark mode. We are trying to reduce the size of bazel and move features to starlark not the other way around.  I can look at it later this week.

fmeum:
> I am very hesitant about adding this since it seems to be just a convenience for what can be done with starlark mode. We are trying to reduce the size of bazel and move features to starlark not the other way around. I can look at it later this week.

I generally agree with this position, but I think this case is a bit special: "Where is my build output?" is a very common question that is also quite relevant for developers that are not core Bazel users and just happen to interface with it. Think of an OSS project that needs to build another OSS project that uses Bazel and has to figure out the set of files it needs to import. If we have the barrier to entry somewhere at "knows about `DefaultInfo`", that may already be too high for people to comfortably build their OSS project with Bazel knowing that others less familiar with Bazel will have to answer such questions.

gregestren:
Disclaimer: I haven't reviewed this in any meaningful way yet.

Generally speaking I agree with @aiuto 's comment and default to a position of needing strong justification for new user-facing features.

I see @fmeum  that you appreciate these points and are trying to make such a justification for this change. Whatever we ultimately decide I think it's fair for us to evaluate your points on their merit.

pcjanzen:
A wise man once [said](https://docs.google.com/document/d/17snvmic26-QdGuwVw55Gl0oOufw9sCVuOAvHqGZJFr4/edit#):
>The major weakness of this model is that it makes paths completely [opaque](https://docs.google.com/document/d/1fZI7wHoaS-vJvZy9SBxaHPitIzXE_nL9v4sS4mErrG4/edit#heading=h.mgfh432qhe7s). This is a calculated compromise: it assumes new APIs can be developed to help users find the subset of outputs they care about, and that these APIs will be important no matter what path model Bazel uses.

gregestren:
Still support that statement 100%! 

I'll hold back from saying more until I've properly read through this, and how all the themes connect.

aiuto:
I think the value is not just from the output paths, but attaching a meaningful labeling to each. The output group name is generally good, but not everything produces that. Sometimes you have to go into the rule specific providers to understand the meaning of a file. Once that happens, we are back at writing rule specific starlark code.

gregestren:
Reviewed further. I support this, with some caveats.

- As @pcjanzen wisely pointed out, I've long supported a proper API for finding build outputs.
- That's because it's strongly in the dev team's interest to have flexibility to change output path patterns (`bazel-out/arm-fastbuild-<hash>-...`). This lets us focus on API and performance improvements freely. So builds can get faster and more cacheable at a faster pace.
- Historically, users rely on exact path patterns (like looking at what `bazel build` produced and copying the exact path into downstream tools). That's the complete opposite of what we want. The only alternative is a canonical API.
- This is important enough to be a) universal, b) easy to use, c) works out of the box
- I think that overcomes the "keep the API simple" concerns mentioned
- I never knew quite what that API would look like. This is a nice approach to it: leveraging existing concepts (`cquery`, output groups) clearly and I think effectively. And with room for further expansion
- I still want to be careful to support backwards compatible API expansion
- I agree `IdentifyingName`: `output` would be a more powerful API
- That requires API expansion (I suppose extending the Starlark calls that declare outputs). Sounds manageable but would require its own design
- For now, is it reasonable to output a raw list? Or a dictionary where every key is empty to offer forward-facing structure for consuming tools?
- I also agree more complex cases still require custom Starlark. But I think it's reasonable to offer a trivial API for the common case and custom Starlark for advanced cases
- Conceptually we could still implement the logic in Starlark, if there was some way to bundle that behind the canonical API. But I don't think we have the infrastructure for that. And, again, I think working cleanly, out of the box, without requiring build expertise is a powerful argument.

gregestren:
One more comment:

There's also some awkwardness with multi-target builds. I can imagine someone wanting to just replace their `bazel build ...` command with `bazel cquery ...`. But:

- Multiple targets all produce output in a big blog: you can't tell which output was from which target
- Syntax isn't interchangeable: `bazel build //:foo //:bar` -> `bazel cquery '//:foo+//:bar'`

site/en/docs/cquery.md
```diff
@@ -357,6 +357,23 @@ This option generates output as a Graphviz-compatible .dot file. See `query`'s
...
+</pre>
+
+This option prints a list of the output files produced by each target matched
+by the query similar to the list printed at the end of a `bazel build`
+invocation. The output contains only the files advertised in the requested
+output groups as determined by the
+[`--output_groups`](/reference/command-line-reference#flag--output_groups) flag.
+
+Note: The output of `bazel cquery --output=files //pkg:foo` contains the output
+files of `//pkg:foo` in *all* configurations that occur in the build (also see
```
gregestren, line 373: Which pattern do you mean here? When you build `//pkg:foo` and it has a transitive dependency on itself? Or configurations over multiple builds?

fmeum: I'm not sure I fully understood the `cquery` semantics, but I think that this would already show up in the following situation:
* `//pkg:foo` is any target
* `//pkg:bar` depends on `//pkg:foo` via a transition

Then, a simple `bazel cquery //pkg:foo` would return two results and thus also two sets of files when used with `--output=files`.

fmeum:
> * Syntax isn't interchangeable: `bazel build //:foo //:bar` -> `bazel cquery '//:foo+//:bar'`

In fact, even before working on this change, I have always wondered about why `bazel (c)query` doesn't support a strict superset of the target specification grammar supported by `bazel build`. How about supporting `bazel (c)query <expr 1> <expr 2> -- -<expr 3>` and simply turning it into `(<expr 1>) + (<expr 2>) - (<expr 3>)` in the command logic before handing it to the query engine? `bazel (c)query` currently errors when it receives more than one positional argument, so this change would be backwards compatible and - at least in my opinion - useful beyond the situation of the current PR.

fmeum:
> * That's because it's strongly in the dev team's interest to have flexibility to change output path patterns (`bazel-out/arm-fastbuild-<hash>-...`). This lets us focus on API and performance improvements freely. So builds can get faster and more cacheable at a faster pace.

With that aim, it is important to keep the following in mind: *If* content-based output paths are used to make builds more cacheable and *if* these paths are used throughout Bazel and not just on the level of `Spawn`s, then `cquery` would no longer be the right tool to show output paths since knowing the paths would require executing (or recalling) a partial build.

> * Historically, users rely on exact path patterns (like looking at what `bazel build` produced and copying the exact path into downstream tools). That's the complete opposite of what we want. The only alternative is a canonical API.

Fully agreed. It's not completely uncommon to see `ST` hashes on the build outputs of top-level targets, at which point relying on exact paths is no longer reliable.

> * I agree `IdentifyingName`: `output` would be a more powerful API

Could you explain what you envision as `IdentifyingName` here? Would it contain more information than just the target (or possibly owner) label?

> * For now, is it reasonable to output a raw list? Or a dictionary where every key is empty to offer forward-facing structure for consuming tools?

I don't have a strong opinion here. When in doubt, I would prefer an output format that plays well with other Unix tools, given that this command will be a common interface to the non-Bazel world. There is `jq` however and it wouldn't be too uncommon to require it (or commit to emitting canonically indented JSON, which could also be parsed with `grep` in a pinch).

> * Conceptually we could still implement the logic in Starlark, if there was some way to bundle that behind the canonical API. But I don't think we have the infrastructure for that. And, again, I think working cleanly, out of the box, without requiring build expertise is a powerful argument.

I could implement `--output=files` as a thin wrapper around `--output=starlark` pointing at a shipped Starlark file, but for that would have to:

1. Parse the value of the `--output_groups` option in Starlark, duplicating the parsing logic.
2. Duplicate the filtering logic used in `BuildResultPrinter`.

That would probably defeat the purpose of decoupling this logic from core Bazel.

fmeum:
> * That's because it's strongly in the dev team's interest to have flexibility to change output path patterns (`bazel-out/arm-fastbuild-<hash>-...`). This lets us focus on API and performance improvements freely. So builds can get faster and more cacheable at a faster pace.

With that aim, it is important to keep the following in mind: *If* content-based output paths are used to make builds more cacheable and *if* these paths are used throughout Bazel and not just on the level of `Spawn`s, then `cquery` would no longer be the right tool to show output paths since knowing the paths would require executing (or recalling) a partial build.

> * Historically, users rely on exact path patterns (like looking at what `bazel build` produced and copying the exact path into downstream tools). That's the complete opposite of what we want. The only alternative is a canonical API.

Fully agreed. It's not completely uncommon to see `ST` hashes on the build outputs of top-level targets, at which point relying on exact paths is no longer reliable.

> * I agree `IdentifyingName`: `output` would be a more powerful API

Could you explain what you envision as `IdentifyingName` here? Would it contain more information than just the target (or possibly owner) label?

> * For now, is it reasonable to output a raw list? Or a dictionary where every key is empty to offer forward-facing structure for consuming tools?

I don't have a strong opinion here. When in doubt, I would prefer an output format that plays well with other Unix tools, given that this command will be a common interface to the non-Bazel world. There is `jq` however and it wouldn't be too uncommon to require it (or commit to emitting canonically indented JSON, which could also be parsed with `grep` in a pinch).

> * Conceptually we could still implement the logic in Starlark, if there was some way to bundle that behind the canonical API. But I don't think we have the infrastructure for that. And, again, I think working cleanly, out of the box, without requiring build expertise is a powerful argument.

I could implement `--output=files` as a thin wrapper around `--output=starlark` pointing at a shipped Starlark file, but for that would have to:

1. Parse the value of the `--output_groups` option in Starlark, duplicating the parsing logic.
2. Duplicate the filtering logic used in `BuildResultPrinter`.

That would probably defeat the purpose of decoupling this logic from core Bazel.

src/test/shell/integration/configured_query_test.sh
```diff
@@ -1416,4 +1416,73 @@ EOF
   expect_not_log "QueryException"
 }
 
+function test_files_output_mode() {
```
gregestren, line 1419: We're trying to move tests into Java where possible.

Would the patterns in https://github.com/bazelbuild/bazel/blob/c3f20f00d43a4c3811137cb4740623e254e45b1e/src/test/java/com/google/devtools/build/lib/query2/cquery/BuildOutputFormatterCallbackTest.java#L120-L129 and related files be able to capture this coverage?

src/test/shell/integration/configured_query_test.sh
```diff
@@ -1416,4 +1416,73 @@ EOF
...
+  local -r pkg=$FUNCNAME
+  mkdir -p $pkg
+  cat > $pkg/rules.bzl <<'EOF'
+def _r_impl(ctx):
```
gregestren, line 1423: Just glancing, but does this cover dupe removal and input file skipping?

gregestren:
Broadly speaking, and after more discussion with @brandjon , I support the simplest syntax:

```
$ bazel cquery --output=files '//tt:arm' 
bazel-out/arm-fastbuild-ST-7d684b6a9cec/bin/tt/libcb.a
...
```
which you're already doing.

The more complex structures quickly get speculative, where the core need is just a simple, easy API. Any extra syntax we add now would make future API changes that much more complicated if we get it wrong. @brandjon had some interesting ideas about separating the content request (outputs vs. BUILD vs. ...) with the format (JSON vs. plain vs. ...).

Now's not the time to get caught up in the subtleties.

gregestren:
Stated differently, lacking more experience with this, I think it's fair to assume users know which targets they're requesting outputs for. So we don't have to preemptively add API features to contextualize.

src/main/java/com/google/devtools/build/lib/analysis/TopLevelArtifactHelper.java
```diff
@@ -236,6 +238,36 @@ static ArtifactsToBuild getAllArtifactsToBuild(
...
 
+  /**
+   * Returns false if the build outputs provided by the target should never be shown to users.
+   *
+   * <p>Always returns false for hidden rules and source file targets.
+   */
+  public static boolean shouldConsiderForDisplay(ConfiguredTarget configuredTarget) {
+    // TODO(bazel-team): this is quite ugly. Add a marker provider for this check.
+    if (configuredTarget instanceof InputFileConfiguredTarget) {
+      // Suppress display of source files (because we do no work to build them).
```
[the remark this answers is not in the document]
fmeum: No, the warnings are emitted directly by a skyframe event, see https://cs.opensource.google/bazel/bazel/+/54575fb306e67775c4f73187f1d255d9c2000727:src/main/java/com/google/devtools/build/lib/skyframe/TargetCompletor.java;l=147.

I considered for a moment whether `query` output *should* contain source files, but then decided against it as that may require traversing transitive structures. I will clarify in the documentation that the command lists *outputs*.

src/test/shell/integration/configured_query_test.sh
fmeum: Definitely, will redo the test and cover all the features. I just didn't want to write extensive tests for something that may not get merged, but now that this is viewed favorably, I'm going to test it thoroughly.

src/main/java/com/google/devtools/build/lib/analysis/TopLevelArtifactHelper.java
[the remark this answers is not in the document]
fmeum: Clarified this in the docs.

fmeum:
I think I addressed all comments, PTAL.

gregestren (commented):
Back online again!

Overall this looks great. Thanks so much - I'm thrilled to see this feature happen.

Just one more tiny but important nit: can you add an appropriate `RELNOTES` to the PR description? (https://bazel.build/contribute/release-notes). Definitely want to make sure this gets announced with the right release.

fmeum:
@bazel-io flag

gregestren (approved):

ckolli5:
Hey @fmeum, we are trying to [cherry pick](https://github.com/bazelbuild/bazel/pull/15955) this change to release-5.3.0. But presubmit checks are failing. Could you please help us in cherry-picking this PR to release-5.3.0? Thanks!

gregestren:
That might have been a merge discrepancy. I think it's just a matter of the right dependency in a BUILD file being added.

gregestren:
There's some kind of baseline discrepancy. 

For example, for `ProtoOutputFormatterCallback`, [this commit](https://github.com/bazelbuild/bazel/commit/681f534ea8d79747c2d8f53cabe26eadedd004bc#diff-4f1f5a5f095fc900af3c7a6ce9217da68db4c1d2b3b23644a0a38ff6acf2adaf
) only adjusts the line

    super(eventHandler, options, out, skyframeExecutor, accessor, /*uniquifyResults=*/ false);

But [the cherrypick](https://github.com/bazelbuild/bazel/pull/15955/files#diff-4f1f5a5f095fc900af3c7a6ce9217da68db4c1d2b3b23644a0a38ff6acf2adaf) also includes

     OutputType outputType,
     @Nullable TransitionFactory<RuleTransitionData> trimmingTransitionFactory) {

which was added by https://github.com/bazelbuild/bazel/commit/9994c3277607a3b7bd452512e1ff6e5ba73bbc4a.

aiuto:
It should usually be the job of the original PR author to create a PR that
backports a feature to the previous release.
The release team is not responsible for figuring out a complex PR when it
is on a single patch.

fmeum:
I submitted https://github.com/bazelbuild/bazel/pull/15979 with the merge conflicts resolved.

Issue #13639, opened 2021-07-02
Databinding generates non deterministic artifacts and breaks caching


arunkumar9t2:
### Description of the problem / feature request:

Few databinding generated artifacts are non deterministic and leads to reduced build performance in both incremental and cached scenario. 

1. **baseClassSrc.srcjar**/**class-info.zip** -  The zip is generated by databinding CLI and it currently does not handle zip time stamps. Manually fixing timestamp in databinding exec compiler fixes the issue.
2. **layout-info.zip** - This zip also has the timestamp problem + absolute path encoded in the generated xmls inside the zip file. For example,
    ```xml
     <?xml version="1.0" encoding="UTF-8"?>
    <Layout layout="activity_main" modulePackage="com.google.android.databinding.libA" 
   absoluteFilePath="/private/var/tmp/_bazel_<redacted>/3c972a5b37390c53dbacb3ca696a0a81/sandbox/darwin- 
    sandbox/162/execroot/__main__/src/libA/res/layout/activity_main.xml" directory="layout" isMerge="false">
   <Variables declared="true" type="int" name="visibility">
      <location startLine="4" startOffset="8" endLine="6" endOffset="24" />
   </Variables>
   <Targets>
   </Targets>
</Layout>
 

Because of this both artifacts are non deterministic and influence build performance significantly.

### Feature requests: what underlying problem are you trying to solve with this feature?

Cross machine caching and incremental build performance.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
[databinding_caching.zip](https://github.com/bazelbuild/bazel/files/6756313/databinding_caching.zip)

Run 
```bash
bazelisk build A
zipinfo bazel-bin/databinding/A/class-info.zip
zipinfo bazel-bin/databinding/A/layout-info.zip
zipinfo bazel-bin/databinding/A/baseClassSrc.srcjar
# Notice time stamp embedded in both of the artifacts

unzip bazel-bin/databinding/A/layout-info.zip -d /tmp/
cat /tmp/activity_main-layout.xml

# Observe absolute path containing /private/var/tmp
```

### What operating system are you running Bazel on?

MacOs Catalina 10.15.7

### What's the output of `bazel info release`?

release 4.1.0

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

NA

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

NA

[...]

PR #15567, opened 2022-05-24
Ensure Databinding `layout-info.zip` entries have consistent timestamps

arunkumar9t2:
This is part of https://github.com/bazelbuild/bazel/issues/13639 and fixes the timestamp issue which makes `layout-info.zip` non deterministic. 

Fixed tests. 

Note that absolute path issue described [here](https://github.com/bazelbuild/bazel/issues/13639#issue-935986425) still exists and that needs fixes in databinding jar.

### DIFF
--- a/src/test/java/com/google/devtools/build/android/AndroidDataBindingProcessingActionTest.java
+++ b/src/test/java/com/google/devtools/build/android/AndroidDataBindingProcessingActionTest.java
@@ -16,24 +16,28 @@ package com.google.devtools.build.android;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.io.MoreFiles;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Collections;
-import java.util.List;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipFile;
+import com.google.common.io.RecursiveDeleteOption;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


sgowroji:
@arunkumar9t2, Can you please check build failures. Thank you!

arunkumar9t2:
Hi @sgowroji I am working on fixing the build failures on Windows.

arunkumar9t2:
Hi @ahumesky, windows test was failing due to different permission issue which I was not able to workaround hence disabled it for now on windows.

src/tools/android/java/com/google/devtools/build/android/AndroidResourceProcessor.java
```diff
@@ -434,7 +435,8 @@ static Path processDataBindings(
 
     ProcessXmlOptions options = new ProcessXmlOptions();
     options.setAppId(packagePath);
-    options.setResInput(inputResourcesDir.toFile());
+    // Passing absolute path is required for databinding to write relative paths in layout-info.zip
+    options.setResInput(inputResourcesDir.toFile().getAbsoluteFile());
```
ahumesky, line 439: This requires changes to the databinding tools, correct? ("LayoutXmlProcessor.java" mentioned in #13639)
Let's separate this fix, because updating the databinding dependencies can take a fair bit of effort.

ahumesky (changes requested):

src/tools/android/java/com/google/devtools/build/android/AndroidResourceProcessor.java
arunkumar9t2: Yes correct, I think the change is backwards compatible but let's fix is separately as it makes sense.

arunkumar9t2:
Seems addressed by https://github.com/bazelbuild/bazel/commit/10af780af88b794c1497e771c6d8bab925297c3c so closing this. However the https://github.com/bazelbuild/bazel/issues/13639 absolute path issue still remains

PR #15569, opened 2022-05-24
Add missing word to error code.

benjaminp:


benjaminp:
I'm assuming these constants are safe to change?

haxorz:
> I'm assuming these constants are safe to change?

Thanks for posing this question. I am pretty sure it is safe to change them. @michajlo , is it safe? If so, consider making that explicit in a code comment in `failure_details.proto`.

Issue #15571, opened 2022-05-25
Reduce dependencies for generated python wrapper in common cases.


adam-azarchs:
### Description of the feature request:

The [python stub template](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt) used for `py_binary` and `py_test` imports a bunch of things which are not used in most common situations.  I would like for those to only be imported as necessary.

Specifically,
```python
import shutil
import tempfile
import zipfile
```
are only used if the template parameter `%is_zipfile%` is `True`.

`subprocess` is _mostly_ only used when running from a zip archive, but is also used on Windows or if `%enable_host_version_warning%` is `True`, which is maybe a sufficiently complex set of conditions to not warrant separating it out, since `subprocess` is present in `python3-minimal`.

It's also true that `re` is only used when _not_ running from a zipfile, but again it's part of `python3-minimal` so it's not as much of an issue as far as I'm concerned.

### What underlying problem are you trying to solve with this feature?

Python's processing of imports can take a non-negligible amount of time, when multiplied by thousands of tests in a large repo.  More to the point, especially in remote build contexts it would be nice to be able to use e.g. `python3-minimal` rather than the full `python3`, for smaller container images, which this change would permit.

### Which operating system are you running Bazel on?

Linux

### What is the output of `bazel info release`?

release 5.1.1

### Have you found anything relevant by searching the web?

See also https://github.com/bazelbuild/rules_python/issues/673.

This would also be mostly solved with https://github.com/bazelbuild/rules_python/issues/42, but that is more complex to implement than simply putting an `if` guard in front of a few of a few `import` statements, and wouldn't address the performance cost of processing those additional imports.

PR #15572, opened 2022-05-25
Reduce imports processed by python stubs.

adam-azarchs:
There are several import statements in the python wrapper stub used by
`py_binary` and `py_test` rules which are only used when running from a
zip archive.  Processing these imports is a waste of time in the common
case where they aren't going to  be used.  They also might not be
present at all in some stripped-down python environments (e.g. the
Debian `python3-minimal` package), which might be the only python
available on a remote builder when using a hermetic python toolchain
(even when using a hermetic python toolchain, the wrapper is executed
using the system python environment).

Closes https://github.com/bazelbuild/bazel/issues/15571


comius:
cc @brandjon do you see any problems with this?

comius (approved):

Issue #3415, opened 2017-07-19
Update documentation on deprecating/removing --android_cpu


chaoren:
Please provide the following information. The more we know about your system and use case, the more easily and likely we can help.

### Description of the problem / feature request / question:

_If asking a question or requesting a feature, also tell us about the underlying problem you're trying to solve._

`--android_cpu` appears to have no effect, while `--fat_apk_cpu` seems to do the job.

### If possible, provide a minimal example to reproduce the problem:

```
$ bazel mobile-install :binary
stderr: Failed to install ./bazel-out/local-fastbuild/bin/binary_incremental.apk: Failure [INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]
# fails as expected

$ bazel mobile-install :binary --fat_apk_cpu=x86
# success

$ bazel mobile-install :binary --android_cpu=x86
stderr: Failed to install ./bazel-out/local-fastbuild/bin/binary_incremental.apk: Failure [INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]
# still fails, not expected

$ bazel mobile-install :binary --android_cpu=foo
stderr: Failed to install ./bazel-out/local-fastbuild/bin/binary_incremental.apk: Failure [INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]
# bogus cpu name not detected

$ bazel mobile-install :binary --android_cpu=x86 --android_compiler=foo
ERROR: No toolchain found for --cpu='armeabi-v7a' --compiler='foo'. Valid toolchains are: [
# cpu value is still armeabi-v7a for some reason

$ bazel mobile-install :binary --fat_apk_cpu=x86 --android_compiler=foo
ERROR: No toolchain found for --cpu='x86' --compiler='foo'. Valid toolchains are: [
# doesn't happen with fat_apk_cpu
```

### Environment info
* Operating System: Ubuntu 14.04.5 LTS

* Bazel version (output of `bazel info release`): release 0.5.2

* If `bazel info release` returns "development version" or "(@non-git)", please tell us what source tree you compiled Bazel from; git commit hash is appreciated (`git rev-parse HEAD`):

###  Have you found anything relevant by searching the web?
(e.g. [StackOverflow answers](http://stackoverflow.com/questions/tagged/bazel),
[GitHub issues](https://github.com/bazelbuild/bazel/issues),
[...]

PR #15583, opened 2022-05-26
supplement android_cpu flag documentation

sugarmanz:
Add additional information captured in #3415 to flag documentation.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
@@ -230,7 +230,9 @@ public class AndroidConfiguration extends Fragment implements AndroidConfigurati
           OptionEffectTag.LOADING_AND_ANALYSIS,
           OptionEffectTag.LOSES_INCREMENTAL_STATE,
         },
-        help = "The Android target CPU.")
+        metadataTags = {OptionMetadataTag.DEPRECATED},
+        help = "The Android target CPU. Will not have an effect if --fat_apk_cpu has a value, "
+                + "even a default one. It is recommended to just use --fat_apk_cpu.")
     public String cpu;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
```diff
@@ -230,7 +230,9 @@ public static class Options extends FragmentOptions {
           OptionEffectTag.LOADING_AND_ANALYSIS,
           OptionEffectTag.LOSES_INCREMENTAL_STATE,
         },
-        help = "The Android target CPU.")
+        metadataTags = {OptionMetadataTag.DEPRECATED},
+        help = "The Android target CPU. Will not have an effect if --fat_apk_cpu has a value, "
+                + "even a default one. It is recommended to just use --fat_apk_cpu.")
```
aiuto, line 235: We should not recommend fat_apk_cpu, as we will be removing it sometime this year in favor of using --android_platforms.
"even a default one" doesn't make sense.  fat_apk_cpu has a default, so that would mean android_cpu does nothing.

The safest correct thing is to just stop after "... --fat_apk_cpu has a value."

aiuto (changes requested):
I should not be the approver on this. I just spotted the problems while glancing at other things. 
@ahumesky @katre

src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
katre: +1 to this comment.

katre: This is fair. I'm going to leave it to @ahumesky to be the decider: what documentation should we have on a flag that will be obsolete in the future?

ted-xie:
We've deprecated --android_cpu in favor of --android_platforms, so I'm going to close this PR.

Issue #8446, opened 2019-05-23
Don't require a system-installed Python interpreter for bootstrapping


brandjon:
Forked from discussion in #7947. On windows, we can have the native launcher take on the [responsibilities](https://github.com/bazelbuild/bazel/issues/7947#issuecomment-494981489) of the stub script, in particular extracting the runfiles. This means it can use any Python runtime to launch the user code, including a runtime in runfiles.

We should also consider creating a native launcher for non-Windows platforms to do the same, so we don't need the `#!/usr/bin/env python` shebang.

Finally, we can consider merging the zip file into the launch to create a self-extracting native executable.

abergmeier:
> We should also consider creating a native launcher for non-Windows platforms to do the same, so we don't need the #!/usr/bin/env python shebang.

Can you write up a proposal for goals and rough ideas?

PR #15590, opened 2022-05-28
rules/python: Add a `coverage_tool` attribute to `py_runtime`.

adam-azarchs:
This allows users to specify a target providing the coveragepy tool (and its dependencies).  This is essential for hermetic python builds, where an absolute path will not really work.  It's also superior to other potential methods using environment variables because the runfiles dependency on the coverage tool and its files is only incurred when building with coverage enabled.

This also builds on the work @TLATER began with https://github.com/bazelbuild/bazel/pull/14677 to integrate with `coveragepy`'s `lcov` support, with an additional step of at least attempting to convert the absolute paths which `coveragepy` uses in the lcov output into the relative paths which the rest of bazel can actually consume.

This is my first time touching Java code professionally, so I'll admit to mostly cargo-culting those parts, and would welcome any feedback on how to improve things there.  I also would have no objections to someone else taking over this PR to get it over the finish line.  I've tested this out with our own team's internal monorepo, and have successfully generated a full combined coverage report for most of our python and go code.  There's still a bunch of things which don't quite work, in particular when it comes to compiled extension modules or executables run from within python tests, but those will need to be addressed separately, and this is already a giant step forward for our team.

Closes https://github.com/bazelbuild/bazel/issues/14436.


src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -332,54 +455,55 @@ def Main():
...
-  if cov_tool:
-    # Inhibit infinite recursion:
-    del os.environ['PYTHON_COVERAGE']
-    if not os.path.exists(cov_tool):
+  # COVERAGE_DIR is set iff the instrumentation is configured for the
+  # file and coverage is enabled.
+  if os.environ.get('COVERAGE_DIR'):
+    cov_tool = FindCoverageTool(module_space)
+    if not cov_tool:
+      if 'PYTHON_COVERAGE_TARGET' in os.environ:
```
TLATER, line 463: We should probably get rid of this particular hack if we're going to allow providing the coverage tool through the runtime. Having that path forwards is sufficiently nice that, assuming we get proper support from the python rules, it can just supersede the old behavior. We'll probably want to remove the old env variable in the long term.

The current implementation has some annoying edge cases, and nobody has yet suggested an alternative implementation that wouldn't. It's a footgun and introduces long-term tech debt. In fact, this hack is why my old PR was a draft.

My commits should work fine with your changes if you simply rebase on top of 72a0f9ed13e4b1f561aee4268e4919b84ac04395 instead (as an aside, avoid making merge commits on a feature branch :) ).

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -352,12 +377,21 @@ def ExecuteFile(python_program, main_filename, args, env, module_space,
...
       cwd=workspace
     )
-    output_filename = os.environ.get('COVERAGE_DIR') + '/pylcov.dat'
+    output_filename = os.environ.get('COVERAGE_DIR') + '/pylcov.dat.tmp'
     ret_code = subprocess.call(
       [python_program, coverage_tool, "lcov", "-o", output_filename] + args,
       env=env,
       cwd=workspace
     )
+    # Coveragepy uses absolute file names in the output.  We need to strip off
```
TLATER, line 386: An alternative may be asking coveragepy to report relative paths. This issue suggests there already is some sort of support for that, albeit incomplete: https://github.com/nedbat/coveragepy/issues/963

adam-azarchs: That would be preferable, though I'd note that first of all "relative to what?" still applies; we'd probably still need to strip off at least the workspace name.  This approach here was mostly a "it's Friday evening and I just want to see if I can get a useful coverage report before going to bed" solution - I'm definitely not thrilled with it, but it does work.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
adam-azarchs: Merge commits have the advantage of retaining proper attribution for the original commits.  I'd probably rebase/squash before promoting this from a draft in any case.

Yes, adding this `PYTHON_COVERAGE_TARGET` environment variable is unnecessary (and probably a bad idea to introduce)  - I was mainly just trying to reduce the diff between your changes and mine for now.  Thus, still a draft :-).  I think `PYTHON_COVERAGE` needs to stay for backwards compatibility but longer-term I think maybe the solution there would be to have a `coverage_tool_path` option on `py_runtime` similar to how there's an `interpreter_path` option for non-hermetic python toolchains.  That said, I'm not particularly interested in working on features to support non-hermetic toolchains, since I don't use them.  Apparently Google does?  But they have a lot more engineering time to spare than we do.

TLATER: > Merge commits have the advantage of retaining proper attribution for the original commits.

So do rebases, as long as you don't need to squash, which I assumed you wouldn't need to, but ignore me; obviously you would have, it touches the same code. I should do more than glance over things, but as you say later, Friday evening (well past midnight, too, in my case).

> That said, I'm not particularly interested in working on features to support non-hermetic toolchains, since I don't use them. Apparently Google does?

As far as I can tell, most, or at least a significant portion of, organizations that use Bazel still use non-hermetic toolchains - just look through public Bazel repositories. There are a lot of users that haven't yet adopted toolchains, and for which changing that would be a significant amount of effort. For the time being we'll have to play nice.

Either way, changes as-is look pretty ok then, treat this comment as a "please remove that commit before you resolve draft status", and thanks for taking this all forward :)

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
adam-azarchs: The problem here is that when it resolves the symlinks in a runfiles tree then it ends up getting a path in either the original source tree or `$(bazel info execution_root)/{external,bazel-out}`, none of which are are useful when made relative to the current working directory.  The problem with my hacky solution to this problem is that it will only work for one of the above roots.

From my read of the code (I haven't tried it, yet) it _does_ look like if coveragepy is [configured to use relative paths](https://github.com/nedbat/coveragepy/blob/3a83e41a3668aaf8d6d9f913c5a4506c36346152/coverage/control.py#L290) then it doesn't [resolve symlinks](https://github.com/nedbat/coveragepy/blob/3a83e41a3668aaf8d6d9f913c5a4506c36346152/coverage/files.py#L148), which would fix the problem here.  The problem is that behavior seems to only be controllable through the config file rather than a command line flag, which can be tricky to get working properly in a bazel sandbox environment.  Various solutions to that problem:
1. Generate a config file in a temp directory for each test invocation.
2. Include a config file probably as a target somewhere in `bazel_tools` so it could be shared by all test instantiations
3. Add a way for the user to add specify a config file in toolchain configuration.  This has some advantages, but is also prone to causing problems.

I don't love any of those but would probably go with the first approach.  I'll try it out on Tuesday.

adam-azarchs (commented):
Looks like I'll need to fix up a bunch of unit tests.  A bit annoying because `bazel test //...` doesn't actually pass for me on master, which isn't too surprising given all of the various configuration and edge cases which it's testing, but I should still be able to track down everything that needs updating.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
adam-azarchs: Unfortunately it seems like coveragepy's relative_paths feature doesn't actually work, as it still canonicalizes paths in the lcov report (as that bug you link pointed out), though it does at least seem to use relative paths in the intermediate data and xml output.  That said, there _is_ a better way to do what I was doing, which I'll push up here once I've tested it properly.

adam-azarchs:
Well, in its current state this works fine for our internal setup, which uses a hermetic python but with additional python packages installed (as required) into that python's search path, rather than some separate `PYTHONPATH` entry.  Unfortunately when I tried setting up a test that would work the way most people are more likely to have things set up, it doesn't work so well:

1. The `pip_install` rule depends on having a configured python toolchain, which results in a circular dependency if you try to use it to fetch the `coverage` package.
2. You can fetch coverage with e.g.
```starlark
http_archive(
    name = "coverage_linux_x64",
    build_file_content = """
filegroup(
    name = "coverage",
    srcs = ["coverage/__main__.py"],
    data = glob(["coverage/**"]),
)
""",
    sha256 = "84631e81dd053e8a0d4967cedab6db94345f1c36107c71698f746cb2636c63e3",
    type = "zip",
    urls = [
        "https://files.pythonhosted.org/packages/74/0d/0f3c522312fd27c32e1abe2fb5c323b583a5c108daf2c26d6e8dfdd5a105/coverage-6.4.1-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl",
    ],
)
```
The problem with that being that you get
```
  File "/.../my_test.runfiles/coverage_linux_x64/coverage/__main__.py", line 7, in <module>
    from coverage.cmdline import main
ModuleNotFoundError: No module named 'coverage'
```

The obvious solution would be to check if the `coverage_tool` attribute has a `PyInfo` provider and, if so, all it to augment the `PYTHONPATH` (e.g. with the `imports` attribute on `py_library`).  There are two problems with that approach:
1. Ideally we wouldn't be adding stuff to the test's `PYTHONPATH` for this.  But I suppose that's unavoidable.
2. `py_library` doesn't have a way to declare an executable entry point.  So we'd need to either declare a separate attribute on `py_runtime` for the entry point, or execute it as `python -m coverage ...`.  The latter has some advantages, but also some significant disadvantages.  If we prepend the PYTHONPATH for `coverage`, we might cause unexpected behavior for the package under test, and if we append it to the end then we might not get the `coverage` tool we expect (though, maybe packages which put a module named `coverage` into the global namespace kind of deserve broken behavior).

Using `py_binary` for this would obviously be a different kind of problematic.

adam-azarchs:
Never mind, the problem is a different one: it's not including the runfiles for `coverage_tool`, though it needs to.

adam-azarchs:
Ok, with the addition of an integration test, I now feel like this is about ready to transition out of draft status.  @TLATER would appreciate it if you have any further feedback to offer before I do that.

src/test/shell/bazel/bazel_coverage_hermetic_py_test.sh
```diff
@@ -0,0 +1,183 @@
...
+# Fetch hermetic python and register toolchain.
+function set_up() {
+    cat >>WORKSPACE <<EOF
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_python",
+    sha256 = "cdf6b84084aad8f10bf20b46b77cb48d83c319ebe6458a18e9d2cebf57807cdd",
+    strip_prefix = "rules_python-0.8.1",
+    url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
```
adam-azarchs, line 33: Fetching this appears to fail on the CI builders, presumably because they're blocking network access for tests.  I'm not really sure how to get around that; without fetching a toolchain from a remote, I guess either those things need to be vendored in to the bazel tree or it's not a hermetic build...

TLATER: I think common practice is to skip the tests for these kind of use cases. An alternative could be mocking all of rules_python, but I don't think that's particularly feasible.

That said, I'm far from a Bazel maintainer, and have the same opinion as you do about not adding tests for these kinds of things.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -281,6 +294,137 @@ def Deduplicate(items):
...
+    unfixed_file = output_filename + '.tmp'
+    os.rename(output_filename, unfixed_file)
+    with open(unfixed_file, "r") as unfixed:
+      with open(output_filename, "w") as output_file:
+        for line in unfixed:
+          if line.startswith('SF:'):
+            for (realpath, filename) in substitutions:
+              line = line.replace(realpath, filename)
+          output_file.write(line)
+    os.unlink(unfixed_file)
```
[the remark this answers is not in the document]
TLATER: I'd still prefer an upstream fix, but this seems better than the previous solution for now.

TLATER (commented):
Overall looks pretty good to me, from what I remember of my foray into this part of Bazel all those months ago. Obviously still needs review from someone with actual authority around here, though :)

comius:
cc @c-mita

adam-azarchs:
I'll need to rebase this due to conflicts.  I think once I'm doing that, I might actually split this PR into two, one to deal with getting pycoverage to actually produce useful results (building off of @TLATER 's change and adding in the bits for fixing up the paths, which we're going to continue to need until that part is fixed upstream) and a second change to add the attribute to `py_runtime`, since the later represents an API change, and for which it is apparently a bit tricky to write a test that will work on bazel's CI infrastructure, which currently seems to have little to no testing for properly hermetic python toolchains.  However I'm on PTO this week so it'll have to wait.

src/test/shell/bazel/bazel_coverage_hermetic_py_test.sh
c-mita: I don't think it's feasible to keep these tests; the automated testing framework cannot run them.
All we can reasonably test is that a provided coverage tool given to py_runtime is used in the way that is expected; a dummy tool could be created and provided for the integration tests.

adam-azarchs:
Yes, sorry I just haven't had a chance to circle back around to this.  I'll fix the conflicts and then see what I can do about the tests.

adam-azarchs:
As was recommended, I updated the unit test to use a mock coverage script.  After fixing various issues with other platforms, tests are now passing.  I can clean up the commit log if anyone wishes me to do so.

comius:
@c-mita, @rickeylev can you please review?

rickeylev:
I'll have a look.
(coincidentally, we made essentially this same change internally a couple months ago, but since we don't use toolchains and do embedded hermetic python, our impl looks entirely different; same basic problems encountered and solved though)

site/en/configure/coverage.md
```diff
@@ -188,8 +188,60 @@ py_test(
...
+configuration.
+
+Because the `pip_install` rule depends on the python toolchain, you cannot use
+it to fetch the `coverage` module.  Instead, add in your `WORKSPACE` e.g.
+
+```starlark
+http_archive(
+    name = "coverage_linux_x86_64"",
+    build_file_content = """
+filegroup(
```
rickeylev, line 203: Rather than everyone copy/pasting this same filegroup target, it'd be better if there was a simple, shared, function people could use. That it's a filegroup, and the details of what's in the file group, are implementation details of the coverage library itself; users shouldn't have to worry about that.

site/en/configure/coverage.md
```diff
@@ -188,8 +188,60 @@ py_test(
     ],
 )
 ```
-<!-- TODO: Allow specifying a target for `PYTHON_COVERAGE`, instead of having to use `$(location)` -->
 
+If you are using a hermetic python toolchain, instead of adding an attribute to
+every `py_test` rule you can instead add the coverage tool to the toolchain
+configuration.
+
+Because the `pip_install` rule depends on the python toolchain, you cannot use
```
rickeylev, line 196: I'm not familiar with pip_install, can you explain more?
Maybe this is avoidable with some adjustments somewhere.

rickeylev (changes requested):
FYI: I probably won't be able to give this another deep look until after Sept 1st

site/en/configure/coverage.md
adam-azarchs: I could add a link to the docs for that rule.  But `pip` uses `python`, and it finds it via the toolchain, but it can't do that if you haven't configured the python toolchain yet.

site/en/configure/coverage.md
adam-azarchs: Agreed, in general.  However those details will also depend on the details of how one is obtaining the coverage library.  The example I put here is what I'd expect would be one of the more common ways people might obtain it, but it is _not_ how our internal teams would do it.  Some teams might for example vendor it into their workspace, others might get it from a local repository rooted in `/usr/lib/python3.x/site-packages` or whatever.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -213,6 +226,136 @@ def Deduplicate(items):
...
+          file=sys.stderr)
+        continue
+      if realpath != filename:
+        print("Fixing up {} -> {}".format(realpath, filename), file=sys.stderr)
+        yield (realpath, filename)
+
+def UnresolveSymlinks(output_filename):
+  """Replace realpath of instrumented files with the relative path in the lcov output.
+  
+  Though we are asking coveragepy to use relative file names, current
```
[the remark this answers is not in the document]
adam-azarchs: Yes, will fix.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -213,6 +226,136 @@ def Deduplicate(items):
...
+  ignore that for purposes of generating the lcov report, so we need to go
```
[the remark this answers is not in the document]
adam-azarchs: Yes, this is working around a bug in coveragepy. I'll look up the link to the issue.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -213,6 +226,136 @@ def Deduplicate(items):
...
+    main_filename: The python file to execute
+    args: Additional args to pass to the python file
+    env: A dict of environment variables to set for the execution
+    module_space: The module space/runfiles tree
+    coverage_tool: The coverage tool to execute with
+    workspace: The workspace to execute in. This is expected to be a
+               directory under the runfiles tree, and will recursively
+               delete the runfiles directory if set.
+  """
+  # We want to use os.execv instead of subprocess.call, which causes
```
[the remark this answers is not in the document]
adam-azarchs: I didn't actually write this doc.  I think it got brought back in by a bad merge, actually.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -213,6 +226,136 @@ def Deduplicate(items):
...
+    with open(rcfile_name, "w") as rcfile:
+      rcfile.write('''[run]
+relative_files = True
+''')
+    # Coveragepy wants to frst create a .coverage database file, from
+    # which we can then export lcov.
+    ret_code = subprocess.call(
+      [
+        python_program,
+        coverage_tool,
```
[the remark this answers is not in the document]
adam-azarchs: This is equivalent to running the `coverage` command directly (due to the `#!` line) except it's guaranteeing that we're running with the right python interpreter.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -269,47 +412,50 @@ def Main():
...
-    # entry point is coverage.coverage_main, so we need to do twice the dirname.
-    new_env['PYTHONPATH'] = \
-        new_env['PYTHONPATH'] + ':' + os.path.dirname(os.path.dirname(cov_tool))
-    new_env['PYTHON_LCOV_FILE'] = os.environ.get('COVERAGE_DIR') + '/pylcov.dat'
+  # COVERAGE_DIR is set iff the instrumentation is configured for the
+  # file and coverage is enabled.
+  if os.environ.get('COVERAGE_DIR'):
+    cov_tool = FindCoverageTool(module_space)
+    if cov_tool is None:
+      print('Coverage was requested, but no python coverage tool was configured.', file=sys.stderr)
```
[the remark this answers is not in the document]
adam-azarchs: Making `bazel coverage //...` fail when no python coverage tool has been configured would be a breaking change, which I wanted to avoid.  The other case is for when you think you've enabled python coverage but didn't do so correctly.

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -269,47 +412,50 @@ def Main():
...
+    else:
+      # Inhibit infinite recursion:
+      if 'PYTHON_COVERAGE' in os.environ:
+        del os.environ['PYTHON_COVERAGE']
+
+      if not os.path.exists(cov_tool):
+        raise EnvironmentError('Python coverage tool %s not found.' % cov_tool)
```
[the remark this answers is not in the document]
adam-azarchs: That would definitely have been useful to know about when I was wrestling with this...  Will add!

src/main/java/com/google/devtools/build/lib/rules/python/PyRuntime.java
```diff
@@ -61,6 +66,24 @@ public ConfiguredTarget create(RuleContext ruleContext)
       ruleContext.attributeError("interpreter_path", "must be an absolute path.");
     }
 
+    Artifact coverageTool = null;
+    NestedSet<Artifact> coverageFiles = null;
+    if (coverageTarget != null) {
+      FilesToRunProvider filesToRun = coverageTarget.getProvider(FilesToRunProvider.class);
```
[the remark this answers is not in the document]
adam-azarchs: I also have other deadlines...

Using `filegroup` in the example docs was a "quick fix" but I think in most real-life situations that's not actually what people will end up using.  `sh_binary` might have been more appropriate.

site/en/configure/coverage.md
```diff
@@ -188,8 +188,61 @@ py_test(
...
+    ],
+)
+```
+
+Then configure your python toolchain as e.g.
+
+```starlark
+py_runtime(
+    name = "py3_runtime_linux_x86_64",
+    coverage_tool = "@coverage_linux_x86_64//:coverage",
```
rickeylev, line 224: There is probably an optimization to be made somewhere in here to only have the coverage dependency when coverage is enabled. The isCodeCoverage() check in the Java code prevents it from being included at runtime, but it still causes Bazel to do various target and config resolution steps.

Coverage being enabled can be detected with a config setting, e.g.

```
config_setting(
    name = "coverage_enabled",
    values = {"collect_code_coverage": "true"},
)

py_library(
    name = "coverage_if_enabled",
    deps = select({
        ":coverage_enabled": [
            ":coverage",
        ],
        "//conditions:default": [],
    }),
)
```

I'm not sure how that fits into toolchain registration; can selects be used within WORKSPACE?

In any case, I'm find with figuring this out in a separate PR.

rickeylev (changes requested):

site/en/configure/coverage.md
adam-azarchs: Declaration of the toolchain is in a BUILD file, not the WORKSPACE.  However, I am not sure either about how `select` gets evaluated in toolchain resolution.  As I commented elsewhere I think eventually this bit of the doc should be just baked into `python_register_toolchains` (in `rules_python`) so that most users don't have to care about it.

src/main/java/com/google/devtools/build/lib/rules/python/PyRuntime.java
```diff
@@ -68,15 +67,23 @@ public ConfiguredTarget create(RuleContext ruleContext)
...
-      FilesToRunProvider filesToRun = coverageTarget.getProvider(FilesToRunProvider.class);
-      if (filesToRun == null) {
-        ruleContext.attributeError("coverage_tool", "must be an executable target.");
+      NestedSet<Artifact> coverageToolFiles =
+          coverageTarget.getProvider(FileProvider.class).getFilesToBuild();
+      if (coverageToolFiles.isSingleton()) {
+        coverageTool = coverageToolFiles.getSingleton();
       } else {
-        coverageTool = filesToRun.getExecutable();
+        FilesToRunProvider filesToRun = coverageTarget.getProvider(FilesToRunProvider.class);
```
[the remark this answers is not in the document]
adam-azarchs: Perhaps more generally applicable, requiring `PyInfo` and a single file, without allowing an executable target, would rule out using a `py_binary` target here, as that doesn't produce a `PyInfo` and its output contains at least two files (including the stub).

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -356,6 +333,74 @@ relative_files = True
...
+      coverage_entrypoint,
+      "lcov",
+      "--rcfile=" + rcfile_name,
+      "-o",
+      output_filename
+    ],
+    env=env,
+    cwd=workspace
+  ) or ret_code
+  os.unlink(rcfile_name)
```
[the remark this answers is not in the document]
rickeylev: > Do you think it's sufficient to just ignore `OsError` here (with a comment)?

Yes.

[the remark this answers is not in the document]
adam-azarchs: done.

src/main/java/com/google/devtools/build/lib/rules/python/PyRuntime.java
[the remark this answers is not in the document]
rickeylev: I appreciate the honesty =).

-----

re: using a py_binary here: I don't think that's even possible -- a py_binary depends on the toolchain, so putting a py_binary as a dependency of the toolchain will result in circular dependencies.

(aside: I think there's a variety of current or potential circular dependency/bootstrapping issues we need to figure out; let's take that convo outside this PR though)

re: running `coverage run main.py` instead of `python coverage.py run main.py`: I don't think it's as far fetched as you might think :). But let's dive into that outside this PR; it's enabling that sort of thing I'd like to bring to the Bazel Python rules.

-----

I'm willing to accept an executable here because I can appreciate your situation, and as long as we make sure that (via docs, in this case), whatever executable is put here, it abides by the conceptual contract of "(1) its executable output is a Python entry point (e.g. .py file), not an e.g. application, and (2) the binary is compatible with being treated as a dependency of other binaries and merged in as such".

I'll take a look at the doc here tonight and suggest changes, if any are needed.

That said, I still want get across an important point about Bazel design and how targets interact with one another. Sorry that I've pontificated a bunch below; I'm sincerely just interested in moving the Python rules forward and making sure they have a solid design base we can build upon.

-----

re: Allowing an executable for the coverage dependency:
First, to make sure we're on the same page, when I say "binary" I mean a top-level executable problem. I don't mean e.g. a shared library. I say this because something I've seen in OSS+Bazel+Python is, e.g. using `cc_binary` to produce a Python-loadable shared-library, and I've seen more C-centric people call them a binary.

Second, the way I see things, when we build e.g. a `py_binary`, things like the Python interpreter/runtime and coverage are inputs to that binary. This is because, essentially, that's how binary rules in Bazel are implemented: the binary rule gathers pieces from the toolchain, dependencies, etc and puts it all together into a final executable. But also because binaries, as the top-level entry point, should get the final say in how they work. (To be fair, I probably have some bias because, within Google, we use embedded Python, so Python is more a library used within a C program than separate runtime).

This is distinct from another way of thinking of things: that the `py_binary` and coverage are inputs to the Python interpreter (or that the py binary and interpreter are inputs to the coverage tool). This is more conceptually how non-Bazel Python works (you install a Python runtime, the add stuff into that runtime).

Insofar as Bazel is concerned, mixing together executables is problematic. In this case, we're mixing together a coverage executable and the py_binary executable. There are several ways this can interact badly, but it all comes down to each target specifying something that is mutually exclusive with something else. Examples include:
  * Specifying different malloc implementations (in Google, we allow `py_binary` to specify a malloc implementation)
  * Specifying or having different versions of the same the same library
  * Having dependencies that are mutually exclusive (e.g. I think boringssl/openssl are mutually exclusive; internally we have other such cases).
  * etc

A similar problem can happen at the Starlark-provider metadata level: an inner binary's settings (e.g. C++ info) can propagate up to an outer binary. This is one of the reasons `py_binary` _doesn't_ provide PyInfo. We saw this sort of issue internally very frequently when doing our Python 2 to 3 transition (a python 2 binary in the dependencies of a Python 3 binary, or vice versa, would fubar the build).

A couple other problems also occur when a binary is added to the dependencies (i.e. `deps` or equiv; not `data`) of another binary: build bloat and unnecessary overhead. This comes in a couple of different forms.

Using `py_binary` as an example again, recall that it (currently) has two outputs: the executable stub script and main .py file. We actually don't care about that executable stub script in this case, but it still has to be built and included into the runfiles (we're merging in its runfiles, as we must). If we built for e.g. windows, then we might get a `.exe` launcher that is entirely unused. In today's rules, this isn't too much overhead (the stub script is small), so nobody really notices. With something like embedded Python (or e.g. pyoxider etc), that overhead is pretty large.

This also segues into the next issue: the inner binary respects the flag settings intended for the outer binary. This is because it's being built in the same configuration (target, as it must). Going back to `py_binary`, if you specified, I think its the `--build_zip` flag (sp), then the inner binary would also build a (largely self-contained) zip file and use that as its output instead of e.g. the typical stub script. This isn't uncommon for binaries. While libraries can also have flags, it's not really an issue -- they still abide by the "I'm a thing to be included in another thing" contract.

I use `py_binary` as a convenient, if presently impossible, example. The same could be true of e.g. `sh_binary` etc.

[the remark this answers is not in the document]
adam-azarchs: > re: using a py_binary here: I don't think that's even possible -- a py_binary depends on the toolchain, so putting a py_binary as a dependency of the toolchain will result in circular dependencies.

Yes, you're absolutely right, and I'd thought of that about half an hour after making the comment.  But there are other similar rules which take in some script and output a version of that script with some kind of wrapper or whatever.  The point, here, is that one can imagine a number of different rules which output more than one file, and that's fine so long as the rule designates no more than one file as the entry point.

> I'm willing to accept an executable here because I can appreciate your situation, and as long as we make sure that (via docs, in this case), whatever executable is put here, it abides by the conceptual contract of "(1) its executable output is a Python entry point (e.g. .py file), not an e.g. application, and (2) the binary is compatible with being treated as a dependency of other binaries and merged in as such".
>
> I'll take a look at the doc here tonight and suggest changes, if any are needed. 

That's very reasonable.  I'll grant the docstring could use some expansion; I didn't expand on it much between my initial draft and now.

> That said, I still want get across an important point about Bazel design and how targets interact with one another. Sorry that I've pontificated a bunch below; I'm sincerely just interested in moving the Python rules forward and making sure they have a solid design base we can build upon.
> 
> re: Allowing an executable for the coverage dependency: First, to make sure we're on the same page, when I say "binary" I mean a top-level executable problem. I don't mean e.g. a shared library. I say this because something I've seen in OSS+Bazel+Python is, e.g. using `cc_binary` to produce a Python-loadable shared-library, and I've seen more C-centric people call them a binary.

Yes, one doesn't have to spend very much time writing (starlark) bazel rules to get that beaten into you :-).

> Second, the way I see things, when we build e.g. a `py_binary`, things like the Python interpreter/runtime and coverage are inputs to that binary. This is because, essentially, that's how binary rules in Bazel are implemented: the binary rule gathers pieces from the toolchain, dependencies, etc and puts it all together into a final executable. But also because binaries, as the top-level entry point, should get the final say in how they work. (To be fair, I probably have some bias because, within Google, we use embedded Python, so Python is more a library used within a C program than separate runtime).
> 
> This is distinct from another way of thinking of things: that the `py_binary` and coverage are inputs to the Python interpreter (or that the py binary and interpreter are inputs to the coverage tool). This is more conceptually how non-Bazel Python works (you install a Python runtime, the add stuff into that runtime).

In this case it's actually that the interpreter and the coverage tool are inputs to `py_binary`.  The `py_binary` stub (which could very reasonably not be written in python at all; making it a small C executable for example would solve more than [one problem](https://github.com/bazelbuild/bazel/issues/8446), and somewhat ironically, probably make things more portable) has the job of gluing those pieces together into some arraignment that produces the desired result: to run the given python script, with the given python interpreter, optionally using the given tool to collect coverage along the way.

> Insofar as Bazel is concerned, mixing together executables is problematic. In this case, we're mixing together a coverage executable and the py_binary executable. There are several ways this can interact badly, but it all comes down to each target specifying something that is mutually exclusive with something else.

Yes, and it's pretty clear in any case, from the implementation here, that the coverage tool does need to be something that  python can interpret (though as you mention, that could be an extension module or a `.pyc` bytecode file).  Even beyond that, it specifically expects the command line API to conform to that of [`coveragepy`](https://coverage.readthedocs.io/); it would be pretty far-fetched to call the implementation generic.  That said, I do believe that this is what ~99% of users outside of google would want anyway.  There aren't many organizations out there with enough SWE capacity to justify reinventing that particular wheel.

Perhaps using `FilesToRun` for the purpose of designating an entry point for a rule that outputs several files, when the target isn't _really_ a standalone binary, is a bit of an abuse of the mechanism, but if so I at least consider it to be a small one.  In our rule set at least, `bazel run .../coverage_tool`, _will_ work, though it'll run it through the system python interpreter, which is not ideal.

> * Specifying different malloc implementations (in Google, we allow `py_binary` to specify a malloc implementation)
> * Specifying or having different versions of the same the same library
> * Having dependencies that are mutually exclusive (e.g. I think boringssl/openssl are mutually exclusive; internally we have other such cases).
> * etc

Again, while it's true the API here is specified as if it were, we're not actually talking about arbitrary tools here.  Coveragepy _does_ include extension modules, however, which could bring up "interesting" issues with e.g. malloc implementations; I don't think there's really much of a way around that; if you're building `coveragepy` from source within bazel, which is probably a better idea that just using the precompiled wheel, then at least it's _likely_ that it'll be built with compatible options.

> A similar problem can happen at the Starlark-provider metadata level: an inner binary's settings (e.g. C++ info) can propagate up to an outer binary. This is one of the reasons `py_binary` _doesn't_ provide PyInfo. We saw this sort of issue internally very frequently when doing our Python 2 to 3 transition (a python 2 binary in the dependencies of a Python 3 binary, or vice versa, would fubar the build).

I think this issue is largely sidestepped here because we're not actually adding anything to the explicit dependencies of the `py_bianry` or `py_test` rule that's getting coverage enabled - it's only through the implicit toolchain dependency, and then only when coverage is enabled.  Meanwhile the toolchain provider metadata does _not_ propagate any metadata from the coverage tool, and as far as I know aspects don't follow implicit (toolchain) dependencies so there's not really any way for that to be visible to a starlark rule.

> A couple other problems also occur when a binary is added to the dependencies (i.e. `deps` or equiv; not `data`) of another binary: build bloat and unnecessary overhead. This comes in a couple of different forms.
> 
> Using `py_binary` as an example again, recall that it (currently) has two outputs: the executable stub script and main .py file. We actually don't care about that executable stub script in this case, but it still has to be built and included into the runfiles (we're merging in its runfiles, as we must). If we built for e.g. windows, then we might get a `.exe` launcher that is entirely unused. In today's rules, this isn't too much overhead (the stub script is small), so nobody really notices. With something like embedded Python (or e.g. pyoxider etc), that overhead is pretty large.

I can't really imagine many situations where someone would want to make bundled distributions of a tool (e.g. pyoxidizer) with coverage enabled.  Maybe for python code that needed to run on some specific hardware?  But maybe that's just my imagination being too limited.  At any rate if you did that you'd be having a bad time because that `.exe` entry point is what the stub would be feeding to the python interpreter, which probably wouldn't go very well.

> This also segues into the next issue: the inner binary respects the flag settings intended for the outer binary. This is because it's being built in the same configuration (target, as it must). Going back to `py_binary`, if you specified, I think its the `--build_zip` flag (sp), then the inner binary would also build a (largely self-contained) zip file and use that as its output instead of e.g. the typical stub script. This isn't uncommon for binaries. While libraries can also have flags, it's not really an issue -- they still abide by the "I'm a thing to be included in another thing" contract.
> 
> I use `py_binary` as a convenient, if presently impossible, example. The same could be true of e.g. `sh_binary` etc.

Thank you, these are fascinating insights which help to crystalize some things about binary targets in bazel which were, admittedly, fuzzy in my understanding.  That said, I don't really think they apply here, for reasons already discussed above.

rickeylev (approved):
Mostly just doc fixes. Otherwise LGTM

src/main/java/com/google/devtools/build/lib/rules/python/PyRuntime.java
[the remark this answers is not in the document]
adam-azarchs: Our team actually hasn't had any trouble integrating extension modules into our bazel build, though we do need to depend explicitly on our python build target.  It would be pretty cool if python toolchains provided `CcInfo` for those purposes, to make it a bit more portable.  (don't want to create noise there by linking it but issue 24665 in grpc is somewhat related to this).

Yes, you are correct that `pytest` has its own coverage API (or maybe plugin? Not quite sure), however that doesn't support reporting in `lcov` format, which makes it less useful in a bazel context.  Agree it would be neat to add some kind of provider to make coverage more configurable, but looking at the implementation in `python_stub_template.txt` there's a lot in there that would be difficult to make configurable.

That functionality could be added at some point in the future without breaking backwards compatibility, with `PyRuntimeRule` detecting if such an provider was present and responding accordingly, but otherwise falling back on the behavior implemented here.

adam-azarchs:
Thank you for the thorough and informative review!

rickeylev:
@c-mita Can you review/approve/merge, please?

adam-azarchs:
Test failure looks to be caused by efa196c7857945f4774a8930e6fd63c9bad698e6, unrelated to my changes.  Once master is passing again I'll merge that in.

adam-azarchs:
Friendly ping!  I would love for our internal CI to be able to start using this feature in official weekly builds some time in the not so distant future, rather than resorting to building off of my dev branch.

rickeylev:
@c-mita @comius signal boost

c-mita:
I'm not familiar with the Python rules or how coverage works for them, but with that caveat, LGTM. ;)

c-mita (approved):

adam-azarchs:
Oops, accidentally pushed a button on the github UI that made it merge from master.  Not that that's a bad thing exactly, but I'm going to assume that that clobbers the approvals.

Related note, I'm not sure how stuff gets from having the [awaiting-PR-merge](https://github.com/bazelbuild/bazel/labels/awaiting-PR-merge) label to actually being merged?

rickeylev:
It says "changed approved" still, so I think the approval is still good. Thanks for bumping the thread -- if something lingers for more than a day or so please tag me.

@c-mita Can you please merge? I don't have permission to do so.

rickeylev:
Oh, wait, I think I see now. ["After your review is complete, a Bazel maintainer applies your patch to Google's internal version control system."](https://bazel.build/contribute/patch-acceptance)

I'll see if I can figure out how to do that internally.

Wyverald:
> Related note, I'm not sure how stuff gets from having the [awaiting-PR-merge](https://github.com/bazelbuild/bazel/labels/awaiting-PR-merge) label to actually being merged?

A member of the Bazel support team will import it into Google. This label is constantly monitored by them. @sgowroji just in case this fell off your radar, could you import this please? Thanks :)

PR #15594, opened 2022-05-30
Implement coverage reports for Starlark

fmeum:
If the new flag `--starlark_coverage_report` is set to a non-empty value, the Starlark coverage recorder will be enabled for all `.bzl` files under the output base executed during the Blaze command. At the end of the command, the collected coverage data will be emitted to the file path given as the flag argument in LCOV format, containing function, line, and branch coverage information.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
@@ -294,6 +294,15 @@ public class CommonCommandOptions extends OptionsBase {
   public String starlarkCpuProfile;
 
   @Option(
+      name = "starlark_coverage_report",
+      defaultValue = "",
+      documentationCategory = OptionDocumentationCategory.LOGGING,
+      effectTags = {OptionEffectTag.BAZEL_MONITORING},
+      help = "Writes into the specified file an LCOV coverage report for all Starlark files "
+          + "executed for the requested Bazel command.")
+  public String starlarkCoverageReport;
+
+  @Option(
       name = "record_full_profiler_data",
       defaultValue = "false",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
--- a/src/main/java/net/starlark/java/eval/Eval.java
+++ b/src/main/java/net/starlark/java/eval/Eval.java
@@ -29,6 +29,7 @@ import net.starlark.java.syntax.BinaryOperatorExpression;
 import net.starlark.java.syntax.CallExpression;
 import net.starlark.java.syntax.Comprehension;
 import net.starlark.java.syntax.ConditionalExpression;
+import net.starlark.java.syntax.CoverageRecorder;
 import net.starlark.java.syntax.DefStatement;
 import net.starlark.java.syntax.DictExpression;
 import net.starlark.java.syntax.DotExpression;
@@ -45,6 +46,7 @@ import net.starlark.java.syntax.LambdaExpression;
 import net.starlark.java.syntax.ListExpression;
 import net.starlark.java.syntax.LoadStatement;
 import net.starlark.java.syntax.Location;
+import net.starlark.java.syntax.Parameter;
 import net.starlark.java.syntax.Resolver;
 import net.starlark.java.syntax.ReturnStatement;
 import net.starlark.java.syntax.SliceExpression;
@@ -131,6 +133,7 @@ final class Eval {
             continue;
           case BREAK:
             // Finish loop, execute next statement after loop.
+            CoverageRecorder.getInstance().recordVirtualJump(node);
             return TokenKind.PASS;
           case RETURN:
             // Finish loop, return from function.
@@ -145,6 +148,7 @@ final class Eval {
     } finally {
       EvalUtils.removeIterator(seq);
     }
+    CoverageRecorder.getInstance().recordVirtualJump(node);
     return TokenKind.PASS;
   }
 
@@ -159,7 +163,9 @@ final class Eval {
     int nparams =
         rfn.getParameters().size() - (rfn.hasKwargs() ? 1 : 0) - (rfn.hasVarargs() ? 1 : 0);
     for (int i = 0; i < nparams; i++) {
-      Expression expr = rfn.getParameters().get(i).getDefaultValue();
+      Parameter parameter = rfn.getParameters().get(i);
+      CoverageRecorder.getInstance().recordCoverage(parameter.getIdentifier());
+      Expression expr = parameter.getDefaultValue();
       if (expr == null && defaults == null) {
         continue; // skip prefix of required parameters
       }
@@ -168,6 +174,10 @@ final class Eval {
       }
       defaults[i - (nparams - defaults.length)] =
           expr == null ? StarlarkFunction.MANDATORY : eval(fr, expr);
+    }
+    // Visit kwargs and varargs for coverage.
+    for (int i = nparams; i < rfn.getParameters().size(); i++) {
+      CoverageRecorder.getInstance().recordCoverage(rfn.getParameters().get(i).getIdentifier());
     }
     if (defaults == null) {
       defaults = EMPTY;
@@ -206,6 +216,7 @@ final class Eval {
     } else if (node.getElseBlock() != null) {
       return execStatements(fr, node.getElseBlock(), /*indented=*/ true);
     }
+    CoverageRecorder.getInstance().recordVirtualJump(node);
     return TokenKind.PASS;
   }
 
@@ -262,6 +273,7 @@ final class Eval {
     if (++fr.thread.steps >= fr.thread.stepLimit) {
       throw new EvalException("Starlark computation cancelled: too many steps");
     }
+    CoverageRecorder.getInstance().recordCoverage(st);
 
     switch (st.kind()) {
       case ASSIGNMENT:
@@ -330,6 +342,7 @@ final class Eval {
 
   private static void assignIdentifier(StarlarkThread.Frame fr, Identifier id, Object value)
       throws EvalException {
+    CoverageRecorder.getInstance().recordCoverage(id);
     Resolver.Binding bind = id.getBinding();
     switch (bind.getScope()) {
       case LOCAL:
@@ -477,6 +490,7 @@ final class Eval {
     if (++fr.thread.steps >= fr.thread.stepLimit) {
       throw new EvalException("Starlark computation cancelled: too many steps");
     }
+    CoverageRecorder.getInstance().recordCoverage(expr);
 
     // The switch cases have been split into separate functions
     // to reduce the stack usage during recursion, which is
@@ -533,9 +547,17 @@ final class Eval {
     // AND and OR require short-circuit evaluation.
     switch (binop.getOperator()) {
       case AND:
-        return Starlark.truth(x) ? eval(fr, binop.getY()) : x;
+        if (Starlark.truth(x)) {
+          return eval(fr, binop.getY());
+        }
+        CoverageRecorder.getInstance().recordVirtualJump(binop);
+        return x;
       case OR:
-        return Starlark.truth(x) ? x : eval(fr, binop.getY());
+        if (Starlark.truth(x)) {
+          CoverageRecorder.getInstance().recordVirtualJump(binop);
+          return x;
+        }
+        return eval(fr, binop.getY());
       default:
         Object y = eval(fr, binop.getY());
         try {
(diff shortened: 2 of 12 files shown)


fmeum:
CC @Wyverald 

I generated reports for some rulesets and found the results to be relatively accurate, but this would definitely benefit from more testing (it doesn't come with test cases yet). I wanted to open a PR to get early feedback and will happily fix bugs and supply tests if this is generally viewed favorably.

src/main/java/net/starlark/java/eval/Eval.java
```diff
@@ -131,6 +133,7 @@ private static TokenKind execFor(StarlarkThread.Frame fr, ForStatement node)
             continue;
           case BREAK:
             // Finish loop, execute next statement after loop.
+            CoverageRecorder.getInstance().recordVirtualJump(node);
```
fmeum, line 136: I don't have any data to assess whether this causes a performance degradation even if coverage is not being recorded (which means that the instance returned will have a no-op `recordVirtualJump`). If this should turn out to be prohibitive, I could make the instance `static final` and populate it e.g. via an environment variable rather than a Bazel flag.

Wyverald:
Thanks for the PR, Fabian! I'll leave it to Jon & Sasha to review this. (NB: Jon is OOO until 9 June.)

Also cc @adonovan

fmeum:
I fixed some bugs and refactored the branch, but will not make further changes.

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
```diff
@@ -293,6 +293,14 @@ public String getTypeDescription() {
       help = "Writes into the specified file a pprof profile of CPU usage by all Starlark threads.")
   public String starlarkCpuProfile;
 
+  @Option(
+      name = "starlark_coverage_report",
+      defaultValue = "",
+      documentationCategory = OptionDocumentationCategory.LOGGING,
+      effectTags = {OptionEffectTag.BAZEL_MONITORING},
+      help = "Writes into the specified file an lcov coverage report for all Starlark files.")
```
sluongng, line 301: From a discussion on Slack: It's not immediately clear to end-users that we are using this flag to report coverage of starlark during execution of a bazel command, similar to `--starlark_cpu_profile`.

My initial impression was that this, similar to other coverage flags, are used with `bazel coverage` when testing starlark rules with Skylib's `unittest.bzl`. This is NOT the case.

Might worth being a bit more specific in this description so that Bazel's end users are clear on intended usages.

sluongng (commented):
I definitely am not familiar with Google Java Style guide or the readability process, but here are some nits I have after reading through the PR. Please take them with a grain of salt 🙇‍♂️

Thanks for working on this 🙏 I can see how this would be very useful for my debugging process.

src/main/java/net/starlark/java/syntax/CoverageRecorder.java
```diff
@@ -0,0 +1,648 @@
...
+  @Override
+  public void dump(PrintWriter out) {
+    ImmutableList<Program> programs = registeredPrograms.stream()
+        .sorted(Comparator.comparing(Program::getFilename))
+        .collect(ImmutableList.toImmutableList());
+    for (Program program : programs) {
+      CoverageNodeVisitor visitor = new CoverageNodeVisitor(program);
+      visitor.visitAll(program.getResolvedFunction().getBody());
+      visitor.dump(out);
+    }
```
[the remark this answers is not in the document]
fmeum: I replaced it with a stream-only implementation. I ended up not using `map` since the `visitAll` call requires `program`.

src/main/java/net/starlark/java/syntax/NodePrinter.java
```diff
@@ -482,4 +476,18 @@ private void printExpr(Expression expr) {
...
+      buf.append("for ");
+      printExpr(forClause.getVars());
+      buf.append(" in ");
+      printExpr(forClause.getIterable());
+    } else {
+      Comprehension.If ifClause = (Comprehension.If) clause;
+      buf.append("if ");
+      printExpr(ifClause.getCondition());
+    }
+  }
```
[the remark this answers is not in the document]
fmeum: In this particular case I personally find the code more readable without the early return, which destroys the symmetry between the branches. Ideally, this would be a switch over a sealed class, but we don't have those.

src/main/java/com/google/devtools/build/lib/runtime/CommonCommandOptions.java
fmeum: I reworded this a little to hopefully clarify the scope. I would be interested in integrating a less low-level version of this with `bazel coverage`, but don't really know how to accomplish this yet.

src/main/java/net/starlark/java/syntax/NodePrinter.java
```diff
@@ -14,6 +14,7 @@
 package net.starlark.java.syntax;
 
 import java.util.List;
+import net.starlark.java.syntax.Comprehension.Clause;
```
adonovan, line 17: Please don't import nested classes; always refer to Comprehension.Clause outside that class.

src/main/java/net/starlark/java/syntax/NodePrinter.java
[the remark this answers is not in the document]
adonovan: I agree, the branches have equal weight and so 'else' is appropriate.

src/main/java/net/starlark/java/syntax/NodeVisitor.java
```diff
@@ -174,9 +174,7 @@ public void visit(@SuppressWarnings("unused") Comment node) {}
   public void visit(ConditionalExpression node) {
     visit(node.getCondition());
     visit(node.getThenCase());
-    if (node.getElseCase() != null) {
```
adonovan, line 177: Why is this change safe?

src/main/java/net/starlark/java/syntax/CoverageVisitor.java
```diff
@@ -0,0 +1,331 @@
...
+// limitations under the License.
+
+package net.starlark.java.syntax;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+abstract class CoverageVisitor extends NodeVisitor {
```
adonovan, line 22: This class doesn't belong in the syntax package: it's a client of that package, and it seems like part of the interpreter (eval).  It appears there is only a single actual class that implements this abstraction, so I suggest you merge them together.

adonovan (commented):
Thanks for sending this, but a change this big needs a design discussion before code. What problem are we trying to solve? What design do you propose? How can we make it simpler?

My initial questions are:
- Does this make the interpreter slower? A lot of work has gone into making it as fast as it is (although there is still room for improvement). Benchmarking is time consuming because the JVM is noisy---execution time may vary by 5% or more---making it hard to find regressions down at the 1-2% level.
- Do we need all this new machinery? In principle, a coverage tool could be implemented as a debugger that simply steps through the whole program one instruction at a time recording the current pc. We already have a primitive debugger. Can its hooks be used to record coverage, with minimal changes?
- Do we need branch coverage? I notice the tool records control jumps such as break statements and the short-circuit in `if true || true {`, and information about individual parameters. Is this necessary? Statement coverage is good enough for most purposes and is less invasive.

src/main/java/net/starlark/java/syntax/NodeVisitor.java
fmeum: `ConditionalExpression#getElseCase()` is not annotated as `@Nullable`, which is correct since the lexer [only accepts a `ConditionalStatement` if it has an `else`](https://cs.opensource.google/bazel/bazel/+/6201dd161aa2878efc68cfcd6a2acf9193f548e8:src/main/java/net/starlark/java/syntax/Parser.java;l=944;drc=7e3755d3bf1443a0f2a780bc8efff658c464c3a5).

fmeum:
> Thanks for sending this, but a change this big needs a design discussion before code. What problem are we trying to solve? What design do you propose? How can we make it simpler?

Sorry, I certainly didn't intend to skip proper process here. I wrote this code out of a concrete need to better understand the edge cases of rulesets I maintain and to improve my understanding of the Java Starlark implementation. I just thought about sharing it hoping that it could be useful for others. I am completely open to rewriting this from scratch or scrapping it entirely, depending on what the outcome of the design discussion is.

>
> My initial questions are:
>
> * Does this make the interpreter slower? A lot of work has gone into making it as fast as it is (although there is still room for improvement). Benchmarking is time consuming because the JVM is noisy---execution time may vary by 5% or more---making it hard to find regressions down at the 1-2% level.

The way the hooks are currently implemented, I won't be able to dismiss this concern with confidence. As you said, benchmarking doesn't necessarily catch all regressions on such a low level.

I do think that there is a way to implement the coverage hooks without any overhead in the case where coverage isn't being collected:

1. Make the static `CoverageRecorder` instance final and populate it based on e.g. a system property `starlark.lcov_report_file` being set to a non-empty value.
2. Make `--starlark_coverage_profile` a startup option that sets this system property for the Bazel server JVM.
3. Add logic that dumps the report on server shutdown. 

The `static final` field should be inlined by the JIT - if we want to be certain, we could also make it a simple boolean that triggers a coverage slow path if true. I could then double-check that the code emitted by the JIT doesn't change with the PR merged. 

Do you like this approach better?
 
> * Do we need all this new machinery? In principle, a coverage tool could be implemented as a debugger that simply steps through the whole program one instruction at a time recording the current pc. We already have a primitive debugger. Can its hooks be used to record coverage, with minimal changes?

After a first read of the code, I'm inclined to say this would work for statement coverage, but not for more fine-grained notions of coverage.

> * Do we need branch coverage? I notice the tool records control jumps such as break statements and the short-circuit in `if true || true {`, and information about individual parameters. Is this necessary? Statement coverage is good enough for most purposes and is less invasive.

I agree that statement coverage would already be quite useful. However, given the general conciseness of Starlark (especially with list/dict comprehensions and conditional expressions), a single statement can encode rather complex logic. With the rulesets I have tried this on, statement coverage has almost always been nearing 100%, but branch coverage revealed some interesting uncovered cases (e.g., handling of workspace vs. external repository, checks for missing/superfluous files). While branch coverage information for `break`s may not be as useful, I found it better to cover all branches rather than just some of them.

Assuming we find a way to ensure that coverage support introduces no overhead at all (as possibly described above), would you still prefer to remove some or all of the branch coverage hooks?.

fmeum:
@adonovan Would you like me to prototype the `static final` approach and take a look at the JIT output? I think that this has a fair chance of alleviating the performance worries entirely.

brandjon:
Finally taking a look at this, as I promised Fabian back at BazelCon.

It looks like there's quite a bit more code to this PR than I originally thought. I'm interested in supporting more tooling for Starlark in the new year, coverage included, but in light of

* the added code complexity (who will maintain all this extra logic?),
* concerns of performance regression, and
* uncertainty about the extent of the benefit of this feature to users, and the relative benefits of expression granularity vs statement granularity,

it's clear that a quick acceptance of this PR is off the table. I'd love to have a more involved discussion centered around the benefit vs maintainability tradeoff in Q1. I'll keep the PR open for now.

fmeum:
@brandjon @tetromino Friendly ping on this. 

I could trim this down to just line coverage and ensure that there is no runtime overhead if coverage is disabled.

jjmaestro:
Just my +1 to this feature, I began writing some modules with Starlark, mostly library / functionality I will use in other projects, and I would love to see how well I've managed to test these modules. I'd love to have Starlark coverage! Also, +1 to branch coverage, I think it adds quite a lot of value to coverage tests so I would like if that functionality remains in the PR.

Thanks!!

albertocavalcante:
Thanks for the work you've done here, @fmeum! 

If it's ok to ask: Is it linked to an issue somewhere? I was about to raise an issue for this then tried to search on https://github.com/bazelbuild/proposals as well but didn't find. Slack brought me here.

Then I would like to confirm, this would apply only for starlark-java right? There would still a need to implement in `starlark-go` and `starlark-rust`.
If so, I was wondering if an issue should be raised in the starlark spec repo and then in each implementation as well. I'm not sure on the best practice or convention/etiquette here.

https://github.com/bazelbuild/starlark/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aopen%20coverage

albertocavalcante:
@brandjon @tetromino as Fabian suggested, is there anything that could be done or changed to get this PR landed? 
Also, @fmeum if I can help (or try to) in any capacity, please let me know.

Thanks!

Issue #12902, opened 2021-01-26
Global shutdown / clean commands


guibou:
### Description of the problem / feature request:

I'd like to be able to `bazel shutdown` or `bazel clean` including all my repos, and possible disk cache.

### Feature requests: what underlying problem are you trying to solve with this feature?

I'm working on multiples bazel workspace at the same time for different repo, or even for the same repo using `git workspace`. The result is that I do have many bazel server / bazel base directory used at the same time.

I'd like to kill a few of the bazel server in order to reclaim. My current workflow is as following:

- using `ps` (on linux) or any other process managemen ttool, I'm locating the different bazel server instances. They are named `bazel(xxx)` where `xxx` is linked to the name of the root directory of the current workspace.
- `cd` in the associated directory. This task is difficult to automate.
- `bazel shutdown`

And I need to repeat this process for all the different bazel server there is.

Same when I want to reclaim disk space, I need to go to all my bazel workspace and type `bazel clean [--expunge]`.

I'd like global commands, such as:

- `bazel shutdown_all`
- `bazel clean_all [--expunge]`
- ...

### What operating system are you running Bazel on?

Linux

### What's the output of `bazel info release`?

```
INFO: Invocation ID: dc694622-27a5-4f89-848e-c832f77fcd37
release 3.3.1- (@non-git)
```

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

nixpkgs.

PR #15608, opened 2022-06-02
Add support for cleaning all outputBase under output root.

KaneBetter:
Support [Issue#12902](https://github.com/bazelbuild/bazel/issues/12902)

With this feature, `bazel clean --all` can remove the execroot tree created by all Bazel instances in your local machine.
Also `all` is compatible with `async` and `expunge`.


meteorcloudy:
Sorry, we cannot accept this change, letting one Bazel instance to delete output directories for another sounds a bit risky. And I don't think this would work if other Bazel instance specific different output_user_root.

Issue #13771, opened 2021-07-29
Documentation issue: Concepts and terminology


devjgm:
Documentation URL: https://docs.bazel.build/versions/main/build-ref.html

This documentation says:
> Directories containing a file called WORKSPACE are considered the root of a workspace. Therefore, Bazel ignores any directory trees in a workspace rooted at a subdirectory containing a WORKSPACE file (as they form another workspace).

Which really sounds like subdirectories containing a WORKSPACE file will not be descended when building with `bazel build ...` from within the parent workspace. However, this appears to be untrue.  Given

```
repo/
├── BUILD
├── foo
│   ├── BUILD
│   ├── foo.cc
│   └── WORKSPACE
└── WORKSPACE
```
```console
$ bazelisk query ...
//foo:test2
//:test
Loading: 0 packages loaded

$ bazelisk build ...
INFO: Analyzed 2 targets (14 packages loaded, 57 targets configured).
INFO: Found 2 targets...
INFO: Elapsed time: 1.287s, Critical Path: 0.07s
INFO: 6 processes: 3 internal, 3 linux-sandbox.
INFO: Build completed successfully, 6 total actions
```

Which *appears* to contradict the documentation.

PR #15615, opened 2022-06-03
Fix grammar mistake in build-ref

gboer:


gregestren (commented):
Interesting change. I think I can see the point being made with both wordings? Do you see just ambiguous wording or a clear doc error being fixed?

gboer:
Well, at least 'form' should be 'from' right? And I'm no native English speaker, but for me, the 'are' should be there as well.

gregestren:
I think the original wording is saying that a subdirectory tree with its own `WORKSPACE` file actually *is* another workspace, which means that subtree forms its own workspace.

The new wording is subtly different in that it implies they're part of, but not necessarily the entire, workspace.

But I don't know the exact history. I'll link some folks I think can offer more authoritative input

@Wyverald 

Also see https://github.com/bazelbuild/bazel/issues/13771.

Wyverald:
The wording here is indeed a bit unclear. The terminology we're going for right now is that the "workspace" is the whole thing you're running Bazel in (i.e. the "root" workspace), whereas the subdirectories with WORKSPACE files are just external repos that happen to reside in the same directory structure. I should fix the glossary page some time.

In any case, I don't believe the specific change in this PR clarifies anything. "A forms B" is a perfectly clear English sentence.

PR #15618, opened 2022-06-03
Fix and test for null value bug when using --repo_env

SalmaSamy:
Fix and test for null value bug when using --repo_env
[Issue Link](https://github.com/bazelbuild/bazel/issues/15430)

### DIFF
--- a/src/test/py/bazel/bazel_external_repository_test.py
+++ b/src/test/py/bazel/bazel_external_repository_test.py
@@ -427,5 +427,34 @@ class BazelExternalRepositoryTest(test_base.TestBase):
         args=['build', '@other_repo//:file'], cwd=work_dir)
     self.AssertExitCode(exit_code, 0, stderr)
 
+  def testRepoEnv(self):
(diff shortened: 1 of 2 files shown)


src/test/py/bazel/bazel_external_repository_test.py
```diff
@@ -427,5 +427,34 @@ def testExternalBazelignoreContainingRepoName(self):
...
+      '    print("FOO", val)',
+      '    ctx.file("BUILD")',
+      'dump_env = repository_rule(',
+      '    implementation = _dump_env,',
+      '    local = True,',
+      ')'
+    ])
+
+    exit_code, _, stderr = self.RunBazel(
+      args=['build', '@debug//:all'],
```
meteorcloudy, line 452: Please addd `-repo_env=FOO` here

SalmaSamy: Done

meteorcloudy (approved):
Thanks for the fix!

meteorcloudy (approved):

meteorcloudy:
@bazel-io fork 5.3.0

Issue #10910, opened 2020-03-05
Add -no_deduplicate to crosstool for ld64 flags for dbg/fastbuild builds


kastiglione:
### Description of the problem / feature request:

For debug builds, Xcode uses ld64's `-no_deduplicate` flag. This has a decent speed up for link times. Bazel does not use this flag in any compilation mode.

### Feature requests: what underlying problem are you trying to solve with this feature?

Using `-no_deduplicate` creates more parity with Xcode builds and results in faster linking times. In our project, `-no_deduplicate` reduced link time by about 8%.

### What operating system are you running Bazel on?

macOS

### What's the output of `bazel info release`?

release 2.1.0

PR #15635, opened 2022-06-07
Add -no_deduplicate for darwin dbg/fastbuild

keith:
The -no_deduplicate linker flag saves some time for debug builds at the
cost of binary size. In testing envoy the difference was 50s -> 9s to
link a debug build with this flag enabled. This change enables it as a
separate feature so it can be opted out of by users if needed, and
leaves the default for opt to get this optimization for release builds.

Fixes https://github.com/bazelbuild/bazel/issues/10910

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -1161,6 +1161,22 @@ def _impl(ctx):
             ],
         )
 
+    no_deduplicate_feature = feature(
+        name = "no_deduplicate",
+        enabled = True,
+        flag_sets = [
+            flag_set(
+                actions = all_link_actions +
+                          ["objc-executable", "objc++-executable"],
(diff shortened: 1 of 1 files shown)


keith:
@sgowroji bump

googlewalt (approved):

brentleyjones (approved):

googlewalt:
FYI I made a couple of NFCs to the PR, to be more consistent with our internal toolchain.

jwnimmer-tri:
If I'm reading correctly, this change was first released in bazel 6.0.0?  I don't see it mentioned in the [release notes](https://github.com/bazelbuild/bazel/releases/tag/6.0.0), though.

This caused a 9% increase in our build farm disk use, which was severe enough that we needed to opt-out of this feature.  For that reason, I think it might be worth adding this to the release notes retroactively, so that it's easier for others to discover what might have changed to cause the bloat they're seeing when they upgrade.

PR #15637, opened 2022-06-08
Fix param_file_arg error message

fmeum:
The error message incorrectly escaped the missing format specifier. As
a result, the 'Expected string with a single "%s"' read 'Expected string
with a single "<value of the arg>"' instead.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/Args.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/Args.java
@@ -526,7 +526,7 @@ public abstract class Args implements CommandLineArgsApi {
       Starlark.checkMutable(this);
       if (!SingleStringArgFormatter.isValid(paramFileArg)) {
         throw Starlark.errorf(
-            "Invalid value for parameter \"param_file_arg\": Expected string with a single \"%s\"",
+            "Invalid value for parameter \"param_file_arg\": Expected string with a single \"%%s\", got \"%s\"",
             paramFileArg);
       }
       this.flagFormatString = paramFileArg;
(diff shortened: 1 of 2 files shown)


fmeum:
@meteorcloudy Could you review this? It's a trivial fix.

meteorcloudy (approved):
Thanks!

meteorcloudy:
@sgowroji I'm importing this one, so you don't have to

src/main/java/com/google/devtools/build/lib/analysis/starlark/Args.java
```diff
@@ -526,8 +526,7 @@ public CommandLineArgsApi useParamsFile(String paramFileArg, Boolean useAlways)
       Starlark.checkMutable(this);
       if (!SingleStringArgFormatter.isValid(paramFileArg)) {
         throw Starlark.errorf(
-            "Invalid value for parameter \"param_file_arg\": Expected string with a single \"%s\"",
-            paramFileArg);
+            "Invalid value for parameter \"param_file_arg\": Expected string with a single \"%%s\"");
```
Wyverald, line 529: should we still print out `paramFileArg` to ease debugging? after the `a single \"%%s\"` part

fmeum: Good point, changed.

fmeum: @meteorcloudy In case you haven't hit merge yet, could you import the updated version?

meteorcloudy: Sure!

meteorcloudy: Hmm, looks like it somehow broke the bazel_module_test

meteorcloudy: Oh, no, it's my change to the BCR repo, I deleted the `testing` branch, but apparently it's used by the test.

meteorcloudy: I see, it's your base commit is too old, which still uses `https://registry.bazel.build` in the bazel_module_test
https://github.com/bazelbuild/bazel/blob/0529fdb18f42c1381cdefe7534e5b5d86ac1e673/src/test/py/bazel/bzlmod/bazel_module_test.py#L45

But it doesn't matter, the internal presubmit check should pass, since it's always rebased.

PR #15638, opened 2022-06-08
Add a flag to force Bazel to download certain artifacts when using --remote_download_minimal

lfpino:
When using --remote_download_minimal Bazel only downloads to the client the minimum number of artifacts necessary for the build to succeed. This can be sometimes problematic when local development (i.e. IDE) requires some artifacts to do local work (e.g. syntax completion). We add a flag --experimental_remote_download_regex that allow specifying a regular expression that will force certain artifacts to be forced to downloaded.

Tested locally that syntax errors disappear in IntelliJ when compiling Bazel itself with --remote_download_minimal and --experimental_remote_download_regex=".*jar$", also included an integration test with several cases.

Signed-off-by: Luis Fernando Pino Duque <luis@engflow.com>


src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -215,6 +220,16 @@ public RemoteExecutionService(
     this.captureCorruptedOutputsDir = captureCorruptedOutputsDir;
 
     this.scheduler = Schedulers.from(executor, /*interruptibleWorker=*/ true);
+
+    String regex = remoteOptions.experimentalForceDownloadsRegex;
+    // TODO(bazel-team): Consider adding a warning or more validation if the experimentalForceDownloadsRegex is used
+    // without RemoteOutputsMode.MINIMAL
+    this.shouldForceDownloads = !regex.isEmpty() && remoteOptions.remoteOutputsMode == RemoteOutputsMode.MINIMAL;
```
coeuvre, line 227: I think we can also use this flag for `--remote_download_toplevel`.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -1021,24 +1036,9 @@ public InMemoryOutput downloadOutputs(RemoteAction action, RemoteActionResult re
             /* exitCode = */ result.getExitCode(),
             hasFilesToDownload(action.getSpawn().getOutputFiles(), filesToDownload));
 
+    ImmutableList.Builder<ListenableFuture<FileMetadata>> forcedDownloadsBuilder = ImmutableList.builder();
```
coeuvre, line 1039: why not reuse `downloadsBuilder`?

coeuvre (changes requested):
Thanks for your PR!

I think this is working towards the point 3 I mentioned in https://github.com/bazelbuild/bazel/issues/12665.

The new flag works on output-wise, i.e. filter each output of a spawn. I didn't look into IDE usages yet, but since we are on the topic, I want to learn more about the requirements. Have you consider other options, like, filter at spawn level (that's how `--remote_download_toplevel` works) or use an aspect with output group?

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -574,6 +574,16 @@ public RemoteOutputsStrategyConverter() {
       help = "Maximum number of open files allowed during BEP artifact upload.")
   public int maximumOpenFiles;
 
+  @Option(
+          name = "experimental_force_downloads_regex",
```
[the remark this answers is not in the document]
lfpino: Done

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: I was trying to steer away from the current downloads codepath since I'm not an expert in this part of the code. Wanted to make sure this change is isolated to avoid any potential problems but if you want I can try to unify them using downloadsBuilder.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: Done

lfpino (commented):
> Thanks for your PR!
> 
> I think this is working towards the point 3 I mentioned in #12665.
> 
> The new flag works on output-wise, i.e. filter each output of a spawn. I didn't look into IDE usages yet, but since we are on the topic, I want to learn more about the requirements. Have you consider other options, like, filter at spawn level (that's how `--remote_download_toplevel` works) or use an aspect with output group?

You're welcome! :-) and thanks for the prompt reply!

Yeah, I hadn't seen #12665 but this does contribute towards that goal.

The requirement is quite simple: Being able to use build without the bytes without breaking IDE integration. By using this flag, one can use IntelliJ with --remote_download_minimal and ask Bazel to also download the jar files.

I haven't explored any other options since, as usual, there's too many ways of doing this and I tried something that made sense when I looked at the code. Of course if you think that this is not ok architecturally I'm happy to look at other options but I'll probably need a couple of code pointers. If you think this approach is ok, I think we should merge and improve this with some follow-up PRs. WDYT?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
coeuvre: Yes, please unify them.

coeuvre (commented):
It seems that the new changes haven't been pushed.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: I tried to unify them in a few ways but I always ended up finding issues and making the code more unreadable. Do you have any suggestions? Since this is experimental I'm inclined to leave it as it is right now and if we can confirm this is a good direction then we can refactor the code to make it more readable (I still find it very confusing the code path that depends on the downloadOutputs boolean)

lfpino (commented):
Addressed comments PTAL

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
coeuvre: `downloadOutputs` basically means whether we are using BwoB. The two mods were merged into one function because eventually we want to only have BwoB (but the behaviour is still the same by default that Bazel download ALL outputs but in the background, so that the boolean is removed and only one code path left).

As you said this is an experimental feature, I am fine to move forward without unifying them at this PR if that introduce other issues.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -1068,6 +1075,15 @@ public InMemoryOutput downloadOutputs(RemoteAction action, RemoteActionResult re
       throw e;
     }
 
+    ImmutableList<ListenableFuture<FileMetadata>> forcedDownloads = forcedDownloadsBuilder.build();
+    try (SilentCloseable c = Profiler.instance().profile("Remote.forcedDownload")) {
```
coeuvre, line 1079: Consider extracting this into a function to share the code above.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -1079,6 +1095,10 @@ public InMemoryOutput downloadOutputs(RemoteAction action, RemoteActionResult re
     tmpOutErr.clearOut();
     tmpOutErr.clearErr();
 
+    if (!forcedDownloads.isEmpty()) {
```
coeuvre, line 1098: Consider moving this into `else` block below to make it clear that this is only available in BwoB.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -215,6 +220,19 @@ public RemoteExecutionService(
     this.captureCorruptedOutputsDir = captureCorruptedOutputsDir;
 
     this.scheduler = Schedulers.from(executor, /*interruptibleWorker=*/ true);
+
+    String regex = remoteOptions.remoteDownloadRegex;
+    // TODO(bazel-team): Consider adding a warning or more validation if the experimentalForceDownloadsRegex is used
```
coeuvre, line 225: Comment needs update.

coeuvre (commented):
Sorry for the delay. I think this PR is good now. Just address remaining comments and we are ready to merge.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: Done.

lfpino:
@coeuvre this PR should be ready now, PTAL!

lfpino:
There are some failures, let me check that first.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
lfpino: @coeuvre unfortunately the generic Exception catch here is making this code very brittle. Exception handling needs to be improved before we can move this to a generic function, I've added a TODO to avoid mixing things up. That was causing the test failures from my previous attempts.

lfpino:
@coeuvre you should be able to take a look now

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -1133,6 +1159,47 @@ public InMemoryOutput downloadOutputs(RemoteAction action, RemoteActionResult re
     return null;
   }
 
+//  private void waitForDownloads(ImmutableList<ListenableFuture<FileMetadata>> downloads,
```
coeuvre, line 1162: Please remove commented code.

coeuvre (approved):
LGTM! I will import after following comment is addressed.

lfpino:
Done. Feel free to merge @coeuvre 

Thanks a lot!

PR #15640, opened 2022-06-08
Cpu timec to bep

alexei-grigoriev:
This PR adds support to BEP for collecting aggregated CPU time of spawned subprocesss, grouped by mnemonic.

Main use case is to identify which mnemonics are bottlenecks and should be prioritized for optimization. Such optimizations can be done by improving associated tool chains or improving cache hit ratio for those mnemonics. Collecting the data via BEP, from all user and all CI builds, ensure that the metrics are representative in general (not only cherry-picked specific scenarios). Trends can be visualized on dashboards grouped by mnemonic.

When observing increased average builds time trends (wall clock time) for an organization, it is often not obvious to identify the cause. CPU time metrics can then not only help identifying regressions for a particular mnemonic, but by correlating CPU time with wall clock time, it can also indicate if the cause is an environmental issue or not.


crydell-ericsson:
Is there any update on the reviewing of this PR?

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -156,6 +157,16 @@ public void onActionComplete(ActionCompletionEvent event) {
   @AllowConcurrentEvents
   public void actionResultReceived(ActionResultReceivedEvent event) {
     spawnStats.countActionResult(event.getActionResult());
+    ActionStats actionStats =
+            actionStatsMap.computeIfAbsent(event.getAction().getMnemonic(), ActionStats::new);
+    Duration systemTime = event.getActionResult()
+            .cumulativeCommandExecutionSystemTime()
+            .orElse(Duration.ZERO);
```
larsrc-google, line 164: It is potentially quite misleading that the times can silently be zero if the ActionResult doesn't have any times on it.

alexei-grigoriev: I have pushed the update addressing @larsrc-google's concern. What do you think about my latest update?

larsrc-google:
Nice that you explicitly handle the case of missing times. Do you expect a given mnemonic to either always have times or never, or for it to be more mixed? And how is the `Empty` state materially different from having a zero duration? You could drop the tristate class and instead have a boolean indicating validity, we already know how many actions were executed.

alexei-grigoriev:
That's correct. For a particular mnemonic I expect mixed i.e. some actions to have times and other no not have them. For example we could have a bunch of tests which are executed remotely but some of them are tagged for local execution and only locally executed would bring times. And in this case I consider timings totally not consistent.

alexei-grigoriev:
@larsrc-google Do You mean something like that:
systemTimeValid = false;
if (numActions ==0) systemTimeValid = true;
if (actionStats.systemTime.isEmpty()) systemTimeValid = false;
else { accumulate time....}

so that instead of Empty state handling we rely on the actions counter

looks a bit more tangled but way less verbose to me.

larsrc-google:
A wrapper around the Accumulator/Adder that tracks validity would make sense. Then it could also have an interface that takes Duration and just use the long value internally.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +353,50 @@ private TimingMetrics finishTimingMetrics() {
     return timingMetrics.build();
   }
 
+  static class DurationAdder {
```
larsrc-google, line 356: Could you add a Javadoc? In particular about the invalidation logic.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +353,50 @@ private TimingMetrics finishTimingMetrics() {
...
+    public void add(Duration duration) {
+      if (this.valid.get())
+        this.duration.add(duration.toMillis());
+    }
+
+    public void invalidate() {
+      this.valid.set(false);
+    }
+
+    public boolean isEmpty() {
```
larsrc-google, line 375: Please rename this. Or make `get()` return an Optional. If you do an Optional, the new BuildActionData parts could be nicely expressed with lambdas.

src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
```diff
@@ -862,6 +862,10 @@ message BuildMetrics {
       // When the last action of this type ended being executed, in
       // milliseconds from the epoch.
       int64 last_ended_ms = 4;
+
+      // Accumulated CPU time of all spawned actions of this type.
```
larsrc-google, line 866: Document that this is not set if any of the actions did not report a time.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +353,50 @@ private TimingMetrics finishTimingMetrics() {
...
+    final LongAdder duration;
+    final AtomicBoolean valid;
+
+    public DurationAdder() {
+      valid = new AtomicBoolean(true);
+      duration = new LongAdder();
+    }
+
+    public void add(Duration duration) {
+      if (this.valid.get())
```
alexei-grigoriev, line 367: @larsrc-google  Do You think this check makes sense? It might be cheaper to sum all values even if the result is inconsistent than executing his check.

larsrc-google: It'd waste some memory to not have this check. Summing should definitely have it. I would not micro-optimize without data behind it.

alexei-grigoriev: @larsrc-google actually, I already pushed a version without that check

larsrc-google: That's probably slower, now that I think about it. The `.get()` method is probably faster than the `.add()` method. `.add()` needs to do some fancy things to handle multithreading, possibly including acquiring a lock, while `.get()` is non-atomic, which is fine in this case.

larsrc-google: Since invalidate only ever sets to true, it doesn't need to be compareAndSet. The valid bit is a latch, essentially, but there is no existing data structure for that.

alexei-grigoriev: Changed it to the simple 'if' check

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +353,63 @@ private TimingMetrics finishTimingMetrics() {
...
+    final AtomicBoolean valid;
+
+    public DurationAdder() {
+      valid = new AtomicBoolean(true);
+      duration = new LongAdder();
+    }
+
+    public void add(Duration duration) {
+      if (valid.get())
+        this.duration.add(duration.toMillis());
```
alexei-grigoriev, line 376: @larsrc-google What do You think about the precision loss here? Wouldn't it be better to switch to microseconds or even nanoseconds here? Initially I was concerned that it might be overflown but maxlong is over 100 days in nanoseconds which should be enough for builds. On the other hand I can imagine some massively parallelized software testing farm

larsrc-google: Microseconds might be an idea, I know of a few cases where time gets down in the low millis. Nanoseconds would be useless for anything I can think of. Just the fact that we're switching to a different process, let alone creating one, will take microseconds.

alexei-grigoriev: My bad, I was wrong. It is almost 300 years actually.

alexei-grigoriev:
@larsrc-google what do you think about my latest commit using microseconds? Would you like something more from me to land this?

larsrc-google:
Looks good overall, I'll add some smaller comments soon.

src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
```diff
@@ -862,6 +862,11 @@ message BuildMetrics {
...
+      // This is not set if any of the actions did not report a time
```
larsrc-google, line 867: Better wording: "This is only set if all the actions reported a time".

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -235,11 +253,34 @@ private BuildMetrics createBuildMetrics() {
...
+    ActionData.Builder builder = ActionData.newBuilder()
+        .setMnemonic(actionStats.mnemonic)
+        .setFirstStartedMs(
+            nanosToMillisSinceEpochConverter.toEpochMillis(
+                actionStats.firstStarted.longValue()))
+        .setLastEndedMs(
+            nanosToMillisSinceEpochConverter.toEpochMillis(
+                actionStats.lastEnded.longValue()))
+        .setActionsExecuted(actionStats.numActions.get());
+    if (actionStats.systemTime.sum().isPresent()) {
```
larsrc-google, line 268: Nit: I would prefer the result of `sum()` assigned to a variable, mainly to decrease verbosity, though also to make it clear that there can't be any race conditions.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +354,62 @@ private TimingMetrics finishTimingMetrics() {
...
+  /**
+   * Millisecond based duration adder which maintain initially zero long sum of microseconds.
+   * May be accessed from multiple threads.
+   * As it maintains duration as long microseconds nanoseconds will be ignored.
+   * The value can be marked invalid with {@code invalidate()}.
+   * For example if the system receives some actions with unknown duration the value is getting unknown (invalid).
+   */
+  static class DurationAdder {
+
+    final LongAdder duration;
```
larsrc-google, line 366: Nit: Rename to `duration_ms` to make the unit clear.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -322,17 +354,62 @@ private TimingMetrics finishTimingMetrics() {
     return timingMetrics.build();
   }
 
+  /**
+   * Millisecond based duration adder which maintain initially zero long sum of microseconds.
```
larsrc-google, line 358: Should be "Microsecond-based duration adder..."

larsrc-google (approved):

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
alexei-grigoriev: I guess, you mean 'us', as the unit is microseconds (μs).

alexei-grigoriev:
I am a bit confused with it. 
The darwin xcode tests are failing, despite I didn't make any specific changes neither for xcode nor for darwin.

larsrc-google:
Ugh, not again. There has been a number of failures on the Darwin tests since a recent update. @fweikert 

Don't worry about trying to fix that, just let me know when you're ready for further review.

alexei-grigoriev:
> just let me know when you're ready for further review

It is ready.

larsrc-google (approved):

sgowroji:
Hi @larsrc-google, Since I can see that this PR has been approved, please let me know whether I should proceed with importing it. Thanks!

larsrc-google:
@sgowroji Yes, you can go ahead, if syncing to head fixes the failing test.

@alexei-grigoriev I'm guessing the test failure is unrelated, if it sticks around when moved to head you'll need to fix it.

larsrc-google:
Unfortunately, we made a change in the metrics code (replacing Duration with ints to save both memory and time), and now your change needs to be updated. I would suggest changing your DurationAdder to just work with the ints, even though you lose micro-second precision.

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
```diff
@@ -246,11 +257,32 @@ private BuildMetrics createBuildMetrics() {
...
+        .setFirstStartedMs(
+            nanosToMillisSinceEpochConverter.toEpochMillis(
+                actionStats.firstStarted.longValue()))
+        .setLastEndedMs(
+            nanosToMillisSinceEpochConverter.toEpochMillis(
+                actionStats.lastEnded.longValue()))
+        .setActionsExecuted(actionStats.numActions.get());
+    long systemTime = actionStats.systemTime.get();
+    if (systemTime > 0) {
+      builder.setSystemTime(com.google.protobuf.util.Durations.fromMillis(systemTime));
```
larsrc-google, line 274: Please use an import for Durations, also below.

larsrc-google (approved):

src/main/java/com/google/devtools/build/lib/metrics/MetricsCollector.java
alexei-grigoriev: done

alexei-grigoriev:
Hi @larsrc-google 
Could You please check why it has been closed unmerged?

sgowroji:
Hi @alexei-grigoriev, PR is merged at 5f1a5709b34ca578f9bbb995d62cd028dac3b03c.

PR #15648, opened 2022-06-09
Deduplicate paths between old and new PYTHONPATH

ericsong:
Resolves https://github.com/bazelbuild/bazel/issues/15649 by deduplicating the PYTHONPATH after the old and new one has been joined.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -298,13 +298,15 @@ def Main():
   # keep first occurrence only.
   python_path_entries = [
     GetWindowsPathWithUNCPrefix(d)
-    for d in Deduplicate(python_path_entries)
+    for d in python_path_entries
   ]
 
   old_python_path = os.environ.get('PYTHONPATH')
   python_path = os.pathsep.join(python_path_entries)
   if old_python_path:
     python_path += os.pathsep + old_python_path
+
+  python_path = os.pathsep.join(Deduplicate(python_path.split(os.pathsep)))
 
   if IsWindows():
     python_path = python_path.replace('/', os.sep)
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -302,11 +302,10 @@ def Main():
   ]
 
   old_python_path = os.environ.get('PYTHONPATH')
-  python_path = os.pathsep.join(python_path_entries)
   if old_python_path:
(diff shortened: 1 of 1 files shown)


ericsong:
Not sure how to run rerun the `cla/google` check but I have since signed the CLA (with the correct email this time).

ckolli5:
Hello @ericsong, I rerun the cla/google check and it is passed. Thanks!

comius:
I have a question. Is there a way to do this deduplication one level higher, that is when stub parameters are being substituted? Bazel has a lot of machinery to do this, and it might be nicer if the duplicates never get into the template.

ericsong:
I haven't tried it out yet but with that approach, we would be swapping it out [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L175) ?

I'm not sure if that would be possible since the duplicates are between that and the PYTHONPATH at runtime. We wouldn't know what [old_python_path](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt#L304) would be when the python script is being generated.

ericsong:
@comius thoughts on the above?

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -298,14 +298,16 @@ def Main():
...
+    for d in python_path_entries
   ]
 
   old_python_path = os.environ.get('PYTHONPATH')
   python_path = os.pathsep.join(python_path_entries)
   if old_python_path:
     python_path += os.pathsep + old_python_path
 
+  python_path = os.pathsep.join(Deduplicate(python_path.split(os.pathsep)))
+
```
comius, line 310: Current implementation with joining, splitting, joining, is not the most readable.

Perhaps rewrite this block, by adding old_python_path to python_path_entries and then deduplicate and join with path separator.

comius (changes requested):

comius:
> I haven't tried it out yet but with that approach, we would be swapping it out [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L175) ?
> 
> I'm not sure if that would be possible since the duplicates are between that and the PYTHONPATH at runtime. We wouldn't know what [old_python_path](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt#L304) would be when the python script is being generated.

Thanks @ericsong. Since it's coming from the environment, it in fact isn't possible to move it one layer higher.

ericsong:
thanks for the review @comius . I made some changes per your comment.

comius (approved):

PR #15655, opened 2022-06-10
Pass singleton list to transition for allow_multiple setting default

### LINKED ISSUE #15653: string flag with allow_multiple=True represented as string or list depending on the number of times it was used
### Description of the bug:

```
multistring_flag = rule(
    implementation = _multistring_impl,
    build_setting = config.string(flag = True, allow_multiple = True),
)
```
When such flag is used in transition corresponding setting is string if the flag was used once or it is list if the flag was used more than once.

I provided compact repro here: https://github.com/Bazel-snippets/string_flag
Command to run: `bazel build test`

The repro defines two very similar flags:
```
string_list_flag = rule(
    build_setting = config.string_list(flag = True),
)
```
and
```
multistring_flag = rule(
    build_setting = config.string(flag = True, allow_multiple = True),
)
```
Rule `test` prints the value of both flags and in both cases it is a list.

Rule `test` also uses those flags in a transition like this:
```
def _test_transition_impl(settings, attr):
    string_list_flag = settings["//:string_list_flag"]
    multistring_flag = settings["//:multistring_flag"]
```
In this case `string_list_flag` is again a list as it should be, but `multistring_flag` is a string, which is a bug.
Interestingly if `multistring_flag` is used more than once on the command line then it is a list too.
Expected behavior: it should be a list always.

Mac, Linux, Windows

5.1.1

fmeum:
Previously, the default value of a user-defined string setting with
`allow_multiple = True`, which is internally represented as a string,
was passed to transitions as a string, not a list containing the single
string.

Fixes #15653

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/StarlarkBuildSettingsDetailsFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/StarlarkBuildSettingsDetailsFunction.java
@@ -18,6 +18,7 @@ import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.devtools.build.lib.packages.RuleClass.Builder.STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.devtools.build.lib.analysis.starlark.StarlarkBuildSettingsDetailsValue;
@@ -87,7 +88,13 @@ final class StarlarkBuildSettingsDetailsFunction implements SkyFunction {
               .collect(
                   toImmutableMap(
                       Rule::getLabel,
-                      rule -> rule.getAttr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME)));
+                      rule -> {
+                        if (rule.getRuleClassObject().getBuildSetting().allowsMultiple()) {
+                          return ImmutableList.of(
+                              rule.getAttr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME));
+                        }
+                        return rule.getAttr(STARLARK_BUILD_SETTING_DEFAULT_ATTR_NAME);
+                      }));
       ImmutableMap<Label, Type<?>> buildSettingToType =
           actualRules.stream()
               .collect(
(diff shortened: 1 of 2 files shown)


fmeum:
@gregestren Another edge case that wouldn't arise if `allow_multiple` flags used `attr.string_list`. Do you think that alternatives such as https://github.com/bazelbuild/bazel/pull/14911 could be considered in time for Bazel 6? I'm a bit worried that usage of `allow_multiple` is rising and migrating over to a better representation will become more painful over time.

gregestren:
Would it help for you, I, and @aranguyen to sync on this?

The proliferation of issues, PRs, and designs both illustrate the relevance and is making it harder for me to keep up (all my own fault!) I could find a sync recap helpful.

fmeum:
I think it would. I'm back on June 30, feel free to send an invite.

gregestren:
Is https://github.com/bazelbuild/bazel/issues/15653#issuecomment-1152598263 still active even with this PR?

fmeum:
> Is https://github.com/bazelbuild/bazel/issues/15653#issuecomment-1152598263 still active even with this PR?

Yes, I don't expect the current PR to fix that bug.

gregestren:
Where were we at with this PR, in the context of the wider discussion we all had last month?

fmeum:
> Where were we at with this PR, in the context of the wider discussion we all had last month?

It's another reason for getting rid of `allow_multiple`, but is unfortunately needed as long as that parameter is still used on `attr.string()`. Although we may of course just call it broken and refrain from fixing it now that it has been deprecated. @gregestren What do you think?

gregestren:
Since the code is here, we might as well have it work a bit better as long as this API exists.

gregestren:
Do you know what's up with the branch conflicts?

fmeum:
@gregestren https://github.com/bazelbuild/bazel/commit/2f7d965287ddfa056b169cf16144d05f78d03c7d caused a large conflict. CI will tell whether I found the right new place for the updated logic.

fmeum:
@aranguyen Could you review this PR? It has been stuck for a while, maybe we can move it forward.

aranguyen:
@fmeum sure I can help. I have no more capacity this week but I can review next week if that is okay with you.

fmeum:
@aranguyen Friendly ping, would be great to get this into Bazel 6.

aranguyen:
@fmeum it's on my list. I'm also working on another 6.0 release blocker. Sorry for the delay.

aranguyen:
LGTM! @sgowroji FYI, I've already started the internal merge.

Issue #14654, opened 2022-01-27
`--experimental_remote_cache_compression` causes incomplete writes


BalestraPatrick:
### Description of the problem / feature request:

We upgraded to Bazel 5.0.0 and are currently trying out the `--experimental_remote_cache_async`. During our build, I see over 1000 messages printed out to stdout that look like this possibly caused by `BulkTransferException`s:
```
WARNING: Remote Cache: 3 errors during bulk transfer
com.google.devtools.build.lib.remote.common.BulkTransferException: 3 errors during bulk transfer
	at com.google.devtools.build.lib.remote.util.RxUtils$BulkTransferExceptionCollector.onResult(RxUtils.java:91)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableCollectSingle$CollectSubscriber.onNext(FlowableCollectSingle.java:94)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableFlatMapSingle$FlatMapSingleSubscriber.innerSuccess(FlowableFlatMapSingle.java:173)
	at io.reactivex.rxjava3.internal.operators.flowable.FlowableFlatMapSingle$FlatMapSingleSubscriber$InnerObserver.onSuccess(FlowableFlatMapSingle.java:342)
	at io.reactivex.rxjava3.internal.operators.single.SingleDoFinally$DoFinallyObserver.onSuccess(SingleDoFinally.java:73)
	at io.reactivex.rxjava3.internal.observers.ResumeSingleObserver.onSuccess(ResumeSingleObserver.java:46)
	at io.reactivex.rxjava3.internal.operators.single.SingleJust.subscribeActual(SingleJust.java:30)
	at io.reactivex.rxjava3.core.Single.subscribe(Single.java:4855)
	at io.reactivex.rxjava3.internal.operators.single.SingleResumeNext$ResumeMainSingleObserver.onError(SingleResumeNext.java:80)
	at io.reactivex.rxjava3.internal.operators.completable.CompletableToSingle$ToSingle.onError(CompletableToSingle.java:73)
	at io.reactivex.rxjava3.internal.operators.completable.CompletableCreate$Emitter.tryOnError(CompletableCreate.java:91)
	at io.reactivex.rxjava3.internal.operators.completable.CompletableCreate$Emitter.onError(CompletableCreate.java:77)
	at com.google.devtools.build.lib.remote.util.RxFutures$OnceCompletableOnSubscribe$1.onFailure(RxFutures.java:102)
[...]

PR #15669, opened 2022-06-13
Chunker: Always seek on the uncompressed stream.

benjaminp:
The `WriteRequest.write_offset` field has bizarre semantics during compressed uploads as documented in the remote API protos: https://github.com/bazelbuild/remote-apis/blob/3b4b6402103539d66fcdd1a4d945f660742665ca/build/bazel/remote/execution/v2/remote_execution.proto#L241-L248 In particular, the write offset of the first `WriteRequest` refers to the offset in the uncompressed source.

This change ensures we always seek the uncompressed stream to the correct offset when starting an upload call. The old code could fail to resume compressed uploads under some conditions. The `progressiveCompressedUploadShouldWork` test purported to exercise this situation. The test, however, contained the same logic error as the code under test.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
@@ -144,17 +144,17 @@ public class Chunker {
   }
 
   /**
-   * Seek to an offset, if necessary resetting or initializing
+   * Seek to an offset in the source stream.
    *
-   * <p>May close open resources in order to seek to an earlier offset.
+   * <p>May close and reopen resources in order to seek to an earlier offset.
    */
   public void seek(long toOffset) throws IOException {
-    if (toOffset < offset) {
+    if (initialized && toOffset >= offset && !compressed) {
+      ByteStreams.skipFully(data, toOffset - offset);
+    } else {
       reset();
+      initialize(toOffset);
     }
-    maybeInitialize();
-    ByteStreams.skipFully(data, toOffset - offset);
-    offset = toOffset;
     if (data.finished()) {
       close();
     }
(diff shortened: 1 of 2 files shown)


brentleyjones:
Would this fully fix https://github.com/bazelbuild/bazel/issues/14654 then? Might be worth an inclusion in a 5.3 if one gets made.

coeuvre (approved):

brentleyjones:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -144,17 +144,17 @@ public void reset() throws IOException {
...
-   * Seek to an offset, if necessary resetting or initializing
+   * Seek to an offset in the source stream.
    *
-   * <p>May close open resources in order to seek to an earlier offset.
+   * <p>May close and reopen resources in order to seek to an earlier offset.
    */
   public void seek(long toOffset) throws IOException {
-    if (toOffset < offset) {
+    if (initialized && toOffset >= offset && !compressed) {
+      ByteStreams.skipFully(data, toOffset - offset);
```
jgao54, line 153: Was just reviewing the release notes today for bazel 5.3.0 and came across this.

it looks like with this change, offset no longer updated here when skipFully is called. Just want to sanity check that this is the intentional behavior? (I am not super familiar with bazel internals and how seek is called, but just worried this would result in extra bytes discarded if offset is not updated)

benjaminp: Thanks; this is a bug. It think it's unlikely to be triggered in practice, since seeking an initialized chunker forward is rare.

jgao54: got it, thanks for clarifying!

PR #15680, opened 2022-06-15
Include symlink targets in inputs to Android data binding actions.

### LINKED ISSUE #15678: Action input symlink into directory doesn't work with --remote_download_minimal
### Description of the bug:

Actions whose inputs contain a symlink into a directory (tree artifact) fail to build remotely with `--remote_download_minimal`.

Minimal repro at https://github.com/tjgq/bazel-symlink-into-dir.

Some notes:

* Doesn't repro without `--remote_download_minimal`.
* Doesn't repro when run locally or with a local `--disk_cache`.
* Symlinks into files (i.e., regular artifacts) work fine.

Linux

development version

Bazel built near head @ d7eaa0b63fee7a2878a8cc78f780bda5f59e1c96. However, issue also exists in 5.2.0 (i.e., it's not a regression).

```text
git@github.com:tjgq/bazel-symlink-into-dir.git
141c22a32af5ed47276e7a76921637ec373774e6
141c22a32af5ed47276e7a76921637ec373774e6
```

_No response_

_No response_

tjgq:
The intended semantics of a symlink as an action input is that only the
symlink itself is an input, not the file/directory it points to. However,
due to sandboxing limitations, such an action still executes successfully
locally, though not remotely.

Note that this change is necessary but not sufficient to make these actions
compatible with remote execution; we also need to fix #15678.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidCommon.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidCommon.java
@@ -515,17 +515,17 @@ public class AndroidCommon {
       jarsProducedForRuntime.add(resourceApk.getResourceJavaClassJar());
     }
 
-    ImmutableList<Artifact> additionalJavaInputsFromDatabinding = null;
+    NestedSet<Artifact> additionalJavaInputsFromDatabinding = null;
     if (shouldCompileJavaSrcs) {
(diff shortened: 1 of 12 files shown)


tjgq:
Per internal discussion, including the symlinks on their own in the action inputs ought to work, so #15678 is actually the only issue that needs fixing here. Therefore, I'm abandoning this PR.

PR #15700, opened 2022-06-18
Fix artifact/symlink mismatch detection 

fmeum:
The check in ActionMetadataHandler that an output declared to be a
symlink is indeed created as a symlink by the action was ineffective as
it would only run if a stat of the output showed that is already was a
symlink. The test only passed by accident since it runs sandboxed and
the sandbox setup would call Path.readSymbolicLink on what it expects to
be a symlink. As this does not happen on Windows, the test correctly
fails there.

This is fixed by calling Path.readSymbolicLink on the output path of
an expected symlink and handling errors appropriately rather
than
assuming the type returned by stat matches the artifact type (file vs symlink).

With this issue fixed, bazel_symlink_test can be enabled on Windows with
the following test-only changes:
* Pass --windows_enable_symlinks as a startup option.
* Use relative instead of absolute symlink targets as these have
  consistent shape under Unix and Windows.
* Use python to create symlinks rather than `ln -s`, which doesn't seem
  to be able to create unresolved symlinks on Windows.

Work towards #10298

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java
@@ -600,6 +600,12 @@ final class ActionMetadataHandler implements MetadataHandler {
     checkState(!artifact.isTreeArtifact() && !artifact.isMiddlemanArtifact(), artifact);
 
     Path pathNoFollow = artifactPathResolver.toPath(artifact);
+    // If we expect a symlink, we can readlink it directly and handle errors appropriately - there
+    // is no need for the stat below.
+    if (artifact.isSymlink()) {
+      return FileArtifactValue.createForUnresolvedSymlink(pathNoFollow);
+    }
+
     RootedPath rootedPathNoFollow =
         RootedPath.toRootedPath(
             artifactPathResolver.transformRoot(artifact.getRoot().getRoot()),
@@ -616,10 +622,6 @@ final class ActionMetadataHandler implements MetadataHandler {
     if (statNoFollow == null || !statNoFollow.isSymbolicLink()) {
       return fileArtifactValueFromStat(
           rootedPathNoFollow, statNoFollow, digestWillBeInjected, xattrProvider, tsgm);
-    }
-
-    if (artifact.isSymlink()) {
-      return FileArtifactValue.createForUnresolvedSymlink(pathNoFollow);
     }
 
     // We use FileStatus#isSymbolicLink over Path#isSymbolicLink to avoid the unnecessary stat
--- a/src/test/shell/bazel/bazel_symlink_test.sh
+++ b/src/test/shell/bazel/bazel_symlink_test.sh
@@ -101,6 +101,22 @@ def _write_impl(ctx):
 write = rule(implementation = _write_impl, attrs = {"contents": attr.string()})
 EOF
 
+  # We use python rather than a simple ln since the latter doesn't handle dangling symlinks on
+  # Windows.
+  mkdir -p symlink_helper
+  cat > symlink_helper/BUILD <<EOF
+py_binary(
+    name = "symlink_helper",
+    srcs = ["symlink_helper.py"],
+    visibility = ["//visibility:public"],
+)
+EOF
+
+  cat > symlink_helper/symlink_helper.py <<EOF
+import os
+import sys
+os.symlink(*sys.argv[1:])
+EOF
 }
 
 function test_smoke() {
@@ -228,17 +244,6 @@ EOF
 
 function test_file_instead_of_symlink() {
   mkdir -p a
-  cat > a/MakeSymlink.java <<'EOF'
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
-public class MakeSymlink {
-  public static void main(String[] args) throws IOException {
-    Files.createSymbolicLink(Paths.get(args[0]), Paths.get(args[1]));
-  }
-}
-EOF
-
   cat > a/a.bzl <<'EOF'
 def _bad_symlink_impl(ctx):
     symlink = ctx.actions.declare_symlink(ctx.label.name)
@@ -266,12 +271,6 @@ EOF
   cat > a/BUILD <<'EOF'
 load(":a.bzl", "bad_symlink", "bad_write")
 
-java_binary(
-    name = "MakeSymlink",
-    srcs = ["MakeSymlink.java"],
-    main_class = "MakeSymlink",
-)
-
 bad_symlink(name="bs", link_target="bad/symlink")
 genrule(name="bsg", srcs=[":bs"], outs=["bsgo"], cmd="echo BSGO > $@")
 
@@ -282,8 +281,8 @@ genrule(
     name="bg",
     srcs=[],
     outs=["bgo"],
-    exec_tools = [":MakeSymlink"],
-    cmd = "$(execpath :MakeSymlink) $@ bad/symlink",
+    cmd = "$(location //symlink_helper) bad/symlink $@",
+    tools = ["//symlink_helper"],
 )
 EOF
 
@@ -329,25 +328,14 @@ EOF
 
 function test_symlink_created_from_spawn() {
   mkdir -p a
-  cat > a/MakeSymlink.java <<'EOF'
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
-public class MakeSymlink {
-  public static void main(String[] args) throws IOException {
-    Files.createSymbolicLink(Paths.get(args[0]), Paths.get(args[1]));
-  }
-}
-EOF
-
   cat > a/a.bzl <<'EOF'
 def _a_impl(ctx):
     symlink = ctx.actions.declare_symlink(ctx.label.name + ".link")
     output = ctx.actions.declare_file(ctx.label.name + ".file")
     ctx.actions.run(
         outputs = [symlink],
-        executable = ctx.executable._make_symlink,
-        arguments = [symlink.path, ctx.attr.link_target],
+        executable = ctx.executable._link,
+        arguments = [ctx.attr.link_target, symlink.path],
         inputs = depset([]),
     )
     ctx.actions.run_shell(
@@ -361,11 +349,11 @@ a = rule(
     implementation = _a_impl, 
     attrs = {
         "link_target": attr.string(),
-        "_make_symlink": attr.label(
-            default = ":MakeSymlink",
+        "_link": attr.label(
+            default = "//symlink_helper",
             executable = True,
             cfg = "exec",
-        )
+        ),
     }
 )
 EOF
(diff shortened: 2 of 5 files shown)


fmeum:
@lberki Could you take a look? This should handle the Windows part of #10298 and fixes an actual bug along the way.

fmeum:
@coeuvre Could you take over the review while @lberki is OOO?

fmeum:
@coeuvre Friendly ping

src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java
```diff
@@ -600,6 +600,12 @@ private static FileArtifactValue fileArtifactValueFromArtifact(
     checkState(!artifact.isTreeArtifact() && !artifact.isMiddlemanArtifact(), artifact);
 
     Path pathNoFollow = artifactPathResolver.toPath(artifact);
+    // Handle the case where a symlink is expected before stat-ing the output so that the case where
+    // it incorrectly is a file triggers the expected NotASymlinkException.
+    if (artifact.isSymlink()) {
+      return FileArtifactValue.createForUnresolvedSymlink(pathNoFollow);
```
alexjski, line 606: I think that comment is not useful as is (it talks about the history of the code more than the current state). What I would say is that we start by eagerly running readSymbolicLink to avoid an unnecessary stat in the correct case. My point is that stat could be used to detect whether that is a symbolic link, it just happened not to be used this way. Your approach is nice in that it saves IO (stat then readlink vs just readlink and analyze errors)--I think that is the clue here.

alexjski (changes requested):

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -107,170 +107,141 @@ EOF
...
-
   mkdir -p a
   cat > a/BUILD <<EOF
 load("//symlink:symlink.bzl", "dangling_symlink")
-dangling_symlink(name="a", link_target="/nonexistent")
+dangling_symlink(name="a", link_target="non/existent")
 EOF
 
-  bazel build --experimental_allow_unresolved_symlinks //a:a || fail "build failed"
+  bazel --windows_enable_symlinks build --experimental_allow_unresolved_symlinks //a:a || fail "build failed"
```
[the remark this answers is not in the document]
fmeum: Thanks, I didn't know about this. Changed.

src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java
fmeum: Good point, modified the comment.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -343,25 +326,32 @@ load(":a.bzl", "bad_symlink")
...
 
 function test_symlink_created_from_spawn() {
-  if "$is_windows"; then
-    # TODO(#10298): Support unresolved symlinks on Windows.
-    return 0
-  fi
-
   mkdir -p a
+  cat > a/MakeSymlink.java <<'EOF'
+import java.io.IOException;
```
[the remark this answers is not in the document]
fmeum: `ln -s` and `mklink` have different expected argument order. Wouldn't I also have to put that code in an `sh_binary`? At that point, I personally prefer the "works the same everywhere" appeal of the slightly more verbose Java tool.

I could turn it into a data dep, import it via `java_import` and then wrap it in a `java_binary` for execution, but again to me that seems more complicated to follow.

What do you think of just extracting this into a test helper function that contains the source code only once?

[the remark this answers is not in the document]
alexjski: > `ln -s` and `mklink` have different expected argument order. Wouldn't I also have to put that code in an `sh_binary`?

Good point about the parameters order. You would not need to wrap that in a sh_binary, you can just plug in the string to run.

> At that point, I personally prefer the "works the same everywhere" appeal of the slightly more verbose Java tool.

That's fair, is there a reason that more verbose Java is preferred over Python here then?

> 
> I could turn it into a data dep, import it via `java_import` and then wrap it in a `java_binary` for execution, but again to me that seems more complicated to follow.
> 
> What do you think of just extracting this into a test helper function that contains the source code only once?

A helper could work. And again, please consider using something less verbose than Java.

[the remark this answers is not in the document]
alexjski: Another thing which could help is a `genrule` with `cmd_bash` and `cmd_bat` (https://bazel.build/reference/be/general#genrule).

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -282,8 +265,7 @@ genrule(
     name="bg",
     srcs=[],
     outs=["bgo"],
-    exec_tools = [":MakeSymlink"],
-    cmd = "$(execpath :MakeSymlink) $@ bad/symlink",
+    cmd = "python3 -c 'import os, sys; os.symlink(*sys.argv[1:])' bad/symlink $@",
```
alexjski, line 268: Looking good, thanks! If you can, please add a comment why we are using Python here instead of `ln` so nobody tries to "fix it" in the future.

fmeum: Added the two comments.

alexjski (approved):

src/test/shell/bazel/bazel_symlink_test.sh
alexjski: That unfortunately doesn't work since python3 is not on path for windows machines and we likely don't want to rely on that. That leaves a `py_binary` tool or something like I did here:

```
function test_symlink_created_from_spawn() {
  mkdir a
  cat > a/a.bzl <<'EOF'
def _a_impl(ctx):
    symlink = ctx.actions.declare_symlink(ctx.label.name + ".link")
    output = ctx.actions.declare_file(ctx.label.name + ".file")
    ctx.actions.run_shell(
        outputs = [symlink],
        inputs = depset([]),
        # Fall back to mklink on Windows since Cygwin cannot create danging
        # symlinks.
        # TODO(#15194): Use `command_bat` instead.
        command = r'ln -s "$1" "$2" || "${SYSTEMROOT}"/system32/cmd //C mklink "${2//\\//\\\\}" "${1//\\//\\\\}"',
        arguments = [ctx.attr.link_target, symlink.path],
    )
    ctx.actions.run_shell(
        outputs = [output],
        inputs = depset([symlink]),
        command = "echo input link is $(readlink " + symlink.path + ") > " + output.path,
    )
    return DefaultInfo(files = depset([output]))

a = rule(
    implementation = _a_impl,
    attrs = {
        "link_target": attr.string(),
    }
)
EOF

  cat > a/BUILD <<'EOF'
load(":a.bzl", "a")

a(name="a", link_target="somewhere/over/the/rainbow")
EOF

  bazel build --experimental_allow_unresolved_symlinks //a:a || fail "build failed"
  assert_contains "input link is somewhere/over/the/rainbow" bazel-bin/a/a.file
}
```

We would either need the same for the genrule or we are back to adding a tool of sorts (`py_binary`). Unfortunately `mklink` seems to want paths with `\` not `/`, therefore we cannot easily leverage that there (please see the crazy slash substitution going on there). If you like that, by all means, please copy-paste that and proceed this way. Otherwise, looks like a tool will be necessary (you can just create one out of that python script).

So sorry about that, I approved this change without seeing that the tests failed and changes needed here exceed what I feel comfortable doing with you in the author tag.

alexjski (changes requested):

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -104,34 +101,40 @@ def _write_impl(ctx):
 write = rule(implementation = _write_impl, attrs = {"contents": attr.string()})
 EOF
 
+  # We use python rather than a simple ln since the latter doesn't handle dangling symlinks on
+  # Windows.
+  mkdir -p symlink_helper
```
fmeum, line 106: Renamed it to `//symlink_helper`.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -101,6 +101,22 @@ def _write_impl(ctx):
...
+  cat > symlink_helper/BUILD <<EOF
```
alexjski, line 107: nit: 'EOF' is generally preferred unless you need variable substitution.

src/test/shell/bazel/bazel_symlink_test.sh
alexjski: Thank you!

alexjski:
Another minor edit could be useful in the commit message:

```
This is fixed by calling Path.readSymbolicLink on the output path of an
expected symlink before rather than after stat-ing it and trusting the
file type contained in the stat info.
```

I think we established that trusting stat output is actually OK, it's just that we didn't look whether it matches the expectation, not that it cannot be used.

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -370,25 +357,35 @@ def _a_impl(ctx):
     )
     return DefaultInfo(files = depset([output]))
 
-a = rule(implementation = _a_impl, attrs = {"link_target": attr.string()})
+a = rule(
+    implementation = _a_impl,
```
alexjski, line 349: FYI: I will remove the trailing space from this line as part of importing this change. You can delete that yourself if you prefer.

src/test/shell/bazel/bazel_symlink_test.sh
fmeum: Could you take care of it? I'm on mobile atm. Thanks!

src/test/shell/bazel/bazel_symlink_test.sh
fmeum: Thanks!

fmeum:
> Another minor edit could be useful in the commit message:
> 
> ```
> This is fixed by calling Path.readSymbolicLink on the output path of an
> expected symlink before rather than after stat-ing it and trusting the
> file type contained in the stat info.
> ```
> 
> I think we established that trusting stat output is actually OK, it's just that we didn't look whether it matches the expectation, not that it cannot be used.

I updated the PR description and am fine with you taking care of the nits.

Thanks for the review!

alexjski:
> > Another minor edit could be useful in the commit message:
> > ```
> > This is fixed by calling Path.readSymbolicLink on the output path of an
> > expected symlink before rather than after stat-ing it and trusting the
> > file type contained in the stat info.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think we established that trusting stat output is actually OK, it's just that we didn't look whether it matches the expectation, not that it cannot be used.
> 
> I updated the PR description and am fine with you taking care of the nits.
> 
> Thanks for the review!

more nitting in the new sentence I would replace "trusting" with something like "assuming the type matches the artifact type (file vs symlink)."

fmeum:
Changed. Not sure I wrapped the lines correctly though, so maybe reflow the text during the import.

alexjski:
> Changed. Not sure I wrapped the lines correctly though, so maybe reflow the text during the import.

Thank you! FYI, I am in the process of merging that internally now. All tests passed, so looking good. Will update the PR once it is merged. Thank you so much for creating this change (and enduring my nitpicking)! It was a pleasure to work with you on that!

alexjski:
The change was merged in https://github.com/bazelbuild/bazel/commit/666fce514b87e6901b033f3399e2d4b56a856429. Thank you for the contribution!

PR #15701, opened 2022-06-19
Adds PYTHONSAFEPATH to python_stub_template.txt

groodt:
See: https://github.com/bazelbuild/rules_python/issues/382

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -314,7 +314,7 @@ def Main():
   if runfiles_envkey:
     new_env[runfiles_envkey] = runfiles_envvalue
 
-  # Don’t prepend a potentially unsafe path to sys.path
+  # Don't prepend a potentially unsafe path to sys.path
   # See: https://docs.python.org/3.11/using/cmdline.html#envvar-PYTHONSAFEPATH
   new_env['PYTHONSAFEPATH'] = 'non_empty'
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -314,6 +314,10 @@ def Main():
   if runfiles_envkey:
     new_env[runfiles_envkey] = runfiles_envvalue
 
+  # Don't prepend a potentially unsafe path to sys.path
+  # See: https://docs.python.org/3.11/using/cmdline.html#envvar-PYTHONSAFEPATH
+  new_env['PYTHONSAFEPATH'] = 'non_empty'
```
hrfuller, line 319: Using "1" or "True" would be clearer IMHO

groodt, line 319:
```suggestion
  new_env['PYTHONSAFEPATH'] = '1'
```

src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -314,6 +314,10 @@ def Main():
   if runfiles_envkey:
     new_env[runfiles_envkey] = runfiles_envvalue
 
+  # Don't prepend a potentially unsafe path to sys.path
```
rickeylev, line 317: fwiw, i'd recommend also linking to the rules_python issue. The unsafe behavior this is trying to prevent isn't immediately obvious.
(apologies for the drive-by comment)

groodt: 👍 Good idea.

comius (approved):

PR #15711, opened 2022-06-21
Fix additional outputs in BEP when remote_download_minimal is used

exoson:
Injects optional test outputs when using --remote_download_minimal and
uses the list of injected artifacts to list them in BEP TestResult.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -35,4 +35,4 @@ build --java_language_version=11
 build --tool_java_language_version=11
 
 # User-specific .bazelrc
-try-import user.bazelrc
+try-import %workspace%/user.bazelrc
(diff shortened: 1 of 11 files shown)


exoson:
Will be fixing the tests later. Would appreciate comments on if a change along these lines would make sense.

coeuvre:
Thanks for your PR!

Since this PR is mainly about test strategy and is already beyond my limited understanding of it, I will try to find an appropriate reviewer.

Out of my head, are these additional outputs already referenced in BEP for normal build / remote build?

exoson:
@coeuvre Any updates on finding a reviewer?

coeuvre:
I went another way (#16123) to fix this issue. It requires no changes outside remote module and allow us to remove the hack in `StandaloneTestStrategy`. That is also the way how we solve similar problems internally.

Issue #5419, opened 2018-06-17
Seemingly duplicate labels


excitoon:
### Description of the problem / feature request:

Basically it is good that Bazel forbids duplicate labels. But it should not make false-positives.
For example:
```
cc_binary(
    name = "hello",
    srcs = select({ "//conditions:default" : [], "//:x86_32_windows" : ["sources/hello.cpp"] })
        + [ "sources/common.cpp" ]
        + select({ "//conditions:default" : [], "//:x64_windows" : ["sources/hello.cpp"] })
)
```

This target looks good, but in reality it won't compile:
```
ERROR: C:/users/chebotarev_v/documents/src/bazel-issues/seemingly-duplicate-labels/BUILD:17:1: Label '//:sources/hello.cpp' is duplicated in the 'srcs' attribute of rule 'hello'
```

### Feature requests: what underlying problem are you trying to solve with this feature?

Label lists can be combined with (`+`) and come from different points of BUILD files, so these false-positives can be extremely hard to workaround in a project. I believe that should be fixed.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please find an example here:
https://github.com/excitoon/bazel-issues/tree/master/seemingly-duplicate-labels

### What operating system are you running Bazel on?

Windows 10 x64

### What's the output of `bazel info release`?

```
Build label: 0.14.0
Build target: bazel-out/x64_windows-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Jun 1 13:06:29 2018 (1527858389)
Build timestamp: 1527858389
Build timestamp as int: 1527858389
```

cushon:
Consider using `collections.uniq` from https://github.com/bazelbuild/bazel-skylib to deduplicate the `srcs` of that target, e.g.:

Add this to your `WORKSPACE`:

```
git_repository(
    name = "bazel_skylib",
    remote = "https://github.com/bazelbuild/bazel-skylib.git",
    tag = "0.1.0",  # change this to use a different release
)
```

And then use:

```
load("@bazel_skylib//:lib.bzl", "collections")

cc_binary(
    name = "hello",
    srcs = collections.uniq(select({ "//conditions:default" : [], "//:x86_32_windows" : ["sources/hello.cpp"] })
        + [ "sources/common.cpp" ]
        + select({ "//conditions:default" : [], "//:x64_windows" : ["sources/hello.cpp"] }))
[...]

PR #15714, opened 2022-06-21
Docs: Include exceptions to combining selects

FaBrand:
Coming across https://github.com/bazelbuild/bazel/issues/5419 it makes sense to document these issues as exceptions to the intuitive usage of combining selects with duplicate targets

Reasoning about this is provided in https://github.com/bazelbuild/bazel/commit/497ef110c344b2a309210339f125916f355530f9 and imho it makes sense to document this behaviour

### DIFF
--- a/site/en/docs/configurable-attributes.md
+++ b/site/en/docs/configurable-attributes.md
@@ -364,11 +364,6 @@ Platforms are still under development. See the
 
 `select` can appear multiple times in the same attribute:
 
-Note: Some restrictions apply on what can be combined in the select's values:
- - Duplicate labels can appear in different paths of the same select.
- - Duplicate labels can *not* appear within the same path of a select.
- - Duplicate labels can *not* appear across multiple combined selects (no matter what path)
-
 ```python
 sh_binary(
     name = "my_target",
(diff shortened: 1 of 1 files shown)


site/en/docs/configurable-attributes.md
```diff
@@ -364,6 +364,11 @@ Platforms are still under development. See the
 
 `select` can appear multiple times in the same attribute:
 
+Note: Some restrictions apply on what can be combined in the select's values:
```
gregestren, line 367: Would this paragraph make more sense after the example? 

Also, as a grammar nit I'd omit "Note:" to get the same message through more concisely.

And one more grammar nit: add backtick around `select` so it appears code-style.

FaBrand: I put it before the example that one would not easily overlook these restrictions.

I assumed that the `Note:` would yield something like this Note-box: https://bazel.build/reference/exec-groups#using-exec-groups-for-exec-properties

image

Maybe if it's formatted like this, moving it after the example is the right choice. Then it would still catch the readers eye. wdyt?

gregestren: You're right, I'm sorry! I'd definitely support that formatting. Still after the example IMO.

gregestren (approved):

FaBrand:
@gregestren It seems like the `note` box turned out nicely. However the list got a bit mangled.
Do you know if there is a way to generate a preview of the docs locally? The wysiwyg kinda type.

I would like to fix it but trial and error via pull request is not really ideal.

image
https://bazel.build/docs/configurable-attributes#combining-selects

gregestren:
I'm not sure what the latest is, since Bazel's doc hosting infrastructure changed a few months back. @fweikert can advise.

PR #15720, opened 2022-06-22
Chunker: Always seek on the uncompressed stream.

ckolli5:
The `WriteRequest.write_offset` field has bizarre semantics during compressed uploads as documented in the remote API protos: https://github.com/bazelbuild/remote-apis/blob/3b4b6402103539d66fcdd1a4d945f660742665ca/build/bazel/remote/execution/v2/remote_execution.proto#L241-L248 In particular, the write offset of the first `WriteRequest` refers to the offset in the uncompressed source.

This change ensures we always seek the uncompressed stream to the correct offset when starting an upload call. The old code could fail to resume compressed uploads under some conditions. The `progressiveCompressedUploadShouldWork` test purported to exercise this situation. The test, however, contained the same logic error as the code under test.

Closes #15669.

PiperOrigin-RevId: 455083727
Change-Id: Ie22dacf31f15644c7a83f49776e7a633d8bb4bca

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/Chunker.java
@@ -142,17 +142,23 @@ public final class Chunker {
   }
 
   /**
-   * Seek to an offset, if necessary resetting or initializing
+   * Seek to an offset in the source stream.
    *
-   * <p>May close open resources in order to seek to an earlier offset.
+   * <p>May close and reopen resources in order to seek to an earlier offset.
+   *
+   * @param toOffset the offset from beginning of the source stream. If the source stream is
+   *     compressed, it refers to the offset in the uncompressed form to align with `write_offset`
+   *     in REAPI.
    */
   public void seek(long toOffset) throws IOException {
-    if (toOffset < offset) {
+    // For compressed stream, we need to reinitialize the stream since the offset refers to the
+    // uncompressed form.
+    if (initialized && toOffset >= offset && !compressed) {
+      ByteStreams.skipFully(data, toOffset - offset);
+    } else {
       reset();
+      initialize(toOffset);
     }
-    maybeInitialize();
-    ByteStreams.skipFully(data, toOffset - offset);
-    offset = toOffset;
     if (data.finished()) {
       close();
     }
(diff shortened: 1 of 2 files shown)


brentleyjones:
This probably requires the refactor commit. @benjaminp can you help @ckolli5 get the right commits for this cherry-pick?

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -245,18 +251,26 @@ private void maybeInitialize() throws IOException {
...
+    initialize(0);
+  }
+
+  private void initialize(long srcPos) throws IOException {
+    checkState(!initialized);
     checkState(data == null);
     checkState(offset == 0);
     checkState(chunkCache == null);
     try {
+      var src = dataSupplier.get();
```
benjaminp, line 263: You're probably going to have to write `InputStream src = dataSupplier.get()`, since Bazel 5 requires the Java 8 language level.

ckolli5: Hello @benjaminp, I tried that but still the builds are failing. Here is the [PR](https://github.com/bazelbuild/bazel/pull/15788) with the suggested changes. Could you please help us in cherry picking these changes. Thanks!

src/main/java/com/google/devtools/build/lib/remote/Chunker.java
```diff
@@ -142,17 +142,23 @@ public void reset() throws IOException {
...
+   * @param toOffset the offset from beginning of the source stream. If the source stream is
+   *     compressed, it refers to the offset in the uncompressed form to align with `write_offset`
+   *     in REAPI.
    */
   public void seek(long toOffset) throws IOException {
-    if (toOffset < offset) {
+    // For compressed stream, we need to reinitialize the stream since the offset refers to the
+    // uncompressed form.
+    if (initialized && toOffset >= offset && !compressed) {
+      ByteStreams.skipFully(data, toOffset - offset);
```
jgao54, line 157: Was just reviewing the release notes today for bazel 5.3.0 and came across this.

it looks like with this change, `offset` no longer updated here when `skipFully` is called. Just want to sanity check that this is the intentional? (I am not super familiar with bazel internals)

PR #15724, opened 2022-06-22
Remote: Always download outputs and outErr even if declared outputs are missing.

coeuvre:
So that if the unusual case described by https://github.com/bazelbuild/bazel/issues/14543 happens, we can still see the error message of the spawn (if any).


coeuvre:
@tjgq friendly ping!

tjgq (approved):

PR #15743, opened 2022-06-25
Set `--incompatible_strict_action_env` to `true` by default

vardaro:


vardaro:
> 

Hi @sgowroji, yeah i'm aware. I plan to chase down downstream consumers and flip the flag in their CI, and set the default to `true` here. I expect it to take a while though..

Issue #15677, opened 2022-06-15
Some python libraries do not support extended-length path.


dhmemi:
### Description of the bug:

Bazel modified the python packages path into [extended-length path](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd) to support paths bigger than 256 characters in this PR: https://github.com/bazelbuild/bazel/pull/2496 . But there are many python libraries that don't support extended-length path，include some python built-in packages like `os.path` .

see related issues for more ditails:
- https://github.com/bazelbuild/rules_python/issues/680
- https://github.com/sphinx-doc/sphinx/pull/10374

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

According to the [Microsoft's documentation](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd#enable-long-paths-in-windows-10-version-1607-and-later), Maybe we can support paths bigger than 256 characters without prefix `\\?\` which is not compatible with forward slash on windows 10.1607 or later version.

But it's need to be test, and I don't know if there is any other place in bazel related to this.

### Which operating system are you running Bazel on?

Windows 10

### What is the output of `bazel info release`?

release 4.1.0

### Have you found anything relevant by searching the web?

https://github.com/bazelbuild/bazel/issues/2504

UlrichEckhardt:
Interesting, I just stumbled across [this issue](https://github.com/pallets/jinja/issues/1675) in the context of Gcovr or, rather the Jinja2 template renderer it uses.

comius:
https://github.com/bazelbuild/bazel/pull/15745

PR #15745, opened 2022-06-27
Remove UNC path prefix on Windows-10.1607 or later

dhmemi:
Remove UNC path prefix on Windows 10, Version 1607, and Later platform to resolve https://github.com/bazelbuild/bazel/issues/15677 .

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
@@ -41,6 +41,13 @@ def GetWindowsPathWithUNCPrefix(path):
   if not IsWindows() or sys.version_info[0] < 3:
     return path
 
+  # Starting in Windows 10, version 1607(OS build 14393), MAX_PATH limitations have been 
+  # removed from common Win32 file and directory functions.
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt
```diff
@@ -41,6 +41,13 @@ def GetWindowsPathWithUNCPrefix(path):
   if not IsWindows() or sys.version_info[0] < 3:
     return path
 
+  # Starting in Windows 10, version 1607(OS build 14393), MAX_PATH limitations have been 
+  # removed from common Win32 file and directory functions.
+  # Related doc: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd#enable-long-paths-in-windows-10-version-1607-and-later 
+  import platform
+  if platform.win32_ver()[1] >= '10.0.14393':
+    return path
+
```
UlrichEckhardt, line 50: I honestly like the idea of this change. :+1: 

However, I'm honestly wondering what this file here represents! It looks like Python, but with a txt extension. It's indented inconsistently, using two or three spaces. Also, below, the `unicode_prefix` has _nothing_ to do with Unicode. In summary, I don't think your change will have any effect on actual code being executed.

comius (approved):

PR #15747, opened 2022-06-27
Build java_tools with jdk8

hvadehra:


Wyverald (approved):
cc @meteorcloudy @comius

Issue #15742, opened 2022-06-25
CompatDx cannot fit requested classes in a single dex file


Bencodes:
### Description of the bug:

After updating to rolling release `6.0.0-pre.20220526.1`, `CompatDx` produces a massive amount of warnings and eventually fails with the following error:

```
Error: Cannot fit requested classes in a single dex file (# methods: 227417 > 65536 ; # fields: 124517 > 65536). Try supplying a main-dex list
Exception in thread "main" java.lang.AssertionError: Failure
        at com.google.devtools.build.android.r8.CompatDx.main(CompatDx.java:478)
Caused by: com.android.tools.r8.CompilationFailedException: Compilation failed to complete, position: null
        at Version.fakeStackEntry(Version_3.3.28.java:0)
        at com.android.tools.r8.internal.Bj.a(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:75)
        at com.android.tools.r8.internal.Bj.a(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:28)
        at com.android.tools.r8.internal.Bj.a(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:27)
        at com.android.tools.r8.internal.Bj.b(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:2)
        at com.android.tools.r8.D8.run(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:6)
        at com.google.devtools.build.android.r8.CompatDx.run(CompatDx.java:624)
        at com.google.devtools.build.android.r8.CompatDx.main(CompatDx.java:472)
Caused by: com.android.tools.r8.internal.f: Cannot fit requested classes in a single dex file (# methods: 227417 > 65536 ; # fields: 124517 > 65536)
        at com.android.tools.r8.internal.DT.a(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:14)
        at com.android.tools.r8.internal.DT.a(R8_3.3.28_2aaf796388b4e9f6bed752d926eca110512a53a3f09a8d755196089c1cfdf799:22)
[...]

PR #15752, opened 2022-06-27
Add Android integration test for large multidexing

ted-xie:
Test for multidexing with >64k classes. Reproduces #15742. Merge after the multidexing error is fixed (currently this integration test fails with `bazel test src/test/shell/bazel/android:android_integration_test ` on baze 5.2.0).

@Bencodes @ahumesky


ted-xie:
@ahumesky already added a test that does this.

PR #15773, opened 2022-06-30
WIP: Preserve unnormalized unresolved symlinks

fmeum:
Fixes #14224
Work towards #10298

Issue #15454, opened 2022-05-10
--experimental_allow_unresolved_symlinks does not work when declaring a folder


GWLeal:
Hello,

I've noticed some strange behavior regarding how bazel evaluates invalid symlinks.

UC1:
When I declare an invalid symlink, in a custom rule, and I build a bazel target of that kind bazel prints out the message:
`Error in declare_symlink: actions.declare_symlink() is not allowed; use the --experimental_allow_unresolved_symlinks command line option`
As expected when the suggested option is added the target builds successfully.

UC2:
However when I declare a folder and add a invalid symlink in that folder I get:
`ERROR: (...): Failed to resolve relative path invalid_symlink inside TreeArtifact (...) The associated file is either missing or is an invalid symlink.`
Adding the --experimental_allow_unresolved_symlinks does not work in this case.

This seems like inconsistent behavior. Also it is pretty common to pack build artifacts that contain invalid symlinks that will only become valid when the package is installed on the intended target.
I looked in the documentation and did not find a way to circumvent the issue.

I prepared a minimal working example to replicate the problem: https://github.com/GWLeal/bazel_invalid_symlinks_ex

UC1:
`bazel build //test:symlink_test`

UC2:
`bazel build //test:dir_symlink_test`

I'm currently using bazel version 5.1.1

Thank you in advance for taking the time to look into this :)

lberki:
I'm not too surprised. `--experimental_allow_unresolved_symlinks` only works for "regular" artifacts not those that are part of `TreeArtifacts`. This is one of the reasons why it's still experimental...

/cc @coeuvre @alexjski 

@coeuvre maybe this is something that could conceivably be fixed as a drive-by fix while you are poking at metadata handling for remote execution? (I dorealize this is a long shot)

alexjski:
> --experimental_allow_unresolved_symlinks

What is the use case for having unresolved symlinks?

GWLeal:
> > --experimental_allow_unresolved_symlinks
> 
> What is the use case for having unresolved symlinks?

@alexjski in this case we generate an image with the application filesystem that is mounted in one of the target's partitions as read-only. One of the symlinks points to a file in the OS partition in the target. But I would say that having unresolved symlinks in the context of packaging is common.

lberki:
After some discussion with @lfpino we came to the conclusion that there are two missing pieces before we can stabilize this:

1. Support on Bazel's RE interface (this is known to not work)
[...]

PR #15781, opened 2022-06-30
Add RBE support for generated unresolved symlinks

fmeum:
Also adds support for such symlinks to the remote worker implementation.

Work towards #10298

### DIFF
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -2964,4 +2964,55 @@ EOF
       @other_repo//test >& $TEST_log || fail "Test should pass"
 }
 
+function test_unresolved_symlink_input() {
+  mkdir -p symlink
+  touch symlink/BUILD
+  cat > symlink/symlink.bzl <<EOF
+def _dangling_symlink_impl(ctx):
+    symlink = ctx.actions.declare_symlink(ctx.label.name)
+    ctx.actions.symlink(
+        output = symlink,
+        target_path = ctx.attr.link_target,
+    )
+    return DefaultInfo(files = depset([symlink]))
+
+dangling_symlink = rule(
+    implementation = _dangling_symlink_impl,
+    attrs = {"link_target": attr.string()},
+)
+EOF
+
+  mkdir -p pkg
+  cat > pkg/BUILD <<'EOF'
+load("//symlink:symlink.bzl", "dangling_symlink")
+dangling_symlink(name="a", link_target="non/existent")
+genrule(
+    name = "b",
+    srcs = [":a"],
+    outs = ["b.txt"],
+    cmd = "readlink $(location :a) > $@",
+)
+EOF
+
+  bazel \
+    --windows_enable_symlinks \
+    build \
+    --experimental_allow_unresolved_symlinks \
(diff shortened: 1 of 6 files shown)


fmeum:
@coeuvre Can you review this PR?

This takes care of 1. in https://github.com/bazelbuild/bazel/issues/15454#issuecomment-1171405393 for generated symlinks. 

Supporting source symlinks would require additional, possibly backwards incompatible changes. Since these are actually quite important for the new `rules_js` approach to NPM builds, I would be interested in tackling this aspect of RBE symlink. @lfpino Have you perhaps already thought about the source symlink case?

src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/OnDiskBlobStoreCache.java
```diff
@@ -52,6 +54,14 @@ public void downloadTree(
       getFromFuture(downloadFile(context, dst, file.getDigest()));
       dst.setExecutable(file.getIsExecutable());
     }
+    for (SymlinkNode symlink : directory.getSymlinksList()) {
+      Path dst = rootLocation.getRelative(symlink.getName());
+      // TODO(fmeum): The following line is not generally correct: The remote execution API allows
+      //  for non-normalized symlink targets, but the normalization applied by PathFragment.create
+      //  does not take directory symlinks into account. However, Bazel's file system API does not
+      //  currently offer a way to specify a raw String as a symlink target.
+      dst.createSymbolicLink(PathFragment.create(symlink.getTarget()));
```
fmeum, line 63: This can be fixed after https://github.com/bazelbuild/bazel/pull/15773.

fmeum:
@coeuvre Friendly ping

bazaglia:
@coeuvre Many Node.js users are waiting for this for a long time to have RBE with rules_js. It would be nice to get it reviewed. 😄

coeuvre:
Sure. I will definitely review it once I get the time.

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -2792,4 +2792,53 @@ EOF
...
+  bazel build \
+    --experimental_allow_unresolved_symlinks \
+    --spawn_strategy=remote \
+    --remote_executor=grpc://localhost:${worker_port} \
+    //pkg:b &>$TEST_log || fail "expected build to succeed"
+  [[ $(cat bazel-bin/pkg/b.txt) =~ .*/non/existent ]] || fail "expected symlink target to be non/existent"
+
+  bazel clean --expunge
+  bazel \
+    --windows_enable_symlinks \
```
coeuvre, line 2834: Can we check whether it is running on windows to conditionally add the flag `--windows_enable_symlinks` instead of running the build twice?

coeuvre (approved):

coeuvre:
LGTM. Please rebase and update the test.

src/test/shell/bazel/remote/remote_execution_test.sh
fmeum: I just added `--windows_enable_symlinks` to both builds - it is a no-op on non-Windows platforms and the difference between the two builds should only be `--experimental_remote_merkle_tree_cache`.

Issue #6443, opened 2018-10-18
Support both Python 2 and 3 in tools (i.e. host configuration)


brandjon:
We need to provide a way for `genrule`s to invoke python 3 binaries as tools. Currently the `tools` attribute uses the special HOST config, which prevents switching the PY2/PY3 mode. Design doc to come.

Targeting 2018 Q4.

brandjon:
(P1 for a feature request essentially means "this quarter", so not as urgent as P1 for a bug.)

brandjon:
This crosses over into Configurability team's domain and I understand they're prioritizing it, so until we understand more I'm marking this P2.

brandjon:
This is blocked on work by Configurability team to add an "[execution transition](https://github.com/bazelbuild/proposals/blob/master/designs/2019-02-12-execution-transitions.md)". Unlike the host transition, this transition will allow for subsequent transitions such as changes to the Python version.

Once that's in, we can add new attributes to rules to use this transition in place of the host transition, e.g. create `genrule.exec_tools` to replace `genrule.tools`. Users can incrementally migrate dependencies from `tools` to `exec_tools` and likewise for other rules.

Over the long term, the difference between the host configuration and the "execution platform configuration" will decrease, as more things are migrated to use platforms and toolchains rather than ad hoc flags.

brandjon:
Update: This is being actively worked on by configurability team but there likely won't be anything in the 0.27 release, which is the release that will turn on Python toolchains. See #7899 for migration-related concerns.

The new plan may in fact be to not create an `exec_tools` attribute but just change `genrule`'s `tools` attribute to use the execution config instead of host config.

brandjon:
Latest update is that `exec_tools` will be added to `genrule` until blockers for migrating `tools` to exec config can be worked out. For starlark rules there's already a `cfg="exec"` value for attribute definitions.

brandjon:
FYI `exec_tools` for `genrule` will be in 0.28 (implemented in d97392fb03a1deacbe655883564530ab856073c1).

speedskater:
@brandjon Thanks for this hint of using `cfg="exec"`. It safed my day. Do you know if there is a specific reason, why this is not part of the official documentation?

lberki:
I'll close this since Python 2 support (at least for the native rules) is going away soon.

PR #15783, opened 2022-07-01
Remove ancient PY2 warning messages from python_stub_template.txt

jvolkman:
These messages are rarely helpful and often misleading. The associated TODO (#6443) was closed years ago.


jvolkman:
Turns out there was a test case. I've removed that as well and the tests are passing now.

groodt:
Should this be removed as well? https://github.com/bazelbuild/bazel/blob/c3d43ec123cf9a0d214a950eca3e15637b3fd325/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L181

jvolkman:
> Should this be removed as well? https://github.com/bazelbuild/bazel/blob/c3d43ec123cf9a0d214a950eca3e15637b3fd325/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java#L181

Good catch.

groodt (approved):
LGTM

comius:
cc @rickeylev can you take a look at this?

comius (approved):

PR #15784, opened 2022-07-01
Fix cherrypick from e4bc370b226eb0cc536b55641640266345a214ec

oquenchil:
Cherrypick was missing line where we add the interface library to the output group.


meteorcloudy:
How did this happen, something went wrong in https://github.com/bazelbuild/bazel/pull/15754?

oquenchil:
Yes, the cherrypick was missing a line.

PR #15785, opened 2022-07-01
Replace uses of `cfg = "host"` with `cfg = "exec"`

Yannic:
This increases the compatibility for builds where host and exec platform
are different (e.g., with remote execution).

### DIFF
--- a/site/en/docs/creating-workers.md
+++ b/site/en/docs/creating-workers.md
@@ -166,12 +166,13 @@ level of the build, this might be the attribute definition:
 "worker": attr.label(
     default = Label("//work:worker"),
     executable = True,
-    cfg = "host",
+    cfg = "exec",
 )
 ```
 
-`cfg = "host"` indicates that the worker should be built to run on your host
-platform.
+`cfg = "exec"` indicates that the worker should be built to run on your
+execution platform rather than on the target platform (i.e., the worker is used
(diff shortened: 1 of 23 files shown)


Yannic:
@comius @susinmotion @gregestren PTAL

Yannic:
@bazel-io flag

(not a hard blocker for 5.3, but users can't test whether they are compatible with `--incompatible_disable_starlark_host_transitions` so we should cherry pick to make migration easier)

Yannic:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

meteorcloudy:
@gregestren @comius Can you help review this one?

aiuto:
AFAIK, this is all fine. 
I was concerned about the potential of fallout from //tools/builds_defs:hash changing config, but I don't think it will matter in any substantial way.

katre (approved):
Because [`cfg = "host"` is treated identically to `cfg = "exec"`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkAttrModule.java;drc=f194684586a214e6e2f77390dfe4bdee96ff0d46;l=240), this change is effectively a no-op but makes the correct usage much more clear.

PR #15792, opened 2022-07-02
Add is_root struct field to bazel_module

fmeum:
This allows module extensions to check whether a given module is the
root module, which is necessary to implement the analogues of
--check_direct_dependencies and archive_override/git_override for
dependencies managed by extensions.


fmeum:
@meteorcloudy @Wyverald This seemed straightforward enough for me to go for a PR right away. If you want to discuss the approach more generally, I could create a feature request instead.

meteorcloudy:
@Wyverald I thought we already have a way to tell which one is the root module. IIUC, the modules are listed in BFS order, so the first module must be root module, is that right?

meteorcloudy:
Oh, I see! Then this change looks good to me, but waiting for @Wyverald to confirm!

meteorcloudy:
@bazel-io fork 5.3.0

Issue #13316, opened 2021-04-08
Implement Bzlmod in Bazel


meteorcloudy:
Implement the new dependency management system proposed in https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit?usp=sharing

meteorcloudy:
We'll first build a MVP in the bzlmod branch.

Psoffie009:
> Implement the new dependency management system proposed in https://docs.google.com/document/d/1moQfNcEIttsk6vYanNKIy3ZuK53hQUFq1b1r0rmsYVg/edit?usp=sharing

cpsauer:
@Wyverald, for toolmakers/library maintainers that should be sure to adopt before bzlmod is enabled by default:
Is there a new issue we should follow to stay abreast of when bzlmod will fully launch?

Wyverald:
Bzlmod will "fully" launch with the release of Bazel 6.0, at which point the flag --experimental_enable_bzlmod will be renamed to just --enable_bzlmod.

Bzlmod will be enabled by default in Bazel 7.0 at the earliest. That would mean deprecating WORKSPACE support, which we're still a rather long way from.

Both of these will be featured prominently in the release notes of the major releases, so hopefully it's unlikely for you to miss them :)

cpsauer:
Definitely. Super good to know. Thank you! Great to know the forward roadmap--and thanks for all your great efforts.

PR #15793, opened 2022-07-02
Bzlmod: Better canonical repo names for modules with overrides

fmeum:
(https://github.com/bazelbuild/bazel/issues/13316)

Use "foo.override" instead of just "foo" as the canonical repo name for the module "foo" when it has a non-registry override. Context: https://groups.google.com/a/bazel.build/g/external-deps/c/GcV1cpRd-Ls/m/o446O3vGAAAJ

PiperOrigin-RevId: 423049588


fmeum:
@ckolli5

PR #15795, opened 2022-07-03
Fix behavior of `print()` in module extensions

### LINKED ISSUE #14526: `print` in a module appears to be different from `print` in a workspace rule

`print` appears to behave differently between workspace rules and when used in a module.

When called from a workspace, the output of `print` is prefixed by a `DEBUG` statement, and then the path to the file, whereas this is missing when the same function is called from a module.

5.0.0rc3

The way that `print` works when called from a workspace is nicer: it makes the output easy to identify and it's easy to grep for.

fmeum:
Previously, messages printed with `print()` didn't have the usual
`DEBUG:` prefix.

Fixes #14526

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/SingleExtensionEvalFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/SingleExtensionEvalFunction.java
@@ -165,6 +165,7 @@ public class SingleExtensionEvalFunction implements SkyFunction {
     try (Mutability mu =
         Mutability.create("module extension", usagesValue.getExtensionUniqueName())) {
       StarlarkThread thread = new StarlarkThread(mu, starlarkSemantics);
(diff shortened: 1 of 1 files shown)


fmeum:
@meteorcloudy @Wyverald I noticed that the `bzlmod` parts of the codebase don't have you listed as `CODEOWNERS`, so reviewers don't get assigned automatically. Would it make sense to add you for e.g. `src/main/java/com/google/devtools/build/lib/bazel/bzlmod`, `src/test/java/com/google/devtools/build/lib/bazel/bzlmod` and `src/test/py/bazel/bzlmod`?

meteorcloudy:
> I noticed that the bzlmod parts of the codebase don't have you listed as CODEOWNERS

Will do it, thanks for the notice!

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 5.3.0

Wyverald (approved):

PR #15796, opened 2022-07-03
Fail the build on repository rule errors in a module extension

fmeum:
Fail the build on repository rule errors in a module extension
    
Also emit a Starlark stack trace pointing to the particular call site
of the repository rule.
    
Previously, the root cause error would be printed, but the build would
continue.
    
Fixes https://github.com/bazelbuild/bazel/issues/14526#issuecomment-1149683214

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/BzlmodRepoRuleCreator.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/bzlmod/BzlmodRepoRuleCreator.java
@@ -31,6 +31,8 @@ import com.google.devtools.build.lib.packages.RuleFactory.InvalidRuleException;
 import com.google.devtools.build.lib.vfs.Root;
 import com.google.devtools.build.lib.vfs.RootedPath;
 import java.util.Map;
+import net.starlark.java.eval.EvalException;
+import net.starlark.java.eval.Starlark;
 import net.starlark.java.eval.StarlarkSemantics;
 import net.starlark.java.eval.StarlarkThread.CallStackEntry;
 import net.starlark.java.syntax.Location;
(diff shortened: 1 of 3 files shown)


fmeum:
@Wyverald @meteorcloudy

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 5.3.0

src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -335,6 +335,36 @@ def testCheckDirectDependencies(self):
...
+        'def _module_ext_impl(ctx):',
+        '    repo_rule(name = "foo", invalid_attr = "value")',
+        'module_ext = module_extension(implementation = _module_ext_impl)',
+    ])
+    exit_code, _, stderr = self.RunBazel(
+        ['run', '@foo//...'],
+        allow_failure=True)
+    self.AssertExitCode(exit_code, 48, stderr)
+    self.assertIn(
+        "ERROR: <builtin>: //pkg:@.module_ext.foo: no such attribute 'invalid_attr' in 'repo_rule' rule",
```
Wyverald, line 360: > `//pkg:@.module_ext.foo`

wow, is this what it prints right now? That's unfortunate. Could you add a TODO on me to look into fixing this error message?

src/main/java/com/google/devtools/build/lib/bazel/bzlmod/BzlmodRepoRuleCreator.java
```diff
@@ -79,6 +81,10 @@ public static Rule createRule(
       // This literally cannot happen -- we just created the package!
       throw new IllegalStateException(e);
     }
+    if (rule.containsErrors()) {
+      throw Starlark.errorf(
+          "failed to instantiate '%s' from this module extension", ruleClass.getName());
```
Wyverald, line 86: this is not exactly accurate -- this method can be called when we're creating repos from Bazel modules too. I think just "failed to instantiate repo '%s'" is probably good enough.

Wyverald (approved):

Wyverald (approved):

ckolli5:
Hey @fmeum , we are trying to [cherry-pick](https://github.com/bazelbuild/bazel/pull/15884) these changes to release-5.3.0. But pre-submit checks are failing due to missing dependencies. So, could you please help us in cherry picking these changes to release-5.3.0?

fmeum:
@ckolli5 The patch doesn't apply cleanly. You can close #15802.

PR #15798, opened 2022-07-04
Allow relative paths as arguments to `path.get_child`

fmeum:
This allows for usages such as:

```
some_path.get_child("nested", "directory")

relative_path = "user/provided"
some_path.get_child(relative_path)
```

RELNOTES: The `get_child` method of `path` now accepts an arbitrary
number of relative path strings as positional arguments.


fmeum:
@fweikert Could you take a look?

fmeum:
@fweikert Friendly ping

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkPath.java
```diff
@@ -93,12 +93,12 @@ public StarlarkPath getDirname() {
...
-      doc = "Append the given path to this path and return the resulted path.",
+      doc = "Append the given basename to this path and return the resulting path.",
       parameters = {
-        @Param(name = "child_path", doc = "The path to append to this path."),
+        @Param(name = "child_name", doc = "The basename to append to this path."),
       })
-  public StarlarkPath getChild(String childPath) {
-    return new StarlarkPath(path.getChild(childPath));
+  public StarlarkPath getChild(String childName) {
+    return new StarlarkPath(path.getChild(childName));
```
Wyverald, line 101: I'm thinking we should maybe just call [`path.getRelative`](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/vfs/Path.java;l=121;drc=a380a157f658cc392f66f86b6aba340c461dc408) here -- the limitation to only use a basename feels rather arbitrary.

Wyverald (commented):
I don't know how to put a comment on an empty file, so leaving it here: what's with the empty local_repository.bzl file?

fmeum:
Dropped the empty file, that was just there by accident.

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkPath.java
fmeum: I agree that this behavior would be much more useful. Since `get_child` would currently hard fails on such paths, it wouldn't even be a breaking change. Should I submit this change instead?

Wyverald: Sounds good to me!

fmeum: Having thought about this some more, what do you think of merging this PR and adding a `get_relative` method that accepts any relative path rather than just a basename? That's because I would usually understand the "children" of a directory to be the entries directly contained in that directory, not a nested subdirectory. We could then mark `get_child` as deprecated if needed.

Wyverald: that's also good. I personally don't think it matters much since for anyone expecting "children" to have the narrower meaning, `get_child` with the broader impl would still work. But I won't complain either way :)

fmeum: I found an API design I'm happy for `get_child`. Please take a look.

Wyverald (approved):
nice!

PR #15799, opened 2022-07-04
Add critical path time to Build Event Protocol

crydell-ericsson:
We want to be able to look at information about the critical path time in consumers of the Build Event Protocol, and have thereby added a field for it in the TimingMetrics. This field is populated by the MetricsCollector, which subscribes to information about the critical path that is emitted by the BuildSummaryStatsModule. 

While the MetricsCollector would originally emit its metrics information upon the BuildPrecompleteEvent, it now waits until both BuildPrecompleteEvent and CriticalPathEvent have been emitted. The new CriticalPathEvent is emitted right after the critical path has been aggregated. The behavior of MetricsCollector remains the same as before upon receiving a NoBuildRequestFinishedEvent.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
+++ b/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
@@ -946,6 +946,8 @@ message BuildMetrics {
     int64 wall_time_in_ms = 2;
     // The elapsed wall time in milliseconds during the analysis phase.
     int64 analysis_phase_time_in_ms = 3;
+    // The critical path time in milliseconds during this build.
+    int64 critical_path_time_in_ms = 4;
   }
   TimingMetrics timing_metrics = 5;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
```diff
@@ -946,6 +946,8 @@ message BuildMetrics {
     int64 wall_time_in_ms = 2;
     // The elapsed wall time in milliseconds during the analysis phase.
     int64 analysis_phase_time_in_ms = 3;
+    // The critical path time in milliseconds during this build.
+    int64 critical_path_time_in_ms = 4;
```
Yannic, line 950: Please use `google.protobuf.Duration` instead.

Yannic:
cc @lfpino @saraadams @willengflow

src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto
crydell-ericsson: Done

meisterT:
We are working internally on open sourcing more of our metrics which would likely give more insight into the critical path, so I propose to close this PR for now.

meisterT:
We have to make some internal changes before we actually work on open sourcing. I assume end of this quarter.

I recommend using the data from the JSON trace profile in the meantime.

crydell-ericsson:
We have some ongoing work in our internal fork of Bazel that also concerns metrics, specifically adding information about the number of local cache hits both overall and per mnemonic to the BEP. Is something similar also covered by the upcoming work on your end? Would there be any point in submitting a pull request for it if this one ends up closed?

meisterT:
Obsolete after https://github.com/bazelbuild/bazel/commit/41b72aa247a89d395e4a764ee47477757ad35b17

PR #15800, opened 2022-07-04
Extend "Deploying Rules" to cover Bazel modules

fmeum:
Adds conventions for the placement of module extensions and test modules
in rulesets deployed as Bazel modules.

### DIFF
--- a/site/en/rules/deploying.md
+++ b/site/en/rules/deploying.md
@@ -44,6 +44,7 @@ For example, when writing new rules for the (make-believe)
   LICENSE
   README
   WORKSPACE
+  MODULE.bazel *
   mockascript/
     constraints/
       BUILD
@@ -52,7 +53,13 @@ For example, when writing new rules for the (make-believe)
       runfiles.mocs
     BUILD
     defs.bzl
+    extensions.bzl *
+    repositories.bzl
   tests/
+    bcr/ *
+      WORKSPACE
+      MODULE.bazel
+      BUILD
     BUILD
     some_test.sh
     another_test.py
@@ -62,6 +69,9 @@ For example, when writing new rules for the (make-believe)
     lib.mocs
     test.mocs
 ```
+
+Files and directories marked with `*` are used by Bazel's new and currently
+experimental external dependency management system [*Bzlmod*](/docs/bzlmod).
 
 ### WORKSPACE
 
@@ -80,6 +90,18 @@ In the following sections, assume the repository belongs to the
 ```
 workspace(name = "rules_mockascript")
 ```
+
+### MODULE.bazel *
+
+Refer to the [Bzlmod user guide](/docs/bzlmod#module-bazel) for detailed
+information on how rules can specify their dependencies using the `MODULE.bazel`
+file.
+
+Since the repository name under which a Bazel module is added can be freely
+chosen by the user via the `repo_name` attribute of
+[`bazel_dep`](/rules/lib/globals#bazel_dep), it is no longer necessary to follow
+the `<org>_rules_<lang>` naming scheme for rules outside the
+[bazelbuild](https://github.com/bazelbuild) organization.
 
 ### README
 
@@ -158,6 +180,12 @@ in the form of a library target located at `//<LANG>/runfiles` (an abbreviation
 of `//<LANG>/runfiles:runfiles`). User targets that need to access their data
 dependencies will typically add this target to their `deps` attribute.
 
+### Module extensions *
+
+If you provide [module extensions](/docs/bzlmod#module-extensions) to allow
+users of your rules to declare domain-specific external dependencies, these
+extensions should all be exported from a a file named `<LANG>/extensions.bzl`.
+
 ### Repository rules
 
 #### Dependencies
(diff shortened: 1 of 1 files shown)


site/en/rules/deploying.md
```diff
@@ -81,6 +91,18 @@ In the following sections, assume the repository belongs to the
 workspace(name = "rules_mockascript")
 ```
 
+### MODULE.bazel *
+
+Refer to the [Bzlmod user guide](/docs/bzlmod#module-bazel) for detailed
+information on how rules can specify their dependencies using the `MODULE.bazel`
+file.
+
+Since the repository name under which a Bazel module is added can be freely
```
Yannic, line 100: I'm not so sure about that. What if there are multiple `rules_<?>`? I think the recommendation should be to use `<org>_rules_<lang>` also for `repo_name`.

We should probably also mention in this guide how to get the rules into BCR (i.e., linking to another doc)

site/en/rules/deploying.md
```diff
@@ -158,6 +180,12 @@ in the form of a library target located at `//<LANG>/runfiles` (an abbreviation
 of `//<LANG>/runfiles:runfiles`). User targets that need to access their data
 dependencies will typically add this target to their `deps` attribute.
 
+### Module extensions *
+
+If you provide [module extensions](/docs/bzlmod#module-extensions) to allow
+users of your rules to declare domain-specific external dependencies, these
+extensions should all be exported from a a file named `<LANG>/extensions.bzl`.
```
Yannic, line 187: `extensions.bzl` seems quite generic. Maybe `<lang>/bzlmod.bzl` instead?

site/en/rules/deploying.md
```diff
@@ -230,6 +258,32 @@ can either be in the standard location for the language the rules are for or a
 It is useful to users to have an `examples/` directory that shows users a couple
 of basic ways that the rules can be used.
 
+### Test module for the Bazel Central Registry *
+
+All rules available as a [Bazel module](https://bazel.build/docs/bzlmod#modules)
+from the [Bazel Central Registry (BCR)](https://github.com/bazelbuild/bazel-central-registry)
+should provide a Bazel test module under `tests/bcr/` that references the main
+module and exercises some basic functionality of the rules. This test should be
```
Yannic, line 266:
```suggestion
module and exercises some basic functionality of the rules. This test must be
```

site/en/rules/deploying.md
```diff
@@ -230,6 +258,32 @@ can either be in the standard location for the language the rules are for or a
 It is useful to users to have an `examples/` directory that shows users a couple
 of basic ways that the rules can be used.
 
+### Test module for the Bazel Central Registry *
+
+All rules available as a [Bazel module](https://bazel.build/docs/bzlmod#modules)
+from the [Bazel Central Registry (BCR)](https://github.com/bazelbuild/bazel-central-registry)
+should provide a Bazel test module under `tests/bcr/` that references the main
```
Yannic, line 265:
```suggestion
must provide a Bazel test module under `tests/bcr/` that references the main
```

I think BCR should enforce this (because how else would they verify that a module is "good"?)

@Wyverald @meteorcloudy WDYT?

site/en/rules/deploying.md
fmeum: Since `repo_name` is chosen by end users, wouldn't a recommendation for it be a bit out of place in this doc? 

There is definitely some tension between "slick module names" and "prevent namegrabbing" that is probably best for the SIG to discuss and resolve. Maybe we can learn from other registries (NPM, cargo) here?

site/en/rules/deploying.md
fmeum: I actually think that generic is what we should go for here: `bzlmod` is just the codename for a currently experimental feature that is planned to become the only way for Bazel projects to declare their dependencies, after `WORKSPACE` has been deprecated and removed. By that time, `<LANG>/bzlmod.bzl` would look pretty weird and `<LANG>/extensions.bzl` would seem natural as everybody will be accustomed to module extensions. I agree that `bzlmod.bzl` would probably be more self-explanatory during this transition period, but IMO we should optimize for the long term.

site/en/rules/deploying.md
Yannic: Sure, for non-ruleset modules, the story might be different. I fully agree on that. But this guide is about rules so I think enforcing is preferable.

I don't feel too strongly about making this a `must`.

I'm wondering if it makes sense to use the existing `//examples` as test module?

Yannic:
I added this to the agenda for the next SIG meeting

alexeagle:
@fmeum would you like to be in the CODEOWNERS as a future reviewer of this file? Maybe someone from the rules authors SIG instead/in addition?

fmeum:
@alexeagle You can add me, but the SIG should definitely be represented there as well (or even instead if there are sufficiently many volunteers).

site/en/rules/deploying.md
Wyverald: I don't think BCR should really enforce having a test module, though it's definitely recommended. So `should` is the right level of nudging, I think.

meteorcloudy:
Do we still need this PR?

fmeum:
I think it's obsolete now with the docs we have in the BCR.

PR #15805, opened 2022-07-04
Add factory for creating paths relative to well-known roots

Yannic:
This change adds a factory for creating `PathFragments` relative to
pre-defined (named) roots (e.g., relative to `%workspace%`).

The syntax is choosen to match existing ad-hoc solutions for `%workspace%`,
or `%builtins%` in other places (so that we can ideally migrate them in
a follow-up).

We'll use this for parsing paths from the command-line (e.g.,
`--credential_helper=%workspace%/foo`).

Progress on https://github.com/bazelbuild/proposals/blob/main/designs/2022-06-07-bazel-credential-helpers.md


Yannic:
@tjgq PTAL

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Factory for creating {@link PathFragment}s from command-line options.
+ *
+ * <p>The difference between this and using {@link PathFragment#create(String)} directly is that
+ * this factory replaces values starting with {@code %<name>%} with the corresponding (named) roots
+ * (e.g., {@code %workspace%/foo} becomes {@code </path/to/workspace>/foo}).
+ */
+public final class CommandLinePathFactory {
+  private static final Pattern REPLACEMENT_PATTERN = Pattern.compile("^(%([a-z_]+)%/)?([^%].*)$");
```
tjgq, line 20: Make the last `*` a `+` so that an empty path is rejected.

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
```diff
@@ -0,0 +1,46 @@
...
+    Preconditions.checkArgument(matcher.matches());
+
+    String rootName = matcher.group(2);
+    String path = matcher.group(3);
+    if (!Strings.isNullOrEmpty(rootName)) {
+      PathFragment root = roots.get(rootName);
+      if (root == null) {
+        throw new IllegalArgumentException(String.format(Locale.US, "Unknown root %s", rootName));
+      }
+      return root.getRelative(path);
```
tjgq, line 42: Out of curiosity, what's the result of

```
new CommandLinePathFactory(ImmutableMap.of("workspace", "/path/to/workspace")).create("%workspace%/../bar");
```

Should we make it a requirement that the path must resolve to a descendant of the root (or possible even disallow `.` and `..` segments entirely)?

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+        .isEqualTo(PathFragment.create("/path/to/workspace/foo/bar"));
+
+    assertThat(factory.create("%output_base%/foo"))
+        .isEqualTo(PathFragment.create("/path/to/output/base/foo"));
+    assertThat(factory.create("%output_base%/foo/bar"))
+        .isEqualTo(PathFragment.create("/path/to/output/base/foo/bar"));
+  }
+
+  @Test
+  public void UnknownRoot() {
```
tjgq, line 51: Method name should start with lowercase.

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @Test
+  public void createFromAbsolutePath() {
+    CommandLinePathFactory factory = new CommandLinePathFactory(ImmutableMap.of());
+
+    assertThat(factory.create("/absolute/path/1"))
+        .isEqualTo(PathFragment.create("/absolute/path/1"));
+    assertThat(factory.create("/absolute/path/2"))
+        .isEqualTo(PathFragment.create("/absolute/path/2"));
+  }
+
```
tjgq, line 24: Add a test to verify that the empty path is rejected.

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.devtools.build.lib.vfs.PathFragment;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link CommandLinePathFactory}. */
+@RunWith(JUnit4.class)
+public class CommandLinePathFactoryTest {
```
tjgq, line 14: Is this class also going to be used in the PATH lookup case? (i.e., `--credential_helper=somehelper`)? If so, we should have tests here for paths that don't start with a slash, or without slashes at all.

tjgq (changes requested):

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
tjgq: Sorry, let me clarify:

- If we're going to use this class for PATH lookup (e.g. `--credential_helper=foo`), we should have a test for them.
- Since we agreed in the proposal to disallow non-absolute paths *with* slashes (e.g. `--credential_helper=foo/bar`) unless they're relative to a `%root%`, we should reject them (and have a test to verify that we do).

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
Yannic: `*` is actually correct. We reject the empty path because of the required `[^%]` before. Changing it to `+` would wrongly reject single-char paths (which are fine)

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
Yannic: Changed it to reject upwards references. `PathFragment.create()` will take care of normalizing it.

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
Yannic: of course...

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
Yannic: done

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
Yannic: Added `PATH` lookup (I was originally planning to do it somewhere else, but we may as well do it in this factory)

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
tjgq: Oops, you're right :)

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
tjgq: Actually, I'm not so sure PATH lookup belongs here. (What I had in mind was to simply store the non-absolute path, and resolve it later.)

I expect this code will run once at options parsing time, which means the result will be cached during the lifetime of the Bazel server process. Hence, if you add or remove a binary from one of the PATH directories that could affect the resolution of a particular helper, it won't take effect until you restart the server, which might be surprising to users.

WDYT? Should we defer PATH resolution until the helper is invoked?

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
```diff
@@ -0,0 +1,97 @@
...
+  public Path create(Map<String, String> env, String value) throws IOException {
+    Preconditions.checkNotNull(env);
+    Preconditions.checkNotNull(value);
+
+    Matcher matcher = REPLACEMENT_PATTERN.matcher(value);
+    Preconditions.checkArgument(matcher.matches());
+
+    String rootName = matcher.group(2);
+    PathFragment path = PathFragment.create(matcher.group(3));
+    if (path.containsUplevelReferences()) {
```
tjgq, line 46: Should we just use `PathFragment.isNormalized` here? I can't imagine a good reason to include a `.` component either.

tjgq (changes requested):

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
tjgq: Actually, I think what I said above only applies to startup options. If we can make `--credential_helper` a non-startup option, then I think it's fine to do the resolution here.

src/main/java/com/google/devtools/build/lib/runtime/CommandLinePathFactory.java
Yannic: I don't think we should. IIUC, `PathFragment.isNormalized` would mean that you need to pass `/` instead of `\` on Windows, which feels like an unnecessary restriction.

tjgq: Ack, I thought it only checked for `.` and `..`.

tjgq (approved):

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
Yannic: Ack. I think `query` actually implies `build` for historical reasons.

Yannic:
@bazel-io flag

ckolli5:
Hey @Yannic, we are trying to [cherry-pick](https://github.com/bazelbuild/bazel/pull/15885) these changes to release-5.3.0. But pre-submit checks are failing due to missing dependencies. So, could you please help us in cherry picking these changes to release-5.3.0?

PR #11385, opened 2020-05-13
Implement namespaced R classes

jongerrish:
Enable rudimentary R class namespacing where each library only contains
references to the resources it declares instead of declarations plus all
transitive dependency references.

This basically implements the `android.namespacedRClass=true` Gradle feature in Bazel.

For our build this saves a total of 215MB outputs + 145 Actions for a clean build in addition to previous open PRs.

No resource merging for android_library targets
Smaller R.{class,txt}
Less desugaring + dexing

PR #15806, opened 2022-07-04
Allow disabling android resource merging via `android_non_transitive_r_class` configuration option

arunkumar9t2:
With this change each `android_library` target will generate `R` class with entries only from current local target instead of including entries from transitive closure of dependencies usually created via a separate merging action.

`R` entries from libraries should be referenced by their fully qualified name otherwise build would fail.

Reference: https://developer.android.com/studio/build/optimize-your-build#use-non-transitive-r-classes

In our large android app, we save several thousand actions and reduce cache output by 1.1 GB. 

Adapted from https://github.com/bazelbuild/bazel/pull/11385 and renamed flag to be in line with Android Gradle plugin flag.

Fixes https://github.com/bazelbuild/rules_android/issues/10


ted-xie:
@arunkumar9t2 Is this still an issue in the Starlark Android rules? If so, shall we migrate this to rules_android?

meisterT:
Let's close this and re-open in the Android rules repo if it still matters.

PR #15366, opened 2022-04-28
WIP: Create tree artifact outputs

fmeum:
Fixes #6393

PR #15818, opened 2022-07-06
Create output directories for remote execution

### LINKED ISSUE #6393: TreeArtifacts output directories aren't created with remote execution enabled

fmeum:
By explicitly declaring output directories as inputs to a remote action,
this commit ensures that these directories are created by the remote
executor prior to the execution of the action. This brings the behavior
of remote execution regarding tree artifacts in line with that of local
or sandboxed execution.

Getting the tests to pass requires modifying a check in Bazel's own
remote worker implementation: Previously, the worker explicitly verified
that output directories don't exist after the inputs have been staged.
This behavior is not backed by the spec and has thus been modified: Now,
it is only checked that the output directories either don't exist or are
directories.

Fixes #6393

Closes #15366.

PiperOrigin-RevId: 447451303

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
@@ -139,6 +139,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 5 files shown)


fmeum:
@ckolli5 Not sure why the cherry pick failed last time, it now applied cleanly (up to a `BUILD` file conflict).

PR #15823, opened 2022-07-06
RemoteExecutionService: fix outputs not being uploaded

### LINKED ISSUE #15682: Remote-cache not work in 5.2.0
### Description of the bug:

When upgrade to the latest version of bazel. Bazel remote seems not to work. Bazel can only put but can not get. When use bazel 5.1.1 bazel remote works again. More strange, bazel 5.2.0 can use the cache of 5.1.1. It seems that the artifact put by bazel 5.2 has problem.

Use bazel 5.2.0 and use `https://github.com/buchgr/bazel-remote` as remote cache like `bazel build --remote_cache=http://localhost:9090`

Ubuntu 20.04

release 5.2.0

_No response_

_No response_

_No response_

_No response_

sluongng:
In 4d900ceea12919ad62012830a95e51f9ec1a48bb we introduced validation in
DiskAndRemoteCacheClient.uploadActionResult() where context's step must
be UPLOAD_OUTPUTS to trigger the upload.

However, this value was never set in RemoteExecutionService before hand
thus led to outputs not being uploaded and cause remote cache misses.

Fix #15682

Thanks @adam-singer for doing the investigation 🙏

### DIFF
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -1583,6 +1583,33 @@ EOF
   rm -rf $cache
 }
 
+function test_combined_cache_upload() {
+  mkdir -p a
+  echo 'bar' > a/bar
+  cat > a/BUILD <<'EOF'
+genrule(
+  name = "foo",
+  srcs = [":bar"],
+  outs = ["foo.txt"],
+  cmd = """
+    echo $(location :bar) > "$@"
+  """,
+)
+EOF
+
+  CACHEDIR=$(mktemp -d)
+
+  bazel build \
+    --disk_cache=$CACHEDIR \
+    --remote_cache=grpc://localhost:${worker_port} \
+    //a:foo >& $TEST_log || "Failed to build //a:foo"
+
+  remote_ac_files="$(count_remote_ac_files)"
+  [[ "$remote_ac_files" == 1 ]] || fail "Expected 1 remote action cache entries, not $remote_ac_files"
+  remote_cas_files="$(count_remote_cas_files)"
+  [[ "$remote_cas_files" == 3 ]] || fail "Expected 3 remote cas entries, not $remote_cas_files"
+}
+
 function test_combined_cache_with_no_remote_cache_tag_remote_cache() {
   # Test that actions with no-remote-cache tag can hit disk cache of a combined cache but
   # remote cache is disabled.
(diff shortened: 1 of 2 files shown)


coeuvre:
Good catch! Can you please also add an integration test for this case so we won't break combined cache again in the future?

fmeum:
@bazel-io flag

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -1056,6 +1056,33 @@ EOF
...
+
+  CACHEDIR=$(mktemp -d)
+
+  bazel build \
+    --disk_cache=$CACHEDIR \
+    --remote_cache=grpc://localhost:${worker_port} \
+    //a:foo >& $TEST_log || "Failed to build //a:foo"
+
+  remote_ac_files="$(count_remote_ac_files)"
+  [[ "$remote_ac_files" == 1 ]] || fail "Expected 1 remote action cache entries, not $remote_ac_files"
```
sluongng, line 1081: Before this PR, this would be 0 indicating cache misses.

Worth to note that this behavior could only be observed when both `--disk_cache` AND `--remote_cache` are both enabled, thus triggering `DiskAndRemoteCacheClient.java` logic.

Personally I would consider enabling both `--disk_cache` and `--remote_cache` somewhat of a miss-use since enabling `--remote_cache` alone would keep a copy of cas and ac on local disk within execroot 🤔. But perhaps we can discuss/document the tradeoffs between these 2 in a separate change

sluongng:
```
bazel test src/test/shell/bazel/remote:remote_execution_test --test_arg=test_remote_cache_upload
```

Handy command for those who are not familiar with Bazel code base's shell test setup to try out the test and see the result with/without the one-liner fix.

sluongng:
CI passed

@coeuvre PTAL

src/test/shell/bazel/remote/remote_execution_test.sh
```diff
@@ -1056,6 +1056,33 @@ EOF
   expect_not_log "remote cache hit"
 }
 
+function test_remote_cache_upload() {
```
coeuvre, line 1059: Please rename this to `test_combined_cache_upload` for consistency.

coeuvre (approved):
Thanks! I will import it after the comment is addressed.

src/test/shell/bazel/remote/remote_execution_test.sh
sluongng: Done.

I also moved the test a bit down so that it is close to other `test_combined_cache_*` tests.

ckolli5:
@bazel-io fork 5.3.0

PR #15825, opened 2022-07-06
Add --collect_coverage_wrapper CLI flag

futt-bucker:
Bazel has --coverage_support flag which allows to add custom label(s) as
dependencies of `bazel coverage` command. This PR also adds
--collect_coverage_wrapper CLI flag, which allows users to override
"collect_coverage.sh" wrapper script.

One use-case for this is to let windows users override .sh script (which
cannot be interpreted without bash) with their custom tools which setup
coverage environment.


futt-bucker:
Bump

c-mita:
I understand the desire, but I'm uncomfortable continuing further in this direction (adding more flags to allow overriding parts of coverage collection).

I really think coverage collection would benefit by taking a step back and thinking about how such a system should work what API should be give to rules for them to be able to implement it; ideally working across languages (without things like `py_test` needing to know how C++ coverage works).

Adding these flags increases external dependence on what should be internal parts of the system and I really don't want to encourage it at this time.

oquenchil:
@c-mita should this PR be closed?

c-mita:
I'm going to close this PR because I really think the coverage API needs some cleaning up and I don't want to increase its current surface area anymore at this time; if there is a compelling reason to accept this then I can re-open this (I don't think being a "step towards" Windows support is sufficient right now).

PR #15833, opened 2022-07-07
Add --experimental_execution_log_explain_file

rsalvador:
This option is similar to --explain but uses the information in the execution log to show which inputs and outputs changed. It is implemented by keeping a history of the previous execution logs spawns keyed by the target + action progress message and comparing the digests of the new inputs/outputs to the old ones.

Sample explain file generated:
Screen Shot 2022-07-07 at 5 10 38 PM

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/bazel/BUILD
@@ -110,7 +110,7 @@ java_library(
     ],
     deps = [
         "//src/main/java/com/google/devtools/build/lib:runtime",
-        "//src/main/java/com/google/devtools/build/lib/bazel/execlog:stable_sort",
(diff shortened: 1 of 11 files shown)


rsalvador:
> I wonder whether this should be part of Bazel or better a separate tool that takes two execlogs and spits out this information.

I had it implemented as a separate tool for me to use it, but it is cumbersome and/or requires wrappers.
Didn't see a way to make it available to my coworkers in a way that was easy so they could use it regularly.

sluongng:
Seems to be very close with https://github.com/bazelbuild/bazel/tree/master/src/tools/execlog

meisterT:
I am hesitant to merge this since it will increase Bazel's maintenance surface and we don't have the bandwidth to support that right now. I suggest indeed building an external tool that uses the execlog instead. If there are fields missing, please let me know.

Issue #15835, opened 2022-07-07
bazel coverage fails on windows because test_wrapper cannot find "collect_coverage.sh"


futt-bucker:
### Description of the bug:

`bazel coverage //cpp/test:mytest`

```
ERROR(tools/test/windows/tw.cc:1294) ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\users\user\_bazel_user\jh7tqry3\execroot\project\bazel-out\x64_windows-fastbuild\bin\cpp\test\mytest.exe.runfiles\project\external\bazel_tools\tools\test\collect_coverage.sh"  cpp/test/mytest.exe
```

My .bazelrc **does not** contain 
```
startup --windows_enable_symlinks
build --enable_runfiles
```

Note how test_wrapper resolved path to `collect_coverage.sh` as to a data file under RUNFILES_DIR. The bug is that on Windows this file does not exist, because runfiles are not enabled.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. have bazel project with `cc_test`
2. run `bazel coverage :test` on windows (does not matter if you use MSVC or MSYS compiler).
3. see the error above.

### Which operating system are you running Bazel on?

Windows 10

### What is the output of `bazel info release`?

development version

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

`bazel build //src:bazel-dev.exe`

Then used `bazel-bin/src/bazel-dev.exe`.

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
https://github.com/futt-bucker/bazel.git
300111212f984587415ad8a1f00bf9d60b31e430
89301f45a7d2bcc3b82b918f96b9ff0037599e4c
```

### Have you found anything relevant by searching the web?

No

fmeum:
The logic in https://github.com/bazelbuild/bazel/blob/d35f923b098e4dc9c90b1ab66b413c216bdee638/tools/test/windows/tw.cc#L438
seems to be causing the chdir to `RUNFILES_DIR`. But it's intended not to run with coverage, see
https://github.com/bazelbuild/bazel/blob/d35f923b098e4dc9c90b1ab66b413c216bdee638/tools/test/windows/tw.cc#L468
I don't have access to a Windows machine, but have you been able to confirm that this check doesn't work as intended?

futt-bucker:
@fmeum I found that this func does not work properly for coverage:
[...]

PR #15836, opened 2022-07-07
If test bin is not found, try to use @bazel_tools

futt-bucker:
Fixes https://github.com/bazelbuild/bazel/issues/15835

In short: on windows if you run `bazel coverage :mytest`, bazel fails because test_wrapper (tw.exe) cannot resolve path to `collect_coverage.sh`. It expects this script to be under `runfiles` directory. But runfiles on Windows are not enabled by default, thus this script will not be where it is expected to be.

We solve the problem by checking - if a test_path does not exist, then we use `$(cwd)/$(test_path)` - which will point to @bazel_tools under execroot/.../external/bazel_tools/...

Before this PR: `bazel coverage ...` fails because `collect_coverage.sh` was not found under runfiles dir.
After this PR: `bazel coverage ...` fails because `collect_coverage.sh` is not Win32 application (it's a bash script). So the script is found, but cannot be interpreted (will be solved in next PRs)

### DIFF
--- a/tools/test/windows/tw.cc
+++ b/tools/test/windows/tw.cc
@@ -1130,6 +1130,7 @@ inline void ComputeRunfilePath(const std::wstring& test_workspace,
 
 bool FindTestBinary(const Path& argv0, const Path& cwd, std::wstring test_path,
                     const Path& abs_test_srcdir, Path* result) {
+  std::wstring test_path_copy = test_path;
   if (!blaze_util::IsAbsolute(test_path)) {
     std::string argv0_acp;
     if (!WcsToAcp(argv0.Get(), &argv0_acp)) {
@@ -1195,6 +1196,32 @@ bool FindTestBinary(const Path& argv0, const Path& cwd, std::wstring test_path,
   }
 
   (void)result->Absolutize(cwd);
+
+  if (!IsReadableFile(*result)) {
+    // Entire routine was no able to resolve into an existing test path.
+    // Try to concat $(cwd)/$(test_path). There's high chance the file
+    // may be there.
+    // Example (bazel coverage ...) produces this:
+    //   cwd=C:\users\user\_bazel_user\jh7tqry3\execroot\projectname
+    //   test_path=external\bazel_tools\tools\test\collect_coverage.sh
+    //
+    // test_path is changed, so use test_path_copy (original value)
+
+    if (!result->Set(test_path_copy)) {
+      LogErrorWithArg2(__LINE__, "Could not join paths", cwd.Get(),
+                       test_path_copy);
+      return false;
+    }
+
+    (void)result->Absolutize(cwd);
+
+    if (!IsReadableFile(*result)) {
+      // the file still does not exist
+      LogErrorWithArg(__LINE__, "Failed to find test path", result->Get());
+      return false;
+    }
+  }
+
   return true;
 }
 


futt-bucker:
Bump

futt-bucker:
@lberki Can you take a look?

Warchant:
cc @c-mita

futt-bucker:
@c-mita Please also look.

meteorcloudy:
> We solve the problem by checking - if a test_path does not exist, then we use $(cwd)/$(test_path) - which will point to @bazel_tools under execroot/.../external/bazel_tools/...

This sounds very brittle, can you instead use the runfiles library to look up the script, just check all the `Rlocation` usages in the code.

Warchant:
> > We solve the problem by checking - if a test_path does not exist, then we use (cwd)/(test_path) - which will point to @bazel_tools under execroot/.../external/bazel_tools/...
> 
> This sounds very brittle, can you instead use the runfiles library to look up the script, just check all the `Rlocation` usages in the code.

That's the point - Rlocation returns a path which does not exist. So this additional check ensures path to test script is correct.

fmeum:
If the runfiles library doesn't return the right path, then something isn't right at a higher level and we should fix that. What is the path that `Rlocation` returns and what would the correct path be?

Warchant:
> If the runfiles library doesn't return the right path, then something isn't right at a higher level and we should fix that. What is the path that `Rlocation` returns and what would the correct path be?

If you run `bazel coverage ... -s` on a bazel project with gtest on Windows, all tests fail with this error:

```
ERROR(tools/test/windows/tw.cc:1294) ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\ba\execroot\bazel_playground\bazel-out\x64_windows-fastbuild\bin\cpp\test\gmock_example.exe.runfiles\bazel_playground\external\bazel_tools\tools\test\collect_coverage.sh"  cpp/test/gmock_example.exe): The system cannot find the file specified.
```

My .bazelrc:
```
startup --output_user_root=C:/b
startup --output_base=C:/ba
```

Pay attention that path to `collect_coverage.sh` is under runfiles (`gmock_example.exe.runfiles`) directory. 
By default, [**runfiles are disabled on windows**](https://bazel.build/configure/windows#symlink), and runfiles dir is always empty (it will only contain MANIFEST):
```powershell
$ tree /F C:\ba\execroot\bazel_playground\bazel-out\x64_windows-fastbuild\bin\cpp\test\gmock_example.exe.runfiles
Folder PATH listing for volume Acer
Volume serial number is C432-FCCE
C:\BA\EXECROOT\BAZEL_PLAYGROUND\BAZEL-OUT\X64_WINDOWS-FASTBUILD\BIN\CPP\TEST\GMOCK_EXAMPLE.EXE.RUNFILES
│   MANIFEST
│   
└───bazel_playground
```

`bazel_playground` - is an empty folder. (it's a name of my project).


----

If I add these 2 to my .bazelrc:
```
startup --windows_enable_symlinks
build --enable_runfiles
```

Then `bazel_playground` is not empty dir, it has this structure:

```powershell
$ tree /F C:\ba\execroot\bazel_playground\bazel-out\x64_windows-fastbuild\bin\cpp\test\test_for_cpp_file.exe.runfiles
Folder PATH listing for volume Acer
Volume serial number is C432-FCCE
C:\BA\EXECROOT\BAZEL_PLAYGROUND\BAZEL-OUT\X64_WINDOWS-FASTBUILD\BIN\CPP\TEST\TEST_FOR_CPP_FILE.EXE.RUNFILES   
│   MANIFEST
│
└───bazel_playground
    └───cpp
        └───test
                test_for_cpp_file.exe
```

But `collect_coverage.sh` is still not there. 

If you drill down to the code, the issue is that Rlocation returns a path that is missing (path to a file `collect_coverage.sh`).

fmeum:
`Rlocation` [falls back to using the runfiles directory](https://github.com/bazelbuild/bazel/blob/658ba15d9f37969edfaae507d267ee3aaba8b44e/tools/cpp/runfiles/runfiles_src.cc#L233) if it can't find the file in the runfiles manifest. 

The paths seem correct, the problem is that `collect_coverage.sh` isn't added to the runfiles of the test action. Finding it under the exec root wouldn't be hermetic anymore: Bazel doesn't track the dependency on the script, so `bazel coverage` wouldn't know to recollect coverage if the script changes.

I think that this could be fixed by making changes to two places:
1. https://github.com/bazelbuild/bazel/blob/658ba15d9f37969edfaae507d267ee3aaba8b44e/src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java#L219 should emit a runfiles and not an exec path, similar to https://github.com/bazelbuild/bazel/blob/658ba15d9f37969edfaae507d267ee3aaba8b44e/src/main/java/com/google/devtools/build/lib/analysis/test/TestStrategy.java#L230. That may require changes to both the Unix and the Windows testrunner though.
2. The script has be added to the runfiles of https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/analysis/test/TestRunnerAction.java. I am not sure what the best (read: doesn't cause memory regressions) way to do this is.

But this is just a guess based on a quick glance at the code.

comius:
Closing, because there was no agreement Amon reviewers and because it’s now pretty stale.

PR #15863, opened 2022-07-12
Expand and clarify XML_OUTPUT_FILE description in test encyclopedia

eriwen:
This change adds a couple sentences to documentation based on my reading of https://github.com/bazelbuild/bazel/blob/master/tools/test/test-setup.sh#L178-L215

### Motivation
To clarify the usage and default behavior of Bazel regarding test result XML output files. 

This doc update might have saved me a few hours of pain — I'm hoping to save the next developer this time.

### DIFF
--- a/site/en/reference/test-encyclopedia.md
+++ b/site/en/reference/test-encyclopedia.md
@@ -447,9 +447,12 @@ The initial environment block shall be composed as follows:
   </tr>
   <tr>
     <td><code>XML_OUTPUT_FILE</code></td>
-    <td>Location of the test result XML output file. The XML schema is based on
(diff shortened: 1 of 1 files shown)


eriwen:
Looks like the failing builds may be flaky and not caused by this documentation change (https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/11600#0181f40d-19f9-433f-8612-a899342e9bc0) — how should we handle this? 

Is there something else you're waiting on me to do here?

fmeum:
@eriwen You can rerun CI checks by amending without changes and then force pushing.

sgowroji (approved):
Thank you!

Issue #14224, opened 2021-11-04
ctx.actions.symlink doesn't make relative symlink


pauldraper:
### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

**WORKSAPCE.bazel**

```py
```

**BUILD.bazel**

```py
load(":rules.bzl", "example")

example(
  name = "example",
)
```

**rules.bzl**

```py
def _example_impl(ctx):
    example1 = ctx.actions.declare_symlink("example1")
    ctx.actions.symlink(output = example1, target_path = "../..")

    example2 = ctx.actions.declare_symlink("example2")
    ctx.actions.run(executable = "ln", arguments = ["-s", "../..", example2.path], outputs = [example2])

    return DefaultInfo(
        files = depset([example1, example2])
    )

example = rule(
  implementation = _example_impl
)
```

```
$ bazel build --experimental_allow_unresolved_symlinks :example
$ ls -l bazel-bin/
lrwxrwxrwx 1 paul paul 68 Nov  4 01:06 example1 -> /home/paul/.cache/bazel/_bazel_paul/fc54cd89f1e9b3d4a2b86f2842cad4ca
lrwxrwxrwx 1 paul paul  5 Nov  4 01:06 example2 -> ../..
```

I expected both symlinks to be ''../.." but only the latter (using executable ln) is.

### What operating system are you running Bazel on?

Ubuntu 18.04

### What's the output of `bazel info release`?

> Replace this line with your answer.

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

4.2.1

philsc:
I ran into this as well. The issue is that, despite the documentation saying otherwise, the first thing [`SymlinkAction`](https://github.com/bazelbuild/bazel/blob/a6d2c623f5ba20910db6b3842fd980f754a1eefa/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkAction.java#L137-L142) does is to turn the unresolved path into a [`PathFragment`](https://github.com/bazelbuild/bazel/blob/a6d2c623f5ba20910db6b3842fd980f754a1eefa/src/main/java/com/google/devtools/build/lib/vfs/PathFragment.java#L32-L41`]. The first thing a `PathFragment` does is to normalize the path.

I don't know what the right answer to fix this is, but I think that's the reason for the issue.

pauldraper:
[...]

PR #15866, opened 2022-07-13
[WIP] Unresolved symlinks

fmeum:


aiuto:
Is there a bug and design behind this?
@alexjski : you might be interested in this.

fmeum:
This PR combines the changes of three other PRs (https://github.com/bazelbuild/bazel/pull/15781, https://github.com/bazelbuild/bazel/pull/15773, https://github.com/bazelbuild/bazel/pull/15700) that are currently awaiting review. Their ultimate goal is to stabilize `--experimental_allow_unresolved_symlinks` (https://github.com/bazelbuild/bazel/issues/10298).

src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java
```diff
@@ -1537,8 +1538,15 @@ private boolean checkOutputs(
           if (output.isTreeArtifact()) {
             reportOutputTreeArtifactErrors(action, output, reporter, e);
           } else {
-            // Are all exceptions caught due to missing files?
-            reportMissingOutputFile(action, output, reporter, output.getPath().isSymbolicLink(), e);
+            if (e instanceof NotASymlinkException) {
+              reporter.handle(Event.error(
+                  action.getOwner().getLocation(),
+                  String.format("declared output '%s' is not a symlink", output.prettyPrint())));
```
larsrc-google, line 1544: How about "Output '%s' was declared as a symlink, but is a '%s'"?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -185,14 +185,14 @@ protected PathFragment readSymbolicLink(PathFragment path) throws IOException {
   }
 
   @Override
-  protected void createSymbolicLink(PathFragment linkPath, PathFragment targetFragment)
+  protected void createSymbolicLink(PathFragment linkPath, String rawTarget)
```
larsrc-google, line 188: I am none too happy about turning this into just taking a raw string. What is it about the PathFragment normalization that needs to be addressed in this commit? Passing around simple strings increases the ways things can go wrong.

src/main/java/com/google/devtools/build/lib/skyframe/SkyframeActionExecutor.java
fmeum: Good point, will update the corresponding PR if possible.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
fmeum: A symlink with target `bar/../foo` is not the same as a symlink with target `foo`: If `bar` is itself a symlink (or a junction on Windows), then `bar/../foo` will resolve relative to the target of `bar`.

Given these complications around symlinks, there has to be a way for Bazel to create "raw" symlinks via its filesystem abstractions. The respective PR is still WIP, but I plan to add comments explaining that the `PathFragment` overload should be preferred in almost all cases - except where it doesn't work, e.g. when explicitly creating a raw symlink.

alexjski:
> Is there a bug and design behind this?

I agree that the context would be really useful here. That also goes together with @larsrc-google's excellent question about using raw string for symlinks--the explanation is sufficient for why that technically is needed, but it would be interesting to know what particular use cases need the symlinks not to be normalized.

fmeum:
@alexjski Please see the individual PRs for context:

* https://github.com/bazelbuild/bazel/pull/15773 fixes https://github.com/bazelbuild/bazel/issues/14224 (discussing the raw vs. normalized symlink issue)
* https://github.com/bazelbuild/bazel/pull/15781 fixes https://github.com/bazelbuild/bazel/issues/10298#issuecomment-1103810996
* https://github.com/bazelbuild/bazel/pull/15700 enables tests on Windows and fixes a bug as explained in the PR description.

More recent context can also be found at https://github.com/bazelbuild/bazel/issues/15454.

The current PR is merely a testbed to verify all the PRs function together, it is not a good starting point for a review (hence is marked as Draft).

alexjski:
> The current PR is merely a testbed to verify all the PRs function together, it is not a good starting point for a review (hence is marked as Draft).

Thank you for the explanation! Should we move the discussion to the corresponding PRs instead?

aiuto:
@alexjski   You might see some context in https://github.com/bazelbuild/rules_pkg/issues/115.

The basic requirement is that when you are building distribution packages, you almost always want symlinks preserved rather than followed.  For example, you want a binary and support files to live under /usr/share/my_package/{bin, lib, messages, ...}, and you want a symlink from `/usr/bin/my_binary` to `/usr/share/my_package/bin/my_binary`.   The issue I pointed to talks about this from the perspective of NodeJS packages where, they have large symlink forests.

I've been approaching this from the rules_pkg side. My latest thinking is that
- Rule authors need to be able to tell if a source file is a symlink or not, so they can offer the ability to the users to either resolve or keep intact
- Some rules might also need the ability to handle symlinks in tree artifacts in one of two ways, but that is not Bazel's problem.  When they walk the tree they can look for symlinks and do what they wan.

PR #15887, opened 2022-07-15
Change py_binary's main attribute type to STRING

### LINKED ISSUE #15850: Expand template outputs being reported as `source file`
### Description of the bug:

I'm not 100% confident this issue is specific to [ctx.actions.expand_template](https://docs.bazel.build/versions/main/skylark/lib/actions.html#expand_template) but in certain circumstances, files generated in rules are incorrectly reported as "source file" in queries. They should instead be reported as "generated file" if the file is the output of an action.

## Workspace Example

The code below results in unexpected behavior when performing:

```
bazel query 'kind("source file", deps(//:pybin, 1))'
```

This outputs:
```
//:expanded.py
```

But this is unexpected. `expanded.py` is a generated file and should not be reported as "source file". It should instead be "generated file".

### BUILD.bazel

```starlark
load(":defs.bzl", "expand_template")

expand_template(
    name = "expanded_src",
    template = "template.txt",
)

py_binary(
    name = "pybin",
    srcs = [":expanded_src"],
    main = "expanded.py",
)
```

### defs.bzl

```starlark
def _expand_template_impl(ctx):
    output = ctx.actions.declare_file("expanded.py")

    ctx.actions.expand_template(
        output = output,
        template = ctx.file.template,
        substitutions = {},
    )

    return [DefaultInfo(
        files = depset([output]),
    )]

expand_template = rule(
    implementation = _expand_template_impl,
    attrs = {
        "template": attr.label(
            allow_single_file = True,
            mandatory = True,
        ),
    }
)
```

### template.txt

```text
print("Hello World")
```

### WORKSPACE.bazel

```starlark
workspace(name = "generated_file")
```

Linux, MacOS

release 5.2.0

_No response_

_No response_

_No response_

_No response_

fmeum:
py_binary's `main` attribute being a label results in a somewhat
unexpected dependency on the file target provided in the attribute.
In the case where the main file happens to be an implicit output file,
mentioning it in a label causes it to be reported as a "source file"
dependency, which is misleading.

This is fixed by changing the type of the `main` attribute to `STRING`.
In order to preserve backwards compatibility, only the portion of the
attribute after the last occurrence of `:` (if any) is used,
corresponding to the previous logic that only considered the name of the
label.

Fixes #15850

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPyRuleClasses.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPyRuleClasses.java
@@ -143,7 +143,7 @@ public final class BazelPyRuleClasses {
           <code>name</code> is used instead (see above). If <code>name</code> does not
           match any filename in <code>srcs</code>, <code>main</code> must be specified.
           <!-- #END_BLAZE_RULE.ATTRIBUTE --> */
-          .add(attr("main", LABEL).allowedFileTypes(PyRuleClasses.PYTHON_SOURCE))
+          .add(attr("main", STRING))
           /* <!-- #BLAZE_RULE($base_py_binary).ATTRIBUTE(python_version) -->
           Whether to build this target (and its transitive <code>deps</code>) for Python 2 or Python
           3. Valid values are <code>"PY2"</code> and <code>"PY3"</code> (the default).
--- a/src/main/java/com/google/devtools/build/lib/rules/python/PyCommon.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/python/PyCommon.java
@@ -877,7 +877,14 @@ public final class PyCommon {
     Rule target = ruleContext.getRule();
     boolean explicitMain = target.isAttributeValueExplicitlySpecified("main");
     if (explicitMain) {
-      mainSourceName = ruleContext.attributes().get("main", BuildType.LABEL).getName();
+      // The "main" attribute used to be of type BuildType.LABEL, but this logic only used the name
+      // part. For backwards compatibility, only take the part after the last label delimiter - if
+      // the string is a valid label, this will give its name.
+      String maybeLabel = ruleContext.attributes().get("main", Type.STRING);
+      int lastLabelDelimiter = Math.max(
+          Math.max(maybeLabel.lastIndexOf('@'), maybeLabel.lastIndexOf('/')),
+          maybeLabel.lastIndexOf(':'));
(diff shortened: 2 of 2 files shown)


fmeum:
@rickeylev Could you take a look?

fmeum:
@sgowroji I think that the label `team-Rules-Python` would be more appropriate here: I'm changing an attribute and the implementation of code common to all Python rules.

rickeylev:
Yeah, it being a label attribute is weird given how it's being used. I decided to dig through the history to find out why. Back in Dec 2010 (Googlers: CL/18472834), it was changed from a string to a label. From what I can ascertain, I think the general rationale was...
  * Because the main file _is_ a real file that needs to be included, it should be a label (and thus an included input dependency, as with srcs).
  * It being a string means one has to repeat the main file name (once in srcs, again in main), which isn't ergonomic.
  * A review comment also says, "This [str->label] change is necessary to eliminate a possible duplicate inclusion declaration for the file from sources."

I've no idea what that last point is referring to. Perhaps back in 2010, putting the same file multiple times into a depset/runfiles/whatever-it-was-back-then was an error? It obviously wouldn't be today.

The first two points still make sense, well, in most respects. The vast majority of the time, main _does_ refer to a source file. This query behavior + implicit outputs in srcs, as mentioned, is a case that ends up confusing and is misleading.

While I'm sympathetic to the idea of `main` being a label that refers to an actual file, I'm not sure it's actually practical. The current behavior simply isn't that (main is treated as a suffix-search string), nor has it been for decades, and making it be an input-file makes using an implicit output for main...uh...pretty hard (maybe not entirely _impossible_, but hard enough with the current rule logic). There's also edge cases introduced by having what was in the srcs attribute that is now only in another attribute (e.g. an aspect visiting srcs no longer sees it, make var expansion doesn't include it, etc); a change that isn't worth the headache imho.

The ergonomics issue for the common case (having to repat main in srcs and main) can probably be addressed with macro level logic that basically treats "main set and srcs unset" as "set srcs to main". That sort of change will have to wait until the Starlark rewrite, though.

Anyways, I'm +1 on this change in concept.

fmeum:
> It being a string means one has to repeat the main file name (once in srcs, again in main), which isn't ergonomic.

I missed that point. You are right that this is best dealt with by a Starlark macro. I will close this PR and keep an eye on the progress of Starlarkification.

> Because the main file is a real file that needs to be included, it should be a label (and thus an included input dependency, as with srcs).

I see one more issue with this: If `srcs` mentions a source directory or a tree artifact, then `main` may not be able to refer to the main file via a label, but could do so via a string.

Issue #15892, opened 2022-07-15
Config transition turns label-type attributes into lists


adam-azarchs:
### Description of the bug:

Setting `cfg` with a transition in `attr.Label` makes `ctx.attr.<attribute>` return a label list instead of a single label.  See example below.

It's quite possible that this is intended behavior.  Certainly for a one-to-many transition it would need to have this effect.  In that case this is really a documentation problem, because it certainly isn't obvious that changing the configuration should change the type of the attribute value (and in fact the [docs for the `cfg` parameter](https://bazel.build/rules/lib/attr#label.cfg) make no mention of transitions, though it does link to the [docs on configurations](https://bazel.build/rules/rules#configurations), which do, but still don't mention anything about the type change).

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

`WORKSPACE`:

```starlark
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "bazel_skylib",
    sha256 = "f7be3474d42aae265405a592bb7da8e171919d74c16f082a5457840f06054728",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.2.1/bazel-skylib-1.2.1.tar.gz",
        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.2.1/bazel-skylib-1.2.1.tar.gz",
    ],
)
```

`BUILD.bazel`:

```starlark
load("@bazel_skylib//rules:common_settings.bzl", "string_setting")
load(":my_rule.bzl", "my_rule", "my_transition_rule")

string_setting(
    name = "kind",
    build_setting_default = "baz",
    values = [
        "bar",
        "baz",
    ],
)

filegroup(
    name = "empty",
)

my_rule(
    name = "without_transition",
    src = ":empty",
)

my_transition_rule(
    name = "with_transition",
    src = ":empty",
)
```

`my_rule.bzl`:

```starlark
def _foo_transition_impl(settings, attr):
    _ignore = settings  # buildifier: disable=unused-variable
    return {
        "//:kind": "bar",
    }

_foo_transition = transition(
    implementation = _foo_transition_impl,
    inputs = [],
    outputs = ["//:kind"],
)

def _rule_impl(ctx):
    print(type(ctx.attr.src))

my_rule = rule(
    attrs = {
        "src": attr.label(
            cfg = "target",
        ),
    },
[...]

PR #15895, opened 2022-07-16
Move link to make document a bit easier to follow.

adam-azarchs:
Move the link to `#accessing-attributes-with-transitions`
from the section about how to define 1:2 transitions into the section
on how to attach an outgoing transition to an attribute, since that's
the section people are most likely to be reading when they need to
see that link.

This got me, leading to me opening https://github.com/bazelbuild/bazel/issues/15892, because I wasn't implementing a 1:2+ transition and hadn't kept reading to the end of the document.

### DIFF
--- a/site/en/rules/config.md
+++ b/site/en/rules/config.md
@@ -512,9 +512,6 @@ multi_arch_transition = transition(
 )
 ```
 
-See [Accessing attributes with transitions](#accessing-attributes-with-transitions)
-for how to read these keys.
-
 ### Attaching transitions {:#attaching-transitions}
 
 [End to end example](https://github.com/bazelbuild/examples/tree/HEAD/configurations/attaching_transitions_to_rules){: .external}
@@ -559,6 +556,9 @@ drink_rule = rule(
     ...
 ```
 Outgoing edge transitions can be 1:1 or 1:2+.
+
+See [Accessing attributes with transitions](#accessing-attributes-with-transitions)
+for how to read these keys.
 
 ### Transitions on native options {:#transitions-native-options}
 


gregestren:
I appreciate your confusion but I think there's an important conceptual connection between the "Defining 1:2+ transitions" and "Accessing attributes with transitions" sections. They're both explaining how to write the transition's implementation function.

"Outgoing edge transitions" is, of course, focused on attaching the transition itself. 

I 100% support anything that can make the docs cleaner and less overwhelming. But I'd like to dig some more on these subtleties. What's the full logical path you follow? i.e. to make any of this work you have to know a) how to define a transition implementation b) how to reach the attributes in the implementation and c) how to attach the transition. If you exclusively read the "Outgoing edge transitions" section it doesn't cover enough of that.

I hope what I'm saying makes sense. I appreciate the opportunity to trace through these docs from someone else's vantage point.

adam-azarchs:
> I appreciate your confusion but I think there's an important conceptual connection between the "Defining 1:2+ transitions" and "Accessing attributes with transitions" sections. They're both explaining how to write the transition's implementation function.

While the code examples do include transition functions, the text of that section is exclusively discussing the implementation of a rule which has attached an outgoing transition.  That's why it opens with a link back to the section on attaching a transition.

The direction I'm coming at this from is writing a 1:1 transition, which meant I never read the section on 1:2+ transitions, and thus missed that there was a separate section discussing how to access the attributes, and thus was confused why the attribute suddenly turned into a list of targets instead of just a single target.  The idea with this change is that anyone who is going to want to access an attribute with a transition will always have to first attach the transition to the attribute, but they might not have any interest on reading up on 1:2+ transitions.

> I 100% support anything that can make the docs cleaner and less overwhelming. But I'd like to dig some more on these subtleties. What's the full logical path you follow? i.e. to make any of this work you have to know a) how to define a transition implementation b) how to reach the attributes in the implementation and c) how to attach the transition. If you exclusively read the "Outgoing edge transitions" section it doesn't cover enough of that.

The section on 1:2+ transitions doesn't provide any useful context for a someone who is writing a 1:1 transition.  If that's what you're doing (like I was) then you'd read the `Defining` section (which explains 1:1 transitions), then skip the discussion on 1:2+ before reading about how to attach the transition.  Then there's several sections which aren't of interest, signaling to me that I'd read far enough to do what I came there to figure out how to do.  That attaching a transition would change the type of value of the attribute is actually pretty surprising behavior, at least if you're focused on the 1:1 case - it's not like passing other values for `cfg` will do that.

gregestren:
> > I appreciate your confusion but I think there's an important conceptual connection between the "Defining 1:2+ transitions" and "Accessing attributes with transitions" sections. They're both explaining how to write the transition's implementation function.
> 
> While the code examples do include transition functions, the text of that section is exclusively discussing the implementation of a rule which has attached an outgoing transition. That's why it opens with a link back to the section on attaching a transition.
> 

You're right. I read the logic wrong. Both transition and rule implementations can read attributes. I was referring to the wrong one.

gregestren (approved):
Thanks for the elaboration. All makes sense to me.

PR #15896, opened 2022-07-17
[WIP] Add clang target support for Mac Catalyst

imWildCat:


keith:
You also need to add the new CPU here https://github.com/bazelbuild/bazel/blob/86b800e4066389d4535fbdf280cfe9447b06df15/tools/osx/crosstool/BUILD.toolchains and I would also grep around for other cpu switches with `darwin` and see if there are other missing places

imWildCat:
@Keith Cool! Thanks!

I'm trying to build bazel locally. It does not seem to work on my M1 machine. I'm going to do it on my Intel Mac.
```
cp: external/local_jdk/include/darwin/jni_md.h: No such file or directory
Target //src:bazel-dev failed to build
INFO: Elapsed time: 0.562s, Critical Path: 0.14s
```

keith:
you shouldn't have a problem on the m1 machine either. I'd try bazelisk if you haven't to make sure you're on a recent version

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -23,6 +23,14 @@ OSX_TOOLS_CONSTRAINTS = {
         "@platforms//os:osx",
         "@platforms//cpu:x86_64",
     ],
+    "maccatalyst_x86_64": [
+        "@platforms//os:ios",
+        "@platforms//cpu:x86_64",
+    ],
+    "maccatalyst_arm64": [
+        "@platforms//os:ios",
```
imWildCat, line 31: maybe we should use `macOS` here? or defining new platform for Catalyst?

Not sure about this.

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -93,6 +93,10 @@ def _impl(ctx):
         target_system_name = "arm64-apple-macosx{}".format(target_os_version)
     elif (ctx.attr.cpu == "darwin_arm64e"):
         target_system_name = "arm64e-apple-macosx{}".format(target_os_version)
+    elif (ctx.attr.cpu == "maccatalyst_x86_64"):
+        target_system_name = "x86_64-apple-ios{}-macabi".format(target_os_version)
```
imWildCat, line 97: I think this is for clang?

For example,

```shell
clang -target arm64-apple-ios13.1-macabi catalyst_main.c
```

src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
```diff
@@ -55,7 +55,7 @@ public enum ApplePlatform implements ApplePlatformApi {
   private static final ImmutableSet<String> TVOS_DEVICE_TARGET_CPUS =
       ImmutableSet.of("tvos_arm64");
   private static final ImmutableSet<String> CATALYST_TARGET_CPUS =
```
imWildCat, line 57: Looks like this is different from iOS and macOS?

imWildCat:
> you shouldn't have a problem on the m1 machine either. I'd try bazelisk if you haven't to make sure you're on a recent version

Thanks! It is the problem of my `JAVA_HOME`. Now it's fixed.

tools/osx/crosstool/BUILD.toolchains
keith: So here's where we might have a problem. The ideal solution here is that we add a third entry to this list to disambiguate this, but the value we need to add is in the apple_support repo, which bazel does not currently depend on. I've been lightly talking with @gregestren about what to do for this case, but we haven't reached a conclusion.

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -23,6 +23,14 @@ OSX_TOOLS_CONSTRAINTS = {
         "@platforms//os:osx",
         "@platforms//cpu:x86_64",
     ],
+    "maccatalyst_x86_64": [
```
keith, line 26: Given the CPUs in java code we might need this to be `catalyst_x86_64` without the `mac` prefix.

tools/osx/crosstool/cc_toolchain_config.bzl
keith: Yep looks right, it's used for clang, swift, and the linker.

tools/osx/crosstool/BUILD.toolchains
keith: But in the meantime I'm not sure if this should be ios or macOS. Probably macOS given that's what it's targeting running on?

imWildCat: I think `@platforms//os:osx` (macOS) makes sense! Because when we're using `xcodebuild` to compile an iOS project for Catalyst, the value of `-sdk` flag is `macosx`.

imWildCat: gentle ping @gregestren 

do we want to move forward on Mac Catalyst support? 
I can take the action when you have any feedback.

gregestren: I expect to discuss the bazel -> apple_support dependency at a meeting at the end of October. I don't have strong opinions in the meantime - it reads like this conversation has an interim consensus?

imWildCat: @gregestren Thanks for your reply! No pressure on this. Because I'm in a self doubt on this. I can hardly see any interest about community support of Catalyst. This is not only limited to Bazel. While using Rust, I'm facing the same issue.

I'm thinking about writing macOS app instead of a Catalyst app.

meisterT:
Closing old draft PR - I assume it's obsolete by now.

PR #15903, opened 2022-07-18
target pattern file: allow comments

sluongng:
`--target_pattern_file` could be a great tool assisting migration of a
code base from one Bazel config to another.  User could define a list of
targets that can be built with the new configs and incrementally expand
that list with each change that was introduced.  Multiple users / teams
could collaborate on such migration by adding / removing their targets
into / from the pattern file.

Having the ability to provide comments to annotate the targets in the
pattern file with extra information during the migration process added a
great value and context for build maintainers to track migration
progress.

Add support for Bash-style comments to the target-pattern file.


src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternsHelper.java
```diff
@@ -54,7 +56,11 @@ public static List<String> readFrom(CommandEnvironment env, OptionsParsingResult
       Path residuePath =
           env.getWorkingDirectory().getRelative(buildRequestOptions.targetPatternFile);
       try {
-        targets = FileSystemUtils.readLines(residuePath, UTF_8);
+        targets = FileSystemUtils.readLines(residuePath, UTF_8).stream()
+            .map(s -> s.split("#")[0])
+            .map(String::trim)
+            .filter(Predicate.not(String::isEmpty))
+            .collect(Collectors.toList());
```
Wyverald, line 63: prefer using ImmutableList.toImmutableList with a static import

Wyverald (changes requested):

src/main/java/com/google/devtools/build/lib/runtime/commands/TargetPatternsHelper.java
sluongng: fixed

sluongng:
unclear why `LocalDiffAwarenessIntegrationTest` failed on darwin https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/11687#018215db-540d-4499-a0fc-8724f35afacf

gona rebase on top of latest master to see if it gets better.

Wyverald:
That's a flaky test, no need to worry :)

Wyverald (approved):

brentleyjones:
@bazel-io flag

Issue #15856, opened 2022-07-11
Implement the credential helpers proposal


tjgq:
### Description of the feature request:

Tracking issue for implementing https://github.com/bazelbuild/proposals/blob/main/designs/2022-06-07-bazel-credential-helpers.md.

### What underlying problem are you trying to solve with this feature?

N/A

### Which operating system are you running Bazel on?

N/A

### What is the output of `bazel info release`?

N/A

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

N/A

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
N/A
```

Wyverald:
@bazel-io flag

ckolli5:
@bazel-io fork 5.3.0

meteorcloudy:
Since this feature is implemented and cherry picked to 5.3.0, should we close this one?

tjgq:
The proposal hasn't been fully implemented yet (e.g. we're missing support for repository fetching).

meteorcloudy:
@tjgq What's the ETA for implementing this? Ideally, we don't want to delay the 6.0 release further more, and since this should be a backwards compatible feature, we can cherry pick it in 6.x releases?

tjgq:
I was planning to get this submitted in time for 6.0.0, but other things got in the way. Let's do it for a 6.x release instead.

tjgq:
This has been fully implemented in 7.x and backported to 6.x.

PR #15930, opened 2022-07-20
Add netrc support to --bes_backend

### LINKED ISSUE #15928: --bes_backend doesn't support reading credentials from .netrc
`--remote_{downloader,cache,executor}` all support reading credentials from `.netrc`: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java;l=1098;drc=1bc1ce3e0d73c35047c29ef85e832db041e0f238

`--bes_backend` doesn't: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/buildeventservice/BazelBuildEventServiceModule.java;l=79;drc=4cd266aa1dfa53d8c8de44f7895edcfd46f74725 calling into https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java;l=194;drc=4cd266aa1dfa53d8c8de44f7895edcfd46f74725

Found while working on #15856

Yannic:
Progress on https://github.com/bazelbuild/bazel/issues/15856

Fixes #15928


Yannic:
@tjgq PTAL

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
```diff
@@ -217,12 +237,51 @@ public static CallCredentialsProvider newCallCredentialsProvider(@Nullable Crede
    */
   @Nullable
   public static Credentials newCredentials(@Nullable AuthAndTLSOptions options) throws IOException {
+    Optional<Credentials> credentials = newGoogleCredentials(options);
+
+    return credentials.orElse(null);
```
tjgq, line 242: Can we make `newGoogleCredentials` public and call it directly, avoiding this indirection? (It's fine if it returns a `@Nullable` for now; I'd rather fix all callers in a single CL than introduce temporary forwarding methods.)

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
```diff
@@ -191,7 +197,21 @@ private static NettyChannelBuilder newNettyChannelBuilder(String targetUrl, Stri
    * @throws IOException in case the call credentials can't be constructed.
    */
   @Nullable
```
tjgq, line 199: Copy the comment from `newCallCredentials` here, to make it clear where we attempt to retrieve credentials from.

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
```diff
@@ -191,7 +197,21 @@ private static NettyChannelBuilder newNettyChannelBuilder(String targetUrl, Stri
...
+    Credentials creds = newCredentials(reporter, clientEnv, fileSystem, options);
+    if (creds != null) {
+      return MoreCallCredentials.from(creds);
+    }
+    return null;
+  }
+
+  @Nullable
+  @VisibleForTesting
+  public static CallCredentials newCallCredentialsForTesting(AuthAndTLSOptions options) throws IOException {
```
tjgq, line 214: Nit: should this be called `newGoogleCallCredentialsForTesting`? This would makes the names in this file a bit more consistent (`Google` in the name means we only look at flags, otherwise we also look at the `.netrc`).

tjgq (changes requested):

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
```diff
@@ -243,7 +302,7 @@ public static Credentials newCredentials(@Nullable AuthAndTLSOptions options) th
    * @throws IOException in case the credentials can't be constructed.
    */
   @VisibleForTesting
-  public static Credentials newCredentials(
+  public static Credentials newGoogleCredentials(
```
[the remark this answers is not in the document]
Yannic: sure, done

src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
Yannic: done

tjgq (changes requested):
Looks like `//src/test/java/com/google/devtools/build/lib/buildeventservice:BazelBuildEventServiceModuleTest` needs fixing?

Yannic:
Fixed, PTAL

tjgq (approved):

tjgq:
FYI, I'll have to remove a couple of `@VisibleForTesting` annotations when importing, because of Google-internal callers that aren't part of the open source release.

PR #15935, opened 2022-07-21
updating patch files + adding external transitive deps definitions for grpc upgrade

aranguyen:
1st pr to be merge for grpc 1.47.x upgrade

PR #15944, opened 2022-07-22
Use new grpc version

aranguyen:
Use grpc 1.47.x. This should be the second pr to be merged.


katre (approved):
#15935 is merged, this looks good to import.

PR #15950, opened 2022-07-22
Introduce persistent_multiplex_android_resource_processor expansion flag

Bencodes:
Alternative approach to https://github.com/bazelbuild/bazel/pull/15321#issuecomment-1192772584 that leverages an expansion CLI flag instead of proliferating unique experimental multiplex flags.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
@@ -861,8 +861,40 @@ public class AndroidConfiguration extends Fragment implements AndroidConfigurati
           "--strategy=AndroidManifestMerger=worker",
           "--strategy=Aapt2Optimize=worker",
           "--strategy=AARGenerator=worker",
+          "--strategy=Desugar=worker",
+          "--strategy=DexBuilder=worker",
         })
     public Void persistentResourceProcessor;
+
+    @Option(
+        name = "persistent_multiplex_android_resource_processor",
+        defaultValue = "null",
+        documentationCategory = OptionDocumentationCategory.EXECUTION_STRATEGY,
+        effectTags = {
+          OptionEffectTag.HOST_MACHINE_RESOURCE_OPTIMIZATIONS,
+          OptionEffectTag.EXECUTION,
+        },
+        help = "Enable the persistent Android resource processor by using workers.",
+        expansion = {
+          "--persistent_android_resource_processor",
+          "--modify_execution_info=AaptPackage=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidResourceParser=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidResourceValidator=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidResourceCompiler=+supports-multiplex-workers",
+          "--modify_execution_info=RClassGenerator=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidResourceLink=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidAapt2=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidAssetMerger=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidResourceMerger=+supports-multiplex-workers",
+          "--modify_execution_info=AndroidCompiledResourceMerger=+supports-multiplex-workers",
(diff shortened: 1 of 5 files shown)


src/test/shell/bazel/android/resource_processing_integration_test.sh
```diff
@@ -123,7 +123,7 @@ function test_persistent_multiplex_resource_processor() {
   setup_font_resources
 
   assert_build //java/bazel:bin --persistent_android_resource_processor \
-    --experimental_persistent_multiplex_busybox_tools
+    --persistent_multiplex_android_resource_processor
```
Bencodes, line 126: With this approach we also wouldn't need to add a unique test case just for `DexBuilder` multiplex workers as it would be covered by the `persistent_multiplex_android_resource_processor` expansion here.

meisterT:
cc @larsrc-google

Bencodes:
@larsrc-google is this something you'd be interested in merging to clean up the multiplex worker flags on the Android side?

src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
```diff
@@ -861,9 +861,41 @@ public static class Options extends FragmentOptions {
           "--strategy=AndroidManifestMerger=worker",
           "--strategy=Aapt2Optimize=worker",
           "--strategy=AARGenerator=worker",
+          "--strategy=Desugar=worker",
+          "--strategy=DexBuilder=worker",
         })
     public Void persistentResourceProcessor;
 
+    @Option(
+        name = "persistent_multiplex_android_resource_processor",
```
larsrc-google, line 870: If this is meant to be used just while rolling the multiplex implementation, it ought to be an `experimental_` flag. If not, why not just set `support-multiplex-workers` in BusyBoxActionBuilder?

Bencodes: I wouldn't be against just setting `support-multiplex-workers` everywhere but that might be on the riskier side given that 6.x will be the first release where there's actual multiplex support for all of the tools that the busybox wrapper runs.

meisterT: I don't think we want the flag around forever, so what is the testing strategy?

ahumesky: Addressing a few things:

- There's a test for this in resource_processing_integration_test.sh, so I think that's covered
- I think setting `supports-multiplex-workers` in BusyBoxActionBuilder would also turn this on internally, and I don't think we want to sign up for figuring out how to roll that out right now. I think we should consider setting workers to be on for all this by default, but that can be a separate change
- I'm ok with leaving this without `experimental_`, because this isn't really a new thing that's being actively changed or developed, and it's not an API or similar

src/test/shell/bazel/android/resource_processing_integration_test.sh
ahumesky: we can consolidate in a followup change

ahumesky (approved):

ahumesky:
I'll start importing this

src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
ahumesky: I realized that with desugar and dexing in the list, this isn't strictly about resource processing. Maybe a more accurate name would be `--persistent_multiplex_android_tools`

Similarly, adding desugar and dexing to `--persistent_android_resource_processor` changes the behavior of the flag, and seems to make that name less accurate. We can't easily rename that though flag since it's been around for a while already. The solution would seem to be to introduce `--persistent_android_dex_desugar`.

If that sounds good I can make the change while importing this PR

ahumesky: Yeah so even though we were trying to simplify things, I seem to be now proposing:
```
--persistent_android_resource_processor
--persistent_multiplex_android_resource_processor
--persistent_android_dex_desugar
--persistent_multiplex_android_dex_desugar
--persistent_multiplex_android_tools
```
`--persistent_android_resource_processor` and `--persistent_multiplex_android_tools` adds all the previous ones (directly or indirectly).

But this seems to be the most accurate and safest.

PR #15953, opened 2022-07-22
Fix `toolchain` typo in bazel-and-java.md

jonshea:
The "Java and Bazel" page twice referenced a method
`register_toolchain` which contains a typo. The method's correct name is
`register_toolchains`, with a plural "s".


gregestren (approved):
Nice catch.

PR #15905, opened 2022-07-18
Add CommandLinePathFactory to CommandEnvironment

Yannic:
Progress on #15856

PR #15954, opened 2022-07-22
Add CommandLinePathFactory to CommandEnvironment

ckolli5:
Progress on #15856

Closes #15905.

PiperOrigin-RevId: 462369707
Change-Id: If8dd493227b67dd4ffe30c9e0af4d8f71de4fb9f


tjgq:
The cause of the build errors is a bad merge in `CommandEnvironment.java`. Compare with the changes to that file in the original PR #15905. The `getBuildEventListener` method and associated logic should not be part of this change.

PR #15963, opened 2022-07-24
Do not force unresolved symlinks to be absolute

### LINKED ISSUE #14224: ctx.actions.symlink doesn't make relative symlink

fmeum:
The documentation of `ctx.actions.symlink(target_path = ...)` states
that the given path is used as the symlink target without any changes,
but in reality this path has so far always been converted into an
absolute path by prepending the exec root. This is changed by this
commit, which gets very close to the documented behavior by preserving
the target path as is except for the standard normalization applied by
`PathFragment`. Improving the situation even further would require
modifying or adding to Bazel's core file system API, which may be done
in a follow-up PR.

Since relative symlinks only resolve correctly at the location they were
originally created, the logic for staging symlinks as runfiles in the
sandbox necessarily becomes more complicated: Analogously to the case of
local unsandboxed execution, symlinks in runfiles are now staged as
symlinks to the original exec path of the relative symlink under the
sandboxed exec root as well as the relative symlink itself in that
location.

As a side-effect of the switch to relative symlinks, this commit
resolves a non-hermeticity issue observed in
https://github.com/bazelbuild/bazel/issues/10298#issuecomment-1192939423
Integration tests are added to verify that symlinks staged in the
sandbox are no longer resolved non-hermetically, including the case
where they are contained in runfiles.

Fixes #14224

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BUILD
@@ -249,6 +249,7 @@ java_library(
         "starlark/StarlarkRuleContext.java",
         "starlark/StarlarkRuleTransitionProvider.java",
         "starlark/StarlarkTransition.java",
+        "starlark/UnresolvedSymlinkAction.java",
         "test/AnalysisTestActionBuilder.java",
         "test/BaselineCoverageAction.java",
         "test/CoverageCommon.java",
--- a/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/actions/SymlinkAction.java
@@ -36,7 +36,10 @@ import com.google.devtools.build.lib.vfs.PathFragment;
 import java.io.IOException;
 import javax.annotation.Nullable;
 
-/** Action to create a symbolic link. */
+/**
+ * Action to create a symlink to a known-to-exist target with alias semantics similar to a true copy
+ * of the input (if any).
+ */
 public final class SymlinkAction extends AbstractAction {
   private static final String GUID = "7f4fab4d-d0a7-4f0f-8649-1d0337a21fee";
 
@@ -150,13 +153,6 @@ public final class SymlinkAction extends AbstractAction {
     Preconditions.checkState(!execPath.isAbsolute());
     return new SymlinkAction(
         owner, execPath, primaryInput, primaryOutput, progressMessage, TargetType.FILESET);
-  }
-
-  public static SymlinkAction createUnresolved(
-      ActionOwner owner, Artifact primaryOutput, PathFragment targetPath, String progressMessage) {
-    Preconditions.checkArgument(primaryOutput.isSymlink());
-    return new SymlinkAction(
-        owner, targetPath, null, primaryOutput, progressMessage, TargetType.OTHER);
   }
 
   /**
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkActionFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkActionFactory.java
@@ -273,7 +273,7 @@ public class StarlarkActionFactory implements StarlarkActionFactoryApi {
             ? (String) progressMessageUnchecked
             : "Creating symlink " + outputArtifact.getExecPathString();
 
-    SymlinkAction action;
+    Action action;
     if (targetFile != Starlark.NONE) {
       Artifact inputArtifact = (Artifact) targetFile;
       if (outputArtifact.isSymlink()) {
@@ -324,10 +324,10 @@ public class StarlarkActionFactory implements StarlarkActionFactoryApi {
       }
 
       action =
-          SymlinkAction.createUnresolved(
+          new UnresolvedSymlinkAction(
               ruleContext.getActionOwner(),
               outputArtifact,
-              PathFragment.create((String) targetPath),
+              (String) targetPath,
               progressMessage);
     }
     registerAction(action);
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
@@ -1,0 +1,112 @@
+// Copyright 2022 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.analysis.starlark;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.devtools.build.lib.actions.AbstractAction;
+import com.google.devtools.build.lib.actions.ActionExecutionContext;
+import com.google.devtools.build.lib.actions.ActionExecutionException;
+import com.google.devtools.build.lib.actions.ActionKeyContext;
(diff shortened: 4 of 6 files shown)


fmeum:
@alexjski Would you be available to review?

CC @gregmagolan This may affect the symlink targets generated by the experimental branch of rules_js.

gregmagolan:
I tried this out on rules_js and I hit a new issue with the sandbox and runfiles related to the change. Will do my best to explain,

A symlink in the sandbox in the runfiles tree now has a relative path to its corresponding location outside of the runfiles tree in the sandbox,

```
$ ls -la /private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/sandbox/darwin-sandbox/1/execroot/aspect_rules_js/bazel-out/darwin-opt-exec-2B5CBBC6/bin/examples/js_binary/bin.sh.runfiles/aspect_rules_js/node_modules
total 0
drwxr-xr-x  4 greg  wheel  128 25 Jul 09:01 .
drwxr-xr-x  5 greg  wheel  160 25 Jul 09:01 ..
drwxr-xr-x  3 greg  wheel   96 25 Jul 09:01 .aspect_rules_js
lrwxr-xr-x  1 greg  wheel   81 25 Jul 09:01 acorn -> ../../../../../../../../bazel-out/darwin-opt-exec-2B5CBBC6/bin/node_modules/acorn
```

Looking at where this is pointing to,

```
$ ls -la /private/var/tmp/_bazel_greg/7aed427a991d86f99332ab79b9b11780/sandbox/darwin-sandbox/1/execroot/aspect_rules_js/bazel-out/darwin-opt-exec-2B5CBBC6/bin/node_modules
total 0
drwxr-xr-x  3 greg  wheel   96 25 Jul 09:01 .
drwxr-xr-x  4 greg  wheel  128 25 Jul 09:01 ..
lrwxr-xr-x  1 greg  wheel   47 25 Jul 09:01 acorn -> .aspect_rules_js/acorn@8.4.0/node_modules/acorn
```

The problem is that in the sandbox `.aspect_rules_js/acorn@8.4.0/node_modules/acorn` (a TreeArtifact that is the actual npm package in the symlinked node_module structure) doesn't exist at that location. It only exists in the sandbox in the runfiles tree.

This seems like the correct sandboxing behaviour since Bazel is only pulling the runfiles tree into the sandbox and not corresponding files outside of the runfiles tree; tho I'm not sure why the symlink `execroot/aspect_rules_js/bazel-out/darwin-opt-exec-2B5CBBC6/bin/node_modules` is in the sandbox.

Could symlinks in the runfiles tree be relative to the runfiles tree? If this was done in the sandbox should resolve this issue. If symlinks were runfiles relative both in the sandbox and in the runfiles tree outside of the sandbox that would also make resolving within the runfiles tree outside of the sandbox more hermetic.

fmeum:
@gregmagolan That's a great suggestion! I like the additional hermiticity properties for local execution. I will submit a separate PR.

alexjski:
> @gregmagolan That's a great suggestion! I like the additional hermiticity properties for local execution. I will submit a separate PR.

Does this obsolete this PR? Is the idea to just plant the symlinks unresolved (which would effectively deem regular symlinks relative to the sandbox and runfiles to the runfiles)?

fmeum:
That's a good question - it turns out that staging symlinks as is as runfiles in local execution is more complicated than I thought at first: The `SourceManifestAction` doesn't have the runfiles as inputs and thus can't read the symlinks. All of this is much easier in the sandboxes or with RBE.

I prepared https://github.com/bazelbuild/bazel/pull/15982, but that requires going through a rather restrictive new Starlark API.

@alexjski Can you explain why `AbstractAction` forbids overriding `getInputs` in non-test code? The ideal API (namely none) would require making `getInputs` of `SourceManifestAction` lazily compute something like `NestedSet.wrap(runfiles.getArtifacts(), Iterables.filter(Artifact::isSymlink))`. Would that have the potential to cause regressions?

fmeum:
@alexjski To make my last comment more concrete (and not open yet another PR), here is what I have in mind: https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#diff-5735787cb0cd600c97d3b645c63e48429f198355b53979b724a61e07683785b2R144.

That change would get us unresolved symlinks in runfiles without the need for additional Starlark API to differentiate them from regular artifacts (as done in https://github.com/bazelbuild/bazel/pull/15982). I don't have the means to benchmark this properly and certainly don't want to propose a change that would regress performance for symlink-free runfiles.

alexjski:
> @alexjski Can you explain why AbstractAction forbids overriding getInputs in non-test code? The ideal API (namely none) would require making getInputs of SourceManifestAction lazily compute something like NestedSet.wrap(runfiles.getArtifacts(), Iterables.filter(Artifact::isSymlink)). Would that have the potential to cause regressions?

Some code archaeology took me [here](https://github.com/bazelbuild/bazel/commit/5ea2b14a9cece2e42779a8b3e4e8f3483e991ee1). It seems it is a code hygiene thing--the only reason when the inputs aren't trivially known for an action is inputs discovery, which the code reflects since this change. I still don't understand why you need to override that, but I hope that at least that helps understand the comment.

> @alexjski To make my last comment more concrete (and not open yet another PR), here is what I have in mind: https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#diff-5735787cb0cd600c97d3b645c63e48429f198355b53979b724a61e07683785b2R144.
>
> That change would get us unresolved symlinks in runfiles without the need for additional Starlark API to differentiate them from regular artifacts (as done in https://github.com/bazelbuild/bazel/pull/15982). I don't have the means to benchmark this properly and certainly don't want to propose a change that would regress performance for symlink-free runfiles.

Personally, I am wondering why the metadata for symlinks/articacts themselves don't carry the information about where they point. I also do not understand why we stage those as relative to the execroot:

1. If we want to have a symlink to a source/output artifact, we don't need to do that by path (arguably, we shouldn't)
2. If we want a symlink to a parent directory of an artifact, that should be possible to do with unresolved relative paths (say `../../tools` from `foo/bar/symlink`), no need to anchor that in execroot.
3. For unresolved relative symlinks, the relativizing them to the execroot is obviously a problem.

I presume that this is just to make to simplify the situation so that we don't need to have de-normalized paths.

Wouldn't not resolving the symlinks just work for all cases: local/sandbox/runfiles? In each case, the location of the symlink itself would express the information.

The other thing I am confused by, is why `ctx.actions.symlink` was added as opposed to extending the `runfiles::symlinks` feature to allow unresolved symlinks. That would certainly make the runfiles case easier.

At this point, I think it may be worth to ask @lberki for his opinion here--he should be back in couple of weeks. Please let me know if you would like stronger inputs on that earlier, I can then look for others who already worked on that/ramp up more on the existing work in the area.

gregmagolan:
Tested these changes against rules_js and besides unearthing a latent bug due in the sandbox guards everything worked as expected. Will try RBE with these changes as well and report back.

gregmagolan:
I tested the changes in https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#diff-5735787cb0cd600c97d3b645c63e48429f198355b53979b724a61e07683785b2R144 against rules_js. Besides unearthing a latent rules_js bug in the sandbox guards, everything worked as expected. I will test with RBE next and report back.

PR #15969, opened 2022-07-25
Minor fixes to support OpenBSD

3405691582:
Allow Bazel to build and run on openbsd.
* Darwin and OpenBSD both support kevent and the relevant sysctl, so use the darwin process-tools for both.
* They just differ in the internal structure of `struct kinfo_proc`, so account for that.
* Use the `env` trick to locate `bash`, for portability.
 
`process-tools-darwin.cc` should probably be renamed to `process-tools-kevent.cc`, but that can come later.

### DIFF
--- a/src/main/tools/BUILD
+++ b/src/main/tools/BUILD
@@ -15,6 +15,7 @@ cc_library(
     name = "process-tools",
     srcs = ["process-tools.cc"] + select({
         "//src/conditions:darwin": ["process-tools-darwin.cc"],
+        "//src/conditions:openbsd": ["process-tools-darwin.cc"],
         "//conditions:default": ["process-tools-linux.cc"],
     }),
     hdrs = ["process-tools.h"],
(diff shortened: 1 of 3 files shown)


3405691582:
20.04 presubmit failure looks unrelated.

fmeum:
Seems so, you can retry with an amend and force push.

aiuto (approved):
Thanks for this fix.

aiuto:
Merge CL in flight.  But it is doing some funny stuff. https://buildkite.com/bazel/google-bazel-presubmit/builds/58472

3405691582:
Fairly sure this is because of https://github.com/bazelbuild/bazel/commit/8e259a340d4196d64650aaf88797a06603cb3392 -- I can reproduce this test bug at HEAD on Linux and the filepaths mention "third_party/android_dex/" in the logs.

aiuto:
Thanks.  Do you have a fix?

3405691582:
Not at the moment, it might take me a bit to trace out what's going on -- whether there's a dependency missing or whether the srcs_list genquery should be updated. I can continue to investigate, though.

aiuto:
Oh. sorry. Wrong thread. It is up to Alex to find the fix.
You can just wait it out.

PR #15984, opened 2022-07-26
Prevent aspects from executing on incompatible targets

### LINKED ISSUE #15271: Aspects get falsely evaluated against incompatible targets

philsc:
This patch prevents aspect functions from being evaluated when their
associated targets are incompatible. Otherwise aspects can generate
errors that should never be printed at all. For example, an aspect
evaluating an incompatible target may generate errors about missing
providers. This is not the desired behaviour.

The implementation in this patch short-circuits the `AspectValue`
creation if the associated target is incompatible.

I had tried to add an `IncompatiblePlatformProvider` to the
corresponding `ConfiguredAspect` instance, but then Bazel
complained about having duplicate `IncompatiblePlatformProvider`
instances.

Fixes  #15271

(cherry picked from commit 6d71850f6d5e9989826114395d58377769f44e4b)

This cherry-pick required a few tweaks.
* `state.transitivePackagesForPackageRootResolution` is not available
  in `AspectFunction` on 5.3. I changed this call to match what the
  code does a little further down when required providers are missing.
* The error message for explicitly requesting an incompatible target is
  different on 5.3. That required adjusting the regexes in the test a bit.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
@@ -302,7 +302,9 @@ final class AspectFunction implements SkyFunction {
           aspect,
           target.getLocation(),
           ConfiguredAspect.forNonapplicableTarget(),
-          state.transitivePackagesForPackageRootResolution.build());
+          /*transitivePackagesForPackageRootResolution=*/ NestedSetBuilder
+              .<Package>stableOrder()
+              .build());
     }
 
     if (AliasProvider.isAlias(associatedTarget)) {
--- a/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/AspectFunction.java
@@ -36,6 +36,7 @@ import com.google.devtools.build.lib.analysis.DependencyKey;
 import com.google.devtools.build.lib.analysis.DependencyKind;
 import com.google.devtools.build.lib.analysis.DuplicateException;
 import com.google.devtools.build.lib.analysis.ExecGroupCollection.InvalidExecGroupException;
+import com.google.devtools.build.lib.analysis.IncompatiblePlatformProvider;
 import com.google.devtools.build.lib.analysis.InconsistentAspectOrderException;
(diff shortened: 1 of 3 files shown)


philsc:
Oh boy. Looks like I shouldn't have assumed that a clean cherry-pick meant that the code would actually work.

PR #15990, opened 2022-07-27
Let labels in --aspect go through repo mapping

### LINKED ISSUE #15989: Labels passed to `--aspect` don't go through repo mapping
### Description of the bug:

The labels passed to the `--aspect` CLI option don't go through repository mapping, which e.g. leads to issues with provider identity when using repo-absolute labels.

```shell
$ git checkout https://github.com/bazelbuild/rules_go
$ cd rules_go && git checkout eb03b81312c3d459fb1ffdb18da9aa122e649905
$ bazel build --aspects=//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect --output_groups=go_pkg_driver_json_file,go_pkg_driver_stdlib_json_file,go_pkg_driver_srcs //tests/core/...
... passes ...
$ bazel build --aspects=@io_bazel_rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect --output_groups=go_pkg_driver_json_file,go_pkg_driver_stdlib_json_file,go_pkg_driver_srcs //tests/core/...
ERROR: /home/fhenneke/git/rules_go/tests/core/cgo/BUILD.bazel:300:10: in @io_bazel_rules_go//go/tools/gopackagesdriver:aspect.bzl%go_pkg_info_aspect aspect on go_binary rule //tests/core/cgo:cc_deps: 
	File "/home/fhenneke/.cache/bazel/_bazel_fhenneke/da2ee2360ba14ac857bdb47dfbba9b7d/external/io_bazel_rules_go/go/tools/gopackagesdriver/aspect.bzl", line 116, column 47, in _go_pkg_info_aspect_impl
		stdlib_json_file = ctx.attr._go_stdlib[GoStdLib]._list_json
Error: <target @io_bazel_rules_go//:stdlib> (rule 'stdlib') doesn't contain declared provider 'GoStdLib'
```

Linux

5.2.0 / HEAD

_No response_

_No response_

_No response_

_No response_

fmeum:
The labels passed in via `--aspect` now go through the main repository's
repository mapping, which ensures that Starlark provider identity is
established correctly even when e.g. the main workspace is referred to
via it's name rather.

Fixes #15989

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/starlark/StarlarkDefinedAspectsTest.java
+++ b/src/test/java/com/google/devtools/build/lib/starlark/StarlarkDefinedAspectsTest.java
@@ -8418,6 +8418,32 @@ public class StarlarkDefinedAspectsTest extends AnalysisTestCase {
     assertThat(aspectAResult).isEqualTo("aspect_a on target //test:dep_target, p = p_v1");
   }
 
+  @Test
+  public void testAspectLabelIsRepoMapped() throws Exception {
+    scratch.appendFile("WORKSPACE", "workspace(name = 'my_repo')");
+    scratch.file(
+        "test/aspect.bzl",
+        "load(':rule.bzl', 'MyInfo')",
+        "def _impl(target, ctx):",
+        "   if MyInfo not in target:",
+        "       fail('Provider identity mismatch')",
+        "   return struct()",
+        "MyAspect = aspect(implementation=_impl)");
+    scratch.file(
+        "test/rule.bzl",
+        "MyInfo = provider()",
+        "def _impl(ctx):",
+        "    return [MyInfo()]",
+        "my_rule = rule(implementation=_impl)");
+    scratch.file("test/BUILD",
+        "load(':rule.bzl', 'my_rule')",
+        "my_rule(name = 'target')");
+
+    AnalysisResult result = update(ImmutableList.of("@my_repo//test:aspect.bzl%MyAspect"),
+        "//test:target");
+    assertThat(result.hasError()).isFalse();
+  }
+
   private ConfiguredAspect getConfiguredAspect(
       Map<AspectKey, ConfiguredAspect> aspectsMap, String aspectName) {
     for (Map.Entry<AspectKey, ConfiguredAspect> entry : aspectsMap.entrySet()) {
--- a/src/main/java/com/google/devtools/build/lib/analysis/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BUILD
@@ -641,6 +641,7 @@ java_library(
         "//src/main/java/com/google/devtools/build/lib/skyframe:skyframe_cluster",
         "//src/main/java/com/google/devtools/build/lib/skyframe:target_pattern_phase_value",
         "//src/main/java/com/google/devtools/build/lib/util",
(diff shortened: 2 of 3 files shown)


fmeum:
@meteorcloudy Could you take a look or should this wait for @Wyverald?

meteorcloudy:
Thanks for the fix! Can you try to add a test in https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/starlark/StarlarkDefinedAspectsTest.java?

@gregestren Can you also take a look?

gregestren:
I don't think I have special insight into this - certainly not the repo mapping part.  What are we looking for in the review?

fmeum:
@meteorcloudy I added a test case.

meteorcloudy:
> I don't think I have special insight into this - certainly not the repo mapping part. What are we looking for in the review?

@gregestren I was hoping you could double check if the exception handling is fine, but I think it looks good ;)

> I added a test case.

@fmeum Nice, thanks!

src/test/java/com/google/devtools/build/lib/starlark/StarlarkDefinedAspectsTest.java
```diff
@@ -8418,6 +8418,32 @@ public void testRuleAspectWithMandatoryParameterProvided() throws Exception {
...
+        "test/rule.bzl",
+        "MyInfo = provider()",
+        "def _impl(ctx):",
+        "    return [MyInfo()]",
+        "my_rule = rule(implementation=_impl)");
+    scratch.file("test/BUILD",
+        "load(':rule.bzl', 'my_rule')",
+        "my_rule(name = 'target')");
+
+    AnalysisResult result = update(ImmutableList.of("@my_repo//test:aspect.bzl%MyAspect"),
```
meteorcloudy, line 8442: Hmm, what repo mapping is this label going through? How would this test fail without your change?

fmeum: I manually verified that it fails without my change.

It goes through the main repositories repo mapping, which AFAIU maps the workspace name `my_repo` to the canonical name, which is just the empty string. Before my change, the label would not go through repo mapping at all, which would cause the `MyInfo` provider to be loaded from the (non-existent) `my_repo` repo and thus differ from the one defined in the main repo.

meteorcloudy: I see, in a real WORKSPACE, there will be a local_repository called "my_repo" that pointing to the main project's root directory to make that "external repo" available, which is an unfortunate hack, but it seems that's not set up for this test case which makes this test case work. Thanks for the clarifying!

meteorcloudy (approved):

fmeum:
@meteorcloudy It should also work with the hack - this test case is essentially my original reproducer trimmed down to the essentials. Even if the external repo exists, I think the `RepositoryName` instance will still be different from that of the main repo, meaning that loaded providers will be distinct.

meteorcloudy:
> I think the RepositoryName instance will still be different from that of the main repo, meaning that loaded providers will be distinct.

Oh, got it, thanks!

meteorcloudy:
@sgowroji Can you help merge this, thanks!

PR #15991, opened 2022-07-27
Patching upb

aranguyen:
Eventhough upb already uses constraint values from `@platforms` in their latest version. However, recent grpc upgrade to 1.47.x uses an older commit of `upb` which still use `@bazel_tools//platforms_x`. I will send a separate pr to use the patch in `distdir_deps.bzl`


meteorcloudy:
Can we just use a newer version of upb in `distdir_deps.bzl` (if that's not breaking grpc)?

aranguyen:
@meteorcloudy I think that might not work. We can have `upb` in the `WORKSPACE` file preceding the loading of grpc deps to override that definition, but wouldn't it still try to load the same `upb` commit at this step https://github.com/bazelbuild/bazel/blob/master/WORKSPACE#L685. And if we do not have the same commit/version, the build would fail? 

I think the better fix would be updating the `upb` deps in grpc to use the new `upb` version -> their new release -> updating to the new release on our side. This way would take longer and might block us from flipping `incompatible_use_platforms_repo_for_constraints` for 6.0 so I rather take this route and file a ticket to `grpc` to let them upgrade their deps to a newer version. What do you think?

meteorcloudy:
> I think that might not work. We can have upb in the WORKSPACE file preceding the loading of grpc deps to override that definition, but wouldn't it still try to load the same upb commit at this step https://github.com/bazelbuild/bazel/blob/master/WORKSPACE#L685. And if we do not have the same commit/version, the build would fail?

I believe the build won't fail, the first definition would win and the one in grpc_deps will be ignored. I had a comment here to explain which repo definition is respected when a repo is defined multiple times: https://github.com/bazelbuild/examples/blob/12e1ab69e6b239f63908113cd6936a6de77f4477/bzlmod/05-integrate_third_party_package_manager/WORKSPACE#L39-L44

aiuto:
You should definitely try updating distdir_deps.bzl first.

In general, it seems we have made this patching style needlessly difficult for ourselves by putting them in third_party.

What if we started a `patches` folder at the top level. Then the patch sections of distdir_deps (or other places) could include them directly. The PRs could all be normal rather than through the third_party magic.
That would also reflect ownership more precisely, since the patches we create are authored by the Bazel team.

aranguyen:
Thank you both for the discussions. This pr is now obsolete as I was able to upgrade disdir_deps via an internal cl. 

@meteorcloudy to @aiuto 's point above, was there any other reasons why we have the patches in third_party? If there isn't, I can make a task to follow up in my own time to make the process of patching a bit easier.

aranguyen:
This is now obsolete. Closing

PR #15992, opened 2022-07-27
Use upb patch

aranguyen:
This should be imported after https://github.com/bazelbuild/bazel/pull/15991 is merged. The import should only include files that are not from third_party.


aranguyen:
This is obsolete. Closing

PR #16000, opened 2022-07-29
Draft code and test for http module extension

SalmaSamy:


src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -364,6 +364,38 @@ def testRepositoryRuleErrorInModuleExtensionFailsTheBuild(self):
             in line for line in stderr
         ]))
 
+  def testHttpfileModuleExtensionSuccess(self):
+      self.ScratchFile('MODULE.bazel', [
+          'file_ext = use_extension("//tools/build_defs/repo:http.bzl", "http_file_ext")',
```
fmeum, line 369: If the extension were just called `http`, this would make it easy to add an `archive` tag class later.

tools/build_defs/repo/http.bzl
```diff
@@ -523,3 +527,41 @@ Examples:
...
+
+#///////////////////////////// -- Module Extensions -- ///////////////////////////////#
+
+def _http_file_ext_impl(module_ctx):
+    """Implementation of the http_file module extension."""
+    files = []
+    for mod in module_ctx.modules:
+        for file in mod.tags.file:
+            # Will use the first file that appears in the tree
+            if file.name in files:
```
fmeum, line 539: I would be worried about some transitive dependency deep down the tree breaking the main module due to a name collision. Maybe non-root modules could be required to prefix the name with their module name? That would make collisions mich less likely while still allowing the root module to use essentially any name.

src/test/py/bazel/bzlmod/bazel_module_test.py
SalmaSamy: Good point

tools/build_defs/repo/http.bzl
Wyverald: This is a very good point, and one that I think basically kills the proposal.

The requirement to prepend the module name can work, but is rather unwieldy IMO. We could also automatically prepend the module name, but that would be a bit surprising.

My preference now is to use home-made extensions every time, which completely eradicates the name collision problem, and is actually not very verbose IMO. Compare the follow two snippets:

```bzl
# MODULE.bazel
module(name="foo")
http = use_extension("@bazel_tools//tools/build_defs/repo:http.bzl", "http")
http.archive(name="foo_archive", ...)
http.file(name="foo_file", ...)
use_repo(http, archive="foo_archive", file="foo_file")
```

```bzl
# MODULE.bazel
my_deps = use_extension("//:my_deps.bzl", "my_deps_ext")
use_repo(my_deps, "archive", "file")

# my_deps.bzl
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")

def my_deps():  # a WORKSPACE macro
  http_archive(name = "archive", ...)
  http_file(name = "file", ...)

my_deps_ext = module_extension(implementation=lambda ctx: my_deps())
```

The former one does not actually bring much more convenience in my opinion, especially as the `my_deps` macro in the latter scenario often already exists.

PR #16018, opened 2022-08-02
Warn about source directory being unsound

fmeum:
When a source artifact is a directory and source directory tracking is
not enabled, a warning is printed that instructs the user to either use
an archive instead or enable the (experimental) source directory
tracking functionality.

See #15774 for additional context.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/ArtifactFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/ArtifactFunction.java
@@ -278,6 +278,13 @@ public final class ArtifactFunction implements SkyFunction {
 
     if (fileValue.isDirectory()) {
       env.getListener().post(SourceDirectoryEvent.create(artifact.getExecPath()));
+      if (!TrackSourceDirectoriesFlag.trackSourceDirectories()) {
+        env.getListener().handle(Event.warn(String.format(
+            "input '%s' is a directory; dependency checking of source directories is unsound. "
+                + "Either use an archive instead or enable highly experimental source directory "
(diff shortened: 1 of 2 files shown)


fmeum:
@alexjski While the fate of `BAZEL_TRACK_SOURCE_DIRECTORIES` is undecided, I think that adding a warning about the incorrect incremental behavior of source directories is a necessary step. The warning message is only meant as a suggestion, feel free to propose different wording (including removing the mention of `BAZEL_TRACK_SOURCE_DIRECTORIES` entirely, if you prefer).

alexjski:
I am afraid that this is a way to pop up internally now (in fact, quite a lot; to make matters worse, in that case it is benign), so if we were to do that, we would need to have that only for Bazel. Given we were discussing enabling that feature by default for Bazel, maybe that is a better way to go?

benjaminp:
This sort of warning already exists: https://github.com/bazelbuild/bazel/blob/86bc93563102fb410b308ede345beeb4be7bdd69/src/main/java/com/google/devtools/build/lib/actions/AbstractAction.java#L512-L538
It's just only enabled for `genrule`.

fmeum:
> I am afraid that this is a way to pop up internally now (in fact, quite a lot; to make matters worse, in that case it is benign), so if we were to do that, we would need to have that only for Bazel. Given we were discussing enabling that feature by default for Bazel, maybe that is a better way to go?

I would prefer enabling source directory tracking in Bazel by default. However, I understood that you weren't sure yet whether that is the right approach or not and what the time frame for that decision would look like. 

Do you think that the path to source directory tracking in Bazel is long enough that a warning would make sense in the meantime? If so, I would hide it behind a flag. If not, I would close this PR to focus our efforts on enabling the feature.

meisterT:
@fmeum I assume this PR is obsolete?

PR #16019, opened 2022-08-02
Fix Linux to Windows cross compilation of py_binary, java_binary and sh_binary using MinGW

jesseschalken:
This branch
- Fixes various compilation and runtime errors when building the C++ Windows launcher using a MinGW toolchain on Linux
- Fixes Windows target platform detection to use the target platform instead of the host platform
- Removes the execution platform configuration from the `$launcher` attributes so that the launcher is compiled for the target platform
- Adds target platform constraints to the autogenerated Windows toolchains so that they can be selected with target platform constraints as a replacement for `--compiler=...` when using `--incompatible_enable_cc_toolchain_resolution`


src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
```diff
@@ -1243,6 +1246,24 @@ public ExecGroupCollection getExecGroups() {
     return execGroupCollection;
   }
 
+  private final static ConstraintSettingInfo osConstraintSetting;
+  private final static ConstraintValueInfo windowsConstraintValue;
+
+  static {
+    try {
+      Label osConstraintSettingLabel = Label.parseAbsolute("@platforms//os:os", ALWAYS_FALLBACK);
```
fmeum, line 1254: This will probably fail with Bzlmod as it doesn't use a repository mapping - with Bzlmod, the `platforms` module will manifest as a repository named something like `@platforms~0.0.4`.

You should be able to use:
 
```
Label.parseWithRepoContext("@platforms//os:os", repoContext.of(RepositoryName.MAIN, getRule.getPackage().getRepositoryMapping()));
```
but that won't work from a static context.

jesseschalken: Thanks. Should I just parse the labels directly in `isTargetOsWindows`?

fmeum: I unfortunately don't know enough about constraint handling in Java code to answer that question - I would leave it for the reviewer.

fmeum: Thinking about this some more, it's more involved and what I suggested above is incorrect: The rule may be instantiated in *any* repository, not necessarily the main repository. But these repositories may not depend on the `platforms` module and thus may not even see `@platforms` in their repository mapping.

@Wyverald can probably help here, but may still be OOO.

jesseschalken: @Wyverald Do you happen to know the best way to test for the `@platforms//os:windows` constraint on the target platform inside a Java rule implementation?

jesseschalken: I noticed the way to use `@platforms//os:windows` in a Starlark rule is to express a dependency on it which allows pulling out a `ConstraintValueInfo` provider in the rule implementation to pass to `target_platform_has_constraint`: https://github.com/aspect-build/bazel-lib/blob/5edb10a0d8d207ba9379322bcebb58f541dc8060/lib/private/copy_file.bzl#L116

Is the right thing to do something similar on the Java side?

Wyverald: @comius would be the best person to answer this, I think. I really don't know whether we're supposed to grab labels like this in Java code.

With that said, what @fmeum pointed out still stands -- we can't rely on the current repository having visibility into the `@platforms` repo. Depending on the answer to the previous question, we might have ways to work around this.

jesseschalken:
@sgowroji Thanks. I've merged master but I am yet to fix the Windows test failures (I'm still trying to get them to pass locally on master). Hoping someone beats me to it.

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
comius: Please move all references to Windows out of RuleContext. 

RuleContext should remain a general device, and is no place for specifics. You can either put the code into into each rule or providing a helper class under `rules` or `bazel/rules` package.

This will allso make it possible to Starlarkify the rule's your modifying. There's already available API in Starlark `ctx.target_platform_has_constraint`.

Creating a providers instead of using an implicit dependency to `@platforms//os:windows` is also not rewritable to Starlark (you can't create a ConstraintInfoProvider). Please use the example you found in copy_file.

src/main/java/com/google/devtools/build/lib/analysis/test/AnalysisTestActionBuilder.java
```diff
@@ -37,25 +37,24 @@ private AnalysisTestActionBuilder() {}
    */
   public static void writeAnalysisTestAction(
       RuleContext ruleContext, AnalysisTestResultInfo testResultInfo) {
-    // TODO(laszlocsomor): Use the execution platform, not the host platform.
-    boolean isExecutedOnWindows = OS.getCurrent() == OS.WINDOWS;
+    boolean isTargetOsWindows = ruleContext.isTargetOsWindows();
```
comius, line 40: cc @brandjon @tetromino 
No need to do anything (maybe a TODO) otherwise just a driveby comment. 

Do you know why we need special casing`AnalysisTestResultInfo` for Windows? This should also be general like `RuleContext`.

src/main/java/com/google/devtools/build/lib/util/OsUtils.java
```diff
@@ -26,7 +26,11 @@ private OsUtils() {
   }
 
   public static String executableExtension(OS os) {
-    return os == OS.WINDOWS ? ".exe" : "";
+    return executableExtension(os == OS.WINDOWS);
```
comius, line 29: Please mark this function private. We don't want to encourage people to use it.

src/main/java/com/google/devtools/build/lib/util/OsUtils.java
```diff
@@ -26,7 +26,11 @@ private OsUtils() {
...
+  }
+
+  public static String executableExtension(boolean isWindows) {
+    return isWindows ? ".exe" : "";
   }
 
   /**
```
comius, line 36: Please mark the following function `executableExtension()` @Deprecated. There are some uses left, that should be cleaned up if we want C++ platformization to fully work.

comius (changes requested):
Thanks for your changes! I'm just working on C++ plaformisation and your change probably fixes a couple of more things expected, judging from the title :)

comius:
Tests might be failing because `--platforms=windows` isn't passed (or some autodetection is failing).

src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java
jesseschalken: @comius Thanks. How do I get a `Label` for `@platforms//os:windows` to use in `Bazel{Sh,Py,Java}BinaryRule.build`? It looks like `RuleDefinitionEnvironment` can only give Labels for things in `@bazel_tools`. Will this work?

```java
.add(
    attr("_windows_constraint", LABEL)
        .value(Label.parseAbsoluteUnchecked("@platforms//os:windows")))
```

Wyverald: the above will likely not work with Bzlmod (same reason as @fmeum described earlier). `@bazel_tools` is special since it's a "built-in module" which is shipped with the Bazel binary; `@platforms` is not.

You would be able to use labels from `@platforms` if you were defining these rules in Starlark (in the builtins_bzl stuff, for example). Not sure where we are on that front -- I think `java_binary` is already Starlarkified?

comius: It think this would work, except if I'm missing some bzlmod machinery that is invoked on Labels in Starlark.

There's another problem: it will only work in Bazel, not in a monorepo. We use getToolsRepository to map it to absolute label. Can you extend it with `getPlatfromsRepository()`? This should make it possible to remap @platforms as well.

comius:
I'm closing this, becuase there's a lack of response for about a year. Feel free to reopen it, if you want to move it along.

PR #16022, opened 2022-08-02
Correctly report errors thrown by CommandLinePathFactory#create.

tjgq:
This method is called while initializing the remote module [1]. Any exceptions
not derived from java.io.IOException cause a silent server crash.

[1] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java;l=436;drc=f3211f00ae08746b5794ab01d404c32b43146aba

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
+++ b/src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.assertThrows;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.devtools.build.lib.runtime.CommandLinePathFactory.CommandLinePathFactoryException;
 import com.google.devtools.build.lib.vfs.DigestHashFunction;
 import com.google.devtools.build.lib.vfs.FileSystem;
 import com.google.devtools.build.lib.vfs.Path;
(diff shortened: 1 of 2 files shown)


tjgq:
@bazel-io fork 5.3.0

tjgq:
@Yannic

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
```diff
@@ -99,6 +100,9 @@ public void createWithNamedRoot() throws Exception {
         .isEqualTo(filesystem.getPath("/path/to/output/base/foo"));
     assertThat(factory.create(ImmutableMap.of(), "%output_base%/foo/bar"))
         .isEqualTo(filesystem.getPath("/path/to/output/base/foo/bar"));
+
+    assertThat(factory.create(ImmutableMap.of(), "%workspace%//foo//bar"))
```
Yannic, line 104: Why allow this?

Yannic:
@bazel-io flag

src/test/java/com/google/devtools/build/lib/runtime/CommandLinePathFactoryTest.java
tjgq: It seems gratuitous to disallow double-slashes when most other command-line utilities handle them correctly; in addition, we get it for free through `PathFragment.create`. We could even avoid the edge case by converting the whole string to a `PathFragment` before regex-matching the first segment, but I wanted to do the bare minimum to avoid complicating the cherry pick.

Yannic: Fair enough

PR #16039, opened 2022-08-03
Refactor combined cache.

### LINKED ISSUE #15934: --noremote_accept_cached cannot be used with remote downloader as of Bazel 5.0
### Description of the bug:

We use the [Remote Downloader](https://github.com/bazelbuild/proposals/blob/main/designs/2020-01-14-remote-downloads.md) (the `--experimental_remote_downloader` flag) to enforce security invariants in our builds and reduce dependencies on external hosts.

We _also_ occasionally run builds with `--noremote_accept_cached`.

This flag is documented as:

> Whether to accept remotely cached action results.

This means we still expect to be able to use the CAS (which underlies the remote downloader service), but we do _not_ want actions themselves to be cached.

As of Bazel 5.0 (specifically, #13769, from the `git bisect` I did), this fails with an error that Bazel retrieved an _empty file_ (the e3b0c... hash), rather than the one specified in the WORKSPACE file.

```
Error in download_and_extract: com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException: 
Checksum was e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
but wanted bc81f1ba47ef5cc68ad32225c3d0e70b8c6f6077663835438da8d5733f917598
```

https://github.com/clint-stripe/remote-downloader-cache

We're using a different implementation of the remote cache and downloader internally, but this repo shows the same issue occurs with the open source implementation at https://github.com/buchgr/bazel-remote.

(This repro doesn't have Bazel run with `--incompatible_remote_results_ignore_disk`, which was added in that CL, but it does not change the behavior here either.)

darwin, linux

release 5.0.0

_No response_

_No response_

_No response_

_No response_

coeuvre:
Instead of guessing when to use remote/disk part, combined cache now uses read/write cache policy provided by the context. The call sites can change the policy based on the requirements.

Fixes #15934. But unfortunately, I am not able to add an integration test for it because our own remote worker doesn't support the asset API.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
@@ -30,7 +30,7 @@ import com.google.devtools.build.lib.buildeventstream.PathConverter;
 import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
 import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext;
-import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext.Step;
+import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext.CachePolicy;
 import com.google.devtools.build.lib.remote.util.DigestUtil;
 import com.google.devtools.build.lib.remote.util.TracingMetadataUtils;
 import com.google.devtools.build.lib.vfs.Path;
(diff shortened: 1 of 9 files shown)


brentleyjones:
@bazel-io flag

kshyanashree:
Hi @coeuvre, we are trying to [cherry-pick] (https://github.com/bazelbuild/bazel/pull/16073) this change to the release 5.3.0. But pre-submits checks are failing due to missing dependencies. So could you please help us  in cherry-picking this change with appropriate commits? Thanks!

coeuvre:
cherry pick is ready https://github.com/bazelbuild/bazel/pull/16110.

Allenqel:
Allenel3

PR #16052, opened 2022-08-05
repository_ctx: Allow renaming archive entries during extraction.

jmillikin:
Adds a `rename_files=` parameter to the `extract()` and `download_and_extract()` methods of `repository_ctx`. This new parameter takes a dict of archive entries to rename, and is applied prior to any prefix adjustment.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorDescriptor.java
+++ b/src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorDescriptor.java
@@ -19,6 +19,7 @@ import com.google.devtools.build.lib.bazel.repository.DecompressorValue.Decompre
 import com.google.devtools.build.lib.rules.repository.RepositoryFunction.RepositoryFunctionException;
 import com.google.devtools.build.lib.vfs.Path;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.util.Map;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
@@ -33,17 +34,21 @@ public class DecompressorDescriptor {
   private final Path repositoryPath;
   private final Optional<String> prefix;
   private final boolean executable;
+  private final Map<String, String> renameFiles;
   private final Decompressor decompressor;
 
   private DecompressorDescriptor(
       String targetKind, String targetName, Path archivePath, Path repositoryPath,
-      @Nullable String prefix, boolean executable, Decompressor decompressor) {
+      @Nullable String prefix, boolean executable,
+      @Nullable Map<String, String> renameFiles,
+      Decompressor decompressor) {
     this.targetKind = targetKind;
     this.targetName = targetName;
     this.archivePath = archivePath;
     this.repositoryPath = repositoryPath;
     this.prefix = Optional.fromNullable(prefix);
     this.executable = executable;
+    this.renameFiles = renameFiles;
     this.decompressor = decompressor;
   }
 
@@ -71,6 +76,10 @@ public class DecompressorDescriptor {
     return executable;
   }
 
+  public Map<String, String> renameFiles() {
+    return renameFiles;
+  }
+
   public Decompressor getDecompressor() {
     return decompressor;
   }
@@ -92,12 +101,13 @@ public class DecompressorDescriptor {
         && Objects.equals(repositoryPath, descriptor.repositoryPath)
         && Objects.equals(prefix, descriptor.prefix)
         && Objects.equals(executable, descriptor.executable)
+        && Objects.equals(renameFiles, descriptor.renameFiles)
         && decompressor == descriptor.decompressor;
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(targetKind, targetName, archivePath, repositoryPath, prefix);
+    return Objects.hash(targetKind, targetName, archivePath, repositoryPath, prefix, renameFiles);
   }
 
   public static Builder builder() {
@@ -115,6 +125,7 @@ public class DecompressorDescriptor {
     private Path repositoryPath;
     private String prefix;
     private boolean executable;
+    private Map<String, String> renameFiles;
     private Decompressor decompressor;
 
     private Builder() {
@@ -125,7 +136,7 @@ public class DecompressorDescriptor {
         decompressor = DecompressorValue.getDecompressor(archivePath);
       }
       return new DecompressorDescriptor(
-          targetKind, targetName, archivePath, repositoryPath, prefix, executable, decompressor);
+          targetKind, targetName, archivePath, repositoryPath, prefix, executable, renameFiles, decompressor);
     }
 
     @CanIgnoreReturnValue
@@ -161,6 +172,12 @@ public class DecompressorDescriptor {
     @CanIgnoreReturnValue
     public Builder setExecutable(boolean executable) {
       this.executable = executable;
+      return this;
+    }
+
+    @CanIgnoreReturnValue
+    public Builder setRenameFiles(Map<String, String> renameFiles) {
+      this.renameFiles = renameFiles;
       return this;
     }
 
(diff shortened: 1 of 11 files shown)


jmillikin:
Use cases include:
* Archives where multiple files have the same name in different case. On platforms with case-insensitive filesystems (macOS, Windows) these archives can't be extracted as-is, and the files can't be fixed up after extraction.
  * One such project is the Linux kernel, which has files named like `net/netfilter/xt_dscp.c` and `net/netfilter/xt_DSCP.c` (among others).
* Archives containing non-UTF8 filenames fail to extract on platforms or filesystems that enforce Unicode paths.
  * See https://github.com/bazelbuild/rules_go/issues/2771 (Go toolchains from golang.org fail to extract)

A longer example of extracting the Linux kernel on macOS (which is case-insensitive by default):
```python
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

def _linux_repository(ctx):
    ctx.file("WORKSPACE", "workspace(name = {name})\n".format(
        name = repr(ctx.name),
    ))
    ctx.file("BUILD.bazel", """
genrule(
    name = "nf_srcs",
    srcs = glob(["net/netfilter/*"]),
    outs = ["nf_srcs.txt"],
    cmd = "ls -1 external/linux/net/netfilter/ > $@",
)
""")

    ctx.download_and_extract(
        url = ["https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.19.tar.xz"],
        sha256 = "ff240c579b9ee1affc318917de07394fc1c3bb49dac25ec1287370c2e15005a8",
        stripPrefix = "linux-5.19",
        renameFiles = _rename_files("5.19"),
    )

def _rename_files(version):
    netfilter_srcs = {
        "xt_DSCP.c": "xt_DSCP_target.c",
        "xt_HL.c": "xt_HL_target.c",
        "xt_RATEEST.c": "xt_RATEEST_target.c",
        "xt_TCPMSS.c": "xt_TCPMSS_target.c",
    }
    renames = {}
    for orig_name, new_name in netfilter_srcs.items():
        prefix = "linux-{}/net/netfilter/".format(version)
        renames[prefix + orig_name] = prefix + new_name
    # and similar logic for netfilter headers under include/uapi/linux/
    return renames

linux_repository = repository_rule(
    implementation = _linux_repository,
)
```

```
$ bazel build @linux//:nf_srcs
[...]
Target @linux//:nf_srcs up-to-date:
  bazel-bin/external/linux/nf_srcs.txt
$ cat bazel-bin/external/linux/nf_srcs.txt | grep -i dscp
xt_DSCP_target.c
xt_dscp.c
```

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -749,8 +749,22 @@ public StructImpl download(
                     + " archive. Instead of needing to specify this prefix over and over in the"
                     + " <code>build_file</code>, this field can be used to strip it from extracted"
                     + " files."),
+        @Param(
+            name = "renameFiles",
```
fmeum, line 753: This is the first time I actually noticed that `stripPrefix` has a camel case name - that looks like a bug as Starlark parameters usually use snake case. Using snake case for `renameFiles` at least wouldn't complicate future efforts to clean up this inconsistency.

jmillikin: Done.

jmillikin:
Gentle ping

src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorDescriptor.java
```diff
@@ -33,17 +34,21 @@ public class DecompressorDescriptor {
...
   private final Optional<String> prefix;
   private final boolean executable;
+  private final Map<String, String> renameFiles;
   private final Decompressor decompressor;
 
   private DecompressorDescriptor(
       String targetKind, String targetName, Path archivePath, Path repositoryPath,
-      @Nullable String prefix, boolean executable, Decompressor decompressor) {
+      @Nullable String prefix, boolean executable,
+      @Nullable Map<String, String> renameFiles,
```
Wyverald, line 43: this doesn't need to be nullable (it's never null anyway, and null has the same semantics as empty). also remove all the null checks

src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorDescriptor.java
```diff
@@ -164,6 +175,12 @@ public Builder setExecutable(boolean executable) {
       return this;
     }
 
+    @CanIgnoreReturnValue
+    public Builder setRenameFiles(Map<String, String> renameFiles) {
+      this.renameFiles = renameFiles;
```
Wyverald, line 180: defensively copy here (ImmutableMap.copyOf)

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -749,8 +749,22 @@ public StructImpl download(
...
                     + " files."),
+        @Param(
+            name = "rename_files",
+            defaultValue = "{}",
+            named = true,
+            positional = false,
+            doc =
+                "An optional dict specifying files to rename during the extraction. Archive entries"
+                    + " with names exactly matching a key will be renamed to the value, prior to"
+                    + " any directory prefix adjustment."),
```
Wyverald, line 760: could you add more information on how this parameter can be used, like you did in the PR comments? or maybe even link to this PR

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
```diff
@@ -881,6 +900,15 @@ public void extract(Object archive, Object output, String stripPrefix, StarlarkT
...
                     + " easier but should be set before shipping."),
+        @Param(
+            name = "rename_files",
+            defaultValue = "{}",
+            named = true,
+            positional = false,
+            doc =
+                "An optional dict specifying files to rename during the extraction. Archive entries"
+                    + " with names exactly matching a key will be renamed to the value, prior to"
+                    + " any directory prefix adjustment."),
```
Wyverald, line 911: same here

src/test/java/com/google/devtools/build/lib/bazel/repository/CompressedTarFunctionTest.java
```diff
@@ -64,6 +66,26 @@ public void testDecompressWithPrefix() throws Exception {
     archiveDescriptor.assertOutputFiles(outputDir, INNER_FOLDER_NAME);
   }
 
+  /**
+   * Test decompressing a tar.gz file, with some entries being renamed during the
+   * extraction process.
+   */
+  @Test
+  public void testDecompressWithRenamedFiles() throws Exception {
```
Wyverald, line 74: can we add a test case where strip_prefix is also present, to test that rename_files happens before strip_prefix?

Wyverald (changes requested):
thanks!

src/main/java/com/google/devtools/build/lib/bazel/repository/DecompressorDescriptor.java
jmillikin: Done.

src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkRepositoryContext.java
jmillikin: Done.

src/test/java/com/google/devtools/build/lib/bazel/repository/CompressedTarFunctionTest.java
jmillikin: Done.

Wyverald (approved):
thanks!

PR #16058, opened 2022-08-06
Allow remote retry max delay to be user configurable

joeljeske:
This introduces a new option `--remote_retry_max_delay` can be used to change the existing maximum exponential backoff interval used when retrying remote requests. Before this change, there was a hardcoded value controlling this maximum exponential backoff interval, set to `5s`. 

Rational
`remote_retries` is useful in masking over temporary disruptions to a remote cluster. If a cluster experiences temporary downtime, it is useful to allow bazel clients to wait for a period of time for the cluster to recover before bailing and giving up. If users cannot configure the maximum exponential backoff delay, one must set a large number for `remote_retries`, each retry eventually waiting for up to 5s. This allows the bazel client to wait for a reasonable amount of time for the cluster to recover. 

The problem here is that under certain cluster failure modes, requests may not be handled and failed quickly, rather they may wait until `remote_timeout` before failing. A large `remote_timeout` combined with a large `remote_retries` could lead to waiting for a very long time before finally bailing on a given action. 

If a user can bump the `remote_retry_max_delay`, they can control the retry waiting semantics to their own needs.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteRetrier.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteRetrier.java
@@ -158,7 +158,7 @@ public class RemoteRetrier extends Retrier {
       Preconditions.checkArgument(jitter >= 0 && jitter <= 1, "jitter must be in the range (0, 1)");
       Preconditions.checkArgument(maxAttempts >= 0, "maxAttempts must be >= 0");
       nextDelayMillis = initial.toMillis();
-      maxMillis = max.toMillis();
+      maxMillis = Math.max(max.toMillis(), nextDelayMillis);
       this.multiplier = multiplier;
       this.jitter = jitter;
       this.maxAttempts = maxAttempts;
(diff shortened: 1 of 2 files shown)


joeljeske:
@tjgq could I trouble you for a quick review of this configuration option? 🙇 🙏

src/main/java/com/google/devtools/build/lib/remote/RemoteRetrier.java
```diff
@@ -154,6 +154,7 @@ public static class ExponentialBackoff implements Backoff {
      */
     ExponentialBackoff(Duration initial, Duration max, double multiplier, double jitter,
         int maxAttempts) {
+      Preconditions.checkArgument(max.compareTo(initial) > 0, "max must be > initial");
```
tjgq, line 157: As a rule, Bazel should not crash due to malformed input. To avoid additional validation logic, how about we use `max(initial, remoteRetryMaxDelay)` as the argument to `ExponentialBackoff` below?

tjgq (changes requested):
Sorry about the delay. This PR seems reasonable to me, I have just a small comment.

joeljeske:
Thanks for checking this out, @tjgq! I made your suggested change, preventing Bazel from crashing on an invalid value here.

brentleyjones:
@bazel-io flag

keertk:
@bazel-io fork 6.2.0

PR #16059, opened 2022-08-07
Use env trick for these test scripts.

3405691582:
Attempting to run tests on projects will fail when bash is not in /bin,
so use the env trick again here.


3405691582:
Ping.

meisterT (approved):

PR #16064, opened 2022-08-07
Add option to git_repository to provide build file / content

limdor:
This is just a first draft from the discussion we had with @aiuto in https://groups.google.com/g/bazel-dev/c/sfeRE3Ucj7E just to see what the CI says about it.
Before this can be proposed to be merged, I see several topics to tackle:

- When new_git_repository rule should be removed?
- Can it be deprecated in a way that is not just text in the documentation?
- Should this PR already replace the internal references to new_git_repository to git_repository or should be done when removing the new_git_repository?

Probably we should also create an issue to track it.

### DIFF
--- a/tools/build_defs/repo/git.bzl
+++ b/tools/build_defs/repo/git.bzl
@@ -135,9 +135,6 @@ _common_attrs = {
               "applied. If this attribute is not set, patch_cmds will be executed on Windows, " +
               "which requires Bash binary to exist.",
     ),
-}
-
-_new_git_repository_attrs = dict(_common_attrs.items() + {
     "build_file": attr.label(
         allow_single_file = True,
         doc =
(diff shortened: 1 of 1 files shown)


tools/build_defs/repo/git.bzl
```diff
@@ -165,34 +162,29 @@ _new_git_repository_attrs = dict(_common_attrs.items() + {
             "Either `workspace_file` or `workspace_file_content` can be " +
             "specified, or neither, but not both.",
     ),
-}.items())
-
-def _new_git_repository_implementation(ctx):
```
aiuto, line 170: We should just get rid of this so there is no documenetntation about new_git_repository.
An assignment at the end can do it.  

I'll make that happen in the merge.

aiuto (approved):

aiuto:
I'll merge this.

PR #16065, opened 2022-08-07
Buildfix Netty filter on this select default case.

3405691582:
The `filter_netty_dynamic_libs` genrule has no srcs when not Darwin,
Linux, or Windows. In this case, attempting to invoke the genrule breaks
the build because the `cp` command has no input. Instead, only depend on
this genrule output when the platform is Darwin, Linux, or Windows.

There might be a nicer way of doing this, but this works.

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -668,8 +668,14 @@ distrib_java_import(
     enable_distributions = ["debian"],
     jars = [
         "netty_tcnative/netty-tcnative-classes-2.0.51.Final.jar",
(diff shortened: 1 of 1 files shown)


3405691582:
Ping.

3405691582:
Ping.

3405691582:
Ping.

meteorcloudy (approved):

sgowroji:
Merged at [b7fc152](https://github.com/bazelbuild/bazel/commit/b7fc15234e9e8c97e08fa1431ad5424d66516d28)

PR #16067, opened 2022-08-08
Implement worker support for `GenerateDataBindingBaseClasses`

arunkumar9t2:
- Moved `GenerateDataBindingBaseClasses` to `ResourceProcessorBusyBox` since Android databinding classes were already a dependency there. 
- `ProcessDatabinding` is already part of `ResourceProcessorBusyBox` but it was not enabled in `persistent_android_resource_processor` expansion flag. Added that as well.

This change is multiplex compatible and probably could use expansion flags as done here https://github.com/bazelbuild/bazel/pull/15950. 

Results:
In a synthetic generated project with 30 modules here https://github.com/arunkumar9t2/bazel-playground, the worker can be turned off/on with `.bazel.sh build //android/databinding_workers:android_data_binding --strategy=GenerateDataBindingBaseClasses=[sandboxed|worker]` and worker is about 9 to 10% faster.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/AndroidConfiguration.java
@@ -861,6 +861,8 @@ public class AndroidConfiguration extends Fragment implements AndroidConfigurati
           "--strategy=AndroidManifestMerger=worker",
           "--strategy=Aapt2Optimize=worker",
           "--strategy=AARGenerator=worker",
+          "--strategy=ProcessDatabinding=worker",
+          "--strategy=GenerateDataBindingBaseClasses=worker"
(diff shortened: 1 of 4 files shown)


ahumesky:
> We probably want this to be ported over to `rules_android` on the `pre_alpha` branch

this is the corresponding code in the starlark rules:
https://github.com/bazelbuild/rules_android/blob/69a3fd5ea29d0283c948222ae95546a2437a04cf/rules/data_binding.bzl#L57-L97

arunsampathkumar-grabtaxi:
Thanks @ahumesky @nkoroste I will port this to rules_android as well

arunkumar9t2:
Thanks @ahumesky @nkoroste I will port this to rules_android as well

PR #16071, opened 2022-08-09
Include exception message for "Source forest creation failed" error

kmicklas:
We experienced this error from integration tests which failed to clean up after themselves and left write-protected files behind. This is just a small messaging improvement which would have made finding the root cause easier.

The change here is analogous to this error: https://github.com/bazelbuild/bazel/blob/d1820a75ea3eb182ef2c2319bb77be19c4113a47/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java#L875


haxorz (approved):

haxorz:
Very sorry about the delay, @kmicklas . Thanks for your contribution to Bazel!

PR #16072, opened 2022-08-09
Verify gRPC cache downloads after flushing the output.

benjaminp:
Make sure any internal buffers (especially compression) are flushed before attempting verify the digest of the downloaded file.


coeuvre (approved):

brentleyjones:
Does this fix any known bug? Is it worth cherry-picking into 5.3?

coeuvre:
I am not aware of any bugs this fixes. but should be easy to cherry-pick into 5.3 if time allows.

brentleyjones:
@bazel-io flag

meteorcloudy:
This should be in 6.0, removing the "potential release blocker" label.

PR #16077, opened 2022-08-10
Add RunfilesLibraryInfo provider

fmeum:
The new Starlark-defined RunfilesLibraryInfo provider may be used in the
future to decide whether additional information has to be emitted by a
rule for a target to find its runfiles at runtime. This may include
serializing repository mapping information to a runfile and/or
generating code that provides the canonical name of the current
repository.

The provider is added now rather than later so that third-party runfiles
libraries can rely on it being present earlier - they have to wait for
their minimum supported Bazel version to offer it. If it should turn out
not to be needed, it can safely be removed before any ruleset would have
a reason to depend on it.

Work towards https://github.com/bazelbuild/proposals/pull/269

### DIFF
--- a/tools/BUILD
+++ b/tools/BUILD
@@ -87,6 +87,7 @@ filegroup(
         "//tools/jdk:bzl_srcs",
         "//tools/osx:bzl_srcs",
         "//tools/python:bzl_srcs",
+        "//tools/runfiles:bzl_srcs",
         "//tools/sh:bzl_srcs",
         "//tools/test:bzl_srcs",
         "//tools/windows:bzl_srcs",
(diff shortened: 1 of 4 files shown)


fmeum:
@bazel-io flag

fmeum:
@oquenchil I just realized that it might be problematic for builtins to access a provider defined in `@bazel_tools`. Do you see a way to add a Starlark-defined provider to Bazel that can be loaded by third-party rulesets but still be usable from builtin rules such as `cc_library`?

kshyanashree:
Hi @fmeum! This PR is in draft state, Can you please provide an update on this since we are very close to our release date. Thanks!

fmeum:
@kshyanashree I am stuck and looking for feedback from @oquenchil. If I don't get this done in time, please just ignore this PR.

PR #16080, opened 2022-08-10
Show a better error when MODULE.bazel is missing

fmeum:
Previously, Bazel would error out with this message if it couldn't find
a MODULE.bazel (e.g. in the root module):

```
ERROR: unknown error during computation of main repo mapping
```


Wyverald (approved):
thanks!

PR #16086, opened 2022-08-10
Include stack trace in all gRPC errors when --verbose_failures is set.

tjgq:
Also refactor a couple places where the stack trace was included in an ad-hoc
manner, and force Utils.grpcAwareErrorMessage callers to be explicit to avoid
future instances.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -507,7 +507,8 @@ public final class RemoteModule extends BlazeModule {
       }
     } catch (IOException e) {
       String errorMessage =
-          "Failed to query remote execution capabilities: " + Utils.grpcAwareErrorMessage(e);
+          "Failed to query remote execution capabilities: "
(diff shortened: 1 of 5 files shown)


coeuvre:
what's the status of this PR?

tjgq:
> what's the status of this PR?

Let's do something slightly different here: we already had logic to include the stack trace when `--verbose_failures` is set, but we weren't consistently using it.

coeuvre (approved):
LGTM. Please fix the tests.

werkt:
I'm requesting incorporation in the 6.3.0 release of this commit in https://github.com/bazelbuild/bazel/pull/18959, as it is required for interop with other fixes made under build/lib/remote.

iancha1992:
The changes in this PR have been included in [Bazel 6.4.0 RC1](https://releases.bazel.build/6.4.0/rc1/index.html). Please test out the release candidate and report any issues as soon as possible. If you're using Bazelisk, you can point to the latest RC by setting USE_BAZEL_VERSION=last_rc.
Thanks!

Issue #8622, opened 2019-06-13
incompatible_use_platforms_repo_for_constraints: Don't use constraints from @bazel_tools, use @platforms instead


hlopko:
Available since: 0.28
Tracking issue: https://github.com/bazelbuild/bazel/issues/6516

### Motivation

Bazel currently provides common constraints for [platforms](https://docs.bazel.build/versions/master/platforms.html) and [toolchains](https://docs.bazel.build/versions/master/toolchains.html) in `@bazel_tools//platforms`. We are migrating these out of the Bazel binary to a principled, standalone repository over at https://github.com/bazelbuild/platforms which can be released independently from the Bazel binary and which defines a process for adding more constraints.

### Migration

Ideally, declare an explicit dependency on https://github.com/bazelbuild/platforms, name the repository as `@platforms`, and use constraints from this repository. In cases where you cannot depend on https://github.com/bazelbuild/platforms (please tell us the reason in the comment), you can use the snapshot of https://github.com/bazelbuild/platforms in Bazel - Bazel implicitly provides this repository for Bazel's needs.

The actual migration in BUILD files is simple - use `@platforms//setting:value` instead of `@bazel_tools//platforms:value`:

```
sed 's$@bazel_tools//platforms:(linux|osx|windows|android|freebsd|ios|os)$@platforms//os:\1$' -E -i **/*
sed 's$@bazel_tools//platforms:(cpu|x86_32|x86_64|ppc|arm|aarch64|s390x)$@platforms//cpu:\1$' -i -E **/*
```

dslomov:
What is the status of this for 1.0?

hlopko:
Not flipping in time for 1.0

ittaiz:
maybe migration can be added to https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md ?

katre:
No one is currently working on this (it's, unfortunately, too low-priority and isn't currently hurting anything).

Once someone starts addressing the migration updating buildtools definitely makes sense.

ittaiz:
👍 @philwo its a bit annoying that the bazel flag bot adds a non actionable issue.
I like the concept a lot but I think the implementation needs to change

alexeagle:
This fails in rules_nodejs with
```
[...]

PR #16093, opened 2022-08-12
Keep usePlatformsRepoForConstraints value in host cfg

AlessandroPatti:
The value gets reset to the default (true) so setting `--noincompatible_use_platforms_repo_for_constraints` has no effect if the target using the constraint is build for the host config.

Related issue: https://github.com/bazelbuild/bazel/issues/8622


gregestren:
@aranguyen could you review this? I think this makes sense.

aranguyen:
@AlessandroPatti is this pr still in progress? There seems to be some failures in the presubmit

AlessandroPatti:
@aranguyen No, it is ready for review. The failure was just a flake, a new run passed.

aranguyen (approved):

Issue #14937, opened 2022-03-02
targeting  java language level 17 or using a java_runtime_level 17 causes build failures on arm64 linux


chancila:
### Description of the problem / feature request:

seems that there's no jdk for linux arm64 for java 17...I didn't find any open or closed issues about this so opening this one
```
_JDKS = [
    "remotejdk11_macos",
    "remotejdk11_macos_aarch64",
    "remotejdk11_win",
    "remotejdk11_win_arm64",
    "remotejdk11_linux_aarch64",
    "remotejdk11_linux",
    "remotejdk11_linux_ppc64le",
    "remotejdk11_linux_s390x",
    "remotejdk17_macos",
    "remotejdk17_macos_aarch64",
    "remotejdk17_win",
    "remotejdk17_win_arm64",
    "remotejdk17_linux",
]
```

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

see below for examples...

### What operating system are you running Bazel on?

can fail on any system if targeting linux arm64, but will always fail on an arm64 linux host not cross-compiling for jdk 17 or using the java 17 runtime.

### What's the output of `bazel info release`?

5.0

###  Have you found anything relevant by searching the web?

nothing in the issue tracker

### Any other information, logs, or outputs that you want to share?

forcing linux x86_64 works fine:

```
bzl build --host_platform=//:linux_x86_64 --java_language_version=17 @bazel_tools//tools/jdk:remotejdk_17
INFO: Invocation ID: c77977bd-cdfd-41fe-880c-ae2bdd1eb9e0
INFO: Analyzed target @bazel_tools//tools/jdk:remotejdk_17 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target @bazel_tools//tools/jdk:remotejdk_17 up-to-date (nothing to build)
INFO: Elapsed time: 0.191s (0.00m), Critical Path: 0.00s (0.00m)
INFO: 1 process: 1 internal.
INFO: Build Event Protocol files produced successfully.
INFO: Build completed successfully, 1 total action
```

forcing linux arm64 does not:

```
bzl build --host_platform=//:linux_arm64 --java_language_version=17 @bazel_tools//tools/jdk:remotejdk_17
INFO: Invocation ID: b7b1bedc-6e5e-4776-930f-f40c0f75118e
INFO: Build options --host_platform and --platforms have changed, discarding analysis cache.
[...]

PR #16095, opened 2022-08-12
Add support for remotejdk17 for linux-aarch64 (arm64)

nresare:
This closes https://github.com/bazelbuild/bazel/issues/14937 (support java language level 17 on linux-aarch64)

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE.tmpl
+++ b/src/main/java/com/google/devtools/build/lib/bazel/rules/java/jdk.WORKSPACE.tmpl
@@ -301,6 +301,21 @@ maybe(
     version = "17",
 )
 
+maybe(
+    remote_java_repository,
+    name = "remotejdk17_linux_aarch64",
+    exec_compatible_with = [
+        "@platforms//os:linux",
+        "@platforms//cpu:aarch64",
+    ],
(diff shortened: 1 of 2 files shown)


nresare:
Obviously, it would be nice if the artefact was mirrored to mirror.bazel.build to lessen the load on the upstream distribution server :)

nresare:
My understanding of the build failures is that the automated tests are executing in an environment that doesn't have the ability to connect to cdn.azul.com, failing with the following error:
```
Error in download_and_extract: java.io.IOException: Error downloading .../zulu17.28.13-ca-jdk17.0.0-linux_aarch64.tar.gz: Unknown host: cdn.azul.com
```
Build logs are available [here](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/01829288-214e-4504-99ed-40bcc807b785/src/test/shell/bazel/workspace_resolved_test/shard_7_of_22/test_attempts/attempt_2.log)

I guess the proper fix is to mirror the artefact to mirror.bazel.build

meteorcloudy:
The failing test is marked as `block-network` https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/BUILD;l=904?q=workspace_resolved&ss=bazel, so mirroring those artifacts won't help. You should also add them in https://cs.opensource.google/bazel/bazel/+/master:distdir_deps.bzl;l=579 so that our test setup up will make sure those artifacts are cached in advance to avoid access the internet.

But yes, we still should also mirror the artifacts on mirror.bazel.build and add the URL.

PR #16098, opened 2022-08-12
Prevent eventfd and timerfd leaks when UNIX sockets are used

EdSchouten:
On Linux, looks like Bazel leaks about 300 file descriptors every time
it's invoked and has to connect to a UNIX socket. All of the leaked file
descriptors are all of type eventfd and timerfd. This change prevents
that from happening by memoizing the event loop group across all gRPC
clients.

I'm not sure how to add testing coverage for this. It looks like all the
tests for --remote_executor=unix:* are disabled right now due to Google
internal bug b/211478955.

Cc @jmillikin

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/GoogleAuthUtils.java
@@ -33,6 +33,7 @@ import io.grpc.auth.MoreCallCredentials;
 import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
+import io.netty.channel.EventLoopGroup;
 import io.netty.channel.epoll.Epoll;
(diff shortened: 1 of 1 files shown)


coeuvre:
Any thoughts why Bazel didn't release the resources? Instead of caching the event loop group, I would like to release it after build is done.

EdSchouten:
That is a very good question. I guess that `.eventLoopGroup()` doesn't take ownership of the event loop group that's provided. It only specifies the one to use.

If it's desirable that the event loop group is freed, what would be the cleanest way to do that? Do we somehow return it separately to the caller and make that one responsible for releasing it?

EdSchouten:
Friendly ping!

coeuvre:
Sorry for the late replay, it somehow got lost in my inbox.

I just checked the API:

> The channel won't take ownership of the given EventLoopGroup. It's caller's responsibility to shut it down when it's desired.

I don't have a "clean" way to manually release the event loop group. I think the workaround in this PR is good for now and I'm going to import it.

coeuvre (approved):

Issue #16094, opened 2022-08-12
Crash on using TEST_NAME as an environment variable name in rules


varungandhi-src:
### Description of the bug:

Passing an environment variable with the name TEST_NAME causes Bazel to crash. Renaming it to (say) MY_TEST_NAME fixes the issue.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Repo with minimal example: https://github.com/varungandhi-src/bazel-crash/tree/042d1747b4e76c3e9e089cd72222bd3b5a36fdf5

Test with `bazel test //...`.

The main problematic part seems to be this is this in `x.bzl`:

```
    native.sh_test(
      name = name,
      srcs = ["hello.sh"],
      deps = [],
      env = {
        "TEST_NAME": "ohno", # Using TEST_NAME creates a collision
      },
    )
```

### Which operating system are you running Bazel on?

macOS 12.5

### What is the output of `bazel info release`?

release 5.2.0-homebrew

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
❯ git remote get-url origin; git rev-parse main; git rev-parse HEAD
https://github.com/varungandhi-src/bazel-crash.git
042d1747b4e76c3e9e089cd72222bd3b5a36fdf5
042d1747b4e76c3e9e089cd72222bd3b5a36fdf5
```

### Have you found anything relevant by searching the web?

There is a similar error in https://github.com/bazelbuild/bazel/issues/14681

```
Caused by: java.lang.IllegalArgumentException: Multiple entries with same key: @bazel_tools//tools/cpp:empty=[] and @bazel_tools//tools/cpp:empty=[]
```

But the cause seems to be different.

illicitonion:
Currently bazel hard-codes setting the `TEST_NAME` env var itself. It's unclear whether whether this should be overridable - on the one hand, `TEST_NAME` isn't specified at https://bazel.build/reference/test-encyclopedia so it feels like it should be overrideable. On the other hand, other tooling may be surprised by this override.

At the very least, though, we shouldn't _crash_. Either we should allow overriding (and have it work), or forbid it (with documentation and a clear error message).

varungandhi-src:
Not crashing and emitting a useful error message would be a reasonable solution IMO. 👍🏽

PR #16102, opened 2022-08-15
Report error instead of crash if test uses reserved env variable

blindpirate:
This closes https://github.com/bazelbuild/bazel/issues/16094

Previously Bazel will crash if test uses reserved env variable
(like `TEST_NAME`). Let's report an error in this case.


meisterT:
Sorry for only looking at this now, it slipped through the cracks - do you want to rebase this so it can be merged?

meisterT (approved):

PR #16123, opened 2022-08-18
Save injected metadata in RemoteActionFileSystem

coeuvre:
So that spawn outputs can be accessed among Spwans within the same action using the `FileSystem` API.

This allow us to revert the hack we introduced in #12590. Also fixes the issue described by #15711.

Closes #15711.

### DIFF
--- a/src/test/shell/bazel/remote/build_without_the_bytes_test.sh
+++ b/src/test/shell/bazel/remote/build_without_the_bytes_test.sh
@@ -556,8 +556,7 @@ sh_test(
 )
 EOF
   cat > a/foo.sh <<'EOF'
-echo foo
-echo bar > $TEST_UNDECLARED_OUTPUTS_DIR/bar.txt
+touch $TEST_UNDECLARED_OUTPUTS_DIR/bar.txt
 EOF
   chmod a+x a/foo.sh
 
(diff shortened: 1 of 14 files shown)


src/test/shell/bazel/remote/build_without_the_bytes_test.sh
```diff
@@ -545,6 +545,34 @@ EOF
...
+  mkdir -p a
+  cat > a/BUILD <<EOF
+sh_test(
+  name = "foo",
+  srcs = ["foo.sh"],
+)
+EOF
+  cat > a/foo.sh <<'EOF'
+echo foo
+echo bar > $TEST_UNDECLARED_OUTPUTS_DIR/bar.txt
```
tjgq, line 560: Nit: this can be simply `touch $TEST_UNDECLARED_OUTPUTS_DIR/foo.txt`, since we're not verifying the test stdout or the file contents anywhere?

tjgq (approved):

brentleyjones:
Should we cherry-pick this into 5.3?

src/test/shell/bazel/remote/build_without_the_bytes_test.sh
coeuvre: Done.

coeuvre:
I am fine to cherry-pick this into 5.3 but not sure whether it is still possible now to do so given that we are releasing 5.3 soon. cc @kshyanashree

brentleyjones:
I'll flag it, and it can get rejected/moved to the next release if it's too late.

brentleyjones:
@bazel-io flag

Issue #13185, opened 2021-03-09
Using --experimental_split_coverage_postprocessing causes a RuntimeException


finn-ball:
### Description of the problem:

With Bazel 4.0, using the `--experimental_split_coverage_postprocessing` flag causes errors and hard crashes. This probably shouldn't happen.

Experimenting with `bazel coverage` on the abseil-cpp project.

Using both flags without remote execution:
```
bazel coverage //absl/flags:config_test --experimental_split_coverage_postprocessing --experimental_fetch_all_coverage_outputs
```
```
ERROR: /<redacted>/abseil-cpp/absl/flags/BUILD.bazel:336:8: Testing //absl/flags:config_test failed: I/O exception during sandboxed execution:  /<redacted>/.cache/bazel/_ /<redacted>/execroot/com_google_absl/bazel-out/k8-fastbuild/testlogs/absl/flags/config_test/_coverage (File exists)
Target //absl/flags:config_test up-to-date:
  bazel-bin/absl/flags/config_test
INFO: Elapsed time: 5.745s, Critical Path: 3.71s
INFO: 153 processes: 54 internal, 98 linux-sandbox, 1 worker.
FAILED: Build did NOT complete successfully
//absl/flags:config_test                                              NO STATUS

FAILED: Build did NOT complete successfully
```
Using just one flag causes a stack-trace regardless of if you use this with remote execution or not:
```
bazel coverage //:test_hello_world --experimental_split_coverage_postprocessing 
```
```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'UnshareableActionLookupData{actionLookupKey=ConfiguredTargetKey{label=//absl/flags:config_test, config=BuildConfigurationValue.Key[2a87b5057419310409dbaa77df700ae2acbe8a53d5e2f613593e99ea39e39754]}, actionIndex=11}' (requested by nodes 'TestCompletionKey{configuredTargetKey=ConfiguredTargetKey{label=//absl/flags:config_test, config=BuildConfigurationValue.Key[2a87b5057419310409dbaa77df700ae2acbe8a53d5e2f613593e99ea39e39754]}, topLevelArtifactContext=com.google.devtools.build.lib.analysis.TopLevelArtifactContext@524aae23, exclusiveTesting=false}')
        at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:563)
[...]

PR #16140, opened 2022-08-21
Throw explicit exception if experimental_split_coverage_postprocessing is used without experimental_fetch_all_coverage_outputs

blindpirate:
This closes https://github.com/bazelbuild/bazel/issues/13185
According to https://github.com/bazelbuild/bazel/commit/e411fa74176cdc54f84d62c189d042e9fb6d5c79,
The flag --experimental_split_coverage_postprocessing depends on the flag
--experimental_fetch_all_coverage_outputs being enabled, but if the former one
is set without the latter one, Bazel throws a quite confusing NullPointerException.
Now we throw an explicit exception with proper error message.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
+++ b/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
@@ -102,14 +102,10 @@ public class StandaloneTestStrategy extends TestStrategy {
       TestRunnerAction action, ActionExecutionContext actionExecutionContext)
       throws ExecException, InterruptedException {
     if (action.getExecutionSettings().getInputManifest() == null) {
-      String errorMessage = "cannot run local tests with --nobuild_runfile_manifests";
-      throw new TestExecException(
-          errorMessage,
-          FailureDetail.newBuilder()
-              .setTestAction(
-                  TestAction.newBuilder().setCode(TestAction.Code.LOCAL_TEST_PREREQ_UNMET))
-              .setMessage(errorMessage)
-              .build());
+      throw createTestExecException(
+          TestAction.Code.LOCAL_TEST_PREREQ_UNMET,
+          "cannot run local tests with --nobuild_runfile_manifests"
+      );
     }
     Map<String, String> testEnvironment =
         createEnvironment(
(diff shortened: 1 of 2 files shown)


blindpirate:
@sgowroji The failure seems to be because CI machine is disconnected from Internet, can you give me some hints how to handle external dependencies which need to be downloaded in integration tests?

oquenchil (approved):
LGTM

oquenchil:
Adding @c-mita

src/test/shell/bazel/bazel_coverage_cc_test_gcc.sh
```diff
@@ -875,6 +875,79 @@ EOF
...
+  return "Hello " + who;
+}
+EOF
+
+  cat << EOF > hello-test.cc
+#include <gtest/gtest.h>
+#include "hello-greet.h"
+
+TEST(HelloTest, GetGreet) {
+  EXPECT_EQ(get_greet("Bazel"), "Hello Bazel");
```
c-mita, line 942: This could just be `return 0;` or something - the actual test result doesn't matter, and we don't want to try pulling in gtest.

src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java
```diff
@@ -111,13 +111,10 @@ public TestRunnerSpawn createTestRunnerSpawn(
       throws ExecException, InterruptedException {
     if (action.getExecutionSettings().getInputManifest() == null) {
       String errorMessage = "cannot run local tests with --nobuild_runfile_manifests";
```
c-mita, line 113: Can remove this now unused local variable.

src/test/shell/bazel/bazel_coverage_cc_test_gcc.sh
```diff
@@ -875,6 +875,79 @@ EOF
...
+#endif
+EOF
+
+  cat << EOF > hello-greet.cc
+#include <string>
+
+std::string get_greet(const std::string& who) {
+  return "Hello " + who;
+}
+EOF
```
c-mita, line 935: If you actually want a file under test, there's no need to make it complicated; a simple `int foo() { return 42; }` is more than sufficient here; we really don't care about the coverage reports, only that the flag combination raises an error,

c-mita (changes requested):
The test setup can be greatly simplified; it should be unnecessary to pull in the `rules_cc` stuff.

blindpirate:
Hi @sgowroji @oquenchil @c-mita now it's all green, PTAL!

sgowroji:
Thank you @blindpirate !

PR #16158, opened 2022-08-24
Add tool name for objcopy

m3rcuriel:
It is not clear to me why this constant is in [src/main/starlark/builtins_bzl/common/cc/action_names.bzl](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/cc/action_names.bzl) but not in [tools/build_defs/cc/action_names.bzl](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/cc/action_names.bzl), but it makes `get_tool_for_action` really annoying.

Furthermore, it seems that if we are willing to define `llvm-cov` for everyone we should do the same for `objcopy`.

### DIFF
--- a/tools/build_defs/cc/action_names.bzl
+++ b/tools/build_defs/cc/action_names.bzl
@@ -94,6 +94,9 @@ OBJC_FULLY_LINK_ACTION_NAME = "objc-fully-link"
 # A string constant for the clif actions.
 CLIF_MATCH_ACTION_NAME = "clif-match"
 
+# A string constant for the obj copy actions.
+OBJ_COPY_ACTION_NAME = "objcopy_embed_data"
+
 ACTION_NAMES = struct(
     c_compile = C_COMPILE_ACTION_NAME,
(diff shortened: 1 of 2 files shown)


oquenchil (approved):

oquenchil (approved):

tpudlik:
Following up on this, is there a reason we have no action name associated with `objdump`? That means you can't use `get_tool_for_action` to get the `objdump` tool and need to use the apparently-deprecated (per https://bazel.build/configure/integrate-cpp#generate-command-lines-toolchain) tool-specific getter. @comius would you mind if I added it?

oquenchil:
Is the question related to: https://github.com/bazelbuild/bazel/issues/8438#issuecomment-1766793970 ?

If it is, it doesn't look like you need the action name in action_names.bzl.

oquenchil:
In any case it should be fine to add the action name if you want it for the default toolchain.

comius:
I don’t mind if objdump is added

PR #16163, opened 2022-08-25
Fix site/en/BUILD.bazel so it can co-exist with site/en/build on OSX

cheister:
When checking out the master branch on OSX it is case insensitive and thinks that `site/en/build` and `site/en/BUILD` are the same file resulting in a dirty git branch with
```
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    site/en/BUILD

no changes added to commit (use "git add" and/or "git commit -a")
```
This PR renames `site/en/BUILD` to `site/en/BUILD.bazel` so there isn't a conflict


fweikert (approved):

Wyverald:
This might not be that easy to import since we don't support "BUILD.bazel" in our internal codebase. @fweikert do you know how to set up Copybara so the import/export works both ways?

fweikert:
Yeah, I actually started a CL after approving.

fweikert:
This is a bit problematic since the actual file on GitHub does not really exist internally (it's being generated from a different file during export).

Due to a different issue the file move got merged into the next exported commit (https://github.com/bazelbuild/bazel/commit/424fe04ee322d0a3a1ddbcf3d0b51ac9b6415fb2). However, due to the move some of the computed dependencies were broken, which should be fixed with the next exported commit. Sorry for this mess, this is something we should fix.

fweikert:
Ok, file has been moved and the post-submit is green again. Unfortunately the change wasn‘t attributed to this PR, sorry for that.

PR #16165, opened 2022-08-25
Use True instead of 1 in the documentation

limdor:
The parameter is a boolean and because of that True can also be used.
From readibility point of view, it is better to use True to know that
this is a boolean value and cannot be any arbitrary integer.
Having 1 in the documentation makes the users also to use 1 instead
of True, that is why the proposal to change it here.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
@@ -246,7 +246,7 @@ public class GenRuleBaseRule implements RuleDefinition {
 
         /* <!-- #BLAZE_RULE(genrule).ATTRIBUTE(output_to_bindir) -->
         <p>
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/devtools/build/lib/rules/genrule/GenRuleBaseRule.java
```diff
@@ -246,7 +246,7 @@ expanded to Windows style paths (with backslash).
 
         /* <!-- #BLAZE_RULE(genrule).ATTRIBUTE(output_to_bindir) -->
         <p>
-          If set to 1, this option causes output files to be written into the <code>bin</code>
+          If set to True, this option causes output files to be written into the <code>bin</code>
```
limdor, line 249: Note that the documentation is rendered saying default is False and not default is 0
https://bazel.build/reference/be/general#genrule.local

Wyverald (approved):

Issue #16145, opened 2022-08-22
dynamic execution: tree artifact intermittently contains .tmp files


beasleyr-vmw:
### Description of the bug:

We are trying to prove out an internal Buildfarm deployment. One of my builds fails intermittently because tree artifacts _sometimes_ contain errant `.tmp` files. Sometimes the `.tmp` files are identical to the non-`.tmp` counterpart, but other times they are significantly smaller as if a download was interrupted (e.g. 1MiB instead of 7MiB).

- We're using dynamic execution via `--internal_spawn_scheduler` and `--strategy={Action,Javac,GoCompilePkg,CppCompile}=dynamic`.
- We're also using `--remote_local_fallback` and (the purportedly deprecated/no-op) `--remote_local_fallback_strategy=sandboxed`.
- As we're proving out the RBE deployment, we've _disabled_ caching and will never see cache hits (`--noremote_upload_local_results`, `--noremote_accept_cached`).
- We *are not* using `--experimental_local_lockfree_output`.
- About the generating action(s):
    - The action class doesn't define a mnemonic, so (I assume) it falls under the dynamic strategy via the `Action` mnemonic.
    - The action declares a tree artifact output containing copies of its inputs into a single directory (roughly `cp -t $OUT $SRCS`). It doesn't create any temporary files of its own.
    - The remote execution gRPC log indicates that the client _does not_ download any files with a `.tmp` extension. Rather, these (presumably) are the temporary outputs created by the remote execution strategy.
- About the consuming action(s):
    - Consumers are genrules which invoke a utility with `$(execpath :tool) $(execpath :treeartifact)`.
    - If the action fails, we run `find -L . -ls; find -L . -type f | xargs sha256sum`.
        - This is how I discovered our inputs include `.tmp` shadows of the expected files.

I have just added `--experimental_debug_spawn_scheduler` with the hope that there will be more clues in the log the next time this occurs.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

~Still working on determining an actual repro case.~ I can't reliably repro unless I'm running under the debugger and intentionally delay the local branch. (See "Any other information" below.)

**Update**: See https://github.com/beasleyr-vmw/repro-bazelbuild-bazel-16145 .

### Which operating system are you running Bazel on?

CentOS 8

### What is the output of `bazel info release`?

5.2.0-vmware

### Have you found anything relevant by searching the web?

[...]

PR #16170, opened 2022-08-26
Download remote outputs into a separate temporary directory.

tjgq:
RemoteExecutionService writes downloaded files into temporary paths derived by
appending .tmp to the output filename. Writing to these temporary locations
isn't guarded by the output lock, as otherwise a slow remote download would
prevent the local branch of a dynamically executed action from completing.
This creates a problem for actions that output tree artifacts, as partial .tmp
files might be present in the output directory when the local branch finishes.
(See https://jmmv.dev/2019/12/bazel-dynamic-execution-download-times.html for
a longer explanation.)

Instead, write the temporary files into a separate temporary directory where
they cannot be accidentally picked up as legitimate output files.

Fixes https://github.com/bazelbuild/bazel/issues/16145.


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -245,21 +247,30 @@ private void initHttpAndDiskCache(
...
     Preconditions.checkState(remoteOptions == null, "remoteOptions must be null");
+    Preconditions.checkState(tempPathGenerator == null, "tempPathGenerator must be null");
 
     RemoteOptions remoteOptions = env.getOptions().getOptions(RemoteOptions.class);
     if (remoteOptions == null) {
       // Quit if no supported command is being used. See getCommandOptions for details.
       return;
     }
 
+    this.tempPathGenerator =
```
coeuvre, line 271: We need to clear stale tmp files just in case something went wrong. Please check the code here https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/remote/AbstractActionInputPrefetcher.java;l=81;drc=cfea5d36056c71da49984d511262ed18a596862b.

coeuvre (changes requested):

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
tjgq: Done. I also had to move the `TmpPathGenerator` instantiation to `RemoteModule#executorInit`. `RemoteModule#beforeCommand` is too early to do that; we need the execroot to be known, which in turn needs the workspace name to have been parsed from the `WORKSPACE` file.

tjgq:
@coeuvre see cl/471782069 internally

PR #16172, opened 2022-08-26
Linux sandbox: ignore ENODEV when making the filesystem read-only.

cf-natali:
The code remounting existing mounts as read-only already ignores various
errors, such as `EINVAL`, `ESTALE` etc.

Unfortunately, it doesn't ignore `ENODEV`, which can occur in case of
`autofs`/`automount` failure.
Currently, having a single broken automount on the host can cause all
sandboxes to fail with the following error:

```
src/main/tools/linux-sandbox-pid1.cc:281: "remount(nullptr,
/data/foo, nullptr, 2101281, nullptr)": No such device
```

So simply also add `ENODEV` to the list of errors ignored.


cf-natali:
See e.g. https://github.com/systemd/systemd/blob/528ba54bbe3cf6c454985952d49e7a49ba62e244/src/core/automount.c#L556 and https://patchwork.kernel.org/project/linux-nfs/patch/87tw43fgrf.fsf@notabene.neil.brown.name/#20551311 for background on automount and `ENODEV`.

meisterT (approved):

PR #16173, opened 2022-08-26
Use True instead of 1 in the documentation

limdor:
The parameter is a boolean and because of that True can also be used.
From readibility point of view, it is better to use True to know that
this is a boolean value and cannot be any arbitrary integer.
Having 1 in the documentation makes the users also to use 1 instead
of True, that is why the proposal to change it here.


aiuto:
lberki is on baby bonding leave, so I reassigned this to the leads of "build semantics" group. IMO, it's sort of a no brainer, but I wanted to give others a chance to chime in.

comius (approved):

comius:
@aiuto feel free to review and import next time :)

PR #16181, opened 2022-08-29
[apple] add support for `arm64` watchOS device

### LINKED ISSUE #16182: add support for `arm64` architecture for watchOS devices (Xcode 14)

Xcode 14 adds `arm64` architecture for watchOS devices:

Screenshot 2022-08-29 at 12 33 42

Currently Bazel [only supports `armv7k` and `arm64_32`](https://github.com/bazelbuild/bazel/blob/26509a7b6c684adcab34d00ef765f069f4c8ac13/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java#L51-L52).

When you would like to build for watchOS device using `arm64` architecture (`watchOS_arm64`) currently it will get [incorrectly mapped to watchOS simulator](https://github.com/bazelbuild/bazel/blob/26509a7b6c684adcab34d00ef765f069f4c8ac13/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java#L141-L142) which already [supports `arm64` architecture](https://github.com/bazelbuild/bazel/blob/26509a7b6c684adcab34d00ef765f069f4c8ac13/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java#L49-L50).

I've already submitted PR (#16181) with proposed changes:
* Adds `watchos_arm64` to supported watchOS device architectures.
* Renames `watchos_arm64` to `watchos_sim_arm64` (following the convention as for tvOS and iOS).

I would like to build in Xcode 14 for watchOS devices with `arm64` architecture.

macOS Monterey 12.5.1

release 5.3.0

_No response_

_No response_

_No response_

_No response_

aleksandergrzyb:
Fixes #16182.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
@@ -47,9 +47,9 @@ public enum ApplePlatform implements ApplePlatformApi {
   private static final ImmutableSet<String> IOS_DEVICE_TARGET_CPUS =
       ImmutableSet.of("ios_armv6", "ios_arm64", "ios_armv7", "ios_armv7s", "ios_arm64e");
   private static final ImmutableSet<String> WATCHOS_SIMULATOR_TARGET_CPUS =
-      ImmutableSet.of("watchos_i386", "watchos_x86_64", "watchos_arm64");
+      ImmutableSet.of("watchos_i386", "watchos_x86_64", "watchos_sim_arm64");
   private static final ImmutableSet<String> WATCHOS_DEVICE_TARGET_CPUS =
-      ImmutableSet.of("watchos_armv7k", "watchos_arm64_32");
+      ImmutableSet.of("watchos_armv7k", "watchos_arm64_32", "watchos_arm64");
   private static final ImmutableSet<String> TVOS_SIMULATOR_TARGET_CPUS =
       ImmutableSet.of("tvos_x86_64", "tvos_sim_arm64");
   private static final ImmutableSet<String> TVOS_DEVICE_TARGET_CPUS =
--- a/tools/osx/crosstool/BUILD.toolchains
+++ b/tools/osx/crosstool/BUILD.toolchains
@@ -59,13 +59,17 @@ OSX_TOOLS_CONSTRAINTS = {
         "@platforms//os:tvos",
         "@platforms//cpu:arm64",
     ],
-    "watchos_arm64": [
+    "watchos_sim_arm64": [
         "@platforms//os:watchos",
         "@platforms//cpu:arm64",
     ],
     "watchos_arm64_32": [
         "@platforms//os:watchos",
         "@platforms//cpu:arm64_32",
+    ],
+    "watchos_arm64": [
+        "@platforms//os:watchos",
+        "@platforms//cpu:arm64",
     ],
     "watchos_armv7k": [
         "@platforms//os:watchos",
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
```diff
@@ -47,9 +47,9 @@ public enum ApplePlatform implements ApplePlatformApi {
   private static final ImmutableSet<String> IOS_DEVICE_TARGET_CPUS =
       ImmutableSet.of("ios_armv6", "ios_arm64", "ios_armv7", "ios_armv7s", "ios_arm64e");
   private static final ImmutableSet<String> WATCHOS_SIMULATOR_TARGET_CPUS =
-      ImmutableSet.of("watchos_i386", "watchos_x86_64", "watchos_arm64");
+      ImmutableSet.of("watchos_i386", "watchos_x86_64", "watchos_sim_arm64");
   private static final ImmutableSet<String> WATCHOS_DEVICE_TARGET_CPUS =
-      ImmutableSet.of("watchos_armv7k", "watchos_arm64_32");
+      ImmutableSet.of("watchos_armv7k", "watchos_arm64_32", "watchos_arm64");
```
brentleyjones, line 52: Since this is a breaking change, we should add something to the release notes at the minimum, though maybe we don't change the name at all? People can have existing configuration that uses `watchos_arm64` to mean the simulator, and some rulesets would have to do a version check or something similar to get the right version of this: https://github.com/buildbuddy-io/rules_xcodeproj/blob/47d9510b64878ae1893d355cbf134d6a4b6eb715/xcodeproj/internal/xcodeproj_rule.bzl#L490-L493

tools/osx/crosstool/BUILD.toolchains
```diff
@@ -59,14 +59,18 @@ OSX_TOOLS_CONSTRAINTS = {
...
         "@platforms//cpu:arm64",
     ],
     "watchos_arm64_32": [
         "@platforms//os:watchos",
         "@platforms//cpu:arm64_32",
     ],
+    "watchos_arm64": [
+        "@platforms//os:watchos",
+        "@platforms//cpu:arm64",
+    ],
```
brentleyjones, line 73: This would also have to change in apple_support. Also a breaking change: https://github.com/bazelbuild/apple_support/blob/e81e1ea6d2a089337955c6a956dc3f96ed67ed09/platforms/BUILD#L129-L136

brentleyjones:
@keith

src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
aleksandergrzyb: > Since this is a breaking change, we should add something to the release notes at the minimum, though maybe we don't change the name at all?

I think breaking change would be suitable for 6.0.0 release? 👀 I wanted to be consistent with the other names, otherwise it can get confusing.

tools/osx/crosstool/BUILD.toolchains
aleksandergrzyb: Should I also make the change in `platforms/BUILD`?

src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java
brentleyjones: It's the best time to break it, yes, but I was also starting how it's tricky for rulesets to do things based on the bazel version.

keith (approved):
don't see how we can avoid the breaking-ness of this without causing worse issues, so I think we should merge and add in the release notes

brentleyjones:
We should try to get this into Bazel 6.0. @sgowroji

brentleyjones:
Can we add this to the 6.0 milestone?

lberki:
@meteorcloudy I'll add this to the list of issues for the 6.0 milestone to be _considered_ as a requirement for 6.0 .

brentleyjones:
Seems this was already addressed a different way upstream? https://github.com/bazelbuild/bazel/commit/ce611646969cfe72ae5e22083708684eae11e478

meteorcloudy:
Looks like the alternative change is already in 6.0.0, so I'll close this one

PR #16184, opened 2022-08-29
Do not crash on URIs without a host component.

### LINKED ISSUE #16171: CredentialHelper change broke unix domain socket support
We use a unix domain socket for our remote_executor and remote_cache endpoints, because we use a side-car proxy to handle SPIFFE mTLS negotiation.  This stopped working with the release of bazel 5.3.0.  With 5.3.0, bazel fails before issuing its first GetCapabilitiesRequest.  The error message was "ERROR: Failed to query remote execution capabilities: UNAUTHENTICATED: Failed computing credential metadata."  When enabling verbose failures, the stack trace pointed to a failed precondition because the host portion of the URI was null.  I played around with the code and got it working again, but the fix isn't production-quality -- it doesn't handle the case where the remote cache endpoint differs from the remote executor endpoint, and might need different credentials.  Because we're using unix domain sockets, this shouldn't affect us at all.  Anyway, consider the following patch a place to start the conversation about what the correct fix might look like.

```diff --git a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
index a7d793ad56..d7db02bd6f 100644
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -1055,6 +1055,22 @@ public final class RemoteModule extends BlazeModule {
       AuthAndTLSOptions authAndTlsOptions,
       RemoteOptions remoteOptions)
       throws IOException {
+
+    // TODO: this is a temporary hack.  Because remoteCache and remoteExecutor could be different, they might need
+    // different credentials.  If the endpoints used to access remote services are unix domain sockets, credential
+    // helpers aren't needed -- the service is running on the same machine.
+    //
+    // Also note that URI("unix://...") will replace "unix" with "https" for some reason that escapes me, but that
+    // makes it tricky to use.  Somewhere (not here because of the checks in the try clause below), that was
+    // happening, because by the time findCredentialHelper() is called, these URIs have been changed from "unix://"
+    // to "https://".  According to the bazel documentation, the "unix:" scheme is supposed to disable TLS.
+    if (remoteOptions.remoteCache != null && Ascii.toLowerCase(remoteOptions.remoteCache).startsWith("unix:/")) {
+           return null;
+    }
+    if (remoteOptions.remoteExecutor != null && Ascii.toLowerCase(remoteOptions.remoteExecutor).startsWith("unix:/")) {
+           return null;
+    }
+
     Credentials credentials =
         GoogleAuthUtils.newCredentials(
             credentialHelperEnvironment, commandLinePathFactory, fileSystem, authAndTlsOptions);
```

tjgq:
URIs such as `unix://...` (Unix domain socket) may legitimately be missing a
host component. We should gracefully handle them when looking for a credential
helper (no such helper can exist for them since the lookup is host-based, but
this could change in the future).

Fixes #16171 which is a regression in 5.3. We had some tests for this in
remote_execution_test.sh, but they were disabled due to flakiness in 8e3c0df.
I've filed #16185 to reenable them.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProvider.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProvider.java
@@ -16,6 +16,7 @@ package com.google.devtools.build.lib.authandtls.credentialhelper;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.devtools.build.lib.vfs.Path;
 import com.google.errorprone.annotations.Immutable;
@@ -66,7 +67,12 @@ public final class CredentialHelperProvider {
   public Optional<CredentialHelper> findCredentialHelper(URI uri) {
     Preconditions.checkNotNull(uri);
 
-    String host = Preconditions.checkNotNull(uri.getHost());
+    String host = uri.getHost();
+    if (Strings.isNullOrEmpty(host)) {
+      // Some URIs (e.g. unix://) legitimately have no host component.
+      return Optional.empty();
+    }
+
     Optional<Path> credentialHelper =
         findHostCredentialHelper(host)
             .or(() -> findWildcardCredentialHelper(host))
--- a/src/test/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProviderTest.java
+++ b/src/test/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProviderTest.java
@@ -105,6 +105,15 @@ public class CredentialHelperProviderTest {
   }
 
   @Test
+  public void uriWithoutHostComponent() throws Exception {
+    Path helper = fileSystem.getPath(EXAMPLE_COM_HELPER_PATH);
+    CredentialHelperProvider provider =
+        CredentialHelperProvider.builder().add("example.com", helper).build();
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProvider.java
```diff
@@ -66,7 +66,12 @@ private CredentialHelperProvider(
   public Optional<CredentialHelper> findCredentialHelper(URI uri) {
     Preconditions.checkNotNull(uri);
 
-    String host = Preconditions.checkNotNull(uri.getHost());
+    String host = uri.getHost();
+    if (host == null) {
```
Yannic, line 70: Maybe use Guava's `Strings.isNullOrEmpty()` here?

Yannic (approved):
A regression test in `CredentialHelperProviderTest` would be nice :)

src/main/java/com/google/devtools/build/lib/authandtls/credentialhelper/CredentialHelperProvider.java
tjgq: Done.

tjgq:
> A regression test in `CredentialHelperProviderTest` would be nice :)

Done.

kshyanashree:
@bazel-io fork 5.3.1

Issue #16087, opened 2022-08-10
Annotation processors are incorrectly run multiple times in android_local_test when using Kotlin


Chang-Eric:
### Description of the bug:

If two annotation processors are triggered on Kotlin sources within the same compilation target and then used as deps within an `android_local_test`, the annotation processors are incorrectly triggered multiple times. This usually results in a failure because the annotation processor hits conflicts trying to write the same file twice.

This occurs with Bazel 5.2.0, but not Bazel 4.2.0. Because this issue requires Kotlin to reproduce, I am using the Kotlin rules from https://github.com/bazelbuild/rules_kotlin. However, only changing the Bazel version from 4.2.0 to 5.2.0 is enough to reproduce this error. I'm using v1.6.0 for the Kotlin rules.

This seems to happen with all sorts of annotation processors. I've tried Dagger, Hilt, AutoValue, and AutoService. It requires two annotation processors to be triggered (i.e. it does not trigger just when the plugins are included with the exported plugins targets being in deps).

For example:

```
android_local_test(
    name = "FooTest",
    srcs = ["FootTest.java"],
    manifest_values = {
        "minSdkVersion": "14",
    },
    deps = [
        ":TestKtLib",
        ...
    ],
)

kt_android_library(
    name = "TestKtLib",
    testonly = True,
    srcs = [
        "TestAutoService.kt",
        "TestAutoValue.kt",
    ],
    deps = [
        # auto service dep
        # auto value dep
    ],
)
```

```kotlin
package foo

import com.google.auto.service.AutoService

@AutoService(String::class)
public class Foo {}
```

```kotlin
package foo

import com.google.auto.value.AutoValue

@AutoValue
abstract class TestAutoValue {
  abstract fun foo(): Long

  companion object {
    fun of(foo: Long): TestAutoValue {
      return AutoValue_TestAutoValue(foo)
    }
  }
}
```

Note that interestingly building `TestKtLib` does not cause the error, just running the test does, which is strange in itself. Running the test in this case however will give an error like the following due to the processors being run twice and both hitting errors from writing the same file twice.

```
warning: /tmp/pwd4144296603390009720/_kotlinc/javatests_foo-TestAutoValue_kt_jvm/temp/stubs/foo/TestAutoValue.java:7: warning: [AutoValueCouldNotWrite] Could not write generated class foo.AutoValue_TestAutoValue: javax.annotation.processing.FilerException: Attempt to recreate a file for type foo.AutoValue_TestAutoValue
[...]

Issue #8006, opened 2019-04-11
improve turbine cold-start performance


benjaminp:
It seems to me that using a persist worker for turbine Java header compilations would be beneficial for the same reasons as running javac as a persistent worker is. Has this been considered?

iirina:
cc: @cushon

cushon:
I considered it, but decided to use a graalvm native-image for turbine instead. That approach has shown good startup performance without the overhead of additional workers.

benjaminp:
I see. So maybe this issue is actually feature request to graal turbine in Bazel's java tools. :)

meisterT:
cc @larsrc-google

larsrc-google:
That, or use multiplex workers.

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

fmeum:
@sgowroji Could you reopen?

benjaminp:
#19361

PR #16186, opened 2022-08-29
Support persistent worker for generating java ABI from source

oliviernotteghem:
Incremental build time on XL projects shows that using ABI jar generated from full jar (ijar) is faster than generating ABI jar from source (hjar), which is counter-intuitive as hjar should allow pipelining. This could be attributed to the lack of persistent worker support for generating ABI jar from source. This diff enables it, i.e you can build now with `--java_header_compilation --strategy=Turbine=worker` to generate ABI jar from source using workers.

Unfortunately, due to bugs in Turbine, build time couldn't be benchmarked. Verification was done only on simple project (Javalib1-> Javalib2 -> Javalib3 -> Javalib4) via build traces : 

before

after

UPDATE: we were able to build our XL apps with source ABI persistent workers. The win for incremental scenario (the ones we care the most about) are only 2-4%. Also, using ABI generated from full jar is still faster for us...

Screen Shot 2022-09-01 at 7 12 19 PM


cushon:
> due to bugs in Turbine, build time couldn't be benchmarked

Which bugs?

oliviernotteghem:
@cushon : at Uber, we use our own annotation processor for dependency injection ([motif](https://github.com/uber/motif)). When switching from `--nojava_header_compilation` to `--java_header_compilation` on our driver app, I noticed more files than expected are generating by the AP with Turbine, and therefore heading compilation may fail because of these extra java file getting compiled and sometimes needed additional (missing) entry on classpath.

Also, a second issue I noticed, is that *sometimes* the AP-generated files from Motif are simply missing from the hjar (this could be specific to kotlin android targets).

I could possibly investigate further, however it's not clear how to build Turbine as I don't see any build files within project?

cushon:
> I could possibly investigate further, however it's not clear how to build Turbine as I don't see any build files within project?

The source is at https://github.com/google/turbine, the jars checked in to bazel are produced by the maven build in the turbine repo

oliviernotteghem:
Got it - didn't realize the pom.xml file before. Any idea what could be wrong here? Are we the only one to experience this? If so, this could be an issue with our AP..

cushon:
Does the AP use any internal javac APIs? Otherwise I don't know of any open issues here, but I've seen APs depend on javac implementation details (e.g. the order of elements returned by APIs), or it could be a turbine or javac bug. I can try to help investigate if it's possible to provide a repro.

benjaminp:
(Let me link an old issue about this: https://github.com/bazelbuild/bazel/issues/8006)

arunkumar9t2:
> Also, a second issue I noticed, is that sometimes the AP-generated files from Motif are simply missing from the hjar (this could be specific to kotlin android targets).


@oliviernotteghem I suspect this could be related to https://github.com/bazelbuild/bazel/issues/16087 and especially Kotlin. We noticed sometimes when using kt_android_library, even transitive plugins appear as direct plugins in the processor classpath. Not sure if that causes issues like annotations processors running twice or generated files not getting picked up correctly. A possible angle to investigate.

oliviernotteghem:
@cushon : I could narrow down the issue we were facing when using Turbine Vs generating ABI from full jar. 
- First, the issue occurs only with kotlin rules, which register the java compile action [here](https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/compile.bzl#L690).
- Second, it happens only with our (internal) optimization to skip KAPT and perform annotation processing on kotlin classes during java compilation - see (*) for a quick explanation of how this is done. 

In this situation, the classes generated from annotation processor (from kotlin classes) are not included in the hjar output. This is a bit of a special situation, so I am not sure if this is something generic enough to fix in Turbine?

(*)
We registered a custom javac plugin, that adds target's kotlin compiled classes to the list of classes on which AP needs to be run (see abbreviated source below). The plugin is registered via the `javac_opts` argument of the java_common.compile call (`javac_opts.append('-Xplugin:"Kaptish %s"' % kt_runtime_jar.path)`)

```
@AutoService(Plugin::class)
class KotlinClassInjectorPlugin : Plugin {
  override fun getName() = "Kaptish"

  override fun init(task: JavacTask, vararg args: String?) {

    if (task is BasicJavacTask) {
      if (Options.instance(task.context).get(Option.PROC) == "none") {
        return
      }

      val classes =
          if (args.isNotEmpty()) getClassesFromJar(args.first()) else getClasses(task.context)
      Arguments.instance(task.context).classNames.addAll(classes)
    }
  }
```

cushon:
> ```
> Arguments.instance(task.context).classNames.addAll(classes)
> ```

Is that `com.sun.tools.javac.main.Arguments`? Turbine only implements the JSR-269 APIs, anything that relies on javac implementation details isn't going to be supported.

oliviernotteghem:
@cushon : yes, it is com.sun.tools.javac.main.Arguments

arunkumar9t2:
After importing this we saw `too many open files` error on a large Android project during Turbine action. We saw similar error in `KotlinKapt` which is fixed [here](https://github.com/bazelbuild/rules_kotlin/pull/848).

Increasing shell limit and `-XX:-MaxFDLimit` would ideally fix it but I realized both JavaBuilder and Turbine does not add user specified jvm flags to their command line. I opened [this](https://github.com/bazelbuild/bazel/issues/16486) to track that. 

Manually fixing the JVM flags allows us to compile just fine.

comius:
Hi @oliviernotteghem! Thank you for contributing to the Bazel repository. We appreciate your time and effort. It seems the review of this PR has stalled, so I will close it. Please feel free to reopen if you'd like to discuss anything further.

PR #16189, opened 2022-08-29
Reenable the UDS proxy tests.

### LINKED ISSUE #16185: Reenable tests for gRPC UDS proxy
https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/remote/remote_execution_test.sh;l=161;drc=b303cd128d9f1d913749927f4a1cd942b10b1ae2
https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/remote/remote_execution_test.sh;l=200;drc=b303cd128d9f1d913749927f4a1cd942b10b1ae2

They were disabled due to flakiness in 8e3c0df. As a consequence, we introduced an undetected regression in 5.3 (#16171).

tjgq:
These tests were disabled due to flakiness in 8e3c0df. As a consequence, we
introduced an undetected regression in 5.3 (#16171).

I've rerun the tests multiple times, both locally and remotely, and I've seen
no evidence of flakiness. Therefore, I'm declaring that this issue has
mysteriously fixed itself.

Notes:

* Use `python3` instead of `python` to run the UDS proxy, as the latter is no
  longer available in some of the environments where the tests run.
* The correct URI syntax for a Unix domain socket URI is unix:///path/to/socket
  rather than unix:/path/to/socket, as the URI technically has no host
  component. The latter form still works due to lenient parsing in the gRPC
  gRPC library, but this may change in the future, so the tests should exercise
  the correct form.

Fixes #16185.

### DIFF
--- a/src/test/shell/bazel/remote/remote_execution_test.sh
+++ b/src/test/shell/bazel/remote/remote_execution_test.sh
@@ -157,8 +157,7 @@ EOF
       || fail "Failed to build //a:foo with remote cache"
 }
 
-# TODO(b/211478955): Deflake and re-enable.
-function DISABLED_test_remote_grpc_via_unix_socket_proxy() {
+function test_remote_grpc_via_unix_socket_proxy() {
   case "$PLATFORM" in
   darwin|freebsd|linux|openbsd)
     ;;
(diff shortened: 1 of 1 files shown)


coeuvre (approved):

coeuvre:
Closing as we discussed offline that the test is still flaky. Feel free to reopen if you have time to look at this again.

PR #16195, opened 2022-08-31
Update on incompatible change process

meteorcloudy:
Our incompatible change process isn't up to date since Bazel adopts LTS release.

This PR update the process to clarify how an incompatible change should be introduced in Bazel.

Related: https://github.com/bazelbuild/continuous-integration/issues/1410 is the effort to fix the Bazelisk + Incompatible flags pipeline to provide useful information for incompatible flags flip.

### DIFF
--- a/site/en/contribute/breaking-changes.md
+++ b/site/en/contribute/breaking-changes.md
@@ -43,8 +43,6 @@ We recommend that:
   update their code. Ideally, when the change is mechanical, include a link to a
   migration tool.
 
-* The description includes the intended length of migration window.
-
 * The description includes an example of the error message users will get if
   they don't migrate. This will make the GitHub issue more discoverable from
   search engines. Make sure that the error message is helpful and actionable.
@@ -71,7 +69,7 @@ should contain the URL of the GitHub issue. As the flag name starts with
 
 In the commit description, add a brief summary of the flag.
 Also add [`RELNOTES:`](release-notes.md) in the following form:
-`RELNOTES: --incompatible_name_of_flag has been added. See #yxz for details`
+`RELNOTES: --incompatible_name_of_flag has been added. See #xyz for details`
 
 The commit should also update the relevant documentation. As the documentation
 is versioned, we recommend updating the documentation in the same commit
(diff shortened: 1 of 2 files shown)


sventiffe (approved):

site/en/contribute/breaking-changes.md
```diff
@@ -80,41 +78,41 @@ as the code change.
...
 
 Before flipping the default value of the flag to true, please make sure that:
 
-  * The migration window is respected.
+  * Core repositories in the ecosystem are migrated. On the
+  [`bazelisk-plus-incompatible-flags` pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags){: .external},
+  the flag should appear under `The following flags didn't break any passing Bazel team owned/co-owned projects`.
   * User concerns and questions have been resolved.
 
+When the flag is ready to flip in Bazel, but blocked on internal migration at Google, please consider setting the flag value to false in the internal `blazerc` file to unblock the flag flip. By doing this, we can ensure Bazel users depend the new behaviour by default as early as possible.
```
fmeum, line 109:
```suggestion
When the flag is ready to flip in Bazel, but blocked on internal migration at Google, please consider setting the flag value to false in the internal `blazerc` file to unblock the flag flip. By doing this, we can ensure Bazel users depend on the new behaviour by default as early as possible.
```

meisterT (approved):

site/en/contribute/breaking-changes.md
```diff
@@ -80,41 +78,41 @@ as the code change.
...
+1. Send PRs to fix downstream projects.
+1. Reach out to the Bazel community for help on migration (e.g. [Bazel Rules Authors SIG](https://bazel-contrib.github.io/SIG-rules-authors/)).
 
 ## Flipping the flag {:#flip-flag}
 
 Before flipping the default value of the flag to true, please make sure that:
 
-  * The migration window is respected.
+  * Core repositories in the ecosystem are migrated. On the
+  [`bazelisk-plus-incompatible-flags` pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags){: .external},
```
Wyverald, line 105: you might need to indent this line and the next

Wyverald (approved):

site/en/contribute/breaking-changes.md
meteorcloudy: Thanks, done!

site/en/contribute/breaking-changes.md
```diff
@@ -71,52 +67,57 @@ should contain the URL of the GitHub issue. As the flag name starts with
...
-  * The migration window is respected.
-  * User concerns and questions have been resolved.
+* Core repositories in the ecosystem are migrated.
+
+    On the [`bazelisk-plus-incompatible-flags` pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags){: .external},
+    the flag should appear under `The following flags didn't break any passing Bazel team owned/co-owned projects`.
+
+* User concerns and questions have been resolved.
+
+When the flag is ready to flip in Bazel, but blocked on internal migration at Google, please consider setting the flag value to false in the internal `blazerc` file to unblock the flag flip. By doing this, we can ensure Bazel users depend on the new behaviour by default as early as possible.
```
aiuto, line 112: This is for Google only.  Should that be highlighted in some way?

site/en/release/backward-compatibility.md
```diff
@@ -60,6 +60,4 @@ For every incompatible change, the issue specifies the following:
 * Description of the changed functionality
 * Migration recipe
 
-The incompatible change issue is closed when the incompatible flag is flipped at
-HEAD. All incompatible changes that are expected to happen in release X.Y
-are marked with a label "breaking-change-X.Y"."
+When an incompatible change is ready for migration with Bazel at HEAD (therefore, also with the next Bazel rolling release), it should be marked with the `migration-ready` label. The incompatible change issue is closed when the incompatible flag is flipped at HEAD.
```
aiuto, line 63: Migration ready is a signal that we are committed to the change and it will become permanent, but it is not necessarily in the next LTS release. We need to account for ones like the non-empty glob change, which has been migration ready for years, but we are still not ready to flip.

We should leave the bug open until we delete the code behind the flag, not just flip it. It is not over until it can't be reverted.

site/en/contribute/breaking-changes.md
```diff
@@ -71,52 +67,57 @@ should contain the URL of the GitHub issue. As the flag name starts with
...
-Later a [Bazel release manager](https://github.com/bazelbuild/continuous-integration/blob/master/docs/release-playbook.md){: .external}
-will update the issue and replace the label with `migration-xx.yy`.
-
-The label `breaking-change-xx.yy` communicates when we plan to flip the flag. If
-a migration window needs to be extended, the author updates the label on GitHub
-issue accordingly.
-
 If a problem is found with the flag and users are not expected to migrate yet:
-remove the flags `migration-xx.yy`.
+remove the flags `migration-ready`.
```
aiuto, line 83: Making people look at the issue to decide if it is migration ready or not it not going to be a great experience.  A retraction of a plan to change behavior has to be acompannied by an announcmented on bazel-discuss that we are changing the plan.

site/en/contribute/breaking-changes.md
```diff
@@ -71,52 +67,57 @@ should contain the URL of the GitHub issue. As the flag name starts with
...
 ## Flipping the flag {:#flip-flag}
 
 Before flipping the default value of the flag to true, please make sure that:
 
-  * The migration window is respected.
-  * User concerns and questions have been resolved.
+* Core repositories in the ecosystem are migrated.
+
+    On the [`bazelisk-plus-incompatible-flags` pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags){: .external},
+    the flag should appear under `The following flags didn't break any passing Bazel team owned/co-owned projects`.
```
aiuto, line 108: How will people know they are bazel team owned. The pipeline should be all or nothing. That is, we should have a mandatory list that block a flag flip and an advisory one, which does not block us. It can't be up to analysis.
Unless we change it around and take the choice out of the developer's hands and give the choice to a release manager, who owns an LTS track and can allow or disallow any specific change.

site/en/contribute/breaking-changes.md
meteorcloudy: Most incompatible change authors are googler, so I think it's fine.

site/en/release/backward-compatibility.md
meteorcloudy: Yes, `migration-ready` doesn't imply it will be flipped in the next major release, it totally depends on the migration status and other potential blockers. Having this label tells the incompatible flags pipeline which flags to test so that we can check the migration statuses for downstream projects.

> We should leave the bug open until we delete the code behind the flag, not just flip it. It is not over until it can't be reverted.

I agree with it, it's just there are already many issues closed without removing the flag, maybe we should reopen this. I'll update this line.

site/en/contribute/breaking-changes.md
meteorcloudy: > Making people look at the issue to decide if it is migration ready or not it not going to be a great experience.

Why? I think it's pretty easy to check what flags are migration ready by just opening [this link](https://github.com/bazelbuild/bazel/issues?q=is%3Aopen+is%3Aissue+label%3Amigration-ready) instead of going through emails.

site/en/contribute/breaking-changes.md
meteorcloudy: This information is recored in the ci downstream projects config: https://github.com/bazelbuild/continuous-integration/blob/534ca201c72b713a36afd017ec6bb38aff9cd4b5/buildkite/bazelci.py#L137

The list might be out of date now, but we can easily update them since @sventiffe @jdob and I already reviewed all repos under bazelbuild.

site/en/contribute/breaking-changes.md
meteorcloudy: I think we can still use the `breaking-change-x.0` label to communicate which flags are meant to be flipped for a major release. Added a sentence about that.

Wyverald: `breaking-change-x.0` will probably only be useful for archiving purposes (i.e. history), since we only do one major release every 9+ months and it's basically impossible for two open `migration-ready` issues to be targeted for different major releases.

meteorcloudy: That is true, but there could be incompatible flags that are not planned to be flipped for the next major release. For example, `--incompatible_disallow_empty_glob` and `--incompatible_strict_action_env`, they are ready for migration, but the migration might be very hard or blocked for other reason. By having this label, we can communicate which flags are actually going to be flipped in next major release, so that users can already start turning them on to prevent regression. 

/cc @aiuto @alexeagle WDYT? Do you still think a custom .bazelrc file is better? Can you parse this label to generate that bazelrc file?

Issue #16191, opened 2022-08-30
Incorrect JUnit parameterized testsuite name in `test.xml`


swarren12:
### Description of the bug:

When running a parameterised JUnit test via Bazel, the output in `test.xml` incorrectly sets the parameters as the name of the testsuite, rather than the fully-qualified test class. For example:

```
simple-test/test.xml
2:<testsuites>
3:  <testsuite name='com.example.SimpleTest' timestamp='2022-08-30T15:59:25.093Z' hostname='localhost' tests='1' failures='1' errors='0' time='0.015' package='' id='0'>
```
vs
```
parameterised-test/test.xml
2:<testsuites>
3:  <testsuite name='com.example.ParameterisedTest' timestamp='2022-08-30T15:59:25.099Z' hostname='localhost' tests='4' failures='2' errors='0' time='0.015' package='' id='0'>
7:  <testsuite name='[true,true]' timestamp='2022-08-30T15:59:25.099Z' hostname='localhost' tests='1' failures='0' errors='0' time='0.013' package='' id='1'>
12:  <testsuite name='[true,false]' timestamp='2022-08-30T15:59:25.113Z' hostname='localhost' tests='1' failures='1' errors='0' time='0.001' package='' id='2'>
55:  <testsuite name='[false,true]' timestamp='2022-08-30T15:59:25.114Z' hostname='localhost' tests='1' failures='1' errors='0' time='0.0' package='' id='3'>
98:  <testsuite name='[false,false]' timestamp='2022-08-30T15:59:25.114Z' hostname='localhost' tests='1' failures='0' errors='0' time='0.0' package='' id='4'
```

In the above snippet, it seems that the class name is included in the file, but it is as an empty `<testsuite>`. It seems as though the parameters are being used as a key for the `<testsuite>` elements, but this is kind of redudant as that information is duplicated on the `<testcase>`s.

**Expected behaviour**: `test.xml` contains a `<testsuite>` using the class name, and then child elements of `<testcase>` that are parameterised.
**Actual behaviour**: `testxml` contains multiple `<testsuite>`s using the parameters, each containing child elements of `<testcase>`.
**Steps to reproduce**: see below

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

See [bug.tar.gz](https://github.com/bazelbuild/bazel/files/9454610/bug.tar.gz), or:

1. Create a new Bazel project;
2. Download the latest `bazel` binary;
[...]

PR #16200, opened 2022-09-01
Fix wrong nested testsuite tags 

blindpirate:
This fixes https://github.com/bazelbuild/bazel/issues/16191

Previously, there was an obvious error in the case of nested test suites.
For example, in the following example:

```
public class Enclosed {
    public static class Inner1 { ... }
}
```

The xml generated by previous `AntXmlResultWriter` was:

```
<testsuite name="Enclosed">
  ...
</testsuite>
<testsuite name="Enclosed$Inner1">
  ...
</testsuite>
```

This is obviously wrong. The correct xml should be:

```
<testsuite name="Enclosed">
  ...
  <testsuite name="Enclosed$Inner1">
    ...
  </testsuite>
</testsuite>
```

I'm a bit surprised it has existed for 6 years unnoticed.

RELNOTES[INC]: xml test results have been changed how nested test classes are displayed. This might either fix or break external integrations.

### DIFF
--- a/src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/AntXmlResultWriter.java
+++ b/src/java_tools/junitrunner/java/com/google/testing/junit/runner/model/AntXmlResultWriter.java
@@ -91,13 +91,13 @@ public final class AntXmlResultWriter implements XmlResultWriter {
     writeTestCases(writer, result, parentFailures);
     writeTestSuiteOutput(writer);
 
-    writer.endElement();
-
     for (TestResult child : result.getChildResults()) {
       if (!child.getChildResults().isEmpty()) {
         writeTestSuite(writer, child, parentFailures);
       }
     }
+
+    writer.endElement();
   }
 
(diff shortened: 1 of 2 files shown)


comius (commented):
I don't think we should make this change. The change would break existing working integrations. It might fix some of them - but I don't have any evidence of that.

The first xsd schema I found online suggest the `testsuite` tags can't be nested.
https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

I would assume most of the users use the first schema they find online.

Malinskiy:
Even though the `testsuite` can't be nested, it can be a child of `testsuites` to simplify the parsing.

I don't get how a seemingly random github repo with a schema is a reference here though. Most CI's use an authorative source, i.e. https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd or https://github.com/jenkinsci/xunit-plugin/blob/84cff8636bc8bc694bbd6e73eb6697901f45d7ba/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd#L4

PR #16207, opened 2022-09-02
errors: noBuildFile suggests BUILD.bazel too

iamricard:
Hi 👋🏼 apologies for the unsolicited submission, this just seemed like a simple QOL improvement for bazel errors so I figured I'd give it a shot (not sure how much of this code is shared with blaze if any at all so I understand if this can't make it in because of google3).

Because there are three kinds of files bazel really cares about (BUILD, BUILD.bazel, and *.bzl) for people unfamiliar with the system it can be confusing that BUILD.bzl, for example, wouldn't work (we've seen this happen). If the noBuildFile error message suggests either BUILD or BUILD.bazel, it becomes more obvious that those are the only two options.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/skyframe/BzlLoadFunction.java
+++ b/src/main/java/com/google/devtools/build/lib/skyframe/BzlLoadFunction.java
@@ -1431,8 +1431,9 @@ public class BzlLoadFunction implements SkyFunction {
       return new BzlLoadFailedException(
           String.format(
               "Every .bzl file must have a corresponding package, but '%s' does not have one."
-                  + " Please create a BUILD file in the same or any parent directory. Note that"
-                  + " this BUILD file does not need to do anything except exist.",
+                  + " Please create a BUILD or BUILD.bazel file in the same or any parent"
+                  + " directory. Note that this BUILD file does not need to do anything except"
+                  + " exist.",
               file),
           Code.PACKAGE_NOT_FOUND);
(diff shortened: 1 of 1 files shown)


iamricard:
ping @sgowroji 👀 wondering if this could get accepted or if i should just close it (so that it doesn't keep showing up on my open PRs)

sgowroji:
@iamricard  Thank you for the patience. We have shared the PR  to our internal team. You will be seeing the significant update very soon.

Wyverald (approved):

brandjon (changes requested):
Not sure why this wasn't merged yet, but looking at it now I have a comment:

We have tons of places where we refer to "BUILD files" -- in code, in error strings, and in user documentation. Should we update virtually all of them to say "BUILD or BUILD.bazel"? I think we'd all agree that's probably overkill, but then what makes it more worthwhile for just this one error message? And what policy should we follow for other occurrences?

(If BUILD.bzl is a common mistake, I'd be tempted to check for the presence of that file and emit a special error in that case. But that's probably too complicated to do in Skyframe, given how many hoops we jump through to lookup where BUILD files are. And it'd need tests for interaction with `--package_path`, etc., to the point that it couldn't be worthwhile.)

brandjon:
I'm going to close this for lack of certainty as to how to address this problem in general for all error messages / documentation. Filed #17054 to track.

PR #16212, opened 2022-09-03
Close Mutability in TemplateDict map_each evaluation

fmeum:
Mutability instances must be wrapped in try-with-resource blocks outside
of test code.


fmeum:
@hvadehra I noticed this while working on a small addition to `template_dict`. Could you take a look?

hvadehra (approved):
Sorry for the delay, and thanks for the fix!

fmeum:
@hvadehra I think you also have to update the labels to get the "awaiting-PR-merge" status.

PR #16213, opened 2022-09-03
Add uniquify parameter to TemplateDict.add_joined

fmeum:
The behavior is analogous to that of Args.add_joined.


fmeum:
@hvadehra With this change, it would among other things become possible to collect the unique user link flags for a hypothetical `cc_static_library` in a memory efficient way (see #1920 and in particular [@oquenchil's example code](https://gist.github.com/oquenchil/3f88a39876af2061f8aad6cdc9d7c045)). Do you think it's worth the additional API in `TemplateDict`?

fmeum:
@hvadehra I rebased onto master, this is ready for review again.

fmeum:
@oquenchil In case @hvadehra should unavailable, could you perhaps review this PR? It's not something that needs to go into 6.0.

hvadehra:
So sorry, this dropped off my radar somehow. Reviewing today.

hvadehra (approved):
I think it does make sense to have this, so thanks! Java rules might start using this for jvm_flags as well

PR #16008, opened 2022-07-30
Fix rpath for binaries in external repositories

fmeum:
The solib directory is located within the subdirectory of the runfiles
directory corresponding to the workspace. Thus, if a binary is contained
in an external repository, its $ORIGIN relative rpath has to first
ascend to the runfiles directory and then descend into the workspace
directory.

Fixes #16003

Issue #16153, opened 2022-08-23
[5.3.1] Bazel 5.3 fails to run external tests


bazel-io:
Forked from #16003

fmeum:
At Bazel HEAD, https://github.com/bazelbuild/bazel/pull/16008#issuecomment-1224267553 could be fixed by adding yet another rpath entry using https://github.com/bazelbuild/bazel/pull/14600 which is going to be merged soon.

For 5.3.1, one option would be to cherry-pick https://github.com/bazelbuild/bazel/pull/14600 and that additional rpath entry.

Alternatively, we could revert both https://github.com/bazelbuild/bazel/commit/e745468461f93839491a4f80d0c1883d9007f9c0 and https://github.com/bazelbuild/bazel/commit/00805727b867d33fd922e63ca82b0d9825ad79fe in Bazel 5.3.1, but that could regress users that started to rely on tests being executed from the runfiles root.

@oquenchil What do you think is the best path forward here? Would having up to three different rpaths on every binary have the potential to cause issues? As far as I understand, the names of dynamically linked binaries are relatively unique within Bazel, but could additional rpath entries that don't apply to the way the binary is actually executed lead to non-hermetic behavior?

oquenchil:
> What do you think is the best path forward here? Would having up to three different rpaths on every binary have the potential to cause issues? As far as I understand, the names of dynamically linked binaries are relatively unique within Bazel, but could additional rpath entries that don't apply to the way the binary is actually executed lead to non-hermetic behavior?

I don't know to be honest. I'd prefer if we do it the cleaner way and revert whatever commits need to be reverted.

Wyverald:
Unfortunately it's unlikely we can revert https://github.com/bazelbuild/bazel/commit/00805727b867d33fd922e63ca82b0d9825ad79fe since it's needed for Roboleaf and has been behavior for over a year now.

According to @oquenchil, @lberki is the best person to talk to about rpaths, but unfortunately he's still on leave for another two weeks.

So, as someone who knows absolutely squat about rpaths (and after consulting @fmeum), I think the best course of action now is merge #14600 into master, cherry-pick it into 5.3.1, cherry-pick another change that adds that extra rpath, and let tests take care of the problem of "do all these combinations work together".

fmeum:
I think that the current issue could be fixed along the lines of #14600 without introducing the breakage on commas. I can have a PR ready by Monday.

pcjanzen:
[...]

PR #16214, opened 2022-09-03
Fix local execution of external dynamically linked cc_* targets

fmeum:
The fix for remote execution of external dynamically linked cc_* targets
in 95a5bfdc3ed regressed local execution of such targets. This is fixed
by emitting two rpaths in the case of an external target.

Fixes https://github.com/bazelbuild/bazel/pull/16008#issuecomment-1224267553

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
@@ -13,6 +13,8 @@
 // limitations under the License.
 package com.google.devtools.build.lib.rules.cpp;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -40,7 +42,6 @@ public class LibrariesToLinkCollector {
   private final PathFragment toolchainLibrariesSolibDir;
   private final CppConfiguration cppConfiguration;
   private final CcToolchainProvider ccToolchainProvider;
-  private final Artifact outputArtifact;
   private final boolean isLtoIndexing;
   private final PathFragment solibDir;
   private final Iterable<? extends LinkerInput> linkerInputs;
@@ -49,7 +50,8 @@ public class LibrariesToLinkCollector {
   private final Artifact thinltoParamFile;
   private final FeatureConfiguration featureConfiguration;
   private final boolean needWholeArchive;
-  private final String rpathRoot;
+  private final ImmutableList<String> potentialExecRoots;
+  private final ImmutableList<String> rpathRoots;
   private final boolean needToolchainLibrariesRpath;
   private final Map<Artifact, Artifact> ltoMap;
   private final RuleErrorConsumer ruleErrorConsumer;
@@ -76,7 +78,6 @@ public class LibrariesToLinkCollector {
     this.cppConfiguration = cppConfiguration;
     this.ccToolchainProvider = toolchain;
     this.toolchainLibrariesSolibDir = toolchainLibrariesSolibDir;
-    this.outputArtifact = output;
     this.solibDir = solibDir;
     this.isLtoIndexing = isLtoIndexing;
     this.allLtoArtifacts = allLtoArtifacts;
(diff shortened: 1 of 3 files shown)


fmeum:
@oquenchil @Wyverald

Wyverald:
Just checking, so this doesn't need #14600 ?

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
```diff
@@ -106,22 +107,67 @@ public LibrariesToLinkCollector(
...
-      // When executed from within a runfiles directory, the binary lies under a path such as
-      // target.runfiles/some_repo/pkg/file, whereas the solib directory is located under
-      // target.runfiles/main_repo.
-      PathFragment runfilesPath = outputArtifact.getRunfilesPath();
-      String runfilesExecRoot;
+      // The runtime location of the solib directory relative to the binary depends on three
+      // independent factors:
+      //
+      // * whether the binary is contained in the main repository or an external repository;
+      // * whether the binary is executed directly or from the runfiles of another target;
```
lberki, line 117: nit: or from its own runfiles tree. AFAIU it's the same as if the runfiles tree was that of another target that depends on it.

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
```diff
@@ -106,22 +107,67 @@ public LibrariesToLinkCollector(
...
+      //    $ORIGIN:    $EXECROOT/pkg
+      //    solib root: $EXECROOT/pkg/file.runfiles/main_repo
+      // 3. main, runfiles, symlink:
+      //    $ORIGIN:    $EXECROOT/pkg
+      //    solib root: $EXECROOT
+      // 4. main, runfiles, regular file:
+      //    $ORIGIN:    other_target.runfiles/main_repo/pkg
+      //    solib root: other_target.runfiles/main_repo
+      // 5. external, direct, symlink:
+      //    $ORIGIN:    $EXECROOT/external/other_repo/pkg
```
lberki, line 140: This (and everything else that starts with `$EXEROOT/external/`) depends on whether `--experimental_sibling_repository_layout` is in effect or not. If not, this is as you say, but if it is, this is `$EXECROOT/../other_repo/pkg`

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
```diff
@@ -106,22 +107,67 @@ public LibrariesToLinkCollector(
...
+      //    solib root: $EXECROOT
+      // 8. external, runfiles, regular file:
+      //    $ORIGIN:    other_target.runfiles/some_repo/pkg
+      //    solib root: other_target.runfiles/main_repo
+      //
+      // Cases 1, 3, 4, 5, and 7 are covered by an rpath that walks up the runfiles path.
+      // Case 8 is covered by walking up some_repo/pkg and then into main_repo.
+      // Cases 2 and 6 are currently not covered as they would require an rpath containing the
+      // binary filename, which may contain commas that would clash with the `-Wl` argument used to
```
lberki, line 155: Are path names containing commas a problem that you encountered or is this just theoretical?

I think it's good practice to be fastidious about quoting, but the right place to worry about that is when the `-Wl` command line argument is constructed. We put a lot of other paths in that argument where we don't worry about quoting commas at all, so it's weird to fix this one special case in a way that doesn't help with fixing all the rest.

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
```diff
@@ -106,22 +107,67 @@ public LibrariesToLinkCollector(
...
+      execRoots.add("../".repeat(runfilesPath.segmentCount() - 1));
       if (runfilesPath.startsWith(LabelConstants.EXTERNAL_RUNFILES_PATH_PREFIX)) {
-        // runfilesPath is of the form ../some_repo/pkg/file, walk back some_repo/pkg and then
-        // descend into the main workspace.
-        runfilesExecRoot = "../".repeat(runfilesPath.segmentCount() - 2) + workspaceName + "/";
-      } else {
-        // runfilesPath is of the form pkg/file, walk back pkg to reach the main workspace.
-        runfilesExecRoot = "../".repeat(runfilesPath.segmentCount() - 1);
+        // Handles cases 8.
+        execRoots.add("../".repeat(runfilesPath.segmentCount() - 2) + workspaceName + "/");
```
lberki, line 164: Do I understand correctly that the `-2` is here because the runfiles path is `../some_repo/pkg/file` and we need to "cancel" `some_repo/pkg` ?

lberki (commented):
In principle, I like this change, even though it results in somewhat unwieldy rpaths. Let's make sure, though, that it works with `--experimental_sibling_repository_layout`. It looks like it does because that flag does not change the layout of the runfiles tree, but better be safe than sorry.

I'm also not very convinced that we need to worry about quoting commas in `-Wl`.

@oquenchil , WDYT?

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
oquenchil: This was breaking internally. lberki pointed out offline that the right fix might be to quote the commas where the -Wl argument is generated.

fmeum:
> Just checking, so this doesn't need #14600 ?

That's right, mostly because it steals large parts of it :-)

fmeum:
@oquenchil The cc_builtin_test is failing with RBE only, but the logs don't contain much information about the failure: https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/01830eb5-da8b-48ab-8206-9cbff9d36d41/src/main/starlark/tests/builtins_bzl/cc_builtin_tests/test.log
Do you have an idea what could cause this failure if all other tests pass?

src/main/java/com/google/devtools/build/lib/rules/cpp/LibrariesToLinkCollector.java
fmeum: @oquenchil I replied to your comment here instead of at https://github.com/bazelbuild/bazel/pull/16214#discussion_r963879539.

oquenchil (approved):
LGTM

fmeum:
@bazel-io flag

Wyverald:
The issue fixed by this PR is tracked for 5.3.1 at https://github.com/bazelbuild/bazel/issues/16153

Issue #16124, opened 2022-08-18
Implement "Locating runfiles with Bzlmod" proposal


fmeum:
The ["Locating runfiles with Bzlmod" proposal](https://github.com/bazelbuild/proposals/blob/main/designs/2022-07-21-locating-runfiles-with-bzlmod.md) consists of the following individual steps:

* [x] ~Add the `RunfilesLibraryInfo` provider (#16125)~
* [x] Collect repository names and mappings of the transitive closure of a target (#16278)
* [x] Emit a repository mapping manifest for every executable (#16321 and https://github.com/bazelbuild/bazel/pull/16652)
* [x] Revert `RunfilesLibraryInfo`
* [x] Expose the current repository name in rules
  * [x] Bash (#16693)
  * [x] C++ (#16216)
  * [x] Java (#16534)
  * [x] Python (#16341)
* [x] Parse and use the repository mapping manifest in runfiles libraries (https://github.com/bazelbuild/bazel/issues/15029)
  * [x] Bash (#16693)
  * [x] C++ (#16623 and https://github.com/bazelbuild/bazel/pull/16701)
  * [x] Java (#16549)
  * [ ] ~Python~ will be updated in rules_python
* [x] Use the Java runfiles library to make Stardoc work with repository mappings (https://github.com/bazelbuild/bazel/issues/14140)

Status of runfiles libraries in third-party rulesets:

* [x] rules_go (https://github.com/bazelbuild/rules_go/pull/3347)
* [ ] [rules_rust](https://github.com/bazelbuild/rules_rust/blob/main/tools/runfiles/runfiles.rs) (???)
* [ ] [rules_haskell](https://github.com/tweag/rules_haskell/blob/master/tools/runfiles/README.md) (???)

fmeum:
@Wyverald Could you assign me and add the appropriate labels?

fmeum:
@comius Do you expect `java_binary` to be starlarkified in time for Bazel 6? That would simplify the changes to the Java rules.

fmeum:
@Wyverald Oops, I missed the use of `Closes ...` in the PR description - we should use `Work towards ...` instead. Could you reopen? For some I lack the permission to do so.

lberki:
The most likely future seems to be that we'll standardize on Python imports _not_ containing the repository name, in which case bzlmod is a no-op.

[...]

PR #16216, opened 2022-09-04
Provide BAZEL_CURRENT_REPOSITORY local define in cc_* rules 

fmeum:
`BAZEL_CURRENT_REPOSITORY` contains the canonical name of the repository
containing the current target and is required to look up runfiles using
apparent repository names when repository mappings are used, e.g. with
Bzlmod.

Work towards https://github.com/bazelbuild/bazel/issues/16124

### DIFF
--- a/src/main/starlark/builtins_bzl/common/cc/cc_binary.bzl
+++ b/src/main/starlark/builtins_bzl/common/cc/cc_binary.bzl
@@ -690,7 +690,7 @@ def cc_binary_impl(ctx, additional_linkopts):
         cc_toolchain = cc_toolchain,
         user_compile_flags = cc_helper.get_copts(ctx, common, feature_configuration, additional_make_variable_substitutions),
         defines = common.defines,
-        local_defines = common.local_defines,
+        local_defines = common.local_defines + cc_helper.get_local_defines_for_runfiles_lookup(ctx),
         loose_includes = common.loose_include_dirs,
         system_includes = common.system_include_dirs,
         private_hdrs = common.private_hdrs,
--- a/src/main/starlark/builtins_bzl/common/cc/cc_helper.bzl
+++ b/src/main/starlark/builtins_bzl/common/cc/cc_helper.bzl
@@ -927,6 +927,9 @@ def _is_stamping_enabled_for_aspect(ctx):
         stamp = ctx.rule.attr.stamp
     return stamp
 
+def _get_local_defines_for_runfiles_lookup(ctx):
+    return ["BAZEL_CURRENT_REPOSITORY=\"{}\"".format(ctx.label.workspace_name)]
+
 cc_helper = struct(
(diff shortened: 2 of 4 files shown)


fmeum:
@oquenchil I went with the slightly less verbose `BAZEL_CURRENT_REPOSITORY` instead of `BAZEL_CURRENT_REPOSITORY_NAME` as suggested in the proposal. Which name do you like best?

fmeum:
Converted to Draft due to the possible changes to the proposal: https://github.com/bazelbuild/proposals/pull/274

oquenchil (approved):
LGTM

oquenchil:
> Converted to Draft due to the possible changes to the proposal: [bazelbuild/proposals#274](https://github.com/bazelbuild/proposals/pull/274)

Saw this too late. Let me know if I need to take another look.

fmeum:
@oquenchil After a discussion with @lberki, we decided to ditch `RunfilesLibraryInfo`. That greatly simplifies the PR. Do you see a problem with having additional `local_define`s for all cc_* targets?

fmeum:
I will flesh out the test a bit more.

oquenchil:
For literally all cc_* targets? Do you mean adding the attribute to all cc_* rules and the targets that need it would use it?

fmeum:
> For literally all cc_* targets? Do you mean adding the attribute to all cc_* rules and the targets that need it would use it?

Yes, literally all of them. But by implicitly adding to the `local_defines` parameter of `cc_common.compile` rather than literally adding to the `local_defines` rule attribute.

oquenchil:
Ah then the latter is totally fine.

fmeum:
@oquenchil I improved the test. Since it relies on running an external test, I temporarily stacked this on #16214. When that has been merged, I will rebase on master and mark as ready. But you can already review.

fmeum:
@oquenchil Test failures are fixed and this PR is no longer stacked.

Wyverald:
Looks good :) Is this ready for merge?

fmeum:
> Looks good :) Is this ready for merge?

Yes, it's ready.

fmeum:
I would have said no: As this is a `local_define`, it's only set when compiling the sources of the current `cc_library` and not exposed transitively. 

@BoleynSu Could you provide an example of a situation you are worried about?

fmeum:
@BoleynSu I agree that that situation would be UB, but is there more to it than "exposing macros via inline functions is potentially dangerous"? Wouldn't the same apply to e.g. the `__FILE__` macro?

fmeum:
Something like that looks pretty good as a modern C++ replacement for what we are doing here. If it requires C++20, we probably have to hide it behind std level checks.

@BoleynSu Since you seem to have this mostly figured out, would you be willing to submit a PR adding this to https://github.com/bazelbuild/bazel/blob/master/tools/cpp/runfiles/runfiles_src.h?

PR #16219, opened 2022-09-05
Enable `oso_prefix_is_pwd` by default in Apple crosstool

thii:
AFAIK this wasn't enabled by default because Xcode older than 11 don't
support this. This change conditionally adds the `OSO_PREFIX_MAP_PWD`
flag to wrapped_clang when Xcode >= 11 is used.


keith:
This also probably breaks debugging by default unless you change the pwd in a lldbinit? It's unclear to me which we should prefer in these cases, hermiticity or out of the box debug-ability.

thii:
Closing since a similar change has been merged in the Apple toolchain maintained in apple_support https://github.com/bazelbuild/apple_support/pull/206.

PR #16224, opened 2022-09-06
Run Android resource processor actions in a single worker process

arunkumar9t2:
Set `worker-key-mnemonic` in execution info to allow android resource processing actions to share the same worker process.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/rules/android/BusyBoxActionBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/rules/android/BusyBoxActionBuilder.java
@@ -13,6 +13,8 @@
 // limitations under the License.
 package com.google.devtools.build.lib.rules.android;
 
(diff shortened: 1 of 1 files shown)


arunkumar9t2:
cc @Bencodes

Bencodes:
Have you done any benchmarking on this to see what kind of performance impact this has? This change seems like it could actually bottleneck a highly parallelized build graph because all Android busybox actions are now being executed under a single `worker-key-mnemonic`.

arunkumar9t2:
> This change seems like it could actually bottleneck a highly parallelized build graph because all Android busybox actions are now being executed under a single worker-key-mnemonic. 

This is true, the goal was to use a single worker process that multiplexes requests and then use https://bazel.build/reference/command-line-reference#flag--worker_max_multiplex_instances to increase the parallelization limit to avoid bottleneck. Initially we considered this to reduce overall Bazel worker memory usage since we saw multiple instances of `ResourceProcessorBusyBox` being active at a given time.
No longer cosidering this internally but happy to circle back if there is interest. Will go ahead and close this.

PR #16233, opened 2022-09-07
Fix grpc 1.47.0 patches

meteorcloudy:
- Added missing `+++`,`---` lines to grpc_1.47.0.patch
- Removed old patch files

The patch doesn't work without those missing lines, but due to a bug, Bazel didn't report this as an error.


aranguyen (approved):
Thank you!

PR #16239, opened 2022-09-08
PatchUtil.java: don't sliently ignore patch chunk without prelude lines

meteorcloudy:
Previously, patch content without ---/+++ prelude lines was silently ignored, which may cause patch file to be incorrectly applied. For example, https://github.com/bazelbuild/bazel/pull/16233

This PR makes sure PatchUtil correctly reports the error to users.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java
+++ b/src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java
@@ -405,26 +405,6 @@ public final class PatchUtilTest {
   }
 
   @Test
-  public void testMissingBothOldAndNewFile() throws IOException {
-    Path patchFile =
-        scratch.file(
-            "/root/patchfile",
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/devtools/build/lib/bazel/repository/PatchUtilTest.java
```diff
@@ -404,26 +404,6 @@ public void testPatchFileNotFound() {
     assertThat(expected).hasMessageThat().contains("Cannot find patch file: /root/patchfile");
   }
 
-  @Test
-  public void testMissingBothOldAndNewFile() throws IOException {
```
meteorcloudy, line 408: With the checking for ---/+++ lines, this error will never be thrown again, so removing this test.

Perim4444221 (approved):

meteorcloudy:
The verify_workspace test is failing on RBE again, I have another CL to disable this test on RBE.

PR #16259, opened 2022-09-13
Remove no-op options already no-op in Bazel 5

limdor:
These no-op options were already no-op in Bazel 5 LTS.


Wyverald (approved):
Thank you!

PR #16260, opened 2022-09-13
[repository/downloader] Add support for multiple header values

Yannic:
Headers can generally be specified multiple times, not just once.

This refactoring is prerequisite for adding support for getting credentials from the credential helper.

Note that this does not change the Starlark API or the qualifier for the gRPC remote downloader (both of which need to go through the process for incompatible changes - which should hopefully be pretty straight forward to do as this feature doesn't seem to be used that much AFAICT).

Progress on #15856

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
@@ -23,6 +23,7 @@ import build.bazel.remote.execution.v2.Digest;
 import build.bazel.remote.execution.v2.RequestMetadata;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.google.common.collect.Iterables;
 import com.google.devtools.build.lib.bazel.repository.downloader.Checksum;
 import com.google.devtools.build.lib.bazel.repository.downloader.Downloader;
 import com.google.devtools.build.lib.bazel.repository.downloader.HashOutputStream;
@@ -105,14 +106,14 @@ public class GrpcRemoteDownloader implements AutoCloseable, Downloader {
 
   @Override
   public void download(
-      List<URL> urls,
-      Map<URI, Map<String, String>> authHeaders,
-      com.google.common.base.Optional<Checksum> checksum,
-      String canonicalId,
-      Path destination,
-      ExtendedEventHandler eventHandler,
-      Map<String, String> clientEnv,
-      com.google.common.base.Optional<String> type)
+          List<URL> urls,
+          Map<URI, Map<String, List<String>>> authHeaders,
+          com.google.common.base.Optional<Checksum> checksum,
+          String canonicalId,
+          Path destination,
+          ExtendedEventHandler eventHandler,
+          Map<String, String> clientEnv,
+          com.google.common.base.Optional<String> type)
       throws IOException, InterruptedException {
     RequestMetadata metadata =
         TracingMetadataUtils.buildMetadata(buildRequestId, commandId, "remote_downloader", null);
@@ -148,7 +149,7 @@ public class GrpcRemoteDownloader implements AutoCloseable, Downloader {
   static FetchBlobRequest newFetchBlobRequest(
       String instanceName,
       List<URL> urls,
-      Map<URI, Map<String, String>> authHeaders,
+      Map<URI, Map<String, List<String>>> authHeaders,
       com.google.common.base.Optional<Checksum> checksum,
       String canonicalId) {
     FetchBlobRequest.Builder requestBuilder =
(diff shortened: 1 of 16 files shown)


Yannic:
@tjgq PTAL

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
```diff
@@ -197,13 +198,17 @@ private OutputStream newOutputStream(
...
-      Map<String, String> orderedHeaders = new TreeMap<>(entry.getValue());
-      for (Map.Entry<String, String> subEntry : orderedHeaders.entrySet()) {
-        subObject.addProperty(subEntry.getKey(), subEntry.getValue());
+      Map<String, List<String>> orderedHeaders = new TreeMap<>(entry.getValue());
+      for (Map.Entry<String, List<String>> subEntry : orderedHeaders.entrySet()) {
+        // TODO(yannic): Introduce incompatible flag for including all headers, not just the first.
+        String value = Iterables.getFirst(subEntry.getValue(), null);
+        if (value != null) {
+          subObject.addProperty(subEntry.getKey(), value);
+        }
```
tjgq, line 211: It's not obvious to me why this is needed. In `StarlarkBaseExternalContext#getAuthHeaders`, the map is populated from a Starlark dict, which by construction cannot contain multiple values for the same key. Am I missing some other code path where the map could be populated with multiple values?

tjgq:
@Wyverald Could you please also take a look at this PR, in particular the bits that affect repository fetching?

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
Yannic: The credential helper returns a `map<string, list<string>>` for `headers`, so we need this change for adding the headers from the helper. We could implement it in a way that only passes the first (or last) value for the header to the remote server, but that doesn't seem like a good idea IMO.

We can allow setting multiple headers from Starlark by allowing `Dict<String, Sequence<String>>` in addition to `Dict<String, String>`, but that would be a follow-up.

tjgq: I understand that this change is necessary. What I don't understand is why `GrpcRemoteDownloader#addHeadersJson` must only call `addProperty` on the first header value for a given key, and ignore the rest.

- If you're *not* using a credential helper, headers should come solely from the Starlark dict (unless I'm missing something), therefore there won't be more than one value per key.
- If you *are* using a credential helper, adding multiple values per key doesn't need to be considered a breaking change, since credential helpers are a new feature (again, unless I'm missing something).

Yannic: The JSON is passed as `Qualifier` in the gRPC download request. The server may or may not look at the JSON (it's not a well-specified part of the assets API), so switching to multiple values here has potential to break things.

We could make the JSON dependent on where a header comes from. But we probably want to merge the headers from Starlark and the credential helper (at least that's my understanding on how the 2 features work together?), and sending a different qualifier based on where credentials come from seems very confusing. That said, with my security hat on, I'm not sure the headers should even be part of the qualifier (esp. in plaintext). The digest of the headers should be sufficient, if sending anything at all.

tjgq: > The JSON is passed as Qualifier in the gRPC download request. The server may or may not look at the JSON (it's not a well-specified part of the assets API), so switching to multiple values here has potential to break things.

But the breakage could only possibly happen if you're already setting `--experimental_credential_helper`, right? I would argue that this falls under "APIs and behavior guarded by an --experimental_* flag can change at any time" (https://bazel.build/release/backward-compatibility) and an --incompatible flag might not be necessary.

> We could make the JSON dependent on where a header comes from. But we probably want to merge the headers from Starlark and the credential helper (at least that's my understanding on how the 2 features work together?), and sending a different qualifier based on where credentials come from seems very confusing.

I agree that the qualifier should not depend on the provenance of the headers. Although something to ponder is what would happen if the same header is set from both Starlark and a credential helper. Would that cause the header to be sent twice, and if so, would that still work?

> That said, with my security hat on, I'm not sure the headers should even be part of the qualifier (esp. in plaintext). The digest of the headers should be sufficient, if sending anything at all.

It might be worth raising this issue with the Remote Execution API working group (but we don't have to block on it).

tjgq: Unfortunately, I think it will be confusing no matter what we do. What's the expected behavior if you set both `--experimental_credential_helper` and the Starlark attribute, but *don't* set the incompatible flag? Forwarding only one of the headers is guaranteed to be a surprise to someone, even if we make a deterministic choice.

That being said, I won't be opposed to adding the incompatible flag if you so insist. The PR otherwise LGTM (but I'd like @Wyverald to chip in as well, in case there's some subtlety around repository fetching that I don't fully understand).

Wyverald: To confirm, we don't need an incompatible flag for the repo fetching case, right? The Starlark API can just accept both types.

Wyverald (approved):
LGTM for the repo fetching part.

src/main/java/com/google/devtools/build/lib/remote/downloader/GrpcRemoteDownloader.java
Yannic: Yes, I think that would work (although, haven't tried).

Yannic: (I assume you're talking about `repository_ctx.download(auth={...})` here)

Issue #16228, opened 2022-09-07
Empty code coverage report is generated when test files are referenced as external repository


VinayDodamani:
### Description of the bug:

an empty code coverage report is generated when the test files exist in an external respository that has its own WORKSPACE file. following is the structure of my code:

local_repository(
name = "sample",
path = "sample",
)

├── BUILD
├── MODULE.bazel
├── sample
│ |── BUILD
│ │   ├──sample_module
│ │   │  ├──test_sample_module.cpp
│ │   ├── BUILD
│ └── WORKSPACE
├── WORKSPACE

and i am using the following coverage relevant flags:
coverage --instrument_test_targets
coverage --instrumentation_filter="[:]"
coverage --combined_report=lcov
coverage --test_env=VERBOSE_COVERAGE=1
coverage --test_output=all

along with this i have also set the appropriate env variables GCOV_PATH and LLVM_COV

coverage:clang --test_env=COVERAGE_GCOV_PATH=external/toolchain.path/usr/local/bin/llvm-profdata
coverage:clang --test_env=LLVM_COV=external/toolchain.path/usr/local/bin/llvm-cov

when i run the following command from the root, i get an empty coverage report:

bazel coverage --config=clang @sample//...

but, I was able to generate a coverage data when I copied the .bazelrc and MODULE.bazel inside the sample repo from the root and ran the following command in the sample repo:

bazel coverage --config=clang //...

the problem that we have is that, we have multiple packages that are external repositories and therefore copying these files everytime into the respective external repositories is not feasible.

I also went through other issues that had similar porblems and most pointed towards the --instrumentation_filter as a potential solution but that also has not worked for me. I have tried instrumenting using the --instrumentation_filter but with no success.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

steps mentioned in the description of the bug

### Which operating system are you running Bazel on?

Ubuntu 18.04.6 LTS 

### What is the output of `bazel info release`?

release 5.1.1

### Any other information, logs, or outputs that you want to share?

Sep 06, 2022 3:04:26 PM com.google.devtools.coverageoutputgenerator.Main getTracefiles
INFO: No lcov file found.
Sep 06, 2022 3:04:26 PM com.google.devtools.coverageoutputgenerator.Main getGcovInfoFiles
INFO: No gcov info file found.
Sep 06, 2022 3:04:26 PM com.google.devtools.coverageoutputgenerator.Main getGcovJsonInfoFiles
[...]

Issue #16208, opened 2022-09-02
No coverage data if testing code from external repositories


celkas:
### Description of the bug:

Using software and tests in the same repository is no problem, coverage data are produced.
Splitting the two makes problems. There is "no coverage data" produced anymore.
It seems that coverage data is generated, but not consumed, maybe because of some path issue/mismatch.
If I also instrument the tests, the test's coverage will be copied into the combined coverage file, but still no software coverage.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Minimal example in [this test repo](https://github.com/celkas/bazel-issue-tests). This will use [this software repo](https://github.com/celkas/bazel-issue-sw) and produce no coverage.

Simply check it out and run
```
bazel coverage -c dbg --combined_report=lcov --instrumentation_filter="@sw_repo//:funclib" --action_env=COVERAGE_GCOV_OPTIONS=-b //:mytest
```
This will say `WARNING: There was no coverage found.`

### Which operating system are you running Bazel on?

Debian (WSL2)

### What is the output of `bazel info release`?

release 5.3.0

### Have you found anything relevant by searching the web?

I did not find anything searching.

In [this Slack discussion](https://bazelbuild.slack.com/archives/CA31HN1T3/p1662129477832939) I was encouraged to create an issue.

### Any other information, logs, or outputs that you want to share?

I can see in the testlog that coverage info is build. Then it is processed but no data is found.
```
Executing tests from //:mytest
-----------------------------------------------------------------------------
File 'test/test.c'
Lines executed:83.33% of 6
Branches executed:100.00% of 2
Taken at least once:50.00% of 2
Calls executed:50.00% of 2

Creating 'test.pic.gcov.json.gz'
Lines executed:83.33% of 6
File 'external/sw_repo/sw/func.c'
Lines executed:53.85% of 13
Branches executed:100.00% of 3
Taken at least once:33.33% of 3
No calls

Creating 'func.pic.gcov.json.gz'
Lines executed:53.85% of 13
Sep 02, 2022 7:49:56 PM com.google.devtools.coverageoutputgenerator.Main getTracefiles
INFO: No lcov file found.
Sep 02, 2022 7:49:56 PM com.google.devtools.coverageoutputgenerator.Main getGcovInfoFiles
INFO: No gcov info file found.
Sep 02, 2022 7:49:56 PM com.google.devtools.coverageoutputgenerator.Main getGcovJsonInfoFiles
[...]

PR #16268, opened 2022-09-13
Allow Java coverage collection for external targets

fmeum:
* Removes a hardcoded filter that results in the LCOV merger filtering
  out coverage of external targets even if these are matched by
  `--instrumentation_filter`.
* Lets `LazyWritePathsFileAction` write out exec rather than
  root-relative paths, as advertised by its javadocs, so that the paths
  match those that the LCOV mergers filters by.

Work towards https://github.com/bazelbuild/bazel/issues/16228
Work towards https://github.com/bazelbuild/bazel/issues/16208

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/actions/LazyWritePathsFileAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/actions/LazyWritePathsFileAction.java
@@ -94,7 +94,7 @@ public final class LazyWritePathsFileAction extends AbstractFileWriteAction {
         continue;
       }
       if (file.isSourceArtifact() || includeDerivedArtifacts) {
-        stringBuilder.append(file.getRootRelativePathString());
+        stringBuilder.append(file.getExecPathString());
         stringBuilder.append("\n");
       }
     }
--- a/src/test/shell/bazel/bazel_coverage_java_test.sh
+++ b/src/test/shell/bazel/bazel_coverage_java_test.sh
@@ -1143,4 +1143,176 @@ end_of_record"
   assert_coverage_result "$coverage_result_num_lib_header" "$coverage_file_path"
 }
 
+function setup_external_java_target() {
+  cat >> WORKSPACE <<'EOF'
+local_repository(
+    name = "other_repo",
+    path = "other_repo",
+)
+EOF
+
+  cat > BUILD <<'EOF'
+java_library(
+    name = "math",
+    srcs = ["src/main/com/example/Math.java"],
+    visibility = ["//visibility:public"],
+)
+EOF
+
+  mkdir -p src/main/com/example
+  cat > src/main/com/example/Math.java <<'EOF'
+package com.example;
+
+public class Math {
+
+  public static boolean isEven(int n) {
+    return n % 2 == 0;
+  }
+}
+EOF
+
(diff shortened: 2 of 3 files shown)


fmeum:
@c-mita Friendly ping

c-mita (approved):

fmeum:
@bazel-io flag

kshyanashree:
Hello @c-mita, As I could see we are good to merge the PR, does it require any further review? Thanks!

meteorcloudy:
@bazel-io fork 6.0.0

c-mita:
> Hello @c-mita, As I could see we are good to merge the PR, does it require any further review? Thanks!

Just forgot to update the labels.

kshyanashree:
merged in [#16637](https://github.com/bazelbuild/bazel/pull/16637)

meteorcloudy:
@kshyanashree Unfortunately, this PR was rolled back in https://github.com/bazelbuild/bazel/commit/4caae75b49e815ad2cf1d805f316bc374f03f2ae at HEAD, can you please also cherry pick the rollback of this change?

meteorcloudy:
@bazel-io fork 6.0.0

kshyanashree:
> @kshyanashree Unfortunately, this PR was rolled back in [4caae75](https://github.com/bazelbuild/bazel/commit/4caae75b49e815ad2cf1d805f316bc374f03f2ae) at HEAD, can you please also cherry pick the rollback of this change?

Sure @meteorcloudy! I will do that.

fmeum:
@meteorcloudy @susinmotion Is there any more context on the nature of the breakage this causes your could provide me with?

meteorcloudy:
@c-mita @hvadehra Can you help provide a minimal reproducible case for the breakage?

c-mita:
Unfortunately not at the moment; we're not really sure why there's a failure.

It might just be a bad test, but unfortunately we won't not be in a position to investigate in detail until after BazelCon.

fmeum:
@c-mita Friendly ping, did you get a chance to investigate the failure? If you can share details about the failing test, I can also look into it.

c-mita:
An internal rule also makes use of `LazyWritePathsFileAction`. I've only looked through the rule, but it seems to make use of this:
https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/actions/Substitution.java;l=230
The rule in question may be the _only_ user of  this).

I would guess this PR would also have to change this, but it's probably best I handle it on our side.

fmeum:
Thanks for looking into it. Let me know if I can help with the roll-forward, but I agree its probably easier if you do it and fix the rule usage in the same CL. I can't reopen the current PR, but feel free to use it for that.

c-mita:
My guess was wrong; there's simply another lump of internal code that assumes these written to the file are root-relative.

Interestingly, when consuming this file, it then appears to stick the exec path in front, so I think it could be simplified and this PR could be accepted as is afterwards, but making the required change is not proving straightforward.

c-mita:
It seems the internal rule really does require the old behaviour to preserved. As a compromise, I've adjusted `LazyWritePathsFileAction` to take an optional converter parameter. By default it writes out the exec path.

00e9af19 is the "roll-forward" with this change.

PR #16272, opened 2022-09-13
Create and stage symlink artifacts with unmodified target path

fmeum:
Unresolved symlink artifacts created with `ctx.actions.symlink(target_path = ...)` are now created without making the target path absolute by prepending the exec root, which diverged from the documentation and intended goal and also gave rise to hermeticity issues as such symlinks would regularly resolve outside the sandbox.

Furthermore, in order to bring local execution in line with other execution types, the runfiles manifest entry (and thus the runfiles directory contents) for symlink artifacts are now their target paths rather than their exec paths, which along the way resolves another soure of non-hermetic resolution outside the runfiles directory.

This requires making symlink artifacts (but not artifacts representing regular files) inputs of `SourceManifestAction`, which previously had no inputs. The change flattens a nested set in `getInputs`, but benchmarks confirm that this does not result in a performance regression. Alternatives considered either result in a substantially worse Starlark API or wouldn't work for symlinks created by spawns.

Work towards #10298

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
@@ -13,16 +13,23 @@
 // limitations under the License.
 package com.google.devtools.build.lib.analysis;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.build.lib.actions.AbstractAction;
 import com.google.devtools.build.lib.actions.ActionExecutionContext;
 import com.google.devtools.build.lib.actions.ActionKeyContext;
 import com.google.devtools.build.lib.actions.ActionOwner;
 import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.actions.Artifact.DerivedArtifact;
 import com.google.devtools.build.lib.analysis.actions.AbstractFileWriteAction;
 import com.google.devtools.build.lib.analysis.actions.DeterministicWriter;
+import com.google.devtools.build.lib.analysis.starlark.UnresolvedSymlinkAction;
+import com.google.devtools.build.lib.collect.nestedset.Depset;
+import com.google.devtools.build.lib.collect.nestedset.NestedSet;
 import com.google.devtools.build.lib.collect.nestedset.NestedSetBuilder;
 import com.google.devtools.build.lib.collect.nestedset.Order;
 import com.google.devtools.build.lib.concurrent.ThreadSafety.Immutable;
@@ -99,6 +106,8 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
 
   private final boolean remotableSourceManifestActions;
 
+  private NestedSet<Artifact> symlinkArtifacts = null;
+
   /**
    * Creates a new AbstractSourceManifestAction instance using latin1 encoding to write the manifest
    * file and with a specified root path for manifest entries.
@@ -129,10 +138,43 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
       Artifact primaryOutput,
       Runfiles runfiles,
       boolean remotableSourceManifestActions) {
+    // The real set of inputs is computed in #getInputs().
     super(owner, NestedSetBuilder.emptySet(Order.STABLE_ORDER), primaryOutput, false);
     this.manifestWriter = manifestWriter;
     this.runfiles = runfiles;
     this.remotableSourceManifestActions = remotableSourceManifestActions;
+  }
+
+  /**
+   * The manifest entry for a symlink artifact should contain the target of the symlink rather
+   * than its exec path. Reading the symlink target requires that the symlink artifact is declared
+   * as an input of this action. Since declaring all runfiles as inputs of the manifest action would
+   * unnecessarily delay its execution, this action exceptionally overrides
+   * {@link AbstractAction#getInputs()} and filters out the non-symlink runfiles by flattening the
+   * nested set of runfiles. Benchmarks confirmed that this does not regress performance.
+   *
+   * <p>Alternatives considered:
+   * <ul>
+   * <li>Having users separate normal artifacts from symlink artifacts during analysis: Makes it
+   * impossible to pass symlink artifacts to rules that aren't aware of them and requires the use of
+   * custom providers to pass symlinks to stage as inputs to actions.</li>
+   * <li>Reaching into {@link ActionExecutionContext} to look up the generating action of symlink
+   * artifacts and retrieving the target from {@link UnresolvedSymlinkAction}: This would not work
+   * for symlinks whose target is determined in the execution phase.</li>
+   * <li>Input discovery: Complex and error-prone in general and conceptually not necessary here -
+   * we already know what the inputs will be during analysis, we just want to delay the required
+   * computations.</li>
+   * </ul>
+   */
+  @Override
+  public synchronized NestedSet<Artifact> getInputs() {
+    if (symlinkArtifacts == null) {
+      ImmutableList<Artifact> symlinks = runfiles.getArtifacts().toList().stream()
+          .filter(Artifact::isSymlink)
+          .collect(toImmutableList());
+      symlinkArtifacts = NestedSetBuilder.wrap(Order.STABLE_ORDER, symlinks);
+    }
+    return symlinkArtifacts;
   }
 
   @VisibleForTesting
@@ -227,7 +269,11 @@ public final class SourceManifestAction extends AbstractFileWriteAction {
         // This trailing whitespace is REQUIRED to process the single entry line correctly.
         manifestWriter.append(' ');
         if (symlink != null) {
-          manifestWriter.append(symlink.getPath().getPathString());
+          if (symlink.isSymlink()) {
+            manifestWriter.append(symlink.getPath().readSymbolicLink().getPathString());
+          } else {
+            manifestWriter.append(symlink.getPath().getPathString());
+          }
         }
         manifestWriter.append('\n');
       }
--- a/src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
@@ -1,0 +1,115 @@
+// Copyright 2022 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.devtools.build.lib.analysis.starlark;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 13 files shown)


src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -129,12 +138,48 @@ public SourceManifestAction(
...
+   * nested set of runfiles. Benchmarks confirmed that this does not regress performance.
+   *
+   * Alternatives considered:
+   * - Having users separate normal artifacts from symlink artifacts during analysis: Makes it
+   *   impossible to pass symlink artifacts to rules that aren't aware of them and requires the use
+   *   of custom providers to pass symlinks to stage as inputs to actions.
+   * - Reaching into {@link ActionExecutionContext} to look up the generating action of symlink
+   *   artifacts and retrieving the target from {@link UnresolvedSymlinkAction}: This would not work
+   *   for symlinks created by spawns.
+   * - Discovered inputs: ???
```
fmeum, line 163: Should I give input discovery a try or do you see a reason why it would perform worse?

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -129,12 +138,48 @@ public SourceManifestAction(
...
+      ImmutableList<Artifact> symlinks = runfiles.getArtifacts().toList().stream()
+          .filter(Artifact::isSymlink)
+          .collect(toImmutableList());
+      symlinkArtifacts = NestedSetBuilder.wrap(Order.STABLE_ORDER, symlinks);
+    }
+    return symlinkArtifacts;
+  }
+
+  @Override
+  public Depset getStarlarkInputs() {
```
fmeum, line 177: I can also make this throw an Exception if you prefer.

fmeum:
@lberki @alexjski

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
fmeum: Dropped it for now as per https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#r83921660.

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
fmeum: Incorporated the points explained in https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#r83921660.

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -129,12 +138,43 @@ public SourceManifestAction(
...
+   * {@link AbstractAction#getInputs()} and filters out the non-symlink runfiles by flattening the
+   * nested set of runfiles. Benchmarks confirmed that this does not regress performance.
+   *
+   * Alternatives considered:
+   * - Having users separate normal artifacts from symlink artifacts during analysis: Makes it
+   *   impossible to pass symlink artifacts to rules that aren't aware of them and requires the use
+   *   of custom providers to pass symlinks to stage as inputs to actions.
+   * - Reaching into {@link ActionExecutionContext} to look up the generating action of symlink
+   *   artifacts and retrieving the target from {@link UnresolvedSymlinkAction}: This would not work
+   *   for symlinks created by spawns.
```
lberki, line 162: nit "symlinks whose target is determined in the execution phase"

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -129,12 +138,43 @@ public SourceManifestAction(
...
+   * - Input discover: Complex and error-prone in general and conceptually not necessary here - we
```
lberki, line 163: nit: typo ("discover" -> "discovery")

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -129,12 +138,43 @@ public SourceManifestAction(
...
 
+  /**
+   * The manifest entry for a symlink artifact should contain the target of the symlink rather
+   * than its exec path. Reading the symlink target requires that the symlink artifact is declared
+   * as an input of this action. Since declaring all runfiles as inputs of the manifest action would
+   * unnecessarily delay its execution, this action exceptionally overrides
+   * {@link AbstractAction#getInputs()} and filters out the non-symlink runfiles by flattening the
+   * nested set of runfiles. Benchmarks confirmed that this does not regress performance.
+   *
+   * Alternatives considered:
```
lberki, line 156: nit: This is Javadoc, mind adding `<p>` and `<ul>` and `<li>` as appropriate?

src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Action to create a possibly unresolved symbolic link to a raw path.
+ *
+ * <p>To create a symlink to a known-to-exist target with alias semantics similar to a true copy of
+ * the input, use {@link SymlinkAction} instead.
+ */
+public final class UnresolvedSymlinkAction extends AbstractAction {
+  private static final String GUID = "0f302651-602c-404b-881c-58913193cfe7";
+
+  private final String target;
```
lberki, line 49: Why is this not a `PathFragment`? (it looks like it's a little bit more economical with GC if the action is invoked more than once during the lifetime of a server)

I mean, it should probably be a String because `PathFragment` is normalized and that's surprising, but it'd be a bit more involved to change that since we can currently only create symlinks from `PathFragment`s in the VFS layer, but for now, why not turn it into a `PathFragment` as early as possible?

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -716,4 +721,179 @@ EOF
...
 }
 
+function setup_unresolved_symlink_as_input() {
+  add_to_bazelrc build --experimental_allow_unresolved_symlinks
+
+  mkdir -p pkg
+  cat > pkg/def.bzl <<'EOF'
+def _r(ctx):
+  symlink = ctx.actions.declare_symlink(ctx.label.name + "_s")
+  ctx.actions.symlink(output = symlink, target_path = ctx.file.file.basename)
```
lberki, line 731: nit: why not make this an actual dangling symlink to better exercise the machinery of Bazel that does *not* resolve them when they are on the input of another action?

src/test/shell/bazel/bazel_symlink_test.sh
```diff
@@ -716,4 +721,179 @@ EOF
...
+  mkdir -p pkg
+  cat > pkg/def.bzl <<'EOF'
+def _r(ctx):
+  symlink = ctx.actions.declare_symlink(ctx.label.name + "_s")
+  target = ctx.file.file.basename
+  ctx.actions.symlink(output=symlink, target_path=target)
+
+  script = ctx.actions.declare_file(ctx.label.name)
+  content = """
+#!/usr/bin/env bash
```
lberki, line 819: Mind extracting this to an input file to avoid quoting shell quoting Starlark quoting shell again?

fmeum: I extracted it into a template for `ctx.actions.expand_template`. Is that what you meant?

src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
fmeum: Changed and moved up the comment. I intend to resolve this in a follow-up effort, but agree that using `PathFragment` is better for now. I would prefer to keep the `String` parameter of the constructor though as that's the API we ultimately want to realize.

src/test/shell/bazel/bazel_symlink_test.sh
fmeum: Whether or not the symlink is dangling here is configured via the `stage_target` attribute. The actual tests using this setup function go over all three cases (file not even built, file built but not declared as an input, file declared as an input). Do you want me to change this?

src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
lberki: As you desire. Given the limitation of VFS it doesn't matter much anyways.

src/test/shell/bazel/bazel_symlink_test.sh
lberki: I was confused for a moment when you mentioned `expand_template` but after looking the code, it makes total sense. I'd probably have used command line arguments instead of template expansion to inject the paths but that's probably just a matter of taste and the main issue of triple-quoting goes away in either case.

src/test/shell/bazel/bazel_symlink_test.sh
lberki: Oh, I see, now I understand why you did this. I thought that it would be better if the fact that the symlink is dangling was obvious from its target, but that would make it impossible to test all three cases so neatly. I'd probably have run one single action with the three kinds of inputs, but I don't think it's worth your time to change to that approach; I don't have an actual argument, it's just my taste, so meh.

lberki:
From my POV, this looks good; @alexjski , mind taking a look whether you can live with the way `getInputs()` is implemented now? If so, let's merge this and call this work well done.

src/main/java/com/google/devtools/build/lib/analysis/starlark/UnresolvedSymlinkAction.java
fmeum: I actually had that limitation removed on a branch and it wasn't so painful. I will revisit this, but would first want to understand the level of Windows support we can achieve.

gregmagolan:
Tested this change against rules_js here https://github.com/aspect-build/rules_js/pull/453 with a release on our bazel fork https://github.com/aspect-build/bazel/releases/tag/6.0.0-pre_20220823_1_aspect. Working as expected. Great work @fmeum!

fmeum:
> From my POV, this looks good; @alexjski , mind taking a look whether you can live with the way `getInputs()` is implemented now? If so, let's merge this and call this work well done.

Based on https://github.com/fmeum/bazel/commit/a2fdd4ad3963aec805d45ae1664a415bcdf2a3c3#r84014073, I would leave this PR as is then, tight?

lberki (approved):

lberki:
Yeah, let's merge this as it is then.

coeuvre:
cc @tjgq as this is relevant to our discussion about metadata handling for symlinks.

larsrc-google:
Going to give this a test against some internal benchmarks.

larsrc-google:
This seems to allow creating symlinks to anywhere, including outside of the Bazel tree. Do we really want that? Are there use cases that require that?

fmeum:
> This seems to allow creating symlinks to anywhere, including outside of the Bazel tree. Do we really want that? Are there use cases that require that?

This has been possible even before this commit and in fact happened more easily by accident since the symlink action created absolute symlinks that pointed outside the sandbox.

Generally speaking, I would expect most symlinks to be relative and stay within the tree. I do see use cases for absolute symlinks (e.g. LLVM ships symlinks to `/usr/bin/ld`).

lberki:
Sponsor of this change here! Yes, this allows creating symlinks to anywhere, but that, as @fmeum says, is not a new behavior; any action was able to do that before this change.

The reason why this is not a problem is that an "unresolved symlink" means that actions should only do `readlink()` and `lstat()` (but not `open()` or `stat()`) on the artifact. It's intended for things like creating file system images in the output tree, where it's perfectly valid to have a symlink `usr/bin/bazel -> /usr/bin/bazel-real` because the intent is not to chase the symlink, but to eventually package it up.

Of course, if an action breaks this contract, it's a problem. That's the same kind of issue as reading a file that's not a declared input of the action: not allowed by contract and if there is a sandbox, it may prevent doing that mistake.

larsrc-google:
Ok, thank you. I have gained further understanding now.

Issue #16044, opened 2022-08-04
Incompatible Target Skipping ignores visibility restrictions


philsc:
### Description of the bug:

With #14096, we're fixing a few usability issues with Incompatible Target Skipping. That is accomplished by moving the code to a much earlier stage of the analysis phase. Unfortunately, that means it runs before visibility is checked.

This is not _inherently_ a problem as it doesn't allow you to bypass visibility restrictions for building code. Compatible targets still obey visibility restrictions. And incompatible targets cannot be built.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create an incompatible target in package `//a`.
```python
# a/BUILD
filegroup(
    name = "private_filegroup",
    visibility = ["//visibility:private"],
    srcs = [...],
)
filegroup(
    name = "private_incompatible_filegroup",
    visibility = ["//visibility:private"],
    target_compatible_with = ["@platforms//:incompatible"],
    srcs = [...],
)
```
Create a target that is either directly incompatible or indirectly incompatible in a different package that depends on one of the private targets.
```python
# b/BUILD
filegroup(
    name = "directly_incompatible",
    srcs = [
        "//a:private_filegroup",
    ],
    target_compatible_with = ["@platforms//:incompatible"],
)
filegroup(
    name = "indirectly_incompatible",
    srcs = [
        "//a:private_incompatible_filegroup",
    ],
)
```
Neither `//b:directly_incompatible` nor `//b:indirectly_incompatible` will generate errors about visibility restrictions.

One possible course of action is to move the visibility checking earlier. https://github.com/bazelbuild/bazel/pull/14096#issuecomment-1189589626

### Which operating system are you running Bazel on?

x86_64 Debian 11

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

This is cbe25324ab but with #14096 merged in.

I'm filing the bug report preemptively.

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
$ git remote get-url upstream; git rev-parse master; git rev-parse HEAD
https://github.com/bazelbuild/bazel
e860453a7be19cc557d02b4edec13819358945fe
21a5fd1b5e091d4fb26ba683bfb3ff31d2f16621
```

### Any other information, logs, or outputs that you want to share?

The hope is to merge #14096 and separately work on this bug.

philsc:
[...]

PR #16274, opened 2022-09-14
Deduplicate incompatible `target_compatible_with` labels

philsc:
The latest refactor unintentionally made it so you can list the same
incompatible constraint multiple times in the `target_compatible_with`
attribute.

It was unintentional, but actually greatly simplifies a discussion
point on bazelbuild/bazel-skylib#381. That skylib patch aims to make
it easier to write non-trivial `target_compatible_with` expressions.
For example, to express that something is compatible with everything
except for Windows, one can use the following:

    foo_binary(
        name = "bin",
        target_compatible_with = select({
            "@platforms//os:windows": ["@platforms//:incomaptible"],
            "//conditions:default: [],
        }),
    )

The skylib patch aims to reduce that to:

    foo_binary(
        name = "bin",
        target_compatible_with = compatibility.none_of("@platforms//os:windows"),
    )

This works fine on its own, but a problem arises when these
expressions are composed. For example, a macro author might write the
following:

    def foo_wrapped_binary(name, target_compatible_with = [], **kwargs):
        foo_binary(
            name = name,
            target_compatible_with = target_compatible_with + select({
                "@platforms//os:none": ["@platforms//:incompatible"],
                "//conditions:default": [],
            }),
        )

A macro author should be able to express their own incompatibility
while also honouring user-defined incompatibility.

It turns out that my latest refactor (#14096) unintentionally made
this work. This happened because we now check for incompatibility
before we perform a lot of sanity checks. That's also one of the
reasons that visibility restrictions are not honoured for incomaptible
targets at the moment (#16044).

Without the deduplicating behaviour, macro authors are stuck with
either not allowing composition, or having to create unique
incompatible constraints for every piece in a composed
`target_compatible_with` expression.

This patch adds a test to validate the deduplicating behaviour to
cement it as a feature.

Unfortunately, this means that `target_compatible_with` behaves
differently from other label list attributes. For other label list
attributes, bazel complains when labels are duplicated. For example,
the following BUILD file:

    py_library(
        name = "lib",
    )

    py_binary(
        name = "bin",
        srcs = ["bin.py"],
        deps = [
        |   ":lib",
        |   ":lib",
        ],
    )

will result in the following error:

    $ bazel build :bin
    INFO: Invocation ID: 3d8345a4-2e76-493e-8343-c96a36185b52
    ERROR: /home/jenkins/repos/test_repo/BUILD:41:10: Label '//:lib' is duplicated in the 'deps' attribute of rule 'bin'
    ERROR: error loading package '': Package '' contains errors
    INFO: Elapsed time: 2.514s
    INFO: 0 processes.
    FAILED: Build did NOT complete successfully (1 packages loaded)


philsc:
@gregestren , @katre , at the risk of disappointing you with my poor decision making, what do you think of this patch?

src/main/java/com/google/devtools/build/lib/analysis/IncompatiblePlatformProvider.java
```diff
@@ -95,6 +105,8 @@ public boolean isImmutable() {
    *
    * <p>This may be null. If it is null, then {@code getTargetsResponsibleForIncompatibility()} is
    * guaranteed to be non-null. It will have at least one element in it if it is not null.
+   *
+   * <p>The list is sorted based on the toString() result of each constraint.
```
philsc, line 109: Probably just reword this to be "stringified label" rather than "toString()".

src/main/java/com/google/devtools/build/lib/analysis/IncompatiblePlatformProvider.java
```diff
@@ -69,7 +70,16 @@ public static IncompatiblePlatformProvider incompatibleDueToConstraints(
       @Nullable Label targetPlatform, ImmutableList<ConstraintValueInfo> constraints) {
     Preconditions.checkNotNull(constraints);
     Preconditions.checkArgument(!constraints.isEmpty());
-    return new AutoValue_IncompatiblePlatformProvider(targetPlatform, null, constraints);
+
+    // Deduplicate and sort the list of constraints. Doing it here means that everyone inspecting
+    // this provider doesn't have to deal with it.
+    ImmutableSet<ConstraintValueInfo> filteredConstraints = ImmutableSet.copyOf(constraints);
```
katre, line 76: It'd be easier to do this in one step (and to re-use the existing name, so that the old data can't be referenced):

```
constraints = constreams.stream()
    .sorted(Comparator.comparing(ConstraintValueInfo::label))
    .distinct()
    .collect(toImmutableList());

src/main/java/com/google/devtools/build/lib/analysis/IncompatiblePlatformProvider.java
katre: +1

katre (changes requested):
I think this makes a lot of sense, and there are probably more places in Bazel where we should deduplicate instead of erroring.

gregestren:
https://github.com/bazelbuild/bazel/commit/2570fe1119b2462ba10790d3bedf53699baecdf5 is a recent change recognizing some subtlety with `select()`. It's not the same purpose as yours, but a reminder that dedupe logic can't be exactly the same everwyhere.

This all reads reasonable to me but my main concern is API consistency. Surely this kind of theme has come up before w.r.t. macros? Let me run this by @comius and/or @brandjon , both of who I'll chat with tomorrow.

src/main/java/com/google/devtools/build/lib/analysis/IncompatiblePlatformProvider.java
philsc: Done

src/main/java/com/google/devtools/build/lib/analysis/IncompatiblePlatformProvider.java
philsc: Done

katre (approved):

katre:
Looks good to me, waitinf ro @gregestren, @comius, and @brandjon to weigh in.

comius (approved):

gregestren (approved):
Sounds like no concerns.

PR #16281, opened 2022-09-15
WIP: Add java_current_repository rule

fmeum:
Work towards #16124


fmeum:
@comius Could you review the overall design of this PR? The context is at https://github.com/bazelbuild/proposals/pull/274#issuecomment-1249285259 (and following comments). The general idea is simply to inject a static final constant into all Java targets without prohibitive overhead.

src/main/starlark/builtins_bzl/common/java/java_library.bzl
```diff
@@ -164,11 +179,21 @@ JAVA_LIBRARY_ATTRS = merge_attrs(
...
+        ),
         "_java_toolchain_type": attr.label(default = semantics.JAVA_TOOLCHAIN_TYPE),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
     },
 )
 
-java_library = rule(
+_java_library = rule(
```
comius, line 196: Can't change the name here, tooling uses it. Solution is to put `java_library` macro into a new file.

src/main/starlark/builtins_bzl/common/java/java_library.bzl
```diff
@@ -180,3 +205,10 @@ java_library = rule(
     compile_one_filetype = [".java"],
     toolchains = [semantics.JAVA_TOOLCHAIN],
 )
+
+def java_library(**kwargs):
+    _java_library(
+        # Skip over the leading '@'.
+        current_repository = repository_name()[1:],
```
comius, line 212: `repository_name()[1:]` creates a new string and will probably cause a memory regression, `repository_name()` is probably ok. If possible process it, that is remove `@` in an action.

src/main/java/com/google/devtools/build/lib/packages/License.java
```diff
@@ -53,7 +54,7 @@ public static class LicenseParsingException extends Exception {
    * <p>Note that the order is important for the purposes of finding the least
    * restrictive license.
    */
-  public enum LicenseType {
+  public enum LicenseType implements StarlarkValue {
```
comius, line 57: Why is this and other `StarlarkValue` needed?

src/main/java/net/starlark/java/eval/Starlark.java
```diff
@@ -206,6 +206,8 @@ public static Object fromJava(Object x, @Nullable Mutability mutability) {
       return StarlarkList.copyOf(mutability, (List<?>) x);
     } else if (x instanceof Map) {
       return Dict.copyOf(mutability, (Map<?, ?>) x);
+    } else if (x instanceof Set && ((Set<?>) x).size() == 1) {
```
comius, line 209: Why is this needed? We are usually more thorough on reviews of Starlark changes. If you need it, perhaps it's better to submit it in a separate PR.

comius (changes requested):
Quick and incomplete review, but I hope it gives you some inputs where you can continue on this PR.

src/main/starlark/builtins_bzl/common/java/java_library.bzl
fmeum: I can definitely change that, but don't we get the regression either way? `repository_name` is backed by [`getNameWithAt`](https://cs.opensource.google/bazel/bazel/+/b750f8c0242d7fcb581d368d8b75e59c51c13a61:src/main/java/com/google/devtools/build/lib/packages/StarlarkNativeModule.java;l=510), so the call will return a new string each time. The returned value isn't retained if we strip off the `@`.

tools/java/runfiles/java_current_repository.bzl
```diff
@@ -0,0 +1,115 @@
...
+def _current_repository_impl(ctx):
+    return _CurrentRepositoryInfo(value = ctx.build_setting_value)
+
+current_repository_setting = rule(
+    implementation = _current_repository_impl,
+    build_setting = config.string(),
+)
+
+def _java_current_repository_impl(ctx):
+    java_file = ctx.actions.declare_file("_java_current_repository/RunfilesConstants.java")
```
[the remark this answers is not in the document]
fmeum: That would be nicer, but I can't see it work: A given `java_binary` potentially bundles `java_library` targets from different repositories with different canonical repo names. How would each library look up its own repo name in a shared manifest?

fmeum:
@cushon As @comius has pointed out, the transition-based approach to generating the repo name Java constant currently has at least the following issues: 

1. Starlark transitions aren't really made to be applied to native rules, but not all Java rules will be starlarkified in time for Bazel 6.
2. The Starlark API for getting the repository name essentially builds in a memory regression as it prepends an `@`.
3. The transition and additional attributes on all Java targets could introduce memory regressions even if 2. is resolved.

Doing this in JavaBuilder seems much more efficient and much less complex (see https://github.com/fmeum/bazel/pull/new/16124-java-builder for a prototype). Do you think that could still be a viable alternative?

cushon:
> Doing this in JavaBuilder seems much more efficient and much less complex. Do you think that could still be a viable alternative?

I'm still not enthusiastic about doing it in JavaBuilder.

Has the [explicit rule per repository](https://github.com/bazelbuild/proposals/pull/274#issuecomment-1242794281) approach been ruled out? I know the ergonomics aren't anyone's first choice, but it's efficient and even less complex, which seems like an appealing combination if there's time pressure to get something into 6.0.0. Doing that doesn't rule out pursing one of the more complicated but easier to use options later.

I'm not asking you to spend more time on the JavaBuilder prototype before there's agreement on the approach, but there were some details I wondered about:

* are `list` and `inferBinaryName` really going to be the only methods that would need to know about the new custom file objects, or will this break other assumptions?
* does the prototype ensure the generated source isn't part of the output jar?
* did you consider writing a file to a tmp dir instead of a custom file object implementation, similar to the source jar handling in `setUpSourceJars`?
* did you consider using an annotation processor for this? If Blaze passed the workspace name to the compilation as an `-A` javacopt, I think it would be possible to do the code generation with an annotation processor. For example, we could write `class Foo { @AutoCurrentRepository private static final String CURRENT_REPOSITORY = AutoCurrentRepository_Foo.CURRENT_REPOSITORY; ... }`. The ergonomics are somewhere in between the explicit rule per repository and magically generating a file in JavaBuilder, and maintenance-wise it uses a standard, already supported mechanism to do the codegen.

fmeum:
Going to close this for now as https://github.com/bazelbuild/bazel/pull/16534 seems to strike a better balance of complexity vs. ergonomics.

PR #16283, opened 2022-09-15
Inject metadata when calling RemoteActionFileSystem#createSymbolicLink for a non-symlink artifact.

tjgq:


src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -101,7 +111,9 @@ void injectTree(SpecialArtifact tree, TreeArtifactValue metadata) {
   void injectFile(ActionInput file, RemoteFileArtifactValue metadata) {
     checkNotNull(metadataInjector, "metadataInject is null");
 
-    injectedMetadata.put(file.getExecPath(), metadata);
+    if (metadata instanceof RemoteFileArtifactValue) {
```
coeuvre, line 114: always true?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -227,11 +239,44 @@ protected PathFragment readSymbolicLink(PathFragment path) throws IOException {
...
+                entry.getValue());
+          }
+          injectTree((SpecialArtifact) output, treeBuilder.build());
+        }
+      } else if (!output.isSymlink()) {
+        RemoteFileArtifactValue metadata = getRemoteInputMetadata(targetFragment);
+        if (metadata != null) {
+          injectFile(output, metadata);
+        }
+      }
```
coeuvre, line 277: What about output that is symlink? I know we are going to handle it differently, but should we throw an error here?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -227,11 +239,44 @@ protected PathFragment readSymbolicLink(PathFragment path) throws IOException {
...
+    FileStatus stat = statNullable(linkPath, /* followSymlinks= */ false);
+    if (stat != null) {
+      throw new IOException(linkPath + " (File exists)");
+    }
+
+    // If the output is a regular file or tree artifact, it has "copy" semantics, i.e., it will be
+    // transparently resolved into the artifact it points to; the symlink is just a filesystem-level
+    // optimization to avoid making an actual copy. We can inject its metadata and avoid the need to
+    // compute it from the filesystem later, which would be slower and require additional filesystem
+    // operations to be implemented.
```
coeuvre, line 251: Potentially, we could download the same files twice. e.g. symlink `a` points to file `b`, but the metadata are same from the perspective of skyframe.

When staging inputs for actions that requires `a`, we download `a` and create files instead of symlinks on the local file system. Then we do the same thing for actions that requires `b`. It is getting worse if the target is directory. This is another reason why we currently creating a dangling symlink for the internal equivalent.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -370,10 +415,22 @@ private RemoteFileArtifactValue getRemoteInputMetadata(PathFragment path) {
...
   }
 
+  @Nullable
+  private TreeArtifactValue getRemoteInputTreeMetadata(PathFragment path) {
+    if (!path.startsWith(outputBase)) {
+      return null;
+    }
+    PathFragment execPath = path.relativeTo(execRoot);
+    TreeArtifactValue m = inputArtifactData.getTreeMetadata(execPath);
+    if (m != null && m.isEntirelyRemote()) {
```
coeuvre, line 428: Do we have the case where the tree is not entirely remote? If so, how do we handle that?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: Oops, it was a leftover from a previous refactor.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: If the output is a symlink, then we should create the symlink and not inject any metadata. Skyframe will later call `readSymbolicLink` and not attempt to resolve it.

Unfortunately, unresolved symlinks don't currently work together with remote execution because of other issues, so I can't add an integration test for it (I did add a unit test to verify that `createSymbolicLink` and `readSymbolicLink` round-trip when the output is of type symlink).

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: No, I don't believe this could ever occur, since two actions cannot produce the same tree artifact. So I think `!isEntirelyRemote` effectively means "is entirely local".

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: That makes sense. Do we allow actions to access undeclared output (i.e. create undeclared symlink in this case)?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: True.

src/test/shell/bazel/remote/build_without_the_bytes_test.sh
```diff
@@ -726,6 +726,182 @@ EOF
     //a:foo || "Failed to build //a:foo"
 }
 
+function test_downloads_minimal_symlink_to_source_file() {
```
coeuvre, line 729: Optional but please consider moving these tests to java.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: Per offline discussion, I've added a `stagingExecPath` field to `RemoteFileArtifactValue`, which is used by the input prefetcher to avoid downloading the same file more than once. This is a bit of a hack, but the alternatives all seem excessively complex to me.

src/test/shell/bazel/remote/build_without_the_bytes_test.sh
tjgq: I will move them once we agree on the overall approach.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Let me write down what we have discussed just now:
- For file, the `stagingExecPath` is used to store the path of final target file. We might want to rename it to a better name to reflect the fact that this remote metadata is actually a symlink to another remote metadata. Alternative way is to create a subclass of `RemoteFileArtifactValue`. I'm fine with either way.
- For tree, we don't want to store `stagingExecPath` in remote metadata for every `TreeFileArtifact`. We can store it in an extra field of `TreeArtifactValue` or in a subclass.
- We resolve symlink chains to the last target when injecting the metadata in the `createSymbolicLink`. So when we staging the symlink in the prefetcher, we know the `execPath` of final target (This is what you are doing now, just write it down for clarification).
- For symlinks whose targets are on the local file system (i.e. remote metadata for the target is `null`), we don't inject metadata but just create symlinks on the local files system.
- We need to update the serialization/deserialization code for remote metadata in https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java.
- We are interested in the test case: `sym` points to a `dir/file` (by `declare_file`) inside a `dir` (by `declare_dir`)
  - are we able to deduplicate the downloads of `dir/file`?

I might miss some other points but you get the idea.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -539,16 +542,27 @@ public static class RemoteFileArtifactValue extends FileArtifactValue {
     private final long size;
     private final int locationIndex;
     private final String actionId;
-
-    public RemoteFileArtifactValue(byte[] digest, long size, int locationIndex, String actionId) {
+    @Nullable private final PathFragment symlinkTargetExecPath;
```
alexjski, line 545: That may not be that useful internally. Both this and the actionid, please check the impact on memory usage due to that.

src/main/java/com/google/devtools/build/lib/remote/AbstractActionInputPrefetcher.java
```diff
@@ -182,10 +218,15 @@ private Completable prefetchInputTree(
...
                   HashSet<Path> dirs = new HashSet<>();
 
-                  // Tree root is created by Bazel before action execution, but the permission is
-                  // changed to 0555 afterwards. We need to set it as writable in order to move
-                  // files into it.
-                  treeRoot.setWritable(true);
+                  if (!treeRoot.exists()) {
+                    // If the tree root doesn't already exist, create it now. This can happen if
+                    // we're prefetching to a different location and symlinking into it.
+                    treeRoot.createDirectory();
```
alexjski, line 224: Consider using [createWritableDirectory](https://github.com/bazelbuild/bazel/blob/d44d4ea960fdc7260977e69b530e053baebf7fb8/src/main/java/com/google/devtools/build/lib/vfs/Path.java#L436-L448) to get it all done in 1 call.

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
```diff
@@ -233,6 +245,11 @@ public Optional<ArchivedRepresentation> getArchivedRepresentation() {
     return Optional.ofNullable(archivedRepresentation);
   }
 
+  /** Return symlinkTarget path (if present). */
+  public Optional<PathFragment> getSymlinkTargetExecPath() {
```
alexjski, line 249: As we talked offline, you may be better off returning a FileMetadataValue from `getMetadata` which includes that so you can just request that using `metadataProvider.getMetadata(treeArtifact)`.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: Yes, undeclared symlinks are created via `ctx.actions.symlink(output = declare_symlink(...))` which works today. They can also be consumed as inputs to other actions.

I suspect we should also inject the metadata for undeclared symlinks into Skyframe (`UndeclaredSymlinkFileArtifactValue` seems to have been introduced for that exact purpose, but is mysteriously unused in the codebase). However, I don't want to expand the scope of this PR to include it, so for now I'm still creating an actual symlink in the local filesystem.

src/test/shell/bazel/remote/build_without_the_bytes_test.sh
tjgq: Moved.

src/main/java/com/google/devtools/build/lib/remote/AbstractActionInputPrefetcher.java
tjgq: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: PTAL. Some points of note:

- Naming: I've settled on `symlinkTargetExecPath` instead of `stagingExecPath`. I think that's as explicit as it gets :)
- Skyframe values: I've added a `symlinkTargetExecPath` field to the existing `TreeArtifactValue` (whose size remains at 40 bytes due to padding) and introduced a new`SymlinkToRemoteFileArtifactValue` subclass (to avoid increasing the size of `RemoteFileArtifactValue`). Following a suggestion by Alex, the metadata for tree artifacts is retrieved via `TreeArtifactValue#getMetadata`, which avoids adding a `MetadataProvider#getTreeArtifactValue` method.
- Nested artifacts: I've added tests for them in `BuildWithoutTheBytesIntegrationTest`, but as we discussed, the prefetcher currently isn't able to deduplicate the downloads (and the downloads might even collide). Since this is a preexisting issue and nested artifacts are relatively rare, I'd prefer to fix it outside of this PR. It has been filed as #16333.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
tjgq: Introduced a separate `SymlinkToRemoteFileArtifact` subclass to avoid increasing the size. For `TreeArtifactValue`, we got lucky and the new field fits within the preexisting padding, so no subclass is required.

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
tjgq: Thanks, this made everything so much simpler!

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -227,14 +237,76 @@ protected PathFragment readSymbolicLink(PathFragment path) throws IOException {
...
+
+  private TreeArtifactValue createTreeArtifactValueForSymlink(
+      SpecialArtifact output, TreeArtifactValue original, PathFragment targetPath) {
+    TreeArtifactValue.Builder builder = TreeArtifactValue.newBuilder(output);
+    for (Map.Entry<TreeFileArtifact, FileArtifactValue> entry :
+        original.getChildValues().entrySet()) {
+      PathFragment parentRelativePath = entry.getKey().getParentRelativePath();
+      builder.putChild(
+          TreeFileArtifact.createTreeOutput(output, parentRelativePath),
+          createRemoteFileArtifactValueForSymlink(
```
coeuvre, line 301: IDK but do we want store symlinks for each child file? What could be wrong if we don't do that?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Sorry for didn't make it clear, but my question was about **undeclared** output not **unresolved** symlink. To be more specific, do we want to move line `super.createSymbolicLink()` inside the check `if (output != null)` and instead throw an error there to prevent actions create **undeclared** outputs?

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
```diff
@@ -190,23 +192,35 @@ public static ArchivedRepresentation create(
...
   @Nullable private final ArchivedRepresentation archivedRepresentation;
 
+  /**
+   * Optional symlink target path.
+   *
+   * <p>If set, this artifact is a copy of another artifact. It is materialized in the local
+   * filesystem as a symlink to the original artifact, whose contents live at this location. This is
+   * used to implement zero-cost copies of remotely stored artifacts.
+   */
+  @Nullable private final PathFragment symlinkTargetExecPath;
```
coeuvre, line 202: I think you measured the memory impact for adding this new field. I'm curious about the numbers.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -931,7 +936,8 @@ public void saveOutputMetadata_treeMetadataWithSameLocalFileMetadata_cached() th
     ActionCache.Entry entry = cache.get(output.getExecPathString());
     assertThat(entry).isNotNull();
     assertThat(entry.getOutputTree(output))
-        .isEqualTo(SerializableTreeArtifactValue.create(children, Optional.empty()));
+        .isEqualTo(
+            SerializableTreeArtifactValue.create(children, Optional.empty(), Optional.empty()));
```
coeuvre, line 940: Add a test case where ActionCacheCheck is able to hit the cache for actions that produce remote tree/file symlinks?

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -466,6 +467,14 @@ private static void encodeRemoteMetadata(
     VarInt.putVarInt(value.getLocationIndex(), sink);
 
     VarInt.putVarInt(indexer.getOrCreateIndex(value.getActionId()), sink);
+
+    Optional<PathFragment> symlinkTargetExecPath = value.getSymlinkTargetExecPath();
```
coeuvre, line 471: Please update `VERSION` number in this file so that we won't read old content.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Totally make sense. Thanks!

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: I think it's reasonable to throw if the symlink doesn't point to an output, yes. This filesystem is only meant to be used by internal actions, and the set of things they can do is fairly restricted.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -626,6 +749,58 @@ public String toString() {
...
           .add("size", size)
           .add("locationIndex", locationIndex)
+          .add("actionId", actionId)
+          .toString();
+    }
+  }
+
+  /** An artifact that should be materialized as a symlink to another remote artifact. */
+  public static class SymlinkToRemoteFileArtifactValue extends RemoteFileArtifactValue {
+    private PathFragment symlinkTargetExecPath;
```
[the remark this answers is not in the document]
tjgq: Done.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -172,6 +173,17 @@ protected boolean couldBeModifiedByMetadata(FileArtifactValue lastKnown) {
     return true;
   }
 
+  /**
+   * Optional symlink target path.
+   *
+   * <p>If present, this artifact is a copy of another artifact. It is materialized in the local
```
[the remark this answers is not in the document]
tjgq: In the future I'd also like to store the (unresolved) symlink target in Skyframe. As I mentioned elsewhere, there's an `UnresolvedSymlinkArtifactValue` class which was created for this exact purpose [1], but we currently only use it in `ActionMetadataHandler` [2].

[1] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java;l=634;drc=de02e3f62598dbe8dcff68b4040f34da3322c720
[2] https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/skyframe/ActionMetadataHandler.java;l=606;drc=5ee37a6b8b7c80db579d0246e2ea88142213b6d8

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -578,6 +598,15 @@ private static byte[] encode(StringIndexer indexer, ActionCache.Entry entry) thr
       } else {
         VarInt.putVarInt(0, sink);
       }
+
+      Optional<PathFragment> symlinkTargetExecPath =
```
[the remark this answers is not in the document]
tjgq: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: I think it's safe to omit them now.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
tjgq: Done.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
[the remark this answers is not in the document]
coeuvre: `maxOutputFilesSize` is also needed to be updated.

src/test/java/com/google/devtools/build/lib/remote/BuildWithoutTheBytesIntegrationTest.java
```diff
@@ -360,6 +360,292 @@ public void changeOutputMode_invalidateActions() throws Exception {
     }
   }
 
+  @Test
+  public void symlinkToSourceFile() throws Exception {
```
coeuvre, line 364: Add a test case that symlink are not created on the local file system if no local actions need them?

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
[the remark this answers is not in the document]
tjgq: Oops, should be fixed now.

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
tjgq: In the end, I didn't benchmark; I just made sure that I'm not increasing the size of `TreeArtifactValue` nor `RemoteFileArtifactValue`, which means the change will be neutral for builds that don't use `ctx.actions.symlink(output = declare_file/declare_directory)`.

(I think there's a way to make it truly neutral by using different specializations of `RemoteFileArtifactValue` for `actionId` and `locationIndex`, since AIUI their presence is mutually exclusive. But I'd rather avoid doing that now if I can.)

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
tjgq: Done.

In passing, I've just realized that the cache checking is a little odd for artifacts with a `symlinkTargetExecPath`: we consider the local filesystem to be up to date if the artifact is materialized by an actual file/directory instead of a symlink pointing to one.

I think this is benign and consistent with how symlinks are transparently followed throughout Bazel; it's a consequence of the fact that the artifacts are declare_file/declare_directory (not declare_symlink) and hence what we really care about is their contents.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -172,6 +173,17 @@ protected boolean couldBeModifiedByMetadata(FileArtifactValue lastKnown) {
...
+   * filesystem as a symlink to the original artifact, whose contents live at this location. This is
+   * used to implement zero-cost copies of remotely stored artifacts.
+   */
+  public Optional<PathFragment> getSymlinkTargetExecPath() {
```
alexjski, line 183: This is out of sync with the `SourceFileArtifactValue`. If that is intended to be short lived, maybe let's move that to a separate interface and case in remote. If that is long lived, it may be worth to keep existing code in sync.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -172,6 +173,17 @@ protected boolean couldBeModifiedByMetadata(FileArtifactValue lastKnown) {
     return true;
   }
 
+  /**
+   * Optional symlink target path.
```
alexjski, line 177: Please add javadoc to explain that this is for optimizations only, maybe even rename the method to indicate that. This API should clearly state that it is an implication only: non-empty response means that this is a symlink, but the opposite may not be true.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -466,6 +467,14 @@ private static void encodeRemoteMetadata(
     VarInt.putVarInt(value.getLocationIndex(), sink);
 
     VarInt.putVarInt(indexer.getOrCreateIndex(value.getActionId()), sink);
+
+    Optional<PathFragment> symlinkTargetExecPath = value.getSymlinkTargetExecPath();
+    if (symlinkTargetExecPath.isPresent()) {
+      VarInt.putVarInt(1, sink);
```
alexjski, line 473: nit: Another option is to put a value which cannot be a correct index (e.g. -1) when it is absent. That may ironically be more than 1 byte for -1 (since varint is 1 byte for 0..127). You could reserve 0 and do index + 1, but it is a little weird.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
```diff
@@ -497,6 +517,7 @@ private static byte[] encode(StringIndexer indexer, ActionCache.Entry entry) thr
     int maxOutputFilesSize =
         VarInt.MAX_VARINT_SIZE // entry.getOutputFiles().size()
             + (VarInt.MAX_VARINT_SIZE // execPath
+                    + (2 * VarInt.MAX_VARINT_SIZE) // symlinkTargetExecPath optional
```
alexjski, line 520: 0 vs 1 you know is 1 byte, so it's more more like 1 + VarInt.MAX_VARINT_SIZE

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
```diff
@@ -190,23 +192,35 @@ public static ArchivedRepresentation create(
...
     this.childData = childData;
     this.totalChildSize = totalChildSize;
     this.archivedRepresentation = archivedRepresentation;
+    this.symlinkTargetExecPath = symlinkTargetExecPath;
     this.entirelyRemote = entirelyRemote;
   }
 
   public FileArtifactValue getMetadata() {
-    return FileArtifactValue.createProxy(digest);
+    return FileArtifactValue.createProxyForTreeArtifact(
```
alexjski, line 222: Given you have a public method for the symlink execpath, I wonder if that would not better be a private static class in here or even an anonymous class here. This is closely tied to the tree artifact, yet the code lives quite far from each other.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -616,6 +624,30 @@ public void saveOutputMetadata_localMetadataIsSameAsRemoteMetadata_cached() thro
     assertThat(entry.getOutputFile(output)).isEqualTo(createRemoteFileMetadata(content));
   }
 
+  @Test
+  public void saveOutputMetadata_localMetadataIsSameAsRemoteMetadata_withTargetExecPath_cached()
```
alexjski, line 628: I would recommend parameterizing the test case above: `@TestParameter({"null", "/target/path"}} String symlinkTargetPath)`.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
```diff
@@ -655,15 +687,16 @@ public void saveOutputMetadata_treeMetadata_remoteFileMetadataSaved() throws Exc
             "file2", createRemoteFileMetadata("content2"));
     Action action =
         new InjectOutputTreeMetadataAction(
-            output, createTreeMetadata(output, children, Optional.empty()));
+            output, createTreeMetadata(output, children, Optional.empty(), Optional.empty()));
```
alexjski, line 690: nit: some `/*foo=*/` comments would aid readability here for the `Optional.empty()`.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
```diff
@@ -134,6 +134,10 @@ public void putTreeArtifact_addsTreeArtifactAndAllChildren() {
     assertContainsEqualMetadata(tree, treeValue.getMetadata());
     assertContainsSameInstance(child1, child1Metadata);
     assertContainsSameInstance(child2, child2Metadata);
+
+    assertThat(map.getMetadata(tree)).isEqualTo(treeValue.getMetadata());
```
alexjski, line 138: That is checked in line 133, delete.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
```diff
@@ -134,6 +134,10 @@ public void putTreeArtifact_addsTreeArtifactAndAllChildren() {
     assertContainsEqualMetadata(tree, treeValue.getMetadata());
     assertContainsSameInstance(child1, child1Metadata);
     assertContainsSameInstance(child2, child2Metadata);
+
```
alexjski, line 137: Remove the nl.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
```diff
@@ -134,6 +134,10 @@ public void putTreeArtifact_addsTreeArtifactAndAllChildren() {
...
+    assertThat(map.getMetadata(tree)).isEqualTo(treeValue.getMetadata());
+    assertThat(map.getTreeMetadata(tree.getExecPath())).isEqualTo(treeValue);
```
alexjski, line 139: Consider adding test cases specific to the new API rather than tagging along a single test. It's likely easy to parameterize as well e.g. tree with some files, empty tree etc.

alexjski (changes requested):

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
tjgq: Done in 5faf917.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
tjgq: Done in 8d76ef1.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
tjgq: Done in 8d76ef1.

src/test/java/com/google/devtools/build/lib/actions/ActionInputMapTest.java
tjgq: Done in 8d76ef1.

src/test/java/com/google/devtools/build/lib/actions/ActionCacheCheckerTest.java
tjgq: Done.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
tjgq: I think using 0 and 1 is fine for what's effectively a presence bit.

src/main/java/com/google/devtools/build/lib/actions/cache/CompactPersistentActionCache.java
tjgq: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -285,11 +301,47 @@ protected PathFragment readSymbolicLink(PathFragment path) throws IOException {
...
+          injectRemoteFile(childExecPath, childMetadata.getDigest(), childMetadata.getSize(), childMetadata.getActionId());
+        }
+        symlinkToTargetExecPath.put(linkPath, metadata.getSymlinkTargetExecPath().orElse(targetFragment.relativeTo(execRoot)));
+        return;
+      }
+    } else if (!output.isSymlink()) {
+      RemoteFileArtifactValue metadata = getRemoteInputMetadata(targetFragment);
+      if (metadata != null) {
+        checkState(targetFragment.isAbsolute());
+        injectRemoteFile(linkPath, metadata.getDigest(), metadata.getSize(), metadata.getActionId());
```
coeuvre, line 337: Since we have a real fs now, can we instead just create the symlink on the in-memory fs? same for tree symlink

src/main/java/com/google/devtools/build/lib/skyframe/TreeArtifactValue.java
tjgq: From our offline discussion, it looks like we might be able to get rid of `actionId` entirely, so that would be preferred.

src/test/java/com/google/devtools/build/lib/remote/BuildWithoutTheBytesIntegrationTest.java
tjgq: I believe this is obsoleted by the fact that `RemoteActionFileSystem#createSymbolicLink` now creates the symlink on both filesystems. This seems unavoidable since `RemoteActionFileSystem` is also used in contexts where modifying the local filesystem is expected (it might be possible to move this to a different filesystem implementation, but not today).

src/main/java/com/google/devtools/build/lib/actions/ActionInputMap.java
```diff
@@ -244,6 +256,11 @@ private FileArtifactValue getMetadataFromTreeArtifacts(PathFragment execPath) {
     return entry != null ? entry.getValue() : null;
   }
 
+  @Nullable
+  public TreeArtifactValue getTreeMetadata(PathFragment execPath) {
+    return treeArtifactsRoot.findTreeArtifactNode(execPath);
```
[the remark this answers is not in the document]
tjgq: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
tjgq: Done.

The changes to `OutputService` and `ActionExecutionFunction` currently included in this PR will be submitted separately as an internal CL.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
alexjski: Makes sense, please consider naming that more specifically to make sure that is clear why this is separate, e.g. `getSymlinkTargetPathForLocalPrefetching` (I am confident you can come up with something better than that).

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -533,12 +555,100 @@ protected boolean couldBeModifiedByMetadata(FileArtifactValue o) {
     }
   }
 
+  private static final class TreeArtifactProxyFileArtifactValue extends FileArtifactValue {
```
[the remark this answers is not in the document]
tjgq: As discussed offline: I don't fully understand the reasons *why* this exists, so I don't think I can do better than the current javadoc.

Regarding the class name: it doesn't make sense to include `Symlink` in the name, since every tree artifact has one of these values associated with it, regardless of whether the `symlinkTargetExecPath` field is set.

(Maybe you meant separate classes for when the symlink is or isn't present? I'm not convinced that pulls its weight; unlike the `RemoteFileArtifactValue` case, there's no need to optimize for memory footprint here.)

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
```diff
@@ -533,12 +555,100 @@ protected boolean couldBeModifiedByMetadata(FileArtifactValue o) {
...
+    public long getModifiedTime() {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add(
+              "digest",
+              digest == null ? "(null)" : BaseEncoding.base16().lowerCase().encode(digest))
```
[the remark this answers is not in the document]
tjgq: It's not nullable. Done.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
tjgq: Done. I think the new name ("materializationExecPath") is distinctive enough to avoid confusion, and the documentation clearly states that it's only used by the prefetcher.

I'm not so sure about the second part of your comment. The artifacts that correspond to these values are *not* symlinks, as far as Bazel is concerned; they're only materialized as such in the local filesystem (which is not even a novel concept - we do the same for sandbox inputs, for example). I don't want folks to erroneously infer that they are, especially considering that there's a different kind of artifact that *is* tracked by Bazel as one.

src/main/java/com/google/devtools/build/lib/actions/FileArtifactValue.java
tjgq: I changed the name to "materialization" which I think is distinctive enough to cause readers to pay attention; it also avoids a potential confusion with unresolved symlinks, which are entirely separate. (I couldn't bring myself to include all of "symlink", "exec path" and "materialization"/"prefetching" in the name, sorry; it's already uncomfortably long as it is.)

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -157,6 +162,91 @@ void flush() throws IOException {
...
+          metadata.getChildValues().entrySet()) {
+        TreeFileArtifact child =
+            TreeFileArtifact.createTreeOutput(parent, entry.getKey().getParentRelativePath());
+        RemoteFileArtifactValue childMetadata = (RemoteFileArtifactValue) entry.getValue();
+        injectedTree.putChild(child, childMetadata);
+      }
+
+      metadataInjector.injectTree(parent, injectedTree.build());
+    } else {
+      RemoteFileArtifactValue metadata = getRemoteInputMetadata(targetPath);
```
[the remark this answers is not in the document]
tjgq: I don't think we ever create symlinks to outputs in any of the scenarios where a `RemoteActionFileSystem` is currently used, but it doesn't hurt to allow it. Generally speaking, I'd like this class to trend towards a fully functional filesystem, instead of implementing just the bits internal actions happen to require.

In that vein, `getRemoteInputTreeMetadata` should also fall back to `remoteOutputTree`; I think this would matter if, for example, someone were to call `renameTo` on a directory. However, this isn't necessary for symlinks, so I just left a TODO there for now.

I've removed `Input` from both names to make the purpose clear.

coeuvre (approved):

Issue #16041, opened 2022-08-03
`UnsatisfiedLinkError` with `--experimental_remote_cache_compression` on Windows


fmeum:
### Description of the bug:

Remote builds with `--experimental_remote_cache_compression` crash Bazel on Windows with:

```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.UnsatisfiedLinkError: 'long com.github.luben.zstd.ZstdOutputStreamNoFinalizer.recommendedCOutSize()'
	at com.github.luben.zstd.ZstdOutputStreamNoFinalizer.recommendedCOutSize(Native Method)
	at com.github.luben.zstd.ZstdOutputStreamNoFinalizer.<clinit>(ZstdOutputStreamNoFinalizer.java:29)
	at com.github.luben.zstd.ZstdOutputStream.<init>(ZstdOutputStream.java:73)
	at com.github.luben.zstd.ZstdOutputStream.<init>(ZstdOutputStream.java:53)
	at com.google.devtools.build.lib.remote.zstd.ZstdCompressingInputStream.<init>(ZstdCompressingInputStream.java:52)
	at com.google.devtools.build.lib.remote.zstd.ZstdCompressingInputStream.<init>(ZstdCompressingInputStream.java:42)
	at com.google.devtools.build.lib.remote.Chunker.maybeInitialize(Chunker.java:254)
	at com.google.devtools.build.lib.remote.Chunker.seek(Chunker.java:153)
	at com.google.devtools.build.lib.remote.ByteStreamUploader$AsyncUpload.lambda$start$0(ByteStreamUploader.java:417)
	at com.google.devtools.build.lib.remote.Retrier.executeAsync(Retrier.java:277)
	at com.google.devtools.build.lib.remote.ByteStreamUploader$AsyncUpload.lambda$start$1(ByteStreamUploader.java:411)
	at com.google.devtools.build.lib.remote.util.Utils.refreshIfUnauthenticatedAsync(Utils.java:491)
	at com.google.devtools.build.lib.remote.ByteStreamUploader$AsyncUpload.start(ByteStreamUploader.java:409)
	at com.google.devtools.build.lib.remote.ByteStreamUploader.startAsyncUpload(ByteStreamUploader.java:343)
	at com.google.devtools.build.lib.remote.ByteStreamUploader.uploadBlobAsync(ByteStreamUploader.java:261)
[...]

PR #16293, opened 2022-09-17
fix zstd on windows

cha5on:
Sandboxing on Windows is more limited than on other platforms, so two jni_md.h headers were ending up in the include path.  This resulted in the wrong jni_md.h header being used on Windows, which meant that the JNIEXPORT macro was not being set to actually export symbols appropriately on Windows.  This caused any attempts to use zstd on windows, whether for http_archive, gRPC cache compression, etc. to result in an UnsatisfiedLinkError being thrown.

Change to use a similar method as the main bazel jni stuff, which is to copy the jni.h and jni_md.h headers over.  This prevents accidentally picking up a header from the wrong architecture.

Relates to #16041

### DIFF
--- a/third_party/zstd-jni/zstd-jni.BUILD
+++ b/third_party/zstd-jni/zstd-jni.BUILD
@@ -1,21 +1,35 @@
+genrule(
+    name = "copy_link_jni_md_header",
+    srcs = select({
+        "@bazel_tools//src/conditions:darwin": ["@bazel_tools//tools/jdk:jni_md_header-darwin"],
+        "@bazel_tools//src/conditions:freebsd": ["@bazel_tools//tools/jdk:jni_md_header-freebsd"],
+        "@bazel_tools//src/conditions:openbsd": ["@bazel_tools//tools/jdk:jni_md_header-openbsd"],
+        "@bazel_tools//src/conditions:windows": ["@bazel_tools//tools/jdk:jni_md_header-windows"],
+        "//conditions:default": ["@bazel_tools//tools/jdk:jni_md_header-linux"],
+    }),
+    outs = ["jni_md.h"],
+    cmd = "cp -f $< $@",
+)
+
+genrule(
+    name = "copy_link_jni_header",
+    srcs = ["@bazel_tools//tools/jdk:jni_header"],
+    outs = ["jni.h"],
+    cmd = "cp -f $< $@",
+)
+
 cc_binary(
     name = "libzstd-jni.so",
     srcs = glob([
         "src/main/native/**/*.c",
         "src/main/native/**/*.h",
-    ]) + select({
-        "@io_bazel//src/conditions:windows": [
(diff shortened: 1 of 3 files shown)


third_party/zstd-jni/zstd-jni.BUILD
```diff
@@ -1,18 +1,32 @@
+genrule(
+    name = "copy_link_jni_md_header",
+    srcs = select({
+        "@io_bazel//src/conditions:darwin": ["@bazel_tools//tools/jdk:jni_md_header-darwin"],
```
meteorcloudy, line 4: `@bazel_tools//src/conditions:*` is better.

meteorcloudy (changes requested):

third_party/zstd-jni/zstd-jni.BUILD
```diff
@@ -1,18 +1,32 @@
...
     name = "libzstd-jni.so",
     srcs = glob([
         "src/main/native/**/*.c",
         "src/main/native/**/*.h",
-    ]) + select({
-        "@io_bazel//src/conditions:windows": [
-            "src/windows/include/jni_md.h",
-            "jni/jni.h",
+    ]) + [
+            "jni_md.h",
```
coeuvre, line 27: prepend `:`? e.g. `:jni_md.h`

meteorcloudy:
We should try to enable the tests on Windows as well: https://cs.opensource.google/bazel/bazel/+/master:.bazelci/presubmit.yml;l=244-245

coeuvre (changes requested):

third_party/zstd-jni/zstd-jni.BUILD
gregoryT5: updated throughout

third_party/zstd-jni/zstd-jni.BUILD
gregoryT5: done

third_party/zstd-jni/zstd-jni.BUILD
cha5on: updated throughout

third_party/zstd-jni/zstd-jni.BUILD
cha5on: done

meteorcloudy:
Can you also update https://cs.opensource.google/bazel/bazel/+/master:.bazelci/postsubmit.yml;l=252 to the same as the presubmit.yml file?

coeuvre (approved):
LGTM. Thanks!

cha5on:
> Can you also update https://cs.opensource.google/bazel/bazel/+/master:.bazelci/postsubmit.yml;l=252 to the same as the presubmit.yml file?

done!

meteorcloudy (approved):
Thank you so much for the fix!

Issue #16151, opened 2022-08-23
`--incompatible_remote_build_event_upload_respect_no_cache` still uploads some `no-cache` outputs


brentleyjones:
### Description of the bug:

Some test target outputs are still uploaded to the cache even when `--incompatible_remote_build_event_upload_respect_no_cache` and `--modify_execution_info='.*=+no-remote'` are set. rules_apple tests upload the binary and the zip, and `swift_test` has the runner uploaded (which it shouldn't), but not the binary...

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

The following will repro the bug:
```
git clone https://github.com/buildbuddy-io/rules_xcodeproj.git

cd rules_xcodeproj

# Modify `examples/command_line/Tests/SwiftGreetingsTests.swift` adding something like
# `public let cache = "123"` and changing as needed to produce a new end output

bazel build --config=cache --modify_execution_info='.*=+no-remote' //examples/command_line/Tests:LibSwiftTests
```

Resulting in bytestream/upload for the test outputs:
```
        "importantOutput": [
            {
                "name": "examples/command_line/Tests/LibSwiftTests",
                "uri": "bytestream://remote.buildbuddy.io/blobs/593de6d72a4f278017c499cffe5cee734144740cffdc7a4cbea2b69eb2d712de/7289",
                "pathPrefix": [
                    "bazel-out",
                    "darwin_arm64-fastbuild",
                    "bin"
                ]
            },
            {
                "name": "examples/command_line/Tests/LibSwiftTests.zip",
                "uri": "bytestream://remote.buildbuddy.io/blobs/a1f751591b0fb98ddd985dbf5bd957e59dd76b5c5ce61ccb7bd6030a400da02d/119243",
                "pathPrefix": [
                    "bazel-out",
                    "applebin_macos-darwin_arm64-fastbuild-ST-3a19c795fefb",
                    "bin"
                ]
            }
        ]
```

The following will somewhat repro the bug:
```
# Modify `tools/generator/test/AddTargetsTests.swift` adding something like
# `public let cache = "123"` and changing as needed to produce a new end output

bazel build --config=cache --modify_execution_info='.*=+no-remote' //tools/generator/test:tests
```

Resulting in bytestream/upload for the test runner, but not the binary:
```
        "importantOutput": [
            {
                "name": "tools/generator/test/tests.test-runner.sh",
[...]

PR #16299, opened 2022-09-19
Add flag --experimental_remote_build_event_upload

coeuvre:
Add flag `--experimental_remote_build_event_upload` which controls the way Bazel uploads files referenced in BEP to remote cache.

It defaults to `all` which maintains current behaviour: Bazel uploads all local files referenced by BEP to remote cache and convert their paths to `bytestream://...`. Additionally, `--incompatible_remote_build_event_upload_respect_no_cache` can be set to avoid uploading outputs that are generated by "non-remote-cachable" spawns.

If set to `minimal`, local outputs are not uploaded to the remote cache, except for files that are **important** to the consumers of BES (e.g. test logs and timing profile). Paths for files that are already uploaded to the remote cache are converted.

`--incompatible_remote_build_event_upload_respect_no_cache` is deprecated in favour of this new flag.

Fixes https://github.com/bazelbuild/bazel/issues/16151.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
@@ -268,15 +268,43 @@ public final class RemoteOptions extends CommonRemoteOptions {
       help = "Whether to upload locally executed action results to the remote cache.")
   public boolean remoteUploadLocalResults;
 
+  @Deprecated
   @Option(
       name = "incompatible_remote_build_event_upload_respect_no_cache",
       defaultValue = "false",
       documentationCategory = OptionDocumentationCategory.REMOTE,
       effectTags = {OptionEffectTag.UNKNOWN},
+      deprecationWarning =
+          "--incompatible_remote_build_event_upload_respect_no_cache has been deprecated in favor"
+              + " of --experimental_remote_build_event_upload=minimal.",
       help =
           "If set to true, outputs referenced by BEP are not uploaded to remote cache if the"
               + " generating action cannot be cached remotely.")
   public boolean incompatibleRemoteBuildEventUploadRespectNoCache;
+
+  @Option(
+      name = "experimental_remote_build_event_upload",
+      defaultValue = "all",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      converter = RemoteBuildEventUploadModeConverter.class,
+      help =
+          "If set to 'all', all local outputs referenced by BEP are uploaded to remote cache.\n"
+              + "If set to 'minimal', local outputs referenced by BEP are not uploaded to the"
+              + " remote cache, except for files that are important to the consumers of BEP (e.g."
+              + " test logs and timing profile).\n"
+              + "file:// scheme is used for the paths of local files and bytestream:// scheme is"
+              + " used for the paths of (already) uploaded files.\n"
+              + "Default to 'all'.")
+  public RemoteBuildEventUploadMode remoteBuildEventUploadMode;
+
+  /** Build event upload mode flag parser */
+  public static class RemoteBuildEventUploadModeConverter
+      extends EnumConverter<RemoteBuildEventUploadMode> {
+    public RemoteBuildEventUploadModeConverter() {
+      super(RemoteBuildEventUploadMode.class, "remote build event upload");
+    }
+  }
 
   @Option(
       name = "incompatible_remote_results_ignore_disk",
--- a/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java
@@ -31,6 +31,7 @@ import com.google.devtools.build.lib.events.Event;
 import com.google.devtools.build.lib.events.ExtendedEventHandler;
 import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext;
 import com.google.devtools.build.lib.remote.common.RemoteActionExecutionContext.CachePolicy;
+import com.google.devtools.build.lib.remote.options.RemoteBuildEventUploadMode;
 import com.google.devtools.build.lib.remote.util.DigestUtil;
 import com.google.devtools.build.lib.remote.util.TracingMetadataUtils;
 import com.google.devtools.build.lib.vfs.Path;
@@ -52,12 +53,14 @@ import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /** A {@link BuildEventArtifactUploader} backed by {@link RemoteCache}. */
 class ByteStreamBuildEventArtifactUploader extends AbstractReferenceCounted
     implements BuildEventArtifactUploader {
+  private static final Pattern TEST_LOG_PATTERN = Pattern.compile(".*/bazel-out/[^/]*/testlogs/.*");
 
   private final Executor executor;
   private final ExtendedEventHandler reporter;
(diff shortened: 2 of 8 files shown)


src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -268,16 +268,41 @@ public String getTypeDescription() {
...
+      name = "experimental_remote_build_event_upload",
+      defaultValue = "all",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      converter = RemoteBuildEventUploadModeConverter.class,
+      help =
+          "If set to 'minimal', local outputs are not uploaded to the remote cache, except for"
+              + " files that are important to the consumers of BES (e.g. test logs and timing"
+              + " profile). Paths for blobs that are already uploaded to the remote cache are"
+              + " converted. Default to 'all'.")
```
brentleyjones, line 295: Does it make sense to mention the paths here? And if so, can the explanation be expanded, since what are they being converted to, and what does it have to do with uploading? (I know the answers to these, but thinking more for the documentation.)

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -268,16 +268,41 @@ public String getTypeDescription() {
...
 
+  @Deprecated
   @Option(
       name = "incompatible_remote_build_event_upload_respect_no_cache",
       defaultValue = "false",
       documentationCategory = OptionDocumentationCategory.REMOTE,
       effectTags = {OptionEffectTag.UNKNOWN},
+      deprecationWarning =
+          "--incompatible_remote_build_event_upload_respect_no_cache has been deprecated in favor"
+              + " of --experimental_remote_build_event_upload.",
```
brentleyjones, line 279: So does `--experimental_remote_build_event_upload=minimal` also imply `--incompatible_remote_build_event_upload_respect_no_cache`? This deprecation should probably mention the `minimal` value?

coeuvre: > So does `--experimental_remote_build_event_upload=minimal` also imply `--incompatible_remote_build_event_upload_respect_no_cache`?

No. I think with `--experimental_remote_build_event_upload=minimal`, there is no need for `incompatible_remote_build_event_upload_respect_no_cache`, right?

Updated message to mention `minimal`.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
coeuvre: Updated doc.

coeuvre:
@brentleyjones can you give this patch a try with your use cases? I think it's good enough but I might miss something.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
```diff
@@ -268,16 +268,42 @@ public String getTypeDescription() {
...
+      name = "experimental_remote_build_event_upload",
+      defaultValue = "all",
+      documentationCategory = OptionDocumentationCategory.REMOTE,
+      effectTags = {OptionEffectTag.UNKNOWN},
+      converter = RemoteBuildEventUploadModeConverter.class,
+      help =
+          "If set to 'minimal', local outputs are not uploaded to the remote cache, except for"
+              + " files that are important to the consumers of BEP (e.g. test logs and timing"
+              + " profile). file:// scheme is used for the paths of local files and bytestream://"
+              + " scheme is used for paths of (already) uploaded files. Default to"
```
brentleyjones, line 295: Do we need to mention the path behavior of `all`? Or adjust it to be agnostic of the setting (since the previous sentence already states what is or is not being uploaded)?

brentleyjones:
@coeuvre Will do.

src/main/java/com/google/devtools/build/lib/remote/options/RemoteOptions.java
coeuvre: Done.

brentleyjones:
Do you want to mention in the PR that this fixes https://github.com/bazelbuild/bazel/issues/16151?

brentleyjones:
This works in my testing.

CleanShot 2022-09-19 at 09 47 10@2x

I've asked a couple others to verify as well though.

erikkerber:
I tested as well and works well for our use case and what we were seeing.

BalestraPatrick:
Looks like this is working for us as well. I tested both on 5.3 without this flag to reproduce that `bes-upload` uploads various MBs of artifacts. Then on this PR with the flag set `bes-upload` doesn't upload the same artifacts, and of course on this PR without this flag the outputs are still uploaded. 👍

brentleyjones:
@coeuvre @meteorcloudy This has been great in our usage. Can we have the name become non-experimental, and the default changed to `minimal`, for 7.0?

coeuvre:
SGTM.

brentleyjones:
@meteorcloudy I don't think anything related to the implementation has changed in 7.0, the "it's worked great" has been 6.1.1 for me. Since a name change isn't a breaking change, can I include that part in 6.2? So I would do two PRs, one to change the name (and cherry-pick that) and one to flip the value (for 7.0 only)?

meteorcloudy:
SGTM, and I believe we have the `oldName` attribute for a flag, so that users still depend on `--experimental_remote_build_event_upload` won't be broken at all.

brentleyjones:
https://github.com/bazelbuild/bazel/pull/17988

lijunsong:
@brentleyjones 

When I set `--experimental_remote_build_event_upload=minimal`, the artifacts are not uploaded as expected, but the artifacts reference links are all `bytestream` instead of `file` even for those targets that are cached. Is this expected?

To reproduce this, you'll need to enable `--remote_executor`, and then:

```
$ echo "int main(){}" > main.cc
$ echo 'cc_binary(name="main", srcs=["main.cc"])' > BUILD
```

## Build with old flag
then build

```
bazel build --build_event_json_file=/tmp/before.json --incompatible_remote_build_event_upload_respect_no_cache :main
```

The event contains
```
{"id":{"namedSet":{"id":"0"}},"namedSetOfFiles":{"files":[{"name":"y4/main","uri":"file:///home/junsong-li/..../execroot/gibraltar/bazel-out/x86_64-opt-out/bin/y4/main","pathPrefix":["bazel-out","x86_64-opt-out","bin"],"digest":"2095f05a9e2118c2a42b6ef03d8e57b5fd7701fab737d0c784e2721a88d89562","length":"7448"}]}}
```

## Build with new flag

```
$ echo >> main.cc
```

and then add the flag

```
bazel build --build_event_json_file=/tmp/after.json --experimental_remote_build_event_upload=minimal  :main
```

The event contains

```
{"id":{"namedSet":{"id":"0"}},"namedSetOfFiles":{"files":[{"name":"y4/main","uri":"bytestream://remote_execution.hostname:port/remote-execution/blobs/be3a30bd7cf1aaf2e77fd492826e35b3a445e43aebf5f3de842ebbcad745096d/7448","pathPrefix":["bazel-out","x86_64-opt-out","bin"],"digest":"be3a30bd7cf1aaf2e77fd492826e35b3a445e43aebf5f3de842ebbcad745096d","length":"7448"}]}}
```

The reference doesn't exist

```
$ /tmp/tools_remote/bazel-bin/remote_client --remote_cache=remote_execution.hostname:port --remote_timeout=10 cat --digest be3a30bd7cf1aaf2e77fd492826e35b3a445e43aebf5f3de842ebbcad745096d/7448 --file=/tmp/myfile
Error: com.google.devtools.build.remote.client.CacheNotFoundException: Missing digest: hash: "be3a30bd7cf1aaf2e77fd492826e35b3a445e43aebf5f3de842ebbcad745096d"
size_bytes: 7448
```

So in sum,

1. The before.json contains `file://` reference
2.The after.json contains `bytestream://` reference even though the blob doesn't exist.

lijunsong:
Oh no, after #16999, BES always uses bytestream.

brentleyjones:
You can use `--nobuild_event_json_file_path_conversion` to get the file paths, right?

PR #16300, opened 2022-09-19
Mark XCode toolchain as `clang` rather than `compiler` 

fmeum:
RELNOTES: The @bazel_tools//tools/cpp:compiler flag now has the value `clang` for the auto-configured XCode toolchain rather than the generic value compiler. A branch for `clang` may have to be added to select statements that do not have a default case that handles this toolchain appropriately.


brentleyjones:
It's `Xcode` instead of `XCode` btw 😄.

src/test/shell/bazel/cc_integration_test.sh
```diff
@@ -1708,4 +1708,48 @@ EOF
   expect_log "in pkg/library.cpp: ''"
 }
 
+function test_compiler_flag_gcc() {
```
keith, line 1711: Looks like your other diff snuck in here

keith (commented):
Related https://github.com/bazelbuild/bazel/pull/14376

src/test/shell/bazel/cc_integration_test.sh
fmeum: It's stacked as the other PR hasn't been merged yet.

fmeum: I rebased onto master, so this is no longer stacked.

keith:
@googlewalt I think we'll want this one as well, related to https://github.com/bazelbuild/bazel/pull/14376/

googlewalt:
Would I be able to submit these two PRs independently, or will tests break?  Should I combine the two?

keith:
Combining is fine, but tests should be fine either way

tools/osx/crosstool/cc_toolchain_config.bzl
```diff
@@ -2898,7 +2898,7 @@ def _impl(ctx):
             target_system_name = target_system_name,
             target_cpu = ctx.attr.cpu,
             target_libc = target_libc,
-            compiler = "compiler",
```
googlewalt, line 2901: I think this should take the value of the input attribute compiler.

googlewalt: I will fix this during import.

PR #16302, opened 2022-09-19
Fix Bazel build with Bzlmod enabled

meteorcloudy:
Fixing https://buildkite.com/bazel/bazel-bazel-with-bzlmod/builds/288


Wyverald (approved):
nice!!

PR #16305, opened 2022-09-20
BES: make uploader retry attempts configurable

sluongng:
Depends on different Build Event Service setup, there could be different failure modes that may tolerate less or more failures for Build Events uploading.

Allow users to tweak the number without having to use a custom JVM args or shipping a fork of Bazel with these number tweaked.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceUploader.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/BuildEventServiceUploader.java
@@ -93,12 +93,6 @@ import javax.annotation.concurrent.GuardedBy;
 public final class BuildEventServiceUploader implements Runnable {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
 
-  /** Configuration knobs related to RPC retries. Values chosen by good judgement. */
-  private static final int MAX_NUM_RETRIES =
-      Integer.parseInt(System.getProperty("BAZEL_BES_NUM_RETRIES_ON_RPC_FAILURE", "4"));
-
-  private static final int DELAY_MILLIS = 1000;
-
   private final BuildEventServiceClient besClient;
   private final BuildEventArtifactUploader buildEventUploader;
   private final BuildEventServiceProtoUtil besProtoUtil;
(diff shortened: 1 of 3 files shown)


sluongng:
Not to sure how to test this just yet. I want to do this in Java but it might need to be a Bash test instead 🤔.

Open for suggestion.

coeuvre:
The changes LGTM but I will let @michaeledgar to make the decision.

sluongng:
I gave the flag names and docs slight tweak to better reflect their usages.

I am looking into `src/test/java/com/google/devtools/build/lib/buildeventservice/BazelBuildEventServiceModuleTest.java` but was unable to run the test locally on an M1 laptop.

```
java.lang.UnsatisfiedLinkError: /private/var/tmp/_bazel_sngoc/3438d92a73240559fc14d42b354b0b3f/sandbox/darwin-sandbox/2373/execroot/io_bazel/_tmp/96e4189e467fdee3386978ad0a337362/bazel-jni.11079868827068369907/libunix_jni.dylib: dlopen(/private/var/tmp/_bazel_sngoc/3438d92a73240559fc14d42b354b0b3f/sandbox/darwin-sandbox/2373/execroot/io_bazel/_tmp/96e4189e467fdee3386978ad0a337362/bazel-jni.11079868827068369907/libunix_jni.dylib, 0x0001): tried: '/private/var/tmp/_bazel_sngoc/3438d92a73240559fc14d42b354b0b3f/sandbox/darwin-sandbox/2373/execroot/io_bazel/_tmp/96e4189e467fdee3386978ad0a337362/bazel-jni.11079868827068369907/libunix_jni.dylib' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64)))

Exception in thread "Thread-0" java.lang.NoClassDefFoundError: Could not initialize class com.google.devtools.build.lib.unix.NativePosixFiles
	at java.base/java.lang.Thread.run(Thread.java:829)
EEE.EEE.EEE.EEE
```

sluongng:
I added a small test for the retry attempt count.

Not quite sure how to test the delay value though, I would need to be able to assert the info log somehow but I could not find a good way to do that.
If anyone could point me to a precedented approach to replicate, it would be very helpful.

sluongng:
@michaeledgar I would appreciate it if you could take a look at this PR.

Just want to know if you folks are ok with the direction.

michaeledgar (commented):
I'll merge this shortly. Thanks for sending this out, and persisting despite my unavailability!

PR #16311, opened 2022-09-20
Delete download temp directory after the build.

benjaminp:
I have observed that since https://github.com/bazelbuild/bazel/commit/bce7db055dd702e8a101f975f8585e0dfeb18a67, `WARNING: Found stale downloads from previous build, deleting...` is printed after every non-clean build. The reason appeared to be that the `remote` staging directory was not deleted after a build.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -856,6 +856,10 @@ public final class RemoteModule extends BlazeModule {
 
     if (actionContextProvider != null) {
       actionContextProvider.afterCommand();
+      var gen = actionContextProvider.getTempPathGenerator();
+      if (gen != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -856,6 +856,10 @@ public void afterCommand() throws AbruptExitException {
 
     if (actionContextProvider != null) {
       actionContextProvider.afterCommand();
+      var gen = actionContextProvider.getTempPathGenerator();
+      if (gen != null) {
+        deleteStaleDownloads(gen.getTempDir());
```
coeuvre, line 861: The downloads could continue in the background if we enable async downloads. Maybe its not necessary to print the warning message?

benjaminp: I wouldn't miss the message, and that's simpler.

PR #16320, opened 2022-09-21
remote: Remove warning about deleting temp downloads directory.

benjaminp:
The remote download directory is not deleted at the end of a build, so it is normal to hit this codepath.


coeuvre:
cc @tjgq since we have discussed this offline and you probably have the code to fix it.

benjaminp:
I see there's https://cr.bazel.build/204475—thank you!

PR #16321, opened 2022-09-21
Write a repo mapping manifest in the runfiles directory

Wyverald:
To ensure we can use repo mappings in the runfiles library, this change writes an extra file "my_binary_target.runfiles/_repo_mapping", which contains a bunch of (base_repo_canonical_name, apparent_repo_name, canonical_repo_name) triples. See https://github.com/bazelbuild/proposals/blob/main/designs/2022-07-21-locating-runfiles-with-bzlmod.md for more information.

Work towards https://github.com/bazelbuild/bazel/issues/16124


Wyverald:
@fmeum @lberki

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,122 @@
...
+      ActionOwner owner,
+      Artifact output,
+      Map<RepositoryName, RepositoryMapping> reposAndMappings,
+      String workspaceName) {
+    super(owner, NestedSetBuilder.emptySet(Order.STABLE_ORDER), output, /*makeExecutable=*/ false);
+    this.reposAndMappings =
+        ImmutableSortedMap.copyOf(
+            Maps.transformValues(
+                reposAndMappings, repoMapping -> ImmutableSortedMap.copyOf(repoMapping.entries())),
+            (a, b) -> OsPathPolicy.getFilePathOs().compare(a.getName(), b.getName()));
```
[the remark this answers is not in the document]
Wyverald: This is just to be consistent with the implementation of RepositoryName#equals, which also uses the OS-dependent thing. In practice this shouldn't really matter. (Is getting consistent outputs across OSes important?)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -495,4 +521,56 @@ public static Path inputManifestPath(Path runfilesDir) {
...
+    NestedSet<Package> transitivePackages =
+        ruleContext.getTransitivePackagesForRunfileRepoMappingManifest();
+    if (transitivePackages == null) {
+      // For environments where transitive packages are not tracked, we don't have external repos,
+      // so don't build the repo mapping manifest in such cases.
+      return null;
+    }
+    Set<NestedSet.Node> visitedNestedSets = new HashSet<>();
+    Map<RepositoryName, RepositoryMapping> reposAndMappings = new HashMap<>();
+    collectReposAndMappings(transitivePackages, visitedNestedSets, reposAndMappings);
```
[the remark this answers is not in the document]
Wyverald: Honestly not sure; I did it this way since I thought the nested set could be very big and it's not great to hold on to it after analysis, but maybe your concern about performance is more pressing. I'll wait for @lberki to weigh in.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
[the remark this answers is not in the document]
fmeum: It's not very important - cross-platform caching doesn't really work yet anyway since the output paths encode the CPU. But since I'm also working on improving that, I'm trying to minimize OS dependencies whenever I see them.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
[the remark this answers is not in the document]
fmeum: Do you have benchmarks for Bazel (since this feature is not enabled in Blaze) that would be able to catch such memory/CPU regressions?

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
lberki, line 62: Why are repository name sorted according to the host OS? Do you really want the contents to be sorted one way if Bazel runs on Windows and another way when it runs on some other OS?

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,122 @@
...
+        for (Entry<String, RepositoryName> mappingEntry : repoAndMapping.getValue().entrySet()) {
+          if (mappingEntry.getKey().isEmpty()) {
+            // The apparent repo name can only be empty for the main repo. We skip this line.
+            continue;
+          }
+          writer.write(repoAndMapping.getKey().getName());
+          writer.write(',');
+          writer.write(mappingEntry.getKey());
+          writer.write(',');
+          if (mappingEntry.getValue().isMain()) {
```
lberki, line 111: nit: a comment might be useful here that explains why this is needed. Let me see if my understanding is correct: the name of the main repository is `""`, but runfiles in it appear in the runfiles tree under the directory whose name is the same as the name of the main workspace. Is that so?

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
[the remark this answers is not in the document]
lberki: +1 for what Fabian says; the nested set is in fact big, but AFAICT we are not disposing of it, since a reference to it just stays in `ConfiguredTargetValue` / `AspectValue` in most of the cases; the notable exception is `--discard_analysis_cache`, but that one is only used within Google where the number of external repositories is zero and where this code is not very useful so meh (holding on to that nested set here would still hurt Google, but see another review comment I'm writing at the moment)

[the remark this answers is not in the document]
lberki: (and yes, we do have benchmarks!)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -495,4 +521,56 @@ public static Path inputManifestPath(Path runfilesDir) {
...
+        .registerAction(
+            new RepoMappingManifestAction(
+                ruleContext.getActionOwner(),
+                repoMappingManifest,
+                reposAndMappings,
+                ruleContext.getWorkspaceName()));
+    return repoMappingManifest;
+  }
+
+  private static void collectReposAndMappings(
```
lberki, line 561: This is delightfully optimal code, but why not just call `toList()`? I understand that it is less optimal, but it's also much simpler and if you do it while doing I/O anyway, it probably doesn't matter and looking at the call sites of `getLeaves()`, it looks like it's intended for places more deeply in the bowels of Bazel. In particular, the comment "may wait for a future" in `getNonLeaves()` scared me.

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
```diff
@@ -0,0 +1,274 @@
...
+    scratch.overwriteFile(
+        moduleRoot.getRelative("simple_rule~1.0/defs.bzl").getPathString(),
+        "def _simple_binary_impl(ctx):",
+        "  exe = ctx.actions.declare_file(ctx.label.name)",
+        "  ctx.actions.write(exe, ctx.toolchains['//:toolchain_type'].simple_info, True)",
+        "  runfiles = ctx.runfiles(files=ctx.files.data)",
+        "  for data in ctx.attr.data:",
+        "    runfiles.merge(data[DefaultInfo].default_runfiles)",
+        "  return DefaultInfo(files=depset(direct=[exe]), executable=exe, runfiles=runfiles)",
+        "simple_binary=rule(",
```
lberki, line 109: nit: I don't think a binary that uses a toolchain is "simple", but  see where you are coming from :)

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /**
+   * Sets up a Bazel module simple_rule@1.0, which provides a simple_binary rule that passes along
+   * runfiles in the data attribute. It also uses a toolchain (type //:toolchain_type, default
+   * toolchain //:simple_toolchain; the macro //:defs.bzl%simple_toolchain can be used to create
+   * more).
+   */
+  @Before
+  public void setupSimpleBinaryRule() throws Exception {
```
lberki, line 93: mind adding test cases for:
1. dependencies caused by aspects
2. host/exec dependencies
3. dependencies caused by an late-bound attribute 

Based on how the machinery that collects the set of transitive packages works, it looks like all these should work, but it's better to test it.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -132,15 +139,18 @@ private static RunfilesSupport create(
       runfilesInputManifest = null;
       runfilesManifest = null;
     }
+    Artifact repoMappingManifest = createRepoMappingManifestAction(ruleContext, owningExecutable);
```
lberki, line 142: Do I understand correctly that the same correctness guarantees will apply to the repo manifest as for the runfiles output manifest (and the runfiles tree), which is "nothing"? (I'm not saying this is a bad thing, I don't think you should be solving the correctness problems of runfiles here, just trying to validate my mental model against reality)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -132,15 +139,18 @@ private static RunfilesSupport create(
...
     Artifact runfilesMiddleman =
-        createRunfilesMiddleman(ruleContext, owningExecutable, runfiles, runfilesManifest);
+        createRunfilesMiddleman(
+            ruleContext, owningExecutable, runfiles, runfilesManifest, repoMappingManifest);
 
     boolean runfilesEnabled = ruleContext.getConfiguration().runfilesEnabled();
 
     return new RunfilesSupport(
```
lberki, line 149: How will repo mapping files work with sandboxes? I suppose they should appear there, but how would that work?

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import net.starlark.java.eval.EvalException;
+
+/** Creates a manifest file describing the repos and mappings relevant for a runfile tree. */
+public class RepoMappingManifestAction extends AbstractFileWriteAction {
```
lberki, line 45: I think it's worthwhile to add at least a smoke integration test; I believe you'll need that anyway to make sure repo mappings appear within sandboxes.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,147 @@
...
+      Writer writer = new BufferedWriter(new OutputStreamWriter(out, ISO_8859_1));
+
+      ImmutableSet<RepositoryName> reposContributingRunfiles =
+          runfilesArtifacts.toList().stream()
+              .filter(a -> a.getOwner() != null)
+              .map(a -> a.getOwner().getRepository())
+              .collect(toImmutableSet());
+      Map<RepositoryName, RepositoryMapping> reposAndMappings = new HashMap<>();
+      for (Package pkg : transitivePackages.toList()) {
+        reposAndMappings.putIfAbsent(
```
fmeum, line 104: Could you add a comment explaining why this is `putIfAbsent` and not `put`? I would assume that there is no difference, is that correct?

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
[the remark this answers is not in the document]
Wyverald: I decided to scrap this, since it's awkward whether or not we sort OS-dependently, and it's better to be less verbose. I have half a mind to just disallow uppercase letters in repo names...

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Override
+  public DeterministicWriter newDeterministicWriter(ActionExecutionContext ctx)
+      throws InterruptedException, ExecException {
+    return out -> {
+      Writer writer = new BufferedWriter(new OutputStreamWriter(out, ISO_8859_1));
+      for (Entry<RepositoryName, ImmutableSortedMap<String, RepositoryName>> repoAndMapping :
+          reposAndMappings.entrySet()) {
+        for (Entry<String, RepositoryName> mappingEntry : repoAndMapping.getValue().entrySet()) {
```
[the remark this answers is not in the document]
Wyverald: Done!

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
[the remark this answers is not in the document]
Wyverald: Done -- moved to the action itself

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
```diff
@@ -0,0 +1,274 @@
...
+        "  native.toolchain(",
+        "    name=name,",
+        "    toolchain=':'+name+'_impl',",
+        "    toolchain_type=Label('//:toolchain_type'),",
+        "  )");
+    scratch.overwriteFile(
+        moduleRoot.getRelative("simple_rule~1.0/BUILD").getPathString(),
+        "load('//:defs.bzl', 'simple_toolchain')",
+        "toolchain_type(name='toolchain_type')",
+        "simple_toolchain(name='simple_toolchain', string='simple')");
```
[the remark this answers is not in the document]
Wyverald: done

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
```diff
@@ -0,0 +1,274 @@
...
+        "load('@simple_rule//:defs.bzl', 'simple_binary')");
+    scratch.appendFile(moduleRoot.getRelative(repo).getRelative("BUILD").getPathString(), lines);
+  }
+
+  @Test
+  public void diamond() throws Exception {
+    rewriteWorkspace("workspace(name='aaa')");
+    scratch.overwriteFile(
+        "MODULE.bazel",
+        "bazel_dep(name='bbb',version='1.0')",
```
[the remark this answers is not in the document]
Wyverald: done

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: done. (yes you're correct)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: done -- switched to toList()

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
Wyverald: it's now called "bare" and doesn't use a toolchain anymore. probably more apt :)

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
Wyverald: I'll get to this next. re "host/exec dependencies" though -- why would we want those in runfiles at all?

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: I don't quite understand what you mean by "correctness guarantees"; any examples?

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: I really don't know how to answer this question :( Does the existing runfiles manifest file work with sandboxes? If so, it doesn't look like the repo mapping manifest file should be any different.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: will do.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: there's no difference. I can change it to `put` if you prefer that, I don't think it matters either way.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
fmeum: Both sandboxed and remote execution go through https://cs.opensource.google/bazel/bazel/+/59e51fe625f485dc664928ae0ad93035808779f3:src/main/java/com/google/devtools/build/lib/analysis/Runfiles.java;l=392;drc=901c75e459d737220cb8e29649c1b6ba24e2221d;bpv=1;bpt=1, so I think that this PR should just work with them. But of course a test will tell ;-)

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
fmeum: Maybe just add a comment saying that repo mappings are uniquely determined by the package and thus this doesn't matter.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -132,15 +135,19 @@ private static RunfilesSupport create(
       runfilesInputManifest = null;
       runfilesManifest = null;
     }
+    Artifact repoMappingManifest =
+        createRepoMappingManifestAction(ruleContext, runfiles, owningExecutable);
```
fmeum, line 139: Just noticed that we aren't adding the repo mapping manifest to the runfiles manifest, which makes it more difficult to find it at runtime on Windows. If we move this call further up, we could pass the Artifact into `SourceManifestAction`.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
lberki: Yep, a test case would dispel all my fears :)

The reason why I'm asking is that because I don't know off the top of my head how the runfiles manifest is plumbed to sandboxes and remote execution workers and since they are very special, I can't say with any confidence that the repo mapping file will be there. Ideally, by the time this changes is merged, we both would understand how runfiles output manifests appear in sandboxes / on RBE and be convinced that the repo mapping file does, too (and there would be a test also, of course!)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
lberki: Erm, sorry for being terse. My understanding is that Bazel doesn't guarantee that if one does

```
bazel build //a:b
rm bazel-bin/src/a/b.runfiles/<some file>
bazel build //a:b
```

the runfiles tree is not rebuilt except if `<some file>` is `MANIFEST`. Do I understand correctly that this would be the same with the repo mapping manifest, i.e. deleting it or changing it would not cause the symlink tree creation action to be re-run?

src/test/java/com/google/devtools/build/lib/analysis/RunfilesRepoMappingManifestTest.java
lberki: Files from host/exec dependencies should never be in the runfiles tree, but to the best of my knowledge, this is not enforced anywhere, which means that it's pretty certain that someone does it anyway.

Therefore, packages of host/exec dependencies must also be in the runfiles repo mapping (unless you convince me that the above property *is* enforced, but I'd be very surprised)

Wyverald: ah I see. I just added a new filtering so that only repos contributing runfiles would appear in the runfiles manifest (https://github.com/bazelbuild/bazel/pull/16321#discussion_r977272663), so I think we should have this case covered (unless someone is actively including stuff from host/exec deps into their runfiles, in which case they can simply... not).

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: I don't really know runfiles well enough to answer this question, but my guess is that yeah, the repo mapping manifest is just like the output manifest.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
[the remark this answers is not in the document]
lberki: Ooh, the `.getOwner().getRepository()` trick is neat!

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: added an integration test in bazel_module_test.

Wyverald:
So _that_ took a while.

* In order to make all tests pass, I eventually moved the repo mapping manifest file up a level, so it's now a sibling (instead of a child) of the runfiles tree. This is to avoid any entanglement with `SymlinkTreeAction`, which nukes the runfiles tree before generating any symlinks.
  * I also tried moving the repo mapping manifest writing logic into `SymlinkTreeAction`, but that breaks the interaction with the flag `--nobuild_runfile_links`, which delays the symlink tree generation to `bazel run` time and doesn't actually execute `SymlinkTreeAction`.
* I also went back to the original approach of flattening the `NestedSet<Package>` in `RunfilesSupport` instead of in `RepoMappingManifestAction` because the latter approach breaks some Google-internal tests.
* Additionally, the repo mapping manifest file is not created unless we're sure Bzlmod is enabled, which is also required make Google happy (as it creates much fewer actions across the repo).
  * This could alternatively be solved by somehow inserting the repo mapping manifest writing logic into `SourceManifestAction` (the one that creates the input manifest), but that action is an `AbstractFileWriteAction` which only produces one output file, and it's rather tiresome to create a custom action class just so that it might write 2 files.

With luck, all the tests (both internal and external) should pass now.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,128 @@
...
+  private static final UUID MY_UUID = UUID.fromString("458e351c-4d30-433d-b927-da6cddd4737f");
+
+  private final ImmutableList<Entry> entries;
+  private final String workspaceName;
+
+  /** An entry in the repo mapping manifest file. */
+  @AutoValue
+  public abstract static class Entry {
+    public static Entry of(
+        RepositoryName sourceRepo, String targetRepoApparentName, RepositoryName targetRepo) {
```
fmeum, line 53: The user-facing Bzlmod docs use the word "local" instead of "apparent". There are many other example of this (starting with "rule"), but if we want to do something about this divergence, now would be the time.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,128 @@
...
+        if (entry.targetRepoApparentName().isEmpty()) {
+          // the apparent repo name can only be empty for the main repo. we skip this line.
+          continue;
+        }
+        writer.write(entry.sourceRepo().getName());
+        writer.write(',');
+        writer.write(entry.targetRepoApparentName());
+        writer.write(',');
+        if (entry.targetRepo().isMain()) {
+          // The canonical name of the main repo is the empty string, but we use the "workspace
```
fmeum, line 117: The "we" here is a bit confusing since it could be understood as "this action", which isn't correct.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @Override
+  public DeterministicWriter newDeterministicWriter(ActionExecutionContext ctx)
+      throws InterruptedException, ExecException {
+    return out -> {
+      Writer writer = new BufferedWriter(new OutputStreamWriter(out, ISO_8859_1));
+      for (Entry entry : entries) {
+        if (entry.targetRepoApparentName().isEmpty()) {
+          // the apparent repo name can only be empty for the main repo. we skip this line.
```
fmeum, line 109: We could explain the underlying reason for this: Rlocation paths can't reference an empty apparent name as paths can't contain empty segments.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
```diff
@@ -0,0 +1,128 @@
...
+        writer.write(entry.targetRepoApparentName());
+        writer.write(',');
+        if (entry.targetRepo().isMain()) {
+          // The canonical name of the main repo is the empty string, but we use the "workspace
+          // name" as the name of the directory under the runfiles tree for it.
+          writer.write(workspaceName);
+        } else {
+          writer.write(entry.targetRepo().getName());
+        }
+        writer.write(System.lineSeparator());
```
fmeum, line 123: This sneaks in a host OS dependency, we should use `\n` instead. Runfiles libraries will be able to handle it.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -495,4 +524,83 @@ public static Path inputManifestPath(Path runfilesDir) {
...
+        ruleContext.getTransitivePackagesForRunfileRepoMappingManifest();
+    if (transitivePackages == null) {
+      // For environments where transitive packages are not tracked, we don't have external repos,
+      // so don't build the repo mapping manifest in such cases.
+      return null;
+    }
+
+    ImmutableList<Entry> entries = collectRepoMappings(transitivePackages, runfiles);
+    if (entries == null) {
+      // If Bzlmod is not enabled, nobody reads the repo mapping manifest anyway, so don't generate
```
fmeum, line 541: Can this branch be taken even if Bzlmod is enabled, for example in repositories defined by `WORKSPACE`?

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -495,4 +524,83 @@ public static Path inputManifestPath(Path runfilesDir) {
...
+
+  /**
+   * Returns the list of entries (unsorted) that should appear in the repo mapping manifest, or null
+   * if Bzlmod is not enabled.
+   */
+  @Nullable
+  private static ImmutableList<Entry> collectRepoMappings(
+      NestedSet<Package> transitivePackages, Runfiles runfiles) {
+    ImmutableSet<RepositoryName> reposContributingRunfiles =
+        runfiles.getAllArtifacts().toList().stream()
```
fmeum, line 574: This being a path not taken in Blaze and not taken in Bazel by default makes me a bit worried it may regress analysis phase performance outside Google. Were you able to get benchmark data on that or have other reasons to believe it's safe?

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -495,4 +524,83 @@ public static Path inputManifestPath(Path runfilesDir) {
...
+            .filter(a -> a.getOwner() != null)
+            .map(a -> a.getOwner().getRepository())
+            .collect(toImmutableSet());
+    Set<RepositoryName> seenRepos = new HashSet<>();
+    ImmutableList.Builder<Entry> entries = ImmutableList.builder();
+    // We use a rather hacky way to detect whether Bzlmod is enabled here: there exist repos that
+    // use strict deps iff Bzlmod is enabled. So if none of the repos we come upon use strict deps,
+    // either Bzlmod is disabled, or we're using a very weird subset of packages that *only* come
+    // from WORKSPACE-repos. In either case, it's unnecessary to output the repo mapping manifest.
+    boolean areWeSureBzlmodIsEnabled = false;
```
fmeum, line 584: It looks like we could simply call this `needsRepoMapping` and get rid of the comment, unless it helps Googlers monitoring action counts sleep soundly :-)

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
```diff
@@ -268,6 +283,16 @@ public Artifact getRunfilesManifest() {
     return runfilesManifest;
   }
 
+  /**
+   * Returns the foo.runfiles/_repo_mapping file if Bazel is run with transitive package tracking
```
fmeum, line 287: This comment is outdated.

src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -575,15 +568,100 @@ def setUpProjectWithLocalRegistryModule(self, dep_name, dep_version,
...
+      scratchFile(
+          projects_dir.joinpath(dir_name, 'BUILD'), [
+              'load("@bare_rule//:defs.bzl", "bare_binary")',
+              'package(default_visibility=["//visibility:public"])',
+              build_file,
+          ])
+
+    # Finally we get to build stuff!
+    self.RunBazel(['build', '//:me', '@bar//:bar'], allow_failure=False)
+    with open(self.Path('bazel-bin/me.repo_mapping'), 'r') as f:
```
fmeum, line 650: Could you build this check into a binary executed as part of an action? That would give use confidence that the file exists in the sandbox, not just in the output tree.

That said, we may want a test that covers remote execution.

src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -575,15 +568,100 @@ def setUpProjectWithLocalRegistryModule(self, dep_name, dep_version,
...
+      self.assertEqual(
+          f.read().strip(), """
+,foo,foo~1.0
```
fmeum, line 653:
```suggestion
          f.read().strip(), """,foo,foo~1.0
```
should fix the test failure

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
fmeum: Reading the definition of `collectRepoMappings` answered that question.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: Hmm. I think we should simply update the docs instead. "Apparent" has been used much more in discussions, and avoids the potential confusion with `local_repository`.

src/test/py/bazel/bzlmod/bazel_module_test.py
Wyverald: I made the `bare_binary` fail if it can't find the repo mapping manifest file, and changed `bazel build` to `bazel run`. The content assertion is still outside.

With the current way the repo mapping manifest file is generated, I don't think remote execution is a concern anymore -- it's just another normal artifact, and has no entanglement with the special handling around the symlink tree.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: Fixed!

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: Done.

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: I didn't get benchmark data. This change was made out of necessity rather than convenience, since things break internally if a `Package` is used as to compute an action key in any way.

Also, this is actually still run in Blaze, so its effects will be visible. It's just not run in certain other modes of builds in Google (where the transitive package tracking doesn't happen).

src/main/java/com/google/devtools/build/lib/analysis/RunfilesSupport.java
Wyverald: Fixed the comment.

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: Done

src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
Wyverald: Added comment

src/test/py/bazel/bzlmod/bazel_module_test.py
Wyverald: and of course the shell script doesn't work on windows... give me a second to fix that

fmeum: Agreed on the remote execution part. By making the binary a test and running `bazel test`, you would get coverage for the sandbox case without more code.

Wyverald:
Merged as https://github.com/bazelbuild/bazel/commit/527308cd3d965855b5a37ef8c0131165f0b10df5

PR #16329, opened 2022-09-22
fix: corporate logo should not have a grey background

alexeagle:
All the logos became badly rendered in the switch from bazelbuild/bazel-website to devsite. From inspecting the DOM, I can see that the problem is coming from this calculated css style

```
.devsite-landing-row-item-image:not([background]) {
  background: var(--devsite-background-4);
}
```

which I think comes from it being classified as a 'button'.

I don't know if there's a way for me to preview this change however. I'm following the same styling as the "Build Concepts" cards on https://bazel.build/start

### DIFF
--- a/site/en/community/experts/_index.yaml
+++ b/site/en/community/experts/_index.yaml
@@ -29,7 +29,7 @@ landing_page:
       buttons:
       - label: Learn more
         path: https://www.aspect.dev/
-        classname: button
+        classname: fully-clickable
     # Item 2
     - image_path: /community/images/buildbuddy-logo.svg
       heading: BuildBuddy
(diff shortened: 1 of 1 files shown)


gregestren:
Acknowledging this PR but I'd like @fweikert 's input.

fweikert:
Unfortunately this change does not fix the images, it simply removes the border of the button. However, I found a solution and will send a CL.

fweikert:
https://github.com/bazelbuild/bazel/commit/0e1d3ee70e7410eedd903907633e718f40031ee2 did the trick

alexeagle:
Thanks @fweikert !

PR #16342, opened 2022-09-27
Move macOS framework name linking into framework feature

keith:
This is a subtle change that moves the `-framework Foo` arguments to the linker after the `-filelist` argument. This matters in the case you have multiple symbols, one that is in your code, where the objects / archives are in the filelist, and one is in one of your pre-compiled framework dependencies. Previously the framework definition would have been preferred. This is inline with the order Xcode uses for this. The difference has cause real issues with prebuilt Swift https://github.com/apple/swift/issues/61287

### DIFF
--- a/tools/osx/crosstool/cc_toolchain_config.bzl
+++ b/tools/osx/crosstool/cc_toolchain_config.bzl
@@ -258,14 +258,6 @@ def _impl(ctx):
                         ],
                     ),
                     flag_group(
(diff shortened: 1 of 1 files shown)


trybka:
LGTM, but I'd like @googlewalt to review

googlewalt (approved):

keith:
thanks! @googlewalt can you land?

googlewalt:
Yes -- waiting for someone to review it internally.

PR #16349, opened 2022-09-28
Fix `//third_party:netty_tcnative` target on OpenBSD & FreeBSD.

aldersondrive:
On these platforms, no tcnative shared library is available. Without this change, the build fails with the following error.

```
ERROR: /home/ad/tree/third_party/BUILD:612:8: in cmd attribute of genrule rule //third_party:filter_netty_dynamic_libs: variable '$<' : no input file

ERROR: /home/ad/tree/third_party/BUILD:612:8: Analysis of target '//third_party:filter_netty_dynamic_libs' failed
```

(The root cause was that the `:filter_netty_dynamic_libs` target [assumed it could copy the shared library](https://github.com/bazelbuild/bazel/blob/c418e9e2635a0afdcbb36f7c14b41fec8417a356/third_party/BUILD#L623) even when [there was no shared library](https://github.com/bazelbuild/bazel/blob/c418e9e2635a0afdcbb36f7c14b41fec8417a356/third_party/BUILD#L620).

### DIFF
--- a/third_party/BUILD
+++ b/third_party/BUILD
@@ -665,8 +665,13 @@ distrib_java_import(
     enable_distributions = ["debian"],
     jars = [
         "netty_tcnative/netty-tcnative-classes-2.0.51.Final.jar",
-        ":netty_tcnative/netty-tcnative-filtered.jar",
-    ],
+    ] + select({
+        # No tcnative dynamic library is available for these platforms.
+        "//src/conditions:freebsd": [],
+        "//src/conditions:openbsd": [],
+        # For other platforms, a tcnative dynamic library is available.
+        "//conditions:default": [":netty_tcnative/netty-tcnative-filtered.jar"],
+    }),
 )
 
 distrib_java_import(


aiuto:
This will certainly fix the build issue. Will the resultant bazel be fully usable on BSD?
I see it used in 2 places
```
$ grep -r netty_tcnative src/
src/main/java/com/google/devtools/build/lib/remote/http/BUILD:        "//third_party:netty_tcnative",
src/main/java/com/google/devtools/build/lib/buildeventservice/client/BUILD:        "//third_party:netty_tcnative",
```

aldersondrive:
I think this change is *probably* an actual fix, because:
- The [FreeBSD port](https://github.com/freebsd/freebsd-ports/tree/main/devel/bazel/) of Bazel 5.3.0 works fine and apparently doesn't incorporate any tcnative shared library.
- FWIW, when I first ported Bazel 3.2.0 to OpenBSD a couple of years ago, Bazel ran fine without any tcnative shared library.

It's certainly possible that something has changed recently, or that I'm missing something. In particular, I don't know why the build bundles tcnative if Bazel apparently doesn't need it. I haven't yet finished sorting through the build failures to get an actual `bazel` binary to test with.

Based on the two points above, however, I think it's reasonable to at least un-break the build with this PR; the result would be not worse than today. In the event of an actual need for tcnative, I can deal with that as porting proceeds. If you prefer we see a `bazel` binary working first, that's also reasonable.

WDYT?

aiuto:
That reasoning makes sense to me.
Since this is third_party, only a small group of people can merge, so I'll let them handle it.

aldersondrive:
Confirmed—I now have a `bazel` binary that was built incorporating the changes in this PR, and it works just fine in my (basic) testing.

meteorcloudy:
https://github.com/bazelbuild/bazel/pull/16065 is sent for the same reason, I think explicitly listing the platforms where the library is available is better?

meteorcloudy:
Close this one since https://github.com/bazelbuild/bazel/pull/16065 is merged.

PR #16358, opened 2022-09-29
[fix] Update google-api-client to 1.35.0

mauriciogg:
google-api-client migrated from jackson to gson
see https://github.com/googleapis/google-api-java-client/pull/1661. The first release that contained this change is 1.32.1. Bazel removed the jackson artifacts but did not update the library accordingly and fails at runtime when google-api-client tries to access jackson classes

see https://github.com/bazelbuild/bazel/pull/15176

Updating to first version after 1.32.1 that does not have any listed vulnerability


aiuto:
This was effectively done at other times in other PRs.

PR #16359, opened 2022-09-29
[bes] Attach `RequestMetadata` to RPC calls

Yannic:
This allows servers to trace the requests similarly to other RPCs.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
+++ b/src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
@@ -102,7 +102,10 @@ public class BuildEventServiceGrpcClient implements BuildEventServiceClient {
           .withInterceptors(
               TracingMetadataUtils.attachMetadataInterceptor(
                   TracingMetadataUtils.buildMetadata(
-                      buildRequestId, commandId.toString(), "publish_lifecycle_event", null)))
+                      buildRequestId,
+                      commandId.toString(),
+                      "publish_lifecycle_event",
+                      /* actionMetadata= */ null)))
           .publishLifecycleEvent(lifecycleEvent);
     } catch (StatusRuntimeException e) {
       Throwables.throwIfInstanceOf(Throwables.getRootCause(e), InterruptedException.class);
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
```diff
@@ -67,6 +81,8 @@ protected BuildEventServiceGrpcClient(
     this.besAsync = besAsync;
     this.besBlocking = besBlocking;
     this.channel = channel;
+    this.buildRequestId = "testing/" + UUID.randomUUID();
```
lfpino, line 84: nit: move the testing string to a constant in the class

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
```diff
@@ -83,6 +99,10 @@ public void publish(PublishLifecycleEventRequest lifecycleEvent)
     try {
       besBlocking
           .withDeadlineAfter(RPC_TIMEOUT.toMillis(), MILLISECONDS)
+          .withInterceptors(
+              TracingMetadataUtils.attachMetadataInterceptor(
+                  TracingMetadataUtils.buildMetadata(
+                      buildRequestId, commandId.toString(), "publish_lifecycle_event", null)))
```
lfpino, line 105: nit: add the relevant comment to the null param, same below

lfpino (approved):

lfpino:
CC @coeuvre

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
Yannic: done

src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java
Yannic: This constructor seems unused: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceGrpcClient.java;l=63;drc=ceac50cc97d35b6e29d2940c1f925940f92be0b6;bpv=1;bpt=1

If it's unused at Google, we should just remove it.

coeuvre (approved):
Thanks!

PR #16366, opened 2022-09-30
Add Objc.static_framework_file to fully link arguments

keith:
When you are creating a fully linked static library that directly depends on static frameworks, you ideally want this framework to be linked, just as if it was an imported static library. Previously it would be silently ignored and you would have to distribute the static framework with the fully linked library. If users need the previous behavior they should use `avoid_deps` when creating the fully linked library.


trybka:
LGTM, but I'd like @googlewalt  to review

googlewalt:
LGTM.  We actually stopped using static_framework_file internally (migrating to imported_library instead), and was surprised to find this difference in behavior.

Issue #16368, opened 2022-10-01
D8 dex merger fails when a synthetic class is placed on a different shard than its container class


mauriciogg:
### Description of the bug:

 a desugared lambda is packed in a different shard than the one
from its original class D8 will fail with the following message:

`Attempt at compiling intermediate artifact without its context
`
e.g.

I'm encountering an issue when compiling my app with the following classes:
`org/chromium/base/MemoryPressureListener
``org/chromium/base/MemoryPressureListener$$ExternalSyntheticLambda0`

bazel fails with the following message 

```
Error in bazel-out/k8-fastbuild/bin/apps/hermosa/hermosa-services/realtime/mirroring-sample-app/dexsplits/apk/9.shard.zip:org/chromium/base/MemoryPressureListener$$ExternalSyntheticLambda0.class.dex:
Attempt at compiling intermediate artifact without its context
Merge failed: Compilation failed to complete, origin: bazel-out/k8-fastbuild/bin/apps/hermosa/hermosa-services/realtime/mirroring-sample-app/dexsplits/apk/9.shard.zip:org/chromium/base/MemoryPressureListener$$ExternalSyntheticLambda0.class.dex
Exception in thread "main" com.android.tools.r8.CompilationFailedException: Merge failed: Compilation failed to complete, origin: bazel-out/k8-fastbuild/bin/apps/hermosa/hermosa-services/realtime/mirroring-sample-app/dexsplits/apk/9.shard.zip:org/chromium/base/MemoryPressureListener$$ExternalSyntheticLambda0.class.dex
at com.google.devtools.build.android.r8.DexFileMerger.main(DexFileMerger.java:389)
Target //apps/hermosa/hermosa-services/realtime/mirroring-sample-app:apk failed to build
```
Here the container class is in shard #8 and the lambda class is in shard #9

Running with d8 dexmerger and native multidex.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Its hard to provide a small reproducible example since it needs to be an app large enough to exceed the size of a dex file and the sharding has to result in a lambda class ending up in a different shard than the container class.

### Which operating system are you running Bazel on?

Linux/Mac

### What is the output of `bazel info release`?

Internal build merged with master at commit https://github.com/bazelbuild/bazel/commit/b6803a2a8367e0a6748256d2e1f162ceb5451636

nkoroste:
@ahumesky

lukaciko:
[...]

PR #16369, opened 2022-10-01
Guarantee classes are packed with their inner synthetic lambdas during sharding

mauriciogg:
If a desugared lambda is packed in a different shard than the one from its original class D8 will fail with the following message:

Attempt at compiling intermediate artifact without its context

Changed sharding code so that it treats a class and synthetic lambda classes as a single unit.

Fixes https://github.com/bazelbuild/bazel/issues/16368

### DIFF
--- a/src/test/java/com/google/devtools/build/android/dexer/BUILD
+++ b/src/test/java/com/google/devtools/build/android/dexer/BUILD
@@ -97,7 +97,7 @@ java_library(
     srcs = [":mixed_srcjar_gen"],
 )
 
-android_library(
+java_library(
     name = "innerclasses_testinput",
     srcs = glob(["testdata/innerclasses/*.java"]),
 )
--- a/src/test/java/com/google/devtools/build/android/dexer/BUILD
+++ b/src/test/java/com/google/devtools/build/android/dexer/BUILD
@@ -34,6 +34,7 @@ java_library(
         "//src/test/java/com/google/devtools/build/lib/testutil",
         "//src/test/java/com/google/devtools/build/lib/testutil:TestSuite",
         "//src/tools/android/java/com/google/devtools/build/android/dexer",
+        "//src/tools/android/java/com/google/devtools/build/android/r8",
         "//third_party:guava",
         "//third_party:junit4",
         "//third_party:mockito",
@@ -47,7 +48,10 @@ java_library(
 
 java_library(
     name = "testdata",
-    srcs = glob(["testdata/**/*.java"]),
+    srcs = glob(
+            ["testdata/**/*.java"],
+            exclude = ["testdata/innerclasses/*"],
+    ),
 )
 
 java_test(
@@ -56,6 +60,7 @@ java_test(
     data = [
         "test_main_dex_list.txt",
         ":mixed_testinput",
+        ":innerclasses_testinput",
         ":testdata",
         ":tests",
     ],
@@ -64,6 +69,7 @@ java_test(
         "-Dtestinputjar=io_bazel/$(location :tests)",
         "-Dtestinputjar2=io_bazel/$(location :testdata)",
         "-Dmixedinputjar=io_bazel/$(location :mixed_testinput)",
+        "-Dinnerclassesinputjar=io_bazel/$(location :innerclasses_testinput)"
     ],
     runtime_deps = [
         ":tests",
@@ -89,4 +95,9 @@ genrule(
 java_library(
     name = "mixed_testinput",
     srcs = [":mixed_srcjar_gen"],
+)
+
+android_library(
+    name = "innerclasses_testinput",
+    srcs = glob(["testdata/innerclasses/*.java"]),
 )
--- a/src/test/java/com/google/devtools/build/android/dexer/DexFileSplitterTest.java
+++ b/src/test/java/com/google/devtools/build/android/dexer/DexFileSplitterTest.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.devtools.build.runfiles.Runfiles;
+import com.google.devtools.build.android.r8.Desugar;
 import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.FileSystems;
@@ -47,6 +48,7 @@ public class DexFileSplitterTest {
   private static final Path INPUT_JAR;
   private static final Path INPUT_JAR2;
   private static final Path MIXED_JAR;
+  private static final Path INNER_JAR;
   private static final Path MAIN_DEX_LIST_FILE;
   static final String DEX_PREFIX = "classes";
 
@@ -58,6 +60,7 @@ public class DexFileSplitterTest {
       INPUT_JAR2 = Paths.get(runfiles.rlocation(System.getProperty("testinputjar2")));
       MIXED_JAR = Paths.get(runfiles.rlocation(System.getProperty("mixedinputjar")));
       MAIN_DEX_LIST_FILE = Paths.get(runfiles.rlocation(System.getProperty("testmaindexlist")));
+      INNER_JAR = Paths.get(runfiles.rlocation(System.getProperty("innerclassesinputjar")));
     } catch (Exception e) {
       throw new ExceptionInInitializerError(e);
     }
@@ -229,6 +232,74 @@ public class DexFileSplitterTest {
     assertThat(dexEntries(outputArchives.get(2))).containsExactly("zzz/Bar.class.dex");
   }
 
+  @Test
+  public void testInnerClasses_keptTogether() throws Exception {
+    Path desugaredJar = desugarJar(INNER_JAR, "inner_together.jar");
+
+    System.out.println(desugaredJar.toString());
+    Path innerArchive = buildDexArchive(desugaredJar, "inner_together.jar.dex.zip");
+    // Pick a number such that OuterClass is packed with its inner classes 
+    // and other classes as well (total number of methods = 34 with OuterClass = 21)
+    ImmutableList<Path> outputArchives =
+        runDexSplitter(24, "inner_input", innerArchive);
+
+    String basepath = "com/google/devtools/build/android/dexer/testdata/innerclasses/";
+    String outerClass = basepath + "OuterClass.class.dex";
+    String outerClassLambdaName = "OuterClass$$ExternalSyntheticLambda";
+    String shouldBeTogether[] = new String[6];
+    for (int i = 0; i < shouldBeTogether.length; i++) {
+        shouldBeTogether[i] = basepath + outerClassLambdaName + i + ".class.dex";
+    }
+
+    // Assert that all inner classes are in the same archive
+    ImmutableSet<String> entries = ImmutableSet.of();
+    for (Path out : outputArchives) {
+        entries = dexEntries(out);
+        if (entries.contains(outerClass)) {
+            break;
+        }
+    }
+
+    for (int i = 0; i < shouldBeTogether.length; i++) {
+        assertThat(entries).contains(shouldBeTogether[i]);
+    }
+
+    assertThat(outputArchives).hasSize(3);
+  }
+
+  @Test
+  public void testInnerClasses_oneShard() throws Exception {
+    Path desugaredJar = desugarJar(INNER_JAR, "inner_one.jar");
+
+    System.out.println(desugaredJar.toString());
+    Path innerArchive = buildDexArchive(desugaredJar, "inner_one.jar.dex.zip");
+
+    // Pick a sufficiently high number to guarantee all classes fit
+    ImmutableList<Path> outputArchives =
+        runDexSplitter(256, "inner_input_one", innerArchive);
+
+    assertThat(outputArchives).hasSize(1);
+  }
+
+  @Test
+  public void testInnerClasses_doNotFit() throws Exception {
+    Path desugaredJar = desugarJar(INNER_JAR, "inner_fail.jar");
+    Path innerArchive = buildDexArchive(desugaredJar, "inner_fail.jar.dex.zip");
+
+    try {
+      ImmutableList<Path> outputArchives =
+      // Pick a sufficiently small number to guarantee fields wont fit
+      // (OuterClass has 21 methods)
+      runDexSplitter(20, "inner_input_fail", innerArchive);
+      fail("IllegalStateException expected");
+    } catch (IllegalStateException e) {
+      assertThat(e).hasMessageThat().contains("Impossible to fit");
+    }
+  }
+
+
+
+
   private static Iterable<String> expectedMainDexEntries() throws IOException {
     return Iterables.transform(
         Files.readAllLines(MAIN_DEX_LIST_FILE),
@@ -345,6 +416,20 @@ public class DexFileSplitterTest {
     DexBuilder.buildDexArchive(options, new Dexing(new DxContext(), dexingOptions));
     return options.outputZip;
   }
+
+  private Path desugarJar(Path inputJar, String outputJar) throws Exception {
+    Desugar.DesugarOptions options = new Desugar.DesugarOptions();
+    options.inputJars = ImmutableList.of(inputJar);
+    options.outputJars =
+        ImmutableList.of(FileSystems.getDefault().getPath(System.getenv("TEST_TMPDIR"), outputJar));
+    options.bootclasspath = ImmutableList.of();
+    options.classpath = ImmutableList.of();
+    options.minSdkVersion = 21;
+    new Desugar(options).desugar();
+    return options.outputJars.get(0);
+  }
+
+
 
   // Can't use lambda for Java 7 compatibility so we can run this Jar through dx without desugaring.
   private enum ZipEntryName implements Function<ZipEntry, String> {
--- a/src/test/java/com/google/devtools/build/android/dexer/DexFileSplitterTest.java
+++ b/src/test/java/com/google/devtools/build/android/dexer/DexFileSplitterTest.java
@@ -233,35 +233,52 @@ public class DexFileSplitterTest {
   }
 
   @Test
-  public void testInnerClasses_keptTogether() throws Exception {
+  public void testLambdaClasses_keptTogether() throws Exception {
     Path desugaredJar = desugarJar(INNER_JAR, "inner_together.jar");
 
-    System.out.println(desugaredJar.toString());
     Path innerArchive = buildDexArchive(desugaredJar, "inner_together.jar.dex.zip");
-    // Pick a number such that OuterClass is packed with its inner classes 
+    // Pick a number such that OuterClass is packed with its generated lambda classes 
     // and other classes as well (total number of methods = 34 with OuterClass = 21)
     ImmutableList<Path> outputArchives =
         runDexSplitter(24, "inner_input", innerArchive);
 
     String basepath = "com/google/devtools/build/android/dexer/testdata/innerclasses/";
     String outerClass = basepath + "OuterClass.class.dex";
-    String outerClassLambdaName = "OuterClass$$ExternalSyntheticLambda";
-    String shouldBeTogether[] = new String[6];
-    for (int i = 0; i < shouldBeTogether.length; i++) {
-        shouldBeTogether[i] = basepath + outerClassLambdaName + i + ".class.dex";
+    String lambdaClassName = "OuterClass$$ExternalSyntheticLambda";
+    String lambdaClasses[] = new String[6];
+    for (int i = 0; i < lambdaClasses.length; i++) {
+        lambdaClasses[i] = basepath + lambdaClassName + i + ".class.dex";
+    }
+
+    String innerClassName = "OuterClass$InnerClass";
+    String innerClasses[]  = new String[6];
+    for (int i = 0; i < innerClasses.length; i++) {
+        innerClasses[i] = basepath + innerClassName + i + ".class.dex";
     }
 
     // Assert that all inner classes are in the same archive
     ImmutableSet<String> entries = ImmutableSet.of();
+    int idx = 0;
     for (Path out : outputArchives) {
         entries = dexEntries(out);
         if (entries.contains(outerClass)) {
-            break;
+          break;
         }
+        idx++;
     }
 
-    for (int i = 0; i < shouldBeTogether.length; i++) {
-        assertThat(entries).contains(shouldBeTogether[i]);
+    // assert all lambdas are together
+    for (int i = 0; i < lambdaClasses.length; i++) {
+        assertThat(entries).contains(lambdaClasses[i]);
+    }
+
+    // Assert that the other inner classes appear in lambdas shard or the next shard
+    ImmutableSet.Builder<String> builder = ImmutableSet.builder();
+    builder.addAll(entries);
+    builder.addAll(dexEntries(outputArchives.get(idx+1)));
+    ImmutableSet<String> combinedEntries = builder.build();
+    for (int i = 0; i < innerClasses.length; i++) {
+        assertThat(combinedEntries).contains(innerClasses[i]);
     }
 
     assertThat(outputArchives).hasSize(3);
@@ -269,36 +286,32 @@ public class DexFileSplitterTest {
 
   @Test
   public void testInnerClasses_oneShard() throws Exception {
-    Path desugaredJar = desugarJar(INNER_JAR, "inner_one.jar");
+    Path desugaredJar = desugarJar(INNER_JAR, "lambdas_one.jar");
 
-    System.out.println(desugaredJar.toString());
-    Path innerArchive = buildDexArchive(desugaredJar, "inner_one.jar.dex.zip");
+    Path innerArchive = buildDexArchive(desugaredJar, "lambdas_one.jar.dex.zip");
 
     // Pick a sufficiently high number to guarantee all classes fit
     ImmutableList<Path> outputArchives =
-        runDexSplitter(256, "inner_input_one", innerArchive);
+        runDexSplitter(256, "lambdas_input_one", innerArchive);
 
     assertThat(outputArchives).hasSize(1);
   }
 
   @Test
-  public void testInnerClasses_doNotFit() throws Exception {
-    Path desugaredJar = desugarJar(INNER_JAR, "inner_fail.jar");
-    Path innerArchive = buildDexArchive(desugaredJar, "inner_fail.jar.dex.zip");
+  public void testLambdaClasses_doNotFit() throws Exception {
+    Path desugaredJar = desugarJar(INNER_JAR, "lambdas_fail.jar");
+    Path innerArchive = buildDexArchive(desugaredJar, "lambdas_fail.jar.dex.zip");
 
     try {
       ImmutableList<Path> outputArchives =
       // Pick a sufficiently small number to guarantee fields wont fit
       // (OuterClass has 21 methods)
-      runDexSplitter(20, "inner_input_fail", innerArchive);
+      runDexSplitter(20, "lambdas_input_fail", innerArchive);
       fail("IllegalStateException expected");
     } catch (IllegalStateException e) {
       assertThat(e).hasMessageThat().contains("Impossible to fit");
     }
   }
-
-
-
 
   private static Iterable<String> expectedMainDexEntries() throws IOException {
     return Iterables.transform(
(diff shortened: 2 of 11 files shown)


Bencodes:
@mauriciogg i'm still able to repro this using this repo:

```console
Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda2.class.dex:
Attempt at compiling intermediate artifact without its context
Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda0.class.dex:
Attempt at compiling intermediate artifact without its context
Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda10.class.dex:
Attempt at compiling intermediate artifact without its context
Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex:
Attempt at compiling intermediate artifact without its context
Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
Exception in thread "main" com.android.tools.r8.CompilationFailedException: Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
        at com.google.devtools.build.android.r8.DexFileMerger.main(DexFileMerger.java:389)
```

mauriciogg:
> @mauriciogg i'm still able to repro this using this repo:
> 
> ```
> Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda2.class.dex:
> Attempt at compiling intermediate artifact without its context
> Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda0.class.dex:
> Attempt at compiling intermediate artifact without its context
> Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda10.class.dex:
> Attempt at compiling intermediate artifact without its context
> Error in bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex:
> Attempt at compiling intermediate artifact without its context
> Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
> Exception in thread "main" com.android.tools.r8.CompilationFailedException: Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip:com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
>         at com.google.devtools.build.android.r8.DexFileMerger.main(DexFileMerger.java:389)
> ```

 are all GalleryService$$ExternalSyntheticLambda in the same shard that gallery service? If so it might be that lambdas itself need to appear in order as well and my code does just sorts alphabetically which misses the ordering of lambdas.

Bencodes:
It looks like they are being split between two dex files.

```console
➜ zipinfo bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip | grep "GalleryService"
-rw----     1.0 fat     1268 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda0.class.dex
-rw----     1.0 fat     1284 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
-rw----     1.0 fat     1308 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda10.class.dex
-rw----     1.0 fat     1392 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda2.class.dex
-rw----     1.0 fat     1332 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda3.class.dex
```

While the main class is in:
```console
➜ zipinfo bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard10.jar.dex.zip | grep "GalleryService"
-rw----     1.0 fat     1344 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda4.class.dex
-rw----     1.0 fat     1272 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda5.class.dex
-rw----     1.0 fat     1288 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda6.class.dex
-rw----     1.0 fat     1396 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda7.class.dex
-rw----     1.0 fat     1400 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda8.class.dex
-rw----     1.0 fat     1448 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda9.class.dex
-rw----     1.0 fat     8956 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService.class.dex
-rw----     1.0 fat     2584 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService_Factory.class.dex
```

mauriciogg:
> It looks like they are being split between two dex files.
> 
> ```
> ➜ zipinfo bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard9.jar.dex.zip | grep "GalleryService"
> -rw----     1.0 fat     1268 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda0.class.dex
> -rw----     1.0 fat     1284 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda1.class.dex
> -rw----     1.0 fat     1308 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda10.class.dex
> -rw----     1.0 fat     1392 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda2.class.dex
> -rw----     1.0 fat     1332 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda3.class.dex
> ```
> 
> While the main class is in:
> 
> ```
> ➜ zipinfo bazel-out/darwin_arm64-dbg/bin/apps/app/_dx/bin_devDebug/shard10.jar.dex.zip | grep "GalleryService"
> -rw----     1.0 fat     1344 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda4.class.dex
> -rw----     1.0 fat     1272 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda5.class.dex
> -rw----     1.0 fat     1288 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda6.class.dex
> -rw----     1.0 fat     1396 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda7.class.dex
> -rw----     1.0 fat     1400 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda8.class.dex
> -rw----     1.0 fat     1448 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService$$ExternalSyntheticLambda9.class.dex
> -rw----     1.0 fat     8956 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService.class.dex
> -rw----     1.0 fat     2584 bx stor 69-Dec-31 16:00 com/lyft/android/camera/gallery/GalleryService_Factory.class.dex
> ```

@Bencodes  I can't repo: I don't see how it could be sharded that way. Looks like dexes are being sort in reverse order. The main class being in shard10 while the inner classes being in shard9. Should be the other way:
```
      ImmutableList<Map.Entry<String, ZipFile>> files =
          deduped
              .entrySet()
              .stream()
                   // ZipEntryComparator::compareClassNames: Ensure inner classes sort second
                   //   a = a.replace('$', '0');
                   //   b = b.replace('$', '0');
              .sorted(Comparator.comparing(e -> e.getKey(), ZipEntryComparator::compareClassNames))
              .collect(ImmutableList.toImmutableList());
```
My code (AFAICT) did not change the iteration order. Can you share the ordering of the dexes building without my code?

Bencodes:
Some follow up here - there is some misleading branching of login inside of Bazel that might trip people up. If you set `dex_shards` on your `android_binary`, the Android binary will enter an entirely different branch of logic for sharding, which fails. When removing that `dex_shards` our builds hit the new sharding logic.

I re-applied the latest patches that you pushed yesterday which seems to have broken `DexFileSplitter`. We are getting this error now at runtime:

```
                      ^
  symbol:   class Dex
  location: class DexEntry
external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:378: error: cannot find symbol
      public Dex getDexFile() {
             ^
  symbol:   class Dex
  location: class DexEntry
external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:114: error: cannot find symbol
      defaultValue = "" + DexFormat.MAX_MEMBER_IDX,
                         ^
```

mauriciogg:
> Some follow up here - there is some misleading branching of login inside of Bazel that might trip people up. If you set `dex_shards` on your `android_binary`, the Android binary will enter an entirely different branch of logic for sharding, which fails. When removing that `dex_shards` our builds hit the new sharding logic.
> 
> I re-applied the latest patches that you pushed yesterday which seems to have broken `DexFileSplitter`. We are getting this error now at runtime:
> 
> ```
> (15:23:25) ERROR: /private/var/tmp/_bazel_blee/499a001013731d09bffd82f8601a3161/external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/BUILD:41:12: Building external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.jar (5 source files) and running annotation processors (AutoValueProcessor) [for tool] failed: (Exit 1): java failed: error executing command (from target @bazel_tools//src/too
> ls/android/java/com/google/devtools/build/android/dexer:DexFileSplitter) external/lyft_jdk11/bin/java -XX:-CompactStrings '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED' '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED' ... (remaining 16 arguments skipped)
> external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:20: error: package com.android.dex does not exist
> import com.android.dex.Dex;
>                       ^
> external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:21: error: package com.android.dex does not exist
> import com.android.dex.DexFormat;
>                       ^
>                       
>                       
>                       ^
>   symbol:   class Dex
>   location: class DexEntry
> external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:378: error: cannot find symbol
>       public Dex getDexFile() {
>              ^
>   symbol:   class Dex
>   location: class DexEntry
> external/bazel_tools/src/tools/android/java/com/google/devtools/build/android/dexer/DexFileSplitter.java:114: error: cannot find symbol
>       defaultValue = "" + DexFormat.MAX_MEMBER_IDX,
>                          ^
> ```

@Bencodes   oh yeah you have to remove dex_shards otherwise it does something completely different.
For the runtime issues do you have these commits?:
https://github.com/bazelbuild/bazel/commit/af1623afbcbbad0c574bc87b8858fa9840a39fec
https://github.com/bazelbuild/bazel/commit/6b25cb20f31d3a0fb86dad25962df94e9a328359

thats the only thing I can think that could be causing those issues

Bencodes:
Yeah we have both commits.

mauriciogg:
ugh [android_remote_tools.WORKSPACE](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/android/android_remote_tools.WORKSPACE#L9)  probably needs to be updated. Although not sure I can do that myself.

Bencodes:
Actually hold up. I think it might have been user error on my part. The `android_tools.tar.gz` was renamed to `android_tools.tar` so my local testing scripts were pulling from a stale output.

Bencodes:
Alright @mauriciogg confirmed that this PR does in fact work for us now.

mauriciogg:
@Bencodes thanks!

src/test/java/com/google/devtools/build/android/dexer/BUILD
```diff
@@ -47,7 +48,10 @@ java_library(
 
 java_library(
     name = "testdata",
-    srcs = glob(["testdata/**/*.java"]),
+    srcs = glob(
```
Bencodes, line 51: Might need to run buildifier and google-java-format against the changes in this PR.

ahumesky: The code will get formatted automatically on import I think, so I don't think we have to manually invoke the formatter

ahumesky:
Regarding android_remote_tools.WORKSPACE, it doesn't look like it needs to be modified, but if it does, you can make changes to it, but we'll need to make a new release of the remote tools and then update Bazel's dependency on it. It's usually not that much work.

For getting the tests to pass on CI, I think you can use the r8 jar from google's maven repo here:
https://github.com/bazelbuild/bazel/blob/9ff90410feb7c4c6b9a80d414c25aa93968d62ad/WORKSPACE#L417

by changing this dependency:
https://github.com/bazelbuild/bazel/blob/9ff90410feb7c4c6b9a80d414c25aa93968d62ad/src/tools/android/java/com/google/devtools/build/android/r8/BUILD#L55-L58

to `@android_gmaven_r8_for_testing//jar`

mauriciogg:
> Regarding android_remote_tools.WORKSPACE, it doesn't look like it needs to be modified, but if it does, you can make changes to it, but we'll need to make a new release of the remote tools and then update Bazel's dependency on it. It's usually not that much work.
> 
> For getting the tests to pass on CI, I think you can use the r8 jar from google's maven repo here:
> 
> https://github.com/bazelbuild/bazel/blob/9ff90410feb7c4c6b9a80d414c25aa93968d62ad/WORKSPACE#L417
> 
> by changing this dependency:
> 
> https://github.com/bazelbuild/bazel/blob/9ff90410feb7c4c6b9a80d414c25aa93968d62ad/src/tools/android/java/com/google/devtools/build/android/r8/BUILD#L55-L58
> 
> to `@android_gmaven_r8_for_testing//jar`

I ended up adding that dependency on the test build file src/test/java/com/google/devtools/build/android/dexer/BUILD because I didnt want to include a test dependency in non test code, but doing it this way the gmaven jar has to always come up first in the classpath which is somewhat brittle- so let me know if you want me to move it into the r8 BUILD file.

ahumesky:
Yeah good point about test code vs non-test code. Luckily `//src/tools/android/java/com/google/devtools/build/android/r8:r8` is visible to only test code at the moment, so I think it would be fine to add `testonly = True`. It's really only for tests anyway, because the R8 source code in there gets embedded into bazel and compiled later to avoid bundling the D8/R8 jars within bazel. This is all for the desugaring classes used in the test which come from the `r8` target, so it seems less surprising to change the deps there, which also avoids the class path brittleness.

mauriciogg:
> Yeah good point about test code vs non-test code. Luckily `//src/tools/android/java/com/google/devtools/build/android/r8:r8` is visible to only test code at the moment, so I think it would be fine to add `testonly = True`. It's really only for tests anyway, because the R8 source code in there gets embedded into bazel and compiled later to avoid bundling the D8/R8 jars within bazel. This is all for the desugaring classes used in the test which come from the `r8` target, so it seems less surprising to change the deps there, which also avoids the class path brittleness.

Done

ahumesky:
I ran this against our internal global presubmit tests, and I'm getting back about 900 test targets failing with
```
Exception in thread "main" java.lang.IllegalStateException: Impossible to fit <foo>.class.dex and all of its inner classes in a single shard
	at com.google.common.base.Preconditions.checkState(Preconditions.java:600)
	at com.google.devtools.build.android.dexer.DexFileSplitter.processDexEntries(DexFileSplitter.java:353)
	at com.google.devtools.build.android.dexer.DexFileSplitter.processDexFiles(DexFileSplitter.java:295)
	at com.google.devtools.build.android.dexer.DexFileSplitter.splitIntoShards(DexFileSplitter.java:188)
	at com.google.devtools.build.android.dexer.DexFileSplitter.main(DexFileSplitter.java:141)
```
I'll look into what's going on. It would be a little surprising if there were classes with so many inner classes that they don't fit into a shard though.

ahumesky:
So the root cause of the vast majority of the failures is from a generated file for protos, and has over 5,000 inner classes, with enough methods that it's bigger than what 1 shard can contain. Not totally sure offhand what can be done about this case. I'll check with the D8 team. There are 2 other root causes for the rest of the failures, they're probably similar situations.

mauriciogg:
> So the root cause of the vast majority of the failures is from a generated file for protos, and has over 5,000 inner classes, with enough methods that it's bigger than what 1 shard can contain. Not totally sure offhand what can be done about this case. I'll check with the D8 team. There are 2 other root causes for the rest of the failures, they're probably similar situations.

I havent confirmed this, but I think the error only happens with classes generated from desugaring lambdas, so in theory those are the only ones needed to be packed with the outer class. If thats, the case we could be smarted here and place sort lambdas right after the outer class and allow splits on other inner classes.

ahumesky:
@mauriciogg were you able to test if including just the classes from desugared lambdas in a shard worked? I can re-run the internal tests with a prototype. I also asked the D8 team if that would work.

mauriciogg:
wo

> @mauriciogg were you able to test if including just the classes from desugared lambdas in a shard worked? I can re-run the internal tests with a prototype. I also asked the D8 team if that would work.

Works in our codebase: I changed `isOuter` to `isLambda` and `processDexFiles` to use that instead :
```
  private boolean isLambda(String classFile) {
    String components[] = classFile.split("/");
    return components[components.length-1].contains("ExternalSyntheticLambda");
  }
 ....

  private void processDexFiles(
    for (Map.Entry<String, ZipFile> entry : filesToProcess) {
      String filename = entry.getKey();
      if (filter.test(filename)) {
        boolean isLambda = isLambda(filename);

	 // TODO: optimize- this currently creates a list of size one for most cases
	 if (!isLambda) {
	    if (!zipEntries.isEmpty()) {
		processDexEntries(zipEntries);
	    }
	    zipEntries.clear();
	}
        zipEntries.add(entry);
      }
    }
    processDexEntries(zipEntries);
  }
```

mauriciogg:
@ahumesky updated to do the split only on lambdas

arunkumar9t2:
We faced this when we migrated to `6.0.0-pre.20220818.1` and we were able to workaround it temporarily by downgrading R8 to 3.0.64. This is the commit that changed the behavior in R8 if that helps https://r8-review.googlesource.com/c/r8/+/60228

ahumesky:
@mauriciogg thanks, i started rerunning the tests, should have results by tomorrow

ahumesky:
I ran this latest version against our internal tests, and everything passed, however in talking with the D8 team, they brought up that there are other kinds of synthetic classes that could be generated, so this only happened to work for now. They suggested to go back to grouping inner classes with their enclosing class, and simply removing the check that they'll all fit into a single dex. With multidex, it shouldn't matter for the dexmerger. I reverted the imported change to group by inner/enclosing class, and took out the check, and I started rerunning all the tests against that.

lukaciko:
We seem to have hit the issue with one of those other kinds of synthetic classes. With Bazel 6.0.0, this PR applied on top and when not setting `dex_shards` we still get:

```
Error in bazel-out/<foo>/15.shard.zip:<bar>/MobileEnhancedDialogsEventFactory$EnhanceButtonTooltip-IA.class.dex:
Attempt at compiling intermediate artifact without its context
Merge failed: Compilation failed to complete, origin: bazel-out/<foo>/dexsplits/app/15.shard.zip:<bar>/MobileEnhancedDialogsEventFactory$EnhanceButtonTooltip-IA.class.dex
Exception in thread "main" com.android.tools.r8.CompilationFailedException: Merge failed: Compilation failed to complete, origin: bazel-out/<foo>/dexsplits/app/15.shard.zip:<bar>/MobileEnhancedDialogsEventFactory$EnhanceButtonTooltip-IA.class.dex
	at com.google.devtools.build.android.r8.DexFileMerger.main(DexFileMerger.java:389)
```

Using `zipinfo` I see `MobileEnhancedDialogsEventFactory$EnhanceButtonTooltip-IA.class.dex` is in `15.shard.zip` and `MobileEnhancedDialogsEventFactory$EnhanceButtonTooltip.class.dex` is in `16.shard.zip` while other `-IA` classes are in the same dex shard as their counterparts.

I don't know what these `-IA` classes are, but when I use a bytecode inspector I see they have a `.source "D8$$SyntheticClass"` line so I'm assuming they're synthetic classes generated by D8.

lukaciko:
We ran into this issue with another kind of synthetic classes, `ExternalSyntheticBackportWithForwarding0`, when using Bazel 6 + this change + handling the `-IA` classes I mentioned above.

```
Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-fastbuild/bin/<foo>/_dx/app/shard21.jar.dex.zip:io/reactivex/rxjava3/internal/operators/observable/ObservableReplay$$ExternalSyntheticBackportWithForwarding0.class.dex
```

Looking forward to a generic fix. Until then it is too risky for us to start using Bazel 6.

mauriciogg:
> We ran into this issue with another kind of synthetic classes, `ExternalSyntheticBackportWithForwarding0`, when using Bazel 6 + this change + handling the `-IA` classes I mentioned above.
> 
> ```
> Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-fastbuild/bin/<foo>/_dx/app/shard21.jar.dex.zip:io/reactivex/rxjava3/internal/operators/observable/ObservableReplay$$ExternalSyntheticBackportWithForwarding0.class.dex
> ```
> 
> Looking forward to a generic fix. Until then it is too risky for us to start using Bazel 6.

A generic fix would be to just group all inner classes. Looks like support for this already exists in mobile-install https://github.com/bazelbuild/rules_android/blob/pre-alpha/src/tools/ak/dex/dex.go#L265

ahumesky:
I ran our internal tests against this, and ran into a lot of failures -- looking through the code it was because this bit of code:

https://github.com/bazelbuild/bazel/pull/16369/files#diff-f01bb745669c01a82d58e3c3cbd6a07f570b297f292cf906634bf4d3a0712089R352-R357

somehow became just
```
    if (tracker.track(dexFiles)) {
      nextShard();
    }
```

leading to a number of dexfiles going over the 64k limit because not all the dexes were counted. That took a while to figure out. Not sure what happened there...

In any event, I've reimported the change with the correct code, and adjusted it to group by inner class again, and started rerunning all our tests. I'll hopefully have results on Monday. We should be able to get this reviewed and imported in time for 6.1.0.

ahumesky:
Looking at the results, and there are many failures like:

```
Impossible to fit _____ and all of its synthetic lambda classes in a single shard
```

I didn't update the error message -- it's really "and all of its inner classes".

I looked into one of the apps with the failure, and looking at all the classes it looked like it didn't have any classes with 64k+ inner classes, so this will need a little more investigating.

ahumesky:
It looks like we have some classes that have enough inner classes to exceed the dex limits for a single dex, so it doesn't look like grouping by just inner classes will work. With help from the D8 folks, I reworked the change so that it groups only classes that are truly synthetic classes, which have an annotation on them from R8.

Since I can't readily repro the original problem, it would be very helpful if those of you who are running into this can try out the updated PR:
https://bazel-review.googlesource.com/c/bazel/+/211231

lukaciko:
@ahumesky thanks for the update. With that patch we're still seeing the issue with proto `-IA` classses:

```
Exception in thread "main" com.android.tools.r8.CompilationFailedException: Merge failed: Compilation failed to complete, origin: bazel-out/darwin_arm64-fastbuild/bin/src/main/java/<foo>/dexsplits/app/8.shard.zip:<bar>/proto/FooClass$Builder-IA.class.dex
	at com.google.devtools.build.android.r8.DexFileMerger.main(DexFileMerger.java:38
```

`FooClass$Builder-IA.class.dex` is annotated with `SynthesizedClassV2`, so that shouldn't be a problem. The next shard contains:

```
FooClass.class.dex
FooClass$1.class.dex
FooClass$Builder.class.dex
FooClassOrBuilder.class.dex
FooClassOuterClass.class.dex
```

ahumesky:
Thanks for trying this and for the detailed report -- I think the problem is that the code doesn't actually make sure that the outer class is with the synthetic classes. The outer class + its synthetic inner classes have to fit into one shard, but the "regular" inner classes can go into any shard. Because the outer class is being put in with the "regular" classes, it could potentially be placed into the next shard.

I think this bit:
```
        if (isOuterClass || !isR8SyntheticClass(dexEntry.getDexFile())) {
          regularClasses.add(dexEntry);
        } else {
          syntheticInnerClasses.add(dexEntry);
        }
```
needs to be:
```
        if (isOuterClass || isR8SyntheticClass(dexEntry.getDexFile())) {
          // Outer classes are never synthetic classes, however since the synthetic classes have to
          // go with their outer class, add the outer class with the synthetic classes.
          syntheticInnerClasses.add(dexEntry);
        } else {
          regularClasses.add(dexEntry);
        }
```
For completeness, I've uploaded a new patch with the above to https://bazel-review.googlesource.com/c/bazel/+/211231

But all the above said, after talking through all of this with the D8 team, they're a bit worried about how tightly coupled this all is to implementation details of D8 (the annotation, the annotation class name, the naming conventions), not to mention it's not really great for performance to have to parse the dex files. (The code is also getting pretty complicated.)

The D8 team proposes:

1. Update CompatDexBuilder to use the D8 DexFilePerClassFile consumer instead of DexIndexed (this should be a no op for the build as the current build there is already doing one dex per class file, but it uses another D8 API such that it can get the synthetic info).

2. Update the D8 DexFilePerClassFile accept function to also receive the "synthetic context" information if the compiled class is a compiler synthesized class.

3. Update CompatDexBuilder to write the synthetic context information to some build meta info for later access by the DexFileSplitter. Something like this is already done for the desugar dependencies which are written to a proto in META-INF/desugar_deps.

4. Update DexFileSplitter to use the meta-inf of synthetic classes to group with the right contexts.

The D8 team would handle 1 and 2, and the bazel team and/or community could do 3 and 4. This avoids coupling to D8 implementation details and having to parse dex files and hopefully reduces the code complexity a bit.

ahumesky:
This should be fixed in Bazel 6.3.0:
https://github.com/bazelbuild/bazel/pull/18808

PR #16393, opened 2022-10-05
Remove --[no]use_action_cache from user manual because it is misleading

wisechengyi:
Per discussion in https://bazelbuild.slack.com/archives/CA31HN1T3/p1664993836113209

> Son Luong Ngoc
I think the doc should remove that flag, it's misleading. It does not completely disable usage of AC but only the local PAC (stored in 2 files)

### DIFF
--- a/site/en/docs/user-manual.md
+++ b/site/en/docs/user-manual.md
@@ -1403,12 +1403,6 @@ for comparisons.
 Temporary flag for testing bazel default visibility changes. Not intended for general use
 but documented for completeness' sake.
 
-#### `--[no]use_action_cache` {:#use-action-cache}
-
-This option is enabled by default. If disabled, Bazel will not use its local action cache.
-Disabling the local action cache saves memory and disk space for clean builds, but will make
-incremental builds slower.
-
 #### `--starlark_cpu_profile=_file_` {:#starlark-cpu-profile}
 
 This flag, whose value is the name of a file, causes Bazel to gather


sluongng:
Just to elaborate a bit: "Action Cache" in Bazel is often known to users as part of the AC that is implemented in Remote Cache (HTTP/ gRPC) or Disk Cache.

This flag does not affect the "Action Cache" mentioned above. Instead, it controls how Persistent Action Cache(PAC) works. PAC is a cache store that involves 2 files that map file digests to some actions and does not contain the usual metadata that is stored in a typical "AC".

Furthermore, this flag is marked as undocumented in the Java code, thus it does not show up in `bazel help` command line doc or anywhere else other than on bazel.build website.

For these reasons, I suggest removing this documentation, for now, to discourage non-advance users from using it. Future work could involve updating the naming conventions and documentation to better distinguish the 2 concepts of AC and PAC apart.

sluongng (approved):

gregestren:
I'm not sure what to think on this one. The flag is communicating something meaningful, and "action cache" has a clear meaning w.r.t. to how it's being used. But if "action cache" has started to become an overloaded term re: remote caches then yeah it sounds fair to clarify.

The stronger argument is how this doesn't appear in `blaze help`. But I don't know what standards are expected for what appears there vs. here.

I CC'ed `team-Remote-Exec` for input on the first issue.

coeuvre:
I agree with the first issue. As least for my observation, "Action Cache" is widely mentioned as AC in the remote cache in the community while the internal action cache is not widely discussed.

We should either remove it from the doc or, at least, make it more clear.

wisechengyi:
Hi @coeuvre, just checking in to see if there's any blocker for this to go in?

coeuvre (approved):
LGTM but I will let @gregestren to make the decision.

gregestren (approved):

Issue #15350, opened 2022-04-26
Bazel architecture changing doesn't rerun repository rules


keith:
### Description of the bug:

When switching between bazel architectures of the same bazel version on macOS M1 machines, repository rules do not get re-run, leading to potential failures. You have to run a `bazel clean --expunge` to recover.

```
% ./bazelisk-darwin run hello
INFO: Analyzed target //:hello (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:hello up-to-date:
  bazel-bin/hello_/hello
INFO: Elapsed time: 0.116s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
hi

% arch -x86_64 ./bazelisk-darwin run hello
Starting local Bazel server and connecting to it...
ERROR: /private/tmp/archexample/BUILD:3:10: While resolving toolchains for target //:hello: no matching toolchains found for types @io_bazel_rules_go//go:toolchain
ERROR: Analysis of target '//:hello' failed; build aborted:
INFO: Elapsed time: 9.674s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (35 packages loaded, 238 targets configured)
FAILED: Build did NOT complete successfully (35 packages loaded, 238 targets configured)
```

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Using rules_go, and this attached project:
[archexample.zip](https://github.com/bazelbuild/bazel/files/8564868/archexample.zip) on an M1 machine run:

```
./bazelisk-darwin run hello
```

and then

```
arch -x86_64 ./bazelisk-darwin run hello
```

Note: the bazelisk binary in this zip is the fat binary, but the same happens if you download the specific arches separately and run them.

### Which operating system are you running Bazel on?

macOS

### What is the output of `bazel info release`?

release 5.1.1

meteorcloudy:
I think the reason is that we never record the current architecture in the marker file of repository rules. Is there any environment variable we can declare as the dependency of repository rule here: https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl#L144 ?

brentleyjones:
Wouldn't we want all repository rules to rerun though? Seems we should put the arch in the marker.

keith:
[...]

PR #16417, opened 2022-10-06
Add host arch to repository rule marker

keith:
Fixes https://github.com/bazelbuild/bazel/issues/15350


jerrymarino:
@keith is there any downsides to this? E.g. having the same output on 2 different cpu types? In macOS you can have a fat binary and CPU independent rule

meteorcloudy:
@jerrymarino We discussed the pros and cons of this approach in https://github.com/bazelbuild/bazel/issues/15350, I think I'm generally fine with this. Since changing architecture on the same machine is a rare thing anyway, but requiring every repository rule author to remember declaring arch as a dependency is probably not very feasible.

meteorcloudy (approved):
Thanks!

PR #16418, opened 2022-10-06
Remove unused symbol_counts feature

keith:
This feature only applied if `symbol_counts_output` was set, which doesn't appear to be possible.


gregestren:
@oquenchil can you domain review?

oquenchil (approved):
LGTM

Issue #16422, opened 2022-10-07
Bazel is hanging when outputs got evicted from remote execution cache with --remote_download_toplevel enabled


k1nkreet:
### Description of the bug:

Since 5.3, Bazel is hanging infinitely when something got evicted from remote execution cache with Build without Bytes enabled. Before 5.3 it was raising `FileNotFoundException` and there is a [PR](https://github.com/bazelbuild/bazel/pull/14126) trying to use this behavior to overcome this issue of BwoB

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

I've created a repo with workspace and scenario reproducing this issue: https://github.com/k1nkreet/bazel-bwob-hanging-issue

### Which operating system are you running Bazel on?

Linux archlinux 5.18.7-arch1-1

### What is the output of `bazel info release`?

release 5.3.1

PR #16423, opened 2022-10-07
Fix bazel hanging when outputs got evicted from remote execution cache with --remote_download_toplevel enabled

k1nkreet:
#16422 This PR removes ignoring exception from upload task which leads to Bazel hanging in scenario represented in this issue


coeuvre:
Thanks! Can you please also add a test for this?

coeuvre:
I created https://github.com/bazelbuild/bazel/pull/16464 and included the test there.

k1nkreet:
Thanks!

Issue #15717, opened 2022-06-21
Add cquery support for --order_output flag


brenden:
### Description of the feature request:

Right now the ordering of the results from cquery cannot be controlled by `--order_output` (as they can be for query). It would be nice if this flag were supported on cquery as well.

### What underlying problem are you trying to solve with this feature?

I am using the output of `bazel query` in a CI system to identify which tests need to be run before merging each pull request. The process is:
  1. run `bazel query --order_output full 'deps(//...)'` on the PR branch
  2. run `bazel query --order_output full 'deps(//...)'` on the master branch
  3. using the two dependency graphs, identify all changed targets and find all test targets that are downstream of them

See [bazel-diff](https://github.com/Tinder/bazel-diff.git) for an implementation of similar logic.

I now want to replace `bazel query` with `bazel cquery` so that changes to configured dependencies will trigger tests to run. However `bazel cquery` does not support the `--order_output` flag and its default results are not in topological order.

### Which operating system are you running Bazel on?

macOS

### What is the output of `bazel info release`?

4.2.2

brenden:
Thanks for the fast reply! I'll take a swing at implementing this then.

gregestren:
If the change looks simple feel free to just put up a PR. If more complicated I suggest reviewing what you find here so we can do some design analysis before going too deeply down one route.

PR #16427, opened 2022-10-07
Adding `--order_output` flag to cquery and aquery

brenden:
This PR makes `--order_output` into a common query flag that can be used by `cquery` and `aquery`, not just `query`. For more context see issue #15717, which this PR fixes.


gregestren:
Huh. How hard did you find the implementation?

brenden:
@gregestren It was trickier than I expected because `query` doesn't share that much with `cquery`/`aquery`. LMK if you think I should back up and do a design review for this change.

src/main/java/com/google/devtools/build/lib/query2/PostAnalysisQueryEnvironment.java
```diff
@@ -575,6 +579,51 @@ public Collection<T> getSiblingTargetsInPackage(T target) throws QueryException
...
+      graph.createNode(result);
+    }
+
+    // Second pass: create edges between nodes
+    for (T result : resultList) {
+      for (T dep : getFwdDeps(ImmutableList.of(result))) {
+        if (graph.getNodeMaybe(dep) != null) {
+          graph.addEdge(
+                  graph.getNode(result),
+                  graph.getNode(dep)
```
gregestren, line 613: Do the results actually match the original list?

Ultimately we want the same nodes here as were in `resultList`. don't we? What happens if `resultList` is only a subset of the original graph? Like it applies a `deps` function but also a `kind` function to filter to only certain rule kinds?

If that is a thing, one option could be to post-filter the graph output to only those nodes in `resultList`.

src/main/java/com/google/devtools/build/lib/query2/PostAnalysisQueryEnvironment.java
```diff
@@ -575,6 +579,51 @@ public Collection<T> getSiblingTargetsInPackage(T target) throws QueryException
...
+    if (orderOutput == OrderOutput.NO) {
+      return resultList;
+    }
+
+    // Behavior of order_output=auto depends on the output format
+    if (orderOutput == OrderOutput.AUTO) {
+      orderOutput = this.getOutputFormat().endsWith("proto") ? OrderOutput.DEPS : OrderOutput.FULL;
+    }
+
+    Digraph<T> graph = new Digraph<>();
```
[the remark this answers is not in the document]
brenden: Unfortunately yes. I followed an approach similar to `GraphOutputFormatterCallback` but clearly this is not ideal.

src/main/java/com/google/devtools/build/lib/query2/aquery/ActionGraphQueryEnvironment.java
```diff
@@ -355,4 +357,22 @@ public ThreadSafeMutableSet<KeyedConfiguredTargetValue> createThreadSafeMutableS
   public void setActionFilters(AqueryActionFilter actionFilters) {
     this.actionFilters = actionFilters;
   }
+
+  class ConfiguredTargetValueOrdering implements Comparator<KeyedConfiguredTargetValue> {
+    @Override
+    public int compare(KeyedConfiguredTargetValue o1, KeyedConfiguredTargetValue o2) {
+      return o1.getConfiguredTarget().getOriginalLabel().compareTo(
+        o2.getConfiguredTarget().getOriginalLabel());
```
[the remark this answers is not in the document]
brenden: Yeah that's right, the configuration hashes need to be compared here too.

src/main/java/com/google/devtools/build/lib/query2/PostAnalysisQueryEnvironment.java
brenden: > Do the results actually match the original list?

I believe so. The only place nodes are created is in the block above (lines 603-605) where one node is created per item in `resultList`. Here we're just adding edges between nodes but not creating any new ones.

gregestren: Oh, I see. Line 610 filters out the possibility of extra nodes?

What happens if the query applies filtering down the transitive closure? Like `kind("cc_library", deps(//foo))`? Where `//foo` has no direct deps on `cc_library` targets but has transitive ones?

brenden: In that case the ordering between `//foo` and the `cc_library` targets would not be preserved because the intermediate targets would not be present in `resultList`. If we need to preserve the pre-filtering ordering we will have to move this logic to an earlier point in the query (which would also address [this issue](https://github.com/bazelbuild/bazel/pull/16427/files/2b8a1c1e9c87b5b263a4264bc95d472a4b3056f3#r1003593251)).

gregestren: My concern is that we want to be able to consolidate what's already complex and subtle behavior. The main documentation - at https://bazel.build/query/language#results-ordering - we shouldn't skew too much in the cquery / aquery case. If cquery/aquery only supports a subset of the settings that's fine. But if each value has a different interpretation that's going to get confusing.

It's a naive followup, but can you clarify your original use case? That you needed a basic topological ordering to be able to do diffs, to identify changes and which tests are relevant? What does a simple example look like?

brenden: Yeah that makes sense. Test selection was the initial motivation for adding this support but that's actually no longer an issue for us. As this PR would need some major changes that I don't have bandwidth to work on now I'll close it out.

Issue #16379, opened 2022-10-03
Runfile collisions across transitions


octalthorpe:
### Description of the bug:

When building up runfiles there can be cases where the 'short_path' is identical
yet the actual file is different due to transitions. So when a rule is building up runfiles using ctx.runfiles
only one of the files that collides will be present in the runfiles.  
The collision can be silent, giving no errors or indications it has occurred.

The attached zip contains a project workspace that recreates an issue with runfiles collisions when there are transitions involved.

For example:
```
DEBUG: /Users/****/repos/overwrote_runfiles/rules.bzl:92:18: File:<generated file hello_world.h> short_path:hello_world.h path:bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/hello_world.h
DEBUG: /Users/****/repos/overwrote_runfiles/rules.bzl:92:18: File:<generated file hello_world.h> short_path:hello_world.h path:bazel-out/darwin_arm64-fastbuild-ST-3af09e09b98c/bin/hello_world.h
```

This workspace is likely more complex than needs to be but it approximates a real world scenario that
exhibited these collisions.

The basics of it are there is a reporting rule that has dependencies on test rules.  These test rules
run binaries but each test uses different transitions, one is the 'exec' and another a custom transition.
So the outputs of the binary are dependant on the incoming transition.

An aspect collects the source from the test targets. The reporting rule in this example does not do
anything with the sources, but in our real implementation the reporting gear uses sources, 
objects and test outputs to create a global report about all the tests.

The issue with the collisions causes the following to happen when runfiles are created:

1. Silent collisions: Last file in the set wins
2. Symlink 'overwrote runfile': When creating runfiles with 'symlinks' then the collisions are logged as a error but do not fail. See https://github.com/bazelbuild/bazel/issues/4965

For our implementation we were able to squelch the error by filtering before creating runfiles, detecting collisions, and
only using the last File in the runfiles.  This only works because the files that are produced as of the transitions
happen to be identical. This may not be true for other situations as the transition might cause differences in the
files created.

To recreate the issue
```
bazel run -s //:report
```

Debug statements show the File objects that ultimately end up as runfiles to the reporting rule.

```
DEBUG: /Users/****/repos/overwrote_runfiles/rules.bzl:92:18: File:<generated file hello_world.h> short_path:hello_world.h path:bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/hello_world.h
[...]

Issue #15029, opened 2022-03-12
Runfiles workspace directories use wrong workspace name if repository renaming is in effect


phst:
### Description of the problem / feature request:

If a workspace B depends on a workspace A, but uses the name C instead of A to name the workspace, the runfiles for workspace A will appear as C in the runfiles directory, breaking runfiles lookup for any library in workspace A.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

- /tmp/test/a/WORKSPACE:

  ```python
  workspace(name = "a")
  ```

- /tmp/test/a/BUILD:

  ```python
  cc_library(
      name = "lib",
      srcs = ["lib.cc"],
      hdrs = ["lib.h"],
      data = ["data.txt"],
      visibility = ["//visibility:public"],
      deps = ["@bazel_tools//tools/cpp/runfiles"],
  )
  ```

- /tmp/test/a/lib.h:

  ```c++
  #include <string>
  void Test(const std::string& argv0);
  ```

- /tmp/test/a/lib.cc:

  ```c++
  #include "lib.h"

  #include <exception>
  #include <fstream>
  #include <iostream>
  #include <string>

  #include "tools/cpp/runfiles/runfiles.h"

  void Test(const std::string& argv0) {
    using bazel::tools::cpp::runfiles::Runfiles;
    const Runfiles* runfiles = Runfiles::Create(argv0);
    if (runfiles == nullptr) throw std::runtime_error("missing runfiles");
    std::string filename = runfiles->Rlocation("a/data.txt");
    if (filename.empty()) throw std::runtime_error("can't find runfile");
    std::ifstream stream(filename);
    stream.exceptions(std::ios::failbit | std::ios::badbit);
    if (!stream.is_open()) throw std::runtime_error("can't open runfile");
    std::cout.exceptions(std::ios::failbit | std::ios::badbit);
    std::cout << stream.rdbuf() << std::endl;
  }
  ```

- /tmp/test/a/data.txt:

  ```none
  hello world
  ```

- /tmp/test/b/WORKSPACE:

  ```python
  workspace(name = "b")

  local_repository(
      name = "c",  ## "a"
      path = "/tmp/test/a",
      repo_mapping = {"@c": "@a"},
  )
  ```

- /tmp/test/b/BUILD:

  ```python
  cc_binary(
      name = "bin",
      srcs = ["bin.cc"],
      deps = ["@c//:lib"],  ## "@a//:lib"
  )
[...]

PR #16428, opened 2022-10-08
Add `$(rlocationpath(s) ...)` expansion

### LINKED ISSUE #10923: Get the name of current repository from macro
I am trying to get the name of the current repository from a macro.
```
.
├── a
│   ├── BUILD.bazel
│   ├── WORKSPACE
│   ├── empty.txt
│   └── txt_files.bzl
└── b
    ├── BUILD.bazel
    ├── WORKSPACE
    └── empty2.txt
```
_(see full example [here](https://github.com/or-shachar/bazel_repository_name_issue) )_

I want that if repo `b` will call the macro in `@a//:txt_files.bzl` then somehow I'd be able to retrieve that it was called from repo `b`.

Right now (not sure if it's a bug or work as intended) - calling `native.repository_name()` will return `@` and not the name of repo as declared in its `WORKSPACE` file.

I want to change the behavior of a macro based on the callee repository.

Can we allow `native.repository_name()` to return the explicit name of the repository and not `@`?  To maintain backward compatibility - I assume it would be better if we add a flag to this method that will be turned off by default.

fmeum:
The new location expansion pattern `rlocationpath` and its plural version `rlocationpaths` resolve to the path(s) suitable for the Rlocation function offered by runfiles libraries.

Compared to the `rootpath` pattern, which can returns `../repo_name/pkg/file` for a file in an external repository and `pkg/file` for a file in the main repository, the path returned by `rlocationpath` is always of the form `repo_name/pkg/file`.

RELNOTES: The new path variable `$(rlocationpath ...)` and its plural form `$(rlocationpaths ...)` can be used to expand labels to the paths accepted by the `Rlocation` function of runfiles libraries. This is the preferred way to access data dependencies at runtime and works on all platforms, even when runfiles are not enabled (e.g., on Windows by default).

Work towards #16124
Fixes #10923

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java
@@ -17,6 +17,7 @@ package com.google.devtools.build.lib.analysis;
 import static java.util.stream.Collectors.joining;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableCollection;
@@ -26,7 +27,9 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.devtools.build.lib.actions.Artifact;
+import com.google.devtools.build.lib.analysis.LocationExpander.LocationFunction.PathType;
 import com.google.devtools.build.lib.cmdline.Label;
+import com.google.devtools.build.lib.cmdline.LabelConstants;
 import com.google.devtools.build.lib.cmdline.LabelSyntaxException;
 import com.google.devtools.build.lib.cmdline.RepositoryMapping;
 import com.google.devtools.build.lib.packages.BuildType;
@@ -63,34 +66,35 @@ public final class LocationExpander {
   private static final boolean EXACTLY_ONE = false;
   private static final boolean ALLOW_MULTIPLE = true;
 
-  private static final boolean USE_LOCATION_PATHS = false;
-  private static final boolean USE_EXEC_PATHS = true;
-
   private final RuleErrorConsumer ruleErrorConsumer;
   private final ImmutableMap<String, LocationFunction> functions;
   private final RepositoryMapping repositoryMapping;
+  private final String workspaceRunfilesDirectory;
 
   @VisibleForTesting
   LocationExpander(
       RuleErrorConsumer ruleErrorConsumer,
       Map<String, LocationFunction> functions,
-      RepositoryMapping repositoryMapping) {
+      RepositoryMapping repositoryMapping,
+      String workspaceRunfilesDirectory) {
     this.ruleErrorConsumer = ruleErrorConsumer;
     this.functions = ImmutableMap.copyOf(functions);
     this.repositoryMapping = repositoryMapping;
+    this.workspaceRunfilesDirectory = workspaceRunfilesDirectory;
   }
 
   private LocationExpander(
-      RuleErrorConsumer ruleErrorConsumer,
+      RuleContext ruleContext,
       Label root,
       Supplier<Map<Label, Collection<Artifact>>> locationMap,
       boolean execPaths,
       boolean legacyExternalRunfiles,
       RepositoryMapping repositoryMapping) {
     this(
-        ruleErrorConsumer,
+        ruleContext,
         allLocationFunctions(root, locationMap, execPaths, legacyExternalRunfiles),
-        repositoryMapping);
+        repositoryMapping,
+        ruleContext.getWorkspaceName());
   }
 
   /**
@@ -204,7 +208,7 @@ public final class LocationExpander {
       // (2) Call appropriate function to obtain string replacement.
       String functionValue = value.substring(nextWhitespace + 1, end).trim();
       try {
-        String replacement = functions.get(fname).apply(functionValue, repositoryMapping);
+        String replacement = functions.get(fname).apply(functionValue, repositoryMapping, workspaceRunfilesDirectory);
         result.append(replacement);
       } catch (IllegalStateException ise) {
         reporter.report(ise.getMessage());
@@ -232,23 +236,29 @@ public final class LocationExpander {
 
   @VisibleForTesting
   static final class LocationFunction {
+    enum PathType {
+      LOCATION,
+      EXEC,
+      RLOCATION,
+    }
+
     private static final int MAX_PATHS_SHOWN = 5;
 
     private final Label root;
     private final Supplier<Map<Label, Collection<Artifact>>> locationMapSupplier;
-    private final boolean execPaths;
+    private final PathType pathType;
     private final boolean legacyExternalRunfiles;
     private final boolean multiple;
 
     LocationFunction(
-        Label root,
-        Supplier<Map<Label, Collection<Artifact>>> locationMapSupplier,
-        boolean execPaths,
-        boolean legacyExternalRunfiles,
-        boolean multiple) {
+      Label root,
+      Supplier<Map<Label, Collection<Artifact>>> locationMapSupplier,
+      PathType pathType,
+      boolean legacyExternalRunfiles,
+      boolean multiple) {
       this.root = root;
       this.locationMapSupplier = locationMapSupplier;
-      this.execPaths = execPaths;
+      this.pathType = Preconditions.checkNotNull(pathType);
       this.legacyExternalRunfiles = legacyExternalRunfiles;
       this.multiple = multiple;
     }
@@ -259,10 +269,11 @@ public final class LocationExpander {
      * using the {@code repositoryMapping}.
      *
      * @param arg The label-like string to be expanded, e.g. ":foo" or "//foo:bar"
-     * @param repositoryMapping map of {@code RepositoryName}s defined in the main workspace
+     * @param repositoryMapping map of apparent repository names to {@code RepositoryName}s
+     * @param workspaceRunfilesDirectory name of the runfiles directory corresponding to the main repository
      * @return The expanded value
      */
-    public String apply(String arg, RepositoryMapping repositoryMapping) {
+    public String apply(String arg, RepositoryMapping repositoryMapping, String workspaceRunfilesDirectory) {
       Label label;
       try {
         label = root.getRelativeWithRemapping(arg, repositoryMapping);
@@ -271,14 +282,14 @@ public final class LocationExpander {
             String.format(
                 "invalid label in %s expression: %s", functionName(), e.getMessage()), e);
       }
-      Collection<String> paths = resolveLabel(label);
+      Collection<String> paths = resolveLabel(label, workspaceRunfilesDirectory);
       return joinPaths(paths);
     }
 
     /**
      * Returns all target location(s) of the given label.
      */
-    private Collection<String> resolveLabel(Label unresolved) throws IllegalStateException {
+    private Collection<String> resolveLabel(Label unresolved, String workspaceRunfilesDirectory) throws IllegalStateException {
       Collection<Artifact> artifacts = locationMapSupplier.get().get(unresolved);
 
       if (artifacts == null) {
@@ -288,7 +299,7 @@ public final class LocationExpander {
                 unresolved, functionName()));
       }
 
-      Set<String> paths = getPaths(artifacts);
+      Set<String> paths = getPaths(artifacts, workspaceRunfilesDirectory);
       if (paths.isEmpty()) {
         throw new IllegalStateException(
             String.format(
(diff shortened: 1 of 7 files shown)


fmeum:
@lberki Can you review this (or delegate to someone else)?

I haven't included any changes to docs yet, but could add them if the rest of the PR looks good. There are no docs on runfiles libraries yet though, so whatever I include in this PR will still be lacking context.

comius (changes requested):
I was thinking of alternatives and I couldn't come up with any.

If I understand correctly, please confirm, rlocation always needs `repo_name/path/in/the/repo` as input. And this can be `_main/path/in/the/main/repo`. Right? What happens in the case this is a generated file - that is in `bazel-bin/reponame/...`?

Possible alternative would be `$(reponame @repo//label)`, but this worse since you'd need to use 2 functions.

About the name. `location(s)` function is deprecated. I'm guessing you chose the name so it fits `rlocation` function (for example the one here https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash). So this then looks like `rlocation $(rlocation //label)`. Does the plural form make sense here?

Because of deprecation of `location`, perhaps consider another name, like `repopath`? Maybe `rlocation $(repopath //label)` looks better?

Slightly out of scope, but do you think there's a way to simplify copy-pasted lines we have in almost all shell tests https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/archive_contents_test.sh#L20-L40 or https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/android/android_helper.sh#L17-L26? 

One more naive question. Can we provide a function that directly returns what `rlocation $(rlocation //label)` would return? Eliminating the need for first/external `rlocation`.

comius:
Also note, the name might cause some confusion when bash's `$()` is used. If you're doing substitutions you'll need something like `$$(rlocation $(rlocation //label))` where inner one is substituted and `$$` is unescaped.

fmeum:
> If I understand correctly, please confirm, rlocation always needs `repo_name/path/in/the/repo` as input.
Exactly.

> And this can be `_main/path/in/the/main/repo`. Right? 

Yes, for all files in the main repo, with Bzlmod enabled. Without Bzlmod, this would be `__main__/path` or `<workspace name>/path` (if a name is set).

> What happens in the case this is a generated file - that is in `bazel-bin/reponame/...`?

Both generated and source file's use the same `repo/package/path/in/repo/name` path structure, just like in the runfiles directory.

> Possible alternative would be `$(reponame @repo//label)`, but this worse since you'd need to use 2 functions.

It's also difficult to get the remaining part since `rootpath` for an external repository doesn't just return `package/path/in/repo/name`, AFAIK it returns `../repo/package/path/in/repo/name`.

> About the name. `location(s)` function is deprecated. I'm guessing you chose the name so it fits `rlocation` function (for example the one here https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash). So this then looks like `rlocation $(rlocation //label)`.

I chose `rlocationpath` instead of `rlocation` for two reasons: 

1. To prevent confusing this with the deprecated `location` function, making it more similar to the recommended `{exec,root}path` functions.
2. To distinguish it from the `rlocation` functions defined by runfiles libraries, which are supposed to take whatever `$(rlocationpath //label)` evaluates to.

Your example would be `rlocation $1` with `args = ["$(rlocationpath //label)"]`. 

> Does the plural form make sense here?

I think so, yes. In Java, it would be easy to do `Arrays.stream(args).map(arg -> r.rlocation(arg))`, making use of the implicit shell tokenization of the `args` attribute.


> Because of deprecation of `location`, perhaps consider another name, like `repopath`? Maybe `rlocation $(repopath //label)` looks better?

See above, do you think `rlocationpath` does a sufficient job at preventing this confusion? This expansion has one very clear purpose, providing the correct path for the `rlocation` function, so I think that `rlocation` should be in the name - otherwise it will be very hard to distinguish `execpath` and `rootpath` from the new expansion.

> 
> Slightly out of scope, but do you think there's a way to simplify copy-pasted lines we have in almost all shell tests https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/archive_contents_test.sh#L20-L40 or https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/android/android_helper.sh#L17-L26?

The latter one is more modern and already quite a bit shorter. I don't see a way to fundamentally improve on that.

fmeum:
> One more naive question. Can we provide a function that directly returns what rlocation $(rlocation //label) would return? Eliminating the need for first/external rlocation.

> Also note, the name might cause some confusion when bash's $() is used. If you're doing substitutions you'll need something like $$(rlocation $(rlocation //label)) where inner one is substituted and $$ is unescaped.

Answering these in one as they are related: You wouldn't commonly do this since `$(rlocationpath //label)` has to be evaluated by Bazel, whereas `$(rlocation repo/some/path)` is an actual Bash function call happening at runtime. 

We *could* think about ways to make `$(rlocation //label)` work from Bash directly, but that would require major changes to the runfiles tree - essentially, we would need to dump an additional manifest file that lists the contents of `DefaultInfo.files` for all Bazel targets contributing runfiles to a given executable. @lberki who has been interested in improving the usability of runfiles in this direction.

src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java
```diff
@@ -348,27 +375,35 @@ static ImmutableMap<String, LocationFunction> allLocationFunctions(
...
+            new LocationFunction(root, locationMap, execPaths ? PathType.EXEC : PathType.LOCATION, legacyExternalRunfiles, EXACTLY_ONE))
         .put(
             "locations",
             new LocationFunction(
-                root, locationMap, execPaths, legacyExternalRunfiles, ALLOW_MULTIPLE))
+                root, locationMap, execPaths ? PathType.EXEC : PathType.LOCATION, legacyExternalRunfiles, ALLOW_MULTIPLE))
         .put(
             "rootpath",
             new LocationFunction(
-                root, locationMap, USE_LOCATION_PATHS, legacyExternalRunfiles, EXACTLY_ONE))
```
comius, line 359: Can you remove USE_EXEC_PATHS, and USE_LOCATION_PATHS. It seems they are not used anymore.

src/test/java/com/google/devtools/build/lib/analysis/LocationExpanderIntegrationTest.java
```diff
@@ -181,8 +189,10 @@ public void otherPathExternalExpansionNoLegacyExternalRunfiles() throws Exceptio
         .matches("foo .*-out/.*/external/r/p/foo\\.txt bar");
     assertThat(expander.expand("foo $(execpaths @r//p:foo) bar"))
         .matches("foo .*-out/.*/external/r/p/foo\\.txt bar");
-    assertThat(expander.expand("foo $(rootpath @r//p:foo) bar")).matches("foo ../r/p/foo.txt bar");
-    assertThat(expander.expand("foo $(rootpaths @r//p:foo) bar")).matches("foo ../r/p/foo.txt bar");
+    assertThat(expander.expand("foo $(rootpath @r//p:foo) bar")).isEqualTo("foo ../r/p/foo.txt bar");
```
comius, line 192: nit: `isEqualTo` seems better here, since it's not using a regex like `matches`. Can you use `isEqualsTo` in all the new code?

The old code should be updated separately, so we don't overload this PR.

src/main/java/com/google/devtools/build/lib/analysis/LocationExpander.java
```diff
@@ -313,24 +327,37 @@ private Collection<String> resolveLabel(Label unresolved) throws IllegalStateExc
...
 
+    private PathFragment getPath(Artifact artifact, String workspaceRunfilesDirectory) {
+      switch (pathType) {
+        case LOCATION:
+          return legacyExternalRunfiles ? artifact.getPathForLocationExpansion() : artifact.getRunfilesPath();
+        case EXEC:
+          return artifact.getExecPath();
+        case RLOCATION:
+          PathFragment runfilesPath = artifact.getRunfilesPath();
+          if (runfilesPath.startsWith(LabelConstants.EXTERNAL_RUNFILES_PATH_PREFIX)) {
```
comius, line 352: I see this works with and without siblings repositories enabled. But how? (I'm just curious, no action needed).

src/test/java/com/google/devtools/build/lib/analysis/LocationExpanderIntegrationTest.java
```diff
@@ -158,6 +162,10 @@ public void otherPathExternalExpansion() throws Exception {
         .matches("foo external/r/p/foo.txt bar");
     assertThat(expander.expand("foo $(rootpaths @r//p:foo) bar"))
         .matches("foo external/r/p/foo.txt bar");
+    assertThat(expander.expand("foo $(rlocationpath @r//p:foo) bar"))
+      .matches("foo r/p/foo.txt bar");
+    assertThat(expander.expand("foo $(rlocationpath @r//p:foo) bar"))
```
comius, line 167: Please add test for `$(rlocationpath @//p:foo)` and `$(rlocationpath @main_repo//p:foo)`, where `main_repo` matches main repo's name. 

Not necessarily here, please pick the best location for those.

comius (changes requested):
Thanks for explanations! This makes it much easier to review. Couple of nits, but overall it looks ok.

fmeum (commented):
I addressed your comments.

I also added `rlocationpath` to the docs and fixed some outdated info about the other location functions.

fmeum:
@bazel-io flag

comius:
@sgowroji there might be some trouble merging `make-variables.vm`

sgowroji:
Hello @comius, Above PR is awaiting to merge. What should be done for the internal fixes which we discussed offline, Can you please have a look. Thanks!

meteorcloudy:
@bazel-io fork 6.0.0

comius (changes requested):
Hey I had one more thought. Is it possible to put this function under an experimental flag or only enable together with bzlmod? What I'm worried is that direction on how to fix runfiles might still change - considering there are new bugs showing up. Having it behind a flag might make it easier to change if needed.

fmeum:
> Hey I had one more thought. Is it possible to put this function under an experimental flag or only enable together with bzlmod? 

Could you explain why you think this is a risky feature to add unconditionally? The paths it returns are tied to a stable, in-use feature, the runfiles libraries. If we introduce a new way to access runfiles in the future and deprecate the libraries, we will have to offer a migration plan for users anyway. Whether we have `$(rlocationpath ..)` around as a stable feature shouldn't affect that migration since it doesn't do anything by itself - we could just make it undocumented and phase it out together with the libraries. Do you see any additional maintenance burden in this new location expansion?

I am asking because guarding this feature behind a flag would make rulesets essentially unable to rely on it. Since `--enable_blzmod` has other effects, we would need a separate flag. In fact, the main use case of this feature would be for existing `WORKSPACE` projects, where the workspace runfiles directory name isn't fixed and thus not easy to obtain from a macro (see https://github.com/bazelbuild/bazel/issues/10923).

> What I'm worried is that direction on how to fix runfiles might still change - considering there are new bugs showing up.

Which new bugs are you referring to? We still have the ongoing discussion in https://github.com/bazelbuild/bazel/issues/15029, but I am not aware of any bugs.

comius:
> Could you explain why you think this is a risky feature to add unconditionally? The paths it returns are tied to a stable, in-use feature, the runfiles libraries. If we introduce a new way to access runfiles in the future and deprecate the libraries, we will have to offer a migration plan for users anyway. Whether we have `$(rlocationpath ..)` around as a stable feature shouldn't affect that migration since it doesn't do anything by itself - we could just make it undocumented and phase it out together with the libraries. Do you see any additional maintenance burden in this new location expansion?

We should consider adding any new features risky. We often didn't do that, arguing the new feature is completely foolproof. For example rule.name. This sometimes resulted in a lot of effort to do a cleanup. The users should be aware this is new and we should have some incubation period to derisk.

I think that things around bzlmod and runfiles might change as we discover new problems. Or are you certain you already have an undoubtedly optimal and perfect solution?

> 
> I am asking because guarding this feature behind a flag would make rulesets essentially unable to rely on it. Since `--enable_blzmod` has other effects, we would need a separate flag. In fact, the main use case of this feature would be for existing `WORKSPACE` projects, where the workspace runfiles directory name isn't fixed and thus not easy to obtain from a macro (see #10923).
> 
> > What I'm worried is that direction on how to fix runfiles might still change - considering there are new bugs showing up.
> 
> Which new bugs are you referring to? We still have the ongoing discussion in #15029, but I am not aware of any bugs.

I was referring to https://github.com/bazelbuild/bazel/issues/16379.

fmeum:
@comius I am definitely _not_ convinced that the current approach to runfiles discovery is optimal and do think that more fundamental changes to this mechanism aren't unlikely. I greatly value your caution when it comes to proposed API changes!

I do think that the _incremental_ risk of adding `$(rlocationpath ...)` is very low though: It is only useful in conjunction with the current approach to runfiles (but in that case, it is very useful), so it would live and die with that without causing additional migration efforts. The code introduced in this PR is also not very complex, I would argue that the switch to an enum even makes the existing code more readable.

I actually think that this PR could simplify potential future migrations as using `$(rlocationpath ...)` sidesteps the entire "repo mapping manifest applied to hardcoded runfiles paths" thing and instead emits a path under our control that could be changed transparently in future releases (we could add a note to the docs stating clearly that the precise nature of the returned path is not defined, but instead is whatever `Rlocation` accepts).

What do you think? If you still aren't convinced, I would introduce this behind an `--experimental_enable_rlocationpath` flag.

comius:
> The code introduced in this PR is also not very complex, I would argue that the switch to an enum even makes the existing code more readable.

I agree.


> I actually think that this PR could simplify potential future migrations as using `$(rlocationpath ...)` sidesteps the entire "repo mapping manifest applied to hardcoded runfiles paths" thing and instead emits a path under our control that could be changed transparently in future releases (we could add a note to the docs stating clearly that the precise nature of the returned path is not defined, but instead is whatever `Rlocation` accepts).

Agreed. It's just a conversion from a label to a string another function accepts. Sidestepping repo mapping sounds inviting :)
 
> What do you think? If you still aren't convinced, I would introduce this behind an `--experimental_enable_rlocationpath` flag.

No this won't be necessary.

fmeum:
@comius If you have no further converns, could you approve? This is the final piece of API I need to be certain about to finalize my BazelCon talk.

meteorcloudy:
@fmeum Ivo is already importing this PR ;)

kshyanashree:
@bazel-io fork 5.4.0

kshyanashree:
@bazel-io fork 5.4.0

kshyanashree:
@bazel-io fork 5.4.0

kshyanashree:
@bazel-io fork 5.4.0

kshyanashree:
@bazel-io fork 5.4.0

PR #16429, opened 2022-10-08
Fix index in `Sequence#cast` error message always being 0

fmeum:
The index into the list wasn't incremented, resulting in a confusing error message.


oquenchil (approved):
LGTM

Issue #12313, opened 2020-10-20
*_binary args get splitted


dmivankov:
### Description of the problem / feature request:

Arguments set by args attribute of *_binary get splitted

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
BUILD:
```python
cc_binary(
  name = "main",
  srcs = [":main.c"],
)
cc_binary(
  name = "main2",
  srcs = [":main.c"],
  args = ["a", "b c"], # "b c" gets split
)
cc_binary(
  name = "main3",
  srcs = [":main.c"],
  args = ["a", "\"b c\""], # excessive quoting is needed
)
cc_binary(
  name = "main4",
  srcs = [":main.c"],
  args = ["a \"b c\""], # we could say singleton lists are preferred and are splitted
)
```
main.c:
```c
#include <stdio.h>
int main(int argc, char** argv) {
	for (int i = 1; i < argc; ++i) {
		printf("%d: %s\n", i, argv[i]);
	}
	return 0;
}
```
Testing:
```sh
$ bazel run :main -- a "b c"
1: a
2: b c
$ bazel run :main2
1: a
2: b
3: c
$ bazel run :main3
1: a
2: b c
$ bazel run :main4
1: a
2: b c
```
### What operating system are you running Bazel on?

Linux/NixOS

### What's the output of `bazel info release`?

release 3.3.1- (@non-git)

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

Comes from NixOS packages
nix-env -iA nixpkgs.bazel

### Any other information, logs, or outputs that you want to share?

Luckily
```python
cc_binary(
  name = "main5",
  srcs = [":main.c"],
  args = ["a \"b c\" >/dev/null"],
)
```
treats ">/dev/null" as string rather than redirection. Could've been an injection vulnerability otherwise

Also tested with recent git version 6bf44ba9339a59c852a5da5f72e14bdb87af6483

dmivankov:
extra info
```sh
$ bazel build :main
$ bazel run --shell_executable=$(pwd)/bazel-bin/main :main2
1: -c
2: SOME_PATH/execroot/__main__/bazel-out/k8-fastbuild/bin/main2 a b c
```

dmivankov:
Looks like the issue is
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java#L178
calling
[...]

PR #16430, opened 2022-10-08
WIP: Add `arguments` parameter to `RunEnvironmentInfo`

### LINKED ISSUE #16076: Add `arguments` parameter to `testing.TestEnvironment`

I would like to write rules which contain an `args` attribute where users can define a test with unique command line arguments. I think it would make the most sense to expand [testing.TestEnvironment](https://bazel.build/rules/lib/testing#TestEnvironment) to support some `arguments` field where users can provide a list of arguments (or, stretch goal, an [Args](https://bazel.build/rules/lib/Args) object) to describe arguments to be passed to the test when it starts.

To accomplish this I'm constantly needing to have my tests execute a platform specific bash/batch script that has the arguments hard coded into it but this has an unfortunate drawback that I can no longer use [--run_under](https://bazel.build/reference/command-line-reference#flag--run_under) on the actual test binary. Over all, the complexity needed to solve for this feel unjustified and that instead, I should be able to describe in a provider some arguments to pass to my test.

Linux, MacOS, Windows

release 5.2.0

_No response_

_No response_

_No response_

_No response_

fmeum:
Executable Starlark rules can use the new `arguments` parameter on `RunEnvironmentInfo` to specify the arguments that Bazel should pass on the command line with `test` or `run`.

If set to a non-`None` value, this parameter overrides the value of the `args` attribute that is implicitly defined for all rules. This allows Starlark rules to implement their own version of this attribute which isn't bound to its proprietary processing (data label expansion and tokenization).

Along the way, this commit adds test coverage and documentation for the interplay between `RunEnvironmentInfo`'s `environment` and `--test_env`.

The value of the `arguments` field of `RunEnvironmentInfo` is intentionally not exposed to Starlark yet: It is not clear how these arguments should be represented and whether rules relying on the magic `args` attribute should also provide this field.

RELNOTES: Executable starlark rules can use the `arguments` parameter of `RunEnvironmentInfo` to specify their command-line arguments with `bazel run` and `bazel test`.

Fixes #16076
Work towards #12313


fmeum:
@comius Would you be available to review this PR?

lberki:
Adding @comius since he reviewed https://github.com/bazelbuild/bazel/pull/15232 .

@comius : do I understand correctly that `RunEnvironmentInfo` is intended to be only consumed within the current rule?

Then it's kind of odd that it's added to the set of providers the rule makes accessible to rules that depend on it; it seems unnecessary and it's Yet Another Little Thing on the interface of Bazel. I suppose we can't remove it anymore without an incompatible flag? :(

fmeum:
@lberki I should add that advertising `RunEnvironmentInfo` was a conscious choice back then: https://github.com/bazelbuild/bazel/issues/15224, https://github.com/bazelbuild/bazel/issues/15225, and my own experience working with wrapper rules that apply transitions lead me to the opinion that Starlark-constructible providers should also be Starlark-retrievable. Otherwise, they can't be tested properly and essentially make any rules that use them "terminal" as wrappers can't read them.

I do agree that this extends the interface, but maybe that extension isn't unnecessary :-)

UebelAndre:
@comius sorry for the ping. Would you happen to have bandwidth this week to review this PR?

comius:
> This allows Starlark rules to implement their own version of this [args] attribute which isn't bound to its proprietary processing (data label expansion and tokenization).

I don't think this is the direction we should be following. Adding a way to customise args attribute will result in "proprietary processing" invented by each rule author for their rules - or copy-pasting the code around that does it.

I'd rather pursue the direction where the `args` attribute is parsed in a general / straight forward non-proprietary way.

Can you close this PR and rather introduce a flag and optional allow list, that fixes the `args` parsing? Unless I'm missing something a list of strings should be straightforward to convert to a list of args.

Are there other arguments for customization? Otherwise, I really don't want users to figure out on which rules args work and on which rules they don't.

UebelAndre:
> Can you close this PR and rather introduce a flag and optional allow list, that fixes the `args` parsing?

Can you expand on what you mean by "flag and optional allow list"?

> Are there other arguments for customization? Otherwise, I really don't want users to figure out on which rules args work and on which rules they don't.

My use case for this functionality is to be able to write tests rules with unique attributes that are translated to arguments to a test binary. `args` is intentionally not exposed in these test rules as it invites users to do things that break the determinism of the targets and introduces cache poisoning. This feature would be very valuable in helping me cleanup code specifically for translating attributes into arguments for a test binary, which is unreasonably difficult to do today.

UebelAndre:
> Why can't this be done with a macro?
> 
> If I understand you correctly, you want to translate "unique attributes to arguments". Isn't that:
> 
> ```
> def my_test_macro(arg1 = 1, arg2 = "str", arg3 = [3,4,5], **kwargs):
>   my_test(args = [str(arg1), arg2] + [str(a) for a in arg3], **kwargs)
> ```

Some arguments accept paths to files which some users would expect to be able to pass a label to. The macro would not work with select statements which is something I need on a number of my rules.

comius:
> Some arguments accept paths to files which some users would expect to be able to pass a label to. 
> The macro would not work with select statements which is something I need on a number of my rules.

You mean something like this?
```
def my_test_macro(file, **kwargs):
  my_test(args = ["$(location %s)" % file], **kwargs)
```

```
my_test_macro(name = "a", file = select({"...": "//file1", "...": "//file2"})
```

UebelAndre:
> > Some arguments accept paths to files which some users would expect to be able to pass a label to.
> > The macro would not work with select statements which is something I need on a number of my rules.
> 
> You mean something like this?
> 
> ```
> def my_test_macro(file, **kwargs):
>   my_test(args = ["$(location %s)" % file], **kwargs)
> ```
> 
> ```
> my_test_macro(name = "a", file = select({"...": "//file1", "...": "//file2"})
> ```

I don't think you can pass selects to expansions like that. Is there evidence of that working?

Additionally, there may be things only knowable within a rule (like whether or not coverage is enabled) that would also need to be translated into an arugment.

UebelAndre:
> You mean something like this?
> 
> ```
> def my_test_macro(file, **kwargs):
>   my_test(args = ["$(location %s)" % file], **kwargs)
> ```
> 
> ```
> my_test_macro(name = "a", file = select({"...": "//file1", "...": "//file2"})
> ```

This also doesn't seem like it'd support having lists of files expanded into `--flag=$(file)` arguments. Is that correct?

UebelAndre:
I also have cases where I have a custom set of `foo_library`, `foo_binary`, and `foo_test` targets where the tests take information from `FooInfo` providers in the library and binary targets and consume them in the tests. This would certainly not be doable via macros.

comius:
> I don't think you can pass selects to expansions like that. Is there evidence of that working?

No, it's not working. What you can do if you need a case like this, is pass dict, and then wrap it into a `select` within a macro.

> This also doesn't seem like it'd support having lists of files expanded into --flag=$(file) arguments. Is that correct?

You can implement expansion like that in a macro.

> I also have cases where I have a custom set of foo_library, foo_binary, and foo_test targets where the tests take information from FooInfo providers in the library and binary targets and consume them in the tests. This would certainly not be doable via macros.

But it's completely doable with a second rule. You can pass foo_* to that rule as a label. Access the providers. And also execute it using `ctx.actions.run`. It seems to me the only difference is having 1 vs. 2 rules.

If the usability is problematic, you can wrap 2 rules together into a macro, to achieve the appearance of a single rule.

UebelAndre:
Something being “doable” I don’t think is reason enough to reject improvements. I’ve proven what I want is doable since have the functionality I want. But I continue to see other developers attempt to solve for the same gap in functionality in different ways which leads to very hard to maintain projects. I have other examples of this but feel sharing them won’t change the outcome of the PR so I respectfully express my disagreement with the decision to reject this PR and leave the rest to you and @fmeum

cameron-martin:
@comius Doesn't the argument of it being achievable via a wrapper macro also apply to environment variables? In this case why was #15232 merged but not this?

comius:
I disagree this is a gap in functionality. Args tokenization is a tech-debt loaded code. This PR seems to provide users to implement workarounds over it. The solution should be a more principle one and accepting the PR could make that harder if not even impossible.

cameron-martin:
I don't think the main purpose of this PR is to fix arguments tokenisation. At least the reasons I want it are not for this, but rather mirror the use cases that @UebelAndre provided.

illicitonion:
@comius can I nudge you at ^^?

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleConfiguredTargetUtil.java
```diff
@@ -643,8 +658,13 @@ private static void addSimpleProviders(
       RunfilesSupport runfilesSupport = null;
       if (!computedDefaultRunfiles.isEmpty()) {
         Preconditions.checkNotNull(executable, "executable must not be null");
-        runfilesSupport =
-            RunfilesSupport.withExecutable(ruleContext, computedDefaultRunfiles, executable);
+        if (starlarkArgsProvided) {
+          runfilesSupport = RunfilesSupport.withExecutableNoArgs(
+              ruleContext, computedDefaultRunfiles, executable);
```
lberki, line 663: What is the argument against creating a `RunfilesSupport.withExecutableAndArgs()` method that overrides the argv computed from the attributes of the rule? That looks simpler than plumbing another set of arguments to `RunCommand` / `TestActionBuilder` and carefully making sure that the one in `RunEnvironmentInfo` overrides the one in `RunfilesSupport`.

lberki:
In the absence of @comius I did my best.

src/main/java/com/google/devtools/build/lib/analysis/starlark/StarlarkRuleConfiguredTargetUtil.java
illicitonion: @fmeum ^^

fmeum:
> I think if this PR were modified to accept Args objects instead-of/in-addition-to Sequences (which should be ~trivial), it could improve this situation significantly: improving expressiveness, and decreasing both memory and runtime overhead.

@illicitonion I wonder whether your use case would be better served by a Starlark `Args` "implementations" that you can feed to your `_test` rule as a drop-in replacement for `ctx.actions.args()`. Something like this (missing `add_all`):

```starlark
def string_args():
    args = []
    def add(o):
        if type(o) == "File":
            args.append(o.path)
        else:
            args.append(str(o))

    def build():
        return list(args)

    return struct(
        add = add,
        build = build,
    )
```

brentleyjones:
Like this? https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/4b19da21da942bb80642cb4f0eef25cc889c0881/test/mock_actions.bzl#L44-L182

fmeum:
I proposed a plan on https://github.com/bazelbuild/bazel/issues/12313 and want to at least dedicate some time to making progress on it before pursuing making arguments configurable from rule implementation functions. 

I do share @comius concern that my PR could end up making `args` parsing less predictable and fragmented across the ecosystem ("how many levels of quotation do I need in this context?"). One way to address this could be to have `RunEnvironmentInfo#arguments` prepend to rather than outright replace `args`. I need to put more thought into this though as a key motivation for me is to ensure proper interaction with wrapping rules.

illicitonion:
> > I think if this PR were modified to accept Args objects instead-of/in-addition-to Sequences (which should be ~trivial), it could improve this situation significantly: improving expressiveness, and decreasing both memory and runtime overhead.
> 
> @illicitonion I wonder whether your use case would be better served by a Starlark `Args` "implementations" that you can feed to your `_test` rule as a drop-in replacement for `ctx.actions.args()`.

If I'm understanding this right, I think this is basically equivalent to giving `Args` a `.to_string_sequence()` function, and requiring it be called before populating the value on `RunEnvironmentInfo` (but possibly doing so in a way which requires up-front opting in to this, rather than just adding a general `to_string_sequence()` to `Args` which may be otherwise problematic)?

If so, then yeah, the two pieces of:
* Adding an `arguments` field on `RunEnvironmentInfo` which accepts a sequence of strings
* Some way of constructing or converting an `Args` so that it can produce a sequence of strings usable in this context

works for my use-case.

comius:
I'll let @lberki handle this.

jfirebaugh:
I want to write a wrapper for [`oci_image`](https://github.com/bazel-contrib/rules_oci/blob/main/docs/image.md) so that you can write the following:

```
go_binary(
    name = "binary",
    srcs = ["main.go"],
    args = [
        "arg1",
        "arg2",
    ],
    data = ["data.txt"],
    env = {"FOO": "bar"},
)

oci_image_wrapper(
  name = "image",
  binary = ":binary"
)
```

The wrapper should:
- Include the binary output and all its runfiles in an image layer (possible)
- Include the binary env values in the image env (possible via `RunEnvironmentInfo`)
- Set the entrypoint array of the image to the appropriate binary path, followed by the binary args (not currently possible)

Would this PR support the last requirement?

katre:
Please remove me from the reviewer list: I can no longer help this PR.

PR #16431, opened 2022-10-08
Remove empty globs

limdor:
There are several globs that are empty and this prevents building with the incompatible flag #8195.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/rules/android/AndroidBinaryTest.java
+++ b/src/test/java/com/google/devtools/build/lib/rules/android/AndroidBinaryTest.java
@@ -183,7 +183,6 @@ public abstract class AndroidBinaryTest extends AndroidBuildViewTestCase {
           "    proguard_specs = [],",
           "    deps = ['lib', 'native-lib'],",
           "    manifest = 'AndroidManifest.xml',",
-          "    resource_files = glob(['res/**']))");
       useConfiguration(
           "--android_platforms=//java/android/platforms:x86,//java/android/platforms:armeabi-v7a",
           "--dynamic_mode=off");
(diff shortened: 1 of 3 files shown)


ahumesky:
Thanks for cleaning this up. I think we can just remove all the attributes that are taking empty globs in the tests, they're probably just copy pasted from other tests.

Could you move the changes to `tools/build_defs/pkg/BUILD` and `tools/jdk/jdk.BUILD` to a separate PR, since that's not test code?

limdor:
> Thanks for cleaning this up. I think we can just remove all the attributes that are taking empty globs in the tests, they're probably just copy pasted from other tests.
> 
> Could you move the changes to `tools/build_defs/pkg/BUILD` and `tools/jdk/jdk.BUILD` to a separate PR, since that's not test code?

Sure, here you have https://github.com/bazelbuild/bazel/pull/16468

limdor:
> Thanks for cleaning this up. I think we can just remove all the attributes that are taking empty globs in the tests, they're probably just copy pasted from other tests.
> 
> Could you move the changes to `tools/build_defs/pkg/BUILD` and `tools/jdk/jdk.BUILD` to a separate PR, since that's not test code?

@ahumesky I removed the empty attributes as requested. Could you take a look? Thanks

ahumesky:
Great, thank you, I'll start importing this now

src/test/java/com/google/devtools/build/lib/rules/android/AndroidBinaryTest.java
```diff
@@ -183,7 +183,6 @@ public void testFatApk_androidPlatformsFlag() throws Exception {
           "    proguard_specs = [],",
           "    deps = ['lib', 'native-lib'],",
           "    manifest = 'AndroidManifest.xml',",
-          "    resource_files = glob(['res/**']))");
```
ahumesky, line 186: this has to be adjusted, but i can do that on the import

Issue #16339, opened 2022-09-26
Crash in `bazel sync` after Ctrl-C


lberki:
### Description of the bug:

Bazel crashed when interrupted with Ctrl-C. The crash dump is as follows:

```
ERROR: An error occurred during the fetch of repository 'android_tools':
   Traceback (most recent call last):
        File "/usr/local/google/home/lberki/.cache/bazel/_bazel_lberki/06874beeaf46f3ea2f26bba6ec6f69f4/external/bazel_tools/tools/build_defs/repo/http.bzl", line 130, column 45, in _http_archive_impl
                download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: thread interrupted
... (this snippet many times)...
 checking cached actions
    Fetching repository @remotejdk17_linux_aarch64; starting 9s
    Fetching repository @fuchsia_sdk; starting 8s
    Fetching ...arch64; Extracting /usr/local/google/home/lberki/.cache/bazel/_bazel_lberki/06874beeaf46f3ea2\
f26bba6ec6f69f4/external/remotejdk17_linux_aarch64/temp4568362422276250630/zulu17.32.13-ca-jdk17.0.2-linux_aa\
rch64.tar.gz
    Fetching ..._arm64; Extracting /usr/local/google/home/lberki/.cache/bazel/_bazel_lberki/06874beeaf46f3ea2\
f26bba6ec6f69f4/external/remotejdk11_win_arm64/temp1937725418544000752/microsoft-jdk-11.0.13.8.1-windows-aarc\
h64.zip
    Fetching ...11_win; Extracting /usr/local/google/home/lberki/.cache/bazel/_bazel_lberki/06874beeaf46f3ea2\
f26bba6ec6f69f4/external/remotejdk11_win/temp7408846882012770473/zulu11.56.19-ca-jdk11.0.15-win_x64.zip
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.NullPointerException
        at com.google.devtools.build.lib.server.FailureDetails$FailureDetail$Builder.setMessage(FailureDetails.java:6620)
        at com.google.devtools.build.lib.util.InterruptedFailureDetails.detailedExitCode(InterruptedFailureDetails.java:33)
        at com.google.devtools.build.lib.bazel.commands.SyncCommand.exec(SyncCommand.java:223)
[...]

PR #16438, opened 2022-10-10
Fix 'bazel sync' crash after interrupt with Ctrl-C

ZabelinK:
Bazel sync command handles InterruptedException, but uses 'getMessage' method which may return 'null' and in this case 'detailedExitCode' raises 'NullPointerException'.

Constructing separate non-empty string together with getMessage's result fixes this problem. This changes similar to what currently used in Fetch and Modquery commands

Fixes: #16339


ZabelinK:
@kshyanashree tests passed. Should I just change lable to "awaiting-review"? Or only maintainers should do this?

Wyverald (approved):
thank you!

PR #16458, opened 2022-10-12
Enable `--experimental_allow_unresolved_symlinks` by default.

### LINKED ISSUE #10298: Stabilize unresolved symbolic links (--experimental_allow_unresolved_symlinks)

meisterT:
Fixes #10298.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -650,7 +650,7 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
 
   @Option(
       name = "experimental_allow_unresolved_symlinks",
-      defaultValue = "false",
+      defaultValue = "true",
       documentationCategory = OptionDocumentationCategory.UNDOCUMENTED,
       effectTags = {
         OptionEffectTag.LOSES_INCREMENTAL_STATE,


src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -650,7 +650,7 @@ public OutputDirectoryNamingSchemeConverter() {
 
   @Option(
       name = "experimental_allow_unresolved_symlinks",
```
fmeum, line 652: How about making this the old name and using just `allow_unresolved_symlinks` as the new name? Having to disable a default option via a flag called `experimental` could be confusing for users.

meisterT: I am not sure the flag will stick around or whether people should actively toggle it; if it doesn't stick around then we should keep the `experimental_` prefix. 

 And I am not the person to make that call - I am just doing the bare minimum to enable this in Bazel 6.0.

meteorcloudy:
Sorry for the breakage! Should be fixed in https://github.com/bazelbuild/bazel/commit/e7b258c3d6f91a6b65136614c4f0360a35727fe2, please rebase and retry.

lberki (approved):

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
lberki: Realistically, I don't think we will flip this flag at Google anytime soon so I guess it'll have to stay?

meisterT: For me the question is more whether we want people (both internally and externally) to change their default value. If not, let's keep the prefix?

lberki: I was scratching my head as to how to decide that. Your line of reasoning is quite nice, so I guess let's keep it?

PR #16463, opened 2022-10-13
Emit Tree objects in topological order

EdSchouten:
remote-apis PR 230 added a way where producers of Tree messages can    indicate that the directories contained within are stored in topological    order. The advantage of using such an ordering is that it permits    instantiation of such objects onto a local file system in a streaming    fashion. The same holds for lookups of individual paths.
    
Even though Bazel currently does not gain from this, this change at    least modifies Bazel's REv2 client to emit topologically sorted trees.    This makes it possible for tools such as Buildbarn's bb-browser to    process them more efficiently.
    
More details:
- https://github.com/bazelbuild/remote-apis/pull/229
- https://github.com/bazelbuild/remote-apis/pull/230

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/UploadManifest.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/UploadManifest.java
@@ -31,6 +31,7 @@ import build.bazel.remote.execution.v2.Tree;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import com.google.devtools.build.lib.actions.ActionExecutionMetadata;
 import com.google.devtools.build.lib.actions.ActionUploadFinishedEvent;
 import com.google.devtools.build.lib.actions.ActionUploadStartedEvent;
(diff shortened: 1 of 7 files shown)


coeuvre (approved):
Thanks!

sgowroji:
Hello @coeuvre, Above PR has ThirdParty changes, Could you please guide which code should be imported first. @EdSchouten Can you please squash above commits into one.  Thanks!

coeuvre:
We need to import third-party changes first because other changes in non third-party directories depend on them.

EdSchouten:
Hey there! So given @coeuvre's response, you don't want me to squash the commits?

sgowroji:
We still need all commits, squash into one.

EdSchouten:
Hey @sgowroji,
It seems the changes to the .proto files have already been merged now. Do I need to do anything on my end (e.g., rebase) for the second half of this change?

sgowroji:
Remaining changes are already in progress through a internal CL. We will update you once it is merged. Thank you for the patience.

brentleyjones:
@bazel-io flag

meteorcloudy:
@bazel-io fork 6.0.0

Issue #16351, opened 2022-09-29
test.xml missing when running flaky test


exoson:
### Description of the bug:

When running a flaky test with the `--flaky_test_attempts=10` and `--remote_download_minimal` flags with remote execution, bazel build execution will stop if the test fails with the following error message: 
ERROR: /data/repos/test_repo/BUILD:9:8: Testing //:python_test failed: unexpected I/O exception: /home/ekattainen/.cache/bazel/_bazel_ekattainen/f465db4dd2df63e33090e2c00fda9ac6/execroot/__main__/bazel-out/k8-fastbuild/testlogs/python_test/test.xml -> /home/ekattainen/.cache/bazel/_bazel_ekattainen/f465db4dd2df63e33090e2c00fda9ac6/execroot/__main__/bazel-out/k8-fastbuild/testlogs/python_test/test_attempts/attempt_1.xml (No such file or directory)

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Have a workspace with empty WORKSPACE file and BUILD and flaky_test.py files with following contents 
```
$ cat BUILD
py_test(
    name = "flaky_test",
    srcs = "flaky_test.py",
)
$ cat flaky_test.py
import random
import sys

if random.random() > 0.5:
    sys.exit(1)
```
Run the following command until the test fails
```
$ bazel test //:python_test --nocache_test_results --config=remote_exec --flaky_test_attempts=4 --remote_download_minimal
```
The `--config=remote_exec` should configure bazel to execute the actions remotely.

### Which operating system are you running Bazel on?

Ubuntu 20.04

### What is the output of `bazel info release`?

development version

### If `bazel info release` returns `development version` or `(@non-git)`, tell us how you built Bazel.

bazel build //src:bazel-bin

### What's the output of `git remote get-url origin; git rev-parse master; git rev-parse HEAD` ?

```text
de02e3f62598dbe8dcff68b4040f34da3322c720
de02e3f62598dbe8dcff68b4040f34da3322c720
```

### Any other information, logs, or outputs that you want to share?

The culprit for this bug seems to be https://github.com/bazelbuild/bazel/pull/16123 Tagging @coeuvre as the PR author if he were to come up with a quick fix.

coeuvre:
Thanks for the report. I am able to reproduce this locally.

[...]

PR #16466, opened 2022-10-13
Inject all metadata at once after the action is executed.

coeuvre:
So that spawns within one action will be able to change the injected metadata using e.g. renameTo.

Also only allows metadata injection for declared action outputs instead of declared spawn outputs.

Related #16351.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
@@ -107,7 +107,7 @@ public class RemoteActionFileSystem extends DelegateFileSystem {
   }
 
   void flush() {
-    checkNotNull(metadataInjector, "metadataInject is null");
+    checkNotNull(metadataInjector, "metadataInjector is null");
 
     for (Map.Entry<PathFragment, Artifact> entry : outputMapping.entrySet()) {
       PathFragment execPath = entry.getKey();
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
@@ -16,6 +16,8 @@
 package com.google.devtools.build.lib.remote;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.Streams.stream;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -84,27 +90,39 @@ public void updateContext(MetadataInjector metadataInjector) {
...
   void injectFile(ActionInput file, RemoteFileArtifactValue metadata) {
-    checkNotNull(metadataInjector, "metadataInject is null");
-
     injectedMetadata.put(file.getExecPath(), metadata);
+  }
 
-    if (file instanceof Artifact) {
-      metadataInjector.injectFile((Artifact) file, metadata);
+  void flush() {
+    checkNotNull(metadataInjector, "metadataInject is null");
```
tjgq, line 110:
```suggestion
    checkNotNull(metadataInjector, "metadataInjector is null");
```

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -84,27 +90,39 @@ public void updateContext(MetadataInjector metadataInjector) {
...
+
+    for (Map.Entry<PathFragment, Artifact> entry : outputMapping.entrySet()) {
```
tjgq, line 112: This seems to be a functional change, as previously `injectFile` and `injectTree` would perform the injection even when the artifact was not an output. Is this intentional? (I'm not aware of a reason to inject non-outputs, but that doesn't mean there isn't one :))

coeuvre: It is intentional. The difference is, before this change, we could only inject declared spawn outputs (not arbitrary output) which is a super set of declared action outputs. In most cases they should be same (i.e one action generates one spawn). In the cases that declared spawn outputs have more outputs than declared action outputs, most of them are instance of `ActionInput` which don't get injected.

From the perspective of reading the injected metadata, all that matters is metadata for declared outputs. I can't find a reason for allowing inject metadata of non-declared outputs (buy I may be wrong).

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Done.

tjgq (approved):

PR #16474, opened 2022-10-14
Match execroot relative path with --experimental_remote_download_regex

fmeum:
Previously, the regex was applied to the full on-disk path of an output, which can lead to non-hermetic matching behavior.


coeuvre (approved):

brentleyjones:
@fmeum We can probably cherry-pick this as well into 5.3.2?

coeuvre:
`file.path` is always under exec root (acquired by `remotePathResolver.outputPathToLocalPath`) so I think it is safe.

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 6.0.0

Issue #4685, opened 2018-02-22
`bazel coverage` fails over remote ex for java_test


werkt:
### Description of the problem / feature request:

Tooling for coverage fails when running under remote execution (LcovMerger specifically)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Run a test target with `bazel coverage --spawn_strategy=remote --remote_executor=... //path/to:test`

### What operating system are you running Bazel on?

Ubuntu 14.04

### What's the output of `bazel info release`?

release 0.10.1

### Any other information, logs, or outputs that you want to share?

Relevant portion of coverage execution

```
+ exec bazel-out/k8-fastbuild/bin/external/bazel_tools/tools/test/LcovMerger --coverage_dir=%execroot%/_coverage/path/to/test/test --output_file=%execroot%/bazel-out/k8-fastbuild/testlogs/path/to/test/coverage.dat
%execroot%/bazel-out/k8-fastbuild/bin/external/bazel_tools/tools/test/LcovMerger: Cannot locate runfiles directory. (Set $JAVA_RUNFILES to inhibit searching.)
```

buchgr:
Is this bug still an issue?cc @lberki  @iirina

vmrob:
I see the same issue when running some of our C++/Python tests with remote execution. Bazel version is a slightly patched version of 0.25.3.

vmrob:
I've come up with a workaround that basically involves the below patch. I have a feeling there is a more correct way to do this--possibly by including some implicit dependencies in the relevant coverage-aware Bazel rules, but I am not familiar enough with the codebase to efficiently tackle that.

```diff
commit ba16394fa207709b2b2a2fdc22de30cc72cf4398 (HEAD -> rbe-coverage)
Author: Victor Robertson <victor.robertson@getcruise.com>
Date:   Wed Sep 11 23:01:49 2019 -0700

    Fix coverage for RBE [EP-12079]

    This patch fixes code coverage in RBE by removing the distinction
    between JAVA_RUNFILES and TEST_SRCDIR in the collect_coverage.sh script.
    This is important as Bazel will not send JAVA_RUNFILES to the remote
    environment. Unfortunately, this means that a Java sdk will be included
    with every coverage test--perhaps this isn't so bad.

    This also means that tests with coverage support must include the
    dependencies normally provided by Bazel. In cruise/cruise, we use the
    following additional dependencies:

    - @embedded_jdk//:jdk
[...]

Issue #6862, opened 2018-12-07
Tracking issue for "Remote Builds without the Bytes"


buchgr:
This issue tracks the progress towards [Remote Builds without the Bytes](https://docs.google.com/document/d/1wXPoGMhd1wmjc_4fqy-bP0RE3AlHk-IHpMUy0nDZLkU/edit?usp=sharing&resourcekey=0-1bw8YMFeUPCeK4lOLWTOBA).

## Milestones
- [x] **Implement basic support for not downloading intermediate artifacts** (In Bazel 0.25)
The feature can be enabled by adding the flag `--experimental_remote_download_outputs=minimal` to the remote caching / execution invocation. The feature mostly works, but comes with the limitations stated in the subsequent tasks.
- [x] **Fix local actions that need to fetch remote outputs** (In Bazel 0.25)
While Bazel can run most actions remotely there are a few dozen or so action implementations that currently can't. This can be problematic because such actions often require Bazel to download remote files which this change tries to avoid. An example of such an action is the `SymlinkAction`.
- [x] **Implement fetching of top level artifacts** (In Bazel 0.26)
Implement `--experimental_remote_download_outputs=toplevel` as described in the design document.
- [x] **Support downloading unzipped undeclared test outputs**
See https://github.com/bazelbuild/bazel/issues/17884 for details
- [x] **Store per output file metadata in the persistent action cache**
The persistent action cache stores the state of Bazel's output base. It does not know about remotely stored files. We need to extend it to do so. See https://github.com/bazelbuild/bazel/issues/8248
- [x] **Builds without the Bytes fails on missing AC result**
See https://github.com/bazelbuild/bazel/issues/10880 for details
- [x] **Support remote cache eviction**
As detailed in the design document it's possible for Bazel to wrongly think that an output file exists on the remote system because it has been evicted ("garbage collected") in the meantime. If a Bazel action then depends on an evicted file there is currently no other way but to fail the build/test. See https://github.com/bazelbuild/bazel/issues/8250
- [x] **Add up- and download stats to the profiler**
A user should have insight into what is being up/downloaded and why.
- [x] **Be compatible with the Build Event Service local file uploader** (In Bazel 0.28.0)
This change currently doesn't interact well with the BEP and BES. See https://github.com/bazelbuild/bazel/issues/8249
- [x] **Don't create a runfiles tree for remote actions**
[...]

PR #16475, opened 2022-10-14
Add support for bazel coverage with BwoB

fmeum:
With `--experimental_split_coverage_postprocessing` and `experimental_fetch_all_coverage_outputs`, all that is needed to get coverage runs to succeed with `--remote_download_minimal` is to explicitly request the raw and processed coverage files. This is made possible by the machinery introduced for
`--experimental_remote_download_regex`.

Work towards #4685


fmeum:
@coeuvre @c-mita Could you take a look? It's a kind of cross-team PR and I would really like someone to check my regexes actually cover all coverage files.

coeuvre:
I am not sure whether using regex to pass the information around is the best option. As I mentioned in https://github.com/bazelbuild/bazel/issues/6862#issuecomment-1278723104, BwoB will be able to download output groups, can we somehow define an output group for those files?

fmeum:
> I am not sure whether using regex to pass the information around is the best option. As I mentioned in [#6862 (comment)](https://github.com/bazelbuild/bazel/issues/6862#issuecomment-1278723104), BwoB will be able to download output groups, can we somehow define an output group for those files?

I agree that regexes are less than ideal. They have the benefit of working with `minimal` though - would we be able to force downloading a particular output group even with that setting or would it require `toplevel`?

I will look into the output group approach, I can't say yet whether it's feasible.

coeuvre:
the problem with `minimal` is the requested files (either through regex or output groups) are only downloaded if the generating actions didn't hit the skyframe cache.

In an incremental build, if you only change the requested regex, no additional intermediate files will be downloaded. However, `toplevel` always works. We can force `minimal` to download requested output groups but it has the same problem with regex.

fmeum:
It seems to be possible for the coverage reports created for individual tests to be included in an output group. We may need to explicitly fetch `bazel-out/_coverage/_coverage_report.dat` as that seems to be requested right from within Skyframe/BuildView and isn't associated with a target, but it would be a static path rather than a regex.

@coeuvre Do you have a branch with a prototype of the output group fetching functionality that I could use to test this approach with?

coeuvre:
not yet, but I plan to create a PR for that in a few days.

fmeum:
> not yet, but I plan to create a PR for that in a few days.

Great, thanks. Just ping me when you have something. I will set this PR to draft for now.

coeuvre:
The PR https://github.com/bazelbuild/bazel/pull/16524 is merged. It now listens to `TestAttempt` event and downloads all the available test outputs.

You can play with it but it may already be able to download coverage outputs in which case can you add an integration test for it?

fmeum:
@coeuvre Just tested it, but the coverage outputs aren't downloaded. How would I go about registering them as test outputs? I have a commit that adds them to a new output group, but how would I register that with the downloader?

coeuvre:
The TestAttempt event is fired [here](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java;l=279;drc=49eb31b1880f56e2762b6cd480ae2cf732469561). You can follow the `testOutputs` to find where to register them.

If you think coverage outputs should be always downloaded by default if they exist, maybe just add them to the test output instead of using output group.

fmeum:
It looks like coverage artifacts and reports are already declared outputs of the test action. However, the spawn that is supposed to take the coverage artifacts and generates a report from them fails to do so as it doesn't see the coverage artifacts.

@coeuvre If a remotely run `SimpleSpawn` declares a tree artifact as an input that is a declared output of a remotely run test, does it need to do anything special to "see" the files in the tree artifact? The files are there in `RemoteExecutionService#injectRemoteArtifacts`, but aren't available on the remote executor when the `SimpleSpawn` is executed.

coeuvre:
> It looks like coverage artifacts and reports are already declared outputs of the test action.

Yes, I remember it is the case so I said it may already works, but apparently, there are some other issues.

> does it need to do anything special to "see" the files in the tree artifact?

I think it **should** just work otherwise there are some bugs in `RemoteActionFileSystem`.

Can you please share a minimal repro so I can look into?

/cc @tjgq

fmeum:
The integration tests added by this PR are reproducers (even if you replace `--remote_download_minimal` with `--remote_download_toplevel`).

I debugged this for a bit and think that the problem is the way the coverage tree artifact is created without properly registering it with Skyframe. That means that `SpawnInputExpander` is not able to expand it and then remote execution stages it as an empty directory. Doesn't seem to be an issue on the RBE side.

fmeum:
Adding slightly more detail: You assessed the situation in https://github.com/bazelbuild/bazel/pull/15276#discussion_r861707965. Either the spawn manually expands the tree artifact and doesn't pick up files that only exist remotely, or it sends a tree artifact that `SpawnInputExpander` doesn't know about und thus can't expand.

Not sure what the right direction is. Maybe looking into how tree artifacts created during the execution of the action could be injected into `SpawnInputExpander`?

coeuvre:
I have a working patch that just read tree artifacts from action fs. I will create a PR soon.

fmeum:
Superseded by https://github.com/bazelbuild/bazel/pull/16556

PR #16476, opened 2022-10-14
Revert to using `matches` instead of `find` with `remote_download_regex`

fmeum:
Reverts a change made in https://github.com/bazelbuild/bazel/commit/e8278edbec1c6be14f01b2ecb078042ee9e753e9 alongside enabling `allowMultiple` for `--experimental_remote_download_regex`.

It is much easier to accidentally write regexes with pathological performance with `find` than with `matches`. If needed, the `find` functionality can always be obtained with `matches` by prepending and appending `.*` as needed.

In addition, common usage scenarios such as matching by file extension are easier to get right: With `matches`, `jar` will visibly fail to have an effect and is easily corrected to `.*jar` (or even `.*\.jar`), whereas with `find` it will silently fetch entire directories that contain the substring `jar`, potentially causing performance regressions.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
@@ -234,7 +234,7 @@ public class RemoteExecutionService {
     this.shouldForceDownloadPredicate =
         path -> {
           for (Pattern pattern : patterns) {
-            if (pattern.matcher(path).find()) {
+            if (pattern.matcher(path).matches()) {
               return true;
             }
           }


fmeum:
@coeuvre I just wanted to flag this. Feel free to close if you intentionally made the switch in https://github.com/bazelbuild/bazel/commit/e8278edbec1c6be14f01b2ecb078042ee9e753e9.

coeuvre (approved):

fmeum:
@kshyanashree In case https://github.com/bazelbuild/bazel/commit/e8278edbec1c6be14f01b2ecb078042ee9e753e9 is cherry-picked into 5.3.2, this should probably be cherry-picked as well.

brentleyjones:
@fmeum I'll include it in the PR.

PR #16477, opened 2022-10-14
Use an in-memory filesystem to track injected remote files

coeuvre:
instead of using Map. This allows implementing more fs operations later.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
@@ -20,7 +20,6 @@ import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Streams.stream;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.devtools.build.lib.actions.ActionInput;
 import com.google.devtools.build.lib.actions.ActionInputMap;
 import com.google.devtools.build.lib.actions.Artifact;
 import com.google.devtools.build.lib.actions.Artifact.SpecialArtifact;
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -438,4 +466,84 @@ protected void createHardLink(PathFragment linkPath, PathFragment originalPath)
...
+    RemoteFileInfo getRemoteFileInfo(PathFragment path, boolean followSymlinks) {
+      InMemoryContentInfo node = inodeStatErrno(path, followSymlinks).inode();
+      if (!(node instanceof RemoteFileInfo)) {
+        return null;
+      }
+      return (RemoteFileInfo) node;
+    }
+  }
+
+  static class RemoteFileInfo extends FileInfo {
```
tjgq, line 504: Will it be possible to reuse `InMemoryFileInfo` once we get rid of `actionId` (as we discussed elsewhere)?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -438,4 +466,84 @@ protected void createHardLink(PathFragment linkPath, PathFragment originalPath)
...
+    @Override
+    protected FileInfo newFile(Clock clock, PathFragment path) {
+      return new RemoteFileInfo(clock);
+    }
+
+    void injectRemoteFile(PathFragment path, byte[] digest, long size, String actionId)
+        throws IOException {
+      createDirectoryAndParents(path.getParentDirectory());
+      InMemoryContentInfo node = getOrCreateWritableInode(path);
+      if (!(node instanceof RemoteFileInfo)) {
```
tjgq, line 486: Could you add a comment as to when this would be the case? My understanding of `InMemoryFileSystem` is that this would happen if a directory or symlink already exists at this location - is that correct?

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
```diff
@@ -751,55 +747,33 @@ private void createSymlinks(Iterable<SymlinkMetadata> symlinks) throws IOExcepti
     }
   }
 
-  private void injectRemoteArtifact(
-      RemoteAction action, ActionInput output, ActionResultMetadata metadata) throws IOException {
+  private void injectRemoteArtifact(RemoteAction action, ActionResultMetadata metadata)
```
tjgq, line 750: Nit: injectRemoteArtifacts (plural)

tjgq (approved):

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: No:
1. we have to override `getFastDigest()` to return the digest of remote file.
2. we want to block operations that don't make sense to remote files e.g. `getOutputStream`.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java
coeuvre: Done.

PR #16480, opened 2022-10-15
Clarify repo visibility error message

fmeum:
Instead of saying `Repository '@foo' is not visible from repository '@bar'`, the message now says `No repository visible as '@foo' from '@bar'`.

The previous message made it sound as if there was a repo known as `@foo` that just happens not to be visible from `@bar` (similar to Java strict deps or the visibility attribute on rules), when in fact `@foo` is just a string made up by `@bar` that has no meaning until it is mentioned in some `use_repo` call.

Also replaces `repository '@@'` with the less cryptic term `main repository`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/cmdline/RepositoryName.java
+++ b/src/main/java/com/google/devtools/build/lib/cmdline/RepositoryName.java
@@ -194,9 +194,14 @@ public final class RepositoryName {
     return ownerRepoIfNotVisible == null;
   }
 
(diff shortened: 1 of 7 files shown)


fmeum:
@Wyverald @meteorcloudy Could you review this? I'm not married to the particular implementation or wording, the existing error message just happened to confuse me.

meteorcloudy (approved):
Indeed sounds much better, thanks!

PR #16482, opened 2022-10-16
Fix query command suggestion in error message with repo mappings

fmeum:
The package label in the query command suggested when a specified target isn't found in a package looked like `@rules_cc~1.0.0//pkg`, which doesn't work as the repo name is canonical, but the label isn't.

This is fixed by reusing the logic from `Label#getUnambiguousCanonicalForm`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
+++ b/src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
@@ -21,6 +21,7 @@ import com.google.devtools.build.lib.concurrent.BlazeInterners;
 import com.google.devtools.build.lib.skyframe.serialization.autocodec.AutoCodec;
 import com.google.devtools.build.lib.vfs.PathFragment;
 import java.util.Objects;
+import java.util.Optional;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -175,6 +176,50 @@ public final class PackageIdentifier implements Comparable<PackageIdentifier> {
   // TODO(bazel-team): Maybe rename to "getDefaultForm"?
   public String getCanonicalForm() {
     return repository.getCanonicalForm() + "//" + getPackageFragment();
+  }
+
+  /**
+   * Returns an absolutely unambiguous canonical form for this package in label form. Parsing this
+   * string in any environment, even when subject to repository mapping, should identify the same
+   * package.
+   */
+  public String getUnambiguousCanonicalForm() {
+    return String.format("@@%s//%s", getRepository().getName(), getPackageFragment());
+  }
+
+  /**
+   * Returns a label representation for this package that is suitable for display. The returned
+   * string is as simple as possible while referencing the current package when parsed in the
+   * context of the main repository.
+   *
+   * @param mainRepositoryMapping the {@link RepositoryMapping} of the main repository
+   * @return <dl><dt><code>//some/pkg</code></dt>
+   * <dd>if this package lives in the main repository</dd>
+   * <dt><code>@protobuf//some/pkg</code></dt>
+   * <dd>if this package lives in a repository with "protobuf" as <code>name</code> of a repository
+   * in WORKSPACE or as local name of a Bzlmod dependency of the main module</dd>
+   * <dt><code>@@protobuf~3.19.2//some/pkg</code></dt>
+   * <dd>only with Bzlmod if the current package belongs to a repository that is not visible from
+   * the main module</dd>
+   */
+  public String getDisplayForm(RepositoryMapping mainRepositoryMapping) {
+    Preconditions.checkArgument(
+        mainRepositoryMapping.ownerRepo() == null || mainRepositoryMapping.ownerRepo().isMain());
+    if (repository.isMain()) {
+      // Packages in the main repository can always use repo-relative form.
+      return "//" + getPackageFragment();
+    }
+    if (mainRepositoryMapping.ownerRepo() == null) {
+      // If the main repository mapping is not using strict visibility, then Bzlmod is certainly
+      // disabled, which means that canonical and local names can be used interchangeably from the
+      // context of the main repository.
+      return repository.getNameWithAt() + "//" + getPackageFragment();
+    }
+    // If possible, represent the repository with a non-canonical label using the local name the
+    // main repository has for it, otherwise fall back to a canonical label.
+    return mainRepositoryMapping.getInverse(repository)
+        .map(localName -> "@" + localName + "//" + getPackageFragment())
+        .orElseGet(this::getUnambiguousCanonicalForm);
   }
 
   /**
--- a/src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
+++ b/src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
@@ -19,6 +19,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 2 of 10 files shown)


fmeum:
@meteorcloudy @Wyverald Could you review this?

I noticed that the output of query consists of non-canonical label literals with canonical repo names. Those can't really be used anywhere with Bzlmod enabled. Should query emit canonical label literals instead, at least with Bzlmod?

Wyverald (approved):

Wyverald:
> I noticed that the output of query consists of non-canonical label literals with canonical repo names. Those can't really be used anywhere with Bzlmod enabled. Should query emit canonical label literals instead, at least with Bzlmod?

I think to begin with, outputting `@@`-style labels when Bzlmod is enabled should suffice; but what you described later is even better, if it's not too much trouble to implement.

Wyverald (changes requested):
Actually, Yun just pointed out to me that this would break Google, where we don't use external repos at all and `@`-prefixed labels are forbidden.

Could you instead only turn on this behavior when Bzlmod is enabled?

fmeum:
Instead of depending on the Bzlmod bit explicitly, I backported the "one label stringification function to rule them all" I came up with for https://github.com/bazelbuild/bazel/pull/16483, which decanonicalizes query output as far as possible, to the current PR. I moved the tests over since the current PR doesn't exercise all its functionality yet.

src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
```diff
@@ -35,7 +40,10 @@ public abstract class RepositoryMapping {
   // Always fallback to the requested name
   public static final RepositoryMapping ALWAYS_FALLBACK = createAllowingFallback(ImmutableMap.of());
 
-  abstract ImmutableMap<String, RepositoryName> repositoryMapping();
+  // A given String maps to exactly one RepositoryName, but a RepositoryName may have multiple
+  // Strings (that is, local repository names), mapping to it. In other words, the entries would
+  // form a valid ImmutableMap, but not necessarily a valid ImmutableBiMap.
+  abstract ImmutableListMultimap<String, RepositoryName> repositoryMapping();
```
Wyverald, line 46: I don't think we should make this an `ImmutableListMultimap` just so that we can call `inverse` on it. Maybe have an `ImmutableListMultimap<...> inverse()` method on `RepositoryMapping` instead, or maybe even just `String inverseLookup(RepositoryMapping)`

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
```diff
@@ -177,6 +178,65 @@ public String getCanonicalForm() {
...
+   * <dd>only with Bzlmod if the current package belongs to a repository that is not visible from
+   * the main module</dd>
+   *
+   * <li>if another repository's mapping is provided:</li>
+   * <dl>
+   * <dt><pre>//some/pkg</pre></dt>
+   * <dd>if this package lives in the main repository</dd>
+   * <dt><pre>@protobuf//some/pkg</pre></dt>
+   * <dd>if this package lives in a repository with "protobuf" as <code>name</code> of a repository
+   * in WORKSPACE or as local name of a Bzlmod dependency of the main module</dd>
```
Wyverald, line 215: "or as local name of a Bzlmod dependency of the main module" does not apply here, does it?

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
```diff
@@ -177,6 +178,65 @@ public String getCanonicalForm() {
...
+    return String.format("@@%s//%s", getRepository().getName(), getPackageFragment());
+  }
+
+  /**
+   * Returns a label representation for this package that is suitable for display. The returned
+   * string is as simple as possible while referencing the current package when parsed in the
+   * context of the main repository.
+   *
+   * @param mainOrOtherRepoMapping the {@link RepositoryMapping} of the main repository or, if that
+   *                               is not available to the caller, the {@link RepositoryMapping} of
```
Wyverald, line 196: it looks like if the repo mapping provided is not of the main repo, then it doesn't matter at all what the repo mapping is, right? it could be for any non-main repo and the result would be the same. We might as well say "grab any repo mapping object and pass it here".

this makes me a bit uneasy about this API, though I don't have a better idea right now.

src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
fmeum: If we don't make it an `Immutable*Multimap`, we would have to iterate over all mapping entries to find the preimage. If you think that that's not going to be costly in terms of CPU time, then I can certainly do it to save memory (and complexity).

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
fmeum: Fixed.

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
fmeum: I wasn't completely sure about that: Are "Bzlmod is enabled" and "for each `RepositoryMapping` instance `ownerRepo()` is non-`null`" equivalent? I was worried of e.g. the `RepositoryMapping`s of repos defined in `WORKSPACE` fallback mode to have `ownerRepo() == null` even though they should really result in canonical labels (or not?). It would be great if you could clarify that point.

src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
fmeum: Pushed that change in a separate commit.

Wyverald: I don't believe the inverse is calculated enough to worry about CPU time (since it's only ever done to produce query output). In any case, it certainly isn't enough to justify the API change, IMO.

src/main/java/com/google/devtools/build/lib/cmdline/PackageIdentifier.java
Wyverald: As you suspected, they're not equivalent, exactly because WORKSPACE repos still aren't subject to strict deps. That doesn't mean they don't go through repo mapping, just that if a local name is missing from the mapping, it'll be used as the canonical name instead. This actually means that line 224 is broken in certain cases.

fmeum: Since the API of `getDisplayForm` was confusing and I would have had to pass in the Bzlmod bit anyway, I am now passing the main repository's mapping into `Package`. This results in the ideal error message and makes it so that `getDisplayForm` can expect to be passed the main repository's rather than any mapping.

Wyverald: Nice!

src/main/java/com/google/devtools/build/lib/packages/Package.java
```diff
@@ -894,6 +903,7 @@ public static Builder newExternalPackageBuilderForBzlmod(
             basePackageId,
             DUMMY_WORKSPACE_NAME_FOR_BZLMOD_PACKAGES,
             starlarkSemantics.getBool(BuildLanguageOptions.INCOMPATIBLE_NO_IMPLICIT_FILE_EXPORT),
+            repoMapping,
```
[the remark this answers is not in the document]
fmeum: I could also pass in `RepositoryMapping.ALWAYS_FALLBACK` if that is less confusing and the value I pass truly doesn't matter.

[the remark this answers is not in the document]
fmeum: I added the comment but can also go with `ALWAYS_FALLBACK` if you find that cleaner.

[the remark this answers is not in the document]
Wyverald: ah, yeah, could you do that please? thanks!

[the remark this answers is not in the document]
fmeum: Done and squashed.

Wyverald (approved):

PR #16483, opened 2022-10-16
Decanonicalize labels emitted by {a,c,}query if possible

fmeum:
Uses the newly added `PackageIdentifier#getDisplayForm` to turn labels
in the output of query, aquery, and cquery into the most concise
representation that allows them to be resolved from the context of the
main repository.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/cmdline/Label.java
+++ b/src/main/java/com/google/devtools/build/lib/cmdline/Label.java
@@ -419,9 +419,19 @@ public final class Label implements Comparable<Label>, StarlarkValue, SkyKey, Co
    * Label.parse*(x.getUnambiguousCanonicalForm(), ...).equals(x)}).
    */
   public String getUnambiguousCanonicalForm() {
-    return String.format(
-        "@@%s//%s:%s",
-        packageIdentifier.getRepository().getName(), packageIdentifier.getPackageFragment(), name);
+    return packageIdentifier.getUnambiguousCanonicalForm() + ":" + name;
+  }
+
+  /**
+   * Returns a label string that is suitable for display, i.e., it resolves to this label when
+   * parsed in the context of the main repository and has a repository part that is as simple as
+   * possible.
+   *
+   * @param mainRepositoryMapping the {@link RepositoryMapping} of the main repository
+   * @return analogous to {@link PackageIdentifier#getDisplayForm(RepositoryMapping)}
+   */
+  public String getDisplayForm(RepositoryMapping mainRepositoryMapping) {
+    return packageIdentifier.getDisplayForm(mainRepositoryMapping) + ":" + name;
   }
 
   /** Return the name of the repository label refers to without the leading `at` symbol. */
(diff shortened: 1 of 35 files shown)


src/main/java/com/google/devtools/build/lib/cmdline/RepositoryMapping.java
```diff
@@ -93,4 +95,17 @@ public RepositoryName get(String preMappingName) {
...
+    if (ownerRepo() == null) {
+      // This can only happen if Bzlmod is not enabled. In this case, assume no repo mappings are
+      // in use, so that canonical and  local names can be used interchangeably.
+      return Optional.of(postMappingName.getName());
+    }
+    Preconditions.checkArgument(ownerRepo().isMain());
+    return repositoryMapping().entrySet().stream()
+        .filter(e -> e.getValue().equals(postMappingName))
+        .map(Entry::getKey)
+        .findFirst();
```
fmeum, line 109: This is `O(num mappings)`, so a proper implementation would instead store the mapping in an `ImmutableSetMultimap`.

fmeum:
Stacked on #16482 and thus in draft state, but otherwise ready for review.

fmeum:
I pushed a commit that updates more formatters. 

This PR doesn't cover the `BUILD`-like output modes, which would have to decanonicalize labels with respect to the current rather than the main repository. That would require more extensive changes though.

fmeum:
@Wyverald I rebased onto master, this is ready for review now.

src/main/java/com/google/devtools/build/lib/cmdline/Label.java
```diff
@@ -416,9 +416,21 @@ public String getCanonicalForm() {
...
+   * Returns a label string that is suitable for display, i.e., it resolves to this label when
+   * parsed in the context of the main repository and has a repository part that is as simple as
+   * possible.
+   *
+   * @param mainOrOtherRepoMapping the {@link RepositoryMapping} of the main repository or, if that
+   *                               is not available to the caller, the {@link RepositoryMapping} of
+   *                               the current package. The former will yield more readable output.
+   * @return analogous to {@link PackageIdentifier#getDisplayForm(RepositoryMapping)}
+   */
+  public String getDisplayForm(RepositoryMapping mainOrOtherRepoMapping) {
```
Wyverald, line 432: this should also be just `mainRepoMapping`, right?

src/main/java/com/google/devtools/build/lib/query2/engine/QueryEnvironment.java
```diff
@@ -536,6 +537,10 @@ ThreadSafeMutableSet<T> getBuildFiles(
    */
   TargetAccessor<T> getAccessor();
 
+  default RepositoryMapping getMainRepoMapping() {
```
Wyverald, line 540: does this need to have a default?

src/main/java/com/google/devtools/build/lib/cmdline/Label.java
fmeum: Fixed, thanks!

Wyverald (approved):

fmeum:
@bazel-io flag

kshyanashree:
Hello @fmeum! As I could see we are good to merge the PR, does it require any other review? Thanks!

fmeum:
@kshyanashree Yes, this is good to merge!

meteorcloudy:
@bazel-io fork 6.0.0

PR #16485, opened 2022-10-17
Proof of concept of license and package information splicing

aiuto:
This is a PoC for the kind of patching discussed in https://docs.google.com/document/d/1PkSk1fHcLNYJX9fk8cEMKueMvrpt22QcVIwcTICCEIk/edit#

The problem we are solving is that dependencies might not come with all the metadata fields needed for producing SBOMs. In some cases there are no `license` targets in the BUILD files for repositories that are Bazel ready. In other cases, the module author has made incorrect assertions about their licenses. They may say MIT, but then include code that is GPL.

To address these problems, we interject a user defined tool into each repository rule, after download and initial patching has been done, but before we return to the main WORKSPACE/bzlmod evaluation.
This tool is passed all the information we might have, for example,
- download url
- sha
- Maven coordinate
- repo name,
- ...
and can examine the entire downloaded tree to scan files for license and provenance information.
Using what it finds and what was passed, it may create/rewrite BUILD files to create needed package metadata rules (`license(`), `package_info()`)

The technical difficulty is that we must do this kind of rewrite only if the user of the workspace wants it. So essentially, we want a repository rule like `http_archive()` to behave differently depending on the existence for the package metadata rewriter.  Since everyone pulls in http_archive via @bazel_tools, we can not simply provide our own in a local repository.  (Well, maybe we can, but that might require having a local repository for all of @bazel_tools.)

The implementation here is
- create a special top level symbol to point to the rewriter. In this case `_bazel_module_patcher`.
- wrap http_archive with a macro that conditionally passes the patcher to the underlying rule if it is present
  - done that way because we can not call native.existing_rule from the http_archive implementation.
  - alternative might be have a command line flag that places an executable path into the repository context
- have http_archive invoke the patcher.

Highlights of this PR
- Only for non-windows right now
- Only for http_archive at this time. No point in doing everything until we like the idea.
- Has to wrap http_archve with a macro to be able to call native.existing_rule("_bazel_module_patcher")
- Skips repositories with no clear top level LICENSE. We can iterate on that.


aiuto:
Hi Xudong.  I may need some help here.

This works for WORKSPACE based builds, but  crashes  in integration tests when bzlmod is on.
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/0183e3ac-444e-408b-a788-0c13b4ea1067/src/test/shell/bazel/execroot_test/test.log

I have a feeling it might be related to the context where bzlmod invokes http_archive being different from the old way, but I'm not sure.
The weird thing is that it should not even come into play for bzlmod, because I don't define the magic symbol in MODULE.bazel or WORKSPACE.bzlmod. All that would be different there is that the bazel executing the intiial build has a different version of http_archive from the one we are building.

PR #16487, opened 2022-10-17
Use the execution time instead of the worker time for the wall time

EdSchouten:
SpawnResult.getWallTime() is documented as follows:

    Returns the wall time taken by the {@link Spawn}'s execution.

With the math performed as is, we also end up including time spent downloading inputs and uploading outputs. I don't think that this is in the spirit of this method. There are also getUserTime() and getSystemTime(), meaning this was likely modeled after the time(1) utility.

Cc @tjgq, as this logic was added in 93029d8f1f01be440d8e06cce9585ad912f9169b.


tjgq (approved):
Yeah, I agree that this is more correct. Thanks!

PR #16489, opened 2022-10-17
Implement renameTo and createDirectory[AndParents] for RemoteActionFileSystem

### LINKED ISSUE #16351: test.xml missing when running flaky test

coeuvre:
Fixes #16351.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
@@ -19,6 +19,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Streams.stream;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.devtools.build.lib.actions.ActionInputMap;
 import com.google.devtools.build.lib.actions.Artifact;
@@ -84,6 +85,16 @@ public class RemoteActionFileSystem extends DelegateFileSystem {
         stream(outputArtifacts).collect(toImmutableMap(Artifact::getExecPath, a -> a));
     this.inputFetcher = checkNotNull(inputFetcher, "inputFetcher");
     this.remoteOutputTree = new RemoteInMemoryFileSystem(getDigestFunction());
+  }
+
+  @VisibleForTesting
+  protected RemoteInMemoryFileSystem getRemoteOutputTree() {
+    return remoteOutputTree;
+  }
+
+  @VisibleForTesting
+  protected FileSystem getLocalFileSystem() {
+    return delegateFs;
   }
 
   /** Returns true if {@code path} is a file that's stored remotely. */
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -438,6 +449,47 @@ private void downloadFileIfRemote(PathFragment path) throws IOException {
     }
   }
 
+  @Override
+  public void renameTo(PathFragment sourcePath, PathFragment targetPath) throws IOException {
```
tjgq, line 453: Should we assert that `sourcePath` and `targetPath` are under `outputBase`? (I can't imagine a legitimate use case for moving to or from the source tree.) And in that case, do we ever need to fall back to the local filesystem?

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
```diff
@@ -438,6 +449,47 @@ private void downloadFileIfRemote(PathFragment path) throws IOException {
...
+  public void createDirectoryAndParents(PathFragment path) throws IOException {
+    super.createDirectoryAndParents(path);
+    if (path.startsWith(outputBase)) {
+      remoteOutputTree.createDirectoryAndParents(path);
+    }
+  }
+
+  @Override
+  public boolean createDirectory(PathFragment path) throws IOException {
+    boolean created = delegateFs.createDirectory(path);
```
tjgq, line 486: Same as above - is it stricly necessary to create the directory in the local filesystem?

tjgq (changes requested):

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: Added assertions for `sourcePath` and `targetPath`. Considering dynamic execution, both local and remote could have files so we need to try to rename on both. Also if the file is in the remote AND local (because it was staged), we need to keep them in sync.

src/main/java/com/google/devtools/build/lib/remote/RemoteActionFileSystem.java
coeuvre: yes, for dynamic execution AND for action logs.

tjgq (approved):

Issue #8162, opened 2019-04-26
import_deps_checker tests aren't compatible with the current Bazel JDK


katre:
Because tests under //src/java_tools haven't been run, the code has decayed badly. I don't understand how to fix these.

# To reproduce:
```
$ bazel build --nokeep_going  //src/java_tools/import_deps_checker/javatests/...
INFO: Analysed 31 targets (0 packages loaded, 0 targets configured).
INFO: Found 31 targets...
ERROR: missing input file '@local_jdk//:jre/lib/rt.jar'
ERROR: /usr/local/google/home/jcater/repos/bazel/src/java_tools/import_deps_checker/javatests/com/google/devtools/build/importdeps/BUILD:84:1: //src/java_tools/import_deps_checker/javatests/com/google/devtools/build/importdeps:DepsCheckerClassVisitorTest: missing input file '@local_jdk//:jre/lib/rt.jar'
ERROR: /usr/local/google/home/jcater/repos/bazel/src/java_tools/import_deps_checker/javatests/com/google/devtools/build/importdeps/BUILD:84:1 1 input file(s) do not exist
INFO: Elapsed time: 0.325s, Critical Path: 0.02s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

I suspect that even cleaning this up to build, the tests still won't actually pass.

katre:
In #8164, I add all tests under //src/java_tools _except_ the tests for `import_deps_checker`. Be sure to re-enable the tests when this is fixed.

katre:
Buildtools also fails under jdk8. I don't know if there's a way to fix this for both.

iirina:
@katre can you also link a log or an error snippet for buildtools?

iirina:
I found the buildkite run: https://buildkite.com/bazel/bazel-bazel-github-presubmit/builds/2717#533453ca-1692-4357-a693-15e7257602b1

cushon:
@kevin1e100 FYI for import_deps_checker.

`@local_jdk` shouldn't be used anymore, but even if the test resolves a JDK correctly `rt.jar` doesn't exist for JDK > 8.

I would try dropping in `@bazel_tools//tools/jdk:platformclasspath` instead of `@local_jdk//:jre/lib/rt.jar` and see if that works.

> `java.lang.NoSuchFieldError: ANNOTATION_PROCESSOR_MODULE_PATH`

[...]

PR #16491, opened 2022-10-17
Re-enable some tests

cushon:
https://github.com/bazelbuild/bazel/issues/8162


meteorcloudy:
This only changed the postsubmit.yml, we'll need to change the presubmit.yml to actually check the result. Also, we need to remove for other platforms.

meteorcloudy (changes requested):
Please also update presubmit.yml and other platforms.

Issue #13275, opened 2021-03-29
Add an ability to print line number or stack trace of the failing assert in unit tests in Starlark


hlopko:
Hello dear Bazel maintainers :)

I'm re-filing this feature request from https://github.com/bazelbuild/bazel-skylib/issues/290.

Right now there is not a way to print an assertion failure with the stack trace. We currently print the assertion error with the location `.../external/bazel_skylib/lib/unittest.bzl:351:10:`. It would be much more useful if we either printed the entire stack trace, or show the location of the callsite.

An argument could be made that there should be just one assertion per unit test, in that scenario this is not a problem. However there is so much boilerplate involved with adding a unit test that this is not practical to have one assertion per test. And even Bazel's internal `BuildViewTestCase` tests have multiple assertions per test.

@tetromino suggested to add something like a `stacktrace()` builtin or perhaps a `with_stacktrace` optional named parameter to `print()`. Both sound good to me.

Thank you!

brandjon:
Interesting idea. It sounds useful, though in general we try to shy away from probing the call stack.

hlopko:
If that helps, this is a test that is super hard to debug without stacktrace/callsite location: https://github.com/bazelbuild/rules_rust/blob/main/test/unit/cc_info/cc_info_test.bzl#L9

hlopko:
> though in general we try to shy away from probing the call stack

What is the reason for that?

anforowicz:
This would be useful not only in unit tests, but also with printf-style debugging of bzl scripts.

hlopko:
This issue is still relevant and should stay open.

PR #16492, opened 2022-10-17
Optionally let `print` emit a stack trace

fmeum:
RELNOTES: The new `with_trace` keyword parameter of the `print` function can be used to emit a stack trace for the location of the call, similar to that of the `fail` function.

Fixes https://github.com/bazelbuild/bazel/issues/13275
Work towards https://github.com/bazelbuild/bazel-skylib/issues/290

### DIFF
--- a/src/main/java/net/starlark/java/eval/EvalException.java
+++ b/src/main/java/net/starlark/java/eval/EvalException.java
@@ -108,7 +108,7 @@ public class EvalException extends Exception {
    */
   public final String getMessageWithStack(SourceReader src) {
     if (callstack != null) {
-      return formatCallStack(callstack, getMessage(), src);
+      return formatCallStack(callstack, getMessage(), src, /*isPrint=*/false);
     }
     return getMessage();
   }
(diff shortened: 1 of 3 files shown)


fmeum:
@tetromino Would you be available to review this?

My motivation for suggesting to add this is two-fold:
1. Without https://github.com/bazelbuild/bazel-skylib/issues/290, it is essentially impossible to write custom asserts or debug assert failures which don't (or can't) have distinctive failure messages.
2. Deprecations are often handled by adding a `print("If you see this, update usage of foo to bar")` to some rule or macro. Without a trace, it is extremely difficult to find the actual usage site. For example, right now `bazel build //src:bazel-dev` prints the following very unhelpful string that would be much simpler to root cause if it had the ability to emit a stack trace:

```
DEBUG: .../external/build_bazel_rules_nodejs/index.bzl:122:10: WARNING: check_rules_nodejs_version has been removed. This is a no-op, please remove the call.
```

brandjon:
This sounds like a useful feature, though I'm not in love with adding more API to `print()`. I'll make a note to bring this up with comius@ in the new year, we can have a design review then.

The stakes are higher than for normal APIs because `print()` is part of the language spec. So perhaps the discussion should occur within bazelbuild/starlark too.

PR #16495, opened 2022-10-18
Don't overwrite errors during `bazel run`.

meisterT:
We output the "Build completed successfully" message via the progress bar (IMHO, as a more long term fix we should fix that). The progress bar clears previous progress bar contents. Previously, we did miss to mark the build as no longer running when it was actually complete. Both at this point of the invocation, and later when the whole invocation was complete we did output the "Build completed successfully" message, deleting what we thought would be the contents of the progress bar, not aware that there was an ERROR message in between.

The new test illustrates this behavior.

In 7c074b5a5f03a05a108fdfed1910fe75a538a323, `buildComplete` was changed into `buildRunning` and the boolean value was reversed. But we missed changing the value on `AfterCommandEvent`.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/UiEventHandler.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/UiEventHandler.java
@@ -652,7 +652,7 @@ public final class UiEventHandler implements EventHandler {
   @Subscribe
   public void afterCommand(AfterCommandEvent event) {
     synchronized (this) {
-      buildRunning = true;
+      buildRunning = false;
     }
     completeBuild();
     try {
--- a/src/test/shell/integration/ui_test.sh
+++ b/src/test/shell/integration/ui_test.sh
@@ -695,4 +695,22 @@ EOF
   expect_log_n "^"$'\e'"\[31m"$'\e'"\[1mFAILED:"$'\e'"\[0m Build did NOT complete successfully" 4
 }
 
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/devtools/build/lib/runtime/UiEventHandler.java
```diff
@@ -652,7 +652,7 @@ public void noBuildFinished(NoBuildRequestFinishedEvent event) {
   @Subscribe
   public void afterCommand(AfterCommandEvent event) {
     synchronized (this) {
-      buildRunning = true;
+      buildRunning = false;
     }
     completeBuild();
```
benjaminp, line 657: From the implementation of `completeBuild`, it looks like this change is exactly equivalent to removing this call.

meisterT: I came to the same conclusion yesterday, but then thought this would be the minimal change to fix the bug in case I didn't spot all the things.

benjaminp (approved):
I certainly agree https://github.com/bazelbuild/bazel/commit/7c074b5a5f03a05a108fdfed1910fe75a538a323#diff-f02379f948b99b60633b3ba4caedade644945f7dba4c0308a91450c03f0b3dfeR507 is erroneous.

meisterT:
Cc @haxorz

Issue #3886, opened 2017-10-10
Default local resources should respect cgroup limits on Linux


nelhage:
Please provide the following information. The more we know about your system and use case, the more easily and likely we can help.

### Description of the problem / feature request / question:

Bazel, by default, looks at available RAM on the system to set `local_resources` defaults, so as to best-use the resources of the machine.

Unfortunately, inside a Docker container or other cgroup environment, the system-wide memory statisics (`/proc/meminfo`, the output of `free`, etc) reflect the memory usage of the host, not the container.

bazel should make a best-effort attempt to find the effective cgroup memory controller limits, and use those.

### If possible, provide a minimal example to reproduce the problem:

I ran into this in a [CircleCI build](https://circleci.com/gh/livegrep/livegrep/25); All builds would fail with

> Server terminated abruptly (error code: 14, error message: '', log file: '/root/.cache/bazel/_bazel_root/f85b6fb5740e6e8c7efea142eec4b6e8/server/jvm.out')

until I added `build --local_resources=4096,4,1.0` to my `.bazelrc`.

Circle's build containers report 60G of RAM, but are cgroup-limited to 4G, so building any large application on Circle ought reproduce the issue.

### Environment info
* Operating System:
Linux; Tested on Ubuntu 16.04

* Bazel version (output of `bazel info release`):
`release 0.5.4`

* If `bazel info release` returns "development version" or "(@non-git)", please tell us what source tree you compiled Bazel from; git commit hash is appreciated (`git rev-parse HEAD`):

###  Have you found anything relevant by searching the web?
(e.g. [StackOverflow answers](http://stackoverflow.com/questions/tagged/bazel),
[GitHub issues](https://github.com/bazelbuild/bazel/issues),
email threads on the [`bazel-discuss`](https://groups.google.com/forum/#!forum/bazel-discuss) Google group)

There are a number of reports online of people puzzling with bazel OOM-ing. It's hard to know how many of them root-cause to this issue, but almost certainly some of them do, since container environments are increasingly popular these days.
- https://github.com/bazelbuild/bazel/issues/3020
- https://github.com/tensorflow/tensorflow/issues/9940
- https://github.com/bazelbuild/bazel/issues/3251

### Anything else, information or logs or outputs that would be helpful?
(If they are large, please upload as attachment or provide link).

https://fabiokung.com/2014/03/13/memory-inside-linux-containers/ has some notes on how to detect memory availability inside containers.

Steve-Munday:
[...]

PR #16512, opened 2022-10-19
Use `Runtime#availableProcessors` on Linux

### LINKED ISSUE #5042: make bazel cpu counter container aware
https://github.com/bazelbuild/bazel/blob/827007729889126cd7e07a08116f9bcb82f819a9/src/main/java/com/google/devtools/build/lib/actions/LocalHostResourceManagerLinux.java 

reads the number of CPUs from /proc/cpuinfo. AFAIK, Docker doesn't virtualize /proc/ , so if you try to control resource usage on a machine with Docker CPU limits, bazel will disregard this an thrash the machine.

fmeum:
As of JDK 11, `Runtime#availableProcessors` on Linux is aware of cgroup resource limits and thus behaves as expected in containers, whereas `/proc/cpuinfo` still reports the number of logical processors available to the host.

Unfortunately, the JVM-internal `os::total_memory` function, which similarly takes cgroup limits into account when computing the available RAM, does not seem to be accessible from Java.

https://hg.openjdk.java.net/jdk/hs/rev/7f22774a5f42#l6.178

Fixes #5042

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/actions/LocalHostResourceManagerLinux.java
+++ b/src/main/java/com/google/devtools/build/lib/actions/LocalHostResourceManagerLinux.java
@@ -14,27 +14,21 @@
 
 package com.google.devtools.build.lib.actions;
 
-import com.google.common.base.Splitter;
-import com.google.common.io.Files;
 import com.google.devtools.build.lib.unix.ProcMeminfoParser;
-import java.io.File;
 import java.io.IOException;
-import java.nio.charset.Charset;
 import javax.annotation.Nullable;
 
 /**
  * This class estimates the local host's resource capacity for Linux.
  */
 public class LocalHostResourceManagerLinux {
-  private static String cpuInfoContent = null;
 
-  private static final Splitter NEWLINE_SPLITTER = Splitter.on('\n').omitEmptyStrings();
-  private static final String CPU_INFO_FILE = "/proc/cpuinfo";
   private static final String MEM_INFO_FILE = "/proc/meminfo";
 
   private static int getLogicalCpuCount() throws IOException {
-    String content = getCpuInfoContent();
-    return getLogicalCpuCountHelper(content);
+    // As of JDK 11, availableProcessors is aware of cgroups as commonly used by containers.
+    // https://hg.openjdk.java.net/jdk/hs/rev/7f22774a5f42#l6.178
+    return Runtime.getRuntime().availableProcessors();
   }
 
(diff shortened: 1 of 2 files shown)


fmeum:
@philwo I just followed the observation you made in https://github.com/bazelbuild/bazel/issues/3886#issuecomment-856057687. Can you take a look?

As described in the PR description, it doesn't seem possible to get the amount of physical memory this way.

philwo (approved):
Very nice, thank you! :)

@meisterT Could you assign this to someone in your team to review & merge it?

meisterT (approved):

meisterT:
@fmeum I already started the import yesterday, but some internal tests ran into timeouts and I have to figure out whether that is related or not.

fmeum:
Thanks for working on that! Just didn't want it to get lost as I thought that label was required.

meisterT:
Update: that specific test needs at least 3 loading phase threads to run correctly and we were running it in a constrained container with just 2 cores.

While the test can be fixed, it points to a deeper problem here: we automatically deduce the number of loading phase threads (see https://cs.opensource.google/bazel/bazel/+/0232dd7cd18bcf807701a5b81367ed794b1fc357:src/main/java/com/google/devtools/build/lib/runtime/LoadingPhaseThreadsOption.java;l=41) based on the host resources available.

Thus, this change can make quite a difference in how Bazel performs in container environments. I would prefer guarding it behind an incompatible flag, so we have a less risky way of rolling this out.

artem-zinnatullin (approved):
This is wholesome, hope it lands in Bazel 6, we run Bazel CI in k8s with CPU limits and this will finally let Bazel know the actual CPU limits in a container it runs in rather than hardware/virtual_machine CPU limits.

meisterT:
@artem-zinnatullin It is very very unlikely that this will land in Bazel 6 given the comment I wrote above. We need to guard it behind a flag.

fmeum:
@meisterT Friendly ping, which approach do you prefer:

1. startup flag
2. JVM property (settable via `--host_jvm_args`)
3. no flag, but users can use `--local_cpu_resources` and `--loading_phase_threads` to override

fmeum:
@meisterT Friendly ping

meisterT:
@fmeum sorry for the delay, it is merged now.

I added RELNOTES[INC] to the commit description (see https://github.com/bazelbuild/bazel/commit/ab2953ca55dfc61595bd465852a83e21d9a3716f) and think that this is the most reasonable path forward.

PR #16519, opened 2022-10-20
Check type of `lines` parameter in `ScratchFile` Python test helper

fmeum:
Since strings are iterable in Python, accidentally passing a string to the `ScratchFile` test helper ends up writing one character per line, which leads to very confusing error messages.


fmeum:
@Wyverald I've been bitten by this in the past as well and would like to fix this kind of mistake for good.

Wyverald (approved):
nice! this also tripped me up many, many times :|

PR #16523, opened 2022-10-21
Bazel profile: improve reporting of critical path components

saraadams:
The Bazel profile reports the critical path. To enable it to be stored in a separate thread, the `tid` is set to 0 throughout. However, this makes it hard to match critical path events to events in the skyframe-evaluator threads.

This change adds the `tid` of the event in the `args` of the `critical path component` event. Like this, the critical path components continue to be listed in the critical path thread, as well as allowing the unique identification of which actual event in the evaluator threads it corresponds to.


saraadams:
@meisterT I believe you've worked on the Bazel profile code in the past. Would you mind taking a look?

meisterT (approved):
Thanks, that makes sense!

PR #16527, opened 2022-10-21
Fix identical gcov json file name problem

krisukox:
**The problem:**

Bazel moves all `.gcov.json.gz` files to one directory. If in a test target, two source files have identical names, then the second `.gcov.json.gz` overwrites the first one.

**The solution:**

I added `gcno_path` to the move destination in order to distinguish multiple `.gcov.json.gz` with the same name.

**Testing:**

In the `test_cc_test_coverage_gcov` test case I added the `different/a.cc` source file, so currently we have the following source tree:
```
coverage_srcs/a.h
coverage_srcs/a.cc
coverage_srcs/b.h
coverage_srcs/t.cc
coverage_srcs/different/a.h
coverage_srcs/different/a.cc
```

gcda and gcno files are created next to the source files. The final `gcov.json` files are placed in the corresponding paths:
```
$COVERAGE_DIR_VAR/coverage_srcs/*a.gcov.json.gz
$COVERAGE_DIR_VAR/coverage_srcs/*t.gcov.json.gz
$COVERAGE_DIR_VAR/coverage_srcs/different/*a.gcov.json.gz
```


src/test/shell/bazel/bazel_cc_code_coverage_test.sh
```diff
@@ -100,37 +115,26 @@ EOF
...
-  if [[ $not_found -ne 0 ]]; then
-    agcno=$(ls *a.gcno)
-    tgcno=$(ls *t.gcno)
-    agcda=$(ls *a.gcda)
-    tgcda=$(ls *t.gcda)
-    mv $agcno coverage_srcs/$agcno
-    mv $tgcno coverage_srcs/$tgcno
-    mv $agcda coverage_srcs/$agcda
-    mv $tgcda coverage_srcs/$tgcda
-  fi
```
krisukox, line 130: This is not the case using the current approach in the `generate_and_execute_instrumented_binary` function.

src/test/shell/bazel/bazel_cc_code_coverage_test.sh
```diff
@@ -100,37 +115,26 @@ EOF
...
+      coverage_srcs/a.cc coverage_srcs/a.o \
+      coverage_srcs/different/a.cc coverage_srcs/different/a.o \
+      coverage_srcs/t.cc coverage_srcs/t.o
+
+  find .
```
c-mita, line 131: Should drop the find; it clutters the logs when there are failures.

src/test/shell/bazel/bazel_cc_code_coverage_test.sh
c-mita: Is this because every file gets it's own invocation with the relevant paths, so g++ doesn't just dump everything in one place?

c-mita (changes requested):
One minor thing, but otherwise LGTM.

src/test/shell/bazel/bazel_cc_code_coverage_test.sh
krisukox: Done

c-mita (approved):
LGTM

Issue #12992, opened 2021-02-10
Can not reenable ui event filters


ptarjan:
### Description of the problem / feature request:

Running this works as expected:
```
$ bazel test //main-street/... --ui_event_filters=-DEBUG,-INFO
INFO: Build completed successfully, 1 total action
Test cases: finished with 4280 passing and 0 failing out of 4280 test cases

INFO: Build completed successfully, 1 total action
```

But if you try to put them back, it doesn't do anything:
```
$ bazel test //main-street/... --ui_event_filters=-DEBUG,-INFO --ui_event_filters=+DEBUG,+INFO
INFO: Build completed successfully, 1 total action
Test cases: finished with 4280 passing and 0 failing out of 4280 test cases

INFO: Build completed successfully, 1 total action
```

I expect the second command to output the same thing as if I just did:
```
$ bazel test //main-street/... --ui_event_filters=+DEBUG,+INFO
INFO: Invocation ID: 1dd3145a-0ce8-424c-aae2-7a3622bef944
DEBUG: /private/var/tmp/_bazel_paul.tarjan/16b823dd05de063e5dc3f5967533d5ec/external/bazel_gazelle/internal/go_repository.bzl:189:18: com_github_confluentinc_confluent_kafka_go: gazelle: /private/var/tmp/_bazel_paul.tarjan/16b823dd05de063e5dc3f5967533d5ec/external/com_github_confluentinc_confluent_kafka_go/kafka/go_rdkafka_generr/go_rdkafka_generr.go: error reading go file: /private/var/tmp/_bazel_paul.tarjan/16b823dd05de063e5dc3f5967533d5ec/external/com_github_confluentinc_confluent_kafka_go/kafka/go_rdkafka_generr/go_rdkafka_generr.go: pkg-config not supported: #cgo pkg-config: --static rdkafka
gazelle: /private/var/tmp/_bazel_paul.tarjan/16b823dd05de063e5dc3f5967533d5ec/external/com_github_confluentinc_confluent_kafka_go/kafka/build_dynamic.go: error reading go file: /private/var/tmp/_bazel_paul.tarjan/16b823dd05de063e5dc3f5967533d5ec/external/com_github_confluentinc_confluent_kafka_go/kafka/build_dynamic.go: pkg-config not supported: #cgo pkg-config: rdkafka
[...]

PR #16531, opened 2022-10-23
Fix passing --ui_event_filters multiple times

AlessandroPatti:
Fixes https://github.com/bazelbuild/bazel/issues/12992

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
@@ -51,15 +51,14 @@ public class UiOptions extends OptionsBase {
   public static class EventFiltersConverter extends Converter.Contextless<EventFiltersConverter.EventKindFilters> {
 
     /**
-     * Container for an EventKind input filter. Added event kinds and removed event kinds are respectively events
-     * this filter wants to enable and disable.
+     * Container for an EventKind input filter.
      */
     @AutoValue
     public static abstract class EventKindFilters {
-      public abstract ImmutableSet<EventKind> getAddedEventKinds();
-      public abstract ImmutableSet<EventKind> getRemovedEventKinds();
-      public static EventKindFilters from(ImmutableSet<EventKind> added, ImmutableSet<EventKind> removed) {
-        return new AutoValue_UiOptions_EventFiltersConverter_EventKindFilters(added, removed);
+      public abstract ImmutableSet<EventKind> getFilteredEventKinds();
+      public abstract ImmutableSet<EventKind> getUnfilteredEventKinds();
+      public static EventKindFilters from(ImmutableSet<EventKind> filtered, ImmutableSet<EventKind> unfiltered) {
+        return new AutoValue_UiOptions_EventFiltersConverter_EventKindFilters(filtered, unfiltered);
       }
     }
     private final CommaSeparatedOptionListConverter commaSeparatedListConverter;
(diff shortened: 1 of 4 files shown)


brentleyjones:
Bump. Can someone review this?

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -46,57 +47,19 @@ public enum UseCurses {
...
-
-    public EventFiltersConverter() {
-      this.delegate = new CommaSeparatedOptionListConverter();
-    }
+  public static class EventFiltersConverter extends CommaSeparatedOptionListConverter {
+    EnumConverter<EventKind> validator = new EnumConverter<>(EventKind.class, "ui event") {};
 
     @Override
-    public List<EventKind> convert(String input) throws OptionsParsingException {
+    public ImmutableList<String> convert(String input) throws OptionsParsingException {
```
meisterT, line 54: What is the reason that you moved the actual conversion out of `convert`?

meisterT (commented):
Please rebase on top of current master.

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
AlessandroPatti: I'm not entirely sure as it's been a long time since I made this change, but I think the reason was that I needed to have the union of all the the inputs to compute the filters. Options with `allowMultiple = true` will aways concatenate the list converted values for each `input`, that is why it wasn't possible to add an event back in: once an `input` was converted to filter an EventKind, a subsequent `input` could not remove the filter from the accumulator. I found it easier to just move the conversion outside, after all the strings have been accumulated.

AlessandroPatti: @meisterT Rebased and moved all the conversions back into the converter. Need to store two separate lists of events (removed and added) thou. Let me know if there is something I'm missing to avoid this.

meisterT (approved):

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -46,52 +49,51 @@ public enum UseCurses {
...
+    public EventKindFilters convert(String input) throws OptionsParsingException {
       if (input.isEmpty()) {
         // This method is not called to convert the default value
         // Empty list means that the user wants to filter all events
-        return new ArrayList<>(EventKind.ALL_EVENTS);
+        return EventKindFilters.from(EventKind.ALL_EVENTS, ImmutableSet.of());
       }
-      List<String> filters = this.delegate.convert(input, /*conversionContext=*/ null);
-      EnumConverter<EventKind> eventKindConverter = new EventKindConverter(input);
+      List<String> filters = commaSeparatedListConverter.convert(input, /*conversionContext=*/ null);
```
meisterT, line 75: change this type to ImmutableList

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -46,52 +49,51 @@ public enum UseCurses {
...
   /** Converter for {@link EventKind} filters * */
-  public static class EventFiltersConverter extends Converter.Contextless<List<EventKind>> {
-
-    /** A converter for event kinds. */
-    public static class EventKindConverter extends EnumConverter<EventKind> {
-
-      public EventKindConverter(String typeName) {
-        super(EventKind.class, typeName);
+  public static class EventFiltersConverter extends Converter.Contextless<EventFiltersConverter.EventKindFilters> {
+    @AutoValue
```
meisterT, line 53: Add some javacdoc for this public class

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -276,6 +278,15 @@ public boolean useCursorControl() {
     return useCursesEnum == UseCurses.YES || (useCursesEnum == UseCurses.AUTO && isATty);
   }
 
+  public ImmutableSet<EventKind> getFilteredEvents() {
```
meisterT, line 281: rename method to make it more clear that this doesn't return events, e.g. to `getCombinedEventFilter` or `getFilteredEventKinds`

src/main/java/com/google/devtools/build/lib/runtime/UiOptions.java
```diff
@@ -46,52 +49,51 @@ public enum UseCurses {
...
-      List<String> filters = this.delegate.convert(input, /*conversionContext=*/ null);
-      EnumConverter<EventKind> eventKindConverter = new EventKindConverter(input);
+      List<String> filters = commaSeparatedListConverter.convert(input, /*conversionContext=*/ null);
+
+      HashSet<EventKind> removedEvents = new HashSet<>();
+      HashSet<EventKind> addedEvents = new HashSet<>();
 
-      HashSet<EventKind> filteredEvents = new HashSet<>();
       for (String filter : filters) {
         if (!filter.startsWith("+") && !filter.startsWith("-")) {
```
meisterT, line 81: Can you please add one more test around this branch?

meisterT (commented):
Added a few minor comments which came up in internal code review.

sgowroji:
Hi @meisterT, Can we import this PR as it is awaiting to merge and got approved. Thanks!

meisterT:
@sgowroji thanks for checking - I already started the import of this PR myself, see cl/515554663 internally.

meisterT:
Thanks @AlessandroPatti for your contribution and patience!

PR #16533, opened 2022-10-23
Add --regex_features option

AlessandroPatti:
The `--features` flag is very useful to experiment but has a very large blast radius. As features often change compilation flags, this causes full rebuilds. The alternative at the moment is to add the feature to the `features` attribute of the individiual targets or packages which seems suboptimal. `--regex_features` would allow adding and removing features for specific targets directly on command line.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/BUILD
+++ b/src/main/java/com/google/devtools/build/lib/analysis/BUILD
@@ -1709,6 +1709,9 @@ java_library(
 java_library(
     name = "config/core_option_converters",
     srcs = ["config/CoreOptionConverters.java"],
+    exports = [
+        "//src/main/java/com/google/devtools/build/lib/exec:regex_filter_assignment_converter",
+    ],
     deps = [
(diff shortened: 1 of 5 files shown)


oquenchil:
We don't have regex in Starlark. They would be have to be available in Starlark before we can approve this change. That would be a big decision that would need a design doc and be discussed by everyone.

AlessandroPatti:
@oquenchil This is not proposing to add regex in starlark. They pr simply adds a new flag to enable features on specific targets based on regexes. There are several flags that already do something similar. (e.g. [--flaky_test_attempts](https://bazel.build/reference/command-line-reference#flag--flaky_test_attempts))

oquenchil:
Our goal is to re-write every C++ specific flag to a Starlark implementation (--flaky_test_attempts is not related to C++) , therefore there would have to be support for regular expressions in Starlark in the future, otherwise the C++ specific flags cannot be re-written. The discussion of regular expressions in Starlark would have to happen before this change goes in.

PR #16534, opened 2022-10-23
Expose current repository name to Java with @AutoBazelRepository 

fmeum:
Java targets depending on `@bazel_tools//tools/java/runfiles` can add the new `@AutoBazelRepository` to a class to have an annotation processor generate a companion class with a `BAZEL_REPOSITORY` constant containing the repository name of the target that compiled the class.

This requires a small addition to JavaBuilder to parse the repository name out of the target label and pass it to javac as a processor option.

Work towards #16124


fmeum:
@cushon I followed your recommendation in https://github.com/bazelbuild/bazel/pull/16281#issuecomment-1276421472 and experimented with annotation processing rather than JavaBuilder magic. The ergonomics seem decent enough and the only change to JavaBuilder is passing the repository name via `-A`.

The integration test is most likely not in the correctly place, I only put it there so that the test only runs against java_tools built from HEAD. How would I do this properly - split up the PR into two parts with a java_tools update in between?

@Wyverald Do you think this is usable?

fmeum:
I don't understand the test failure - locally, `bazel test //src/test/shell/bazel:bazel_java17_test` passes. Any idea what could be up with that?

fmeum:
I changed this commit so that the flag is injected in `JavaCommon` rather than `JavaBuilder`. That does require duplicating it in two places though.

tools/java/runfiles/AutoBazelRepositoryProcessor.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.annotation.processing.SupportedSourceVersion;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.Name;
+import javax.lang.model.element.TypeElement;
+import javax.tools.Diagnostic.Kind;
+
+@SupportedAnnotationTypes("com.google.devtools.build.runfiles.AutoBazelRepository")
+@SupportedOptions("bazel.repository")
+@SupportedSourceVersion(SourceVersion.RELEASE_8)
```
cushon, line 39: Prefer

```java
  @Override
  public SourceVersion getSupportedSourceVersion() {
    return SourceVersion.latestSupported();
  }
```

to avoid warnings for compilations targeting Java > 8

fmeum: Thanks, changed.

tools/java/runfiles/AutoBazelRepositoryProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+    try (PrintWriter out = new PrintWriter(
+        processingEnv.getFiler().createSourceFile(generatedClassName).openWriter())) {
+      out.printf("package %s;\n", generatedClassPackage);
+      out.printf("\n");
+      out.printf("class %s {\n", generatedClassSimpleName);
+      out.printf("  /**\n");
+      out.printf("   * The canonical name of the repository containing the Bazel target that\n");
+      out.printf("   * compiled {@link %s}.\n", annotatedClass.getQualifiedName().toString());
+      out.printf("   */\n");
+      out.printf("  static final String BAZEL_REPOSITORY = \"%s\";\n", repositoryName);
```
fmeum, line 94: @cushon Given that the class is called `AutoBazelRepository_X`, should I maybe replace this with `NAME`? More concise and less repetitive.

Wyverald (approved):
looks good!

tools/java/runfiles/testing/AutoBazelRepositoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Unit tests for {@link AutoBazelRepository}.
+ */
+@RunWith(JUnit4.class)
+@AutoBazelRepository
+public final class AutoBazelRepositoryTest {
```
comius, line 28: 1. This looks like a weird tests, the arrange part is happening by the annotation. I'd be more comfortable the test is moved into the shell test.

2. Why isn't repository name properly injected? This is a java_test and it is in a repository. Shouldn't this be a failure?

comius (changes requested):

tools/java/runfiles/AutoBazelRepositoryProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+        .flatMap(element -> roundEnv.getElementsAnnotatedWith(element).stream())
+        .map(element -> (TypeElement) element)
+        .forEach(this::emitClass);
+    return true;
+  }
+
+  private void emitClass(TypeElement annotatedClass) {
+    // This option is always provided by JavaBuilder.
+    String repositoryName = processingEnv.getOptions()
+        .getOrDefault("bazel.repository", "for testing only");
```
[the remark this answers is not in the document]
fmeum: Made this a hard error.

comius (approved):

tools/java/runfiles/testing/AutoBazelRepositoryTest.java
comius: Agreed.

fmeum:
@bazel-io flag

Wyverald:
@bazel-io fork 6.0.0

src/main/java/com/google/devtools/build/lib/rules/java/JavaCommon.java
```diff
@@ -374,6 +374,8 @@ private ImmutableList<String> computeJavacOpts(Collection<String> extraRuleJavac
     return ImmutableList.<String>builder()
         .addAll(javaToolchain.getJavacOptions(ruleContext))
         .addAll(extraRuleJavacOpts)
+        // Consumed by com.google.devtools.build.runfiles.AutoBazelRepositoryProcessor.
+        .add("-Abazel.repository=" + ruleContext.getRepository().getName())
```
Wyverald, line 378: @cushon asks:

>>> Instead of unconditionally passing it to all builds, could it be passed to builds that use AutoBazelRepository?
>>
>> (me) how would one check whether a build is using AutoBazelRepository?
>
> It's just a java_plugin, right? So they'd have to depend on it, and com.google.devtools.build.runfiles.AutoBazelRepositoryProcessor would be in the list of processors for the compilation, and so on.

fmeum: I have something that compiles and will run the tests, but I don't know whether I'm holding it right. (Seems like `RunfilesLibraryInfo` made a return in the form of an annotation processor :-)).

fmeum: @cushon Now that I'm digging into this deeper, how is it decided whether a plugin is "API generating" or not? I am not sure whether I would need to pass in `-Abazel.repository` for header compilation actions.

fmeum: Added a commit that implements this and has the test passing.

cushon: > it does generate API (a new class), even if we don't really want dependents to use the generated class

I think we probably want `generates_api = False`, unless we actually want dependents to be able to use the generated class, or decide it's not worth trying to discourage.

Using the generated code in the same target is fine, `generates_api = True` is for cases that deliberately support other targets directly accessing the generated code.

fmeum: Okay, then `generates_api = False` sounds right. We really want the generated class to be used only from the annotated class.

src/main/java/com/google/devtools/build/lib/rules/java/JavaInfoBuildHelper.java
```diff
@@ -284,6 +285,22 @@ public JavaInfo createJavaCompileAction(
 
     JavaToolchainProvider toolchainProvider = javaToolchain;
 
+    JavaPluginInfo pluginInfo = mergeExportedJavaPluginInfo(plugins, deps);
+    Builder<String> allJavacOptsBuilder = ImmutableList.<String>builder()
+        .addAll(toolchainProvider.getJavacOptions(starlarkRuleContext.getRuleContext()))
+        .addAll(javaSemantics.getCompatibleJavacOptions(starlarkRuleContext.getRuleContext(),
+            toolchainProvider));
+    if (pluginInfo.plugins().processorClasses().toList()
```
Wyverald, line 293: @cushon asks:

> Are we OK with flattening the nested set here?
>
> Another option that preserves this behaviour would be to be to add the javacopt in JavaBuilder instead (I think an earlier version of the PR did that). But maybe these nested sets are small enough it isn't a big concern, up to you.

fmeum: `processorClasses().toSet()` is called in a few places that I believe are executed during the analysis phase, so I assume it already ends up flattened.

I pushed a commit to use `toSet` rather than `toList` for consistency.

Issue #16232, opened 2022-09-07
Allow starlark rules to pass `--run_under` to executables


UebelAndre:
### Description of the feature request:

I would like to be able to intercept the [--run_under](https://bazel.build/reference/command-line-reference#flag--run_under) flag in a starlark rule to be able to pass it as an environment variable or argument to a test or action to have better control over when it's used. In [rules_rust](https://github.com/bazelbuild/rules_rust), executables are the compiled Rust binary which allows me to pass `--run_under` and start a debugger (see https://github.com/bazelbuild/rules_rust/issues/370#issuecomment-796592288). However, there's some interest in being able to instead generate a process wrapper that would subprocess or exec into the compiled executable (https://github.com/bazelbuild/rules_rust/issues/1303). If this were to be the case, the current use case of `--run_under` would break as the thing being executed by `--run_under` is a process wrapper. I want to be able to use process wrappers and maintain the intent of `--run_under` being run on the correct executable and not a process wrapper.

### What underlying problem are you trying to solve with this feature?

By being able to control when `--run_under` is used, rules authors would be able to more accurately translate the intent of `--run_under` to run the correct executable under the specified command while still handling ephemeral things the process may emit (coverage or profiling reports, etc).

### Which operating system are you running Bazel on?

Linux, MacOS, Windows

### What is the output of `bazel info release`?

release 5.3.0

UebelAndre:
@fmeum wondering if you have thoughts on this one. Could this be handled similar to the proposal for `RunEnvironmentInfo.arguments` (https://github.com/bazelbuild/bazel/issues/16076) where [RunEnvironmentInfo](https://bazel.build/rules/lib/RunEnvironmentInfo) gains a `run_under` field that:
- If set will follow the standard `--run_under` behavior.
- If set to `None`, will explicitly disable `--run_under`.

Additionally, the `--run_under` would then be consumable via [ctx.var](https://bazel.build/rules/lib/ctx#var) where there'd be some RUN_UNDER value equivalent to whatever the user set (perhaps, not sure). This way rule authors could set custom environment variables via `RunEnvironmentInfo` to have their process wrappers invoke the intended binary for the `--run_under` content.

fmeum:
I'm not too familiar with `--run_under` to have a qualified opinion. 

Would your proposal work with both simple commands and Bazel targets passed in as `--run_under`? Could you give an example where a rule would need to act on the particular value of `--run_under` at analysis time?

[...]

PR #16540, opened 2022-10-24
Added `run_under_env` argument to `RunEnvironmentInfo`

UebelAndre:
Executable Starlark rules can use the `run_under_env` parameter of `RunEnvironmentInfo` to define an environment variable with the value of `--run_under` when it's set. This is useful in cases where executable targets execute some sort of process-wrapper which either sub-processes or execs into the expected process. Before, `--run_under` would happen on the process wrapper which made debuggers and profiling tools harder to use as the process they were attached to was not the process users cared about. 

After this change, by having an executable Starlark rule return `RunEnvironmentInfo(run_under_env = "RUN_UNDER_VALUE")`, `run` or `test` invocations which pass `--run_under='strace -c'` will cause the executable to have `RUN_UNDER_VALUE='strace -c'` set in it's environment.

RELNOTES: Executable Starlark rules can use the `run_under_env` parameter of `RunEnvironmentInfo` to define an environment variable with the value of `--run_under` when it's set. This allows Starlark rule authors to control how this value is used.

closes https://github.com/bazelbuild/bazel/issues/16232

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/RuleConfiguredTargetBuilder.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RuleConfiguredTargetBuilder.java
@@ -474,6 +474,7 @@ public final class RuleConfiguredTargetBuilder {
     if (environmentProvider != null) {
       testActionBuilder.addExtraEnv(environmentProvider.getEnvironment());
       testActionBuilder.addExtraInheritedEnv(environmentProvider.getInheritedEnvironment());
+      testActionBuilder.setRunUnderEnv(environmentProvider.getRunUnderEnv());
     }
 
     TestParams testParams =
--- a/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
+++ b/src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
@@ -174,6 +174,90 @@ public class RunCommand implements BlazeCommand {
     return args;
   }
 
+  private String computeRunUnderArgs(
+      RunUnder runUnder, ConfiguredTarget targetToRun, ConfiguredTarget runUnderTarget) {
+    String runUnderValue = runUnder.getValue();
+    if (runUnderTarget != null) {
+      // --run_under specifies a target. Get the corresponding executable.
+      // This must be an absolute path, because the run_under target is only
+      // in the runfiles of test targets.
+      runUnderValue =
+          runUnderTarget
+              .getProvider(FilesToRunProvider.class)
+              .getExecutable()
+              .getPath()
+              .getPathString();
+      // If the run_under command contains any options, make sure to add them
+      // to the command line as well.
+      List<String> opts = runUnder.getOptions();
+      if (!opts.isEmpty()) {
+        runUnderValue += " " + ShellEscaper.escapeJoinAll(opts);
+      }
+    }
+
+    return runUnderValue;
+  }
+
+  private void constructCommandLine(
+      List<String> cmdLine,
+      List<String> prettyCmdLine,
+      CommandEnvironment env,
+      BuildConfigurationValue configuration,
+      ConfiguredTarget targetToRun,
+      ConfiguredTarget runUnderTarget,
+      List<String> args)
+      throws NoShellFoundException {
+    BlazeRuntime runtime = env.getRuntime();
+    String productName = runtime.getProductName();
+    Artifact executable = targetToRun.getProvider(FilesToRunProvider.class).getExecutable();
+
+    BuildRequestOptions requestOptions = env.getOptions().getOptions(BuildRequestOptions.class);
+
+    PathFragment executablePath = executable.getPath().asFragment();
+    PathPrettyPrinter prettyPrinter =
+        OutputDirectoryLinksUtils.getPathPrettyPrinter(
+            runtime.getRuleClassProvider().getSymlinkDefinitions(),
+            requestOptions.getSymlinkPrefix(productName),
+            productName,
+            env.getWorkspace(),
+            requestOptions.printWorkspaceInOutputPathsIfNeeded
+                ? env.getWorkingDirectory()
+                : env.getWorkspace());
+    PathFragment prettyExecutablePath =
+        prettyPrinter.getPrettyPath(executable.getPath().asFragment());
+
+    RunEnvironmentInfo runEnvironmentInfo =
+        (RunEnvironmentInfo) targetToRun.get(RunEnvironmentInfo.PROVIDER.getKey());
+    boolean runUnderEnvDefined =
+        (runEnvironmentInfo != null && runEnvironmentInfo.getRunUnderEnv() != null);
+
+    RunUnder runUnder = env.getOptions().getOptions(CoreOptions.class).runUnder;
+    // Insert the command prefix specified by the "--run_under=<command-prefix>" option at
+    // the start of the command line whenever RunEnvironmentInfo.run_under_env is not used.
+    if (runUnder != null && !runUnderEnvDefined) {
+      String runUnderValue = computeRunUnderArgs(runUnder, targetToRun, runUnderTarget);
+
+      PathFragment shellExecutable = ShToolchain.getPathForHost(configuration);
+      if (shellExecutable.isEmpty()) {
+        throw new NoShellFoundException();
+      }
+
+      cmdLine.add(shellExecutable.getPathString());
+      cmdLine.add("-c");
+      cmdLine.add(runUnderValue + " " + executablePath.getPathString() + " "
+          + ShellEscaper.escapeJoinAll(args));
+      prettyCmdLine.add(shellExecutable.getPathString());
+      prettyCmdLine.add("-c");
+      prettyCmdLine.add(runUnderValue + " " + prettyExecutablePath.getPathString() + " "
+          + ShellEscaper.escapeJoinAll(args));
+    } else {
+      cmdLine.add(executablePath.getPathString());
+      cmdLine.addAll(args);
+      prettyCmdLine.add(prettyExecutablePath.getPathString());
+      prettyCmdLine.addAll(args);
+    }
+  }
+
   @Override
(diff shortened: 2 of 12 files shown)


UebelAndre:
Some open questions for this draft:
- How are integration tests written for Bazel?
- Should the value set to `run_under_env` include the shell executable?

src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
```diff
@@ -496,10 +509,16 @@ public BlazeCommandResult exec(CommandEnvironment env, OptionsParsingResult opti
       }
       RunEnvironmentInfo environmentProvider = targetToRun.get(RunEnvironmentInfo.PROVIDER);
       if (environmentProvider != null) {
-        actionEnvironment =
             actionEnvironment.withAdditionalVariables(
                 environmentProvider.getEnvironment(),
                 ImmutableSet.copyOf(environmentProvider.getInheritedEnvironment()));
+            if (environmentProvider.getRunUnderEnv() != null) {
+                actionEnvironment.withAdditionalVariables(
+                    Map.ofEntries(Map.entry(
```
fmeum, line 517: You an use `ImmutableMap.of(key, value, ...)` instead.

fmeum:
> * How are integration tests written for Bazel?

Given that the existing `RunEnvironmentInfo` tests are in https://github.com/bazelbuild/bazel/blob/b5db37924fc204eff4ed6ebbccba18da972bf538/src/test/shell/bazel/bazel_rules_test.sh#L670, you should probably add shell tests there. 

> * Should the value set to run_under_env include the shell executable?

That is a good question I don't have a good answer to - I yet have to use `--run_under` in a non-trivial way. Do Starlark rules have access to the shell toolchain?

UebelAndre:
> Do Starlark rules have access to the shell toolchain?

I don't think so. I'm not aware of `shell_toolchain` being something that exists. I think if I could get an integration test it might lead to an answer here.

src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
```diff
@@ -194,6 +194,29 @@ private List<String> computeArgs(ConfiguredTarget targetToRun, List<String> comm
     return args;
   }
 
+  private String computeRunUnderArgs(
+    RunUnder runUnder,
+    ConfiguredTarget targetToRun,
+      ConfiguredTarget runUnderTarget
```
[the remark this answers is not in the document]
UebelAndre: Is there a way to format with vscode? I'm surprised there isn't a formatter script I could just run to not think about it.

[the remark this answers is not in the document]
fmeum: A workaround for the formatting problem is to only format changed lines with [this tool](https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py). A Bazel-wide reformat would require all PRs to go through relatively painful rebases, which may be why it hasn't been done yet.

I think that there was an issue about making `bazel build //...` pass, but I can't find it right now. `bazel test //...` would require fetching a hermetic Android NDK, which AFAIK isn't possible due to the license acceptance step. I usually just check that `bazel build //src:bazel-dev` completes and then fix test build breakages reported by CI.

src/main/java/com/google/devtools/build/lib/starlarkbuildapi/RunEnvironmentInfoApi.java
```diff
@@ -61,6 +63,15 @@ public interface RunEnvironmentInfoApi extends StructApi {
       structField = true)
   List<String> getInheritedEnvironment();
 
+//   @StarlarkMethod(
```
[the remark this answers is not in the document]
fmeum: That was intentional: Adding such a method/struct field would give Starlark rules access to this information in the provider. I wasn't sure whether `arguments` is something that other rules should be able to inspect, so I decided not to expose it for now.

Do you see a reason for depending rules to access the value of `run_under_env`? If there is no clear use case, it may be better not to expose it for now. API creep is real and incompatible changes removing functionality are painful.

UebelAndre:
@fmeum with the first passing build (which took quite a bit of time and effort), I think it's worth a thorough review. Would you be willing to take another look?

src/test/shell/bazel/bazel_rules_test.sh
```diff
@@ -743,4 +743,153 @@ EOF
...
+    return [DefaultInfo(
+        executable = executable_sh,
+    )]
+
+my_wrapper = rule(
+    implementation = _my_wrapper_impl,
+    attrs = {
+        "_script": attr.label(
+            allow_single_file = True,
+            default = Label("//pkg:run_under" + _SCRIPT_EXT),
```
fmeum, line 862: The `Label(...)` call isn't needed, you can just pass a string.

src/test/shell/bazel/bazel_rules_test.sh
```diff
@@ -743,4 +743,153 @@ EOF
...
+        ),
+    },
+    executable = True,
+)
+EOF
+
+  cat >.bazelrc <<'EOF'
+build --announce_rc
```
fmeum, line 870: There is a dedicated `add_to_bazelrc` shell function for this.

src/test/shell/bazel/bazel_rules_test.sh
```diff
@@ -743,4 +743,153 @@ EOF
...
+build --subcommands
```
fmeum, line 871: While useful for writing the test in the first place, I think that `--announce_rc` and `--subcommands` should probably be removed before this gets merged.

src/test/shell/bazel/bazel_rules_test.sh
```diff
@@ -743,4 +743,153 @@ EOF
...
+  # Test --run_under as a custom command
+  bazel test //pkg:my_test --run_under="$(pwd)/${run_under_script}" >$TEST_log 2>&1 \
+    || fail "Test binary with command should have exit code 0"
+
+  # Test --run_under as a target
+  bazel test //pkg:my_test --run_under=//pkg:my_wrapper >$TEST_log 2>&1 \
+    || fail "Test binary with target should have exit code 0"
+
+  # Test --run_under as a custom command
+  bazel run //pkg:my_executable --run_under="$(pwd)/${run_under_script}" >$TEST_log 2>&1 \
```
fmeum, line 887: This test looks pretty thorough!

Only thing I see missing is test coverage for `--run_under` options (`--run_under=//pkg:debugger --attach`) with `run_under_env`.

src/test/shell/bazel/bazel_rules_test.sh
```diff
@@ -743,4 +743,153 @@ EOF
...
+
+my_test = rule(
+    implementation = _my_executable_impl,
+    attrs = {},
+    test = True,
+)
+
+def _my_wrapper_impl(ctx):
+    executable_sh = ctx.actions.declare_file(ctx.attr.name + _SCRIPT_EXT)
+    ctx.actions.expand_template(
```
fmeum, line 847: With no substitutions, you should be able to use `ctx.actions.write` instead.

fmeum:
Looks very good now, thanks for investing the time!

Given that he also reviewed the other `RunEnvironmentInfo` PRs, @comius would be in the best position to review this PR.

src/main/java/com/google/devtools/build/lib/analysis/RuleConfiguredTargetBuilder.java
```diff
@@ -474,6 +474,7 @@ private TestProvider initializeTestProvider(FilesToRunProvider filesToRunProvide
     if (environmentProvider != null) {
```
fmeum, line 474: Comment on the PR description: The description will become the commit message after the merge, so you might want to add a few sentences describing the change and include a brief [`RELNOTES` section](https://bazel.build/contribute/release-notes).

src/test/shell/bazel/bazel_rules_test.sh
UebelAndre: I wanted to use `pkg/run_under.(sh|bat)` directly instead of needing to render the same content into a starlark constant to be written in this rule. This is one of the few actions that let me copy files.

src/test/shell/bazel/bazel_rules_test.sh
UebelAndre: Sure, though in general I would expect all default values for rules to use `Label(...)` since it offers more consistent results across repository boundaries. It's not strictly needed for this test but would be a better thing for other devs to pattern match.

src/test/shell/bazel/bazel_rules_test.sh
fmeum: True, I missed that point!

UebelAndre: I replaced it with `ct.actions.symlink` which I was originally nervous about using because of windows but that seems to work 😄

fmeum: It will do a copy if symlinks aren't enabled. Keep calm and symlink on 🙂

src/test/shell/bazel/bazel_rules_test.sh
UebelAndre: Good catch, updated!

src/main/java/com/google/devtools/build/lib/analysis/RuleConfiguredTargetBuilder.java
UebelAndre: Updated the description. Let me know what you think.

UebelAndre:
@fmeum I think I've addressed all the feedback and the PR is ready for another review.

fmeum (approved):
Looks great, nothing to add :-)

This still has to be reviewed and imported by a Googler though.

UebelAndre:
@comius Hello again, would you have time this week to review this as well? 😅

UebelAndre:
I don't have time to do a rebase which is sad because I think these changes would be really good. Closing so the changes don't appear to be in work.

PR #16545, opened 2022-10-25
Make `bazel run` works with minimal mode

### LINKED ISSUE #11920: UX: make --remote_download_output=minimal download the necessary files when invoking `bazel run`

When using remote build exection, calling `bazel run sometarget --remote_download_output=minimal` doesn't make sense: Bazel will not dowload all the files required to run `sometarget`.

I propose changing the behaviour of `--remote_download_output=minimal` to download files needed to run an action that is called with `bazel run`.

I'm trying to improve the user experience of Bazel.

Not downloading the files needed to run a target passed to `bazel run` is never something that a user wants to do.

I think the behavior would still be consistent with the meaning of `--remote_download_output=minimal`. The documentation says:

> If set to 'minimal' doesn't download any remote build outputs to the local machine, except the ones required by local actions.

I believe it is fair to consider `sometarget` a local action when calling `bazel run sometarget`.

coeuvre:
Always use `ToplevelArtifactsDownloader` when building without the bytes.

It checks the combination of current command (e.g. `build`, `run`, etc.) and download mode (e.g. `toplevel`, `minimal`) to decide whether download outputs for the toplevel targets or not.

Also in the `RunCommand`, we wait for the background downloads before checking the local filesystem.

Fixes #11920.

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
+++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
@@ -929,29 +929,21 @@ public final class RemoteModule extends BlazeModule {
       remoteOutputService.setActionInputFetcher(actionInputFetcher);
       actionContextProvider.setActionInputFetcher(actionInputFetcher);
 
-      boolean useToplevelDownloader = remoteOutputsMode.downloadToplevelOutputsOnly();
-
-      // Download toplevel artifacts for `run` command.
-      if ("run".equals(env.getCommandName())) {
-        useToplevelDownloader = true;
-      }
-
-      if (useToplevelDownloader) {
-        toplevelArtifactsDownloader =
-            new ToplevelArtifactsDownloader(
-                env.getCommandName(),
-                env.getSkyframeExecutor().getEvaluator(),
-                actionInputFetcher,
-                (path) -> {
-                  FileSystem fileSystem = path.getFileSystem();
-                  Preconditions.checkState(
-                      fileSystem instanceof RemoteActionFileSystem,
-                      "fileSystem must be an instance of RemoteActionFileSystem");
-                  return ((RemoteActionFileSystem) path.getFileSystem())
-                      .getRemoteMetadata(path.asFragment());
-                });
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -929,7 +929,14 @@ public void executorInit(CommandEnvironment env, BuildRequest request, ExecutorB
       remoteOutputService.setActionInputFetcher(actionInputFetcher);
       actionContextProvider.setActionInputFetcher(actionInputFetcher);
 
-      if (remoteOutputsMode.downloadToplevelOutputsOnly()) {
+      boolean useToplevelDownloader = remoteOutputsMode.downloadToplevelOutputsOnly();
+
+      // Download toplevel artifacts for `run` command.
+      if ("run".equals(env.getCommandName())) {
```
coeuvre, line 935: We might need to support other commands (e.g. `coverage`, `mobile-install`, etc.). Do you think whether it is better if we **always** use toplevel downloader but pass in the download mode, let the downloaders decide what to download in both minimal/toplevel mode?

tjgq (approved):

src/main/java/com/google/devtools/build/lib/runtime/commands/RunCommand.java
```diff
@@ -306,6 +306,14 @@ public BlazeCommandResult exec(CommandEnvironment env, OptionsParsingResult opti
       return BlazeCommandResult.detailedExitCode(result.getDetailedExitCode());
     }
 
+    if (env.getOutputService() != null) {
```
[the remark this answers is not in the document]
coeuvre: Done.

src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
coeuvre: Done.

coeuvre:
@bazel-io flag

PR #16547, opened 2022-10-25
Infer CPUs for tvOS and watchOS when on Apple Silicon

BalestraPatrick:
This PR contains the same logical change as https://github.com/bazelbuild/bazel/commit/aaae8ce2881c8c1a5d4ad64f20d6e71aa372cf2e but for watchOS and tvOS. When on Apple Silicon, we should build for the right CPU architecture. This is `sim_arm64` for tvOS and `arm64` for watchOS.

On a separate note, I'm not sure if `i386` should be the default architecture for watchOS anymore. I'm happy to change that to `x86_64` to align with all the other platforms if people agree.


brentleyjones:
@bazel-io flag

kshyanashree:
@bazel-io fork 6.0.0

PR #16548, opened 2022-10-25
Use "apparent name" instead of "local name" in Bzlmod docs

fmeum:
"Apparent" is used all over the codebase now, so it makes sense to also use it in user-facing docs.

Also slightly rewords the definition of an "apparent repository name" to make it easier to understand.


Wyverald (approved):

Wyverald:
> @fweikert Are cherry-picks the way to get this into the docs versioned for 6.0.0?

yes

fmeum:
@bazel-io flag

meteorcloudy (approved):
Thanks!

Wyverald:
@bazel-io fork 6.0.0

PR #16549, opened 2022-10-25
Make Java runfiles library repo mapping aware

fmeum:
Work towards #16124


fmeum:
@Wyverald Could you review this? It's useless without #16534 and #16321, but can be tested and reviewed in isolation.

Wyverald (approved):
looks good to me!

fmeum:
@comius This pattern is just great for binaries and tests with small scope, which is exactly why I'm pushing so hard to get https://github.com/bazelbuild/bazel/pull/16428 in without a flag: It would already be useful in today's WORKSPACE world and would be required with Bzlmod as the "hardcoded repo name + `$(location)`" approach breaks down there due to repo mapping.

It doesn't work well for binaries with more data deps and in particular doesn't work well with *libraries* that bring in data deps since the top-level binary would need to pass in all their paths explicitly, which gets infeasible pretty quickly.

fmeum:
> The drawback is, that if you go forward with repo_mappings and annotation, you probably cancel out this mechanism.
> WDYT?

This is not a problem: The `$(rlocationpath ...)` or `repo + $(location ...)` strings are still accepted by `Rlocation`. This is possible since canonical repo names aren't valid apparent repo names and vice versa. If a canonical repo name is passed in, it will simply not be repo mapped and used as is.

comius:
> This is not a problem: The `$(rlocationpath ...)` or `repo + $(location ...)` strings are still accepted by `Rlocation`. This is possible since canonical repo names aren't valid apparent repo names and vice versa. If a canonical repo name is passed in, it will simply not be repo mapped and used as is.

True.

What do you think about the other mechanism? Do you think we need two ways of doing it? If we're doing a complete cleanup it would be nice that there are no hardcoded paths in the sources and that they are passed from BUILD file.

comius:
> It doesn't work well for binaries with more data deps and in particular doesn't work well with _libraries_ that bring in data deps since the top-level binary would need to pass in all their paths explicitly, which gets infeasible pretty quickly.

Ah, thanks for the answer!

tools/java/runfiles/Runfiles.java
```diff
@@ -101,34 +177,105 @@ public static Runfiles create() throws IOException {
...
+   * know about this data-dependency.
+   *
+   * @param path             runfiles-root-relative path of the runfile
+   * @param sourceRepository the canonical name of the Bazel repository relative to which apparent
+   *                         repository names in runfiles paths should be resolved. Overrides the
+   *                         repository set when creating this {@link Runfiles} instance.
+   * @throws IllegalArgumentException if {@code path} fails validation, for example if it's null or
+   *                                  empty, or not normalized (contains "./", "../", or "//")
+   */
+  public final String rlocation(String path, String sourceRepository) {
```
comius, line 278: What is the use case for rlocation with sourceRepostory function? If there isn't one please remove.

Canonical names seem to be hard enough to obtain, and users should not care about anything else than their current repo.

There's functionality missing to obtain other repo names. Let's not introduce it.

tools/java/runfiles/Runfiles.java
```diff
@@ -183,20 +340,58 @@ private static String getRunfilesDir(Map<String, String> env) throws IOException
...
     }
 
+    private ManifestBased(ManifestBased existing, String sourceRepository) {
+      super(existing, sourceRepository);
+      this.manifestPath = existing.manifestPath;
+      this.runfiles = existing.runfiles;
+    }
+
+    @Override
+    public Runfiles withSourceRepository(String sourceRepository) {
```
comius, line 391: What's the use case? Please remove.

This is an optimisation, that I doubt anybody will have a need to use. As such it just adds maintenance burden.

tools/java/runfiles/Runfiles.java
```diff
@@ -225,6 +420,18 @@ private static String findRunfilesDir(String manifest) {
       return "";
     }
 
+    private static String findRepositoryMapping(String manifestPath) {
+      if (manifestPath != null && (manifestPath.endsWith(".runfiles/MANIFEST")
+          || manifestPath.endsWith(".runfiles\\MANIFEST")
+          || manifestPath.endsWith(".runfiles_manifest"))) {
+        String path = manifestPath.substring(0, manifestPath.length() - 18) + ".repo_mapping";
```
comius, line 427: Just a comment, nothing to do:`- 18` is "clever". Possibly causing problems in the future. But I see same thing is used elsewhere.

comius (changes requested):

tools/java/runfiles/Runfiles.java
```diff
@@ -67,21 +80,84 @@
...
+  protected Runfiles(Runfiles runfiles, String sourceRepository) {
+    this.repoMapping = runfiles.repoMapping;
+    this.sourceCanonical = sourceRepository;
+  }
 
   /**
    * Returns a new {@link Runfiles} instance.
    *
+   * <p><strong>Deprecated: With {@code --enable_bzlmod}, this function can only resolve runfiles
+   * correctly if called from the main repository. Use {@link #create(String)} instead. </strong>
```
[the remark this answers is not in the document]
fmeum: This should always work when used from the main repository, whether with or without Bzlmod. The canonical name of the main repository is the empty string in both cases, so runfiles lookups can succeed from there without any additional information.

Do you see a situation in which this would break? I hope to be able maintain backwards compatibility with the majority of runfiles lookups this way as most of them typically happen from the main repository.

tools/java/runfiles/Runfiles.java
fmeum: In larger applications, `Runfiles` instances are typically created at the `lang_binary` level and then passed down via some kind of dependency injection pattern. This is crucial since finding the runfiles directory or manifest may require custom logic.

Then, all downstream consumers that need to access runfiles just work with an existing Runfiles instance. However, these downstream consumers may very well live in different repositories and would thus need to specify a custom `sourceRepository` to find *their* runfiles. This also came up in https://github.com/bazelbuild/proposals/pull/269#discussion_r927728158.

tools/java/runfiles/Runfiles.java
fmeum: This is meant to preserve the common pattern of injecting a `Runfiles` instance at the top and feeding it to downstream consumers with Bzlmod. By using this function, users only have to make a change at one central location per repository (where they need to pass in their repository name obtained via the annotation), rather than at every `Rlocation` invocation outside the main repository.

In particular, this makes it possible to fully adopt Bzlmod *without* all of your transitive runfiles-using dependencies having done so. I thus consider it a relatively important part of the migration story.

Do you see a way I could address some of your concerns while keeping this method around?

tools/java/runfiles/Runfiles.java
```diff
@@ -67,21 +80,84 @@
...
+    }
+  }
+
+  private final Map<RepoMappingKey, String> repoMapping;
+  private final String sourceCanonical;
+
   // Package-private constructor, so only package-private classes may extend it.
-  private Runfiles() {}
+  private Runfiles(String repoMappingPath, String sourceRepository) throws IOException {
+    this.repoMapping = loadRepositoryMapping(repoMappingPath);
```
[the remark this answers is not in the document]
fmeum: See my other comments, this info may be needed and used by transitive dependencies that live in other repositories.

tools/java/runfiles/Runfiles.java
```diff
@@ -39,17 +42,27 @@
...
+ *   public class MyClass {
+ *     ...
+ * </pre>
+ *
+ * <p>4. Create a Runfiles object and use rlocation to look up runfile paths:
  *
  * <pre>
  *   public void myFunction() {
- *     Runfiles runfiles = Runfiles.create();
+ *     Runfiles runfiles = Runfiles.create(AutoBazelRepository_MyClass.BAZEL_REPOSITORY);
```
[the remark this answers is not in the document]
fmeum: I will add comments explaining this additional approach.

tools/java/runfiles/Runfiles.java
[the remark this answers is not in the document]
fmeum: Sorry, I didn't understand your last sentence at first. I will clarify that this pattern only works with Bzlmod as long as the repository is part of the root module.

tools/java/runfiles/Runfiles.java
comius: That's a valid use case. 

It's quite error prone though to be passing Runfiles with `sourceRepository` set as you might cross repo boundary unaware. It would make sense to design and structure classes in a way to prevent this.

tools/java/runfiles/Runfiles.java
comius: Ack. Let's keep conversation about this in a single thread. Resolving this one.

tools/java/runfiles/Runfiles.java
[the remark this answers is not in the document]
comius: Ack. Resolving.

tools/java/runfiles/Runfiles.java
fmeum: If we started out with a clean slate, I would probably go for an API where `Create` doesn't take `sourceRepository` and returns a `Runfiles` that doesn't have any `rlocation` functions. Instead, users would need to call `withSourceRepository` on it to get an object of a *distinct* `RunfilesWithRepoContext` type that does have an `rlocation` method. Users would then be encouraged to only ever pass `Runfiles` into other repos, with no repo context. I don't see how to implement this in a backwards compatible way though - if you have any ideas, please let me know!

But it's ultimately impossible to prevent misuse even with these hypothetical changes. After all, if we could programatically determine repo boundaries, we wouldn't have the user pass in `sourceRepository` manually but just do it for them (which we can in some languages, e.g. Python, Bash and Go).

What I am thinking about as a potential followup is a debug mode, similar to what the Bash runfiles library offers: If a certain environment variable is set, runfiles lookups would fail with a descriptive exception if a repo name isn't found in the repo mapping but also doesn't look like a canonical repository name (which, with Bzlmod, are always either `_main` or contain a tilde, but of course that's an implementation detail we could only depend on from a runfiles library shipped with Bazel).

What do you think? Do you want me to make further changes here?

comius: You could add `static createPreloaded()` that returns a `PreloadedRunfiles` without `rlocation` and with `Runfiles withSourceRepository(String)` methods. Usability is not much different, i.e. `Runfiles runfiles = Runfiles.createPreloaded().withSourceRepository(AutoBazelRepository_MyClass.BAZEL_REPOSITORY);`

fmeum: @comius I pushed a new commit that implements this pattern. It is not complete yet (I haven't updated the docs).

If there are any changes to the user-facing API you would like me to make, could you let me know today? I have to finalize my BazelCon talk before Monday and would like to present this API in it.

fmeum: I also added `.unmapped()` in addition to `.withSourceRepository(String)` for the case where someone uses `rlocationpath` exclusively. Having them pass in a non-sense repo name would make the API confusing.

comius: > If there are any changes to the user-facing API you would like me to make, could you let me know today?

No, nothing out of the lines we're already talking about.

fmeum: @comius What do you think of renaming `createPreloaded` to just `preload`? That may differentiate this method more clearly from `create`, which has very different semantics.

comius: SGTM

comius: Stardoc shows some need to provide a public method that returns `RepositoryContext`. That would allow it to store labels in canonical form, using `Label.parseWithRepoContext` method.

Slightly related is, that it seems it's also possible to implement a method similar `rlocation` that takes an absolute label (but with non-canonical repos) as an input.

comius: `RepositoryMapping` and `Label` might be a dependency you don't want to bring into `Runfiles`. What also works for Stardoc is a more minimalistic `getCannonicalName(targetRepoApparentName)` or similar. This way you can hop around different repos by calling  `withSourceRepository(getCannonicalName(repoInLoad))`.

BTW one question. I was playing around with stardoc and I didn't get repo_mappings file in the exec root / sandbox. Is it possible it doesn't work yet or am I missing some flags?

fmeum: > `RepositoryMapping` and `Label` might be a dependency you don't want to bring into `Runfiles`. What also works for Stardoc is a more minimalistic `getCannonicalName(targetRepoApparentName)` or similar. This way you can hop around different repos by calling `withSourceRepository(getCannonicalName(repoInLoad))`.

I have thought about introducing such a function, but I don't know what it would be useful for other than implementing Stardoc. But that may be good enough of a reason to add it - after all, it would just have to be extracted out of this PR.

> BTW one question. I was playing around with stardoc and I didn't get repo_mappings file in the exec root / sandbox. Is it possible it doesn't work yet or am I missing some flags?

It's not staged properly yet, which I intend to fix with https://github.com/bazelbuild/bazel/pull/16652. When it's working, I will also add a small integration test to this PR.

fmeum:
@comius I updated the documentation and made this work in sandboxes by looking up the repository mapping under its root symlink `_repo_mapping` introduced by #16652. Could you review the last three commits? The commits I rebased onto are all part of other PRs.

comius:
> @comius I updated the documentation and made this work in sandboxes by looking up the repository mapping under its root symlink `_repo_mapping` introduced by #16652. Could you review the last three commits? The commits I rebased onto are all part of other PRs.

I did a quick review and the last three commits all look reasonable.

One more thing you should consider is making preload() method a lazy initializer / preloaded runfiles a singleton. Singleton pattern is usually avoided at all costs, however it looks at the moment Bazel's design forces you into a single global instance. 

Making it a singleton might prevent involuntary duplication by the user. Libraries using runfiles might be coming from a different repository, that could follow a different philosophy and there might be no way of passing the Runfiles object down to those libraries.

fmeum:
@comius I can make `preload()` return a singleton but keep `preload(Map)` around in case users rely on custom runfiles locations (e.g. when using manifests at install time). What do you think of that?

Should I use `WeakReference` for the `preload()` singleton? Not being able to free the manifests memory may otherwise be wasteful.

comius:
> @comius I can make `preload()` return a singleton but keep `preload(Map)` around in case users rely on custom runfiles locations (e.g. when using manifests at install time). What do you think of that?

Sounds ok.

> Should I use `WeakReference` or `SoftReference` for the `preload()` singleton? Not being able to free the manifests memory may otherwise be wasteful.

I think with or without it should be fine. Without it the could would be simpler and with it you can arguably save some memory in some cases. I can't say there's a strong case for one or the other decision.

fmeum:
@comius I went with a softly cached singleton instance in a new commit.

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 6.0.0

fmeum:
Rebased on master, but remains stacked on #16534.

fmeum:
Rebased and no longer stacked.

Issue #16551, opened 2022-10-26
Enable setting C/C++ standards as a toolchain feature


jungleraptor:
### Description of the feature request:

Proposal to add a handful of new "features" to the default c++ toolchain that enables setting C/C++ standards (names tentative):
- default_cxx_standard (respects `BAZEL_CXXOPTS`, default `-std=c++0x`)
- cxx14_standard
- cxx17_standard
- etc...

I've got an initial implementation of this capability that currently works on linux & macos, and does not break backwards compatibility:
https://github.com/bazelbuild/bazel/pull/16552

I'd like to get some feedback from the Bazel team on if it's possible to get this merged before doing whatever else is necessary to meet the contribution requirements (design docs, testing, etc...).

### What underlying problem are you trying to solve with this feature?

The original issue I was trying to solve was to get some code to build with bazel that mixes C/C++ code in the same target:
```starlark
cc_library(
    name = "example",
    srcs = [
        "src/foo.cpp", # requires compiling with C++14
        "src/bar.c"
    ],
    copts = [
        "-Werror",
        "-std=c++14",
    ]
)
```

It's currently not possible to build that target with Bazel. This is because there's no way to distinguish between C and C++ flags at the `cc_*` rules level. The compiler will issue a warning when `-std=c++14` is passed while compiling the C file, and -Werror will cause that warning to be propagated as a build error.

There are various alternatives but none of them are desirable:
- Using BAZEL_CXXOPTS: Not able to control the standard at the target level, no equivalent for setting the C standard
- Using --cxxopts/--copts at the command line: Also doesn't allow setting at the target level.
- Splitting the code up into mutliple targets (one for the C portions, and one for the C++ portions). The build system should be flexible/powerful enough for developers to decide how they would like to structure their build/targets.
- Writing a custom toolchain that implements the desired feature - Writing a custom toolchain is a big commitment with many other implications.

I think that the most elegant solution to this problem is to just expose a new set of features in the default toolchain that allow setting the standard. The features API enables distinguishing between C/C++ actions and avoids the scope issues above. Every target can mix and match C/C++ and decide the standard for each.

The POC branch I've linked shows that it should be possible to do this without breaking backwards compatibility of existing behavior, assuming that the order of flags is not part of the API.

### Which operating system are you running Bazel on?

Windows, Linux, and Mac

### What is the output of `bazel info release`?

5.3.2

### Have you found anything relevant by searching the web?

Lots of research, still prefer my suggestion above.

[...]

PR #16552, opened 2022-10-26
Enable setting C/C++ standards as a toolchain feature

jungleraptor:
Adds a handful of new features to the default toolchain that enables setting the C/C++ standards, with or without gnu extensions (`-std=gnu`) at the target level.

The issue associated with the PR is linked, and goes into further detail about the motivation for introducing this feature.

https://github.com/bazelbuild/bazel/issues/16551

~~This is an initial implementation. I'd like to get feedback on whether it's possible to accept this change before I add tests/documentation etc..~~

# Usage

```sh
# build with -std=c++17
bazel build --features=c++17

# build with -std=gnu++17
bazel build --features=c++17 --features=gnu_extensions
```

# Testing

Testing was done using this project: https://github.com/isaactorz/bazel-features-test

Let me know if there's an appropriate place in this repo to add these tests.

Testing done w/ 
- gcc-11, gcc-12, and clang-14 on ubuntu 22.04
- apple clang on intel macos (note that I add to bump the `-mmacos-version-min` flag to get the c++17 tests to build as the default was to low to use `optional` api.)

# TODO

Assuming this can be merged I'd think the next steps would be:
- [x] Add equivalent features for setting C standard
- [x] Add tests
- [x] Finalize the names of the feature
- [ ] Mingw support
- [x] Testing with msvc


tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -144,6 +149,7 @@ lto_index_actions = [
 ]
 
 def _impl(ctx):
+    print("I'm HERE")
```
keith, line 152: debug print made it in

tools/cpp/unix_cc_configure.bzl
```diff
@@ -528,6 +521,12 @@ def configure_unix_toolchain(repository_ctx, cpu_value, overriden_tools):
                 cpu_value,
                 False,
             )),
+            "%{cxx_standard}": escape_string(get_env_var(
+                repository_ctx,
+                "BAZEL_CXXOPTS",
```
keith, line 526: I don't think this is a safe change as previously this env var could have contained much more than just this. I think to keep compatibility you'll have to move this back up to where it was, just default it to nothing, and hardcode the C++ default here (although I guess it also could have affected the CXX include directories output? I'm not sure if that's good or bad)

isaactorz: Ah, of course 🤦🏼 .

So since that's the case I've left everything in `unix_cc_configure.bzl` alone and only added the new features that set the standard.

If the features are activated then they take precedent over the `-std=c++0x` opt defined by `BAZEL_CXX_OPTS` (since they come later on the cmd line).

If one really doesn't want duplicate options then I guess they could set `BAZEL_CXXOPTS=""`.

Does that sound better? Certainly simpler implementation wise.

keith (approved):

tools/cpp/unix_cc_configure.bzl
keith: yea sg

monika566:
If one really doesn't want duplicate options then I guess they could set BAZEL_CXXOPTS="".

buildbreaker2021:
Hey, is this ready for a review?

isaactorz:
> Hey, is this ready for a review?


Whoops. Sorry for letting this go stale.

See the issue I linked for the background on why I initially proposed this. Internally we ended up just settling on splitting up building c and c++ code because we also need to set things like `-fno-exceptions` and `-fno-rtti`. If you're curious on our struggles see: https://github.com/swift-nav/rules_swiftnav/pull/52

However I still think this is a good and useful change for Bazel in general. I'll rebase and test this weekend and get it ready for review.

Are there any tests I should be adding for this stuff someone can point me to?

Wyverald:
You may want to see https://docs.google.com/document/d/132vEDnQZY0PtF9ko6HoLQ2dqk6meZnSrUfc1gsKuBkk/edit?usp=share_link

@fmeum

isaactorz:
> Hi @isaactorz, Lets us know once it is ready for review. Thanks!

Will do. Been very busy lately. Only thing left I want to do is add support for this feature to the mingw toolchain.

junyer:
RTYI, @jsharpe. :)

Mizux:
Why not adding C++ **23** and C++ **26** ? they have already landed in CMake and it will take months (years?) to have bazel 7.0 widely available so....

```
23
New in version 3.20.
C++23

26
New in version 3.25. 
C++26. CMake 3.25 and later recognize 26 as a valid value, no version has support for any compiler.
```
ref: https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

ps:
Nov 16, 2022: v3.25.0
Mars 23, 2021: v3.20.0
ref: https://github.com/Kitware/CMake/releases

isaactorz:
> Why not adding C++ **23** and C++ **26** ? they have already landed in CMake and it will take months (years?) to have bazel 7.0 widely available so....
> 
> ```
> 23
> New in version 3.20.
> C++23
> 
> 26
> New in version 3.25. 
> C++26. CMake 3.25 and later recognize 26 as a valid value, no version has support for any compiler.
> ```
> 
> ref: https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html
> 
> Also you have C **23**
> 
> ```
> 23
> New in version 3.21.
> C23
> ```
> 
> see: https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html
> 
> ps: Nov 16, 2022: v3.25.0 Mars 23, 2021: v3.20.0 ref: https://github.com/Kitware/CMake/releases

I was going for a consistent set of features to expose across the "big 3" (gcc/clang/msvc)

with gcc and clang this would probably be fine since they have explicit codenames for the newer standards (i.e. `std=c++2b`, `std=c++2c`, etc..), but what to do for msvc? What `/std:c++latest` means now may not be what it means in the future. I guess the simple option would be to just not enable support for those features for the msvc toolchain.

Side note: maybe we even should consider using `-std=c++2a` instead of `-std=c++20` to support a wider range of clang/gcc

Mizux:
> with gcc and clang this would probably be fine since they have explicit codenames for the newer standards (i.e. `std=c++2b`, `std=c++2c`, etc..), but what to do for msvc? What `/std:c++latest` means now may not be what it means in the future. I guess the simple option would be to just not enable support for those features for the msvc toolchain.
> 
> Side note: maybe we even should consider using `-std=c++2a` instead of `-std=c++20` to support a wider range of clang/gcc

FYI CMake adapt what it pass to the command line according to the compiler version...
https://gitlab.kitware.com/cmake/cmake/-/blob/0183956d30283212bc66cde17d9756f18bc5db27/Modules/Compiler/Clang.cmake

peakschris:
Hi @jungleraptor what is the status of this, it is a sorely missed feature!

jungleraptor:
> Hi @jungleraptor what is the status of this, it is a sorely missed feature!

Hello!

At work we ended up forking `unix_cc_toolchain_config.bzl` and implementing this ourselves so this fell off my radar, but it is annoying that it doesn't work with the autoconfigured toolchain.

@sgowroji @buildbreaker2021 Is this something we could land in mostly it's current shape ?

jungleraptor:
@buildbreaker2021 @sgowroji @Mizux friendly ping

Recently came across another case in `rules_rust` that's currently having to work around this issue:
https://github.com/bazelbuild/rules_rust/blob/bad49c4a1139afdb257fed0af6bd1dc20a6c21a1/bindgen/3rdparty/patches/llvm-project.cxx17.patch#L403

Also tagging @keith since you did the initial review.

benjamin-branchware:
Is there really no way of having the actual C/CXX standard specified in a string (which may be parsed in some .bzl function) instead of having individual boolean features? We just about get away with it here with the number of standard revisions, but for other controls/flags which may hold arbitrary N values, this won't scale.

Summoning @fmeum on this point because it's an architectural question.

fmeum:
> Is there really no way of having the actual C/CXX standard specified in a string (which may be parsed in some .bzl function) instead of having individual boolean features?

This could probably be realized as an API, but in the end that would still need to be just syntactic sugar on top of boolean features: If you want to match on the availability of a certain feature in `c++11` with a `select` or dependent `feature`, you would need e.g. `c++17` to imply `c++11` so that the check succeeds. There is no way to check on a range of features or anything like that.

peakschris:
Not stale


________________________________

bjm9991:
Not stale - but long since moved away from Bazel myself....

peakschris:
Not stale

________________________________

peakschris:
Not stale


________________________________

keith:
this file has moved to rules_cc, so idk if there is a path forward here or not, but if there is, it has to be reapplied over there

PR #16556, opened 2022-10-26
Make `bazel coverage` work with minimal mode

coeuvre:
This PR solves the problem in a different way that #16475 tries to solve:

1. Use `ActionFileSystem` to read directory output for coverage data.
2. Fire event `CoverageReport` in the action after the coverage report is generated and listen to it in `ToplevelArtifactsDownloader` to download the report.


src/main/java/com/google/devtools/build/lib/remote/RemoteModule.java
```diff
@@ -929,7 +929,19 @@ public void executorInit(CommandEnvironment env, BuildRequest request, ExecutorB
...
-      if (remoteOutputsMode.downloadToplevelOutputsOnly()) {
+      boolean useToplevelDownloader = remoteOutputsMode.downloadToplevelOutputsOnly();
+
+      // Download toplevel artifacts for `run` command.
+      if ("run".equals(env.getCommandName())) {
+        useToplevelDownloader = true;
+      }
+
+      // Download toplevel artifacts for `coverage` command.
+      if ("coverage".equals(env.getCommandName())) {
```
fmeum, line 940: @coeuvre This looks great, thanks for picking it up! 

One thing I noticed: Could this make `bazel coverage` download not only the coverage reports, but also the test binaries itself (see https://github.com/bazelbuild/bazel/commit/b6f3111d8e460de35a5e5570bd784919fc30c5f8#diff-1f3936a41b59af9df49bf16acbe44524fbd503e8ed7cd76b62d5164867e0032aR160)? If that's the case, it's something I may figure out by myself and could submit as a follow-up.

coeuvre: I had this in mind but didn't make any change to that line yet because I want to have the discussion in that PR and finalize how do we handle different commands. This is also the reason why this PR is still in draft.

coeuvre: Now the test binaries are not downloaded with `bazel coverage`

brentleyjones:
@coeuvre @tjgq Do you think we can get this in, and possibly into 6.0?

brentleyjones:
@coeuvre @tjgq friendly ping

brentleyjones:
@coeuvre Can this get cherry-picked into 6.1.0?

coeuvre:
Yes, I believe this is a safe change.

coeuvre:
@bazel-io fork 6.1.0

PR #16560, opened 2022-10-26
Escape tilde more gracefully

comius:
Tilde needs to be escaped only when it's the first character after the white space. Otherwise, we can keep the string unescaped.

This supports bzlmod better, because tilde is used in the directory names.

Users often already escape location function, for example `SJ="$(location @bazel_tools//tools/jdk:singlejar)"; $SJ`. Without the change this becomes `'external/repo~name/singlejar'` and the script fails (no file found).

Location function shouldn't be escaped. But without this change we risk a lot of users will need to fix their scripts.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/util/ShellEscaperTest.java
+++ b/src/test/java/com/google/devtools/build/lib/util/ShellEscaperTest.java
@@ -41,6 +41,8 @@ public class ShellEscaperTest {
     assertThat(escapeString("\\'foo\\'")).isEqualTo("'\\'\\''foo\\'\\'''");
     assertThat(escapeString("${filename%.c}.o")).isEqualTo("'${filename%.c}.o'");
     assertThat(escapeString("<html!>")).isEqualTo("'<html!>'");
+    assertThat(escapeString("~not_home")).isEqualTo("'~not_home'");
+    assertThat(escapeString("external/protobuf~3.19.6/src/google")).isEqualTo("external/protobuf~3.19.6/src/google");
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/devtools/build/lib/util/ShellEscaperTest.java
```diff
@@ -41,6 +41,8 @@ public void shellEscape() throws Exception {
     assertThat(escapeString("\\'foo\\'")).isEqualTo("'\\'\\''foo\\'\\'''");
     assertThat(escapeString("${filename%.c}.o")).isEqualTo("'${filename%.c}.o'");
     assertThat(escapeString("<html!>")).isEqualTo("'<html!>'");
+    assertThat(escapeString("~not_home")).isEqualTo("'~not_home'");
+    assertThat(escapeString("external/protobuf~3.19.6/src/google")).isEqualTo("external/protobuf~3.19.6/src/google");
```
fmeum, line 45: Just to be safe, could you add a test case for `external/~install_dev_dependencies~foo/pkg`? This is a path that Bzlmod can generate and it's slightly more special as the tilde is the first character of a path segment.

fmeum (approved):
Thanks for catching this!

fmeum:
@bazel-io flag

src/test/java/com/google/devtools/build/lib/util/ShellEscaperTest.java
fmeum: Luckily, I don't think that Bazel will generate paths where the first segment starts with a tilde in any situation. At the moment that is, I think that this may change with `--experimental_sibling_repository_layout`, where paths can start with repository names. These paths would not work without escaping.

@Wyverald Do you see other cases where this could be problematic?

Wyverald: Bash seems to consider paths starting with '~' special: https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html

Wyverald (approved):

fmeum:
@bazel-io flag

src/test/java/com/google/devtools/build/lib/util/ShellEscaperTest.java
fmeum: https://github.com/bazelbuild/bazel/pull/16564 should ensure that paths emitted by Bazel never start with `~`.

PR #16563, opened 2022-10-26
Improve error on invalid `-//foo` and `-@repo//foo` options

fmeum:
`-//foo` and `-@repo//foo` are always invalid Bazel options, but are usually meant to be either negative target patterns (which have to come after the `--` separator) or Starlark options (which always start with `--`).

With this commit, the error shown to the user mentions these two situations and how to fix the issue in each case.

### DIFF
--- a/src/main/java/com/google/devtools/common/options/OptionsParserImpl.java
+++ b/src/main/java/com/google/devtools/common/options/OptionsParserImpl.java
@@ -476,6 +476,16 @@ class OptionsParserImpl {
         continue; // not an option arg
       }
 
+      if (arg.startsWith("-//") || arg.startsWith("-@")) {
+        // Fail with a helpful error when an invalid option looks like an absolute negative target
+        // pattern or a typoed Starlark option.
+        throw new OptionsParsingException(String.format("Invalid options syntax: %s\n"
+                + "Note: Negative target patterns can only appear after the end of options marker"
+                + " ('--'). Flags corresponding to Starlark-defined build settings always start"
+                + " with '--', not '-'.",
+            arg));
+      }
+
       arg = swapShorthandAlias(arg);
 
       if (arg.equals("--")) { // "--" means all remaining args aren't options
--- a/src/test/java/com/google/devtools/common/options/OptionsParserTest.java
+++ b/src/test/java/com/google/devtools/common/options/OptionsParserTest.java
@@ -2369,6 +2369,30 @@ public final class OptionsParserTest {
         .containsExactly("--second=hello");
   }
 
+  @Test
+  public void negativeTargetPatternsInOptions_failsDistinctively() {
+    OptionsParser parser = OptionsParser.builder().optionsClasses(ExampleFoo.class).build();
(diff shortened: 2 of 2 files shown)


fmeum:
@aranguyen Could you review this PR?

michajlo:
Is there any more discussion around this and why putting the patterns after `--` isn't sufficient?

fmeum:
> Is there any more discussion around this and why putting the patterns after `--` isn't sufficient?

It's sufficient if you know about it, but the error message you get if you don't doesn't give any indication of this being how to fix your command line. Instead of catching this case and emitting a more helpful error messages, it seemed simpler to just make the most common, unambiguous case work out of the box.

I don't have more context than https://github.com/CodeIntelligenceTesting/cifuzz/pull/383#discussion_r1005694516 and similar past experiences that myself or coworkers had with this limitation.

Given his reaction to the PR, maybe @brentleyjones can provide a separate opinion.

brentleyjones:
I've just run into the same situation myself, and agree that it's hard to know when you need `--`, and if we can make this work without it :+1:.

michajlo:
IIUC this is a quite common convention for CLI tools, and while it is confusing I fear changing things will only move confusion around, not reduce it. The one example that I believe is cited in the code comments is that relative labels (which I personally use more often) behave differently. It's also liable to cause confusion around starlark-defined flags, which can look like target patterns but are indeed flags.

My interest here is keeping command-line parsing semantics as tight as possible so that any attempted automated introspection of command lines doesn't need to be too smart.

fmeum:
I agree that not supporting all negative target patterns has potential to create confusion in different places and that supporting any kind of new syntax will make command line parsing more difficult.

Starlark flags shouldn't add to the confusion though as they always starts with two dashes whereas negative target patterns start with one - isn't that a clean separation?

Given that, would you support a version of this PR that doesn't add support for absolute negative target patterns without a `--`, but instead fails with a more descriptive error?

michajlo:
> Starlark flags shouldn't add to the confusion though as they always starts with two dashes whereas negative target patterns start with one - isn't that a clean separation?

I'm thinking of how one might be tricked into thinking `-//foo` is a valid syntax because the parser accepted it, or typo-ing `-//foo` and not noticing because the parser accepted it. 

> Given that, would you support a version of this PR that doesn't add support for absolute negative target patterns without a --, but instead fails with a more descriptive error?

As long as it can be done tastefully sure. Thought it may not be me stamping it, I'm mainly here because I have an interest in parsing semantics.

aranguyen:
@fmeum Thank you very much for sending the PR. I just had the sync with the team and brought up your PR. We agree with https://github.com/bazelbuild/bazel/pull/16563#issuecomment-1292536637 and will be supportive of reviewing another PR that shows a more descriptive error.

fmeum:
@aranguyen I updated the commit and the PR description, please take another look.

aranguyen (approved):
LGTM. Thanks!

iancha1992:
@bazel-io fork 6.3.0

PR #16569, opened 2022-10-26
Relnotes python script

Wyverald:
The old shell script no longer works in the LTS setup. This new script is written in Python so hopefully more maintainable, and reflects what we need to do in the post-LTS world.

I'll look into hooking this up into the release pipeline later.

### DIFF
--- a/scripts/release/relnotes.py
+++ b/scripts/release/relnotes.py
@@ -16,48 +16,11 @@ import re
 import requests
 import subprocess
 
-GCS_BASE_URL = "https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/"
 
-
-def _paginated_get(base_url):
-  """GETs a URL from GCS, respecting the pagination token if any."""
-  next_page_token = None
-  while True:
-    url = base_url
-    if next_page_token is not None:
-      url += f"&pageToken={next_page_token}"
-    r = requests.get(url)
-    r.raise_for_status()
-    json = r.json()
-    yield json
-    if "nextPageToken" in json:
-      next_page_token = json["nextPageToken"]
-    else:
-      break
-
-
-def _version_sort_key(v):
-  # Converts "5.3.2" into [5, 3, 2] so that we get a natural lexicographical
-  # sort. Note that we have a version "0.2.2b" so we just convert any non-digit
-  # segment into 9999.
-  return [int(s) if s.isdigit() else 9999 for s in v.split(".")]
-
-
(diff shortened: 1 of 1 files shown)


scripts/release/relnotes.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _version_sort_key(v):
+  # Converts "5.3.2" into [5, 3, 2] so that we get a natural lexicographical
+  # sort. Note that we have a version "0.2.2b" so we just convert any non-digit
+  # segment into 9999.
+  return [int(s) if s.isdigit() else 9999 for s in v.split(".")]
+
+
+def get_latest_release():
+  """Discovers the latest stable release name from GCS."""
```
meteorcloudy, line 47: If you want to simplify this, you can use 
```
#!/bin/env python3

from urllib import response
import requests
response = requests.get('https://github.com/bazelbuild/bazel/releases/latest')
print(response.url)
print(response.url.split("/")[-1])
```
output is
```
https://github.com/bazelbuild/bazel/releases/tag/5.3.2
5.3.2
```

meteorcloudy: We are using similar approach at https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/setup-windows.ps1#L185-L191

scripts/release/relnotes.py
```diff
@@ -0,0 +1,139 @@
...
+    if relnote is not None:
+      relnotes.append(relnote)
+  return relnotes
+
+
+if __name__ == "__main__":
+  # Get the last stable release.
+  last_release = get_latest_release()
+  print("last_release is", last_release)
+  last_release_branch = "origin/release-" + last_release
```
meteorcloudy, line 118: This assume `origin/release-<version>` is already fetched, but might not be true.
Also maybe it's slightly better to calculate the base commit from the tag instead a the branch?
We can do
```
git fetch origin 5.3.2
git merge-base FETCH_HEAD HEAD
```

scripts/release/relnotes.py
Wyverald: wow, that's so much simpler :) done.

scripts/release/relnotes.py
Wyverald: changed to fetch the tag first

meteorcloudy (approved):

Wyverald:
@bazel-io fork 6.0.0

PR #16578, opened 2022-10-27
Relax `Label` repo visibility validation

fmeum:
Failing the build eagerly when `Label` is passed a label string referencing an unknown apparent repository name turned out to be too strict and causes failures on invalid, but unused labels: https://buildkite.com/bazel/bazel-bazel-with-bzlmod/builds/667#01841861-3d3a-4b65-91b7-b7083ee6b42b

Instead, fail when a method that requires a valid repository name is called on an invalid `Label` instance.


fmeum:
@meteorcloudy Could you check that this fixes the Bazel build?

Wyverald (approved):

fmeum:
@bazel-io flag

meteorcloudy:
Yes, I tested this fixed worked!

Wyverald:
@bazel-io fork 6.0.0

PR #16601, opened 2022-10-28
[remote/downloader] Migrate `Downloader` to take `Credentials`

Yannic:
Progress on https://github.com/bazelbuild/bazel/issues/15856

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/authandtls/staticcredentials/StaticCredentials.java
+++ b/src/main/java/com/google/devtools/build/lib/authandtls/staticcredentials/StaticCredentials.java
@@ -1,0 +1,49 @@
+package com.google.devtools.build.lib.authandtls.staticcredentials;
+
+import com.google.auth.Credentials;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+
+/** Implementation of {@link Credentials} which provides a static set of credentials. */
+public final class StaticCredentials extends Credentials {
+  public static final StaticCredentials EMPTY = new StaticCredentials(ImmutableMap.of());
+
+  private final ImmutableMap<URI, Map<String, List<String>>> credentials;
+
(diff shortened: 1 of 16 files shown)


src/main/java/com/google/devtools/build/lib/authandtls/staticcredentials/StaticCredentials.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public StaticCredentials(Map<URI, Map<String, List<String>>> credentials) {
+    Preconditions.checkNotNull(credentials);
+
+    this.credentials = ImmutableMap.copyOf(credentials);
+  }
+
+  public Map<URI, Map<String, List<String>>> getMapForMigration() {
+    return credentials;
+  }
```
tjgq, line 23: I couldn't find any callers for this method in this PR. Will it be used in a followup? If so, consider adding it in the same PR where it gets used, so we can later trace back why it was added.

tjgq (changes requested):

src/main/java/com/google/devtools/build/lib/authandtls/staticcredentials/StaticCredentials.java
Yannic: Eh, yeah. I used this when creating this PR. Removing

src/test/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloaderTest.java
```diff
@@ -370,7 +371,7 @@ public void downloadOneUrl_ok() throws IOException, InterruptedException {
       httpDownloader.download(
           Collections.singletonList(
               new URL(String.format("http://localhost:%d/foo", server.getLocalPort()))),
-          Collections.emptyMap(),
+          new StaticCredentials(ImmutableMap.of()),
```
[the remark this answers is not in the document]
Yannic: done

Yannic:
PTAL

tjgq (approved):
Importing.

I think I'd prefer to put StaticCredentials.java in `authandtls` instead of creating another subpackage (unless you give me a strong reason not to). Thoughts?

PR #16602, opened 2022-10-28
Include source files with `cquery --output=files`

fmeum:
RELNOTES[INC]: `cquery --output=files` also outputs source files.

### DIFF
--- a/src/test/java/com/google/devtools/build/lib/query2/cquery/FilesOutputFormatterCallbackTest.java
+++ b/src/test/java/com/google/devtools/build/lib/query2/cquery/FilesOutputFormatterCallbackTest.java
@@ -106,8 +106,7 @@ public class FilesOutputFormatterCallbackTest extends ConfiguredTargetQueryTest 
     this.reporter = new Reporter(new EventBus(), events::add);
   }
 
-  private List<String> getOutput(String queryExpression, List<String> outputGroups,
-      boolean includeSourceFiles)
+  private List<String> getOutput(String queryExpression, List<String> outputGroups)
       throws Exception {
     QueryExpression expression = QueryParser.parse(queryExpression, getDefaultFunctions());
     Set<String> targetPatternSet = new LinkedHashSet<>();
@@ -115,7 +114,6 @@ public class FilesOutputFormatterCallbackTest extends ConfiguredTargetQueryTest 
     PostAnalysisQueryEnvironment<KeyedConfiguredTarget> env =
         ((ConfiguredTargetQueryHelper) helper).getPostAnalysisQueryEnvironment(targetPatternSet);
 
-    options.includeSourceFiles = includeSourceFiles;
     ByteArrayOutputStream output = new ByteArrayOutputStream();
     FilesOutputFormatterCallback callback =
         new FilesOutputFormatterCallback(
@@ -136,14 +134,7 @@ public class FilesOutputFormatterCallbackTest extends ConfiguredTargetQueryTest 
 
   @Test
   public void basicQuery_defaultOutputGroup() throws Exception {
-    List<String> output = getOutput("//pkg:all", ImmutableList.of(), false);
-    assertContainsExactlyWithBinDirPrefix(
-        output, "pkg/main_default_file", "pkg/other_default_file");
-  }
-
-  @Test
-  public void basicQuery_defaultOutputGroup_includeSourceFiles() throws Exception {
-    List<String> output = getOutput("//pkg:all", ImmutableList.of(), true);
+    List<String> output = getOutput("//pkg:all", ImmutableList.of());
     var sourceAndGeneratedFiles = output.stream()
         .collect(Collectors.<String>partitioningBy(path -> path.startsWith("bazel-out/")));
     assertThat(sourceAndGeneratedFiles.get(false)).containsExactly("pkg/BUILD", "defs/rules.bzl");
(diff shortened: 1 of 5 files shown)


fmeum:
@gregestren Friendly ping :-)

gregestren:
Overall looks good.

Main feedback is: do we actually need the flag? Or can we justify universally including source files?

I'm asking because Bazel has a real issue with too many flags, and in general we want to reduce them unless really needed. Do we have a compelling use case to want to preserve access to the no-source variation? I realize this is technically an incompatible change but I don't think it's significant enough to worry about.

fmeum:
@gregestren I also prefer this being the default behavior. I was worried about backwards incompatibility, but since it should be rare for targets to advertise both source and non-source files, it's probably not a big deal to change.

Updated the PR and description.

gregestren (approved):

illicitonion (approved):
Perfect, thanks so much! 🎉

fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 6.0.0

fmeum:
@gregestren Friendly ping, is this in the process of being merged? Let me know if there are any blockers, it would be great to get this into Bazel 6.

gregestren:
@sgowroji and I have been trying but some merge conflicts popped up. I think it's just related to bazel vs. blaze naming. One sec...

gregestren:
See https://buildkite.com/bazel/google-bazel-presubmit/builds/61054#01848eae-8c2b-4791-b46e-e8297a6fdc3e.

src/test/java/com/google/devtools/build/lib/query2/cquery/FilesOutputFormatterCallbackTest.java
```diff
@@ -134,21 +135,31 @@ private List<String> getOutput(String queryExpression, List<String> outputGroups
   @Test
   public void basicQuery_defaultOutputGroup() throws Exception {
     List<String> output = getOutput("//pkg:all", ImmutableList.of());
-    assertContainsExactlyWithBinDirPrefix(
-        output, "pkg/main_default_file", "pkg/other_default_file");
+    var sourceAndGeneratedFiles = output.stream()
+        .collect(Collectors.<String>partitioningBy(path -> path.startsWith("bazel-out/")));
```
gregestren, line 139: Can we get the equivalent without hard-coding "bazel-out"? I'm not sure if this path is exposed elsewhere as some test constant. Otherwise we could find some output path / artifact and parse its segments. 

I'm pretty sure this literal is causing the conflicts.

fmeum: I went with a regex that should be good enough for this test.

fmeum:
> See https://buildkite.com/bazel/google-bazel-presubmit/builds/61054

I have never seen this pipeline before - in which way does it differ from the regular GitHub presubmit? It does seem to run the Bazel tests, but somehow behaves as if executed by Blaze?

src/test/java/com/google/devtools/build/lib/query2/cquery/FilesOutputFormatterCallbackTest.java
gregestren: Thanks. @sgowroji can we re-import the latest version of this PR?

gregestren:
> > See https://buildkite.com/bazel/google-bazel-presubmit/builds/61054
> 
> I have never seen this pipeline before - in which way does it differ from the regular GitHub presubmit? It does seem to run the Bazel tests, but somehow behaves as if executed by Blaze?

1. PRs produced here.
2. Regular GitHub presubmit vets them
3. On approval, @sgowroji copies them into internal PRs
4. That triggers both the remaining Google tests (internally) and https://buildkite.com/bazel/google-bazel-presubmit. It also creates a public Gerritt review for reference: https://bazel-review.googlesource.com/c/bazel/+/207117

These are mostly the same. Technically the internal PR could deviate from the GitHub PR, so the extra presubmits are just caution. In this case they caught an actual difference.

fmeum:
> > > See https://buildkite.com/bazel/google-bazel-presubmit/builds/61054
> > 
> > 
> > I have never seen this pipeline before - in which way does it differ from the regular GitHub presubmit? It does seem to run the Bazel tests, but somehow behaves as if executed by Blaze?
> 
> 1. PRs produced here.
> 2. Regular GitHub presubmit vets them
> 3. On approval, @sgowroji copies them into internal PRs
> 4. That triggers both the remaining Google tests (internally) and https://buildkite.com/bazel/google-bazel-presubmit. It also creates a public Gerritt review for reference: https://bazel-review.googlesource.com/c/bazel/+/207117
> 
> These are mostly the same. Technically the internal PR could deviate from the GitHub PR, so the extra presubmits are just caution. In this case they caught an actual difference.

I see. So the imported PR had `bazel` replaced with `blaze`, which then caused the test to fail? I at first thought it would be running the tests differently, but it's just that it runs against the imported PR?

gregestren:
Yes, that seems to be the case.

src/test/java/com/google/devtools/build/lib/query2/cquery/FilesOutputFormatterCallbackTest.java
sgowroji: Okay!

sgowroji:
Thank you @fmeum for the new changes. It was successfully imported.

PR #16606, opened 2022-10-28
[6.0.0] Remove logic that increases delay between progress updates over time

### LINKED ISSUE #16119: Progress updates become ever more scarce without terminal cursor control
### Description of the bug:

[The docs](https://bazel.build/docs/user-manual#--progress_report_intervaln) for the option `--progress_report_interval=n` read:
`This option sets the reporting frequency, progress will be printed every n seconds.`

But this seems to not work when Bazel can not update the terminal output in place. The delay between progress updates grows linearly as time increases.

I had a quick look at the code and according [to the comments](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/UiEventHandler.java#L858) this was implemented intentionally. [The commit message](https://github.com/bazelbuild/bazel/commit/fa2917522b99dd28aee757b41b0e087eb6cfcb1d ) also mentions this.

The behavior as described in the docs would be nice for us, as then we can automatically check whether Bazel is still running correctly for some very long-running actions. There also seems no real benefit to increasing the update interval to 10 minutes or more, as we have encountered.

Of course, if I'm missing why this works this way, it would still be useful if the docs could be updated with the behavior for non-overwritable output.

Run a test that just sleeps and redirect stderr to a file.

```
cat > sleep.sh << 'EOF'
#!/usr/bin/env bash
sleep 10000
EOF
chmod +x sleep.sh
cat > BUILD.bazel << 'EOF'
sh_test(
    name = "sleep_test",
    size = "enormous",
    srcs = ["sleep.sh"],
)
EOF
touch WORKSPACE
bazel test //:sleep_test --progress_report_interval=2 2> output.txt
```

Initially the progress updates will be written every 2 seconds as requested, but after 40 seconds the delay between them will start increasing noticeably.

Ubuntu 20.04

release 5.2.0

_No response_

_No response_

_No response_

_No response_

kshyanashree:
Removes a bit of undocumented behavior that increases the delay between progress updates the longer the build is running. Potentially resolves https://github.com/bazelbuild/bazel/issues/16119

Also renamed `MAXIMAL_UPDATE_DELAY_MILLIS` to `MINIMAL_UPDATE_INTERVAL_MILLIS` because it is used as a minimum and doesn't affect the delay between messages.

Fixes #16119.

Closes #16221.

RELNOTES[INC]: Bazel no longer increases the delay between progress updates when there is no cursor control.

PiperOrigin-RevId: 484203851
Change-Id: I33f7302ca85b75135744eb6093b618196199a49f

### DIFF
--- a/site/en/docs/user-manual.md
+++ b/site/en/docs/user-manual.md
@@ -790,6 +790,9 @@ The default is 0, that means an incremental algorithm: the first
 report will be printed after 10 seconds, then 30 seconds and after
 that progress is reported once every minute.
 
+When bazel is using cursor control, as specified by
+[`--curses`](#curses), progress is reported every second.
+
 #### `--local_{ram,cpu}_resources {{ "<var>" }}resources or resource expression{{ "</var>" }}` {:#local-resources}
 
 These options specify the amount of local resources (RAM in MB and number of CPU logical cores)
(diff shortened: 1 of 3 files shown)


tom-cnops:
It seems like this wasn't added to the 6.0.0 release notes, though the functionality was included. Not sure if this is intentional? @kshyanashree

kshyanashree:
> It seems like this wasn't added to the 6.0.0 release notes, though the functionality was included. Not sure if this is intentional? @kshyanashree

Hi @tom-cnops, My apologies! I have added it in our Github 6.0.0 release note. Thank you!

tom-cnops:
Thanks @kshyanashree 
Would it also be possible for me to be added in the acknowledgements?

kshyanashree:
> Thanks @kshyanashree Would it also be possible for me to be added in the acknowledgements?

Sure!

PR #16609, opened 2022-10-30
Fix import instructions for Python runfiles library

fmeum:
The rules_python version of the runfiles library has diverged and, since the import statement mentions a repository name, would not work with Bzlmod.

Instead, use a repo root relative import from `@bazel_tools` and only switch to rules_python once it has become the source of truth for the Python rules.

Work towards #16124

### DIFF
--- a/tools/python/runfiles/runfiles.py
+++ b/tools/python/runfiles/runfiles.py
@@ -20,12 +20,12 @@ USAGE:
       py_binary(
           name = "my_binary",
           ...
-          deps = ["@rules_python//python/runfiles"],
+          deps = ["@bazel_tools//tools/python/runfiles"],
       )
 
 2.  Import the runfiles library.
 
-      from rules_python.python.runfiles import runfiles
+      from tools.python.runfiles import runfiles
 
 3.  Create a Runfiles object and use rlocation to look up runfile paths:
 
@@ -49,7 +49,7 @@ USAGE:
     the right environment variables for them:
 
       import subprocess
-      from rules_python.python.runfiles import runfiles
+      from tools.python.runfiles import runfiles
 
       r = runfiles.Create()
       env = {}


fmeum:
@Wyverald @brandjon Could you review this?

Wyverald (approved):

Wyverald:
@rickeylev could you also take a look and see if this is fine?

fmeum:
@bazel-io flag

rickeylev:
Er, sorry, I'm a bit confused by this PR in several regards.

In the prior PR that changed the logic, you said that the "bazel_tools" repo was special? The impression I got was that, for whatever reason, it wasn't subject to the repo renaming, so e.g. "import bazel_tools.python.runfiles" still worked? Is that true?

Or is this only about the rules_python import? Which would be subject to renaming because it's not special like bazel_tools.

> Since the import statement mentions a repository name, would not work with Bzlmod.

So, something to understand is that the names used in the import statement are part of the public API for a library. By changing it from "import rules_python" to "import tools", all users of it are broken. This seems rather gratuitous.

Additionally, the new top level name, "tools" is very generic -- this seems like a name prone to being already taken or likely to collide. Actually, wouldn't rules_python and bazel_tools conflict in this regard? So depending on both and doing "import tools" will get one or the other.

> The rules_python version of the runfiles library has diverged ... 

Ugh, this sounds like a giant mess. With the name collision above, I don't see how doing "import tools" can really reliably work.

A better fix might be to do something in rules_python so that "import rules_python" still works with and without bzlmod enabled.

fmeum:
You are entirely right, `bazel_tools` is special in that it is both the apparent and the canonical repository name, which isn't true for `rules_python`with Bzlmod. Thus, `import bazel_tools.tools.python.runfiles` would also work (and may be preferable as it prevents the ambiguity), but the current `import rules_python.python.runfiles` would not.

But given that the instructions currently provided for the runfiles library under `bazel_tools` result in the one from `rules_python` being used, maybe making changes to the one in `bazel_tools` just doesn't make sense? Following https://bazelbuild.slack.com/archives/CA306CEV6/p1667242843747579, it seems that having the `rules_python` version adopted instead, perhaps coupled with it being published on `pip`, may be the better choice. If you think that's better, I could close this PR and open an issue over at `rules_python` to add Bzlmod support to its runfiles library instead.

fmeum:
@rickeylev Following the discussion in the Slack thread I linked, I will close this PR. I'm open to contributing the changes needed for Bzlmod support to the upstream runfiles library instead once the import/distribution question has been settled.

Wyverald:
@bazel-io fork 6.0.0

fmeum:
@Wyverald I ended up closing this since I took away from my discussions with @rickeylev and @alexeagle that having the runfiles library be part of `rules_python` and being available via pip would be preferable. As a result, these changes can (and need to) be made independent of a Bazel release.

PR #16611, opened 2022-10-30
Move critical path computation out of runtime

Yannic:
... so that we can use it for posting a machine-readable form of the critical path as build metric in the build event stream.


Yannic:
> Sorry for the delay here - please ping us next time.

Will do!

> The change itself looks good, but needs to be rebased on top of master.

Done. PTAL

PR #16612, opened 2022-10-30
Add critical path to BEP

Yannic:
Additional info will be added in follow-ups.


Yannic:
@meisterT PTAL

Yannic:
ping?

meisterT:
@Yannic I am sorry that I didn't get to this one earlier.

What other metrics do you plan to include here in the future? If it is just the total time, adding it here might be better? https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto;l=955;drc=58eeea17db80cc85d685d9319655f2dc398f063d

meisterT:
Are you aware of https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/runtime/ExecutionGraphModule.java?  See also https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/execution_graph.proto

Yannic:
Is that uploaded to BES like the Bazel profile?

brentleyjones:
> Yes, it is. That's also how we use it internally.

I assume we need to modify `--remote_build_event_upload=minimal` to upload this as well, as that only uploads logs and the profile currently:

https://github.com/bazelbuild/bazel/blob/c9c28a9b4324f1eebbbe2e23f471e0880fabd559/src/main/java/com/google/devtools/build/lib/remote/ByteStreamBuildEventArtifactUploader.java#L222-L236

cc: @coeuvre

meisterT:
@brentleyjones I assume this is done by `--experimental_stream_log_file_uploads`, isn't it?

brentleyjones:
I'm not sure, I don't use that flag. But I do use `--remote_build_event_upload=minimal`, and it has logic around which files it will upload in minimal mode. In particular it currently is:
* `.*/bazel-out/[^/]*/testlogs/.*`
* .*/bazel-out/_tmp/actions/std(err|out)-.*
* `file.type == LocalFileType.LOG || file.type == LocalFileType.PERFORMANCE_LOG`

But I just realized all this is doing is adjusting the BEP, not adding a new type of log file, so we should be good.

Issue #16613, opened 2022-10-31
Bazel Windows Binary cannot run without embedded JDK


meteorcloudy:
### Description of the bug:

https://github.com/bazelbuild/bazel/issues/16159#issuecomment-1292560531
https://github.com/bazelbuild/bazel/issues/16159#issuecomment-1295993565

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Download https://releases.bazel.build/6.0.0/rc1/bazel_nojdk-6.0.0rc1-windows-x86_64.exe and run it on Windows

### Which operating system are you running Bazel on?

Windows

meteorcloudy:
I cannot reproduce https://github.com/bazelbuild/bazel/issues/16159#issuecomment-1292560531by building the windows Bazel nojdk binary on my windows machine. Also the nojdk Windows binary at `last_green` doesn't have this issue. I suspect there is some strange infrastructure flakiness? 

@fmeum Did the rules_jni failure happens consistently after https://github.com/bazelbuild/bazel/commit/902a0b5763dddc445f75637c08a1396de8395411?

meteorcloudy:
I rerun the release pipeline at the same commit for release-5.0.0rc1:

Previous:
```
2022-10-24 19:20:07 INFO   Uploading artifact 01840b70-1ba9-4b0f-97f9-ac34367b8e80 bazel_nojdk-6.0.0rc1-windows-x86_64.exe (30310526 bytes)
```

Now
```
2022-10-31 10:47:38 INFO   Uploading artifact 01842da7-6d67-4353-a8b2-1497e9d4df4e bazel_nojdk-6.0.0rc1-windows-x86_64.exe (31858134 bytes)
```

The nojdk windows binary has significantly size difference, and it runs correctly. so there must be some infra problem while building the Windows binary.

meteorcloudy:
Maybe the https://cs.opensource.google/bazel/bazel/+/master:src/package-bazel.sh script is not hermetic enough, at least on Windows? 🧐

fmeum:
@meteorcloudy Yes, it kept happening consistently. Note that I am not using the nojdk build, it's just that the regular with-JDK build doesn't seem to be able to find its embedded JDK.

meteorcloudy:
@fmeum Can you provide a minimal reproduce case?

meteorcloudy:
[...]

PR #16614, opened 2022-10-31
Fix race condition in package-bazel.sh

meteorcloudy:
`bazel build -c //src:bazel.exe //src:bazel_nojdk.exe` sometimes output corrupted binaries on Windows.

The reason is when we are executing the genrule for packaging the bazel zip files, we are writing a "file.list" file into the execroot, however there is no sandbox on Windows. So two actions are actually sharing the same path for the "file.list", this PR fixes the issue by writing the "file.list" file under the tmp dir.

Fixes https://github.com/bazelbuild/bazel/issues/16613


meteorcloudy:
@bazel-io fork 6.0.0

PR #16616, opened 2022-10-31
Remove warnings about ignored starlark options

keith:
Neither clean nor info really make sense to have semantics changed from starlark flags, but if you have a starlark flag in your .bazelrc in the `build` section you see this warning.


fmeum:
@bazel-io flag

meteorcloudy:
@bazel-io fork 6.0.0

Issue #15041, opened 2022-03-15
iOS platform constraints locations


keith:
In order to use bazel toolchains with Apple platforms, we have to define the core set of platforms it supports. This is mostly done by https://github.com/bazelbuild/bazel/blob/96c8a9073807c9e97635ddafe2ed0365a9318d6f/tools/osx/crosstool/BUILD.toolchains

The issue is that since M1 macs were released there are now overlapping OS + CPU combos, like `ios` + `arm64` that need more specific information for whether or not the build is for a physical device, or the simulator. To support this, these custom constraints exist: https://github.com/bazelbuild/apple_support/blob/2ec0ef3eb2954371596fbb1a225fb244fcde31a9/constraints/BUILD#L12-L26

Since they live in apple_support today, and bazel does not take them into account, as far as I can tell this setup isn't able to do what we want. It seems to me that we either need to move these constraints into bazel, and annotate the toolchains here, or move the toolchain definitions out of bazel and into apple_support (or another location?).

I'm hoping we can use this issue to decide on this, which will help make some more progress on toolchain support for Apple platforms.

keith:
In the short term if you want to try to use this, I believe you'd have to manually override `local_config_cc_toolchains` and define some valid toolchains yourself.

keith:
Related https://github.com/bazelbuild/bazel/pull/14844

gregestren:
As next step our team should discuss.

Since these definitions are "special" - owned by Apple but intermingled with C++, my intuition is the most practical way to continue supporting now is to move whatever extra constraints we need into Bazel, I suppose next to https://github.com/bazelbuild/bazel/blob/96c8a9073807c9e97635ddafe2ed0365a9318d6f/tools/osx/crosstool/BUILD.toolchains.

i.e. cleanly breaking Apple toolchain ownership from core Bazel is its own issue, with its own scope. For better or worse it's in Bazel now. If we need to support Apple toolchains, that's where we need to support them. Moving *everything* sounds like biting off a much larger commitment.

gregestren:
Update: I expect us to discuss this on Wednesday.

gregestren:
We chatted today and support the opinion at https://github.com/bazelbuild/bazel/issues/15041#issuecomment-1068495694.

[...]

Issue #12439, opened 2020-11-09
objc_library rules fail to link (empty *.objlist) when position-independent code is enabled


erikmchut:
# Description of the problem / feature request:

objc_library rules fail when the position-independent code feature is enabled.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Build an objc_library rule with `--force_pic` or `-c dbg --features=supports_pic`

### What operating system are you running Bazel on?

Mac OSX with a custom apple crosstool

### What's the output of `bazel info release`?

release 1.2.1

###  Have you found anything relevant by searching the web?

I've looked into the bazel codebase. It appears that there is a hardcoded line in the call to `registerObjFilelistAction` to includes only object files build without pic in the intermediate objlist file. As a result, the `*.objlist` file that is generated is empty if pic is enabled.

https://github.com/bazelbuild/bazel/blob/1.1.0/src/main/java/com/google/devtools/build/lib/rules/objc/CompilationSupport.java#L1097

### Any other information, logs, or outputs that you want to share?

The error is the following
`fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: file list file: bazel-out/apl-darwin_x86_64-opt/bin/c8/gui/imgui-app-archive.objlist is empty
`
This error stems from the fact that `imgui-app-archive.objlist` is indeed empty. I believe that it is empty because the object files generated in the compile step are all .pic.o, but the hardcoded line (linked above) is only looking for non-pic object files to include. It perhaps should be updated to use the value of the current bazel pic setting.

susinmotion:
@googlewalt is this in your wheelhouse?

googlewalt:
Sure.  I can take this.

epertoso:
@googlewalt any update on this one? On the surface, it seems it's a matter of adding the pic object files to the action that emits the `.objlist` file, but maybe is there some other reason why this isn't the case?

As a side note, al the default macOS `apple_cc_toolchain`s don't have the `supports_pic` feature at all (it's declared [here](https://github.com/bazelbuild/bazel/blob/a04cb1bfad4734f801c48bae3070a799067bda4e/tools/osx/crosstool/cc_toolchain_config.bzl#L3516) but is not used by any toolchain except `armeabi-v7a`, which IIUC is relevant for Android, therefore never being selected for Objective-C-related actions), maybe that's why this bug isn't so noticeable.

[...]

PR #16619, opened 2022-11-01
Move Apple toolchain setup to apple_support

keith:
This moves the CC toolchain for building Apple platforms besides macOS to the apple_support repo https://github.com/bazelbuild/apple_support/pull/113

The default unix toolchain is now used if someone wants to build for macOS without the apple_support toolchain, but it doesn't handle as many platform specific features as the previous toolchain.

Fixes https://github.com/bazelbuild/bazel/issues/15041


keith:
This is failing because rules_cc references the primary file we're deleting here

keith:
Yea I definitely can, the problem will be doing that in lockstep, but I guess it's ok since it's breaking either way

keith:
https://github.com/bazelbuild/rules_cc/pull/154

keith:
I think the order of operations has to be:

1. remove from rules_cc
2. create rules_cc release, update it in bazel
3. remove here
4. merge into apple_support

keith:
@katre ok I moved out all of the apple crosstool setup that I can, and moved it into https://github.com/bazelbuild/apple_support/pull/113/

one thing I'm unclear on currently is that I currently have the apple_support using `local_config_cc`, and falling back in the case the Apple toolchain doesn't apply. It seems ideally I would use `local_config_apple_cc` or something, but currently that would require we force a `--crosstool_top` argument that I don't think we want to. Based on https://github.com/grailbio/bazel-toolchain, this might change if we were using `--incompatible_enable_cc_toolchain_resolution`, but I believe that's still blocked on this detail https://github.com/bazelbuild/bazel/issues/7260#issuecomment-1255255323

katre:
The right answer is for non-toolchains users to use `--crosstool_top=@rules_apple//whatever`, and for toolchains users to pick it up in the WORKSPACE.

That sequencing looks right but we should confirm with @comius.

keith:
I wonder if adding that in our transition would be enough instead of requiring that change on folks before we transition to the workspace way

keith:
it seems like if we could change the default of `apple_crosstool_top` to the new toolchain _when it exists_ users wouldn't have to know about this. we could do that always in rules_apple, but that also is used in a transition in this repo, and defaults to the default CC toolchain

keith:
I think this is pretty close to ready, but I can't make CI green until we have a new release of rules_cc, it seems too difficult to correctly override all the tests cases for it at the moment.

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -1368,6 +1385,11 @@ cc_toolchain_config = rule(
         "coverage_link_flags": attr.string_list(),
         "supports_start_end_lib": attr.bool(),
         "builtin_sysroot": attr.string(),
+        "_xcode_config": attr.label(default = configuration_field(
+            fragment = "apple",
+            name = "xcode_config_label",
+        )),
     },
     provides = [CcToolchainConfigInfo],
+    fragments = ["apple"],
```
brentleyjones, line 1394: nit: should be above `provides`?

tools/cpp/unix_cc_configure.bzl
```diff
@@ -488,7 +488,8 @@ def configure_unix_toolchain(repository_ctx, cpu_value, overriden_tools):
             cc,
             "-xc++",
             cxx_opts + no_canonical_prefixes_opt + ["-stdlib=libc++"],
-        ),
+        ) +
+        ["/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"],
```
keith, line 492: This is required because now this toolchain can be used when the user has Xcode + the CLT installed, in which case the Xcode paths are automatically collected, but this SDK can still be used.

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -285,6 +292,17 @@ def _impl(ctx):
                 with_features = [with_feature_set(features = ["opt"])],
             ),
         ],
+        env_sets = [
+            env_set(
+                actions = all_link_actions + lto_index_actions + [ACTION_NAMES.cpp_link_static_library],
+                env_entries = ([
+                    env_entry(
+                        key = "ZERO_AR_DATE",
```
keith, line 300: This is required to get reproducible archives / links on macOS, it shouldn't hurt applying to other platforms

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -1275,6 +1293,17 @@ def _impl(ctx):
         ],
     )
 
+    macos_minimum_os_feature = feature(
+        name = "macos_minimum_os",
+        enabled = True,
+        flag_sets = [
+            flag_set(
+                actions = all_compile_actions,
+                flag_groups = [flag_group(flags = ["-mmacos-version-min={}".format(_target_os_version(ctx))])],
```
keith, line 1302: this is required because when you have Xcode + the CLT installed, it seems that the minimum os version behavior changes somehow, without explicitly passing this I saw it defaulting to the oldest version. This also means `--macos_minimum_os` works for C* binaries

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -1350,9 +1379,9 @@ def _impl(ctx):
             ),
         ]
         features = [
+            macos_minimum_os_feature,
             libtool_feature,
             archiver_flags_feature,
-            supports_pic_feature,
```
keith, line 1355: This was not supposed to be here, but was just not getting exercised. see https://github.com/bazelbuild/bazel/issues/12439

tools/cpp/unix_cc_toolchain_config.bzl
```diff
@@ -1367,7 +1396,6 @@ def _impl(ctx):
             user_link_flags_feature,
             default_link_libs_feature,
             fdo_optimize_feature,
-            supports_dynamic_linker_feature,
```
keith, line 1370: This was not supposed to be here and not in the default Xcode toolchain

MODULE.bazel
```diff
@@ -2,8 +2,17 @@
...
     repo_name = "io_bazel",
+    version = "7.0.0-pre",
+)
+
+archive_override(
+    integrity = "sha256-L9ASXKv/pEv9VKLmueHxobnl5aCWk4MNTZywE0MKIe4=",
+    module_name = "rules_cc",
+    strip_prefix = "rules_cc-ffaaf6bfb168eef354ae3b089085ccb86b47d1da",
+    urls = [
+        "https://github.com/keith/rules_cc/archive/ffaaf6bfb168eef354ae3b089085ccb86b47d1da.tar.gz",
```
katre, line 14: Presumably this needs to be a released rules_cc at some point, just flagging so it's not forgotten.

src/main/starlark/builtins_bzl/common/objc/objc_library.bzl
```diff
@@ -39,6 +39,15 @@ def _objc_library_impl(ctx):
...
     cc_toolchain = cc_helper.find_cpp_toolchain(ctx)
+    feature_configuration = cc_common.configure_features(
+        ctx = ctx,
+        cc_toolchain = cc_toolchain,
+        requested_features = ctx.features,
+        unsupported_features = ctx.disabled_features,
+    )
+
+    if not cc_common.action_is_enabled(feature_configuration = feature_configuration, action_name = "objc-compile"):
+        fail("Compiling objc_library targets requires the Apple CC toolchain which can be found here: https://github.com/bazelbuild/apple_support")
```
katre, line 50: Nice!

src/test/java/com/google/devtools/build/lib/pkgcache/BUILD
```diff
@@ -278,10 +278,10 @@ java_library(
...
 )
 
-java_test(
-    name = "LoadingPhaseRunnerTest",
-    test_class = "com.google.devtools.build.lib.pkgcache.LoadingPhaseRunnerTest",
-    runtime_deps = [
-        ":LoadingPhaseRunnerTest_lib",
-    ],
-)
+# java_test(
```
katre, line 281: Should this be removed, or is the test going to be fixed?
Prefer tagging as "noci" if the test doesn't pass on CI.

src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -23,6 +23,17 @@
 from src.test.py.bazel.bzlmod.test_utils import BazelRegistry
 from src.test.py.bazel.bzlmod.test_utils import scratchFile
 
+_OVERRIDE = """
```
katre, line 26: Is this still needed after rules_cc has your changes?

src/test/shell/bazel/BUILD
```diff
@@ -369,14 +369,14 @@ sh_test(
...
 
-sh_test(
-    name = "bazel_rules_cc_test",
-    srcs = ["bazel_rules_cc_test.sh"],
-    data = [
-        ":test-deps",
-        "@bazel_tools//tools/bash/runfiles",
-    ],
-)
+# sh_test(
```
katre, line 372: Same comment about using "noci", and whether this test can be re-enabled.

src/test/shell/bazel/BUILD
```diff
@@ -1055,44 +1055,44 @@ sh_test(
...
-        "//src:embedded_jdk_allmodules",
-        "@bazel_tools//tools/bash/runfiles",
-    ],
-    exec_compatible_with = ["//:highcpu_machine"],
-    tags = [
-        "block-network",
-        "no_windows",
-    ],
-)
+# sh_test(
```
katre, line 1058: And here.

src/test/shell/bazel/apple/bazel_apple_test.sh
```diff
@@ -40,101 +40,6 @@ function set_up() {
   create_new_workspace
 }
 
-function test_fat_binary_no_srcs() {
```
katre, line 43: Are these tests moving to apple_support with the toolchain, or are they no longer needed at all?

tools/cpp/unix_cc_configure.bzl
katre: Can you add that as a comment in the code?

tools/cpp/unix_cc_toolchain_config.bzl
katre: Definitely make this a comment.

MODULE.bazel
keith: yep! Hoping that @oquenchil can help with a release after https://github.com/bazelbuild/rules_cc/pull/154 is merged (which hopefully is in progress)

src/test/py/bazel/bzlmod/bazel_module_test.py
keith: nope, just need that update

src/test/shell/bazel/BUILD
keith: nope, just need that update

keith: I'll just leave these as broken since they shouldn't ever land this way

src/test/shell/bazel/BUILD
keith: I'll just leave these as broken since they shouldn't ever land this way

src/test/shell/bazel/apple/bazel_apple_test.sh
keith: these all moved to apple_support. anything that relied on the real toolchain and non macOS basic features has to move there. there are still many tests here using the mock toolchain for coverage

keith:
Threw the comments in here, didn't change the broken tests since we shouldn't land until we have the rules_cc  update here, at which point I'll remove them from this diff

src/test/java/com/google/devtools/build/lib/packages/util/mock/osx_cc_toolchain_config.bzl
```diff
@@ -2319,6 +2330,7 @@ def _impl(ctx):
         )
     elif (ctx.attr.cpu == "darwin_x86_64"):
         objc_compile_action = action_config(
+            enabled = True,
```
brentleyjones, line 2333: Why are these needed?

tools/cpp/unix_cc_toolchain_config.bzl
brentleyjones: Probably too much effort, but should we land these changes first?

src/test/java/com/google/devtools/build/lib/packages/util/mock/osx_cc_toolchain_config.bzl
keith: this is what objc_library now uses to tell if it's supported to build objc

oquenchil (approved):

brentleyjones:
@oquenchil This can't be merged yet, it has commented out stuff waiting on the new 0.0.5 rules_cc release. Can you change the labels to prevent it accidentally landing early?

keith:
Yep good idea, just reverted all the test workarounds so this will be red and blocked on https://github.com/bazelbuild/rules_cc/issues/174 until we update bazel with that new version

sgowroji:
Hi @keith, Above PR is awaiting to merge. Could you please fix the above branch conflicts and  Pending [changes](https://github.com/bazelbuild/bazel/pull/16619#pullrequestreview-1264017133). Thanks!

brentleyjones:
@sgowroji still blocked on https://github.com/bazelbuild/rules_cc/issues/174. The red statuses are expected until that.

brentleyjones:
rules_cc 0.0.5 has been released. Keith is OOO for another week, so no progress will be made on this PR until at least then.

keith:
We're all green here 🎉 although we should probably merge https://github.com/bazelbuild/bazel/pull/17475 first to simplify reverts if needed.

@oquenchil @katre could you run the downstream pipeline job with this change? Also is there anything else on your mind that we should do before merging?

meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2873

keith:
thanks @meteorcloudy can you run it one more time with the fixes I pushed for those? (note that many of those issues were not introduced by this PR)

oquenchil:
New one here: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2876

keith:
@oquenchil @katre I submitted PRs for all the downstream breakages from this, can we talk about a merge plan?

oquenchil:
Thanks Keith. What are the reasons for a specific plan instead of just merging the PR normally?

keith:
I guess I just wanted to hear from you all if you have any specific concerns. I imagine there will be some straggler issues that come up and I'm happy to field them. Assuming they're not catastrophic I think it would be nice if we could avoid reverting since it's a conflict-prone change

keith:
but if you all have no other concerns then this is ready to go!

oquenchil:
Still working on merging this.

googlewalt:
I have a concern about migrating the OSS apple crosstool from bazel to a separate repository.  In the past I have had to make changes to the OSS apple crosstool so that it works with changes I'm making to bazel.  How do you envision this being handled in the future, if such a need arises?  Are the apple_support repo owners prepared to handle these changes, and any issues that may arise with keeping the toolchain compatible with multiple versions of bazel.

A couple options that would lessen/eliminate my concern:

- keep a copy of the apple crosstool in bazel, so that me and other bazel developers can update the crosstool in sync with bazel changes, that apple_support can import as needed.

- keep the crosstool in bazel, but proceed with the rest of the proposed migration.

keith:
Given the historical examples I don't think the rate or complexity of changes that you're referring to will be a problem for this living in another repo. I think depending on the changes you can submit PRs, or we could adapt them ourselves.

I don't think duplicating or leaving any of the moved files here achieves the goals we set out to solve. Duplicating the toolchain would increase the burden on the rules maintainers, and both of those options would continue to block supporting platforms with this toolchain in the future.

I also wonder how these concerns apply to other custom toolchains folks are using as well? For example if we duplicated the toolchain, the downstream pipeline would still break if bazel made breaking changes that also required toolchain changes, similar to bazel-toolchain or other popular CC toolchains that aren't part of this repo.

googlewalt:
We see a bunch of failures: https://buildkite.com/bazel/google-bazel-presubmit/builds/64142. We're fixing it in the internal import patch, but I want to check whether this should have been caught by bazel presubmits earlier?

keith:
i imagine something broke since i rebased last?

keith:
let me know if you want me to fix on this branch!

keith:
Thanks all! @googlewalt just so I know for the future, are the files you ended up keeping libtool.sh / make_hashed_objlist.py still there because they're used in those paths internally? (since i don't see any external references)

meteorcloudy:
image
https://buildkite.com/bazel/bazel-bazel/builds?branch=master&page=3
https://buildkite.com/bazel/bazel-bazel/builds/22421#0186ca61-268c-4135-b44d-f1123f11d11c

It looks like this PR is causing a lot of tests to time out in Bazel's postsubmit. @keith Can you take a look?

meteorcloudy:
I guess we need to increase the test size of those timing out tests, e.g. https://cs.opensource.google/bazel/bazel/+/master:src/test/shell/bazel/BUILD;l=937

But this does indicate that there is a potential performance regression?

googlewalt:
> Thanks all! @googlewalt just so I know for the future, are the files you ended up keeping libtool.sh / make_hashed_objlist.py still there because they're used in those paths internally? (since i don't see any external references)

Yes.  I suspect we should get rid of that dependency but left the cleanup for another day.

keith:
@meteorcloudy I'll look at this today

keith:
https://github.com/bazelbuild/bazel/pull/17763

meteorcloudy:
This change was rollbacked at https://github.com/bazelbuild/bazel/commit/a50cca561fe0a77c3750101e505171ddcc16db72, @zhengwei143 can you provide more details about what was broken?

brentleyjones:
I believe @googlewalt is working on a roll forward.

keith:
@googlewalt how's the forward fix looking?

googlewalt:
I just fixed one small remaining issue.  Awaiting review @oquenchil.

googlewalt:
Submitted as https://github.com/bazelbuild/bazel/commit/699e40373f95e42390a85f29dfa1098636336103.

keith:
Thanks!

PR #16621, opened 2022-11-01
Move Boost into C++ Docs; Add Libraries Section

cpsauer:
Hi wonderful Bazelers,

This is just a docs change.

Backstory: I'd been looking to make HTTPS requests across platforms from C++. A classic problem if there ever were one, networking being perhaps the most glaring omission in the C++ standard library. Thankfully, this is a problem Bazel can solve well, since most of the problem is the friction of using 3rd party libraries from C++. So, I spun up some build rules to make network requests easy, inspired by collaborating on the boost ones, and set off to add them to the docs.

Along the way, I noticed that the boost rules were in an odd spot: Listed at the language level alongside C++, rather than nested within C++. So I fixed that by nesting Boost inside, added Abseil, and then (hoping you'll forgive my hubris), I'd love to add the rules I just released, since I think they're a solution to a very real need. Perhaps rules for more famous, critical libraries can accumulate there over time, helping Bazel users get set up with the essential tools they need.

Chris
(ex-Googler and author of [bazel-compile-commands-extractor](https://github.com/hedronvision/bazel-compile-commands-extractor), also in the docs.)

### DIFF
--- a/site/en/docs/bazel-and-cpp.md
+++ b/site/en/docs/bazel-and-cpp.md
@@ -14,6 +14,10 @@ The following resources will help you work with Bazel on C++ projects:
 *  [Tutorial: Building a C++ project](/tutorials/cpp)
 *  [C++ common use cases](/tutorials/cpp-use-cases)
 *  [C/C++ rules](/reference/be/c-cpp)
+*  Essential Libraries
+   -  [Abseil](https://abseil.io/docs/cpp/quickstart){: .external}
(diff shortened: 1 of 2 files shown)


cpsauer:
[Tagging @gregestren, since I see him listed as code owner, and because we've worked together on changes in the past, so he might have the most context.]

gregestren:
@oquenchil does this look reasonable to you?

oquenchil (approved):

PR #16622, opened 2022-11-01
redact 'token' strings from logging

mattem:
It's common for users to set 'TOKEN' as an env var. While this is a little like whack-a-mole and we can't cover everything, this seems like a common string to redact.


alexeagle:
ping @meisterT ?

meisterT:
I requested it to be cherry-picked in https://github.com/bazelbuild/bazel/pull/16764

PR #16623, opened 2022-11-01
Make C++ runfiles library repo mapping aware

fmeum:
Also removes a comment mentioning a `Create` overload that does not exist.

Work towards #16124


fmeum:
@oquenchil @Wyverald Could you review?

tools/cpp/runfiles/runfiles_src.h
```diff
@@ -160,21 +180,48 @@ class Runfiles {
     return envvars_;
   }
 
+  // Returns a new Runfiles instance that by default uses the provided source
+  // repository as a default for all calls to Rlocation.
+  //
+  // The current instance remains valid.
+  Runfiles* WithSourceRepository(const std::string& source_repository) const {
```
Wyverald, line 187: is it too late to make this return a `std::unique_ptr<Runfiles>`?

tools/cpp/runfiles/runfiles_src.h
```diff
@@ -160,21 +180,48 @@ class Runfiles {
...
+                     std::move(repo_mapping), std::move(envvars),
+                     std::string(source_repository));
+    return new_runfiles;
+  }
+
  private:
-  Runfiles(const std::map<std::string, std::string>&& runfiles_map,
-           const std::string&& directory,
-           const std::vector<std::pair<std::string, std::string> >&& envvars)
+  Runfiles(std::map<std::string, std::string>&& runfiles_map,
```
Wyverald, line 201: just make these take the bare type (`std::map<...>` instead of `std::map<...>&&`)

tools/cpp/runfiles/runfiles_src.h
```diff
@@ -160,21 +180,48 @@ class Runfiles {
...
+    std::map<std::string, std::string> runfiles_map(runfiles_map_);
+    std::string directory(directory_);
+    std::map<std::pair<std::string, std::string>, std::string>
+        repo_mapping(repo_mapping_);
+    std::vector<std::pair<std::string, std::string> > envvars(envvars_);
+    auto* new_runfiles =
+        new Runfiles(std::move(runfiles_map), std::move(directory),
+                     std::move(repo_mapping), std::move(envvars),
+                     std::string(source_repository));
+    return new_runfiles;
```
Wyverald, line 197: you don't need the separate copies. just `return new Runfiles(runfiles_map, directory, ...);` works.

Wyverald (approved):

tools/cpp/runfiles/runfiles_src.h
fmeum: The function is new and even though this deviates from `Create`, I agree it's just better.

tools/cpp/runfiles/runfiles_src.h
fmeum: Fixed.

tools/cpp/runfiles/runfiles_src.h
fmeum: Changed.

oquenchil (approved):

Wyverald (approved):

Wyverald:
@bazel-io fork 6.0.0

kshyanashree:
Hi @fmeum! I am trying to cherry-pick this change to the release-6.0.0, but pre-submits checks are failing due to missing dependencies, So could you please help us in cherry-picking this change with appropriate commits. Thanks!

fmeum:
@kshyanashree I think that the test failure is caused by [this Git update](https://lore.kernel.org/lkml/xmqq4jw1uku5.fsf@gitster.g/). It likely also fails on master.

@meteorcloudy

meteorcloudy:
Yes, we just pushed new docker image for Linux VMs, there seems to be some problem, I have reverted the image version to the last working version, rerun the presubmit should fix the problem.

kshyanashree:
> @kshyanashree I think that the test failure is caused by [this Git update](https://lore.kernel.org/lkml/xmqq4jw1uku5.fsf@gitster.g/). It also fails on master with Git 2.38.1.
> 
> @meteorcloudy

Ohh I see! Thanks for the info @fmeum

kshyanashree:
> Yes, we just pushed new docker image for Linux VMs, there seems to be some problem, I have reverted the image version to the last working version, rerun the presubmit should fix the problem.

Thanks @meteorcloudy!

PR #16624, opened 2022-11-01
[remote] Respect whether the server supports action cache updates

Yannic:
Only a subset of users may be allowed to update the action cache (e.g., only CI but not devs).

Today, there are 2 ways to achive the desired behavior:
  - `GetCapabilities` returning that all users are allowed to update, and `UpdateActionResult` returning an error that Bazel prints and ignores, or
  - have the users that are not allowed to update the action cache set `--remote_upload_local_results=false`.

Why don't we instead respect whether the remote cache supports uploading action results?

Note that this requires support from the remote system to fully work (i.e., it needs to return `update_enabled = false` for users that don't have permission). Otherwise, Bazel's behavior will be the same as before this change: failed `UpdateActionResult` do not cause the build to fail. The only change this introduces is that Bazel will no longer error if `--remote_upload_local_results=true` and `GetCapabilities` returning `update_enabled = false`.


bduffany:
@Yannic what do you think about printing an info- or warning-level message if remote_upload_local_results=true but the server returns update_enabled=false? Maybe something like "INFO: remote cache has disabled action result uploads, so --remote_upload_local_results setting will be ignored."

Asking because users might be confused by action uploads silently not working, especially if they are intentionally passing --remote_upload_local_results at the command line. "Why isn't <X> a cache hit?" is one of our most common support requests, so just want to minimize confusion there.

Yannic:
@bduffany That sounds reasonable, at least for a transition period. It's also known server-side, so it would be possible to add a warning server-side for that (e.g., on the result page).

Yannic:
@coeuvre @meisterT PTAL

Yannic:
@bazel-io flag

Yannic:
I'd like to get this into 6.0.0 if at all possible

coeuvre (approved):

Issue #13839, opened 2021-08-12
Separate `--features` and `--host_features`


keith:
Currently if you flip a feature like `--features=tsan` that affects compilation, any host / exec tools you depend on for your build will be recompiled with the tsan features, even though you likely only care about that for your target build, not your host tools.

I think there should be some way to separate these, for example having `--features` only apply to your `target` configuration and having a separate `--host_features` for the host / exec configuration similar to `--copt` and `--host_copt` (I'm sure there are many other potential solutions to this though)

### Have you found anything relevant by searching the web?

I really thought there was a tracking issue for this somewhere, but here are some slightly related issues:

- https://github.com/bazelbuild/bazel/issues/13315
- https://github.com/bazelbuild/bazel/issues/13308
- https://github.com/bazelbuild/bazel/issues/13048

jyknight:
There's a bug in this change: it doesn't pass on `--host_features` _itself_ to the exec config, so a further transition within exec will drop all features. That is, it's missing a line like:
`exec.hostFeatures = ImmutableList.copyOf(hostFeatures);`

keith:
@jyknight ah yes thanks, do you have a transition example where you saw this that I can use for a test case

keith:
Submitted https://github.com/bazelbuild/bazel/pull/17625 in the meantime, but an example would be helpful

jyknight:
Sorry, I don't have a standalone example. This issue broke builds with an action conflict in my codebase when I tried to pass `--incompatible_use_host_features --host_features foo`.

I'll guess it probably arises from a binary "X" requiring two exec tools "A" and "B" to generate code, where exec tool "B" also requires "A" as an exec tool. I think that'd result in `X -exec> A; X -exec> B -exec> A`. And with this bug, the second A will have a different feature set than the first A.

PR #16626, opened 2022-11-01
Add --host_features

keith:
Previously there was no way to have a feature only apply to the entire transitive target closure without `--features` which also applied to the host / exec configuration. This is undesirable for many features such as C++ sanitizers where you often only need them to affect targets in the target configuration, and you don't want to invalidate all host tools when switching between these build configurations.

RELNOTES[INC]: `--features` only applies to targets built in the target configuration, and `--host_features` is used for the host / exec configuration (gated behind `--incompatible_use_host_features`)

Fixes https://github.com/bazelbuild/bazel/issues/13839

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
@@ -646,12 +646,34 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
       documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
       effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
       help =
-          "The given features will be enabled or disabled by default for all packages. "
-              + "Specifying -<feature> will disable the feature globally. "
+          "The given features will be enabled or disabled by default for targets "
+              + "built in the target configuration. "
+              + "Specifying -<feature> will disable the feature. "
               + "Negative features always override positive ones. "
-              + "This flag is used to enable rolling out default feature changes without a "
-              + "Bazel release.")
+              + "See also --host_features")
   public List<String> defaultFeatures;
+
+  @Option(
+      name = "host_features",
+      allowMultiple = true,
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
+      help =
+          "The given features will be enabled or disabled by default for targets "
+              + "built in the exec configuration. "
+              + "Specifying -<feature> will disable the feature. "
+              + "Negative features always override positive ones.")
+  public List<String> hostFeatures;
+
+  @Option(
+      name = "incompatible_use_host_features",
+      defaultValue = "false",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
+      metadataTags = {OptionMetadataTag.INCOMPATIBLE_CHANGE},
+      help = "If true, use --features only for the target configuration and --host_features for the exec configuration.")
+  public boolean incompatibleUseHostFeatures;
 
   @Option(
       name = "target_environment",
@@ -989,7 +1011,12 @@ public class CoreOptions extends FragmentOptions implements Cloneable {
     exec.checkLicenses = checkLicenses;
 
     // === Pass on C++ compiler features.
-    exec.defaultFeatures = ImmutableList.copyOf(defaultFeatures);
+    exec.incompatibleUseHostFeatures = incompatibleUseHostFeatures;
+    if (incompatibleUseHostFeatures) {
+      exec.defaultFeatures = ImmutableList.copyOf(hostFeatures);
+    } else {
+      exec.defaultFeatures = ImmutableList.copyOf(defaultFeatures);
+    }
 
     // Save host options in case of a further exec->host transition.
     exec.hostCpu = hostCpu;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -624,6 +624,19 @@ public OutputDirectoryNamingSchemeConverter() {
...
 
+  @Option(
+      name = "host_features",
+      allowMultiple = true,
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
+      help =
+          "The given features will be enabled or disabled by default for packages "
+              + "built in the host or exec configurationsx. "
```
luispadron, line 635:
```suggestion
              + "built in the host or exec configurations. "
```

luispadron (approved):

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
keith: thanks!

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -624,6 +624,19 @@ public OutputDirectoryNamingSchemeConverter() {
...
   public List<String> defaultFeatures;
 
+  @Option(
+      name = "host_features",
+      allowMultiple = true,
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
+      help =
+          "The given features will be enabled or disabled by default for packages "
```
fmeum, line 634:
```suggestion
          "The given features will be enabled or disabled by default for targets "
```

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -624,6 +624,19 @@ public OutputDirectoryNamingSchemeConverter() {
...
+              + "built in the host or exec configurations. "
+              + "Specifying -<feature> will disable the feature globally. "
```
fmeum, line 636: "globally" can be confusing here given that the previous sentence take about the "exec configuration" only.

src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java
```diff
@@ -624,6 +624,19 @@ public OutputDirectoryNamingSchemeConverter() {
...
 
+  @Option(
+      name = "host_features",
+      allowMultiple = true,
+      defaultValue = "null",
+      documentationCategory = OptionDocumentationCategory.OUTPUT_PARAMETERS,
+      effectTags = {OptionEffectTag.CHANGES_INPUTS, OptionEffectTag.AFFECTS_OUTPUTS},
+      help =
+          "The given features will be enabled or disabled by default for packages "
+              + "built in the host or exec configurations. "
```
fmeum, line 635:
```suggestion
              + "built in the exec configurations. "
```

Host is gone.

fmeum (commented):
Given that this an incompatible change, the 
PR description should have a `RELNOTES[INC]` section explaining the impact.

It may also be helpful to add a section to the `--features` help text saying that it only affects the target config.

fmeum (approved):
@oquenchil Could you review this? It does make sanitizers much more usable with Bazel, which we are also interested in.

oquenchil (dismissed):

oquenchil:
I can't merge this because of the incompatibility internally. I can't really give you an accurate estimate but given how many times --features is used in different bazelrc files, I suspect it'd be a week of full time work minimum to migrate those configurations while making sure that nothing breaks.

We can justify that time if you tell me this is of huge benefit to OSS users. Ideally though you could add the same functionality while keeping the change compatible.

In any case if we move forward with the incompatible change, this would have to be done in two stages. First one that adds the flag and is a no-op so that I can update bazelrc files. Then the incompatible one.

brentleyjones:
For the example reason listed in the PR description, only applying the sanitizer features to targets, not tools, it's a pretty big deal.

I'll leave the rest of your comment for Keith. He's OOO until mid-February.

fmeum:
Just came across another use case for this: `protoc` doesn't build with `layering_check`, which is otherwise a very useful feature.

pierfale:
I just came across a case where I needed this PR: 
I use a custom toolchain where I use features to enable or not instruction sets (e.g. avx512).
The target should enable this feature but not the host since my build system do not support this instruction set (especially to execute protoc).
Just adding --host_features=-avx512 did the trick

keith:
@oquenchil fair questions

> Ideally though you could add the same functionality while keeping the change compatible.

I imagine we could come up with some other solution, like a new pair of flags for this, but I guess that's a short-term vs long-term maintainability question. I think the ideal end state is what we have here, but if you think that's too tough of a migration then we could do something else. (like `--features_new=foo`, `--host_features_new=bar` and keep the current behavior of `--features`)

Personally I actually think the times where you _want_ to have `--features` also apply to the exec configuration is less often than when you don't, so that combined with the easy migration path if you do (dup your `--features` args and add a `--host_features` for each), seems to me like this with an incompatible flag is a fine path forward given enough time for users to migrate before the flip.

I've added the off by default incompatible flag `--incompatible_use_host_features` here

oquenchil:
I agree that having 3 flags would not be ideal. 

I will shoot an email to the internal toolchain teams list to see if this is something they'd really like and would be willing to migrate.

oquenchil:
So... as far as I have understood there is plenty of motivation for this. Great idea about rolling this with an incompatible flag first. I will get it merged.

Thank you!

oquenchil:
Yes, please go ahead.

brentleyjones:
@bazel-io flag

kshyanashree:
@bazel-io fork 6.1.0

Issue #16635, opened 2022-11-02
Bazel NPE when setting `--java_classpath=bazel`


arunkumar9t2:
### Description of the bug:

Setting `--java_classpath=bazel` on the command line and building a simple project crashes bazel with following stacktrace

```java
java.lang.RuntimeException: Unrecoverable error while evaluating node 'ActionLookupData{actionLookupKey=ConfiguredTargetKey{label=//classpath:B_Java, config=BuildConfigurationKey[62ff62eac06d5b2a3021fd76848185e8623f28c27a49346ecc5997f2c301b7b9]}, actionIndex=0}' (requested by nodes 'TargetCompletionKey{topLevelArtifactContext=com.google.devtools.build.lib.analysis.TopLevelArtifactContext@90904c3b, actionLookupKey=ConfiguredTargetKey{label=//classpath:A_Java, config=BuildConfigurationKey[62ff62eac06d5b2a3021fd76848185e8623f28c27a49346ecc5997f2c301b7b9]}, willTest=false}')
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:642)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:382)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.NullPointerException
	at com.google.devtools.build.lib.rules.java.JavaCompileAction.getReducedClasspath(JavaCompileAction.java:262)
	at com.google.devtools.build.lib.rules.java.JavaCompileAction.beginExecution(JavaCompileAction.java:404)
	at com.google.devtools.build.lib.actions.Action.execute(Action.java:133)
	at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$5.execute(SkyframeActionExecutor.java:936)
	at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$ActionRunner.continueAction(SkyframeActionExecutor.java:1103)
	at com.google.devtools.build.lib.skyframe.SkyframeActionExecutor$ActionRunner.run(SkyframeActionExecutor.java:1061)
[...]

PR #16636, opened 2022-11-02
Fix NPE when setting `--java_classpath=bazel`

arunkumar9t2:
Fixes https://github.com/bazelbuild/bazel/issues/16635

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
+++ b/src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
@@ -80,6 +80,7 @@ import com.google.devtools.build.lib.profiler.ProfilePhase;
 import com.google.devtools.build.lib.profiler.Profiler;
 import com.google.devtools.build.lib.profiler.ProfilerTask;
 import com.google.devtools.build.lib.profiler.SilentCloseable;
+import com.google.devtools.build.lib.rules.java.JavaCompileActionContext;
 import com.google.devtools.build.lib.runtime.BlazeModule;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/devtools/build/lib/buildtool/ExecutionTool.java
```diff
@@ -184,6 +187,10 @@ public class ExecutionTool {
         .restrictTo(WorkspaceStatusAction.Context.class, "")
         .restrictTo(SymlinkTreeActionContext.class, "");
 
+    // Register JavaCompileActionContext for java_classpath=bazel to work
+    actionContextRegistryBuilder.register(
+        JavaCompileActionContext.class, new JavaCompileActionContext());
+
```
comius, line 193: A better location to do this registration is `BazelStrategyModule` overriding `registeringActionContext`. The module allows for better extensibility, that is potential use of a different compilation context.

Please move it there.

src/main/java/com/google/devtools/build/lib/rules/java/BUILD
```diff
@@ -79,6 +79,19 @@ java_library(
     ],
 )
 
+java_library(
```
comius, line 82: It's a bad practice to have a source file part of two rules. Please remove `JavaCompileActgionContext.java` from the other rule where it's still used and fix dependencies if needed.

comius (changes requested):
Does the bazel classpath work after this fix?

Could you also add a regression test for this NPE?

arunkumar9t2:
Sorry could not follow up on review comments, thanks @comius @benjaminp!

PR #16640, opened 2022-11-02
Emit executable target's repository name into repo mapping manifest

fmeum:
Code in the top-level binary target that generates an executable needs to know its canonical repository name in order to look up runfiles with Bzlmod. If this information can not be compiled into the binary, e.g. because it is a shell script and doesn't use a launcher, bootstrapping runfiles discovery requires having this information available at a known location next to the executable.

This commit emits the executable target's canonical repository name as the first line of the repository mapping manifest.

Work towards #16124

### DIFF
--- a/src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
+++ b/src/main/java/com/google/devtools/build/lib/analysis/RepoMappingManifestAction.java
@@ -101,6 +101,12 @@ public class RepoMappingManifestAction extends AbstractFileWriteAction {
       throws InterruptedException, ExecException {
     return out -> {
       PrintWriter writer = new PrintWriter(out, /*autoFlush=*/ false, ISO_8859_1);
+
+      // The executable needs to be able to learn its own repository name. This is especially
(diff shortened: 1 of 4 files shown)


fmeum:
Missed that this means we always have to generate the file.

fmeum:
I got confused by https://github.com/bazelbuild/bazel/pull/16640/files#diff-1b62f0799c4c0e96415e905208d05beb5a53bd73ae77a656449828c459e3f36cL291, but it looks like we are indeed always creating the repo mapping manifest when Bzlmod is enabled now. I updated the comment to reflect that.

Wyverald:
cc @lberki

Wyverald:
This could potentially be fixed by introducing a simple launcher for sh_binaries that doesn't do not much more than setting a few environment variables. @lberki It'd be nice to hear your thoughts.

lberki:
Thoughts dispensed in person. Since `$0` is under the execroot, that can be used to tell both the name of the repository the `sh_binary` is in and the name of the main repository.

PR #16652, opened 2022-11-03
Stage repository mapping manifest as a root symlink

### LINKED ISSUE #16643: Repository mapping manifest does not exist in test sandbox
### Description of the bug:

The `.repo_mapping` file does not appear in test sandboxes.

```
# WORKSPACE
# MODULE.bazel
# pkg/BUILD
sh_test(
  name = "test",
  srcs = ["test.sh"],
)
# pkg/test.sh
#!/usr/bin/env bash

runfiles_dir="$TEST_SRCDIR/../test.runfiles"
if [[ ! -d "$runfiles_dir" ]]; then
  echo "$runfiles_dir does not exist"
  exit 1
fi
repo_mapping="$TEST_SRCDIR/../test.repo_mapping"
if [[ ! -f "$repo_mapping" ]]; then
  echo "$repo_mapping does not exist"
  exit 1
fi
```

`USE_BAZEL_VERSION=last_green bazel test //pkg:test --enable_bzlmod` fails
`USE_BAZEL_VERSION=last_green bazel run //pkg:test --enable_bzlmod` passes
`USE_BAZEL_VERSION=last_green bazel test //pkg:test --enable_bzlmod --spawn_strategy=local` passes

Linux

HEAD

_No response_

_No response_

_No response_

_No response_

fmeum:
By adding the repository mapping manifest to runfiles as a root symlink, it is staged as `foo.runfiles/_repo_mapping` with all execution strategies. This includes sandboxed and remote execution, which previously did not stage the manifest at all.

As a side effect, runfiles libraries can now find the repository mapping manifest via `rlocation("_repo_mapping")`.

Fixes #16643
Work towards #16124


fmeum:
@lberki @Wyverald

src/test/py/bazel/bzlmod/bazel_module_test.py
```diff
@@ -633,40 +635,51 @@ def testRunfilesRepoMappingManifest(self):
...
+    bazel_command = 'build' if self.IsWindows() else 'test'
 
     # Finally we get to build stuff!
-    self.RunBazel([bazel_command, '//:me'], allow_failure=False)
-    with open(self.Path('bazel-bin/me.repo_mapping'), 'r') as f:
-      self.assertEqual(
-          f.read().strip(), """,foo,foo~1.0
+    exit_code, stderr, stdout = self.RunBazel(
+        [bazel_command, '//:me', '--test_output=errors'],
+        allow_failure=True)
```
Wyverald, line 662: why can't we keep using `allow_failure=False`? I thought test failures propagated to the `bazel test` command

fmeum: We can use it, but the error you get in the logs if the test fails will not include the test output since `RunBazel` only prints stderr, not stdout where the test output is printed to.

fmeum: I can revert this part of the change, it's just meant to make the test easier to debug in case of future failures.

Wyverald: i see, it's okay -- was just curious

Wyverald (approved):

fmeum:
@Wyverald Had to fix a check on Windows and added another one for good measure.

lberki:
This looks reasonable:

1. `_repo_mapping` cannot clash with repository names
2. Runfiles manifests are not currently in sandboxes / remote execution so there is no prior art with which we would be inconstent with
3. I just learned that we do create "symlink trees" on Windows by copying, which also works with this approach

I have one suggestion: this creates an extra `Runfiles` object, which may be a memory hog. How about adding an extra field to   `RunfilesSupplier` and tweaking it so that it adds that to the runfiles trees it returns instead of creating that extra `Runfiles` instance?

Wyverald:
to confirm, does this mean we might in some cases have exactly one of `foo.runfiles/_repo_mapping` and `foo.repo_mapping` present? And that's fine because it'll always be found by `rlocation("_repo_mapping")`?

fmeum:
> I have one suggestion: this creates an extra `Runfiles` object, which may be a memory hog. How about adding an extra field to `RunfilesSupplier` and tweaking it so that it adds that to the runfiles trees it returns instead of creating that extra `Runfiles` instance?

I will try to implement it that way as I share this concern.

> to confirm, does this mean we might in some cases have exactly one of foo.runfiles/_repo_mapping and foo.repo_mapping present? And that's fine because it'll always be found by rlocation("_repo_mapping")?

Yes, with `--noenable_runfiles`, only `foo.repo_mapping` will be present, but pointed to by an entry in the runfiles manifest. `rlocation("_repo_mapping")` will always work (the integration test should ensure this).

fmeum:
@lberki There are number of places that directly consume `Runfiles` and don't go through `RunfilesSupplier`, e.g. `SymlinkTreeStrategy`. Do you have a suggestion for how I should deal with them?

fmeum:
@lberki I pushed a commit that wires up the root symlink without allocating a new `Runfiles` in the most straightforward way I could find. It does seem to require passing the manifest into `SourceManifestAction` and `SymlinkTreeAction`, which is not so nice.

fmeum:
@bazel-io flag

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
```diff
@@ -118,28 +118,32 @@ void writeEntry(
   @VisibleForTesting
   SourceManifestAction(
       ManifestWriter manifestWriter, ActionOwner owner, Artifact primaryOutput, Runfiles runfiles) {
-    this(manifestWriter, owner, primaryOutput, runfiles, /*remotableSourceManifestActions=*/ false);
+    this(manifestWriter, owner, primaryOutput, runfiles, /*remotableSourceManifestActions=*/ null,
```
lberki, line 121: nit: this comment now has the wrong parameter name (maybe just remove it?)

lberki (approved):
Yeah, this is kind of ugly, but I don't have a better idea, either (other than the previous "create new `Runfiles` instance" one, which I think is worse on the balance. Maybe add a comment to `Runfiles.getRunfilesInputs()` that explains why the repo mapping manifest can't be there?

Conceptually, the issue is that Runfiles is now two things:
1. What a configured target propagates upwards (and there is never a repo mapping manifest there)
2. What is put in the symlink tree (there may be a repo mapping manifest here)

So maybe the best would be to create two separate classes for these two use cases, but I won't insist.

src/main/java/com/google/devtools/build/lib/analysis/SourceManifestAction.java
fmeum: Removed it.

fmeum:
> Maybe add a comment to `Runfiles.getRunfilesInputs()` that explains why the repo mapping manifest can't be there?

I added a comment.

> So maybe the best would be to create two separate classes for these two use cases, but I won't insist.

Fully agree, but that would be a bit over my head right now.

meteorcloudy:
@bazel-io fork 6.0.0

